PR #1123, opened 2018-04-24
Add option to disable help text on pl_number_input (Close #1118)

### LINKED ISSUE #1118: Is it possible to add an option to remove the "hint" inside the input box?
For the input element, I think it will be beneficial if we can disable the message inside the text box indicating what is the type of tolerance used. I understand this is an important feature, and most people would use it anyways, but I think adding the option would be nice (I have cases in which I would prefer to remove it. BTW, anyone out there planning to implement an "integer" box?

coatless:
This PR address #1118 by including:

- Enabled `hide_help_text` on `pl_number_input`
- Updated `doc/elements.md` with new change
- Added `pl_number_input` demo of options to the PL Element Gallery.

Example: 

screen shot 2018-04-23 at 9 23 03 pm

Reviewer request: @mwest1066 @tbretl

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -8,6 +8,10 @@
   * Fix HTML rendering by reverting `cheerio.js` to `0.22.0` (Matt West).
 
   * Fix Google auth using new API (Matt West).
+  
+  * Add option to toggle placeholder help text for `pl_number_input` (James Balamuta).
+
+  * Add demo question showcasing all options for `pl_number_input` (James Balamuta).
 
 * __3.0.0__ - 2018-05-23
 
@@ -80,10 +84,8 @@
   * Add `sympy.ImmutableMatrix` to list of types accepted by `prairielearn.to_json()` (Tim Bretl).
 
   * Add form help text indicating multiple answer can be selected for `pl_checkbox` (James Balamuta).
-  
-  * Add option to toggle placeholder help text for `pl_number_input` (James Balamuta).
 
-  * Add demo questions showcasing all options for `pl_checkbox` and `pl_number_input` (James Balamuta).
+  * Add demo question showcasing all options for `pl_checkbox` (James Balamuta).
 
   * Add example of how to use PL to learn student names (Tim Bretl).
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -80,8 +80,10 @@
   * Add `sympy.ImmutableMatrix` to list of types accepted by `prairielearn.to_json()` (Tim Bretl).
 
   * Add form help text indicating multiple answer can be selected for `pl_checkbox` (James Balamuta).
+  
+  * Add option to toggle placeholder help text for `pl_number_input` (James Balamuta).
 
-  * Add demo question showcasing all options for `pl_checkbox` (James Balamuta).
+  * Add demo questions showcasing all options for `pl_checkbox` and `pl_number_input` (James Balamuta).
 
   * Add example of how to use PL to learn student names (Tim Bretl).
 
--- a/doc/elements.md
+++ b/doc/elements.md
@@ -79,6 +79,7 @@ Attribute | Type | Default | Description
 `atol` | number | 1e-8 | Absolute tolerance for `comparison="relabs"`.
 `digits` | integer | 2 | number of digits that must be correct for `comparison="sigfig"` or `comparison="decdig"`.
 `allow_complex` | boolean | False | Whether or not to allow complex numbers as answers. If the correct answer `ans` is a complex object, you should use `import prairielearn as pl` and `data['correct_answer'][answers_name] = pl.to_json(ans)`.
+`hide_help_text` | boolean | False | Hide help text stating the default comparison. 
 
 ## `pl_integer_input` element
 
--- a/elements/pl_number_input/pl_number_input.py
+++ b/elements/pl_number_input/pl_number_input.py
@@ -10,7 +10,7 @@ import random
 def prepare(element_html, element_index, data):
     element = lxml.html.fragment_fromstring(element_html)
     required_attribs = ['answers_name']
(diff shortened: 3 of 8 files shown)


coatless:
@tbretl if you have a moment can you give @mwest1066 a thumbs up or down on how I've opted to suppress the hint in `pl_number_input`?

coatless:
@tbretl good point. Let me disable that as well.

coatless:
Okay, with that latest edition, the `pl_number_input` with help text disabled looks like:

screen shot 2018-04-29 at 9 45 29 pm

coatless:
@mwest1066 could this PR on disabling help text be folded in? I'd like to bump the documentation branch.

ChangeLog.md
```diff
@@ -46,8 +46,10 @@
   * Add `sympy.ImmutableMatrix` to list of types accepted by `prairielearn.to_json()` (Tim Bretl).
 
   * Add form help text indicating multiple answer can be selected for `pl_checkbox` (James Balamuta).
+
```
nwalters512: Can you rebase on master and put these under the next version header?

exampleCourse/questions/numberExamples/question.html
```diff
@@ -0,0 +1,69 @@
+<pl_question_panel>
+    <p>
+        Questions below were designed to showcase the different features of 
+		<a href="http://prairielearn.readthedocs.io/en/latest/elements/#pl_number_input-element"><code>pl_number_input</code></a>.
```
nwalters512: switch to spaces instead of tabs for consistency

elements/pl_number_input/pl_number_input.py
```diff
@@ -51,7 +51,9 @@ def render(element_html, element_index, data):
             info = chevron.render(f, info_params).strip()
         with open('pl_number_input.mustache', 'r', encoding='utf-8') as f:
             info_params.pop('format', None)
-            info_params['shortformat'] = True
+            # Within mustache, the shortformat generates the shortinfo that is used as a placeholder inside of the numeric entry.
+            # Here we opt to not generate the value, hence the placeholder is empty.
+            info_params['shortformat'] = not pl.get_boolean_attrib(element, 'hide_help_text', False)
```
nwalters512: Any reason you don't just immediately default to the empty string if help text is disabled? Seems wasteful to go through the template for that. In fact, you could probably skip `# Get comparison parameters and info strings` and down if help text is disabled, since `info` won't be used either.

coatless: Yes, I think that is definitely accurate. However, I'm picking up a weird error on the example problem after rebasing. Any thoughts as to why the test values are `null`?

screen shot 2018-05-30 at 11 13 43 pm
screen shot 2018-05-30 at 11 00 33 pm

ChangeLog.md
```diff
@@ -8,6 +8,10 @@
   * Fix HTML rendering by reverting `cheerio.js` to `0.22.0` (Matt West).
 
   * Fix Google auth using new API (Matt West).
+  
+  * Add option to toggle placeholder help text for `pl_number_input` (James Balamuta).
+
+  * Add demo question showcasing all options for `pl_number_input` (James Balamuta).
```
mwest1066: Needs to be updated with kebab-case rather than snake_case. Sorry!

doc/elements.md
```diff
@@ -79,6 +79,7 @@ Attribute | Type | Default | Description
 `atol` | number | 1e-8 | Absolute tolerance for `comparison="relabs"`.
 `digits` | integer | 2 | number of digits that must be correct for `comparison="sigfig"` or `comparison="decdig"`.
 `allow_complex` | boolean | False | Whether or not to allow complex numbers as answers. If the correct answer `ans` is a complex object, you should use `import prairielearn as pl` and `data['correct_answer'][answers_name] = pl.to_json(ans)`.
+`hide_help_text` | boolean | False | Hide help text stating the default comparison.
```
mwest1066: We are also about to move to kebab-case for attributes. Sorry again!

exampleCourse/questions/numberExamples/info.json
```diff
@@ -0,0 +1,7 @@
+{
+    "uuid": "933849d9-d7ee-4462-b743-b50e43b03b9e",
+    "title": "Examples Showing Real and Imaginary Input via pl_number_input",
```
mwest1066: We are using sentence case (leading capital only) for question titles.

Also, I think it's more accurate to say "real or complex". Maybe we should also drop the "Examples showing" and just say "Real or complex number input via pl-number-input"?

exampleCourse/questions/numberExamples/question.html
```diff
@@ -0,0 +1,69 @@
...
+    <hr />
+</pl_question_panel>
+
+<pl_question_panel>
+    <p>Under this variant, we opt to change the default comparison of <a href="http://prairielearn.readthedocs.io/en/latest/elements/#pl_number_input-element"><code>pl_number_input</code></a>  from relative and absolute tolerances (<code>comparison="relabs"</code>) to significant digits (<code>comparison="sigfigs"</code>), which has a default of two digits. In addition, we've opted to add a suffix or a unit classification to the input via <code>suffix="$units$"</code>. </p>
+    <p> </p>
+    <p> Solve for the value of $c$ given $c^2 = {{params.a}}^2 + {{params.b}}^2$ </p>
+</pl_question_panel>
+
+<pl_number_input answers_name="c" label = "Enter the Value for $c$:" comparison="sigfig" suffix="$units$"/>
```
mwest1066: Units should be in roman font (upright, not italic: see https://physics.nist.gov/cuu/pdf/typefaces.pdf), so this should be `suffix="units"`.

coatless:
Merged in with #1330

Issue #1121, opened 2018-04-23
[Feature Request] Provide Questions from a Template


coatless:
Current status-quo is to copy and paste an existing question variant to create a new one leading to a bunch of UUID issues or naming faults. 

I was thinking adding something along the lines of:

```bash
pl --qtype="pl_element" --name= "Version Control" --copy-to="questions/"
```

Spawn a `pl_element` with `name` and attempt to resolve into the `questions/` directory? Probably need some initial `pwd` check if it matches `pl-courseXXX` setup.

coatless:
Any objection to a NodeJS app?

**package.json**
```js
{
  "name": "pl",
  "version": "1.0.0",
  "description": "Create a PrairieLearn Question from Element Templates",
  "bin" : {
    "pl" : "./pl-question-template.js"
  },
  "license": "AGPL"
}
```

**pl-question-template.js**

```js
#!/usr/bin/env node
const [,, ... args] = process.args

console.log(`Arguments passed: ${args}`)
```

Add CLI options via `Commander.js`

Install would then be done with:

```
npm link
```

This would give:

```bash
pl --question-type="mc" --name="DBProperties"
```

c.f.

- https://docs.npmjs.com/creating-node-js-modules
- https://scotch.io/tutorials/build-an-interactive-command-line-application-with-nodejs
- https://codeburst.io/building-a-node-js-interactive-cli-3cb80ed76c86
- https://www.sitepoint.com/javascript-command-line-interface-cli-node-js/

trombonekenny:
Similar but could be wrapped into this:

I'd wanted something that bootstrapped a whole PL course directory. Something you could point at an empty directory and have it make the minimum required files to start loading. Bonus points for adding sample questions how you described. It would need to be careful not to overwrite things, etc.

I was thinking about running it as a command-line interaction from the PrairieLearn dev-mode console. @nwalters512 had talked about something standalone. Maybe we could make both work? (A mostly external tool that can be run from inside the PL console too). Running these tools from inside PL is helpful because it already would have the needed software installed in the container, doesn't vary by OS (newlines, paths) and has known locations for /course, etc. The `tools/generate_uuids.py` script is useful but it's not easy for new people to be able to run it.

There are other things that would be useful to trigger from inside PL, like pointing to multiple /course directories (rewriting the config.json), running npm update, checking a cloud source to see if the docker version is the latest version or if an update is needed, etc.

coatless:
How about:

```
[...]

PR #1141, opened 2018-04-30
Makefile improvements (close #1124)

### LINKED ISSUE #1124: Improved Makefile
Right now, the `Makefile` for the project is hidden underneath the `doc/` directory.  

I'd like to add a `Makefile` into the main directory that focuses on:

**docs**
1. Build the `doc/` folder using `mkdocs build`
1. Serve the `doc/` folder using `mkdocs serve`
1. Render automatically generated images. 

**pl-dev**
1. Build the PL Development Docker Container
1. Launch a PL Development Docker Container

**pl-dev-tests**
1. Run the linters against the PL Development Docker Container
1. Run the test suite against the PL Development Docker Container

I think naming the recipes along the lines of `build-docker`, `run-docker`, et cetera would be appropriate.

coatless:
This PR has two goals per #1124:

1. improve the functionality of the `doc/Makefile` by adding `mkdocs build` and `mkdocs serve` to render and preview documentation in a local environment prior to the update on prairielearn.readthedocs.org
2. provide initial structure for automatically building, previewing, and testing a local development copy of prairielearn


coatless:
Reviewers: @mwest1066, @nwalters512, @trombonekenny ?

coatless:
Thinking about this more, I'll probably change from `build-docker` to `build-dev`.

I'd also like to add a binding where docker container could launch the course along the lines of `make run-course` with a course name being supplied in a `deploy.env`, e.g.

```bash
COURSE_NAME=pl-stat385
```

coatless:
Thoughts on the revised `Makefile`: @mwest1066, @nwalters512, @trombonekenny ?

Makefile
```diff
@@ -0,0 +1,70 @@
...
+# Generates a development docker from current directory
+build-dev:
+	docker build -t pltestlocal .
+
+# Show the development version of the prairielearn
+dev: build-dev
+	docker run -it --rm -p 3000:3000 pltestlocal
+
+dcourse: build-dev
+	docker run -it --rm -p 3000:3000 -v ${COURSE_REPO_PATH}:/course prairielearn/prairielearn
```
mwest1066: Should this run `pltestlocal` rather than `prairielearn/prairielearn`?

Makefile
```diff
@@ -0,0 +1,70 @@
...
+# Show a live version of the prairielearn
+live:
+	docker run -it --rm -p 3000:3000 prairielearn/prairielearn
+
+lcourse:
+	docker run -it --rm -p 3000:3000 -v ${COURSE_REPO_PATH}:/course prairielearn/prairielearn
+
+# Remove the docker container
+clean:
+	docker stop pltestlocal; docker rm pltestlocal
```
mwest1066: Isn't `pltestlocal` the name of the image, whereas these commands want the name of a container?

Also, maybe we should additionally delete the image with `docker rmi pltestlocal`?

Makefile
```diff
@@ -0,0 +1,70 @@
+#!make
+
+# Include config variables for custom course
+include envfile
+export $(shell sed 's/=.*//' envfile)
```
mwest1066: I don't understand the intention with `envfile`. Are users supposed to edit it? If not, then it seems to not do very much. If they are supposed to edit it then it looks like it's going to interfere with git?

ChangeLog.md
```diff
@@ -5,6 +5,8 @@
 
   * Add student name and clickable e-mail address information to issue reports (James Balamuta).
 
+  * Add `Makefile` to main directory to handle docker builds and previews (James Balamuta).
```
mwest1066: I like this Makefile, but I'm slightly nervous about the fact that it won't work out of the box on Windows. Is this somewhere where we should instead be using more npm scripts? See https://www.keithcirkel.co.uk/how-to-use-npm-as-a-build-tool/

Makefile
coatless: Yes. Good catch!

Makefile
coatless: The idea behind the `envfile` was indeed to have users edit it with a location to the user's course. Thus, we avoid having users type out:

```bash
docker run -it --rm -p 3000:3000 -v ~/Google\ Drive/Courses/pl/pl-stat385:/course prairielearn/prairielearn
```

Perhaps, this can be folded into a "template" directory and then on call to a CLI client the `envfile` can be added to `~/` c.f. #1121

mwest1066: I'm completely on board with this idea, but I don't like files that are tracked by git but which are supposed to be edited. Can we just ship this without `envfile` committed to git, but still have the Makefile import it if it exists (so a conditional import)?

ChangeLog.md
nwalters512: I'm very much on board with working towards eliminating our Makefile - npm scripts were designed for these kind of things.

Makefile
coatless: Instead of merging this PR, let's opt to kill it and move over to the `npm` approach mentioned below. This served as a great starting place in terms of features needed.

ChangeLog.md
mwest1066: Unfortunately I just read more about npm scripts. They kind of suck compared to Makefiles -- why can't people at least retain the same functionality when reinventing the wheel?!? :-)

But I'm still generally in favor of them wherever possible.

coatless: I'll have to look a bit more into the process. But, I agree with this sentiment since interoperability between macOS, Windows, and Linux distros are paramount.

Makefile
mwest1066: I'm on board with that. I think that we might be able to get away with using npm config variables (see "NPM Config Variables" at https://www.keithcirkel.co.uk/how-to-use-npm-as-a-build-tool/) so users could run:

```
npm set prairielearn:course_dir "/Users/mwest/git/pl-tam212" # once only

npm run pl    # uses the config variable from able
```

(or something like that)

PR #1143, opened 2018-05-01
Add optional async render for all requests

nwalters512:
This allows us to prevent long-running page renders from blocking the main thread. It creates a pool of Node workers that will render pages and then pass the results back to the main process to be sent to the client.

### DIFF
--- a/middlewares/injectRenderAsync.js
+++ b/middlewares/injectRenderAsync.js
@@ -1,0 +1,28 @@
+const Pool = require('threads').Pool;
+
+const pool = new Pool(10);
+
+pool.run(([filename, locals], done) => {
+    const ejs = require('ejs');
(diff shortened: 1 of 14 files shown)


middlewares/injectRenderAsync.js
```diff
@@ -0,0 +1,28 @@
+const Pool = require('threads').Pool;
+
+const pool = new Pool(10);
```
nwalters512: I just picked a number that seemed reasonable - this should probably be configurable though?

trombonekenny: Default without specifying is number of CPUs.

nwalters512:
If we ever want to land this in the future, it'll be based on #2394 instead - closing.

PR #1145, opened 2018-05-02
Cluster support

trombonekenny:
Adds nodejs cluster support for one fork per processor.


trombonekenny:
We decided not to go down this route, right? This is a stale PR that can be closed?

nwalters512:
Yep, I think we can close this.

PR #1146, opened 2018-05-05
Improve instructor issues page

nwalters512:
Improves (in my opinion) the display of issues on the instructor issues page by putting the student message on its own line, separate from the QID:

screen shot 2018-05-05 at 3 07 23 pm

The `is:automatically-reported` filter wasn't ever actually being applied; this PR removes that option, since `is:automatically-reported` can be expressed simply with `-is:manually-reported`.


mwest1066:
This looks awesome! While I agree that `is:automatically-reported = -is:manually-reported`, I think that this is the sort of thing that might not be so obvious to a user who wants to find auto-reported issues. Related to this, should we explicitly tag the auto-reported issues? Maybe we should also change the text to say "automatically reported XX ago while username was active" or something like that? The current text makes it sounds like the student really reported it themselves.

nwalters512:
This still includes a filter for automatically reported issues in the pre-built filter drop-down, it's just now expressed as a negation. And the old flag didn't work and wasn't documented, so I don't think we're losing muxh here. Although, we have both is:open and is:closed, so maybe we want two filters for all boolean things.

Those are good suggestions; maybe "reported for X" instead of "reported by X"? I'll fold that into this PR.

nwalters512:
I added a badge for automatically-reported issues and change the text to "reported XX ago **for** YY". Here's how that looks:

screen shot 2018-05-06 at 12 01 27 pm

PR #1163, opened 2018-05-22
[WIP] Update to support running PG locally with docker-compose

nwalters512:
This PR makes a number of changes to allow us to run the actual grading stack alongside PL with docker-compose.

* Add support for RabbitMQ for queueing; we'll run an instance of that with docker-compose
* Add support for using the disk instead of S3 to fetch + report external grading files
* Add an option to disable that Postgres server that runs in the PL container by default
* Switch to using the config model from PrairieGrader, which has been moved into PrairieLib. This allows us to easily override config options with environment variables, which is valuable when configuring things with `docker-compose`. This required refactoring `server.js`, as loading the config is now an async operation.
* Add a `docker-compose.yml` file that can spin up PL, PG, RabbitMQ, and Postgres.

TODO before merge:

* [ ] Completely remove old local grader support
* [ ] Update documentation to include instructions for docker-compose and get rid of references to the old local grader support


lib/external-grading/file-store-disk.js
```diff
@@ -0,0 +1,54 @@
...
+        (callback) => {
+            tmp.dir({ unsafeCleanup: true }, (err, path, cleanupCallback) => {
+                if (ERR(err, callback)) return;
+                dir = path;
+                cleanup = cleanupCallback;
+                callback(null);
+            });
+        },
+        (callback) => {
+            externalGraderCommon.buildDirectory(dir, submission, variant, question, course, callback);
```
mwest1066: Although it's clearly longer and more horrible, I prefer to do the explicit callback here with `ERR(...)` so that we get a proper stacktrace on errors. Hopefully we can make this all go away with async/await soon!

lib/external-grading/file-store-disk.js
```diff
@@ -0,0 +1,54 @@
...
+        },
+        (callback) => {
+            fs.mkdirs(storageDir, (err) => {
+                if (ERR(err, callback)) return;
+                callback(null);
+            });
+        },
+        (callback) => {
+            // Now that we've built up our directory, let's zip it up and send
+            // it off to S3
```
mwest1066: This is the non-S3 version.

lib/external-grading/file-store-s3.js
```diff
@@ -0,0 +1,70 @@
...
+        (callback) => {
+            tmp.dir({ unsafeCleanup: true }, (err, path, cleanupCallback) => {
+                if (ERR(err, callback)) return;
+                dir = path;
+                cleanup = cleanupCallback;
+                callback(null);
+            });
+        },
+        (callback) => {
+            externalGraderCommon.buildDirectory(dir, submission, variant, question, course, callback);
```
mwest1066: Same comment as above about `callback`.

lib/external-grading/file-store-s3.js
```diff
@@ -0,0 +1,70 @@
...
+const tar = require('tar');
+const AWS = require('aws-sdk');
+const tmp = require('tmp');
+const sqldb = require('@prairielearn/prairielib/sql-db');
+const sqlLoader = require('@prairielearn/prairielib/sql-loader');
+
+const config = require('../config');
+const externalGraderCommon = require('./common');
+
+const sql = sqlLoader.loadSqlEquiv(__filename);
```
mwest1066: Is there a `file-store-s3.sql` included with this PR?

lib/external-grading/queue-factory.js
```diff
@@ -0,0 +1,89 @@
...
+                    break;
+                default:
+                    callback(new Error(`Unknown queue type: ${type}`));
+                    return;
+            }
+            callback(null, queue);
+        } catch (e) {
+            callback(e);
+        }
+    })();
```
mwest1066: Haha, I like the async/await!

lib/external-grading/queue-rabbitmq.js
```diff
@@ -0,0 +1,18 @@
...
+
+    async init(conn) {
+        this.ch = await conn.createChannel();
+    }
+
+    submitJob(message, callback) {
+        (async () => {
+            await this.ch.assertQueue(this.name, { durable: true });
+            this.ch.sendToQueue(this.name, new Buffer(JSON.stringify(message)), { persistent: true });
+        })().then(() => callback(null)).catch(callback);
```
mwest1066: Not that I really mind, but I guess I find the try/catch version above a bit easier to follow.

schemas/serverConfig.json
```diff
@@ -46,7 +46,7 @@
         },
         "serverPort": {
             "description": "The TCP port number for the server.",
-            "type": "string"
+            "type": "integer"
```
mwest1066: This is going to break production, so we should be ready for it.

server.js
```diff
@@ -1,24 +1,16 @@
...
-const express = require('express');
-const bodyParser = require('body-parser');
-const cookieParser = require('cookie-parser');
-const passport = require('passport');
 const http = require('http');
 const https = require('https');
 const blocked = require('blocked-at');
-const onFinished = require('on-finished');
-const uuidv4 = require('uuid/v4');
 const argv = require('yargs-parser') (process.argv.slice(2));
```
mwest1066: While you are making changes to all the server stuff, could you also fix it to just use `yargs` properly rather than directly using the low-level `yargs-parser`? This has been bothering me :-) There is an example of `yargs` usage in `tools/load-test.js`.

lib/external-grading/file-store-s3.js
nwalters512: Very good catch! I'll move that over from the appropriate old file.

lib/external-grading/queue-factory.js
nwalters512: It's a little awkward to glue between callbacks and promises, but I want to start using async/await when possible 🙂

mwest1066: I'm one thousand percent in favor of this.

schemas/serverConfig.json
nwalters512: Right. Is this change worth making? I don't think we can really have a non-integer port, but maybe there's a reason this was a string to start with?

mwest1066: I can't remember, it's been this way for so long :-) I believe that `port` must be an integer, but we also don't really rely on it in any way -- we just concat the strings. It might be easier to just leave it as is.

server.js
nwalters512: Sure thing. We currently have
```
// If there is only one argument, legacy it into the config option
```
Can we remove support for passing a config file name without the `--config` option?

mwest1066: Yes, so long as the docker files aren't using it.

nwalters512: `init.sh` seems to be using `--config`, so we should be good.

mwest1066: 👍

doc/running-docker-compose.md
```diff
@@ -0,0 +1,38 @@
...
+
+4. Within the `PrairieLearn` directory you just cloned, run
+
+```sh
+docker-compose up
+```
+
+Note that you may see error messages about connections. It's possible that PrairieLearn or PrairieGrader started up before the RabbitMq or Postgres containers. Wait until you see the error messages stop; this will indicate that everything is now up and running!
+
+You should now be able to access PrairieLearn at [http://localhost:3000/pl](http://localhost:3000/pl).
```
mwest1066: This can actually be just `http://localhost:3000` now.

doc/running-docker-compose.md
```diff
@@ -0,0 +1,38 @@
...
+
+2. Depending on your operating system and the version of Docker you have, you may have to install Docker Compose separately. Visit [https://docs.docker.com/compose/install/](https://docs.docker.com/compose/install/) to determine if this is necessary for you.
+
+3. If you haven't already, clone the [https://github.com/PrairieLearn/PrairieLearn](PrairieLearn repository) to your machine:
+
+```sh
+git clone https://github.com/PrairieLearn/PrairieLearn.git
+```
+
+4. Within the `PrairieLearn` directory you just cloned, run
```
mwest1066: Unfortunately the RTD markdown flavor doesn't correctly handle numbered lists with text in between. I was using `*` rather than `1.` as a workaround.

doc/running-docker.md
```diff
@@ -1,22 +1,21 @@
...
 
-* Step 2: Run PrairieLearn using the example course with:
+2. Run PrairieLearn using the example course with:
 
 ```sh
 docker run -it --rm -p 3000:3000 prairielearn/prairielearn
 ```
 
-* Step 3: Open a web browser and connect to [http://localhost:3000/pl](http://localhost:3000/pl)
+3. Open a web browser and connect to [http://localhost:3000/pl](http://localhost:3000/pl)
```
mwest1066: Same as previous comment. The use of `*` is a workaround.

doc/running-docker.md
```diff
@@ -1,22 +1,21 @@
+# Running the PrairieLearn docker image
 
-# Installing and running PrairieLearn
-
-* Step 1: Install [Docker Community Edition](https://www.docker.com/community-edition). It's free. 
+1. Install [Docker Community Edition](https://www.docker.com/community-edition). It's free.
```
mwest1066: Remove this link -- I've noticed it's confusing for people because we have better links on the sub-bullet points below.

doc/running.md
```diff
@@ -0,0 +1,48 @@
+
+# Running PrairieLearn
+
+Running PrairieLearn on your own machine is the easiest and best way to get started with developing course content. PrairieLearn supports several different ways of running on your own machine; these are outlined below.
+
+## Running the PrairieLearn docker image
```
mwest1066: capital-D Docker for consistency

doc/running-docker.md
```diff
@@ -1,22 +1,21 @@
+# Running the PrairieLearn docker image
```
mwest1066: capital-D Docker for consistency. Maybe better as "Running PrairieLearn with Docker"

Maybe we should repeat the text here about this being the recommended way to run PL?

doc/running-native.md
```diff
@@ -1,27 +1,27 @@
...
+# Running natively
 
 This page describes the procedure to install and run PrairieLearn without any use of Docker. This means that PrairieLearn is running fully natively on the local OS. This installation method is tested and supported on MacOS and Linux, but not on Windows. It should also work on Windows, but it is not tested.
 
-* Install the pre-requisites:
+1. Install the pre-requisites:
 
     * [Node.js](http://nodejs.org/) version  or higher
     * [npm](https://npmjs.org/) (included with Node.js on Windows)
     * [PostgreSQL](https://www.postgresql.org) version 9.6 or higher
```
mwest1066: Now postgres 10

doc/running-native.md
```diff
@@ -1,27 +1,27 @@
...
-# Installing natively
+# Running natively
 
 This page describes the procedure to install and run PrairieLearn without any use of Docker. This means that PrairieLearn is running fully natively on the local OS. This installation method is tested and supported on MacOS and Linux, but not on Windows. It should also work on Windows, but it is not tested.
 
-* Install the pre-requisites:
+1. Install the pre-requisites:
 
     * [Node.js](http://nodejs.org/) version  or higher
     * [npm](https://npmjs.org/) (included with Node.js on Windows)
```
mwest1066: node 10 or higher

doc/running-native.md
```diff
@@ -1,27 +1,27 @@
...
     * [PostgreSQL](https://www.postgresql.org) version 9.6 or higher
     * [Python 3](https://www.python.org) version 3.6 or higher
     * command-line git or [GitHub Desktop](https://desktop.github.com)
 
-On OS X these can be installed with [MacPorts](http://www.macports.org/) or [Homebrew](http://brew.sh/). On Linux these should all be standard packages from the OS distrbution.
+On OS X these can be installed with [MacPorts](http://www.macports.org/) or [Homebrew](http://brew.sh/). On Linux these should all be standard packages from the OS distribution.
 
 Note that with MacPorts you need to select the active version of PostgreSQL, for example `port select postgresql postgresql96`.
```
mwest1066: we can probably delete this now -- anyone reading this should know how to use their package manager

doc/running-docker-compose.md
```diff
@@ -0,0 +1,38 @@
+# Running PrairieLearn + PrairieGrader with `docker-compose`
+
+If you need to develop or test [externally-graded questions](externalGrading.md), you should use this method of running PrairieLearn. This method will spin up PrairieLearn, PrairieGrader, and several other services, and ensure they are connected properly to enable external grading to work.
+
+## Instructions
+
+1. Install [Docker Community Edition](https://www.docker.com/community-edition). It's free.
```
mwest1066: Could probably replace the first few points here with `* Install Docker (see the [PrairieLearn Docker instructions] for details)", to avoid duplication.

doc/running-docker-compose.md
```diff
@@ -0,0 +1,38 @@
...
+Note that you may see error messages about connections. It's possible that PrairieLearn or PrairieGrader started up before the RabbitMq or Postgres containers. Wait until you see the error messages stop; this will indicate that everything is now up and running!
+
+You should now be able to access PrairieLearn at [http://localhost:3000/pl](http://localhost:3000/pl).
+
+## Using your own course content
+
+If you have your own course and you'd like to work on it locally, you can mount the course directory into the PrairieLearn container. To do this, specify the `COURSE_DIR` environment variable when running `docker-compose`:
+
+```sh
+COURSE_DIR=/absolute/path/to/course/ docker-compose up
```
mwest1066: Do you know what the Windows syntax for this is?

doc/running-docker-compose.md
nwalters512: Ahhh, gotcha. That's unfortunate.

doc/running-docker-compose.md
nwalters512: I have absolutely no idea 🙂

doc/running-native.md
nwalters512: You'd be surprised how many people still don't know about HomeBrew 🙂 I'd rather leave this in for now

doc/running-docker-compose.md
coatless: Glancing over this PR, I think the transition over to `docker-compose` takes the need out of running a `makefile`. Thoughts?

doc/index.md
```diff
@@ -1,13 +1,13 @@
 
 # PrairieLearn
 
-PrairieLearn is an online problem-driven learning system for creating homeworks and tests. It allows questions to be written using arbitrary HTML/JavaScript, thus enabling very powerful questions that can randomize and autograde themselves, and can access client- and server-side libraries to handle tasks such as graphical drawing, symbolic algebra, and student code compilation and execution.
+PrairieLearn is an online problem-driven learning system for creating homeworks and tests. It allows questions to be written using arbitrary HTML, JavaScript, and Python, thus enabling very powerful questions that can randomize and autograde themselves, and can access client- and server-side libraries to handle tasks such as graphical drawing, symbolic algebra, and student code compilation and execution.
```
coatless: Add _R_ language?

PR #1165, opened 2018-05-24
Cache question renders

nwalters512:
This enables caching of the panels that are rendered for a question. This is pretty fine-grained; essentially, each call to `renderPanel` is memoized based on the hash of the `data` object. This means we don't have to do any manual cache invalidation (yay - one hard CS problem avoided).

By default, an in-memory LRU cache is used. If `redisUrl` is provided via config, then redis will be used as a cache instead.

We should add some kind of tracking to figure out if this is actually useful - a CloudWatch metric showing cache hit rate would be awesome.


lib/cache.js
```diff
@@ -0,0 +1,64 @@
...
+
+module.exports.init = function(callback) {
+    if (config.redisUrl) {
+        cacheType = 'redis';
+        const client = redis.createClient({
+            url: config.redisUrl,
+        });
+        client.on('error', (err) => logger.error(err));
+        // This max is completely arbitrary
+        cache = redisLru(client, 200);
```
mwest1066: Seems low? Make it a `config` option with a much higher default? 300 students * 20 questions * 5 submissions = 30,000, so maybe 100,000 would be a good limit? At 10k per cached render that's still only 1 GB.

lib/courseUtil.js
```diff
@@ -0,0 +1,73 @@
...
+
+const config = require('./config');
+
+const sql = sqlLoader.loadSqlEquiv(__filename);
+
+function getCommitHash(course, callback) {
+    const gitEnv = process.env;
+    if (config.gitSshCommand != null) {
+        gitEnv.GIT_SSH_COMMAND = config.gitSshCommand;
+    }
```
mwest1066: Not that it hurts, but I think we don't care about GIT_SSH_COMMAND here because this will be a purely local operation.

lib/courseUtil.js
```diff
@@ -0,0 +1,73 @@
...
+    const spawnOptions = {
+        cwd: course.path,
+        env: gitEnv,
+    };
+    const proc = child_process.spawn('git', ['rev-parse', 'HEAD'], spawnOptions);
```
mwest1066: Could use `child_process.exec` to avoid having to do all the event handling ourselves?

migrations/121_pl_courses__commit_hash__add.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE pl_courses ADD COLUMN IF NOT EXISTS commit_hash VARCHAR(40);
```
mwest1066: Using TEXT is probably safer for when they eventually do https://lwn.net/Articles/715716/

(in postgres, VARCHAR is actually implemented as TEXT with a CONSTRAINT, so there is no speed difference)

package.json
```diff
@@ -34,8 +34,9 @@
         "jju": "^1.3.0",
         "jquery": "^3.3.1",
         "json-stringify-safe": "^5.0.1",
-        "lodash": "^4.17.10",
-        "mathjax": "^2.7.4",
+        "lodash": "^4.17.4",
+        "lru-cache": "^4.1.3",
+        "mathjax": "^2.7.2",
```
mwest1066: downgraded lodash and mathjax?

pages/courseSyncs/courseSyncs.js
```diff
@@ -45,11 +46,20 @@ var pullAndUpdate = function(locals, callback) {
...
         // First define the jobs.
+        //
+        // After either cloning or pulling from Git, we'll need to load and
+        // store the current commit hash in the database
+        const updateCommitHash = () => {
+            courseUtil.updateCourseCommitHash(locals.course, (err) => {
+                ERR(err, (e) => logger.error(e));
+                syncStage2();
+            });
+        };
```
mwest1066: I was keeping these blocks in the execution order in the code and numbering them sequentially, so we run 1A->2->3 (or 1B->2->3).

pages/studentInstanceQuestionExam/studentInstanceQuestionExam.js
```diff
@@ -114,14 +114,15 @@ router.post('/', function(req, res, next) {
...
+    logPageView(req, res, (err) => {
         if (ERR(err, next)) return;
-        logPageView(req, res, (err) => {
+        question.getAndRenderVariant(variant_id, res.locals, function(err) {
             if (ERR(err, next)) return;
             res.render(__filename.replace(/\.js$/, '.ejs'), res.locals);
         });
     });
+    const variant_id = null;
+
```
mwest1066: This seems wrong now? We need to pass `variant_id = null` to `getAndRenderVariant()` to tell it to generate a new variant.

Also, we need to do `logPageView()` after `getAndRenderVariant()` because `getAndRenderVariant()` is going to add the new variant to `res.locals` from where `logPageView()` will read the `variant_id` for the page view log. Pretty nasty, I know!

pages/instructorQuestion/instructorQuestion.js
```diff
@@ -154,6 +156,12 @@ router.get('/', function(req, res, next) {
             // res.locals.question_attempts_before_giving_up_histogram_hw = res.locals.result.question_attempts_before_giving_up_histogram_hw;
             callback(null);
         },
+        (callback) => {
+            logPageView(req, res, (err) => {
+                if (ERR(err, next)) return;
+                callback(null);
+            });
+        },
```
mwest1066: See below for why `logPageView()` can't be shifted up to here.

pages/studentInstanceQuestionHomework/studentInstanceQuestionHomework.js
```diff
@@ -103,10 +103,10 @@ router.post('/', function(req, res, next) {
 
 router.get('/', function(req, res, next) {
     if (res.locals.assessment.type !== 'Homework') return next();
-    // req.query.variant_id might be undefined, which will generate a new variant
-    question.getAndRenderVariant(req.query.variant_id, res.locals, function(err) {
+    logPageView(req, res, (err) => {
         if (ERR(err, next)) return;
-        logPageView(req, res, (err) => {
+        // req.query.variant_id might be undefined, which will generate a new variant
+        question.getAndRenderVariant(req.query.variant_id, res.locals, function(err) {
```
mwest1066: As above, we need `logPageView()` to come after `getAndRenderVariant()` so that it can access the newly set `res.locals.variant`.

question-servers/freeform.js
```diff
@@ -624,10 +640,62 @@ module.exports = {
...
+            };
+
+            // This function will check the cache for the specified cache key
+            // and either return the cached render for a cache hit, or render
+            // the panel for a cache miss
+            const getFromCacheOrRender = (cacheKey) => {
+                cache.get(cacheKey, (err, cachedData) => {
+                    // We don't actually want to fail if the cache has an error; we'll
+                    // just render the panel as normal
+                    if (ERR(err, callback)) logger.error(err);
```
mwest1066: I think this should be `if (ERR(err, () => {}) logger.error(err);` so that we won't call `callback`?

question-servers/freeform.js
```diff
@@ -624,10 +640,62 @@ module.exports = {
...
+                // In dev mode, we should skip caching so that we'll immediately
+                // pick up new changes from disk
+                doRender(null);
+            } else {
+                module.exports._getCacheKey(course, data, (err, cacheKey) => {
+                    // If for some reason we failed to get a cache key, don't
+                    // actually fail the request, just skip the cache entirely
+                    // and render as usual
+                    ERR(err, e => logger.error(e));
+                    logger.info(`Using cache key ${cacheKey}`);
```
mwest1066: Maybe a bit to spammy for the logs?

question-servers/freeform.js
```diff
@@ -624,10 +640,62 @@ module.exports = {
...
+                    if (err || !cacheKey) {
+                        doRender(null);
+                    } else {
+                        getFromCacheOrRender(cacheKey);
```
mwest1066: The control-flow here certainly works, but I'd be slightly more comfortable with explicit callbacks so that we can unwind the stack if there is an error. That is, have `doRender(null, (err) => {...})` and call `callback` explicitly from inside there.

question-servers/freeform.js
```diff
@@ -624,10 +640,62 @@ module.exports = {
...
+                    // just render the panel as normal
+                    if (ERR(err, callback)) logger.error(err);
+                    if (!err && cachedData !== null) {
+                        const {
+                            courseIssues,
+                            html,
+                            renderedElementNames,
+                        } = cachedData;
+
+                        callback(null, courseIssues, html, renderedElementNames, true /* cache hit */);
```
mwest1066: I'd prefer the more verbose `const cacheHit = true; callback(..., cacheHit);` to make it easier to search for the `cacheHit` variable. Or at least change the comment to `/* cacheHit */` to match the later variable name.

tests/helperServer.js
```diff
@@ -66,6 +67,12 @@ module.exports = {
                     callback(null);
                 });
             },
+            function(callback) {
+                cache.init(function(err) {
+                    if (ERR(err, callback)) return;
+                    callback(null);
+                });
+            },
```
mwest1066: Do we need a corresponding `cache.finish()` for tests where we start and stop the server many times? It may not matter if we are leaking.

lib/courseUtil.js
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * This should only ever really need to happen at max once per course - in the
+ * future, the commit hash will already be in the course object and will be
+ * updated during course sync.
+ */
+module.exports.getOrUpdateCourseCommitHash = function(course, callback) {
+    if (course.commit_hash) {
+        callback(null, course.commit_hash);
+    } else {
+        module.exports.updateCourseCommitHash(course, (err, hash) => {
```
mwest1066: It may not really matter, but if `course` doesn't have a `commit_hash` then we will insert it into the DB here, but we won't put it into `course`, so I think that we'll end up inserting it many times for the same request (every time we render a panel, I think).

middlewares/logPageView.js
```diff
@@ -8,20 +8,19 @@ var sql = sqlLoader.loadSqlEquiv(__filename);
...
         const user_id = res.locals.user ? res.locals.user.user_id : res.locals.authn_user.user_id;
-        // only log if we are the user who owns the viewed page
-        if (res.locals.instance_user && res.locals.instance_user.user_id != user_id) return;
-        
+
+        // Originally, we opted to only record page views for assessments if
+        // the authn'ed user is also the owner of the assessment instance.
+        // However, we now track all page views, so be sure to filter by
+        // authn_user_id if you only want page views from the student taking
+        // the assessment.
```
mwest1066: I think we'll need such a filter in `instructorAssessmentInstance.sql` in the `BLOCK select_log`?

lib/courseUtil.js
nwalters512: Gotcha - I wasn't completely sure what the purpose of that was, I was just copying from the other places that git operations are done.

lib/cache.js
nwalters512: Should definitely be configurable - I want to hold off on landing #1163 before making config changes.

lib/courseUtil.js
nwalters512: Absolutely - forgot that that existed :)

question-servers/freeform.js
nwalters512: Yep - left over from debugging, forgot to remove it.

lib/cache.js
mwest1066: Ack

lib/courseUtil.js
mwest1066: It's for the containerization code. In current production we are running git as the root user and we've configured the root ssh keys to allow access to github. In the containerized versions we are mounting these keys into the the container and so we need to pass a git command line argument to point at the keys. This avoids hard-coding root ssh keys into our cloudformation scripts.

migrations/121_pl_courses__commit_hash__add.sql
nwalters512: Ah, right, forgot they were doing that with git.

package.json
nwalters512: Don't know why that happened - rebase gone wrong, perhaps? Will fix.

pages/courseSyncs/courseSyncs.js
nwalters512: I did this so that this wouldn't become a separate stage, that feels like unnecessary clutter in the frontend. I can make that change though if you want.

pages/studentInstanceQuestionExam/studentInstanceQuestionExam.js
nwalters512: Mutability has its perks, but man, these global context objects are a mess :)

pages/courseSyncs/courseSyncs.js
mwest1066: ah... I misread the control flow and didn't notice that updateCommitHash() calls syncStage2() without spawning another job to do so. All clear now!

But thinking further, how about moving this inside syncFromDisk.syncDiskToSql()? Seems like it's naturally part of syncing everything from disk into the SQL DB.

middlewares/logPageView.js
nwalters512: It looks like that's already being done: https://github.com/PrairieLearn/PrairieLearn/blob/master/pages/instructorAssessmentInstance/instructorAssessmentInstance.sql#L308 (from master)

mwest1066: Isn't that that JOIN just getting the user corresponding to the authn_user_id in the page_view_log? I don't see where it's actually checking that the authn_user_id is the same as the user who owns the assessment? Maybe something like a WHERE clause on `pvl.authn_user_id = ai.user_id`?

nwalters512: Gah, you're absolutely right :) not sure what I was thinking there.

nwalters512:
Latest few commits added a card for the question render cache to the admin overview page.

screen shot 2018-07-03 at 1 31 42 am

PR #1170, opened 2018-05-28
Add simple html validator

nwalters512:
From the discussion had with @mwest1066, we're going to transition to being more strict when parsing question HTML files. This PR adds a simple validator based on the tokenizer from `htmlparser2` that will error if any elements are self-closing but not specifically marked as void by the HTML spec. This is consistent with how most HTML parsers work, and specifically with the behavior of Web Components: https://github.com/w3c/webcomponents/issues/624

This shouldn't be merged until after we've updated course code and given courses time to update their tooling, as it will deliberately break any questions that use non-compliant HTML.

Example of something that will fail parsing:

```
<pl_submission_panel>
  <pl_file_preview />
  <pl_external_grader_results />
</pl_submission_panel>
```

The corrected version:
```
<pl_submission_panel>
  <pl_file_preview></pl_file_preview>
  <pl_external_grader_results></pl_external_grader_results>
</pl_submission_panel>
```


nwalters512:
Tested with the example course. Modified a question to have a `<pl-number-input ... />` element, which failed with the new renderer and succeeded with the old one.

mwest1066:
@nwalters512 what's the status of this PR? Did we want to try and get this merged, or has the functionality been subsumed by more recent code?

trombonekenny:
It's a little harsh in that it fails to load the question and creates error issues. I was looking for something that was more informative to give an idea of how much work it would be to tweak a course to use the new parser.

I've noodled on this branch to add HTML validation to question sync but only print out an error on the sync page and a running count. Using ECE 470 as an example:

image

The bad news is this course has 136 questions to fix. But we have a harmless way to evaluate it?

mwest1066:
@trombonekenny note that there is https://github.com/nwalters512/pl-question-migration to automatically fix questions -- I should have mentioned this earlier

mwest1066:
The converter above ^^^ that @nwalters512 wrote round-trips question HTML via the renderer to produce valid HTML. I believe it uses the _old_ renderer, so it should be pretty safe. (I should check it to confirm that it's using the old renderer, not the new one.)

I think @nwalters512 wrote the above converter using the _old_ renderer, so it should generally be pretty safe to use (I should confirm that it uses the _old_ one

mwest1066:
We need to come up with a plan to convert everyone to the new renderer. @trombonekenny is this something you could think about how to execute? Some combination of auto-running the converter on everyone's course just after semester, merging this PR (perhaps with more graceful errors), communicating with instructors before and after the transition, etc?

echuber2:
> We need to come up with a plan to convert everyone to the new renderer. @trombonekenny is this something you could think about how to execute? Some combination of auto-running the converter on everyone's course just after semester, merging this PR (perhaps with more graceful errors), communicating with instructors before and after the transition, etc?

I think Nathan pitched an plan to announce and deprecate the old renderer within the next six months (somewhere).

eliotwrobson:
@nwalters512 would this be a good candidate for inclusion with the experimental renderer (https://github.com/PrairieLearn/PrairieLearn/pull/7128) or behind another feature flag? I think something like this would be very useful in combination with that, since it would make it easier to debug unexpected behavior for large courses (which potentially have significant amounts of issues).

eliotwrobson:
@nwalters512 I think even if courses had to manually hit a button to check all HTML in the course, this would be a very useful tool. I've been trying to set up HTML linting for GitHub CI for a while in our course, but kept running into weird issues. So even being able to manually check all HTML would be something very handy. This would be great for local development as well.

nwalters512:
Closing for the time being. We do still eventually want to deprecate/remove the legacy question renderer and provide tooling to help with that; we can revisit the functionality in this PR when we're ready to actively push towards deprecation and removal.

PR #1171, opened 2018-05-28
Change element names to use dashes instead of underscores

nwalters512:
This changes all element names to use dashes instead of underscores in their name - this is more consistent with how Web Components name things, and editor syntax highlighting will finally work for questions 🙏

This change is so far backwards-compatible - I've aliased `***_***_***` to `***-***-***` for all core elements. I'm leaving courses to do what they want with their elements.


elements/pl-checkbox/pl-checkbox.py
```diff
@@ -15,7 +15,7 @@ def prepare(element_html, element_index, data):
     incorrect_answers = []
     index = 0
     for child in element:
-        if child.tag == 'pl_answer':
+        if child.tag in ['pl-answer', 'pl_answer']:
```
mwest1066: I think you are writing python :-) I believe JS needs [includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes), not [in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in) for arrays.

elements/pl-threejs/pl-threejs.py
```diff
@@ -38,8 +38,8 @@ def get_objects(element, data):
     obj_list = []
 
     for child in element:
-        is_stl = (child.tag == 'pl_threejs_stl')
-        is_txt = (child.tag == 'pl_threejs_txt')
+        is_stl = (child.tag in ['pl-threejs-stl', 'pl_threejs_stl'])
+        is_txt = (child.tag == ['pl-threejs-txt', 'pl_threejs_txt'])
```
mwest1066: Both wrong, for different reasons :-)

question-servers/freeform.js
```diff
@@ -699,29 +707,21 @@ module.exports = {
...
-                        let isCourseElement = false;
-                        if (_.has(context.course_elements, elementName)) {
-                            elementDependencies = context.course_elements[elementName].dependencies || {};
-                            isCourseElement = true;
-                        } else {
-                            elementDependencies = coreElementsCache[elementName].dependencies || {};
-                        }
-
-                        elementDependencies = _.cloneDeep(elementDependencies);
+                        let resolvedElement = module.exports.resolveElement(elementName, context);
```
mwest1066: Could throw?

elements/pl-checkbox/pl-checkbox.py
nwalters512: This is python though, right? I'm confused 😂

mwest1066: Hahahaha! I was the most confused of all :-)

elements/pl-threejs/pl-threejs.py
mwest1066: Only one wrong, actually :-)

PR #1178, opened 2018-06-06
Handle missing `assessments` directory on course instance sync

### LINKED ISSUE #999: Assessments required for a course instance?
Check to make sure we're not erroring when a courseInstance exists but doesn't have any assessments defined (or the structure for assessments-- it was unclear from Andre's report).

rbessick5:
Currently an empty `assessments` directory syncs correctly, but a missing `assessments` causes syncs to fail. This explicitly checks for the existence of `assessments` and warns if the directory is missing (skipping the sync).

Fixes #999

### DIFF
--- a/lib/course-db.js
+++ b/lib/course-db.js
@@ -189,23 +189,21 @@ module.exports.loadFullCourse = function(courseDir, logger, callback) {
             var assessmentsDir = path.join(course.courseInfo.courseInstancesDir, courseInstanceDir, 'assessments');
             courseInstance.assessmentDB = {};
             // Check that the assessments folder exists and is accessible before loading from it
-            fs.lstat(assessmentsDir, function(err, fileStats) {
-                // If lstat errors, determine why
+            fs.access(assessmentsDir, function(err) {
                 if (err) {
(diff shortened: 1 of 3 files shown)


lib/course-db.js
```diff
@@ -188,9 +188,28 @@ module.exports.loadFullCourse = function(courseDir, logger, callback) {
...
+            fs.access(assessmentsDir, function(err) {
+                if (err) {
+                    // If the directory does not exist, warn and skip assessment loading
+                    if (err.code == 'ENOENT') {
+                        logger.warn('Warning: ' + courseInstanceDir + ' has no `assessments` directory.');
+                    
+                    }
+                    // Report the error code for any other error and skip assessment loading
+                    else {
+                        logger.warn('Warning: ' + courseInstanceDir + ' `assessment` directory inaccessible (lstat error code: ' + err.code + ').');
```
nwalters512: nit: `assessment` -> `assessments`. Also a good spot for template strings!

mwest1066: Agreed!

rbessick5: I've not used template strings before -- I'll take a look later today.

PR #1184, opened 2018-06-12
Matrix component input

mfsilva22:
This element allows matrices to be entered in a grid format. The shape of the grid will correspond to the shape of the correct_answer. There is an option for partial credit (each entry is graded separately). I added one test question: gradientVectorFunction. 

Update: Deleted the question gradientVectorFunction and instead added the question examplesMatrixInput

I also added one function to display latex matrices inside prairielearn.py (it is used to display the submission and answer panels, and also in another element "latex_display_matrix"


exampleCourse/questions/gradientVectorFunction/info.json
```diff
@@ -0,0 +1,11 @@
+{
+    "uuid": "acff9a90-0145-486a-87b0-65200c4e0253",
+    "title": "Obtain the gradient of a vector",
```
mwest1066: Can we change this question to be named and formatted as an example, like `examplesString`, and put it in the `plelementexamples` homework?

Also add the question to the list for automatic testing in `tests/testQuestions.js`.

elements/pl-matrix-component-input/pl-matrix-component-input.mustache
```diff
@@ -0,0 +1,101 @@
...
+    <a class="btn btn-secondary btn-sm" role="button" data-placement="auto" data-trigger="focus" data-toggle="popover" data-html="true" title="Error" tabindex="0" data-content="{{parse_error}}{{#raw_submitted_answer}}<hr><h5><small class='text-uppercase'>What you submitted</small></h5><pre>{{raw_submitted_answer}}</pre>{{/raw_submitted_answer}}">INVALID
+      <span><i class="fa fa-question-circle" aria-hidden="true"></i></span>
+    </a>
+</span>
+{{/parse_error}}
+{{^parse_error}}
+{{#label}}<span>{{label}}</span>{{/label}}
+<samp>{{a_sub}}</samp>
+{{#correct}}<span class="badge badge-success"><i class="fa fa-check" aria-hidden="true"></i> 100%</span>{{/correct}}
+{{#partial}}<span class="badge badge-warning"><i class="fa fa-circle-o" aria-hidden="true"></i> {{partial}}%</span>{{/partial}}
```
mwest1066: I think `fa-circle-o` doesn't exist? See https://fontawesome.com for the FA icons.

elements/pl-matrix-component-input/pl-matrix-component-input.mustache
```diff
@@ -0,0 +1,101 @@
...
+    $(function(){
+        $('#pl-matrix-component-input-submission-{{uuid}} [data-toggle="popover"]').popover({
+            container: 'body',
+            template: '<div class="popover pl-matrix-component-input-popover" role="tooltip"><div class="arrow"></div><h3 class="popover-header"></h3><div class="popover-body"></div></div>',
+        });
+    });
+</script>
+<span id="pl-matrix-component-input-submission-{{uuid}}">
+    {{#label}}<span>{{label}}</span>{{/label}}
+    <a class="btn btn-secondary btn-sm" role="button" data-placement="auto" data-trigger="focus" data-toggle="popover" data-html="true" title="Error" tabindex="0" data-content="{{parse_error}}{{#raw_submitted_answer}}<hr><h5><small class='text-uppercase'>What you submitted</small></h5><pre>{{raw_submitted_answer}}</pre>{{/raw_submitted_answer}}">INVALID
```
mwest1066: The display of the raw submitted answer didn't work for me when I submitted invalid input.

exampleCourse/infoCourse.json
```diff
@@ -21,6 +21,7 @@
         {"name": "mwest", "color": "gray1", "description": "Question originally written by mwest"},
         {"name": "tbretl", "color": "blue1", "description": "Question originally written by tbretl"},
         {"name": "balamut2", "color": "orange1", "description": "Question originally written by balamut2"},
+        {"name": "mfsilva", "color": "gray1", "description": "Question originally written by mfsilva"},
```
mwest1066: In the `string_element` branch you used `orange1` for your color, which is now causing a conflict. Can you change this one to `orange1`?

elements/pl-matrix-component-input/pl-matrix-component-input.mustache
mfsilva22: I found another bug when I was running a different example, that may fix the problem you found. However, I was not able to get errors when submitting invalid answers in the gradient question. Do you remember which type of invalid answer you tried?

elements/pl-matrix-component-input/pl-matrix-component-input.mustache
mfsilva22: I am confused what do you mean by "doesn't exist". `fa-circle-o` is used in many other elements, such as `pl-number-input`, or `pl-matrix-input`. Am I missing something?

mwest1066: I see the image below, where the icon with an explanation mark indicates that the requested icon was not found. Do you see something different?

2018-06-19 at 17 32

elements/pl-matrix-component-input/pl-matrix-component-input.mustache
mwest1066: Here's an example below. This is tested in the instructor question view, in case that matters.

2018-06-19 at 17 34

elements/pl-matrix-component-input/pl-matrix-component-input.mustache
mfsilva22: I see the same thing, and I always thought it was a weird icon, blinking different than others. But as I said, it appears in other elements, so I thought I should just keep it consistent (and maybe there was a reason for it). What do you suggest? I can just choose another icon, if you want.

elements/pl-matrix-component-input/pl-matrix-component-input.mustache
mfsilva22: I had done it intentionally (can't remember the exact reason :-)) I think I was having difficulties printing the submitted matrix in latex format when one of the entries is left blank (but okay to print when it is a string like the example above). I will take a look at this again.

mwest1066: For printing the raw submitted answer I think it might be better to not use latex at all, and just print out the raw data in a <pre> block or something? I'm wondering about something like:

<pre>
element 1,1: "3"
element 1,2: "5.3"
element 2,1: "" INVALID BLANK ENTRY
element 2,2: "4"
element 3,1: "sdlfkj" INVALID ENTRY
element 3,2: "5,2" INVALID ENTRY
</pre>

The main point here is not to look pretty, but to help a student realize when they've entered an invalid character (such as the comma in 5,2 in the last entry).

Anyway, this is just an idea. Anything else that clearly shows all the entered data and which one is wrong would also be fine.

I think printing out exactly what the student entered is helpful for both the student and a TA or instructor looking at the problem later (although they can also view the raw submission data in the event log).

elements/pl-matrix-component-input/pl-matrix-component-input.mustache
mwest1066: I think maybe it's supposed to be https://fontawesome.com/icons/circle?style=regular ?

I guess the other elements are also wrong in that case. Want to submit a PR to fix them? :-) @tbretl might be able to confirm this as well.

elements/pl-matrix-component-input/pl-matrix-component-input.mustache
mfsilva22: thanks for the input. I will work on it!

elements/pl-matrix-component-input/pl-matrix-component-input.mustache
mfsilva22: okay, I will use the suggested icon for this element. Once @tbretl gives his input, I could also submit a PR for the other elements.

elements/pl-matrix-component-input/pl-matrix-component-input.mustache
mfsilva22: I added the detailed description of all entries as suggested above when there is an invalid answer. If all answers are valid, I decided to keep the display using latex format (I think it is cleaner). Also, I am only giving feedback if each entry is correct or not when `allow-feedback=true`.

elements/pl-matrix-component-input/pl-matrix-component-input.mustache
tbretl: Matt is correct, the upgrade from FA4 to FA5 is why a change is needed:
https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4

elements/pl-matrix-component-input/pl-matrix-component-input.mustache
tbretl: Two more quick things:

1) I suggest putting the invalid raw submitted answer inside the "invalid" popover - this is what's done in both `pl-number-input` and `pl-matrix-input`, for example.

2) I suggest displaying the invalid raw submitted answer (inside the popover) in table format, with the exact same code that's used to create the array of inputs in the question panel - then you wouldn't have to write "element 3, 1" or whatever, wouldn't have to put the submission in quotes, wouldn't have to rely on students knowing that "3, 1" meant "third column and first row," etc. It seems like this would be easier to interpret?

elements/pl-matrix-component-input/pl-matrix-component-input.py
```diff
@@ -0,0 +1,426 @@
...
+                if score >= 1:
+                    html_params['correct'] = True
+                elif score > 0:
+                    html_params['partial'] = math.floor(score * 100)
+                else:
+                    html_params['incorrect'] = True
+            except Exception:
+                raise ValueError('invalid score' + score)
+
+        if allow_feedback:
```
tbretl: When `allow_feedback` is `true`, the code here adds a card to the HTML titled "Feedback" that has a list of the incorrect elements (e.g. `(1, 2) (3, 1)`).

As an alternative, what about displaying the submitted answer in table format (with the same code that was used to generate the array of inputs in the question panel) and putting a red "x" next to each incorrect entry?

This would be similar to multiple choice and checkbox elements, where the "x" comes from this code:
`'&nbsp;<span class="badge badge-danger"><i class="fa fa-times" aria-hidden="true"></i></span>'`

It seems like this would be easier for students to interpret, and wouldn't require them to know that `(1, 2)` (for example) meant "first row and second column."

elements/pl-matrix-component-input/pl-matrix-component-input.py
```diff
@@ -0,0 +1,426 @@
...
+    optional_attribs = ['weight', 'label', 'comparison', 'rtol', 'atol', 'digits', 'format-type', 'allow-partial-credit', 'allow-feedback']
+    pl.check_attribs(element, required_attribs, optional_attribs)
+
+
+def render(element_html, element_index, data):
+    element = lxml.html.fragment_fromstring(element_html)
+    # get the name of the element, in this case, the name of the array
+    name = pl.get_string_attrib(element, 'answers-name')
+    label = pl.get_string_attrib(element, 'label', None)
+    format_type = pl.get_string_attrib(element, 'format-type', 'latex')
```
tbretl: If it is an option to display both the correct answer and the submitted answer in python format, it might be best to display these answers in both python **and** matlab format, as in `pl-matrix-output`, where the student can switch back and forth between these two.

Would it be best to avoid the entire issue and display each answer in a table? (With the same code that is used to create the array of inputs in the question panel.)

question-servers/freeformPythonLib/prairielearn.py
```diff
@@ -810,6 +810,42 @@ def matlab_to_numpy(a):
             return (None, 'Invalid format (missing square brackets and not a real number).')
 
 
+def latex_array_from_numpy_array(A, presentation_type='f', digits=2):
```
tbretl: It will probably be best to get the `pl-matrix-latex` element merged first, so this function can come from there?

PR #1069, opened 2018-03-11
[WIP] Reworking PrairieLearn element docs (close #1050) 

coatless:
This PR is a WIP. 

There are three goals behind this PR.

1. Reorganize the HTML element documentations into two groups: **Input fields** and **Decorative**
2. Include screenshots of the element in action. 
3. Provide _links_ to: 1. questions using element features and 2. source code on GitHub of the question.

PR #1185, opened 2018-06-12
String element

mfsilva22:
This element allows user to enter strings in the input box. There is an option to leave the box empty without raising an error. I added two examples: addBinaryStringElement (where binary numbers are graded as strings), and stringOperations (which shows some of the attributes of the element, such as removal of trailing and leading spaces).


elements/pl-string-input/pl-string-input.mustache
```diff
@@ -0,0 +1,74 @@
...
+{{/answer}}
+
+{{#format}}
+<p>
+Your answer must be a string. Any symbolic expressions or numbers will be intepreted as a string.
+</p>
+{{/format}}
+
+{{#shortformat}}
+string
```
nwalters512: It's probably cleaner just to hardcode this in Python? I know we have the pattern elsewhere, it just seems silly to have to get it out of the template.

exampleCourse/questions/stringOperations/question.html
```diff
@@ -0,0 +1,66 @@
...
+
+
+<pl-question-panel>
+<p> 5) The attribute <code>allow_blank=True</code> allows users to submit blank answers, without
+  receiving the <code>INVALID</code> error. Leave the box below empty, and notice
+that you will not receive the <code>INVALID</code> feedback (and hence the question is graded). </p>
+</pl-question-panel>
+
+<p>
+<pl-string-input answers_name="ans5" label = "blank input box = " allow_blank=True >
```
nwalters512: Remove the spaces on either side of the =; boolean attributes should be like `allow_blank="true"` (this might work now, but will probably fail on a future stricter parser.

elements/pl-string-input/pl-string-input.mustache
mfsilva22: I can go one way or the other. I don't have an opinion about this, I was just following the same idea from other elements. Anyone else would like to comment on this? Should I make the change?

exampleCourse/questions/stringOperations/question.html
mfsilva22: Fixed!

doc/elements.md
```diff
@@ -95,6 +95,24 @@ Attribute | Type | Default | Description
...
+
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`answers_name` | string | — | Variable name to store data in.
+`weight` | integer | 1 | Weight to use when computing a weighted average score over elements.
+`correct_answer` | string | special | Correct answer for grading. Defaults to `data["correct_answers"][answers_name]`.
+`label` | text | — | A prefix to display before the input box (e.g., `label="$x =$"`).
+`suffix` | text | — | A suffix to display after the input box (e.g., `suffix="items"`).
+`display` | "block" or "inline" | "inline" | How to display the input field.
+`remove_leading_trailing` | boolean | False | Whether or not to remove leading and trailing blank spaces from the input string.
```
mwest1066: We are about to change all element attributes to use kebab case (`variables-like-this`) rather than snake case (`variables_like_this`) so it might be a good idea to do this already for these attributes.

elements/pl-string-input/pl-string-input.mustache
mwest1066: I don't mind either way.

elements/pl-string-input/pl-string-input.py
```diff
@@ -0,0 +1,228 @@
...
+
+    if data['panel'] == 'question':
+        editable = data['editable']
+        raw_submitted_answer = data['raw_submitted_answers'].get(name, None)
+
+        # Get info strings
+        info_params = {'format': True}
+        with open('pl-string-input.mustache', 'r', encoding='utf-8') as f:
+            info = chevron.render(f, info_params).strip()
+        with open('pl-string-input.mustache', 'r', encoding='utf-8') as f:
```
mwest1066: It would be a bit more efficient to not read the file twice. I think you can do:
```
with open(...) as f:
    template = f.read()
    info = chevron.render(template, info_params).strip()
    ...
    shortinfo = chevron.render(template, info_params).strip()
```

elements/pl-string-input/pl-string-input.py
```diff
@@ -0,0 +1,228 @@
...
+    allow_blank = pl.get_string_attrib(element, 'allow_blank', False)
+
+    # Get submitted answer or return parse_error if it does not exist
+    a_sub = data['submitted_answers'].get(name, None)
+    if a_sub is None:
+        data['format_errors'][name] = 'No submitted answer.'
+        data['submitted_answers'][name] = None
+        return
+
+    try:
```
mwest1066: This try/except block is more complicated than needed. I think you can just do a simple `if` statement:
```
if not a_sub and not allow_blank:
    data['format_errors'][name] = 'Invalid format. The submitted answer was left blank.'
    data['submitted_answers'][name] = None
else:
    data['submitted_answers'][name] = pl.to_json(a_sub)
```

elements/pl-string-input/pl-string-input.py
```diff
@@ -0,0 +1,228 @@
...
+
+    result = random.choices(['correct', 'incorrect'], [5, 5])[0]
+
+    if result == 'correct':
+        data['raw_submitted_answers'][name] = a_tru
+        data['partial_scores'][name] = {'score': 1, 'weight': weight}
+    elif result == 'incorrect':
+        data['raw_submitted_answers'][name] = a_tru + str((random.randint(1, 11) * random.choice([-1, 1])))
+        data['partial_scores'][name] = {'score': 0, 'weight': weight}
+    # No test for 'invalid' implemented at this time
```
mwest1066: Can you add an invalid test? Returning a blank submission would test invalid, I believe?

doc/elements.md
```diff
@@ -95,6 +95,24 @@ Attribute | Type | Default | Description
...
+`remove_between_spaces` | boolean | False | Whether or not to remove blank spaces between characters from the input string.
```
mwest1066: I think this might be better named `remove_spaces`, because looking at the code it actually removes all spaces, not just spaces between characters?

exampleCourse/questions/stringOperations/question.html
```diff
@@ -0,0 +1,66 @@
...
+  <p> 1) For the following python code snippet, enter below the resulting string:</p>
+
+<pl-code language="python">
+a = "{{params.stringname}}"*{{params.a}}
+</pl-code>
+
+</pl-question-panel>
+
+<p>
+<pl-string-input answers_name="ans1"  label = "a = ">
```
mwest1066: No spaces around the equal sign for attributes, so it should be `label="a = "`

exampleCourse/questions/stringOperations/question.html
```diff
@@ -0,0 +1,66 @@
...
+<pl-question-panel>
+<p> 2) For the following python code snippet, what is d = b + c?</p>
+<pl-code language="python">
+  b = "{{params.b}}"
+  c = "{{params.c}}"
+</pl-code>
+</pl-question-panel>
+
+<p>
+<pl-string-input answers_name="ans2" label = "b + c = " >
```
mwest1066: Spaces around equals again.

exampleCourse/questions/stringOperations/question.html
```diff
@@ -0,0 +1,66 @@
...
+</pl-string-input>
+</p>
+
+<pl-question-panel>
+<p>3) Enter the string below, however include trailing and leading blanks.
+ When using the attribute <code>remove_leading_trailing=True</code>, the answer will
```
mwest1066: Should be `remove_leading_trailing="true"`, I think?

This also occurs in more places below (in both the actual HTML and in the descriptive text).

exampleCourse/questions/stringOperations/info.json
```diff
@@ -0,0 +1,7 @@
+{
+    "uuid": "c07e099f-0149-471e-9c44-ab55622bedec",
+    "title": "String Operations",
```
mwest1066: Can we name this question `String answers with pl-string-input` and add this question to homework4 (which is questions showing how to use different elements)?

exampleCourse/questions/addBinaryStringElement/info.json
```diff
@@ -0,0 +1,7 @@
+{
+    "uuid": "0387c7f5-0c36-447c-9739-e6794d234a6b",
+    "title": "Add binary numbers",
```
mwest1066: I suggest we not bother with this question, because the other question already shows how to use the the element really clearly and I'd rather keep the number of questions under control :-)

doc/elements.md
mfsilva22: so I should modify all attributes, including `answers_name` and `correct_answer`? And I assume you are saying I should change this in the documentation file, and also html and python files.

mwest1066: Yes, sorry! Blame @nwalters512 for making us do things the right way :-)

mfsilva22: Thanks :)

exampleCourse/questions/stringOperations/info.json
coatless: @mfsilva22 also, could you update folder name from `stringOperations` to `examplesStringInput`. This allows for questions configured to show _all_ the possible features of an element.

The HW @mwest1066 is referring to can be found here:

https://github.com/PrairieLearn/PrairieLearn/blob/37b150c83341d66e499ddb8afd05b348323550b9/exampleCourse/courseInstances/Sp15/assessments/plelementexamples/infoAssessment.json#L36

Note: Naming conventions will be updated later this weekend in #1069

mwest1066:
Thanks for all the cleanups! One last change, can you add your example question to the list in `tests/testQuestion.js`? This will make it be automatically tested by Travis.

PR #1186, opened 2018-06-12
Element to display matrices with latex format

mfsilva22:
This element uses a function latex_array_from_numpy_array (added to prairielearn.py) to display matrices in the question panel using latex format. The element allows for groups of matrices. Modified documentation and added two examples: findSingularValues and displayNumpyArraysUsingLatex


exampleCourse/questions/elementsDisplayLatexArrays/info.json
```diff
@@ -0,0 +1,11 @@
+{
+    "uuid": "4d2cf71b-869f-40d3-905a-a67cc855e4be",
+    "title": "Display numpy arrays in the question panel in latex format",
```
mwest1066: Can we change this question to have QID/Title/etc like examplesString?

doc/elements.md
```diff
@@ -169,6 +169,39 @@ A = np.array([[1.23], [4.56]])
...
+    <variable params-name="B"  digits="0"> </variable>
+</pl-latex-matrix-display>
+```
+
+Attributes for `<variable>` (one of these for each variable to display):
+
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`params-name` | string | — | Name of variable in `data['params']` to display.
+`digits` | integer | — | Number of digits to display after the decimal.
```
mwest1066: There is also a `presentation-type` attribute?

doc/elements.md
```diff
@@ -169,6 +169,39 @@ A = np.array([[1.23], [4.56]])
 If a variable `v` is a complex object, you should use `import prairielearn as pl` and `data['params'][params_name] = pl.to_json(v)`.
 
 
+## `pl-latex-matrix-display` element
+
+```html
+<pl-latex-matrix-display>
+    <variable params-name="A"  digits="2"> A = </variable>
+    <variable params-name="B"  digits="0"> </variable>
+</pl-latex-matrix-display>
```
mwest1066: Why do we have this system with the element `pl-latex-matrix-display` containing multiple `variable` elements? It seems to me we could just have:
```
$A = <pl-latex-matrix params-name="A" digits="2"> <pl-latex-matrix params-name="B" digits="0">$
```
or something like this?

I know that `<pl-matrix-output>` uses a syntax like this, but I think that it's needed in that case because all the variables are going to end up displayed in a custom frame. For `<pl-latex-matrix-display>` we aren't doing this, so it just seems like extra complexity?

I'd be interested to hear @tbretl's thoughts about this.

mwest1066: Cool, thanks!

question-servers/freeformPythonLib/prairielearn.py
```diff
@@ -810,6 +810,42 @@ def matlab_to_numpy(a):
             return (None, 'Invalid format (missing square brackets and not a real number).')
 
 
+def latex_array_from_numpy_array(A, presentation_type='f', digits=2):
```
tbretl: Maybe call this `latex_from_2darray` to be consistent with `string_from_2darray`?

question-servers/freeformPythonLib/prairielearn.py
```diff
@@ -810,6 +810,42 @@ def matlab_to_numpy(a):
...
+
+    """latex_array_from_numpy_array
+    This function assumes that A is one of these things:
+            - a number (float or complex)
+            - a 2D ndarray (float or complex)
+    It returns A as a Latex-formatted matrix
```
tbretl: It might be helpful to clarify the meaning of "Latex-formatted matrix" - for example, saying that this function returns a string with the form `\begin{bmatrix} ... & ... \\ ... & ... \end{bmatrix}`.

Maybe also say what the arguments `presentation_type` and `digits` do?

question-servers/freeformPythonLib/prairielearn.py
```diff
@@ -810,6 +810,42 @@ def matlab_to_numpy(a):
...
+    """
+    # if A is a scalar
+    if np.isscalar(A):
+        if presentation_type == 'sigfig':
+            return string_from_number_sigfig(A, digits=digits)
+        else:
+            return '{:.{digits}{presentation_type}}'.format(A, digits=digits, presentation_type=presentation_type)
+    else:
```
tbretl: Maybe make this function consistent with `string_from_2darray` by getting rid of `else:` (and the indent)? The function would have returned already if A was a scalar.

question-servers/freeformPythonLib/prairielearn.py
```diff
@@ -810,6 +810,42 @@ def matlab_to_numpy(a):
...
+                'float_kind': lambda x: to_precision.to_precision(x, digits),
+                'complex_kind': lambda x: _string_from_complex_sigfig(x, digits)
+            }
+        else:
+            formatter = {
+                'float_kind': lambda x: '{:.{digits}{presentation_type}}'.format(x, digits=digits, presentation_type=presentation_type),
+                'complex_kind': lambda x: '{:.{digits}{presentation_type}}'.format(x, digits=digits, presentation_type=presentation_type)
+            }
+
+        if len(A.shape) > 2:
```
tbretl: I'd be tempted to strictly enforce a 2d shape instead of also allowing a 1d shape, similar to `string_from_2darray`. What do you think?

question-servers/freeformPythonLib/prairielearn.py
```diff
@@ -810,6 +810,42 @@ def matlab_to_numpy(a):
...
+            raise ValueError('bmatrix can at most display two dimensions')
+        lines = np.array2string(A, formatter=formatter).replace('[', '').replace(']', '').splitlines()
+        rv = [r'\begin{bmatrix}']
+        rv += ['  ' + ' & '.join(l.split()).replace("'", '') + r'\newline' for l in lines]
```
tbretl: I'm not sure what this does:
`.replace("'", '')`

Is it necessary?

Also, maybe replace `\newline` with `\\`?

exampleCourse/questions/elementsMatrixLatexDisplay/server.py
```diff
@@ -0,0 +1,33 @@
...
+def generate(data):
+
+    sf = 2
+
+    # Matrix shape
+    M = 3
+
+    # Generating the orthogonal matrix U
+    # (numbers rounded with 2 decimal digits)
+    X = np.random.rand(M,M)
```
tbretl: Maybe `M, M` (and `sf + 1`, `M + 3`, `M, 1`, etc., in subsequent lines)? This is the style we often use elsewhere.

exampleCourse/questions/elementsMatrixLatexDisplay/server.py
```diff
@@ -0,0 +1,33 @@
...
+    Q,R = sla.qr(X)
+    U = np.around(Q,sf+1)
+
+    b = np.random.rand(M+3)
+    c = np.random.rand(M,1)
+
+    x = np.array([['x'],['y'],['z']])
+
+    data["params"]["sf"] = sf
```
tbretl: Maybe single quotes?

elements/pl-latex-matrix-display/pl-latex-matrix-display.py
```diff
@@ -0,0 +1,56 @@
...
+    var_data = data['params'].get(var_name, None)
+    if var_data is None:
+        raise Exception('No value in data["params"] for variable %s in pl-latex-matrix-display element' % var_name)
+
+    # If the variable is in a format generated by pl.to_json, convert it
+    # back to a standard type (otherwise, do nothing)
+    var_data = pl.from_json(var_data)
+
+    if np.isscalar(var_data):
+        raise Exception('Value in data["params"] for variable %s in pl-latex-matrix-display element must be an 1D or 2D array' % var_name)
```
tbretl: Isn't it ok for the data to be a scalar? I thought it had to be either a scalar or a 2d array.

elements/pl-latex-matrix-display/pl-latex-matrix-display.py
```diff
@@ -0,0 +1,56 @@
...
+    else:
+        # Wrap the variable in an ndarray (if it's already one, this does nothing)
+        var_data = np.array(var_data)
+        # Check shape of variable
+        if var_data.ndim > 2:
```
tbretl: Maybe check that it is exactly a 2d array? (So, disallow 1d arrays?) This would be consistent with `pl-matrix-output`.

elements/pl-latex-matrix-display/pl-latex-matrix-display.py
```diff
@@ -0,0 +1,56 @@
...
+            raise Exception('Value in data["params"] for variable %s in pl-latex-matrix-display element must be an 1D or 2D array' % var_name)
+
+    # Get the label. Start with blank string if no label is given
+    label = pl.get_string_attrib(element, 'label', '')
+
+    # Create string for latex matrix format
+    latex_data = label + '$' + pl.latex_array_from_numpy_array(var_data, presentation_type=presentation_type, digits=digits) + '$'
```
tbretl: Would it be best if this element returned a string without `$` signs, for example so this string could be combined with other strings and all wrapped in something that may or may not be `$`? I'm confused, in particular, by suggesting the use of the label `$A = $` in the docs, which would result in:

`$A = $ $\begin{bmatrix} ... \end{bmatrix}$`

I would normally write:

`$A = \begin{bmatrix} ... \end{bmatrix}$`

doc/elements.md
```diff
@@ -187,6 +187,32 @@ A = np.array([[1.23], [4.56]])
 If a variable `v` is a complex object, you should use `import prairielearn as pl` and `data['params'][params_name] = pl.to_json(v)`.
 
 
+## `pl-latex-matrix-display` element
```
tbretl: What do you think about the alternative name `pl-matrix-latex`? The idea would be that we've got elements that do things with 2d numpy arrays (or scalars) - input, output, or in this case create a latex string.

doc/elements.md
```diff
@@ -187,6 +187,32 @@ A = np.array([[1.23], [4.56]])
...
+
+The variable in `data['params']` must be a 1D or 2D numpy array. For example, if we want to display the following matrix operations
+```
+x = [A][b] + [c]
+```
+we write
+
+```html
+<pl-latex-matrix-display params-name="A"  digits="1" label="${\bf x} = $"></pl-latex-matrix-display>
+<pl-latex-matrix-display params-name="b"  digits="1"></pl-latex-matrix-display>
```
tbretl: I don't remember - is it possible to use this syntax instead?
`<pl-latex-matrix-display params-name="b"  digits="1" />`

mfsilva22: As far as I know, we should use the syntax that I implemented, where all tags are explicitly closed.

nwalters512: yep - self-closing tags that aren't explicitly self-closing in the HTML spec will fail under an upcoming parser change.

elements/pl-latex-matrix-display/pl-latex-matrix-display.py
mfsilva22: The goal of the element is to return a latex matrix (and not the string, that is done by the function `latex_from_2darray` (new name!)). If I don't wrap the variable `latex_data` with `$` sign, where else would I add `$`?


Someone may not want the label with latex fonts, and that is why I didn't add the label after the `$`. 

Please let me know if you had something specific in mind here. I can't think of a better solution at this time.

tbretl: I suggest removing the `$` signs. The user would then be responsible for wrapping the element in `$` signs for inline math or in `$$` signs for displayed math (https://docs.mathjax.org/en/latest/start.html#tex-and-latex-input).

I understand your concern about the label. Options include (1) removing the label, so requiring the user to prepend the element with a label of their own - this is how I would prefer to use the element; or (2) suggesting the syntax `\text{...}` if the user wants non-math content as the label. I'm sure there are other options.

Maybe another way to put this is that I would often use the element without the `$` signs, but would rarely use the element with the `$` signs.

What do you think?

PR #1187, opened 2018-06-12
Checkbox partial credit

mfsilva22:
Updated checkbox element to allow from partial credit (done before by Pearl), and remove the feedback when partial credit is used (only allow feedback to show if question is disabled).

Update: implemented another grading method for partial credit. The two methods are:

1) "Every Decision Counts" (EDC), following edX (https://edx.readthedocs.io/projects/edx-partner-course-staff/en/latest/exercises_tools/checkbox.html#example-checkbox-problem) 
2) "Percent Correct" (PC) - used by Canvas, but they don't give a name. I am not sure what to call it, so open to suggestions. (https://community.canvaslms.com/docs/DOC-6674-understanding-multiple-answers-questions)

Documentation has a brief explanation for each method.

When EDC is used, students need to submit at least one correct answer (to avoid getting points for no submission), otherwise they get invalid error. For the other grading options, "no submission" is allowed.


exampleCourse/infoCourse.json
```diff
@@ -20,11 +20,13 @@
...
         {"name": "tbretl", "color": "blue1", "description": "Question originally written by tbretl"},
+        {"name": "mfsilva", "color": "blue1", "description": "Question originally written by mfsilva"},
         {"name": "balamut2", "color": "orange1", "description": "Question originally written by balamut2"},
         {"name": "tpl101", "color": "gray1", "description": "Question originally written for the XC 101 example course"},
         {"name": "fa17", "color": "gray1", "description": "Question written in Fall 2017"},
         {"name": "sp18", "color": "gray1", "description": "Question written in Spring 2018"},
         {"name": "v2", "color": "red1", "description": "v2 generation question"},
-        {"name": "v3", "color": "green1", "description": "v3 generation question (freeform)"}
+        {"name": "v3", "color": "green1", "description": "v3 generation question (freeform)"},
+        {"name": "pearl", "color": "yellow3", "description": "Question originally written by pearl"}
```
tbretl: Maybe move this up to follow the other named users, instead of following v2/v3?

trombonekenny: Are you suggesting we shouldn't partially grade an element that has nothing submitted? Do we have scenarios where 0 correct answers are included (i.e. they must check something)?

trombonekenny:
Poke that STAT 448 is interested in this functionality. Do we think this will be merged by start of fall semester?

exampleCourse/infoCourse.json
mfsilva22: Do we want to mark as "invalid" when the submitted answer is left blank (and so it is not graded)? In that case, instructors will always have to make at least one of the answers correct, which I think is reasonable. 

Should I go ahead and make this implementation? Or any other suggestions?

coatless: @mfsilva22 I think we do want to mark the answer as invalid. I say this because the logic was never implemented to check for empty submissions, c.f.

https://github.com/PrairieLearn/PrairieLearn/blob/12ca790b3c93adadb0cdfdc43962d96ca54e2b37/elements/pl-checkbox/pl-checkbox.py#L216-L218

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -178,11 +190,12 @@ def render(element_html, element_index, data):
                     if score >= 1:
                         html = html + '&nbsp;<span class="badge badge-success"><i class="fa fa-check" aria-hidden="true"></i> 100%</span>'
                     elif score > 0:
-                        html = html + '&nbsp;<span class="badge badge-warning"><i class="fa fa-circle-o" aria-hidden="true"></i> {:d}%</span>'.format(math.floor(score * 100))
+                        html = html + '&nbsp;<span class="badge badge-warning"><i class="far fa-circle" aria-hidden="true"></i> {:d}%</span>'.format(math.floor(score * 100))
```
coatless: Shouldn't this be `fa` instead of `far`?

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -207,22 +220,37 @@ def render(element_html, element_index, data):
 
 
 def parse(element_html, element_index, data):
-    # FIXME: check for invalid answers
-    pass
+
+    element = lxml.html.fragment_fromstring(element_html)
```
coatless: Woohoo! No more empty checkboxes.

tbretl:
Thanks for the updates to this element. It looks great. I have a few quick questions.

(1) Why does EDC - but not PC - require at least one checked box, both in the correct answer (error when nothing is "true") and in the submitted answer (invalid when nothing is "true")? If anything, it seems like PC should be the partial credit type that requires at least one checked box.

(2) We had talked about the merits of providing feedback - for example, of telling students some or all of the boxes they incorrectly did or didn't check, particularly in the case of PC. Had you decided that providing feedback would be a bad idea?

(3) Given that there are now three different grading schemes available (no partial credit, EDC, PC) and that different instructors may make different choices, it may be helpful to have a popover or something that explains how the student's answer will be graded, and also explains minimum requirements for an answer, when applicable (e.g., that at least one box must be checked). What do you think?

(4) Do you think it might be helpful to combine your new example question with the other example checkbox question, the purpose of which was to show all the various options? (I'm not sure.)

(5) Do you think it might be better to make PC the default grading scheme? (I'm not sure.)

You may want to rebase onto the current master before proceeding.

Sorry for all the questions! I think we can do the merge quickly after they get resolved.

mfsilva22:
Response to @tbretl 

1) From what I understood from previous conversations, we want to avoid students receiving partial credit points for just submitting a "blank" answer (which is what would happen with the EDC option. I don't think guessing a "blank" answer will give partial credit points when using the option PC.

2) I remember the discussion about feedback, but at the end of that meeting, before @mwest1066 left the room, he asked me to implement the other grading scheme at this time (which I called PC), but not worry about feedback. 

3) I like the idea of the popover. I will think about it and respond soon.

4) I could combine the examples. I assume you mean I should add a couple more items to the existing question illustrating both partial credit options, correct? 

5) I agree that PC should be default (I will particularly use that grading scheme more often)

tbretl:
With respect to submitting blank answers... it seems less confusing if *all* checkbox questions (regardless of grading type) have at least one "true" box, so that a blank answer is always an invalid answer. This avoids students "hitting grade by accident," which is more of a concern to me than students gaming the partial credit system. In any case, with EDC, students could check one box at random and still get just about as much partial credit (on average) as they would have received by submitting a blank answer.

mfsilva22:
okay, I will go back to previous implementation where blank answers (not clicking at least one option) always get marked as invalid, regardless of grading method.

Going back to popover, it seems that I should add a mustache file for that, correct? 

Also, I am not sure where to add the button. Should we just add it below the options, and even add the "hide-text" inside as well. I think it looks a little funny, but I am not sure where else it could go:

image

tbretl:
This looks really good. I like it. I suggest two more small changes.

First, I would slightly revise the popover text.

I suggest that you always make the header say “Checkbox” (instead of “Grading algorithm”). This is not necessarily the best choice, but is consistent with other elements.

Then, I suggest something like this text:

For no partial credit…
```
You must select {{at least one option | between X and Y options}}.
You will receive a score of 100% if you select all options that are true
and no options that are false. Otherwise, you will receive a score of 0%.
```

For EDC…
```
You must select {{at least one option | between X and Y options}}.
You will receive a score of <code>100% * (t + f) / {{number of options}}</code>,
where <code>t</code> is the number of true options that you select and
<code>f</code> is the number of false options that you do not select.
```

For PC…
```
You must select {{at least one option | between X and Y options}}.
You will receive a score of <code>100% * (t - f) / n</code>, where <code>t</code
is the number of true options that you select, <code>f</code> is the number of
false options that you select, and <code>n</code> is the total number of true
options. At minimum, you will receive a score of 0%.
```

Here, the popover would show "at least one option" or "between X and Y options" (with X and Y replaced with appropriate numbers) depending on whether or not `detailed-help-text="true"`. Also, `{{number of options}}` would be replaced with the total number of options (i.e., of checkboxes).

Second, when `detailed-help-text="true"`, I would add a check that the number of options selected is in range, and declare the answer invalid if the number is out of range (just as you already do when the number of options selected is zero).

Does that make sense?

I think everybody is going to be really happy with the new version of checkbox. This is great.

PR #1188, opened 2018-06-14
[WIP] Rewrite question renderer

nwalters512:
This changes how panels are rendered. Instead of using `cheerio` to do a fake traversal, we do a proper breadth-first traversal that allows us to cleanly handle nested elements and is also more efficient (elements that will never be shown in a certain panel will never be rendered).

This is unfortunately a breaking change for any question that has malformed HTML, as this new parser is significantly less lenient than the old one. We should auto-fix all questions (which is backwards-compatible) and ensure that everyone has those changes merged before merging + deploying this one.


nwalters512:
Closing in favor of #1209

PR #1193, opened 2018-06-19
Use hyphens for all element attribute names

nwalters512:
Going along with our theme of hyphenating element things, this changes all element attributes to use hyphenated names. It keeps backwards compatibility by checking for the attribute `/s/-/_` if the hyphenated version isn't found.

Before:

```html
<my-element answers_name="asdf" correct_answer="fdas"></my-element>
```

After:
```html
<my-element answers-name="asdf" correct-answer="fdas"></my-element>
```


question-servers/freeformPythonLib/prairielearn.py
```diff
@@ -128,11 +128,33 @@ def inner_html(element):
...
+    if attrib in object:
+        return True
+    new_attrib = attrib.replace('-', '_')
+    return new_attrib in object
+
+
+def compat_get(object, attrib, default):
+    if attrib in object:
+        return object[attrib]
+    new_attrib = attrib.replace('-', '_')
```
mwest1066: I think the overall logic here is correct, but shouldn't this be called `old_attrib` rather than `new_attrib`? As in, this variable is going to be in the "old" (i.e., with underscores) format, right? Same comment applies to `compat_has()` above.

question-servers/freeformPythonLib/prairielearn.py
```diff
@@ -128,11 +128,33 @@ def inner_html(element):
...
+    for i in arr:
+        new_arr.append(i)
+        new_arr.append(i.replace('-', '_'))
+    return new_arr
+
+
 def check_attribs(element, required_attribs, optional_attribs):
     for name in required_attribs:
-        if name not in element.attrib:
+        if not compat_has(element.attrib, name):
```
mwest1066: Why do we call `compat_has()` here rather than the existing `has_attrib()`? It looks to me like they do the same thing, but I'm probably missing something.

question-servers/freeformPythonLib/prairielearn.py
nwalters512: Good call, I think I had the logic reversed at first. Will fix.

question-servers/freeformPythonLib/prairielearn.py
nwalters512: Nope, good catch! Fixed.

PR #1196, opened 2018-06-22
Fixes "Finish exam" button showing up in all exams

trombonekenny:
Should have been only SEB and password protected exams.
Adds an example course exam with a password.

### DIFF
--- a/pages/studentAssessmentInstanceExam/studentAssessmentInstanceExam.ejs
+++ b/pages/studentAssessmentInstanceExam/studentAssessmentInstanceExam.ejs
@@ -152,7 +152,7 @@
         Submit your answer to each question with the <strong>Grade</strong> button on the question page.
         Look at <strong>Best submission</strong> to confirm that each question has been graded.
 
-        <% if (authz_result.mode == 'SEB' || 'password' in authz_result) { %>
+        <% if (authz_result.mode == 'SEB' || authz_result.password !== null) { %>
         <hr>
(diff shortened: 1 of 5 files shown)


pages/studentAssessmentInstanceExam/studentAssessmentInstanceExam.ejs
```diff
@@ -152,7 +152,7 @@
         Submit your answer to each question with the <strong>Grade</strong> button on the question page.
         Look at <strong>Best submission</strong> to confirm that each question has been graded.
 
-        <% if (authz_result.mode == 'SEB' || 'password' in authz_result) { %>
+        <% if (authz_result.mode == 'SEB' || authz_result.password !== null) { %>
```
mwest1066: I suggest making this `authz_password != null` (note one equals not two). This will catch both `null` and `undefined`, just for added safety in case we change things in the future.

trombonekenny: Just a habit I picked up somewhere of always using === with null. I'll change it here.

mwest1066: Yeah, JS is kind of horrible here. Using triple-equals with null checks for exactly null, whereas double-equals null checks for either null or undefined, which is often safer. I used to do triple-equals but then switch to double.

PR #1199, opened 2018-06-30
Redesign login screen

nwalters512:
This redesigns the login page to be prettier, while remaining just as functional. Run locally and go to http://localhost:3000/pl/login to test.


coatless:
Great redesign of the login page! 

Though, I'm a bit nervous about adding any kind of _hide_ animation to a key portal. Glancing at the underlying animation, there is a cost of ~800ms prior to usable functionality. Not sure this would be ideal in a testing environment.

nwalters512:
I'm trying to be more conscious of UX in my work - see Issues, which deviate pretty heavily from the standard table design used across the site, but overall results in a better experience.

@trombonekenny I'll add that explanatory text back in. We should think about how these Illinois-specific buttons + text will scale beyond UIUC, but that's a conversation for another time.

@coatless fair point. My thoughts are that there's research that shows animations like those can inspire trust and delight users, which may not be a bad way to start an exam. I can play around with the timing a bit - 800ms is probably too long.

mwest1066:
I agree that I'm nervous about having any animations or JS running on such a basic page. Not to argue against "delight" for users 😄 but I'd rather just have the page render?

nwalters512:
😢 never let it be said that I didn't try to make things prettier!

nwalters512:
This should be ready for re-review.

nwalters512:
Tested on `pl-dev`, things look good.

PR #1200, opened 2018-07-06
Add (incomplete) PrairieDraw documentation

rbessick5:
This is an incomplete list of drawing functions available to PrairieDraw users, mainly written for the TAM 2XX summer development staff.

### DIFF
--- a/doc/PrairieDraw.md
+++ b/doc/PrairieDraw.md
@@ -1,6 +1,175 @@
 
 # PrairieDraw
 
+The PrairieDraw library (`PrairieDraw.js`) is a figure-drawing library which allows dynamic figures to be rendered through the `pl_prairiedraw_figure` custom element.
+
+## PrairieDraw figure element configuration
+
+The `pl_prairiedraw_figure` element is used to place figures within `question.html`. To make this element available for your course, a copy of the `pl_prairiedraw_figure` directory must be placed in the `elements` directory for your course.
+
+## `pl_prairiedraw_figure` element
+
+```html
+<pl_prairiedraw_figure script_name="drawFigure.js" params_names="r1,r2,isHorizontal" width="900" height="400" />
+```
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`script_name` | string | - | Name of PrairieDraw script.
+`params_names` | string | `None` | Comma-separated list of parameters to make available to PrairieDraw.
+`width` | integer | 500 | Width of the drawing element.
+`height` | integer | 500 | Height of the drawing element.
+
+Parameter names are those stored in `data["params"]["m"]` in `server.py` and accessed as `{{params.m}}` in `question.html`.
+
+## `Sylvester.js` library and selected methods
+
+PrairieDraw uses the `Sylvester.Vector` class to specify positions and directions for drawing; any Vector-type arguments listed below are of this type.
+
+Below is an incomplete selection of useful operations on Vector; the complete documentation is found [here](http://sylvester.jcoglan.com/docs.html).
+
+`Sylvester.Vector.create(elements)`
+Creates and returns a new Vector from the array `elements`. It is often useful to alias this command in your script as `var $V = Sylvester.Vector.create` so you can write `$V([x,y])` later on.
+
+`add(vector)`
+Returns a new Vector which is the sum of the receiver and the argument.
+
+`subtract(vector)`
+Returns a new Vector which is the difference between the receiver and argument; i.e., `a.subtract(b)` returns the vector difference `a-b`.
+
+`multiply(scalar)` or `x(scalar)`
+Returns a new Vector which is the receiver scaled by the argument. `a.multiply(k)` and `a.x(k)` are equivalent.
+
+`dot(vector)`
+Returns the scalar product of the receiver and argument.
+
+`cross(vector)`
+Returns a new Vector which is the cross product in left-to-right order; i.e., `A.cross(B)` returns `A x B`.
+
+`modulus()`
+Returns the (2)-norm of the vector.
+
+`toUnitVector()`
+Returns a new Vector which is the receiver scaled to have modulus one.
+
+`rotate(angle, axis)`
+Returns a new Vector formed by rotating the receiver counterclockwise through `angle` (in radians), using the Vector `axis` as a center point. E.g.,
+```javascript
+a = $V([10,5]);
+b = $V([5,5]);
+c = a.rotate(Math.PI / 2, b) // c = [5, 10]
+```
+
+## `PrairieDraw.js` library
+
+The script provided by `script_name` may access the PrairieDraw figure via `this.pd`. The operations below allow for changing the orientation or scaling of the figure, as well as adding drawing elements to it.
+
+In general, drawing objects are opaque, so the order in which items are added affects the final appearance of the figure.
+
+Below is an incomplete list of operations available from PrairieDraw; the library source code can be found [here](https://github.com/PrairieLearn/PrairieLearn/blob/master/public/javascripts/PrairieDraw.js).
+
+## Constants provided by PrairieDraw
+
+`goldenRatio`
+Returns the golden ratio; useful for creating pleasantly-proportioned figures in the arguments to `setUnits`.
+
+`milesPerKilometer`
+Returns a conversion factor of 0.621371
+
+## Setting and changing the coordinate system
+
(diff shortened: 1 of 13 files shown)


doc/PrairieDraw.md
```diff
@@ -1,6 +1,175 @@
...
+The PrairieDraw library (`PrairieDraw.js`) is a figure-drawing library which allows dynamic figures to be rendered through the `pl_prairiedraw_figure` custom element.
+
+## PrairieDraw figure element configuration
+
+The `pl_prairiedraw_figure` element is used to place figures within `question.html`. To make this element available for your course, a copy of the `pl_prairiedraw_figure` directory must be placed in the `elements` directory for your course.
+
+## `pl_prairiedraw_figure` element
+
+```html
+<pl_prairiedraw_figure script_name="drawFigure.js" params_names="r1,r2,isHorizontal" width="900" height="400" />
```
coatless: In the element naming and attribute specification, the PL group has switched from snake-case in favor of kebab-case. (PR's https://github.com/PrairieLearn/PrairieLearn/pull/1171 and https://github.com/PrairieLearn/PrairieLearn/pull/1193) 

e.g.

`pl_prairiedraw_figure` => `pl-prairiedraw-figure`

`script_name` => `script-name`

nwalters512:
We should only be documenting features that are offered by core PL. If you feel that the PrairieDraw element is functioning and stable, feel free to put up a PR for that first.

rbessick5:
As long at `PrairieDraw.js` is still considered part of core PL, then this element should probably also be (with the appropriate underscore-to-hyphen transition).

@nwalters512 , would you prefer a separate PR for incorporating `pl-prairiedraw-figure` as an element, or should I withdraw this PR and include the documentation update and element in a single PR?

nwalters512:
@rbessick5 feel free to just push element code as part of this PR, we've been doing docs and elements together in the same PRs.

rbessick5:
Ok, will do.

Since this is a custom element in a few courses, will there be any issues with copies existing both in `/PrairieLearn/elements/` and `/course/elements/` post-merge? I guess I don't know what the PATH looks like for resolving PL elements.

nwalters512:
PL will always use the course version of an element if it exists, so there shouldn't be any issues with that.

rbessick5:
Ok, cool. I'll have to spread the word to deprecate the course version once we merge (so that any future updates PL-side don't get masked by the course version).

rbessick5:
OK, this should be ready for final review and merge.

nwalters512:
I'll try to give this a closer look later tonight. Can you add an example question for this element and add it to the example homework?

doc/elements.md
```diff
@@ -216,6 +216,23 @@ def file(data):
 
 If `file()` does not return anything, it will be treated as if `file()` returned the empty string.
 
+## `pl-prairiedraw-figure` element
+
+```html
+<pl_prairiedraw_figure script_name="drawFigure.js" params_names="r1,r2,isHorizontal" width="900" height="400" />
```
nwalters512: Update to use dashed names for everything

doc/elements.md
```diff
@@ -216,6 +216,23 @@ def file(data):
...
+```
+
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`script-name` | string | - | Name of PrairieDraw script.
+`params-names` | string | `None` | Comma-separated list of parameters to make available to PrairieDraw.
+`width` | integer | 500 | Width of the drawing element.
+`height` | integer | 500 | Height of the drawing element.
+
+The provided `script-name` corresponds to a file located within the director for the question. Parameter names are those stored in `data["params"]["m"]` in `server.py` and accessed as `{{params.m}}` in `question.html`.
```
nwalters512: I read this as saying that it's hard-coded to use the `m` key; can you clarify the behavior?

doc/elements.md
```diff
@@ -216,6 +216,23 @@ def file(data):
...
+## `pl-prairiedraw-figure` element
+
+```html
+<pl_prairiedraw_figure script_name="drawFigure.js" params_names="r1,r2,isHorizontal" width="900" height="400" />
+```
+
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`script-name` | string | - | Name of PrairieDraw script.
+`params-names` | string | `None` | Comma-separated list of parameters to make available to PrairieDraw.
```
nwalters512: Nit: `param-names` seems to me to be a better name for this and is more consistent with attributes like `file-names` in `pl-file-upload`.

elements/pl-prairiedraw-figure/info.json
```diff
@@ -0,0 +1,14 @@
+{
+    "controller": "pl-prairiedraw-figure.py",
+    "dependencies": {
+        "coreScripts": [
```
nwalters512: Are all of these scripts required for PrairieDraw?

elements/pl-prairiedraw-figure/pl-prairiedraw-figure.py
```diff
@@ -0,0 +1,45 @@
+import prairielearn as pl
+import lxml.html
+import chevron
+import os
+
+
+def prepare(element_html, element_index, data):
+    element = lxml.html.fragment_fromstring(element_html)
+    required_attribs = ['script_name']
```
nwalters512: Attributes should use `kebab-case`

elements/pl-prairiedraw-figure/pl-prairiedraw-figure.py
```diff
@@ -0,0 +1,45 @@
...
+    optional_attribs = ['params_names', 'width', 'height']
+    pl.check_attribs(element, required_attribs, optional_attribs)
+    return data
+
+
+def render(element_html, element_index, data):
+    element = lxml.html.fragment_fromstring(element_html)
+    script_name = pl.get_string_attrib(element, 'script-name', None)
+    if script_name is None:
```
nwalters512: This shouldn't be necessary, we'd never get to the `render` phase if this was missing since it would fail in the `prepare` check.

doc/elements.md
rbessick5: I'll change this to read "parameter names are keys stored in `data["params"]` in `server.py` (i.e., those available for templating within `question.html`)

doc/elements.md
rbessick5: Changed to dashed names.

doc/elements.md
rbessick5: I agree with your reasoning on the change; my concern is backwards compatibility. The TAM 2XX classes have used this element (course-level) for several semesters, so changing this attribute name will break a large collection of existing problems.

Does it make sense to tweak the element code to search for `param-names` and then `params-names` so older questions work without changes, but let the documentation reflect the new name?

elements/pl-prairiedraw-figure/pl-prairiedraw-figure.py
rbessick5: Changed to kebab case as intended

elements/pl-prairiedraw-figure/pl-prairiedraw-figure.py
rbessick5: I'll remove the redundant check from `render`.

doc/elements.md
nwalters512: You could do that, but I would also take this as a good opportunity to make breaking changes. This is effectively a brand-new element; courses that already use it have to explicitly opt-in to this version by deleting the copy from their course. A part of that process could be updating the usage; thankfully, this change is probably trivial to do with find/replace.

elements/pl-prairiedraw-figure/info.json
rbessick5: It seems so:
- `PrairieDraw.js` directly uses
-- `PrairieGeom.js` for some computations
-- `sylvester.js` for representing positions as Vectors
-- `sha1.js` for the TeX-to-image rendering
- `PrairieGeom.js` uses `numeric-1.2.6.min.js` for some calculations
- Since some PD figures are input-responsive (students draw lines on the figures), `underscore.js` and `backbone.js` are needed for responsive figures.

doc/elements.md
rbessick5: That seems like a reasonable solution; since everyone using this currently has a course-level version of the element, nothing existing should break (I hope).

nwalters512: You can verify it with your local course + PL if you have concerns about it breaking :)

PR #1209, opened 2018-07-31
New question renderer with feature flag

nwalters512:
I told @mfsilva22 I'd have this ready ASAP, but my parents will be in town for the next week, meaning that most of my free time will be family time 🙂this is a continuation of my work from #1188 - it's basically the same code, but leaving the legacy question parsing + rendering in place. The thought is to have the new renderer behind a feature flag so that courses who know what they're doing can opt in while we get our act together re: question migrations.

The remaining thing to be done is to actually add the feature flag. It looks like the `pl_courses` table has an unused column `options` that isn't ever populated during sync, which could probably be used for this purpose. For an example (abbreviated) `infoCourse.json`:

```json
{
    "name": "CS 225",
    "options": {
        "useNewQuestionRenderer": true
    }
}
```

This would then be available as `context.course.options.?useNewQuestionRenderer` inside the question server functions. I don't have time to wrap this up now, but it should be pretty quick to get this PR to completion. To test, I've been putting a `<pl-figure>` inside a `<pl-multiple-choice>` answer. It won't render with the old renderer, but it will with the new.

This probably also requires testing support, as we'll want to ensure both renderers keep working for now.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -14,6 +14,7 @@
   * Add `pl-prairiedraw-figure` element and update PrairieDraw graphics documentation. (Ray Essick).
 
   * Add `pl-matrix-component-input` element (Mariana Silva).
+
   * Add new question renderer behind feature flag (Nathan Walters).
 
   * Fix HTML rendering by reverting `cheerio.js` to `0.22.0` (Matt West).
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -14,6 +14,7 @@
   * Add `pl-prairiedraw-figure` element and update PrairieDraw graphics documentation. (Ray Essick).
 
   * Add `pl-matrix-component-input` element (Mariana Silva).
+  * Add new question renderer behind feature flag (Nathan Walters).
 
   * Fix HTML rendering by reverting `cheerio.js` to `0.22.0` (Matt West).
 
--- a/doc/question.md
+++ b/doc/question.md
@@ -84,7 +84,6 @@ def generate(data):
 
 ```
 
-
 ## Question `question.html`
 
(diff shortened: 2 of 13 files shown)


trombonekenny:
Manually configuring to use a different parser seems....unsatisfying. Both for configuration management and for testing. Can we somehow use the new one, and fall back to the old one if there are errors? Or vice versa?

I guess it depends on the impact of changing the renderer - if that's worth doing it in a staged way, or if we can roll it out even if we know there's some breakage?

nwalters512:
@mfsilva22 needs the new parser ASAP. The plan is to move all courses to be compatible with the new parser (by round-tripping their code through the old parser to ensure tags are properly closed), but that will take time an potentially require a lot of manual effort. Old questions will not "fail" loudly with the new parser, as HTML parsers by design don't crash, but questions will absolutely be rendered incorrectly if their HTML is malformed.

This PR exists specifically so that people who know what they're doing can opt into the breaking behavior until we can ensure that 100% of existing content has been updated for compatibility with the new parser.

ChangeLog.md
```diff
@@ -14,6 +14,7 @@
   * Add `pl-prairiedraw-figure` element and update PrairieDraw graphics documentation. (Ray Essick).
 
   * Add `pl-matrix-component-input` element (Mariana Silva).
+  * Add new question renderer behind feature flag (Nathan Walters).
```
tbretl: Separate with blank line - no evil change-log.

nwalters512: Good catch - merging gone wrong!

nwalters512:
IIRC I made all questions compliant when I switched to use dashed element names. I'll add some docs and a question that actually uses element nesting 👍

PR #1210, opened 2018-07-31
LTI support for Coursera

trombonekenny:
Adds in LTI auth callbacks so we authenticate users coming from LMSes, make user accounts for them, auto-enroll them in the course. Including mapping a "link id" from the LMS to a specific assessment so students are brought right into the homework. Also maps LTI outcomes, which lets PL callback to the LMS when the assessment score is updated.

Useful links for devel and testing:
- https://www.imsglobal.org/learning-tools-interoperability-verifying-launch-messages
- http://ltiapps.net/test/tc.php
- As well as Coursera sandbox test space

Remaining things to track/figure out are in the comments below.


trombonekenny:
Storing LTI authentication

- Currently I'm overloading the `users.uin` field with unique user identifiers from LTI, but we should make them their own columns (`lti_user_id`, `lti_context_id`) We also get `lis_person_contact_email_primary` and `lis_person_name_full` which we could add to the `uid` and `name` user table columns.
- The current `users_select_or_insert` sproc updates based on `uid` and shares a common `user.id` so if I login as "mussulma@illinois.edu" via Gmail or Azure or Shib it points to the same user. We probably don't want that for LTI, as users are to interface with PL only via their LTI links.

I'm just not sure the best way to fix that. I'm leaning towards adding columns to the `user` table as needed and having LTI use its own SQL or sproc for updates that updates those columns and does not overload to update or use other `institution` rows.

trombonekenny:
Stop LTI users from accessing enrollments

- How to stop LTI users from adding other courses? Should we block that in the enrollment page?
- I'm not quite sure how the `institution` field protects the courses that are shown.

trombonekenny:
Reporting scores back to Coursera

- Ideally, we'd have a JS hook that ties into any score updates... whether it's student, or cron, or instructor triggered. Most of that's handled in a sproc, but maybe we should wrap that in a `lib` function call so we have a consistent place to add to the callback?
- When that's ready, it can call `lib/ltiOutcomes.updateScore(assessment_instance_id, callback)` which does the needful for LTI.
- I'm not sure what logging or error reporting for this library is needed.
- We might want to add a button to instructorAssessment that allows manually triggering running that LTI update for all the assessment instances? Shouldn't be needed if it's hooked in all the other places.

trombonekenny:
instructor course instance nav bar

Course instances need a place to configure LTI, so I made a new page. Right now I'm calling it instructorAdmin with `Admin` in the navbar. The only thing that's in it is the LTI configuration, but maybe we'd add more things later? It didn't seem worth it to have a tab at the top just for LTI configuration.

mwest1066:
Some responses to questions above:
- I agree that it's probably cleanest to write a new `users_select_or_insert_lti` that is completely separately from the existing `users_select_or_insert`. Adding columns to the `users` table also seems like the way to go here.
- To stop LTI users from adding other courses I think we should add some more checks to `sprocs/check_course_instance_access_rule.sql`. This will mean we need to pass some more data done to this function. This is where the current institution protection is done. How do we want the instructor to specify LTI access? They add a course instance access rule like `"LTI": "some magic string from Coursera"`?
- I still need to add a JS hook for score updates.
- I like the new instructorAdmin page. There will undoubtedly be other things we want to add here over time. Good idea.

trombonekenny:
The authCallbackLti workflow automatically makes them a user account and enrolls them in the course. An LTI identity shouldn't need to enroll in anything. At least that was my thinking. So maybe the check on the `check_course_instance_access_rule` just looks for `user.institution: lti` and denies off of that?

migrations/130_lti_links__create.sql
```diff
@@ -0,0 +1,15 @@
...
+    resource_link_id TEXT, -- should be unique?
+    resource_link_title TEXT,
+    resource_link_description TEXT,
+    created_at timestamptz DEFAULT current_timestamp,
+    deleted_at timestamptz NULL DEFAULT NULL
+);
+
+CREATE INDEX lti_links_resource_link_id_idx ON lti_links (resource_link_id);
+ALTER TABLE lti_links
+    ADD CONSTRAINT course_instance_context_link UNIQUE (course_instance_id, context_id, resource_link_id);
```
mwest1066: Should be called `course_instance_context_id_resource_link_id_key` for consistency with postgres naming conventions.

migrations/131_lti_outcomes__create.sql
```diff
@@ -0,0 +1,13 @@
...
+    user_id BIGINT REFERENCES users ON DELETE CASCADE ON UPDATE CASCADE,
+    assessment_id BIGINT REFERENCES assessments ON DELETE CASCADE ON UPDATE CASCADE,
+    lti_credential_id BIGINT REFERENCES lti_credentials(id) ON DELETE CASCADE ON UPDATE CASCADE,
+    lis_result_sourcedid TEXT,
+    lis_outcome_service_url TEXT
+);
+
+CREATE INDEX IF NOT EXISTS lti_outcomes_assessment_id ON lti_outcomes(assessment_id);
+ALTER TABLE lti_outcomes
+    ADD CONSTRAINT lti_outcomes_user_assessment UNIQUE (user_id, assessment_id);
```
mwest1066: Should be called `lti_outcomes_user_id_assessment_id_key`

pages/authCallbackLti/authCallbackLti.js
```diff
@@ -0,0 +1,201 @@
...
+            }
+        });
+        if (nonceReused) return next(error.make(500, 'Invalid nonce reuse'));
+
+        //
+        // OAuth validation succeeded, next look up and store user authn data
+        //
+
+        //console.log(parameters);
+        //console.log(ltiresult);
```
mwest1066: delete leftover debug code

pages/authCallbackLti/authCallbackLti.js
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        /* Conflicts with other UIDs, potentially
+           Also means we have no way to store the email coming from the consumer
+        if (parameters.lis_person_contact_email_primary) {
+            authUid = parameters.lis_person_contact_email_primary;
+        } else if (parameters.lis_person_sourcedid) {
+            authUid = parameters.lis_person_sourcedid;
+        }
+        */
```
mwest1066: What's the status of the above commented-out code? I'd rather just delete it.

pages/authCallbackLti/authCallbackLti.js
```diff
@@ -0,0 +1,201 @@
...
+            // e.g. UIUC Degree Sandbox user
+        }
+        var authName = parameters.lis_person_name_full || fallbackName;
+
+        var params = [
+        authUid,
+        authName,
+        ltiresult.course_instance_id,
+        parameters.user_id,
+        parameters.context_id,
```
mwest1066: should be indented

pages/authCallbackLti/authCallbackLti.js
```diff
@@ -0,0 +1,201 @@
...
+                provider: 'lti',
+                /*
+                lti_launch_presentation_return_url: parameters.launch_presentation_return_url,
+                resource_link_id: parameters.resource_link_id,
+                resource_link_title: parameters.resource_link_title,
+                resource_link_description: parameters.resource_link_description,
+                context_id: parameters.context_id,
+                context_label: parameters.context_label,
+                context_title: parameters.context_title,
+                */
```
mwest1066: delete commented-out code?

pages/authCallbackLti/authCallbackLti.js
```diff
@@ -0,0 +1,201 @@
...
+                        // No linked assessment
+
+                        if (role != 'Student') {
+                            redirUrl = `${res.locals.urlPrefix}/course_instance/${ltiresult.course_instance_id}/instructor/admin`;
+                        } else {
+                            // Show an error that the assignment is unavailable
+                            return next(error.make(400, 'Assignment not available yet'));
+
+                            // Default them into the course instance
+                            //redirUrl = res.locals.urlPrefix + '/course_instance/' + ltiresult.course_instance_id;
```
mwest1066: delete unused code?

pages/instructorAdmin/instructorAdmin.sql
```diff
@@ -0,0 +1,102 @@
...
+                'tid', a.tid
+            ) ORDER BY (aset.number, a.order_by, a.id)) as assessments
+    FROM assessments AS a
+    JOIN assessment_sets AS aset ON (aset.id = a.assessment_set_id)
+    WHERE a.deleted_at IS NULL
+    AND a.course_instance_id = $course_instance_id
+),
+lti_credentials AS (
+    SELECT
+    --jsonb_agg(lti_credentials) as lti_credentials
```
mwest1066: delete?

pages/instructorAdmin/instructorAdmin.sql
```diff
@@ -0,0 +1,102 @@
...
+        'created_at', lc.created_at
+    ) ORDER BY (lc.created_at)) as lti_credentials
+    FROM lti_credentials AS lc
+    JOIN course_instances AS ci ON(ci.id = lc.course_instance_id)
+    WHERE course_instance_id = $course_instance_id
+
+),
+lti_links AS (
+    SELECT
+--        jsonb_agg(lti_links ORDER BY created_at) AS lti_links
```
mwest1066: delete?

pages/instructorAdmin/instructorAdmin.ejs
```diff
@@ -0,0 +1,180 @@
...
+                        <% }); %>
+                        <% } else { %>
+                            <tr><td colspan=99><em>No credentials have been established.</em></td></tr>
+                        <% } %>
+                    <tr>
+                        <td colspan="99">
+                            <form method="post">
+                                <input type="hidden" name="__action" value="lti_new_cred">
+                                <input type="hidden" name="__csrf_token" value="<%= __csrf_token %>">
+                                <input type="submit" class="btn btn-success" value="Create new">
```
mwest1066: Change text to "Create new LTI credential"

mwest1066:
I think we want to include `lti` in the list of methods in `pages/partials/logout.ejs` that display the "Logout" button.

server.js
```diff
@@ -305,8 +306,17 @@ app.use('/pl/course_instance/:course_instance_id/instructor/grading_job', requir
...
+// admin pages
+app.use(/^(\/pl\/course_instance\/[0-9]+\/instructor\/admin)\/?$/, (req, res, _next) => {
+    res.redirect(`${req.params[0]}/access`);
+});
+app.use('/pl/course_instance/:course_instance_id/instructor/admin/access', [
+    require('./pages/instructorAdminAccess/instructorAdminAccess'),
+]);
+app.use('/pl/course_instance/:course_instance_id/instructor/admin/lti', [
+    require('./pages/instructorAdminLti/instructorAdminLti'),
+]);
```
mwest1066: We should also mount `./pages/courseOverview/courseOverview` at the route `/pl/course_instance/:course_instance_id/instructor/admin/course` so that it can appear as a real tab, not as a weird fake tab where you click the tab but then go to a totally different page. This will require some changes to `navbarInstructor.ejs` to render the dropdown correctly for this page, and also some changes to `courseOverview.ejs` to include the tabbar when mounted on the `instructor/admin` route. We can detect which route we are on by either setting a special variable in `server.js` just before the `require`, or by looking for whether `locals.course_instance` is defined.

PR #1218, opened 2018-08-23
Include input pl editor

mfsilva22:
This PR addresses the following for the `pl-file-editor` element
- Allows adding an empty text editor (at least I was not able to start from empty editors with the original code)
- Allows displaying text inside editor from another file


doc/elements.md
```diff
@@ -358,6 +358,7 @@ Attribute | Type | Default | description
 `file-name` | string | - | The name of this file; will be used to store this file in the `_files` submitted answer
 `ace-mode` | string | None | Specifies an Ace editor mode to enable things like intelligent code indenting and syntax highlighting; see the full list of modes [here](https://github.com/ajaxorg/ace/tree/master/lib/ace/mode).
 `ace-theme` | string | `ace/theme/chrome` | Specifies an Ace editor theme; see the full list of themes [here](https://github.com/ajaxorg/ace/tree/master/lib/ace/theme).
+`input-file-name` | string | None | Name of the file with content to be displayed inside the text editor (instead - or in addition to - writing the text between the open/closing tags).
```
nwalters512: Similar to #1219, I'd consider `source-file-name` and adding support for a `source-directory`.

gchallen:
@trombonekenny: if it works, sure! I had observed that an empty editor led to "None" being displayed, which is obviously a Python-based bug. Thanks @mfsilva22!

elements/pl-file-editor/pl-file-editor.py
```diff
@@ -47,7 +49,16 @@ def render(element_html, element_index, data):
...
+    if element.text is not None:
+        text_display = str(element.text)
+    else:
+        text_display = ''
+
+    if source_file_name is not None:
+        file_path = os.path.join(data['options']['question_path'], source_file_name)
+        html_params['original_file_contents'] = base64.b64encode(str(text_display + open(file_path).read()).encode('UTF-8').strip()).decode()
+    else:
+        html_params['original_file_contents'] = base64.b64encode(text_display.encode('UTF-8').strip()).decode()
```
mwest1066: Can we replace the above if/else with the following?
```
    if source_file_name is not None:
         file_path = os.path.join(data['options']['question_path'], source_file_name)
         text_display += open(file_path).read()
    html_params['original_file_contents'] = base64.b64encode(text_display.encode('UTF-8').strip()).decode()
```

exampleCourse/questions/fibonacciEditor/question.html
```diff
@@ -1,22 +1,21 @@
...
+
+<pl-file-editor file-name="fib.py" ace-mode="ace/mode/python" source-file-name="tests/initial_code.py"></pl-file-editor>
+
 </pl-question-panel>
 
 <pl-submission-panel>
-  <pl-file-preview></pl-file-preview>
-  <pl-external-grader-results></pl-external-grader-results>
+<pl-file-preview></pl-file-preview>
+<pl-external-grader-results></pl-external-grader-results>
```
mwest1066: Most of the changes to this file seem to just be removing indenting. Was this intentional or did an editor do it? I'd rather change only the single <pl-file-editor> line that really needed changing, if possible.

exampleCourse/questions/fibonacciEditor/tests/initial_code.py
```diff
@@ -0,0 +1,3 @@
+def fib(n):
+
```
mwest1066: Spurious blank line?

exampleCourse/questions/fibonacciEditor/tests/initial_code.py
```diff
@@ -0,0 +1,2 @@
+def fib(n):
```
nwalters512: Nit: the `tests` directory is really reserved for external grading support - could you put this file in the question root instead?

doc/elements.md
```diff
@@ -358,6 +358,7 @@ Attribute | Type | Default | description
 `file-name` | string | - | The name of this file; will be used to store this file in the `_files` submitted answer
 `ace-mode` | string | None | Specifies an Ace editor mode to enable things like intelligent code indenting and syntax highlighting; see the full list of modes [here](https://github.com/ajaxorg/ace/tree/master/lib/ace/mode).
 `ace-theme` | string | `ace/theme/chrome` | Specifies an Ace editor theme; see the full list of themes [here](https://github.com/ajaxorg/ace/tree/master/lib/ace/theme).
+`source-file-name` | string | None | Name of the file with content to be displayed inside the text editor (instead - or in addition to - writing the text between the open/closing tags).
```
nwalters512: What's the use case for having starting contents come from both a file and the element contents? Not opposed to this necessarily, just curious.

mfsilva22: I didn't think I could remove the option to add content from inside the element, so I just added the option to add content from a source file as well.

In my case, I would just use one or the other, but I think keeping both adds more flexibility?

nwalters512: I don't take issue with that, it's the appending behavior that's potentially confusing. I feel like it should be one or the other, probably enforced in code.

mfsilva22: okay, I will make it be one or the other. How do I go about `pl-code`. Do I start another branch and do another PR?

mfsilva22: also, since it is one or the other, I am assuming the file content will override the element content when someone uses both? or should it throw an error?

exampleCourse/questions/fibonacciEditor/tests/initial_code.py
mfsilva22: should I change the example back to original format, where the code is inserted inside the element, and not from a file? In that case, I will just delete the file `initial-code.py`. Or do you think it is better to keep the source file as the example?

nwalters512: I would just put two elements in that question, one sourced from a file, one sourced from the children of the element. The question doesn't necessarily have to make sense - I view them mostly as demos of how to use the code.

doc/elements.md
nwalters512: Yep, just open a new PR. @mwest1066 what do you think about throwing an error? I think enforcing that it's only one just so behavior isn't silently wrong is probably best.

exampleCourse/questions/fibonacciEditor/tests/initial_code.py
mfsilva22: ok!

mfsilva22: If I add two elements, would't that mess up the autograder?

nwalters512: Ah, sorry, forgot which PR I'm in :) I would say to prefer the `source-file-name` attribute in this example.

doc/elements.md
mwest1066: Yeah, new PR for the `pl-code` fix.

I think throwing an error if both are provided is a good idea, to save anyone from surprises. For the file it's easy to tell whether it's provided (do we have the `source-file-name` attribute or not). For the inline text, I think it's a bit harder to tell whether it's provided. I think we need to check whether there is text, and if so, whether the text is just whitespace?

PR #1223, opened 2018-08-27
activate Logout button for google authn

mwest1066:


pages/partials/logout.ejs
```diff
@@ -1,6 +1,8 @@
 <% if (typeof authn_user != 'undefined') { %>
    <% if (authn_user.provider == 'shibboleth') { %>
        <a class="btn btn-secondary" href="<%= plainUrlPrefix %>/logout">Log out</a>
+   <% } else if (authn_user.provider == 'google') { %>
```
nwalters512: Why this as opposed to `|| authn_user.provider == 'google'` on the conditional above?

mwest1066: laziness? :-)

mwest1066: fixed now

PR #1229, opened 2018-08-30
[Ready to merge] Access denied error rather than "Incorrect row count"

### LINKED ISSUE #1221: Incorrect Row Count
Frequently my attempts to link directly to assessments on PL produce results like this for students:

image

I have a hard time reproducing this problem (the links work for me, that's why I send them in the first place). But it seems to be caused by either:

1. accidentally providing an instructor link to a student, or
1. cases where the student isn't yet signed up for the class on PL

In both cases it would be helpful if a more useful error message was displayed. In the first case it seems like a simple redirect to the student version of the assessment would be possible, and in the second perhaps a redirect to a page where they can sign up for the course.

trombonekenny:
I ran into this error again recently, so might as well PR and make it more descriptive.

Closes #1221

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -61,6 +61,8 @@
 
   * Change assessment password protection method (Dave Mussulman).
 
+  * Change "0 rows" error to be more descriptive (Dave Mussulman).
+
 * __3.0.0__ - 2018-05-23
 
   * Add improved support for very large file downloads (Nathan Walters).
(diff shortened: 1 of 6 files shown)


trombonekenny:
The challenge is by changing the error HTTP response code from the default (500) to the more specific access denied (403 -- I borrowed the log logic and lines from the course instance access middleware) now all of the tests are broken.

Design choice: Change the tests to 403? Or change the error message to use response 500 and don't touch the tests? @mwest1066?

mwest1066:
I agree that changing the tests is the right thing to do. It shouldn't be very hard -- just some helper functions will need changing in `tests/testAccess.js`, I think.

mwest1066:
Nice!

Issue #1234, opened 2018-09-01
per-zone maxPoints, per-zone “best N of M questions” grading


coatless:
From @mwest1066 on Slack regarding the maxPoints per-zone:

1. Write a migration to add a database column to store the new maxPoints in the zone table, sort of like this: https://github.com/PrairieLearn/PrairieLearn/blob/master/migrations/121_pl_courses__commit_hash__add.sql 
1. Change the `infoAssessment.json` schema to allow the new value to be configured: https://github.com/PrairieLearn/PrairieLearn/blob/master/schemas/infoAssessment.json 
1. Enable the configuration and storage in the DB: https://github.com/PrairieLearn/PrairieLearn/blob/master/sync/fromDisk/assessments.js 
1. Modify the grading scheme https://github.com/PrairieLearn/PrairieLearn/blob/master/sprocs/assessment_instances_points.sql

PR #1238, opened 2018-09-06
Per zone grading

tbretl:
Zones have two new options:

* If a zone has `maxPoints`, then, of the points that are awarded for answering questions in this zone, at most `maxPoints` will count toward the total points.

* If a zone has `numberGrade`, then, of the questions in this zone, only `numberGrade` with the highest number of awarded points will count toward the total points.

These options can, of course, be used at the same time.

Two things that have not yet been done that probably should be done before merging are:

* Add an instructor page UI (information is only provided on `studentAssessmentInstance*` pages).

* Add unit tests.

Closes #1234


migrations/124_instance_questions__used_for_grade__add.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE instance_questions ADD COLUMN IF NOT EXISTS used_for_grade BOOLEAN;
```
mwest1066: Should we make this `NOT NULL` and `DEFAULT FALSE`?

doc/assessment.md
```diff
@@ -84,6 +84,10 @@ An assessment is broken down in to a list of zones, like this:
 
 * An assessment question can be specified by either a single `id` or by a list of alternatives, in which case one or more of these alternatives is chosen at random. Once the question `id` is determined, then a random variant of that question is selected. Question alternatives inherit the points of their parent group, if specified.
 
+* If a zone has `maxPoints`, then, of the points that are awarded for answering questions in this zone, at most `maxPoints` will count toward the total points.
+
+* If a zone has `numberGrade`, then, of the questions in this zone, only `numberGrade` with the highest number of awarded points will count toward the total points.
```
mwest1066: I agree with @nwalters512 in that I'm not wild about the name of `numberGrade`. Perhaps it's because we are grading all of them, but only using some of them? I'm aware we are probably well into bike-shedding territory here.

Looking at the HTML we show to users, we say "maximum X points", which sounds a lot like `maxPoints`. Then we say "best N questions", so maybe we the attribute should be `bestQuestions`?

exampleCourse/courseInstances/Sp15/assessments/perzonegrading_hw/infoAssessment.json
```diff
@@ -0,0 +1,50 @@
...
+            "questions": [{
+                    "id": "partialCredit6_no_partial",
+                    "points": 4,
+                    "maxPoints": 8
+                }
+            ]
+        },
+        {
+            "title": "Questions to test numberGrade",
+            "numberGrade": 2,
```
mwest1066: I see that `perzonegrading_exam` also has `numberGrade = 2`. Would it be worth making this one different to (slightly) improve test coverage?

sprocs/assessment_instances_points.sql
```diff
@@ -34,81 +29,84 @@ BEGIN
...
-        max_possible_score_perc := greatest(max_possible_score_perc, score_perc);
-    END IF;
-
-    -- compute score_perc_in_grading
-    score_perc_in_grading := max_possible_score_perc - score_perc;
+    -- compute the points by zone
+    RETURN QUERY
+        WITH all_questions AS (
+            SELECT
+                iq.id AS qid,
```
mwest1066: Can we call this `iq_id` rather than `qid`? We are already using `qid` to specifically mean the short names of questions (the column `questions.qid`).

doc/assessment.md
tbretl: done - went with `bestQuestions`

sprocs/assessment_instances_points.sql
tbretl: good point, done.

exampleCourse/courseInstances/Sp15/assessments/perzonegrading_hw/infoAssessment.json
tbretl: agreed, done.

migrations/124_instance_questions__used_for_grade__add.sql
tbretl: If I understand correctly, doing so would populate the new column with `FALSE` everywhere. I don't think this is what we want.

Leaving this migration as is allows us to distinguish between cases in which instance questions have or have not been used (`TRUE` or `FALSE`) and cases in which - because they occurred prior to this PR - we don't know (`NULL`).

Does that make sense?

tbretl:
@nwalters512 - I had already added docs - were they not enough?

I like the suggestion about making the entire label the button. It's more natural to click, and actually looks better too. Perhaps further UI improvement could be left to later?

Good call on the name. I went with `bestQuestions`, as suggested by Matt.

migrations/124_instance_questions__used_for_grade__add.sql
mwest1066: Yep, agreed.

mwest1066:
Oops, I commented on an earlier commit. Please disregard if already addressed. I’ll look in more detail in the morning.

mfsilva22:
I tested maxPoints and bestQuestions for an assessment in CS 357, and I could not find any issue. I was able to combine maxPoints and bestQuestions, use one or the other, or neither, and the results were as expected.

I am not sure this would look better or not, but what about having the maximum number of points for a given zone to always be displayed next to the zone title (instead of only displaying when maxPoints option is used? I think this can be useful when you have many questions in the assessment, and a quick glance can indicate how many points is available in each zone (yeah... I am getting lazy here and don't want to do some simple summation :-))

mwest1066:
Thanks @mfsilva22 for the testing! Most appreciated! I think I'll merge this as is and we can add those good UI suggestions later.

PR #1239, opened 2018-09-06
Add option to disable selecting on pl-code element

### LINKED ISSUE #1224: Add obfuscation to pl-code element
In 225, we convert all source code in questions into images with the goal of preventing copy-paste to just run the code during exams. It's of questionable utility in the first place (anyone with a decent WPM can just retype the code in under a minute), but assuming this is a desirable property sometimes, we should add optional similar protections to the `<pl-code>` element. I'm thinking something as simple as disabling right-click and preventing text selection will be enough for most people. This doesn't prevent people from just viewing the source, of course, but I'm guessing that's ok.

Open to suggestions for attribute names. `obfuscate="true"`? This should definitely default to false.

nwalters512:
This adds an option to attempt to prevent students from copy/pasting the source code of a `pl-code` element. It disables selection, clicking, and right-clicking on the code. It's still relatively easy to see the source - just going to "View source" on the page and grepping for the code will reveal it. But frankly, I don't care that much 🙂 

Resolves #1224

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -23,6 +23,8 @@
 
   * Add partial credit option to `pl-checkbox` element (Mariana Silva).
 
+  * Add `prevent-select` attribute to `pl-code` element (Nathan Walters).
+
   * Fix `pl-file-editor` to allow display empty text editor and add option to include text from source file (Mariana Silva).
 
   * Fix HTML rendering by reverting `cheerio.js` to `0.22.0` (Matt West).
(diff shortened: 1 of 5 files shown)


mwest1066:
Very nice!

If we wanted to obscure it even more, could we put the text in as base64-encoded with a little bit of JS to decode it and insert it into the DOM? That way the view-source would be obscured, and the DOM view in the browser is already quite obscured by all the extra markup that the highlighting introduces. This is just me musing -- I think we should merge&deploy this as is.

coatless:
Missing a corresponding entry in the element help doc for `prevent-select`

PR #1242, opened 2018-09-07
Eliminate element_index from arguments passed to element functions

### LINKED ISSUE #1240: change in behavior of element_index with new renderer
All element functions (e.g., `prepare`) are given the integer `element_index` as an argument.

With the old renderer, `element_index` came from **one** counter over **all** elements. So, it was unique for each instance of any element.

With the new renderer, it appears that `element_index` comes from a **separate** counter for **each** element. So, it is only unique for instances of a given element.

This is causing problems, because some elements rely on `element_index` being (globally) unique. For example, both `pl-matrix-input` and `pl-matrix-output` both use `element_index` in the `id` for copying to clipboard. So, with the new renderer, instances of these two elements cannot reliably be used together on the same page.

For an example of the bug, see the question `multiplyTwoMatrices` in `exampleCourse`.

tbretl:
The argument `element_index` is no longer passed to element functions. All element functions now have this signature:
```python
def fcn(element_html, data)
```
This change is backward compatible, in the sense that course elements that continue to use the old signature will still receive `element_index`, but the value of this argument will be `None`.

Instead of relying on `element_index` as a unique id, all core elements now rely on a UUID, generated within element code.

This eliminates any issues with the new renderer, with which `element_index` was no longer unique.

Fixes #1240

### DIFF
--- a/lib/python-caller-trampoline.py
+++ b/lib/python-caller-trampoline.py
@@ -72,12 +72,11 @@ with open(3, 'w', encoding='utf-8') as outf:
             # get the desired function in the loaded module
             method = getattr(mod, fcn)
 
-            # check if the desired function is a legacy element function - if so,
-            # we assume that ... and add argument for element_index
-            sig = signature(method)
-            if len(sig.parameters) == 3:
-                if all(p in sig.parameters for p in ['element_html', 'element_index', 'data']):
-                    args.insert(1, None)
+            # check if the desired function is a legacy element function - if
+            # so, we add an argument for element_index
+            arg_names = list(signature(method).parameters.keys())
+            if len(arg_names) == 3 and arg_names[0] == 'element_html' and arg_names[1] == 'element_index' and arg_names[2] == 'data':
+                args.insert(1, None)
 
             # call the desired function in the loaded module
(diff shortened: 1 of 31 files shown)


lib/python-caller-trampoline.py
```diff
@@ -69,8 +69,17 @@ def try_dumps(obj, sort_keys=False, allow_nan=False):
...
         if hasattr(mod, fcn):
-            # call the desired function in the loaded module
+            # get the desired function in the loaded module
             method = getattr(mod, fcn)
+
+            # check if the desired function is a legacy element function - if so,
+            # we assume that ... and add argument for element_index
+            sig = signature(method)
+            if len(sig.parameters) == 3:
+                if all(p in sig.parameters for p in ['element_html', 'element_index', 'data']):
```
mwest1066: nitpick: it would be slightly more specific to do `if sig.parameters[0] == 'element_html' and sig.parameters[1] == 'element_index' and sig.parameters[2] == 'data':`

tbretl: I agree. Fixed.

Issue #1243, opened 2018-09-08
Add Programmatic Grade Retrieval API


gchallen:
Those of us that build our own courseware would appreciate a programmatic API to retrieve data from PrairieLearn. This is particularly pressing given that the coming 2FA requirement is going to severely complicate and possible entirely prohibit data retrieval scripts that rely on web automation—which is what I'm doing currently—since it will become more difficult to automate login.

An API _should_ also reduce the load of data retrieval on the PrairieLearn frontend. At the moment retrieving (kind of) detailed assessment data requires issuing N + 1 GET requests, where N is the number of assessments. Given that our intention in CS 125 is to assign one PrairieLearn homework problem per day, N is not going to scale well as the semester goes on.

I would imagine that this would work something like this:

1. Users with role Instructor or TA would be able to retrieve an API access token from their profile page. I'd suggest using a UUIDv4 for this, since they are heavily used in the rest of the PL codebase. This token could be reset at any time. (Does PL even have a profile page? The closest thing I can find is the page that allows me to change my effective role. That might be suitable, or you might prefer to put it elsewhere.)
1. HTTPS GET requests that include that API token in their header would be authenticated as the user that owns the token.
1. New endpoints would be set up to return data via JSON. At first I think it would be reasonable to mimic some of the data available via CSV download, but eventually it might be better to include more information. I'd like a single endpoint that returns a fairly detailed summary for an entire course instance. For example, for all users and all assessments and all submitted answers: the received timestamp and score. (It's currently hard to maintain timeseries data from PL. I do this via regular snapshots, but this both creates unnecessary load for PL and fails when my daemon crashes... which does, in fact, happen, like when someone decides to add a fancy new login page 😄.)

So my questions are:

1. Is this something that is on the PL roadmap for the near future?
1. If not, would PL accept a set of patches to add this functionality? Or would a core PL developer prefer to develop and deploy this functionality, assuming it can be done fairly quickly. (The 2FA page claims that this will start to be required for staff in November.)
1. If so, would some of the current developers like to comment on the proposal above and offer feedback to try and ensure that this feature can be smoothly integrated? This does touch a few different parts of the site, including endpoint layout, database access, and authentication.

nwalters512:
1. It's been on our to-do list, but was not necessarily being planned for the near future. Given that the upcoming 2FA requirement will break any existing workflows that rely on web scraping, I think it would be reasonable to move this up in our queue.

[...]

PR #1245, opened 2018-09-08
Add personal access token management

nwalters512:
This PR adds the ability to create, view, and delete personal access tokens; this is the first part of #1243. Tokens aren't currently used for anything, but I felt it best to break #1243 into multiple PRs to avoid a single monolithic PR that's impossible to review.

## Security concerns

I did my best to design this with security in mind. The generated token is only persisted in the database for a short period of time; this is necessary to display it back to the user on the new request. Immediately after the token list is rendered, all tokens for that user are removed from the database. I originally stored the token only in memory, but that will break in a multi-server architecture.

We only permanently store a SHA256 hash of the token in the database. SHA256 is not _necessarily_ good for passwords, but given that the tokens are guaranteed to be 36 characters long, I'm not too concerned about the weakness of SHA256 when used to hash passwords (it's a very fast algorithm and can thus be brute-forced comparatively faster than something like bcrypt). I would have used bcrypt itself, but its reliance on a native component means that the "mount the PL repo into Docker" development paradigm would break.

Here's what the page looks like to a user:

screen shot 2018-09-08 at 3 41 52 pm

### DIFF
--- a/pages/userSettings/tokenDeleteForm.ejs
+++ b/pages/userSettings/tokenDeleteForm.ejs
@@ -1,0 +1,10 @@
+<form name="add-user-form" method="POST">
+  <input type="hidden" name="__action" value="token_delete">
+  <input type="hidden" name="__csrf_token" value="<%= __csrf_token %>">
+  <input type="hidden" name="token_id" value="<%= token_id %>">
+  <p>Once you delete this token, any applications using it will no longer be able to access the API. You cannot undo this action.</p>
+  <button type="button" class="btn btn-secondary" onclick="$('#<%= id %>').popover('hide')">
+    Cancel
+  </button>
+  <button type="submit" class="btn btn-danger">Delete token</button>
+</form>
--- a/pages/userSettings/userSettings.ejs
+++ b/pages/userSettings/userSettings.ejs
@@ -15,11 +15,11 @@
             type="button" class="btn btn-light btn-sm ml-auto"
             data-toggle="popover" data-container="body"
             data-html="true" data-placement="auto" title="Generate new token"
-            data-content="<%= include('generateTokenForm', {id: 'generateTokenButton'}) %>"
+            data-content="<%= include('tokenGenerateForm', {id: 'generateTokenButton'}) %>"
             data-trigger="manual" onclick="$(this).popover('show')">
             <i class="fa fa-plus" aria-hidden="true"></i>
             <span class="d-none d-sm-inline">Generate new token</span>
-          </a>
+          </button>
         </div>
         <div class="card-body">
           <p class="mb-0">You can generate tokens in order to access the PrairieLearn API.</p>
@@ -38,11 +38,20 @@
             <span class="text-muted">You don't currently have any access tokens</span>
           </li>
           <% } else { %>
-          <% accessTokens.forEach(function (accessToken) { %>
+          <% accessTokens.forEach(function (token) { %>
           <li class="list-group-item d-flex align-items-center">
-            <span class="mr-3"><%= accessToken.name %></span>
-            <span><%= accessToken.token %></span>
-            <button type="button" class="btn btn-outline-danger ml-auto">Delete</button>
+            <div class="d-flex flex-column mr-3">
+              <strong><%= token.name %></strong>
+              <span class="text-muted">Created at <%= token.created_at %></span>
+            </div>
+            <button id="generateTokenButton"
+              type="button" class="btn btn-outline-danger btn-sm ml-auto"
+              data-toggle="popover" data-container="body"
+              data-html="true" data-placement="auto" title="Delete this token"
+              data-content="<%= include('tokenDeleteForm', {id: 'deleteTokenButton' + token.id, token_id: token.id }) %>"
+              data-trigger="manual" onclick="$(this).popover('show')">
+              <span class="d-none d-sm-inline">Delete</span>
+            </button>
           </li>
           <% })}; %>
         </ul>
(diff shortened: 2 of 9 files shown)


trombonekenny:
UI looks sharp. (I'd probably combine the blue and green bands into one single band with new lines.) I might steal some of that for the LTI credential management.

"Last used" might be an interesting thing to track and display from that interface.

I'm not sure hashing the token (instead of storing it directly) in the database is necessary. If someone did compromise our database, learning some personal tokens isn't the issue.

nwalters512:
The blue/green was copied from GitHub's token UI; I don't necessarily agree that they should be part of the same box.

I was thinking about tracking last used time - that's a good idea, I can add that.

I've just had "don't store plaintext password" drilled into my head by 461 :) but true, it might not be an issue for us.

pages/userSettings/userSettings.js
```diff
@@ -0,0 +1,72 @@
...
+
+        const params = {
+            user_id: res.locals.authn_user.user_id,
+            name,
+            // The token will only be persisted until the next page render
+            // After that, we'll remove it from the database
+            token,
+            token_hash: tokenHash,
+        };
+        sqldb.queryOneRow(sql.insert_access_token, params, (err) => {
```
mwest1066: Can we insert a row into `audit_logs` whenever we do anything with tokens? For example, insert like `https://github.com/PrairieLearn/PrairieLearn/blob/master/sprocs/assessment_instances_update.sql#L73` or delete like `https://github.com/PrairieLearn/PrairieLearn/blob/master/sprocs/assessment_instances_delete.sql#L31`. We don't use `audit_logs` for anything in the UI, but I've found it helpful in the past when tracking down bugs and also when figuring out what some weird user had been up to :-)

A pattern we were using elsewhere is to make a file like `sprocs/access_tokens_insert.sql` that does both the insert itself and the audit log update. I don't mind whether we continue that pattern or just call again from JS to update the audit log.

pages/userSettings/userSettings.ejs
```diff
@@ -0,0 +1,63 @@
...
+            <span class="text-muted">You don't currently have any access tokens</span>
+          </li>
+          <% } else { %>
+          <% accessTokens.forEach(function (token) { %>
+          <li class="list-group-item d-flex align-items-center">
+            <div class="d-flex flex-column mr-3">
+              <strong><%= token.name %></strong>
+              <span class="text-muted">Created at <%= token.created_at %></span>
+            </div>
+            <button id="generateTokenButton"
```
mwest1066: `id` should be `"deleteTokenButton" + token.id`?

pages/userSettings/tokenDeleteForm.ejs
```diff
@@ -0,0 +1,10 @@
+<form name="add-user-form" method="POST">
```
mwest1066: form name should be `token-delete-form`?

pages/userSettings/userSettings.js
```diff
@@ -0,0 +1,72 @@
...
+            if (ERR(err, next)) return;
+            res.redirect(req.originalUrl);
```
mwest1066: Rather than doing a redirect, we could actually just render the page directly here. That way we wouldn't need to do this dance of store-token-then-delete. I normally prefer the POST/redirect/GET pattern (see https://prairielearn.readthedocs.io/en/latest/dev-guide/#state-modifying-post-requests and https://en.wikipedia.org/wiki/Post/Redirect/Get) but this is special case where rendering the page in the POST handler might actually be preferable?

nwalters512: If we don't use PRG here, refreshing after creating a token would cause another token to be created in most user agents. I would definitely like to not be doing this little dance, but I do think we still need PRG here.

mwest1066: Agreed.

sprocs/access_tokens_delete.sql
```diff
@@ -15,7 +15,7 @@ BEGIN
         a.* INTO old_row;
 
     INSERT INTO audit_logs
-        (user_id, table_name,
+        (authn_user_id, table_name,
```
mwest1066: I think we should actually store both `user_id` and `auth_user_id` into the two columns with those names in `audit_logs`.

mwest1066:
Other than the couple of minor comments I just made, I think this is ready to merge.

sprocs/access_tokens_delete.sql
nwalters512: Are they ever going to be different on this page? IIRC you can't have an effective user outside the context of course instances.

mwest1066: Say an instructor changes their effective user id (`user_id`) to that of a TA in the class, and then creates/deletes access tokens for the TA. Then `authn_user_id` would be the instructor and `user_id` would be the TA.

nwalters512: Right, what I'm saying is that I can't find a way to make the effective user that I set on a course instance take effect on the settings page. I can become person X in a course instance, but once I go back to the homepage or visit my settings page, my identity reverts to its actual one.

mwest1066: Oh right, good point, I think the settings page is special in this respect. In this case I agree it’s not necessary to distinguish the two types of user ids, although we could always store both for completeness :-)

nwalters512: Good idea, that'll potentially make querying easier in the future. Done 🙂

gchallen:
Awesome! Great work.

PR #1252, opened 2018-09-14
Yet another symbolic input fix (handling E and I)

tbretl:
This time the issue was handling `sympy.E`.

Before, if the submitted answer were to contain something like `exp(1)`, then it would be parsed to `sympy.E`, which would be converted to the string `E`, which - upon the double-check parse - would raise an exception, because `E` wouldn't be in the list of variables (and if it were, that would be even worse).

Now, we handle this properly, by substituting a hidden helper variable for `sympy.E`.

We also handle `sympy.I` properly now, and - as a bonus - add an option to allow symbolic inputs with complex numbers, which wasn't ever allowed before.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -29,6 +29,8 @@
 
   * Add personal access token management (Nathan Walters).
 
+  * Add `allow-complex` attribute for `pl-symbolic-input` (Tim Bretl).
+
   * Fix `pl-file-editor` to allow display empty text editor and add option to include text from source file (Mariana Silva).
 
   * Fix HTML rendering by reverting `cheerio.js` to `0.22.0` (Matt West).
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -67,6 +67,8 @@
 
   * Fix bug in `pl-symbolic-input` to handle submissions that simplify to invalid expressions (Tim Bretl).
 
+  * Fix bug in `pl-symbolic-input` to handle the sympy constants I and E properly (Tim Bretl).
+
   * Change `pl-code` to display code from a source file OR inline text (Mariana Silva).
 
   * Change element names to use dashes instead of underscores (Nathan Walters).
--- a/elements/pl-symbolic-input/pl-symbolic-input.py
+++ b/elements/pl-symbolic-input/pl-symbolic-input.py
@@ -49,7 +49,7 @@ def render(element_html, data):
         raw_submitted_answer = data['raw_submitted_answers'].get(name, None)
 
         operators = ', '.join(['cos', 'sin', 'tan', 'exp', 'log', 'sqrt', '( )', '+', '-', '*', '/', '^', '**'])
-        constants = ', '.join(['pi'])
+        constants = ', '.join(['pi, e'])
 
         info_params = {
             'format': True,
--- a/elements/pl-symbolic-input/pl-symbolic-input.py
+++ b/elements/pl-symbolic-input/pl-symbolic-input.py
@@ -19,7 +19,7 @@ def get_variables_list(variables_string):
 def prepare(element_html, data):
     element = lxml.html.fragment_fromstring(element_html)
     required_attribs = ['answers-name']
-    optional_attribs = ['weight', 'correct-answer', 'variables', 'label', 'display', 'allow-complex']
+    optional_attribs = ['weight', 'correct-answer', 'variables', 'label', 'display', 'allow-complex', 'imaginary-unit-for-display']
     pl.check_attribs(element, required_attribs, optional_attribs)
     name = pl.get_string_attrib(element, 'answers-name')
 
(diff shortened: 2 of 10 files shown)


ChangeLog.md
```diff
@@ -67,6 +67,8 @@
 
   * Fix bug in `pl-symbolic-input` to handle submissions that simplify to invalid expressions (Tim Bretl).
 
+  * Fix bug in `pl-symbolic-input` to handle the sympy constants I and E properly (Tim Bretl).
+
```
mwest1066: Add a second ChangeLog line something like "Add `allow-complex` attribute for `pl-symbolic-input`"?

elements/pl-symbolic-input/pl-symbolic-input.py
```diff
@@ -130,10 +142,13 @@ def render(element_html, data):
             html = chevron.render(f, html_params).strip()
 
     elif data['panel'] == 'answer':
-        a_tru = pl.from_json(data['correct_answers'].get(name, None))
+        a_tru = data['correct_answers'].get(name, None)
         if a_tru is not None:
             if isinstance(a_tru, str):
-                a_tru = phs.convert_string_to_sympy(a_tru, variables)
+                # this is so instructors can specify the true answer simply as a string
```
mwest1066: Want to give an example of correct-answer-as-string in the docs, if we don't have one already?

elements/pl-symbolic-input/pl-symbolic-input.py
```diff
@@ -172,10 +188,17 @@ def parse(element_html, data):
...
 
         # Convert safely to sympy
-        a_sub_parsed = phs.convert_string_to_sympy(a_sub, variables)
+        a_sub_parsed = phs.convert_string_to_sympy(a_sub, variables, allow_complex=allow_complex)
 
-        # Store result as a string.
-        a_sub_string = str(a_sub_parsed)
+        # If complex numbers are not allowed, raise error if expression has the imaginary unit
+        if (not allow_complex) and (a_sub_parsed.has(sympy.I)):
+            s = 'Your answer was simplified to this, which contains a complex number (denoted $i$): $${:s}$$'.format(sympy.latex(a_sub_parsed))
```
mwest1066: If the student inputs a $j$, will it be rendered as $i$ when displayed back to the student? The ECE folks would rather retain $j$ as $j$, I guess.

lib/python_helper_sympy.py
```diff
@@ -204,3 +246,37 @@ def point_to_error(s, ind, w = 5):
...
+    variables = [str(v) for v in a.free_symbols]
+
+    # Check that variables do not conflict with reserved names
+    reserved = {**const.helpers, **const.variables, **const.hidden_variables, **const.functions}
+    if allow_complex:
+        reserved = {**reserved, **const.complex_variables, **const.hidden_complex_variables}
+    for k in reserved.keys():
+        for v in variables:
+            if k == v:
+                raise ValueError('sympy expression has a variable with a reserved name: {:s}'.format(k))
```
mwest1066: Not that I mind the use of `format()`, but I've been starting to convert everything to f-string syntax: https://cito.github.io/blog/f-strings/

elements/pl-symbolic-input/pl-symbolic-input.py
```diff
@@ -45,7 +46,13 @@ def render(element_html, data):
         operators = ', '.join(['cos', 'sin', 'tan', 'exp', 'log', 'sqrt', '( )', '+', '-', '*', '/', '^', '**'])
```
mwest1066: Is `e` a default constant as well? I guess I think it should be if it isn't.

elements/pl-symbolic-input/pl-symbolic-input.py
```diff
@@ -45,7 +46,13 @@ def render(element_html, data):
...
         constants = ', '.join(['pi'])
 
-        info_params = {'format': True, 'variables': variables_string, 'operators': operators, 'constants': constants}
+        info_params = {
+            'format': True,
+            'variables': variables_string,
+            'operators': operators,
+            'constants': constants,
+            'allow_complex': allow_complex,
+        }
```
mwest1066: Do we explicitly check somewhere that the instructor has not asked for `pi` or `e` to be a variable?

ChangeLog.md
tbretl: done

elements/pl-symbolic-input/pl-symbolic-input.py
tbretl: done

elements/pl-symbolic-input/pl-symbolic-input.py
tbretl: done

lib/python_helper_sympy.py
tbretl: good to know. next time.

elements/pl-symbolic-input/pl-symbolic-input.py
tbretl: done

elements/pl-symbolic-input/pl-symbolic-input.py
tbretl: We check for conflicts, yep.

Issue #1276, opened 2018-09-21
pl-string-input Content Hardcoded to "string"


gchallen:
This is a bit confusing. Starting the input blank would probably be a better choice.

![image](https://user-images.githubusercontent.com/445443/45908769-cec3a200-bdc3-11e8-9900-2a06848fed90.png)

nwalters512:
I'm ok with this - maybe we can add an attribute to set the placeholder text too?

mwest1066:
I agree that we should probably default to blank for this particular element, given that there is really no format that the answer has to have. An option to set the placeholder text would also be great!

mfsilva22:
I can submit the PR. Suggestions for the attribute name?

trombonekenny:
placeholder ref: https://www.w3schools.com/TAGS/att_input_placeholder.asp

mwest1066:
Fixed by #1277

PR #1277, opened 2018-09-23
String element update

mfsilva22:
Removes the word "string" as a default placeholder. Creates an attribute `placeholder-text` allowing users to choose their own text. By default, it displays an empty input box.

This PR addresses issue #1276


elements/pl-string-input/pl-string-input.py
```diff
@@ -52,7 +51,7 @@ def render(element_html, data):
             'remove-spaces': remove_spaces,
             'editable': editable,
             'info': info,
-            'shortinfo': shortinfo,
+            'shortinfo': placeholder,
```
mwest1066: Rename the `shortinfo` variable to `placeholder` to make the code easier to follow?

Issue #1114, opened 2018-04-22
[Feature Request] Adding answers to Instructor's Questions page


coatless:
On the question page, it would be nice if we could have a section under the question panel that shows the answers for each part of the question, c.f. LON-CAPA's implementation:

![screen shot 2018-04-22 at 10 16 34 am](https://user-images.githubusercontent.com/833642/39096494-6cade60e-4616-11e8-8eed-49ed97c37bc7.png)

In addition, it may be useful to provide a "script-vars" option that display the `params` array instead of clicking into `Test Once with full details`

mwest1066:
See #1560 for essentially the same thing.

mwest1066:
closing in favor of #1560

PR #1281, opened 2018-09-23
Always display the correct answer for a question in dev, instructor, and admin modes

coatless:
Related to #1114. 

This PR changes when the "correct answer" is displayed from being on input to always being active regardless of whether a user has answered the question in dev / instructor / or admin mode. The hope here is to easy debugging. The next stage is freeing question data from the "issue" ticket generation (c.f. "Script Vars") 

always-enabled

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -129,6 +129,7 @@
 
   * Change `pl-string-input` to include an attribute for the placeholder (Mariana Silva).
 
+  * Change question pages to always include the answer if in instructor, dev, or administrator mode (James Balamuta).
 
 * __3.0.0__ - 2018-05-23
 
--- a/lib/question.js
+++ b/lib/question.js
@@ -1104,6 +1104,9 @@ module.exports = {
             locals.showSaveButton = true;
             locals.allowAnswerEditing = true;
             locals.showNewVariantButton = true;
+
+            // Always display the answer to the instructor
+            locals.showTrueAnswer = true;
         } else {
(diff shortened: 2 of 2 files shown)


trombonekenny:
I understand that in question development, seeing all the metadata about a question is useful. But I think this page could be used for other things, like demoing a question to a student or class (outside of an assessment), and so by-default showing the answer seems harmful.

Playing around with things, any Save or Grade action would show the correct answer with the submission. I'm not sure if that's desirable either, but currently you're just one click away from seeing the answer with the question. 

That's as good as my other branch which adds a card to question pages with instructor info, including the raw answer, but that's a hidden div with a Show/Hide button.

trombonekenny:
@coatless If #1342 goes ahead, do we still need this PR? Does the other approach work for you?

mwest1066:
@trombonekenny the advantage of this PR is that it displays the rendered correct answer box. Is there any way in #1342 to display the correct answer panel in the question itself, in addition to displaying the raw data? I'm thinking about appending a query string (`?show_correct_answer=yes`) or something?

trombonekenny:
I feel like if we're putting it in the instructor panel, that PR makes sense. If we're changing the behavior of `instructorQuestion`, we continue this one?

mwest1066:
I’m closing this because always-display is not what we want when checking questions or simulating the student UX. I opened #1560 as a way to implement this functionality without negatively impacts (although that will be more work).

PR #1292, opened 2018-09-26
PL/PS linked exams tested, fixed, big testing speed improvement

### LINKED ISSUE #987: Mode exam assessment visibility with PS checkin can't handle multiple semesters
If the `courses` table has >1 row with the same `pl_course_id`, PL doesn't show the exam to the students.

That means each semester we need to remove the old `pl_course_id` mappings and add them to the new PS `course_id`.  Then we lose the history and it's just more effort.

It's probably better to modify the pl_course_id check to see if there are any active reservations for any `course_id`s that are linked with the right `pl_course_id`  We're not going to have live reservations in expired `semester_id`s, so this seems pretty safe. It just sounds like the query isn't flexible enough for that.

trombonekenny:
The `check_assessment_access_rule` sproc is the component that evaluates PL exam access against PS reservations. It was broken in two ways:

- PL/PS linking over multiple PS courses (semesters) wasn't working. Each semester I had to remove the old PS `pl_course_id` links and add them to the new courses. (Closes #987)
- Linked exams weren't working as expected for a course that was already PL/PS linked. (No bug on this one?)

This PR:
- adds unit tests for the `check_assessment_access` and `check_assessment_access_rule` sprocs and tests them. They fail in 6 places with the original sproc code.
- Updates the sproc so all tests pass.
- Changes `tests/helperDb` to make a template database first and copy it each time the database needs to be reset. This is much faster than redoing all the migrations between tests. My dev box did the full test suite in about half the time as before.
- (For development) Added an optional `tests/keepdb` file toggle to not drop databases after tests finish (so you can go inspect them)


trombonekenny:
The previous `helperDb` ran the tests on my dev box in 6m19s. Running it with this new one finished in 3m20s. I'm a little sad Travis doesn't see a bigger speed boost, but it looks like it was already doing jobs around the 5m mark ... and it can't be as I/O starved as my old Linux box. :)

.gitignore
```diff
@@ -27,6 +27,7 @@
 /v1/backend/config.json.*
 /config.json
 /aws-config.json
+/tests/keepdb
```
mwest1066: Rather than using a magic file, can we use an environment variable? The `fs.existsSync(keepFile)` can instead be `_.has(process.env, 'PL_KEEP_TEST_DB')` or something like that.

We should also document this in `https://prairielearn.readthedocs.io/en/latest/dev-guide/#unit-tests-and-integration-tests`

ChangeLog.md
```diff
@@ -135,6 +137,7 @@
 
   * Change element documentation to add placeholder attribute to `pl-string-input` (Mariana Silva).
 
+  * Change test cases to use templated DB for faster performance (Dave Mussulman).
```
mwest1066: Shift ChangeLog entries up to the new version area (this sucks, I know -- I'm trying to think of a better way to handle this).

sprocs/check_assessment_access_rule.sql
```diff
@@ -59,86 +59,79 @@ BEGIN
...
+                authorized := FALSE;
+                EXIT schedule_access;
+            END IF;
+
+        ELSE -- no rule.exam_id defined, so search the long way
+
+            -- only needed for exams
+            EXIT schedule_access WHEN assessment_access_rule.mode IS DISTINCT FROM 'Exam';
+
+            -- is there a corresponding PrairieSchedule course?
```
mwest1066: remove spurious `?` at end of line

sprocs/check_assessment_access_rule.sql
```diff
@@ -59,86 +59,79 @@ BEGIN
         ps_course_id bigint;
         reservation reservations;
     BEGIN
-        -- only needed for Exams when we care about the date
-        EXIT schedule_access WHEN assessment_access_rule.mode IS DISTINCT FROM 'Exam';
+        -- reservation checking implies we are given a date to check
```
mwest1066: Do we really want to skip PS checking when `use_date_check = FALSE`? We set `use_date_check = FALSE` when we want to generate the list of access rules to show to students (so they can see what is coming up or was in the past). I think we don't want to show them this information for linked exams?

trombonekenny: That's a carry over from the previous iteration of the function. I might have put that check in the wrong IF stanza when I moved things around.

The student won't see the exam at all unless they're checked in, and then it's just what the "Available Credit" and question mark popover show.

The "examId" rule could be with "startDate" and "endDate", or by itself, so those are the edge cases to think about. If there's no date associated and just the examId, a checked in student for Exam 3 would see:

screen shot 2018-10-11 at 9 50 25 am

If there is a date range, then they'd see the whole date range (which I think you want to avoid):
screen shot 2018-10-11 at 9 52 13 am

If I move the `EXIT WHEN NOT use_date_check` into the ELSE part that only covers non-examId linked exam reservation checking, here's what I see:

If a date range is also config'd:
screen shot 2018-10-11 at 9 46 11 am

Otherwise if it's just an `examId` with no date ranges:
screen shot 2018-10-11 at 9 48 11 am

I'm not sure which is preferable.

.gitignore
trombonekenny: I looked for previous examples in PL of environment variables vs magic files (or config.json settings) but didn't really see anything. I liked the file because it had more visibility and permanence in the test directory and was easy to toggle. I definitely want the functionality; I don't really care what triggers it.

+1 for referencing the docs. Thanks for the pointer.

mwest1066: I'd rather go with an env var in keeping with the `DEBUG=*` style of requesting different behavior from the running app. I like this because you just set it for a single run, so you are less likely to accidentally leave it lying around. In this case I would just do `PL_KEEP_TEST_DB=yes npm test` whenever I wanted to keep the DB.

sprocs/check_assessment_access_rule.sql
trombonekenny: @mwest1066 Thoughts on this?

.gitignore
trombonekenny: Works for me.

sprocs/check_assessment_access_rule.sql
mwest1066: I mainly want to stop students seeing exams that they are not currently checked in for. With the `EXIT schedule_access WHEN NOT use_date_check` I think that a student will see a list all exams that are available to them in the CBTF, but then when they click on an individual exam they will be denied access if they aren't checked in for it. I'd prefer that they never see the exams at all.

Or am I misunderstanding what will happen here?

Just to be clear, I'm proposing that we completely delete the `EXIT schedule_access WHEN NOT use_date_check` line, not just move it around. I know this is a change from what we had before, but I think it's better?

tests/index.js
```diff
@@ -16,3 +21,4 @@ require('./testExam');
 require('./testAccess');
 require('./testZoneGradingHomework');
 require('./testZoneGradingExam');
+require('./testSprocs');
```
mwest1066: I really like the addition of sproc unit tests!

One small request: I think it might be cleaner if we call this file `testAccessSprocs` or something, and then in the future when we add more sproc testing we can put it in separate files. This file is already fairly large and complex.

tests/testSprocs.js
```diff
@@ -0,0 +1,304 @@
...
+            'Exam',
+            'Student',
+            1002,
+            'instructor@school.edu',
+            '2010-07-07 06:06:06-00',
+            'US/Central',
+        ];
+
+        sqldb.call(`check_assessment_access`, params, (err, result) => {
+            if (ERR(err, result)) return;
```
mwest1066: Should be `ERR(err, callback)`. There are more of these below.

tests/testSprocs.js
```diff
@@ -0,0 +1,304 @@
...
+            assert.strictEqual(result.rows[0].authorized, expectedBool);
+            callback();
+        });
+    });
+
+    describe(`with checked-in reservation for student for exam ${exam_id}`, () => {
+
+        before(`create checked-in reservation for student for exam ${exam_id}`, function(callback) {
+            sqldb.query(sql.insert_ps_reservation, {exam_id}, (err, _result) => {
+                if (ERR(err,callback)) return;
```
mwest1066: Very minor nitpick: should be a space after comma.

tests/testSprocs.js
```diff
@@ -0,0 +1,304 @@
...
+                callback();
```
mwest1066: By convention we normally write this as `callback(null)`, which is technically slightly different. I guess mocha doesn't care, but I'd rather stick with the convention anyway.

sprocs/check_assessment_access_rule.sql
trombonekenny: The studentAssessments page does an authz_assessment() call to create the list, so it only shows authorized (i.e. checked-in) exams on the assessment page. So the scope of what `use_date_check` does in that context is just what times are shown when the user already has access.

trombonekenny: What it looks like for the example course, student, exam with the `EXIT schedule_access WHEN NOT use_date_check` line removed:

screen shot 2018-10-18 at 9 26 15 am

PR #1295, opened 2018-09-27
increase default blocked event-loop warning threshold to 200 ms

mwest1066:
We are seeing a lot of log-spam with BLOCKED-AT warnings just slightly over 100 ms. Increasing this to 200 ms should allow us to focus on events that are actually important.


mwest1066:
Actually I'm wondering if merging this would just be hiding the symptoms. I'm now inclined to discard this and fix #1301 instead.

mwest1066:
I'm closing this PR without merging in favor of fixing #1301.

PR #1309, opened 2018-10-05
pl-string change attribute

### LINKED ISSUE #1288: pl-string-input: remove-spaces on by default
After a support case with CHEM 104 not grading correctly a `pl-string-input` because students put an accidental space at the end of the text, I propose the following changes:

- We make remove-spaces `true` by default. If someone needs blank spaces on the perimeter of the string, they can enable it but I think the much more common case is the other way around.
- Include in the input help popup bubble a blurb about this config and its status. i.e. `Leading and trailing spaces will be allowed.`
- Could we do some `<span>` highlighting in the submission and correct answer displays of the element that would better show the trailing white space?
- Maybe clean up the rest of the student facing interface for this to something kinder to the CS layperson than String? Use the word text instead? I'm not suggesting we need to change the element name, but just how it's referenced in the student UI.

mfsilva22:
This PR makes `remove-leading-trailing=true` by default, adds hint to popover message, and modifies some of the user-interface text to use the word "text" instead of "string". Submitted answers show with highlighted gray background, making blank spaces visible.

Closes #1288


ChangeLog.md
```diff
@@ -137,6 +137,12 @@
 
   * Change element documentation to add placeholder attribute to `pl-string-input` (Mariana Silva).
 
+  * Change instructor assessment page into multiple sub-pages (Matt West).
```
tbretl: Fix the changelog (this entry and the next one don't belong in this PR).

elements/pl-string-input/pl-string-input.mustache
```diff
@@ -48,7 +48,7 @@
 {{/parse_error}}
 {{^parse_error}}
 {{#label}}<span>{{label}}</span>{{/label}}
-<samp>{{a_sub}}</samp>
+<samp style="white-space: pre; background-color: #e8e8e8;">{{a_sub}}</samp>
```
tbretl: It's best to avoid inline CSS styling, if possible.

1) Is the `white-space` style necessary?
2) Could you replace the `background-color` style with `class="bg-something"` as in https://getbootstrap.com/docs/4.0/utilities/colors/?

elements/pl-string-input/pl-string-input.mustache
```diff
@@ -65,6 +65,7 @@
 
 {{#format}}
 <p>
-Your answer must be a string. Any symbolic expressions or numbers will be intepreted as a string.
+Your answer must be a text. Any symbolic expressions or numbers will be intepreted as a text (string type).
```
tbretl: In these two sentences, I think you mean just "text" and not "a text"? (Perhaps "plain text"?) Also, I would delete ` (string type)`.

elements/pl-string-input/pl-string-input.py
```diff
@@ -27,16 +27,21 @@ def render(element_html, data):
...
     placeholder = pl.get_string_attrib(element, 'placeholder', None)
 
     if data['panel'] == 'question':
         editable = data['editable']
         raw_submitted_answer = data['raw_submitted_answers'].get(name, None)
 
+        if remove_leading_trailing:
+            trailing_hint = 'Leading and trailing spaces will be removed from your answer.'
+        else:
+            trailing_hint = 'Leading and trailing spaces will be allowed (included as part of your answer).'
```
tbretl: Perhaps: `will be left as part of your answer.`

elements/pl-string-input/pl-string-input.py
```diff
@@ -27,16 +27,21 @@ def render(element_html, data):
     label = pl.get_string_attrib(element, 'label', None)
     suffix = pl.get_string_attrib(element, 'suffix', None)
     display = pl.get_string_attrib(element, 'display', 'inline')
-    remove_leading_trailing = pl.get_string_attrib(element, 'remove-leading-trailing', False)
-    remove_spaces = pl.get_string_attrib(element, 'remove-spaces', False)
+    remove_leading_trailing = pl.get_boolean_attrib(element, 'remove-leading-trailing', True)
+    remove_spaces = pl.get_boolean_attrib(element, 'remove-spaces', False)
```
tbretl: Should there also be help text that describes what `remove_spaces` will do, if this flag is `True`?

elements/pl-string-input/pl-string-input.py
```diff
@@ -140,7 +145,7 @@ def parse(element_html, data):
     element = lxml.html.fragment_fromstring(element_html)
     name = pl.get_string_attrib(element, 'answers-name')
     # Get allow-blank option
-    allow_blank = pl.get_string_attrib(element, 'allow-blank', False)
+    allow_blank = pl.get_boolean_attrib(element, 'allow-blank', False)
```
tbretl: It might be a good idea to revisit the `allow_blank` option. The code here is problematic:
https://github.com/PrairieLearn/PrairieLearn/blob/2df88e38ed39c4026253afe8ae639d7a76ec5ee2/elements/pl-string-input/pl-string-input.py#L157

Consider two different student actions:
1) Open question, click grade.
2) Open question, enter text, delete all of this text, click grade.

In both cases, the input box will be blank, i.e., empty.

If there is a difference between these cases, then it's confusing for the user, because they can't tell the difference between empty and empty-after-deleting-text.

If there is no difference between these cases, then we will never arrive at the `if not a_sub and not allow_blank` line, because we will have found that `a_sub is None`.

elements/pl-string-input/pl-string-input.mustache
mfsilva22: The `whits-space: pre` is needed, otherwise by default, it uses `white-space: normal`, where "Sequences of whitespace will collapse into a single whitespace".

I don't particularly like the color options from the link you indicated. They are mostly to dark or distracting. `bg-light` is the only one I would choose, but I think that is a little too light, so I still prefer the `background-color` suggested by Matt.

elements/pl-string-input/pl-string-input.py
mwest1066: @tbretl: In talking with @mfsilva22 in office hours, I think it wasn't entirely clear what the issue was here. Is it that we should be treating a null or missing value the same as a zero-length string value, but we are failing to do that?

elements/pl-string-input/pl-string-input.py
mfsilva22: done

elements/pl-string-input/pl-string-input.py
mfsilva22: done

elements/pl-string-input/pl-string-input.mustache
mfsilva22: done

elements/pl-string-input/pl-string-input.py
mfsilva22: I never really understood the need of the allow_blank option... I cant imagine a situation where it will be needed. So initially I thought it should be just considered invalid, but I was asked to implement this option. So if we are to keep the option, can you please suggest how you would like to address the issue above?

elements/pl-string-input/pl-string-input.mustache
mfsilva22: okay, I will add this soon and push again

elements/pl-string-input/pl-string-input.py
tbretl: @mfsilva22 Sorry for the confusion!

If I understand correctly, `allow_blank` means that the empty string is an acceptable answer. One reasonable way for a student to input the empty string is to input nothing - that is, to leave the input box blank and just click `Grade`. When the student does this, `data[‘submitted_answers’][name]` will have the value `None`, and the submission will be declared invalid.

If that makes sense (?), then I would do one of three things:

1. Eliminate the `allow_blank` option.
2. Change `if a_sub is None` to `if a_sub is None and not allow_blank`.
3. Add a button (e.g., in the input group between the input box and the question mark) that says something like “submit empty string” or “blank” and that, if pressed, will disable the input box and will (in `parse`) cause the submitted answer to be interpreted as blank.

I suggest option 1. I would also be on board with something like option 3.

I don’t like option 2, since it conflicts with our general philosophy of trying to avoid “I accidentally clicked `Grade`” situations.

tbretl:
I have one more suggestion. These lines currently appear in `grade`:

```
    # Remove the leading and trailing characters
    if (remove_leading_trailing):
        a_sub = a_sub.strip()

    # Remove the blank spaces between characters
    if (remove_spaces):
        a_sub = a_sub.replace(' ', '')
```

It may be best to move these lines to `parse`. The philosophy here is that `parse` takes the raw submission and puts it in a form that can be compared to the correct answer. This would (I think) be consistent with other elements, for example the various matrix/number input elements that substitute a minus sign for a “calculator hyphen” and so forth in `parse`. What do you think?

elements/pl-string-input/pl-string-input.py
mwest1066: I'm fine with 1.

mwest1066:
I agree with moving the stripping to `parse()`.

mwest1066:
@tbretl this PR still has changes requested from you. Can you check whether they are all done and sign off on them if so? Thanks!

nicknytko:
This should be closed in favour of #2065?

mwest1066:
All changes (except the default switch) were implemented in #2065

PR #1320, opened 2018-10-10
fix load closing for unit tests; close #1284

### LINKED ISSUE #1284: Load reporting errors during testing
I noticed load reporting errors on the console while running tests. Since the server and database change state so much during tests, maybe we should disable background tasks during testing?

mwest1066:


mwest1066:
Good idea to test together with #1292. I just did it and the merge of these two worked fine, so I think this really fixed the problem (it was a race condition to do with the creation of the load reporters).

PR #1330, opened 2018-10-14
Add option to disable help text text on pl-number-input and add size attribute

nicknytko:
Fixes issue #1118.  Rebased James' PR (#1123) and updated to work on latest codebase.


exampleCourse/courseInstances/Sp15/assessments/plelementexamples/infoAssessment.json
```diff
@@ -33,11 +33,12 @@
     "zones": [
         {
             "questions": [
-				{"id": "examplesCheckbox",           "points": 2, "maxPoints": 10},
-				{"id": "examplesStringInput",           "points": 2, "maxPoints": 10},
-				{"id": "elementsMatrixLatexDisplay",        "points": 2, "maxPoints": 10},
-                {"id": "examplesMatrixInput",        "points": 2, "maxPoints": 10},
-                {"id": "examplesSymbolicInput",        "points": 2, "maxPoints": 10}
+		{"id": "examplesCheckbox", "points": 2, "maxPoints": 10},
```
tbretl: Maybe retain the original indents, so only the new line is part of this commit?

elements/pl-number-input/pl-number-input.mustache
```diff
@@ -15,16 +15,25 @@
...
                 <i class="fa fa-question-circle" aria-hidden="true"></i>
                 {{#correct}}&nbsp;<span class="badge badge-success"><i class="fa fa-check" aria-hidden="true"></i> 100%</span>{{/correct}}
                 {{#partial}}&nbsp;<span class="badge badge-warning"><i class="fa fa-circle-o" aria-hidden="true"></i> {{partial}}%</span>{{/partial}}
                 {{#incorrect}}&nbsp;<span class="badge badge-danger"><i class="fa fa-times" aria-hidden="true"></i> 0%</span>{{/incorrect}}
             </a>
+            {{/shortinfo}}
         </span>
+        {{/display_append_span}}
+        {{^shortinfo}}
+            {{#correct}}&nbsp;<span class="badge badge-success pl-number-input-inline-badge"><i class="fa fa-check" aria-hidden="true"></i> 100%</span>{{/correct}}
```
tbretl: These lines duplicate what’s above. It would be nice to have them appear only once. I think you can change the template a bit to make this work?

elements/pl-number-input/pl-number-input.css
```diff
@@ -2,3 +2,14 @@
...
+.pl-number-input-inline-badge {
+    position: relative;
+    top: 9px;
+    height: 17px;
+}
+
+.pl-number-input-force-radius-border {
+    border-top-right-radius: 0.25em !important;
+    border-bottom-right-radius: 0.25em !important;
+}
```
tbretl: These two CSS styles (`pl-number-input-inline-badge` and `pl-number-input-force-radius-border`) seem really specific, and brittle with respect to things like future bootstrap updates. Are they necessary?

elements/pl-number-input/pl-number-input.mustache
```diff
@@ -15,16 +15,25 @@
             <span class="input-group-text">{{label}}</span>
         </span>
         {{/label}}
-        <input name="{{name}}" type="text" class="form-control" size="35" {{^editable}}disabled{{/editable}} {{#raw_submitted_answer}}value="{{raw_submitted_answer}}"{{/raw_submitted_answer}} aria-describedby="basic-addon1" placeholder="{{shortinfo}}" />
+        <input name="{{name}}" type="text" class="form-control {{^display_append_span}}pl-number-input-force-radius-border{{/display_append_span}}" size="{{size}}" {{^editable}}disabled{{/editable}} {{#raw_submitted_answer}}value="{{raw_submitted_answer}}"{{/raw_submitted_answer}} aria-describedby="basic-addon1" placeholder="{{shortinfo}}" />
+        {{#display_append_span}}
         <span class="input-group-append">
-            {{#suffix}}<span class="input-group-text">{{suffix}}</span>{{/suffix}}
+            {{#suffix}}<span class="input-group-text {{^shortinfo}}pl-number-input-force-radius-border{{/shortinfo}}">{{suffix}}</span>{{/suffix}}
+            {{#shortinfo}}
```
tbretl: Just to confirm - is there consensus around the decision to hide the question mark along with the placeholder text on `hide-help-text`?

@mfsilva22 ?

coatless: 100% agree there should be some sort of standardization for these options. 

Right now, `hide-help-text` disables _all_ help prompts on `pl-checkbox` (c.f. https://prairielearn.readthedocs.io/en/latest/elements/#pl-checkbox-element). I added this option due to the need to differentiate between Multiple Choice (MC) and Multiple Selection (MS) questions as students didn't detect subtle differences between checkboxes and radio buttons.

nicknytko: Following everyone's advice I've set `hide-help-text` to disable the question mark and `placeholder` to disable the placeholder text inside the input box.

exampleCourse/questions/examplesNumberInput/question.html
```diff
@@ -0,0 +1,81 @@
+<div class="card my-2">
+  <div class="card-body">
+    Questions below were designed to showcase the different features of
```
nwalters512: Extremely minor nit - move this intro outside of the first card

exampleCourse/questions/examplesNumberInput/question.html
```diff
@@ -0,0 +1,81 @@
...
+  </div>
+</div>
+
+<div class="card my-2">
+  <div class="card-body">
+    <pl-question-panel>
+      <p>This question modifies the default relative tolerance (of <code>1e-3</code>) and absolute tolerance (of <code>1e-8</code>) underneath <a href="http://prairielearn.readthedocs.io/en/latest/elements/#pl-number-input-element"><code>pl-number-input</code></a> element. Furthermore, we add a label to the front of the input field to specify what the field should receive. </p>
+      <p> Solve for the value of $c$ given $c^2 = {{params.a}}^2 + {{params.b}}^2$ </p>
+    </pl-question-panel>
+    <pl-number-input answers_name="c_2" label = "Enter the Value for $c$:" rtol = "1e-3" atol="1e-10"/>
```
nwalters512: The more canonical label from the rest of examples would be `$c = $`

exampleCourse/questions/examplesNumberInput/question.html
```diff
@@ -0,0 +1,81 @@
...
+    <a href="http://prairielearn.readthedocs.io/en/latest/elements/#pl-number-input-element"><code>pl-number-input</code></a>. 
+    The goal of <a href="http://prairielearn.readthedocs.io/en/latest/elements/#pl-number-input-element"><code>pl-number-input</code></a>
+    is to allow users to enter <b>real or complex values</b> to be checked within a specific tolerance.
+  </div>
+
+  <div class="card-body">
+    <pl-question-panel>
+      <p>Within this question, you can observe the default behavior of the <a href="http://prairielearn.readthedocs.io/en/latest/elements/#pl-number-input-element"><code>pl-number-input</code></a> element.
+        In particular, the default input field has help text as a placeholder as well as a question button and is set to use <code>"relabs"</code> to check the tolerance, which is a combination of relative (<code>"rtol"</code>) and absolute (<code>"atol"</code>) tolerances.</p>
+      <p> Solve for the value of $c$ given $c^2 = {{params.a}}^2 + {{params.b}}^2$ </p>
```
nwalters512: Might be nice to mention the solution to this equation at the very start of this question so that lazy people (like me) can quickly get a sense of what the right answer is so they can observe how the input behaves with a correct answer.

exampleCourse/courseInstances/Sp15/assessments/plelementexamples/infoAssessment.json
```diff
@@ -35,9 +35,10 @@
             "questions": [
 				{"id": "examplesCheckbox",           "points": 2, "maxPoints": 10},
 				{"id": "examplesStringInput",           "points": 2, "maxPoints": 10},
-				{"id": "elementsMatrixLatexDisplay",        "points": 2, "maxPoints": 10},
+		{"id": "elementsMatrixLatexDisplay",        "points": 2, "maxPoints": 10},
                 {"id": "examplesMatrixInput",        "points": 2, "maxPoints": 10},
-                {"id": "examplesSymbolicInput",        "points": 2, "maxPoints": 10}
+                {"id": "examplesSymbolicInput",        "points": 2, "maxPoints": 10},
```
nwalters512: Can you tidy this chunk up to use spaces (the first 2 questions) and a consistent indent?

doc/elements.md
```diff
@@ -88,6 +88,10 @@ Attribute | Type | Default | Description
 `atol` | number | 1e-8 | Absolute tolerance for `comparison="relabs"`.
 `digits` | integer | 2 | number of digits that must be correct for `comparison="sigfig"` or `comparison="decdig"`.
 `allow-complex` | boolean | False | Whether or not to allow complex numbers as answers. If the correct answer `ans` is a complex object, you should use `import prairielearn as pl` and `data['correct_answer'][answers-name] = pl.to_json(ans)`.
+`hide-help-text` | boolean | False | Hide question mark at the end of the input displaying required input parameters.
+`placeholder` | boolean | True | Show the placeholder text that shows the default comparison.
```
nwalters512: I think I'd prefer `show-placeholder` to go along with the `[verb]-[noun]` naming of other boolean attributes ("allow complex", "hide-help-text", etc.)

doc/elements.md
```diff
@@ -88,6 +88,10 @@ Attribute | Type | Default | Description
...
+`size` | integer | 35 | Size of the input box.
+`correct-answer-feedback` | boolean | False | Whether to show the correct answer in the submitted answers panel.
```
nwalters512: Ditto with naming - maybe `show-correct-answer`?

doc/elements.md
```diff
@@ -88,6 +88,10 @@ Attribute | Type | Default | Description
 `atol` | number | 1e-8 | Absolute tolerance for `comparison="relabs"`.
 `digits` | integer | 2 | number of digits that must be correct for `comparison="sigfig"` or `comparison="decdig"`.
 `allow-complex` | boolean | False | Whether or not to allow complex numbers as answers. If the correct answer `ans` is a complex object, you should use `import prairielearn as pl` and `data['correct_answer'][answers-name] = pl.to_json(ans)`.
+`hide-help-text` | boolean | False | Hide question mark at the end of the input displaying required input parameters.
```
nwalters512: Naming: maybe `show-help-text` for consistency with other potential boolean attribute names? Not sure if I like naming with affirmative or positive verbs :/ @mwest1066 any thoughts?

exampleCourse/questions/examplesNumberInput/question.html
mfsilva22: I completely agree with you! I didn't ask Nicolas to modify this because I was not the one that created this sample question. If there is a consent here, I agree that we should change all questions to have the label `$c = $`

exampleCourse/questions/examplesNumberInput/question.html
mfsilva22: I can modify the sample question to be something easier (sorry, I thought the above was simple enough). What about ask to enter the dimensions of a matrix?

doc/elements.md
mfsilva22: if we change this to `show-placeholder`, should we also change the attribute in `pl-string-input`? I think we should keep the element attributes as consistent as possible

exampleCourse/questions/examplesNumberInput/question.html
nwalters512: It's not too hard as is! I was just like, ever so slightly too lazy to actually get out my calculator :) just mentioning at the top of the question "if you don't feel like getting out your calculator, the correct answer is 10.xxxxxx" would be enough for me

doc/elements.md
mfsilva22: I like that better too. Nicolas, can you change that please?

doc/elements.md
mfsilva22: Again, I think we should keep consistency. `pl-checkbox` has `hide-help-text`. So if we change here, I think we should change there as well.

doc/elements.md
nwalters512: The difference here is that `placeholder` for `pl-string-input` actually specifies placeholder text, whereas this one is a boolean on/off switch. I think it's actually more consistent to name them different things.

doc/elements.md
nwalters512: Agreed, though we don't have the best migration/deprecation plan for elements and attributes. I'd be happy to rename the attribute to the "right" thing on `pl-checkbox` and update the docs but support both versions in code indefinitely.

doc/elements.md
mfsilva22: oh yes.. you are correct! makes sense to change to `show-placeholder` :-)

exampleCourse/questions/examplesNumberInput/question.html
```diff
@@ -0,0 +1,83 @@
...
+  </div>
+</div>
+
+<div class="card my-2">
+  <div class="card-body">
+    <pl-question-panel>
+      <p>This question modifies the default relative tolerance (of <code>1e-3</code>) and absolute tolerance (of <code>1e-8</code>) underneath <a href="http://prairielearn.readthedocs.io/en/latest/elements/#pl-number-input-element"><code>pl-number-input</code></a> element. Furthermore, we add a label to the front of the input field to specify what the field should receive. </p>
+      <p> Solve for the value of $c$ given $c^2 = {{params.a}}^2 + {{params.b}}^2$ </p>
+    </pl-question-panel>
+    <pl-number-input answers_name="c_2" label = "$c = $" rtol = "1e-3" atol="1e-10"/>
```
nwalters512: Missed this in the first round though, but we should update all attributes to used the dashed version (e.g. `answers_name`). @nicknytko definitely realize you didn't write this originally but we should correct this before introducing this to master.

mfsilva22: I also noticed that some of the elements are not "properly closed" in this example. 

`<pl-number-input ....></pl-number-input>`

@nicknytko can you also make these changes please?

elements/pl-number-input/pl-number-input.py
```diff
@@ -124,6 +152,13 @@ def render(element_html, data):
             if raw_submitted_answer is not None:
                 html_params['raw_submitted_answer'] = escape(raw_submitted_answer)
 
+        # Add true answer to be able to display it in the submitted answer panel
+        ans_true = None
+        if (pl.get_boolean_attrib(element, 'show-correct-answer', False)):
+            ans_true = format_true_ans(element, data, name)
+        if (ans_true is not None):
```
mwest1066: Python style nitpick: no extra parentheses around `if` statement conditions (also above)

elements/pl-number-input/pl-number-input.mustache
```diff
@@ -50,7 +64,27 @@
...
+            template: '<div class="popover pl-number-input-popover" role="tooltip"><div class="arrow"></div><h3 class="popover-header"></h3><div class="popover-body"></div></div>',
+        });
+    });
+</script>
+<span id="pl-number-input-correct-{{uuid}}">
+  {{^a_tru}}
+    <span class="badge badge-success"><i class="fa fa-check" aria-hidden="true"></i> 100%</span>
+  {{/a_tru}}
+  {{#a_tru}}
+    <a class="btn btn-light border" role="button" data-toggle="popover" data-html="true" title="Correct Answer" data-content="The correct answer used for grading is <samp>{{a_tru}}</samp>" data-placement="auto" data-trigger="focus" tabindex="0">
```
mwest1066: Can we change "The correct answer used for grading is" to "The exact correct answer is"? I think this would be clearer?

elements/pl-number-input/pl-number-input.mustache
```diff
@@ -15,16 +15,30 @@
...
-                {{#partial}}&nbsp;<span class="badge badge-warning"><i class="fa fa-circle-o" aria-hidden="true"></i> {{partial}}%</span>{{/partial}}
-                {{#incorrect}}&nbsp;<span class="badge badge-danger"><i class="fa fa-times" aria-hidden="true"></i> 0%</span>{{/incorrect}}
-            </a>
-        </span>
+        <input name="{{name}}" type="text" class="form-control {{^display_append_span}}rounded-right{{/display_append_span}}" size="{{size}}" {{^editable}}disabled{{/editable}} {{#raw_submitted_answer}}value="{{raw_submitted_answer}}"{{/raw_submitted_answer}} aria-describedby="basic-addon1" placeholder="{{shortinfo}}" />
+          
+          <span class="input-group-append">
+            {{#suffix}}<span class="input-group-text {{^questionmark}}rounded-right{{/questionmark}}">{{suffix}}</span>{{/suffix}}
+            {{#questionmark}}
+              <a class="btn btn-light border" role="button" data-toggle="popover" data-html="true" title="Number" data-content="{{info}}" data-placement="auto" data-trigger="focus" tabindex="0">
```
mwest1066: If `show-correct-answer == True` and the element has `#correct` then can we add the display of `{{a_tru}}` to the end of the `{{info}}`? That is, make the popover contain both the current `{{info}}` and also the identical text that we display for correct answers in the submission panel, with the same logic? I think this would help to avoid confusion for students by putting the correct answer information in both places.

exampleCourse/courseInstances/Sp15/assessments/plelementexamples/infoAssessment.json
```diff
@@ -33,11 +33,12 @@
     "zones": [
         {
             "questions": [
-				{"id": "examplesCheckbox",           "points": 2, "maxPoints": 10},
-				{"id": "examplesStringInput",           "points": 2, "maxPoints": 10},
-				{"id": "elementsMatrixLatexDisplay",        "points": 2, "maxPoints": 10},
+                {"id": "examplesCheckbox",           "points": 2, "maxPoints": 10},
+                {"id": "examplesStringInput",        "points": 2, "maxPoints": 10},
+                {"id": "elementsMatrixLatexDisplay", "points": 2, "maxPoints": 10},
```
mwest1066: no more spaces! yay!

ChangeLog.md
```diff
@@ -3,12 +3,20 @@
...
 
   * Add course instance admin subpages (Dave Mussulman).
 
   * Add networkx Python library to plbase (Nathan Bowman).
 
+  * Add option to toggle placeholder help text for `pl-number-input` (James Balamuta).
+
+  * Add `size` attribute to `pl-number-input` (Nicolas Nytko).
+
+  * Add demo question showcasing all options for `pl-number-input` (James Balamuta).
```
mwest1066: @nicknytko please add yourself to all these ChangeLog entries -- we want to credit you! You can say "James Balamuta and Nicolas Nytko" if you want.

doc/elements.md
```diff
@@ -88,6 +88,10 @@ Attribute | Type | Default | Description
...
+`show-placeholder` | boolean | True | Show the placeholder text that shows the default comparison.
+`size` | integer | 35 | Size of the input box.
+`show-correct-answer` | boolean | False | Whether to show the correct answer in the submitted answers panel.
```
mwest1066: I think we should default this to `True`. It's a change from the current behavior, but I can't really think of a situation where we wouldn't want to display it. Can anyone see any objection to default True?

nwalters512: Agreed that this is a very sensible default.

elements/pl-number-input/pl-number-input.mustache
eliving2: I think this wording was chosen because sometimes multiple solution methods can work/will be within tolerance. That being said, I don't know if that is too specific to TAM 251.

mwest1066: That's a fair point. I'm happy to leave the text as it is.

doc/elements.md
mfsilva22: tam 251 people will be happy with that :-)

elements/pl-number-input/pl-number-input.mustache
nicknytko: Sure, but what do you suggest for when the popup is disabled?  If `show-correct-answer` is going to default to True then it may make sense to hide it as well.  Just wondering what people's thoughts on this were.

mwest1066: One straightforward option would be to enable the popover if `show-correct-answer` is True, but don't put the `{{info}}` string in it in that case, but rather only the true answer (so it would work like the popover in the submission panel).

That is, have two conditions for showing the popover next to the input element. First, if `show-help-text` is True then the popover displayed and contains `{{info}}`. Second, if `show-correct-answer` is True and we should be displaying the correct answer then also display the popover, containing the true answer. If both conditions hold then display the popover with both the info text and the true answer.

Does this system make sense? Any corner cases I'm forgetting?

PR #1331, opened 2018-10-14
LTI html fixup part 2

trombonekenny:


mwest1066:
Did something equivalent with #1332.

Issue #1128, opened 2018-04-24
Non-feedback Exam Mode


coatless:
Per a [reddit](https://www.reddit.com/r/UIUC/comments/8ecp7t/cbtf_tests/dxva3f9/) comment on the CBTF:

> More like on written exams you aren’t given the chance to slowly watch your grade drop with each wrong attempt

I think it may be advantageous to setup a mode where correct/incorrect responses are not shown if there is only one attempt allotted. This allows for the matching of a traditional paper-based exam and may influence the effects of slipping (Wrong Answer | Know Material) and guessing (Right Answer | Don't Know Material) on a given exam.

mwest1066:
Fixed by @trombonekenny in 4b88ed0ea7a5017746fc807cb0016efc566a35e2

PR #1336, opened 2018-10-16
Grade all button

### LINKED ISSUE #1087: Bring the "Grade all" button back
Would it be possible to have a "grade all" button again in Exams/quizzes?

trombonekenny:
- Change question button text to "Save & Grade" and "Save only" to be clearer on what they do.
- Move buttons closer to each other so one or the other isn't ignored.

screen shot 2018-10-16 at 9 15 29 am

- Adds "Grade all saved answers" button to exam assessment overview (closes #1087).

screen shot 2018-10-16 at 9 16 15 am

Telling the student to save all answers and grade at the end addresses #1128.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -11,6 +11,10 @@
 
   * Add read-only API for instructors to access assessment data (Nathan Walters).
 
+  * Add ability to "Grade all saved answers" on exam assessment overview (Dave Mussulman).
+
+  * Change "Save & Grade" button text and alignment (Dave Mussulman).
+
   * Fix load-reporting close during unit tests (Matt West).
 
   * Fix responsiveness and centering of images displayed with `pl-figure` (James Balamuta, h/t Dave Mussulman).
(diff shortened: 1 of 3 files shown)


trombonekenny:
Button inline with the instructions doesn't bother me, but I agree it can look a little weird. I tried to use bullets to break it up a little better.

My experience is students don't really read the block of text. But they do click random buttons....

nwalters512:
Should probably update the text in the "Submit your answers..." bullet to reflect the new button names. Agree that button inline with bulleted text looks a little strange.

trombonekenny:
Incorporating feedback:
image

nwalters512:
I think that looks better :+1:

mwest1066:
This is great! Two small suggestions:

1. We could check whether any instance question has `status = 'saved'` and only display the "Grade all saved answers" button in this case. If there aren't any saved questions then we could still display the button, but in a [disabled state](http://getbootstrap.com/docs/4.1/components/buttons/#disabled-state) and with the button text changed to "No ungraded answers" or something like that.

2. We could add some text to the final bullet point, something like: "If you have any ungraded saved answers when you leave, they will be automatically graded before your final score is computed."

trombonekenny:
huh Travis disappeared off this PR?

PR #1338, opened 2018-10-16
Fix GitHub links to exampleCourse questions

trombonekenny:
When you visited the exampleCourse (in a container, probably on the live site too but no one has an instructor view there) the "view in GitHub" links were missing.

This fixes that part in the instructorQuestion display and adds a new infoCourse.json option `isExampleCourse`

### DIFF
--- a/exampleCourse/infoCourse.json
+++ b/exampleCourse/infoCourse.json
@@ -3,7 +3,8 @@
     "name": "XC 101",
     "title": "Example Course",
     "options": {
-        "useNewQuestionRenderer": true
+        "useNewQuestionRenderer": true,
+        "isExampleCourse": true
     },
     "assessmentSets": [],
     "topics": [
--- a/exampleCourse/infoCourse.json
+++ b/exampleCourse/infoCourse.json
@@ -3,8 +3,7 @@
     "name": "XC 101",
     "title": "Example Course",
     "options": {
-        "useNewQuestionRenderer": true,
-        "isExampleCourse": true
+        "useNewQuestionRenderer": true
     },
     "assessmentSets": [],
     "topics": [
--- a/lib/course-db.js
+++ b/lib/course-db.js
@@ -78,9 +78,14 @@ module.exports.loadCourseInfo = function(courseInfo, courseDir, logger, callback
         });
 
         // Course options
-        courseInfo.options = info.options || {};
-        // Set default course options
+        courseInfo.options = {};
         courseInfo.options.useNewQuestionRenderer = _.get(info, 'options.useNewQuestionRenderer', false);
+        courseInfo.options.isExampleCourse = false;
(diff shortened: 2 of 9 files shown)


lib/course-db.js
```diff
@@ -78,7 +78,8 @@ module.exports.loadCourseInfo = function(courseInfo, courseDir, logger, callback
         });
 
         // Course options
-        courseInfo.options = {};
+        courseInfo.options = info.options || {};
+        // Set default course options
```
mwest1066: I think that it's probably cleaner here to stick with @nwalters512's existing pattern of starting with `{}` and then explicitly copying over each option that we know about, with a default explicitly set.

exampleCourse/infoCourse.json
```diff
@@ -3,7 +3,8 @@
     "name": "XC 101",
     "title": "Example Course",
     "options": {
-        "useNewQuestionRenderer": true
+        "useNewQuestionRenderer": true,
+        "isExampleCourse": true
```
mwest1066: I'm worried that people are going to copy this. What about an evil hack that directly detects the example course? For example, when loading the course we could check the `uuid`, `name`, and `title` and set `isExampleCourse` if they all match. Then we won't need an option in the JSON file and no one will accidentally copy it. There is a slight risk if someone copies the entire example course, including the UUID, but that will break anyway sooner or later when we have a UUID conflict with the real example course.

trombonekenny: I agree there's a risk the file would get copied. It just seemed cleaner, more controllable with a variable than trying to make decisions based on other contents. I'll work on it and see what we can do.

lib/course-db.js
trombonekenny: It seems like wasteful duplication to define a strict schema, and then duplicate that checking and validation with a different library on load. I get that adding missing default valuables is important, and missing in our json validator-- but if that existed, would we need to be so strict here?

exampleCourse/infoCourse.json
mwest1066: I agree that detecting exampleCourse directly is a hack, but it's only this one case that we'll ever have to worry about, so I don't mind so much. I really want to avoid confusion with people copying things from the example course.

lib/course-db.js
mwest1066: I agree that if we had really strict JSON schema checking then we wouldn't need to do it here. Until that time I'd rather be strict.

exampleCourse/infoCourse.json
trombonekenny: Do you have a preference about doing the dirty hack at some higher level (`lib-coursedb`) or putting it in `instructorQuestion` where it's used?

mwest1066: I have a slight preference for putting it at a higher level. I suspect that @tbretl's editor patches might also want to handle the example course specially?

PR #1341, opened 2018-10-19
Correct answer feedback

bmercer486:
Adds popovers next to correct submitted answers in the Submitted Answers panel for `pl-number-input` submissions. The popovers display the exact numerical answer. This feedback helps students understand how close their input answer is to "true" and can help avoid slightly out-of-tolerance answer submissions that are due to error propagation in multi-part submissions. No such feedback is given for incorrect submissions.

feedback_example


mwest1066:
Merged as part of #1330

PR #1355, opened 2018-10-26
Change assessment access rule examId linking to examUuid

trombonekenny:
This PR changes the PS exam-PL assessment linking from using the `exam.id` to a new `exam.uuid`.

Any allowAccess rules that used examId will need to be updated, will no longer work, and the config will fail schema checks on sync. That appears to just be the CS125 proficiency exam so no big deal there.

Exam uuid generation and display on the PS side still needs to be done.


mwest1066:
We talked about this on Slack, but I think it would be best to NOT have `assessement_access_rules.exam_uuid` be a foreign key to `exams.uuid`.

This would mean that we can delete the `config.syncExamIdAccessRules` option and remove the code that makes fake `exams` on sync.

Instead, we could add an option like `config.checkAccessRulesExamUuid` that would cause an error to be thrown if an invalid `exam_uuid` is encountered during sync. This is pretty harsh (breaking sync), but I think it's appropriate in this case.

trombonekenny:
We'll want to set the migration number lower before merging, but this is almost to that point?

migrations/900_assessment_access_rules__exam_uuid__add.sql
```diff
@@ -1,2 +1,2 @@
 ALTER TABLE exams ADD COLUMN IF NOT EXISTS uuid uuid NULL UNIQUE;
-ALTER TABLE assessment_access_rules ADD COLUMN exam_uuid uuid REFERENCES exams(uuid) ON UPDATE CASCADE ON DELETE SET NULL;
+ALTER TABLE assessment_access_rules ADD COLUMN exam_uuid uuid NULL;
```
mwest1066: Not that it matters, but we don't need the NULL on the two lines above, right?

mwest1066:
Seems like this is ready to go (modulo migration renumbering and maybe removing the unnecessary NULL specifications). Let me know when it's ready to merge.

trombonekenny:
gwa emailed about being the only assessment access rule that's using the old examId linking in production. He's got the info it would take to fix that.

If we want to enable the checking in production, we need to add `checkAccessRulesExamUuid: true` in `config.json`. It defaults to false.

I think this one is ready to ship.

migrations/133_assessment_access_rules__exam_uuid__add.sql
```diff
@@ -1,2 +1,4 @@
 ALTER TABLE exams ADD COLUMN IF NOT EXISTS uuid uuid UNIQUE;
-ALTER TABLE assessment_access_rules ADD COLUMN exam_uuid uuid;
+ALTER TABLE assessment_access_rules
+    ADD COLUMN exam_uuid uuid,
+    DROP COLUMN exam_id;
```
mwest1066: I'm afraid that you're going to need to also delete the column from `database/tables/assessment_access_rules.pg` to keep Travis happy.

PR #1357, opened 2018-10-27
Add section on common pl errors to FAQ

coatless:
Per slack discussion...

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -1,7 +1,18 @@
 
 # FAQ (Frequently Asked Questions)
 
-## How do I give a student access to homeworks or an exam after the semester is over?
+Topics
+
+- [Assessments](#assessments)
+- [Common Errors](#common-errors)
+
+Have a question or issue that wasn't listed here but you think it should be? 
+
+Consider **[adding the question or issue](https://github.com/PrairieLearn/PrairieLearn/edit/master/doc/faq.md)** to the FAQ.
+
+## Assessments
+
+### How do I give a student access to homeworks or an exam after the semester is over?
 
 You need to give the student access to both the course instance itself for the completed semester as well as the specific assessments within that course instance.
 
@@ -48,13 +59,9 @@ See [Access control](accessControl.md) for more details.
 Within this section of the FAQ, we will go over different errors that may occur
 when using PrairieLearn to write questions, sync information, and more.
 
-Have an error that wasn't listed here but you think it should be? 
-
-Consider [adding the error](https://github.com/PrairieLearn/PrairieLearn/edit/master/doc/faq.md) using GitHub's web editor.
-
 ### Why is my QID invalid?
 
-QID stands for the *Q**uestion **Id**entifier given to the folder that contains
+QID stands for the **Q**uestion **Id**entifier given to the folder that contains
 the question information -- e.g. `info.json`, `question.html`, and `server.py` 
 (optional) -- in `questions/`. QIDs can be invalid if they are spelled incorrectly
 or simply are not present in the directory. On sync, these issues are picked up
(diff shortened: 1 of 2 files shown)


coatless:
@mwest1066 @trombonekenny @nwalters512

doc/faq.md
```diff
@@ -42,3 +42,119 @@ Second, edit the assessment `pl-exp101/courseInstance/Fa17/assessments/final/inf
...
+Within this section of the FAQ, we will go over different errors that may occur
+when using PrairieLearn to write questions, sync information, and more.
+
+Have an error that wasn't listed here but you think it should be? 
+
+Consider [adding the error](https://github.com/PrairieLearn/PrairieLearn/edit/master/doc/faq.md) using GitHub's web editor.
+
+### Why is my QID invalid?
+
+QID stands for the *Q**uestion **Id**entifier given to the folder that contains
```
mwest1066: Missing one asterisk before the Q in Question?

doc/faq.md
```diff
@@ -42,3 +42,119 @@ Second, edit the assessment `pl-exp101/courseInstance/Fa17/assessments/final/inf
 ```
 
 See [Access control](accessControl.md) for more details.
+
+## Common Errors
```
mwest1066: Rather than having a separate section here, how about just listing everything as level-2 headings in the FAQ? I like the text about "Have an error? Add it in the editor" -- this could just be moved to the top of the FAQ?

doc/faq.md
```diff
@@ -42,3 +42,119 @@ Second, edit the assessment `pl-exp101/courseInstance/Fa17/assessments/final/inf
...
+As a result, we'll end up receiving a duplication error when we try to sync
+the question base to the website.
+
+```bash
+<<<<<<< Local Changes
+Error message: Error: UUID 7a323f4c-cafd-40b2-8576-7db18e3f439b is used in multiple questions: Q1, Q2
+=======
+Error message: Error: UUID 7a323f4c-cafd-40b2-8576-7db18e3f439b is used in multiple questions: question_name_uuid_original, question_name_uuid_duplicate
+>>>>>>> External Changes
+```
```
mwest1066: Is the text above output from a `git diff`? I'm not sure why we have the `<<<<<` stuff. I'm also not sure why we have two lines, one with `Q1, Q2` and one with `question_name_uuid_original, question_name_uuid_duplicate`.

doc/faq.md
```diff
@@ -42,3 +42,119 @@ Second, edit the assessment `pl-exp101/courseInstance/Fa17/assessments/final/inf
...
+
+Generating a new UUID via <https://www.uuidgenerator.net/> and substituting it 
+in new question's `info.json` will resolve this issue. That said, care must be
+taken to avoid changing the existing question's UUID that the new question was
+derived from. Modifying the UUID for an existing question being used by students
+will create a new variant of the question.
+
+See [UUIDs in JSON files](uuid.md) for more details.
+
+### Why can't I launch PrairieLearn with docker?
```
mwest1066: Might be worth adding something to this section that if you have run `docker run prairielearn` inside a terminal window, just closing the terminal window is not enough to quit PL. Instead you need to Ctrl-C docker first, then close the terminal window.

doc/faq.md
coatless: I think as the FAQ grows it would make sense to have distinct sections afforded by the level-2 headings. Take for instance a course FAQ, there are registration, grading, OH sections. If all sections were placed together, there would be an accessibility issue.

coatless: So, the main difference is the questions are pulled more out from the side bar...

**New**

screen shot 2018-10-27 at 12 43 48 pm

**Old**

screen shot 2018-10-27 at 12 44 27 pm

-----

In addition, the head mast provides some context:

**New**

screen shot 2018-10-27 at 12 45 18 pm

**Old**

screen shot 2018-10-27 at 12 45 36 pm

mwest1066: Yep, this is the main reason that I'd rather only use two levels of heading in our docs (where possible). Top level is the file, then we have level-2 headings within that (which get shown in the UI). If our FAQ gets too big in the future then I'd prefer to just split it into multiple files. So for now let's just go with:

```
# FAQ

## Single question

## Another common question
```

etc.

coatless: @mwest1066 addressed. Feel free to merge.

PR #1370, opened 2018-11-15
fix users.lti_course_instance_id delete action

mwest1066:


trombonekenny:
Did you notice something broken because of this? I'm curious about the late night PR. :)

mwest1066:
Yeah, it broke `tools/dump_filter.sh` that I use for exporting per-course database dumps. I was doing a new dump of 233 for a grad student who is studying retake exams.

PR #1372, opened 2018-11-15
[WIP] Autosave + restore with localStorage

### LINKED ISSUE #1371: Sync code buffer with localStorage and rehydrate on load
navigating to another page causes me to lose my work. by autosaving and rehydrating, I can always come back to the page and still have my code

As per a discussion with @nwalters512

nwalters512:
So far, just a proof of concept illustrating the use of `window.localStorage` to autosave + restore files. Some things to consider before actually merging:

* Should restore be automatic, or should it prompt first?
* We should probably use a better unique question identifier than `client_files_question_dynamic_url`. It is currently unique per variant, but we probably want the variant ID itself. This required modifying the element/question renderer.

cc @pranaygp

Fixes #1371

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -22,6 +22,8 @@
   * Add documentation on common development errors to FAQ. (James Balamuta).
 
   * Add R Data Packages and SQLite connection to centos-plbase. (James Balamuta).
+
+  * Add `window.localStorage` autosave/restore to `pl-file-editor` (Nathan Walters).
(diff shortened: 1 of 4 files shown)


pranaygp:
> Should restore be automatic, or should it prompt first?

Definitely automatic. No need for an extra step and intuitive UI thereafter if the person wanted a fresh copy is to hit "Restore original question"

PR #1382, opened 2018-12-08
Fix inner_html helper function

nwalters512:
Previously, there was a bug where HTML entities were being pre-escaped in the `inner_html` helper function. This is because the `text` attribute on lxml elements was pre-escaped. We now deliberately escape `text` to account for this.

Also adds the beginning of tests for `prairielearn.py` to avoid future regressions. We can add to these tests as we see fit.


question-servers/freeformPythonLib/prairielearn.py
```diff
@@ -117,13 +118,13 @@ def from_json(v):
...
 
 def inner_html(element):
-    html = element.text
-    if html is None:
-        html = ''
-    html = str(html)
+    inner = element.text
+    if inner is None:
+        inner = ''
+    inner = html.escape(str(inner))
```
mwest1066: good catch

question-servers/freeformPythonLib/prairielearn_test.py
```diff
@@ -1,13 +1,14 @@
+# flake8: noqa
```
mwest1066: Could replace with `# noqa: E402` on lines 7, 8, and 9?

PR #1401, opened 2019-01-15
Adding conf.py to try to fix Sphinx documentation root path detection

echuber2:


mwest1066:
@echuber2 could you help me understand what's going on with the `doc` versus `docs` issue here? I actually wasn't aware that we are using Sphinx. I thought ReadTheDocs is just running `mkdocs`?

echuber2:
Ah, you're right; I found some other users who had the same issue with the Edit on GitHub link breaking on ReadTheDocs but their build system was different. We should be able to fix it just using `mkdocs.yml`. I'll close this request.

PR #1403, opened 2019-01-15
Fix "Edit on GitHub" links in mkdocs.yml

echuber2:
Have to add a line to mkdocs.yml to make the "Edit on GitHub" links point to real pages.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -3,6 +3,8 @@ site_url: https://github.com/PrairieLearn/PrairieLearn
 repo_url: https://github.com/PrairieLearn/PrairieLearn
 theme: readthedocs
 docs_dir: doc
+# Fix "Edit on GitHub" link: (defaults to edit/master/docs/, note the s)
+edit_uri: blob/master/doc/
 
 pages:
 - Main: 'index.md'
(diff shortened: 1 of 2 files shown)


echuber2:
Reference for the edit: https://www.mkdocs.org/user-guide/configuration/

mkdocs.yml
```diff
@@ -3,6 +3,8 @@ site_url: https://github.com/PrairieLearn/PrairieLearn
 repo_url: https://github.com/PrairieLearn/PrairieLearn
 theme: readthedocs
 docs_dir: doc
+# Fix "Edit on GitHub" link: (defaults to edit/master/docs/, note the s)
+edit_uri: blob/master/doc/
```
mwest1066: Should this be `edit/master/doc/`?

echuber2: `edit/master/doc/` will lead directly to the page source opened in the web editor, whereas `blob/master/doc/` leads to the viewing page instead (like normal browsing on GitHub would). My thought was that this would be more appropriate if someone just wanted to locate the source file on GH quickly, and also it doesn't require the user to be logged in to work.

echuber2: Of course someone can click the normal GitHub edit button after arriving at the viewing page, too.

mwest1066: Sounds good to me, thanks!

PR #1404, opened 2019-01-15
change native install doc to be macos-specific

tbretl:
The docs for installing natively were a little out-of-date and, as far as I know, were only ever used to install on MacOS. I changed the docs to be MacOS-specific and improved them a little bit (and made sure they worked on my own machine).

One thing these docs do not include are clear instructions for installing R packages. One would want these packages to be installed in the same `conda` environment that is used for the python packages. My understanding is that this could be achieved by calling `Rscript environments/centos7-plbase/r-requirements.R` within this environment - however, the script `r-requirements.R` may require slight modification to specify the install directory (https://stackoverflow.com/a/35023854). Perhaps @coatless knows how to do this - I'll leave it to him.


doc/installingNative.md
```diff
@@ -1,78 +1,100 @@
...
 pg_ctl -D ~/defaultdb -l ~/logfile start
 ```
 
+    Note that you can put `defaultdb` and `logfile` anywhere you like.
+
 * Make sure the `postgres` database user exists and is a superuser (these might error if the user already exists):
 
 ```sh
 psql -c "CREATE USER postgres;"
 psql -c "ALTER USER postgres WITH SUPERUSER;"
```
mwest1066: If postgres is installed then you should be able to run `createuser -s postgres` at the commandline to do both of these.

doc/installingNative.md
```diff
@@ -1,78 +1,100 @@
...
 
 ```sh
-python3 --version     # should return "Python 3.6" or higher
-python3.6 --version   # should return "Python 3.6" or higher
+cd PrairieLearn
+python -m pip install -r environments/centos7-plbase/python-requirements.txt
 ```
 
-* Install the Python libraries:
+* (Optionally, install the R packages that are used on the production server, as in `environments/centos7-plbase/r-requirements.R`.)
```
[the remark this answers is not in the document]
tbretl: (I used `Rscript ...` without any need for `chmod +x ...` - is it essential?)

When I tried this, I encountered some "permission denied" errors, like the ones discussed here:
https://github.com/Homebrew/homebrew-core/issues/17098

Thoughts?

Also, is it correct that this will install packages globally? If so, is there any alternative that will contain this install to an environment, for example a `conda` environment?

Thanks!

doc/installingNative.md
```diff
@@ -106,6 +128,6 @@ cd PrairieLearn
 node server
 ```
 
-   This should end with `PrairieLearn server ready` and will remain running in the foreground, so this terminal can't be used for anything else. Stopping or restarting the server can be done with `Crtl-C`.
+    This should end with `PrairieLearn server ready` and will remain running in the foreground, so this terminal can't be used for anything else. Stopping or restarting the server can be done with `Crtl-C`.
```
mwest1066: Spurious space at start of line?

mwest1066:
Did the various questions above get resolved?

mwest1066:
@tbretl bump on this PR

mwest1066:
@tbretl bump again

coatless:
So, procrastination on this PR paid off... R 4.0 is set to be released this week! So, everything gets simplified a bit.

nwalters512:
This hasn't seen any movement for a few years now - going to close this. Thankfully, the native install docs are now much better than they were three years ago! If there's anything here that's worth adding to the docs, let's add it in a new PR.

PR #1411, opened 2019-01-18
Add GraphViz dependency to Python, CentOS

nicknytko:
Needed for graph drawing

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -29,7 +29,7 @@
 
   * Add default tags for semester Sp19 - Fa21 (James Balamuta).
 
-  * Add Graphviz Yum package and Python library to centos7-plbase and centos7-python (Nicolas Nytko)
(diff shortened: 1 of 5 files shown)


environments/centos7-python/requirements.txt
```diff
@@ -5,3 +5,4 @@ pandas==0.22.0
 scipy==1.0.0
 sympy==1.1.1
 Pillow==5.2.0
+pygraphviz==1.5
```
mwest1066: To make this work in `centos7-python` I think you'll also need to add the yum packages to `centos7-python/Dockerfile`.

ChangeLog.md
```diff
@@ -29,6 +29,8 @@
 
   * Add default tags for semester Sp19 - Fa21 (James Balamuta).
 
+  * Add Graphviz Yum package and Python library to centos7-plbase and centos7-python (Nicolas Nytko)
```
mwest1066: Add a period at the end.

PR #1414, opened 2019-01-18
Add Graphviz rendering element

nwalters512:
This is an element that can render Graphviz DOT notation on the client with [viz.js](https://github.com/mdaines/viz.js). We've successfully used this in CS225, and CS357 is interested in something similar.


mfsilva22:
I was able to take a quick look at the example. Yes, CS357 needs a way of creating graphs that look the same. However, I was planning to have a PL element that would take a numpy array (the adjacency matrix) and then make the graph. if `display-weight = true`, then the entries of the array would be displayed as the weights. Instead of letters for the graph nodes, we have numbers corresponding to the row/column index. 

It seems that it would be better if we submit a different PR for that?

mfsilva22:
I know it will render any graph, I am just not sure it will be easy to use (call the element). 

Maybe we can just modify your element to take an attribute like the `params-name` used in other PL matrix elements. For my questions, the server.py will be generating the adjacency matrix, that we want to be illustrated as a graph.

nwalters512:
I would prefer to just template it directly into the element:

```html
<pl-graphviz-render>{{{ params.mygraphstring }}}</pl-graphviz-render>
```

I don't want to build in support for any specific graph "sources" - if you want an element that can go straight from numpy array to graph, you could always build an abstraction on top of this element to do so.

PR #1421, opened 2019-02-03
Fix escape of code under pl-code's file input (close #1420)

### LINKED ISSUE #1420: Escape HTML Read In by File with pl-code
When attempting to read in a code file via `pl-code`, any HTML or XML files contents are not escaped. e.g.

```html
<pl-code language='xml' source-file-name='code.html'>
</pl-code>
```

Would display XML data with: 

screen shot 2019-02-03 at 4 34 01 pm

Looking at the source, we can see that the XML was evaluated and not escaped: 

screen shot 2019-02-03 at 4 34 17 pm

coatless:
Within this PR, we:

- Ensured that code read in via `source-file-name` in `pl-code` was escaped.
- Added an example of reading in an XML data file (note, `xml` is not a language supported.)

Details in #1420.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -33,6 +33,8 @@
 
   * Add Graphviz Yum package and Python library to centos7-plbase and centos7-python (Nicolas Nytko).
 
+  * Add a second example of reading XML code in from a file with `pl-code` (James Balamuta).
+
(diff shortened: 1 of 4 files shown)


coatless:
Paging: @mwest1066 || @nwalters512

PR #1431, opened 2019-02-23
Change 'pl-number-input' to accept fractional answers

henryg3:
Added the ability to use fractional inputs by checking if slash is present in string. If present, the quotient will be computed and then sent as an answer. Adding fractional input functionality will be useful in the event a student does not have a calculator to convert a difficult fraction to decimal form during a test or if they're studying.


elements/pl-number-input/pl-number-input.py
```diff
@@ -183,21 +183,51 @@ def parse(element_html, data):
...
-        a_sub_parsed = pl.string_to_number(a_sub, allow_complex=allow_complex)
-        if a_sub_parsed is None:
-            raise ValueError('invalid submitted answer (wrong type)')
-        if not np.isfinite(a_sub_parsed):
-            raise ValueError('invalid submitted answer (not finite)')
+        # Accept fractional input, then convert it to a float or decimal
+        # (only works for certain integer addition/subtraction problems)
+    elif '/' in a_sub:
+        a_sub = a_sub.strip()
+        a_sub_list = a_sub.split('/')
```
mwest1066: What if there is more than one `'/'` character in the string?

elements/pl-number-input/pl-number-input.py
```diff
@@ -183,21 +183,51 @@ def parse(element_html, data):
...
-    try:
-        a_sub_parsed = pl.string_to_number(a_sub, allow_complex=allow_complex)
-        if a_sub_parsed is None:
-            raise ValueError('invalid submitted answer (wrong type)')
-        if not np.isfinite(a_sub_parsed):
-            raise ValueError('invalid submitted answer (not finite)')
+        # Accept fractional input, then convert it to a float or decimal
+        # (only works for certain integer addition/subtraction problems)
+    elif '/' in a_sub:
+        a_sub = a_sub.strip()
```
mwest1066: is this `strip()` necessary?

elements/pl-number-input/pl-number-input.py
```diff
@@ -183,21 +183,51 @@ def parse(element_html, data):
...
+        a_sub_list = a_sub.split('/')
+        try:
+            a_sub_parsed_numerator = pl.string_to_number(a_sub_list[0])
+            if a_sub_parsed_numerator is None:
+                raise ValueError('invalid submitted answer (wrong type)')
+            if not np.isfinite(a_sub_parsed_numerator):
+                raise ValueError('invalid submitted answer (not finite)')
+                return
```
mwest1066: why do we have a `return` after a `raise`?

elements/pl-number-input/pl-number-input.py
```diff
@@ -183,21 +183,51 @@ def parse(element_html, data):
...
+        except Exception:
+            data['format_errors'][name] = 'Invalid format. The submitted answer could not be interpreted as a double-precision floating-point number or fraction.'
+            data['submitted_answers'][name] = None
+            return
+        try:
+            a_sub_parsed_denominator = pl.string_to_number(a_sub_list[1])
+            if a_sub_parsed_denominator is None:
+                raise ValueError('invalid submitted answer (wrong type)')
+            if not np.isfinite(a_sub_parsed_denominator):
+                raise ValueError('invalid submitted answer (not finite)')
```
mwest1066: I'm having trouble following the Exception logic here -- it looks like we raise a ValueError with an error message, but then we catch it below and discard this error message?

elements/pl-number-input/pl-number-input.mustache
```diff
@@ -71,7 +71,7 @@ Your answer must be either a real number between <code>-1e308</code> and <code>1
...
+Fractions (e.g., <code>3/4</code> and <code>1.0/2.5</code>) are accepted, while no other symbolic expressions are accepted. Scientific notation is accepted (e.g., <code>1.2e03</code>). {{#relabs}}<p>Your answer must be accurate to within relative tolerance <code>{{rtol}}</code> and absolute tolerance <code>{{atol}}</code>.</p>{{/relabs}}{{#sigfig}}<p>Your answer must be accurate to {{digits}} significant figures. This means (for example) that if the true answer is <code>1.234</code>, then the submitted answer must be between <code>1.234 - {{comparison_eps}}</code> and <code>1.234 + {{comparison_eps}}</code> to be counted correct.</p>{{/sigfig}}{{#decdig}}<p>Your answer must be accurate to {{digits}} digits after the decimal. This means (for example) that if the true answer is <code>1.234</code>, then the submitted answer must be between <code>1.234 - {{comparison_eps}}</code> and <code>1.234 + {{comparison_eps}}</code> to be counted correct.</p>{{/decdig}}
```
mwest1066: How do fractions interact with complex numbers? Do we handle fractions in complex numbers? If not, we should tell the user this in the help text.

elements/pl-number-input/pl-number-input.py
```diff
@@ -183,21 +183,51 @@ def parse(element_html, data):
...
-            raise ValueError('invalid submitted answer (not finite)')
+        # Accept fractional input, then convert it to a float or decimal
+        # (only works for certain integer addition/subtraction problems)
+    elif '/' in a_sub:
+        a_sub = a_sub.strip()
+        a_sub_list = a_sub.split('/')
+        try:
+            a_sub_parsed_numerator = pl.string_to_number(a_sub_list[0])
+            if a_sub_parsed_numerator is None:
+                raise ValueError('invalid submitted answer (wrong type)')
```
mwest1066: In these errors, should we specify that it's the numerator that's wrong? And similarly for other errors.

elements/pl-number-input/pl-number-input.py
henryg3: There will be an error message that shows after the user grades their answer 
screen shot 2019-02-23 at 5 44 02 pm

elements/pl-number-input/pl-number-input.py
henryg3: I just realized .split() makes a separated list, so it is not necessary. Removed .strip() in the code.

elements/pl-number-input/pl-number-input.mustache
henryg3: They should not be handled with complex numbers. Noted.

elements/pl-number-input/pl-number-input.py
henryg3: Just noticed the redundancy. Noted and removed.

elements/pl-number-input/pl-number-input.py
henryg3: Noted and added.

henryg3: For the denominator, what would we do in the case that it is a large, near-infinite value?

henryg3: If the denominator was an infinitely large value, wouldn't the fraction simply equate to 0?

mwest1066:
I'm testing this on the `addNumbers` question in the example course. Entering `1/0` causes an uncaught Exception to be thrown.

elements/pl-number-input/pl-number-input.py
mwest1066: Entering `20/5/10` is accepted as a valid number, which it should not be.

mwest1066:
Entering `1e200/1e-200` throws an uncaught exception.

mwest1066:
@henryg3 did you want to continue working on this PR? It's fine if you don't have time, and in that case I'll go ahead and close it out.

henryg3:
> @henryg3 did you want to continue working on this PR? It's fine if you don't have time, and in that case I'll go ahead and close it out.

Hi Matt, I've been really busy with schoolwork and midterms next week, so I have not been able to continue working on this PR, but I am still interested in working on it, perhaps over the summer.

mwest1066:
Thanks Henry! I'll close this PR for now, but we can always re-open it again if you have time to work on it. Just let me know!

PR #1453, opened 2019-03-07
Improvements to pl-matrix-output element

### LINKED ISSUE #1365: Add MATLAB comment option and a Mathematica tab to the pl-matrix-output element
Add a "params-comment" attribute to < variable >

Will need to consider case where comments are not specified, that will be the existing concatenation.

Given _varName_, _digit_, and _comment_, the changes will be as follows:
Matlab commented string syntax:
     _varName_ = _digit_; %_comment_
Mathematica commented string syntax:
     _varName_ = _digit_; (* _comment_ * )

Expected minimum changes needed:
ADD: `var_comment = pl.get_string_attrib(child, 'params-comment')`
UPDATE: `matlab_data += pl.inner_html(child) + ' = ' + pl.string_from_2darray(var_data, language='matlab', digits=digits) + '; %' + var_comment +'\n'`

eliving2:
Added Mathematica tab to matrix-output
Added comment capability as a local (optional) parameter named `params-comment`, works with all three coding languages


eliving2:
image

Add optional parameter `params-digits` to change number of display digits for each variable

elements/pl-matrix-output/pl-matrix-output.py
```diff
@@ -45,13 +59,27 @@ def render(element_html, data):
...
+            mathematica_reserved = ['C', 'D', 'E', 'I', 'K', 'N', 'O']
+            if pl.inner_html(child) in mathematica_reserved:
+                mathematica_suffix = 'm'
+            else:
+                mathematica_suffix = ''
+
             # Create string for matlab and python format
-            matlab_data += pl.inner_html(child) + ' = ' + pl.string_from_2darray(var_data, language='matlab', digits=digits) + ';\n'
-            python_data += pl.inner_html(child) + ' = ' + prefix + pl.string_from_2darray(var_data, language='python', digits=digits) + suffix + '\n'
+            if commentExists == 0:
```
mwest1066: I suggest replacing this with `if pl.has_attrib(child, 'params-comment'):` and deleting the `commentExists` variable.

(In case we keep `commentExists`, we should change it to be snake_case and use boolean True/False rather than 0/1)

elements/pl-matrix-output/pl-matrix-output.py
```diff
@@ -45,13 +59,27 @@ def render(element_html, data):
...
+                matlab_data += pl.inner_html(child) + ' = ' + pl.string_from_2darray(var_data, language='matlab', digits=var_digits) + ';\n'
+                mathematica_data += pl.inner_html(child) + ' = ' + pl.string_from_2darray(var_data, language='matlab', digits=var_digits) + ';\n'
+                python_data += pl.inner_html(child) + ' = ' + prefix + pl.string_from_2darray(var_data, language='python', digits=var_digits) + suffix + '\n'
+            else:
+                matlab_data += pl.inner_html(child) + ' = ' + pl.string_from_2darray(var_data, language='matlab', digits=var_digits) + '; % ' + var_comment + '\n'
+                mathematica_data += pl.inner_html(child) + mathematica_suffix + ' = ' + pl.string_from_2darray(var_data, language='matlab', digits=var_digits) + '; (* ' + var_comment + ' *)\n'
+                python_data += pl.inner_html(child) + ' = ' + prefix + pl.string_from_2darray(var_data, language='python', digits=var_digits) + suffix + ' # ' + var_comment + '\n'
 
     html_params = {
         'default_is_matlab': True,
```
mwest1066: Is this hard-coded? I thought we got to choose?

elements/pl-matrix-output/pl-matrix-output.mustache
```diff
@@ -8,6 +8,7 @@
     <div class="card-header">
         <ul class="nav nav-tabs card-header-tabs" role="tablist">
             <li class="nav-item" role="presentation"><a class="nav-link{{#default_is_matlab}} active{{/default_is_matlab}}" href="#matlab-{{uuid}}" aria-controls="matlab-{{uuid}}" role="tab" data-toggle="pill">matlab</a></li>
+            <li class="nav-item" role="presentation"><a class="nav-link{{#default_is_mathematica}} active{{/default_is_mathematica}}" href="#mathematica-{{uuid}}" aria-controls="mathematica-{{uuid}}" role="tab" data-toggle="pill">mathematica</a></li>
```
mwest1066: Can we capitalize the user-visible tab labels to "Mathematica", "Matlab", "Python"?

mwest1066:
@eliving2 would you mind adding a new example question that shows off this element? See `examplesNumberInput` for an example question that demonstrates all the features of a single element. It would be great to have an `examplesMatrixOutput` that does the same thing, and this would also make easier for me to test and review. Thanks!

elements/pl-matrix-output/pl-matrix-output.py
mfsilva22: It has been been hard-coded, and I agree that we should choose! It is okay to have just another attribute :-)

mwest1066: Yep, an attribute for this would be great!

mfsilva22:
> @eliving2 would you mind adding a new example question that shows off this element? See `examplesNumberInput` for an example question that demonstrates all the features of a single element. It would be great to have an `examplesMatrixOutput` that does the same thing, and this would also make easier for me to test and review. Thanks!

mwest1066:
@mfsilva22 any particular reason this PR was closed?

ChangeLog.md
```diff
@@ -151,6 +151,10 @@
 
   * Remove old temporary upgrade flag `tmp_upgraded_iq_status` (Matt West).
 
+  * Add Mathematica tab to `pl-matrix-output` (Liz Livingston)
+
+  * Add comment and local digit control for `pl-matrix-output`, optional `params-comment` and `params-digits` (Liz Livingston)
+
```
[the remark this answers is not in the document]
eliving2: Moved and added periods.

doc/elements.md
```diff
@@ -213,19 +213,26 @@ Attributes for `<variable>` (one of these for each variable to display):
 Attribute | Type | Default | Description
 --- | --- | --- | ---
 `params-name` | string | — | Name of variable in `data['params']` to display.
+`params-comment` | string | — | Name of variable in `data['params']` to display.
```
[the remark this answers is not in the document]
eliving2: Fixed description

doc/elements.md
```diff
@@ -213,19 +213,26 @@ Attributes for `<variable>` (one of these for each variable to display):
...
+`params-digits` | integer | — | Number of digits to display after the decimal for the variable.
 
-This element displays a list of variables inside `<pre>` tags that are formatted for import into either MATLAB or python (the user can switch between the two). Each variable must be either a scalar or a 2D numpy array (expressed as a list). Each variable will be prefixed by the text that appears between the `<variable>` and `</variable>` tags, followed by ` = `.
+This element displays a list of variables inside `<pre>` tags that are formatted for import into either MATLAB, Mathematica, or python (the user can switch between the two). Each variable must be either a scalar or a 2D numpy array (expressed as a list). Each variable will be prefixed by the text that appears between the `<variable>` and `</variable>` tags, followed by ` = `.
```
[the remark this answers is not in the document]
eliving2: Capitalized

elements/pl-matrix-output/pl-matrix-output.py
eliving2: Doesn't make a difference to me, removed `commentExists` variable with suggested line

elements/pl-matrix-output/pl-matrix-output.py
eliving2: What do you mean hardcoded? Choosing which tabs appear?

elements/pl-matrix-output/pl-matrix-output.mustache
eliving2: Capitalized

elements/pl-matrix-output/pl-matrix-output.py
coatless: @eliving2 I think they're discussing what should be the "default" tab shown on question load. Within `html_params`, only ` 'default_is_matlab': True` is set. However, the mustache template has: 

-`#default_is_matlab`
-`#default_is_python`
- and now `#default_is_mathematica`

eliving2:
@mfsilva22 @mwest1066 I noticed a mistake while adding attributes for tab display and finishing the example question (`examplesMatrixOutput`). 

I use `pl.string_from_2darray(var_data, language='matlab')` for Mathematica because I wasn't thinking about outputting matrices. Apparently, Mathematica syntax is: `mat = {{1, 2, 3}, {4, 5, 6}, {7, 8, 9}}`...should I edit `pl.string_from_2darray` as part of this PR?

mwest1066:
I agree that adding `language='mathematica'` support to `pl.string_from_2darray()` would be the right thing to handle this correctly. Including it in this PR would make sense. Thanks!

elements/pl-matrix-output/pl-matrix-output.py
eliving2: Ended up implementing an option to hide tabs. No idea how I would change the display order in the mustache file. If someone knows how to implement this, I am happy to fix it

exampleCourse/questions/examplesMatrixOutput/question.html
```diff
@@ -0,0 +1,82 @@
...
+    </div>
+  </div>
+
+  <div class="card my-2">
+    <div class="card-body">
+      <pl-question-panel>
+        <p>Another option is to display a tab for Mathematica. This is very similar to Matlab syntax, but it takes into account Mathematica reserved variables (C, D, E, I, K, N, O) and formats comments correctly. This uses the optional parameter <code>display-mathematica-tab</code>.</p>
+      </pl-question-panel>
+      <pl-matrix-output digits="0" display-mathematica-tab="True">
+        <variable params-name="x" params-comment="The Mathematica tab adds an 'm' to the variable name">C</variable>
```
mwest1066: When I run this, I don't see the `m` being added. I just see the plain variable names (like `C`).

exampleCourse/questions/examplesMatrixOutput/server.py
```diff
@@ -0,0 +1,23 @@
+import random, math
+
+def generate(data):
+
+    # Simulate values
+    a = random.randint(2, 10)
+    b = random.randint(2, 10)
+    c = random.randint(20,100)/100
```
mwest1066: maybe we should add an example that's just a random real number, so it has lots of digits displayed by default?

question-servers/freeformPythonLib/prairielearn.py
```diff
@@ -341,6 +341,10 @@ def string_from_2darray(A, language='python', presentation_type='f', digits=2):
 
         [ ... ... ; ... ... ]
 
+    If language is 'mathematica' and A is a 2D ndarray, the string looks like this:
+
+        {{ ..., ... }{ ..., ... }}
```
mwest1066: missing a comma in the middle?

question-servers/freeformPythonLib/prairielearn.py
```diff
@@ -375,8 +379,23 @@ def string_from_2darray(A, language='python', presentation_type='f', digits=2):
             return numpy_to_matlab_sf(A, ndigits=digits)
         else:
             return numpy_to_matlab(A, ndigits=digits, wtype=presentation_type)
+    elif language == 'mathematica':
+        if presentation_type == 'sigfig':
```
mwest1066: How do we trigger `sigfig` presentation? Is there an option for this? I don't think it's documented.

mwest1066: Also, the docs say that `digits` specifies "Number of digits to display after the decimal", which isn't true if we are doing `sigfig`.

elements/pl-matrix-output/pl-matrix-output.py
```diff
@@ -6,19 +6,37 @@
...
+    active_tab_matlab = True
+    active_tab_mathematica = False
+    active_tab_python = False
+    # If Matlab Display is False, will cycle through to next displayed tab
+    if display_matlab_tab is False:
+        active_tab_matlab = False
+        if display_mathematica_tab is True:
+            active_tab_mathematica = True
+        elif display_python_tab is True:
+            active_tab_python = True
```
mwest1066: I think we should add a parameter to allow explicit specification of the default tab. Maybe `default-tab="matlab"`?

elements/pl-matrix-output/pl-matrix-output.py
```diff
@@ -45,13 +73,32 @@ def render(element_html, data):
...
+            if pl.has_attrib(child, 'params-comment'):
+                matlab_data += pl.inner_html(child) + ' = ' + pl.string_from_2darray(var_data, language='matlab', digits=var_digits) + '; % ' + var_comment + '\n'
+                mathematica_data += pl.inner_html(child) + mathematica_suffix + ' = ' + pl.string_from_2darray(var_data, language='mathematica', digits=var_digits) + '; (* ' + var_comment + ' *)\n'
+                python_data += pl.inner_html(child) + ' = ' + prefix + pl.string_from_2darray(var_data, language='python', digits=var_digits) + suffix + ' # ' + var_comment + '\n'
+            else:
+                matlab_data += pl.inner_html(child) + ' = ' + pl.string_from_2darray(var_data, language='matlab', digits=var_digits) + ';\n'
+                mathematica_data += pl.inner_html(child) + mathematica_suffix + ' = ' + pl.string_from_2darray(var_data, language='mathematica', digits=var_digits) + ';\n'
+                python_data += pl.inner_html(child) + ' = ' + prefix + pl.string_from_2darray(var_data, language='python', digits=var_digits) + suffix + '\n'
```
mwest1066: Can we avoid the code duplication here? Maybe first make a variable like `var_formatted_comment` which is either the empty string (if there is no comment) or something like `'(*' + comment + '*)'`.

elements/pl-matrix-output/pl-matrix-output.py
```diff
@@ -45,13 +73,32 @@ def render(element_html, data):
...
+            if pl.inner_html(child) in mathematica_reserved:
+                mathematica_suffix = 'm'
+            else:
+                mathematica_suffix = ''
+
             # Create string for matlab and python format
-            matlab_data += pl.inner_html(child) + ' = ' + pl.string_from_2darray(var_data, language='matlab', digits=digits) + ';\n'
-            python_data += pl.inner_html(child) + ' = ' + prefix + pl.string_from_2darray(var_data, language='python', digits=digits) + suffix + '\n'
+            if pl.has_attrib(child, 'params-comment'):
+                matlab_data += pl.inner_html(child) + ' = ' + pl.string_from_2darray(var_data, language='matlab', digits=var_digits) + '; % ' + var_comment + '\n'
```
mwest1066: can we just set variables like `var_name = pl.inner_html(child)`, `var_formatted_data = pl.string_from_2darray(var_data, language=var_language, digits=var_digits)` etc, so that we don't have to repeat all this code?

mwest1066: I suggest using python f-strings to make this cleaner.

mwest1066:
The example question is awesome! It really helped me to review the code better. Thanks!

mwest1066:
Two other things:

1. As @mfsilva22 points out above, the name for this element is not really appropriate anymore. Should we rename it to `pl-variable-output` or something like that?

2. Is there any reason we only support scalars and 2D matrices? Any reason we couldn't easily add 1D vector support?

Sorry to expand the scope of your work by asking these questions :-) But I think it would make this a much more powerful element if it could just do a little bit more?

mfsilva22:
> Two other things:
> 
> 1. As @mfsilva22 points out above, the name for this element is not really appropriate anymore. Should we rename it to `pl-variable-output` or something like that?
> 2. Is there any reason we only support scalars and 2D matrices? Any reason we couldn't easily add 1D vector support?
> 
> Sorry to expand the scope of your work by asking these questions :-) But I think it would make this a much more powerful element if it could just do a little bit more?

eliving2:
@mfsilva22 You closed the PR again :(

question-servers/freeformPythonLib/prairielearn.py
eliving2: I have no idea how this works and actually noticed with my example question. The default "digits" for the element is `3`, when I set it to `2`, i think it displays the same number of digits after the decimal.

eliving2:
> Two other things:
> 
> 1. As @mfsilva22 points out above, the name for this element is not really appropriate anymore. Should we rename it to `pl-variable-output` or something like that?
> 2. Is there any reason we only support scalars and 2D matrices? Any reason we couldn't easily add 1D vector support?
> 
> Sorry to expand the scope of your work by asking these questions :-) But I think it would make this a much more powerful element if it could just do a little bit more?

1. I agree and think renaming it would be better since this has a lot more applications than just matrix output. How would that work? Would this just be a "new" element and `pl-matrix-output` would be slowly removed just like PLv2?

2. I was actually confused when trying to get the matrix display to work because I needed two sets of brackets. I've never used this for matrices, but it would make sense to also support vectors. I can try to implement this?

exampleCourse/questions/examplesMatrixOutput/info.json
```diff
@@ -0,0 +1,7 @@
+{
+    "uuid": "cc99f57c-1a13-4161-b052-1cd795011d98",
+    "title": "Examples showing matrix and scalar output via pl-number-output",
```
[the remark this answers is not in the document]
eliving2: Yup, copy and paste error.

exampleCourse/questions/examplesMatrixOutput/question.html
```diff
@@ -0,0 +1,82 @@
+<pl-question-panel>
+  <p>
+    Questions below were designed to showcase the different features of
```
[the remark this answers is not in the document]
eliving2: Went back through and tried to fix this. Let me know if there are any other nonsensical or incomplete sentences.

exampleCourse/questions/examplesMatrixOutput/question.html
eliving2: The `m` is only added in the Mathematica tab, I have clarified this in the example question explanation

exampleCourse/questions/examplesMatrixOutput/server.py
eliving2: Added this, but the default display is still not acting as expected.

elements/pl-matrix-output/pl-matrix-output.py
```diff
@@ -6,19 +6,37 @@
 
 def prepare(element_html, data):
     element = lxml.html.fragment_fromstring(element_html)
-    pl.check_attribs(element, required_attribs=[], optional_attribs=['digits'])
+    required_attribs = []
+    optional_attribs = ['digits', 'display-matlab-tab', 'display-mathematica-tab', 'display-python-tab']
```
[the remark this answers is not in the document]
eliving2: Changed this, left the internal html_params names as they were, not sure if they should also be "show_matlab" instead of "display_matlab_tab" for consistency

elements/pl-matrix-output/pl-matrix-output.py
```diff
@@ -6,19 +6,37 @@
...
+            active_tab_python = True
 
     matlab_data = ''
+    mathematica_data = ''
     python_data = 'import numpy as np\n\n'
     for child in element:
         if child.tag == 'variable':
             # Raise exception of variable does not have a name
-            pl.check_attribs(child, required_attribs=['params-name'], optional_attribs=[])
+            pl.check_attribs(child, required_attribs=['params-name'], optional_attribs=['params-comment', 'params-digits'])
```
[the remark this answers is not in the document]
eliving2: Ah, that makes sense. Any reason to rename digits since it is also an attribute of the element?

elements/pl-matrix-output/pl-matrix-output.py
eliving2: I added this, I still hate how long and kind of "hard-coded" this is. Open to suggestions for better implementation strategies

elements/pl-matrix-output/pl-matrix-output.py
eliving2: Changed to f-strings

eliving2: Couldn't think of how to make `var_language` work with the current implementation. I made separate `var_language_data` which still cleans up some of the repeated code

elements/pl-matrix-output/pl-matrix-output.py
eliving2: I did this, but again it has to be custom for each language, so I made `var_matlab_comment`, `var_mathematica_comment` and `var_python_comment`.

Not ideal but helps prevent the duplicate code?

question-servers/freeformPythonLib/prairielearn.py
eliving2: Yup, fixed this

eliving2:
@mwest1066 , taking @mfsilva22 's comment into consideration, should I rename the element to `pl-variable-output` and add 1D vector support? 

If so, how would that work? Would this just be a "new" element and `pl-matrix-output` would be slowly removed just like PLv2 will eventually not be supported?

elements/pl-matrix-output/pl-matrix-output.py
eliving2: Added a default-tab option, resolving conversation.

elements/pl-matrix-output/pl-matrix-output.py
```diff
@@ -6,19 +6,49 @@
...
     element = lxml.html.fragment_fromstring(element_html)
     digits = pl.get_integer_attrib(element, 'digits', 2)
+    display_matlab_tab = pl.get_boolean_attrib(element, 'show-matlab', True)
+    display_mathematica_tab = pl.get_boolean_attrib(element, 'show-mathematica', True)
+    display_python_tab = pl.get_boolean_attrib(element, 'show-python', True)
+    default_tab = pl.get_string_attrib(element, 'default-tab', 'matlab')
+
+    # Default active tab
+    if default_tab == 'matlab':
+        active_tab_matlab = True
```
coatless: How about setting these before the `if` to be `False` and then inside the appropriate `if` body set the specific value to `True`? 

e.g. 
```py

# Initialize values to false
active_tab_matlab = False
active_tab_mathematica = False
active_tab_python = False

# Set starting tab
if default_tab == 'matlab':
    active_tab_matlab = True
elif default_tab == 'mathematica': 
    active_tab_mathematica = True
... 
```

eliving2: This is way more elegant and less repetitive. I also changed up my default tab handling. 

Do I need to throw an error if someone tries to set the default tab to one that they have hidden? Right now the code just handles it and basically ignores the impossible `default-tab` setting.

elements/pl-matrix-output/pl-matrix-output.py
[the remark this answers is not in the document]
mwest1066: No, `digits` is fine. We generally prefer short name. The only reason for `params-name` has `params` on it is for consistency with other elements. We often want to specify variables from inside `params`, `correct-answers`, etc, so we try and name these `params-name`, `correct-answers-name`, etc, so that we are very clear about which structure the name should be used in.

elements/pl-matrix-output/pl-matrix-output.py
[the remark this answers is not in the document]
mwest1066: I’m inclined to also change the internal `html_params` names. The less changing of names we have the better (although the internal names should be snake_case).

elements/pl-matrix-output/pl-matrix-output.py
mwest1066: I think it’s fine to just silently work the best we can in this case.

mwest1066:
@eliving2 I think doing a rename to `pl-variable-output` and adding 1D support is the way to go. It would make this element much more useful. The easiest thing to preserve backwards compatability is to leave a copy of `pl-matrix-output` in place. This can be the old version (before any of your changes). We should remove all references to `pl-matrix-output` in the docs and example questions, so no one will know about it. At some point in the future we can get everyone using it to switch to `pl-variable-ouput` and then delete `pl-matrix-output` (probably over summer).

doc/elements.md
```diff
@@ -207,25 +207,36 @@ Attributes for `<pl-matrix-output`:
 Attribute | Type | Default | Description
 --- | --- | --- | ---
 `digits` | integer | — | Number of digits to display after the decimal.
+`default-tab` | string | 'matlab' | Select the active tab.
+`show-matlab` | boolean | True | Toggles the display of the Matlab tab.
+`show-mathetmatica` | boolean | True | Toggles the display of the Mathematica tab.
```
mwest1066: Typo in mathematica.

elements/pl-matrix-output/pl-matrix-output.py
```diff
@@ -6,19 +6,45 @@
...
 def render(element_html, data):
     element = lxml.html.fragment_fromstring(element_html)
     digits = pl.get_integer_attrib(element, 'digits', 2)
+    display_matlab_tab = pl.get_boolean_attrib(element, 'show-matlab', True)
+    display_mathematica_tab = pl.get_boolean_attrib(element, 'show-mathematica', True)
+    display_python_tab = pl.get_boolean_attrib(element, 'show-python', True)
+    default_tab = pl.get_string_attrib(element, 'default-tab', 'matlab')
 
+    # Handles cases when Matlab tab hidden, if python is default tab don't overwite active tab setting
+    if display_matlab_tab is False and default_tab != 'python':
```
mwest1066: I think this code will work, but it seems weird. I think it’s trying to say:
```
if not display_matlab_tab and default_tab == ‘matlab’:
```

elements/pl-matrix-output/pl-matrix-output.py
```diff
@@ -6,19 +6,45 @@
...
+    pl.check_attribs(element, required_attribs, optional_attribs)
 
 
 def render(element_html, data):
     element = lxml.html.fragment_fromstring(element_html)
     digits = pl.get_integer_attrib(element, 'digits', 2)
+    display_matlab_tab = pl.get_boolean_attrib(element, 'show-matlab', True)
+    display_mathematica_tab = pl.get_boolean_attrib(element, 'show-mathematica', True)
+    display_python_tab = pl.get_boolean_attrib(element, 'show-python', True)
+    default_tab = pl.get_string_attrib(element, 'default-tab', 'matlab')
```
mwest1066: We should do something like this:
```
if default_tab is not in [‘matlab’, ‘python’, ‘mathematica’]:
    raise Exception(f’invalid default-tab: {default_tab}’)
```

mwest1066: We should do something like this:
```
if default_tab is not in [‘matlab’, ‘python’, ‘mathematica’]:
    raise Exception(f’invalid default-tab: {default_tab}’)
```

elements/pl-matrix-output/pl-matrix-output.py
```diff
@@ -6,19 +6,45 @@
...
 
+    # Handles cases when Matlab tab hidden, if python is default tab don't overwite active tab setting
+    if display_matlab_tab is False and default_tab != 'python':
+        if display_mathematica_tab is True:
+            default_tab = 'mathematica'
+        elif display_python_tab is True:
+            default_tab = 'python'
```
mwest1066: From PEP8, in Python we should write `if X` and `if not X`, rather than `if X is True` or `if X is False`. This occurs elsewhere in the code as well.

elements/pl-matrix-output/pl-matrix-output.py
```diff
@@ -32,6 +58,23 @@ def render(element_html, data):
...
 
+            # Get comment, if it exists
+            var_matlab_comment = ''
+            var_mathematica_comment = ''
+            var_python_comment = ''
+            if pl.has_attrib(child, 'comment'):
+                var_comment = pl.get_string_attrib(child, 'comment')
+                var_matlab_comment = '% {}'.format(var_comment)
+                var_mathematica_comment = '(* {} *)'.format(var_comment)
+                var_python_comment = '# {}'.format(var_comment)
```
mwest1066: To just be explicit about f-strings, they will look like this:
```
var_python_comment = f’# {var_comment}’
```

See 
https://docs.python.org/3/reference/lexical_analysis.html#f-strings

Also: 
https://realpython.com/python-f-strings/

exampleCourse/courseInstances/Sp15/assessments/plelementexamples/infoAssessment.json
```diff
@@ -1,45 +1,46 @@
...
-                {"id": "examplesCheckbox",           "points": 2, "maxPoints": 10},
-                {"id": "examplesStringInput",        "points": 2, "maxPoints": 10},
-                {"id": "elementsMatrixLatexDisplay", "points": 2, "maxPoints": 10},
-                {"id": "examplesMatrixInput",        "points": 2, "maxPoints": 10},
-                {"id": "examplesSymbolicInput",      "points": 2, "maxPoints": 10},
-                {"id": "examplesNumberInput",        "points": 2, "maxPoints": 10}
-            ]
-        }
-    ]
+  "uuid": "69aa799b-9492-4db2-9102-7753263f3863",
```
mwest1066: This whole file had its indenting changed, which doesn’t seem necessary.

question-servers/freeformPythonLib/prairielearn.py
mwest1066: Adding an option to switch between sigfig digits and digits-after-decimal would be really nice. I think we have examples in other elements of doing this, and if possible we should copy the attribute names and code (I think this will be fairly easy).

exampleCourse/courseInstances/Sp15/assessments/plelementexamples/infoAssessment.json
eliving2: Yeah, that was an accident, I auto-indented. Do you want me to try to revert it?

elements/pl-matrix-output/pl-matrix-output.py
eliving2: My Google search was very bad. Fixed now.

elements/pl-matrix-output/pl-matrix-output.py
eliving2: There is a block of code after this that sets the default tab to be the active tab. 

This code block is meant to set the first existing tab to be active. So if there is a Mathematica and Python tab, Mathematica will be the active tab. However, if the user specifies that Python should be the active tab, it should skip this code block.

elements/pl-variable-output/pl-variable-output.py
```diff
@@ -0,0 +1,126 @@
...
+        default_tab = tab_list[default_tab_index]
+    else:
+        first_display = displayed_tab.index(True)
+        default_tab = tab_list[first_display]
+    default_tab_index = tab_list.index(default_tab)
+
+    # Active tab should be the default tab
+    default_tab_list = [False, False, False]
+    default_tab_list[default_tab_index] = True
+    [active_tab_matlab, active_tab_mathematica, active_tab_python] = default_tab_list
```
mwest1066: This new array-based code is really nice! Great job!

elements/pl-variable-output/pl-variable-output.py
```diff
@@ -0,0 +1,126 @@
...
+
+    # Process parameter data
+    matlab_data = ''
+    mathematica_data = ''
+    python_data = 'import numpy as np\n\n'
+    for child in element:
+        if child.tag == 'variable':
+            # Raise exception of variable does not have a name
```
mwest1066: "of" -> "if"

elements/pl-variable-output/pl-variable-output.py
```diff
@@ -0,0 +1,126 @@
...
+
+            # Create string for matlab and python format
+            var_name_disp = pl.inner_html(child)
+            var_matlab_data = pl.string_from_2darray(var_data, language='matlab', digits=var_digits)
+            var_mathematica = pl.string_from_2darray(var_data, language='mathematica', digits=var_digits)
+            var_python_data = pl.string_from_2darray(var_data, language='python', digits=var_digits)
+
+            matlab_data += f'{var_name_disp} = {var_matlab_data}; {var_matlab_comment}\n'
+            mathematica_data += f'{var_name_disp}{mathematica_suffix} = {var_mathematica}; {var_mathematica_comment}\n'
+            python_data += f'{var_name_disp} = {prefix}{var_python_data}{suffix} {var_python_comment}\n'
```
mwest1066: It's a really minor point, but the current comment variable system will produce a trailing space if there is no comment. To avoid this, we could make the `var_matlab_comment` variable include a space at the start of it (if it's not the empty string) and change the template to end with `{var_matlab_data};{var_matlab_comment}` (no space between the semicolon and the comment variable.

exampleCourse/questions/examplesVariableOutput/server.py
```diff
@@ -0,0 +1,27 @@
...
+    data["params"]["a"] = a
+    data["params"]["b"] = b
+    data["params"]["c"] = c
+    data["params"]["d"] = d
+    data["params"]["x"] = x
+    data["params"]["x1"] = x1
+    data["params"]["x2"] = x2
+    data["params"]["xC"] = xC
+
+    # Release correct answer
```
mwest1066: spurious comment here?

question-servers/freeformPythonLib/prairielearn.py
```diff
@@ -341,6 +351,10 @@ def string_from_2darray(A, language='python', presentation_type='f', digits=2):
 
         [ ... ... ; ... ... ]
 
+    If language is 'mathematica' and A is a 2D ndarray, the string looks like this:
+
+        {{ ..., ... },{ ..., ... }}
+
```
mwest1066: does the comment below need updating now that 1D arrays are supported?

question-servers/freeformPythonLib/prairielearn.py
```diff
@@ -375,8 +389,23 @@ def string_from_2darray(A, language='python', presentation_type='f', digits=2):
...
+                'float_kind': lambda x: '{:.{digits}{presentation_type}}'.format(x, digits=digits, presentation_type=presentation_type),
+                'complex_kind': lambda x: '{:.{digits}{presentation_type}}'.format(x, digits=digits, presentation_type=presentation_type)
+            }
+        result = np.array2string(A, formatter=formatter, separator=', ').replace('\n', '')
+        result = result.replace('[', '{')
+        result = result.replace(']', '}')
+        return result
     else:
-        raise Exception('language "{:s}" must be either "python" or "matlab"'.format(language))
+        raise Exception('language "{:s}" must be either "python","matlab", or "mathematica"'.format(language))
```
mwest1066: needs a space between the comma and `"matlab"`

question-servers/freeformPythonLib/prairielearn.py
mwest1066: actually, it looks like the comments above need updating as well

mwest1066: And the function name should not be `string_from_2darray()` anymore. Maybe `string_from_array()`?

mwest1066:
@tbretl could you take a look at the code in this PR that adds 1D array output to `prairielearn.py`? I suggested up-thread that we rename `string_from_2darray()`, so any suggestions about what to call it now would also be good.

Issue #1091, opened 2018-04-02
generate_uuids.py adds uuid to course elements


echuber2:
Running `generate_uuids.py -n .` tends to naively add a uuid property to e.g. `element/course_element/info.json` which results in Load From Disk throwing an error.

mwest1066:
Fixed by #1454

PR #1454, opened 2019-03-09
Closes generate_uuid bug #1091 and added verbose flag

Pavitra122:
Fixed a bug in tools/generate_uuid.py that added a uuid for files in the element subdirectory. 

Also added a verbose flag that can be passed as a command line argument to print all the files that were changed by the script to help in debugging.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -151,9 +151,9 @@
 
   * Remove old temporary upgrade flag `tmp_upgraded_iq_status` (Matt West).
 
-  * Fixed a bug in tools/generate_uuid.py that added a uuid for files in the element subdirectory (Pavitra Shadvani).
+  * Fix `tools/generate_uuid.py` to not add UUID in element subdirectory (Pavitra Shadvani).
 
-  * Added a verbose flag to tools/generate_uuid.py to show all the files changed.(Pavitra Shadvani).
+  * Add verbose flag to `tools/generate_uuid.py` to show all the files changed by script (Pavitra Shadvani).
 
 * __3.1.0__ - 2018-10-08
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -45,6 +45,8 @@
 
   * Add ability to highlight the background of specific lines of text in `pl-code` (Nathan Walters).
 
+  * Add verbose flag to `tools/generate_uuid.py` to show all the files changed by script (Pavitra Shadvani).
+  
   * Change "Save & Grade" button text and alignment (Dave Mussulman).
 
   * Change Ace editor to use source files from npm and upgrade to 1.4.1 from 1.2.8 (Nathan Walters).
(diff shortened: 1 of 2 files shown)


ChangeLog.md
```diff
@@ -151,6 +151,10 @@
 
   * Remove old temporary upgrade flag `tmp_upgraded_iq_status` (Matt West).
 
+  * Fixed a bug in tools/generate_uuid.py that added a uuid for files in the element subdirectory (Pavitra Shadvani).
+
+  * Added a verbose flag to tools/generate_uuid.py to show all the files changed.(Pavitra Shadvani).
+
```
mwest1066: Start ChangeLog entries with exactly "Fix", "Add", etc, as already there.

Group all the "Fix" entries together, all the "Add" entries together, etc.

Put backticks around filenames (the ChangeLog is Markdown).

You have an extra period before the `(` character in the last entry above.

Capitalize UUID.

tools/generate_uuids.py
```diff
@@ -4,10 +4,12 @@
 
 parser = argparse.ArgumentParser(description="Generate UUIDs for info*.json files that don't already have them.")
 parser.add_argument("directory", help="the directory to search for info*.json files")
+parser.add_argument("-v", "--verbose", action="store_true", help="List all the files for which uuid's were generated")
```
mwest1066: UUID should be capitalized.

No apostrophe in `UUIDs` (it's a plural, not a possessive).

Argument help should start with a lowercase character.

tools/generate_uuids.py
```diff
@@ -4,10 +4,12 @@
...
 parser.add_argument("-n", "--new", action="store_true", help="generate new UUIDs for all files, even if they already have one")
 args = parser.parse_args()
 
-skip_dirs = [".git"]
+# Added elements in the skip directory because elements should not contain a uuid
```
mwest1066: Don't comment on what was changed. Comments should reflect the current status of the file. It's also a bit redundant to say that we are skipping `elements` because they shouldn't contain UUIDs. I suggest simply deleting this comment.

tools/generate_uuids.py
```diff
@@ -4,10 +4,12 @@
...
+skip_dirs = [".git","elements"]
```
mwest1066: Follow PEP8 style guidelines: https://www.python.org/dev/peps/pep-0008/

In this case, space after comma.

tools/generate_uuids.py
```diff
@@ -47,6 +49,8 @@ def add_uuid_to_file(filename):
             out_f.write(new_contents)
         os.remove(filename) # needed on Windows
         os.rename(tmp_filename, filename)
+        if  args.verbose:
+            print("Created a UUID for ",filename)
```
mwest1066: space after comma

PR #1459, opened 2019-03-12
Fix gradebook download link for special characters

nwalters512:
Course names like "STAT/CS/IS 107" and "TAM 210/211" would have gradebook download links with the `/` character in the link, which would confuse the Express router. This adopts a similar approach to that used in the assessment downloads page and replaces any character not in `[a-zA-Z0-9]` with an underscore.

### DIFF
--- a/pages/instructorGradebook/instructorGradebook.js
+++ b/pages/instructorGradebook/instructorGradebook.js
@@ -5,18 +5,14 @@ var express = require('express');
 var router = express.Router();
 
 var error = require('@prairielearn/prairielib/error');
+const sanitizeName = require('../../lib/sanitize-name');
 var sqldb = require('@prairielearn/prairielib/sql-db');
 var sqlLoader = require('@prairielearn/prairielib/sql-loader');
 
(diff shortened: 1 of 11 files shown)


pages/instructorGradebook/instructorGradebook.js
```diff
@@ -10,10 +10,12 @@ var sqlLoader = require('@prairielearn/prairielib/sql-loader');
 
 var sql = sqlLoader.loadSqlEquiv(__filename);
 
+const sanitizeStringForUrl = (str) => str.replace(/[^a-zA-Z0-9]/g, '_');
```
wadefagen: It would be nice to retain `-` instead of making it an underscore (ex: "CS 296-25").  This also more consistently maintains the existing behavior: `str.replace(/[^a-zA-Z0-9-]/g, '_');`

mwest1066: @nwalters512 I'm on board with Wade's suggestion to include hyphens in the whitelist, but if you make this change then we should probably also change the equivalent whitelist in `assessment.js` for consistency.

PR #1463, opened 2019-03-14
Added in-browser option to pl-file-download.

shreyas208:
Added a `force-download` boolean attrib to `pl-file-download`. Default is true to maintain current behavior. If false, files are opened in a new tab without the `download` HTML attribute.


coatless:
@shreyas208 need to add the new attribute in the `prepare()` function under `optional_attribs=` parameter, c.f.

https://github.com/PrairieLearn/PrairieLearn/blob/afcf81f420277b2e98d36eb97c908b4c9d598967/elements/pl-file-download/pl-file-download.py#L8

coatless:
@shreyas208 also need to add a line to `ChangeLog.md`:

https://github.com/PrairieLearn/PrairieLearn/blob/ae525495e8a130627c6a8c6d8efdd1c3e7dac0ef/ChangeLog.md

Probably:

> Added `force-download` attribute to `pl-file-download` to specify if the file should be  downloaded or is able to be viewed in browser (Your Name)."

mwest1066:
@shreyas208 can you also add an example of using this to `exampleCourse/questions/downloadFile`?

mwest1066:
@shreyas208 thanks for the PR!

PR #1473, opened 2019-03-31
Issue should cause test to fail

### LINKED ISSUE #1471: should fail when issues are logged in npm test
One test is to check that a set of questions in `exampleCourse` load and render (see `testQuestions.js`, which calls `helperQuestion.js`).

This test should fail if an issue is logged. It doesn't.

As an example, add `raise Exception()` to line 13 of `pl-matrix-output.py` (the first line of the `render` function). On load of `multiplyTwoMatrices`, an issue ("Error rendering question") is logged, but the test does not fail.

Probably a copy/paste of the `should produce no issues` test from later in `helperQuestion.js` will fix this issue.

tbretl:
Added a test to `autoTestQuestion()` in `helperQuestion.js` that will fail if issues are generated on question load.

Fixes #1471


mwest1066:
Awesome catch! It’s very bad when we discover that our tests were not actually testing things 🙂

PR #1474, opened 2019-04-01
Add native Markdown support

nwalters512:
This PR adds Markdown support in questions. This is done with the `<markdown>` tag. For instance, in my question, I could have:

`````html
<pl-question-panel>
<markdown>
# Hello world!

```cpp
this is code!
```
</markdown>
</pl-question-panel>
`````

This would be transformed into the following before it went down the normal question pipeline:

```html
<pl-question-panel>
<h1>Hello world!</h1>
<pre><code class="language-cpp">this is code!</code></pre>
</pl-question-panel>
```

The implementation is simple: do a non-greedy regex for `<markdown>(.*)</markdown>`, transform the contents into HTML, and replace the original string with that. The only kind of weird bit is to account for cases where we want to use markdown inline in other elements, like:

```html
<pl-multiple-choice ...>
<pl-answer><markdown>**Hello!*</markdown</pl-answer>
<pl-answer><markdown>**Goodbye!**</markdown></pl-answer>
</pl-multiple-choice>
```

By default, the rendered HTML is wrapped in `<p>...</p>`, which breaks formatting of the resulting checkboxes since `<p>`, a block element, is attempted to be displayed inside an inline-block element next to a bare string (`(a)`, etc.). To account for that, I strip the `<p>` off rendered text if there's exactly one `<p>` block in the rendered HTML. This is just one way of fixing this - a better way is probably to just *allow* block content inside of multiple choice answers. In fact, writing this all out, that's exactly what we should probably do lol. @mwest1066 agree?

**TODO**
- [x] Docs


trombonekenny:
I think if you do something non-intuitive to markdown stripping (or any filtering/rendering) you're going to confuse things later. That special edge case probably isn't alone. I'd vote for leaving the markdown tag to a vanilla environment (through whatever markup library you're using) and fixing the wrapping issues inside other elements as they come up.

nwalters512:
Fair - that was mostly an easy way to get it working for multiple choice answers, but the right way to fix that is to let the answers render block elements.

trombonekenny:
Why `<markdown>` instead of `<pl-markdown>`? Are doing things that make that parser our own? Is there a standard for this?

nwalters512:
> Why `<markdown>` instead of `<pl-markdown>`?

`<pl-*>` only refers to elements so far. We could use `<pl-markdown>`, but it's not an element and I don't want users getting confused about that.

> Are doing things that make that parser our own?

The only thing we're doing is rendering triple-backtick code fences as `<pl-code>` elements.

> Is there a standard for this?

If by this you mean "embedding markdown in HTML", then not really. Piazza actually does largely the same thing we're doing here (minus escaping, and we're doing parsing better than them).

PR #1476, opened 2019-04-09
Update to support uid change with same uin

trombonekenny:
Changed the sproc to support updating `uid` changes if the `uin` is consistent.
Added sproc tests for `users_select_or_insert`


trombonekenny:
I confirmed that the updated function will let us create a "stub" user account (without a netid or proper name, but with correct UIN) and it gets updated when the user first logs in. This would be useful for the freshmen proficiency testing and letting proctors create "stub" accounts.

@mwest1066 review?

PR #1478, opened 2019-04-12
Disable 'Test 100 times' button for externally graded questions

nwalters512:
Because any button that allows a single person to slow our grading throughput to a crawl is Bad.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -103,6 +103,8 @@
 
   * Change element documentation to have a separation between submission and format elements (James Balamuta).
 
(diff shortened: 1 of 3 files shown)


trombonekenny:
It would be safer to trap that at the post handler too. Hiding the button doesn't hide the functionality.

mwest1066:
@trombonekenny that’s true, but it’s not a security issue or anything, it’s just slightly annoying — if someone wants to hand-POST to the endpoint, I can’t be bothered to stop them 🙂

PR #1489, opened 2019-04-21
Fix how binary files are handled during external grading

nwalters512:
For text files, eagerly decoding the base64 contents to a string was fine. For binary files (tar archives, specifically) that breaks things. We now decode the files into a `Buffer` and write that directly to disk.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -195,6 +195,8 @@
 
   * Fix (or at least attempt to) S3 file uploads for external grading (Nathan Walters).
 
+  * Fix handling of binary files during external grading (Nathan Walters).
+
   * Remove `allowIssueReporting` option in `infoCourseInstance.json` (Matt West).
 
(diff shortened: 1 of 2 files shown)


trombonekenny:
Is this a good place to hook in a test case for external grading that the supplied file actually matches what's seen on the grader?

nwalters512:
We should definitely have a test case for that, though I'm still not sure what the best way to do that kind of integration test for external grading would be. Testing async things is tricky.

nwalters512:
To be fair, we could probably just test `buildDirectory` directly if we can get the right data in place beforehand.

nwalters512:
Scratch that, the function doesn't even read from the database! I'll pretend like I thought ahead to make it easily unit-testable. Let me see if I can write some tests for this.

nwalters512:
I'll punt the tests to another branch, don't know how long those will take me and I want to ship this soon for STAT 430.

PR #1495, opened 2019-05-15
Use new JSON schema validator and publish schemas to npm

nwalters512:
I recently moved all of our schemas to https://github.com/PrairieLearn/schemas so that they can be consumed by other tools/components/etc. This PR switches us to relying on those schemas and removes the ones from this repository. This also switches us to using [`ajv`](https://github.com/epoberezkin/ajv) to validate JSON with our schemas. The combination of no longer reading the schemas once per question/assessment/etc. and ajv being significantly faster and caching compiled validation code means that I saw a roughly 50% speedup (~5s to ~2.5s) in the "Loading info.json files from git repository..." phase of sync for CS 225, a pretty large course repo.


ChangeLog.md
```diff
@@ -113,6 +113,8 @@
   
   * Change element documentation to follow a common structure (James Balamuta).
 
+  * Change JSON validation to use schemas from the [`@prairielearn/schemas`](https://github.com/PrairieLearn/schemas) package (Nathan Walters).
```
trombonekenny: What's the rational for a new repository, rather than fixing in PrairieLearn or moving to something like PrairieLib? If you're moving the schemas, why not move the code bits too?

trombonekenny:
That sounds great! I know the courses will definitely appreciate sync speed improvements! Thanks for putting this together!

ChangeLog.md
nwalters512: The schemas and code bits are only really "related" at this point because of how they're used. There are at least three other places I can think of off the top of my head that people might want to consume them from:

* Our VSCode extension
* The online editing experience for in-browser validation
* A custom course CI setup that validates file against our schemas

In those cases, they might want just the schemas without the (minimal) code that does validation and file reading; the first two definitely doesn't need it at all. That's why I opted a) for a second package, and b) to not collocate the validation code with the schemas.

We could definitely move the validation bits to PrairieLib too if you feel they'd be useful in the scheduler or the like. But our functions are already really thin wrappers around standard Node file APIs and a JSON schema validator, of which there are many. I don't know if there's a ton of value in exposing our own functions vs. letting others just pick the tools and read the documentation they want.

lib/course-db.js
```diff
@@ -236,21 +241,25 @@ module.exports.loadFullCourse = function(courseDir, logger, callback) {
...
     var defaultCourseInstanceInfo = {};
     var defaultAssessmentInfo = {};
+    const startTime = new Date();
     async.series([
         that.loadCourseInfo.bind(that, course.courseInfo, courseDir, logger),
         function(callback) {
-            that.loadInfoDB(course.questionDB, 'qid', course.courseInfo.questionsDir, 'info.json', defaultQuestionInfo, 'schemas/infoQuestion.json', 'schemas/questionOptions', '.json', course.courseInfo, logger, function(err) {
+            const questionLoadStart = new Date();
+            that.loadInfoDB(course.questionDB, 'qid', course.courseInfo.questionsDir, 'info.json', defaultQuestionInfo, schemas.infoQuestion, 'questionOptions', course.courseInfo, logger, function(err) {
+                logger.info(`Question info elapsed time: ${(new Date()) - questionLoadStart}ms`);
```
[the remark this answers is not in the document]
nwalters512: Oh this shouldn't even have been left in here - thanks for catching this.

nwalters512:
Yeah, I think tests are definitely reasonable. I can work on that before we merge this.

nwalters512:
After talking with @mwest1066, we decided it would be best if the schemas were still located in the PrairieLearn repo but were then published from within the repo so other things could still consume them. I've updated the PR to reflect this new approach.

schemas/README.md
```diff
@@ -0,0 +1,2 @@
+# schemas
+Schemas for various PrairieLearn JSON files. They are designed to be published independently of PrairieLearn for consumption by other tools or applications. They are published on npm in the `@prairielearn/schemas` package.
```
coatless: @nwalters512 could you extend the README to note the procedure to update the schemas on `npm`?

mwest1066:
Could we leave the schema files with a `.json` extension? I think that would more clearly signal that the schemas are themselves just JSON objects. All the other code can be pretty much identical — we can still `require()` the files with a `.json` extension.

doc/assessment.md
```diff
@@ -19,7 +19,7 @@ Each assessment is a single directory in the `assessments` folder. The directory
 }
 ```
 
-* [Format specification for assessment `infoAssessment.json`](https://github.com/PrairieLearn/PrairieLearn/blob/master/schemas/infoAssessment.json)
+* [Format specification for assessment `infoAssessment.json`](https://github.com/PrairieLearn/PrairieLearn/blob/master/schemas/schemas/infoAssessment.js)
```
mwest1066: I’m afraid that the doc links also need to be changed back to `.json`

PR #1496, opened 2019-05-24
Add "clientFilesElement" folder for loading element files

nicknytko:
Will be helpful for CS357/TAM251 for loading images and icons for an element being developed.

Files can be put into a `clientFilesElement` folder located in the root of a course element, and accessed from the base URL `/pl/course_instance/:course_instance_id/instructor/element/:element_id/clientFilesElement/`.

The `customElement` question was updated as an example:
Screen Shot 2019-05-24 at 9 45 42 AM


trombonekenny:
Nick, forgive me for the silly question as I haven't done much with custom elements but is there a reason you need an element for this? Can you describe more about what you're trying to do that doesn't work with the existing clientFiles or clientFilesCourse dirs?

nicknytko:
> Nick, forgive me for the silly question as I haven't done much with custom elements but is there a reason you need an element for this? Can you describe more about what you're trying to do that doesn't work with the existing clientFiles or clientFilesCourse dirs?

Working on an existing element that needs to load external files for button icons, backgrounds, etc. and would rather have these resources in one place tied to the element code than in the course files.  Though, if you think it's unnecessary I can just leave these files in clientFilesCourse.

trombonekenny:
I wonder if we should just adopt a clientFiles directory under an element dir, and make that dir work with https://github.com/PrairieLearn/PrairieLearn/blob/master/pages/elementFiles/elementFiles.js ? i.e. widening the scope of element included paths from CSS and JS files.

exampleCourse/elements/course-element/course-element.py
```diff
@@ -9,7 +9,8 @@ def prepare(element_html, data):
 
 def render(element_html, data):
     html_params = {
-        'number': data['params']['random_number']
+        'number': data['params']['random_number'],
+        'image_url': data['options']['elements_url'] + '/course-element/clientFilesElement/block_i.png'
```
trombonekenny: I wonder if there's a way to do this that automatically maps the `elements_url` to include the element name folder and clientFilesElement without having to hard code it in the element in the element python code. i.e. 
'image_url': data['options']['elements_url'] + '/block_i.png'

pages/elementFiles/elementFiles.js
```diff
@@ -4,21 +4,25 @@ const router = express.Router();
...
+    const valid = pathSpl[1] == CLIENT_FOLDER ||
+          _.some(EXTENSION_WHITELIST, (extension) => filename.endsWith(extension));
     if (!valid) {
         res.status(404);
         const err = new Error('Unable to serve that file');
         err.status = 404;
         return next(err);
     }
-
+
```
nwalters512: nit: extra spaces

exampleCourse/elements/course-element/course-element.py
nwalters512: +1 - this would require mutating `data.options` during HTML traversal which we don't currently do for any other case, but I don't see a problem with that. Avoiding hardcoding element names in element code would make them easier to copy/fork/rename/etc.

question-servers/freeform.js
```diff
@@ -165,7 +165,12 @@ module.exports = {
             const resolvedElement = module.exports.resolveElement(elementName, context);
             const cwd = resolvedElement.directory;
             const controller = resolvedElement.controller;
-            const pythonArgs = [elementHtml, data];
+
+            let dataCopy = _.cloneDeep(data);
+            if ('base_url' in data.options) {
+                dataCopy.options.client_files_element_url = path.join(data.options.base_url, "elements", elementName, 'clientFilesElement');
```
nwalters512: when running on Windows, this would give us backslashes in a URL: https://nodejs.org/api/path.html#path_windows_vs_posix. consider `path.posix.join` or just a template literal like
```js
`${data.options.base_url}/elements/${elementName}/clientFilesElement`
```

question-servers/freeform.js
```diff
@@ -165,7 +165,12 @@ module.exports = {
             const resolvedElement = module.exports.resolveElement(elementName, context);
             const cwd = resolvedElement.directory;
             const controller = resolvedElement.controller;
-            const pythonArgs = [elementHtml, data];
+
+            let dataCopy = _.cloneDeep(data);
+            if ('base_url' in data.options) {
```
nwalters512: is there ever a case where `base_url` wouldn't be present?

nicknytko: I'm not familiar with the specifics of how the PL internals work, but I've found `elementFunction` gets called occasionally with a blank `data.options` so the check is there so nothing weird happens

question-servers/freeform.js
```diff
@@ -165,7 +165,12 @@ module.exports = {
...
+                dataCopy.options.client_files_element_url = path.posix.join(data.options.base_url, 'elements', elementName, 'clientFilesElement');
```
nwalters512: nit: a comment about why we use `path.posix` specifically would be good for future maintainers, as it might not be immediately obvious

PR #1499, opened 2019-06-13
Adding modal dialogue support to enroll.ejs

echuber2:
Add/remove modal popup will replace the old popover, if enabled. This should have better behavior on some touchscreen devices, as well as offering more verbose warnings about what it means to add/remove a course on PL.

~(There is a toggle in the EJS code, `var useModals=true`, so it's easy to undo this if it seems to cause some unforeseen problems, or if A/B testing is necessary.)~ (Removed popover cruft as per Dave Mussulman's suggestion on Slack.)

The more verbose messages clearly state that add/remove on PL does not register for or drop courses with the University:

> ~Are you sure you want to add this course content to your PrairieLearn account? This does not register you in the actual course. You need to formally register for courses with the University separately.~

> ~Are you sure you want to remove this course content from your PrairieLearn account? This does not drop you from the actual course. You need to formally drop classes through the University registration system.~

(revised:)

> Are you sure you want to add this course content to your PrairieLearn account? Adding or removing courses here only affects what is visible to you on PrairieLearn. This does not change your university course registration.

> Are you sure you want to remove this course content from your PrairieLearn account? Adding or removing courses here only affects what is visible to you on PrairieLearn. This does not change your university course registration.

~I'm not sure if this part of the "remove" warning message is necessary:~ (removed this as per suggestion)
> ~If you have recently used this PrairieLearn course to take an exam for proficiency, placement, or entrance qualification, please do not remove yourself from the course.~

I imagine some students removing themselves from a course after taking a one-time exam like a placement exam. I don't think this affects our records of their results, right? However, we may need to make new instances on a monthly basis to control how often students can re-take certain exams. In that case, maybe it's more clear from a UX perspective if the students can see which instances are the ones they _haven't_ previously taken, at a glance. So, they should just leave themselves added to past courses.

### DIFF
--- a/pages/enroll/enroll.ejs
+++ b/pages/enroll/enroll.ejs
@@ -2,6 +2,12 @@
 <html>
   <head>
     <%- include('../partials/head'); %>
+
+    <style>
+      .modal-course-title-style {
+        font-size: 125%;
+      }
+    </style>
   </head>
   <body>
     <script>
@@ -15,6 +21,17 @@
         <div class="card-header bg-primary text-white">Courses</div>
         <table class="table table-sm table-hover table-striped">
           <tbody>
+            <%# If useModals=true, then the UI will use modal confirmation
+                dialogs. If false, then the old popover tooltips with buttons
+                will be generated instead. %>
+            <% var useModals=true; %>
+            <%# Modal configuration preamble. This has to come before the
+                following forEach: %>
+            <%
+              var modalIdx=0;
+              var dataToggleVal="popover";
+              if (useModals) {dataToggleVal="modal";}
+            %>
             <% course_instances.forEach(function(course_instance) { %>
             <tr>
               <td class="align-middle">
@@ -23,8 +40,10 @@
               <td>
                 <% if (!course_instance.enrolled) { %>
                 <a tabindex="0" class="btn btn-sm btn-info" role="button"
-                   data-toggle="popover" data-trigger="focus" data-container="body"
+                   data-toggle="<%= dataToggleVal %>"
+                   data-trigger="focus" data-container="body"
                    data-html="true" data-placement="auto" title="Confirm add"
+                   data-target="#addModal<%= modalIdx %>"
                    data-content="
                                  <form name=&quot;enroll-form&quot; method=&quot;POST&quot;>
                                    <input type=&quot;hidden&quot; name=&quot;__action&quot; value=&quot;enroll&quot;>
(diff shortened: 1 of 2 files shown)


echuber2:
enroll_modal_add

enroll_modal_remove

echuber2:
Made revisions as per suggestions from @trombonekenny on Slack

echuber2:
For the sake of completeness, here's the behavior of the previous Boostrap popover UI with a button placed on the tooltip on Linux, for both Firefox and Chromium: If the user quickly taps the popover button, it works as expected. However, a long press dismisses the tooltip without activating the button. I assume that for some devices/drivers, a click or tap was interpreted as a long press.

Buttons on Bootstrap modal dialogs fortunately do not have this problem at all.

pages/enroll/enroll.ejs
```diff
@@ -15,53 +21,93 @@
...
+                      </div>
+                    </div>
+                  </div>
+                </div>
                 <% } %>
               </td>
             </tr>
-            <% }); %>
+            <%# incrementing modalIdx must be done before closing the forEach function block: %>
+            <% ; modalIdx++; }); %>
```
mwest1066: Why do we have a leading semi-colon here?

pages/enroll/enroll.ejs
```diff
@@ -15,53 +21,93 @@
         <div class="card-header bg-primary text-white">Courses</div>
         <table class="table table-sm table-hover table-striped">
           <tbody>
-            <% course_instances.forEach(function(course_instance) { %>
+            <%# var modalIdx=0; must come before the forEach: %>
```
mwest1066: While we are switching to modals, we could also switch to using just one copy of each modal at the top of the page, and then call them with parameters. For an example of this, see the modals that @nwalters512  cleaned up on https://github.com/PrairieLearn/PrairieLearn/blob/master/pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs

I don’t insist on doing this, and I’m happy to merge the code as is, but I think this modal-with-parameters approach that @nwalters512 used is generally cleaner than the way I was doing it originally. It also makes the HTML much smaller, so faster page loads! And it means we don’t need the `modalIdx` variable.

mwest1066:
@echuber2 very nice PR! I’m so happy that you tracked down the “can’t add course” bug! That is this the longest unfixed bug in PrairieLearn, so you should win some some sort of award for fixing it.

pages/enroll/enroll.ejs
echuber2: I think that's left over from an earlier draft of the code where it seemed to make the EJS easier to read/more defensive against someone putting an IIFE in front of it by accident. We can remove it.

pages/enroll/enroll.ejs
echuber2: That's a good idea. The thought had crossed my mind but I wasn't sure how drastic to make the first draft of this. I can try to amend the PR a little later.

pages/enroll/enroll.ejs
mwest1066: makes sense, removing is probably a good idea

pages/enroll/enroll.ejs
mwest1066: excellent! I’ll hold of on merging this PR for now, if you plan on making that change soon. If you’d rather not make the change and want to merge as is, then just let me know.

echuber2:
Sorry for the delay -- I'll try to implement the parameterized version sometime this week. I'll make a new issue about it.

PR #1018, opened 2018-02-05
Course questions

rbessick5:
Currently, this addresses the first checkpoint of #988 . There are no links to individual questions yet (I need to do the right thing with variants/submisisons that don't belong to a course_instance before this can be added).

Closes #988 , closes #912 , and fixes #1019

PR #1503, opened 2019-06-17
Add course-level Questions page

### LINKED ISSUE #912: Add Questions to Course view
Let people who can view the Course page also see the Questions tab (currently just linked under course instances)

Confirm that viewer level access can see those too.

(Also add a section to Courses that list course instances without having to go back to the main page)

### LINKED ISSUE #988: Compare question use across semesters at course level
Create a "Questions" page for the course view which:

- [x] Lists of all course questions (equivalent to the one available for a course instance)
- [x] Create "courseQuestion" view (like instructorQuestion) and enable the hyperlinks from each question title.
- [x] Displays a column for each course_instance, with the assessment(s) using that question listed underneath (equivalent to the rightmost column of the Questions view in a course instance)
- [ ] (Extra) Allow individual course instances to be toggled on/off in this view, so specific course instances can be compared (for comparisons like, e.g., "last two years" or "semesters I taught the class")
- [x] (Extra) Create "courseIssues" view and enable links on the course navbar and from Questions/Question view.
The first two checkpoints are identical to #912

### LINKED ISSUE #1019: "Issues" badge on instructorQuestion page should pre-filter by qid
Clicking on the badge on instructorQuestions (i.e., the question list) takes us to instructorIssues with a qid filter already applied (to show issues with that question), but doing so on instructorQuestion (i.e., individual question view) does not apply this filter.

rbessick5:
This PR creates a "Questions" page for the course view, along with some supporting pages:
- A courseQuestions page to view all questions, including one column per course instance showing the assessments which use the question.
- A courseQuestion page to view individual question details outside of a course instance
- A courseIssues page to view issues outside a course instance

This replaces PR #1018 (which got broken during a rebase); that PR should go away.

This does everything in #988 except a view/hide toggle on individual course instances (so you can compare a subset of course instances like "last two years" or "times I taught"). I'd suggest that get spun out into a new Enhancement issue.

Closes #988, closes #912, fixes #1019

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -72,6 +72,8 @@
   * Add example questions for `pl-multiple-choice` and `pl-integer-input` customizations (James Balamuta).
 
   * Add `clientFilesElement` folder for loading element-specific client files (Nicolas Nytko).
+  
+  * Add course-level pages for viewing all questions, individual questions, and issues (Ray Essick).
 
   * Change "Save & Grade" button text and alignment (Dave Mussulman).
 
--- a/middlewares/selectAndAuthzCourseQuestion.js
+++ b/middlewares/selectAndAuthzCourseQuestion.js
@@ -1,0 +1,21 @@
+var ERR = require('async-stacktrace');
+var _ = require('lodash');
+
+var sqldb = require('@prairielearn/prairielib/sql-db');
+var sqlLoader = require('@prairielearn/prairielib/sql-loader');
+var error = require('@prairielearn/prairielib/error');
+
+var sql = sqlLoader.loadSqlEquiv(__filename);
+
+module.exports = function(req, res, next) {
+    var params = {
+        question_id: req.params.question_id,
+        course_id: res.locals.course.id,
+    };
+    sqldb.queryZeroOrOneRow(sql.select_and_auth, params, function(err, result) {
+        if (ERR(err, next)) return;
+        if (result.rowCount == 0) return next(error.make(403, 'Access denied'));
+        _.assign(res.locals, result.rows[0]);
+        next();
+    });
+};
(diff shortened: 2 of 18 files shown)


rbessick5:
There is significant code duplication in this PR, as I've essentially created parallel instructor* and course* versions of several pages. I can (probably should) DRY these out in the following way:
- courseIssues and instructorIssues
These are effectively identical in appearance and function. I suggest removing instructorIssues and using courseIssues in all of its routings
- courseQuestion and instructorQuestion
These are nearly identical -- courseQuestion omits the "Assessments" row in the question details (as this is course-instance-specific), and omits the "Tests" row/buttons (these buttons don't work without a valid course instance). I've written courseQuestion to conditionally display these rows whenever course_instance and/or assessments are defined (which happens in instructor view), so we could eliminate instructorQuestion and use courseQuestion.
- selectAndAuthzCourseQuestion and selectAndAuthzInstructorQuestion
The only difference between these two is whether "assessments" is populated (by using a course instance id that exists in the InstructorQuestion version). We could combine these into a single .js/.sql that decides which of the two SQL queries to perform. This (and above) would make the files used by both instructor- and course-routing the same

rbessick5:
In principle, we could also consolidate instructorQuestions and courseQuestions, but right now the former displays only the current course instance 's assessment, while the latter displays all of them. Without a way to control what semesters are visible/hidden, this could be too unwieldy (we talked about this in the thread for #1018 ).

Adding a view/hide toggle would let us make the reasonable-ish choice of "default all course instances visible from course view; default only current course instance visible in instructor view" but let the user turn things on/off to suit them.

jonatanschroeder:
@mwest1066 @rbessick5 based on how the pages work now and the changes in #4653 I believe this PR is outdated, any issues with closing this PR?

mwest1066:
> @mwest1066 @rbessick5 based on how the pages work now and the changes in #4653 I believe this PR is outdated, any issues with closing this PR?

I agree it can be closed in favor of #4653

PR #1504, opened 2019-06-19
Speed up tags syncing

nwalters512:
Speeds up tags syncing by avoiding doing a round-trip to the DB for every tag and question-tag. Tags and question-tags are now batched into just two sproc calls, allowing the bulk of the work to take place at the DB itself. Data is also encoded in arrays instead of objects when being sent to the DB to eliminate some of the overhead imposed by JSON's verbosity.

With this PR, I was able to cut the time for syncing 225's tags from 3700ms to just 84ms on my machine :tada:

This PR was also a chance to introduce a bit more async/await control flow into PL, and I think the logic is much nicer to read. Currently that's done with a few helpers I wrote inline in the tags syncing file, but they can be moved out into PLib at a later time if that's desired.


sprocs/sync_question_tags.sql
```diff
@@ -0,0 +1,35 @@
+-- Accepts a course ID and a list of objects containins qids and tag IDs,
```
mwest1066: I don’t see the course ID being passed here?

sprocs/sync_question_tags.sql
```diff
@@ -0,0 +1,35 @@
...
+        WITH new_tags AS (
+            INSERT INTO question_tags (
+                question_id,
+                tag_id,
+                number
+            ) SELECT
+                (question->>0)::bigint,
+                tag_id::bigint,
+                number
+            FROM JSONB_ARRAY_ELEMENTS_TEXT(question->1) WITH ORDINALITY AS t(tag_id, number)
```
mwest1066: Why do we need the `_TEXT` variant here? This is generally less efficient, I think?

sprocs/sync_question_tags.sql
nwalters512: copy/paste gone wrong 🙂

sprocs/sync_question_tags.sql
nwalters512: AFAIK, the only way to go from something of type `jsonb` to `bigint` is to go through `text` first, hence this. This is the equivalent of what we do at https://github.com/PrairieLearn/PrairieLearn/blob/master/sprocs/authz_assessment.sql#L34.

PR #1506, opened 2019-06-25
Fix hljs syntax highlighting compatibility issue in pl-code.py

echuber2:
hljs has a compatibility issue with pl-code, where a single-line C++ comment can't be parsed correctly if the line is "highlighted" by pl-code and it is not terminated with a newline; the result is that the line following the single-line comment is also marked as a comment by hljs. Inserting a terminating \n on the highlighted line, inside the span, seems to fix this.

GitHub isn't letting me attach an image here but I'll try in a followup.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -204,6 +204,8 @@
   * Remove old temporary upgrade flag `tmp_upgraded_iq_status` (Matt West).
 
   * Remove `string_from_2darray_sf()` from `freeformPythonLib/prairielearn.py` (Liz Livingston)
+  
+  * Fix hljs syntax highlighting compatibility issue in `pl-code.py` (Eric Huber)
 
 * __3.1.0__ - 2018-10-08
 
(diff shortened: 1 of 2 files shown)


echuber2:
Before the fix:
hljsbugbefore

After the fix:
hljs_bug_after

ChangeLog.md
```diff
@@ -204,6 +204,8 @@
   * Remove old temporary upgrade flag `tmp_upgraded_iq_status` (Matt West).
 
   * Remove `string_from_2darray_sf()` from `freeformPythonLib/prairielearn.py` (Liz Livingston)
+  
+  * Fix hljs syntax highlighting compatibility issue in `pl-code.py` (Eric Huber)
```
nwalters512: Nit: changelog entries should be grouped by prefix ("fix") and end with a period

echuber2: OK, recommitted that adjustment

PR #1510, opened 2019-07-01
WIP: Add LTIOutcomeRequired access rule

trombonekenny:
- Stop users that come from LTI with outcomes enabled from browsing to assessments where outcomes are not already present (i.e. they must navigate to the assessments from their origin LMS instead of through PL)

### DIFF
--- a/pages/instructorAssessmentAccess/instructorAssessmentAccess.ejs
+++ b/pages/instructorAssessmentAccess/instructorAssessmentAccess.ejs
@@ -26,6 +26,7 @@
                 <th>Time limit</th>
                 <th>Password</th>
                 <th>CBTF exam</th>
+                <th>LTI Outcome Required</th>
               </tr>
             </thead>
             <tbody>
@@ -40,6 +41,7 @@
                 <td><%= access_rule.time_limit %></td>
                 <td><%= access_rule.password %></td>
                 <td><%- access_rule.exam %></td>
+                <td><%= access_rule.lti_outcome_required %></td>
               </tr>
               <% }); %>
             </tbody>
--- a/pages/instructorAssessmentAccess/instructorAssessmentAccess.sql
+++ b/pages/instructorAssessmentAccess/instructorAssessmentAccess.sql
@@ -33,6 +33,10 @@ SELECT
         ELSE aar.password
     END AS password,
     CASE
+        WHEN aar.lti_outcome_required IS NULL THEN '—'
+        ELSE aar.lti_outcome_required::TEXT
+    END AS lti_outcome_required,
+    CASE
         WHEN aar.exam_uuid IS NULL THEN '—'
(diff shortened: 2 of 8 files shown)


echuber2:
I think there are cases where a student ends up populating both types of user ID record (normal email address vs. on-demand ID). For example, if they first go through Assessments as a side channel to access an assessment, thus resulting in no LTI connection for that homework, but they then go through the LTI link for the same homework, I think it results in two different shadow copies of the same homework for that student, with two different submission statuses. It sounds like students can see one "view" or the other depending on which way they navigate back to the homework.

I'm wondering if this may still not fix the issue. Suppose that if they navigate with the Assessments menu even after going through the LTI link, it still results in creating the two different homework instances for the same user. Then instead of seeing a list with missing assessments as this proposed warning message would indicate, the student _would_ still see it listed, but the link would lead to the wrong instance.

The other solution that we had previously discussed would be to somehow restrict the UI entirely when students first navigate through LTI. That is, disallow them the ability to use the back hallway at all when they arrive through LTI.

trombonekenny:
> I think there are cases where a student ends up populating both types of user ID record (normal email address vs. on-demand ID). For example, if they first go through Assessments as a side channel to access an assessment, thus resulting in no LTI connection for that homework, but they then go through the LTI link for the same homework, I think it results in two different shadow copies of the same homework for that student, with two different submission statuses. It sounds like students can see one "view" or the other depending on which way they navigate back to the homework.

Do you have example students/situations to share for me to investigate? (Feel free to email or slack instead of posting them publicly here.) The only reasoning I can figure that would make two different assessments depending on how the student accesses things is if they are logging into PL and accessing the course instance using a different authentication source (google or Illinois, etc.) and also LTI. You would be able to confirm that in the gradebook; they would be multiple students listed. If that's so, and not desired, the right way to fix that is to adjust the course instance allowAccess to only allow `"institution": "LTI"`

> The other solution that we had previously discussed would be to somehow restrict the UI entirely when students first navigate through LTI. That is, disallow them the ability to use the back hallway at all when they arrive through LTI.

Yeah, I think that's reasonable but I'm not sure how to best configure it. This PR is thinking about that as a per-assessment basis but while I was writing it I started thinking about that as a per-course-instance configuration instead. I'm not sure the best/easiest route to continue.

mwest1066:
@trombonekenny can you describe a bit more what you are thinking with the per-assessment versus per-course-instance restrictions?

trombonekenny:
@nwalters512 This can be archived. We've lived this long with the way LTI 1.1 works in PL that I'm not sure this is worth continuing. We've had later work that tries to handle cases when LTI users try to do what traditional login users do, and have error messages there. I don't think it's worth cluttering up the access rules at this point, unless we do it as a part of 1.3.

PR #1514, opened 2019-07-02
Ignore string case (close #1511)

### LINKED ISSUE #1511: Case sensitivity for string input
Would it be possible to add a flag for whether a string input should be case sensitive or not? We have some situations where case shouldn't matter and others where it does.

coatless:
Adds the ability to disable case comparison on `pl-string-input`.

/ cc @glherman


exampleCourse/questions/examplesStringInput/question.html
```diff
@@ -66,3 +66,17 @@
 <pl-string-input answers-name="ans5" label="blank input box = " allow-blank="true">
 </pl-string-input>
 </p>
+
+<pl-question-panel>
+<p> 6) To allow for case insensitivity, use <code>ignore-case="true"</code>. 
+    This will autoconvert both the student answer and the correct answer to lower case. </p>
```
mwest1066: Can we add an instruction here on how to answer this question? The questions above all do this. For example, we could add “Enter the string below with any choice of uppercase and lowercase letters.”

coatless:
```suggestion
Enter the string below with any choice of uppercase and lowercase letters.
```

PR #1517, opened 2019-07-03
Let generate_uuids.py find uppercase UUIDs

echuber2:
I encountered this issue recently. The script wasn't matching UUIDs with uppercase letters although it could see the `uuid` property was there, failing with an error message.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -191,9 +191,9 @@
 
   * Fix editing popovers to work with the new sanitization defaults (Matt West).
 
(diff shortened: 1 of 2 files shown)


echuber2:
@mwest1066  I'm feeling deja vu about this. Is there an intentional reason why the script was set to fail to match when uppercase UUIDs were present in json files?

mwest1066:
@echuber2 no reason this was like it was. Just an oversight. Thanks for fixing!

PR #1581, opened 2019-07-10
Syncing rewrite

### LINKED ISSUE #473: Remove question_tags.number and sort by tags.number instead

### LINKED ISSUE #1564: Make UUIDs only locally unique
As @wadefagen and others have pointed out, there is no need to have UUIDs be globally unique and it only causes headaches for instructors when they are copying things around. We should make UUIDs unique only within the smallest possible local scope, which is the course for questions and course_instance for assessments.

### LINKED ISSUE #1565: Auto-create tags and topics as needed
For new instructors copying example questions, it’s a significant barrier that tags and topics need to match those in `courseInfo.json`. We should auto-create tags and topics as needed, assigning them a gray color and a description that says something like “Please put this tag/topic in your courseInfo.json file to set its color and description.”

nwalters512:
A continuation of #1504 that rewrites the remainder of the syncing process to use batched sprocs. Running on my machine, this reduces sync times for CS 225 down to 1-2s, and the example course can sync in ~200ms. As a part of this, I'm working on a proper test suite for the syncing process to ensure that all the new code behaves as expected.

Also fixes #1564; UUIDs are now only locally unique instead of being globally unique.

**TODO**
* [x] Fixes #1565 - auto-create tags and topics when syncing
* [x] Fixes #473 
* [x] Remove profiling code
* [x] Finish tests


sync/fromDisk/questions.js
```diff
@@ -1,124 +1,115 @@
...
+                external_grading_timeout: (q.externalGradingOptions && q.externalGradingOptions.timeout),
+                external_grading_enable_networking: (q.externalGradingOptions && q.externalGradingOptions.enableNetworking),
+            };
         });
-    },
-};
+
+        // Sync the questions to the DB. This will create/update all questions,
+        // soft-delete any unused questions, and ensure that all questions have
+        // assigned numbers (TODO clarify what these numbers are?)
```
mwest1066: These are the numbers referred to here: https://prairielearn.readthedocs.io/en/latest/assessment/#changing-question-order-randomization

sync/tests.md
```diff
@@ -0,0 +1,133 @@
+# Testing plan
```
mwest1066: This is an awesome document!

sync/tests.md
```diff
@@ -0,0 +1,133 @@
...
+* Course name/title/options are modified
+  * Modification should be reflected in the DB
+
+* An assessment set is added
+  * The assessment set should be added to the `assessment_sets` table
+* An assessment set is modified
+  * The assessment set should be updated in the `assessment_sets` table
+* An assesment set is removed
+  * The assesment set should be deleted from the `assessment_sets` table
+  * Sync should error if any assessments still reference this assessment set
```
mwest1066: I guess that an assessmentSet having its `name` changed is a special case of “remove and add” — I’m not sure if that’s worth testing separately.

tests/sync/initialSync.js
```diff
@@ -0,0 +1,48 @@
+const { assert } = require('chai');
+const util = require('./util');
+const helperDb = require('../helperDb');
+
+describe('Initial Sync', () => {
+  beforeEach('set up testing database', helperDb.before);
```
mwest1066: I think this needs to now be `before()` after Tim’s changes were merged.

tests/sync/initialSync.js
```diff
@@ -0,0 +1,48 @@
...
+    const topics = await util.dumpTable('topics');
+    assert.lengthOf(topics, 1);
+    const [topic] = topics;
+    const [syncedTopic] = courseData.course.topics;
+    assert.equal(topic.name, syncedTopic.name);
+    assert.equal(topic.color, syncedTopic.color);
+    assert.equal(topic.description, syncedTopic.description);
+  });
+
+  it('is idempotent when syncing the exact same course twice', async () => {
```
mwest1066: I like this!

tests/sync/initialSync.js
```diff
@@ -0,0 +1,48 @@
...
+const helperDb = require('../helperDb');
+
+describe('Initial Sync', () => {
+  beforeEach('set up testing database', helperDb.before);
+  afterEach('tear down testing database', helperDb.after);
+
+  it('correctly syncs content from disk to the database', async () => {
+    const courseData = util.getCourseData();
+    const courseDir = await util.writeCourseToTempDirectory(courseData);
+    await util.syncCourseData(courseDir);
```
mwest1066: Very slick!

tests/sync/util.js
```diff
@@ -0,0 +1,194 @@
...
+    for (const assessmentName of Object.keys(courseInstance.assessments)) {
+      const assessmentPath = path.join(courseInstancePath, assessmentName);
+      await fs.ensureDir(assessmentPath);
+      const assessmentInfoPath = path.join(assessmentPath, 'infoAssessment.json');
+      await fs.writeJSON(assessmentInfoPath, courseInstance.assessments[assessmentName]);
+    }
+  }
+};
+
+const course = {
```
mwest1066: I’m not sure how final this currently is, but I feel like we’d get a more complete test if we had at least two of each object (e.g., two questions, two tags, etc).

tests/sync/util.js
```diff
@@ -0,0 +1,194 @@
...
+
+  // Write all questions
+  const questionsPath = path.join(coursePath, 'questions');
+  await fs.ensureDir(questionsPath);
+  for (const qid of Object.keys(courseData.questions)) {
+    const questionPath = path.join(questionsPath, qid);
+    await fs.ensureDir(questionPath);
+    const questionInfoPath = path.join(questionPath, 'info.json');
+    await fs.writeJSON(questionInfoPath, courseData.questions[qid]);
+  }
```
mwest1066: This a/a code is so nice!

tests/sync/util.js
```diff
@@ -0,0 +1,194 @@
+const fs = require('fs-extra');
+const tmp = require('tmp-promise');
+const path = require('path');
+const sqldb = require('@prairielearn/prairielib/sql-db');
+
+const syncFromDisk = require('../../sync/syncFromDisk');
+
+module.exports.course = {
```
mwest1066: Unused?

lib/course-db.js
```diff
@@ -188,42 +302,53 @@ module.exports.checkInfoValid = function(idName, info, infoFile, courseInfo, log
...
+            jsonLoad.readInfoJSON(infoFile, schema, function(err, info) {
+                if (err && err.code && err.path && (err.code === 'ENOTDIR') && err.path === infoFile) {
+                    // In a previous version of this code, we'd pre-filter
+                    // all files in the parent directory to remove anything
+                    // that may have accidentally slipped in, like .DS_Store.
+                    // However, that resulted in a huge number of system calls
+                    // that got really slow for large directories. Now, we'll
+                    // just blindly try to read a file from the directory and assume
+                    // that if we see ENOTDIR, that means the directory was not
+                    // in fact a directory.
```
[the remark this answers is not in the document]
nwalters512: I don't think it's so much that the failed open was faster than an `lstat()` as it was that in practice we were doing a lot of unnecessary `lstat()`s since most things were in fact directories. Even if a failed open is actually slower, we're not going to get many of them and effectively halving the number of system calls is what really speeds things up here!

lib/course-db.js
```diff
@@ -103,7 +101,123 @@ module.exports.loadCourseInfo = function(courseInfo, courseDir, logger, callback
     });
 };
 
-module.exports.checkInfoValid = function(idName, info, infoFile, courseInfo, logger) {
+const counts = {};
+const nsTimes = {};
+const lastStarts = {};
```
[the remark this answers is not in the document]
nwalters512: 100% temporary, I'm going to rip this all out once I'm happy with everything.

exampleCourse/courseInstances/Sp15/assessments/this-should-fail
```diff
@@ -0,0 +1 @@
+testing!
```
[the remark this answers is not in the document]
nwalters512: Oops, this was left over from when I was iterating on the new code to catch files when loading `info.json` from a directory. Will remove (this should not fail a sync, btw).

tests/sync/util.js
nwalters512: Still incomplete - thinking about how best to scale this to N of each thing without having to manually repeat the checking code everywhere.

tests/sync/util.js
nwalters512: Unused - wrote that object to get a sense of how to structure the `CourseData` type.

sprocs/sync_assessments.sql
```diff
@@ -0,0 +1,303 @@
...
+                assessment->>'title',
+                assessment->'config',
+                (assessment->>'multiple_instance')::boolean,
+                (assessment->>'shuffle_questions')::boolean,
+                (assessment->>'max_points')::double precision,
+                (assessment->>'auto_close')::boolean,
+                NULL,
+                sync_assessments.new_course_instance_id,
+                assessment->>'text',
+                COALESCE((SELECT id FROM assessment_sets WHERE name = assessment->>'set_name' AND assessment_sets.course_id = sync_assessments.course_id), NULL),
```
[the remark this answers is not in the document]
nwalters512: Yep, should enforce that at the DB level.

Regarding checking for valid assessment sets, we could probably validate that up in JS. I think JS errors tend to be clearer for users, at least how we generally present them in job logs.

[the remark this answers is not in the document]
nwalters512: Never mind, looks like we already do this check in `lib/course-db.js`!

sprocs/sync_assessments.sql
```diff
@@ -0,0 +1,303 @@
...
+        DELETE FROM zones
+        WHERE
+            assessment_id = new_assessment_id
+            AND number > jsonb_array_length(assessment->'zones');
+
+        -- Delete excess alternative groups for this assessment
+        DELETE FROM alternative_groups
+        WHERE
+            assessment_id = new_assessment_id
+            AND ((number < 1) OR (number > (assessment->>'lastAlternativeGroupNumber')::integer));
```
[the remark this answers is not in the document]
nwalters512: The tricky thing here is that the alternative groups are in an array of arrays so we can't just use the length of something. We'll need to do some kind of calculation somewhere as we loop over the outer array. I guess I don't have a strong preference if we do this in JS and forward it into SQL or if it's done in the SQL. In this case, since I was computing the numbers in JS, I computed the last number there too.

tests/sync/initialSync.js
nwalters512: I think that's only for `helperServer` - we don't need a full server here, so I'm just using `helperDb`.

exampleCourse/courseInstances/Sp15/assessments/this-should-fail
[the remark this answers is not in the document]
mwest1066: I think this used to throw an error with the old sync code? That is, we ignore files, but if there is a directory then it’s an error for it to not contain an `info*.json` file. The original intention with this was to avoid the case where someone accidentally renamed something like an `infoCourseInstance.json` file (perhaps just a single letter change), which would cause us to delete the entire course instance (this is recoverable, but it might be a bit scary 😄).

I’m not convinced that we need to preserve this behavior. If we did want to do it then we could always `lstat()` only on failed `open()`, which should still be fast.

What do you think? Is this just worrying about an edge case that’s very unlikely in practice?

sprocs/sync_assessments.sql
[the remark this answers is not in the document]
mwest1066: Ah, excellent point, I had missed this. All good!

tests/sync/initialSync.js
mwest1066: Right!

tests/sync/util.js
mwest1066: It would be awesome to have some way of saying “assert(everything is like this old version, except for this specific set of rows)”. I’m not sure how best to achieve this.

nwalters512: I'm hoping some nice generic function like that is just going to kind of naturally fall out of the tests as I write them 🙂 but something something no premature abstractions.

exampleCourse/courseInstances/Sp15/assessments/this-should-fail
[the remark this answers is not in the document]
nwalters512: Ah, that's a different case than what I was testing with this file. If you look in `lib/course-db.js`, I'm careful to only handle `ENOTDIR` which is what we'll get if we try to read a file from a directory that doesn't exist (which is what would happen if there was a text file in`questions`, for instance). `ENOENT` for a missing `info.json` file still counts as an error. I'll write test cases for these too 🙂

sprocs/sync_course_staff.sql
```diff
@@ -0,0 +1,48 @@
...
+        SET
+            role = EXCLUDED.role;
+    END LOOP;
+
+    -- Downgrade all other enrollments
+    UPDATE enrollments AS e
+    SET role = 'Student'
+    FROM
+        users AS u,
+        course_instances AS ci
```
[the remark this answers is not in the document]
nwalters512: Here's hoping the query optimizer would have handled that for us :wink: but that's cleaner anyways, I'll make that change.

tests/sync/questionsSync.js
```diff
@@ -10,16 +12,24 @@ describe('Question syncing', () => {
   beforeEach('set up testing database', helperDb.before);
   afterEach('tear down testing database', helperDb.after);
 
-  it('soft-deletes questions that are removed from the course', async () => {
+  it('sof-deletes and restores questions', async () => {
```
mwest1066: `sof`

tests/sync/util.js
```diff
@@ -190,5 +391,53 @@ module.exports.captureDatabaseSnapshot = async function() {
...
+ * Asserts that two snapshots match each other. Two snapshots are defined as
+ * matching if they both contain the same keys and if for each key, the array
+ * of values contains the same elements. Elements may be in different orders.
+ * Optionally, a subset of the keys in the snapshot can be ignored.
+ * 
+ * @param {{ [key: string]: any[] }} snapshotA - The first snapshot
+ * @param {{ [key: string]: any[] }} snapshotB - The second snapshot
+ * @param {string[]} [ignoreKeys=[]] An optional list of keys to ignore
+ */
+module.exports.assertSnapshotsMatch = function(snapshotA, snapshotB, ignoreKeys = []) {
```
mwest1066: nice!

sync/tests.md
nwalters512: I did add an independent test for this.

sprocs/sync_questions.sql
```diff
@@ -0,0 +1,126 @@
...
+            question->>'external_grading_image',
+            (SELECT ARRAY_AGG(external_grading_files)::text[] FROM JSONB_ARRAY_ELEMENTS_TEXT(COALESCE(question->>'external_grading_files', '[]')::jsonb) external_grading_files)::text[],
+            question->>'external_grading_entrypoint',
+            (question->>'external_grading_timeout')::integer,
+            (question->>'external_grading_enable_networking')::boolean
+        FROM JSONB_ARRAY_ELEMENTS(sync_questions.new_questions) WITH ORDINALITY AS t(question, number)
+        -- TODO: is this necessary? does postgres make any guarantees about a) order when
+        -- selecting from JSONB_ARRAY_ELEMENTS or b) insertion order when inserting from a
+        -- SELECT expression?
+        ORDER BY number
```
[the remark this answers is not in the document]
nwalters512: Yep, I think your `ORDER BY number` suggestion is perfect, I went ahead and made that change in all the relevant places.

lib/json-load.js
```diff
@@ -22,12 +22,14 @@ module.exports.readJSON = function(jsonFilename, callback) {
...
             return;
         }
+        let json;
         try {
-            const json = jju.parse(data, {mode: 'json'});
-            callback(null, json);
+            json = jju.parse(data, {mode: 'json'});
         } catch (e) {
             callback(`Error in JSON file format: ${jsonFilename} (line ${e.row}, column ${e.column})\n${e.name}: ${e.message}`);
+            return;
```
mwest1066: Should we be using `ERR()` here so we get a full stacktrace?

nwalters512: Yes! Good call.

lib/async.js
```diff
@@ -0,0 +1,17 @@
+
+/**
+ * @param {<T>() => Promise<T>} func - The async function to execute
+ * @param {(error: Error | null, result: T | undefined) => void} callback - Callback to receive the result or the error
+ */
+module.exports.safeAsync = function(func, callback) {
+    (async () => {
+        let res;
+        try {
+            await func();
```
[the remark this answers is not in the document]
nwalters512: Ooh, TIL about `util.callbackify`. Yep, it's the same thing - maybe it's best to lean on the node standard library here?

lib/async.js
```diff
@@ -0,0 +1,17 @@
+
```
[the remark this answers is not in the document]
nwalters512: Removed this file, no longer a concern.

mwest1066:
Any idea why test coverage has decreased due to this PR. It seems that an enormous amount of tests were added, so it’s a bit depressing that measured coverage went down 😄

sprocs/sync_assessments.sql
```diff
@@ -0,0 +1,302 @@
...
+        AND a.id NOT IN (SELECT unnest(new_assessment_ids));
+
+    -- Delete unused assessment access rules
+    DELETE FROM assessment_access_rules AS aar
+    WHERE NOT EXISTS (
+        SELECT 1 FROM assessments AS a
+        WHERE
+            a.id = aar.assessment_id
+            AND a.deleted_at IS NULL
+            AND a.course_instance_id = sync_assessments.new_course_instance_id
```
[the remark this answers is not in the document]
nwalters512: 100% - fixed this.

PR #1587, opened 2019-07-11
Partially fix flaky tests in pl-number-input

ZhaoMJ:
This PR partially fixes the flaky test cases in pl-number-input elements, which didn't take care of comparison options.
🚨Note: this PR only adds test cases concerning different comparison options, but doesn't fix other issues - e.g. lack of complex number tests, lack of randomization in 'correct' and 'invalid' test cases, and incorrect render with respect to rtol and atol, etc.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -179,8 +179,6 @@
 
   * Fix Python linter errors in (Matt West).
 
-  * Fix flaky test cases in `pl-number-input` which didn't handle comparison options properly (Mingjie Zhao).
-
   * Fix `pl-code` HTML escaping (Nathan Walters).
 
   * Fix legacy question renderer by explicitly using `htmlparser2` for cheerio (Nathan Walters).
@@ -256,6 +254,8 @@
   * Fix help text CSV upload example with `points` (Matt West, h/t Mariana Silva and James Balamuta).
 
   * Fix CSRF checking for external grading live updates (Nathan Walters).
+
+  * Fix flaky test cases in `pl-number-input` which didn't handle comparison options properly (Mingjie Zhao).
 
   * Remove `allowIssueReporting` option in `infoCourseInstance.json` (Matt West).
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -179,7 +179,7 @@
 
   * Fix Python linter errors in (Matt West).
 
-  * Fix flaky test cases in pl-number-input which didn't handle comparison options properly
+  * Fix flaky test cases in `pl-number-input` which didn't handle comparison options properly (Mingjie Zhao).
 
   * Fix `pl-code` HTML escaping (Nathan Walters).
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -179,6 +179,8 @@
 
   * Fix Python linter errors in (Matt West).
 
+  * Fix flaky test cases in pl-number-input which didn't handle comparison options properly
+
   * Fix `pl-code` HTML escaping (Nathan Walters).
 
   * Fix legacy question renderer by explicitly using `htmlparser2` for cheerio (Nathan Walters).
--- a/elements/pl-number-input/pl-number-input.py
+++ b/elements/pl-number-input/pl-number-input.py
@@ -323,8 +323,35 @@ def test(element_html, data):
         data['raw_submitted_answers'][name] = str(a_tru)
         data['partial_scores'][name] = {'score': 1, 'weight': weight}
     elif result == 'incorrect':
-        data['raw_submitted_answers'][name] = str(a_tru + (random.uniform(1, 10) * random.choice([-1, 1])))
         data['partial_scores'][name] = {'score': 0, 'weight': weight}
+        # Get method of comparison, with relabs as default
+        comparison = pl.get_string_attrib(element, 'comparison', 'relabs')
+        if comparison == 'relabs':
+            rtol = pl.get_float_attrib(element, 'rtol', 1e-2)
+            atol = pl.get_float_attrib(element, 'atol', 1e-8)
+            # Get max error according to numpy.allclose()
+            eps = np.absolute(a_tru) * rtol + atol
+            eps += random.uniform(1, 10)
+            answer = a_tru + eps * random.choice([-1, 1])
+        elif comparison == 'sigfig':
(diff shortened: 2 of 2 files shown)


nwalters512:
Is this different from #1584?

elements/pl-number-input/pl-number-input.py
```diff
@@ -323,12 +323,39 @@ def test(element_html, data):
...
+            raise ValueError('method of comparison "%s" is not valid' % comparison)
+        data['raw_submitted_answers'][name] = str(answer)
     elif result == 'invalid':
         # FIXME: add more invalid expressions, make text of format_errors
         # correct, and randomize
         data['raw_submitted_answers'][name] = '1 + 2'
         data['format_errors'][name] = 'invalid'
     else:
-        raise Exception('invalid result: %s' % result)
+        raise Exception('invalid result: %s' % result)
```
nwalters512: Needs a newline at the end of the file

elements/pl-number-input/pl-number-input.py
```diff
@@ -323,12 +323,39 @@ def test(element_html, data):
         data['raw_submitted_answers'][name] = str(a_tru)
         data['partial_scores'][name] = {'score': 1, 'weight': weight}
     elif result == 'incorrect':
-        data['raw_submitted_answers'][name] = str(a_tru + (random.uniform(1, 10) * random.choice([-1, 1])))
         data['partial_scores'][name] = {'score': 0, 'weight': weight}
+        # Get method of comparison, with relabs as default
+        comparison = pl.get_string_attrib(element, 'comparison', 'relabs')
```
nwalters512: The defaults in this and other calls to `pl.get_*_attrib` are now repeated several times throughout the file; I'm thinking we should have some dict of defaults that we can reference for consistency and DRYness. @mwest1066 thoughts?

ZhaoMJ:
@nwalters512 Nothing different. I just accidentally uploaded something private from CS 233 so I had to delete that branch.

elements/pl-number-input/pl-number-input.py
ZhaoMJ: Maybe for now I can move the defaults to the head of the file as (fake) constant variables? Like CONST_COMPARISON_DEFAULT = 'relabs'

mwest1066: Yep, that’s a good idea

mwest1066:
@ZhaoMJ: could you describe in a bit more detail why the existing test case was broken? Thanks!

elements/pl-number-input/pl-number-input.py
nwalters512: By that I just meant something like

```python
defaults = {
  'comparison': 'relabs'
}

# ...

comparison = pl.get_string_attrib(element, 'comparison', defaults['comparison'])
```

But yeah, this is functionally the same as just using `CONSTANTS` at the top 🙂

ZhaoMJ:
@mwest1066 Sure! The original code didn't handle anything about tolerance and it only generated a random number fluctuated around the correct answer. If the tolerance was small, then it would probably work; but if the tolerance was large, it would be very possible that the small fluctuation was within the tolerance and the generated answer was actually correct. Therefore, some of the tests would fail with the error 'partial credits mismatch'.
I found this issue when I created new questions and tried to 'test for 100 time'. You can reproduce this issue by setting atol to some relatively large number like 2.

ChangeLog.md
```diff
@@ -179,6 +179,8 @@
 
   * Fix Python linter errors in (Matt West).
 
+  * Fix flaky test cases in pl-number-input which didn't handle comparison options properly
```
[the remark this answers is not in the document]
ZhaoMJ: Oops I forgot that when fixing my repo. Thanks!

elements/pl-number-input/pl-number-input.py
ZhaoMJ: For now I'm just using `CONSTANTS` at the top. If you decide to use something like `get_attribs(element)`, I can implement that.

ZhaoMJ:
Also, if you can tell me more details on those two FIXME tags, I can try to fix them.

elements/pl-number-input/pl-number-input.py
```diff
@@ -7,6 +7,20 @@
 import random
 
 
+RTOL_DEFAULT = 1e-2
+ATOL_DEFAULT = 1e-8
+SIZE_DEFAULT = 35
+NONE_DEFAULT = None
```
nwalters512: Not sure if this is necessary, I think `None` is close enough to a constant as it.

ChangeLog.md
```diff
@@ -179,6 +179,8 @@
 
   * Fix Python linter errors in (Matt West).
 
+  * Fix flaky test cases in `pl-number-input` which didn't handle comparison options properly (Mingjie Zhao).
```
nwalters512: The changelog is ordered chronologically, so this should be at the bottom of the "Fix" section.

elements/pl-number-input/pl-number-input.py
ZhaoMJ: I agree 😄

ChangeLog.md
ZhaoMJ: Hmmm I thought I was ordered lexigraphically...

ZhaoMJ: Thanks!

nwalters512:
@ZhaoMJ can you point me at a CS 357 question where you saw this problem? I'm going to deploy this to pl-dev.engr.illinois.edu just to double-check that everything is working properly :pray:

ZhaoMJ:
@nwalters512 I don't know where to find CS 357 questions but you can modified the addNumbers question in exampleCourse. Change the `pl-number-input` to `<pl-number-input answers-name="c" atol=5 label="$c=$"></pl-number-input>` and then run test 100 times. You will find some errors like this:
image
image

nwalters512:
My bad - I don't know where I got 357 from, I see now that you mentioned 233. Thanks, will try this out and merge if all looks good :+1:.

nwalters512:
Looks great, thanks @ZhaoMJ! Regarding the other FIXMEs, feel free to work on them in another branch if you get more details about them.

PR #1609, opened 2019-07-18
Improve in-browser file editor

tbretl:
This PR improves the in-browser file editor.
- It simplifies the interface, removing all buttons except "Save and Sync".
- It simplifies the use of drafts, using them only to recover from a failed "Save and Sync".
- It improves the interface (e.g., autosizes the editor, shows a banner with the results of "Save and Sync" available upon request, etc.).

It also eliminates the `number` column from the `jobs` table, along with the corresponding uniqueness constraint. This column was never used, and the uniqueness constraint could cause unnecessary errors when two jobs with the same course and user were created at the same time.


mwest1066:
@tbretl you may have already found this, but there is a second copy of `ace` at `public/localscripts/calculationQuestion/ace`. Unfortunately, there are 286 v2 questions that load this second copy of `ace`. These are:

* `pl-cs418` - one question
* `pl-cs125project` - one question
* `pl-cs498sl` - one question
* `pl-cs421` - lots of questions
* `pl-mse304` - lots of questions
* `pl-mse485` - lots of questions

Options that occur to me off the top of my head:

1. Leave `public/localscripts/calculationQuestion/ace` alone — expedient but ugly, and likely to confuse someone in the future.

2. Delete `public/localscripts/calculationQuestion/ace`, but add an extra route in `server.js` to serve `node_modules/ace` at the `public/localscripts/calculationQuestion/ace` endpoint.

lib/named-locks.js
```diff
@@ -62,37 +78,48 @@ module.exports = {
      *
      * @param {string} name - The name of the lock to acquire.
      * @param {boolean} wait - Whether to wait for the lock.
+     * @param {integer} timeout - If we are waiting for the lock, how many milliseconds before we stop (null, or anything other than a positive number, means forever).
      * @param {function} callback - A callback(err, lock) function.
      */
-    _getLock(name, wait, callback) {
+    _getLock(name, wait, timeout, callback) {
```
nwalters512: Thought for making this more extensible in the future: use an options object instead of separate parameters so you'd use it like:

```js
const options = {
  wait: true,
  timeout: 10,
};
_getLock('mylock', options, (err, lock) => { ... });
```

pages/instructorFileEditor/instructorFileEditor.ejs
```diff
@@ -1,78 +1,155 @@
...
-<html class="h-100">
-
 <head>
     <%- include('../partials/head'); %>
     <script src="/javascripts/lodash.min.js"></script>
     <script src="/javascripts/d3.min.js"></script>
     <link href="/stylesheets/theme.bootstrap.css" rel="stylesheet" />
-    <script src="/javascripts/ace/ace.js"></script>
+    <script src="/node_modules/ace-builds/src-min-noconflict/ace.js"></script>
+    <script src="/node_modules/crypto-js/crypto-js.js"></script>
```
nwalters512: Is this used on the page?

pages/instructorFileEditor/instructorFileEditor.ejs
```diff
@@ -1,78 +1,155 @@
-<!DOCTYPE html>
-<html class="h-100">
-
 <head>
     <%- include('../partials/head'); %>
     <script src="/javascripts/lodash.min.js"></script>
     <script src="/javascripts/d3.min.js"></script>
```
nwalters512: Is this used on the page?

pages/instructorFileEditor/instructorFileEditor.ejs
```diff
@@ -1,78 +1,155 @@
...
+                                </div>
+                                <% } %>
+                                <button type="button" class="close" data-dismiss="alert" aria-label="Close">
+                                    <span aria-hidden="true">&times;</span>
+                                </button>
+                            </div>
+                            <% if (fileEdit.failedPush) { %>
+                            <div class="row mt-4">
+                                <div class="col-auto">
+                                    In particular, it looks like there was a failure to push your changes to the remote git repo. The most likely cause is that another user made changes to other course files while you were editing. In this case, you should open the sync page in a new browser window, pull from remote git repo, close the window, and continue editing.
```
nwalters512: Could we add a button to directly kick off a sync in a new tab? So we could say "Click here to sync from GitHub in a new tab, then return here and continue editing." or the like.

pages/instructorFileEditor/instructorFileEditor.js
```diff
@@ -127,11 +158,11 @@ router.post('/', (req, res, next) => {
     if (!res.locals.authz_data.has_course_permission_own) return next(new Error('Insufficient permissions'));
 
     let fileEdit = {
-        editID: req.body.file_edit_id,
         userID: req.body.file_edit_user_id,
         courseID: req.body.file_edit_course_id,
         dirName: req.body.file_edit_dir_name,
         fileName: req.body.file_edit_file_name,
+        origHash: req.body.file_edit_orig_hash,
```
mwest1066: Apologies for missing this the first time around, but this pattern of getting data from the client is generally risky (I didn’t check if it’s actually a security hole in this specific case). The problem is that, for example, `courseID` could have been changed by the client to point to a different course, and it’s very easy for us to fail to check for this. My rule of thumb is that:

* variables in `res.locals` have been through our `authz` functions and so can all be trusted.
* variables in `req.body` are basically just whatever random thing a potentially-evil client sent to us, and they should be treated with extreme skepticism 😄

In this case, I suggest using `res.locals.user.id` and `res.locals.course.id` rather than values from the client. I think that `dir_name`, `file_name`, etc are fine because you are already treating them as untrusted below (e.g., checking that `dir_name` is inside the course dir).

This type of security flaw is one that I’ve made many times — I find it easy to forget that the variable we put into the EJS is not necessarily the variable we will get back in the POST request.

pages/instructorFileEditor/instructorFileEditor.ejs
tbretl: If you are referring to `crypto-js`, then yes - it's used for hashing in the client. Perhaps something lighter-weight would do, this was convenient (used on both client and server side).

pages/instructorFileEditor/instructorFileEditor.ejs
mwest1066: I definitely agree that what users will want here is a button that means “just fix it”. Ideally we would have a button that POSTs back to the current page to trigger the sequence of pull/sync/save/sync.

However, let’s leave that for another branch (hopefully soon!). Could we just make the text “sync page” in the paragraph above be a link to the sync page? I know that triggering a sync via that would be a bit ugly, so we could also skip that for now.

pages/instructorFileEditor/instructorFileEditor.ejs
```diff
@@ -1,78 +1,155 @@
...
+                                        <% } else { %>
+                                            File was saved, but failed to sync.
+                                        <% } %>
+                                    <% } else { %>
+                                        Failed to save and sync file.
+                                    <% } %>
+                                </div>
+                                <% if (fileEdit.jobSequenceId != null) { %>
+                                <div class="col-auto">
+                                    <a class="btn btn-secondary btn-sm" data-toggle="collapse" href="#results-<%= fileEdit.uuid %>" role="button">Detail</a>
```
mwest1066: Maybe change “Detail” to “Show details of error”?

pages/instructorFileEditor/instructorFileEditor.ejs
```diff
@@ -1,78 +1,155 @@
...
-                <% } %>
 
-                <div class="card d-flex flex-column flex-fill">
-                    <div class="card-header">
-                        <%= fileEdit.dirName %>/<%= fileEdit.fileName %>
+                    <div id="file-editor-<%= fileEdit.uuid %>-draft" class="col">
+                        <div class="card p-0">
+                            <% if (fileEdit.alertChoice) { %>
+                            <div class="card-header bg-white text-center">
+                                <button id="choose" class="btn btn-primary" type="button">Choose this version (continue editing)</button>
```
mwest1066: Could we add prominent labels to the left and right sides like “My version” and “Their version”? The terminology of “my” and “theirs” is what git uses, and I think it’s fairly clear.

We could then also label the buttons “Choose my version (continue editing)” and “Choose their version (discard my changes)”.

pages/instructorFileEditor/instructorFileEditorClient.js
```diff
@@ -1,22 +1,38 @@
...
+        mode: options.aceMode,
+        readOnly: options.readOnly,
+    });
 
-    if (options.aceMode) {
-        this.editor.getSession().setMode(options.aceMode);
+    if (options.altElementId) {
+        this.editor.setReadOnly(true);
+        this.altElement = $(`#${options.altElementId}`);
+        this.chooseElement = this.element.find('button[id=choose]');
```
mwest1066: Choosing my version didn’t work for me on local dev. Steps to reproduce:
1. I opened two windows editing the same file.
2. In window 1 I made change 1 and save-and-synced it.
3. In window 2 I made a different change 2 and clicked save-and-sync.
4. It gave me the “choose” dialog, and I chose “Choose this version (continue editing)”.
5. It gave me the editor with change 2 still present.
6. I clicked save-and-sync.
7. It gave me the same error and “choose” dialog as in step 4.

Instead, I expected that after step 6 the save-and-sync would succeed, overwriting change 1 with change 2.

mwest1066: I took a quick look at the tests and it looked to me like we are testing the case above that didn’t work for me, so I’m not sure whether I just did something weird, or we don’t actually have a test for this?

pages/instructorFileEditor/instructorFileEditor.ejs
mwest1066: Maybe an extra-small button for these “detail” buttons would look better?

tbretl: I chose to show the results both on success and failure - that's why I'm not referring to "error" in the button (adding a switch b/w "Show details of success" and "Show details of error" or whatever seems unnecessary). Totally open to a better choice than "Detail".

With respect to extra-small buttons - that's no longer a default option in BS4 (as far as I understand, anyway).

pages/instructorFileEditor/instructorFileEditor.ejs
mwest1066: Seems fine to me

pages/instructorFileEditor/instructorFileEditor.ejs
mwest1066: We do have quite a few `btn-xs` classes in our templates, and they seem to work. This is just a very weak suggestion, though.

Fair enough on “error”. I’d prefer “Show details”. Is it possible to add a close-icon button inside the collapsable that will close it? The alternative is to label the details-button with “Show/hide details”, but I’ve always found that a bit ugly. Of course we could add some JS to change “Show” to “Hide”, but I think that’s more mess than is warranted.

Way down in the weeds here, of course, so just do whatever you want (or nothing)!

server.js
```diff
@@ -143,6 +143,10 @@ app.use(express.static(path.join(__dirname, 'public')));
 app.use('/MathJax', express.static(path.join(__dirname, 'node_modules', 'mathjax')));
 app.use('/node_modules', express.static(path.join(__dirname, 'node_modules')));
 
+// Support legacy use of ace by v2 questions
+app.use('/public/localscripts/calculationQuestion/ace', express.static(path.join(__dirname, 'node_modules/ace-builds/src-min-noconflict')));
```
mwest1066: Does this work? It looks to me like this route will be preemptively handled by the static route on line 142.

mwest1066: Or does express static fall through if the file isn’t there? That would be both neat and slightly scary.

tbretl: I've tested one of the courses in question locally - seems to work fine.

pages/instructorFileEditor/instructorFileEditor.ejs
tbretl: `btn-xs` is defined in a custom class, so I'll not use it unless there's a compelling reason.

The show/hide thing is done, in a small bit of JS.

pages/instructorFileEditor/instructorFileEditorClient.js
tbretl: Fixed by replacing `origHash` with `diskHash` on "Choose my version".

tbretl: We don't have a test for this, because it's a client-side thing. In any case, the problem has now been addressed.

pages/instructorFileEditor/instructorFileEditor.ejs
tbretl: One-click recovery = implemented.

pages/instructorFileEditor/instructorFileEditor.ejs
```diff
@@ -1,78 +1,193 @@
...
-                            <span aria-hidden="true">&times;</span>
-                        </button>
+                        <% if ((fileEdit.alertChoice) && (!fileEdit.failedPush)) { %>
+                        <div class="alert alert-danger alert-dismissible fade show m-2" role="alert">
+                            <% if (fileEdit.hasSameHash) { %>
+                            You were editing this file and made changes. You may choose either to continue editing or to discard your changes.
+                            <% } else { %>
+                            Both you and another user made changes to this file. You may choose either to continue editing your draft or to discard your changes.
+                            <% } %>
+                            In particular, if you click <strong>Choose my version</strong> and then click <strong>Save and sync</strong>, you will overwrite the version of this file that is on disk. If you instead click <strong>Choose their version</strong>, any changes you have made to this file will be lost.
```
mwest1066: Even after the user clicks "Choose my version" this error message is still displayed. Could we dismiss this alert after the user chooses a version? (Just this alert should be dismissed, I think, not the other alert saying that we failed to sync.)

mwest1066: I think this alert does correctly vanish when the user clicks "Choose their version" because we reload the page.

tbretl: Done.

PR #1627, opened 2019-08-04
Disable autocomplete on v3 questions

### LINKED ISSUE #1624: Consider adding autocomplete="off" to elements with inputs
Chrome (and presumably other browsers) will aggressively remember and suggest previous inputs to form inputs by the `name` attribute. This is of limited, if any, utility to PL, and can be disabled to some degree with `autocomplete="off"` on input elements. We should investigate if this would have any adverse consequences and consider adding it to all input elements.

nwalters512:
Adds `autocomplete="off"` to the v3 question for to disable autocomplete. Fixes #1624.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -163,6 +163,8 @@
 
   * Change file editor to simplify the use of drafts and to improve the user interface (Tim Bretl).
 
+  * Change v3 questions to disable autocomplete on the question form (Nathan Walters).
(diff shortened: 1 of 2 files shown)


nwalters512:
Tested in Chrome, Firefox, and Safari with `addNumbers` and `examplesStringInput` questions from `exampleCourse`.

PR #1636, opened 2019-08-08
Add reporting of unsuccessful cron jobs

mwest1066:


cron/sendUnfinishedCronWarnings.js
```diff
@@ -0,0 +1,34 @@
...
+
+module.exports = {};
+
+module.exports.run = (callback) => {
+    if (!opsbot.canSendMessages()) return callback(null);
+    sqldb.query(sql.select_unfinished_cron_jobs, [], (err, result) => {
+        if (ERR(err, callback)) return;
+        let msg;
+        if (result.rowCount <= 0) {
+            msg = `_All cron jobs successfully finished_\n`;
```
nwalters512: Probably no need to send this if we don't have any problems?

mwest1066: Done.

nwalters512:
Lol, so much for me thoroughly reviewing the code :joy: looks like there are syntax errors in opsbot.

PR #1637, opened 2019-08-14
Assessment password cookie clearing fix

### LINKED ISSUE #1579: Password exam locking behavior needs changes
A potential UI issue for people taking an exam remotely with a proctor (password-protected): returning to the front page of PL causes the exam to re-lock. This makes the top navbar dangerous; students accidentally click "Assessments" trying to return to the exam overview instead of using the "Assessment Overview" button. Although re-locking the exam doesn't stop the exam timer, proctors may balk at logging students back into the exam if this happens, or the student may panic and not know what to do.

It would be good if there were some way to disable parts of the top navbar in specific situations. We already noticed it caused confusion for LTI students, but I think @trombonekenny has a solution drafted for that, see #1494

If hiding the navbar entirely would cause issues, perhaps we could add a modal warning popup when a student taking an exam clicks one of the links in the top navbar.

trombonekenny:
Do not clear the assessment password cookie when visiting the `/pl` homepage, but do clear it when logging out.

Closes #1579

### DIFF
--- a/middlewares/clearCookies.js
+++ b/middlewares/clearCookies.js
@@ -1,8 +1,13 @@
-var _ = require('lodash');
+const _ = require('lodash');
+
+var cookies_to_ignore = ['pl_authn', 'pl_assessmentpw'];
 
 module.exports = function(req, res, next) {
     _(req.cookies).each(function(value, key) {
-        if (/^pl_/.test(key) && key != 'pl_authn') {
+        if (/^pl_/.test(key)) {
+            if (cookies_to_ignore.includes(key)) {
+                return;
+            }
             res.clearCookie(key);
         }
     });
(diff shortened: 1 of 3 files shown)


trombonekenny:
This one will need to be tested on `pl-dev`. Auth testing in devmode is cumbersome.

middlewares/clearCookies.js
```diff
@@ -1,8 +1,13 @@
-var _ = require('lodash');
+const _ = require('lodash');
+
+var cookies_to_ignore = ['pl_authn', 'pl_assessmentpw'];
```
mwest1066: May as well make this one `const` as well?

trombonekenny:
Test OK on `pl-dev`. I'd say this is safe to ship.

PR #1638, opened 2019-08-14
Grading containers

coatless:
Within this PR, we propose adding a new top-level folder `graders` that has a sample Docker container for different programming languages.

Naming scheme is:

```
graders/<language>
```

Thus, _R_  and _Python_ would have:

```
graders/r
graders/python
```

As a result, we moved the `centos-python` image from the `environments/` folder.


exampleCourse/questions/fibonacciEditorWithR/info.json
```diff
@@ -0,0 +1,16 @@
...
+    "uuid": "8f3e0727-336b-4388-9410-565951b0beff",
+    "title": "Fibonacci Autograding with R",
+    "topic": "Functions",
+    "tags": ["code", "rstats", "v3", "balamut2", "Su19"],
+    "type": "v3",
+    "singleVariant": true,
+    "gradingMethod": "External",
+    "externalGradingOptions": {
+        "enabled": true,
+        "image": "prairielearn/grader-r",
```
coatless: Requires @mwest1066 to setup the image on DockerHub.

coatless:
Before this can be merged, two steps are required:

1. DockerHub will need to be enabled for the `prairielearn/grader-r` image
2. The `centos-python` image listed in `info.json` will need to be changed in downstream questions. (I think @nwalters512 has a script for this?)

graders/r/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+      org.label-schema.vcs-url="https://github.com/PrairieLearn/PrairieLearn" \
+      org.label-schema.vendor="PrairieLearn" \
+      maintainer="James Balamuta <balamut2@illinois.edu>"
+
+# Add a script to install any additional R packages
+COPY r-requirements.R /
+
+# Modify the image to meet with PrairieLearn specifications
+RUN useradd ag && \                        # Enable autograde user for PrairieLearn
+    su root -c "Rscript /r-requirements.R" # Install the additional packages
```
mwest1066: Have you tested this? I think the comment on line 14 will cause line 15 to not execute.

mwest1066:
@coatless to let us test this, could you remove the changes to `exampleCourse` from this PR and put them in a separate PR? That is, this PR would only contain the changes to `grader/`. That way I can merge the new graders and set up the build rules for them, while will make the new PR (containing the `exampleCourse` changes) be testable. Sorry for not thinking of this sooner! Thanks!

graders/r/Dockerfile
coatless: Mmm, I did test the image creation. You can see that the `Dockerfile` successfully built: 

https://cloud.docker.com/repository/registry-1.docker.io/jamesbalamuta/ubuntu-pl-r/builds/f6ade2cb-e3a5-4930-a73a-b9ac50bbd4dd

However, I didn't test whether `RcppArmadillo` was enabled. I'll setup something that uses commenting from the `centos-plbase` image: 

https://github.com/PrairieLearn/PrairieLearn/blob/master/environments/centos7-plbase/Dockerfile#L53

coatless:
@mwest1066 implemented the changes. Note, this might initially error as we changed the image name used in Fibonacci from `centos7-python` to `grader-python`.  

We used the following to update questions:

```bash
grep -rl centos7-python . | xargs sed -i '' 's/centos7-python/grader-python/g'
```

(Used `-i ''` because macOS uses an old bash shell (v3?) requiring `-i`.)

ChangeLog.md
```diff
@@ -5,8 +5,12 @@
 
   * Add UUIDs to cron job logs for better debugging (Matt West).
 
+  * Add example docker grading images for externally graded questions (James Balamuta).
+  
   * Change v3 questions to disable autocomplete on the question form (Nathan Walters).
 
+  * Change `centos-python` to `grader-python` and place it under `graders/`  (James Balamuta).
```
mwest1066: nit: it used to be `centos7-python` (note `7`)

graders/r/Dockerfile
mwest1066: The change to comment format fixes this, thanks!

ChangeLog.md
coatless:
```suggestion
  * Change `centos7-python` to `grader-python` and place it under `graders/`  (James Balamuta).
```

Issue #1369, opened 2018-11-12
"Natural Sort" for Assessment Filter


coatless:
The sorting for assessment sets doesn't appear to be **natural**, c.f. https://blog.codinghorror.com/sorting-for-humans-natural-sort-order/

<img width="188" alt="screen shot 2018-11-12 at 4 34 44 pm" src="https://user-images.githubusercontent.com/833642/48379187-e3703800-e698-11e8-9595-cdaea9257845.png">

Natural sort would give:

<img width="187" alt="screen shot 2018-11-12 at 4 34 11 pm" src="https://user-images.githubusercontent.com/833642/48379161-cfc4d180-e698-11e8-9508-824e6ecb8622.png">

Note the placement of `Q10` and `Q11` at the end vs. beginning.

mwest1066:
Absolutely agreed. We already do this for the assessment order on the Assessments page. We use the package `javascript-natural-sort`. See `sync/fromDisk/assessments.js` for an example of where we do this.

It should be easy to use this same code to do a natural sort of `all_assessments` in `instructorQuestions.js`.

mwest1066:
It would actually be best to implement this in a new SQL sproc. There is code for this in https://github.com/PrairieLearn/PrairieLearn/pull/1640/commits/bcd991f92446130600e1a159bbffd1da2a1cd127#r317405340

This code would need to be pulled out into a sproc, tests added, and everywhere that we currently use `javascript-natural-sort` could be converted to use this new sproc.

nwalters512:
For anyone looking at this issue who might want to pick it up: we're moving away from sprocs, so we'd likely want to do this in JavaScript instead.

PR #1640, opened 2019-08-14
Syncing rewrite, part 2

### LINKED ISSUE #2256: Renaming assessment sets deletes entries from gradebook
Steps to reproduce: 

1. Have a custom assessment set that you have been using, e.g. "Project".
2. In `infoCourse.json`, change the name and heading, e.g. "Project" -> "Projects".
3. In an `infoAssessment.json` from a previous assignment, change the `"set"`, e.g. "Project" -> "Projects".
4. Commit, push, sync.
5. The grades for the assessment are now gone.  Reverting the commit and re-syncing does not bring the grades back.  Be very nervous until Matt assures you the data is still there :)

### LINKED ISSUE #2793: Invalid working directory in file browser after changing directory name with broken sync
Do these things with the in-browser editor:

- Make a copy of a course instance
- Make a change to the `infoCourseInstance.json` file of the new course instance that breaks sync (e.g., remove a comma)
- Make a change to the CIID of the new course instance (from the instance admin "Settings" tab)
- Try to access `infoCourseInstance.json` of the new course instance from the file browser (the instance admin "Files" tab)

Then you will see an error saying "Invalid working directory":

Screen Shot 2020-07-31 at 5 31 18 PM

It's likely the same thing would happen with an existing course instance and with anything that breaks sync. It's likely the same thing would happen with other "objects" like questions and assessments.

nwalters512:
This PR initially started as an effort to be able to sync individual questions/assessments/etc. at once, but evolved into a very fundamental way to the way that things are synced.

### Old way of syncing
Files were loaded and validated one by one. If any file failed validation, sync as a whole would be aborted. Once files were loaded, syncing to the DB began. As we were syncing, we'd be performing more validation on models as well, such as checking if the QIDs in an assessment were valid. Again, if any validation failed, syncing would be aborted.

### Motivation for changing this
The primary motivation for changing how this works is to support our future goal of being able to work entirely in the browser. If multiple people are working on course content at once, a syntax error in one question shouldn't stop someone else from being able to sync and view their work in another.

### New way of syncing
Syncing is now split into two phases: loading/validation, and writing. During the loading/validation phase, all course content is loaded as before. However, instead of throwing errors, we capture them and associate them with the file. We also validate during this phase, again associating errors and warnings with the appropriate file. Once the loading/validation phase completes, we'll have either errors or valid data for each JSON file in the course. Then, we actually write all of that data (and possibly errors/warnings) to the database. This model ensures that syncing should never fail from an unexpected exception and that we'll write as much valid data as we can.

There are some nuances to this model, such as how UUIDs and names are handled, and what happens to existing data if errors are encountered. I still need to document this.

**TODO**
- [x] Handle `order_by` for assessments
- [x] Handle exam UUID checking for assessment access rules
- [x] Add tests for duplicate assessment sets, tags, and topics
- [x] Document syncing process more fully, including the aforementioned nuances (there are at least some docs and the code is nice and clear)
- [x] Write complete tests for all strange JSON situations (we did at least some of them)
- [x] Make sure these changes incidentally fix #2256
- [x] Add a test to make sure #2256 is really fixed
- [x] Audit DB foreign key constraints like #2256 and make sure we don't have `ON DELETE CASCADE` from soft-delete tables to hard-delete ones
- [x] Add a test to delete a used course topic before syncing and check that a question using it still has a topic. This will currently fail.
- [x] Add a test to delete a used course tag before syncing and check that a question using it still has it as a tag. This will currently fail.
- [x] Check that this PR fixes #2793 
- [x] Enable TypeScript type checking for all new files + tests
- [x] Completely delete all old syncing code


lib/json-load.js
```diff
@@ -13,9 +14,8 @@ ajv.addMetaSchema(require('ajv/lib/refs/json-schema-draft-04.json'));
 /**
  * Asynchronously reads the specified JSON file.
  * 
- * @param jsonFilename {string} The name of the file to read
- * @param callback {Function} Invoked with the resolved JSON data or an error
- * @returns {void}
+ * @param {string} jsonFilename The name of the file to read
+ * @param {(err: Error | null | undefined, data?: any) => void} callback Invoked with the resolved JSON data or an error
```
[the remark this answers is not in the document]
nwalters512: Hmm, I thought I was copying this pattern from the Node typings, but it looks like they're using `Error | null`. Idk where I got this from, will change 🙂

lib/json-load.js
```diff
@@ -50,19 +49,21 @@ module.exports.validateJSON = function(json, schema, callback) {
...
     }
     if (!valid) {
-        callback(new Error(`Error in JSON file: ${ajv.errorsText(validate.errors)}`));
+        callback(new Error(ajv.errorsText(validate.errors)));
     } else {
         callback(null, json);
     }
 };
 
+module.exports.validateJSONAsync = util.promisify(module.exports.validateJSON);
```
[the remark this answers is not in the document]
nwalters512: Separate types would make IDE features more useful, so I'm in favor of adding :+1:

migrations/148_assessments__assessment_set_id__nullable.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE assessments ALTER COLUMN assessment_set_id DROP NOT NULL;
+-- This will verify that any non-deleted assessments have a valid assessment set ID
+-- NOTE: This is expensive, as all rows will be checked. Keep this in mind when
+-- deploying.
+-- ALTER TABLE assessments ADD CONSTRAINT assessment_set_id_check CHECK (
+  -- deleted_at IS NOT NULL
+  -- OR (deleted_at IS NULL AND assessment_set_id IS NOT NULL)
+-- );
```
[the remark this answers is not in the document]
nwalters512: I think this was just a temporary thing as I was developing, let me put it back in and make sure tests still look good.

lib/named-locks.js
```diff
@@ -1,117 +1,153 @@
...
-     * @param {string} name - The name of the lock to acquire.
-     * @param {function} callback - A callback(err, lock) function. If lock is null then it was not acquired.
-     */
-    tryLock(name, callback) {
-        this._getLock(name, {wait: false}, (err, lock) => {
-            if (ERR(err, callback)) return;
-            callback(null, lock);
-        });
-    },
+module.exports.tryLockAsync = util.promisify(module.exports.tryLock);
```
mwest1066: Following up on the resolved comment above, there are more of these `...Async()` versions without their own docs. I’m not necessarily pushing to add docs here, just pointing it out in case you missed them 😄

migrations/148_assessments__assessment_set_id__nullable.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE assessments ALTER COLUMN assessment_set_id DROP NOT NULL;
+-- This will verify that any non-deleted assessments have a valid assessment set ID
+-- NOTE: This is potentially very expensive, as all rows will be checked.
+-- Keep this in mind when deploying.
+ALTER TABLE assessments ADD CONSTRAINT assessment_set_id_check CHECK (
+  deleted_at IS NOT NULL
+  OR (deleted_at IS NULL AND assessment_set_id IS NOT NULL)
+);
```
mwest1066: Just looking ahead to minimizing downtime during deployment, I’m wondering if we should change this SQL to only add this constraint if it doesn’t already exist (may need plpgsql code?) and run the ALTER TABLE ahead of time by hand?

pages/partials/assessmentSyncErrorsAndWarnings.ejs
```diff
@@ -0,0 +1,21 @@
+<% if (assessment.sync_errors) { %>
+  <div class="alert alert-danger" role="alert">
+    <h2 class="h5 alert-heading">Sync error</h2>
+    <p>
+      There was an error syncing this assessment; the information you see below
+      may be inconsistent with this assessment's <code>infoAssessment.json</code> file.
+      Please correct the error and sync again.
```
mwest1066: Would it be easy to add a link to the `edit` page for this `infoAssessment.json` file here? And similarly for course, courseInstance, etc.

pages/partials/assessmentSyncErrorsAndWarnings.ejs
```diff
@@ -0,0 +1,21 @@
...
+    </p>
+    <pre class="bg-dark text-white rounded p-3 mb-0"><code><%= assessment.sync_errors %></code></pre>
+  </div>
+<% } %>
+<% if (assessment.sync_warnings) { %>
+  <div class="alert alert-warning" role="alert">
+    <h2 class="h5 alert-heading">Sync warning</h2>
+    <p>
+      These warnings do not impact the ability to sync this assessment, but they
+      should still be reviewed and corrected.
```
mwest1066: Also link to `edit` page?

pages/instructorQuestions/instructorQuestions.ejs
```diff
@@ -30,6 +30,11 @@
               <% questions.forEach(function(question) { %>
               <tr>
                 <td class="align-middle">
+                  <% if (question.sync_errors) { %>
+                    <i class="fa fa-times text-danger" aria-hidden="true"></i>
+                  <% } else if (question.sync_warnings) { %>
+                    <i class="fa fa-exclamation-triangle text-warning" aria-hidden="true"></i>
+                  <% } %>
```
mwest1066: Can we make these link to the `edit` page for the corresponding `info.json` file, or have a popover containing a message about the error and link to `edit`, or something like that? I guess I’m thinking that we want users to see the error, and be able to click it to figure out what it means and how to fix it. A similar current pattern is the issues counter on questions that can be clicked to go straight to the issues list: https://github.com/PrairieLearn/PrairieLearn/blob/master/pages/instructorQuestions/instructorQuestions.ejs#L34

pages/instructorFileEditor/instructorFileEditor.ejs
```diff
@@ -8,7 +8,7 @@
     <script src="/node_modules/ace-builds/src-min-noconflict/ace.js"></script>
     <script src="/node_modules/crypto-js/crypto-js.js"></script>
     <script src="./edit/instructorFileEditorClient.js"></script>
-    <script src="/javascripts/socket.io.js"></script>
+    <script src="/node_modules/socket.io-client/dist/socket.io.slim.js"></script>
```
mwest1066: yay!

sync/fromDisk/tags.js
```diff
@@ -1,98 +1,71 @@
...
-            if (unknownTags.length > 0) {
-                throw new Error(`Question ${qid} has unknown tags: ${unknownTags.join(', ')}`);
-            }
-            const duplicateTags = getDuplicates(tags);
-            if (duplicateTags.length > 0) {
-                throw new Error(`Question ${qid} has duplicate tags: ${duplicateTags.join(', ')}`);
-            }
-            paramQuestionTags.push([question.id, tags.map(tag => tagIdsByName[tag])]);
-        }
+    /** @tyle {} */
```
mwest1066: `tyle`

sync/infofile.js
```diff
@@ -0,0 +1,131 @@
+// @ts-check
+/**
+ * Represents the result of attempting to load and validate an info file. May
+ * contain any combination of errors, warnings, data, and a UUID.
+ * @template T
```
mwest1066: Wow, I didn’t know typescript had templates. This is very clever use of them!

tests/sync/courseInstancesSync.js
```diff
@@ -9,10 +10,18 @@ const helperDb = require('../helperDb');
 const { assert } = chai;
 
 describe('Course instance syncing', () => {
-  before('remove the template database', helperDb.dropTemplate);
+  // before('remove the template database', helperDb.dropTemplate);
```
mwest1066: leftover commented-out code (and similarly elsewhere)

tests/sync/questionsSync.js
```diff
@@ -77,6 +79,8 @@ describe('Question syncing', () => {
     // We'll just sync the same course from two different directories.
     // Since courses are identified by directory, this will create two
     // separate courses.
+    // TODO: we should actually assert that two independent copies of all data exist.
+    // While developing, I found some cases where data wasn't correctly scoped to one course/course instance.
```
mwest1066: I assume this TODO and the ones below are temporary, but flagging them here just in case.

tests/sync/courseInstancesSync.js
nwalters512: These are intentionally left in here, these need to be run every time the DB changes when running individual tests with `npx mocha path/to/file.js`. It was easiest to leave them there and uncomment when needed.

sprocs/sync_single_question.sql
```diff
@@ -0,0 +1,157 @@
+
+CREATE OR REPLACE FUNCTION
+    sync_single_question(
```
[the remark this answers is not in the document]
nwalters512: I haven't thought carefully about this either, and actually this code should be removed until partial syncing is properly implemented in another PR. This was a remnant of when the focus of this branch was actually partial syncing 🙂 we can revisit how that's done in the future.

sprocs/sync_assessments.sql
```diff
@@ -258,15 +333,39 @@ BEGIN
...
+    SET order_by = assessments_with_ordinality.order_by
+    FROM (
+        SELECT
+            tid,
+            row_number() OVER (ORDER BY (
+                SELECT string_agg(convert_to(coalesce(r[2],
+                    length(length(r[1])::text) || length(r[1])::text || r[1]),
+                    'SQL_ASCII'),'\x00')
+                FROM regexp_matches(number, '0*([0-9]+)|([^0-9]+)', 'g') r 
+            ) ASC) AS order_by
```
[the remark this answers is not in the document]
nwalters512: #1369 is all done with client-side JS IIRC, so I don't know if that would be relevant here. If this proves to be useful in other places in the future, I'd be happy to move to a sproc, but for the time being I think this is best left inline.

pages/partials/assessmentSyncErrorsAndWarnings.ejs
nwalters512: Yep, I think this would be pretty easy.

sprocs/sync_assessments.sql
```diff
@@ -20,82 +21,155 @@ DECLARE
     new_assessment_question_id bigint;
     new_assessment_question_ids bigint[];
 BEGIN
-    -- The outermost structure of the JSON blob is an array of assessments
-    FOR assessment IN SELECT * FROM JSONB_ARRAY_ELEMENTS(sync_assessments.assessments) LOOP
-        new_assessment_question_ids := array[]::bigint[];
+    -- Move all our data into a temporary table so it's easier to work with
+    DROP TABLE IF EXISTS disk_assessments;
```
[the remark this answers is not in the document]
nwalters512: Correct! Do you see any problems with this thinking?

pages/instructorQuestions/instructorQuestions.ejs
nwalters512: I'm kind of torn on this - simply clicking on the question will show that information, and said information will probably be a lot more legible in a full-width browser window than in a tooltip. I'm not sure if much is gained by putting it inline on the question/assessment page.

mwest1066: You are right, popovers would suck for this 😄. I do still think that the `edit` link (plus showing errors there) would be awesome.

sprocs/sync_assessments.sql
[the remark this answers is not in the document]
mwest1066: Nope, seems solid to me!

sprocs/sync_assessments.sql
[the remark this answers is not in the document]
mwest1066: My thought was that we could fix #1369 by changing https://github.com/PrairieLearn/PrairieLearn/blob/master/pages/instructorQuestions/instructorQuestions.sql#L45 from:
```
ORDER BY aset.number,a.number;
```
to:
```
ORDER BY aset.number,naturalsort(a.number);
```

tests/sync/courseInstancesSync.js
mwest1066: sounds good! maybe a comment to that effect, if this is going to be a long-term thing?

nwalters512:
Taking notes as I resolve merge conflicts:

* I'm not confident that the issue described in https://github.com/PrairieLearn/PrairieLearn/pull/2063 will be fixed here; both that branch and this branch contain diverging bits of code. Ideally we'd write tests against master to prevent regressions, and then merge those changes into this branch to test for regressions.
* I'm not entirely confident that the changes from https://github.com/PrairieLearn/PrairieLearn/pull/2201, https://github.com/PrairieLearn/PrairieLearn/pull/2208, and https://github.com/PrairieLearn/PrairieLearn/pull/2228 have survived the merge. This should be verified by hand.
* All the validation from https://github.com/PrairieLearn/PrairieLearn/pull/2201 (such as https://github.com/PrairieLearn/PrairieLearn/pull/2201/files#diff-ca1588431ac4a19548590ca69553fe2fR48) will need to be re-added; these things will likely need tests (or possibly refactorings of the existing tests) as well.

nwalters512:
> * I'm not confident that the issue described in #2063 will be fixed here; both that branch and this branch contain diverging bits of code. Ideally we'd write tests against master to prevent regressions, and then merge those changes into this branch to test for regressions.

@mwest1066 going to need your help to test this, or at least provide test cases that I could use to reproduce a potential bug.

> * I'm not entirely confident that the changes from #2201, #2208, and #2228 have survived the merge. This should be verified by hand.

These have tests and the tests are passing!

> * All the validation from #2201 (such as https://github.com/PrairieLearn/PrairieLearn/pull/2201/files#diff-ca1588431ac4a19548590ca69553fe2fR48) will need to be re-added; these things will likely need tests (or possibly refactorings of the existing tests) as well.

I've implemented the validation with the `infofile.addError` API and updated the tests to work correctly.

> * #1646 produced some conflicts; my PR description there alluded to this maybe being fixed on this branch, but we should verify that that's the case. That PR unfortunately did not come with tests, so we should consider adding some.

I've verified that the functionality exists, but there aren't currently tests for this. I'll added a checkbox to the PR description for this task.

package.json
```diff
@@ -20,13 +31,15 @@
...
         "byline": "^5.0.0",
+        "chalk": "^2.4.2",
         "chart.js": "^2.9.3",
         "cheerio": "^0.22.0",
         "clipboard": "^2.0.6",
         "cookie-parser": "^1.4.5",
         "crypto-js": "^4.0.0",
         "csv": "^5.3.2",
         "csvtojson": "^2.0.10",
+        "date-fns": "^2.0.0",
```
mwest1066, line 42: Is this doing something different/better than `moment.js`? I'm just checking in a losing attempt to keep our library dependencies under control 😄

pages/instructorAssessments/instructorAssessments.ejs
```diff
@@ -9,6 +9,7 @@
   <body>
     <%- include('../partials/navbar'); %>
     <div id="content" class="container-fluid">
+      <%- include('../partials/courseInstanceSyncErrorsAndWarnings'); %>
```
mwest1066, line 12: If I understand correctly, this partial should only be included on pages that have a single `locals.assessment`? This page has many assessments, so this partials code won't do anything?

pages/instructorAssessmentAccess/instructorAssessmentAccess.ejs
```diff
@@ -6,6 +6,7 @@
   <body>
     <%- include('../partials/navbar'); %>
     <div id="content" class="container-fluid">
+      <%- include('../partials/assessmentSyncErrorsAndWarnings'); %>
```
mwest1066, line 9: I wonder whether it would be cleaner/easier to put this code in the navbar? That is, I think we want to display the "this assessment is not synced" message on all pages for the assessment? The navbar code is the place where we know which list of pages this is. Otherwise I think we need to add an `include()` like this to essentially all of the instructor pages?

pages/instructorFileEditor/instructorFileEditor.js
```diff
@@ -142,13 +144,22 @@ router.get('/*', (req, res, next) => {
...
-                serverJobs.getJobSequence(fileEdit.jobSequenceId, res.locals.course.id, (err, job_sequence) => {
+                serverJobs.getJobSequenceWithFormattedOutput(fileEdit.jobSequenceId, res.locals.course.id, (err, job_sequence) => {
                     if (ERR(err, callback)) return;
                     fileEdit.jobSequence = job_sequence;
                     callback(null);
                 });
             }
         },
+        (callback) => {
+            const promisified = callbackify(editorUtil.getErrorsAndWarningsForFilePath);
```
mwest1066, line 155: I'm a bit confused about the name here. I would have thought we'd call this `callbackified`? And use `promisfied` in the case that we did `const promisified = util.promisify(...)`?

pages/instructorFileEditor/instructorFileEditor.js
```diff
@@ -142,13 +144,22 @@ router.get('/*', (req, res, next) => {
...
+            promisified(res.locals.course.id, relPath, (err, data) => {
+                if (ERR(err, callback)) return;
+                fileEdit.sync_errors = data.errors;
+                fileEdit.sync_warnings = data.warnings;
+                callback(null);
+            });
+        },
```
mwest1066, line 162:
```suggestion
        callbackify(async () => {
            data = await editorUtil.getErrorsAndWarningsForFilePath(res.locals.course.id, relPath);
            fileEdit.sync_errors = data.errors;
            fileEdit.sync_warnings = data.warnings;
        },
```

I believe this is equivalent to the above pattern, and might be cleaner?

pages/instructorJobSequence/instructorJobSequence.ejs
```diff
@@ -2,7 +2,7 @@
 <html>
   <head>
     <%- include('../partials/head'); %>
-    <script src="/javascripts/socket.io.js"></script>
+    <script src="/node_modules/socket.io-client/dist/socket.io.slim.js"></script>
```
mwest1066, line 5: Awesome! Can we delete `/javascripts/socket.io.js`?

pages/instructorQuestions/instructorQuestions.ejs
```diff
@@ -55,8 +55,12 @@
                         <% questions.forEach(function(question) { %>
                         <tr>
                             <td class="align-middle">
-                                <a href="<%= urlPrefix %>/question/<%= question.id %>/">
-                                    <%= question.title %></a>
+                                <% if (question.sync_errors) { %>
+                                    <i class="fa fa-times text-danger" aria-hidden="true"></i>
```
mwest1066, line 59: Should we add a tooltip or popover to these icons? Maybe a popover that says "This question is not synced. Click here to see the errors."?

mwest1066:
I'm still working through the actual syncing code, but the above comments are some minor things I noticed in passing.

mwest1066:
I added a checkbox to the issue for #2256. I think that we if are doing things correctly in the new sync code then that issue should get fixed for free, but I guess we should add a test to verify it.

mwest1066:
> > * I'm not confident that the issue described in #2063 will be fixed here; both that branch and this branch contain diverging bits of code. Ideally we'd write tests against master to prevent regressions, and then merge those changes into this branch to test for regressions.
> 
> @mwest1066 going to need your help to test this, or at least provide test cases that I could use to reproduce a potential bug.

The old code had the (incorrect) behavior that a JSON array with value `[]` would be synced to a DB array with value `NULL`.

The new code should handle this case correctly and sync it to a zero-length non-NULL array in the DB. In particular, we should get that JSON -> DB sync correctly maps:
```
NULL::jsonb         -> NULL::text[]
'null'::jsonb       -> NULL::text[]
'[]'::jsonb         -> array[]::text[]
'["a"]'::jsonb      -> array['a']::text[]
'["a", "b"]'::jsonb -> array['a', 'b']::text[]
```

(Of course the `text[]` type is only if we are using an array of strings.)

The simplest test here would be to have a course instance access rule with `"uids": []` in the JSON and check that this is correctly synced to a zero-length non-NULL array in the DB. An access rule without `"uids"` specified at all should sync to a NULL array in the DB.

sprocs/sync_assessments.sql
```diff
@@ -20,88 +21,157 @@ DECLARE
     new_assessment_question_id bigint;
     new_assessment_question_ids bigint[];
 BEGIN
-    -- The outermost structure of the JSON blob is an array of assessments
-    FOR assessment IN SELECT * FROM JSONB_ARRAY_ELEMENTS(sync_assessments.assessments) LOOP
-        new_assessment_question_ids := array[]::bigint[];
+    -- Move all our data into a temporary table so it's easier to work with
```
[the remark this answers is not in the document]
mwest1066: Note that we don't need to (and maybe shouldn't) DROP the table first. TEMP tables will locally hide any global tables with the same name.

sprocs/sync_assessments.sql
```diff
@@ -20,88 +21,157 @@ DECLARE
...
+        errors,
+        warnings,
+        data
+    ) SELECT
+        entries->>0,
+        (entries->>1)::uuid,
+        entries->>2,
+        entries->>3,
+        (entries->4)::JSONB
+    FROM UNNEST(disk_assessments_data) AS entries;
```
mwest1066, line 45: I think it's safer to use JSON objects rather than arrays with `jsonb_to_recordset()`, as described in the final two bullet points of the doc section: https://prairielearn.readthedocs.io/en/latest/dev-guide/#database-access

For an example, see: https://github.com/PrairieLearn/PrairieLearn/blob/1b249374d6f15e1468d53ce4e2b1a7efa02650cb/sprocs/sync_news_items.sql#L33

database/tables/questions.pg
```diff
@@ -17,20 +17,24 @@ columns
...
     topic_id: bigint
     type: enum_question_type
-    uuid: uuid not null
+    uuid: uuid
 
 indexes
     questions_pkey: PRIMARY KEY (id) USING btree (id)
-    questions_course_id_number_key: UNIQUE (course_id, number) USING btree (course_id, number)
-    questions_course_id_uuid_key: UNIQUE (course_id, uuid) USING btree (course_id, uuid)
+    questions_course_id_number_nondeleted_key: USING btree (course_id, uuid) WHERE deleted_at IS NULL
```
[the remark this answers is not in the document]
nwalters512: > Is this index supposed to be on `(course_id, number)`?

No, it should be on `uuid`, I just named it wrong 🙂 

> And why is this not UNIQUE?

It's declared as `UNIQUE` in `migrations/174_resilient_syncing.sql` - I don't know why it doesn't show up as that here.

lib/json-load.js
```diff
@@ -1,3 +1,4 @@
+// @ts-check
```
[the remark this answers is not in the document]
nwalters512: This just tells `tsc` that it should check this file!

database/tables/questions.pg
```diff
@@ -17,20 +17,24 @@ columns
...
+    questions_course_id_uuid_nondeleted_key: USING btree (course_id, uuid) WHERE deleted_at IS NULL
```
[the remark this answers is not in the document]
nwalters512: > Why do we want `WHERE deleted_at IS NULL`? Won't this make it expensive to find-and-undelete questions by UUID? 

This new syncing code will allow UUIDs to be duplicated within a course's deleted questions. I frankly don't remember exactly why this is 😂 

> Also, shouldn't this be UNIQUE?

It's declared as such in the migration, I don't know why that isn't captured here.

lib/named-locks.js
```diff
@@ -60,28 +64,62 @@ module.exports.tryLockAsync = util.promisify(module.exports.tryLock);
...
+/**
+ * Convenience function to do some unit of work with a lock. Automatically
+ * releases the lock once `acquiredCallback` finishes executing. If the lock
+ * could not be acquired, `failedCallback` is called with the error.
+ * 
+ * @param {string} name The name of the lock to acquire
+ * @param {(lock: Lock) => Promise<void>} acquiredCallback
+ * @param {(err: Error | null) => void} failedCallback
+ */
+module.exports.doWithLockAsync = async function(name, acquiredCallback, failedCallback) {
```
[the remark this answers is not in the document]
nwalters512: Great suggestion!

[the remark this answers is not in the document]
nwalters512: Actually, looks like this function isn't being used at all - just going to remove this.

package.json
nwalters512: The "different/better" is "functional interface that's more ergonomic" 😂 I forget exactly why I installed this instead of using moment, happy to back this out.

migrations/173_assessments__assessment_set_id__nullable.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE assessments ALTER COLUMN assessment_set_id DROP NOT NULL;
+-- This will verify that any non-deleted assessments have a valid assessment set ID
+-- NOTE: This is potentially very expensive, as all rows will be checked.
+-- Keep this in mind when deploying.
+ALTER TABLE assessments ADD CONSTRAINT assessment_set_id_check CHECK (
```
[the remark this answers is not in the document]
nwalters512: I believe that's right, think it's worth revisiting the `pg_describe` code to capture this?

pages/instructorAssessments/instructorAssessments.ejs
nwalters512: Note that this is `courseInstanceSyncErrorsAndWarnings`, not `assessmentSyncErrorsAndWarnings` - there's actually four different `*SyncErrorsAndWarnings.ejs` partials.

pages/instructorAssessmentAccess/instructorAssessmentAccess.ejs
nwalters512: Good call, that makes sense to me. I think this code predates some of the navbar refactoring that was done "recently", so that probably didn't occur to me the first time around.

pages/instructorFileEditor/instructorFileEditor.js
nwalters512: Yep, this is just 100% wrong - just fixed this.

pages/instructorFileEditor/instructorFileEditor.js
nwalters512: Resolved above.

pages/instructorJobSequence/instructorJobSequence.ejs
nwalters512: It's still in use by a few pages, but I'll replace it with the new script provided by `node_modules`. I'll assume there aren't any courses still using this.

nwalters512:
> The simplest test here would be to have a course instance access rule with `"uids": []` in the JSON and check that this is correctly synced to a zero-length non-NULL array in the DB. An access rule without `"uids"` specified at all should sync to a NULL array in the DB.

I added test for this! See f7fc50d, f115408, and 40f489c. The only bug they uncovered was in the `sync_questions` sproc, which I fixed in 19cf861.

sprocs/sync_assessments.sql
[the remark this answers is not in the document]
nwalters512: Unfortunately we can do exactly this; see https://www.postgresql.org/docs/8.1/sql-createtableas.html:

> The standard defines an ON COMMIT clause; this is not currently implemented by PostgreSQL.

That being said, I agree that it's best to not drop the table first; I'll take that out.

tests/sync/util.js
```diff
@@ -50,21 +50,21 @@ const syncFromDisk = require('../../sync/syncFromDisk');
 
 /**
```
mwest1066, line 51: All of the type definitions seem to appear in both `tests/sync/util.js` and `sync/course-db.js`. Is there any way to have just one copy of these somewhere? I think they might already be out of sync (`Question.dependencies` is only in `sync/course-db.js`?).

mwest1066:
@nwalters512 do you remember which of the un-checked ToDo items are actually important?

tests/sync/util.js
nwalters512: Yeah, I think we can consolidate most of them in `sync/course-db.js`!

nwalters512:
@mwest1066 This is my off-the-cuff take on those items:

* "Document syncing process more fully, including the aforementioned nuances" and "Write complete tests for all strange JSON situations" should almost definitely be done before this lands.
* "Add tests for duplicate assessment sets, tags, and topics" sounds important, though I don't completely remember what this was meant to be.
* "Make sure these changes incidentally fix #2256 and add a test for it" would be nice, but since *not* fixing that wouldn't be a regression, I don't think it needs to be an acceptance criteria for landing this PR.
* "Enable TypeScript type checking for all new files + tests" appears to have been done already! Going to check that off.
* "Completely delete all old syncing code" I *believe* is not done - should probably get done before this merges?

tests/sync/util.js
nwalters512: Ahhh, I remember why I had them separate - the test utils types have many properties marked as optional so that you don't have to include as many of them in type data, whereas the types for the syncing code itself do *not* have as many optional properties because the defaults applied when loading JSON will generally the properties are defined.

If this were actually TypeScript, we could use the full type system and define the syncing types in terms of the other types but with particular properties as optional, buuuut it's not actually TypeScript, so we can't 🙁.

mwest1066: Haha, no worries.

mwest1066:
> * "Document syncing process more fully, including the aforementioned nuances" and "Write complete tests for all strange JSON situations" should almost definitely be done before this lands.

Agreed. That does sound rather important!

> * "Add tests for duplicate assessment sets, tags, and topics" sounds important, though I don't completely remember what this was meant to be.

I guess it just means that we should raise an error if `infoCourse.json` contains:
```
tags = [
        {"name": "mwest", "color": "gray1", "description": "Question originally written by mwest"},
        {"name": "mwest", "color": "gray1", "description": "Question originally written by mwest"}
],
```

> * "Make sure these changes incidentally fix #2256 and add a test for it" would be nice, but since _not_ fixing that wouldn't be a regression, I don't think it needs to be an acceptance criteria for landing this PR.

Let me see if I can make sure we've fixed this.

> * "Completely delete all old syncing code" I _believe_ is not done - should probably get done before this merges?

Yep, no reason not to do this.

mwest1066:
I converted `sprocs/sync_questions.sql` to use the new sync algorithm (Algorithm 1 from `paper_sync`). I implemented this by copying the `code/query.sql` function from `paper_sync` (also shown in Figure 2) and renaming the generic `src`/`dest` variables to `disk_questions`/`questions`. There were a few small tweaks needed (e.g., restricting queries to just the `syncing_course_id`). I also added the consistency checks from `code/check.sql` in `paper_sync` as assertions in `sync_questions.sql`, and I converted the function to use JSON objects for data input and output (which removes the need to use positional parameters in arrays).

All the tests still pass, but I haven't added any new tests yet.

@nwalters512 if the new `sprocs/sync_questions.sql` looks ok to you then I can work on converting the other functions. Is the list of sync functions that use the uuid/string_id syncing algorithm the following functions?

[ ] `sync_assessments.sql`
[ ] `sync_course_instances.sql`

If you are happy with the change to using JSON objects (not arrays) input/output, then I'll convert the input/output for the following functions.

[ ] `sync_assessment_sets.sql`
[ ] `sync_course_tags.sql`
[ ] `sync_question_tags.sql`
[ ] `sync_topics.sql`

nwalters512:
@mwest1066 haven't looked too closely at the syncing code, but as long as the tests are still passing, I feel good about doing everything that way!

> Is the list of sync functions that use the uuid/string_id syncing algorithm the following functions?

Yep, that looks complete to me.

> If you are happy with the change to using JSON objects (not arrays) input/output, then I'll convert the input/output for the following functions.

My goal with using arrays instead of objects was to reduce the size of the (potentially very large) JSON payloads that need to be sent to Postgres. If you don't see any significant increase in syncing times for large courses like CS225 on the prod instance (where DB round trips are more expensive than locally), then I'm fine with that!

mwest1066:
> My goal with using arrays instead of objects was to reduce the size of the (potentially very large) JSON payloads that need to be sent to Postgres. If you don't see any significant increase in syncing times for large courses like CS225 on the prod instance (where DB round trips are more expensive than locally), then I'm fine with that!

Makes sense to me. I reverted the object-data changes, so now we are back to using arrays for all JSON transfer to sync SQL functions.

mwest1066:
In d60c1bbcefe11abd7902060eddf040eed2e52e07 I added a new test that:
* creates an assessment set
* creates an assessment using the assessment set
* syncs
* deletes the assessment set
* syncs
* checks whether the assessment has an auto-created assessment set
This failed (the assessment had a NULL assessment set), but after the changes it passes. The change also removed the need to do a second-pass deletion of assessment sets.

We need to do make the same changes to topic and tag syncing. I added these as checkboxes to the main PR comment.

PR #1645, opened 2019-08-17
Add sample R autograder for externally graded questions

### LINKED ISSUE #970: R autograder
- [x] Add docker image based off of [`prairielearn/centos7-base`](https://hub.docker.com/r/prairielearn/centos7-base/) ([source](https://github.com/PrairieLearn/PrairieLearn/blob/9374dd1d2d65d863be9df21f35b8285166ae5500/environments/centos7-base/Dockerfile), [CS 225](https://github.com/PrairieLearn/PrairieLearn/blob/9374dd1d2d65d863be9df21f35b8285166ae5500/environments/centos7-cs225/Dockerfile)) and [`rocker`](https://github.com/rocker-org/rocker/blob/7c99ef6c45053e703d43f0bcecd9e30f56c72df6/r-base/Dockerfile)
    - [ ] minor port from debian to RHEL-based infrastructure
- [x] Verify code highlighting for _R_ exists c.f. 

https://github.com/PrairieLearn/PrairieLearn/blob/9374dd1d2d65d863be9df21f35b8285166ae5500/exampleCourse/questions/fibonacciEditor/question.html#L15

Mode file located at [`/public/javascripts/ace/mode-r.js`](https://github.com/PrairieLearn/PrairieLearn/blob/9374dd1d2d65d863be9df21f35b8285166ae5500/public/javascripts/ace/mode-r.js)

- [x] Provide a modified [`testthat`](https://cran.r-project.org/package=testthat) (see [`R/reporter-junit.R`](https://github.com/r-lib/testthat/blob/eda2d54a78e75e76f5544b78a14b50e88d926689/R/reporter-junit.R)) that outputs unit test results to JSON c.f. ["grade result" section of external autograders](http://prairielearn.readthedocs.io/en/latest/externalGrading/#grading-result). 

Did I forget anything?

coatless:
Within this PR, we implement a native _R_ autograder for PrairieLearn. Previous implementations relied on the `Python` unit testing framework and manipulation of _R_ objects via `rpy2`. 

This PR is a complement to #1638 and closes #970 

If all goes **well** when checking code, then the student will see:

Screen Shot 2019-08-17 at 10 56 32 PM

If the student does _not_ have a correct answer, then the student will see:
Screen Shot 2019-08-17 at 10 56 41 PM

If we hit an **irrecoverable error**, then we have: 

Screen Shot 2019-08-17 at 10 56 50 PM


exampleCourse/serverFilesCourse/r_autograder/pltest.R
```diff
@@ -0,0 +1,157 @@
...
+# Format the results of each inline test.
+pl_test_list_format = function(i, x) {
+
+    pl_message_statement = function(error, fail) {
+        if(isTRUE(error)) {
+            "Your code encounted an error. Please verify it works."
+        } else if(isTRUE(fail)) {
+            "The code ran but did not produce the correct result."
+        } else {
+            "No errors!"
```
mwest1066: The failing tests report a message of “No errors!” for me, but then the output is “Your output did not match the expected output.” This felt a bit contradictory. Was the message supposed to be “The code ran but did not produce the correct result”?

exampleCourse/serverFilesCourse/r_autograder/pltest.R
```diff
@@ -0,0 +1,157 @@
...
+)
+
+## Format the Results to Match External Grading Format ----
+
+# Format the results of each inline test.
+pl_test_list_format = function(i, x) {
+
+    pl_message_statement = function(error, fail) {
+        if(isTRUE(error)) {
+            "Your code encounted an error. Please verify it works."
```
mwest1066: You don’t show the student the error? I feel like errors should passed back to the student so they can fix them?

exampleCourse/questions/fibonacciEditorWithR/tests/tests.R
```diff
@@ -0,0 +1,47 @@
...
+# In particular, we make the following assumptions about unit tests:
+
+# 1. `describe()` function indicates a single problem steps worth.
+# 2. `it()` function describes the specification being assessed.
+#
+# Avoid using multiple comparisons within the same `describe()`
+
+# Sample test
+describe("1", {                               # Specify points per check
+    it("Does `fib()` exist?", {               # Description of test
```
mwest1066: I don’t think we support markdown here (the backticks around fib()).

exampleCourse/serverFilesCourse/r_autograder/pltest.R
```diff
@@ -0,0 +1,157 @@
...
+        }
+    }
+
+    pl_output_statement = function(passed) {
+        condition = if(passed == 1) {
+            "matched"
+        } else {
+            "did not match"
+        }
+        sprintf("Running test...\nYour output %s the expected output.", condition)
```
mwest1066: You don’t want to tell them the expected and actual output? Seems like debugging would be easier if they were told.

mwest1066: At the moment this information seems very redundant with the message above.

exampleCourse/questions/fibonacciEditorWithR/tests/tests.R
```diff
@@ -0,0 +1,47 @@
...
+        source("../student/student_answer.R")
+        expect_equal(fib(13), 233)
+    })
+})
+
+describe("3", {
+    it("Check fib(n)", {
+        source("../student/student_answer.R")
+        student_ans = fib(15)
+        expect_equal(student_ans, 610)
```
mwest1066: Best practice here would be to randomly generate a few different test values to check, to prevent students hard-coding answers.

ChangeLog.md
```diff
@@ -11,7 +11,9 @@
 
   * Add integration tests for daily cron jobs (Matt West).
 
-  * Add example docker grading images for externally graded questions (James Balamuta).
+  * Add example docker grading images for externally graded questions in `graders/` (James Balamuta).
```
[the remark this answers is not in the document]
coatless: This is just a correction to the previous changelog entry that introduced the `grader/` directory in #1638. The bullet now emphasizes the location of grading images.

exampleCourse/questions/fibonacciEditorWithR/tests/tests.R
coatless: Agreed. I was trying to mimic here the `Python` implementation with unit tests.

exampleCourse/serverFilesCourse/r_autograder/pltest.R
coatless: Agreed. I'll probably add a new layer to parse the error messages.

exampleCourse/questions/fibonacciEditorWithR/tests/tests.R
coatless: This was just styling to pop the function name.

exampleCourse/serverFilesCourse/r_autograder/pltest.R
coatless: I wasn't sure how indepth the error was given the entry on readthedocs. I also wanted to provide a means to avoid having this be viewed as a live code editor.

exampleCourse/serverFilesCourse/r_autograder/pltest.R
coatless: Let me double check this. I think the type strength check `isFALSE` triggers might not match.

exampleCourse/questions/fibonacciEditorWithR/info.json
```diff
@@ -0,0 +1,16 @@
+{
+    "uuid": "8f3e0727-336b-4388-9410-565951b0beff",
+    "title": "Fibonacci Autograding with R",
```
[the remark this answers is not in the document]
coatless: 👍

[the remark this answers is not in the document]
coatless:
```suggestion
    "title": "Fibonacci autograding with the R programming language",
```

coatless:
Out of pure curiosity, how would you feel if I moved the `pltest.R` logic into an _R_ package and deployed it with the `grader-r` image? This would allow for more seamless updates.

coatless:
So, here is the data available for each test:

```r
  .. .. ..$ message          : chr "class(fib) is equal to \"function\".\nEqual"
  .. .. ..$ srcref           : 'srcref' int  16 9 16 44 9 44 16 16
  .. .. .. ..- attr(*, "srcfile")=Classes 'srcfilecopy', 'srcfile' <environment: 0x7fe7eb262788> 
  .. .. ..$ expectation_calls:List of 1
  .. .. .. ..$ : language expect_equal(class(fib), "function")
  .. .. .. .. ..- attr(*, "srcref")= 'srcref' int  16 9 16 44 9 44 16 16
  .. .. .. .. .. ..- attr(*, "srcfile")=Classes 'srcfilecopy', 'srcfile' <environment: 0x7fe7eb262788> 
  .. .. .. ..- attr(*, "start_frame")= num 41
  .. .. ..$ call             :List of 1
  .. .. .. ..$ : language expect_equal(class(fib), "function")
  .. .. .. .. ..- attr(*, "srcref")= 'srcref' int  16 9 16 44 9 44 16 16
  .. .. .. .. .. ..- attr(*, "srcfile")=Classes 'srcfilecopy', 'srcfile' <environment: 0x7fe7eb262788> 
  .. .. .. ..- attr(*, "start_frame")= num 41
  .. .. ..$ start_frame      : num 41
  .. .. ..$ end_frame        : num 41
  .. .. ..$ test             : chr "1: Does `fib()` exist?"
  .. .. ..- attr(*, "class")= chr  "expectation_success" "expectation" "condition"
```

I can expose `message` -- which surfaces the result check information -- and `expectation_calls` -- which surfaces the exact check that is occurring.

coatless:
If all goes **well** when checking code, then the student will see:

Screen Shot 2019-08-17 at 10 56 32 PM

If the student does _not_ have a correct answer, then the student will see:
Screen Shot 2019-08-17 at 10 56 41 PM

If we hit an **irrecoverable error**, then we have: 

Screen Shot 2019-08-17 at 10 56 50 PM

ChangeLog.md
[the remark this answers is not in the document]
mwest1066: Got it!

exampleCourse/questions/fibonacciEditorWithR/tests/tests.R
mwest1066: Looking at `elements/pl-external-grader-results/pl-external-grader-results.mustache`, we include the description with `{{description}}`, so we’ll escape any HTML before it reaches the student. I’m not sure whether we want styling to make it through? @nwalters512?

exampleCourse/questions/fibonacciEditorWithR/tests/tests.R
mwest1066: Yep, I see what you were doing here. I’d like to push these example questions to implement best practices, though, because I think people will copy them.

I think a good practical strategy here would be to add about 3 tests that each generate a random input. For example, replace the `fib(13)` and `fib(15)` checks with ones that choose a random value in ranges `[5,14]`, `[15,24]`, and a new one in `[25,34]`. You could add a comment saying that we choose random values so that students can’t hard-code solutions.

exampleCourse/questions/fibonacciEditorWithR/question.html
```diff
@@ -0,0 +1,30 @@
...
+  <p>
+    The Fibonacci numbers are 1, 1, 2, 3, 5, 8, ..., where each
+    number is the sum of the two before it.
+    \[
+    F_n = F_{n-1} + F_{n-2}
+    \]
+  </p>
+  <p>
+    Write an <b><i>R</i></b> function <tt>fib</tt> that takes a number <tt>n</tt>
+    and returns the n<sup>th</sup> Fibonacci number.
```
mwest1066: As a further improvement over the current Python question, we should be more explicit about numbering (0-indexed versus 1-indexed) by specifying `... the n<sup>th</sup> Fibonacci number, starting from <tt>n = 1</tt>. For example, <tt>fib(3) = 2</tt>.`

exampleCourse/questions/fibonacciEditorWithR/tests/tests.R
```diff
@@ -0,0 +1,47 @@
...
+    it("Does `fib()` exist?", {               # Description of test
+        source("../student/student_answer.R") # Load student submission
+        expect_equal(class(fib), "function")  # Perform test
+    })
+})
+
+describe("1", {
+    it("Check fib(0)", {
+        source("../student/student_answer.R")
+        expect_equal(fib(0), 0)
```
mwest1066: Given that we didn’t actually specify what should happen for `fib(0)` in the question, can we remove this test? I tend to think we should have fixed checks for `fib(1) = 1`, `fib(2) = 1`, `fib(3) = 2`, `fib(4) = 3`, and then the randomized tests listed below.

exampleCourse/serverFilesCourse/r_autograder/pltest.R
mwest1066: Looks great now!

exampleCourse/serverFilesCourse/r_autograder/pltest.R
mwest1066: I think it’s better to err on the side of giving students too much information. While it’s suboptimal if they do live debugging in the editor, I think this is better than them getting frustrated when they can’t pass a test, and they also can’t see what’s wrong. For example, I’ve seen students with working code in a local dev environment, who have a copy-paste problem (e.g., utf8 issues when pasted from Windows) and would be really unable to debug it if they can’t see detailed information about the return values and errors from their code when run in the grader.

I guess we should document this somewhere? 😄

exampleCourse/serverFilesCourse/r_autograder/pltest.R
mwest1066: looks good now.

mwest1066:
The new error reporting looks awesome! Much clearer. The new comments I left above are all about improving things over the Python question (which we should also fix 😄).

exampleCourse/serverFilesCourse/r_autograder/pltest.R
```diff
@@ -0,0 +1,171 @@
...
+# Format the results of each inline test.
+pl_test_list_format = function(i, x) {
+    
+    pl_message_statement = function(x) {
+        if (isTRUE(x$error)) {
+            test_stack = unlist(x$result, recursive = FALSE)
+            error_message = if (!is.null(test_stack) && length(test_stack) != 0) {
+                format(test_stack[[1]])
+            } else {
+                "Error message was unrecoverable."
```
mwest1066: I’m getting this “unrecoverable” error message on any syntax error in the code. For example, leaving off the closing brace in the function, or having `retrn` rather than `return`. Can we display the actual error message to the student in this case? I would have expect some sort of stack trace with `Syntax error: unknown identifier “retrn”` or whatever.

mwest1066:
Sorry for not being clearer above: I don’t think we should change anything about the Python AG or questions as part of this PR. But I _do_ think we should change this R question to be “best practices” (e.g., clarify 0/1 indexing in the question, add some random-input tests).

coatless:
@mwest1066 if it's okay with you, I'd like to potentially move a large portion of `pltest.R` in this PR into an _R_ package called [`plearn`](https://github.com/illinois-r/plearn). This would allow us a bit more flexibility.

mwest1066:
sure!

coatless:
@mwest1066 I've packaged it under `pltest`, c.f. https://github.com/illinois-r/pltest.

I'm going to remove `pltest.R` and directly call it within the `run.sh`, 

```bash
Rscript -e 'pltest::run_testthat()' > results.json
```

As a result, I'll need to modify the `grader-r` image. During the modification, I may transition away from the pre-built versioned `tidyverse` container supplied by rocker to a more slim version via @eddelbuettel.

eddelbuettel:
@coatless and I discussed that today in light of my 'pdf sketch'.  It is six of one versus half-a-dozen of another whether you prefer the Rscript one-liner with a redirect etc, or a mostly constant and clear 10 to 20 line script that is directly accessible, rather than tucked away in a package. Keeping it a script like `run.sh` _may_ make it easier to read for others.  Could be argued either way.

Also in the above `> results.json` is fixed so you can fold that into `pltest::run...()` directly.

And then we're just one step away of making it a shebang script for Rscript (or [littler](https://github.com/eddelbuettel/littler)) taking maybe two cmdline args for input dir (or question dir) and output json.  I do that a _ton_ with daily scripts, cronjobs, scripts in Rocker projects, ...

Anyway, just meant to say I'd be happy to help with containers for this. Let me know if/how I can help.  Our current one is [here](https://github.com/stat430dspm/rocker-pl).

mwest1066:
I’ll let you guys figure out the best way to arrange this code. Just so long as it’s easy for a non-expert to make an R-autograding question 😄

eddelbuettel:
It's pretty close. One approach prefers `testthat` but appears to be in waiting for an upstream change, the other uses `tinytest`, a lighter alternative, and works now. We have a corpus of R question, and that will grow.  I know too little about what you want/need/do inside the PL container.  What we have now is _perfect_ for our needs, runs so much faster (locally, in prod of course the s3 part comes in) and is tuneable as we need it.

coatless:
From the edit....

There are three components to _R_ on PL:

1. Generating data for a question and/or grading it using `rpy2`.
    - This is controlled by [`environments/centos7-plbase`](https://github.com/PrairieLearn/PrairieLearn/blob/f0e36d2df5178b02ff1023d14cacdeca444bc8c0/environments/centos7-plbase/Dockerfile)
2. Externally grading R code. 
3. Both.

Previously, 1.) was used to both **randomize** question parameters and **grade** submissions via `pl-numeric-input`/`pl-integer-input`/ ... As a result, there is a need to have a copy of _R_ associated with PrairieLearn. Otherwise, generating random data as well as computing the answer isn't possible. Since PL uses `python` the ability to generate questions was free.

With this being said, the purpose of the PR was to propose a "best practices" variant of 2.) to serve as an example to build off of. Part of the reason for moving `pltest.R` out of the `serverCourseFiles` was to simplify its adoption and better provide updates via the `pltest` _R_ package. Largely, the system also tried to be similar in scope to the `python` AG variant that had individual tests with varying point values in a single file. 

Again, there needs not be a single solution. This is just what I planned to use and communicated such to you at the start of the semester. At that point, I also indicated why I preferred the approach developed. One of the only issues that was present at the time was the script was already set to launch as being the `ag` user instead of from `root`, c.f. 

https://github.com/PrairieLearn/PrairieLearn/pull/1645/files#diff-76b9f074a65cc7ac6214fb77d489a5c5R47-R52

This impacted the student vs. solution environment issue I mentioned privately in Slack. The solution to this was trivial in the sense that a new _clean_ session with lower permissions is spun up. 

Anyhow, I think that addresses the issue.

eddelbuettel:
We randomize within the external grader. That is not a constraint as far as I can tell, at least we're not bound by.  PL has a hook for something to call, and gets a JSON file back.

coatless:
> We randomize within the external grader. 

That's different though. That's testing a function's _input_. c.f. @mwest1066's remark on 

https://github.com/PrairieLearn/PrairieLearn/pull/1645/files/54d3ee0e87451133fe689f7ab0fb7f2939e0af32#diff-4b16ea519c6fee773650a2ac37ff7e04

> That is not a constraint as far as I can tell, at least we're not bound by. 

No, the goal of the externally graded questions are to supply a function that meet a design standard. Sometimes a simple variant exists that can be answered using `pl-multiple-choice`/`pl-checkbox`/`pl-*-input`. In essence, these latter set of questions benefit the most from having _R_ on PL's core.

eddelbuettel:
Yep. Looks very similar to what we do in `plr`.

PR #1647, opened 2019-08-21
Change blocked-event-loop detection to be more lightweight

mwest1066:


lib/config.js
```diff
@@ -62,6 +62,7 @@ config.serverLoadAverageIntervalSec = 30;
 config.serverUsageIntervalSec = 10;
 config.PLpeekUrl = 'https://cbtf.engr.illinois.edu/sched/proctor/plpeek';
 config.blockedWarnEnable = false;
+config.blockedAtWarnEnable = false;
```
nwalters512: I know the `serverConfig.json` schema has drifted a bit from what the actual config is, but it would be good to add both properties there.

PR #1669, opened 2019-08-31
Add python libraries sklearn and nltk

chen386:


coatless:
Need a changelog entry. What class is this for?

chen386:
CS 105. Why is the checking failing?

coatless:
Need to add an entry into the `ChangeLog.md` file describing the change.

Also, do you need it in `plbase` or `grader-python`?

mwest1066:
@chen386 are you going to also preload these in the python trampoline?

mwest1066:
@chen386 I get a docker build failure with `centos7-plbase/Dockerfile`. Have you tested this?

lib/python-caller-trampoline.py
```diff
@@ -15,7 +15,7 @@
 # Errors are signaled by exiting with non-zero exit code
 # Exceptions are not caught and so will trigger a process exit with non-zero exit code (signaling an error)
 
-import sys, os, json, importlib, copy, base64, io, matplotlib, signal
+import sys, os, json, importlib, copy, base64, io, matplotlib, signal, sklearn, nltk
```
nwalters512: I want someone to audit this for impact on trampoline startup time - of course it's free via forking after the initial load, but I'd just rather know what the impact is.

mwest1066: I believe that local testing by Binglin showed a +200ms impact on worker startup time.

nwalters512:
The current test failures are caused by how we handle images in CI - #1688 will address that. Once that's on master, we can merge back into this branch and verify that the tests are then green.

chen386:
@nwalters512 Thanks for fixing the issue. What issues have `numpy` caused in the past on random seeding?

nwalters512:
@chen386 every forked worker would end up with the same seed, so we would effectively kill randomization. If you look at https://github.com/PrairieLearn/PrairieLearn/blob/42dde84826d6a8263f36d0be5b28c0d1bef6ffa5/lib/python-caller-trampoline.py#L43, we now re-seed `numpy` and the built-in `random` on every fork so we're guaranteed to have "proper" randomness. I looked around a bit and it seems like `sklearn` uses `numpy` for randomness, so this should hopefully not be an issue.

nwalters512:
@chen386 the other PR got merged, if you can pull in master and push the changes, we should be good to go!

nwalters512:
@chen386 this will be available for use locally once the image builds - keep an eye on #announce on Slack to see when it lands on prod.

PR #1690, opened 2019-09-01
Compile native R code in parallel in Docker image

nwalters512:
Experimental change based on https://www.r-bloggers.com/speeding-up-r-packages-installation-process/ - I saw much better build times locally, but I want to see how this performs on CI.


nwalters512:
No real improvement seen, plus it sounds like this isn't really useful for what we're doing.

PR #1696, opened 2019-09-08
Add student-attached files to assessments and questions

mwest1066:
This allows students to attach files to either an assessment instance or to specific question instances.

The main use-case for this is to allow students to attach files to their exams, for later reference during exam review sessions. For example, they could attach the Matlab code they used to solve a question, so they can get help from a TA later about why the code didn’t work.

The core code is also what we will will use for allowing scanned scratch paper to be viewed by students. The proctor-upload page still needs to be implemented.

Despite the only real use-case being for exams, I added the functionality to homeworks as well (it’s almost free). This is intended to allow students to practice with it before the exam.

78F1B77A-5182-4F01-922B-8356B5DBD425


lib/file-store.js
```diff
@@ -0,0 +1,68 @@
+const path = require('path');
+const fsPromises = require('fs').promises;
+const uuidv4 = require('uuid/v4');
+const debug = require('debug')('prairielearn:' + path.basename(__filename, '.js'));
+
+const sqldb = require('@prairielearn/prairielib/sql-db');
+const config = require('../lib/config');
+
+module.exports = {
```
nwalters512: Nit: I've been preferring direct assignments to `module.exports.XXX` recently - it shaves a level of indentation off of function implementations so more can fit on the screen. E.g.

```js
module.exports.upload = ...
```

vs

```js
module.exports = {
    upload: ...
}
```

Not blocking, just for your consideration.

lib/file-store.js
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Upload a file into the file store.
+     *
+     * @param {string} display_filename - The display_filename of the file.
+     * @param {buffer} contents - The file contents.
```
nwalters512: Capital-B `Buffer` for typechecking in the future 😉

lib/file-store.js
```diff
@@ -0,0 +1,68 @@
...
+            instance_question_id,
+            user_id,
+            authn_user_id,
+        ];
+        await sqldb.callAsync('files_insert', params);
+        debug('upload(): inserted files row into DB');
+    },
+
+    /**
+     * Delete a file from the file store.
```
nwalters512: Based on the final `debug()` here, it seems like this is only a soft-deletion and the file is left on disk. If that's the case, that behavior should be documented here as intended.

pages/shared/studentAssessmentInstance.js
```diff
@@ -0,0 +1,28 @@
...
+    async processTextUpload(req, res) {
+        if (!res.locals.assessment_instance.open) throw new Error(`Assessment is not open`);
+        await fileStore.upload(req.body.filename, Buffer.from(req.body.contents), 'student_upload', res.locals.assessment_instance.id, null, res.locals.user.user_id, res.locals.authn_user.user_id);
+    },
+
+    async processDeleteFile(req, res) {
+        if (!res.locals.assessment_instance.open) throw new Error(`Assessment is not open`);
+
+        // Check the requested file belongs to the current assessment instance
+        const l = _.filter(res.locals.file_list, file => (file.id == req.body.file_id));
```
nwalters512: Nit: would prefer a better name than `l`

pages/shared/studentInstanceQuestion.js
```diff
@@ -0,0 +1,74 @@
+const _ = require('lodash');
+const sqldb = require('@prairielearn/prairielib/sql-db');
+
+const fileStore = require('../../lib/file-store');
+
+module.exports = {
+    async getVariantId(req, res) {
```
nwalters512: It's a little unclear to me what this function is doing - it seems to be relying on `callOneRowAsync` to throw if something is invalid? A comment explaining why this is necessary would be good, and maybe an explicit `try`/`catch` with the sproc call would be good too.

tests/testExam.js
```diff
@@ -119,7 +120,72 @@ describe('Exam assessment', function() {
 
     helperExam.startExam(locals);
 
-    describe('6. save correct answer to question addVectors', function() {
+    describe('6. assessment_instance: set attach files page URL', function() {
```
nwalters512: Numbering these seems to require a lot of extraneous diffs when tests are added or removed - is there a reason these were numbered in the past?

nwalters512:
Did a first pass without looking much at functionality/design, going to test this out on my machine and leave additional comments based on that. To continue our recent push for docs, some docs for users explaining this feature would be good to have!

pages/partials/attachFilePanel.ejs
```diff
@@ -0,0 +1,99 @@
...
+    <br>
+    <button class="btn btn-xs btn-secondary" type="button" data-toggle="collapse" data-target="#attachTextCollapse" aria-expanded="false" aria-controls="attachTextCollapse">
+      Attach text <i class="far fa-caret-square-down"></i>
+    </button>
+    <div class="collapse" id="attachTextCollapse">
+      <form method="POST" class="attach-text-form">
+        <p class="small mt-3">
+          Attached text will be saved here for your reference.
+        </p>
+        <input name="filename" value="notes.txt">
```
nwalters512: Adding `class="form-control"` will enable Bootstrap styling.

nwalters512: Accessibility: filename and content inputs should have corresponding labels.

pages/shared/studentAssessmentInstance.js
```diff
@@ -0,0 +1,28 @@
+const _ = require('lodash');
+
+const fileStore = require('../../lib/file-store');
+
+module.exports = {
+    async processFileUpload(req, res) {
+        if (!res.locals.assessment_instance.open) throw new Error(`Assessment is not open`);
+        await fileStore.upload(req.file.originalname, req.file.buffer, 'student_upload', res.locals.assessment_instance.id, null, res.locals.user.user_id, res.locals.authn_user.user_id);
```
nwalters512: I'm going to assume that our upload middleware will handle capping the size of uploaded files?

pages/partials/attachFilePanel.ejs
```diff
@@ -0,0 +1,99 @@
...
+    </p>
+    <% } else { /* assessment_instance is open */ %>
+    <!-- UPLOAD FILE -->
+    <button class="btn btn-xs btn-secondary" type="button" data-toggle="collapse" data-target="#attachFileCollapse" aria-expanded="false" aria-controls="attachFileCollapse">
+      Attach a file <i class="far fa-caret-square-down"></i>
+    </button>
+    <div class="collapse" id="attachFileCollapse">
+      <form class="attach-file-form" name="attach-file-form" method="POST" enctype="multipart/form-data" class="mb-3">
+        <p class="small mt-3">
+          Attached files will be saved here for your reference.
```
nwalters512: This should mention the maximum allowable filesize, if any.

pages/partials/attachFilePanel.ejs
```diff
@@ -0,0 +1,99 @@
...
+          </a>
+        </td>
+        <% if (assessment_instance.open) { %>
+        <td style="width:1%; text-align:right;">
+          <% if (file.type == 'student_upload') { %>
+          <a id="attachFileDeleteButton<%= iFile %>"
+             tabindex="0" class="btn btn-xs btn-secondary attachFileDeleteButton" role="button"
+             data-toggle="popover" data-container="body"
+             data-html="true" data-placement="auto" title="Confirm delete"
+             data-content="<%= include('attachFileDeleteForm',
```
nwalters512: Nit: this is less important for smaller lists like this one, but I'd still prefer that we use the popover style used by the instructor gradebook popovers (attach data to button with `data-*` attributes, create and launch popover in code). Not blocking merge.

pages/partials/attachFilePanel.ejs
```diff
@@ -0,0 +1,99 @@
...
+  <div class="card-header bg-secondary text-white">Attached files</div>
+  <table class="table table-sm">
+    <tbody>
+    <% if (file_list.length == 0) { %>
+    <tr><td><i>No attached files</i></td></tr>
+    <% } else { %>
+      <% file_list.forEach(function(file, iFile) { %>
+      <tr>
+        <td style="word-break:break-all;">
+          <i class="far fa-file-alt"></i>
```
nwalters512: Icon seems to be superfluous, could probably be removed.

lib/file-store.js
mwest1066: Avoiding indentation 👍 

I switching in all new `.js` files introduced in this PR.

pages/shared/studentAssessmentInstance.js
mwest1066: done

lib/file-store.js
mwest1066: done

lib/file-store.js
mwest1066: done

pages/shared/studentInstanceQuestion.js
mwest1066: done

tests/testExam.js
mwest1066: The numbering was originally because I was annoyed at grepping for the test that was failing, and unique numbers made this easier. I agree the diff noise (and need to edit) is very annoying as well 😄. I guess this is all part of my general dissatisfaction with Jasmine-type test runners (mocha, jest, etc). I wish they would be less clever and just print error messages right when they happen. I also wish they didn’t do their async contortions, which force us into the horrible `locals` hack inside our tests, rather than us being able to simply pass variables around. Ok, enough of me ranting about JS test frameworks. I’m happy to drop the test numbering anytime, but I won’t both removing it now unless you think this is a good time to do it?

pages/partials/attachFilePanel.ejs
mwest1066: done

pages/shared/studentAssessmentInstance.js
mwest1066: yep, it already does (`multer` handles this in `server.js`)

pages/partials/attachFilePanel.ejs
mwest1066: done

mwest1066: I was worried that adding labels above these is just going to spread out the whole form vertically, and make it less clear what’s going on. I was hoping that the placeholder text in the content input was enough to clarify its purpose. As an example of using a placeholder rather than a label, see the “reply” box I’m typing in right now 😄

The filename input is a stronger case for a label, but I couldn’t really see where to put one and not be confusing. I’m open to suggestions on this!

pages/partials/attachFilePanel.ejs
mwest1066: I’m also not a huge fan of this icon. However, the reason I put it there is that on the assessment overview page, when a question has attached files, we put the same icon at the end of the question title to indicate that there are attachments. To make it visually clear how that overview-page icon connects to the actual files, I put the same icon next to each file. Without this, I worried that the link wouldn’t be clear. More feedback on this is welcome!

nwalters512: What about a paperclip? I think that's a pretty universal symbol for "attachment".

pages/partials/attachFilePanel.ejs
mwest1066: Legit. I will skip for now.

pages/partials/attachFilePanel.ejs
mwest1066: Agreed. I switched the icon to a paperclip. We now show a paperclip next to the question title on the assessment overview page, and next to the attached files themselves on the question page. Does this work for you? An alternative would be to put the icon next to the card title on the question page, for example, rather than next to each file.

nwalters512: I think It'd prefer it in the card header - I do think that it makes sense to have the icon be *somewhere* when attached files are involved.

pages/partials/attachFilePanel.ejs
nwalters512: I'll compromise on an `aria-label` attribute on each (which the box I'm typing in now has 😛)

pages/partials/attachFilePanel.ejs
```diff
@@ -0,0 +1,100 @@
...
+        </p>
+        <input class="form-control" name="filename" value="notes.txt">
+        <div class="form-group">
+          <textarea class="form-control" rows="5" name="contents" placeholder="Type or paste text here"></textarea>
+        </div>
+        <% if (typeof variant != 'undefined') { %>
+        <input type="hidden" name="__variant_id" value="<%= variant.id %>">
+        <% } %>
+        <input type="hidden" name="__csrf_token" value="<%= __csrf_token %>">
+        <div class="form-group text-right">
```
nwalters512: Nit: I'd prefer a left-aligned button. For a single-column form, the eye wants to naturally keep moving down the page after the inputs, and putting the button on the right makes it slightly harder to notice IMO.

pages/shared/studentInstanceQuestion.js
```diff
@@ -3,72 +3,83 @@ const sqldb = require('@prairielearn/prairielib/sql-db');
...
-    async getVariantId(req, res) {
-        const variant_id = req.body.__variant_id;
-        const params = [
-            variant_id,
-            res.locals.instance_question.id,
-        ];
+/*
+ * Get a validated variant_id from a request, or throw an exception.
+ *
+ * This function assumes req.body.__variant_id has been sent by the client, but
```
nwalters512: Good comment - thanks for this!

server.js
```diff
@@ -128,11 +130,14 @@ if (config.hasAzure) {
...
-        fieldSize: 1e7, // max in bytes
-        fileSize: 1e7, // max in bytes
-        parts: 100, // max number of fields + files
+        fieldSize: config.fileUploadMaxBytes,
+        fileSize: config.fileUploadMaxBytes,
+        parts: config.fileUploadMaxParts,
     },
 });
+config.fileUploadMaxBytesFormatted = filesize(config.fileUploadMaxBytes, {base: 10, round: 0});
+app.use(function(req, res, next) {res.locals.config.fileUploadMaxBytes = config.fileUploadMaxBytes; next();});
```
nwalters512: I'm a little confused on how this works - down on L167, we assign `res.locals.config = config`. Could you double check that?

pages/partials/attachFilePanel.ejs
mwest1066: Excellent point!

pages/partials/attachFilePanel.ejs
mwest1066: done

pages/shared/studentInstanceQuestion.js
mwest1066: Thanks for the prompt to do it! These subtle authz things really need clear docs.

pages/partials/attachFilePanel.ejs
mwest1066: good point, done

server.js
mwest1066: Haha, many thanks! I completely failed to realized that we were already making `config` available as `locals.config`. I thought to myself “it would be very convenient if config was available inside locals” 😄 — for whatever reason I only made a few fields explicitly available.

I removed all this duplication.

PR #1697, opened 2019-09-11
Show connecting IP

trombonekenny:
Adds the Express connecting IP address to the authenticated user information shown in the instructor "Change effective user page"

There hasn't been a way for the user (Instructor) to see which IP address they are connecting from. Making that more visible than grepping through server logs will make troubleshooting exam mode network configurations easier.

### DIFF
--- a/pages/instructorEffectiveUser/instructorEffectiveUser.js
+++ b/pages/instructorEffectiveUser/instructorEffectiveUser.js
@@ -19,6 +19,7 @@ router.get('/', function(req, res, next) {
     sqldb.queryOneRow(sql.select, params, function(err, result) {
         if (ERR(err, next)) return;
         _.assign(res.locals, result.rows[0]);
+        
         res.locals.ipaddress = req.headers['x-forwarded-for'] || req.ip;
         // Trim out IPv6 wrapper on IPv4 addresses
         if (res.locals.ipaddress.substr(0, 7) == '::ffff:') {
--- a/pages/instructorEffectiveUser/instructorEffectiveUser.js
+++ b/pages/instructorEffectiveUser/instructorEffectiveUser.js
@@ -19,7 +19,7 @@ router.get('/', function(req, res, next) {
     sqldb.queryOneRow(sql.select, params, function(err, result) {
         if (ERR(err, next)) return;
         _.assign(res.locals, result.rows[0]);
-        res.locals.ipaddress = req.ip;
+        res.locals.ipaddress = req.headers['x-forwarded-for'] || req.ip
         // Trim out IPv6 wrapper on IPv4 addresses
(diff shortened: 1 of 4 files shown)


pages/instructorEffectiveUser/instructorEffectiveUser.js
```diff
@@ -19,6 +19,11 @@ router.get('/', function(req, res, next) {
     sqldb.queryOneRow(sql.select, params, function(err, result) {
         if (ERR(err, next)) return;
         _.assign(res.locals, result.rows[0]);
+        res.locals.ipaddress = req.ip;
```
mwest1066: Our current reverse-proxy setup sends proxy IP as `req.ip` and puts the real IP in the `x-forwarded-for` header. Our current method for getting the real IP is:
https://github.com/PrairieLearn/PrairieLearn/blob/master/middlewares/authzCourseInstance.js#L18

trombonekenny: I had thought we had `app.set('trust proxy', true);` added to PL, but maybe not? That's the magic to make Express do the proxying translations for us. http://expressjs.com/en/guide/behind-proxies.html

server.js
```diff
@@ -88,6 +88,7 @@ if (config.blockedAtWarnEnable) {
 const app = express();
 app.set('views', path.join(__dirname, 'pages'));
 app.set('view engine', 'ejs');
+app.set('trust proxy', true);
```
mwest1066: Won’t this enable any client to spoof a CBTF IP by setting `X-Forwarded-For` to a CBTF address?

That is, if a client sends a request with `X-Forwarded-For = [Some-CBTF-IP]`, Apache will add the true client IP to the right-hand end of the `X-Forwarded-For` list, but `trust proxy` will make Express trust the first (left-hand) entry (the spoofed entry set by the client).

I think maybe we want `app.set(‘trust proxy’, ‘localhost’)`, but I haven’t tested it (or thought very hard about it).

An alternative would be to add a config variable and do `app.set(‘trust proxy’, config.trustProxyHops)`, where for us `config.trustProxyHops = 1`. I haven’t tested this either.

trombonekenny: That's fair. PS uses `app.set('trust proxy', 'loopback')` which should be safe here too?

mwest1066: Maybe? What does that do exactly?

trombonekenny:
In the spirit of getting something deployed today/tonight to help Wayne in ZJU, I reverted the Express level changes and copied the IP identifying code that's already in place to the instructorEffectiveUser page.

mwest1066:
sounds sensible — it’s failing CI for some reason?

trombonekenny:
Linter caught a missing semicolon.

PR #1702, opened 2019-09-13
Add https infrastructure support inside Docker

trombonekenny:
In case we needed `https://localhost:3000` support working with the Docker development containers, this PR adds the package and script support to setup the keys PrairieLearn is expecting. It generates a self-signed localhost key where PrairieLearn is expecting it.

This purposefully doesn't configure PL to use https. That needs a `config.json` `"serverType": "https"` config. But if that config exists, this should take care of everything needed in the container to support that.

My primary use for this is testing LTI endpoints which might balk at insecure connections.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -27,6 +27,8 @@
 
   * Add connecting IP address report in instructor effective user page (Dave Mussulman).
 
+  * Add SSL https support inside the Docker container (Dave Mussulman).
+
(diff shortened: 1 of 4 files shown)


trombonekenny:
This might also be a good test for your GitHub actions work, since it needs a centos7-plbase rebuild to work.

mwest1066:
@trombonekenny can you rebase this to trigger the checks?

Issue #1625, opened 2019-08-04
Unicode is not handled correctly by in-browser file editing


mwest1066:
An example is `questions/derivativeUpload/question.html` in `pl-dev101`. It has a unicode `MINUS SIGN` in it (code point `0x2212`) which is displayed as the UTF-8 encoding of `0xE2 0x88 0x92` rather than the unicode character.

PR #1707, opened 2019-09-16
Fix unicode bug in editor

### LINKED ISSUE #1705: Instructor file editor crashes server with invalid data
At 00:55:05 UTC today, PL crashed due to an uncaught exception in the instructor file editor:

```
/PrairieLearn/pages/instructorFileEditor/instructorFileEditor.js:250
return decodeURIComponent(Buffer.from(str, 'base64').toString().split('').map((c) => {
^
URIError: URI malformed
at decodeURIComponent (<anonymous>)
at b64DecodeUnicode (/PrairieLearn/pages/instructorFileEditor/instructorFileEditor.js:250:12)
at writeEdit (/PrairieLearn/pages/instructorFileEditor/instructorFileEditor.js:443:19)
at /PrairieLearn/pages/instructorFileEditor/instructorFileEditor.js:427:13
at /PrairieLearn/node_modules/async/dist/async.js:2952:28
at replenish (/PrairieLearn/node_modules/async/dist/async.js:444:21)
at iterateeCallback (/PrairieLearn/node_modules/async/dist/async.js:428:21)
at /PrairieLearn/node_modules/async/dist/async.js:325:20
at /PrairieLearn/node_modules/async/dist/async.js:2957:17
at /PrairieLearn/pages/instructorFileEditor/instructorFileEditor.js:422:17
at /PrairieLearn/node_modules/@prairielearn/prairielib/lib/sql-db.js:452:9
at Query.<anonymous> (/PrairieLearn/node_modules/@prairielearn/prairielib/lib/sql-db.js:423:17)
at Query.handleReadyForQuery (/PrairieLearn/node_modules/@prairielearn/prairielib/node_modules/pg/lib/query.js:125:10)
at Connection.<anonymous> (/PrairieLearn/node_modules/@prairielearn/prairielib/node_modules/pg/lib/client.js:244:19)
at Connection.emit (events.js:203:13)
at Connection.EventEmitter.emit (domain.js:471:20)
```

tbretl:
Changes the way that base64 is encoded and decoded to properly handle unicode.

Fixes #1705 and #1625.


tests/testFileEditor.js
```diff
@@ -212,6 +215,20 @@ const verifyEditData = [
 describe('test file editor', function() {
     this.timeout(20000);
 
+    describe('test base64-util lib, on which file editor depends', function() {
```
nwalters512: I'd prefer a test of the editor itself with such a string so that we don't rely on current implementation details.

ChangeLog.md
```diff
@@ -341,6 +341,8 @@
 
   * Fix flaky test cases in `pl-number-input` which didn't handle comparison options properly (Mingjie Zhao).
 
+  * Fix bug in unicode encode/decode in file editor (Tim Bretl).
```
nwalters512: Entry should be under the **__next version__** section.

pages/instructorFileEditor/instructorFileEditor.js
```diff
@@ -236,24 +237,8 @@ router.post('/', (req, res, next) => {
...
-function b64DecodeUnicode(str) {
-    // Going backwards: from bytestream, to percent-encoding, to original string.
-    return decodeURIComponent(Buffer.from(str, 'base64').toString().split('').map((c) => {
-        return '%' + ('00' + c.charCodeAt(0).toString(16)).slice(-2);
-    }).join(''));
-}
-
 function getHash(contents) {
-    return b64EncodeUnicode(sha256(contents).toString());
+    return b64Util.b64EncodeUnicode(sha256(contents).toString());
```
nwalters512: Not really relevant to this PR, but why is the hash base-64 encoded?

tests/testFileEditor.js
```diff
@@ -212,6 +215,20 @@ const verifyEditData = [
...
+        it('should encode a challenging string without crashing', function() {
```
nwalters512: Nit: I think we should be more conservative with our usage `it(...)` blocks; this is the main reason our test output is so long. Not as important for just three tests, but in general I want to move towards grouping more testing inside a single `it(...)` block.

mwest1066:
Rather than including this blob of code from MDN, could we use the `base-64` npm library [0], together with the `utf8` library [1] to first convert to UTF-8? Alternatively, we could also accept corrupted unicode using WTF-8 [2].

[0] https://www.npmjs.com/package/base-64
[1] https://github.com/mathiasbynens/utf8.js
[2] https://github.com/mathiasbynens/wtf-8

tests/testFileEditor.js
tbretl: done

ChangeLog.md
tbretl: done

pages/instructorFileEditor/instructorFileEditor.js
tbretl: you're right, let's revisit this later ;-)

tests/testFileEditor.js
tbretl: fair enough, and this test has been removed anyway

PR #1710, opened 2019-09-17
Add <pl-hide-in-panel> element.

### LINKED ISSUE #1694: Add a more flexible way to specify which panels something is rendered in
Right now, a portion of a question can only show up in either a) every panel, or b) exactly 1 panel; there isn't a way to render something in exactly 2 panels. I propose a new `<pl-panel>` element that's more flexible than this. Example:

```html
<pl-panel question="true" answer="true">
  <p>I will only show up in the question and answer panels!</p>
</pl-panel>
```

For this question, the inner `<p>` will only be rendered in the question and answer panels.

We could provide a few variants of the interface for better user ergonomics:

* `<pl-panel question="false">`: shows in submission/answer, but not question
* `<pl-panel question="true">`: shows only in question; equivalent to `<pl-question-panel>`

More specifically, if some panel is specified as `false`, we'll default to showing in any remaining panels; if some panel is specified as `true`, we'll default to showing only in panels that are specifically opted in to.

However, this may be complicating the interface too much. Feedback welcome!

mwest1066:
Closes #1694

### DIFF
--- a/doc/elements.md
+++ b/doc/elements.md
@@ -1119,7 +1119,7 @@ Common reasons that trigger the display of this element are:
 
 ## `pl-hide-in-panel` element
 
-Hide the contents in any of the panels ("question", "submission", or "answer").
+Hide the contents so that it is **not** displayed in specific panels ("question", "submission", or "answer").
 
 #### Sample Element
 
--- a/doc/elements.md
+++ b/doc/elements.md
@@ -212,11 +212,11 @@ Attribute | Type | Default | Description
 `rtol` | number | 1e-2 | Relative tolerance for `comparison="relabs"`.
 `atol` | number | 1e-8 | Absolute tolerance for `comparison="relabs"`.
 `digits` | integer | 2 | number of digits that must be correct for `comparison="sigfig"` or `comparison="decdig"`.
-`allow-complex` | boolean | False | Whether or not to allow complex numbers as answers. If the correct answer `ans` is a complex object, you should use `import prairielearn as pl` and `data['correct_answer'][answers-name] = pl.to_json(ans)`.
-`show-help-text` | boolean | True | Show the question mark at the end of the input displaying required input parameters.
-`show-placeholder` | boolean | True | Show the placeholder text that shows the default comparison.
+`allow-complex` | boolean | false | Whether or not to allow complex numbers as answers. If the correct answer `ans` is a complex object, you should use `import prairielearn as pl` and `data['correct_answer'][answers-name] = pl.to_json(ans)`.
+`show-help-text` | boolean | true | Show the question mark at the end of the input displaying required input parameters.
+`show-placeholder` | boolean | true | Show the placeholder text that shows the default comparison.
 `size` | integer | 35 | Size of the input box.
-`show-correct-answer` | boolean | True | Whether to show the correct answer in the submitted answers panel.
+`show-correct-answer` | boolean | true | Whether to show the correct answer in the submitted answers panel.
 
 #### Example implementations
 
@@ -292,7 +292,7 @@ Attribute | Type | Default | Description
 `label` | text | — | A prefix to display before the input box (e.g., `label="$F =$"`).
 `display` | "block" or "inline" | "inline" | How to display the input field.
 `variables` | string | — | A comma-delimited list of symbols that can be used in the symbolic expression.
-`allow-complex` | boolean | False | Whether complex numbers (expressions with `i` or `j` as the imaginary unit) are allowed.
+`allow-complex` | boolean | false | Whether complex numbers (expressions with `i` or `j` as the imaginary unit) are allowed.
 `imaginary-unit-for-display` | string | `i` | The imaginary unit that is used for display. It must be either `i` or `j`. Again, this is *only* for display. Both `i` and `j` can be used by the student in their submitted answer, when `allow-complex="true"`.
 
 #### Details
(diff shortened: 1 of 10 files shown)


mwest1066:
I’m trying out a less flexible design here than that outlined in #1694. My thinking is to name the element according to the problem people are trying to solve with it. In this case, someone wants to hide information and stop it from showing up in certain panels. Opinions on naming and attributes would be most welcome!

elements/pl-hide-in-panel/pl-hide-in-panel.py
```diff
@@ -0,0 +1,28 @@
+import prairielearn as pl
+import lxml.html
+
+
+QUESTION_DEFAULT = False
```
coatless: Thanks for making defaults available!

doc/elements.md
```diff
@@ -1104,7 +1108,53 @@ Common reasons that trigger the display of this element are:
...
+- [`pl-hide-in-panel` to hide contents in one or more display panels.](#pl-hide-in-panel-element)
+- [`pl-external-grader-results` for showing the results from an externally graded code question.](#pl-external-grader-results-element)
+
+
+-----
+
+
+## `pl-hide-in-panel` element
+
+Hide the contents in any of the panels ("question", "submission", or "answer").
```
coatless: Maybe:

```suggestion
Set when contents should be displayed depending the current view state ("question", "submission", or "answer").
```

exampleCourse/questions/examplesPanels/question.html
```diff
@@ -0,0 +1,77 @@
...
+      <li><b><tt>&lt;pl-hide-in-panel&gt;</tt></b> has attributes that allow content to be hidden in one or more panels.</li>
+  </ul>
+
+  <p>In normal usage, we put the question prompt in a <tt>&lt;pl-question-panel&gt;</tt> element and all inputs requesting student input outside of this element.</p>
+
+  <h5>Example HTML</h5>
+
+  <p>As an example, this question contains the follow panel elements.  Click the "Save &amp; Grade" button to make the submission and answer panels appear.</p>
+<pre>
+&lt;pl-question-panel&gt;
```
coatless: I think the use of `&lt;`/`&gt;` is to avoid the tag being picked up. If that's not the case, it might be a bit better to wrap this within the new `<markdown></markdown>` tag.

exampleCourse/questions/examplesPanels/question.html
```diff
@@ -0,0 +1,77 @@
...
+  <p>(D) This text is hidden in the question panel, and shown in the others.</p>
+</pl-hide-in-panel>
+
+<pl-hide-in-panel submission="True">
+  <p>(E) This text is hidden in the submission panel, and shown in the others.</p>
+</pl-hide-in-panel>
+
+<pl-hide-in-panel answer="True">
+  <p>(F) This text is hidden in the answer panel, and shown in the others.</p>
+</pl-hide-in-panel>
```
coatless: Maybe add an example with two states?  e.g.

`<pl-hide-in-panel question="True" answer = "True">`

and, then show it matches

`<pl-submission-panel>`?

exampleCourse/questions/examplesPanels/info.json
```diff
@@ -0,0 +1,7 @@
+{
+    "uuid": "81a3e61d-9ecc-457c-9dac-db5b6b451d3c",
+    "title": "Show or hide text in specific panels",
```
coatless: I think examples normally have the name of the element in the title? e.g.

`Element action via pl-element`

`Real and complex numeric input via pl-number-input`

exampleCourse/questions/examplesPanels/question.html
mwest1066: You are correct, this is just HTML escaping. Markdown would be an alternative, but would still require escaping to avoid switching into HTML.

exampleCourse/questions/examplesPanels/info.json
mwest1066: I tried this initially, but we discuss four elements and the title got crazy long with everything listed 😄

I didn’t want to only discuss `pl-hide-in-element` because I wanted to show the list of alternatives.

exampleCourse/questions/examplesPanels/question.html
mwest1066: It’s cute that these match, but it doesn’t seem very practical? I wanted to focus on “here’s what to do to get the effect you want”, if that makes sense?

doc/elements.md
```diff
@@ -1104,7 +1108,53 @@ Common reasons that trigger the display of this element are:
...
+<pl-hide-in-panel submission="True" answer="True">
+  This text will be hidden in the submission panel and answer panel.
+</pl-hide-in-panel>
+```
+
+#### Customizations
+
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`question` | boolean | False | Whether to hide the element contents in the question panel.
```
coatless: One nit: I think booleans are listed as `true`/`false elsewhere in the doc?

exampleCourse/questions/examplesPanels/question.html
```diff
@@ -0,0 +1,77 @@
...
+
+&lt;pl-submission-panel&gt;
+  &lt;p&gt;(B) This text is only shown in the submission panel.
+&lt;/pl-submission-panel&gt;
+
+&lt;pl-answer-panel&gt;
+  &lt;p&gt;(C) This text is only shown in the answer panel.
+&lt;/pl-answer-panel&gt;
+
+&lt;pl-hide-in-panel question="True"&gt;
```
coatless: Maybe lowercase `true` to match other elements use of bools?

doc/elements.md
mwest1066: done

mwest1066:
@coatless no big refactors are in progress, hence this PR 😄

doc/elements.md
mwest1066: Good catch! In fact we were being totally inconsistent. I standardized everything on lowercase.

exampleCourse/questions/examplesPanels/question.html
mwest1066: done

Issue #1708, opened 2019-09-16
Instructor file editing crashes server (v2)


mwest1066:
Current `master` (with #1707 merged) on `pl-dev` crashes when attempting to edit a file:

```
/PrairieLearn/node_modules/aws-sdk/lib/request.js:31
throw err;
^
Error [TypeError]: string.charCodeAt is not a function
at ucs2decode (/PrairieLearn/node_modules/utf8/utf8.js:14:19)
at Object.utf8encode [as encode] (/PrairieLearn/node_modules/utf8/utf8.js:88:20)
at Object.module.exports.b64EncodeUnicode (/PrairieLearn/lib/base64-util.js:5:31)
at Response.<anonymous> (/PrairieLearn/pages/instructorFileEditor/instructorFileEditor.js:309:57)
at Request.<anonymous> (/PrairieLearn/node_modules/aws-sdk/lib/request.js:364:18)
at Request.callListeners (/PrairieLearn/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
at Request.emit (/PrairieLearn/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
at Request.emit (/PrairieLearn/node_modules/aws-sdk/lib/request.js:683:14)
at Request.transition (/PrairieLearn/node_modules/aws-sdk/lib/request.js:22:10)
at AcceptorStateMachine.runTo (/PrairieLearn/node_modules/aws-sdk/lib/state_machine.js:14:12)
at /PrairieLearn/node_modules/aws-sdk/lib/state_machine.js:26:10
at Request.<anonymous> (/PrairieLearn/node_modules/aws-sdk/lib/request.js:38:9)
at Request.<anonymous> (/PrairieLearn/node_modules/aws-sdk/lib/request.js:685:12)
at Request.callListeners (/PrairieLearn/node_modules/aws-sdk/lib/sequential_executor.js:116:18)
at Request.emit (/PrairieLearn/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
at Request.emit (/PrairieLearn/node_modules/aws-sdk/lib/request.js:683:14) {
message: 'string.charCodeAt is not a function',
code: 'TypeError',
time: 2019-09-16T12:59:45.424Z
}
```

mwest1066:
I suspect that this issue is showing up here because CI tests use local file storage, while prod uses S3. We should change to use `lib/file-store.js` in both cases, so that CI and prod are identical.

[...]

PR #1711, opened 2019-09-18
Stop using s3 in editor

### LINKED ISSUE #1625: Unicode is not handled correctly by in-browser file editing

tbretl:
This PR fixes two bugs.

(1) The server-side base64 encoding and decoding was implemented incorrectly, and crashed the server when applied to text containing certain unicode characters (e.g., unicode minus). The implementation is now correct. It is also wrapped in a try/catch so if anything ever does go wrong, it won't crash the server.

(2) The first attempt at a fix introduced an even worse bug, as a consequence of forgetting that the AWS SDK reads and writes to S3 with a buffer and not a string. In production, this crashed the server on every attempt to edit a file. This second bug is also now fixed.

The reason the second bug wasn't caught by CI is that CI used local storage and not S3. To avoid this issue in future, this PR also switches from using S3 (in prod) and local storage (in dev) to using only `file-store.js` (in both prod and dev). This required a slight change to the `file-store` API.

For real this time, I hope:
Fixes #1625 and #1708


lib/base64-util.js
```diff
@@ -0,0 +1,30 @@
...
+    // (1) use encodeURIComponent to get percent-encoded UTF-8
+    // (2) convert percent encodings to raw bytes
+    // (3) convert raw bytes to Base64
+    try {
+        return btoa(encodeURIComponent(str).replace(/%([0-9A-F]{2})/g, (match, p1) => {
+            return String.fromCharCode('0x' + p1);
+        }));
+    } catch(e) {
+        logger.error(`b64EncodeUnicode: failed to encode ${str}`);
+        return null;
```
mwest1066: Are callers prepared to handle `null`? Might be safer to return an empty string?

lib/file-store.js
```diff
@@ -63,3 +66,27 @@ module.exports.delete = async (file_id, authn_user_id) => {
     await sqldb.callAsync('files_delete', params);
     debug('delete(): soft-deleted row in DB');
 };
+
+/**
+ * Get a file from the file store.
+ *
+ * @param {number} file_id - The file to get.
```
mwest1066: Specify that the return value is a `Buffer` (or an object containing a Buffer and a file object)?

doc/installingNative.md
```diff
@@ -88,6 +88,8 @@ npm run lint-js -s
 npm run lint-python -s
 ```
 
+* If you want to use the in-browser file editor while running natively, set `config.filesRoot` in `PrairieLearn/lib/config.js` to a convenient, existing directory. Temporary files will be saved there. Note that it is not necessary to set `config.filesRoot` to use `npm run test` - a separate directory will be assigned during testing.
```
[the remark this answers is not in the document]
tbretl: done

lib/file-store.js
```diff
@@ -63,3 +66,27 @@ module.exports.delete = async (file_id, authn_user_id) => {
...
+    if (result.rows.length < 1) {
+        throw new Error(`No file with file_id ${file_id}`);
+    }
+
+    if (config.filesRoot == null) throw new Error(`config.filesRoot must be non-null to get file_id ${file_id} from file store`);
+    const filename = path.join(config.filesRoot, result.rows[0].storage_filename);
+
+    debug(`get(): readFile ${filename} and return buffer with contents`);
+    const contents = await fsPromises.readFile(filename);
+    return contents;
```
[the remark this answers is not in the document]
tbretl: done

lib/base64-util.js
tbretl: they are, but fair enough - empty string it is

lib/file-store.js
tbretl: done

Issue #817, opened 2017-09-17
PL v3 roadmap to release


tbretl:
# Roadmap
- Development
- Examples
- Documentation
- Announcement

# Development

## Code auto-graders
- [x] Change the way docker containers are run to make them faster (switch from AWS Batch to a custom docker-runner)
- [x] Change the way we pass information to docker (currently, we use networking to pull things in from S3, which forces the docker to have network enabled, which we don't want, so we're going to add stuff to the docker image from the custom docker runner, code pushing/pulling will be done by this custom docker runner)
- [x] Change the directory structure inside the docker container (finalize the conventions for names inside the docker container - pieces of code that are pulled into this container get put in a bunch of confusing places)
- [x] For a small number of standard languages (e.g., python) provide a PrairieLearn auto-grader that improves on what's used in the example Fibonacci questions (#940)
- [x] Modify Fibonacci example questions to use the PrairieLearn auto-grader (#940)
- [x] Create tools so both instructors and admins can see things like: job times, queue length, etc.
- [x] Parameterize `test/` files before copying to the container somehow? Jake has questions with many test cases, but each variant only wants a small subset of them.
- [x] Re-submitting for a variant should cancel the outstanding jobs, which should websocket back to show that they were canceled.
- [x] Shut off networking on the grading Docker containers.
- [x] Report an issue whenever an autograder has an error.
- [x] Run multiple jobs on each grader instance.
- [x] Report "job start" back to PrairieLearn so we can estimate the number of pending jobs.
- [x] Add per-question Docker config
  - [x] Time limits
  - [x] Memory limits

## Javascript support
- Question code
    - [x] Enable the use of `server.js` as an alternative to `server.py` (error if both are present)
    - [x] Enable header dependencies, same as elements (e.g., see `info.json` in `pl_file_editor`)
    - [x] Create at least one example question with client-side code (e.g., a drawing question)
- Element code
    - [x] Enable the use of `<pl_elementname.js>` as an alternative to `<pl_elementname.py>` (error if both are present)

## Error handling and debugging
- [x] Implement error handling with websockets
- [x] Distinguish stdout - e.g., calls to `print()` - from errors and implement display with websockets
- [x] Create linter that does not allow code with `print()` calls (for example) to be pushed to production
- [x] Add a debug button (visible to instructors) that toggles display of lots of information about exactly what is happening (including, at minimum, the contents of `data` as it changes)
- Log more things to the errors table:
[...]

PR #1715, opened 2019-09-21
Create `pl-python-variable` element

nicknytko:
Creating an element to display the values of Python variables, specifically Pandas dataframes.  I'm not too attached to the name, so I'm open to suggestions to change it.
It will pull variables from `data['params']`, and has the option to display as text, or as a dataframe.

Example screenshots:
Screen Shot 2019-09-20 at 8 05 38 PM
Screen Shot 2019-09-20 at 8 27 27 PM


doc/elements.md
```diff
@@ -709,6 +709,50 @@ Please let the PrairieLearn developers know if you need a language that is not o
...
+`variable-name` | string | — | The name of the key in `data['params']` to get a value from
+`variable-type` | string | `text` | The type of variable that should be shown.  Either `text` for a textual representation as defined by `repr(var)`, or `dataframe` for a pandas dataframe.
+`variable-prefix` | string | (empty) | Any prefix to append to the output in `text` mode.
+`variable-suffix` | string | (empty) | Any suffix to append to the output in `text` mode.
+`no-highlight` | string | False | Disable syntax highlighting in `text` mode.
+
+#### Details
+
+As of right now, the element supports displaying either Pandas dataframes as an HTML table or Python objects via `repr()`.  When setting a parameter to a dataframe, use Pandas' built in `dataframe.to_json()`.
+
```
coatless: Please add:

```
#### Example implementations

- [`examplesCodeOutput`: Examples of all customization options for the element.](https://github.com/PrairieLearn/PrairieLearn/tree/master/exampleCourse/questions/examplesCodeOutput)

#### See also

- [`pl-code` to display blocks of code with syntax highlighting](#pl-code-element)
- [`pl-variable-output` for displaying a matrix or element in code form.](#pl-variable-output-element)
```

---

Instead of Sample 1 / Sample 2, just `Sample Element`

exampleCourse/questions/exampleCodeOutput/question.html
```diff
@@ -0,0 +1,19 @@
+<pl-question-panel>
```
coatless: File name should be `examplesCodeOutput` instead of `exampleCodeOutput` (note the `s`).

exampleCourse/questions/exampleCodeOutput/question.html
```diff
@@ -0,0 +1,19 @@
...
+  <p>
+    The <code>pl-code-output</code> element allows you to easily format the values of parameters.
+  </p>
+  <p>
```
coatless: Please provide a paragraph at the top describing the element. For each feature, please try to block them using 

```html
<div class="card my-2">
  <div class="card-body">
    <pl-question-panel>
      <p>Some details about what feature the element is displaying.</p>
      <pl-element></pl-element>
    </pl-question-panel>
  </div>
</div>
```

Sample implementation question would be `examplesNumberInput` c.f. 

https://github.com/PrairieLearn/PrairieLearn/blob/a8c431440408dc5cce6a934cc1043a5860201319/exampleCourse/questions/examplesNumberInput/question.html#L1-L19

elements/pl-code-output/pl-code-output.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def prepare(element_html, data):
+    element = lxml.html.fragment_fromstring(element_html)
+    pl.check_attribs(element, required_attribs=['variable-name'], optional_attribs=['variable-type', 'no-highlight', 'variable-prefix', 'variable-suffix'])
+
+
+def render(element_html, data):
+    element = lxml.html.fragment_fromstring(element_html)
+    vartype = pl.get_string_attrib(element, 'variable-type', 'text')
```
coatless: Please define a global variable for each optional attributes above the `prepare()` function. 

As an example, the`variable-type` attribute would have a global variable of `VARIABLE_TYPE_DEFAULT = 'text'`.

elements/pl-code-output/pl-code-output.py
```diff
@@ -0,0 +1,37 @@
...
+
+    varout = data['params'][varname]
+    html = ''
+
+    # render the output variable
+    if vartype == 'dataframe':
+        varout = pd.read_json(varout)
+        html += varout.to_html(classes=['pl-code-output-table']) + '<p class="pl-code-output-table-dimensions">' + str(varout.shape[0]) + ' rows x ' + str(varout.shape[1]) + ' columns</p><br>'
+    elif vartype == 'text':
+        no_highlight = pl.get_boolean_attrib(element, 'no-highlight', False)
```
coatless: Globals, see above.

exampleCourse/questions/exampleCodeOutput/breast-cancer-train.dat
```diff
@@ -0,0 +1,300 @@
+842302,M,17.99,10.38,122.8,1001,0.1184,0.2776,0.3001,0.1471,0.2419,0.07871,1.095,0.9053,8.589,153.4,0.006399,0.04904,0.05373,0.01587,0.03003,0.006193,25.38,17.33,184.6,2019,0.1622,0.6656,0.7119,0.2654,0.4601,0.1189
```
coatless: Is there a smaller data set possible? Or could this data set be downsampled? (~20 observations?)

elements/pl-code-output/pl-code-output.py
```diff
@@ -0,0 +1,40 @@
+import prairielearn as pl
+import lxml.html
+import pandas as pd
+
+VARIABLE_TYPE_DEFAULT = 'text'
+NO_HIGHLIGHT_DEFAULT = False
+
```
coatless: Missing:

'variable-prefix'
 'variable-suffix'

exampleCourse/courseInstances/Sp15/assessments/plelementexamples/infoAssessment.json
```diff
@@ -44,7 +44,8 @@
                 {"id": "examplesMatrixLatex",              "points": 2, "maxPoints": 10},
                 {"id": "examplesPrairieDrawFigure",        "points": 2, "maxPoints": 10},
                 {"id": "examplesVariableOutput",           "points": 2, "maxPoints": 10},
-                {"id": "examplesFileEditor",               "points": 2, "maxPoints": 10}
+                {"id": "examplesFileEditor",               "points": 2, "maxPoints": 10},
+                {"id": "examplesCodeOutput",               "points": 2, "maxPoints": 10}
```
coatless: Thanks for adding this!

elements/pl-code-output/pl-code-output.py
```diff
@@ -0,0 +1,40 @@
...
+    html = ''
+
+    # render the output variable
+    if vartype == 'dataframe':
+        varout = pd.read_json(varout)
+        html += varout.to_html(classes=['pl-code-output-table']) + '<p class="pl-code-output-table-dimensions">{} rows x {} columns</p><br>'.format(str(varout.shape[0]), str(varout.shape[1]))
+    elif vartype == 'text':
+        no_highlight = pl.get_boolean_attrib(element, 'no-highlight', NO_HIGHLIGHT_DEFAULT)
+        prefix = pl.get_string_attrib(element, 'variable-prefix', '')
+        suffix = pl.get_string_attrib(element, 'variable-suffix', '')
```
coatless:
```suggestion
        suffix = pl.get_string_attrib(element, 'variable-suffix', VARIABLE_SUFFIX_DEFAULT)
```

elements/pl-code-output/pl-code-output.py
```diff
@@ -0,0 +1,40 @@
...
+    varout = data['params'][varname]
+    html = ''
+
+    # render the output variable
+    if vartype == 'dataframe':
+        varout = pd.read_json(varout)
+        html += varout.to_html(classes=['pl-code-output-table']) + '<p class="pl-code-output-table-dimensions">{} rows x {} columns</p><br>'.format(str(varout.shape[0]), str(varout.shape[1]))
+    elif vartype == 'text':
+        no_highlight = pl.get_boolean_attrib(element, 'no-highlight', NO_HIGHLIGHT_DEFAULT)
+        prefix = pl.get_string_attrib(element, 'variable-prefix', '')
```
coatless:
```suggestion
        prefix = pl.get_string_attrib(element, 'variable-prefix', VARIABLE_PREFIX_DEFAULT)
```

nicknytko: just added this now :)

doc/elements.md
```diff
@@ -709,6 +709,59 @@ Please let the PrairieLearn developers know if you need a language that is not o
...
+```
+
+![](elements/pl-code-output2.png)
+
+#### Customizations
+
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`variable-name` | string | — | The name of the key in `data['params']` to get a value from
+`variable-type` | string | `text` | The type of variable that should be shown.  Either `text` for a textual representation as defined by `repr(var)`, or `dataframe` for a pandas dataframe.
```
mwest1066: Now that you extended `pl.to_json()` to handle DataFrames, I think that we can automatically determine the type, right? But I can see that we might want to either output `repr()` or the HTML table for a DataFrame, so we need an attribute to control this. Would it be a good idea to replace `type` with `formatted=“true”` to indicate that we want “formatted” output (the HTML table). The `formatted` attribute would default to false. This would also allow us to add special handling of numpy arrays (for example) in the future, where they could be rendered also as HTML tables.

I’m not set on the `formatted` name — this could also be `text` or `repr` or something else. The main idea is just that it shouldn’t be specific to DataFrames, and it’s not really about the variable type, but rather about how it’s formatted for display.

doc/elements.md
```diff
@@ -709,6 +709,59 @@ Please let the PrairieLearn developers know if you need a language that is not o
...
+`variable-prefix` | string | (empty) | Any prefix to append to the output in `text` mode.
+`variable-suffix` | string | (empty) | Any suffix to append to the output in `text` mode.
```
mwest1066: Is the `variable-` prefix on the attributes `variable-prefix` and `variable-suffix` important, or would it be equally good to make them shorter by just calling them `prefix` and `suffix`?

doc/elements.md
```diff
@@ -709,6 +709,59 @@ Please let the PrairieLearn developers know if you need a language that is not o
...
+`no-highlight` | string | False | Disable syntax highlighting in `text` mode.
```
mwest1066: Could we make this just `highlight`, defaulting to true?

nicknytko: This can be changed, I was doing `no-highlight` to maintain consistency with `pl-code`, though.

mwest1066: Oh cool, good point, let’s be consistent and leave it as is.

nwalters512:
I'm not completely sold on the name - I think something emphasizing that this is Python-specific (for if/when we add more languages) and that it's showing a value (and not output from running code, which is what first came to mind) would be good. Some ideas:

* `pl-show-python-variable`
* `pl-display-python-variable`
* `pl-python-variable`
* `pl-python-repr`

Open to other ideas, but I'd really like to see the name changed before shipping.

nicknytko:
> I'm not completely sold on the name - I think something emphasizing that this is Python-specific (for if/when we add more languages) and that it's showing a value (and not output from running code, which is what first came to mind) would be good. Some ideas:
> 
> * `pl-show-python-variable`
> * `pl-display-python-variable`
> * `pl-python-variable`
> * `pl-python-repr`
> 
> Open to other ideas, but I'd really like to see the name changed before shipping.

I’m fond of `pl-python-variable` unless someone else has any ideas? @mwest1066

mwest1066:
I agree with @nwalters512 on the name, and I also like `pl-python-variable` the best!

coatless:
RE: element name. 

I'm not sure differentiating the tag by language is needed as `server` is going to always be python-based. So, no matter what the result will have to end up in Python. Thoughts?

doc/elements.md
nicknytko: I've set this to just `text`, defaulting to `False`.

nwalters512:
@coatless Server will definitely not always be Python-based, we have explicit plans to add support for JS-based servers in the future.

coatless:
@nwalters512 are the plans publicly documented?

nwalters512:
@coatless I think "JavaScript support" on https://github.com/PrairieLearn/PrairieLearn/issues/817 is the most we have.

nicknytko:
Is there any consensus on keeping python in the name or making it more generic?  I don’t mind either way and I can change the element name.

coatless:
@nicknytko the `python` name change is preferred. No need to switch it back. (Unless @mwest1066 or @nwalters512 asks.)

nicknytko:
Sure, in that case the name change should be in the PR already and we should be good for a (hopefully) final round of review 🙂

elements/pl-python-variable/pl-python-variable.py
```diff
@@ -0,0 +1,44 @@
+import prairielearn as pl
+import lxml.html
+
+NO_HIGHLIGHT_DEFAULT = False
+PREFIX_DEFAULT = ''
+SUFFIX_DEFAULT = ''
+FORCE_TEXT_DEFAULT = False
```
mwest1066: We are trying to keep the `_DEFAULT` names exactly the same as `<attribute-name>_DEFAULT`, so remove `FORCE_`?

PR #1717, opened 2019-09-22
Pl drawing

mfsilva22:
This element adds a canvas where pre-defined images can be added (for example, the rectangle and triangle below), and can also expect objects to be inserted for grading. 

image


doc/drawingElement.md
```diff
@@ -0,0 +1,1186 @@
...
+`answers-name` | string | — | Variable name to store data in. This variable is required when `preview = false`.
+`width` | integer | 580 | Horizontal width of the canvas.
+`height` | integer | 320 | Vertical height of the canvas.
+`grid-size` | integer | 20 | Size of the square grid for the canvas background. If `grid-size = 0`, then the background is white.
+`snap-to-grid` | boolen | true | Objects placed in the canvas snap to closest grid point.
+`correct-answer` | string | special | Correct answer for grading. Defaults to `data["correct-answers"][answers-name]`.
+`tol` | float | `0.5*grid-size` | Tolerance to check the position of objects. The error is calculated as the absolute difference between expected position and submitted one.
+`angle-tol` | float | 10 | Tolerance to check angles (in degrees). The error is calculated as the absolute difference between expected angle and submitted one.
+`show-tolerance-hint` | boolean | true | Show tolerance hint under the canvas. The default is `true` when `preview = false`
```
mwest1066:
```suggestion
`show-tolerance-hint` | boolean | true | Show tolerance hint under the canvas. The default is `true` when `preview = false`.
```

doc/drawingElement.md
```diff
@@ -0,0 +1,1186 @@
+
+# Elements to setup canvas (drawing space)
+
+## `pl-drawing` element
+
+A `pl-drawing` element displays a canvas, which can be used to display initial drawing objects or expect drawing objects for grading.
+
+The system of coordinates of the canvas is located at the top/left corner.
```
mwest1066: Could we add an image to illustrate the coordinate system?

doc/elements.md
```diff
@@ -1014,6 +1016,48 @@ Attribute | Type | Default | Description
 
 -----
 
+## `pl-drawing` element
+
+Creates a canvas (drawing space) that can display images from a collection of pre-defined drawing objects.
+Users can also add drawing objects to the canvas for grading.
```
mwest1066: I suggest deleting all the text below this point and just having a sentence that explicitly says “See the pl-drawing documentation for details” (with a link).

elements/pl-drawing/info.json
```diff
@@ -0,0 +1,18 @@
+{
+    "controller": "pl-drawing.py",
+    "dependencies": {
+        "coreScripts": [
+            "sylvester.js",
+            "sha1.js",
+            "underscore.js"
+        ],
+        "elementScripts": [
+            "fabric.require.js",
```
mwest1066: Is it possible to install `fabric` with `npm`, rather than including a packaged source distribution directly?

package-lock.json
```diff
@@ -627,7 +627,7 @@
             "dependencies": {
                 "buffer": {
                     "version": "4.9.1",
-                    "resolved": "https://registry.npmjs.org/buffer/-/buffer-4.9.1.tgz",
+                    "resolved": "http://registry.npmjs.org/buffer/-/buffer-4.9.1.tgz",
```
mwest1066: I’m not sure why there are changes to `package-lock.json`? Can they be reverted? Maybe revert them and then run `npm ci` to make sure you are back to the default package install?

doc/drawingElement.md
```diff
@@ -0,0 +1,1186 @@
...
+
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`preview` | boolean | false | `preview = true` is used when canvas is used for display only. `preview = false` expects objects to be placed in the canvas for grading.
+`answers-name` | string | — | Variable name to store data in. This variable is required when `preview = false`.
+`width` | integer | 580 | Horizontal width of the canvas.
+`height` | integer | 320 | Vertical height of the canvas.
+`grid-size` | integer | 20 | Size of the square grid for the canvas background. If `grid-size = 0`, then the background is white.
+`snap-to-grid` | boolen | true | Objects placed in the canvas snap to closest grid point.
+`correct-answer` | string | special | Correct answer for grading. Defaults to `data["correct-answers"][answers-name]`.
```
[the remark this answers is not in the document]
coatless: Doesn't this go against the `correct-answer` naming scheme established here: 

https://prairielearn.readthedocs.io/en/latest/devElements/#element-functions

doc/drawingElement.md
```diff
@@ -0,0 +1,1186 @@
...
+#### Sample Element
+
+```html
+<pl-drawing preview=true width="200" height="200">
+    <pl-drawing-initial>
+        <pl-coordinates x1="60" y1="80" label="0" angle="30"></pl-coordinates>
+    </pl-drawing-initial>
+</pl-drawing>
+```
+<img src="elements/pl-coordinates.png" width=50%>
```
coatless: I think the overarching doc style primarily uses Markdown to include images instead of the HTML `` tag.

```md
![Example of `pl-coordinates`](elements/pl-coordinates.png)
```

 The downside is the [`mkdocs` build](https://github.com/mkdocs/mkdocs/issues/1385) doesn't allow for attributes as it uses [`Python-Markdown`](https://python-markdown.github.io/), which adheres to the [original Markdown spec](https://daringfireball.net/projects/markdown/syntax#img). What is the need to resizing the image? Could it be accomplished via CSS?

doc/drawingElement.md
```diff
@@ -0,0 +1,1186 @@
...
+`height` | integer | 320 | Vertical height of the canvas.
+`grid-size` | integer | 20 | Size of the square grid for the canvas background. If `grid-size = 0`, then the background is white.
+`snap-to-grid` | boolen | true | Objects placed in the canvas snap to closest grid point.
+`correct-answer` | string | special | Correct answer for grading. Defaults to `data["correct-answers"][answers-name]`.
+`tol` | float | `0.5*grid-size` | Tolerance to check the position of objects. The error is calculated as the absolute difference between expected position and submitted one.
+`angle-tol` | float | 10 | Tolerance to check angles (in degrees). The error is calculated as the absolute difference between expected angle and submitted one.
+`show-tolerance-hint` | boolean | true | Show tolerance hint under the canvas. The default is `true` when `preview = false`
+`tolerance-hint` | text | "The expected tolerance is 1/2 square grid." | Hint explaining tolerance used when grading objects.
+
+
```
coatless: Many entries are missing:

````
# Example Implementations

- [link to example question in exampleCourse]().

# See also
- related elements
````

I would say the most helpful is link to fill out is the example implementation in a working question.

File: doc/drawingElement.md (hunk 6 above)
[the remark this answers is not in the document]
Comment (mfsilva22) [reply doc/drawingElement.md]: I agree. I think we should keep `correct-answer`, which is consistent with other elements, for example, `pl-number-input`.

File: elements/pl-drawing/colors.py [hunk 9]
```diff
@@ -0,0 +1,41 @@
+colors = {
+    'red1': '#ffccbc',
```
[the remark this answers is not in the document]
mfsilva22: Yes, I did get these from PL (following @mwest1066 suggestion). I agree that it would be better to have a shared configuration.

nicknytko:
> The drawing element and its components are _amazing_.
> 
> Few minor comments regarding docs, defaults (to anyone), and files.
> 
> Most notably, is there any reason for choosing `.png` over `.svg` for each element listed in `pl-drawing/clientFilesElement`?
> 
> There might be a desire to also remove the PTX from git that contains image screenshots. c.f. `elements/pl-drawing/elements/WIP-pl-figures.pptx`

Replaced icons with SVG's and removed the extra ppt

nwalters512:
This is so cool! Haven't looked at the implementation yet, but have some UX notes/thoughts based on playing around with the questions. These are not necessarily blockers, just things I noticed as a first-time user of this feature, and students are likely to notice them too. Very open to landing this as-is and committing to enhancement over time.

* I tried a couple times to drag a line/point/etc onto the canvas before realizing that they're buttons that you click to add something.
* Using the "help line" (not sure what we're calling that) isn't super intuitive - I'd expect to be able to grab an endpoint and move it wherever I want while keeping the other endpoint in one place. However, it looks like the only way to manipulate is either through translation, scaling, or rotation, not the more "natural" method I'm describing. This seems to function differently than the "graded line with circle endpoints". Could we unify behind one form of interaction, or else add a "help line with circle endpoints"?
* It's not immediately obvious that the "X" button under "help buttons" will remove the currently selected element. I spent a while right clicking, hitting the "delete" key, etc. trying to get rid of things before noticing that button. It seems to be unique among the buttons in that it removes something instead of adding. It also seems like deleting should be built into this element instead of being something that an instructor has to manually add to every question.
* I'd argue that the same is true for the help objects - those should be available by default. It shouldn't be up to the instructor to predict how a student might want to solve a problem.

nicknytko:
> This is so cool! Haven't looked at the implementation yet, but have some UX notes/thoughts based on playing around with the questions. These are not necessarily blockers, just things I noticed as a first-time user of this feature, and students are likely to notice them too. Very open to landing this as-is and committing to enhancement over time.
> 
>     * I tried a couple times to drag a line/point/etc onto the canvas before realizing that they're buttons that you click to add something.
> 
>     * Using the "help line" (not sure what we're calling that) isn't super intuitive - I'd expect to be able to grab an endpoint and move it wherever I want while keeping the other endpoint in one place. However, it looks like the only way to manipulate is either through translation, scaling, or rotation, not the more "natural" method I'm describing. This seems to function differently than the "graded line with circle endpoints". Could we unify behind one form of interaction, or else add a "help line with circle endpoints"?
> 
>     * It's not immediately obvious that the "X" button under "help buttons" will remove the currently selected element. I spent a while right clicking, hitting the "delete" key, etc. trying to get rid of things before noticing that button. It seems to be unique among the buttons in that it removes something instead of adding. It also seems like deleting should be built into this element instead of being something that an instructor has to manually add to every question.
> 
>     * I'd argue that the same is true for the help objects - those should be available by default. It shouldn't be up to the instructor to predict how a student might want to solve a problem.

I agree with your points that the interface could probably be made more intuitive, and I'd be happy to work on that in a future PR(s) since I'd like to get the functionality out sooner rather than later so courses in TAM can start creating questions with this.

mfsilva22:
> * I tried a couple times to drag a line/point/etc onto the canvas before realizing that they're buttons that you click to add something.

You are not alone! Students had mentioned this several times when I first implemented the drawing tool in TAM 251 :-)

> * Using the "help line" (not sure what we're calling that) isn't super intuitive - I'd expect to be able to grab an endpoint and move it wherever I want while keeping the other endpoint in one place. However, it looks like the only way to manipulate is either through translation, scaling, or rotation, not the more "natural" method I'm describing. This seems to function differently than the "graded line with circle endpoints". Could we unify behind one form of interaction, or else add a "help line with circle endpoints"?
> * It's not immediately obvious that the "X" button under "help buttons" will remove the currently selected element. I spent a while right clicking, hitting the "delete" key, etc. trying to get rid of things before noticing that button. It seems to be unique among the buttons in that it removes something instead of adding. It also seems like deleting should be built into this element instead of being something that an instructor has to manually add to every question.
> * I'd argue that the same is true for the help objects - those should be available by default. It shouldn't be up to the instructor to predict how a student might want to solve a problem.

Most of the tools made sense for TAM 251, and were created to address specific needs. But I agree that it doesn't make sense for other cases, and there is definitely lots of room for improvement.

I agree with Nicolas above, it would be better if we can move this to PL sooner, to allow TAM 2XX instructors to start using the tool. They are currently using an old version, and we have already fixed some bugs in the current version. We can work on improvements later (and I am sure we will get more feedback from instructors as well).

I will be working on the documentation tomorrow, to address some of the comments above.

nwalters512:
Glad to hear it isn't just me! As I said, I'm happy shipping this without these changes, just wanted to provide my perspective as a first-time user. This is some great work and I'm excited for it to be iterated on!

As an aside: looking at these huge amounts of code makes me wish we had a build process for client code where it could be transpiled + minified. Of course, what we have now is totally adequate. Just an idea for the future.

doc/drawingElement.md
```diff
@@ -0,0 +1,1186 @@
...
+<pl-drawing answers-name="fbd">
+</pl-drawing>
+```
+<img src="elements/pl-drawing.png" width=70%>
+
+#### Customizations
+
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`preview` | boolean | false | `preview = true` is used when canvas is used for display only. `preview = false` expects objects to be placed in the canvas for grading.
```
mfsilva22: Should we agree on a name before we make a change? What about `gradable` and set it to `false` by default?

doc/drawingElement.md
mfsilva22: I think we should keep the name as `correct-answer`. Can someone confirm this?

doc/drawingElement.md
mfsilva22: added the figure

doc/elements.md
mfsilva22: done.

doc/drawingElement.md
coatless: From Slack... Turns out there is an option:

```md
![alt](path/to/image.png){: width="30%" }
```
c.f. https://python-markdown.github.io/extensions/attr_list/#inline

We might need to add to the `mkdocs.yml` config:

```yaml
markdown_extensions:
  - markdown.extensions.attr_list: true
```

c.f.
https://python-markdown.github.io/extensions/
https://www.mkdocs.org/user-guide/configuration/#formatting-options

doc/drawingElement.md
mfsilva22: I added links to the documentation. If possible, other improvements to the documentation could be done later.

doc/drawingElement.md
mfsilva22: As discussed via Slack, we will work on these changes at a later time. Thanks for the input!

doc/drawingElement.md
mwest1066: `gradable` (default `false`) sounds excellent to me!

doc/drawingElement.md
mwest1066: agreed!

PR #1724, opened 2019-09-26
make devMode login a full auth class

### LINKED ISSUE #1509: dev_user login mode
A challenge of development of auth systems is the config.authType == 'none' overrides some other options - for example, if you wanted to easily test LTI or password based authentication.

This issue proposes stripping out the magic in `middlewares/authn.js` that makes "Dev User" auth work and make it a full `pages/authCallbackDev` method. We'd still want to make sure it only triggers in certain cases, but having it be peers with other auth systems would make testing and developing other auth systems easier.

If we want to make the Docker user experience the same, perhaps default unprompted to Dev auth (like it does now), but on logout you can choose other systems?

trombonekenny:
When testing other authentication modes in development, it was difficult to toggle between the new auth class and the devMode "Dev User" hard-wired authentication. Sometimes it was helpful to have both at the same time.

This PR breaks the devMode login out to look like all the other auth classes, so it can co-exist with them. If `devMode` is true and `config.authType` is `none` (the default local docker experience) it redirects through to login as Dev User.

I don't think this is any less secure than it was before, but now it's more configurable.

I need help figuring out what to change to enable the tests to pass as they did before.

Still to do:
- [x] library-ize the sql call and its returns so it's consistent across the files that call it
- [x] fix tests

Closes #1509


trombonekenny:
I figured out that `request` doesn't store cookies by default and that's why my auth changes weren't being honored by the test suite. Rather than change all of the request calls to store a dev login cookie, I created an `authType = 'testrun'` setting which doesn't need cookie auth (like our previous `none`) and set it to that value when doing the tests.

trombonekenny:
Here's a different take on this PR - changing the defaults so Logout works for all methods, redirects back to the login page directly, and `middlewares/authn.js` only triggers the authType none special handling if an auth cookie isn't already present.

This seems to let different auth systems interact without changing the defaults or breaking testing. i.e. I can be in authType `none` and have an effective LTI session. Logout of that, select the Dev Login link, and I have admin rights.

trombonekenny:
Closes #1509 (just noticed the issue and PR weren't linked). Bump to revisit now that the new edit-in-browser PR shipped.

trombonekenny:
Still to do: library-ize the sql call and its returns so it's consistent across the files that call it

middlewares/authn.js
```diff
@@ -60,44 +60,36 @@ module.exports = function(req, res, next) {
...
             });
-        });
-        return;
-    }
+            return;
+        }
 
-    // otherwise look for auth cookies
-    if (req.cookies.pl_authn == null) {
+        // not in autoType == none
```
mwest1066, line 92:
```suggestion
        // not in authType == none
```

middlewares/authzCourseInstance.js
```diff
@@ -14,7 +14,7 @@ module.exports = function(req, res, next) {
         course_instance_id: req.params.course_instance_id,
         is_administrator: res.locals.is_administrator,
         req_date: res.locals.req_date,
-        ip: req.ip,
+        ip: req.headers['x-forwarded-for'] || req.ip,
```
mwest1066, line 17: Is this still needed, now that #1913 is merged?

pages/authLogout/authLogout.js
```diff
@@ -7,7 +7,7 @@ router.get('/', function(req, res, _next) {
     if (res.locals.authn_provider_name == 'Shibboleth') {
         res.redirect('/Shibboleth.sso/Logout');
     } else {
-        res.redirect('/');
+        res.redirect('/pl/login');
```
mwest1066, line 10: Does this interfere with homepage cookie-clearing? https://github.com/PrairieLearn/PrairieLearn/blob/517be1007699db7ab2d5c823a5912fa00b39d045/server.js#L228

server.js
```diff
@@ -206,6 +206,7 @@ app.use('/pl/shibcallback', require('./pages/authCallbackShib/authCallbackShib')
 app.use('/pl/azure_login', require('./pages/authLoginAzure/authLoginAzure'));
 app.use('/pl/azure_callback', require('./pages/authCallbackAzure/authCallbackAzure'));
 app.use('/pl/lti', require('./pages/authCallbackLti/authCallbackLti'));
+app.use('/pl/devlogin', require('./pages/authLoginDev/authLoginDev'));
```
mwest1066, line 209: Should we wrap this in `if (config.devMode) {}`?

mwest1066:
I added some todo items (using the `[ ]` syntax) to the main description, so that it shows up in the PR list as still needing work.

server.js
trombonekenny: We could, but the first line of that file checks for devMode and errors.

trombonekenny: https://github.com/PrairieLearn/PrairieLearn/blob/e214adadb9f3fdbdb93b30819aa99b36deb19a3a/pages/authLoginDev/authLoginDev.js#L10

mwest1066: Yep, I checked 😄. But I find it nice when skimming through `server.js` to see what is mounted in different modes: https://github.com/PrairieLearn/PrairieLearn/blob/517be1007699db7ab2d5c823a5912fa00b39d045/server.js#L241

I would also keep the check in place in the page code too. We _really_ don't want this route to accidentally be enabled in prod!

pages/authLogout/authLogout.js
trombonekenny: I think this is required so a logout specifically triggers another login. If it didn't and redirected to /, the "if no cookie and devMode, assume dev login" would stop any other login.

I think it's independent from the homepage cookie-clearing, but that's something we'd need to test. I'm not sure how.

server.js
trombonekenny: It's consistent with how the other auth methods work if they're enabled or not (we don't alter the route, just if it fires anything in the target page). I'm okay to add a check in server.js, though, if the skimming makes it more obvious?

pages/authLogout/authLogout.js
mwest1066: I'm fine to redirect to `/pl/login`, but that's going to prevent the cookie-clearing code from working. I think we should add some other way of triggering that.

server.js
mwest1066: It's not crucial, but I do like to be able to easily see what's active when we are in prod versus dev.

pages/authLogout/authLogout.js
trombonekenny: I think we might be triggering this because the test that's failing is when it's trying to emulate a student. I'm investigating.

Could logout just clear all of the cookies?

mwest1066: Sure, clearing cookies on logout would be fine, but I think you might need to bounce via another page to do it?

mwest1066:
> Closes #1509 (just noticed the issue and PR weren't linked). Bump to revisit now that the new edit-in-browser PR shipped.

I think that GitHub only picks up these comments if they are in the top-level description box on the issue page, so you need to edit that in place. The right-hand column has a section called "Linked issues" where it shows you what it's detected.

pages/authLogin/authLogin.js
```diff
@@ -23,7 +20,12 @@ router.get('/', function(req, res, next) {
         // this will also need to depend on which institution we have
         // detected (e.g., UIUC doesn't want Azure during exams, but
         // ZJUI does want it).
-        res.render(__filename.replace(/\.js$/, '.ejs'), res.locals);
+
+        sqldb.query(sql.get_course_instances, {}, (err, result) => {
```
mwest1066, line 24: Can we avoid running this in production?

pages/authLoginDev/authLoginDev.js
```diff
@@ -18,4 +22,41 @@ router.get('/', function(req, res, next) {
...
+
+    let course_instance_id = req.body.course_instance_id ? req.body.course_instance_id : 1;
+    var params = [
+        'devMode-LTI-user',
+        'devMode-LTI-user',
+        course_instance_id,
+        'devMode-LTI-user',
+        'devMode-contextId',
+    ];
+
```
mwest1066, line 36:
```suggestion
```

pages/authLogin/authLogin.sql
```diff
@@ -2,14 +2,19 @@
 SELECT COALESCE($force_mode, ip_to_mode($ip, $req_date)) AS mode;
 
 -- BLOCK get_course_instances
+WITH lti_institution AS (
+  SELECT * FROM institutions WHERE short_name = 'LTI'
```
mwest1066, line 6:
```suggestion
    SELECT * FROM institutions WHERE short_name = 'LTI'
```

nwalters512:
@trombonekenny would you still like to get this merged? It's not clear if the new permissions model makes the obsolete - I admittedly haven't looked too closely at either.

trombonekenny:
Closing in favor of #7094

PR #1726, opened 2019-09-27
Add tzlocal as a python library under core

coatless:
New dependency of `rpy2`

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -33,6 +33,8 @@
 
   * Add student Gradebook page (Matt West).
 
+  * Add python library `tzlocal` (James Balamuta).
+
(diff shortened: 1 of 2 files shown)


mwest1066:
I’m not necessarily opposed to adding this library, but for my own curiosity, why do you want it? The local timezone is UTC on the PL servers. If you just want the TZ database then `pytz` might be what you want?

coatless:
@mwest1066 it's an "optional hard" dependency for `rpy2` when using `pandas`

c.f. https://bitbucket.org/rpy2/rpy2/issues/458/tzlocal-dependency

mwest1066:
@coatless can you rebase this to current master to fix the tests?

coatless:
@mwest1066 this is good to go!

PR #1727, opened 2019-09-27
R questions

coatless:
Adds:

- [x] Grabbing a data set in _R_ 
- [x] RNG seed usage in _R_
- [x] "Hello to R PL world" with adding 2 numbers
- [x] Using pre-generated _R_ data to fit a linear regression.


mwest1066:
Does this replace #1645? If so, could we close that earlier PR?

coatless:
@mwest1066 probably keep the two separate. This will likely be `rpy2`-type questions and the other is the AG.

coatless:
Added two publically facing sample question.

exampleCourse/questions/examplesRAddTwoNumbers/info.json
```diff
@@ -0,0 +1,7 @@
+{
+    "uuid": "69994f6e-0150-4c29-bc6f-87017719f7b1",
+    "title": "Add Two Numbers through R",
+    "topic": "R",
+    "tags": ["numeric", "balamut2", "v3", "Fa19", "rstats"],
+    "type": "v3"
+}
```
mwest1066: nit: all the JSON files are missing a newline at the end of file. It’s no big deal, but it makes it weird to cat a file or things like that. The Python files are all fine.

exampleCourse/questions/examplesRAddTwoNumbers/info.json
```diff
@@ -0,0 +1,7 @@
+{
+    "uuid": "69994f6e-0150-4c29-bc6f-87017719f7b1",
+    "title": "Add Two Numbers through R",
```
mwest1066: I think we’ve standardized on sentence case for titles.

exampleCourse/questions/examplesRGetData/server.py
```diff
@@ -0,0 +1,62 @@
+import io
+import numpy
+import random
+import rpy2.robjects as robjects
+from pandas import DataFrame
+
+def generate(data):
+    
+    # Get list of data sets in base R
+    # val = robjects.r('d = data(package = "datasets")$results[,"Item"]; d[!grepl("[[:space:]]", d)]')
```
mwest1066: Why do we have commented-out code in this file? Is it illustrating something, or is it just left over? If the latter, maybe remove it?

exampleCourse/questions/examplesRRNGSample/question.html
```diff
@@ -0,0 +1,31 @@
+<pl-question-panel>
+  <p>
+    With the seed of <b>{{params.seed}}</b>, generate a sample under replacement of size <b>{{params.sample_size}}</b> from a set of <b>{{params.set_size}}</b> integers.
```
mwest1066: I’m not familiar with the terminology “sample _under_ replacement”. I would word this:
```suggestion
    With the seed of <b>{{params.seed}}</b>, generate a sample of size <b>{{params.sample_size}}</b> by sampling with replacement from a set of <b>{{params.set_size}}</b> integers.
```

exampleCourse/courseInstances/Sp15/assessments/examplesRGallery/infoAssessment.json
```diff
@@ -0,0 +1,24 @@
...
+            "startDate": "2014-07-07T00:00:01",
+            "endDate": "2024-07-10T23:59:59"
+        }
+    ],
+    "zones": [
+        {
+            "questions": [
+                {"id": "examplesRAddTwoNumbers",   "points": 2, "maxPoints": 10},
+                {"id": "examplesRGetData",   "points": 2, "maxPoints": 10},
+                {"id": "examplesRRNGSample", "points": 2, "maxPoints": 10}
```
mwest1066: Also add these questions to `tests/testQuestions.js`?

exampleCourse/questions/examplesRGetData/server.py
```diff
@@ -0,0 +1,62 @@
...
+    # Simulate a location
+    #loc_row = random.randint(1, ds_nrow)
+    #loc_col = random.randint(1, ds_ncol)
+    
+    # Retrieve location value
+    #loc_value = robjects.r("get('%s')[%s, %s]" % (ds_name, loc_row, loc_col))
+    #data['correct_answers']['ds_loc_val'] = loc_value
+    
+    # Output Data Table
+    data['params']['display_data'] = str(robjects.r("knitr::kable(head(get('%s'), n = %s), align = rep('c', %s), format = 'html', row.names = FALSE)" %
```
mwest1066: Would the R code be easier to read with f-strings? Like this:
```suggestion
    data['params']['display_data'] = str(robjects.r(f"knitr::kable(head(get('{dn_name}'), n = {ds_nrow}), align = rep('c', {ds_ncol}), format = 'html', row.names = FALSE)"))
```

exampleCourse/questions/examplesRRNGSample/server.py
```diff
@@ -0,0 +1,50 @@
...
+  #            n_prop = n_hits / sample_size
+  #            return(c(n_hits, n_prop))
+  #        }
+  #        ''')
+  # sample_py = robjects.r("sample_r") # Retrieve function
+  # vals = sample_py(sample_size, set_size, seed, observed_val)
+  #
+  # # Unpack data
+  # n_occurrence = vals[0]
+  # n_prop = vals[1]
```
mwest1066: Commented-out code has a bad habit of bit-rotting. Is there some way we could run this code, or just delete it? We could also add another example question to demonstrate writing functions in R?

exampleCourse/questions/examplesRAddTwoNumbers/server.py
```diff
@@ -0,0 +1,26 @@
...
+    values = robjects.r(
+      '''
+       x = runif(2, 5, 10)
+       x = floor(x)
+       x
+      '''
+    )
+    # Cast to R
+    a = values[0] + 0
+    b = values[1] + 0
```
mwest1066: What’s the `+ 0` doing here? In my tests it seems unnecessary, because `values[0]` already has type float.

exampleCourse/questions/examplesRGetData/server.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # Provide data set name
+    data['params']['ds_name'] = ds_name
+
+    # Provide data set name
+    #data['params']['loc_row'] = loc_row
+    #data['params']['loc_col'] = loc_col
+    
+    # Simulate some noise
+    data['params']['bad_row_lo'] = ds_nrow + random.randint(-8, -1)
```
mwest1066: nit: write `low` instead of `lo`, to match `high` and `medium`

exampleCourse/questions/examplesRRNGSample/question.html
coatless: Done.

exampleCourse/questions/examplesRRNGSample/server.py
coatless: I re-wrote this question to use the new style. It's a bit faster than multiple calls into _R_.

exampleCourse/courseInstances/Sp15/assessments/examplesRGallery/infoAssessment.json
coatless: How about the `examplesRRNGSample` variant? I'm not sure all three are required.

exampleCourse/questions/examplesRGetData/server.py
```diff
@@ -0,0 +1,55 @@
...
+    list(params = list(display_data = df, 
+                       bad_row_low = ds_nrow + sample(-8:-1, 1),
+                       bad_row_medium = ds_nrow + sample(1:3, 1),
+                       bad_row_high = ds_nrow + sample(4:6, 1),
+                       bad_row_highest = ds_nrow + sample(6:9, 1),
+                       bad_col_low = ds_ncol + sample(-2:-1, 1), 
+                       bad_col_medium = ds_ncol + sample(1:3, 1),
+                       bad_col_high = ds_ncol + sample(4:7, 1), 
+                       bad_col_highest = ds_ncol + sample(7:10, 1)),
+         ans    = list(ds_nrow = ds_nrow, ds_ncol = ds_ncol))
```
mwest1066: The above randomization strategy means that the true answer is always the second-lowest answer out of the five options, if I'm understanding correctly.

An alternative secure version (written in Python) is:
```
# calculate the true answer somehow
true_answer = 42

# parameters for generating distractors
number_answers = 5  # 1 true answer plus 4 distractors
min_gap = 1   # minimum distance between answers
max_gap = 4   # maximum distance between answers

# generate distinct random answer offsets from random gaps between answers
# this is not essential to the provably-secure generation (e.g., could use accept/reject instead)
gaps = np.random.randint(min_gap, max_gap, number_answers)
offsets = np.cumsum(gaps)

# choose one of the answers to be the true answer
true_answer_index = np.random.randint(number_answers)

# offset of the true answer from the smallest answer
true_answer_offset = offsets[true_answer_index]

# calculate the answers by computing the offsets from the known true answer
answers = true_answer + offsets - true_answer_offset

# eliminate the true answer from the answers array to get just the distractors
distractors = np.delete(answers, true_answer_index)

# the list of distractors are now in distractors[0], distractors[1], etc.
```

I believe that the above algorithm is provably secure, in the sense that:
```
Prob(c_i = T | v_1,...,v_n) = Prob(c_i = T) = 1/n for all i = 1,...,n
```
where
```
n = number of available answers
v_1,...,v_n = numerical values of each answer
c_1,...,c_n = T/F for each answer, indicating whether it's a correct answer
```
That is, the probability that a particular answer is correct is independent of the numerical value for that answer, and it is also independent of the numerical values of the other answers.

Proving this seems like a straightforward application of Bayes, assuming an uniformative prior given by a uniform likelihood over the reals (technically, an improper prior) that satisfies translation invariance. That is, translation invariance of the prior means that
```
P(v_1,...,v_n | c_i = T) / P(v_1,...,v_n) = 1
```
and from the uniform selection of the true answer offset we have
```
P(c_i = T) = 1/n
```
so combining these with Bayes' theorem gives
```
P(c_i = T | v_1,...,v_n) = P(v_1,...,v_n | c_i = T) P(c_i = T) / P(v_1,...,v_n)
                         = (P(v_1,...,v_n | c_i = T) / P(v_1,...,v_n)) P(c_i = T)
                         = 1/n
```
which proves the result. We have a slight technical issue in that the translation invariance requires an improper prior (does not satisfy the law of total probability). This can be fixed using a uniform likelihood function `L(c_i = T, {v_1,...,v_n}) = 1` and Bayes' theorem written in likelihood form:
```
P(c_i = T | v_1,...,v_n) \propto L(c_i = T, {v_1,...,v_n}) P(c_i = T)
                         = 1/n
```
and then recognizing that this is already normalized gives `P(c_i = T | v_1,...,v_n) = 1/n`.

We can also write an equivalent algorithm for continuous-valued answers:
```
# calculate the true answer somehow
true_answer = 3.1415

# parameters for generating distractors
number_answers = 5  # 1 true answer plus 4 distractors
min_gap = 0.7   # minimum distance between answers
max_gap = 1.8   # maximum distance between answers

# generate distinct random answer offsets from random gaps between answers
# this is not essential to the provably-secure generation (e.g., could use accept/reject instead)
gaps = np.random.uniform(min_gap, max_gap, number_answers)
offsets = np.cumsum(gaps)

# choose one of the answers to be the true answer
true_answer_index = np.random.randint(number_answers)

# offset of the true answer from the smallest answer
true_answer_offset = offsets[true_answer_index]

# calculate the answers by computing the offsets from the known true answer
answers = true_answer + offsets - true_answer_offset

# eliminate the true answer from the answers array to get just the distractors
distractors = np.delete(answers, true_answer_index)

# the list of distractors are now in distractors[0], distractors[1], etc.
```

This satisfies the same provably-secure result as the discrete version above, using an identical proof.

I wonder whether we should include these algorithms in the `prairielearn.py` library so questions can easily use them? I feel like these methods must be in the literature somewhere, but I don't know where to look. If they aren't, then we should publish them 😄

exampleCourse/questions/examplesRRNGSample/server.py
```diff
@@ -0,0 +1,47 @@
+import random
+import rpy2.robjects as robjects
+
+def generate(data):
+
+  # Pick a seed
+  seed_val = random.randint(1000, 9999)
```
mwest1066: Is there any reason to not just use `data["variant_seed"]` here?

exampleCourse/courseInstances/Sp15/assessments/examplesRGallery/infoAssessment.json
coatless: Good point. I'll add them.

exampleCourse/questions/examplesRRNGSample/server.py
coatless: Lack of knowledge that the `variant_seed` existed. Though, the question is what is the range of the `variant_seed`? Documentation only suggests `Integer`.

https://prairielearn.readthedocs.io/en/latest/devElements/#element-functions

exampleCourse/questions/examplesRGetData/server.py
coatless: Surprisingly, there is little literature under ["question distractors numeric"](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C14&q=question+distractors+numeric&btnG=) and ["question distractors"](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C14&q=question+distractors&btnG=) on a precursory gaze. The majority of literature is focused on generating stems or re-using words within a given domain [c.f. "Automatic Generation of Cloze Question Distractors"](https://www.isca-speech.org/archive/L2WS_2010/papers/lw10_P2-11.pdf). I think this was largely left off of the MC formulation because the input is traditional `numeric`. 

Regarding the provably secure proof, let me think about it a bit more and get back to you.

mwest1066: Go for it!

exampleCourse/questions/examplesRRNGSample/server.py
mwest1066: This is undocumented, but it's in the range 0 to 2^{32}-1 (inclusive). That is, it's a random unsigned 32-bit integer.

coatless: Unfortunately, the range of [`Integers`](https://stat.ethz.ch/R-manual/R-patched/library/base/html/integer.html) in _R_ is restricted to `+/-2*10^9`. So, we cannot use the `data["variant_seed"]`. Instead, we have to pre-generate a seed. 

```r
set.seed(2^(32)-1)
# Warning in set.seed(2^(32) - 1) :
#   NAs introduced by coercion to integer range
# Error in set.seed(2^(32) - 1) : supplied seed is not a valid integer
```

exampleCourse/questions/examplesRGetData/server.py
coatless: One issue I think that creeps up with the distractor generation is the inability to set a lower bound for values to take on. As an example, the number of variables can now be negative.
bounds

coatless:
@mwest1066 bump.

coatless:
@mwest1066 bump.

PR #1730, opened 2019-09-28
Add CloudWatch stats for external grader lifecycles

mwest1066:


mwest1066:
Here are the new stats. This shows:

* Start from 1 grader
* Autoscale up to 3 graders, where we see them first Launching and then InService
* Make one of them unhealthy at 11:31 by killing docker on it
* Autoscaling adds a replacement grader 6 minutes later (why so long?!)
* Autoscaling removes the unhealthy grader a few minutes after that

IMG_0097

Watching the EC2 console, the reason that it took so long to replace the unhealthy grader was that EC2/ELB didn’t mark it as unhealthy until about 11:37. I verified that the `/ping` endpoint was returning `Unhealthy` immediately after 11:31, so maybe ELB isn’t pinging very often?

PR #1733, opened 2019-10-04
Add console stack trace on Node warnings

mwest1066:


trombonekenny:
Just the strack trace? Why not the whole error? i.e.

```javascript
process.on('warning', (warning) => {
  console.warn(warning.name);    // Print the warning name
  console.warn(warning.message); // Print the warning message
  console.warn(warning.stack);   // Print the stack trace
});
```

Is there something in particular you're looking for from these `warning`s?

nwalters512:
> Just the strack trace? Why not the whole error?

To this point, just logging the error as a whole should be enough:

```
> console.log(new Error())
< VM325:1 Error
      at <anonymous>:1:13
```

PR #1737, opened 2019-10-06
Fix CI to merge origin/master

mwest1066:


nwalters512:
What's the reasoning behind this change?

mwest1066:
This is attempting to replicate what Travis does (if I understand correctly). To quote:

“Rather than build the commits that have been pushed to the branch the pull request is from, we build the merge between the source branch and the upstream branch.”

(see https://docs.travis-ci.com/user/pull-requests/#how-pull-requests-are-built )

My understanding is that this attempts to test what the state of the code will be _after_ merging, rather than simply the code in the PR. This seems generally sensible to me, because the PR might be ok on it’s own, but will break is some weird way after merging due to other changes on `master`.

PR #1745, opened 2019-10-10
Fix assessments.assessment_set_id delete cascade rule

mwest1066:
The assessments.assessment_set_id foreign key constraint had ON DELETE
SET NULL, but the assessment_set_id column had a NOT NULL constraint.
This makes no sense and gave an error when we deleted courses (and hence
assessment_sets) in the tools/dump_filter.sh script.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -99,6 +99,8 @@
 
   * Fix jobsPerInstance grader statistic (Matt West).
 
+  * Fix `assessments.assessment_set_id` to cascade on deletes (Matt West).
+
 * __3.2.0__ - 2019-08-05
 
   * Add openpyxl to the centos7-python for Excel .xlsx autograding (Craig Zilles).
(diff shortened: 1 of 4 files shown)


mwest1066:
Great point! I think the key thing here is Transactional DDL: https://julien.danjou.info/why-you-should-care-that-your-sql-ddl-is-transactional/

I believe the two ALTER TABLE commands are executed within a single transaction: https://www.postgresql.org/docs/11/protocol-flow.html#PROTOCOL-FLOW-MULTI-STATEMENT

As an alternative to an explicit transaction, I changed this migration to use a single ALTER TABLE statement with the ADD and DROP both in it. According to the docs the entire ALTER TABLE will take an ACCESS EXCLUSIVE lock, which I understand to be around the ADD/DROP as a pair: https://www.postgresql.org/docs/current/sql-altertable.html

nwalters512:
Awesome, thanks for checking the docs! Let's ship this 🚀

PR #1751, opened 2019-10-12
Add support for rendering adjacency matrices as graphs

nicknytko:
PR to add support for matrix input to `pl-graphviz-render`, which has been renamed to `pl-graph` as per @mwest1066 's suggestion.  Support for questions with the old tag has been introduced to reduce breakage. This is an element that was used in CS357 as a fork of the original graphviz elem that we'd like to merge back into master.

*Example:*

Screen Shot 2019-10-12 at 3 23 07 PM

```html
<pl-matrix-latex params-name="matrix"></pl-matrix-latex>
<pl-graph params-name-matrix="matrix" params-name-labels="labels"></pl-graph>
```

```python
def generate(data):
  mat = np.random.random((3, 3))
  mat = mat / la.norm(mat, 1, axis=0)
  data['params']['labels'] = pl.to_json(['A', 'B', 'C'])
  data['params']['matrix'] = pl.to_json(mat)
```


schemas/index.js
```diff
@@ -1,7 +1,7 @@
 module.exports.infoAssessment = require('./schemas/infoAssessment.json');
 module.exports.infoCourse = require('./schemas/infoCourse.json');
 module.exports.infoCourseInstance = require('./schemas/infoCourseInstance.json');
-module.exports.infoElementCore = require('./schemas/infoElementCourse.json');
+module.exports.infoElementCore = require('./schemas/infoElementCore.json');
```
coatless: Nice catch.

question-servers/freeform.js
```diff
@@ -106,6 +106,13 @@ module.exports = {
                             // TODO remove once everyone is using the new version
                             if (elementType === 'core') {
                                 elements[elementName.replace(/-/g, '_')] = info;
+                                
+                                if ('additionalNames' in info) {
+                                    info.additionalNames.forEach(name => {
+                                        elements[name] = info;
+                                        elements[name.replace(/-/g, '_')] = info;
```
coatless: Is replacing `-` to `_` required here?

elements/pl-graph/pl-graph.py
```diff
@@ -0,0 +1,105 @@
+import prairielearn as pl
```
coatless: If possible, in the future, could you `git mv old_file new_file`? History was lost on this file :/  (Though, glancing at the other diff, I'm not sure that's a bad thing.)

doc/elements.md
```diff
@@ -1041,28 +1041,49 @@ The provided `script-name` corresponds to a file located within the director for
...
 
-## `pl-graphviz-render` element
+## `pl-graph` element
 
 Using the [viz.js](https://github.com/mdaines/viz.js/) library, create 
-Graphviz DOT visualizations .
+Graphviz DOT visualizations.
 
-#### Sample Element
+#### Sample Element 1
```
coatless: Remove the 1 and 2. Switch to  Sample Elements

Use `---` to create a horizontal rule between two sections if need be.

question-servers/freeform.js
nicknytko: Not sure, are underscores still in the process of being phased out?  I have it there just in case but I'm fine taking It out if we think it's unneeded.

mwest1066: We have completely phased out underscores, but we will support them indefinitely for backwards compatibility so we want this code.

elements/pl-graph/pl-graph.py
mwest1066: Git actually doesn’t differentiate between `git mv` and the pair of `git rm` + `git add`. It only stores file contents, and does “rename detection” for display. See https://stackoverflow.com/questions/7938582/how-does-git-detect-similar-files-for-its-rename-detection

mwest1066:
I’ll add a vote for an example question to show off how the new adjacency matrix code works.

nicknytko:
Added sample as `examplesGraph`

exampleCourse/questions/examplesGraph/info.json
```diff
@@ -0,0 +1,11 @@
+{
+    "uuid": "3c4fc412-7bf2-45e2-bae4-f8e7a5faede5",
+    "title": "Rendering Graphs using pl-graph",
```
mwest1066:
```suggestion
    "title": "Rendering graphs using pl-graph",
```

PR #1755, opened 2019-10-15
WIP: Tools to test external graded questions outside of PL

trombonekenny:
Somewhat of a proof-of-concept and somewhat of a development tool, these scripts are designed to simplify mimicking the external grader environment without needing to run the question through PrairieLearn. 

This is helpful if you want to test a submission against the question, especially while updating the grading scripts. You can run the two scripts without any interaction for a quick test, or run them with interactive shells to check the environment as the scripts are running. `question-container-debugging.sh` runs on the host and setups the grader container to use against a QID. `run-question-in-container` sets things up in `/grade` and runs the entrypoint commands.

WIP to get some documentation and ChangeLog added. Reviewers are welcome.


tools/question-container-debugging.sh
```diff
@@ -0,0 +1,83 @@
...
+
+PLMOUNT=""
+if [ -z "$PLPATH" ]; then
+    echo ""
+    echo "Environment variable PLPATH not set."
+    echo "Set this to the absolute path to your PrairieLearn directory to use it inside the container"
+    echo "Things will likely error without this."
+    echo ""
+else
+    PLMOUNT="-v $PLPATH:/PrairieLearn"
```
mwest1066: Do we want to cope with spaces in filenames? If so, this should be
```
PLMOUNT='-v "$PLPATH:/PrairieLearn"'
```
or we should use bash arrays:
```
PLMOUNT=(-v "$PLPATH:/PrairieLearn")
```

trombonekenny: It's hard to fix that case. I'm having trouble finding an easy way because if we trap the file name (so bash treats it as `'-v /home/mussulma/Prairie Learn:/PrairieLearn'` we still have to pass that into the docker command line which needs escaping. I'm prepared to say "don't have spaces in file names"?

PR #1760, opened 2019-10-20
add R package kableExtra

daviddalpiaz:
add R package kableExtra for much easier html table creation

### DIFF
--- a/environments/centos7-plbase/r-requirements.R
+++ b/environments/centos7-plbase/r-requirements.R
@@ -47,7 +47,8 @@ pkg_list = c(
   'randomForest',
   'rpart',
   'rpart.plot',
-  'rsample'
+  'rsample',
+  'kableExtra'
 )
 
 # Determine what packages are NOT installed already.


coatless:
Maybe try out the new `pl-python-variable` output? 

https://prairielearn.readthedocs.io/en/latest/elements/#pl-python-variable-element

daviddalpiaz:
This wouldn't be for outputting data stored in a data frame. My use case here would be something more along the lines of specifying the joint distribution of two discrete random variables as a table. (Where the probabilities were generated at random.) Would need the kableExtra flexibility to do things like suppressing full width and bolding the first "column" which would be the values of a random variable.

nwalters512:
Could this be implemented as a PL element with Python? It's entirely possible that this package is fully necessary but I'm trying to be more cognizant of the impacts of adding more and more packages to our base image.

daviddalpiaz:
Probably, I'm just very python deficient. And now I have a setup where I basically never touch `server.py` anymore, so I've reverted back to thinking only in R.

Basically I could pass a data frame back to Python, which can easily be captured with pandas. I just don't know how I'd do the styling I'm looking for. (In the R world, I know how to pass back a string containing the exact html that I want, by using that package.)

There might be another conversation to have here around R packages. I've added a lot to the base image that are needed to programmatically generate questions and answers, and eventually I'd like to add more. I suppose @coatless could comment on the necessity and impact of these.

Also, no rush on merging this particular package. I certainly would like it for the future, but I don't actually have an immediate need.

coatless:
@daviddalpiaz can you provide a minimal example of the output using _R_?

daviddalpiaz:
```
# load packages
library(dplyr)
library(knitr)
library(kableExtra)

# "randomly" define a joint distribution
joint_probs = round(1:9 / sum(1:9), 2)
joint_probs = sample(joint_probs)
joint_dist = matrix(joint_probs, 3, 3)
rownames(joint_dist) = c("X = 5", "X = 6", "X = 7")
colnames(joint_dist) = c("Y = 1", "Y = 2", "Y = 3")

# matrix view in R
joint_dist

# styled and output as HTML table
joint_dist %>% 
  kable() %>% 
  kable_styling("striped", full_width = FALSE) %>% 
  column_spec(column = 1, bold = TRUE, background = "white", border_right = TRUE)
```

PR #1819, opened 2019-10-28
Add libraries for CS199 Data Science w/ Python external grader

nicknytko:
Adding:

- `scikit-image`
- `scikit-learn`
- `requests`
- `folium`
- `colormath`
- `nltk`
- `beautifulsoup4`
- `bokeh`

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -48,6 +48,8 @@
   * Add support for rendering graphs using adjacency matrices in `pl-graph` (Nicolas Nytko).
 
   * Add `gvsu.edu` as an institution option (Matt West).
+  
+  * Add `scikit-learn`, `scikit-image`, and other data science libraries to external grader (Nicolas Nytko).
 
   * Change v3 questions to disable autocomplete on the question form (Nathan Walters).
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -49,7 +49,7 @@
 
   * Add `gvsu.edu` as an institution option (Matt West).
   
-  * Add `scikit-learn`, `scikit-image`, and other data science libraries to external grader (Nicolas Nytko).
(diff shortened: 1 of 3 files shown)


graders/python/Dockerfile
```diff
@@ -10,6 +10,6 @@ RUN yum -y update \
     && yum install -y python36u python36u-devel python36u-pip \
     && yum install -y graphviz graphviz-devel \
     && ln -s /usr/bin/python3.6 /usr/bin/python3 \
-    && python3 -m pip install --no-cache-dir -r /requirements.txt
+     ; python3 -m pip install --no-cache-dir -r /requirements.txt
```
mwest1066: What's the reason for this change?

ChangeLog.md
```diff
@@ -48,6 +48,8 @@
   * Add support for rendering graphs using adjacency matrices in `pl-graph` (Nicolas Nytko).
 
   * Add `gvsu.edu` as an institution option (Matt West).
+  
+  * Add `scikit-learn`, `scikit-image`, and other data science libraries to external grader (Nicolas Nytko).
```
mwest1066:
```suggestion
  * Add `scikit-learn`, `scikit-image`, and other data science libraries to external Python grader (Nicolas Nytko).
```

graders/python/Dockerfile
nicknytko: When I was building the docker image locally on my machine it was error-ing out because python3 already existed, so I made it fail silently if it can't create the symlink

mwest1066: Good point -- I think the right fix here is to apply the same `python3` update as @trombonekenny did for `plbase`: https://github.com/PrairieLearn/PrairieLearn/pull/1719

PR #1855, opened 2019-11-08
Update grader-r image with pltest dependency

coatless:


graders/r/Dockerfile
```diff
@@ -1,17 +1,19 @@
...
 
-# Add a script to install any additional R packages
+# Add a script to install any additional R packages from CRAN or GitHub
 COPY r-requirements.R /
 
 # Modify the image to meet with PrairieLearn specifications
-RUN useradd ag `# Enable autograde user for PrairieLearn` \                        
-    && su root -c "Rscript /r-requirements.R" `# Install the additional packages`
+# Enable autograde user for PrairieLearn and create home directory
+RUN adduser --disabled-password --gecos "" ag \
```
coatless: The alternative here would be to use:

```bash
useradd -m ag
```

I switched it over to `adduser` as I think this is the preferred approach.

graders/r/r-requirements.R
```diff
@@ -16,13 +16,16 @@ if(Ncpus > 4) {
...
 to_install_pkgs = pkg_list[!(pkg_list %in% installed.packages()[,"Package"])]
 
 # Install the missing packages
 if(length(to_install_pkgs)) {
-    install.packages(to_install_pkgs)
+    install.packages(to_install_pkgs, quiet = TRUE, verbose = FALSE)
 }
+
+# Install pl testing framework package from GitHub
+remotes::install_github('illinois-r/pltest')
```
coatless: This allows us to easily update the underlying autograding code as needed.

Previously, each course had a copy of the code under `pltest.R` within `serverFilesCourse`.

PR #1867, opened 2019-11-18
Demo related to using survey questions within PrairieLearn

### LINKED ISSUE #1291: Request for "do not grade this element" feature
For a couple research studies, I'd like to have a consent form where people can choose between "I agree to participate" and "I do not agree to participate" and then also have a text box to enter their name.
Of course, it would probably be better if the element was more flexible and allowed for more or fewer inputs.

Here's an example of what I'd like to be able to have, except that it would not be graded and that the multiple-choice would display both "true" answers.
<pl-multiple-choice answers-name="consent" weight="0" fixed-order="true" number-answers="2">
 <pl-answer correct="true">I agree to participate in this study</pl-answer>
 <pl-answer correct="true">I do not agree to participate in this study</pl-answer>
</pl-multiple-choice>

<pl-string-input answers_name="name" allow-blank="true" weight = "1"></pl-string-input>

coatless:
Sample approach for collecting student feedback created to resolve #1291 by @glherman.

I added this because of the discussion on Slack.

### Question text

Question text

### Correct Answer and Submission Panel

Correct Answer & Submission Panel


trombonekenny:
For a v3 question, you shouldn't need an `answer.html` right?

coatless:
@trombonekenny great comment. I've fixed that by moving the text into a `<pl-answer-panel>` element. I've also added new screenshots in the main post.

coatless:
@mwest1066 bump.

eliotwrobson (approved):
LGTM!

coatless:
It's.. It's... Alive!

jonatanschroeder:
About using inputs without elements, would these be included in downloads? I presume that for surveys this would be necessary.

exampleCourse/questions/demo/surveyQuestion/info.json
```diff
@@ -0,0 +1,8 @@
+{
+    "uuid": "f532aa25-511e-42a9-be4c-0daec9ef45e1",
+    "title": "Demo: Obtaining feedback from students",
+    "topic": "Demo",
+    "tags": ["Fa18", "v3", "glherman", "balamut2"],
+    "type": "v3",
```
jonatanschroeder, line 6:
```suggestion
    "type": "v3",
    "singleVariant": true,
```

exampleCourse/questions/demo/surveyQuestion/question.html
```diff
@@ -0,0 +1,47 @@
+<pl-question-panel>
+
+    <p>Please answer the following demographic survey questions. </p>
+    <p>As this is a survey, there are no correct or incorrect responses.</p>
+
+    <hr>
+
+    <h6>Race/Ethnicity: Check all that apply</h6>
+
+    <p><input type="checkbox" name="race" value="white" id="1" /> <label for="1">White or European American</label></p>
```
jonatanschroeder, line 10: HTML ids cannot start with a digit. Note that you can use label around the input, in that case an id is not necessary:

```suggestion
    <p><label><input type="checkbox" name="race" value="white" /> White or European American</label></p>
```

jonatanschroeder (commented):
> About using inputs without elements, would these be included in downloads? I presume that for surveys this would be necessary.

It seems the submitted answers are populated, so this should be fine. That said, the fact that an element is not being used here means the look-and-feel of the elements (e.g., spacing, style, etc.) does not match the rest of PL. I'm not saying an element must be used, but this aspect is something to consider for an official template.

eliotwrobson:
> That said, the fact that an element is not being used here means the look-and-feel of the elements (e.g., spacing, style, etc.) does not match the rest of PL. I'm not saying an element must be used, but this aspect is something to consider for an official template.

This is a good point, and I thought about it too while putting this together. As mentioned in the original issue, I don't think that any of the current elements are really appropriate for this type of survey, and if people are using their own inputs to do such surveys, we should match that style in the example. As long as the example is high quality, then I think using inputs is fine (people's custom elements may not conform to PL visual standards anyway). @coatless any input?

exampleCourse/questions/demo/surveyQuestion/question.html
eliotwrobson: This totally slipped my mind, great catch! Updated here https://github.com/PrairieLearn/PrairieLearn/pull/1867/commits/9d370d64f0be83c6055411ef822b2db5a6903043

exampleCourse/questions/demo/surveyQuestion/question.html
```diff
@@ -0,0 +1,47 @@
...
+            Native</label></p>
+    <p>Other (please specify): <input type="text" name="otherrace" /></p>
+
+    <hr>
+
+    <h6>Gender: Select one</h6>
+
+    <p><label><input type="radio" name="gender" value="male" />Male</label></p>
+    <p><label><input type="radio" name="gender" value="female" />Female</label></p>
+    <p><label><input type="radio" name="gender" value="nonbin" />Non-Binary</label></p>
```
jonatanschroeder, line 27: May add a "prefer not to answer"?

exampleCourse/questions/demo/surveyQuestion/question.html
```diff
@@ -0,0 +1,47 @@
...
+    <p><label><input type="checkbox" name="race" value="white" />White or European American</label></p>
+    <p><label><input type="checkbox" name="race" value="black" />Black or African American</label></p>
+    <p><label><input type="checkbox" name="race" value="hispanic" />Hispanic, Latina/o, or
+            Spanish</label></p>
+    <p><label><input type="checkbox" name="race" value="asian" />Asian American</label></p>
+    <p><label><input type="checkbox" name="race" value="pacific" />Native Hawaiian or Other Pacific
+            Islander</label></p>
+    <p><label><input type="checkbox" name="race" value="native" />Native American or Alaskan
+            Native</label></p>
+    <p>Other (please specify): <input type="text" name="otherrace" /></p>
```
jonatanschroeder, line 19:
```suggestion
    <p><label>Other (please specify): <input type="text" name="otherrace" /></label></p>
```

exampleCourse/questions/demo/surveyQuestion/question.html
eliotwrobson: Done here https://github.com/PrairieLearn/PrairieLearn/pull/1867/commits/f41c4e6e0e49da34753af17132c3fb0484b9df2c

coatless:
@eliotwrobson & @jonatanschroeder I can agree about the look and feel aspect. However, I think a second part of the survey inclusion was showcasing that PL still functioned as a data collection mechanism outside of using the curated `pl` elements.

jonatanschroeder:
> I think a second part of the survey inclusion was showcasing that PL still functioned as a data collection mechanism outside of using the curated `pl` elements.

It may be a good idea to add that as a note to the template question.

exampleCourse/questions/demo/surveyQuestion/info.json
nwalters512, line 1: Nit: there's no need for `Question` to appear in the question name. It's in the `questions` directory; it's unambiguously a question.

exampleCourse/questions/demo/surveyQuestion/question.html
```diff
@@ -0,0 +1,48 @@
+<pl-question-panel>
+
+        <p>Please answer the following demographic survey questions. </p>
```
nwalters512, line 3: Nit: we don't use 8 spaces of indentation elsewhere. Let's try to remain consistent.

exampleCourse/questions/demo/surveyQuestion/question.html
```diff
@@ -0,0 +1,48 @@
...
+        <p>As this is a survey, there are no correct or incorrect responses.</p>
+
+        <hr>
+
+        <h6>Race/Ethnicity: Check all that apply</h6>
+
+        <p><label><input type="checkbox" name="race" value="white" /> White or European American</label></p>
```
nwalters512, line 10: Nit: Bootstrap's default bottom margin on `<p>` makes things really spaced out:

Screenshot 2023-05-30 at 11 30 31

Let's just use divs instead?

nwalters512 (commented):
Looks good overall! A few nits that we can discuss before merging.

exampleCourse/questions/demo/surveyQuestion/question.html
```diff
@@ -0,0 +1,48 @@
+<pl-question-panel>
+
+        <p>Please answer the following demographic survey questions. </p>
+        <p>As this is a survey, there are no correct or incorrect responses.</p>
+
+        <hr>
+
+        <h6>Race/Ethnicity: Check all that apply</h6>
```
nwalters512, line 8: Nit: this is an inappropriate usage of `<h6>`, as it doesn't follow the proper hierarchy (that is, there's no `<h5>`/`<h4>`/etc. above it in the DOM).

eliotwrobson: Should this just be in a `<p>` tag then?

exampleCourse/questions/demo/surveyQuestion/question.html
eliotwrobson: Fixed here https://github.com/PrairieLearn/PrairieLearn/pull/1867/commits/1fe5f5fe4ef56d022cb77a461a1d73e7ab80596e

exampleCourse/questions/demo/surveyQuestion/question.html
eliotwrobson: Fixed here https://github.com/PrairieLearn/PrairieLearn/pull/1867/commits/1fe5f5fe4ef56d022cb77a461a1d73e7ab80596e

exampleCourse/questions/demo/surveyQuestion/question.html
eliotwrobson: Done here https://github.com/PrairieLearn/PrairieLearn/pull/1867/commits/1fe5f5fe4ef56d022cb77a461a1d73e7ab80596e

exampleCourse/questions/demo/surveyQuestion/info.json
eliotwrobson: Done here https://github.com/PrairieLearn/PrairieLearn/pull/1867/commits/59dda490b54e3d64d368f83cf7528c3ac511fea0

nwalters512 (approved):

PR #1868, opened 2019-11-18
Add external grader result delay logging

trombonekenny:
If an external grading result process takes more than 10 seconds to execute, log the details.

I'm not sure if `setTimeout()` is the best way to do this, and I couldn't figure out how to easily make it log every 10 seconds (resetting the Timeout, it only works once) but I think this will work as a rough picture to see maybe where we're hanging?


lib/externalGraderResults.js
```diff
@@ -101,6 +101,11 @@ function loadQueueUrl(sqs, callback) {
     }
 }
 
+function logAtTimeout(jobId, logBundle) {
+    logger.info('externalGradingResults timeout logs for ' + jobId);
+    logger.info(logBundle);
```
mwest1066: Logger statements using winston should always have a string as the first argument. I suggest upgrading this to an error and replacing the two lines above with:
```
logger.error('externalGradingResults timeout', {jobId, logBundle});
```

lib/externalGraderResults.js
```diff
@@ -112,18 +117,25 @@ function processMessage(data, callback) {
         return callback(error.makeWithData('Message does not contain a valid grading job id.', data));
     }
 
+    var logBundle = [];
+    var loggingTimeout = setTimeout(logAtTimeout, 10000, jobId, logBundle);
+
+    logBundle.push({ timestamp: Date.now(), msg: `externalGraderResults: ${jobId} processMessage ${data}`});
```
mwest1066: I suggest just logging data as actual objects, rather than as strings. Also no need to re-log `jobId`.
```suggestion
    logBundle.push({ timestamp: Date.now(), msg: 'processMessage', data});
```

lib/externalGraderResults.js
```diff
@@ -112,18 +117,25 @@ function processMessage(data, callback) {
...
     logger.verbose('Processing external grading job result message', {grading_job_id: jobId, ...data});
     if (data.event === 'job_received') {
+        logBundle.push({ timestamp: Date.now(), msg: `${jobId} processMessage, job_received`});
```
mwest1066:
```suggestion
        logBundle.push({ timestamp: Date.now(), msg: 'job_received'});
```

lib/externalGraderResults.js
```diff
@@ -112,18 +117,25 @@ function processMessage(data, callback) {
...
             received_time: data.data.receivedTime,
         };
         sqldb.queryOneRow(sql.update_grading_received_time, params, (err, _result) => {
             if (ERR(err, callback)) return;
             externalGradingSocket.gradingJobStatusUpdated(jobId);
+            clearTimeout(loggingTimeout);
             callback(null);
         });
     } else if (data.event === 'grading_result') {
+        logBundle.push({ timestamp: Date.now(), msg: `${jobId} processMessage, grading_result`});
```
mwest1066:
```suggestion
        logBundle.push({ timestamp: Date.now(), msg: 'grading_result'});
```

lib/externalGraderResults.js
```diff
@@ -150,7 +165,10 @@ function processMessage(data, callback) {
                 // have to download it from S3 first.
                 if (data.data) {
                     // We have the data!
+                    logBundle.push({ timestamp: Date.now(), msg: `${jobId} processMessage, data in data`});
```
mwest1066:
```suggestion
                    logBundle.push({ timestamp: Date.now(), msg: 'data in data'});
```

lib/externalGraderResults.js
```diff
@@ -150,7 +165,10 @@ function processMessage(data, callback) {
...
                     processResults(jobId, data.data);
+                    logBundle.push({ timestamp: Date.now(), msg: `${jobId} processMessage, callback`});
```
mwest1066:
```suggestion
                    logBundle.push({ timestamp: Date.now(), msg: 'finished with data included'});
```

lib/externalGraderResults.js
```diff
@@ -160,9 +178,12 @@ function processMessage(data, callback) {
                         Key: `${s3RootKey}/results.json`,
                         ResponseContentType: 'application/json',
                     };
+                    logBundle.push({ timestamp: Date.now(), msg: `processMessage, fetching from s3`});
```
mwest1066:
```suggestion
                    logBundle.push({ timestamp: Date.now(), msg: 'fetching from s3'});
```

lib/externalGraderResults.js
```diff
@@ -160,9 +178,12 @@ function processMessage(data, callback) {
...
                     new AWS.S3().getObject(params, (err, s3Data) => {
                         if (ERR(err, callback)) return;
                         processResults(jobId, s3Data.Body);
+                        logBundle.push({ timestamp: Date.now(), msg: `${jobId} processMessage, callback`});
```
mwest1066:
```suggestion
                        logBundle.push({ timestamp: Date.now(), msg: 'finished via S3'});
```

lib/externalGraderResults.js
```diff
@@ -160,9 +178,12 @@ function processMessage(data, callback) {
                         Key: `${s3RootKey}/results.json`,
                         ResponseContentType: 'application/json',
                     };
+                    logBundle.push({ timestamp: Date.now(), msg: `processMessage, fetching from s3`});
                     new AWS.S3().getObject(params, (err, s3Data) => {
                         if (ERR(err, callback)) return;
```
mwest1066: logBundle.push({ timestamp: Date.now(), msg: 'completed S3 fetch'});

mwest1066:
The `setTimeout()` approach seems great to me. It should definitely show up the problem.

PR #1913, opened 2019-12-05
Change to express 'trust proxy' for client IP parsing

### LINKED ISSUE #1911: Users behind web proxies cannot login
This results in an error message like:
```
invalid input syntax for type inet: "70.95.162.46, 104.129.198.95"
```

The reason is that a web proxy sets `X-Forwarded-For`, and the PL reverse proxy adds to the list:
```
X-Forwarded-For: client, proxy1, proxy2
```
(see https://en.wikipedia.org/wiki/X-Forwarded-For)

But PL is treating `X-Forwarded-For` as a single IP address: https://github.com/PrairieLearn/PrairieLearn/blob/master/middlewares/authzCourseInstance.js#L18

This results in passing a list of IPs to Postgres, which throws the `invalid input syntax for type inet` error.

To fix this, we should pull off just the correct (last-hop) IP from `X-Forwarded-For`.

trombonekenny:
Changes to use the built in Express logic to get the client IP `req.ip` when connecting through a proxy. Fixes #1911 where we weren't processing the raw headers properly.

### DIFF
--- a/server.js
+++ b/server.js
@@ -91,6 +91,7 @@ if (config.blockedAtWarnEnable) {
 const app = express();
 app.set('views', path.join(__dirname, 'pages'));
 app.set('view engine', 'ejs');
+app.set('trust proxy', 'loopback');
 
 config.devMode = (app.get('env') == 'development');
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -155,6 +155,8 @@
   
   * Fix use of `data["correct_answers"]` in documentation (James Balamuta, h/t Eric Huber).
 
+  * Fix authorization for users behind web proxies (Dave Mussulman).
+
 * __3.2.0__ - 2019-08-05
 
   * Add openpyxl to the centos7-python for Excel .xlsx autograding (Craig Zilles).
--- a/middlewares/authzCourseInstance.js
+++ b/middlewares/authzCourseInstance.js
@@ -15,7 +15,7 @@ module.exports = function(req, res, next) {
         course_instance_id: req.params.course_instance_id,
         is_administrator: res.locals.is_administrator,
         req_date: res.locals.req_date,
-        ip: req.headers['x-forwarded-for'] || req.ip,
+        ip: req.ip,
         force_mode: (config.authType == 'none' && req.cookies.pl_requested_mode) ? req.cookies.pl_requested_mode : null,
     };
     sqldb.queryZeroOrOneRow(sql.select_authz_data, params, function(err, result) {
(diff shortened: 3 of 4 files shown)


trombonekenny:
It's live right now on pl-dev if you want to poke at it.

server.js
```diff
@@ -91,6 +91,7 @@ if (config.blockedAtWarnEnable) {
 const app = express();
 app.set('views', path.join(__dirname, 'pages'));
 app.set('view engine', 'ejs');
+app.set('trust proxy', 'loopback');
```
mwest1066: Doesn't this have the same issue as the same proposed change to #1697? See https://github.com/PrairieLearn/PrairieLearn/pull/1697#discussion_r323416628

trombonekenny: My preference was to piggy back on existing tools that do this kind of manipulation rather than doing it ourselves, that's why I brought it back. I took a few minutes today to dig under the layers and see what it's doing.

`express` calls `proxy-addr` to do the parsing, which calls `forwarded` (these are both part of https://jshttp.github.io/) Digging back through the code gave me the ability to setup a quick test to see what happens if we run the `trust proxy` line above with a spoofed x-forwarded-for header, both in direct connection mode (like with docker) and when running behind the apache proxy.

```
Direct with spoofed x-forwarded-for
req { headers: { 'x-forwarded-for': '9.9.9.9' },
  connection: { remoteAddress: '128.174.55.55' } }
forwarded(req) [ '128.174.55.55', '9.9.9.9' ]
proxyaddr(req, 'loopback') 128.174.55.55


Proxy with spoofed x-forwarded-for trusting loopback
req { headers: { 'x-forwarded-for': '9.9.9.9, 128.174.55.55' },
  connection: { remoteAddress: '::1' } }
forwarded(req) [ '::1', '128.174.55.55', '9.9.9.9' ]
proxyaddr(req, 'loopback') 128.174.55.55
```

The proxy case mimics the headers received by PL under `pl-dev`. In both cases, it came out with the correct (non-spoofed) proxyaddr address. Showing the forwarded step probably isn't necessary, but it shows how the order of "forwarded" IPs is flipped between the HTTP header (appends to the end of the header string when new forwards are added) vs the JS array (which puts the most-recently seen IP at index 0).

Because apache interjects the connected IP before anything else it received in the HTTP header, it's "closest" to localhost. The 'loopback' argument tells proxyaddr to shift from the front of the array the first entry that's not one of the localhost addresses.

So I think we're good?

mwest1066: Cool, thanks for digging into the details here!

I agree that this is safe, given that we are only trusting `loopback` as a proxy.

PR #1917, opened 2019-12-11
Fix institution error in administrator overview page

mwest1066:


mwest1066:
Yep, I've got a PR in dev that makes `institution` required. I'll try and push it later today.

PR #1918, opened 2019-12-12
Fix buttons after bootstrap upgrade

tbretl:
After bootstrap upgrade from v4.3 to v4.4, buttons of the form
```
<a role="button" ... > ... </a>
```
render incorrectly.

This PR changes all of these buttons to
```
<button ... > ... </button>
```
so they render correctly.

In cases where a button is inside a form but is not intended to submit the form, `type="button"` is added to the tag.

I assume the reason we were ever using `<a>` tags and not `<button>` tags was due to some old feature of bootstrap.


nwalters512:
We should also be able to remove the `tabindex="0"` since buttons can receive keyboard focus without any additional configuration.

tbretl:
@nwalters512 What about `tabindex="-1"`? (In `pl-file-editor` if I recall.)

nwalters512:
I likely wrote that element before learning more about web accessibility 🙂 That attribute prevents elements from being focused with the keyboard, which hurts accessibility. We should remove that and make sure things still work.

Issue #1373, opened 2018-11-26
infoCourseInstance.json: add ps_linked


trombonekenny:
`course_instances.ps_linked` boolean exists in the database and is used for exam mode checking against PrairieSchedule. Let's make that configurable via the infoCourseInstance.json file.

trombonekenny:
Bump again for SP19.

mwest1066:
@trombonekenny is this still needed with the new UUID linking?

mwest1066:
bumping for @trombonekenny — I think we don’t need this anymore with UUID linking

trombonekenny:
I use this flag with the ZJUI courses because they use exam mode but not PS checking. I guess it depends how we deprecate the PS linking which would determine if we need this db column anymore.

The issue was to make that field configurable in the JSON files so that we can easier document to the ZJUI courses what they need to set. Better than the current "setup your courseInstance, sync, and then tell me so I can fiddle something behind the scenes to make the exams work"

trombonekenny:
Another bump for Fall 2020. We should figure out how to not have to toggle `ps_linked` for the non-CBTF mode:Exam exams.

mwest1066:
I think we can just delete this entire ELSE block: https://github.com/PrairieLearn/PrairieLearn/blob/180124d964913d64bd9c6c3df6d2b9a00c8b2f30/sprocs/check_assessment_access_rule.sql#L91

And then also delete the `ps_linked` column.

All the Illinois exams are using `exam_uuid` so they don't use the ELSE block above anyway. And the ZJUI courses don't have `ps_linked` set, so we should be fine with the above deletions.

Alternatively, we could leave the above code in place and just set `ps_linked = NULL` and `pl_course_id = NULL` everywhere, which should also just work?

trombonekenny:
I agree it's probably much easier to rip things out than understand all of the specific edge cases that grew in over the years. CBTF has switched entirely to examUuid linking so anything else it used can go away. I always need to go back to the unit tests when it comes to this sproc to make sure the default cases work as we expect.

mwest1066:
I think on the PL side it's fine to just rip this all out. I grepped in PS and there are a number of hits for `pl_course_id`. It wasn't immediately clear to me whether things would continue to work if we just set `pl_course_id = NULL` everywhere (or delete this column)?

trombonekenny:
Bump for Spring 2022

trombonekenny:
Bump for Fall 2022

[...]

PR #1974, opened 2020-01-13
Improve check_assessment_access tests

trombonekenny:
This PR started as the following:

> This PR changes the way the CBTF scheduler enforces reservation-to-PL-assessment matching. It removes the date range matching (and fuzzy PS to PL matching), and exclusively uses the `exam_uuid` matching that's been in place for a few semesters now.
>
> Courses that do not use the scheduler for their "mode": "Exam" exams can disable the reservation checking by setting course_instance.ps_linked to false. (ZJUI)

However, the removal of PrairieSchedule code was done in other PRs, so this PR now just adds some additional tests for the `check_assessment_access` sproc.


nwalters512:
We've since moved away from using PrairieSchedule in favor of PrairieTest; closing this.

trombonekenny:
Actually this one is still relevant. We need to remove the access rules that expect PS linking. This comes up with the ZJUI mode=Exam situations.

trombonekenny:
But it's old code that probably needs review/refactoring.

nwalters512:
@trombonekenny gotcha, sorry for closing prematurely. Can you resolve the conflicts and get this ready for review?

trombonekenny:
Closes #1373 (not sure why that wasn't linking here, sorry the names are a bit out of whack)

trombonekenny:
@mwest1066 The re-written sproc tests are passing, but `access.test.js` where it did things with PrairieSchedule integration. Should we remove that and rely only on the sproc testing? Or update access.test.js to do the PT test and logic?

trombonekenny:
I went ahead and removed the PS tests. I'll leave it to @mwest1066 and his review (poke) to decide if he'd like to add PT tests here. Those could also be added under a different PR.

trombonekenny:
@mwest1066 *poke* I updated this to work with the new PL repo structure but it still needs review.

apps/prairielearn/src/sprocs/check_assessment_access_rule.sql
```diff
@@ -45,103 +43,48 @@ BEGIN
...
+            AND e.user_id = check_assessment_access_rule.user_id;
+
+        IF FOUND THEN
+            IF assessment_access_rule.exam_uuid = exam_uuid THEN
+                -- exam_uuid matches, so don't keep going to authorized := FALSE
                 EXIT schedule_access;
+            ELSE
+                -- checked-in so deny any exams that is not linked
+                authorized := FALSE;
             END IF;
```
jonatanschroeder, line 98: I just tested, and it is possible for a student to have more than one active reservation at a time, so I assume the exam_uuid will need to go to the query instead of to a separate test.

trombonekenny:
Notes from a dev meeting today:

- This PR refactors some things that are PrairieTest related. Avoid touching that.
- Move ahead with stripping out PrairieScheduler code and tests.
- Get tests to pass against master

Take out things in one PR to the level that it passes.
Make another PR for any refactoring, if necessary.

apps/prairielearn/src/tests/sprocs/check_assessment_access.test.ts
```diff
@@ -1,59 +1,225 @@
...
+      //     authz_mode: 'Exam',
+      //     course_role: 'None',
+      //     course_instance_role: 'None',
+      //     user_id: '1000',
+      //     uid: 'valid@school.edu',
+      //     date: '2010-07-07 06:06:06-00',
+      //     display_timezone: 'US/Central',
+      //   });
+      //   assert.isFalse(authorized);
+      // });
```
nwalters512, line 195: @trombonekenny from what I can tell, this test tries to assert behavior that doesn't currently exist. Let's discuss. I don't think we necessarily want the behavior that this test expects, as for better or for worse we still support legacy `exam_mode_networks` independently of PT and `exam_uuid`.

trombonekenny: Is there an edge case where someone is in mode: Exam and there's an exam with mode: Exam protections but no exam_uuid?

nwalters512: I wouldn't call it an edge case, but yes, that's what would happen if someone is using `exam_mode_networks`, which some self-hosters are.

nwalters512: Or are you thinking about the case where a student is in a PT center and they end up with access to one exam with an `examUuid` and one without?

trombonekenny: Yeah, I was looking for tests for all the possible permutations of access like mode: Exam, checked in reservation, exam protected by uuid (and their not counterparts.)

nwalters512: I agree that it's useful to have assertions in place; the question is what behavior we actually want to assert. Exam-mode assessments without UUIDs is indeed a thing we support, so I think we should assert that access is allowed in that case.

mwest1066: At the dev meeting today we concluded that we do care about the case when we are in exam mode but don't have an examUuid. However, to handle this case correctly we'll need to remember whether we are in exam mode because of PT or because of exam mode networks. Passing this information down to the assessment access checks will require passing down an extra variable.

mwest1066 (approved):

PR #1975, opened 2020-01-14
WIP: code coverage ignore for ERR()

trombonekenny:
Checking to see if we can ignore code coverage for the async-stacktrace functions anytime they're called.

Not ready to merge, but PR will call the coveralls bit.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -183,6 +183,8 @@
 
   * Fix redirects to question preview page by maintaining query parameters (Nathan Walters).
 
+  * Fix code coverage tool to ignore async-stacktrace error handler (Dave Mussulman).
+
 * __3.2.0__ - 2019-08-05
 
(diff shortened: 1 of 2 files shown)


mwest1066:
I guess that maybe its because `ERR()` isn't a class method? I guess we could put `/* instanbul ignore next */` in front of all `ERR()` calls. That's amazingly horrible.

trombonekenny:
I couldn't find (good) documentation on the `ignore-class-method` function. Hopefully it's something in the syntax that could make it work better? Otherwise, yeah, we're unlikely to go back and tag each logic branch.

Unless we did for a file or two (with lots of ERR() calls) that showed an appreciable increase in coverage by changing them. It would be nice to know if those calls make a difference or are just a red herring.

nwalters512:
We're moving away from using callbacks and `ERR`, and we also no longer report test coverage to PRs. I'm going to close this.

PR #1979, opened 2020-01-15
Add "News" page with announcements of new PL features

mwest1066:
News items are stored in the PL code repo in the `news_items` directory. When new items are added to this, users get a badge in the menubar that leads them to the unread items. Clicking on a news item clears the "unread" status. I put two initial news items in the list:

1. New page layout

2. Full in-browser editing

My intention is to merge this before deploying current master, which includes these two features.

Note that in local development mode you never get the "new unread items" badge, because we don't have any stable storage to remember when a user has viewed them. You can still view the news items in dev mode by clicking on your username (Dev) in the top navbar and selecting "News" from the dropdown.


database/tables/news_items.pg
```diff
@@ -0,0 +1,15 @@
+columns
+    date: timestamp with time zone not null default now()
+    directory: text not null
+    for_students: boolean not null default false
+    id: bigint not null default nextval('news_items_id_seq'::regclass)
+    order_by: integer not null default 0
```
tbretl: Is it necessary to have an `order_by` when it seems natural to order by `date`?

news_items/001_new_page_layout/index.html
```diff
@@ -0,0 +1,29 @@
...
+  The layout of instructor pages and navigation bars has been
+  massively improved by <a href="https://github.com/tbretl">Tim
+    Bretl</a>.
+</p>
+
+<p>
+  <b>Easy switching.</b> You can now quickly switch between different
+  courses and different instances (semesters/terms) of a course. To do
+  this, click on the downwards-pointing triangle to the right of the
+  course name or instance name.
```
tbretl: Maybe add "in the navigation bar".

news_items/001_new_page_layout/index.html
```diff
@@ -0,0 +1,29 @@
...
+</p>
+
+<p>
+  <b>Shortcut tabs in the top menu bar.</b> All pages can be accessed
+  from the second-level menu bar, but for commonly used pages like
+  "Questions" and "Assessments" there are also shortcut tabs in the
+  top menu bar.
+</p>
+
```
tbretl: The language here is a little confusing, but perhaps only because I'm used to what you call things in html/bootstrap. 😄 

I might try to stay consistent about...
* Calling what is at the top of the page the "navigation bar" (not the "top menu bar" for example).
* Calling items in the navigation bar "buttons" or "items" or something that is not "tabs".
* Calling what is in the page the "tabbed menu bar" or something, which has "tabs".

news_items/001_new_page_layout/index.html
```diff
@@ -0,0 +1,29 @@
...
+<p>
+  <b>Wide view on instructor pages.</b> All instructor pages now use
+  the full width of the web browser window. This is particularly
+  helpful on pages like the Questions list, where there are many
+  columns in a table. Student pages have not changed width and there
+  is no plan to ever change this. When viewing a question preview in
```
tbretl: Maybe just "When previewing a question..."

news_items/002_in_browser_editing/index.html
```diff
@@ -0,0 +1,39 @@
+
+<p>
+  Thanks to a huge amount of work
+  by <a href="https://github.com/tbretl">Tim Bretl</a>, PrairieLearn
+  now allows you to directly edit all course content directly within
```
tbretl: Maybe change "to directly edit" -> "to edit" (too many directly's)

pages/news_item/news_item.ejs
```diff
@@ -0,0 +1,29 @@
...
+    <link href="/stylesheets/issues.css" rel="stylesheet" />
+  </head>
+  <body>
+    <%- include('../partials/navbar', {navPage: 'news_item'}); %>
+    <div id="content" class="container">
+      <%- include('../partials/navtabs'); %>
+
+      <a class="btn btn-primary mb-4" href="<%= urlPrefix %>/news_items" role="button">
+        <i class="fa fa-arrow-left" aria-hidden="true"></i>
+        Back to news_items
```
tbretl: Maybe something human-readable like `Back to list of news items`, `Back to news items`, or whatever - anything other than `news_items`.

mwest1066:
> > You've chosen to define each news item as a file in the main repo. Maybe you could say a little bit about why you've made that choice?

My thinking is that this "News" functionality is playing the same role as a ChangeLog or Release Notes, but only for major features and with more explanation. This is not intended to be a general notification system, although it would be fine to add that at some point as well.

The reason that I went with news items in the repo is that they are intended to be like ChangeLog entries, so they should be written as part of features PRs and the text can be commented on using the normal PR comment methods.

An alternative would have been to make this a feature where Admins can post messages to other users, or something like that. That's also a completely viable thing to have, but I think it serves a different purpose than this "Feature Release Notes" functionality.

> It means that all news will be public (since the repo is public). I assume this is why you didn't bother to prevent students from viewing news items that aren't marked `for_students` - they just aren't notified. (You may want to make this clear in docs or in the JSON schema - or call it `notify_students` instead of `for_students` or something.)

Yep, the intention is definitely to be public. The `for_students` restriction is simply to avoid spamming students with things that won't make any sense because they've never seen the instructor interface. If we make significant student-visible changes then I'd like to make a news item that students can see. An example of this in the past would have been the "attach files" feature where students can store files during exams for use in review sessions.

> It also means that, in practice, news will be infrequent and will be associated primarily with other PRs (other code changes).

Yep, this is precisely the type of news I was intending. Basically "feature announcements". Of course it's always possible to have a PR that only adds a news item, but I'm thinking this will be rare.

> I think this is all fine and makes sense, but as long as you are creating a notification system... isn't there often a need for more frequent news/notifications that are more targeted (e.g., specific to role or to course instance) and easier to post (don't require merge/deploy to master)? Would it be easy to allow news/notifications to come from other sources as well?

All totally reasonable ideas and I'd be supportive of extending in these directions in the future. The current implementation is just an MVP focused on the "feature release notes" use case.

mwest1066:
I'm inclined to not change `for_students` to `notify_students` because this flag controls both notifications and also visibility. I'd be fine with something like `visible_to_students` or similar, if that would be better?

database/tables/news_items.pg
mwest1066: Maybe the column name could be better -- the sort order used is actually `date, order_by`. The need for `order_by` is to disambiguate cases where we have two items with the same date. An example of this is the current news items (two of them) which will both be inserted into the DB with the same date field. I wanted the sort order to be stable, which is why I added `order_by`.

news_items/001_new_page_layout/index.html
mwest1066: done

news_items/001_new_page_layout/index.html
mwest1066: done

news_items/002_in_browser_editing/index.html
mwest1066: done

pages/news_item/news_item.ejs
mwest1066: Oops, good catch, this was a search-and-replace failure. Fixed.

news_items/001_new_page_layout/index.html
mwest1066: done

mwest1066:
Further to the comments above about the purpose of this feature, I was at least partially inspired by:

https://github.blog

https://aws.amazon.com/blogs/aws/

PR #1981, opened 2020-01-16
Order instances by date

tbretl:
When displayed in a list, course instances were being ordered by `number`, which was defined by hand as a property of `infoCourseInstance.json`.

This PR orders course instances from newest to oldest by earliest and latest access date. "Earliest access date" means the earliest `startDate` that appears in any `accessRule` for the course instance. "Latest access date" means the latest `endDate` that appears in any `accessRule` for the course instance. Course instances having no access dates are listed last.

When these dates are the same, course instances are ordered by `id` from largest to smallest. This means that when a copy of a course instance is made - so, a new instance with the same dates as an old instance - the copy will appear first in any list of course instances.

This PR also drops `number` from the `course_instances` table and from the `infoCourseInstance.json` schema.


pages/instructorCourseAdminInstances/instructorCourseAdminInstances.sql
```diff
@@ -3,19 +3,30 @@ SELECT
...
     ci.long_name,
     ci.id,
+    CASE
+        WHEN d.start_date IS NULL THEN '—'
+        ELSE format_date_full_compact(d.start_date, ci.display_timezone)
+    END AS start_date,
+    CASE
+        WHEN d.end_date IS NULL THEN '—'
+        ELSE format_date_full_compact(d.end_date, ci.display_timezone)
+    END AS end_date,
```
mwest1066: Elsewhere we distinguish between the actual date fields (e.g., `end_date`) and the formatted ones (which we call `formatted_end_date`). I think this is helpful to make sure we sort on the read dates, and not on the formatted date strings (which do weird things with timezones, for example). In this particular case I doubt it's very important in practice, of course.

In this query, I'd suggest calling the computed value above `formatted_end_date`, but leaving the `ORDER BY` clause using the real `end_date`.

tbretl: good call, done

PR #1993, opened 2020-01-20
Fix authorization of effective user

### LINKED ISSUE #1984: Instructor mode with emulated user may have wrong nav bar
Reported by @daviddalpiaz 

Steps to reproduce:

1. Go to the instructor view of a course instance.

2. Select "Your name" -> "Change effective user" and change effective user to a student in the current course instance.

3. Now you will still be in Instructor view for the course instance, but the top navbar will no longer display the course shortcuts like "Issues" and "Questions"

### LINKED ISSUE #1990: Emulated users have no course permissions
We explicitly set course and course instance permissions to `false` when emulating another user: https://github.com/PrairieLearn/PrairieLearn/blob/master/middlewares/authzCourseInstance.js#L93

This can cause confusion. The case that @daviddalpiaz and @mfsilva22 reported is when emulating a TA (even one with course "Viewer" role) you can't see the "Questions" tab.

### LINKED ISSUE #1991: Access denied error page has incorrect navbar
Steps to reproduce:

1. As an instructor, go into a course instance for which students do not currently have access (e.g., an old course instance).

2. Select "Change effective user", then set "Change effective role" to "Student".

3. This will correctly give an error page with the message "Access denied", which is true because students do not have access.

4. The bug is that the navbar is incorrect. For example, the spacing is wrong, the role is incorrectly listed as "(Instructor)", and the link to the "Change effective user" page is wrong.

Related issues include #535 and #759

tbretl:
This PR sets the course permissions `has_course_permission_view`, `has_course_permission_edit`, and `has_course_permission_own` correctly when emulating another user (formerly these were always set to `false`), and leaves the `course_role` alone (formerly, this was set to `None`).

This PR also:
- adds helpful information to the the "Access denied" error page if the effective user does not have access to the current course instance;
- fixes the navbar appearance on this same error page.

**NOTE**

If a different `uid` has been requested, then course permissions are generated for the user with that `uid`. Otherwise, course permissions are generated for the original authenticated user. This means that, in development mode (i.e., in docker on a local machine), switching the `role` or the `mode` without switching the `uid` will result in losing course permissions, because the original authenticated user will be `dev@illinois.edu`, who typically will not have been granted any course permissions. This behavior is not necessarily "wrong," but it may be unexpected - it would probably be best to revisit this whole thing if improvements are made to how permissions are handled in PL.

Fixes #1990 
Fixes #1984
Fixes #1991

### DIFF
--- a/middlewares/authzCourseInstance.js
+++ b/middlewares/authzCourseInstance.js
@@ -120,11 +120,11 @@ module.exports = function(req, res, next) {
                 res.locals.req_date = result.rows[0].req_date;
 
                 // Make sure that we never grant extra permissions
-                if (res.locals.authz_data.has_instructor_view) res.locals.authz_data.has_instructor_view = result.rows[0].has_instructor_view;
-                if (res.locals.authz_data.has_instructor_edit) res.locals.authz_data.has_instructor_edit = result.rows[0].has_instructor_edit;
-                if (res.locals.authz_data.has_course_permission_view) res.locals.authz_data.has_course_permission_view = result.rows[0].permissions_course.has_course_permission_view;
-                if (res.locals.authz_data.has_course_permission_edit) res.locals.authz_data.has_course_permission_edit = result.rows[0].permissions_course.has_course_permission_edit;
-                if (res.locals.authz_data.has_course_permission_own) res.locals.authz_data.has_course_permission_own = result.rows[0].permissions_course.has_course_permission_own;
+                res.locals.authz_data.has_instructor_view = res.locals.authz_data.has_instructor_view && result.rows[0].has_instructor_view;
+                res.locals.authz_data.has_instructor_edit = res.locals.authz_data.has_instructor_edit && result.rows[0].has_instructor_edit;
+                res.locals.authz_data.has_course_permission_view = res.locals.authz_data.has_course_permission_view && result.rows[0].permissions_course.has_course_permission_view;
+                res.locals.authz_data.has_course_permission_edit = res.locals.authz_data.has_course_permission_edit && result.rows[0].permissions_course.has_course_permission_edit;
+                res.locals.authz_data.has_course_permission_own = res.locals.authz_data.has_course_permission_own && result.rows[0].permissions_course.has_course_permission_own;
 
                 // NOTE: When this code is all rewritten, you may want to throw an error if
                 // the user tries to emulate another user with greater permissions, so that
--- a/middlewares/authzCourseInstance.js
+++ b/middlewares/authzCourseInstance.js
@@ -88,13 +88,22 @@ module.exports = function(req, res, next) {
                 if (ERR(err, next)) return;
                 if (result.rowCount == 0) return next(error.make(403, 'Access denied'));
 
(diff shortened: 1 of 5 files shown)


middlewares/authzCourseInstance.js
```diff
@@ -86,15 +86,52 @@ module.exports = function(req, res, next) {
...
-                res.locals.authz_data.has_course_permission_view = false;
-                res.locals.authz_data.has_course_permission_edit = false;
-                res.locals.authz_data.has_course_permission_own = false;
+
+                // Make sure that we never grant extra permissions
+                if (res.locals.authz_data.has_instructor_view) res.locals.authz_data.has_instructor_view = result.rows[0].has_instructor_view;
+                if (res.locals.authz_data.has_instructor_edit) res.locals.authz_data.has_instructor_edit = result.rows[0].has_instructor_edit;
+                if (res.locals.authz_data.has_course_permission_view) res.locals.authz_data.has_course_permission_view = result.rows[0].permissions_course.has_course_permission_view;
+                if (res.locals.authz_data.has_course_permission_edit) res.locals.authz_data.has_course_permission_edit = result.rows[0].permissions_course.has_course_permission_edit;
+                if (res.locals.authz_data.has_course_permission_own) res.locals.authz_data.has_course_permission_own = result.rows[0].permissions_course.has_course_permission_own;
```
mwest1066: I'm more familiar with using logical-and to take the "minimum" of booleans. That is:
```
bool_var = bool_var && override_bool_var;
```

server.js
```diff
@@ -230,7 +230,7 @@ if (config.devMode) {
 }
 
 // clear cookies on the homepage to reset any stale session state
-app.use(/^\/pl\/?/, require('./middlewares/clearCookies'));
+app.use(/^(?:\/?)$|^(?:\/pl\/?)$/, require('./middlewares/clearCookies'));
```
mwest1066: Why are we using non-capturing parentheses (ones that start with `?:`) here?

trombonekenny:
Doesn't `dev@illinois.edu` have administrator rights, so it has course permissions already?

middlewares/authzCourseInstance.js
```diff
@@ -86,15 +86,52 @@ module.exports = function(req, res, next) {
...
+
+                // NOTE: When this code is all rewritten, you may want to throw an error if
+                // the user tries to emulate another user with greater permissions, so that
+                // it is clear why these permissions aren't granted.
+
                 // FIXME: debugging for #422
                 logger.debug('Overridden authz_data', res.locals.authz_data);
```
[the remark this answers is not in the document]
tbretl: done

middlewares/authzCourseInstance.js
tbretl: haha, that would be a little nicer, wouldn't it

server.js
tbretl: habit - reverted to capturing

tbretl:
> Doesn't `dev@illinois.edu` have administrator rights, so it has course permissions already?

We have chosen to revoke any administrator rights when changing the effective user - this seemed the most sensible.

PR #1996, opened 2020-01-21
Fix news item display on error page navbar

mwest1066:


trombonekenny:
The fix works (I tested by putting a hard coded `next(Error())` in server.js where it failed on master and passed OK on this branch.

I also noticed this fix branch isn't based off of master but another fix branch. I'm guessing that was accidental?

mwest1066:
Thanks for checking!

Where do you see that this branch isn't based off of master? You are correct that I originally had it based off of another branch, but I thought I fixed that, and github shows only a single commit?

trombonekenny:
`git diff master` is a part of my local workflow, and it included a lot more than just the changes GitHub.com referenced. That's how I spotted it - well, that and `git la`

PR #2002, opened 2020-01-23
Fix permissions on issues page

### LINKED ISSUE #1995: Instructor close all issues button no longer working
The "Close all issues" button throws a Not Found error and doesn't remove the open issues. This is new with the recent deploy.

tbretl:
This PR allows course owners and editors to open and close issues, and nobody else (except administrators).

Formerly, access was being granted only to instructors - this broke the ability to open and close issues through the `pl/course` route, where we don't check whether or not users are instructors.

This PR also adds a test that you can report an issue and close it either at the course level or the course instance level.

Fixes #1995

### DIFF
--- a/pages/instructorIssues/instructorIssues.js
+++ b/pages/instructorIssues/instructorIssues.js
@@ -143,7 +143,7 @@ router.get('/', function(req, res, next) {
 });
 
 router.post('/', function(req, res, next) {
-    if (!res.locals.authz_data.has_instructor_edit) return next();
+    if (!res.locals.authz_data.has_course_permission_edit) return next();
     if (req.body.__action == 'open') {
         let params = [
             req.body.issue_id,
--- a/pages/instructorIssues/instructorIssues.js
+++ b/pages/instructorIssues/instructorIssues.js
@@ -143,7 +143,7 @@ router.get('/', function(req, res, next) {
 });
 
 router.post('/', function(req, res, next) {
-    if (!res.locals.authz_data.has_course_permission_edit) return next();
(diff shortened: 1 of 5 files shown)


pages/instructorIssues/instructorIssues.js
```diff
@@ -143,7 +143,7 @@ router.get('/', function(req, res, next) {
 });
 
 router.post('/', function(req, res, next) {
-    if (!res.locals.authz_data.has_instructor_edit) return next();
+    if (!res.locals.authz_data.has_course_permission_edit) return next();
```
mwest1066: Won't this fail for users who are instructors but not course editors?

tbretl: Yes.

I could go either way on this (I had an "or" with `has_instructor_edit` for a while, then got rid of it). On the one hand, these issues are a course-level thing, resolving them requires editing course content and syncing these edits, etc. On the other hand, current practice might be to ask non-editors to resolve issues by pushing/pulling content to the course git repo. Thoughts?

mwest1066: I don't have any strong thoughts about the "right" answer here.

But just in practice at the moment, I believe instructors-who-aren't-course-editors can see issues, right? And they used to be able to close issues? If so, I'm generally in favor of letting them continue to do this until the new perm system is merged, to avoid confusion. Once the new perm system is merged we can do a big news item explaining how it works, but until then I think it would be best to minimize the removal of capabilities.

tbretl: Sure, done

PR #2004, opened 2020-01-24
Fix angle checks for vectors in `pl-drawing` element

nicknytko:
As pointed out by Vadim on Slack changing the `angle-tol` parameter previously did not change anything, this fix makes the grader use that parameter when checking tolerances.

### DIFF
--- a/elements/pl-drawing/pl-drawing.py
+++ b/elements/pl-drawing/pl-drawing.py
@@ -1925,15 +1925,15 @@ def grade(element_html, data):
         rang = ref['angle']
         rang_bwd = ref['angle'] + 180
         rang_rad = rang * (np.pi / 180.0)
-        
+
         # Defining the error box limit in the direction of the vector
         max_backward = ref['offset_backward'] + tol
         max_forward = ref['offset_forward'] + tol
-        
+
         # Check the angles
         error_fwd = abserr_ang(rang, eang)
         error_bwd = abserr_ang(rang_bwd, eang)
-        
+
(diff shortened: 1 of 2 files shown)


elements/pl-drawing/pl-drawing.py
```diff
@@ -1876,6 +1879,11 @@ def grade(element_html, data):
     def abserr(x, xapp):
         return np.abs(x - xapp)
 
+    def abserr_ang(ref, x):
+        ref = ref % 360
+        x = x % 360
```
mwest1066: Do we need the two lines above?

elements/pl-drawing/pl-drawing.py
```diff
@@ -1912,26 +1920,25 @@ def comp_vector(ref, st):
...
         max_forward = ref['offset_forward'] + tol
 
         # Check the angles
-        error1 = abserr(rang, eang)
-        error2 = abserr(rang2, eang)
-        error3 = abserr(rang3, eang)
-        error4 = abserr(rang4, eang)
+        error_fwd = abserr_ang(rang, eang)
+        error_bwd = abserr_ang(rang_bwd, eang)
+
```
mwest1066: very nice cleanup here --- much easier to verify correctness!

elements/pl-drawing/pl-drawing.py
nicknytko: woops, you're totally right I don't think the mod is actually needed here

PR #2010, opened 2020-01-30
Fix MathJax rendering for `pl-drawing` element on Safari

nicknytko:
Fixed math rendering on Safari, as brought up on TAM slack. 

Safari did not like setting `.outerHTML` to update SVG references, so this has been switched to using the less yucky `.replaceChild()`.


nicknytko:
@mwest1066 I will keep bugging you about this until you review 🙂

Issue #2062, opened 2020-02-06
Empty uids in access rule allows anyone to access


mwest1066:
Reported by @butala

Make an access rule with `"uids": []` (no UIDs listed). This allows anyone to access, whereas it should logically prevent anyone from accessing.

mwest1066:
Fixed by #2063

PR #2063, opened 2020-02-06
Fix handling of empty JSON arrays during sync

mwest1066:
Fixees #2062

The SQL expression that we were using to convert JSON arrays to SQL
arrays was converting zero-length arrays to NULL, rather than to
zero-length SQL arrays. This PR adds explicit PLPGSQL functions to do
this conversion. These map:
```
NULL::jsonb         -> NULL::text[]
'null'::jsonb       -> NULL::text[]
'[]'::jsonb         -> array[]::text[]
'["a"]'::jsonb      -> array['a']::text[]
'["a", "b"]'::jsonb -> array['a', 'b']::text[]
```


sprocs/jsonb_array_to_double_precision_array.sql
```diff
@@ -0,0 +1,27 @@
...
+    END IF;
+
+    len := jsonb_array_length(jsonb_array);
+
+    IF len IS NULL THEN
+        double_precision_array := NULL;
+    ELSIF len = 0 THEN
+        double_precision_array := array[]::double precision[];
+    ELSE
+        double_precision_array := (SELECT array_agg(v::double precision) FROM jsonb_array_elements_text(jsonb_array) AS tmp(v));
```
tbretl: The reason for `AS tmp(v)` is not obvious to me here.

sprocs/jsonb_array_to_text_array.sql
```diff
@@ -0,0 +1,27 @@
...
+    END IF;
+
+    len := jsonb_array_length(jsonb_array);
+
+    IF len IS NULL THEN
+        text_array := NULL;
+    ELSIF len = 0 THEN
+        text_array := array[]::text[];
+    ELSE
+        text_array := (SELECT array_agg(v) FROM jsonb_array_elements_text(jsonb_array) AS tmp(v));
```
tbretl: Again, I don't understand the `AS tmp(v)` syntax.

tests/testSproc-check_assessment_access.js
```diff
@@ -161,6 +161,42 @@ describe('sproc check_assessment_access* tests', function() {
             });
         });
 
+        it('pass if all parameter match match except UID with no uids', function(callback) {
```
tbretl: I think you mean `pass if all parameters match except UID with no uids`

sprocs/jsonb_array_to_double_precision_array.sql
mwest1066: It's a table alias, so I don't have to write
```
array_agg(jsonb_array_elements_text::double precision)
```

The general table alias syntax is `AS tablename(col1name, col2name)`. You don't have to provide the column aliases if you don't want them. In this case they aren't strictly necessary, so I just pushed a change `tmp(v) -> v`.

tests/testSproc-check_assessment_access.js
mwest1066: fixed

PR #2077, opened 2020-02-14
Fix nightly calculation of question statistics

### LINKED ISSUE #1323: Consider changing question stats calculation interval to 1 hour

### LINKED ISSUE #1635: Daily cron job timeout
We had a previous case of transaction timeouts in #1623, which we fixed in #1629 

Now we’ve still got transaction timeouts on the daily cron jobs, but only in production (not for `pl-dev`).

mwest1066:
The nightly cron job to update question statistics used to iterate over assessments inside the SQL. However, this now takes more than 10 minutes so we timeout the query. This PR changes it to iterate over assessments in JS, so each individual SQL query should be below 10 minutes.

Fixes #1635
Fixes #1323

### DIFF
--- a/cron/calculateAssessmentQuestionStats.sql
+++ b/cron/calculateAssessmentQuestionStats.sql
@@ -1,4 +1,3 @@
-SELECT
-    assessment_questions_calculate_stats_for_assessment(a.id)
-FROM
-    assessments AS a;
+-- BLOCK select_assessments
+SELECT *
+FROM assessments;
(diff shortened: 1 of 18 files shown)


cron/calculateAssessmentQuestionStats.sql
```diff
@@ -1,4 +1,3 @@
-SELECT
-    assessment_questions_calculate_stats_for_assessment(a.id)
-FROM
-    assessments AS a;
+-- BLOCK select_assessments
+SELECT *
```
trombonekenny: Do we want a smarter query that only updates on the assessments that have had data changed since the last time stats were run? Or is that not worth it if it churns over all the assessments in negligible time?

mwest1066:
See #1323

PR #2079, opened 2020-02-15
Added demo element with hidden tag

dowobeha:
The existing custom element doesn't show how to accept dynamic user input and compare it to a correct answer. This one does.


nicknytko:
Can this question be added to one of the demo assessments?  To `E1` or `HW1`?

exampleCourse/elements/clickable-image/clickable-image.mustache
```diff
@@ -0,0 +1,29 @@
+{{#question}}
+<div class="example-course-custom-element">
+    <p class="big example-shared-style">Click on the image this many times: {{number}}</p>
+    <input id="my_hidden_field" name="{{answers_name}}" type="hidden" value="0" />
+    <p><img src="{{image_url}}" width="50px" id="IMark"></p>
```
nicknytko: Can we give these components unique ids?  I know this is a demo element but I think it would be good practice to prepend the element uuid to the id fields.

nicknytko:
@tdy can you take a quick look at the updated code as a sanity check?

exampleCourse/questions/demo/custom/elementHiddenTag/question.html
```diff
@@ -0,0 +1,6 @@
+<pl-question-panel>
+    <p> This question has an example of an element that uses hidden input tags to send information to the server. Hidden inputs are useful for custom JavaScript elements that have custom data to be graded. </p>
```
tdy, line 2:
```suggestion
    <p> This question has an example of an element that uses hidden input tags to send information to the server. Hidden inputs are useful for custom JavaScript elements that have custom data to be graded. Keep in mind that users can manipulate tags using browser devtools, so make sure that all grading is handled server-side when designing questions with hidden inputs. For example, do not grade the question client-side and return the number of earned points in a hidden input. </p>
```

tdy:
Works well on my end. Just added a warning about not using hidden inputs to send grade-sensitive info.

PR #2080, opened 2020-02-16
[WIP] RFC: Question Sets

nwalters512:
This PR contains a work-in-progress RFC for question sets, a proposed way of sharing content across institutions and courses.

Structure is based on React's RFC template: https://github.com/reactjs/rfcs/blob/master/0000-template.md


rfcs/0000-question-sets.md
```diff
@@ -0,0 +1,150 @@
+# Summary
+
+A mechanism for allowing sets of questions to be published and used by multiple courses.
```
mwest1066, line 3: I guess I'm wondering whether we should call these "sharable courses" by a different name ("questionSets") or whether we should just allow any course to set a flag like `"sharable": true`?

One advantage of making full courses sharable is that they could include not just lists of questions, but also template assessments that could be copied. For example, if I made a bank of questions for "Learning Python 101", I could also put together sample homeworks (and even exams). We could have a "copy assessment" button that mimics the current "copy question" button, so that another instructor could copy the homeworks into their course and the questions would all be automatically linked back to the source course. Then they'd just need to adjust the dates, maybe add or delete a question or two, and they'd be ready to go.

I could even provide entire sample course instances that could be copied. Maybe I make course instances for a "Full 15-week Python Course" and another "4-week intro course". Then we could have a button so an instructor could just copy the entire course instance, and then go through and make small tweaks as they want (editing dates, at a minimum).

I think this "course in a box" could be very powerful.

rfcs/0000-question-sets.md
```diff
@@ -0,0 +1,150 @@
...
+  "uuid": "754dac56-fc1f-4888-a74a-1b365fd48376",
+  "questionSet": true
+}
+```
+
+Question sets will have a unique, human readable name, like `uiuc-data-structures`.
+
+Questions in a question set are defined the same way that they are for courses. That is, a directory in a `questions` directory. For example, the file `questions/avl-tree/info.json` defines a question with the ID `avl-tree`.
+
+Questions from a question set can be referenced with a combination of question set ID and question ID. For instance, in the following simplified `infoAssessment.json`, `uiuc-data-structures/avl-tree` references the `avl-tree` question from the `uiuc-data-structures` question set.
```
mwest1066, line 20: We frequently have requests for #2360. To enable that, perhaps we should use some special character to indicate source course questions? Like `@uiuc-data-structures/avl-tree`?

rfcs/0000-question-sets.md
```diff
@@ -0,0 +1,150 @@
...
+* The person running PrairieLearn locally is able to clone the question set repository. In this case, they can add the course to their local Docker container. Then, questions from question sets can be accessed exactly like they would be in a production environment.
+
+In the future, one could build a feature that allows an authorized user to export a copy of a question set for use with local development.
+
+## Safety guarantees
+
+The following guarantees should exist:
+
+* There is no ambiguity between whether a question belongs to a course or a question set.
+* Syncing a new version question set should not break any existing assessments that reference questions from it.
```
mwest1066, line 66:
```suggestion
* Syncing a new version of a question set should not break any existing assessments that reference questions from it.
```

rfcs/0000-question-sets.md
```diff
@@ -0,0 +1,150 @@
...
+* There should never be an assessment on a production instance that references a question that does not exist.
+
+The following facts and behaviors ensure those guarantees:
+
+* The presence of a `/` in a question ID in an assessment definition will unambiguously differentiate question set questions from course-local questions. Since question IDs are derived from the name of a directory in a `questions` directory and `/` is not a valid character in a directory name, questions are implicitly prohibited from containing.
```
mwest1066, line 71: See above for a proposal to use `@` instead.

rfcs/0000-question-sets.md
```diff
@@ -0,0 +1,150 @@
...
+
+* When a course is synced, all questions that are referenced by its assessments will be checked. If a referenced question set is not available on that instance, or a question does not exist, the sync will error. *(Note that this restriction is relaxed in local installations of PrairieLearn, where a user may not have the necessary permissions to clone the question set.)*
+
+## Access control
+
+Even though question sets are "courses", they will not be visible in PrairieLearn for students to register for.
+
+On a given PrairieLearn instance, the ability to sync a question set is controlled the same way that it is for courses: via the "course access roles" configuration.
+
+# Drawbacks
```
mwest1066, line 85: I like this section!

rfcs/0000-question-sets.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Access control within a PrairieLearn instance
+
+As with normal courses, access to question sets should be carefully controlled. It should also be consistent: if I'm staff for a course instance, I should be able to access question sets that are referenced by my courses' assignments. It's not clear how this access should be requested, granted, or managed.
+
+These permissions should likely not live in a question set's repo - they're liekly going to be specific to the instance of PrairieLearn a course is running on.
+
+Perhaps permissions should exist at the course level - a PrairieLearn admin can grant course X the ability to use and view question set Y. That permission would be transitive: if I'm course staff for course X, and course X has the ability to use question set Y, then I should be able to view and use anything in question set Y.
```
mwest1066, line 146: I think the risk with transitive permissions is that some child courses might have very large and permissive course staffs. I don't really know what to do about this 😄

rfcs/0000-question-sets.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Discovery
+
+At first, the existencen of question sets will probably be shared directly between staff from courses. However, this word-of-mouth does not scale well. There should be a way to see what question sets exist, both globally and on a particular instance of PrairieLearn.
```
mwest1066, line 150:
```suggestion
At first, the existence of question sets will probably be shared directly between staff from courses. However, this word-of-mouth does not scale well. There should be a way to see what question sets exist, both globally and on a particular instance of PrairieLearn.
```

rfcs/0000-question-sets.md
nwalters512: I like this idea! I'd still want the same semantics + restrictions as outlined below for questions in a shareable course, but that wouldn't impact our ability to provide assessments like you're suggesting.

rfcs/0000-question-sets.md
nwalters512: Really good point - agreed! We can prevent people from syncing questions that start with `@` to prevent any ambiguity.

rfcs/0000-question-sets.md
mwest1066: Yep! I like all those restrictions below for sharable courses.

rfcs/0000-question-sets.md
mwest1066: Yes! I was thinking earlier that we might want to start enforcing some rules about allowed characters in names. I noticed that CS 233 has some exams this semester that start with `@` (they are called `@Home1`, etc).

rfcs/0000-question-sets.md
nwalters512: I wasn't sure about this initially either. There are definitely some security risks with this, but I don't see how to prevent an insane amount of complexity or unexpected behavior without this transitive model. There are a lot of valid use cases for your standard member of course staff accessing all questions (including shared ones):

* Building an exam
* Reviewing an exam prior to launching it
* Reviewing student-submitted issues on an assessment
* Helping a student in office hours

The only case where I really see problems here is if a) a student is course staff for X, b) a student is taking course Y, and c) X and Y both use the same set of shared questions. That seems unlikely to me, so I'm willing to overlook it, tbh 🙂

mwest1066: I agree this is a horrible can of worms to try and lock it down further.

Do shared questions appear in the big list of questions within a child course?

rfcs/0000-question-sets.md
tdy: > whether we should just allow any course to set a flag like `"sharable": true`?

I'm still not too familiar with PL access rules, so apologies if this was actually addressed in the Access Control section (course access roles?), but:

If an instructor sets `sharable`, they probably won't want it to just be open season for anyone to click "clone question" (esp other schools). This has come up with MCS/Coursera, where instructors seemed fine with sharing their content with other schools but only on their own terms.

rfcs/0000-question-sets.md
```diff
@@ -0,0 +1,150 @@
...
+
+As PrairieLearn becomes more widely used, courses want to be able to easily share questions. Currently this is accomplished by copying question files between courses. However, this approach requires a lot of manual effort to copy questions around and keep them up-to-date as questions are improved. By adding first-class support for sharing questions, we hope to reduce the burden of sharing questions.
+
+# Detailed design
+
+Question sets will be modeled similarly to courses: as GitHub repositories containing questions.
+
+## Usage across multiple PrairieLearn instances.
+
+By design, PrairieLearn is a decentrilized system. There are now several production instances running at different universities, and dozens (if not hundreds) of individuals run PrairieLearn on their computers for local course development. The architecture of question sets needs to take this decentralized nature into account.
```
tdy, line 48:
```suggestion
By design, PrairieLearn is a decentralized system. There are now several production instances running at different universities, and dozens (if not hundreds) of individuals run PrairieLearn on their computers for local course development. The architecture of question sets needs to take this decentralized nature into account.
```

rfcs/0000-question-sets.md
mwest1066: Oh yes! Totally agreed! This PR doesn't describe anything about how sharing permissions work (although maybe this PR would be a good place to add this?). In our earlier conversations we had talked about having explicit permissions (like "Course X" can use questions from "Course Y") which would be set by the course owner.

rfcs/0000-question-sets.md
nwalters512: I don't think they should, but I do think a separate view like that would be helpful.

rfcs/0000-question-sets.md
nwalters512: Good catch 😂

rfcs/0000-shareable-courses.md
```diff
@@ -0,0 +1,150 @@
...
+* Consistent across instances of PrairieLearn
+* Human-readable
+
+Because they must be human readable, a shareable course's UUID is unsuitable as the user-facing course ID. The `name` value in `infoCourse.json` could be repurposed for this, but it should have additional restrictions (no spaces) and has a different semantic meaning. A new `id` value could be introduced that has no meaning in standard courses.
+
+## Access control within a PrairieLearn instance
+
+As with normal courses, access to shareable courses should be carefully controlled. It should also be consistent: if I'm staff for a course instance, I should be able to access shareable courses that are referenced by my courses' assignments. It's not clear how this access should be requested, granted, or managed.
+
+These permissions should likely not live in a shareable courses's repo - they're liekly going to be specific to the instance of PrairieLearn a course is running on.
```
mwest1066, line 144:
```suggestion
These permissions should likely not live in a shareable courses's repo - they're likely going to be specific to the instance of PrairieLearn a course is running on.
```

rfcs/0000-shareable-courses.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+Deprecated questions could render a warning in instructor views of the question, on a question's row in an assessment overview, or in other useful locations across PrairieLearn.
+
+# Unresolved questions and concerns
+
+## Feature name
+
+With the existence of assessment sets, "set" might become too much of an overloaded term if used for this feature. A name like "question collections" may be more appropriate.
+
```
mwest1066, line 129: I think this subsection can be deleted, now that we've changed to calling this a "shareable" course.

```suggestion

```

rfcs/0000-shareable-courses.md
```diff
@@ -0,0 +1,150 @@
...
+* Syncing a new version of a shareable course should not break any existing assessments that reference questions from it.
+* There should never be an assessment on a production instance that references a question that does not exist.
+
+The following facts and behaviors ensure those guarantees:
+
+* A leading `@` in a question ID in an assessment definition will unambiguously differentiate shareable course questions from course-local questions.
+
+* Once a course has been designated as shareable, it must remain shareable. That is, once a course with `"shareable": true` is synced to an instance of PrairieLearn, attempting to sync the same course with `"shareable": false` will be a sync error.
+
+* Once a question in a shareable course has been synced to an instance of PrairieLearn, attempting to perform a sync where the question would be renamed or deleted will be a sync error. *(Is there any special consideration necessary for local development?)*
```
mwest1066, line 75: I think that in local development all of these safety guarantees should be disabled, so that you can rename new questions before pushing them to production.

mwest1066, line 75: Should we add a notion of question aliases? That is, alternative QIDs for a question specified in the `info.json` file which can also be used to reference this question. In this case, the sync rule might be that renames are allowed, but only if the old name is still listed as an alias.

The in-browser editor could add an alias automatically whenever renaming questions in a shareable course.

rfcs/0000-shareable-courses.md
```diff
@@ -0,0 +1,150 @@
...
+
+# Alternatives
+
+## No formal support for sharing questions
+
+This is the current state of PrairieLearn. "Sharing" a question involves manually copying all of its files from one course to another. From a technical perspective, this solution is simple, as it requires no special treatment in PrairieLearn code. However, it scales poorly as it requires a lot of manual action to propagage question enhancements or fixes.
+
+## Share questions directly from courses
+
+Instead of having a special entity that stores shared questions, one could share questions directly from an existing course. From a technical perspective this is likely easier than the proposal of dedicated shareable courses.
```
mwest1066, line 101:
```suggestion
Instead of having a special flag to indicate a shareable course, one could share questions directly from any existing course. From a technical perspective this is slightly easier than the proposal of dedicated shareable courses.
```

rfcs/0000-question-sets.md
mwest1066: I can see this both ways. Sometimes I'd want to see only the questions that are actually local to the current course, but when I'm are trying to find another question to include on the homework, I'd like to see everything available in a big list that I can search.

Issue #2097, opened 2020-02-27
Changes "News" badge from red colour to something less scary


nicknytko:
When new 'news' posts are added users get a red notification badge displaying how many posts are unread.  Red is a scary colour, lets change this to something nicer like blue.

KaveeshaW:
I'll take this. I found what to change as well. Its under the navbar.ejs in the partials folder.

nicknytko:
Hey!  I can help you create a pull request for this tomorrow.

mwest1066:
Implemented in #2152

PR #2102, opened 2020-02-29
Changed span class of the navbar (#2097)

KaveeshaW:
Changed the span class from badge-danger to badge-primary so that the notification news post is blue instead of red.


mwest1066:
Can we also change the "Unread" badge color on the news pages, so that it matches this change?

https://github.com/PrairieLearn/PrairieLearn/blob/master/pages/news_items/news_items.ejs#L34

zbs2:
I'll be working on this under pl-dev-students

mwest1066:
Implemented in #2152

PR #2105, opened 2020-02-29
Displaying issue status to student open/closed

shresta4:
Displayed whether the issue is open/closed to the student. 
Below, issue status is "open". 

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4920,8 +4920,8 @@
             "from": "git://github.com/mathjax/MathJax-src.git",
             "requires": {
                 "esm": "^3.2.25",
-                "mj-context-menu": "^0.2.0",
-                "speech-rule-engine": "^3.0.0-beta.6"
+                "mj-context-menu": "^0.2.2",
+                "speech-rule-engine": "^3.0.0-beta.8"
             }
         },
         "mdast-util-definitions": {
--- a/pages/partials/question.ejs
+++ b/pages/partials/question.ejs
@@ -27,9 +27,9 @@
       </tbody>
     </table>
 
-    <% if (devMode || authz_data.has_instructor_view || is_administrator) { %>
+    
     <div class="card-body border border-bottom-0 border-left-0 border-right-0">
-      <% if (is_administrator && issue.system_data.courseErrData ) { %>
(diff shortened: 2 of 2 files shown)


package-lock.json
```diff
@@ -4920,8 +4920,8 @@
             "from": "git://github.com/mathjax/MathJax-src.git",
             "requires": {
                 "esm": "^3.2.25",
-                "mj-context-menu": "^0.2.0",
-                "speech-rule-engine": "^3.0.0-beta.6"
+                "mj-context-menu": "^0.2.2",
+                "speech-rule-engine": "^3.0.0-beta.8"
```
mwest1066: Is this change needed in this PR?

pages/partials/question.ejs
```diff
@@ -41,19 +41,33 @@
...
+        <% if (issue.open) { %>
+          <span type="button" class="badge badge-danger" data-toggle="collapse" href="#">
+            Open
+          </span>
+          <% } else { %>
+            <span type="button" class="badge badge-success" data-toggle="collapse" href="#">
+              Closed
+            </span>
+          <% } %>
+      </p>
```
mwest1066: Can we put the issue status in the table above? That table is used for standard issue data, and I think "status" is like this. It would also avoid the need to change around the authorization checking.

pages/partials/question.ejs
```diff
@@ -41,19 +41,33 @@
...
         <button type="button" class="btn btn-xs btn-secondary" data-toggle="collapse" href="#issue-system-data-<%= iIssue %>" aria-expanded="false" aria-controls="#issue-system-data-<%= iIssue %>">
           Show/hide
         </button>
       </p>
+      <% } %>
+      <p><strong>Issue status:</strong>
+        <% if (issue.open) { %>
+          <span type="button" class="badge badge-danger" data-toggle="collapse" href="#">
+            Open
+          </span>
```
mwest1066: Why are these `type="button"`? They don't seem to anything when clicked, if I'm reading the code correctly?

dkim221:
I'll work on this project.

mwest1066:
Re-PRed as #2203

PR #2108, opened 2020-03-01
Darktheme

### LINKED ISSUE #1190: Dark/Night theme
Is there a dark/night theme for prarielearn? If not, I'd like to start working on it!

matttang7:
Closes #1190 

- Created additional table themes that has userid and theme number to query and decide which theme to load

- Added styled toggle button

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1051,6 +1051,11 @@
             "resolved": "https://registry.npmjs.org/bootstrap/-/bootstrap-4.4.1.tgz",
             "integrity": "sha512-tbx5cHubwE6e2ZG7nqM3g/FZ5PQEDMWmMGNrCUBVRPHXTJaH7CBDdsLeu3eCh3B1tzAxTnAbtmrzvWEvT2NNEA=="
         },
+        "bootstrap4-toggle": {
+            "version": "3.6.1",
+            "resolved": "https://registry.npmjs.org/bootstrap4-toggle/-/bootstrap4-toggle-3.6.1.tgz",
+            "integrity": "sha512-eRejcTc9YurhZ64nHY9Ii9DQn+F9/R74H9RPoeANVM3N1+C2lZ2tUuFCx1w3orOJ1y/iG4A7iCwdDZphMDIrYg=="
+        },
         "boxen": {
             "version": "1.3.0",
             "resolved": "https://registry.npmjs.org/boxen/-/boxen-1.3.0.tgz",
@@ -4920,8 +4925,8 @@
             "from": "git://github.com/mathjax/MathJax-src.git",
             "requires": {
                 "esm": "^3.2.25",
-                "mj-context-menu": "^0.2.0",
-                "speech-rule-engine": "^3.0.0-beta.6"
+                "mj-context-menu": "^0.2.2",
+                "speech-rule-engine": "^3.0.0-beta.8"
             }
         },
         "mdast-util-definitions": {
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
         "blocked-at": "^1.2.0",
         "body-parser": "^1.19.0",
         "bootstrap": "^4.4.1",
+        "bootstrap4-toggle": "^3.6.1",
         "byline": "^5.0.0",
         "chart.js": "^2.9.3",
         "cheerio": "^0.22.0",
--- a/pages/partials/head.ejs
+++ b/pages/partials/head.ejs
@@ -6,7 +6,6 @@
 <% if(locals.darkmode == 1){%>
 <link href="/node_modules/bootstrap/dist/css/bootstrap1.min.css" rel="stylesheet" />
 <% }%>
-<%=JSON.stringify(locals.darkmode)%>
 <link href="/stylesheets/colors.css" rel="stylesheet" />
 <link href="/stylesheets/local.css" rel="stylesheet" />
 <script src="/node_modules/jquery/dist/jquery.min.js"></script>
--- a/pages/partials/head.ejs
+++ b/pages/partials/head.ejs
@@ -3,6 +3,10 @@
 <meta name="viewport" content="width=device-width, initial-scale=1">
 <title>PrairieLearn</title>
 <link href="/node_modules/bootstrap/dist/css/bootstrap.min.css" rel="stylesheet" />
+<% if(locals.darkmode == 1){%>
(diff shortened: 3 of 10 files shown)


pages/userSettings/userSettings.ejs
```diff
@@ -2,6 +2,8 @@
 <html>
   <head>
     <%- include('../partials/head'); %>
+    <link href="https://cdn.jsdelivr.net/gh/gitbrent/bootstrap4-toggle@3.6.1/css/bootstrap4-toggle.min.css" rel="stylesheet">
+    <script src="https://cdn.jsdelivr.net/gh/gitbrent/bootstrap4-toggle@3.6.1/js/bootstrap4-toggle.min.js"></script>
```
mwest1066: We can't include anything from a CDN because PL needs to work without a network.

pages/userSettings/userSettings.ejs
```diff
@@ -69,7 +71,25 @@
...
+          <fieldset>        
+              <%if(checked) {%>
+              <input type="checkbox" name="dark_mode" value="0" id="darkid" checked = true checked data-toggle="toggle" 
+              data-on="🌙" data-off="☀" data-onstyle="outline-primary" data-offstyle="outline-warning">dark_mode<br>  
+              <%} else { %>   
+                <input type="checkbox" name="dark_mode" value="0" id="darkid" data-toggle="toggle" 
+                data-on="🌙" data-off="☀" data-onstyle="dark" data-offstyle="primary">dark_mode<br> 
+              <%}  %> 
+              <!-- <input type="checkbox" name="other_thing1" value="2">other_thing1<br>     
+              <input type="checkbox" name="other_thing2" value="3">other_thing2<br>    -->
```
mwest1066: Remove commented-out code.

package.json
```diff
@@ -19,6 +19,7 @@
         "blocked-at": "^1.2.0",
         "body-parser": "^1.19.0",
         "bootstrap": "^4.4.1",
+        "bootstrap4-toggle": "^3.6.1",
```
mwest1066: I'd rather not include novel UI elements if there is a standard HTML element we can use instead. For example, I think we can just use a regular checkbox or radio in this case?

server.js
```diff
@@ -863,8 +863,15 @@ module.exports.insertDevUser = function(callback) {
         var params = {user_id};
         sqldb.query(sql, params, function(err, _result) {
             if (ERR(err, callback)) return;
-            callback(null);
+            var sql
+            = 'INSERT INTO theme (id, user_id, themenum)'
+            + ' VALUES ($user_id, $user_id, 1)';
```
mwest1066: Please put all SQL code in the `.sql` files.

pages/userSettings/userSettings.js
```diff
@@ -37,34 +37,69 @@ router.post('/', (req, res, next) => {
             if (ERR(err, next)) return;
             res.redirect(req.originalUrl);
         });
-    } else {
+    }
+    else if(req.body.__action === 'update_settings'){
```
mwest1066: Please follow code style of other code (e.g., spaces around keywords and parens).

pages/userSettings/userSettings.js
```diff
@@ -37,34 +37,69 @@ router.post('/', (req, res, next) => {
...
+        if('dark_mode' in req.body){
+            var params = {
+                user_id: res.locals.authn_user.user_id,
+                dark_mode: 1,
```
mwest1066: Is this parameter a binary switch (`darkmode = 0/1`) or a theme number? Sometimes we seem to be doing one and sometimes the other.

mwest1066:
The test suite is currently failing.

mwest1066:
Rather than having an explicit toggle, I'd prefer to follow the user's OS-level preferences with the `prefers-color-scheme: dark` media query.

package-lock.json
```diff
@@ -4920,8 +4925,8 @@
             "from": "git://github.com/mathjax/MathJax-src.git",
             "requires": {
                 "esm": "^3.2.25",
-                "mj-context-menu": "^0.2.0",
-                "speech-rule-engine": "^3.0.0-beta.6"
+                "mj-context-menu": "^0.2.2",
+                "speech-rule-engine": "^3.0.0-beta.8"
```
mwest1066: Are these changes needed?

pages/partials/head.ejs
```diff
@@ -3,6 +3,9 @@
 <meta name="viewport" content="width=device-width, initial-scale=1">
 <title>PrairieLearn</title>
 <link href="/node_modules/bootstrap/dist/css/bootstrap.min.css" rel="stylesheet" />
+<% if(locals.darkmode == 1){%>
+<link href="/node_modules/bootstrap/dist/css/bootstrap1.min.css" rel="stylesheet" />
```
mwest1066: The `node_modules` directory is automatically created by `npm ci`, so we can't put things in there by hand. Extra themes should be installed by npm (preferably) or be put in `/public/stylesheets` (if necessary).

pages/partials/head.ejs
```diff
@@ -3,6 +3,9 @@
...
+<% }%>
```
mwest1066: I'm worried about including random bootstrap themes. These have a bad history of becoming unsupported. Is there any more standard way to get darkmode with bootstrap?

mwest1066:
What are the impacts of darkmode on questions? Do custom question elements (like `pl-file-editor` and `pl-drawing`) work correctly?

mwest1066:
I'm closing this PR for now. If anyone wants to address the issues above then please open a new PR based on these commits.

echuber2:
The file editor uses ACE so it could be toggled to use a dark theme itself, but courses may already be styling that. The other features like drawing stuff, MathJax rendering, etc. may be troublesome.

harsh183:
Well it can be a start I suppose? Marked as experimental or Beta and slowly things get better. Ace has a really easy setting for setting up dark mode.

echuber2:
Yes, there's nothing stopping the ACE editor (wrapped by `pl-file-editor` in PL) from using one of its own dark themes. However courses can already style that when they inject the editor, so they may be overriding it. I think it would be good to add a dropdown in the corner of the file editor for selecting a color scheme. Later, we could add an option for "default" if PL overrides that.

harsh183:
Sure. Even if there are some light elements just having a dark background with light text makes a large difference. That plus ace dark support should work for like 95% of questions as is (not many people do fancy stuff really).

PR #2144, opened 2020-03-06
Fix course instance access LTI check

trombonekenny:
Updates the logic for course instances protected with `"institution": "LTI"` to force unauthorized unless the LTI course instance matches. Prior to this we weren't restricting access when comparing a NULL value, so users not associated with an LTI course instance could still access the course when protected with `"institution": "LTI"`.

### DIFF
--- a/sprocs/check_course_instance_access_rule.sql
+++ b/sprocs/check_course_instance_access_rule.sql
@@ -43,14 +43,16 @@ BEGIN
         -- we have an explicit institution restriction
 
         IF course_instance_access_rule.institution = 'LTI' THEN
+            available := FALSE;
+
             -- get the uid row from users
             SELECT *
             INTO user_result
             FROM users
             WHERE users.uid = check_course_instance_access_rule.uid;
 
-            IF user_result.lti_course_instance_id != course_instance_access_rule.course_instance_id THEN
-                available := FALSE;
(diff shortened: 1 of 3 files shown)


trombonekenny:
It looks the logic changes to LTI and institutions in #1958 missed this case, so the open access should have been since ~Jan 10, 2020 until this pushes.

sprocs/check_course_instance_access_rule.sql
```diff
@@ -43,14 +43,16 @@ BEGIN
...
             -- get the uid row from users
             SELECT *
             INTO user_result
             FROM users
             WHERE users.uid = check_course_instance_access_rule.uid;
 
-            IF user_result.lti_course_instance_id != course_instance_access_rule.course_instance_id THEN
-                available := FALSE;
+            IF user_result.lti_course_instance_id = course_instance_access_rule.course_instance_id THEN
+                available := TRUE;
```
mwest1066: I think this will override any earlier failed checks. For example, suppose one of the `date` checks failed and set `available = FALSE`. Then we get here and set it `TRUE` again.

Was the original bug caused by regular users having `user_result.lti_course_instance_id = NULL`, which won't trigger the `!=` comparison? If so, I think an alternative fix would be to replace `!=` with `IS DISTINCT FROM`.

PR #2146, opened 2020-03-07
Add dev docs about passing JSON object arrays to SQL

mwest1066:


doc/dev-guide.md
```diff
@@ -458,6 +458,28 @@ sqldb.query(sql.select_questions, params, ...);
 SELECT * FROM questions WHERE id IN (SELECT unnest($id_list::INTEGER[]));
 ```
 
+* To pass a lot of data to SQL a useful pattern is to send a JSON object array and unpack it in SQL to the equivalent of a table. This is the pattern used by the "sync" code, such as [sprocs/sync_news_items.sql](https://github.com/PrairieLearn/PrairieLearn/blob/master/sprocs/sync_news_items.sql). For example, given the JavaScript:
+```javascript
+let data = [
+    {a: 5, b: "foo"},
+    {a: 9, b: "bar"}
+];
+let params = {data};
```
nicknytko: This is cool, didn’t know this syntax was a thing to create both a key and value!

doc/dev-guide.md
```diff
@@ -458,6 +458,28 @@ sqldb.query(sql.select_questions, params, ...);
...
+sqldb.query(sql.insert_data, params, ...);
+```
+We can insert this into a DB table in a single operation with:
+```sql
+-- BLOCK insert_data
+INSERT INTO my_table (a, b, order_by)
+SELECT *
+FROM
+    ROWS FROM(
```
nicknytko: Is the `ROWS FROM()` needed here since we’re only calling one function?

doc/dev-guide.md
mwest1066: It's needed for the `WITH ORDINALITY`. I split this out as a second bullet point to make it clearer.

doc/dev-guide.md
mwest1066: It's needed for the `WITH ORDINALITY`. I split that out into a separate bullet point for clarity.

mwest1066:
@trombonekenny good catch on the `JSON.stringify()`!

PR #2106, opened 2020-02-29
Added Version Selecting

sean94011:
Allow version sorting for v2 and v3 for the issue: Add v2/v3 type to Questions list #2094. Possible addition to add color to text for displaying purposes.

PR #2151, opened 2020-03-09
Version Selecting with color, issue #2106

kevin88tu:
Updated code from Hackathon

### DIFF
--- a/pages/instructorQuestions/instructorQuestions.ejs
+++ b/pages/instructorQuestions/instructorQuestions.ejs
@@ -3,9 +3,9 @@
 
 <head>
     <%- include('../partials/head'); %>
-        <link href="/stylesheets/theme.bootstrap.css" rel="stylesheet" />
-        <script src="/javascripts/jquery.tablesorter.min.js"></script>
-        <script src="/javascripts/jquery.tablesorter.widgets.js"></script>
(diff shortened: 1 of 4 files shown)


kevin88tu:
Screen Shot 2020-03-12 at 8 02 51 PM
Demonstration of this code. When you try to select a version, you have to select between "v3", "v2(Multiple Choice)", and "v2(Checkbox)". Is this the desired behavior or should it only select between "v2" and "v3"?

pages/instructorQuestions/instructorQuestions.ejs
```diff
@@ -49,6 +49,7 @@
                             <% if (locals.course_instance) { %>
                             <th data-placeholder="Choose assessment">Assessments</th>
                             <% } %>
+                            <th class="filter-select filter-exact" data-placeholder="Choose Version">Version</th>
```
mwest1066:
```suggestion
                            <th class="filter-select filter-exact" data-placeholder="Choose version">Version</th>
```

mwest1066:
> Demonstration of this code. When you try to select a version, you have to select between "v3", "v2(Multiple Choice)", and "v2(Checkbox)". Is this the desired behavior or should it only select between "v2" and "v3"?

This looks great!

mwest1066:
Can we re-order the columns so "Version" comes before "Assessments"?

Issue #890, opened 2017-11-01
make the return / enter key map to "save" by default and not to "grade"


tbretl:
The return key and the enter key map to "grade" by default (e.g., after typing a number into a `pl_number_input` element). This means that if a student presses one of these keys "by accident," then their submission gets graded.

(A wonderful recent version of this is - "the mouse wire pressed enter on the numeric keypad.")

It would be better if the return key and the enter key both mapped to "save" (or nothing) by default.

mwest1066:
Fixed in #2156

PR #2156, opened 2020-03-13
Save on enter, Issue #890

kevin88tu:
During testing, this saved the submission instead of grading for pl-integer-input, pl-number-input, pl-symbolic-input, pl-string-input, pl-matrix-component-input, and pl-checkbox-input. To have the enter key do nothing, remove lines 167-170.

### DIFF
--- a/pages/instructorQuestions/instructorQuestions.ejs
+++ b/pages/instructorQuestions/instructorQuestions.ejs
@@ -3,9 +3,9 @@
 
 <head>
     <%- include('../partials/head'); %>
-        <link href="/stylesheets/theme.bootstrap.css" rel="stylesheet" />
-        <script src="/javascripts/jquery.tablesorter.min.js"></script>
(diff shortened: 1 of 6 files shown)


mwest1066:
I'd prefer to avoid client-side JS to implement this, if possible. Could we instead simply re-order the buttons in the document order? https://stackoverflow.com/questions/1963245/multiple-submit-buttons-on-html-form-designate-one-button-as-default

kevin88tu:
When I tried float: right, it moved to buttons into the middle for some reason, so I used the "float-left" class from bootstrap. When I did that, the buttons were pressed next to each other, so I manually put a margin of 5px between them. I don't know if this was a good idea but it seems to work.

mwest1066:
Looks great, much simpler than the custom JS! 😄 

Rather than explicitly setting CSS rules for margins, you can use the bootstrap spacing utilities: https://getbootstrap.com/docs/4.4/utilities/spacing/

PR #2171, opened 2020-03-18
Add question converters under tools directory

coatless:
This PR takes previously written question converters for various LMS systems into the `tools` directory with extended documentation. 

In particular, there is now:

- randexam (via e-mail)
- loncapa (via repository with kebob case changes)
- smartPhysics (via #2167)

Once merged, the [`question-converters`](https://github.com/PrairieLearn/question-converters) repository can be deleted.


tools/lms-question-converters/README.md
```diff
@@ -0,0 +1,8 @@
+## Convert Learning Management System Questions to PrairieLearn
+
+Within this directory are a set of conversion tools for moving previously written questions
+in a different Learning Management System (LMS) to the question format for PrairieLearn. 
+
+- Craig Zilles: [LON-CAPA](lon-capa/)
+- Lawrence Angrave: [randexam](randexam/)
+- Lucas Wagner: [smartPhysics (FlipIt Physics)](smart-physics/)
```
mwest1066: Let's not associate people's names with these. We don't do it elsewhere in the code. I'm fine to add names to the commit notes or somewhere like that. If we want to get really creative, you could actually make commits with these people as authors for each individual converter.

tools/lms-question-converters/README.md
```diff
@@ -0,0 +1,8 @@
+## Convert Learning Management System Questions to PrairieLearn
```
mwest1066: Maybe just `question_converters` as a directory name? We've been a bit inconsistent with snake_case versus kebab-case, but `node_modules` is probably a good reference.

coatless: Done.

tools/lms-question-converters/README.md
coatless: Done. Switched the committers to Craig and Lawerence.

PR #2176, opened 2020-03-19
Live Markdown preview in `pl-file-editor` element

### LINKED ISSUE #2163: Enable live markdown rendering using ace editor
Have some way to live preview markdown that is input through the `pl-file-editor` element.  Markdown is already being rendered serverside through `unified.js`, this could probably be pulled clientside to provide preview functionality.

nicknytko:
Resolves #2163 

image

Adds question `demoMarkdownEditorLivePreview` to the example course.


elements/pl-file-editor/pl-file-editor.js
```diff
@@ -43,6 +44,35 @@ window.PLFileEditor = function(uuid, options) {
...
+                preview.innerHTML = conv.makeHtml(editor.getValue());
+                if (text.includes('$') ||
+                    text.includes('\\(') || text.includes('\\)') ||
+                    text.includes('\\[') || text.includes('\\]')) {
+                    MathJax.typesetPromise();
+                }
+            }
+        };
+        editor.session.on('change', function() {
+            update_preview();
```
mwest1066: If you put a huge markdown document in the editor and type some more text in it, then the previewing gets very slow.

One way to get around this is to have two flags: one for `preview_in_progress` and one for `preview_needed`. 

On change, if `preview_in_progress == True` then we set `preview_needed = True` and return. If `preview_in_progress == False` then we set `preview_in_progress = True` and call `update_preview()`.

At the end of `update_preview()` we set `preview_in_progress = False` and also check `preview_needed`. If `preview_needed == True` then we set `preview_in_progress = True, preview_needed = False` and we call `update_preview()` again.

elements/pl-file-editor/pl-file-editor.js
```diff
@@ -43,6 +44,35 @@ window.PLFileEditor = function(uuid, options) {
         this.editor.setOption('maxLines', Infinity);
     }
 
+    const default_preview_text = '<p>Begin typing to preview</p>';
```
mwest1066:
```suggestion
    const default_preview_text = '<p>Begin typing above to preview</p>';
```

mwest1066:
News item:

[PrairieLearn.pdf](https://github.com/PrairieLearn/PrairieLearn/files/4356097/PrairieLearn.pdf)

news_items/007_markdown_live_preview/index.html
```diff
@@ -0,0 +1,38 @@
...
+  set <tt>preview="markdown"</tt> like this:
+</p>
+
+<div class="d-flex justify-content-center">
+  <div class="card mb-4 style="width: 50%;"">
+    <div class="card-header">
+      <tt>question.html</tt>
+    </div>
+    <div class="card-body">
+      <pre class="mb-0">&lt;pl-file-editor file-name="ans.md" ace-mode="ace/mode/markdown" preview="markdown"&gt;
```
coatless: Can you add a highlighted background here?

news_items/007_markdown_live_preview/index.html
```diff
@@ -0,0 +1,38 @@
...
+  </div>
+</div>
+
+<p>
+  When students type their answers in the text editor, their file will
+  be saved as Markdown for grading and they will see a live-updating
+  preview:
+</p>
+
+<img src="markdown_preview.png" width="833px">
```
coatless: Add that this is useful for manually graded assignments? Link out to documentation?

mwest1066: Great idea! Done.

news_items/007_markdown_live_preview/index.html
mwest1066: Done

PR #2185, opened 2020-03-20
Moving multiple instances of assessments into their own subpage

### LINKED ISSUE #2068: Aggregate multiple assessment instance grading
If a student takes a practice exam in multiple-instance mode, provide a gradebook view to the students that provides an overall version of performance across the instances (on whatever metric is being scored). For example, we provide "best overall" CSV downloads for the instructor to assess all questions across all instances, but we don't have a representation for that visible to the student.

From Lane Schwartz in the PL workshop 7 Feb 2020

### LINKED ISSUE #2086: Make multiple-instance assessments collapsible
Hi! I'm a student at UIUC using PrairieLearn and I rarely make issues on GitHub, so I hope you'll forgive me if I mess this up. I think multiple-instance assessments should be collapsible, so you can hide all your previous instances if you'd like. Maybe after the due date, it could be collapsed by default. This is sort of related to #2068, it could also show you your aggregate score
For example, this is from my CS 357 page, it got pretty long pretty quickly:
image
I propose that each of those numbered instances could be hidden underneath the main entry that has the "New instance" button.

astrosticks:
Fixes #2086 by moving links to instances of assessments that allow multiple instances into a student view that was previously only used to generate a new instance. @nikitaa4 and I worked on this at HackIllinois 2020. Screenshots showing the changes are below:

### pages/studentAssessments
Previously | Now*
-------|------
pl_instancesCondensed_before_assessments | pl_instancesCondensed_after_assessments

*~~_Currently shows the maximum of all of the score percentages, but Dave's wording in #2068 seems to imply that this might not always be the final score that the student receives._~~

### pages/studentAssessmentExam
Previously | Now*
-------|------
pl_instancesCondensed_before_instances | pl_instancesCondensed_after_instances

*~~_May want to switch order so that the most recent attempt is at the top of the list? This would keep the most relevant information near the top of the page._~~

Still to-do (based on conversation from Jun 18): 
- [x] Display the student's maximum score among all multiple instances in student gradebook (closes #2068)
    - [x] Move individual multipleInstance instances under expanded region in student gradebook (collapsed by default)


pages/studentAssessmentExam/studentAssessmentExam.ejs
```diff
@@ -35,7 +36,13 @@
             <input type="hidden" name="__action" value="newInstance">
             <input type="hidden" name="__csrf_token" value="<%= __csrf_token %>">
             <div class="d-flex justify-content-center">
-              <button id="start-assessment" type="submit" class="btn btn-primary">Start assessment</button>
+              <button id="start-assessment" type="submit" class="btn btn-primary">
+                <% if (assessment.multiple_instance) { %>
+                  New instance 
+                <% } else { %>
+                  Start Assessment
```
nicknytko: Looks like the structure of this was changed slightly, the corresponding test case here should be updated to match: https://github.com/PrairieLearn/PrairieLearn/blob/master/tests/testShowClosedAssessment.js#L54.

This is *probably* as simple as just changing that line to:

```javascript
assert.equal(response.$('#start-assessment').text().trim(), 'Start Assessment');
```

astrosticks: @nicknytko done! Sorry about the messy commit history :S

nicknytko (dismissed):
Looks okay, I will defer review of the SQL stuff to @mwest1066 since I'm not sure what he's wanting in terms of code style here.

mfsilva22:
This looks great and will be very helpful to students. 
I like the order as it is. At least I think of order of assessments going from old to new.
I am not sure about how the point system works. I would think displaying the max score seems the right way to go.

mwest1066:
Can you add a skeleton news item to this PR? With some hires screenshots of before and after? I'll write the text of the news item, but having some images and an outline of the main points would be very helpful.

nwalters512:
Discussed with the dev team: while we still like the concept, this has drifted sufficiently far from `master` that we think it'd be easiest to reimplement this from scratch when we're ready for this.

PR #2186, opened 2020-03-20
Fix buggy float behavior via flexbox

tdy:
pl-question-footer-float-flex


mwest1066:
I love the image in the PR! Makes it super easy to understand what's changed.

PR #2187, opened 2020-03-20
Enable disable-on-submit JS for v3 (#1812)

### LINKED ISSUE #1812: Disable-on-submit/click button classes are not used for v3 questions
We add classes `disable-on-submit` and `disable-on-click` to buttons in the question footer:

https://github.com/PrairieLearn/PrairieLearn/blob/master/pages/partials/questionFooter.ejs#L19

But the JS that uses these classes is only added for non-Freeform (that is, old-style v2) questions:

https://github.com/PrairieLearn/PrairieLearn/blob/master/pages/partials/questionFooter.ejs#L84

It looks like this might have been an error in the original implementation from Nov 2017, but since we haven't heard any complaints since then it's not clear that it's worth fixing?

tdy:
Copies the disable-on-submit `script` from v2 to v3 to fix #1812. Originally I could double-submit and now I can't (the disabling is also visibly happening "immediately").


pages/partials/questionFooter.ejs
```diff
@@ -35,6 +35,29 @@
...
+              });
+          } else {
+              e.preventDefault();
+          }
+      });
+
+      $('a.disable-on-click').on('click', function() {
+          $(this).addClass('disabled');
+      });
+  </script>
```
mwest1066: Rather than duplicating the `<script>` for both the `Freeform` and non-`Freeform` question types, could we shift it outside the entire footer `<div>` so that it's included for both question types?

tdy: Yes! Actually I was kinda wondering why there was so much duplication already, but now I'm guessing it's to make the v2 deprecation easier... just wipe out the `else`? Anyway I will update the branch.

tdy: Moved `<script>` outside. Squashed with old commit.

pages/partials/questionFooter.ejs
```diff
@@ -74,30 +74,28 @@
       </div>
       <% } %>
     </form>
+  </div>
+<% } %>
+<script>
+    $('form').on('submit', function(e) {
```
mwest1066: I'm a bit worried by the fact that this will also find other forms on the page (like the one for submitting issues). It shouldn't cause any damage, but I don't like having random JS triggering if we can avoid it.

Perhaps we could add a class like `main-question-form` to the question form and change the line above to only use `$('form.main-question-form')` or similar?

mwest1066:
Minor point, but when we merge we always do a full squash, so it's not necessary to worry about commit cleanliness or squashing in PRs.

pages/partials/questionFooter.ejs
tdy: I was wondering about this also for the "alert for unsaved" issue. It would be helpful if we have a unique identifier for question forms. Should we add an `id` instead? Or you prefer using a class for this.

nwalters512: The main form already has `class="question-form"`, seems like we could use that here?

tdy: Yeah that's what I'm currently using in my other PR (`.question-form`). Just wasn't sure since conceptually I've always treated `class` for styling and `id` for unique id'ing elements.

tdy: Went ahead and updated to use `form.question-form`

echuber2:
@tdy @mwest1066 If the issue is double-clicking on "submit", what about having a cooldown timer on it, or having it diff with the previous submission and ask the user to confirm in a modal popup if it appears they're resubmitting the same data?

PR #2192, opened 2020-03-20
Add display of R vector/matrix object in pl-variable-output

### LINKED ISSUE #1460: R matrix representation
Once #1453 is merged. 

The _R_ implementation for exporting a matrix would require the modification of [`string_from_2darray()`](https://github.com/PrairieLearn/PrairieLearn/blob/eca00c02848560d1aff19f1c5f4ae7311c833a3a/question-servers/freeformPythonLib/prairielearn.py#L326-L379) to create a string in the form of:

```
X = matrix(c(data1, data2), nrow = , ncol = )) # matrix
```

## Extension possibilities

Note: `array2string` print prefix/suffix doesn't override `[]`.
c.f. https://github.com/numpy/numpy/issues/11142

### Joining

```py
import numpy as np

# Sample data
x = np.array([[1, 2, 3], [4, 5, 6]], np.int32)

#  Dump to comma separated list
y_star = ','.join(','.join('%0.3f' %a for a in y) for y in x)
                                  # ^^^ formatting.

# Convert to R format
r_variant = 'matrix(c(' + y_star + '))'

# Display variant
print(r_variant)
# matrix(c(1.000,2.000,3.000,4.000,5.000,6.000))
```

### Regular expression route

```py
import numpy as np
import re

# Sample data
x = np.array([[1, 2, 3], [4, 5, 6]], np.int32)

# Array conversion
y = np.array2string(x, precision=2, separator=',')

# Remove brackets with regular expressions
y_star = re.sub('[\[\]]', '', y) 

# Modify with R syntax
r_variant = "matrix(c(" + y_star + "))")

# Display
print(r_variant)
# matrix(c(1,2,3, 4,5,6))
```

### Misc

Subset index trick, e.g. `y[1:-1]` would breaks for a 2D array... 

```py
# matrix(c([1,2,3],
# [4,5,6]))
 #^        ^ :'(
```

coatless:
Adds a tab to the `pl-variable-output` element that shows pre-formatted _R_ code in scalar/vector/matrix form. Close #1460

Screen Shot 2020-03-20 at 2 04 20 PM

Note: `byrow = TRUE` is required because _R_ stores data in column-format whereas _NumPy_ stores it in row-format.

### DIFF
--- a/doc/elements.md
+++ b/doc/elements.md
@@ -991,7 +991,7 @@ If `file()` does not return anything, it will be treated as if `file()` returned
 ## `pl-variable-output` element
 
 Displays a list of variables that are formatted for import into the
-supported programming languages (e.g. MATLAB, Mathematica, or Python).
+supported programming languages (e.g. MATLAB, Mathematica, Python, or R).
 
 #### Sample Element
 
@@ -1034,6 +1034,7 @@ Attribute | Type | Default | Description
 `show-matlab` | boolean | true | Toggles the display of the Matlab tab.
 `show-mathematica` | boolean | true | Toggles the display of the Mathematica tab.
 `show-python` | boolean | true | Toggles the display of the Python tab.
+`show-r` | boolean | true | Toggles the display of the R tab.
 
 Attributes for `<variable>` (one of these for each variable to display):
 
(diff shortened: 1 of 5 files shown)


coatless:
@daviddalpiaz ^^^

PR #2198, opened 2020-03-21
Enable instructor panel on course question pages (#2189)

### LINKED ISSUE #2189: Instructor information panel is missing on course Question pages
Steps to reproduce:

1. From the PL homepage, click on a course (not a course instance).

2. Click on "Questions".

3. Click a question from the list.

4. The question preview page displays the question correctly, but there is no yellow "Instructor information" panel on the right hand side.

5. For comparison, select a Course instance from the navbar dropdown and navigate back to the same question. Now there will be a yellow "Instructor information" panel.

tdy:
From Matt:
> The problem is that `has_instructor_view` is only associated with course *instances* (because "Instructor" is a course-instance-level role, not a course role). See [authzCourseInstance.js#L47](https://github.com/PrairieLearn/PrairieLearn/blob/master/middlewares/authzCourseInstance.js#L47) to see where `has_instructor_view` is set from the course instance permissions.
> 
> The problem with the current bug is that we don't have an active course *instance*, only an active course. The corresponding *course* permission is `has_course_permission_view`.

Close #2189:
* Add `authz_data.has_course_permission_view` to the instructor panel check
* Fix the QID URL to check for `locals.course_instance`
* Unchanged: "show answer" and "report an issue" work as is


tdy:
Or should we make it switch to Dev User (Instructor) for viewing `course` questions? (Don't know how to do this one.)

pages/partials/instructorInfoPanel.ejs
```diff
@@ -1,4 +1,4 @@
-<% if (authz_data.has_instructor_view) { %>
+<% if (authz_data.has_instructor_view || locals.is_administrator) { %>
```
mwest1066: I'm a bit perplexed about why an administrator user doesn't automatically have instructor view?

tdy: So that's the real bug? Where to start debugging that?

mwest1066: Ah, ok, now I understand more.

The problem is that `has_instructor_view` is only associated with course _instances_ (because "Instructor" is a course-instance-level role, not a course role). See https://github.com/PrairieLearn/PrairieLearn/blob/master/middlewares/authzCourseInstance.js#L47 to see where `has_instructor_view` is set from the course instance permissions.

The problem with the current bug is that we don't have an active course _instance_, only an active course. The corresponding _course_ permission is `has_course_permission_view`.

So I think the correct fix is to use `authz_data.has_instructor_view || authz_data.has_course_permission_view` above.

This complicated mess is something that @tbretl's work on the new permissions system should clean up, but for now we need to survive with the janky course/course-instance distinction.

tdy: Wait... just realized my admin check only makes sense when running locally 🤦‍♂

PR #2202, opened 2020-03-21
More template/tutorial questions for the exampleCourse

mfsilva22:
Adding new template questions. I am also considering adding questions that are following a tutorial format. I added one for rounding with this format.

A new organization of the assessments may be needed in the exampleCourse


exampleCourse/questions/templateRandomMultipleChoiceWithFigure/server.py
```diff
@@ -0,0 +1,22 @@
...
+        data['params']['filename'+str(i+1)] = fig
+
+    #Create a list with the file name of all non-UIUC figures
+    allnonUIUCfigures = ["b"+str(i)+".jpeg" for i in range(1,5)]
+
+    #Make a random selection for the non-UIUC figure
+    falseFig = random.choice(allnonUIUCfigures)
+
+    #Store the false figure in data['params']
+    data['params']['filename'] = falseFig
```
mwest1066: I found the naming confusing here, because the `falseFig` is the one that's the _correct_ answer. Could we name the figures something like `data['params']['correct_filename']` and `data['params']['wrong_filename1']`?

Where `correct_filename` is the non-UIUC fig.

exampleCourse/questions/templateRandomMultipleChoiceWithFigure/server.py
```diff
@@ -0,0 +1,22 @@
...
+
+    #Create a list with the file name of all UIUC figures
+    allUIUCfigures = ["fig"+str(i)+".jpeg" for i in range(1,8)]
+
+    #Make a random selection of 3 figures from UIUC
+    trueFig = random.sample(allUIUCfigures,3)
+
+    #Store the true figures in data['params']
+    for i,fig in enumerate(trueFig):
+        data['params']['filename'+str(i+1)] = fig
```
mwest1066:
```suggestion
    #Create a list with the file name of all UIUC figures
    UIUCfigures = ["fig1.jpeg", "fig2.jpeg", "fig3.jpeg", "fig4.jpeg", "fig5.jpeg", "fig6.jpeg", "fig7.jpeg", "fig8.jpeg"]

    # Shuffle the order of the UIUC figures
    random.shuffle(UIUCfigures)

    # Store the first three shuffled UIUC figures as the wrong answers
    data['params']['wrong_filename0'] = UIUCfigures[0]
    data['params']['wrong_filename1'] = UIUCfigures[1]
    data['params']['wrong_filename2'] = UIUCfigures[2]
```

I'm thinking that something like this ^^^ might be easier for non-Python-experts to follow?

exampleCourse/questions/templateRandomNumberInput/question.html
```diff
@@ -0,0 +1,24 @@
+<pl-question-panel>
+  <p>The comparison between the correct and submitted answers can be done using
```
mwest1066: Do we want to add some explanation about when you should use which method? What do you normally use?

exampleCourse/questions/templateRandomMultipleChoiceWithFigure/server.py
mfsilva22: haha. This is actually funny! I never thought I would one day get that type of advice (the non_Python expert here speaking). Yes, sure, we can change that.

exampleCourse/questions/templateRandomNumberInput/question.html
mfsilva22: I think that is so personal! I avoid using sigfig. I think many students find it confusing. They think you are asking them to give that exact number of sigfigs, and if they don't know how to do it correctly, they get it marked wrong for something silly. I think tolerance is clearer to students.

mwest1066: Ok, sounds good, let's go with it how it is.

exampleCourse/questions/templateRandomIntegerInput/info.json
```diff
@@ -0,0 +1,7 @@
+{
+    "uuid": "5E64666F-1588-4EBB-8E8D-C1FA4141F20D",
+    "title": "Template pl-integer-input: randomized input parameters",
+    "topic": "Template",
+    "tags": ["mfsilva", "Sp20", "tpl101"],
```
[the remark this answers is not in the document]
mfsilva22: should I change all the other tags, including the questions created by @coatless

exampleCourse/questions/templateRandomIntegerInput/info.json
```diff
@@ -0,0 +1,7 @@
+{
+    "uuid": "5E64666F-1588-4EBB-8E8D-C1FA4141F20D",
```
nicknytko: this uuid is conflicting with `templateIntegerInput` and that is why travis is not happy :slightly_smiling_face:

mfsilva22: Interesting.... I thought `templateRandomIntegerInput` was deleted (or actually renamed as  `templateIntegerInput`.

exampleCourse/questions/templateIntegerInputWithDynamicFigure/question.html
```diff
@@ -0,0 +1,35 @@
+<pl-question-panel>
+
+<p>What is the size of the {{params.dname}}?</p>
+
+<!-- This element will generate the f -->
```
mwest1066:
```suggestion
<!-- This element will generate the figure -->
```

exampleCourse/questions/templateNumberInputTips/question.html
```diff
@@ -0,0 +1,78 @@
+<pl-question-panel>
+
+<p>Suppose you define a random number in <code>server.py</code> using the following:</p>
+
+<pl-code language=python>a = random.uniform(1,10)</pl-code>
+
+<p> and you want display the value of <code>a</code> in your question panel. We often
+    prefer to display a rounded value of <code>a</code> , such that instead of providing </p>
```
mwest1066:
```suggestion
    prefer to display a rounded value of <code>a</code>, such that instead of providing </p>
```

Issue #1438, opened 2019-02-26
Prototype: issue status visibility


trombonekenny:
Right now when a student submits an issue, they don't get any acknowledgement or visibility into its next steps. Karrie suggested we would gain more student trust in these issues if they had a page that could show their open issues and their status (un-opened, opened, resolved with message, etc.)

Perhaps a new global page, or another tab in the course instance for student roles?

coatless:
👍 but avoid showing the _contents_ of the issue message. Perhaps: Assessment / Date

echuber2:
Is this related (not coincidentally, but actually) to the list at #1483 ?

mwest1066:
This would all be awesome. We should show the student messages if they have currently have access to the linked variant/assessment. This would allow them to see full details for their issues when they are back in the CBTF for a review session, for example, but they would only be able to see the status when they are at home (for exam issues, that is).

PR #2203, opened 2020-03-21
Displaying issue status to student open/closed v2

dkim221:


pages/partials/question.ejs
```diff
@@ -38,6 +38,17 @@
           Show/hide
         </button>
       </p>
+      <p><strong> Issue status: </strong>
+        <% if (issue.open) { %>
+        <span class="badge badge-danger" data-toggle="collapse" href="#">
```
mwest1066: I don't think the `data-toggle` and `href` attributes are doing anything here?

mwest1066: Can we use the same icon styling as the issue badges elsewhere? https://github.com/PrairieLearn/PrairieLearn/blob/master/pages/instructorIssues/instructorIssues.ejs#L168

That is, add the `fa-exclamation-circle` and `fa-check-circle` icons.

dkim221:
I just pushed a new commit using the same icon styling as issue badges elsewhere.

mwest1066:
It looks like you copied too much when you copied the styling. This code is now failing CI tests.

mwest1066:
Closing this PR for now. See #1438

PR #2205, opened 2020-03-22
added css overflow-wrap: normal for pre tag Issue #1661

### LINKED ISSUE #1661: elements with <pre> tags need overflow-wrap:normal
As @nwalters512 discovered in #1660, Safari needs `<pre>` tags to have `overflow-wrap: normal` set to make them not wrap their contents. This has been done for `<pl-external-grader-results>`, but is still needed for other elements that use `<pre>`.

kevin88tu:
Before:
Screen Shot 2020-03-21 at 10 22 58 PM
After:
Screen Shot 2020-03-21 at 10 23 12 PM
Note: in the pl-code element, the code is not written in the \<pre> tag but in \<span> tags, and when I tested the code would still wrap in Safari but would not in chrome. Should I add the span class to the css?

Resolves #1661

### DIFF
--- a/public/stylesheets/local.css
+++ b/public/stylesheets/local.css
@@ -109,6 +109,10 @@ table.two-column-description-no-header-with-help tr td:last-of-type {
     white-space: nowrap;
 }
 
+pre {
+    overflow-wrap: normal;
+}
+
 /**************
  * Submissions
  **************/
@@ -202,12 +206,12 @@ table td > .badge {
     white-space: pre;
     color: #444;
     background: #F0F0F0;
-    
+
     padding-top: 4px;
     padding-bottom: 4px;
     padding-left: 2px;
     padding-right: 2px;
-    
+
     border-radius: 2px;
     border-color: #ccc;
     border-width: 1px;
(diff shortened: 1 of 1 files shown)


nicknytko:
Resolves #1661

public/stylesheets/local.css
```diff
@@ -202,12 +206,12 @@ table td > .badge {
     white-space: pre;
     color: #444;
     background: #F0F0F0;
-    
+
```
nicknytko: are these whitespace changes necessary?

public/stylesheets/local.css
```diff
@@ -202,12 +206,12 @@ table td > .badge {
...
     padding-top: 4px;
     padding-bottom: 4px;
     padding-left: 2px;
     padding-right: 2px;
-    
+
```
nicknytko: Are these whitespace changes necessary?

kevin88tu: I think my editor made those changes without me realizing since I don't remember removing any whitespace. Should I fix them?

nwalters512: I'm a fan of empty lines not having whitespace, so this change is fine by me!

nicknytko: if possible, yes

nicknytko: I guess either is fine then 😆

PR #2206, opened 2020-03-22
Allow invalid attempts in externally graded questions

### LINKED ISSUE #2160: Allow externally graded question to return status of not gradable
Consider the following case:

- A student is working on a coding question (possibly in an exam setting)
- The student submits their code
- The code fails to compile

In this setting, there should be a mechanism whereby the question is not graded (in the same way that a blank submission is not graded).

### LINKED ISSUE #2195: Grading errors during `grade()` still use up attempts
submission.gradable gets updated in `lib/question.js:465`https://github.com/PrairieLearn/PrairieLearn/blob/master/lib/question.js#L465, but this does not get checked later.  Should probably update `sprocs/grading_jobs_insert_internal.sql` to check if `gradable` is false.

nicknytko:
Adds the `gradable` attribute to external grading `results.json`, which when set to `False` marks the submission as invalid.  This can be used to not penalize students for syntax errors.

Screen Shot 2020-03-22 at 3 38 14 PM

Also, when `succeeded = False` the submission will be marked invalid as well and will auto-generate a course issue:

Screen Shot 2020-03-22 at 3 38 48 PM

Resolves #2160 

Screen Shot 2020-03-22 at 4 06 00 PM

Also fixes #2195, which occurs when format errors are set during `grade()`.


sprocs/grading_jobs_insert_internal.sql
```diff
@@ -53,9 +54,16 @@ BEGIN
 
     new_correct = (new_score >= 1.0);
 
+    -- Only update the graded_at time if the submission is gradable
```
nwalters512: Are these changes meant to fix existing behavior or are they somehow related to this PR? A little confused since this is the sproc that handles internal grading jobs.

sprocs/grading_jobs_process_external.sql
```diff
@@ -64,21 +66,45 @@ BEGIN
...
-    -- ######################################################################
-    -- update all parent objects
+        IF assessment_instance_id IS NOT NULL THEN
+            UPDATE instance_questions
+            SET status = 'saved'::enum_instance_question_status
+            WHERE id = instance_question_id;
+        END IF;
+
+        IF succeeded = FALSE THEN
+            -- Automatically add an issue if the grading job was unsuccessful
```
nwalters512: Ahh, I like this! Nice improvement.

elements/pl-external-grader-results/pl-external-grader-results.mustache
```diff
@@ -11,15 +11,20 @@
...
-      <span style="color: {{results_color}}">
-      {{#points}}
-      <strong>{{points}}/{{max_points}} ({{score}}%)</strong>
-      {{/points}}
-      {{^points}}
-      <strong>{{score}}%</strong>
-      {{/points}}
-      </span>
+      {{#invalid}}
+          <p> Invalid, not graded </p>
```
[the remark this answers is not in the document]
mwest1066: I think this is for consistency with the badge that we put on internally-graded submissions are are not gradable (the badge says "Invalid, not graded"). For example, someone types a word into a `pl-number-input`.

sprocs/grading_jobs_insert_internal.sql
nicknytko: Sorry, this is for #2195 , I’ve lumped both in here because the changes are similar.

nwalters512: Ahh, makes sense, just wanted to double-check.

lib/assessment.js
```diff
@@ -297,13 +297,34 @@ module.exports = {
                 if (_(content.grading).has('feedback') && !_(content.grading.feedback).isObject()) {
                     return callback(error.makeWithData('invalid grading.feedback', {content: content}));
                 }
+
+                let succeeded = true;
```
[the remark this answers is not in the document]
nicknytko: Gotcha, if this is mostly bad code then I’m okay not catching it.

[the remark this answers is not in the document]
nwalters512: @mwest1066 good catch - I forgot we had two notions of `succeeded` (I think that was my fault haha).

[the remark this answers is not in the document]
mwest1066: I think we basically have three cases:
1. Grading failed: we have no idea what went wrong, but we are going to assume it was student error, give them `score = 0`, and count it as an attempt. This is treated exactly as if grading in fact succeeded with `score = 0`.
2. Grading succeeded, `gradable = false`: this is the new case, where we don't want to count it as an attempt.
3. Grading succeeded, `gradable = true`: proceed as normal.

sprocs/grading_jobs_process_external.sql
```diff
@@ -64,21 +65,35 @@ BEGIN
...
+        SET
+            gradable = FALSE,
+            feedback = grading_jobs_process_external.feedback
+        WHERE id = grading_job.submission_id;
 
-    -- ######################################################################
-    -- update all parent objects
+        IF assessment_instance_id IS NOT NULL THEN
+            UPDATE instance_questions
+            SET status = 'saved'::enum_instance_question_status
```
mwest1066: Same here. I'm worried that `saved` is not really what we want?

Should we add `invalid` as a new value for `enum_instance_question_status`?

nicknytko: Yes, I’m all for adding an `invalid` value.  Setting it to `saved` is what I observed the current behaviour to be for invalid questions, but I can see how it would be misleading.

lib/assessment.js
```diff
@@ -297,13 +297,32 @@ module.exports = {
...
+                        succeeded = results.succeeded;
+                    }
+                    if (!succeeded) {
+                        content.grading.score = 0;
+                    }
+                    if (succeeded && _(results).has('gradable')) {
+                        gradable = results.gradable;
+                    }
+                }
+
```
mwest1066:
```suggestion
                const succeeded = !!_.get(content, 'grading.feedback.results.succeeded', true);
                const gradable = !!_.get(content, 'grading.feedback.results.succeeded', true);
                if (!succeeded) {
                    content.grading.score = 0;
                }
```

lib/question.js
```diff
@@ -1406,9 +1411,10 @@ module.exports = {
                     // Render the submission panel
 
                     // If the submission wasn't graded yet, return null to indicate that
-                    // the panel shouldn't be rerendered.
-                    if (submission.graded_at === null) {
-                        callback(null);
+                    // the panel shouldn't be rerendered.  Check if 'feedback' is set because
+                    // that will indicate the grading code was run (even if it is ungradable)
```
mwest1066: This feels a bit too sneaky to me. Should we be adding a new property like `grading_succeeded_at` that's separate to `graded_at`, so `graded_at` can mean "ran the grading code"?

Or we could add a new property like `grading_code_run_at` to indicate that we ran the grading code, even if `graded_at = NULL`.

But maybe I'm misunderstanding what's going on here. Adding a whole other DB column just for one check seems pretty heavy-weight.

nicknytko: The question is, is this check even needed?  This line used to be `callback(null)` (it did not `return`), so it would still render anyway if it got to this point.  I'll experiment with dropping the check and seeing if anything breaks.

sprocs/grading_jobs_process_external.sql
nicknytko: image

The red "invalid" badge looks awfully close to the "unanswered" badge imo.  What if we changed those to a different, happier colour?

image

image

lib/question.js
nicknytko: Removed, seems to be unnecessary.

sprocs/grading_jobs_process_external.sql
mwest1066: I think the idea here was that "invalid" and "unanswered" are functionally identical. Both mean that we don't have an answer that can be graded (which is also whey they are red). Does that argument make sense?

nicknytko: I like blue, but I suppose that makes sense.

sprocs/grading_jobs_insert_internal.sql
```diff
@@ -69,6 +77,18 @@ BEGIN
     WHERE
         s.id = submission_id;
 
+    IF new_gradable = FALSE THEN
+        -- If the submission is ungradable then we should stop here
```
mwest1066: I would like to still insert a `grading_job` even in the case that the submission was ungradable. That would preserve the idea that `grading_jobs` correspond to attempts to grade for both internal and external questions. I guess we should add a new `gradable` column to `grading_jobs` to record whether the grading job returned `gradable=false`.

sprocs/grading_jobs_insert_internal.sql
```diff
@@ -69,6 +77,18 @@ BEGIN
...
+            UPDATE instance_questions
+            SET status = 'invalid'::enum_instance_question_status
+            WHERE id = instance_question_id;
+        END IF;
+
+        RETURN;
+    END IF;
+
     -- ######################################################################
     -- update the variant
```
mwest1066: Should we also run the `variant` update in the case that `gradable=false`? I guess so.

The `grading_job` that we insert below seems to have `graded_at=NULL`? But in the assessment log we explicitly try and find grading jobs with `graded_at IS NOT NULL` for internal grading jobs: https://github.com/PrairieLearn/PrairieLearn/blob/master/pages/instructorAssessmentInstance/instructorAssessmentInstance.sql#L248

I think these entries are just not showing up in the log, which is actually a bug. To fix this, we should insert the new grading job with `graded_at=now()`. I think we should do this even in the case that `gradable=false`.

lib/question.js
```diff
@@ -505,6 +505,11 @@ module.exports = {
                 ];
                 sqldb.callWithClientOneRow(client, 'grading_jobs_insert', params, (err, result) => {
                     if (ERR(err, callback)) return;
+
+                    /* If the submission was marked invalid during grading the sproc will return a null record,
+                       and we should bail here to prevent LTI updates. */
+                    if (result.rows[0].id === null) return callback(new NoSubmissionError());
```
mwest1066: If we add `grading_jobs.gradable` then we can use that here?

doc/externalGrading.md
```diff
@@ -167,11 +167,14 @@ Your grading process must write its results to `/grade/results/results.json`. Th
 
 As long as those two fields are present, you may add any additional data to that object that you want. This could include information like detailed test results, stdout/stderr, compiler errors, rendered plots, and so on.
 
+The optional boolean `gradable` can be added to the results object and indicates that the input was invalid or formatted incorrectly.  For example, `gradable` can be set to `false` if the student code has a syntax error, in which case the code will not be graded and the student will not be penalised an attempt.  The omission of this field is equivalent to assuming that the input was gradable (`"gradable": true`).
+
 The `<pl-external-grader-results>` element is capable of rendering a list of tests with associated test names, descriptions, point values, output, and messages. Here's an example of well-formed results that can be rendered by this element:
 
 ```json
 {
    "succeeded": true,
```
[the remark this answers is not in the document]
nicknytko: Yep, since it's basically unused anyway we can remove it.

nicknytko:
@mwest1066 ping!

docs/externalGrading.md
```diff
@@ -163,15 +163,17 @@ The following is an example of a well-structured course layout:
 
 ## Grading results
 
-Your grading process must write its results to `/grade/results/results.json`. The result only has 2 mandatory fields: `succeeded` and `score`. `succeeded` indicates if the tests were able to run successfully, or if they failed due to some error. `score` is the score, and should be a floating point number in the range [0.0, 1.0].
+Your grading process must write its results to `/grade/results/results.json`. The result only has one mandatory field: `score`, which is the score for the submitted attempt, and should be a floating point number in the range [0.0, 1.0].
```
mwest1066, line 166: If `gradable = false` then should we make the `score` optional? It seems weird to be forced to set a score when it's going to be ignored.

docs/externalGrading.md
```diff
@@ -163,15 +163,17 @@ The following is an example of a well-structured course layout:
...
 
-As long as those two fields are present, you may add any additional data to that object that you want. This could include information like detailed test results, stdout/stderr, compiler errors, rendered plots, and so on.
+As long as this field is presetn you may add any additional data to that object that you want. This could include information like detailed test results, stdout/stderr, compiler errors, rendered plots, and so on.
```
mwest1066, line 168:
```suggestion
As long as this field is present you may add any additional data to that object that you want. This could include information like detailed test results, stdout/stderr, compiler errors, rendered plots, and so on.
```

docs/externalGrading.md
```diff
@@ -163,15 +163,17 @@ The following is an example of a well-structured course layout:
...
+
+The optional boolean `gradable` can be added to the results object and indicates that the input was invalid or formatted incorrectly.  For example, `gradable` can be set to `false` if the student code has a syntax error, in which case the code will not be graded and the student will not be penalized an attempt.  The omission of this field is equivalent to assuming that the input was gradable (`"gradable": true`).
```
mwest1066, line 170:
```suggestion
The optional boolean `gradable` can be added to the results object and set to `false` to indicate that the input was invalid or formatted incorrectly, for example if it has a syntax error that prevented compilation.  If `"gradable": false` is set then the submission will be marked as "invalid, not gradable", no points will be awarded or lost, and the student will not be penalized an attempt on the question.  The omission of this field is equivalent to assuming that the input was gradable (`"gradable": true`).
```

migrations/169_grading_jobs__add_gradable_attribute.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE grading_jobs ADD COLUMN gradable BOOLEAN;
+UPDATE grading_jobs SET gradable=true;
```
mwest1066, line 2: Do we need to set old grading jobs to have `gradable=true`? Maybe leaving them with `NULL` is clearer, because we don't actually know their status?

sprocs/grading_jobs_process_external.sql
```diff
@@ -57,28 +58,43 @@ BEGIN
...
-    -- ######################################################################
-    -- update all parent objects
+        IF assessment_instance_id IS NOT NULL THEN
+            UPDATE instance_questions
+            SET status = 'invalid'::enum_instance_question_status
+            WHERE id = instance_question_id;
+        END IF;
+    ELSE
+        UPDATE submissions
+        SET
```
mwest1066, line 83: Should we set `gradable = TRUE` here?

docs/externalGrading.md
```diff
@@ -163,15 +163,17 @@ The following is an example of a well-structured course layout:
 
 ## Grading results
```
mwest1066, line 164: Should we also add documentation to `server.py` and/or element-writing docs to say that `grade()` can set `gradable = false`?

sprocs/grading_jobs_process_external.sql
nicknytko: Yep, if it is not already gradable here then this code won't execute.

docs/externalGrading.md
nicknytko: Since we're no longer parsing, should we still be setting `data['format_errors']`?  Should we have a new key like `data['grading_errors']`?

sprocs/grading_jobs_process_external.sql
mwest1066: Excellent.

docs/externalGrading.md
mwest1066: I'm not in love with the name `format_errors`, but I'm tentatively inclined to stick with only having one set of errors that correspond to `gradable = false`. My thinking is that it's sort of an implementation detail in the question about whether the "invalid" status of the submission is detected during `parse()` or `grade()`. We try and detect it during `parse()`, because that's friendlier for the student, but in any case we will treat the outcome the same (that is, `gradable = false` and `format_errors = something`). This has the advantage that we don't need to add separate code paths to handle `grading_errors`.

Thinking about it more, I could even talk myself into liking the name `format_errors`. Certainly for code that doesn't compile, that sounds a lot like a format error of some kind 😄

What do you think? Or is there some advantage to separating `grading_errors` from `format_errors`?

nicknytko: I don't think there's an advantage to having it separate &mdash; I just don't know how I feel about the name either.  I suppose it's okay to keep it as-is, just sorta thinking out loud here.

sprocs/grading_jobs_process_external.sql
```diff
@@ -57,28 +58,43 @@ BEGIN
...
-        graded_at = grading_job.graded_at,
-        score = grading_job.score,
-        correct = grading_job.correct,
-        feedback = grading_job.feedback
-    WHERE id = grading_job.submission_id;
+    IF new_gradable = FALSE THEN
+        UPDATE submissions
+        SET
+            gradable = FALSE,
+            feedback = grading_jobs_process_external.feedback
```
mwest1066, line 73: Should we be setting `format_errors` rather than `feedback` here?

I'm thinking that the rule is that "valid" submissions must have `format_errors = {}` and may have non-empty `feedback`, while "invalid" submissions must have `feedback = {}` and may have non-empty `format_errors`.

Does this seem reasonable? It would fit with the big flowchart at the bottom of https://prairielearn.readthedocs.io/en/latest/dev-guide/

Perhaps we could update that flowchart by adding `submission.format_errors = <value>` to the "Ungradable submission" ellipse?

Or are we setting `feedback` here because external graders are putting compile errors into feedback?

sprocs/grading_jobs_insert_internal.sql
```diff
@@ -83,20 +83,32 @@ BEGIN
...
+        new_partial_scores, authn_user_id, grading_method, new_gradable)
     RETURNING gj.*
     INTO grading_job;
 
-    -- ######################################################################
-    -- update all parent objects
+    IF new_gradable = FALSE THEN
+        -- ######################################################################
+        -- If the submission is ungradable then we shouldn't update grades
+        -- and use up an attempt
```
mwest1066, line 96: Should we be setting `submission.gradable = false` here? And also setting `submission.format_errors` (and possibly `submission.feedback`?)

docs/externalGrading.md
mwest1066: Yep, I hear you. I added a couple more comments below. I'd like to be clearer about exactly which properties are being set on the `grading_job` and `submission` when we have valid/invalid submissions. Ideally the rules should be the same whether we discover the invalid status during `parse()` or `grade()`.

Also, it occurs to me that it would be nice to add something about the `grading_job` to the question flow diagram at the bottom of https://prairielearn.readthedocs.io/en/latest/dev-guide/

I think we just need to add information about the `grading_job` to the "Graded submission" and "Ungraded submission" ellipses.

sprocs/grading_jobs_process_external.sql
nicknytko: Yep, external grader results contain everything in `feedback`, they don't have a `format_errors` key.

sprocs/grading_jobs_insert_internal.sql
nicknytko: Ah yup, good catch :sweat_smile:

sprocs/grading_jobs_process_external.sql
mwest1066: Perhaps we should add `format_errors` as an allowed key in `results.json`? I'm thinking that the rule will basically be that external graders work just like internal graders in terms of which variables they can access and which ones they can modify?

sprocs/grading_jobs_insert_internal.sql
nicknytko: Oh, no actually this will be already set in the `UPDATE` on lines 56-70

mwest1066: Oh right!

sprocs/grading_jobs_process_external.sql
nicknytko: Sure, I'd be on board with that.  I'm thinking if there are any `format_errors` they would be displayed between the "Score" and "Test Results" sections (I think this is where any messages or output would currently be displayed anyways)

image

mwest1066: What do we want our exact rules to be for `gradable`, `score`, and `format_errors`?

I think at the moment we [only allow](https://github.com/PrairieLearn/PrairieLearn/blob/master/question-servers/freeform.js#L376) a `server.py` `grade()` function to set `format_errors`, and then we [enforce](https://github.com/PrairieLearn/PrairieLearn/blob/master/question-servers/freeform.js#L1166) the rule that if `format_errors` is not empty then `gradable = false`. We don't enforce `score = null` in this case, but maybe we should? Also maybe we should force `partial_scores` to `{}` to avoid leaking any information?

I'm thinking about something like:

* After both `parse()` and `grade()`, if we have `format_errors` then set `gradable = false`, `score = null`, and `partial_scores = {}`.
* If we don't have `format_errors` then set `gradable = true`.

The above rules would apply to both internal and external grading.

Thoughts?

nicknytko: Yeah, I like `score = null` and `partial_scores = {}`, it's better than having some random garbage in there.

I also do like having an explicit `gradable` flag for the grader, though I guess for consistency with the internal grading scheme we should just get that from `format_errors`.

mwest1066: I'm also fine with having `gradable` as an explicit flag. Would we add it to the internal grade() as well, to keep consistency?

nicknytko: I would think so, that might be best to do in a separate PR as I could see many of the python elements needing to be updated.

nicknytko: image

Thoughts? Unrelated, but I think we should not display "Test results" if there are none.

nicknytko: image

nicknytko: @mwest1066 should we be actually setting `partial_scores` to null here?  I see that's what the value is when `parse()` fails.

nicknytko:
@mwest1066 ping

docs/question.md
```diff
@@ -155,6 +155,9 @@ def grade(data):
     # data["score"] is the total score for the question (0 to 1).
     # We can modify or delete any of these if we have a custom grading method.
     # This function only runs if `parse()` did not produce format errors, so we can assume all data is valid.
+	#
```
mwest1066, line 158: Is this a tab?

sprocs/grading_jobs_process_external.sql
mwest1066: I think we should set `partial_scores` to `null` and also `score` to `null` (or 0? whatever it would be if we detected a format error during parse). Otherwise we risk leaking grading information. For example, say there is a pl-number-input that already graded successfully. We don't want to show the student that their answer to that was correct.

I think we should also set `partial_scores` to `null` (and `score` to whatever) here: https://github.com/PrairieLearn/PrairieLearn/blob/master/question-servers/freeform.js#L1179

nicknytko: Yeah, this is what I have on the diagram I think:

https://raw.githubusercontent.com/PrairieLearn/PrairieLearn/7009d4a7e26738e07fc48f11e96bb878c659e94a/docs/question-flow.png

mwest1066: Perfect! Now we just need the code to match the diagram 😄

nicknytko: Oh, that should be in there already!  Internal grading jobs will set score, partials, and correct to `null`, external will set `score` and `correct` to `null` since the partials don't ever have a value anyway I think?

mwest1066: Hahaha, ok, I guess I need to look at the actual code 😄 

I think that it might be a good idea to set `partial_scores = null` as well, just in case we change things in the future to run both the internal and external grader (I actually think this could be a good idea).

PR #2208, opened 2020-03-23
Add showClosedAssessment access rule restriction

### LINKED ISSUE #560: Add access controls to limit question score/detail access in closed exams
Initial version in PR #551, with comments on that PR about implementation change.

nwalters512:
This PR adds a new `showClosedAssessment` flag to assessment access rules that controls whether closed assessments can be opened if their access rules would otherwise permit them to be opened.

Resolves #560

I unfortunately have a day job to go do this week, so I'm passing this off to @mwest1066 to see through.

**TODO**
- [x] Before merging, consider renaming in light of the `completelyHideAllTraceOfClosedAssessment` option
  - Consider `accessWhenClosed` instead of `showClosedAssessment`, and `displayWhenClosed` for `completelyHideAllTraceOfClosedAssessment`
- [x] Add documentation for this option, what it does, when to use it
- [x] Tests, ideally

### DIFF
--- a/database/tables/assessment_access_rules.pg
+++ b/database/tables/assessment_access_rules.pg
@@ -9,6 +9,7 @@ columns
     password: text
     role: enum_role
     seb_config: jsonb
+    show_closed_assessment: boolean default true
     start_date: timestamp with time zone
     time_limit_min: integer
     uids: text[]
--- a/database/tables/assessment_access_rules.pg
+++ b/database/tables/assessment_access_rules.pg
@@ -9,7 +9,7 @@ columns
     password: text
     role: enum_role
     seb_config: jsonb
-    show_closed_assessment: boolean default true
+    show_closed_assessment: boolean not null default true
     start_date: timestamp with time zone
     time_limit_min: integer
     uids: text[]
--- a/doc/assessment.md
+++ b/doc/assessment.md
@@ -143,7 +143,7 @@ PrairieLearn is designed to give students immediate feedback on their work. Howe
 
 To disable real-time grading for an assessment, add `"allowRealTimeGrading": false` to the assessment's `infoAssessment.json` file. This will hide the "Save & Grade" button on student question pages; only the "Save" button will be available. The "Grade saved answers" button on the assessment overview will also be hidden. Note that real-time grading can only be disabled for `Exam` assessments, as immediate feedback is a core part of the `Homework` experience.
 
-An assessment without real-time grading will not show any score information during the exam, but after a [time limit](accessControl.md#time-limits) runs out the assessment with auto-grade and show students exactly which questions they got correct/incorrect. To prevent this, set the [`showClosedAssessment` access rule restriction](accessControl.md#showinghiding-closed-assessments).
+An assessment without real-time grading will not show any score information during the exam, but after the [time limit](accessControl.md#time-limits) runs out the assessment will auto-grade and show students exactly which questions they got correct/incorrect. To prevent this, set the [`showClosedAssessment` access rule restriction](accessControl.md#showinghiding-closed-assessments).
 
 Disabling real-time grading changes a lot of fundamental details of how PrairieLearn is used. To account for that, the student assessment overview page displays less information about points and grading than for usual exams.
 
--- a/doc/assessment.md
+++ b/doc/assessment.md
@@ -143,6 +143,8 @@ PrairieLearn is designed to give students immediate feedback on their work. Howe
 
 To disable real-time grading for an assessment, add `"allowRealTimeGrading": false` to the assessment's `infoAssessment.json` file. This will hide the "Save & Grade" button on student question pages; only the "Save" button will be available. The "Grade saved answers" button on the assessment overview will also be hidden. Note that real-time grading can only be disabled for `Exam` assessments, as immediate feedback is a core part of the `Homework` experience.
 
+An assessment without real-time grading will not show any score information during the exam, but after a [time limit](accessControl.md#time-limits) runs out the assessment with auto-grade and show students exactly which questions they got correct/incorrect. To prevent this, set the [`showClosedAssessment` access rule restriction](accessControl.md#showinghiding-closed-assessments).
+
 Disabling real-time grading changes a lot of fundamental details of how PrairieLearn is used. To account for that, the student assessment overview page displays less information about points and grading than for usual exams.
 
 Here is the assessment page for a normal exam with real-time grading enabled:
--- a/doc/assessment.md
+++ b/doc/assessment.md
@@ -143,7 +143,7 @@ PrairieLearn is designed to give students immediate feedback on their work. Howe
 
 To disable real-time grading for an assessment, add `"allowRealTimeGrading": false` to the assessment's `infoAssessment.json` file. This will hide the "Save & Grade" button on student question pages; only the "Save" button will be available. The "Grade saved answers" button on the assessment overview will also be hidden. Note that real-time grading can only be disabled for `Exam` assessments, as immediate feedback is a core part of the `Homework` experience.
 
-An assessment without real-time grading will not show any score information during the exam, but after the [time limit](accessControl.md#time-limits) runs out the assessment will auto-grade and show students exactly which questions they got correct/incorrect. To prevent this, set the [`showClosedAssessment` access rule restriction](accessControl.md#showinghiding-closed-assessments).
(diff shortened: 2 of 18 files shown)


schemas/schemas/infoAssessment.json
```diff
@@ -166,6 +166,11 @@
                             }
                         }
                     }
+                },
+                "showClosedAssessment": {
+                    "description": "If set, prevents the assessment from being viewed once it has been closed",
```
mwest1066:
```suggestion
                    "description": "If true, prevents the assessment from being viewed once it has been closed",
```

migrations/166_assessment_access_rules__show_closed_assessment__add.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE assessment_access_rules ADD COLUMN show_closed_assessment boolean DEFAULT true;
```
mwest1066:
```suggestion
ALTER TABLE assessment_access_rules ADD COLUMN show_closed_assessment boolean NOT NULL DEFAULT true;
```

database/tables/assessment_access_rules.pg
```diff
@@ -9,6 +9,7 @@ columns
     password: text
     role: enum_role
     seb_config: jsonb
+    show_closed_assessment: boolean default true
```
mwest1066:
```suggestion
    show_closed_assessment: boolean not null default true
```

mwest1066:
This PR is now complete and ready for testing. To test it, Exam 2 in the example course (XC 101) is configured with:
* `"showClosedAssessment": false` --- this is the new option in this PR
* `"allowRealTimeGrading": false` --- this removes all the "Grade" buttons (from the recent PR)
* 5 minute time limit

This exam should mimic the pen-and-paper experience, where the student can't grade their exam in real time, but can only save answers. Once the timer runs out, their exam will auto-close and grade itself. At this point they will be able to see the total score, but they should not be able to see any details about the exam at all. In particular, they should not be able to see which questions they got correct and incorrect.

mwest1066:
When testing, it's important to have `role = Student`. If you have `role = Instructor` (the default when running PL in local dev mode) then you will override all access control.

To change to `role = Student`, click on your name (Dev User) in the top navbar and choose "Change effective user". Then change the effective role to Student.

doc/assessment.md
```diff
@@ -143,6 +143,8 @@ PrairieLearn is designed to give students immediate feedback on their work. Howe
 
 To disable real-time grading for an assessment, add `"allowRealTimeGrading": false` to the assessment's `infoAssessment.json` file. This will hide the "Save & Grade" button on student question pages; only the "Save" button will be available. The "Grade saved answers" button on the assessment overview will also be hidden. Note that real-time grading can only be disabled for `Exam` assessments, as immediate feedback is a core part of the `Homework` experience.
 
+An assessment without real-time grading will not show any score information during the exam, but after a [time limit](accessControl.md#time-limits) runs out the assessment with auto-grade and show students exactly which questions they got correct/incorrect. To prevent this, set the [`showClosedAssessment` access rule restriction](accessControl.md#showinghiding-closed-assessments).
```
nicknytko:
```suggestion
An assessment without real-time grading will not show any score information during the exam, but after the [time limit](accessControl.md#time-limits) runs out the assessment will auto-grade and show students exactly which questions they got correct/incorrect. To prevent this, set the [`showClosedAssessment` access rule restriction](accessControl.md#showinghiding-closed-assessments).
```

schemas/schemas/infoAssessment.json
```diff
@@ -166,6 +166,11 @@
                             }
                         }
                     }
+                },
+                "showClosedAssessment": {
+                    "description": "If true, prevents the assessment from being viewed once it has been closed",
```
nicknytko: Should this be

`If *false*, prevents the assessment from being viewed once it has been closed`?

Otherwise the name seems confusing

middlewares/studentAssessmentAccess.ejs
```diff
@@ -32,13 +32,13 @@
...
-                      <div class="col-md-6 col-sm-12 text-right">
-                          Assessment is <strong>closed</strong> and is no longer available
+                      <div class="col-md-6 col-sm-12 text-right assessment-closed-message">
+                          Assessment is <strong>closed</strong> and is no longer available.
                       </div>
                     </div>
 
                     <% if (authz_data.mode == 'SEB') { %>
-                        <p><a href="/pl/SEBquit">Click here to quit the Safe Exam Browser</a>
+                        <p><a href="/pl/SEBquit">Click here to quit the Safe Exam Browser</a></p>
```
nicknytko: nice catch!

middlewares/studentAssessmentAccess.ejs
```diff
@@ -32,13 +32,13 @@
                       <div class="col-md-3 col-sm-6">
                         <%- include('../pages/partials/scorebar', {score: assessment_instance.score_perc}) %>
                       </div>
-                      <div class="col-md-6 col-sm-12 text-right">
-                          Assessment is <strong>closed</strong> and is no longer available
+                      <div class="col-md-6 col-sm-12 text-right assessment-closed-message">
```
nicknytko: Where is this css class `assessment-closed-message` defined?  I don't see it anywhere in this PR or in master.

nicknytko:
image

This red header looks scary and makes it seem like the student did something wrong, even though it was just the assessment finishing.  Can we change this to blue or some other colour?

middlewares/studentAssessmentAccess.ejs
mwest1066: It's not defined 😄

It's only used by the test suite. The tests use this class to make sure we are on the correct page and have the correct message.

Should we name it something else? `test-suite-assessment-closed-message`?

middlewares/studentAssessmentAccess.ejs
mwest1066: All credit for this goes to @nwalters512 !

schemas/schemas/infoAssessment.json
mwest1066: Oh yes! Good catch!

I reworded it completely.

middlewares/studentAssessmentAccess.ejs
nicknytko: Ooh, yeah that makes sense.  Changing it to that would definitely make this clearer.

doc/assessment.md
mwest1066: I used `a time limit` because there might not be one. Let me reword a bit more.

middlewares/studentAssessmentAccess.ejs
mwest1066: done

mfsilva22:
I didn't see any "issue" with the assessment. It does exactly what it is supposed to do.  

I noticed that if I enter an answer but forget to click "Save', my answers are not automatically saved when the assessment closes. What is the reasoning behind this? I think it would be better if we could save the answers.

Do you want feedback on the questions as well? Just in case you are...

The submitted answer for `demoMatrixComplexAlgebra` does not fit in the display area. It does not look very nice. Maybe you can choose another example question?

I would have the answers in `demoRandomDataFrame` showing in separate lines.

coatless:
Scratch that note, we needed to start the exam while under the effective role of a student from the get-go. Restarting the docker container gave the desired results. 


Screen Shot 2020-03-24 at 7 20 08 PM
Screen Shot 2020-03-24 at 7 20 38 PM

mwest1066:
> I didn't see any "issue" with the assessment. It does exactly what it is supposed to do.

Thanks for checking! Most appreciated!!

> I noticed that if I enter an answer but forget to click "Save', my answers are not automatically saved when the assessment closes. What is the reasoning behind this? I think it would be better if we could save the answers.

This is basically the way that PrairieLearn has always worked -- only explicit "Save" or "Save & Grade" operations will cause an answer to be saved. The problem with doing it implicitly is that, suppose a student had a correct multiple-choice answer saved, and then they accidentally clicked a their mouse on a different answer just before the timer ran out, they might be annoyed if they got auto-saved and marked incorrect.

However, on the other hand, I agree that if a student clicks a multiple-choice answer but doesn't have time to click "Save", they will be annoyed if the timer runs out and they don't get the points.

I'm not sure what the best way is to reconcile these? I'd definitely like to discuss this more and come to a good solution. Can you bring this up on #pl-dev or file an issue?

> Do you want feedback on the questions as well? Just in case you are...

Not really 😄 --- Nathan just grabbed some random questions, and I haven't even looked at which ones they are! I'm hoping that you can fix up the questions!

> The submitted answer for `demoMatrixComplexAlgebra` does not fit in the display area. It does not look very nice. Maybe you can choose another example question?

I'll leave it for now because I want to get this merged, but maybe you can submit a PR to clean some of these things up and have nice example exams? And nice questions!

> I would have the answers in `demoRandomDataFrame` showing in separate lines.

Please submit a PR! 😄

mwest1066:
Resolves #560

PR #2212, opened 2020-03-23
Fix errors in element documentation question

### LINKED ISSUE #2210: Errors in elementElementCodeDocumentation example question
Some elements seem to be behaving badly when dealing with incomplete (invalid) data. Steps to reproduce:

1. Go to the `elementElementCodeDocumentation` question in the exampleCourse (XC 101).

2. Choose an answer for the first multiple choice question, but leave all of the other ones blank.

3. Click "Save & Grade".

4. There will be errors thrown by some element on the page.

nicknytko:
Fixes #2210

The correct answer for the _symbolic input_ field was using a floating point variable `x` instead of symbolic `x`, which breaks the sympy parser because it doesn't like floats.

### DIFF
--- a/doc/elements.md
+++ b/doc/elements.md
@@ -343,7 +343,7 @@ import sympy
 def generate(data):
   
   # Declare math symbols
-  sympy.var('x y')
+  x, y = sympy.symbols('x y')
 
   # Describe the equation
   z = x + y + 1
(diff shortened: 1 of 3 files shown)


coatless:
@nicknytko can you change the corresponding help entry? 

https://github.com/PrairieLearn/PrairieLearn/blob/master/doc/elements.md#pl-symbolic-input-element

coatless:
@nicknytko missing the `var` label on the input box as well 😭

nicknytko:
:laughing: what would I do without you @coatless

PR #2214, opened 2020-03-23
Update docs to use `npm ci` instead of install

### LINKED ISSUE #1317: Docs are missing for running "npm ci" in a container with locally-mounted source

### LINKED ISSUE #2053: Dev docs don't describe how to run npm inside docker
The docs that describe how to use docker and mount in a local copy of PL source assume that you can run `npm` in the host OS. We should change this to run `npm` inside docker.

### LINKED ISSUE #2100: UX: Load from Disk, make my course more prominent
A side effect of having the PrairieLearn docker bring in the Example Course and the Test Course is that the Load from Disk view cluttered. The information the course content developer would want to look at ("how did my course sync?") is at the bottom and you need to scroll to see the overall status and the individual messages. I don't think most course content developers care how the static test and example course sync'd - but they do appreciate having them loaded.

Perhaps reorder the docker `config.json courseDirs` to put `/courses` first? Or do some sort of tabbed-by-course design for the sync page with /course showing?

nicknytko:
`npm install` => `npm ci`

Also fixes #2053, fixes #1317, fixes #2100


doc/installingLocal.md
```diff
@@ -39,7 +47,22 @@ docker run -it --rm -p 3000:3000 -v /path/to/PrairieLearn:/PrairieLearn prairiel
...
 ```
 
 This can be used to, e.g., run scripts distributed with PrairieLearn.
+
+#### Development from the shell
+
+When making local changes to server-side code, it is faster to restart only the node server instead of the whole docker container.  This can be done by starting the container into a shell environment and starting the server manually:
+
+```sh
+docker run -it --rm -p 3000:3000 -v /path/to/PrairieLearn:/PrairieLearn prairielearn/prairielearn /bin/bash
```
coatless: Maybe change this to just initialize the bash session instead of running the script?

From there, go into a brief overview of the possible `docker` scripts described by @mwest1066 ?

```bash
docker run -it --rm -p 3000:3000 -v /path/to/PrairieLearn:/PrairieLearn prairielearn/prairielearn /bin/bash
# inside container:
cd /PrairieLearn
npm ci
./docker/start_postgres.sh
./docker/lint_js.sh
./docker/lint_python.sh
./docker/test_js.sh
./docker/test_python.sh
```

doc/installingLocal.md
```diff
@@ -15,9 +15,17 @@ git clone https://github.com/PrairieLearn/PrairieLearn.git
...
+npm ci
 ```
 
+If you don't have `npm` installed on your computer, you can use the version that comes pre-installed in the Docker image:
+
+```sh
+docker run -w /PrairieLearn -v /path/to/PrairieLearn:/PrairieLearn prairielearn/prairielearn /usr/local/bin/npm ci
+```
+
+The path `/path/to/PrairieLearn` should be replaced with the *absolute* path to the PrairieLearn source on your computer.  If your terminal is currently in this folder, you can replace this with `` `pwd` `` to auto-fill the path.
```
mwest1066: I think `pwd` only works on MacOS/Linux? Some other instructions (maybe the "getting started" ones?) have the Windows equivalent.

doc/installingLocal.md
```diff
@@ -11,13 +11,14 @@ This page describes the procedure to run PrairieLearn within Docker, but using a
...
 
 ```sh
-cd PrairieLearn
-npm install
+docker run -w /PrairieLearn -v /path/to/PrairieLearn:/PrairieLearn prairielearn/prairielearn /usr/local/bin/npm ci
 ```
 
+The path `/path/to/PrairieLearn` should be replaced with the *absolute* path to the PrairieLearn source on your computer.  If you're in the root of the source directory already, you can substitute `%cd%` (on Windows) or `$PWD` (Linux and MacOS) for `/path/to/PrairieLearn`.
+
 * Create the file `PrairieLearn/config.json`:
```
mwest1066: Actually, since #2158 this step is not required. PL will use a default `courseDirs` with just the `exampleCourse` if `config.json` is missing. Maybe we should delete this step from the main instruction sequence but add a note saying that PL will use the exampleCourse by default?

Actually, maybe we should change the default (in `lib/config.js`) to match `docker/config.json`? That is, include `testCourse` and the `config[N]` directories? That would allow us to delete `docker/config.json` and would also make the experience using containerized version for dev more similar to the normal docker-based PL experience.

While we are there, I'm wondering whether we should also change the default for `filesRoot` to match `docker/config.json`. Are there downsides with this?

mwest1066:
If making the above change, maybe consider #2100 and put the `/course[N]` directories first in the list?

doc/installingLocal.md
nicknytko: Absolutely, I think removing the config.json step is a great idea.  I've briefly looked through the source and I don't see any immediate downsides to giving a value for `filesRoot`, so I'll go about updating the values in `lib/config.js`.

doc/configJson.md
```diff
@@ -0,0 +1,37 @@
+# Server Configuration
```
mwest1066: To get this new doc file to show up in the docs list it needs to be added to https://github.com/PrairieLearn/PrairieLearn/blob/master/mkdocs.yml

PR #2215, opened 2020-03-24
Add confirmation for unsaved changes on question pages

### LINKED ISSUE #223: Alert user of unsaved changes when they try navigating away from page

### LINKED ISSUE #506: Warn when leaving question with unsaved answers during exams
When a student tries to navigate away from a question page that has unsaved changes to an answer, we should pop up a dialog asking whether they are sure they want to leave.

To do this, we could use [`onbeforeunload`](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload) or [`onunload`](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onunload). We would then need to get the current submission from the question and compare it to the last saved submission to determine whether there are unsaved changes.

### LINKED ISSUE #1022: On file upload, warn the user if they navigate away before saving or grading
From a student:
> I took quiz 1, in the CBTF today, and in a hurry to finish the quiz in time I uploaded the file but forgot to press the save button on Prairielearn.

Could maybe use the `onbeforeunload` function? https://stackoverflow.com/questions/1119289/how-to-show-the-are-you-sure-you-want-to-navigate-away-from-this-page-when-ch

### LINKED ISSUE #1132: warn student leaving a question instance page without saving
Add the javascript 'are you sure you want to navigate away' warning to question instance pages. I don't think that's custom code; it's just triggering the browser to warn.

From Mariana based on her TA's walk-through of a first time with PrairieLearn. Maybe Relate does this?

### LINKED ISSUE #1475: Require file submission if in CBTF
If a student navigates away from a question without submitting a file, there should probably be a warning.

tdy:
Client-side confirmation script for unsaved changes on question pages
* [x] Support `ace-mode`, `pl-file`, `pl-drawing`, etc.
* [x] Support Firefox
* [x] Support Chrome
* [x] Support Edge
* [x] Support Safari **(seems it never pops up again after "leave" is chosen once?)**
* [x] Suppress for expiring exams
* [ ] Auto-save/submit on expiring exams **(maybe?)**

Fixes #1132 
Fixes #506 
Fixes #223 
Fixes #1475 
Fixes #1022


pages/partials/questionFooter.ejs
```diff
@@ -115,3 +115,23 @@
...
+    $(document).ready(() => {
+        // Set form state on load and submit
+        var initialForm = getForm();
+        form.submit(() => { initialForm = getForm(); });
+
+        // Check form state on unload
+        $(window).bind('beforeunload', (event) => {
+            const currentForm = getForm();
+            if(initialForm != currentForm) {
+                (event || window.event).returnValue = null;
```
nicknytko: Can you throw in a comment explaining what setting `returnValue = null` does?  Are we trying to prevent some default action here?

tdy: Done, with some minor tweaks in the process (will describe below)

pages/partials/questionFooter.ejs
```diff
@@ -115,3 +115,31 @@
...
+            const currentForm = getForm();
+            if(initialForm != currentForm) {
+                // MDN recommendation (not fully supported)
+                (event || window.event).preventDefault();
+
+                // Chrome legacy
+                (event || window.event).returnValue = '';
+
+                // Fallback
+                return '';
```
tdy: New:
* added `preventDefault()` (per MDN)
* added comments to explain the fallback chain
* changed the return values from `null` to `''` (as used by MDN, though seems to work either way)

mwest1066:
News item looks like this:

[PrairieLearn.pdf](https://github.com/PrairieLearn/PrairieLearn/files/4382829/PrairieLearn.pdf)

nicknytko:
> News item looks like this:
> 
> [PrairieLearn.pdf](https://github.com/PrairieLearn/PrairieLearn/files/4382829/PrairieLearn.pdf)

Beautiful, I love how macOS includes the shadow in window screenshots.

mwest1066:
> Beautiful, I love how macOS includes the shadow in window screenshots.

It's awesome, but I was worried it looks too real? Maybe that's just because I'm on a mac. I kept thinking "damn, there's some dialog on top of my page".

tdy:
Wait! Just tested what @nicknytko mentioned in slack. If you have unsaved changes when a timed exam expires, you can click "stay" to stay in the exam.

tdy:
Now uses `window.allowQuestionUnload` in `/pages/partials/countdown.ejs` to bypass the warning for expiring exams.

ChangeLog.md
```diff
@@ -89,6 +89,10 @@
 
   * Add example course question templates (James Balamuta).
 
+  * Add extra `/course*` mount points (Tim Yang).
```
tdy: Not sure how this got here. Thought this was in my old PR.

nicknytko: Probably when you merged master, this is *ok* I think?

mwest1066: Yeah, those lines will come in from master mergers.

Note that we are actually not worrying about updating ChangeLog anymore. We now use the git log as the source of truth. At some point soon I would like to add a bot to automatically mirror the git log into the ChangeLog file, but for now we are just neglecting ChangeLog.

mwest1066:
I'm not sure whether we should pay attention to this, but I see recommendations to not use the pattern we've gone for: https://developers.google.com/web/updates/2018/07/page-lifecycle-api

mwest1066:
On Safari, once you click "Leave" on one warning dialog (meaning to leave the page and discard unsaved changes) it never asks you again on other questions.

I've been googling, but I haven't found a solution for this yet.

tdy:
> On Safari, once you click "Leave" on one warning dialog (meaning to leave the page and discard unsaved changes) it never asks you again on other questions.

As in only on Safari? I can't reproduce on Firefox/Chrome/Edge.

tdy:
Oh wait I'm seeing a "no more dialogs" checkbox. Maybe that's auto-enabled on Safari?

image

mwest1066:
> Oh wait I'm seeing a "no more dialogs" checkbox. Maybe that's auto-enabled on Safari?

I tested with Chrome and Safari. On Chrome it seemed that the dialog kept prompting as it should even after repeated use (I tested maybe three questions). On Safari it worked the first time, but then it never prompted again. I didn't test FF.

Which browser gave you the "no more dialogs" message?

Rather than using `onbeforeunload`, should we hook `onclick` of all the nav buttons (next/prev questions, navbar, etc) and trigger our own dialog? Does that even work? It wouldn't stop students closing the browser, but I don't think that's the main problem anyway. It's our own site nav that we are worried about, I think.

tdy:
> Which browser gave you the "no more dialogs" message?

"no more dialogs" checkbox was in Chrome.

> Rather than using `onbeforeunload`, should we hook `onclick` of all the nav buttons (next/prev questions, navbar, etc) and trigger our own dialog? Does that even work? It wouldn't stop students closing the browser, but I don't think that's the main problem anyway. It's our own site nav that we are worried about, I think.

What about browser nav (esp key shortcuts <kbd>cmd-r</kbd> <kbd>cmd-[</kbd> <kbd>cmd-]</kbd>)? At least for me, I'll sometimes accidentally use a key shortcut instinctively when I didn't actually want to.

I tried implementing what I thought the Lifecycle page meant, but either I have bugs or I misunderstood the page because it's not working atm.

```javascript
<script>
    const form = $('form.question-form');
    const getForm = () => form.serialize();

    // Wait for DOM (ace, file upload, etc.)
    $(document).ready(() => {
        // Set form state on load and submit
        var initialForm = getForm();
        form.submit(() => {
            initialForm = getForm();
        });

        // Unsaved alert listener (added/removed dynamically)
        const beforeUnloadListener = (event) => {
            event.preventDefault();
            return event.returnValue = '';
        };

        // Add/remove alert based on form change state
        form.change(() => {
            const isSameForm = initialForm == getForm();
            if(isSameForm) {
                removeEventListener('beforeunload', beforeUnloadListener, {capture: true});
            } else if(!window.allowQuestionUnload) {
                addEventListener('beforeunload', beforeUnloadListener, {capture: true});
            }
        }
    });
</script>
```

tdy:
The `onclick` approach would basically be the same, right? Except using `$('whatever-site-nav-selector').click`?

mwest1066:
I had the impression that the lifecycle events could be used to run some code (to save state) but you couldn't block the event and prevent the user from leaving the page. But I might have misunderstood that.

mwest1066:
> The `onclick` approach would basically be the same, right? Except using `$('whatever-site-nav-selector').click`?

Exactly. And I guess we could select everything except what's inside the question form?

tdy:
> Exactly. And I guess we could select everything except what's inside the question form?

So you're not concerned about back/reload? or <kbd>cmd-[</kbd> / <kbd>cmd-r</kbd>?

mwest1066:
> 1. So you're not concerned about back/reload? Or cmd-[ / cmd-r?

I'm concerned about it, but I guess I'm not sure what the alternative is? Assuming the `onbeforereload` doesn't work reliably.

The history API (and especially onpopstate) seem that they can do various things to interfere with the back button, but it all feels pretty hacky.

> 2. As far as links go, this would mean giving everything its own id? Or capture blanket selectors? Feels like this is kinda error-prone, especially with any future UI changes?
>    `nav a.nav-link` `nav a.dropdown-item` `nav a.btn` (or just all `nav a`?)
>    `a.nav-link` (Files/Preview/Settings/Statistics)
>    `a.btn` (Assessment Overview, Next Question, Previous Question)
>    `button` (Save & grade, Save only, Report an error, Attach file, Attach text)

Would it work to select all `<a>` with `href` set? For buttons we could add a class like `needs-unsaved-check` to the few buttons on the question page.

As I understand it, we don't seem to have any solution that will work reliably on all browsers for all types of page interaction? What do you think the least-worst option is?

tdy:
> As I understand it, we don't seem to have any solution that will work reliably on all browsers for all types of page interaction? What do you think the least-worst option is?

@mwest1066 will you be at PL OH tomorrow? Discuss there? I'm thinking some stuff is getting lost in the shuffle here. Should be easier at OH to clarify what is/isn't working (and won't work) on both ends.

mwest1066:
> @mwest1066 will you be at PL OH tomorrow? Discuss there? I'm thinking some stuff is getting lost in the shuffle here. Should be easier at OH to clarify what is/isn't working (and won't work) on both ends.

Yep, I will be.

PR #2220, opened 2020-03-24
Add `size` and `show-help-text` attributes to elements

### LINKED ISSUE #2218: Add size parameter for string element and other elements
Make string and other pl elements have a size parameter so that they can more easily be placed inside tables for multi-part problems.

nicknytko:
Adds `size` and `show-help-text` for:

- `pl-string-input`
- `pl-integer-input`
- `pl-symbolic-input`

Add `show-help-text` for:
- `pl-matrix-input` (this is a block element)

Fixes #2218

### DIFF
--- a/doc/elements.md
+++ b/doc/elements.md
@@ -366,6 +366,8 @@ Attribute | Type | Default | Description
 `variables` | string | — | A comma-delimited list of symbols that can be used in the symbolic expression.
 `allow-complex` | boolean | false | Whether complex numbers (expressions with `i` or `j` as the imaginary unit) are allowed.
 `imaginary-unit-for-display` | string | `i` | The imaginary unit that is used for display. It must be either `i` or `j`. Again, this is *only* for display. Both `i` and `j` can be used by the student in their submitted answer, when `allow-complex="true"`.
+`size` | integer | 35 | Size of the input box.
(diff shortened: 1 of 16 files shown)


tdy:
https://github.com/PrairieLearn/PrairieLearn/blob/4461ab81ad8f1b3245486bdbd6bb37c67f864746/elements/pl-symbolic-input/pl-symbolic-input.mustache#L19

I guess wrap symbolic expression placeholder in `show_info`?

`{{#show_info}}placeholder="{{shortinfo}}"{{/show_info}} />`

nicknytko:
> I think the one issue that has to be addressed was brought up by @tdy in Slack.
> 
> What happens to the inner input help text when `size < threshold`?
> 
> image

I think the simplest thing to do here is to just hide the placeholder text once the size of the input box gets small enough.  The text just says `symbolic input` or `integer` anyway so I don't think it's too much information lost?

> My only other note is the name of this PR says: `"hide-help-text"` when the new attribute is `"show-help-text"`.

:man_facepalming:

> I'm genuinely a bit worried about non-standard input box sizes... I can understand hiding help text... But, I feel like inputs should be standardized as much as possible.

Agreed, but I still think the option of being able to do something weird here is useful.

tdy:
Working on my side 👍

PR #2223, opened 2020-03-24
Add online exam with dres accommodations example

trombonekenny:
Includes an example and description of `"mode": "Public"` exams with `uids` exceptions


mwest1066:
Incorporated info into #2224

PR #2242, opened 2020-03-25
Add admin dashboard for usage queries

trombonekenny:
Introduces a light-weight SQL query display interface to see the current state of the database. Works off of folders and files in the `pages/administratorDashboard` folder.


pages/administratorDashboard/main/available_assessments.sql
```diff
@@ -0,0 +1,22 @@
+SELECT c.short_name,ci.long_name,a.title,
+    a.id as assessment_id,
+    a.course_instance_id,
+    ci.course_id,
+    aar.start_date,
```
mwest1066:
```suggestion
    format_date_full_compact(aar.start_date, config_for_timezone.value),
```

pages/administratorDashboard/main/available_assessments.sql
```diff
@@ -0,0 +1,22 @@
...
+    aar.end_date,
```
mwest1066:
```suggestion
    format_date_full_compact(aar.end_date, config_for_timezone.value),
```

pages/administratorDashboard/main/available_assessments.sql
```diff
@@ -0,0 +1,22 @@
...
+    aar.end_date - aar.start_date AS duration,
```
mwest1066:
```suggestion
    format_interval_short(aar.end_date - aar.start_date AS duration),
```

pages/administratorDashboard/main/available_assessments.sql
```diff
@@ -0,0 +1,22 @@
+SELECT c.short_name,ci.long_name,a.title,
```
mwest1066:
```suggestion
WITH config_for_timezone AS (
    SELECT * WHERE key = 'display_timezone'
)
SELECT c.short_name,ci.long_name,a.title,
```

pages/administratorDashboard/main/available_assessments.sql
```diff
@@ -0,0 +1,22 @@
...
+    aar.role,
+    aar.uids,
+    (SELECT count(*) FROM enrollments WHERE course_instance_id=ci.id AND role='Student') as enrollments
+--    to_jsonb(aar.*) as aar_json,
+--    to_json(a.*) as a_json
+
+FROM assessment_access_rules AS aar
+JOIN assessments AS a ON (a.id = aar.assessment_id)
+JOIN course_instances AS ci ON (ci.id = a.course_instance_id)
+JOIN pl_courses AS c ON (c.id = ci.course_id)
```
mwest1066:
```suggestion
JOIN pl_courses AS c ON (c.id = ci.course_id)
LEFT JOIN config_for_timezone
```

pages/administratorDashboard/administratorDashboard.ejs
```diff
@@ -0,0 +1,113 @@
...
+    <div class="col-12">
+
+<% queries.forEach(function(q) { 
+   var columns = [];
+%>
+
+<h2 id="<%= q.SQLFilename %>"><%= q.SQLFilename %>
+    <span class="badge badge-pill badge-info"><%= q.rowCount %></span></h2>
+<span class="badge badge-secondary"
+    title="<%= q.SQLstring %>">SQL</span>
```
mwest1066: I had to read the EJS source to figure out how to use this button. Just make it a collapse element in the main page?

pages/administratorDashboard/main/available_assessments.sql
```diff
@@ -0,0 +1,22 @@
...
+    aar.end_date,
+    aar.end_date - aar.start_date AS duration,
+    aar.role,
+    aar.uids,
+    (SELECT count(*) FROM enrollments WHERE course_instance_id=ci.id AND role='Student') as enrollments
+--    to_jsonb(aar.*) as aar_json,
+--    to_json(a.*) as a_json
+
+FROM assessment_access_rules AS aar
+JOIN assessments AS a ON (a.id = aar.assessment_id)
```
mwest1066:
```suggestion
JOIN assessments AS a ON (a.id = aar.assessment_id)
JOIN assessment_sets AS aset ON (aset.id = a.assessment_set_id)
```

mwest1066:
To be honest, I'm not loving the "multiple queries on one page" layout. It makes it hard to scroll back to the top of a particular query -- I'm always getting lost about which query I'm actually looking at.

How about just doing a single-query-per-page model? So the main page is just "here is a list of queries" and you click on one and it runs?

pages/administratorDashboard/administratorDashboard.js
```diff
@@ -0,0 +1,62 @@
...
+    var dashboard = req.params.dashboard;
+    res.locals.dashboardName = dashboard;
+    res.locals.otherDashboards = [];
+    res.locals.baseUrl = req.baseUrl;
+
+    var params = {
+        req_date: res.locals.req_date,
+    };
+    res.locals.queries = [];
+    res.locals.req_date_present = moment(res.locals.req_date).format('llll');
```
mwest1066: I'd rather pull this from the DB using something like
```
SELECT format_date_full(now(), (SELECT value FROM config WHERE key = 'display_timezone'));
```

pages/administratorDashboard/main/available_assessments.sql
trombonekenny: I fixed this by using a sproc, but I appreciate the pointer!

trombonekenny:
> To be honest, I'm not loving the "multiple queries on one page" layout. It makes it hard to scroll back to the top of a particular query -- I'm always getting lost about which query I'm actually looking at.
> 
> How about just doing a single-query-per-page model? So the main page is just "here is a list of queries" and you click on one and it runs?
> 
> Feel free to push back! It could well be that I'm not understanding the vision of how this will work.
> 
> (I also think the single-query-per-page model will be much easier to extend to parameterized queries, for what it's worth.)

I think it's two different visions that similar enough that it's easy to confuse them. I wanted a quick, at-a-glance dashboard view of different things server/DB related. The multiple queries in one page is a little easier to reconcile than a big CTE and having to pack/unpack the data out of it.

It sounds like you're going for a more elaborate SQL runner, where the all-in-one page doesn't really make sense. I changed it so the page does multiple dashboards, so a bigger query could be run manually on it own. Is that enough of a reconciliation or are we looking at two different features?

mwest1066:
Merged in a slightly reworked form as #2345

PR #2248, opened 2020-03-26
Fix countdown timer bar colors

### LINKED ISSUE #2246: Countdown timer bar does not turn yellow/red near the end
We have code in in the countdown timer bar that is supposed to turn it yellow and then red near the end: https://github.com/PrairieLearn/PrairieLearn/blob/master/pages/partials/countdown.ejs#L74

This used to work, but now it just stays blue until the end.

tdy:
Fixes #2246. Noticed the cause when I was digging around `countdown.ejs` for 2215.

(yellow does work, no screenshot)


mwest1066:
Ah! Bootstrap 4 conversion, I guess.

PR #2261, opened 2020-03-28
Document server.py functions

### LINKED ISSUE #809: Document `server.py` functions
Basically a copy of http://prairielearn.readthedocs.io/en/latest/devElements/#element-functions

The `file()` function is currently missing from the above list.

### LINKED ISSUE #2052: Document server.py functions
Much of this can be copied from `devElements.md`, but other functions like `file()` also need to be added.

tdy:
* [x] Copy (or link to?) functions table in `devElements.md`
    * Having a copy seems easier for end users while linking is easier for maintainers
* [x] Alphabetize keys (can undo if preferred)
* [x] Add `file()`
* [x] ~~Add `test()`~~ Tabled for now
* [x] Add `server.py` skeleton

Closes #809 
Closes #2052


doc/devElements.md
```diff
@@ -100,7 +100,7 @@ Function | Return object | modifiable `data` keys | unmodifiable `data` keys | D
 --- | --- | --- | --- | ---
 `generate()` | `data` (dict) | `params`, `correct_answers` | `variant_seed`, `options` | Generate the parameter and true answers for a new random question variant. Set `data["params"][name]` and `data["correct_answers"][name]` for any variables as needed. Return the modified `data` dictionary.
 `prepare()` | `data` (dict) | `params`, `correct_answers` | `variant_seed`, `options` | Final question preparation after element code has run. Can modify data as necessary. Return the modified `data` dictionary.
-`render()` | `html` (string) | `params`, `correct_answers`, `submitted_answers`, `format_errors`, `partial_scores`, `score`, `feedback` `variant_seed`, `options`, `raw_submitted_answers`, `editable`, `panel` | Render the HTML for one panel and return it as a string.
+`render()` | `html` (string) | | `params`, `correct_answers`, `submitted_answers`, `format_errors`, `partial_scores`, `score`, `feedback` `variant_seed`, `options`, `raw_submitted_answers`, `editable`, `panel` | Render the HTML for one panel and return it as a string.
```
tdy: Column was missing. I believe I added the missing column back to the right place -- based on my interpretation of [freeform.js#L364](https://github.com/PrairieLearn/PrairieLearn/blob/master/question-servers/freeform.js#L364), all keys are unmodifiable in `render()`.

doc/question.md
```diff
@@ -87,6 +87,18 @@ def generate(data):
...
+`grade()` | `data` (dict) | `params`, `correct_answers`, `submitted_answers`, `format_errors`, `partial_scores`, `score`, `feedback` | `variant_seed`, `options`, `raw_submitted_answers` | Grade `data["submitted_answers"][var]` to determine a score. Store the score and any feedback in `data["partial_scores"][var]["score"]` and `data["partial_scores"][var]["feedback"]`. Return the modified `data` dictionary.
+`file()` | string, byte-like, file-like | | `params`, `correct_answers`, `variant_seed`, `options`, `filename` | Generate a file object dynamically in lieu of a physical file. Trigger via `type="dynamic"` in the question element (e.g., `pl-figure`, `pl-file-download`). Access the requested filename via `data['filename']`. If `file()` returns nothing, an empty string will be used.
+`test()` | `data` (dict), `Exception` | `format_errors`, `raw_submitted_answers`, `partial_scores`, `score` | `params`, `correct_answers`, `variant_seed`, `options`, `feedback`, `gradable` |
```
tdy: Not sure about `test()` description

doc/question.md
```diff
@@ -87,6 +87,18 @@ def generate(data):
...
+`prepare()` | `data` (dict) | `params`, `correct_answers` | `variant_seed`, `options` | Final question preparation after element code has run. Can modify data as necessary. Return the modified `data` dictionary.
+`render()` | `html` (string) | | `params`, `correct_answers`, `submitted_answers`, `format_errors`, `partial_scores`, `score`, `feedback` `variant_seed`, `options`, `raw_submitted_answers`, `editable`, `panel` | Render the HTML for one panel and return it as a string.
+`parse()` | `data` (dict) | `submitted_answers`, `format_errors` | `params`, `correct_answers`, `variant_seed`, `options`, `raw_submitted_answers` | Parse the `data["submitted_answers"][var]` data entered by the student, modifying this variable. Return the modified `data` dictionary.
+`grade()` | `data` (dict) | `params`, `correct_answers`, `submitted_answers`, `format_errors`, `partial_scores`, `score`, `feedback` | `variant_seed`, `options`, `raw_submitted_answers` | Grade `data["submitted_answers"][var]` to determine a score. Store the score and any feedback in `data["partial_scores"][var]["score"]` and `data["partial_scores"][var]["feedback"]`. Return the modified `data` dictionary.
```
tdy: Copy this? Or link to `devElements.md`? Having a copying is probably easier for consumers, but linking is easier to maintain.

coatless:
@tdy maybe link to the `server.py` document? Emphasize in the question document that additional server functions are possible.

Alternatively, how about include a fully laid out `server.py` with all the functions?

doc/question.md
mwest1066: I think "copy" is the way to go here. It's more painful for us, but much friendlier for users, as you say.

doc/devElements.md
```diff
@@ -98,11 +98,12 @@ The element functions are:
...
-`parse()` | `data` (dict) | `submitted_answers`, `format_errors` | `params`, `correct_answers`, `variant_seed`, `options`, `raw_submitted_answers` | Parse the `data["submitted_answers"][var]` data entered by the student, modifying this variable. Return the modified `data` dictionary.
-`grade()` | `data` (dict) | `params`, `correct_answers`, `submitted_answers`, `format_errors`, `partial_scores`, `score`, `feedback` | `variant_seed`, `options`, `raw_submitted_answers` | Grade `data["submitted_answers"][var]` to determine a score. Store the score and any feedback in `data["partial_scores"][var]["score"]` and `data["partial_scores"][var]["feedback"]`. Return the modified `data` dictionary.
+`generate()` | `data` (dict) | `correct_answers`, `params` | `options`, `variant_seed` | Generate the parameter and true answers for a new random question variant. Set `data["params"][name]` and `data["correct_answers"][name]` for any variables as needed. Return the modified `data` dictionary.
```
mwest1066: We actually changed things long ago so that functions DON'T return the modified `data` dictionary. They just modify it in place. The return value from these functions is ignored, so they are free to return `data` but it doesn't do anything.

Note that `render()` and `file()` do still return meaningful values (the HTML and file contents, respectively).

mwest1066:
> Alternatively, how about include a fully laid out `server.py` with all the functions?

I think this is an excellent idea! It's often much easier to just look at an example than read docs 😄 

I like skeletons of code with inline comments explaining what the code is doing. Perhaps something like:

```
def generate(data):
    # generate random parameters for the question and store them in the data["params"] dict
    data["params"]["x"] = random.randint(5, 10)
    data["params"]["operation"] = random.choice(["double", "triple"])

    # also compute the correct answer (if there is one) and store in the data["correct_answer"] dict:
    if data["params"]["operation"] == "double"
        data["correct_answer"]["y"] = 2 * data["params"]["x"]
    else:
        data["correct_answer"]["y"] = 3 * data["params"]["x"]

def prepare(data):
    # this function will run after all elements have run `generate()`
    # we can alter any of the element data here, but this is rarely needed

def parse(data):
    # data["raw_submitted_answer"] is a dictionary of exact raw answers submitted by the student
    # data["submitted_answer"] is the answers parsed by elements (e.g., strings converted to numbers)
    # data["format_errors"] is a dictionary of answer format errors (if any) from elements
    # we can modify or delete data["format_errors"] if we have custom logic (this is rarely needed)
    # if there are format errors then the submission is "invalid" and is not graded

    # as an example, we will reject negative numbers for "y"
    if "y" not in data["format_errors"]: # check we don't already have a format error
        if data["submitted_answers"]["y"] < 0
            data["submitted_answers"]["y"] = "Negative numbers are not allowed"

def grade(data):
    # all elements will have already graded their answers (if any) before this point
    # the individual element scores are in data["partial_scores"][NAME]
    # the total score is in data["score"]
    # all scores are between 0 and 1
    # we can modify or delete any of these if we have a custom grading method
    # this function only runs if `parse()` did not produce format errors, so we can assume all data is valid

    # as an example, we will erase all partial scores and automatically grade as correct:
    data["partial_scores"] = {}
    data["score"] = 1
```

doc/devElements.md
mwest1066: That's correct, they are all unmodifiable. The code is `freeform.js` is the ultimate authority on this, because it actually enforces these restrictions 😄

doc/question.md
mwest1066: We've actually been meaning to slightly change `test()`. See #2269. Maybe we should not document this until we've made the change? No one is using `test()` at the moment, as far as I'm aware.

doc/question.md
```diff
@@ -110,6 +88,78 @@ The `question.html` is regular HTML, with four special features:
 
 4. LaTeX equations are available within HTML by using `$x^2$` for inline equations, and `$$x^2$$` or `\[x^2\]` for display equations.
 
+## Question `server.py`
```
tdy, line 91: Moved the `server.py` section below the `question.html` section since the `server.py` skeleton includes a `question.html`.

doc/question.md
```diff
@@ -110,6 +88,78 @@ The `question.html` is regular HTML, with four special features:
...
+`parse()` | | `format_errors`, `submitted_answers` | `correct_answers`, `options`, `params`, `raw_submitted_answers`, `variant_seed` | Parse the `data["submitted_answers"][var]` data entered by the student, modifying this variable. Return the modified `data` dictionary.
+`grade()` | | `correct_answers`, `feedback`, `format_errors`, `params`, `partial_scores`, `score`, `submitted_answers` | `options`, `raw_submitted_answers`, `variant_seed` | Grade `data["submitted_answers"][var]` to determine a score. Store the score and any feedback in `data["partial_scores"][var]["score"]` and `data["partial_scores"][var]["feedback"]`. Return the modified `data` dictionary.
+`file()` | `object` (string, byte-like, file-like) | | `correct_answers`, `filename`, `options`, `params`, `variant_seed` | Generate a file object dynamically in lieu of a physical file. Trigger via `type="dynamic"` in the question element (e.g., `pl-figure`, `pl-file-download`). Access the requested filename via `data['filename']`. If `file()` returns nothing, an empty string will be used.
```
tdy, line 102: Would love to include a `file()` example, but doesn't really fit with the current skeleton. Is it reasonable to make a second skeleton example demo'ing `file()` or is that too much?

tdy: Or maybe I will just add comments to `exampleCourse/questions/demoRandomPlot`

doc/question.md
```diff
@@ -98,28 +101,64 @@ Function | Return object | modifiable `data` keys | unmodifiable `data` keys | D
...
 
-The `question.html` is regular HTML, with four special features:
-
-1. Any text in double-curly-braces (like `{{params.m}}`) is substituted with variable values. If you use triple-braces (like `{{{params.html}}}`) then raw HTML is substituted (don't use this unless you know you need it). This is using [Mustache](https://mustache.github.io/mustache.5.html) templating.
-
-2. Special HTML elements (like `<pl-number-input>`) enable input and formatted output. See the [list of PrairieLearn elements](elements.md).
-   
-3. A special `<markdown>` tag allows you to write Markdown inline in questions.
+```python
+# server.py
```
mwest1066, line 120: `import random`

exampleCourse/questions/demoRandomPlot/question.html
```diff
@@ -1,5 +1,6 @@
 <pl-question-panel>
-    <pl-figure file-name="figure.png" type="dynamic" width="500px"></pl-figure>
+    <!-- "my-mpl-figure" here is not a physical file. Since this pl-figure is the "dynamic" type, the figure is generated via server.py's `file()`. -->
+    <pl-figure file-name="my-mpl-figure" type="dynamic" width="500px"></pl-figure>
```
tdy, line 3: When I first tried to figure out what was going on here, I couldn't figure out where `figure.png` was. I kept thinking "both `question.html` and `server.py` keep referencing this `figure.png`, but I can't find it anywhere." Mariana ended up explaining how it worked in OH.

I think it's easier for new users to understand if the dynamic figure `file-name` does not look like a physical file, but I can revert this you think commenting is sufficient.

doc/question.md
```diff
@@ -98,28 +101,64 @@ Function | Return object | modifiable `data` keys | unmodifiable `data` keys | D
...
+    data["params"]["x"] = random.randint(5, 10)
+    data["params"]["operation"] = random.choice(["double", "triple"])
+
+    # Also compute the correct answer (if there is one) and store in the data["correct_answer"] dict:
+    if data["params"]["operation"] == "double"
+        data["correct_answer"]["y"] = 2 * data["params"]["x"]
+    else:
+        data["correct_answer"]["y"] = 3 * data["params"]["x"]
+
+    def prepare(data):
```
mwest1066, line 133: Something happened to the indenting.

doc/question.md
coatless, line 120:
```suggestion
# server.py
import random
```

doc/question.md
mwest1066: I agree that adding `file()` to the current skeleton wouldn't make much sense, but I think it would be fine to include a second complete example in a separate section. Perhaps something like:

`question.html`
```
<p>Here is a dynamically-rendered figure showing a line of slope $a = {{params.a}}$:</p>
<img src="{{options.client_files_question_dynamic_url}}/fig.png" />
```

`server.py`:
```
import random, io, matplotlib.pyplot as plt

def generate(data):
    data["params"]["a"] = random.choice([0.25, 0.5, 1, 2, 4])

def file(data):
    # We should look at data["filename"], generate the corresponding file,
    # and return the contents of the file as a string, bytes-like, or file-like object.
    # We can access data["params"] 
    # As an example, we will generate the figure "fig.png"

    if data["filename"] == "fig.png":                # check for the appropriate filename
        plt.plot([0, data["params"]["a"]], [0, 1])   # plot a line with slope "a"
        buf = io.BytesIO()                           # make a bytes object (a buffer)
        plt.savefig(buf, format="png")               # save the figure data into the buffer
        return buf
```

(I think the code above is correct, but I haven't tested it.)

exampleCourse/questions/demoRandomPlot/question.html
mwest1066: I definitely like the comment. One reason to prefer a filename that "looks real" is for use with `<pl-download>`, where a generated PDF file should be saved with filename extension `.pdf`, for example. I guess to me it makes sense to always use "real looking" filenames, but I acknowledge that dynamic files are just inherently tricky.

Maybe it's worth adding a special section on dynamic files, where you really clearly explain that the file never exists as a file on disk in the conventional sense, and that it's all "fake"?

doc/question.md
```diff
@@ -110,6 +88,81 @@ The `question.html` is regular HTML, with four special features:
...
+    # Also compute the correct answer (if there is one) and store in the data["correct_answers"] dict:
+    if data["params"]["operation"] == "double":
+        data["correct_answers"]["y"] = 2 * data["params"]["x"]
+    else:
+        data["correct_answers"]["y"] = 3 * data["params"]["x"]
+
+def prepare(data):
+    # This function will run after all elements have run `generate()`.
+    # We can alter any of the element data here, but this is rarely needed.
+    pass
```
tdy, line 138: I actually tested the skeleton this time, which caught several other errors as well (sorry, should've done this before). One of the errors was having an empty `prepare()`, so I added `pass`. Will this mess up the question in some invisible way (seems to work fine by my testing) if we override `prepare()` but make it empty?

mwest1066: It's fine to have the functions present and not doing anything. We aren't actually overriding anything here. The flow is "run all the element functions", "do some internal bookkeeping", "run the server.py" function if it exists. Everything is designed to run correctly without a "server.py" existing at all, and if a function exists but passes then it's the same as if it just wasn't there.

exampleCourse/questions/demoRandomPlot/question.html
tdy: Added a `question.md` subsection and slightly modified the `pl-figure` docs. Not sure if that's what you had in mind.

doc/question.md
```diff
@@ -110,6 +88,115 @@ The `question.html` is regular HTML, with four special features:
...
+`parse()` | | `format_errors`, `submitted_answers` | `correct_answers`, `options`, `params`, `raw_submitted_answers`, `variant_seed` | Parse the `data["submitted_answers"][var]` data entered by the student, modifying this variable. Return the modified `data` dictionary.
+`grade()` | | `correct_answers`, `feedback`, `format_errors`, `params`, `partial_scores`, `score`, `submitted_answers` | `options`, `raw_submitted_answers`, `variant_seed` | Grade `data["submitted_answers"][var]` to determine a score. Store the score and any feedback in `data["partial_scores"][var]["score"]` and `data["partial_scores"][var]["feedback"]`. Return the modified `data` dictionary.
+`file()` | `object` (string, byte-like, file-like) | | `correct_answers`, `filename`, `options`, `params`, `variant_seed` | Generate a file object dynamically in lieu of a physical file. Trigger via `type="dynamic"` in the question element (e.g., `pl-figure`, `pl-file-download`). Access the requested filename via `data['filename']`. If `file()` returns nothing, an empty string will be used.
```
tdy, line 102:
```suggestion
`file()` | `object` (string, bytes-like, file-like) | | `correct_answers`, `filename`, `options`, `params`, `variant_seed` | Generate a file object dynamically in lieu of a physical file. Trigger via `type="dynamic"` in the question element (e.g., `pl-figure`, `pl-file-download`). Access the requested filename via `data['filename']`. If `file()` returns nothing, an empty string will be used.
```

exampleCourse/questions/demoRandomPlot/question.html
mwest1066: Looks great!

doc/question.md
```diff
@@ -110,6 +88,115 @@ The `question.html` is regular HTML, with four special features:
...
+    # As an example, we will give half points for incorrect answers larger than "x":
+    if data["score"] == 0: # only if not already correct
+        if data["submitted_answers"]["y"] > data["params"]["x"]:
+            data["partial_scores"]["y"] = 0.5
+            data["score"] = 0.5
+```
+
+## Generating dynamic files
+
+You can dynamically generate file objects `server.py`. These files never appear physically on the disk. They are generated in `file()` and returned as strings, bytes-like objects, or file-like objects. A complete `question.html` and `server.py` example using a dynamically generated `fig.png` looks like:
```
tdy, line 168:
```suggestion
You can dynamically generate file objects in `server.py`. These files never appear physically on the disk. They are generated in `file()` and returned as strings, bytes-like objects, or file-like objects. A complete `question.html` and `server.py` example using a dynamically generated `fig.png` looks like:
```

doc/question.md
```diff
@@ -110,6 +88,115 @@ The `question.html` is regular HTML, with four special features:
...
+You can dynamically generate file objects in `server.py`. These files never appear physically on the disk. They are generated in `file()` and returned as strings, bytes-like objects, or file-like objects. A complete `question.html` and `server.py` example using a dynamically generated `fig.png` looks like:
+
+```html
+<!-- question.html -->
+
+<p>Here is a dynamically-rendered figure showing a line of slope $a = {{params.a}}$:</p>
+<img src="{{options.client_files_question_dynamic_url}}/fig.png" />
+```
+
+```python
```
tdy, line 177: FYI I did test `question.html` and `server.py` this time. Worked on my end.

mwest1066: Yay!

tdy:
Ok this one should be good to go now (from my end).

PR #2262, opened 2020-03-28
Add support for server-side live-reloading

tdy:
There were already fragments of `nodemon` lying around, but this PR makes it actually usable:

* [x] Add/fix `nodemon` support: `-e NODEMON=true`
* [x] Add docker command to docs
* [x] Clarify that live-reload is not default in PL (I guess I've been spoiled by React and assumed that live-reload was a given in node)

```
$ docker run -it --rm -p 3000:3000 -e NODEMON=true -v /path/to/PrairieLearn:/PrairieLearn prairielearn/prairielearn
```
```
waiting for server to start.... done
server started

> PrairieLearn@3.2.0 start-nodemon /PrairieLearn
> node_modules/nodemon/bin/nodemon.js -L server.js

[nodemon] 2.0.2
[nodemon] to restart at any time, enter `rs`
[nodemon] watching dir(s): *.*
[nodemon] watching extensions: js,mjs,json
[nodemon] starting `node server.js`
info: PrairieLearn server start
info: Running migration 167_assessments__require_honor_code__add.sql
info: Not using Redis for message passing
info: External grader running locally
info: PrairieLearn server ready, press Control-C to quit
info: Go to http://localhost:3000/pl
[nodemon] restarting due to changes...      <----- auto detect+restart when source changes
[nodemon] starting `node server.js`
info: PrairieLearn server start
info: Not using Redis for message passing
info: External grader running locally
info: PrairieLearn server ready, press Control-C to quit
info: Go to http://localhost:3000/pl
```

### DIFF
--- a/docker/init.sh
+++ b/docker/init.sh
@@ -14,7 +14,7 @@ else
     # Uncomment to start redis to test message passing
     # redis-server --daemonize yes
 
-    if [[ -n $NODEMON ]] && [[ $NODEMON == "true" ]]; then
+    if [[ $NODEMON == "true" ]]; then
         npm run start-nodemon
     else
         npm start
--- a/package.json
+++ b/package.json
@@ -114,7 +114,7 @@
     },
     "scripts": {
         "start": "node server.js",
-        "start-nodemon": "node_modules/nodemon/bin/nodemon.js -L server.js",
+        "start-nodemon": "npx nodemon -L server.js",
         "test": "nyc --reporter=lcov mocha tests/index.js",
         "test-nocoverage": "mocha tests/index.js",
         "lint-js": "eslint --ext js \"**/*.js\"",
--- a/package.json
+++ b/package.json
@@ -114,7 +114,7 @@
     },
     "scripts": {
         "start": "node server.js",
-        "start-nodemon": "nodemon server.js",
+        "start-nodemon": "node_modules/nodemon/bin/nodemon.js -L server.js",
         "test": "nyc --reporter=lcov mocha tests/index.js",
         "test-nocoverage": "mocha tests/index.js",
         "lint-js": "eslint --ext js \"**/*.js\"",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -114,7 +114,7 @@
     },
     "scripts": {
         "start": "node server.js",
-        "start-nodemon": "nodemon server.js",
+        "start-nodemon": "node_modules/nodemon/bin/nodemon.js -L server.js",
```
tdy: 1. spell out path since symlinks trip up docker
2. need legacy mode `-L` inside docker

docker/init.sh
```diff
@@ -14,7 +14,7 @@ else
     # Uncomment to start redis to test message passing
     # redis-server --daemonize yes
 
-    if [[ -n $NODEMON ]] && [[ $NODEMON == "true" ]]; then
+    if [[ $NODEMON == "true" ]]; then
```
coatless: Why did you get rid of the bash variable existence check?

tdy: Just removing redundancy since bash already evals `[[ $NODEMON == "true" ]]` to 1 if `NODEMON` is unset.

package.json
mwest1066: I think the "correct" way to run things from inside `node_modules` is to use `npx nodemon ...`? This runs things out of `node_modules/.bin` (note `.`) which is where "scripts" get installed.

mwest1066: Actually, commands in `scripts` inside `package.json` already have `node_modules/.bin` on the path, I believe, so we can just make this `nodemon ...`

tdy: `/PrairieLearn/node_modules/.bin` is in the path, but `nodemon ...` doesn't work for some reason. I think my conclusion about symlinks was a red herring -- there's just something wrong with the way `nodemon` is packaged given how many "nodemon not recognized" google hits there are.

tdy: `npx nodemon -L server.js` does work on my end.

trombonekenny: Seems like we could do better than `-L` but if the performance is good, I guess it doesn't matter?

mwest1066: Seems specifically recommended for containers with mounted source: https://www.npmjs.com/package/nodemon#application-isnt-restarting

PR #1920, opened 2019-12-16
Add more detailed tab titles to student homework pages

astrosticks:
In a page's `.js` file, overriding the value res.locals.htmlTitle now lets a page set a custom string that shows up in the browser's tab (the default is set to "Prairielearn" in `server.js`). 4b1ff82 provides a few examples where the student homework pages have richer tab titles, in order of (question number and name) (assessment) and (class). Closes #1863.

PR #2281, opened 2020-03-30
Adding more detailed html titles

### LINKED ISSUE #1863: Tab titles should reflect the subpage that is open
Currently, every single PrairieLearn page just has "PrairieLearn" as the name that shows up in tabs and window titles. This makes it so that users with multiple PrairieLearn tabs/windows open have to manually keep track of which tab has which page open. 

Here is an example of how my tabs appear to me when I do multiple pre-lectures for CS233 at once:

image

It would be nice to see the page titles contain information about the subpage that is open, formatted with information like the name of the problem that is open, the associated course, and possibly the word "PrairieLearn".

astrosticks:
Closes #1863 by allowing pages to set an array of strings `res.locals.navTrace`, which will appear in browser tabs. `navTrace` is combined with the short name of the current course (if available, e.g. XC 101), and "PrairieLearn" to form the final html title.

b2e0aea implements the following examples: 
image


astrosticks:
I made this a separate pull request from #1920 because I ended up rewriting most of what I had, and some mistakes I made with git a while ago ended up making it impossible to merge anyway.

pages/partials/title.ejs
```diff
@@ -0,0 +1,19 @@
+<%
+    var navTrace = new Array();
```
mwest1066:
```suggestion
    const navTrace = [];
```

pages/partials/title.ejs
```diff
@@ -0,0 +1,19 @@
...
+
+    if (typeof pageTitle !== 'undefined') {
+        navTrace.push(pageTitle);
+    }
+    // If page is associated with a particular course,
+    // always display the short name, e.g. QA 101
+    if (typeof course !== 'undefined'
+    && typeof course.short_name !== 'undefined') {
```
mwest1066: Courses must have a `short_name` so we can skip this check.

pages/studentInstanceQuestionHomework/studentInstanceQuestionHomework.ejs
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <%- include('../partials/head'); %>
+    <%- include('../partials/head', {pageTitle: instance_question_info.question_number}); %>
```
mwest1066: Why is this different to the exam equivalent?

astrosticks: That's a big reason why it took so long to make a new PR for this, it's because `instance_question_info.question_number` is formatted completely differently for questions and for exams. Basically, homeworks format that value like "HW1.1" but exams will have it formatted like "1". I tried to fix the inconsistency myself but I couldn't really guarantee that it wouldn't break things elsewhere, so I decided to leave that out of this PR.

mwest1066: Sounds good!

mwest1066:
Would it be possible to also fix up the titles for all the admin and instructor pages? 😄

astrosticks:
> Would it be possible to also fix up the titles for all the admin and instructor pages? 😄

Possibly, but I might want to choosing those titles for somebody experienced with the instructor views. I'm not a model user for that group :P

pages/studentAssessments/studentAssessments.ejs
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <%- include('../partials/head'); %>
+    <%- include('../partials/head', {pageTitle: "Assessments"}); %>
```
astrosticks: Style: Single-quotes

mwest1066:
> Possibly, but I might want to leave choosing those titles for somebody experienced with the instructor views. I'm not a model user for that group!

It would be awesome if you could take your best go at this. Others can provide feedback in code review once you've done a first pass. As an initial suggestion:

* `pages/instructorAssessment*` could have titles like `HW1 (Access) - XC 101, Sp15 - PrairieLearn`.
* `pages/instructorCourseAdmin*` could have titles like `XC 101 (Access) - PrairieLearn`
* `pages/instructorFile*` could have titles like `FILENAME - XC 101, Sp15 - PrairieLearn`
* `pages/instructorInstanceAdmin*` could have titles like `XC 101, Sp 15 (Access) - PrairieLearn`
* `pages/instructorQuestion*` could have titles like `demoCalculation - XC 101, Sp15 - PrairieLearn`
* other pages could be `PAGENAME - XC 101, Sp15 - PrairieLearn`

As you can see, most instructor pages have a two-level hierarchy. For example, the `instructorAssessment*` pages are all about an assessment. I suggest using `ASSESSMENT_NAME (PAGE_NAME)` as the general way of representing this two-level hierarchy.

Instructor pages will always have a `course` variable and they might have a `course_instance` variable. If they only have `course` then we can use `course.short_name`. If they have both then I suggest `course.short_name, course_instance.short_name`.

All of the above is just a suggestion. Feel free to experiment with other naming conventions!

pages/partials/title.ejs
astrosticks: Fixed in de59867

mwest1066:
Would it be possible to make the page titles reflect the nav hierarchy? Using `file_view` as an example, consider "Files" tabs for different navbar sections in the PR at the moment:

1. `XC 101 -> Files` (while a course instance _is not_ active) - title is `Files - XC 101`
2. `XC 101 -> Files` (while a course instance _is_ active) - title is `Files - XC 101, Sp15`
3. `Sp15 -> Files` - title is `Files - XC 101, Sp15`
4. `XC 101 -> Questions -> [Specific Question] -> Files` - title is `Files - XC 101` (when a course instance is not active) or `Files - XC 101, Sp15` (when a course instance is active)
5. `HW1 -> Files` - title is `Files - XC 101, Sp15`

Ideally, pages 1 and 2 above would have the same title (they are the "same page" from the user's point of view) and pages 4 and 5 would have titles that reflect the current question or assessment.

I think the "right" way to do this is to use the `navPage` and `navSubPage` variables to decide the format of the title. For example, if `navPage = 'question'` and `navSubPage = 'file_view'` then the title would look like `Files - [QUESTION] - XC 101`, whereas if `navPage = 'assessment'` and `navSubPage = 'file_view'` then title would be `Files - [ASSESSMENT] - XC 101, Sp15`.

astrosticks:
Yeah, maybe that would be simpler, but I'm thinking about the cases where the `navSubPage` becomes the new `navPage`: On a course instance page, I can have the `assessment` tab open as my "subpage", but then `assessment` would become the new main page when I click into it, and it would have its own subpage. They also might not be verbose enough (there might need to be an extra flag for instructor vs. student views). So a nav refactor should probably not try to hard-code every possible combination of navPage and navSubPage values to a title format, but set a full description for the navigation hierarchy as paths are requested in `server.js`. For now, this PR doesnt really make any assumptions about what the navigation hierarchy *should* be, but just wraps around the existing system.

mwest1066:
I'm not certain that I understand all of your comment above. Just make sure we are on the same page, I'm not proposing to change the nav hierarchy as part of this PR.

The thing that was mainly bothering me was that quite different pages can have the same title in the current PR. For example, the "Files" pages. Using `navPage` and `navSubpage` to (largely) determine the page title seemed like an easy fix, but I'm happy with other solutions as well.

Did I respond to what you were getting at in your comment? Sorry if I missed what you were saying.

astrosticks:
I think I found a solution that makes the most sense. The script that creates the HTML titles will now try to generate part of the title using the `navSubPage` and `navPage` values. If a page's `navSubPage` isn't descriptive enough, or doesn't format nicely, an individual page can override it by setting `pageTitle`. Assessment and question pages will now automatically insert their abbreviated names into the title, instead of needing to be declared per-page. `navSubPage` and `navPage` values are now declared in `server.js`, too.

astrosticks:
Oh, also, I didn't set `navPage` and `navSubPage` values for `administrator/overview` because of my other pull that heavily edits that page, so once the two of them are merged, I'll make sure they work together.

pages/partials/title.ejs
```diff
@@ -0,0 +1,74 @@
...
+        if (assessment.type == "Homework") {
+          navTrace.push(instance_question_info.question_number);
+        } else if (assessment.type == "Exam") {
+          console.log({
+            "assessment": assessment,
+            "assessment_set": assessment_set,
+            "question": question,
+            "instance_question": instance_question,
+            "instance_question_info": instance_question_info
+          });
```
nicknytko, line 44: Can we remove this `console.log()`?

astrosticks:
> Code looks fine on my end, just one tiny `console.log()` that we can probably remove.

Done!

PR #2282, opened 2020-03-30
Fix logger.error() calls without a string argument

mwest1066:
Calls to `logger.error()` must have a string as the first argument to log correctly. This PR fixes all the calls that were missing this.

### DIFF
--- a/middlewares/authnToken.js
+++ b/middlewares/authnToken.js
@@ -46,7 +46,7 @@ module.exports = (req, res, next) => {
                 token_id: result.rows[0].token_id,
             };
             sqldb.query(sql.update_token_last_used, lastUsedParams, (err) => {
-                if (ERR(err, (e) => logger.error('Error in update_token_last_used', e)));
+                if (ERR(err, (e) => logger.error('Error in sql.update_token_last_used', e)));
             });
         }
     });
--- a/middlewares/authnToken.js
+++ b/middlewares/authnToken.js
@@ -46,7 +46,7 @@ module.exports = (req, res, next) => {
                 token_id: result.rows[0].token_id,
             };
             sqldb.query(sql.update_token_last_used, lastUsedParams, (err) => {
-                if (ERR(err, (e) => logger.error(e)));
+                if (ERR(err, (e) => logger.error('Error in update_token_last_used', e)));
             });
         }
     });
--- a/question-servers/freeform.js
+++ b/question-servers/freeform.js
@@ -845,7 +845,7 @@ module.exports = {
                 cache.get(cacheKey, (err, cachedData) => {
                     // We don't actually want to fail if the cache has an error; we'll
                     // just render the panel as normal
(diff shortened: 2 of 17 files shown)


middlewares/authnToken.js
```diff
@@ -46,7 +46,7 @@ module.exports = (req, res, next) => {
                 token_id: result.rows[0].token_id,
             };
             sqldb.query(sql.update_token_last_used, lastUsedParams, (err) => {
-                if (ERR(err, (e) => logger.error(e)));
+                if (ERR(err, (e) => logger.error('Error in update_token_last_used', e)));
```
tdy:
```suggestion
                if (ERR(err, (e) => logger.error('Error in sql.update_token_last_used', e)));
```
Originally suggested `query()`, but seems like this deviation was intended since I saw it again below. Maybe add `sql.`?

question-servers/freeform.js
```diff
@@ -845,7 +845,7 @@ module.exports = {
                 cache.get(cacheKey, (err, cachedData) => {
                     // We don't actually want to fail if the cache has an error; we'll
                     // just render the panel as normal
-                    ERR(err, (e) => logger.error(e));
+                    ERR(err, (e) => logger.error('Error in cache.get()', e));
```
tdy:
```suggestion
                    ERR(err, (e) => logger.error('Error in get()', e));
```
Not sure if intentional? `get()` is too generic I guess?

webhooks/grading/grading.js
```diff
@@ -36,7 +36,7 @@ router.post('/', function(req, res, next) {
         };
 
         sqldb.queryOneRow(sql.update_grading_received_time, params, (err, _result) => {
-            if (ERR(err, (err) => logger.error(err))) return;
+            if (ERR(err, (err) => logger.error('Error in update_grading_received_time', err))) return;
```
tdy:
```suggestion
            if (ERR(err, (err) => logger.error('Error in sql.update_grading_received_time', err))) return;
```

tools/external-grading-load-test/main.js
```diff
@@ -96,8 +96,7 @@ async.series([
                 return next(null);
             });
             request.on('error', (err) => {
-                logger.error(`Error submitting job ${n}!`);
-                logger.error(err);
+                logger.error(`Error submitting job ${n}!`, err);
```
tdy:
```suggestion
                logger.error(`Error submitting job ${n}`, err);
```
just for consistency -- feel free to disregard though

question-servers/freeform.js
mwest1066: Yep, `get()` just felt super-generic. The exact error string doesn't really matter here too much because it's the stack trace that we really want.

tools/external-grading-load-test/main.js
mwest1066: Oh definitely, I'm not a fan of exclamation marks. Thanks!

PR #2288, opened 2020-03-31
Document how to connect to existing container

tdy:
Expands `installingLocal.md` to show how to open a shell in a running container.


docs/installingLocal.md
```diff
@@ -83,3 +83,21 @@ cd /PrairieLearn
 ./docker/test_js.sh
 ./docker/test_python.sh
 ```
+
+### Connecting to an Existing Docker
```
mwest1066:
```suggestion
### Connecting to an existing Docker container
```

We've aren't totally consistent, but we are trying to standardize on sentence case for titles (leading caps only).

docs/installingLocal.md
```diff
@@ -83,3 +83,21 @@ cd /PrairieLearn
...
+docker ps --format '{{.Image}}: {{.Names}}'
+```
+
+This will output your running container(s) in the form `<container_image>: <container_name>`. Find the `prairielearn/prairielearn: <container_name>` line and copy its `<container_name>` (e.g., `adoring_gauss`).
+
+2. Open a shell in `<container_name>`:
+
+```sh
+docker exec -it <container_name> /bin/bash
+```
```
mwest1066: ### Using tmux in a container

The PrairieLearn Docker images are built with `tmux` installed. If you start a container with a shell then you can first run `tmux` before running other commands.

Tmux creates virtual windows which run simultaneously (you only see one window at a time). Tmux is controlled by typing a `Ctrl-b` and then another key. The basic commands are:

* `Ctrl-b` `c` - create a new window
* `Ctrl-b` `0` - switch to window number 0 (also `Ctrl-b` `1` switches to window 1, etc.)
* `Ctrl-b` `d` - detaches from tmux back to the original shell, which you can exit to terminate the container

Google `tmux` for tutorials that demonstrate many more capabilities.

mwest1066: Sorry, I see now that I wrote this as a comment, I meant to propose it as a change.
=

coatless: @mwest1066 fixed that for you.

Future reference, add an addition backtick, e.g. 

`````md
Extra backtick
````suggestion

```
Documentation
```

````

Comment (mwest1066) [reply docs/installingLocal.md:103]: Ahhhh! Many thanks.

File: docs/installingLocal.md [hunk 3]
```diff
@@ -83,3 +83,21 @@ cd /PrairieLearn
...
 ```
+
+### Connecting to an existing Docker container
+
+The previous shells were launched in their own containers. If you want to open a shell in a Docker container that is *already running*, you can find the container's name and connect to it.
+
+1. Find the name of your running PrairieLearn container:
+
+```sh
+docker ps --format '{{.Image}}: {{.Names}}'
```
tdy, line 94: Not so sure about this line anymore. At the time, I was worried about `docker ps` output having too much gibberish, so this would give users just a simple `prairielearn/prairielearn: adoring_gauss` output. BUT `docker ps --format '{{.Image}}: {{.Names}}'` is much harder to memorize, so it's probably worth just documenting `docker ps`.

mwest1066: Agreed

docs/installingLocal.md
```diff
@@ -83,3 +83,41 @@ cd /PrairieLearn
...
+```
+
+which will output multiple columns of information about your running container(s). Look for the `prairielearn/prairielearn` image and copy its corresponding name. For example, the name of the PrairieLearn container in this `docker ps` output is `upbeat_roentgen`:
+
+```
+CONTAINER ID  IMAGE                      COMMAND              CREATED      STATUS      PORTS                   NAMES
+4be50d160d49  catracker-devcontainer     "sleep infinity"     5 hours ago  Up 5 hours                          determined_cori
+c681714bea9a  catracker-postgres         "docker-entrypoin…"  5 hours ago  Up 5 hours  5432/tcp                tender_wu
+e0f522f41ea4  prairielearn/prairielearn  "/bin/sh -c /Prai…"  2 hours ago  Up 2 hours  0.0.0.0:3000->3000/tcp  upbeat_roentgen
+```
```
tdy, line 104: Too busy? Maybe most users will only see 1 line.

docs/installingLocal.md
```diff
@@ -83,3 +83,43 @@ cd /PrairieLearn
...
```
tdy, line 104: Too busy? Maybe most users will only see 1 line.

mwest1066: I agree it seems unnecessarily busy. I think it's probably best to go with the minimal example that most users will see (single PL container). Users who are running multiple containers can presumably figure things out 😄

PR #2289, opened 2020-03-31
Moving and restyling navtabs

### LINKED ISSUE #2066: Tabs should be outside of content?
The horizontal line under the list of tabs does not extend all the way across the window. I think this is because the tab bar is inside `<div id="content" class="container-fluid">` rather than above it. It seems to me that being above content would be correct?

astrosticks:
Closes #2066 but also refactors navtabs so setting the style is much less work, too. Also refactors code style in navbars to be easier to read.

| Changes summary |
| -------|
| **Before** |
| pl-navtabs-before |
| **After** |
| pl-navtabs-after |

Not sure if I'd keep 19c9a34, since this is what the gray background looks like on skinnier screens: 

pl-navtabs-caveat
Although, recreating this in the current version of PrairieLearn shows that this isn't really any worse than what's already implemented.


pages/partials/navbar.ejs
```diff
@@ -1,126 +1,114 @@
...
-
-<ul class="nav navbar-nav" id="username-nav">
-
-    <% if (!locals.effective_user_has_no_access_to_course_instance) { %>
-    <li class="nav-item dropdown mb-2 mb-md-0 mr-2 <% if (navPage == "effective") { %>active<% } %>">
-        <a class="nav-link dropdown-toggle" href="#" id="navbarDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-            <%= displayedName %>
-            <% if (typeof news_item_notification_count != 'undefined' && news_item_notification_count > 0) { %>
-              <span class="badge badge-pill badge-primary news-item-count"><%= news_item_notification_count %></span>
+    <!-- Home link -->
```
astrosticks: This commit is entirely just balancing some messy indentation and comments in the navbar files.

mwest1066:
I love this PR! I really want to merge this.

However, there are a lot of whitespace changes here that don't seem important for the actual change. This makes it hard to review, even when using `git diff -b` to try and ignore whitespace. Would it be possible to make a version of this PR that doesn't change whitespace unnecessarily?

astrosticks:
Sure thing. Will close this PR if it doesn't work out.

PR #2290, opened 2020-03-31
Add padding to table rows to match bootstrap card headers

### LINKED ISSUE #2090: Text misaligned on tables throughout UI
misaligned
I see this misaligned text on student course instance views, as well as the instructor view and most other tables on the site.

astrosticks:
Fixes #2090. I just copied the values that Bootstrap uses to set horizontal padding in card headers so that the sides of tables are aligned. 

| Before | After |
| ---- | ---- |
| pl-tables-before | pl-tables-after |

### DIFF
--- a/public/stylesheets/local.css
+++ b/public/stylesheets/local.css
@@ -186,10 +186,10 @@ table td > .badge {
 }
 
 /* Used to give table rows the same padding as Bootstrap card headers */
-.card table tr td:first-child, .card table tr th {
+.card table tr td:first-child, .card table tr th:first-child {
     padding-left: 1.25rem;
 }
-.card table tr td:last-child, .card table tr th {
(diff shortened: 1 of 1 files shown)


mwest1066:
Could we use https://www.npmjs.com/package/bootstrap4-card-tables?

nicknytko:
> Could we use https://www.npmjs.com/package/bootstrap4-card-tables?

+1, pulling in a new package is probably okay here

astrosticks:
Sounds good to me and more dependable. If someone wants to show me on Slack how to get this package, I'd love to learn.

astrosticks:
Checking out the package that Nicolas linked, it assumes that card table padding is only wanted in tables without the `table-sm` class, which all of ours do, so it won't work for us.

Issue #2003, opened 2020-01-24
Consider highlighting navbar course when on course pages


mwest1066:
To show more clearly where we are in the navigation hierarchy, I wonder whether we should always highlight the navbar course tab ("TAM 212") when on any of the course pages (e.g., "Questions", "Issues"). At the moment, if I'm on a page like "Course Instances" and I click the "Questions" tab (not the top navbar item, but the tab) then the navbar changes, which is a bit weird. Also, when I'm on a course instance page and I click "Questions", then I jump to the course tabs, but it may not be clear why this happened because the course itself is not highlighted.

I guess I'm advocating here for a more strictly hierarchical nav, where it's always clear exactly where we are in the hierarchy (on all levels, both top navbar and the tabbar).

When on a page like "Questions" that has a shortcut navbar item, I'm not sure whether we should highlight both "TAM 212" and "Questions" or only "TAM 212".

mwest1066:
Yep! Good call, this is implemented by the new nav.

Issue #2263, opened 2020-03-28
Instructor assessment instance view is missing tab-bar


mwest1066:
Clicking "Details" next to an assessment on the "Students" tab takes you to `pages/instructorAssessmentInstance`, but the tab-bar vanishes when you go there. I guess it should still stay with "Students" as the active tab.

ZacWarham:
@mwest1066 is this just changing this

https://github.com/PrairieLearn/PrairieLearn/blob/dd6a70f3fef224645b0fe4befae8f90167aa91ec/apps/prairielearn/src/lib/navPageTabs.ts#L321

To `activeSubPage: ['instances', 'assessment_instance'],`  or am I misunderstanding?

![sample](https://github.com/user-attachments/assets/1175b341-9902-41fe-8126-b3839e4a0931)

mwest1066:
Maybe? CC @miguelaenlle because it might be best to fix this as part of the new nav experience. In any case, @miguelaenlle is the current expert on nav!

miguelaenlle:
Yes, that will resolve the issue! The issue is also present in enhanced nav:

![Image](https://github.com/user-attachments/assets/7208c158-c94c-462a-a18d-ce1d296bef8d)

It is also fixed by changing `activeSubPage: 'instances',` to `activeSubPage: ['instances', 'assessment_instance']`

![Image](https://github.com/user-attachments/assets/80405190-32fb-4385-81fc-3bbbc871f93a)

ZacWarham:
@miguelaenlle - is it worth me adding a PR for this or will you bundle it in your changes?

miguelaenlle:
> [@miguelaenlle](https://github.com/miguelaenlle) - is it worth me adding a PR for this or will you bundle it in your changes?

Hi @ZacWarham . I think it's okay to have this as a separate PR, since I'll be releasing individual feature PRs for each navigation system improvement.

Issue #2014, opened 2020-01-30
Navbar highlighting is wrong when on course question preview


mwest1066:
When on a course question preview page, the navbar has "PrairieLearn" highlighted rather than "[Rubric]" or "Questions".

See also #2003

PR #2292, opened 2020-03-31
Moving and restyling navtabs

### LINKED ISSUE #2066: Tabs should be outside of content?

astrosticks:
Because of readability issues in #2289 I made a new branch for this without whitespace edits. Functionally, everything should be exactly the same. 
___
Closes #2066 but also refactors navtabs so setting the style is much less work, too. 

| Changes summary |
| -------|
| **Before** |
| pl-navtabs-before |
| **After** |
| pl-navtabs-after |

A slight caveat, but this is what these changes look like on thin screens:

pl-navtabs-caveat

This problem is already present in the current version of PL but this change just draws attention to it.


mwest1066:
Ah!! Now I see what's going on! Many thanks.

pages/partials/navbar.ejs
```diff
@@ -124,3 +125,18 @@ if (typeof authz_data !== 'undefined' && authz_data.has_instructor_view) {
...
+const navPagesTabs = {
+    instance_admin: 'instructorInstanceAdminTabs',
+    course_admin: 'instructorCourseAdminTabs',
+    assessment: 'instructorAssessmentTabs',
+    question: 'instructorQuestionTabs'
+};
+// If navPage matches any of the keys in navPagesTabs, include the page given by its value
+if (typeof navPage != 'undefined' && Object.keys(navPagesTabs).includes(navPage)) { %>
+    <!-- Instructor navigation tabs -->
+    <ul class="nav nav-tabs pt-2 px-3 bg-light">
```
coatless: I think this is missing an `id=...` attribute. 

e.g. `id="navtabsInstructorQuestion"`

Not sure if this is needed though.

astrosticks: I did a scan through the codebase for the IDs that each of the instructor navtabs files specified, but didn't find any reference to them anywhere else, so I thought it best to delete them.

tdy:
Haven't reviewed the markup yet, but regarding the tab colors:

I'm inclined to support the original tab color scheme. The inactive tabs are what we actually click (and thus more intuitive to be styled as links). The active tab already has significant visual indication that it's active, so it doesn't seem meaningful to add even more `primary` styling.

astrosticks:
I'll leave the styling options to the reviewers since it's a fairly straightforward change at the bottom of `navbar.ejs` and `navtab.ejs`

mwest1066:
@astrosticks would it make sense to combine #2281 into this PR? I feel like they are both making very similar changes when I was reviewing them. I haven't thought very deeply about combining, so feel free to tell me that it's not a good idea 😄

astrosticks:
@mwest1066 I agree that they do very similar things and it would be far neater to have a well-designed system that guarantees that the right navigation information gets sent to each page. After trying to do that initially, I found so many places in the codebase that could use a refactor that I decided to go with this implementation that requires minimal overhead. I can try to compile a list of all of these refactors needed into an issue sometime.

coatless:
> I found so many places in the codebase that could use a refactor that I decided to go with this implementation that requires minimal overhead. **I can try to compile a list of all of these refactors needed into an issue sometime.**

Please do so. I'm not sure if you're following `#pl-dev`, but this has come up a bit over the past few days.

coatless:
@mwest1066 let's keep these two PRs separate for right now. I think that allow for this to go in smoother and quicker than combining with the dynamic / breadcrumb page titles.

pages/partials/navbar.ejs
```diff
@@ -124,3 +125,18 @@ if (typeof authz_data !== 'undefined' && authz_data.has_instructor_view) {
...
 </nav>
+<% // Mapping navPage to navtab sets
+const navPagesTabs = {
+    instance_admin: 'instructorInstanceAdminTabs',
+    course_admin: 'instructorCourseAdminTabs',
+    assessment: 'instructorAssessmentTabs',
+    question: 'instructorQuestionTabs'
+};
+// If navPage matches any of the keys in navPagesTabs, include the page given by its value
+if (typeof navPage != 'undefined' && Object.keys(navPagesTabs).includes(navPage)) { %>
```
mwest1066, line 136:
```suggestion
if (typeof navPage != 'undefined' && Object.hasOwnProperty(navPagesTabs, navPage)) { %>
```

It doesn't really matter here, but for large objects it's must faster to directly test whether they have a property (using `hasOwnProperty()`) rather than converting their keys to a list and doing an O(N) search with `includes()`.

pages/partials/navbar.ejs
```diff
@@ -124,3 +125,18 @@ if (typeof authz_data !== 'undefined' && authz_data.has_instructor_view) {
 </ul>
 </div>
 </nav>
+<% // Mapping navPage to navtab sets
+const navPagesTabs = {
+    instance_admin: 'instructorInstanceAdminTabs',
+    course_admin: 'instructorCourseAdminTabs',
+    assessment: 'instructorAssessmentTabs',
+    question: 'instructorQuestionTabs'
```
mwest1066, line 133: Can we rename all these files so they have a common prefix? E.g., `navTabsInstructorInstanceAdmin`, `navTabsInstructorQuestion`, etc. I find this makes the directory listings easier to navigate because all the "navTab" file are next to each other.

To bikeshed the naming scheme a bit: are we calling these `tabs`, `navtabs`, or `navTabs`? I don't have a strong opinion, but consistency would be good.

mwest1066:
Would it be possible to also fix #2003, #2014, and #2263 as part of this PR?

pages/partials/navtab.ejs
```diff
@@ -0,0 +1,3 @@
+<li class="nav-item">
+    <a class="nav-link <% if (navSubPage == activeSubPage) { %>active text-primary<% } else { %>text-secondary<% } %>" href="<%= urlPrefix %><%= urlSuffix %>"><i class="<%= iconClasses %>"></i> <%= tabLabel %></a>
```
tdy, line 2:
```suggestion
    <a class="nav-link <% if (navSubPage == activeSubPage) { %>active text-dark<% } else { %>text-secondary<% } %>" href="<%= urlPrefix %><%= urlSuffix %>"><i class="<%= iconClasses %>"></i> <%= tabLabel %></a>
```
So I mentioned earlier that I'm not a fan of making the `.active` text blue since it's too much of a <a href="#">**click-me**</a> UX pattern.

In the spirit of mimicking github, what about making it `text-dark` and customizing `border-top-*` via `public/stylesheets/local.css`:
```
.nav-tabs .nav-link.active {
    border-top-color: var(--primary);
    border-top-width: 0.2rem;
}
```
(or ideally using only bootstrap border utils, but I couldn't get it to work since navtabs leverage transparency: https://getbootstrap.com/docs/4.1/utilities/borders/)

Just my 2c. I won't belabor this anymore if @mwest1066 disagrees 😃 

image

mwest1066: Very nice! I agree that it's bad to use blue for text that is not a link.

That blue bar at the top looks great. Has it resulted in the text "Preview" shifting downwards slightly? The text baseline looks lower than that in the inactive tabs.

astrosticks: I like it a lot too! @mwest1066 the fix would be to add another CSS rule for the inactive tabs that gives them the same tab-border but fully transparent

mwest1066: Works for me. I prefer to avoid custom CSS and use vanilla Bootstrap, but this clearly seems like an area where the Bootstrap defaults are not good and we need to do it by hand.

tdy: Could also switch to `px` and use a negative `margin-top` so we only add 1 custom css rule without modifying inactive tabs:
```
.nav-tabs .nav-link.active {
    border-top-color: var(--primary);
    border-top-width: 3px;
    margin-top: -2px; /* realign text with inactive tabs (give back 1px border-top) */
}
```

tdy: Look like:

image

mwest1066: Extra nice!!

astrosticks:
Looks like @tdy and I stepped on eachother's toes because I did almost the same thing. Tim's implementation didn't make active and inactive tabs the same height, but mine does. My text baselines are slightly misaligned but it's a sub-pixel difference.

image
image 

I also added a new class for navtabs so that the new rule doesn't affect all bootstrap nav-tabs, just ones where it's explicitly requested.

astrosticks:
@mwest1066 I know it'd be nice to fix more nav issues in one PR, but the scope of this PR was just to fix #2066 and add a few style improvements. I'd like to get to those issues but I'd like to do a separate PR for refactoring and revising navigation in general like @coatless has suggested

pages/partials/navbar.ejs
tdy: > To bikeshed the naming scheme a bit: are we calling these tabs, navtabs, or navTabs?

Bootstrap calls them `.nav-tabs`, so maybe `navTabs*`?

astrosticks: I can get that into a follow up refactor

mwest1066: Can you just do it now? I'm hesitant to merge code that I know needs immediate cosmetic improvements.

astrosticks: Sure thing. What would you think about putting them inside a folder navTabs instead of the common prefix?

mwest1066: Sounds good! I think there were also internal variable names that should be changed to match? `navtab` -> `navTab`? I haven't thought about them all other than the desire for consistency 😄 

In cases where we use a mix of kebab-case, snake_case, and camelCase, we try and keep the word boundaries the same. So `instance_questions`, `instanceQuestions`, `instance-questions`, for example.

astrosticks: Done!

mwest1066:
I love the way the new tab bar looks! So much clearer and more elegant. And a code cleanup too 🎉

PR #2342, opened 2020-04-03
Escape special HTML characters in code snippets

### LINKED ISSUE #1098: <= in Question HTML Breaks Rendering
Something in PL seems to be interpreting a `<=` in `question.html` as part of a tag. Unfortunately this breaks rendering of many code snippets that include a `<=`.

Example `question.html`:
````html
<pl_question_panel>                                                                                                                                                                  
  <pre>                                                                                                                                                                              
    <code>                                                                                                                                                                           
for (int i = 0; i <= 10; i++);                                                                                                                                                       
    </code>                                                                                                                                                                          
  </pre>                                                                                                                                                                             
  <p>Is the question above broken?</p>                                                                                                                                               
    <pl_multiple_choice weight="1" answers_name="Sp18_Broken_0" number_answers="2">                                                                                                  
        <pl_answer correct="true">Yes</pl_answer>                                                                                                                                    
        <pl_answer correct="false">No</pl_answer>                                                                                                                                    
    </pl_multiple_choice>                                                                                                                                                            
</pl_question_panel> 
````
Result:

image

If you replace the `<=` with a `<`, all is well:

````html
<pl_question_panel>                                                                                                                                                                  
  <pre>                                                                                                                                                                              
    <code>                                                                                                                                                                           
for (int i = 0; i < 10; i++);                                                                                                                                                       
    </code>                                                                                                                                                                          
  </pre>                                                                                                                                                                             
  <p>Is the question above broken?</p>                                                                                                                                               
    <pl_multiple_choice weight="1" answers_name="Sp18_Broken_0" number_answers="2">                                                                                                  
        <pl_answer correct="true">Yes</pl_answer>                                                                                                                                    
        <pl_answer correct="false">No</pl_answer>                                                                                                                                    
    </pl_multiple_choice>                                                                                                                                                            
</pl_question_panel> 
````
image

[Here's a link to an example on the main PL site](https://prairielearn.engr.illinois.edu/pl/course_instance/8235/instructor/question/1541331/). Note that it's markup is a bit different since we preprocess things through HLJS, but I've verified that the minimal example above still triggers the problem.

I suspect that something is interpreting a `<=` as the start of a template tag. It seems like at minimum this could be disabled in `code` or `pre` blocks.

### PR DESCRIPTION
**Pull request by nicknytko**
Special characters (<, >, etc.) will be automatically escaped when enclosed in `<pl-code>` or `<code>` tags.  Fixes #1098.

The following

```
<p> Any special HTML characters (&lt;, &gt;, etc.) will be automatically escaped for you when using inline code. </p>
<pl-code language="cpp">
void loop() {
    for (int i = 0; i < 10; i++) {
        std::cout << i << std::endl;
    }
}
</pl-code>
```

will get rendered as:

image

### DIFF
--- a/exampleCourse/questions/elementCode/question.html
+++ b/exampleCourse/questions/elementCode/question.html
@@ -1,12 +1,11 @@
 <pl-question-panel>
-<markdown>Questions below were designed to showcase the different features of 
+<markdown>Questions below were designed to showcase the different features of
 [`pl-code`](http://prairielearn.readthedocs.io/en/latest/elements/#pl-code-element). 
  The goal of [`pl-code`](http://prairielearn.readthedocs.io/en/latest/elements/#pl-code-element)
 is to display different programming languages with appropriate syntax highlighting and code line emphasis.
 </markdown>
 </pl-question-panel>
 
-      
 <div class="card my-2">
 <div class="card-body">
 <pl-question-panel>
@@ -19,7 +18,22 @@ python with `language="python"`.
 def square(x):
     return x * x
 </pl-code>
-</pl-question-panel>                
+</pl-question-panel>
+</div>
+</div>
+
+<div class="card my-2">
+<div class="card-body">
+<pl-question-panel>
+<p> Any special HTML characters (&lt;, &gt;, etc.) will be automatically escaped for you when using inline code. </p>
+<pl-code language="cpp">
+void loop() {
+    for (int i = 0; i < 10; i++) {
+        std::cout << i << std::endl;
+    }
+}
+</pl-code>
+</pl-question-panel>
 </div>
 </div>
 
@@ -42,7 +56,7 @@ def square(x):
 def bar(x):
     return 'bar'
 </pl-code>
-</pl-question-panel>                
+</pl-question-panel>
 </div>
 </div>
 
--- a/exampleCourse/questions/elementMarkdown/question.html
+++ b/exampleCourse/questions/elementMarkdown/question.html
@@ -50,7 +50,7 @@ class Example {
     }
 
     Example(std::string name) {
-        stc::cout << "Hello, " << name << std::endl;
+        std::cout << "Hello, " << name << std::endl;
     }
 };
 ```
--- a/question-servers/freeform.js
+++ b/question-servers/freeform.js
@@ -8,6 +8,7 @@ const cheerio = require('cheerio');
 const hash = require('crypto').createHash;
 const parse5 = require('parse5');
 const debug = require('debug')('prairielearn:' + path.basename(__filename, '.js'));
+const he = require('he');
 
 const schemas = require('../schemas');
 const logger = require('../lib/logger');
@@ -285,6 +286,24 @@ module.exports = {
         });
     },
 
+    _escapeCodeSnippets: function(html) {
+        /* Escape special characters inside of code snippets. Specifically
+           handles <pl-code> and <code> tags.  We run this before the markdown
+           renderer because that will automatically escape characters for us. */
+
+        const regex_plcode = /<pl-code(.*?)>(.*?)<\/pl-code>/gms;
+        const regex_code = /<code>(.*?)<\/code>/gms;
+
+        let plcode_replace = (match, attrib, contents) => {
+            return `<pl-code ${attrib}>${he.encode(contents)}</pl-code>`;
+        };
+        let code_replace = (match, contents) => {
+            return `<code>${he.encode(contents)}</code>`;
+        };
+
+        return html.replace(regex_plcode, plcode_replace).replace(regex_code, code_replace);
+    },
+
     execTemplate: function(htmlFilename, data, callback) {
         fs.readFile(htmlFilename, { encoding: 'utf8' }, (err, rawFile) => {
             if (ERR(err, callback)) return;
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by echuber2** [comment]
This does something weird if you have a `pl-code` within an HTML comment in `question.html`. Try this:

```
<!-- <pl-code language="cpp" source-file-name="code.cpp" highlight-lines="2-2" highlight-lines-color="#ffbbbe" prevent-select="true"> -->

<pl-code language="cpp">
#include <iostream>
int main() {
  std::cout << "Hello" << std::endl;
}
</pl-code>

<markdown>
\```cpp
#include <iostream>
int main() {
  std::cout << "Hello" << std::endl;
}
\```
</markdown>
```

(Please note where I had to escape internal backticks here to put markdown within markdown for GitHub's sake.)

It results in all of the `question.html` content being escaped as a big HTML comment when the page is rendered.

**Comment by coatless** [comment]
@echuber2 fixed that for you. Just use one extra backtick (c.f. https://github.github.com/gfm/#fenced-code-blocks)

Use:

`````
````html
<pl-code>
</pl-code>

<markdown>
```cpp
#include <iostream>
```
</markdown>
````

`````

echuber2:
@nicknytko I'm still trying a few solutions, but circling back to the original intent: Do we really _want_ to do this in light of the fact that "<" isn't supposed to appear in HTML when it's not used as part of a tag? That is, it's a nice bonus feature if we can automatically escape characters as it does for markdown, but I think people expect that preprocessing from markdown blocks (given the evidence that we have implemented preprocessing for markdown) whereas they would not expect HTML per se to support that anyway.

TLDR should people just know to replace `<` with `&lt;` (and so on) in their `question.html`?

nicknytko:
> @nicknytko I'm still trying a few solutions, but circling back to the original intent: Do we really _want_ to do this in light of the fact that "<" isn't supposed to appear in HTML when it's not used as part of a tag? That is, it's a nice bonus feature if we can automatically escape characters as it does for markdown, but I think people expect that preprocessing from markdown blocks (given the evidence that we have implemented preprocessing for markdown) whereas they would not expect HTML per se to support that anyway.
> 
> TLDR should people just know to replace `<` with `&lt;` (and so on) in their `question.html`?

This is exactly what I've been flip-flopping between, do we support automatically escaping it or do we just expect people to stick to the spec?  If we're not able to find an elegant solution I'm okay with doing the latter here.

echuber2:
Another factor: If people already *did* manually escape their brackets, then the proposed fix breaks those too by escaping them twice.

echuber2:
I guess I'd vote for leaving things alone. If people want auto-escaping for inline code, the best solution is to use the markdown.

mwest1066:
I can certainly support the idea that we just put as many warnings as we can in the docs about HTML needing to escape `<` and leave things as is. It's a good point that some people might already be correctly escaping (I think CS 125 might be doing this) and that if we add escaping it will break their questions.

tdy:
Instead of directly cleaning the markup, what about using those regexes only to generate an issue? e.g., "Warning: you may have ..."

echuber2:
I think it's hard to do that. How can we tell the difference between already-escaped code and unescaped code? Especially for things like `&amp;` [sic] that might be the escaped result of `&`.

One thing we could try: First, copy the pl-code contents and try removing escaping from it naively. Then apply escaping. See if the result matches the original. E.g.:

<iostream> becomes <iostream> becomes &lt;iostream&gt; and this does not match the original, so the original appears to not be escaped.

&lt;iostream&gt; becomes <iostream> becomes &lt;iostream&gt; and this does match the original, so it appears it was already escaped.

In order for this to work, we'd still need to parse the page rather than trying to just use regex, I think (again for the reason that they may have an un-closed `pl-code` within an HTML comment, or they may have a `<code>` within a `pl-code` that is meant to show HTML, and so on). It's quite a rabbit hole, the more I think about it...

nicknytko:
Agreed, I think I’m just going to make the decision to kill this PR and stick some warnings in the documentation.

tdy:
All of those are reasons why the parsing isn't perfect. But that's the point of generating a fuzzy warning like "hey, there are potential potholes here, pay attention to X" instead of directly manipulating the markup. If they've already escaped everything properly, then no problem. Otherwise it can clue them in on what's wrong.

echuber2:
What I mean is that even to detect the case when a warning should be issued is theoretically complex. It's hard to tell if the user has already escaped properly or not, to the extent that you might almost always end creating the warning/issue.

coatless:
@nicknytko any reason the approach taken by `source-file-name` isn't being done with inline?

https://github.com/PrairieLearn/PrairieLearn/pull/1421/files#diff-ed3701cb4819f7f2f8c1ebce4afc7152R3-R81

nicknytko:
We can escape it in the element, but then we run into issues because markdown rendered cells are already escaped, and I think there was some other wonky behaviour that I can’t remember atm.  One possible solution to this is to add an `escape-inline` (or similar) attribute that controls the escaping, and then turn it off when we render markdown?

echuber2:
@coatless I think when the source is loaded from a separate file, it's much easier conceptually, because we can assume that it is not pre-escaped (just a plain source file in whatever language), and since the source is already isolated, there's no need to parse it out of the `question.html` context.

It _would_ probably be reasonable to add the `escape-inline` attribute to `pl-code`, defaulting to false. Then the existing `pl-code` people have written that already have manual escaping would not be damaged. If someone opts to turn it on, we could escape the `pl-code`, but I think this still needs to happen at the level where the page is parsed as HTML (since someone might have a leftover fragment of question content in an HTML comment, etc., and this breaks the proposed regex-based parsing).

In general it's a no-no to parse with extended regex even where it seems feasible, because there will always be edge cases that can't be parsed properly that way... I'm reminded of [this classic answer](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454#1732454).

PR #2343, opened 2020-04-04
Question thumbnails

kevin88tu:
Adds the capability to add an image thumbnail to the instructor questions page. 
Usage: add 'thumbnail.jpg', 'thumbnail.png', or 'thumbnail.jpeg' to the questions folder and it will appear in the instructor questions overview. If there is no thumbnail the box will be left blank (no default).
Questions:
What size should the image be? I have it set to 100x100 pixels, but is there a standard I should follow?
Are there any other file types I should handle?
Example:
Screen Shot 2020-04-04 at 6 54 49 PM


pages/instructorQuestions/instructorQuestions.ejs
```diff
@@ -60,6 +61,9 @@
                                     <%= question.title %></a>
                                 <%- include('../partials/issueBadge', {count: question.open_issue_count, issueQid: question.qid}); %>
                             </td>
+                            <td class="align-middle">
+                                <image src="<%= urlPrefix %>/course_admin/questionThumbnail/<%= question.qid %>" onerror="this.style.display='none'" width="100" height="100"/>
```
mwest1066, line 65: What happens when the browser renders? Does it initially render with the size set to `100 x 100`, then if it turns out there is no thumbnail, does it hide it and cause the table to resize? If that is what happens then it might be better to not request thumbnails for questions that don't have them.

server.js
```diff
@@ -626,6 +626,18 @@ app.use('/pl/course_instance/:course_instance_id/instance_question/:instance_que
     require('./pages/clientFilesQuestion/clientFilesQuestion'),
 ]);
 
+// questionThumbnail
+app.use('/pl/course/:course_id/course_admin/questionThumbnail', [
```
mwest1066, line 630: Nit: URLs use `snake_case`.

pages/questionThumbnail/questionThumbnail.js
```diff
@@ -0,0 +1,28 @@
+var path = require('path');
+var express = require('express');
+var fs = require('fs');
+var router = express.Router();
+
+router.get('/*', function(req, res, _next) {
```
mwest1066, line 6:
```suggestion
router.get('/:qid', function(req, res, _next) {
```

Use an explicit parameter name whenever possible.

server.js
```diff
@@ -626,6 +626,18 @@ app.use('/pl/course_instance/:course_instance_id/instance_question/:instance_que
...
+    function(req, res, next) {res.locals.navSubPage = 'questions'; next();},
```
mwest1066, line 631: We don't need to set a `navSubPage` because we aren't going to render a navbar.

server.js
```diff
@@ -626,6 +626,18 @@ app.use('/pl/course_instance/:course_instance_id/instance_question/:instance_que
...
+    require('./middlewares/studentAssessmentAccess'),
```
mwest1066, line 632: This seems wrong? These pages are only for instructors.

pages/instructorQuestions/instructorQuestions.ejs
mwest1066, line 65: We generally use numeric IDs in URLs to ensure we get exactly the object we want. Here that would mean using `question.id` rather than `question.qid`.

mwest1066:
This PR looks great! Are you planning on also adding thumbnails to all the other locations where instructors see questions? For example, on `pages/instructorAssessmentQuestions`, `pages/instructorQuestion`, etc?

pages/questionThumbnail/questionThumbnail.js
```diff
@@ -0,0 +1,28 @@
...
+        return;
+    }
+    if (fs.existsSync(path.join(clientFilesDir, 'thumbnail.png'))) {
+        res.sendFile('thumbnail.png', {root: clientFilesDir});
+        return;
+    }
+    if (fs.existsSync(path.join(clientFilesDir, 'thumbnail.jpeg'))) {
+        res.sendFile('thumbnail.jpeg', {root: clientFilesDir});
+        return;
+    }
```
nicknytko, line 24: Could we also add support for `.svg`? (And maybe `.gif` :stuck_out_tongue_closed_eyes:  ?)

mwest1066:
Can we configure the page to cache thumbnails quite aggressively? I'm worried about the cost of requesting hundreds of images on every page load.

pages/instructorQuestions/instructorQuestions.ejs
kevin88tu: That is how it works currently. In order to not request the thumbnail, would that require adding something like a thumbnail tag to the info.json? So that we can check if the question has a thumbnail without having to load in the link. Alternatively, is there a way that it is initially no size, but then resizes to 100x100 upon the image loading?

kevin88tu: The question.qid is conveniently the name of the file the image is stored in. Is there a way to go between the two? So that I can pass in the question.id into the get request and still find the folder.

pages/questionThumbnail/questionThumbnail.js
kevin88tu: How does the '/*' parameter work? I tried your suggestion and I couldn't get it to work. I'm looking at [this page](https://expressjs.com/en/api.html#res.send) but it doesn't seem to have a clear explanation.

kevin88tu:
I'll look into how to cache the thumbnails. As for adding the thumbnails to other pages, I suppose it is possible, but the instructorAssessmentQuestions page seems to be more cramped than the instructorQuestions page. Is having thumbnails to that page something I should do?

mwest1066:
> I'll look into how to cache the thumbnails. As for adding the thumbnails to other pages, I suppose it is possible, but the instructorAssessmentQuestions page seems to be more cramped than the instructorQuestions page. Is having thumbnails to that page something I should do?

Yes please! Any instructor page that references questions should include thumbnails, if possible.

pages/questionThumbnail/questionThumbnail.js
mwest1066: See https://expressjs.com/en/guide/routing.html

pages/instructorQuestions/instructorQuestions.ejs
mwest1066: I think we want to avoid adding thumbnail information to `info.json`, because that would be annoying for instructors. We could either: (1) check for thumbnail presence when rendering the page and embed the exact filename in the HTML, or (2) check for thumbnail presence when syncing the course to the DB and store the filename in the `questions` table.

I would suggest trying (1) and measuring how fast it is on a questions list with many questions. If that's too slow then we might have to resort to (2), which is more complicated.

I also suggest that maybe 64x64 might be a good size. The current size seems visually a bit too big to me.

sync/fromDisk/questions.js
```diff
@@ -24,6 +26,22 @@ module.exports.sync = function(courseInfo, questionDB, jobLogger, callback) {
...
+            }
+            if (fs.existsSync(path.join(courseInfo.questionsDir, qid, 'thumbnail.jpeg'))) {
+                thumbnailPresence = true;
+            }
+            if (fs.existsSync(path.join(courseInfo.questionsDir, qid, 'thumbnail.svg'))) {
+                thumbnailPresence = true;
+            }
+            if (fs.existsSync(path.join(courseInfo.questionsDir, qid, 'thumbnail.gif'))) {
+                thumbnailPresence = true;
+            }
```
nicknytko, line 44: I see that these lines are essentially duplicated here and in `questionThumbnail.js`, would we be able to find the path once and then save that for whenever we request the thumbnail?  Perhaps instead of a thumbnail boolean, we could store the path to the thumbnail in the db and then use that in the express router?

pages/instructorQuestions/instructorQuestions.ejs
```diff
@@ -60,6 +61,11 @@
                                     <%= question.title %></a>
                                 <%- include('../partials/issueBadge', {count: question.open_issue_count, issueQid: question.qid}); %>
                             </td>
+                            <td class="align-middle">
+                                <% if (question.thumbnail) { %>
+                                <image src="<%= urlPrefix %>/course_admin/question_thumbnail/<%= question.qid %>" onerror="this.style.display='none'" width="64" height="64"/>
```
nicknytko, line 66: is the `onerror` needed now that we are checking the thumbnail's existence ahead of time?

pages/questionThumbnail/questionThumbnail.js
```diff
@@ -0,0 +1,16 @@
...
+
+router.get('/:qid/:filename', function(req, res, _next) {
+    var qid = req.params['qid'];
+    var filename = req.params['filename'];
+    var clientFilesDir = path.join(
+        res.locals.course.path,
+        'questions',
+        qid,
+    );
+    res.sendFile(filename, {maxAge: 86400000 * 30, root: clientFilesDir});
```
nicknytko, line 13: I think this is a great simplification, but I'm worried this will serve any file in the same directory.  Could the route just get `/:qid` and then load the thumbnail path via sql query?  I think you can see an example of this pattern in `pages/news_items/news_items.js` and the sql file in same folder.

doc/question.md
```diff
@@ -125,6 +125,30 @@ To disable partial credit for a question, set `"partialCredit": false` in the `i
...
+questions
+|
+|-- fossilFuelsRadio          # first question, id is "fossilFuelsRadio"
+|   |
+|   +-- info.json             # metadata for the fossilFuelsRadio question
+|   +-- thumbnail.jpg         # thumbnail image associated with this question
+|   +-- server.py             # secret server-side code (optional)
+|   `-- question.html         # HTML template for the question
+```
+
```
nicknytko, line 151: Maybe add a small blurb about what pages the thumbnail will show up in?

pages/instructorAssessmentQuestionStatistics/instructorAssessmentQuestionStatistics.ejs
```diff
@@ -132,6 +133,11 @@
                     <%= row.assessment_question_number %>. <%= row.question_title %>
                   </a>
                 </td>
+                <td>
+                  <% if (row.thumbnail != null) { %>
+                  <image src="<%= urlPrefix %>/assessment/<%= assessment.id %>/question_thumbnail/<%= row.question_id %>" onerror="this.style.display='none'" width="64" height="64"/>
```
mwest1066, line 138: Why do we need the `onerror` here?

pages/questionThumbnail/questionThumbnail.sql
```diff
@@ -0,0 +1,9 @@
+-- BLOCK questions
+SELECT
+    q.thumbnail, q.qid, q.id
+FROM
+    questions as q
+    LEFT JOIN course_instances AS ci ON (ci.id = $course_instance_id)
```
mwest1066, line 6: We don't seem to use `ci` for anything here?

pages/questionThumbnail/questionThumbnail.sql
```diff
@@ -0,0 +1,9 @@
...
+    LEFT JOIN pl_courses AS c ON (c.id = $course_id)
```
mwest1066, line 7: Why a `LEFT JOIN` rather than a plain `JOIN`?

server.js
```diff
@@ -622,6 +622,20 @@ app.use('/pl/course_instance/:course_instance_id/instance_question/:instance_que
...
+]);
+app.use('/pl/course_instance/:course_instance_id/instructor/course_admin/question_thumbnail', [
+    function(req, res, next) {next();},
+    require('./pages/questionThumbnail/questionThumbnail'),
+]);
+app.use('/pl/course_instance/:course_instance_id/instructor/assessment/:assessment_id/question_thumbnail', [
+    function(req, res, next) {next();},
+    require('./pages/questionThumbnail/questionThumbnail'),
+]);
+
```
mwest1066, line 638: Can we just use a single route of `/pl/course/:course_id/question/:question_id/thumbnail`? I don't think we need to have three different routes here.

sync/fromDisk/questions.js
```diff
@@ -24,6 +26,22 @@ module.exports.sync = function(courseInfo, questionDB, jobLogger, callback) {
...
+            if (fs.existsSync(path.join(courseInfo.questionsDir, qid, 'thumbnail.jpeg'))) {
+                thumbnailPath = 'thumbnail.jpeg';
+            }
+            if (fs.existsSync(path.join(courseInfo.questionsDir, qid, 'thumbnail.svg'))) {
+                thumbnailPath = 'thumbnail.svg';
+            }
+            if (fs.existsSync(path.join(courseInfo.questionsDir, qid, 'thumbnail.gif'))) {
+                thumbnailPath = 'thumbnail.gif';
+            }
```
mwest1066, line 44: Please don't use `sync` functions anywhere in the code. High performance in NodeJS requires only using asynchronous functions like `fs.exists()` (not `fs.existsSync()`).

mwest1066:
Should we also add thumbnails to student pages? I'm thinking that we would display them using the same rules as for displaying question titles (that is, whenever we display a question title, we would also display the thumbnail if one exists).

mwest1066:
Can you add thumbnails to some of the questions in the exampleCourse, to demonstrate the usage?

docs/question.md
```diff
@@ -212,6 +212,29 @@ To disable partial credit for a question, set `"partialCredit": false` in the `i
 
 In general, it is strongly recommended to leave partial credit enabled for all questions.
 
+## The `thumbnail` image
+
+To add a thumbnail to your question, add a `thumbnail.jpg` image to your question folder. The thumbnail will appear on the `instructorQuestions`, `instructorAssessmentQuestions`, and `instructorAssessmentQuestionStatistics` pages. Supported file types are:
```
mwest1066, line 217: I don't think it's necessary to list the pages here.

mwest1066:
This PR changes the rules for `sync`. Previously, the only thing the `sync` code needed was the `info.json` files. Now it also needs to test for the presence of thumbnails in the question directories. This means we need to be careful about when we trigger a sync operation. I think we should either:

1. Add a check to the in-browser editing code to trigger a sync whenever a thumbnail image is modified, and also add special code to detect the addition of thumbnails when in dev mode.

OR

2. Require that the thumbnail be explicitly listed in the `info.json` file.

I sort of like option (2), but to be friendly we should also add a "Change thumbnail" button to the `instructorQuestionSettings` page that uploads a new images and, if necessary, edits `info.json` and triggers a sync.

mwest1066:
At the moment thumbnails are question-specific, but I think that some instructors will want to have "generic" thumbnails. For example, they might want to have all their coding questions have a "coding" thumbnail, and all their multiple-choice questions might have a "multiple-choice" thumbnail. We could implement this by allowing thumbnails to be loaded from any of these locations:

* The `clientFilesQuestion` directory.
* The `clientFilesCourse` directory.
* A built-in list of thumbnails inside PL.

This could be done by specifying the thumbnail inside the `info.json` file with a format like:
```
"thumbnail": {"filename": "code.png", "location": "clientFilesCourse"}
```

Now that I think about it more, this would mean that we wouldn't need to add any special routes to serve the thumbnail images, but we could just rely on the existing routes to serve files.

sync/fromDisk/questions.js
nicknytko: Oh good catch.  We had this in there just to test the functionality with the intention of moving to async calls.

astrosticks:
Hi Kevin! I came up with some suggestions during the weekly meeting today: 
- I think it's standard in UIs to put icons on the same side of text as the text is aligned (like how Slack puts user icons to the left of left-aligned messages), so I think moving the question icons one column to the left would look neater and help keep the textual information on one side of the screen. 
- If icons can't be sorted/filtered, it would probably be good to disable that interaction for that column, if possible. (Maybe the icon and question title columns can be merged? Idk if this would break sorting/filtering)
- Adding the icons is useful but we probably shouldn't let the icons increase the page size by too much (it looks like your screenshots make the rows 4-5x as tall). Since this is to help users find things faster, we probably shouldn't ask them to scroll much more than they already do.
- Also, making the rows a consistent height with or without an icon would probably make the page look a lot neater too. 

Keep up the good work!

mwest1066:
> * If icons can't be sorted/filtered, it would probably be good to disable that interaction for that column, if possible. (Maybe the icon and question title columns can be merged? Idk if this would break sorting/filtering)

I think that if we add support for "generic icons" (e.g., a "coding icon") then sorting might actually useful to group all icons together. But maybe this is too complicated and we should just disable it as you suggest. I'm ok either way.

kevin88tu:
Changed: 
Added thumbnails to student pages
Added 3 examples to exampleCourse
Added thumbnail tag to info.json (default: none)
Removed special routes, using existing ones like clientFilesCourse
Moved thumbnails one column to the left
Reduced thumbnail size to 50x50
Made row height consistent (min-height=50)

TODO:
Add "Change thumbnail" to instructorQuestionSettings
Manage sorting thumbnails

Questions:
Can I have a list of all directories thumbnails may be stored in? I had to add a line to server.js since the instructorQuestions page couldn't reach clientFilesCourse, so I'd like to know if there are any other places I'd have to include in server.js
Just my two cents, but prior to the thumbnail implementation there were still rows in exampleCourse that were of different size due to all the tags, and there probably still will be depending on how many tags a question might have. I didn't think the differently sized rows made it less neat and it was more space efficient, but I implemented it anyways so you could judge how it looks.
On Monday's meeting, we discussed how we could automatically insert a thumbnail for a question based on its tags or question type. Is this up for consideration? Because it seems like each course could have its own default set of thumbnails, so it might be difficult to set a standard based on tags and such.
I just included some thumbnails in exampleCourse that I used while testing, do you want me to create a new question specifically for demonstrating thumbnails?

mwest1066:
> I just included some thumbnails in exampleCourse that I used while testing, do you want me to create a new question specifically for demonstrating thumbnails?

I think that having thumbnails added to existing questions is great. There is no need to add new questions. But could we have some nicer ones? 😄 For example, `demoDrawingPulley` could have a screenshot of the pulley diagram, `demoFixedCheckbox` could have a screenshot of the graph, or `demoCalculation` could have a screenshot of the `a + b` equation, just as ideas.

mwest1066:
> Just my two cents, but prior to the thumbnail implementation there were still rows in exampleCourse that were of different size due to all the tags, and there probably still will be depending on how many tags a question might have. I didn't think the differently sized rows made it less neat and it was more space efficient, but I implemented it anyways so you could judge how it looks.

I agree with you. Let's just allow the rows to be their natural height.

mwest1066:
> On Monday's meeting, we discussed how we could automatically insert a thumbnail for a question based on its tags or question type. Is this up for consideration? Because it seems like each course could have its own default set of thumbnails, so it might be difficult to set a standard based on tags and such.

I agree with you that I can't see any way to do this that would make sense for different courses. I'm happy to have this added if someone can think of a good way to do it that works across the board, but otherwise let's not do this.

pages/instructorQuestions/instructorQuestions.ejs
```diff
@@ -42,6 +42,7 @@
                 <table id="questionsTable" class="table table-sm table-hover tablesorter">
                     <thead>
                         <tr>
+                            <th width="50">Thumbnail</th>
```
mwest1066, line 45: Why are we setting an explicit width here? It seems too wide, and it also forces the column to take up a lot of space even when there aren't any thumbnails.

Here we can see that the column looks too wide:

image

Here is an example where there aren't any thumbnails:

image

mwest1066:
Can we also add thumbnails to:
* The "Detailed question statistics" table on `pages/instructorAssessmentQuestionStatistics`
* `pages/instructorQuestionPreview` (perhaps in the main panel header, to the left of the question title?)
* `pages/instructorQuestionSettings` (I guess this will come when you add "Change thumbnail")
* `pages/instructorQuestionStatistics` (perhaps in the main panel header?)
* `pages/instructorAssessmentInstance` (in the two tables that list questions, and in the log table)
* `pages/studentInstanceQuestionHomework` (in the main panel header)
* `pages/studentInstanceQuestionExam` (in the main panel header, if we are showing titles)

pages/studentAssessmentInstanceExam/studentAssessmentInstanceExam.ejs
```diff
@@ -149,10 +149,11 @@
           </div>
           <% } %>
         </div>
-  
+
         <table class="table table-sm table-hover">
           <thead>
             <tr>
+              <th width="50">Thumbnail</th>
```
mwest1066, line 156: On student pages I think we shouldn't call out the thumbnail explicitly. That is, don't label the column "Thumbnail", but instead make the "Question" header span two columns to cover both the thumbnail and the question title.

pages/instructorAssessmentQuestionStatistics/instructorAssessmentQuestionStatistics.ejs
```diff
@@ -117,6 +117,7 @@
           <table id="questionsTable" class="table table-sm table-hover tablesorter">
             <thead>
               <tr>
+                <th class="text-center" width="50">Thumbnail</th>
                 <th class="text-center">Question</th>
```
mwest1066, line 121:
```suggestion
                <th colspan="2">Question</th>
```

I think we should just include the thumbnail under the "Question" label here, and left-justify the header.

I generally think we should do this on all pages where we are just listing questions, and the user doesn't really care about the specific parts of the question. This does not apply to `pages/instructorQuestions`, because there we are really separating out thumbnail/title/qid/etc. But almost all other pages should show the thumbnail as part of the "Question" column (even though it might actually be a separate column in the table HTML).

database/tables/questions.pg
```diff
@@ -9,8 +9,10 @@ columns
     external_grading_files: text[] default ARRAY[]::text[]
     external_grading_image: text
     external_grading_timeout: integer
+    filename: text
```
mwest1066, line 12: I think we should give this column a more descriptive name, otherwise it might be very unclear what the "question filename" means.

```suggestion
    thumbnail_filename: text
```

database/tables/questions.pg
```diff
@@ -9,8 +9,10 @@ columns
...
     grading_method: enum_grading_method not null default 'Internal'::enum_grading_method
     id: bigint not null default nextval('questions_id_seq'::regclass)
+    location: text
```
mwest1066, line 15:
```suggestion
    filename_location: text
```

mwest1066:
> Reduced thumbnail size to 50x50

I wonder if we should reduce this further, perhaps to 32x32? Or even 24x24?

I think to judge this it would be good to have some realistic examples of thumbnails for the exampleCourse questions, so we can try out just how small they can go.

mwest1066:
> Can I have a list of all directories thumbnails may be stored in? I had to add a line to server.js since the instructorQuestions page couldn't reach clientFilesCourse, so I'd like to know if there are any other places I'd have to include in server.js

I think there are three locations that thumbnails could come from:
* `"location": "question"`: this means the thumbnail will be in the actual question directory itself (not in `clientFilesQuestion`). Now that I think about it more, we can't use `clientFilesQuestion` because that might allow students to access the thumbnail during exams, which we want to prevent. This will require adding back the custom routes. Sorry about this! I forgot about the security implications when I recommended removing the custom routes.
* `"location": "clientFilesCourseThumbnails"`: the thumbnail is in the `clientFilesCourse/thumbnails` directory. This can use existing routes.
* `"location": "public"`: this refers to thumbnails in the directory `public/images/questionThumbnails` (which doesn't exist yet). This can use existing routes.

When implementing the "change thumbnail" button for instructors, I think we would give them the option of uploading their own thumbnail and storing it in the question directory (and setting `"location": "question"`) or choosing one of the existing thumbnails in `clientFilesCourse/thumbnails` or `public/images/thumbnails`.

server.js
```diff
@@ -607,6 +607,9 @@ app.use('/pl/course_instance/:course_instance_id/clientFilesCourse', [
     require('./middlewares/studentAssessmentAccess'),
     require('./pages/clientFilesCourse/clientFilesCourse'),
 ]);
+app.use('/pl/course/:course_id/clientFilesCourse', [
+    require('./pages/clientFilesCourse/clientFilesCourse'),
+]);
```
mwest1066, line 612: I think this route should be needed. For all student pages you will have a valid `locals.course_instance.id` so you can use the route `/pl/course_instance/:course_instance_id/clientFilesCourse`. Or am I misunderstanding why you needed to add this route?

kevin88tu: I think this course is needed because it is possible to view a course's questions without a :course_instance_id, in which case this route is needed. For example, clicking on 'exampleCourse' and then 'questions' in the top navbar brings you to the questions page without a course_instance_id

nicknytko:
The included thumbnails are all royalty free/creative commons, yes?

kevin88tu:
I switched out a couple of thumbnails since I wasn't completely confident they weren't copyrighted, and added some more public thumbnails. These are 'easy', 'medium', 'hard', 'long', and 'tricky', which I thought could be used for a variety of courses. I just made these using microsoft paint so if you decide they're not useful or they don't look good you can just remove them without asking me.

pages/partials/uploadThumbnailForm.ejs
```diff
@@ -0,0 +1,49 @@
+<form name="upload-thumbnail-form" class="needs-validation" method="POST" enctype="multipart/form-data" novalidate>
+    <div class="container p-0 mb-4">
+        Upload an image of type jpg, jpeg, png, svg, or gif.
```
nicknytko, line 3: Minor stylistic nit: can we wrap the file extensions with `<code>` tags (or better yet, `<code class="user-output">`)

sync/fromDisk/questions.js
```diff
@@ -52,6 +58,8 @@ module.exports.sync = function(courseInfo, questionDB, jobLogger, callback) {
                 external_grading_entrypoint: (q.externalGradingOptions && q.externalGradingOptions.entrypoint),
                 external_grading_timeout: (q.externalGradingOptions && q.externalGradingOptions.timeout),
                 external_grading_enable_networking: (q.externalGradingOptions && q.externalGradingOptions.enableNetworking),
+                thumbnail_filename: filename,
+                filename_location: location,
```
nicknytko, line 62: Can we make this more concise like the above external grading options?

```suggestion
                thumbnail_filename: (q.thumbnail && q.thumbnail.filename),
                filename_location: (q.thumbnail && q.thumbnail.location),
```

migrations/173_questions_filename.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE questions ADD COLUMN thumbnail_filename text;
```
nicknytko, line 1: Will defer to @mwest1066 on this, but can we combine migrations 173 and 174 into one file?

lib/editors.js
```diff
@@ -1414,6 +1460,147 @@ class FileRenameEditor extends Editor {
...
+        }
+
+        const found = this.container.invalidRootPaths.find((invalidRootPath) => contains(invalidRootPath, this.filePath));
+        if (found) {
+            let err = new Error('Invalid file path');
+            err.info =  `<p>The file path</p>` +
+                        `<div class="container"><pre class="bg-dark text-white rounded p-2">${this.oldPath}</pre></div>` +
+                        `<p>must <em>not</em> be inside the directory</p>` +
+                        `<div class="container"><pre class="bg-dark text-white rounded p-2">${found}</pre></div>`;
+            return callback(err);
```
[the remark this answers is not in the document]
kevin88tu: I want to edit when the image is already there but the thumbnail information in the info.json doesn't match, so I'm going to include the shouldEdit() as well.

pages/instructorQuestionSettings/instructorQuestionSettings.js
```diff
@@ -169,6 +230,74 @@ router.get('/', function(req, res, next) {
...
+        (callback) => {
+            const clientFilesCoursePath = encodePath(path.join(res.locals.course.path, 'clientFilesCourse', 'thumbnails'));
+            const questionPath = encodePath(path.join(res.locals.course.path, 'questions', res.locals.question.qid));
+            const publicPath = encodePath('public/images/thumbnails');
+            res.locals.available_thumbnails = [];
+            async.series([
+                (callback) => {
+                    fs.readdir(clientFilesCoursePath, function (err, files) {
+                        // Probably no clientFilesCourse directory
+                        if (err) {
```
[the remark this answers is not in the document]
kevin88tu: When I tested this change, the error triggered on ENOENT and caused the server to exit, so I added a special case for it. I want ENOENT to call callback(null, {}) because a clientFilesCourse/thumbnails not existing is completely OK and shouldn't cause an error.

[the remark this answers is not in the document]
nicknytko: Ah gotcha, disregard my comments then.

mfsilva22:
Few things just from a quick look:
1) I think the "Thumbnail" column in the Questions page is too wide (the default size). I think it should be narrower. I still think the thumbnail size is large. Can we make it smaller?
2) With making it smaller, can we think of a way of hover over the image, which will bring a zoom of it (larger view)? 
3) As of now, the thumbnail appears before the question title in the question panel. At this level, the image of the question is already there, it makes no sense of adding the thumbnail. This should be removed, since it only clutters the question.

kevin88tu:
The extra none option was being caused because when a directory doesn't exist (for example, clientFilesCourse/thumbnails doesn't exist) my js parsed an empty object as an empty string. I filtered out the empty objects so it should not display empty objects. I also increased the size of the thumbnail zoom while also making it zoom towards the upper right, which prevented clipping and makes it look better in my opinion. I also made it so clicking on the image selects the thumbnail in the select thumbnail form.

mfsilva22:
Matt's request above, and a couple others from me:

- [x] add thumbnails to student pages
- [x] add thumbnails to some exampleCourse questions
- [x] require thumbnails be explicitly listed in info.json
- [x] add a "Change thumbnail" button to the instructorQuestionSettings page
- [x] support course-wide thumbnails like "thumbnail": {"filename": "code.png", "location": "clientFilesCourse"}
- [] discuss in the dev meeting the support for clientFilesQuestion
- [x?] remove special routes to serve the thumbnail images and use the existing file routes
- [x] move question icons one column to the left
- [] decide whether to allow sorting/filtering by icons (not done and I personally don't think it is needed)
- [] make sure icons are small enough that they don't make question lists too long (maybe not small enough yet?)
- [] make table rows a consistent height even if there is no icon 
- [] to make the above possible, I think we will need to make the thumbnails even smaller (discuss this in the dev meeting today)
- [] question page seems to be taking longer to load then usual

nicknytko:
I like the thumbnail scaling on hover, but can we change it so that the effect is centered on the y-axis?  And maybe add a shadow on thumbnail hover?  Something that helps separate the thumbnail from the background imo.

pages/partials/questionThumbnail.ejs
```diff
@@ -0,0 +1,9 @@
+<% if (location == "clientFilesCourse") { %>
+<image src="<%= urlPrefix %>/clientFilesCourse/thumbnails/<%= filename %>" class="thumbnail"/>
```
tdy, line 2:
```suggestion
<image src="<%= urlPrefix %>/clientFilesCourse/thumbnails/<%= filename %>" class="thumbnail" alt=""/>
```

pages/partials/questionThumbnail.ejs
```diff
@@ -0,0 +1,9 @@
...
+<% } %>
+<% if (location == "question") { %>
+<image src="<%= urlPrefix %>/question_thumbnail/<%= id %>/<%= filename %>" class="thumbnail"/>
```
tdy, line 5:
```suggestion
<image src="<%= urlPrefix %>/question_thumbnail/<%= id %>/<%= filename %>" class="thumbnail" alt=""/>
```

pages/partials/questionThumbnail.ejs
```diff
@@ -0,0 +1,9 @@
...
+<% } %>
+<% if (location == "public") { %>
+<image src="/images/thumbnails/<%= filename %>" class="thumbnail"/>
```
tdy, line 8:
```suggestion
<image src="/images/thumbnails/<%= filename %>" class="thumbnail" alt=""/>
```

kevin88tu:
I added clientFilesQuestion as a thumbnail location. Quick question: should the thumbnails be stored in `clientFilesQuestion` or `clientFilesQuestion/thumbnails`? I made it the former, but I think it should be the latter because the selectThumbnailForm will display everything inside the `clientFilesQuestion` folder since there isn't a way to select only the thumbnail images. 
<br>
I also made the thumbnails smaller but with a larger zoom, and implemented Nick's code to add a shadow and center the image zoom. I also removed thumbnails from the student exam pages since Matt doesn't want thumbnails to be accessible during exams.

nicknytko:
> 
> 
> I added clientFilesQuestion as a thumbnail location. Quick question: should the thumbnails be stored in `clientFilesQuestion` or `clientFilesQuestion/thumbnails`? I made it the former, but I think it should be the latter because the selectThumbnailForm will display everything inside the `clientFilesQuestion` folder since there isn't a way to select only the thumbnail images.
> 
> 
> I also made the thumbnails smaller but with a larger zoom, and implemented Nick's code to add a shadow and center the image zoom. I also removed thumbnails from the student exam pages since Matt doesn't want thumbnails to be accessible during exams.

I think we should be able to select any image in `clientFilesQuestion`, if I remember Mariana's use case correctly she wanted to reuse existing figures from the `clientFilesQuestion` folder as the thumbnail.  It's probably fine if you're showing all images in the thumbnail picker.

nicknytko:
Can we explore what the thumbnail picker would look like if all the thumbnails were on a grid?  Maybe something akin to:
  
(https://victor-valencia.github.io/bootstrap-iconpicker/ )
Where each thumbnail is a button?  Also what are your thoughts on dismissing the popup when the user clicks in the background?  I know the other popovers don't do this but I feel like it makes more natural to me from a UX standpoint.

mwest1066:
@kevin88tu I'm sorry for taking so long to get your PR reviewed again and merged. I'm afraid that it's now a bit out of date. Could you please fix up the conflicts with master?

nwalters512:
@kevin88tu would you still like to get these changes merged in? There are a number of conflicts that need resolving. If you'd no longer like to work on this, I'll close it out.

nwalters512:
It's been close to a year since this was worked on, and there hasn't been any sign that Kevin will come back to work on this. Closing it out.

Issue #2177, opened 2020-03-19
Add admin dashboard


trombonekenny:
I think we need some kind of dashboard view of the current state of a live PL server. Examples of things it might show:

* Currently running assessments, their expected student count, number started, number completed, etc.
* Assessments opening/closing in the next N days
* List of current instructors (to paste into mass mailings or user management in lists.illinois.edu)

What else might this contain?

PR #2345, opened 2020-04-05
Add administrator query runner

mwest1066:
This PR is based on Dave's dashboard work in #2177 

It extends that work to support parameterized queries, but only allows running single queries at a time.

image

image


lib/csv-maker.js
```diff
@@ -15,7 +16,7 @@ module.exports = {};
       ['Column 1', 'col1'],
       ['Column 2', 'col2'],
   ];
-  rowsToCsv(rows, headers, properties, function(err, csv) {
+  rowsToCsv(rows, columns, function(err, csv) {
```
nicknytko, line 19: Nice catch!

pages/partials/navbarPlain.ejs
```diff
@@ -1,4 +1,5 @@
 
       <% if (is_administrator) { %>
       <li class="nav-item <% if (navPage == "admin") { %>active<% } %>"><a class="nav-link" href="<%= plainUrlPrefix %>/administrator/overview">Admin</a></li>
+      <li class="nav-item <% if (navPage == "adminQueries") { %>active<% } %>"><a class="nav-link" href="<%= plainUrlPrefix %>/administrator/queries">Queries</a></li>
```
nicknytko, line 4: To me it seems a bit strange to have an "admin" tab and a "queries" tab when they're both admin tasks -- if we don't expect to have too many of these queries could we combine them into one page?

mwest1066:
> This looks great! Would it be worthwhile to add a sort of "custom" sql query page where you can spit some lines of sql code into ace editor and it will run it for you?

That would be awesome! I'll add an issue for that, rather than doing it now.

pages/partials/navbarPlain.ejs
mwest1066: Totally agreed! I'd actually like to break up the Admin page into sub-pages for all the individual panels, and add a tab-bar under "Admin" for the sub-pages. Then "Queries" would be one of these tabs.

I didn't want to make this PR more complex by also doing this breakup, but I'll add an issue for it now.

pages/administratorQuery/administratorQuery.ejs
```diff
@@ -0,0 +1,171 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <%- include('../partials/head'); %>
+    <link href="/stylesheets/theme.bootstrap.css" rel="stylesheet" />
+    <link href="/stylesheets/highlight-default.css" rel="stylesheet" />
+    <script src="/javascripts/jquery.tablesorter.min.js"></script>
+    <script src="/javascripts/jquery.tablesorter.widgets.js"></script>
+    <script src="/javascripts/highlight.js"></script>
```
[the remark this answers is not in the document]
nicknytko: Great question, not sure why we have both but I'll dig through the source and try to understand why we have two copies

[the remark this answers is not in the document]
mwest1066: Awesome, I'll file an issue for this an not try and resolve it in the current PR.

PR #2353, opened 2020-04-06
Fix unaligned matrix input

### LINKED ISSUE #2352: `pl-matrix-component` cells are unaligned
Screenshot:

Screen Shot 2020-04-06 at 11 00 03 AM

I am guessing with the recent changes to add padding to tables we missed that the matrices should not be padded.

nicknytko:
Closes #2352.

Before:
Screen Shot 2020-04-06 at 11 00 03 AM

After:
Screen Shot 2020-04-06 at 11 23 56 AM

Recent changes to table padding are shifting entries in the matrix input, so created a `.table-no-row-padding` class and applied it to the component input to turn off the padding.

### DIFF
--- a/public/stylesheets/local.css
+++ b/public/stylesheets/local.css
@@ -186,10 +186,16 @@ table td > .badge {
 }
 
 /* Used to give table rows the same padding as Bootstrap card headers */
-.card table:not(.table-no-row-padding) tr td:first-child, .card table:not(.table-no-row-padding) tr th:first-child {
+.card > table tr td:first-child,
+.card > table tr th:first-child,
+.card .table-responsive > table tr td:first-child,
+.card .table-responsive > table tr th:first-child {
     padding-left: 1.25rem;
 }
-.card table:not(.table-no-row-padding) tr td:last-child, .card table:not(.table-no-row-padding) tr th:last-child {
+.card > table tr td:last-child,
+.card > table tr th:last-child,
+.card .table-responsive > table tr td:last-child,
+.card .table-responsive > table tr th:last-child {
     padding-right: 1.25rem;
 }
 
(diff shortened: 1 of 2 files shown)


public/stylesheets/local.css
```diff
@@ -186,10 +186,10 @@ table td > .badge {
 }
 
 /* Used to give table rows the same padding as Bootstrap card headers */
-.card table tr td:first-child, .card table tr th:first-child {
+.card table:not(.table-no-row-padding) tr td:first-child, .card table:not(.table-no-row-padding) tr th:first-child {
```
mwest1066, line 189: An alternative fix might be to change to `.card > table` to restrict to tables that are immediate descendants of cards? I think that's the case that @astrosticks was worried about in #2290?

nicknytko: That might be cleaner, let me try that.

nicknytko: Changed, I also added `.card .table-responsive > table` since sometimes there's an extra div in the chain.

public/stylesheets/local.css
```diff
@@ -186,10 +186,12 @@ table td > .badge {
 }
 
 /* Used to give table rows the same padding as Bootstrap card headers */
-.card table tr td:first-child, .card table tr th:first-child {
+.card > table tr td:first-child, .card > table tr th:first-child,
+.card .table-responsive > table tr td:first-child, .card .table-responsive > table tr th:first-child {
```
mwest1066, line 190: Do we want that to be `.card > .table-responsive`?

mwest1066: I also think I prefer the 4-line version, fwiw.

nicknytko: Sure, this seems like a safe change to make.

mwest1066: Super minor nit, but the 4-line version was easier for me to see what was going on.

nicknytko: hah, sure

Issue #2348, opened 2020-04-05
Duplicate copies of highlight.js


mwest1066:
`pl-code` uses a downloaded copy of `highlight.js` in `public/javascripts` (see comments in `pl-code.py`), while `pages/instructorFileBrowser` uses `node_modules/highlight`. It's not clear to me why this is the case, or whether we can delete one copy (preferably the `javascripts` copy).

If we do delete the copy in `public/javascripts` we should be careful to add a back-compat route to serve the `node_modules` version on the old path, for the sake of any questions importing it directly. Also note that we'll need to override this in Apache for prod.

nicknytko:
It seems the version in `public/javascripts` contains all languages pre-packaged in one javascript file, while the version in `node_modules` requires languages to be loaded via separate javascript files.  Do you think we should give dynamically loading these language files a try?

mwest1066:
How does the dynamic loading work? Do we need to do anything or does `highlight` figure out which one to load and take care of it all? Assuming so, that sounds much better, because we get to use the updated `node_modules` version and we also don't ship every client a huge set of languages when they just want one.

nicknytko:
We'd need to `import` or `require` (whatever module loader is being used) all the languages manually, which I'm worried could break backwards compatibility.

mwest1066:
Ah, ok, I get it. In `instructorFileBrowser` we are doing server-side highlighting using `highlight.js`, but in `pl-code` we can't do this because we are running in Python. We thus do highlighting client-side. Yuck!

Can you think of a good reason why we don't use Pygments in pl-code?

nicknytko:
I’m totally in favor of using Pygments here, I’ve used it before in the python autograder and it works great.

mwest1066:
If you switch pl-code to Pygments then we can probably delete `/javascripts/highlight`, which would be great! One less thing that slowly bitrots.

PR #2354, opened 2020-04-06
Change to server-side SQL highlighting

mwest1066:
See #2348

This PR changes the SQL query page to use server-side highlighting with the copy of `highlight.js` in `node_modules`.

### DIFF
--- a/pages/administratorQuery/administratorQuery.js
+++ b/pages/administratorQuery/administratorQuery.js
@@ -4,6 +4,7 @@ const asyncHandler = require('express-async-handler');
 const fsPromises = require('fs').promises;
 const path = require('path');
 const _ = require('lodash');
+const hljs = require('highlight.js');
 
 const csvMaker = require('../../lib/csv-maker');
 const jsonLoad = require('../../lib/json-load');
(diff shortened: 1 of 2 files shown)


pages/administratorQuery/administratorQuery.js
```diff
@@ -4,6 +4,7 @@ const asyncHandler = require('express-async-handler');
 const fsPromises = require('fs').promises;
 const path = require('path');
 const _ = require('lodash');
+const hljs = require('highlight.js');
```
nicknytko, line 7:
```suggestion
const hljs = require('highlight.js/lib/highlight.js');
hljs.registerLanguage('sql', require('highlight.js/lib/languages/sql'));
```

I'm thinking we should pull in only the sql language here

mwest1066: I don't mind doing this, but is there really an advantage? We load `highlight` only once when node starts up, and because other pages (`instructorFileBrowser`) are already loading it we get it for free (Node's `require()` only imports one shared copy of each package for the entire process).

nicknytko: Ah, didn't realize we we're already pulling everything in &mdash; in that case I say go ahead and import it all :slightly_smiling_face:

PR #2363, opened 2020-04-08
Use server-side highlighting in `pl-code`

nicknytko:
Removes the need for client-side syntax highlighting with `highlight.js`, this gives us more languages and removes the duplicate copy of `highlight.js` in public.


elements/pl-code/pl-code.py
```diff
@@ -70,18 +55,98 @@
...
+        self.name = "No Highlighting Lexer"
+        self.aliases = ["none"]
+        self.filenames = ["*.none"]
+        self.mimetypes = []
+
+    def get_tokens_unprocessed(self, text):
+        return [(0, Token.Text, text)]
+
+
+class HljsFormatter(pygments.formatter.Formatter):
```
mwest1066, line 71: I'm not very familiar with how Pygments works, but is this class so that we can emulate the look of Highlight.js using Pygments? Does Pygments have a default style that we could use instead? I don't think there is any particular need to maintain backwards compatibility with the previous highlighting?

nicknytko: It does have a default style, I don't think it looks particularly great but it's something.

mwest1066: Ok, I'll leave it up to you. I'm generally in favor of using defaults and not having our own special code unless there's really a significant gain. I haven't seen the Pygments default style so I don't know how bad it is 😄

nicknytko: Sure, I think it actually may come with a bunch of styles built-in so I'll see if any of those look nice

mwest1066: Awesome, if any of them are acceptable then that will let us have less code to maintain in the future 😄

nicknytko:
Used the built-in style `friendly`:

image
image

Which I think looks decently nice, I also added a little bit of custom styling to round out the code boxes because I think that looks nice, thoughts?

I am worried this will break @gchallen 's questions though, looks like he is importing `<pl-code>` just to get the hljs styles?

elements/pl-code/pl-code.mustache
```diff
@@ -1,12 +1,10 @@
...
+<div class="pl-code"><span
 {{#prevent_select}}
 style="-moz-user-select: none; -webkit-user-select: none; -ms-user-select:none; user-select:none;-o-user-select:none;"
 unselectable="on"
 onselectstart="return false;"
 onmousedown="return false;"
 oncontextmenu="return false;"
 {{/prevent_select}}
->{{{code}}}</code></pre>
+>{{{code}}}</span></div>
```
mwest1066, line 10: If we change `<code>` to `<span>`, do we also have to worry about things like wrapping and whitespace behavior being different by default?

mwest1066:
This looks great! I love PRs that achieve a net -100 lines of code!

mwest1066:
Can you ask @gchallen what he's doing with highlight.js so that we don't accidentally break his questions when we switch to Pygments? If he just needs the CSS styles, maybe he could include those himself in `clientFilesCourse`?

elements/pl-code/pl-code.mustache
nicknytko: Don't think so, Pygments will automatically wrap its stuff with a `<pre>` tag so that's why I took it out, I think that's all we need to ensure formatting.

mwest1066: Awesome! In that case the potential breakage for @gchallen is all we need to worry about.

nicknytko:
@mwest1066 we're good to go with @gchallen 's stuff and I've updated the docs to point to all the languages Pygments supports, I'd just rather you be the one to press the "Squash and merge" button here :slightly_smiling_face:

PR #2367, opened 2020-04-09
Add double-submit protection and refactor confirm-unload script

### LINKED ISSUE #854: Double-click on "Grade" can result in a double-submission

### LINKED ISSUE #1812: Disable-on-submit/click button classes are not used for v3 questions

### LINKED ISSUE #2234: Disable grade button on click, also for externally-graded questions
This was implemented in #2187 but it turns out that it broke repeated submissions to externally-graded questions. I reverted that change in #2233. It would be great to re-do #2187 while taking externally-graded questions into account.

Externally-graded questions are different in that they still do a form-submit and page-reload to submit the answers, but because they are not yet graded they use websockets to update the grading information after page load. Something about this (the ungraded status? websocket updates?) seems to interact badly with the logic we were using to prevent double-clicks on the Grade button.

tdy:
Another attempt to fix #854 and #1812 -- hopefully compatible with externally graded questions this time (e.g., `demoCodeEditorAutograded`).

~~Instead of disabling buttons upon form submit (2e0b8eb and #2187), this PR disables the form upon button clicks. Can't find a good way to test this systematically, but I haven't managed to break it yet with click spamming. There's also fewer moving parts (no cloning+hiding+rearranging buttons), so hopefully it's less finicky?~~

**Removes `else` block from disable-on-submit script:**

<a href="https://github.com/PrairieLearn/PrairieLearn/pull/2367/commits/0a13fcfaaf85aaabf4d9ee8da2c9f43be1868191#diff-86879b23fea7876ca06d461a05a9b1c9L109-L110"></a>

Fixes #2234

---

* [x] Fix disable script
* [x] Refactor disable-submit script
* [x] Refactor disable-click script
* [x] Refactor confirm-unload script

---

Double check v2 regressions:

* [x] Disable-submit works (save, save+grade) -- did not verify autograded v2 questions, but Matt is fine with that
* [x] Disable-click works (new variant)
* [x] ~~Confirm-unload does not work~~, but Matt is ok with dropping support as long there's no breakage (#2612)


tdy:
Not sure what will happen with this PR when `tyang15-fix-const-redeclare` gets merged -- guessing I'll end up having to redo this PR to merge into `master` later.

pages/partials/questionFooter.ejs
```diff
@@ -90,34 +90,22 @@
...
   </div>
 <% } %>
-
+<script>
+    // Instead of disabling buttons upon form submit (2e0b8eb and #2187),
+    // disable the form upon button clicks to resolve issue with
+    // externally graded questions (#2234).
+    $(document).ready(() => {
+        const form = $('form.question-form');
+        $('button.disable-on-submit').click(function() {
```
mwest1066, line 101: I worry a bit that if something interferes with JS execution (after we've installed this onclick handler) then we'll be unable to submit the form. But I think that even if a question contains bad JS, this handler will continue to run? Is it possible to test robustness to JS errors inside questions?

pages/partials/questionFooter.ejs
```diff
@@ -90,34 +90,22 @@
...
-          $(this).addClass('disabled');
-      });
-    </script>
   </div>
 <% } %>
-
+<script>
+    // Instead of disabling buttons upon form submit (2e0b8eb and #2187),
+    // disable the form upon button clicks to resolve issue with
+    // externally graded questions (#2234).
```
mwest1066, line 98: Nice!

pages/partials/questionFooter.ejs
tdy: In general it's probably fine (but probably still browser-dependent): https://stackoverflow.com/a/29394912/13138364

> An uncaught runtime error only stops the current execution, which may be
>
> * the execution of a script
> * the call of an event handler
>
> Suppose you have a runtime error while handling an event, the only problem you might have (apart not really handling the event) is a non consistent state of your user variables if your event handler modifies some of them. Other event handlers won't be impacted besides that.
>
> So it can usually be considered as non catastrophic (I guess I don't have to remember it's a good practice to fix errors anyways and that flooding the console with errors isn't a good thing).

tdy: I tested adding a broken script above the double-submit script:
```javascript
<script>
    alert('malformed alert
</script>
```

* [x] Firefox: still works
* [x] Chrome: still works
* [ ] Edge: does **NOT** work

Actually upon further testing, I'm not sure this solution works in Edge even without the broken script. It was working yesterday, but now it's giving me `unknown __action: undefined` 😩

mwest1066: Huh! That's weird that Edge is different. The `unknown __action: undefined` means the server got a POST with the `body` containing `__action: undefined`. Normally `__action` should be set by the submit button `value`: https://github.com/PrairieLearn/PrairieLearn/blob/master/pages/partials/questionFooter.ejs#L10

nicknytko: Interesting, I don't know what was mentioned during OH today since I was working on other things, but my thoughts here are that `__action` only gets set when the form gets submitted through the button, and when the form gets submit via js there's no `__action` value set.  Maybe you can append a hidden input tag to get around this?

tdy: With some hackery, the hidden `input` fixes the POST issue on Edge... but unfortunately the form disabling behavior is somehow totally ignored on Edge. I'm not even sure how/why 😩

nicknytko: Do you have to set the button to have `type=“button”`?  I think by default it will “submit” the page.

tdy: You mean to prevent the bound `button` from "submitting" along with the JS `form.submit()`?

nicknytko: Right, unless I'm mistaken the default behaviour is to submit the form on button press?

mwest1066: I'd been hoping to have the debounce code function as progressive enhancement, so the page would still work without JS running. Of course this might not be possible, given the weird behavior from Edge.

echuber2: Are you talking about the old Edge or the new Chrome-based Edge?

If Edge doesn't respect disabling form elements directly, would it work to layer a semitransparent div on top of the controls to "disable" them, similar to the annoying nag screens that news media pop up?

tdy: Seems that I still have the non-Chromium Edge for some reason. Windows Update is up-to-date, and I'm reading that Chromium-Edge's public release was Jan 2020, so not sure why I'm still on v44.

tdy: Based on the original v2 solution where cloned buttons were not properly removed on externally graded questions, the same would happen with an overlay. It looks like we'll have to go full circle and debug that original problem.

mwest1066: I'm still hoping that the original problem is not too hard to debug (famous last words). Let me know if you need an extra set of eyes on it.

pages/partials/questionFooter.ejs
```diff
@@ -106,8 +106,6 @@
                 $(this).hide();
                 form.prepend($(this));
             });
-        } else {
-            e.preventDefault();
```
tdy: The culprit! (I think)

mwest1066:
Wow! That was a very small change to fix the failure on externally-graded questions!

I think I now understand what's happening. When an externally-graded question returns its grading results, we re-render a number of divs, including the question footer: https://github.com/PrairieLearn/PrairieLearn/blob/8b1051f71061aeb788f7280d9868b7dd6730d77f/pages/partials/externalGradingLiveUpdate.ejs#L98

The problem is that the question footer contains the scripts that attach the on-submit handler, so we end up with two copies of this handler: one from the initial page load, and one from the question footer re-render.

Maybe we should add a comment to the script tag to explain that this is going to execute every time we re-render the footer, which might be more than once for a given page load, so the handler needs to be ready for the fact that it might get installed multiple times?

This also means that the `beforeunload` handler is going to get installed multiple times. This doesn't actually cause errors (as far as I can tell) but it's a bit unexpected and might be worth a comment in the code?

Another possibility would be to shift these scripts out of the footer and into a separate JS file (maybe `public/localscripts/questionPageScripts.js`) and include this in the header of any page that will include `pages/partials/question.ejs`? I think there are only three such pages (`instructorQuestionPreview`, `studentInstanceQuestionExam`, `studentInstanceQuestionHomework`). This would mean that we are absolutely certain that these scripts are all being run only once per page.

pages/partials/questionFooter.ejs
```diff
@@ -90,34 +90,26 @@
...
   </div>
 <% } %>
+<script>
+    $('form.question-form').on('submit', function() {
+        if (!$(this).data('submitted')) {
+            $(this).data('submitted', true);
+
+            $(this).find('.disable-on-submit').each(function() {
+                // Create a disabled clone of the submit button
+                $(this).clone(false).removeAttr('id').prop('disabled', true).insertBefore($(this));
```
nicknytko, line 102: Maybe a dumb question, but what's the need for creating a new disabled button instead of setting the existing one to be disabled?  I see that you're doing that for link buttons so not sure what the difference is.

tdy: Ah that reminds me, I should update the comments. It took me a while to figure out why.

Long story short, disabled buttons don't POST (all browsers, not just Edge). Disabling the links is fine since they don't need to POST anything, but we need the buttons to POST the `__action`.

After trying a bunch of other hacks, Matt's original cloning method seems to be the most cross-compatible, progressive-enhancing workaround. The fix was just removing the `else` clause.

nicknytko:
> Wow! That was a very small change to fix the failure on externally-graded questions!
> 
> I think I now understand what's happening. When an externally-graded question returns its grading results, we re-render a number of divs, including the question footer:
> 
> https://github.com/PrairieLearn/PrairieLearn/blob/8b1051f71061aeb788f7280d9868b7dd6730d77f/pages/partials/externalGradingLiveUpdate.ejs#L98
> 
> The problem is that the question footer contains the scripts that attach the on-submit handler, so we end up with two copies of this handler: one from the initial page load, and one from the question footer re-render.
> 
> Maybe we should add a comment to the script tag to explain that this is going to execute every time we re-render the footer, which might be more than once for a given page load, so the handler needs to be ready for the fact that it might get installed multiple times?
> 
> This also means that the `beforeunload` handler is going to get installed multiple times. This doesn't actually cause errors (as far as I can tell) but it's a bit unexpected and might be worth a comment in the code?
> 
> Another possibility would be to shift these scripts out of the footer and into a separate JS file (maybe `public/localscripts/questionPageScripts.js`) and include this in the header of any page that will include `pages/partials/question.ejs`? I think there are only three such pages (`instructorQuestionPreview`, `studentInstanceQuestionExam`, `studentInstanceQuestionHomework`). This would mean that we are absolutely certain that these scripts are all being run only once per page.

+1 to have the script in a separate JS file, that seems like it would be cleaner and would let us add more client-side functionality in the future if needed.

Besides that, this looks good -- I was able to test on Firefox and Chromium and didn't get any double submits for internal or external questions.

tdy:
> This also means that the `beforeunload` handler is going to get installed multiple times. This doesn't actually cause errors (as far as I can tell) but it's a bit unexpected and might be worth a comment in the code?

I refactored the `disable` script, but I'm not sure we can move the `beforeunload` script. At least part of it needs to stay in the footer (or somewhere that's re-rendered) because we need to reinitialize the form state after each submission. I tried switching to globals and splitting the state-saving part (footer) from the handler (external), but I couldn't get it to work.

mwest1066:
For `disable-on-click` we need to separate the installation of the handler into a separate function, which is called both on page load and on re-render of the question footer in the websocket code: https://github.com/PrairieLearn/PrairieLearn/blob/1b249374d6f15e1468d53ce4e2b1a7efa02650cb/pages/partials/externalGradingLiveUpdate.ejs#L97

mwest1066:
The `unload` script needs to be shifted into the global JS script.

astrosticks:
> Refactor unload script (fixme: global script causes confirmation dialog on every unload now, even if the state hasn't changed)

Done. The reason why this was happening is because some elements (s/a pl-drawing, pl-file-editor) use `$(document).ready()` to populate some inputs. Putting `questionPageScripts.js` at the top of the question files was getting the value for `initialForm` before those inputs existed, so once they appeared, `initialForm` was always inaccurate.

323501c and e22f3b0 both fix this problem, I just thought the second one was easier to understand, even though `defer` is basically the `<script src="">` version of `$(document).ready()`.

public/localscripts/questionPageScripts.js
```diff
@@ -0,0 +1,60 @@
...
+                $(this).clone(false).removeAttr('id').prop('disabled', true).insertBefore($(this));
+
+                // Hide actual submit button
+                $(this).hide();
+            });
+        }
+    });
+}
+
+// also called in pages/partials/externalGradingLiveUpdate.ejs
```
tdy, line 48:
```suggestion
// Also called in pages/partials/externalGradingLiveUpdate.ejs
```

pages/partials/questionFooter.ejs
```diff
@@ -90,63 +90,10 @@
...
-
-                // Chrome legacy
-                event.returnValue = '';
-
-                // Fallback
-                return '';
-            }
-        });
-    });
+  $('a.disable-on-click').on('click', function() {
```
tdy, line 96:
```suggestion
  // Install on page load and reinstall on websocket re-render
  $('a.disable-on-click').on('click', function() {
```

tdy:
I think this one's finally good to go.

pages/studentInstanceQuestionExam/studentInstanceQuestionExam.ejs
```diff
@@ -97,5 +96,6 @@
         </div>
       </div>
     </div>
+    <script src="/localscripts/questionPageScripts.js"></script>
```
tdy, line 99:
```suggestion
    // Needs to be located below element scripts
    <script src="/localscripts/questionPageScripts.js"></script>
```

pages/studentInstanceQuestionHomework/studentInstanceQuestionHomework.ejs
```diff
@@ -76,5 +76,6 @@
         </div>
       </div>
     </div>
+    <script src="/localscripts/questionPageScripts.js"></script>
```
tdy, line 79:
```suggestion
    // Needs to be located below element scripts
    <script src="/localscripts/questionPageScripts.js"></script>
```

pages/instructorQuestionPreview/instructorQuestionPreview.ejs
```diff
@@ -39,5 +39,6 @@
       </div>
 
     </div>
+    <script src="/localscripts/questionPageScripts.js"></script>
```
tdy, line 42:
```suggestion
    // Needs to be located below element scripts
    <script src="/localscripts/questionPageScripts.js"></script>
```

tdy:
Update on `v2` questions:

* [x] Disable-submit works (save, save+grade)
* [x] Disable-click works (new variant)
* [ ] Confirm-unload does **not** work -- the script serializes `form.question-form` inputs, but v2 `input` fields are somehow located outside the form (**note:** confirm-unload is actually an unrelated feature, so it technically isn't in the original scope of this PR -- refactoring this script got lumped in as PR creep)

pages/instructorQuestionPreview/instructorQuestionPreview.ejs
```diff
@@ -39,5 +39,7 @@
       </div>
 
     </div>
+    // Needs to be located below element scripts
```
nicknytko, line 42:
```suggestion
    <!-- Needs to be located below element scripts -->
```

pages/studentInstanceQuestionExam/studentInstanceQuestionExam.ejs
```diff
@@ -97,5 +96,7 @@
         </div>
       </div>
     </div>
+    // Needs to be located below element scripts
```
tdy, line 99:
```suggestion
    <!-- Needs to be located below element scripts -->
```

pages/studentInstanceQuestionHomework/studentInstanceQuestionHomework.ejs
```diff
@@ -76,5 +76,7 @@
         </div>
       </div>
     </div>
+    // Needs to be located below element scripts
```
tdy, line 79:
```suggestion
    <!-- Needs to be located below element scripts -->
```

nicknytko: Aha good catch :slightly_smiling_face:

tdy: Somehow I didn't notice these until you caught the first one 🤦‍♂️

Issue #2154, opened 2020-03-10
Change "close all" issues to only close the visible ones


mwest1066:
On the Issues tab, the "Close all issues" button would be more useful it was "Close all visible issues", where we'd need to use the current filter to select what to close.

cheeren:
Came here to request this exact thing.

armandofox:
We could really use this as well, it's come up on a number of occasions at Cal.

nwalters512:
For anyone subscribed to this issue, #9077 was just deployed, which includes this functionality!

PR #2371, opened 2020-04-10
Change server.py test() to take a test_type parameter

### LINKED ISSUE #2269: Change server.py test() to take a test_type parameter
At the moment the `test()` function in `server.py` (and in elements) is supposed generate a testing submission and the expected parsing and grading result for that submission. PL uses this by calling `test()`, then taking the returned submission and running it through `parse()` and `grade()` to make sure that the result is as expected.

As an example: https://github.com/PrairieLearn/PrairieLearn/blob/master/elements/pl-number-input/pl-number-input.py#L376

`test()` should fill in `raw_submitted_answers` (not `submitted_answers`), `format_errors`, `partial_scores`, and `score` (only for `server.py`, not for elements?).

We want to test the handling of three types of submissions: correct, incorrect, and invalid. At the moment we do this by having each `test()` function randomly decide which of these it's going to generate. This is bad because PL can't ensure that it's tested one submission of each type.

To fix this, we should change `test()` to accept a `test_type` parameter with a value of `correct`, `incorrect`, or `invalid`, and it should generate a submission with this type. This will require changing all elements that currently have `test()` functions.

We can then also change `startTestQuestion()` in `lib/question.js` to always iterate through a cycle of correct/incorrect/invalid submissions.

See also #1214 and #1700

nicknytko:
Creates a new `test_type` key in `data` that is set to either `correct`, `incorrect`, or `invalid` when `test()` is run.  Running tests will cycle through all of these (or a random subset if running less than 3).

I put `test_type` in data as to not break custom elements and existing functions in `server.py`, these will just ignore the new key.

Resolves #2269


mwest1066:
I'm wondering if we should change the "Test once" button to test all three types?

nicknytko:
> I'm wondering if we should change the "Test once" button to test all three types?

That seems like a good change to make, should we also keep the other button at 100 tests?

mwest1066:
> That seems like a good change to make, should we also keep the other button at 100 tests?

I guess so? Or 99? Or leave it at 100 but actually run 100 x 3 tests in total (so 100 times as much as the "single test" button). Whatever you think best.

nicknytko:
> > That seems like a good change to make, should we also keep the other button at 100 tests?
> 
> I guess so? Or 99? Or leave it at 100 but actually run 100 x 3 tests in total (so 100 times as much as the "single test" button). Whatever you think best.

I've set it so that 1 test = 1 of each type.  So "single test" = 1x each incorrect, correct, invalid, and "100 tests" is 100x each.

mwest1066:
> I've set it so that 1 test = 1 of each type. So "single test" = 1x each incorrect, correct, invalid, and "100 tests" is 100x each.

Perfect!

mwest1066:
> I definitely support the switch to testing 3x for a single test run. But, I'm slightly concerned about a button that launches a 100x3 testing scheme -- especially if the button is pressed by mistake.

You can always click away from the test --- it will just continue in the background. Of course, if it finds issues then you might end up with a lot of them. Which would increase the desire for #2154.

> * Perhaps a confirmation should appear prior to testing in the 100x3 case?

Can't hurt, I guess. But I wouldn't insist on it.

> * Also, what would the impact be of someone pressing a 100x button on production? (Negligible I assume, but with the shift to in-browser editing and more students than ever before slamming PL...)

Good load testing for us! But honestly, no problem.

coatless:
> You can always click away from the test --- it will just continue in the background. Of course, if it finds issues then you might end up with a lot of them. 

Maybe in addition to a close all button, the issues themselves should be threaded or grouped by QID? This would allow a more targeted close. 

Instead of a confirmation, maybe just add a sentence to the testing view stating you can move away? The current test view doesn't emphasize that. 
 
Screen Shot 2020-04-14 at 10 08 23 AM

nicknytko:
> > You can always click away from the test --- it will just continue in the background. Of course, if it finds issues then you might end up with a lot of them.
> 
> Maybe in addition to a close all button, the issues themselves should be threaded or grouped by QID? This would allow a more targeted close.
> 
> Instead of a confirmation, maybe just add a sentence to the testing view stating you can move away? The current test view doesn't emphasize that.
> 
> Screen Shot 2020-04-14 at 10 08 23 AM

+1 to add text to the testing page.  

Maybe the issues page can be revamped in a separate PR?  I'm trying to reduce the amount of feature creep -- I've been pretty bad about that in the past 😄 
I'm thinking things like more ways to filter issues and maybe an overview/threaded view like you mentioned.

mwest1066:
> Maybe the issues page can be revamped in a separate PR? I'm trying to reduce the amount of feature creep -- I've been pretty bad about that in the past 😄

Oh yes! Definitely not for this PR.

> I'm thinking things like more ways to filter issues and maybe an overview/threaded view like you mentioned.

So nice! But later 😄

coatless:
> Maybe the issues page can be revamped in a separate PR? I'm trying to reduce the amount of feature creep -- I've been pretty bad about that in the past 😄

@nicknytko who do you think started pushing that line? 😉 

I'll throw the musing in another issue ticket. 

---

In all seriousness, I think this is good to go with just a slight phrase addition to the test view:

> Please feel free to navigate away from this page. If any issues arise during question tests, they will appear under the "Issues" page.

pages/partials/jobSequenceResults.ejs
```diff
@@ -68,6 +68,9 @@
             <p class="mb-1">
                 Started at <%= job.start_date_formatted %><% if (job_sequence.user_uid) { %> by <%= job_sequence.user_uid %><% } %><% if (job_sequence.authn_user_uid != job_sequence.user_uid) { %> (really <%= job_sequence.authn_user_uid %>)<% } %><% if (job.finish_date_formatted) { %>, finished at <%= job.finish_date_formatted %><% } %>
             </p>
+            <p class="mb-1">
+                Please feel free to navigate away from this page. If any issues arise they will appear under the "Issues" page.
```
mwest1066, line 72: I don't think it's generally true that all job issues will be filed on the Issues page. That's true for question issues arising from `test()`, but the job sequence page is also used for jobs like course syncing which will report errors on the Sync page. I'm tempted to just not add this message here at all? As a comparison, the GitHub Action page doesn't say anything about being able to navigate away, they just assume you know you can.

lib/question.js
```diff
@@ -1007,9 +1013,10 @@ module.exports = {
                     return;
                 }
 
-                async.eachSeries(_.range(count), (iter, callback) => {
-                    job.verbose(`Test ${iter + 1}`);
-                    this._runTest(job, showDetails, question, course_instance, course, authn_user_id, (err, ret_success) => {
+                async.eachSeries(_.range(count * test_types.length), (iter, callback) => {
+                    let type = test_types[iter % test_types.length];
+                    job.verbose(`Test ${iter + 1}, type ${type}`);
```
mwest1066, line 1018:
```suggestion
                    job.verbose(`Test ${Math.floor(iter/test_types.length) + 1}, type ${type}`);
```

Very minor, but this makes it so clicking "Test 100 times" actually counts up to 100, not 300.

pages/partials/jobSequenceResults.ejs
nicknytko: I'd rather it only show up on sync, but since all the jobs use the same page maybe I'll just take it out for now.

PR #2378, opened 2020-04-14
Separating admin overview into tabs

### LINKED ISSUE #2347: Split up "Admin" page into tabs
The current "Admin" page is getting quite long, so it would be nice if it was broken up into tabs. This would also allow "Queries" to be a tab underneath "Admin", which makes logical sense.

astrosticks:
Resolves #2347, adding tabs inside the administrator page to break up longer sections.

Screenshots:
image
___
image
___
image
___
image


trombonekenny:
I would prefer institutions and courses maintained in the same page.

mwest1066:
This is awesome!!

Could we go further and break up the "Server" page, so that Administrators, Config, and Networks are all separate pages?

It turns out that we don't actually need the information about the question render cache (we monitor that via AWS CloudWatch) so we could just delete all the code that computes and shows the cache statistics. This might also mean we can remove the `chart.js` package entirely, as I think this is the only page it's used on.

Unfortunately we will still need the "reset cache" button. I'm not sure whether to have a separate page (maybe "Actions"?) with just this one button on it at the moment (there may be more later), or whether we should stick it on one of the other pages? But which one? Maybe it really is better to have an Actions page, even though it's going to look a bit silly with only a single button.

mwest1066:
As a stretch goal, we could add some basic tests to at least GET each of these pages and check it doesn't throw an error when rendering. This would help to get us above the 60% code coverage level that we've been stuck at for a long time. We could even test basic POST functionality, like adding a new course, but the basic GET functions would already be awesome.

An example of the best current test layout we have is https://github.com/PrairieLearn/PrairieLearn/blob/master/tests/testAdministratorQueries.js

A slightly more complicated test layout that also does some POSTs is https://github.com/PrairieLearn/PrairieLearn/blob/master/tests/testShowClosedAssessment.js

astrosticks:
@mwest that's what 64d3d9ce476c0abaa3d4b65358820ddb2d80781a tries to do ;)

mwest1066:
> @mwest that's what [64d3d9c](https://github.com/PrairieLearn/PrairieLearn/commit/64d3d9ce476c0abaa3d4b65358820ddb2d80781a) tries to do ;)

Haha, fair enough! 😄 

In that case, all that's left is the question of whether you can break things up even further?

astrosticks:
Definitely, I'll get that out soon

mwest1066:
> Definitely, I'll get that out soon

Awesome!

astrosticks:
Alright, I've completely split up the admin pages. Unsure of what exactly is preventing the CI builds from passing, and we still need to add a page to contain the "Reset cache" button, but that's where we are right now.

trombonekenny:
My 2c would be to put the Reset Cache button under the AdministratorServer page - that fits there, and there's not much on that page so having separate vertical sections wouldn't be overwhelming.

pages/partials/navTabs/administrator.ejs
```diff
@@ -0,0 +1,25 @@
+<%- include('navTab', {
+    activeSubPage:  'administrators',
+    urlSuffix:      '/administrator/admins',
+    iconClasses:    'fas fa-user-shield',
+    tabLabel:       'Administrators'}); %>
+<%- include('navTab', {
+    activeSubPage:  'server',
+    urlSuffix:      '/administrator/server',
+    iconClasses:    'fas fa-server',
+    tabLabel:       'Server'}); %>
```
mwest1066, line 10: Could we rename the "Server" tab to be "Settings", to match the other PL pages that contain setting/config information (e.g., `pages/instructorCourseAdminSettings` and `instructorAssessmentSettings`).

Also, the convention we were using on the other pages is to put the tabs in alphabetical order. I'm not sure alphabetical is the best order, but at least it makes it possible to find things 😄 --- the tab order here is almost alphabetical, except that this one tab needs to move.

pages/partials/navTabs/administrator.ejs
```diff
@@ -0,0 +1,25 @@
...
+<%- include('navTab', {
+    activeSubPage:  'queries',
+    urlSuffix:      '/administrator/queries',
+    iconClasses:    'fas fa-database',
+    tabLabel:       'Queries'}); %>
+<%- include('navTab', {
+    activeSubPage:  'settings',
+    urlSuffix:      '/administrator/settings',
+    iconClasses:    'fas fa-cog',
+    tabLabel:       'Settings'}); %>
```
mwest1066, line 25: This tab doesn't activate for me when on the Settings page.

pages/administratorServer/administratorServer.ejs
```diff
@@ -0,0 +1,49 @@
...
+          $('[data-toggle="popover"]').popover({sanitize: false})
+      });
+    </script>
+    <%- include('../partials/navbar', {navPage: 'admin', navSubPage: 'server'}); %>
+    <div id="content" class="container-fluid">
+
+      <!-------------------------------------------------------------------------------->
+      <!-------------------------------------------------------------------------------->
+      <!-------------------------------------------------------------------------------->
+      <!-- Config ---------------------------------------------------------------------->
```
mwest1066, line 18: Could delete these comments now.

mwest1066 (approved):

pages/administratorServer/administratorServer.ejs
nwalters512: Turns out this whole page was unused; I removed it in cdf33e0.

pages/partials/navTabs/administrator.ejs
nwalters512: Fixed in 2d9f83b.

PR #2383, opened 2020-04-17
Change the plbase image to use a separate install script

mwest1066:
This shouldn't change the plbase image build in any way, but it will
allow us to re-use this install script for production builds and thus
ensure that prod and test are using the same configuration.


mwest1066:
> Looks fine to me. Do we need to update anything so that CI will pick up changes to the `install.sh` and re-build the base container?

I think it should all just work. The CI runs https://github.com/PrairieLearn/PrairieLearn/blob/master/.github/workflows/main.yml#L14

This checks for changes in `environments/centos7-plbase` and rebuilds if it finds any: https://github.com/PrairieLearn/PrairieLearn/blob/master/tools/build-plbase-if-needed.sh#L3

I think this will all still work with this PR, because the `plbase` docker config is still entirely contained within the `environments/centos7-plbase` directory.

Does that sound right to you?

nicknytko:
> > Looks fine to me. Do we need to update anything so that CI will pick up changes to the `install.sh` and re-build the base container?
> 
> I think it should all just work. The CI runs https://github.com/PrairieLearn/PrairieLearn/blob/master/.github/workflows/main.yml#L14
> 
> This checks for changes in `environments/centos7-plbase` and rebuilds if it finds any: https://github.com/PrairieLearn/PrairieLearn/blob/master/tools/build-plbase-if-needed.sh#L3
> 
> I think this will all still work with this PR, because the `plbase` docker config is still entirely contained within the `environments/centos7-plbase` directory.
> 
> Does that sound right to you?

Sure, this looks good then.

trombonekenny:
It's a minor nit, but I'd probably name it something more descriptive like `plbase-install.sh` or `prairielearn-<image>-install.sh` so when we look at files in / later, we'll know what this installs.

PR #2384, opened 2020-04-17
Allow loading of dependencies per-question

### LINKED ISSUE #2271: Question code should have dependencies like elements
Elements can require dependencies in their `info.json` files. These dependencies are then injected into the page header. We should support the same thing for questions.

nicknytko:
Resolves #2271 

Questions can now load dependencies from the following:
Property | Description
--- | ---
`coreStyles` |  The styles required by this question, relative to `[PrairieLearn directory]/public/stylesheets`.
`coreScripts` | The scripts required by this question, relative to `[PrairieLearn directory]/public/javascripts`.
`nodeModulesStyles` | The styles required by this question, relative to `[PrairieLearn directory]/node_modules`.
`nodeModulesScripts` | The scripts required by this question, relative to `[PrairieLearn directory]/node_modules`.
`clientFilesQuestionStyles` | The scripts required by this question relative to the question's `clientFilesQuestion` directory.
`clientFilesQuestionScripts` | The scripts required by this question relative to the question's `clientFilesQuestion` directory.
`clientFilesCourseStyles` | The styles required by this question relative to `[course directory]/clientFilesCourse`.
`clientFilesCourseScripts` | The scripts required by this question relative to `[course directory]/clientFilesCourse`. 

Also adds a _beautiful_ new demo question, `demoCustomStylingScripts`:

image

### DIFF
--- a/database/tables/questions.pg
+++ b/database/tables/questions.pg
@@ -2,6 +2,7 @@ columns
     client_files: text[] default ARRAY[]::text[]
     course_id: bigint not null
     deleted_at: timestamp with time zone
+    dependencies: jsonb default '{}'::jsonb
     directory: text
     external_grading_enable_networking: boolean
     external_grading_enabled: boolean default false
--- a/database/tables/questions.pg
+++ b/database/tables/questions.pg
@@ -2,7 +2,7 @@ columns
     client_files: text[] default ARRAY[]::text[]
     course_id: bigint not null
     deleted_at: timestamp with time zone
-    dependencies: jsonb default '{}'::jsonb
+    dependencies: jsonb not null default '{}'::jsonb
     directory: text
     external_grading_enable_networking: boolean
     external_grading_enabled: boolean default false
--- a/docs/devElements.md
+++ b/docs/devElements.md
@@ -153,7 +153,7 @@ Property | Description
 `elementStyles` | The styles required by this element relative to the element's directory, which is either `[PrairieLearn directory]/elements/this-element-name` or `[course directory]/elements/this-element-name`.
 `elementScripts` | The scripts required by this element relative to the element's directory, which is either `[PrairieLearn directory]/elements/this-element-name` or `[course directory]/elements/this-element-name`.
 `clientFilesCourseStyles` | The styles required by this element relative to `[course directory]/clientFilesCourse`. *(Note: This property is only available for elements hosted in a specific course's directory, not system-wide PrairieLearn elements.)*
-`clientFilesCourseStyles` | The styles required by this element relative to `[course directory]/clientFilesCourse`. *(Note: This property is only available for elements hosted in a specific course's directory, not system-wide PrairieLearn elements.)*
+`clientFilesCourseScripts` | The scripts required by this element relative to `[course directory]/clientFilesCourse`. *(Note: This property is only available for elements hosted in a specific course's directory, not system-wide PrairieLearn elements.)*
 
 You can also find the types of dependencies defined in these schema files:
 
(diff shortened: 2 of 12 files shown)


docs/devElements.md
```diff
@@ -153,7 +153,7 @@ Property | Description
 `elementStyles` | The styles required by this element relative to the element's directory, which is either `[PrairieLearn directory]/elements/this-element-name` or `[course directory]/elements/this-element-name`.
 `elementScripts` | The scripts required by this element relative to the element's directory, which is either `[PrairieLearn directory]/elements/this-element-name` or `[course directory]/elements/this-element-name`.
 `clientFilesCourseStyles` | The styles required by this element relative to `[course directory]/clientFilesCourse`. *(Note: This property is only available for elements hosted in a specific course's directory, not system-wide PrairieLearn elements.)*
-`clientFilesCourseStyles` | The styles required by this element relative to `[course directory]/clientFilesCourse`. *(Note: This property is only available for elements hosted in a specific course's directory, not system-wide PrairieLearn elements.)*
+`clientFilesCourseScripts` | The scripts required by this element relative to `[course directory]/clientFilesCourse`. *(Note: This property is only available for elements hosted in a specific course's directory, not system-wide PrairieLearn elements.)*
```
coatless, line 156: Nice fix!

docs/question.md
```diff
@@ -62,9 +62,52 @@ Property | Type | Description
 `singleVariant` | boolean | Whether the question is not randomized and only generates a single variant. (Optional; default: `false`)
 `partialCredit` | boolean | Whether the question will give partial points for fractional scores. (Optional; default: `true`)
 `externalGradingOptions` | object | Options for externally graded questions. See the [external grading docs](externalGrading.md). (Optional; default: none)
+`dependencies` | object | External JavaScript or CSS dependencies to load.  See below.  (Optional; default: `{}`)
 
 For details see the [format specification for question `info.json`](https://github.com/PrairieLearn/PrairieLearn/blob/master/schemas/schemas/infoQuestion.json)
 
+### Question Dependencies
+
```
coatless, line 70: I suppose a question would be does a similarly named element load that elements dependencies prior to loading the custom element? Or is it a clean break?

I'm 95% sure it's a clean break. But, this probably should be emphasized here to avoid CSS hackers thinking parts of an element can be re-used/extended.

docs/question.md
```diff
@@ -62,9 +62,52 @@ Property | Type | Description
...
+Your quesition can load client-side assets such as scripts or stylesheets from different sources.  A full list of dependencies will be compiled based on the question's needs and any dependencies needed by page elements, then they will be de-duped and loaded onto the page.
```
mwest1066, line 71:
```suggestion
Your question can load client-side assets such as scripts or stylesheets from different sources.  A full list of dependencies will be compiled based on the question's needs and any dependencies needed by page elements, then they will be de-duped and loaded onto the page.
```

docs/question.md
mwest1066: I'm not sure I understand this question?

database/tables/questions.pg
```diff
@@ -2,6 +2,7 @@ columns
     client_files: text[] default ARRAY[]::text[]
     course_id: bigint not null
     deleted_at: timestamp with time zone
+    dependencies: jsonb default '{}'::jsonb
```
mwest1066, line 5: Is the code able to handle `dependencies = NULL` or should we add a `NOT NULL` constraint here?

mwest1066, line 5: I generally hate storing JSON blobs in the DB, rather than making a proper table to store the data, but in this case I think it's ok because the server code is going to want this data in the same format as the dependency data from elements. Is that right?

nicknytko: Right

nicknytko: No I don’t think it would, nice catch here.

docs/question.md
coatless: @mwest1066 

If there is an implementation of `pl-number-input`  in both the the **core elements/built in to PrairieLearn** and a **local course**, are these two implementations independent from each other? 

That is, I shouldn't expect the CSS from `pl-number-input` in the core to be available within the local course. Thus, if I wanted to change a single CSS rule it would require a complete duplication of all contents for `pl-number-input`.

nicknytko: Right, my understanding is that only the course element will get loaded here.

mwest1066: That's also my understanding, but I haven't checked the code.

PR #2387, opened 2020-04-17
Allow blank reference answer for pl-matrix-component-input

nicknytko:
Fixing bug discovered by Cinda on Slack.  Adds `rows` and `columns` attributes to `pl-matrix-component-input` to be used when there is no value set in `data['correct_answers']`.

### DIFF
--- a/elements/pl-matrix-component-input/pl-matrix-component-input.py
+++ b/elements/pl-matrix-component-input/pl-matrix-component-input.py
@@ -22,7 +22,18 @@ def prepare(element_html, data):
     required_attribs = ['answers-name']
     optional_attribs = ['weight', 'label', 'comparison', 'rtol', 'atol', 'digits', 'allow-partial-credit', 'allow-feedback', 'allow-fractions', 'rows', 'columns']
     pl.check_attribs(element, required_attribs, optional_attribs)
-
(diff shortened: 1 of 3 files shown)


elements/pl-matrix-component-input/pl-matrix-component-input.py
```diff
@@ -39,17 +39,27 @@ def render(element_html, data):
...
+            if m < 1:
+                raise Exception('Number of rows in pl-matrix-component-input must be strictly positive.')
+
+            n = pl.get_integer_attrib(element, 'columns', None)
+            if n is None:
+                raise Exception('Number of columns is not set in pl-matrix-component-input with no correct answer.')
+            if n < 1:
+                raise Exception('Number of columns in pl-matrix-component-input must be strictly positive.')
         else:
             if np.isscalar(a_tru):
```
mwest1066, line 54: Should we check that `rows` and `columns` are `None` here?

nicknytko:
@mwest1066 this took longer than I expected and I’m not even sure Cinda still needs this, but this is ready for review

mwest1066:
> @mwest1066 this took longer than I expected and I’m not even sure Cinda still needs this, but this is ready for review

Thanks! It's definitely needed in any case, so it's great to have it fixed.

PR #2390, opened 2020-04-18
Fixing unclear navbar symbol, text baseline alignment, broken active elements

### LINKED ISSUE #2350: Separator symbols in instructor navbar don't clearly communicate what they are
Update: See https://github.com/PrairieLearn/PrairieLearn/pull/2390#issuecomment-615564050
___

This leads to some confusing wrapping on thin screens: 

image

astrosticks:
[Updated 5/12] Fixes #2350 by replacing the circle separators in the instructor navbar with slashes `/` to more clearly communicate their purpose.

|Comparison|
|---|
|Before|
|image|
|After| 
|image|

Some other minor cosmetic changes: 
- If an item in the navbar is active and has a dropdown arrow attached, the dropdown arrow will no longer show itself as active: 
image
(Hard to see, but the arrow isn't white anymore)
- Items in the navbar are now aligned at their baselines, including the "PrairieLearn" brand text. 
image

### DIFF
--- a/pages/partials/navbar.ejs
+++ b/pages/partials/navbar.ejs
@@ -7,7 +7,7 @@
   <button class="navbar-toggler" type="button" data-toggle="collapse" data-target=".navbar-collapse" aria-expanded="false" aria-label="Toggle navigation">
     <span class="navbar-toggler-icon"></span>
   </button>
-  <div class="collapse navbar-collapse">
+  <div id="course-nav" class="collapse navbar-collapse">
     <ul class="nav navbar-nav mr-auto" id="main-nav">
         <% if (typeof navbarType == 'undefined' || navbarType == 'plain') { %>
         <%- include('navbarPlain'); %>
--- a/pages/partials/navbarInstructor.ejs
+++ b/pages/partials/navbarInstructor.ejs
@@ -30,8 +30,8 @@
         <!-------------------------------------------------------------------------------->
         <!-- Instance Admin -------------------------------------------------------------->
 
+        <span class="navbar-text mx-2 no-select">/</span>
         <li class="nav-item btn-group">
-            <span class="navbar-text mx-2 no-select">/</span>
             <a class="nav-link <% if (navPage == "instance_admin" && !(typeof navSubPage !== 'undefined' && (navSubPage == "assessments" || navSubPage == "gradebook"))) { %>active<% } %>" href="<%= plainUrlPrefix %>/course_instance/<%= course_instance.id %>/instructor/instance_admin" role="button">
                 <%= course_instance.short_name %>
             </a>
@@ -57,8 +57,8 @@
         <li class="nav-item <% if (navPage == "instance_admin" && typeof navSubPage !== 'undefined' && navSubPage == "gradebook") { %>active<% } %>"><a class="nav-link" href="<%= urlPrefix %>/instance_admin/gradebook">Gradebook</a></li>
 
         <% if (typeof assessment_label != 'undefined' && typeof assessment != 'undefined') { %>
+            <span class="navbar-text mx-2 no-select">/</span>
             <li class="nav-item btn-group">
-                <span class="navbar-text mx-2 no-select">/</span>
                 <a class="nav-link <% if (navPage == "assessment") { %>active<% } %>" href="<%= urlPrefix %>/assessment/<%= assessment.id %>" role="button">
                     <%= assessment_label %>
                 </a>
--- a/pages/partials/navbarInstructor.ejs
+++ b/pages/partials/navbarInstructor.ejs
@@ -30,8 +30,8 @@
         <!-------------------------------------------------------------------------------->
         <!-- Instance Admin -------------------------------------------------------------->
 
-        <li class="nav-item btn-group <% if (navPage == "instance_admin" && !(typeof navSubPage !== 'undefined' && (navSubPage == "assessments" || navSubPage == "gradebook"))) { %>active<% } %>">
-            <a class="nav-link" href="<%= plainUrlPrefix %>/course_instance/<%= course_instance.id %>/instructor/instance_admin" role="button">
+        <li class="nav-item btn-group">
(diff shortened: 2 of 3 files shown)


tdy:
What's the purpose of those circles anyway?

astrosticks:
I think the idea is to give them a color to help users tell which specific course / course instance / assessment that they're looking at, but that doesn't currently exist. Assessment sets already have their own color, so that could go in the assessment circle.

mwest1066:
> What's the purpose of those circles anyway?

They are intended as a divider between course-level navbar items and course-instance-level items.

But clearly that isn't evident 😄 

In a navbar like:
PrairieLearn   CS 233   Questions    Issues   Sp20   Assessments

there are three "groups" of items. First is "PrairieLearn" as the main site. Then "CS 233, Questions, Issues" are all course-level. Finally "Sp20, Assessments" are both course-instance-level.

astrosticks:
> They are intended as a divider between course-level navbar items and course-instance-level items.

_That's_ what they are??? I spent all this time thinking that they were attached to the dropdowns hahaha

mwest1066:
> _That's_ what they are??? I spent all this time thinking that they were attached to the dropdowns hahaha

Lol. It's sort of a coincidence that the dropdowns are the first items in each new group.

Logically, we have PrairieLearn -> CS 233 -> Sp20 -> E1 (PL includes a course, which includes a semester, which includes an assessment).

These are really the important navbar items. The other ones (Questions, Issues, Assessments) are really just shortcuts to particular tabs on the second-level tabbar.

They are grouped so that Questions (which is a shortcut to the Questions tab under CS 233) is in the same group as the CS 233 navbar item.

mwest1066:
I'm completely open to eliminating the circles and having some other way to indicate the hierarchy in the navbar. Ideas welcome!

The nav structure overall sort of has two dimensions. First there is the "horizontal" dimension of PL -> CS 233 -> Sp20 -> E1. Then there is a "vertical" dimension under each of those items, which we represent with the tabbar. So under CS 233 there is a tabbar with Questions, Issues, etc. And under Sp20 there is a different tabbar.

mwest1066:
Here's how I view the navigation structure of PL. There is a tree structure of courses, course-instances, assessments. Then any of these things has a set of pages (tabs) corresponding to it.

```
PL
|
+- CS 233 -> Access, Questions, Issues, Tags
|     |
|     +- Sp20 -> Assessments, Access, Gradebook
|     |    |
|     |    +- Exam1 -> Downloads, Files, Statistics
|     |    |
|     |    +- Exam2 -> Downloads, Files, Statistics
|     |
|     +- Fa19 -> Assessments, Access, Gradebook
|          |
|          +- Exam1 -> Downloads, Files, Statistics
|
+- TAM 212 -> Access, Questions, Issues, Tags
      |
      +- Sp20 -> Assessments, Access, Gradebook
      |    |
      |    +- Homework1 -> Downloads, Files, Statistics
      |
      +- Fa19 -> Assessments, Access, Gradebook
           |
           +- Homework1 -> Downloads, Files, Statistics
           |
           +- Homework2 -> Downloads, Files, Statistics
```

mwest1066:
Anything to help users better navigate this structure would be great!

tdy:
I see. Now that it's laid out, it does make sense. For newer users, I think the PL nav is too busy to understand the intended logic.

Gut reaction that may not pass inspection: Feels like there's some mixing of breadcrumb and navbar patterns here. If so, separating them would improve clarity/usability.

tdy:
But I think for starters, just switching to smaller dots would make it clear that they are separators instead of icons/indicators.

tdy:
Another alternative is the megamenu route given how much info the PL nav tries to display, though megamenus also have their cons.

mwest1066:
> But I think for starters, just switching to smaller dots (or pipes or angle brackets) would make it clear that they are separators instead of icons/indicators.

I'm certainly on board with that.

mwest1066:
> I see. Now that it's laid out, it does make sense. For newer users, I think the PL nav is too busy to understand the intended logic.

Totally agreed that it's very dense. It's the usual challenge of trying to allow experienced users to rapidly navigate a complex structure, while also making it clear for new users. I'm not saying we succeeded at this latter objective at all 😄 

> Gut reaction that may not pass inspection: Feels like there's some mixing of breadcrumb and navbar patterns here. If so, separating them would improve clarity/usability.

I agree there is a mixing of the two here. It's actually worse, because the top navbar structure of "PL -> CS 233 -> Sp20 -> E1" is a sort of "recent breadcrumb trail". That is, we maintain this breadcrumb even when we jump back to the "CS 233" level to look at the Questions list. This lets us maintain a "currently active" assessment (here it's "E1") even when we are jumping back to look at a course-level page.

The intention of this "active object" notion is that it helps with fast nav, because if you are looking at an assessment you can jump over to the questions list and then get back to your assessment, without having to re-navigate all the way down the tree again.

But maybe there are other, better ways of allowing nav back to "recently viewed" pages?

echuber2:
> But maybe there are other, better ways of allowing nav back to "recently viewed" pages?

Related suggestion: #2391

tdy:
> The intention of this "active object" notion is that it helps with fast nav, because if you are looking at an assessment you can jump over to the questions list and then get back to your assessment, without having to re-navigate all the way down the tree again.

😲🤯 I've been doing just that -- re-navigating back and forth manually -- without realizing the shortcut design of all the nav items.

mwest1066:
> 😲🤯 I've been doing just that -- re-navigating back and forth manually -- without realizing the shortcut design of all the nav items.

Oh no! Clearly this nav design needs to somehow be made clearer or simpler.

tdy:
Actually I should say that most of the time I've worked around it via multiple browser tabs, but when I do navigate within the same tab, I've been doing it manually 🤦‍♂️ 

I'll mess around sometime with some different nav ideas.

mwest1066:
I like the change to the baseline and the non-active dropdown triangle.

But I don't like the fact that the circle (the separator icon) is now closer to "Sp15" than it is to "Questions". Before the circle was equally spaced between "Questions" and "Sp15", which makes sense given its role as a separator.

I also don't like the fact that the circle is now active (white) when the course instance "Sp15" is active. I think the circle should always be inactive, just like the dropdown triangle.

I guess both of these points are a reflection of the fact that the circle is a section separator, not something attached to the "Sp15" item. I'd be happy to replace the circle with something else (an arrow? A vertical bar?) if that would be clearer.

astrosticks:
Fixed the changes that @mwest1066 requested and following @echuber2's suggestion, changed out the circle icons for slashes.

PR #2392, opened 2020-04-18
WIP symbolic matrix input with example

echuber2:
This is the first version I've shared of the symbolic input matrix. It mashes up the symbolic input and matrix component input. There are some other significant changes. I added an attribute for hiding the correct answer; there's a way to specify placeholders for each cell in the question python; I locally forked copies of the sympy-handling PL libraries in order to allow floating point and symbols to be used together; I made other small changes to reinject symbols to a student's submission when it's displayed as a latex preview after a save. (In-line feedback after grading does not render as latex, but instead shows the ASCII representation along with the feedback for each cell.)

I still need to make more changes to finish re-implementing the relabs comparison at least. I'm not sure if it's worth implementing _all_ of the other modes like sigfigs that the previous floating-point inputs had supported. It is already hard to do approximate checking of equivalence with symbols, except maybe for simple polynomials, and so I mainly intend to make it easier for the question to override the grading function rather than having it checked against a correct answer verbatim.

Since this inherently supports fractions and other expressions, I don't know how this conflicts with the recently-added fraction support from @nicknytko. I know there are other WIPs in progress for things like allowing blank input fields, which if added should be reflected in some of the "?" boxes here that claim all fields must be filled. The lack of that feature was one reason why I implemented the placeholders. If an attempt is made to make the attributes for hiding or showing answers consistent across elements, that would affect this too.


echuber2:
Note to self, this probably needs to be integrated #2387

echuber2:
I reverted the commit about file line endings because it's part of another PR. #2594

nwalters512:
@echuber2 are you still interested in seeing this through?

echuber2:
Only CS 445 is using this currently, and it probably should be rethought before I PR it again. Instead of making the internal fork of the symbolic input field, the underlying symbolic input element should probably be upgraded to add the needed features.

PR #2394, opened 2020-04-18
Add parallel template rendering

mwest1066:
This is a rebased version of #1143, also upgraded to the new version of `threads`.

It allows the use of multi-process EJS template rendering.

Still needs:
[ ] Should we terminate the pool at some point? See `middlewares/renderAsync.js`. This doesn't matter for the live server, but it probably does for tests.


nwalters512:
Talked with @mwest1066 and if we want to do this in the future, we'll need to start from scratch anyways. Closing this out.

PR #2404, opened 2020-04-19
Fix bullets in docs/installingLocal.md

tdy:
Seems like the numbered bullets don't increment as desired when there's text in between. Not sure if there's a way to fix the numbering, but I just changed to dots for now.

https://prairielearn.readthedocs.io/en/latest/installingLocal/#connecting-to-an-existing-docker-container

### DIFF
--- a/docs/installingLocal.md
+++ b/docs/installingLocal.md
@@ -88,7 +88,7 @@ cd /PrairieLearn
 
 The previous shells were launched in their own containers. If you want to open a shell in a Docker container that is *already running*, you can find the container's name and connect to it.
 
-1. Find the name of your running PrairieLearn container by running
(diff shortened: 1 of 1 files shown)


mepstein-gh:
This document:
https://gist.github.com/dupuy/1855764#bulleted-and-enumerated-lists
(third paragraph in that section)
indicates that you need to indent the text of the li by at least four spaces.  I tried this on my local machine, using discount, and it seemed to work.  Hard to tell what the result will look like on github or readthedocs though.
On a similar note, just a little bit earlier in this document, there's a bit of stuff (around the line `and when any modifications are made,`) that looks like it belongs under the second li there, but that's not clear from the way it's rendered.  If indenting is used to fix the numbered bullets issue, maybe it should be used here as well.

PR #2405, opened 2020-04-19
Change time_limit_min to account for end_date

### LINKED ISSUE #2217: Remaining time does not consider endDate
When using a timed exam, at the moment the remaining time does not take the allowAccess endDate into consideration. For example, if a student is taking an exam with `timeLimitMin = 50` in a 1-hour window, if they start the exam 30 minutes late then they will in fact only have 30 minutes to work on it, but the countdown timer will show them the full 50 minutes.

When we compute the remaining time to send to the client, we should limit it to be no longer than the time until the endDate (if any) of the current access rule.

tdy:
* [x] Include `end_date` when calculating `time_limit_min`
* [x] Update docs/remoteExams.md
* [x] Update docs/accessControl.md
* [x] Fix time-limit/end-date race condition
* [x] Comment on race condition
* [x] Make news figure/skeleton
* [x] ~~Change to `time_limit_sec`~~ (more pervasive than anticipated)

Resolves #2217


sprocs/check_assessment_access.sql
```diff
@@ -38,7 +38,7 @@ BEGIN
                     END)
             ELSE 'None'
         END AS credit_date_string,
-        aar.time_limit_min,
+        LEAST(aar.time_limit_min, extract(epoch FROM aar.end_date - now()) / 60)::integer AS time_limit_min,
```
tdy, line 41: Existing sprocs aren't totally consistent with capitalization -- not sure of preferred caps for `extract()`, `epoch`, `now()`

tdy:
@mwest1066 I didn't really understand this race condition. Is it still an issue with my current patch?

From [#2217 / issuecomment-613473094](https://github.com/PrairieLearn/PrairieLearn/issues/2217#issuecomment-613473094):
> One more thing to consider is that there will be a race condition for exams with a time limit that expires exactly at the `end_date`. The "finish exam" event will auto-finish the exam at the time limit, but if this is just after the `end_date` then we will block access and the exam will not actually be closed and the student won't see their final score. One way to handle this (without fixing the underlying race condition) would be to set the time limit to 30 seconds less than the `end_date`, or something like that.

docs/remoteExams.md
```diff
@@ -51,7 +51,7 @@ This configuration is good when:
 
 Some notes about this configuration:
 
-* The exam window (70 minutes, `startDate` to `endDate`) has been set to be 10 minutes longer than the exam time limit (60 minutes). If a student starts the exam late, then the countdown timer on their exam will show the full exam time limit (60 minutes). However, they will not be able to access the exam past the `endDate` time under any circumstances. See [PL issue #2217](https://github.com/PrairieLearn/PrairieLearn/issues/2217) for details.
+* The exam window (70 minutes, `startDate` to `endDate`) has been set to be 10 minutes longer than the exam time limit (60 minutes). However, they will not be able to access the exam past the `endDate` time under any circumstances. If a student starts this exam more than 10 minutes late, then the countdown timer on their exam will reflect the time remaining until `endDate` (previously, the timer would incorrectly show the full exam time limit regardless of `endDate`).
```
mwest1066, line 54:
```suggestion
* The exam window (70 minutes, `startDate` to `endDate`) has been set to be 10 minutes longer than the exam time limit (60 minutes). However, students will not be able to access the exam past the `endDate` time under any circumstances. If a student starts this exam more than 10 minutes late, then the countdown timer on their exam will reflect the time remaining until `endDate`.
```

I think it's best to just describe how things currently work. Talking about how things used to be is helpful in the short run, but it quickly gets confusing over time.

sprocs/check_assessment_access.sql
mwest1066: Our general rule is "use the same formatting rules as the PostgreSQL docs" (I don't think we document this anywhere?). Unfortunately those docs aren't always consistent themselves 😄 

For example: https://www.postgresql.org/docs/12/functions-datetime.html

On that page we see both `interval` and `INTERVAL`. So long as we are not more inconsistent than the postgres docs, I think we are doing ok!

docs/remoteExams.md
mwest1066, line 54: Now that we've fixed this endDate problem, I actually think we should shorten the "extra time" from 10 minutes to 5 minutes.

sprocs/check_assessment_access.sql
mwest1066, line 41: We only want to set a time limit (that is, for `time_limit_min` to be non-NULL) if the access rule actually has a time limit in it. Because `LEAST(NULL, 3)` returns `3`, I think the code above will set a time limit equal to the time-until-endDate even when no explicit time limit was set.

docs/remoteExams.md
mwest1066, line 54: I guess we should also document the `endDate` behavior of time limits at https://prairielearn.readthedocs.io/en/latest/accessControl/#time-limits

sprocs/check_assessment_access.sql
```diff
@@ -38,7 +38,9 @@ BEGIN
                     END)
             ELSE 'None'
         END AS credit_date_string,
-        aar.time_limit_min,
+        CASE WHEN aar.time_limit_min IS NULL THEN NULL
+             ELSE LEAST(aar.time_limit_min, EXTRACT(EPOCH FROM aar.end_date - now()) / 60)::integer
+        END AS time_limit_min,
```
tdy, line 43: Indent style taken from [postgresql conditionals docs](https://www.postgresql.org/docs/7.4/functions-conditional.html) (though I didn't find any PL `CASE` blocks formatted this way)

sprocs/check_assessment_access.sql
tdy: Tested it and you're right. Wrapped it in a `CASE` block. Not sure if there's a better way.

tdy:
I think I understand the race condition now. You're saying that if the "finish exam" event fires at any point after `end_date`, then PL thinks the user is submitting past due. I've gotten "access denied" a couple times, which I guess was caused by this.

mwest1066:
> I think I understand the race condition now. You're saying that if the "finish exam" event fires at any point after `end_date`, then PL thinks the user is submitting past due. I've gotten "access denied" a couple times, which I guess was caused by this.

Exactly!

I'm not really sure the best way to fix this "properly", hence the suggested hack of cutting the time limit down by a little bit 😄

docs/accessControl.md
```diff
@@ -98,7 +98,7 @@ For students taking remote exams it can be helpful to have an enforced time limi
 ]
 ```
 
-The above example will give students 50 minutes for this exam, and they must start (and complete) within the date limits.
+The above example will give students 50 minutes for this exam, and they must start (and complete) within the date limits. If a student starts the exam close enough to `endDate` such that the remaining time to `endDate` is less than `timeLimitMin`, then the exam's countdown timer will be initialized with the remaining time to `endDate`, not `timeLimitMin`.
```
mwest1066, line 101: Nice! This is very clearly explained.

sprocs/check_assessment_access.sql
```diff
@@ -38,7 +38,9 @@ BEGIN
                     END)
             ELSE 'None'
         END AS credit_date_string,
-        aar.time_limit_min,
+        CASE WHEN aar.time_limit_min IS NULL THEN NULL
+             ELSE LEAST(aar.time_limit_min, EXTRACT(EPOCH FROM aar.end_date - now() - INTERVAL '31 seconds') / 60)::integer
```
mwest1066, line 42: Maybe we should add a comment here to explain why we are subtracting 31 seconds? (also, why 31 and not 30? looks suspicious! 😄)

mwest1066:
I think this change would benefit from a news item, because it's pretty crucial to how faculty are communicating with their students. @tdy could you put together a skeleton news item in this PR? I'll edit the text, but if you could give the details and maybe draw a picture then that would be a great start. I'm thinking about maybe including a diagram something like this:

time Limit mini 90

sprocs/check_assessment_access.sql
tdy: Suspicious indeed! 😆 Thing is -- `time_limit_min` only has minutes precision, so a 30s shift gives us 0 net change.

e.g.:
```
time_limit_min = 60m
now = 17:00
end = 18:00
```
If offset 30s, then no change: `end - now - 30s = 60m` (psql interval rounding)
If offset 31s, then change: `end - now - 31s = 59m`

So in practice, the buffer is 1 minute. Does that make sense or have I missed something?

mwest1066: Ahhh!!! Excellent catch!!! I totally overlooked that.

Maybe change to 60 seconds in that case?

tdy: > Maybe change to 60 seconds in that case?

So even though we subtract in 1-minute increments, I believe the exact offset still changes how quickly we steal that minute (higher offset = faster subtraction).

60s offset -> timer drops to 58m at 17:00:30
31s offset -> timer drops to 58m at 17:00:59

docs/accessControl.md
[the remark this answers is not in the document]
echuber2: "roughly" 30-second buffer then? Are you worried students would read the docs and try to abuse it? Seems minor.

tdy:
Might be another issue -- current PR only fixes race conditions _after_ `end_date - time_limit_min`, not before (e.g., `limit=60, launch=16:57, end=18:00`).

sprocs/check_assessment_access.sql
mwest1066: Makes sense!

I'm totally fine to go with the 31-second solution here, but an alternative would be to have `check_assessment_access.sql` return `time_limit_sec` (`min` -> `sec`) so we can precisely encode `end_date - interval '60 seconds'`. This would require changing `studentAssessmentExam.ejs` to display seconds, and the assessment-instance creation code (`assessment.makeAssessmentInstance` and `assessment_instances_insert.sql`) to accept seconds rather than minutes. These are all small and easy changes, but I'm also fine to just stay with the 31-second solution.

docs/accessControl.md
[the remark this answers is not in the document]
mwest1066: I'm inclined to say `minus a 1-minute buffer`. I think it's better to be conservative, so instructors aren't surprised with their student gets a slightly different time to what they might have expected.

mwest1066:
> Might be another issue -- current PR only fixes race conditions _after_ `end_date - time_limit_min`, not before (e.g., `limit=60, launch=16:57, end=18:00`).

I'm not sure I'm understanding -- what's the race condition here?

tdy:
> 
> 
> > Might be another issue -- current PR only fixes race conditions _after_ `end_date - time_limit_min`, not before (e.g., `limit=60, launch=16:57, end=18:00`).
> 
> I'm not sure I'm understanding -- what's the race condition here?

Oops I typo'ed (left out some zeros). I meant something like `limit=60, launch=16:00:59, end=18:00:00`, which gives `limit=60, remaining=60` -- isn't that still racey, or am I getting my wires crossed somewhere?

mwest1066:
> Oops I typo'ed (left out some zeros). I meant something like `limit=60, launch=16:00:59, end=18:00:00`, which gives `limit=60, remaining=60` -- isn't that still racey, or am I getting my wires crossed somewhere?

I think that case is fine. We'll get `((end_date - launch - 31 seconds)/60)::integer`, which is either 58 or 59 depending on which way rounding goes. Either is ok. I can't see how we can get 60 in this case?

tdy:
For `end = 18:00:00`, here's what I'm getting in `psql`:

```sql
SELECT (EXTRACT(EPOCH FROM TIMESTAMP '2020-01-01 18:00:00.00' - TIMESTAMP '2020-01-01 16:59:59.00' - INTERVAL '31 seconds') / 6
0)::integer;
```

| `now` | `offset = 31s` | `offset = 60s` |
| ------ | ----- | ----- |
| 16:58:58 | 61m | 60m |
| 16:58:59 | ***60m*** | 60m |
| ... | ... | ... |
| 16:59:30 | 60m | 60m |
| 16:59:31 | 60m | ***59m*** |
| ... | ... | ... |
| 16:59:59 | 60m | 59m |
| 17:00:00 | ***59m*** | 59m |
| ... | ... | ... |
| 17:00:29 | 59m | 59m |
| 17:00:30 | 59m | ***58m*** |
| ... | ... | ... |
| 17:00:58 | 59m | 58m |
| 17:00:59 | ***58m*** | 58m |

tdy:
Figure preview. If preferred, I can re-plot without the extra stuff (to match your draft drawing):

figure_timer_300

mwest1066:
> Figure preview. If preferred, I can re-plot without the extra stuff (to match your draft drawing):

Wow! This looks amazing! I wouldn't change anything.

tdy:
@mwest1066 ping, in case you're still wanting to deploy this one before finals

mwest1066:
> @mwest1066 ping, in case you're still wanting to deploy this one before finals

I've actually reached my point of extreme conservatism, where I don't want to make any changes at all to the code until after exams are over. Sorry to do this, but I think I'm going to merge/deploy this as soon as finals end.

news_items/011_remaining_time_end_date/index.html
```diff
@@ -1,39 +1,16 @@
...
-  </li>
-</ol>
-
-<p>
-  See the <span aria-hidden="true">documentation on</span>
-  <a href="https://prairielearn.readthedocs.io/en/latest/accessControl/#time-limits"><span class="sr-only">documentation on</span> time limits</a>
-  and
-  <a href="https://prairielearn.readthedocs.io/en/latest/remoteExams/"><span class="sr-only">documentation on</span> timed exams</a>
-  for details.
+  <a href="https://github.com/tty">Tim Yang</a> has fixed the timer countdown for students who start exams late. Previously, even if a student started an exam so late that they would run into the <tt>endDate</tt>, PrairieLearn would still show the full <tt>timeLimitMin</tt> countdown timer for them, but would then cut them off at <tt>endDate</tt>. With Tim's fix, the countdown timer correctly only shows the time remaining until <tt>endDate</tt>.
```
tdy, line 2:
```suggestion
  <a href="https://github.com/tdy">Tim Yang</a> has fixed the timer countdown for students who start exams late. Previously, even if a student started an exam so late that they would run into the <tt>endDate</tt>, PrairieLearn would still show the full <tt>timeLimitMin</tt> countdown timer for them, but would then cut them off at <tt>endDate</tt>. With Tim's fix, the countdown timer correctly only shows the time remaining until <tt>endDate</tt>.
```

mwest1066: Good catch!

Issue #2267, opened 2020-03-29
Store server.py call times in DB objects


mwest1066:
For debugging and robustness it would be very helpful to track how long function calls are taking. We could store:

* `generate()` time in `variants`
* `render()` time in `view_logs`
* `parse()` time in `submissions`
* `grade()` time in `grading_logs`

etc.

eliotwrobson:
@nwalters512 I think I've seen you report times like these in some newer PRs. Would this be a worthwhile change?

mwest1066:
Agreed on this being low priority, but I think it would be a good thing to add as part of adding better debugging support in general for instructors (@reteps was thinking about this, I believe).

PR #2407, opened 2020-04-20
Change question timeout

### LINKED ISSUE #1822: Question timeout limits
Probably should place the timeout for creating a question in `config.js`

I think the code governing it is:

https://github.com/PrairieLearn/PrairieLearn/blob/42dde84826d6a8263f36d0be5b28c0d1bef6ffa5/lib/code-caller.js#L135-L139

tdy:
Fixes #1822 (I think?)

* [x] Refactor timeout to `config.js`
* [x] Lower timeout from 20000 (now ~~5000~~ 10000)

> Probably should place the timeout for creating a question in `config.js`
>
> I think the code governing it is:
>
> https://github.com/PrairieLearn/PrairieLearn/blob/42dde84826d6a8263f36d0be5b28c0d1bef6ffa5/lib/code-caller.js#L135-L139
>
> It would also be good to reduce this value for safety.


coatless:
@tdy let's avoid a timeout that low. I think the `rpy2`-based questions -- e.g. _R_ generated data -- will break with only `5000` ms (~5s) to generate. Could we try ~10s first? Or @mwest1066 can you pull question generation timing?

lib/config.js
```diff
@@ -48,6 +48,7 @@ config.cronIntervalCalculateAssessmentQuestionStatsSec = 10 * 60;
 config.cronDailySec = 8 * 60 * 60;
 config.autoFinishAgeMins = 6 * 60;
 config.questionDefaultsDir = 'question-servers/default-calculation';
+config.questionTimeout = 5000;
```
tdy, line 51:
```suggestion
config.questionTimeout = 10000;
```

tdy:
Bumped to 10000 for now, unless Matt finds a better mark from the data.

mwest1066:
I think we aren't actually tracking this at the moment: https://github.com/PrairieLearn/PrairieLearn/issues/2267

lib/config.js
```diff
@@ -48,6 +48,7 @@ config.cronIntervalCalculateAssessmentQuestionStatsSec = 10 * 60;
 config.cronDailySec = 8 * 60 * 60;
 config.autoFinishAgeMins = 6 * 60;
 config.questionDefaultsDir = 'question-servers/default-calculation';
+config.questionTimeout = 10000;
```
mwest1066, line 51:
```suggestion
config.questionTimeoutMilliseconds = 10000;
```

We typically put the unit on time-related config variables, because they are variously in milliseconds, seconds, minutes, etc, and it's really annoying to have to go and read the code to figure out what the units are 😄 

We were using "MS" for "Milliseconds" on other config variables, but I actually think we should just write it out for clarity.

tdy:
> 
> 
> I think we aren't actually tracking this at the moment: #2267

So maybe I'll add back the FIXME/TODO comment to tweak this after seeing data from #2267

lib/code-caller.js
```diff
@@ -135,7 +135,7 @@ class PythonCaller {
         const localOptions = _.defaults(options, {
             cwd: __dirname,
             paths: [],
-            timeout: 20000, // FIXME: this number (equivalent to 20 seconds) should not have to be this high
+            timeout: config.questionTimeout,
```
tdy, line 138:
```suggestion
            timeout: config.questionTimeoutMilliseconds,
```

lib/config.js
```diff
@@ -48,6 +48,7 @@ config.cronIntervalCalculateAssessmentQuestionStatsSec = 10 * 60;
 config.cronDailySec = 8 * 60 * 60;
 config.autoFinishAgeMins = 6 * 60;
 config.questionDefaultsDir = 'question-servers/default-calculation';
+config.questionTimeoutMilliseconds = 10000;
```
tdy, line 51:
```suggestion
config.questionTimeoutMilliseconds = 10000; // TODO: tweak this value once we see the data from #2267
```

PR #2419, opened 2020-04-22
WIP groups schema 

trombonekenny:
The goal is to make a common database schema for group storage to accommodate different projects developing group-centric features. PR is here for review and discussion. Once merged, the projects doing group-based work can build off of it.

Base schema from https://github.com/frankyhao/PrairieLearn/blob/master/migrations/171_group_change_assessment_inst.sql but added a few different columns to add date logging and deletion, course_instance affiliation.

A third table governing group creation or membership rules is there but commented out because it seems more about controls of groups than their storage.


migrations/800_groups__add.sql
```diff
@@ -0,0 +1,40 @@
...
+-- Suggestion: maybe call group_criteria or group_config or group_contraints?
+-- This group defines additional criteria or 'ownership' of the group, perhaps
+-- if the group was autogenerated based on rules
+--
+-- Haven't implemented this yet because it seems more about group creation or constraints
+-- on autogenerated groups. Because one group_config could be connected to multiple
+-- groups this is pulled out to a different table
+--CREATE TABLE IF NOT EXISTS group_config (
+--    id BIGSERIAL PRIMARY KEY,
+--    course_instance_id BIGINT NOT NULL REFERENCES course_instances(id) ON DELETE CASCADE ON UPDATE CASCADE,
```
tdy, line 13: Weren't we planning to reference `assessments`?

tdy: Did we end up deciding to make it course-wide and override those settings in each layer?

trombonekenny: There will probably need to be some linking between `assessments` and groups, probably in the `group_config` (or group_constraints, whatever we're going to call it). But I hadn't flushed that part out very much. A column might not be needed if the linkage comes from an assessment_access_rule?

migrations/800_groups__add.sql
```diff
@@ -0,0 +1,40 @@
...
+
+-- Abdu students called this group_type
+-- Suggestion: maybe call group_criteria or group_config or group_contraints?
+-- This group defines additional criteria or 'ownership' of the group, perhaps
+-- if the group was autogenerated based on rules
+--
+-- Haven't implemented this yet because it seems more about group creation or constraints
+-- on autogenerated groups. Because one group_config could be connected to multiple
+-- groups this is pulled out to a different table
+--CREATE TABLE IF NOT EXISTS group_config (
```
mwest1066, line 11: I agree that we need a `group_configs` table.

migrations/800_groups__add.sql
mwest1066: I agree we want a `group_configs.assessment_id` column.

migrations/800_groups__add.sql
```diff
@@ -0,0 +1,29 @@
...
+    id BIGSERIAL PRIMARY KEY,
+    course_instance_id BIGINT NOT NULL REFERENCES course_instances(id) ON DELETE CASCADE ON UPDATE CASCADE,
+    name TEXT,      -- visible name of the group
+    group_config_id BIGINT REFERENCES group_configs(id) ON DELETE CASCADE ON UPDATE CASCADE,
+    date timestamp with time zone DEFAULT CURRENT_TIMESTAMP,
+    deleted_at timestamp with time zone
+);
+
+-- simple join table, no extra metadata - that could be stored in audit logs if needed
+CREATE TABLE IF NOT EXISTS group_users (
```
astrosticks, line 25: Maybe this should be called something along the lines of `group_memberships`? `group_users` kind of makes it sound like there's a special user type created for group assessments.

migrations/800_groups__add.sql
```diff
@@ -0,0 +1,29 @@
+-- A Work in Progress to discuss shared schema for group storage in PL
+
+CREATE TABLE IF NOT EXISTS group_configs (
+    id BIGSERIAL PRIMARY KEY,
+    course_instance_id BIGINT NOT NULL REFERENCES course_instances(id) ON DELETE CASCADE ON UPDATE CASCADE,
+    assessment_id BIGINT REFERENCES assessments(id) ON DELETE CASCADE ON UPDATE CASCADE,
+    type TEXT, -- what will these be? This might belong in groups rather than groups_config
+    name TEXT, -- or maybe description is accurate here? Maybe as a base for groups created from it?
+    minimum INT,
+    maximum INT,
```
astrosticks, line 10: Is there a use case where we'd want to impose these restrictions on the number of users in a group? I'm guessing `maximum` is more for students opting-in to a group, to prevent them from making a group that's too large.

migrations/800_groups__add.sql
```diff
@@ -0,0 +1,29 @@
+-- A Work in Progress to discuss shared schema for group storage in PL
+
+CREATE TABLE IF NOT EXISTS group_configs (
+    id BIGSERIAL PRIMARY KEY,
+    course_instance_id BIGINT NOT NULL REFERENCES course_instances(id) ON DELETE CASCADE ON UPDATE CASCADE,
+    assessment_id BIGINT REFERENCES assessments(id) ON DELETE CASCADE ON UPDATE CASCADE,
```
astrosticks, line 6: Why restrict a config to a particular assessment? Especially if instructors want to, say, spread groups over multiple assessments for an end-of-semester ongoing project.

trombonekenny: The group developing that feature has a way to duplicate groups used on one assessment on others, which I think addresses that point. Database-wise, the `assessment_id` isn't required so there might be place for groups used across multiple assignments?

migrations/800_groups__add.sql
trombonekenny: Yes, part of their design includes defined max and minimums for groups (regardless of if they're opted-in or auto assigned.)

migrations/800_groups__add.sql
trombonekenny: Our convention has been that join tables are named after the tables they join (users and groups).

echuber2: This was the nomenclature that Matt originally suggested. I thought people were pronouncing it "groups to users" in past meetings; I would have suggested `groups_to_users` or something. "Memberships" does sound better though. Would `group_members` be fewer letters to type?

echuber2: For some reason Dave's comment wasn't loaded at the time I commented.

migrations/800_groups__add.sql
```diff
@@ -0,0 +1,29 @@
...
+);
+
+CREATE TABLE IF NOT EXISTS groups (
+    id BIGSERIAL PRIMARY KEY,
+    course_instance_id BIGINT NOT NULL REFERENCES course_instances(id) ON DELETE CASCADE ON UPDATE CASCADE,
+    name TEXT,      -- visible name of the group
+    group_config_id BIGINT REFERENCES group_configs(id) ON DELETE CASCADE ON UPDATE CASCADE,
+    date timestamp with time zone DEFAULT CURRENT_TIMESTAMP,
+    deleted_at timestamp with time zone
+);
```
astrosticks, line 22: Should we include a UNIQUE (course_instance_id, name) here, to catch conflicts?

mwest1066: I'm in favor of adding constraints wherever possible.

trombonekenny: I added the uniqueness constraints based on `course_instance_id, name, group_config_id`. It made sense to me that if a `group_config` was used, the name should be unique inside of it but could be reused across two different configurations (i.e. Assessment 1 and Assessment 2 can both have a "Group 1"

Two partial constraints were added because postgresql doesn't see NULL as unique and I wanted to catch the case where no group_config was applied still had uniqueness based on the name and course_instance_id. We might be able to simplify this if we require a linked group_config in groups instead of making it optional.

mwest1066:
Implemented in #2518

PR #2427, opened 2020-04-28
Add docs FAQ entry for special html characters

### LINKED ISSUE #1098: <= in Question HTML Breaks Rendering

nicknytko:
Special characters like `<`, `>`, `&` can break rendering on the page especially when used in inline code segments.  This is common enough that we should probably write something on the docs -- this adds an entry in the FAQ and a small section in the elements page for `pl-code`.

Resolves #1098


echuber2:
Do you want to add a note that this doesn't apply to embedded markdown? E.g. the recommendations are to either use pl-code with manual escaping, or use it with external (raw) code files, or use the embedded markdown (without manual escaping).

echuber2:
Now that I say that, I forget whether this affects embedded markdown too or not. I know it gets parsed out and manually escaped, but maybe that only solves problems with things like ampersands, not angle brackets that look like closing tags.

nicknytko:
> Do you want to add a note that this doesn't apply to embedded markdown? E.g. the recommendations are to either use pl-code with manual escaping, or use it with external (raw) code files, or use the embedded markdown (without manual escaping).

Good point, I think markdown will correctly escape ampersands and angle brackets as well if they're in a code block.  I suspect it will also escape characters outside of a code block, but I'll test this to make sure.

docs/elements.md
```diff
@@ -773,6 +773,10 @@ Attribute | Type | Default | Description
 
 The `pl-code` element uses the _Pygments_ library for syntax highlighting, a full list of supported languages can be found [here](https://pygments.org/languages/).
 
+##### Common Pitfalls
+
+The HTML specification disallows inserting special characters onto the page (i.e. `<`, `>`, `&`), and using these characters with inline code may break rendering.  To fix this, either escape the characters (`&lt;`, `&gt;`, `&amp;`, more [here](https://www.freeformatter.com/html-entities.html)), or load code snippets from external files.
```
coatless, line 778:
```suggestion
The HTML specification disallows inserting special characters onto the page (i.e. `<`, `>`, `&`), and using these characters with inline code may break rendering.  To fix this, either escape the characters (`&lt;`, `&gt;`, `&amp;`, more [here](https://www.freeformatter.com/html-entities.html)), or load code snippets from external files into `pl-code` with `source-file-name` attribute.
```

docs/faq.md
```diff
@@ -282,3 +282,7 @@ To address this, there are a variety of different ways. In particular, we have:
 - Restart your computer.
 - Stop the process in terminal with <kbd>CNTRL</kbd> + <kbd>C</kbd> and, then, 
   close the terminal application.   
+
+## Why do special characters like (<=) break my question display?
+
+The HTML specification disallows inserting special characters onto the page (i.e. `<`, `>`, `&`), and using these characters in your question, for example with inline code, may break rendering.  To fix this, either escape the characters (`&lt;`, `&gt;`, `&amp;`, more [here](https://www.freeformatter.com/html-entities.html)), or load code snippets from external files.
```
coatless, line 288:
```suggestion
The HTML specification disallows inserting special characters onto the page (i.e. `<`, `>`, `&`), and using these characters in your question, for example with inline code, may break rendering.  To fix this, either escape the characters (`&lt;`, `&gt;`, `&amp;`, more [here](https://www.freeformatter.com/html-entities.html)), or load code snippets from external files into `pl-code` with `source-file-name` attribute. For more information, see the [`pl-code` element documentation](elements.md#pl-code-element)
```

PR #2429, opened 2020-04-28
Config toggles for localstack usage

trombonekenny:
Adds the ability to specify AWS endpoints to use PrairieLearn under different frameworks, like https://github.com/localstack/localstack

Adds `config.runningInEc2` to mimic PrairieGrader for some load reporting toggling (that localstack doesn't support).

### DIFF
--- a/schemas/schemas/serverConfig.json
+++ b/schemas/schemas/serverConfig.json
@@ -96,6 +96,10 @@
             "description": "Time after last activity when exams are auto-closed (minutes).",
             "type": "integer"
         },
+        "runningInEc2": {
+            "description": "Toggle if PL is running as an AWS EC2 instance",
+            "type": "boolean"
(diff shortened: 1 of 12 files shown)


mwest1066:
Would you mind submitting a corresponding PR to ansible-pl? (assuming it needs changes)

schemas/schemas/serverConfig.json
```diff
@@ -96,6 +96,10 @@
             "description": "Time after last activity when exams are auto-closed (minutes).",
             "type": "integer"
         },
+        "runningInEc2": {
+            "description": "Toggle if PL is running as an AWS EC2 instance",
```
mwest1066, line 100:
```suggestion
            "description": "Whether PL is running as an AWS EC2 instance",
```

PR #2431, opened 2020-04-29
[WIP] RFC: multi-server course content architecture

nwalters512:


rfcs/0000-multi-server-course-content.md
```diff
@@ -0,0 +1,58 @@
...
+
+When a course instance is synced, we update its git hash in the database. This hash will serve as a unique version identifier for the course's contents. The next time someone tries to access something from the course, we'll check if we already have a working directory for that particular version. If not, we can create one that will store only files for the particular revision, like `/courses/my_course_id/234fae...[hash]...9823efde/`. We'll then start loading files again to this new directory so as not to disrupt any question renders that might be in progress. We can then have a background task that periodically deletes old versions.
```
mwest1066, line 50: One thing I don't quite understand is the distinction between the course-level hash, which uniquely defines the whole course, and chunk-level hashes, which uniquely define the chunk. Most of the time, updating a course (and getting a new hash for it) will not change most of the chunk-level hashes, so it would be a shame to have to re-download them all.

I'm assuming that we store a DB table that records which version (hash) of each chunk we have on each machine, as well as the hash of the latest version of each chunk that's available in S3? This way each machine can run a fast query to make sure that it has up-to-date versions of all needed chunks before it starts processing the query.

When making local copies of updated chunks, it seems to me we could either:
1. Unpack them in a temporary directory and then atomically move them into place in a fixed course directory (without the hash), like `/courses/my_course_id/`.
2. Use course directories with the course-hash in the name (`/courses/my_course_id/234fae.../`) but when we make a new one of these, we hard-link all existing content from the last version so we don't have to re-download existing chunks.

While option (2) is more robust, I'm actually not sure it's needed, especially in a first version. It would be easy to retrofit later. We've been running with in-place updates from sync (option 1) for many years without any noticeable problems.

rfcs/0000-multi-server-course-content.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Local storage exhaustion
+
+Over time, the state of the world will trend towards all course files for all courses existing on disk. Ideally, autoscaled instances are running on smaller, cheaper machines with limited resources. This means that over time, we're likely to run out of disk space. To prevent this (or if this someday becomes a problem), we can build an LRU cache on top of this system that will delete old content that hasn't been used in a while.
+
+# Potential future optimizations
+
+* The hash of a chunk's content could be stored and incorporated into its ID; this could be used to optimize syncs to only upload changes chunks. And for servers, they would only need to load a chunk again if its contents have changed.
```
mwest1066, line 58: Ah, ok, now I understand my earlier confusion. I was thinking that we'd do this optimization from the start. Otherwise any tiny change to any part of a course would require all servers to re-download all questions for that course. I've seen instructors updating access rules as a remote exam is starting (typical case is to add an extra-time accommodation for a student at the last minute), and this happens exactly when the servers are all under high load.

I think we need two DB tables:

Table `chunks` to store which chunks are currently available in S3, with columns:
* `course_id`
* `type` - can be `serverFilesCourse`, `elements`, `question`, etc.
* `question_id` - this is NULL unless `type = 'question'`
* `course_instance_id` - this is NULL unless `type = 'clientFilesCourseInstance'`
* `hash` - the currently active hash for this chunk

Table `instance_chunks` to store which chunk versions are on each machine, with columns: 
* `instance_id` - unique identifier for each machine (we already have this as `config.instanceId`)
* `chunk_id`
* `hash` - this is the hash that we have on local disk for this chunk on this instance

When we want to serve a question route, we'd figure out whether we need new chunks with something like

```sql
SELECT
    chunks.*
FROM
    chunks
    LEFT JOIN instance_chunks ON (chunks.id = instance_chunks.chunk_id)
WHERE
    course_id = $course_id
    AND instance_id = $instance_id
    AND (
        type = 'serverFilesCourse'
        OR type = 'elements'
        OR (type = 'question' and question_id = $question_id)
    )
    AND instance_chunks.hash != chunks.hash;
```

We could do this as part of `middlewares/selectAndAuthzInstanceQuestion.sql`, for example, (just like the `file_list` CTE in that file) so that it will add negligible overhead and no each DB round-trips. We can store the list of returned chunks in `res.locals.chunks_needed` and have a middleware that checks this variable and blocks until we've got any needed chunks downloaded.

rfcs/0000-multi-server-course-content.md
```diff
@@ -0,0 +1,58 @@
...
+When course content is synced, it will be broken up into a number of chunks:
+
+* One chunk containing all course custom elements
+* One chunk for `clientFilesCourse`
+* One chunk for `serverFilesCourse`
+* A chunk for each course instance's `clientFilesCourseInstance`
+* A chunk for each course instance's `serverFilesCourseInstance`
+* One chunk for each question
+* A chunk for all question thumbnails
+
```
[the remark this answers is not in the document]
nwalters512: I did some audits of these two:

There is only one file that exists in a `serverFilesCourseInstance` directory:
* `/pl-cs126/courseInstances/Sp17/serverFilesCourseInstance/sp17secret.txt`

There are no files in a `serverFilesAssessment` directory.

Given that, I propose dropping these from the documentation and not handling them in any new syncing code.

rfcs/0000-multi-server-course-content.md
```diff
@@ -0,0 +1,58 @@
...
+These chunks will be uploaded to some file store with a unique, possibly deterministic ID.
+
+When a server needs to use some course content, it will download the necessary chunks to disk. Here are some common pages and the chunks necessary for them:
+
+* Viewing the course question page
+  * Question thumbnails chunk
```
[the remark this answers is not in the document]
nwalters512: Excellent, that should make things a bit simpler!

rfcs/0000-multi-server-course-content.md
```diff
@@ -0,0 +1,58 @@
...
+* Viewing or grading a question
+  * Custom elements chunk
+  * Question chunk
+  * `serverFilesCourse` chunk
+  * Appropriate `serverFilesCourseInstance` chunk
```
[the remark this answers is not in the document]
nwalters512: This actually makes a lot of sense - a question shouldn't be tied to a particular course instance. Given the above comment where I recommended not supporting `serverFilesCourseInstance`, we can remove this.

rfcs/0000-multi-server-course-content.md
nwalters512: Sorry, I was definitely a bit hand-wavey with how chunks would be stored and identified! Agreed that it makes sense to do this optimization from the start; I can add that explicitly to this RFC.

Storing the chunks used by a particular instance in the DB seems pretty heavyweight, especially considering that we're intending to trend towards many smaller, more ephemeral instances. I know that the DB is the one bit of our stack that will handle anything we can throw at it, but I'd rather not bloat it any more than we have to 🙂. I think just having each instance keep an in-memory cache of which files are on its particular disk would be sufficient. If we can't find it in the in-memory cache, we can fall back to a slightly-slower `stat` to see if it's on disk. If it is, we update the in-memory cache; if it isn't, we go to S3 for the chunk.

This should keep the average case (chunk is on disk already) as fast as possible (in-memory cache is orders of magnitude faster than hitting the disk), while also keeping less common cases (server crashes and restarts; files are still on disk, and a `stat` is still faster than a fresh download from S3.

rfcs/0000-multi-server-course-content.md
nwalters512: I agree that in-place updates should be sufficient, especially since we've been doing it that way since the start. Long-term, I think we could actually do truly atomic updates by introducing some new APIs. Currently, questions must use hardcoded paths like `../../serverFilesCourse/foo/bar.txt` to access assets. If we instead provided an API to questions like `data['options']['server_files_course_path']` to complement existing things like `data['options']['client_files_question_url']`, we could unpack new server files to `/courses/my_course_id/serverFilesCourse-1234fdea...[hash]...4321` instead of just `/courses/my_course_id/serverFilesCourse`. This way, we could keep the files seen by a particular question render/grade/etc. wholly consistent! This would be very backwards incompatible though, so this is definitely a future optimization we could potentially make.

rfcs/0000-multi-server-course-content.md
mwest1066: I agree that in-memory (and on-disk) storage of local chunk state is fine. Although it does make it harder for a central process to generate statistics 😄

rfcs/0000-multi-server-course-content.md
mwest1066: I was thinking about this further and I’m actually a bit tempted to “do it right” from the start with respect to atomic update, if it isn’t too complex. You mentioned earlier about using symlinks to activate new components. This seems relatively easy and robust?

rfcs/0000-multi-server-course-content.md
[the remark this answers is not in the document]
mwest1066: Perfect! I completely agree.

rfcs/0000-multi-server-course-content.md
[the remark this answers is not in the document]
mwest1066: Excellent!

PR #2434, opened 2020-04-30
npm audit fix

trombonekenny:
I'm not sure why dependabot didn't catch and process this, but we've had a GitHub security warning for `minimist` for almost a month. This branch tries to resolve this.

It's a checkout and `npm audit fix` which updates handlebars to get rid of the warned dependency.


mwest1066:
Superseded by #2477

PR #2472, opened 2020-05-01
Vertically align inputs in checkbox and multiple choice elements 

### LINKED ISSUE #1215: Fix alignment on pl-checkbox and friends
When multiple choice questions have very tall options, the placement of the checkboxes looks potentially confusing. We should consider adding custom styles to make this use case look better.

screen shot 2018-08-19 at 7 02 00 pm

nicknytko:
Fixes #1215 by sticking labels in a vertically aligned flexbox div/span

Also cleans up some of the code in the checkbox and multiple choice elements by moving markup out of the python files into their respective mustache templates

Before:
image
image

After:
image
image

Small tweak, but labels are now horizontally aligned such that their left border will line up.

Checkbox/mc items now have more padding between them:

image

### DIFF
--- a/elements/pl-checkbox/pl-checkbox.css
+++ b/elements/pl-checkbox/pl-checkbox.css
@@ -2,3 +2,7 @@
     width: 340px;
     max-width: 100%;
 }
+
+:not(.form-check-inline) > .pl-checkbox-key-label {
+    min-width: 20px;
+}
--- a/elements/pl-checkbox/pl-checkbox.css
+++ b/elements/pl-checkbox/pl-checkbox.css
@@ -4,5 +4,5 @@
 }
 
 :not(.form-check-inline) > .pl-checkbox-key-label {
-    min-width: 20px;
+    min-width: 1.25em;
 }
--- a/elements/pl-checkbox/pl-checkbox.mustache
+++ b/elements/pl-checkbox/pl-checkbox.mustache
@@ -72,7 +72,7 @@
 {{^inline}}<div class="d-block">{{/inline}}
     <span id="pl-checkbox-submission-{{uuid}}">
         <span class="badge text-danger badge-invalid"><i class="fa fa-exclamation-triangle" aria-hidden="true"></i> Invalid</span>
-        <a href="javascript:void(0);" role="button" class="btn btn-sm btn-secondary small border" data-placement="auto" data-trigger="focus" data-toggle="popover" data-html="true" style="margin-left: 5px" title="Format Error" tabindex="0" data-content="{{parse_error}}"> Why <i class="fa fa-question-circle" aria-hidden="true"></i></a>
+        <a href="javascript:void(0);" role="button" class="btn btn-sm btn-secondary small border ml-1" data-placement="auto" data-trigger="focus" data-toggle="popover" data-html="true" title="Format Error" tabindex="0" data-content="{{parse_error}}"> Why <i class="fa fa-question-circle" aria-hidden="true"></i></a>
     </span>
 </div>
 {{#inline}}</span>{{/inline}}
--- a/elements/pl-checkbox/pl-checkbox.mustache
+++ b/elements/pl-checkbox/pl-checkbox.mustache
@@ -13,7 +13,7 @@
 {{^inline}}<div class="d-block">{{/inline}}
 
 {{#answers}}
-    <div class="form-check {{#inline}}form-check-inline d-inline-flex{{/inline}}{{^inline}}d-flex{{/inline}} align-items-center">
+    <div class="form-check {{#inline}}form-check-inline d-inline-flex{{/inline}}{{^inline}}d-flex{{/inline}} align-items-center py-1">
         <input class="form-check-input" type="checkbox"
                name="{{name}}" value="{{key}}" {{^editable}}disabled{{/editable}}
                {{#checked}}checked{{/checked}} id="{{name}}-{{key}}">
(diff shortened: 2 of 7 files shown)


elements/pl-multiple-choice/pl-multiple-choice.mustache
```diff
@@ -0,0 +1,82 @@
+{{#question}}
+
+{{#inline}}
+<span>
+{{/inline}}
+    
+{{#answers}}
+    <div class="form-check {{#inline}}form-check-inline d-inline-flex{{/inline}}{{^inline}}d-flex{{/inline}} align-items-center">
```
tdy, line 8:
```suggestion
    <div class="form-check {{#inline}}form-check-inline d-inline-flex{{/inline}}{{^inline}}d-flex{{/inline}} align-items-center py-1">
```
How about some slight vertical padding?
image

elements/pl-checkbox/pl-checkbox.mustache
```diff
@@ -12,8 +12,28 @@
 {{#inline}}<span class="d-inline-block">{{/inline}}
 {{^inline}}<div class="d-block">{{/inline}}
 
-{{{answerset}}}
+{{#answers}}
+    <div class="form-check {{#inline}}form-check-inline d-inline-flex{{/inline}}{{^inline}}d-flex{{/inline}} align-items-center">
```
tdy, line 16:
```suggestion
    <div class="form-check {{#inline}}form-check-inline d-inline-flex{{/inline}}{{^inline}}d-flex{{/inline}} align-items-center py-1">
```

elements/pl-checkbox/pl-checkbox.mustache
```diff
@@ -47,14 +68,65 @@
         });
     });
 </script>
-{{#parse_error}}
 {{#inline}}<span class="d-inline-block">{{/inline}}
 {{^inline}}<div class="d-block">{{/inline}}
     <span id="pl-checkbox-submission-{{uuid}}">
-        <a role="button" class="btn btn-light btn-sm" data-placement="auto" data-trigger="focus" data-toggle="popover" data-html="true" title="Error" tabindex="0" data-content="{{parse_error}}">INVALID <span><i class="fa fa-question-circle" aria-hidden="true"></i></span></a>
+        <span class="badge text-danger badge-invalid"><i class="fa fa-exclamation-triangle" aria-hidden="true"></i> Invalid</span>
+        <a href="javascript:void(0);" role="button" class="btn btn-sm btn-secondary small border" data-placement="auto" data-trigger="focus" data-toggle="popover" data-html="true" style="margin-left: 5px" title="Format Error" tabindex="0" data-content="{{parse_error}}"> Why <i class="fa fa-question-circle" aria-hidden="true"></i></a>
```
tdy, line 75:
```suggestion
        <a href="javascript:void(0);" role="button" class="btn btn-sm btn-secondary small border ml-1" data-placement="auto" data-trigger="focus" data-toggle="popover" data-html="true" title="Format Error" tabindex="0" data-content="{{parse_error}}"> Why <i class="fa fa-question-circle" aria-hidden="true"></i></a>
```
Unless there was a specific rationale for manual `style`?

elements/pl-multiple-choice/pl-multiple-choice.mustache
```diff
@@ -0,0 +1,82 @@
...
+            sanitize: false,
+            container: 'body',
+            template: '<div class="popover pl-checkbox-popover" role="tooltip"><div class="arrow"></div><h3 class="popover-header"></h3><div class="popover-body"></div></div>',
+        });
+    });
+</script>
+        
+<span id="pl-multiple-choice-submission-{{uuid}}">
+    <span class="badge text-danger badge-invalid"><i class="fa fa-exclamation-triangle" aria-hidden="true"></i> Invalid</span>
+    <a href="javascript:void(0);" role="button" class="btn btn-sm btn-secondary small border" data-placement="auto" data-trigger="focus" data-toggle="popover" data-html="true" style="margin-left: 5px" title="Format Error" tabindex="0" data-content="{{parse_error}}"> Why <i class="fa fa-question-circle" aria-hidden="true"></i></a>
```
tdy, line 61:
```suggestion
    <a href="javascript:void(0);" role="button" class="btn btn-sm btn-secondary small border ml-1" data-placement="auto" data-trigger="focus" data-toggle="popover" data-html="true" title="Format Error" tabindex="0" data-content="{{parse_error}}"> Why <i class="fa fa-question-circle" aria-hidden="true"></i></a>
```

coatless:
@nicknytko Huge fan of the refactoring work that went into separating logic from display in the `*.py` scripts.

nicknytko:
@tdy how does that look on something like `templateMultipleChoiceSingleCorrect` where padding between choices isn't needed?

tdy:
> 
> 
> @tdy how does that look on something like `templateMultipleChoiceSingleCorrect` where padding between choices isn't needed?

image

nicknytko:
That's not bad, I think it could be a worthy trade-off to have more legible large answer choices. @coatless @mwest1066 any thoughts here?

nicknytko:
Or, we could possibly do something clientside where we dynamically set padding based on the height

coatless:
Not sure I'd go down the route of dynamic item spans. I think even now having a little space between options is good. Previously, everything was highly linked.

elements/pl-checkbox/pl-checkbox.css
```diff
@@ -2,3 +2,7 @@
     width: 340px;
     max-width: 100%;
 }
+
+:not(.form-check-inline) > .pl-checkbox-key-label {
+    min-width: 20px;
```
mwest1066, line 7: Not that it will really change anything, but would it be better to use `em` as the units here, given that this is sizing for text?

nicknytko: Good point!

mwest1066:
It looks to me like the radio buttons are actually not quite aligned correctly with the text? The text seems to be set higher than the radio buttons?

image

PR #2479, opened 2020-05-02
CS497 - Header Lock Implementation

shroudie:


trombonekenny:
Hello and thanks for the PR! It doesn't look like it's building properly - I think that perhaps the version you based your `instructorGradebook.ejs` off has changed in PrairieLearn from when you did your original fork, and so those changes need to be incorporated into your code. Can you work on fixing this or do you need assistance?

shroudie:
I just tried to merge the new commits, and checked that the updates are in this branch, was able to build locally. 
Would you mind try again?

trombonekenny:
Good, that seemed to build and run. But it also removed the UIN column that was added in a recent version of PrairieLearn. Your changes seemed to remove it.

The interface isn't very intuitive. On Chrome on Windows, I saw this:

image

Which doesn't give any scroll indicators that there's horizontal scrolling possible. I need to click in the grades grid and use the scroll arrows to move the grid around. That's not intuitive. I'm also not sure why it put horizontal scrollbars on the UID columns. Is it possible to clean that up?

shroudie:
Yes, give us a few days to make the changes

trombonekenny:
FYI, the PL master branch just had a feature added that might be useful in your development. There's a `generate_and_enroll_users` options under the Queries tab (as an admin off the main page) that lets you create random users and enroll them in your course. That might help if you need to easily recreate 300 students or something for the gradebook view.

shroudie:
Hi, `generate_and_enroll_users` was really helpful. There is some difficulty with having variable length of table columns, but we tried our best to do the clean-up, in terms of adding UIN, and hiding the scrolling on UID.

mwest1066:
I wonder whether we'd be better off using something like http://www.tabulator.info ?

shroudie:
I think I better close this PR since the semester is over... 
If there is anything else need, you could message me. 
Thanks.

PR #2488, opened 2020-05-04
Fix MathJax rendering bug

nicknytko:
MathJax 3.0.5 breaks on SVG output when representing double/triple/quad primes with `''`, `'''`, and `''''`.  Here is a fix by one of the MathJax developers to get around this.


mwest1066:
See https://github.com/mathjax/MathJax/issues/2402 for the upstream bug report and workaround.

PR #2490, opened 2020-05-05
Add generate_and_enroll_users sproc and query

trombonekenny:
Gives local developers a quick way to add a lot of student roles to a course instance.


admin_queries/generate_and_enroll_users.sql
```diff
@@ -0,0 +1 @@
+CALL users_randomly_generate($count::int, $course_instance_id);
```
mwest1066, line 1:
```suggestion
SELECT user_id, uid, uin, name, email
FROM users_randomly_generate($count, $course_instance_id);
```
Can we make this return information about the generated users like the above? (I know this will require changing the procedure to a function that returns the newly-inserted user rows.)

Also, I think we don't need the `::int` typecast?

sprocs/users_randomly_generate.sql
```diff
@@ -0,0 +1,60 @@
...
+
+    first_names text[];
+
+    status int;
+    name text;
+    email text;
+    uin text;
```
mwest1066, line 19: We don't seem to actually set the `uin`?

sprocs/users_randomly_generate.sql
```diff
@@ -0,0 +1,60 @@
...
+
+    _user_id bigint;
+
+BEGIN
+    FOR i IN 1..count LOOP
+
+        name := '';
+        status := FLOOR(RANDOM() * 10);
+        --RAISE NOTICE 'Status is %', status;
```
mwest1066, line 28:
```suggestion
```

sprocs/users_randomly_generate.sql
```diff
@@ -0,0 +1,60 @@
...
+            name = name || first_names[FLOOR((RANDOM()*ARRAY_LENGTH(first_names, 1)))::int] || ' ';
+        END IF;
+
+        name = name || first_names[FLOOR((RANDOM()*ARRAY_LENGTH(first_names, 1)))::int] || ' ';
+        name = name || surnames[FLOOR((RANDOM()*ARRAY_LENGTH(surnames, 1)))::int];
+
+        email = LOWER(name);
+        email = REPLACE(email, ' ', '.') || '@example.com';
+        --RAISE NOTICE 'Name is %', name;
+        --RAISE NOTICE 'Email is %', email;
```
mwest1066, line 47:
```suggestion
```

sprocs/users_randomly_generate.sql
```diff
@@ -0,0 +1,60 @@
+CREATE OR REPLACE PROCEDURE
+    users_randomly_generate(
+        count int,
+        course_instance_id text DEFAULT NULL
```
mwest1066, line 4:
```suggestion
        course_instance_id bigint DEFAULT NULL
```

sprocs/users_randomly_generate.sql
```diff
@@ -0,0 +1,60 @@
...
+
+        email = LOWER(name);
+        email = REPLACE(email, ' ', '.') || '@example.com';
+        --RAISE NOTICE 'Name is %', name;
+        --RAISE NOTICE 'Email is %', email;
+
+        INSERT INTO users (name, uid) VALUES (name, email)
+        RETURNING users.user_id INTO _user_id;
+
+        course_instance_id = NULLIF(course_instance_id, '');
```
mwest1066, line 52:
```suggestion
```

I think this is unnecessary if we make the argument a `bigint`?

sprocs/users_randomly_generate.sql
```diff
@@ -0,0 +1,60 @@
...
+        IF course_instance_id IS NOT NULL THEN
+            INSERT INTO enrollments (user_id, course_instance_id, role)
+                VALUES (_user_id, course_instance_id::int, 'Student');
```
mwest1066, line 55:
```suggestion
                VALUES (_user_id, course_instance_id, 'Student');
```

admin_queries/generate_and_enroll_users.sql
trombonekenny: I guess I didn't have a reason to do anything with the output, and I couldn't make a function not have an output, so I switched to a procedure. What use case would you have for seeing the users?

sprocs/users_randomly_generate.sql
trombonekenny: I hesitated on what makes a good random UIN and then gave it up (because it's not important those accounts have one)? I guess we could make a hash or something -- I just didn't want them to conflict and trigger that workflow.

sprocs/users_randomly_generate.sql
trombonekenny: `null` isn't a valid int, so you can't pass it in. Passing in a string and converting it as needed seems to be the way to work around that? Maybe you have a different pattern?

trombonekenny: Most of that complexity is around making the SQL 1) generate random users and 2) optionally enroll them in a specified course instance. If we made the enrollment not optional it might simplify things?

admin_queries/generate_and_enroll_users.sql
mwest1066: Two things: (1) it looks weird when there is no output, and it's unclear whether anything really happened, and (2) I wanted to login as one of the new users, but now I had to go off and query the DB to find their UID, so it would have been much easier to just know a valid UID right there.

Generally it seems like a trivial addition to return the users list?

sprocs/users_randomly_generate.sql
mwest1066: Fair enough, then I'd delete this local variable.

sprocs/users_randomly_generate.sql
mwest1066: In SQL I believe `NULL` is a valid instance of any base type. The problem is that we get an empty string (not null) when the `course_instance_id` input field is blank. I changed the code so that the underlying SQL function takes the correct types (int/bigint) and put the `nullif()` in the `admin_query` SQL query.

admin_queries/generate_and_enroll_users.sql
mwest1066: I added this.

PR #2493, opened 2020-05-08
Add help text for 1D arrays in matrix input element

nicknytko:
image

Clarifies that 1D arrays can be created in the `pl-matrix-input` element, but they are row vectors by default.

### DIFF
--- a/elements/pl-matrix-input/pl-matrix-input.mustache
+++ b/elements/pl-matrix-input/pl-matrix-input.mustache
@@ -122,12 +122,18 @@
 
 {{#format}}
 <p>
-Your answer must be a matrix (i.e., a 2D array). It can be expressed in one of two ways:
+    Your answer should be a matrix (i.e., a 2D array), or a vector (a 1D array).
+</p>
+<p>
+    A <strong>2D array</strong> can be expressed in one of two ways:
 </p>
 <ul>
     <li><strong>Matlab Format.</strong> Enclose it by a single pair of square brackets. Separate entries in each row with a space or comma. Indicate the end of each intermediate row with a semicolon. Examples: <code>[1 2 3; 4 5 6]</code> or <code>[1, 2, 3; 4, 5, 6]</code></li>
     <li><strong>Python Format.</strong> Enclose it by a single pair of square brackets. Enclose each row by a single pair of square brackets. Separate rows with a comma. Separate entries in each row with a comma. Example: <code>[[1, 2, 3], [4, 5, 6]]</code></li>
 </ul>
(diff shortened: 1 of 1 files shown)


elements/pl-matrix-input/pl-matrix-input.mustache
```diff
@@ -122,12 +122,18 @@
 
 {{#format}}
 <p>
-Your answer must be a matrix (i.e., a 2D array). It can be expressed in one of two ways:
+    Your answer should be a matrix (i.e., a 2D array), or a vector (a 1D array).
```
mwest1066, line 125: I think that we should stick with the current wording for the help text, because we really are always entering a 2d array. It's just that we are accepting Matlab OR Python syntax, and `[1,2,3]` is a valid Matlab 2d array so we accept it. That is, we should undo all changes, but add one additional note that I'll write below.

elements/pl-matrix-input/pl-matrix-input.mustache
```diff
@@ -122,12 +122,18 @@
...
+<p>
+    A <strong>2D array</strong> can be expressed in one of two ways:
 </p>
 <ul>
     <li><strong>Matlab Format.</strong> Enclose it by a single pair of square brackets. Separate entries in each row with a space or comma. Indicate the end of each intermediate row with a semicolon. Examples: <code>[1 2 3; 4 5 6]</code> or <code>[1, 2, 3; 4, 5, 6]</code></li>
     <li><strong>Python Format.</strong> Enclose it by a single pair of square brackets. Enclose each row by a single pair of square brackets. Separate rows with a comma. Separate entries in each row with a comma. Example: <code>[[1, 2, 3], [4, 5, 6]]</code></li>
 </ul>
+<p>
+    A <strong>1D array</strong> can be expressed by enclosing the array with a single pair of square brackets.  Separate entries with a space or comma.  Note that this will generate a <i>row vector</i>, if you are trying to create a column vector use the above 2D array syntax.  Example: <code>[1,2,3]</code>
+</p>
```
mwest1066, line 136: I think here we should insert a paragraph here saying something like:
```
Note that [1, 2, 3] is valid Matlab input for a 1x3 matrix (a row vector). If you want a column vector (a 3x1 matrix) then you should enter [1; 2; 3] or [[1], [2], [3]].
```

mwest1066:
For actual 1D arrays I think we should add a `pl-vector-input` element that accepts either Matlab or Python formats for 1D arrays (single square brackets, commas or spaces as separators) and returns a 1D numpy array.

Issue #1843, opened 2019-11-07
Split elements into separate documentation pages.


coatless:
From review in PR #1757,

> We don't have to do this right away, but I was wondering whether we should break up elements.md into a separate file per element? It's a monster page now, and will only be getting longer with time. The pl-drawing element is leading the way with the page-per-element strategy. I was thinking that we could make a new top-level category ("Element reference"?)

reteps:
I particularly like how https://getbootstrap.com/docs/5.0/forms/overview/ is laid out: an overview page, with specific pages per elements. Each element page is in the sidebar, and the overview page provides ways to see what elements you might be interested in.

reteps:
This was done in #13375

PR #2494, opened 2020-05-09
Add pl-overlay element

nicknytko:
Creates the `pl-overlay` element, which allows layering existing elements on top of one another in any specified position.

Adds the `elementOverlay` question to the example course.

image


docs/elements.md
```diff
@@ -1256,6 +1257,60 @@ See the [`pl-drawing` documentation](pl-drawing/index.md) for details.
...
+
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`x` | float | - | The x coordinate of the child element (relative to the left of the canvas)
+`y` | float | - | The y coordinate of the child element (relative to the top of the canvas)
+`center` | boolean | False | If true, the contents will be centered at the given x,y coordinates.  If false, the position defines the top-left corner of the contents.
+
+#### Details
+
+An overlay is pre-defined as a "canvas" with a set width and height.  By default, elements that exceed these canvas boundaries will get partially or totally cut off.  Child elements are wrapped with a `<pl-location>` tag that specifies the position relative to the top-left corner with the `x` and `y` attributes.  Anything inside the location tag will be displayed at that position.  Children are layered in the order they are specified, with later child elements being displayed on top of those defined earlier.
```
mwest1066, line 1307: This isn't an actual `<canvas>`, right? Maybe we should just refer to it as an "overlay area" or something like that?

exampleCourse/questions/elementOverlayFigure/info.json
```diff
@@ -0,0 +1,10 @@
+{
+    "uuid": "F539AB6B-4281-49E5-882D-8FC20C4EBE13",
+    "title": "Element pl-overlay: Adding labels to a figure",
```
mwest1066, line 3:
```suggestion
    "title": "Element pl-overlay: Positioning elements on top of other elements",
```

exampleCourse/questions/elementOverlayFigure/question.html
```diff
@@ -0,0 +1,20 @@
...
+        along with the <code class="user-output">pl-string-input</code> element.
+    </p>
+    <p>
+        Write the name of the missing country:
+    </p>
+</pl-question-panel>
+
+<pl-overlay width="400" height="600" clip="false">
+    <pl-location x="0" y="0">
+       <pl-figure file-name="mapzoom.png" directory="clientFilesQuestion"></pl-figure>
```
coatless, line 14: Mmm, looking at the raw of the image, Ukraine is already labeled. How does this impact rendering of the problem? (e.g. a student right clicks to open the image in a new browser tab?)

exampleCourse/questions/elementOverlayFigure/info.json
```diff
@@ -0,0 +1,10 @@
+{
```
mwest1066, line 1: Can we rename this question to just plain `elementOverlay`, as the basic example of using the element?

docs/elements.md
```diff
@@ -1256,6 +1257,60 @@ See the [`pl-drawing` documentation](pl-drawing/index.md) for details.
 
 -----
 
+## `pl-overlay` element
+
+The overlay element allows existing PrairieLearn and HTML elements to be layered on top of one another in arbitrary positions.
```
coatless, line 1262: Creative re-use of the `pl-drawing` capabilities. 

Maybe emphasize this doesn't allow for an interactive drag/drop approach?

docs/elements.md
```diff
@@ -1256,6 +1257,60 @@ See the [`pl-drawing` documentation](pl-drawing/index.md) for details.
...
+
+#### Sample Element
+
+![](elements/pl-overlay.png)
+
+```html
```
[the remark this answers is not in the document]
mfsilva22: It was intentional. Why do you think we should add this?

exampleCourse/questions/elementOverlayFigure/question.html
mfsilva22: I didn't think of this. I was just trying to find something simple and easy to do. In a real example, the figure will have the missing information. We can edit the figure, if that is important?

exampleCourse/questions/elementOverlayRandom/info.json
```diff
@@ -0,0 +1,10 @@
+{
+    "uuid": "e93af989-ba82-4859-9b86-f850152341ce",
+    "title": "Element pl-overlay: Arbitrary positioning of existing elements",
```
mwest1066, line 3: Can we rename this question to a `demoXXXXX` question? Maybe `demoTriangle`? And change the title to `Demo: randomize a drawing with overlaid inputs` or something like that?

I think we are aiming to have one `elementXXX` question per element, showing basic functionality. And then other more complex questions are `demoXXXX`.

docs/elements.md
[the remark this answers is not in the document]
coatless: @mfsilva22 many users copy and paste from the element documentation. Prior iteration that only showed `question.html` often left people struggling for the randomized portion. 

Plus, the goal in the long term is to spin out each element into its own documentation page (c.f. #1843, #1757)

docs/elements.md
```diff
@@ -1256,6 +1257,60 @@ See the [`pl-drawing` documentation](pl-drawing/index.md) for details.
...
+`height` | float | - | The height of the overlay canvas in pixels.
+`clip` | boolean | True | If true, children will be cut off when exceeding canvas boundaries.
+
+#### `pl-location` Customizations
+
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`x` | float | - | The x coordinate of the child element (relative to the left of the canvas)
+`y` | float | - | The y coordinate of the child element (relative to the top of the canvas)
+`center` | boolean | False | If true, the contents will be centered at the given x,y coordinates.  If false, the position defines the top-left corner of the contents.
```
mwest1066, line 1303: Could we make this a bit more general by having `valign` and `halign` attributes? `valign` could be `top`, `middle`, `bottom`. `halign` could be `left`, `center`, `right`. (copying from the HTML table attributes)

docs/elements.md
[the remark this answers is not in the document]
nicknytko: I took out any randomization in the docs example, so this should work without any `server` code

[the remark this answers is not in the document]
mfsilva22: Feel free to delete. I am not sure I agree that every question needs to be completely randomized. The other example is complicated for new users, and I just thought of adding a simple one, to lower the barrier for new adopters.

docs/elements.md
nicknytko: Sure, if we want more control over the alignment this can be added.  If we do this I think it'd make more sense to have both centered by default?

mwest1066: Yep, makes sense.

docs/elements.md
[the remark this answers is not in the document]
mwest1066: Yep, I think both examples are great!

mwest1066:
Also, can you make a skeleton news item for this element?

nicknytko:
> Also, can you make a skeleton news item for this element?

image

docs/elements.md
```diff
@@ -1256,6 +1257,61 @@ See the [`pl-drawing` documentation](pl-drawing/index.md) for details.
...
+`width` | float | - | The width of the overlay canvas in pixels.
+`height` | float | - | The height of the overlay canvas in pixels.
+`clip` | boolean | True | If true, children will be cut off when exceeding overlay boundaries.
+
+#### `pl-location` Customizations
+
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`x` | float | - | The x coordinate of the child element (relative to the left of the canvas)
+`y` | float | - | The y coordinate of the child element (relative to the top of the canvas)
```
tdy, line 1302: Hmm I find bottom-left to be a more intuitive anchor. I'm used to annotating plots, so my brain still thinks of these locations in that frame of reference, but maybe that's just me.

nicknytko: Not sure I’d necessarily agree with having this as the default, but I’d be happy to add this as an optional attribute.

mwest1066: The `position` attribute actually has `top`, `left`, `right`, `bottom` attributes: https://developer.mozilla.org/en-US/docs/Web/CSS/position

Maybe we should just use those?

tdy: (actually I just realized this would be more like annotating an image, which usually has top-left origin anyway)

nicknytko: Hmm, yeah we can go with @mwest1066 's suggestion here since that'll let people use whatever system they're more comfortable with.

exampleCourse/questions/elementOverlay/question.html
```diff
@@ -0,0 +1,20 @@
+<pl-question-panel>
+    <p>
+        The element <code class="user-output">pl-overlay</code> allows the positioning of pre-existing elements in arbitrary positions on top of other elements.
+        This question combines the <code class="user-output">pl-figure</code> element to serve as a background,
+        along with the <code class="user-output">pl-string-input</code> element.
```
mwest1066, line 5: Wrap the above in italics?

exampleCourse/questions/elementOverlay/question.html
```diff
@@ -0,0 +1,20 @@
...
+    </p>
+    <p>
+        Write the name of the missing country:
```
mwest1066, line 8:
```suggestion
        Write the name of the unlabeled country:
```

docs/elements.md
```diff
@@ -1256,6 +1257,63 @@ See the [`pl-drawing` documentation](pl-drawing/index.md) for details.
...
+#### `pl-location` Customizations
+
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`left` | float | - | The x coordinate of the child element (relative to the left of the overlay)
+`top` | float | - | The y coordinate of the child element (relative to the top of the overlay)
+`right` | float | - | The x coordinate of the child element (relative to the right of the overlay)
+`bottom` | float | - | The y coordinate of the child element (relative to the bottom of the overlay)
+`valign` | string | `middle` | Specifies the vertical alignment of the contents.  Can be one of `top`, `middle`, or `bottom`.
+`halign` | string | `center` | Specifies the horizontal alignment of the contents.  Can be one of `left`, `center`, or `right`.
```
mwest1066, line 1306: A possible future enhancement would be to have another child type of `<pl-base-image>` from which we automatically pull the width/height. At the moment it's a bit of messing around to get an image, find it's dimensions, set those explicitly as width/height for `pl-overlay`, include the image at position (0,0), etc.

Or maybe we should just have a `base-image="mapzoom.png"` attribute for `pl-overlay`? I'm not sure.

Another thought: it would be really cool to have a demo question which has an image with many locations on it. For example, the US map of states. The Python file could have a big list of (x,y) locations and state abbreviations, and we could show how to randomize the placement of the text input box over a state and the setting of the correct answer. This would be an example of a "label parts of an image" question.

Another way to set up a random problem like this would be to choose five random elements from the big list of (x,y,name) tuples, use the first one to overlay the simple text "State X" on the image, and use the five elements as MC alternatives (with the first one marked as correct).

I think both of these are good general patterns for labeling questions and we should probably include both.

mwest1066: It occurs to me that we don't have a `pl-dropdown` element. If we did have one, then it would be cool to have a map of the states with five random states containing dropdowns with all 50 states listed in them. So the student has to pick the names of 5 different states, getting 20% of the points for each one that is correct.

nicknytko:
image

mwest1066:
@nicknytko I pushed some changes to the news item. Please check!

nicknytko:
News article looks good!  With regards to the above issue, that looks to be a bug in `pl-drawing` itself, so I'll try to get a fix in a new PR.

docs/elements.md
```diff
@@ -1268,13 +1268,13 @@ The overlay element allows existing PrairieLearn and HTML elements to be layered
 
 ```html
 <pl-overlay width="400" height="400" clip="false">
-    <pl-location left="0" top="0" halign="left" valign="top">
+    <pl-background>
```
coatless, line 1271: Missing `pl-background` in documentation even though it's a specific version of `pl-location`, still useful to have its own table.

elements/pl-overlay/pl-overlay.py
```diff
@@ -0,0 +1,124 @@
...
+        if child.tag == 'pl-location':
+            pl.check_attribs(child, required_attribs=[], optional_attribs=['left', 'right', 'top', 'bottom', 'valign', 'halign'])
+
+            if ('left' not in child.attrib and 'right' not in child.attrib) or ('left' in child.attrib and 'right' in child.attrib):
+                raise ValueError('pl-location requires exactly one of "left" or "right" attributes.')
+
+            if ('top' not in child.attrib and 'bottom' not in child.attrib) or ('top' in child.attrib and 'bottom' in child.attrib):
+                raise ValueError('pl-location requires exactly one of "top" or "bottom" attributes.')
+
+            valign = pl.get_string_attrib(child, 'valign', None)
```
coatless, line 41:
```suggestion
            valign = pl.get_string_attrib(child, 'valign', VALIGN_DEFAULT)
```

elements/pl-overlay/pl-overlay.py
```diff
@@ -0,0 +1,124 @@
...
+            if valign is not None and valign not in VALIGN_VALUES:
```
coatless, line 42: I think using the default would be better than a special case here...

elements/pl-overlay/pl-overlay.py
```diff
@@ -0,0 +1,124 @@
...
+                raise ValueError(f'Unknown vertical alignment "{valign}"')
+
+            halign = pl.get_string_attrib(child, 'halign', None)
```
coatless, line 45:
```suggestion
            halign = pl.get_string_attrib(child, 'halign', HALIGN_DEFAULT)
```

elements/pl-overlay/pl-overlay.py
```diff
@@ -0,0 +1,124 @@
...
+            if halign is not None and halign not in HALIGN_VALUES:
```
coatless, line 46: ditto

docs/elements.md
nicknytko: Yep, though the `pl-background` child doesn’t have any attributes that can get set.

coatless: Eh, maybe an empty table with a line underneath saying no attributes are necessary? It's handled by `pl-drawing`? I'm just worried that folks will overlook the brief mention in the details section.

elements/pl-overlay/pl-overlay.py
```diff
@@ -0,0 +1,124 @@
...
+    'middle': '-50%',
+    'center': '-50%',
+    'bottom': '-100%',
+    'right': '-100%'
+}
+
+
+def prepare(element_html, data):
+    element = lxml.html.fragment_fromstring(element_html)
+    #
```
coatless, line 26: Should there be a comment here?

nicknytko: I have no idea what I was trying to do here :shrug:

PR #2495, opened 2020-05-11
Add option to fast-error if too busy

mwest1066:


lib/workers.js
```diff
@@ -92,10 +92,18 @@ module.exports = {
...
+            if (config.workerUseQueue) {
+                debug(`getPythonCaller(): adding to waitingCallbacks`);
+                callback.__load_uuid = uuidv4();
+                load.startJob('python_callback_waiting', callback.__load_uuid);
+                waitingCallbacks.push(callback);
+            } else {
+                debug(`getPythonCaller(): no workers available, waiting to error`);
+                setTimeout(() => {
+                    debug(`getPythonCaller(): no workers available, reporting error`);
+                    callback(new Error('Server is overloaded. Please try again.'));
```
nicknytko, line 104: whats the logic here behind waiting to throw an error?

mwest1066: To slow down submissions. This will only be triggered if we are overloaded, so we want to apply [backpressure](https://medium.com/@jayphelps/backpressure-explained-the-flow-of-data-through-software-2350b3e77ce7).

nicknytko: Interesting article, that makes sense

PR #2496, opened 2020-05-12
Add caching for file() requests

mwest1066:
This PR uses the cache for the data generated by file() requests, mirroring the existing usage for HTML question panel rendering. To test it, run with `NODE_ENV=production` and set `"questionRenderCacheType": "memory"` in `config.json`.

### DIFF
--- a/question-servers/freeform.js
+++ b/question-servers/freeform.js
@@ -10,6 +10,7 @@ const parse5 = require('parse5');
 const debug = require('debug')('prairielearn:' + path.basename(__filename, '.js'));
 
 const schemas = require('../schemas');
+const config = require('../lib/config');
 const logger = require('../lib/logger');
 const codeCaller = require('../lib/code-caller');
 const workers = require('../lib/workers');
@@ -768,14 +769,6 @@ module.exports = {
         });
     },
 
-    _getCacheKey: function(course, data, callback) {
-        courseUtil.getOrUpdateCourseCommitHash(course, (err, commitHash) => {
-            if (ERR(err, callback)) return;
-            const dataHash = hash('sha1').update(JSON.stringify(data)).digest('base64');
-            callback(null, `${commitHash}-${dataHash}`);
-        });
-    },
-
     renderPanel: function(panel, pc, variant, question, submission, course, locals, callback) {
         debug(`renderPanel(${panel})`);
         // broken variant kills all rendering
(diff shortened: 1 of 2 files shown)


question-servers/freeform.js
```diff
@@ -1093,17 +1057,36 @@ module.exports = {
...
+                        });
                     });
-                });
-            });
+                },
+                (cachedData, _cacheHit) => {
+                    // function to process the cachedData, whether we
+                    // just rendered it or whether it came from cache
+                    const {courseIssues, fileDataBase64} = cachedData;
+                    const fileData = Buffer.from(fileDataBase64, 'base64');
```
nicknytko, line 1085: I haven't played around with the caching code so maybe this is an obvious question, but is it necessary to convert to/from base64 when storing the file data?

question-servers/freeform.js
```diff
@@ -1255,4 +1238,62 @@ module.exports = {
...
+
+    getCacheKey: function(course, data, context, callback) {
+        courseUtil.getOrUpdateCourseCommitHash(course, (err, commitHash) => {
+            if (ERR(err, callback)) return;
+            const dataHash = hash('sha1').update(JSON.stringify({data, context})).digest('base64');
+            callback(null, `${commitHash}-${dataHash}`);
+        });
+    },
+
+    getCachedDataOrCompute: function(course, data, context, computeFcn, processFcn, errorFcn) {
```
nicknytko, line 1250: Nice refactor! Almost makes me wish there was more we could cache with the freeform server

question-servers/freeform.js
mwest1066: My initial local testing with the memory cache worked fine leaving the `fileData` as a buffer, but then when I tried it on `pl-dev` (using redis) it didn't work. Changing to base64 fixed it for redis, and it still worked with the memory cache as well. I didn't investigate to figure out why redis had a problem with binary data.

question-servers/freeform.js
mwest1066: Yes! Caching is awesome! In typical usage, more than half of all question HTML is served from the cache. @nwalters512 wrote this cache code and convinced me that it was actually worth running 😄

mwest1066:
> Looks good, I was able to confirm it caches dynamic figures on a local instance set to production mode.

Thanks so much for the rapid check!

question-servers/freeform.js
nicknytko: Strange, but I suppose if it works... 🤷

echuber2: Redis strings are supposedly binary safe, so maybe the issue arose in some other part of the toolchain.

PR #2501, opened 2020-05-13
Decluttering the instructor navbar

astrosticks:
The instructor navbar is getting noticeably large, text-heavy, and colorful. This PR combats these by re-ordering, re-designing, and moving some elements around.

|Decluttered Navbar|
|-|
|**Before**|
|image| 
|**After**|
|image|

It would be nice if we could make the "load from disk" button smaller too, but I'm not sure of the best way to do it. Suggestions welcome.


mfsilva22:
I like the new look! 

Two comments:
- I noticed that json info is getting printed in the terminal where docker is running
- after loading page for the first time, I see:
image
However, if I click "Settings" and then go to Dev User dropdown menu again, I see:
image
seems odd to me that one of the options disappear.

astrosticks:
> seems odd to me that one of the options disappear.  

The changes here don't affect the control flow for when these options appear to the user, so that might be out of scope for this PR

PR #2502, opened 2020-05-14
Add extensions support to pl-drawing

### LINKED ISSUE #2510: Old submissions don't render pl-drawing correctly
From #2494 

Steps to reproduce:

1. Go to the `demoOverlayTriangle` question as an instructor.

2. Submit multiple answers.

3. Open the submission panel for the first submission.

4. The `pl-drawing` element does not render:

nicknytko:
Adds support for any element to be placed on manipulated on the canvas, and also an interface for course-level extensions.
Adds `demo/drawing/buttons` to show all buttons:

image

Adds `demo/drawing/customizedButtons` to show how button elements may be customized:

image

Adds `demo/drawing/extensions` to showcase an example of a custom extension:

image

Also fixes #2510


elements/pl-drawing/pl-drawing.mustache
```diff
@@ -6,21 +6,19 @@
...
 </script>
-<span id="drawing-interface-{{uuid}}">
-    <a role="button" class="btn btn-light btn-sm" data-placement="auto"
-       data-trigger="focus" data-toggle="popover" data-html="true" title="Error"
-       tabindex="0" data-content="{{parse_error}}">
-        INVALID <span><i class="fa fa-question-circle" aria-hidden="true"></i></span></a>
-</span>
+<span id="pl-drawing-{{uuid}}">
+    <span class="badge text-danger badge-invalid"><i class="fa fa-exclamation-triangle" aria-hidden="true"></i> Invalid</span>
+    <a href="javascript:void(0);" role="button" class="btn btn-sm btn-secondary small border" data-placement="auto" data-trigger="focus" data-toggle="popover" data-html="true" style="margin-left: 5px" title="Format Error" tabindex="0" data-content="{{parse_error}}"> Why <i class="fa fa-question-circle" aria-hidden="true"></i>
```
mwest1066, line 17: Why do we have `javascript:void(0);` here? I seem to recall that this was also used in a few places in the PL `.ejs` files but it was unnecessary, and not recommended by the bootstrap docs? Also, any chance we could use a class like `ml-3` rather than explicit margins? (see https://getbootstrap.com/docs/4.6/utilities/spacing/)

exampleCourse/courseInstances/Sp15/assessments/hw05-demoDrawingExamples/infoAssessment.json
```diff
@@ -21,7 +21,7 @@
     "zones": [
         {
             "questions": [
-                {"id": "element/drawingGallery",                 "points": 2, "maxPoints": 6},
+                {"id": "element/drawingGallery",                  "points": 2, "maxPoints": 6},
```
mwest1066, line 24: I didn't go through all the demo questions carefully, but I did notice that:

1. `demo/drawing/centroid` allows the student to move the triangle and rectangle -- I think they should be fixed?
2. `demo/drawing/centroid` doesn't show the correct answer
3. `demo/drawing/vmDiagrams` doesn't show the correct answer

I don't want to hold up this PR now because of demo questions, so maybe we should add an issue about these questions and also that we should go through the other demo drawing questions?

mwest1066 (approved):

elements/pl-drawing/pl-drawing.mustache
nicknytko: Good catch, this was in the other elements and I just copy+pasted it over but I agree it can be re-worked.  I've updated it here to be a button, but we should probably stick this into an issue or something.

PR #2504, opened 2020-05-14
Add RFC for container workspaces

mwest1066:
This RFC is a (very slightly) cleaned up version of the notes from @nwalters512 

- [x] "Database" section ([currently L76](https://github.com/PrairieLearn/PrairieLearn/pull/2504/files#diff-43b441d676beb36cd12f4de809bdfeebR76)) - @tdy 
- [x] "Questions" section ([currently L105](https://github.com/PrairieLearn/PrairieLearn/pull/2504/files#diff-43b441d676beb36cd12f4de809bdfeebR105)) - @tdy 
- [x] "Student-facing question interface" section ([currently L143](https://github.com/PrairieLearn/PrairieLearn/pull/2504/files#diff-43b441d676beb36cd12f4de809bdfeebR143)) - @tdy 
- [x] Server code that serves `/workspace/<workspace_id>` serving HTML and JS  - @polohan
  - Should at least contain a hardcoded `iframe` pointing at `/workspace/<workspace_id>/container/`
  - Can contain frontend JS code itself that connects back with a websocket to the server
- [x] Webserver proxy code (`/workspace/<workspace_id>/container/*`) -- See #2508 - @polohan (note that at this point, it should always proxy to localhost:13746 for testing - proxying to different hosts will be done later)

After the above have been completed, we can piece together an intermediate deliverable, where workspaces are functional on a local development machine. We should be able to visit a question with a workspace, click the workspace button, and see the proxied workspace in their browser. PrairieLearn will not yet be dealing with saving files, creating submissions, allocating hosts, etc.

---

### Example workspace questions (tracked in #3300)

* [x] VS Code (code-server): https://hub.docker.com/r/codercom/code-server
* [ ] VS Code (theia): https://hub.docker.com/r/theiaide/theia/
* [x] Jupyter notebooks: https://hub.docker.com/r/jupyter/datascience-notebook
* [ ] RStudio
* [ ] Observable notebooks
* [x] xterm.js: https://hub.docker.com/r/xtermjs/xterm.js/

---

### Old to do list:

- [x] @tdy - Add code to support `HOST_JOBS_DIR` and localhost changes for running `interface.js` inside container
- [x] @tdy - Add sqldb integration with `interface.js` and change docker settings to be queried
- [x] @tdy The outer frame on `pages/workspace/` should not be the regular PL navbar, but instead should just display the current workspace `state` variable and buttons like "Reboot workspace", "Stop workspace". The workspace `state` variable should be updated by the websocket code. The buttons should go to URLs like `/workspaces/<ID>/reboot` which should do the required action and then redirect back to `/workspaces/<ID>/`.
- [x] @tdy Change `interface.js` so that it doesn't always load `config.json` but uses the same command line arguments as `server.js` to allow `--config` to specify an alternative configuration filename.
- [x] @tdy Allow running of multiple copies of `interface.js` locally, with different config filenames to set different port numbers.
- [x] @tdy Rename config variables. `workspaceContainerLocalhost` and `workspaceContainerPort` should be combined into one setting called `workspaceDevHost`. `workspaceNativeLocalhost` should be renamed to `workspaceDevContainer`.
- [x] @tdy Use state machine for all operations in both PL server and workspace host.
- [x] @tdy Unique names for each container launch. On the workspace host, every time it is asked to launch a container, it should allocate a unique name for that container (e.g., `id_workspace_mapper[workspace_id].name = generate_random_uuid()`). Then this name should be used for the local directory and the docker container name, rather than using `workspace-<ID>`. This will mean that repeated launches will be safe. The S3 name will be `workspace-<ID>`, the local directory will be `name`.
- [x] @polohan Rewrite `_getAvailablePort()` in `interface.js` to use a synchronous and non-recursive scheme to look up a free port number. We should add a config option for the starting port (default 9090 or something) and just use the first port number above this that isn't currently allocated in `port_id_mapper`.
- [x] @tdy - Add an option in `workspaceOptions` to set the home directory inside the container. At the moment this is hard-coded to `/home/coder/project` or something like that, but different workspaces will have different requirements.
- [x] @tdy When the workspace host starts a container, it should allocate a port, start the container, and then update the DB to set `workspaces.hostname` (replaces `workspaces.port`) to the correct `hostname:port` address. This way the main PL server can proxy directly to `${workspaces.hostname}` and we can remove all the proxy code from the workspace host. Right now, the value of `workspaces.hostname` will be `${config.workspaceDevContainer}:port`.
- [x] @tdy When running in development mode, the workspace host should set `workspaces_hosts.hostname` to the value from the config variable. The PL server should use `workspaces.workspace_host_id` to look up this value from the DB, rather than using the config variable.
- [x] @polohan The function `s3Sync()` in `pages/workspace/workspace.js` needs to be written to be async using callbacks. It should only use functions like `fs.readdir()`, not `fs.readdirSync()`. Maybe the right way to do this is to unify all the S3 code from `interface.js` and `workspace.js` into a library inside PL.
- [x] Set up the development docker container to use `localstack`, so that we don't need to use real S3 buckets.
- [x] @polohan Deleting files in S3 on file sync, maybe use `fs.watch()` with rate limiter and locking.
- [x] @polohan Fix the bug where you need to reload the workspace page on first launch to get the workspace in the iframe.
- [x] @tdy `workspace_hosts.load_count` to track the total number of running workspaces. Update this every time we start or stop a container.
- [x] @tdy Add `syncIgnore` option to `workspaceOptions` (e.g., ignore code-server's `.local/share/code-server/heartbeat`)

---

### For local-dev MVP:

- [x] @polohan Add a webhook route to `interface.js` to send back the `gradedFiles` when requested.
- [x] @tdy Grading integration in `server.js` after clicking the "Grade" button.
- [x] Fix the Xterm.js question to work with the new flex iframe (or remove this question).
- [x] @nicknytko Set up hub.docker.com auto-builds of the workspace images.
- [x] @polohan Pull the docker image before using it. Like [this code](https://github.com/PrairieLearn/PrairieLearn/blob/master/lib/externalGraderLocal.js#L72). Also add an option `config.workspacePullImagesFromDockerHub` like [`config.externalGradingPullImagesFromDockerHub`](https://github.com/PrairieLearn/PrairieLearn/blob/master/lib/config.js#L72).
- [x] @tdy Add docs. Put a big warning at the top that this is an alpha release and should not be used for students yet. Say that it currently only works with Chrome and Firefox (not Safari). Include instructions on how to run workspaces locally with docker (on MacOS, Windows, Linux). For MacOS, this is:
```
docker run -it --rm -p 3000:3000 -v "$HOME/pl_ag_jobs:/jobs" -e HOST_JOBS_DIR="$HOME/pl_ag_jobs" -v /var/run/docker.sock:/var/run/docker.sock prairielearn/prairielearn /PrairieLearn/docker/start_workspace.sh
```

---

### For initial production MVP:

- [x] @nicknytko Fix workspaces with Safari. It sometimes works but sometimes fails with `Compressed bit must be 0 if no negotiated deflate-frame extension` in the browser console. A page reload normally fixes it.
- [x] @tdy Clean up the help text in the workspace navbar that explains about reload/close.
- [x] @nicknytko  In production, get the real `instanceId` and real hostname/IP for `workspace_hosts`.
- [x] @nicknytko  Fix the `awsConfig` in `interface.js` so that it uses the current one in dev but the auto-created one in prod.
- [x] @nicknytko  Add ansible config.
- [x] @nicknytko Make `interface.js` restartable:
    - [x] Remove all local state from `interface.js` so it can die/restart and keep going. Store all state in the DB. In particular, store the `launch_uuid` in the DB (this is the UUID part of `localName = "workspace-UUID"`.
    - [x] When `interface.js` starts up, it should check whether there are running workspaces on it (meaning we are recovering from a crash) and do a full S3 sync of them in case we missed file changes while crashed.
- [x] @tdy Add shutdown and cleanup of old containers:
    - [x] Add `workspaces.launched_at` and set this every time we launch a container (the initSequence in `interface.js`)
    - [x] Add `workspaces.heartbeat_at`. Add a heartbeat from the socket code that updates this value every minute.
    - [x] In PL add a job to `cron/` to check for old workspaces and set their state to be `stopped`. We define "old" to be `heartbeat_at` older than 10 minutes or `launched_at` older than 12 hours.
    - [x] In `interface.js`, every minute run a check to see whether we have any running containers that are either stopped or for which there isn't a corresponding `launch_uuid` in the DB. For any such containers, kill and remove them.
- [x] @nicknytko Make sure all code is async:
    - [x] Replace `adm-zip` with `archiver` in `interface.js` and `workspace.js`. See `pages/instructorAssessmentDownload` for an example of using `archiver`. The internals of `adm-zip` use sync code all over the place. I think it's possible to use it in an async way, but I spent an hour reading their internal code and I only half-succeeded. Switching to `archiver` seems much safer.
    - [x] In `lib/workspace.js` the `s3Sync()` function needs to be replaced with an async version.
- [x] @tdy Fix S3 race condition where we don't always get all the initial files.
    * [x] When the PL server is copying the initial files to S3, it should make a zip of them and copy this to `workspace-[N]/initial.zip`.
    * [x] The main server should also copy the initial files to `workspace-[N]/current/` as individual files.
    * [x] Modify the `init` webhook in `interface.js` so that it ~~takes a parameter called `useInitialZip`~~ (instead I'm passing in `req.body.state`, then determining `useInitialZip` on the host which passes it to `initSequence`). The PL server should set `useInitialZip=true` if we are starting a workspace that was `uninitialized`. If the workspace is in any other state then it should use `useInitialZip=false`.
    * [x] In the `init` method of `interface.js`, if `useInitialZip` is `true` then we should not copy the files from S3 as usual, but instead we should copy `initial.zip` and unpack it locally to get the files.
    * [x] All file watching/copying code in `interface.js` will need to be modified to copy to `workspace-[N]/current/` rather than just `workspace-[N]` as we are doing at the moment.
- [x] @tdy Change all `logger.info()` calls to `debug()`. We don't want to be spamming the logs in production.
- [x] @tdy Add `workspaces.state_updated_at` and set it to `now()` whenever we update the state.

---

### For full production:

- [x] Add a news item.
- [x] @nicknytko Security: Add a middleware to translate `workspace_id` into `locals.course_id`, `locals.course_instance_id`, and `locals.instance_question_id` (some of these might be `null`). Then call the code from `middlewares/authzCourse` or `middlewares/authzCourseInstance`, and optionally call `middlewares/selectAndAuthzInstanceQuestion`. These calls should be done on access to `pages/workspace` and in the proxy setup code.
- [x] @tdy Add course name and question number in the workspace navbar. Something like "Q3 (Exam 2, CS 105)". I'm sure students are going to have multiple workspaces open at once and it will be easy for them to get confused about which one is which.
- [x] @tdy "Reset to initial state" button for students.
    - [x] add a `workspaces.version` column that defaults to 1
    - [x] the S3 directory name is now of the format `workspace-ID-VERSION` (we'll need to change all the places in the code that compute the S3 workspace directory name)
    - [x] to reset a workspace we set it to `uninitialized`, increment `version`, and then redirect to the workspace URL
    - [x] to reboot a workspace we set it to `stopped`, leave the `version` alone, and then redirect to the workspace URL
    - [x] the launch code will not require any changes at all, because it will see that the workspace is uninitialized, it will make the new `workspace-ID-VERSION` (with the new version), and everything proceeds as normal
    - [x] `workspace_logs` should also have a version column that we set to the current version whenever we write a log entry
- [x] @tdy Write data to `workspace_logs` on all state changes.
- [ ] **(tracked in #3289)** Locking of workspace and workspace_host rows before updating.
- [ ] **(tracked in #3298)** Verify state machine and add checks for correct state before transitions.
- [x] @tdy Add a confirmation dialog to the "reboot" button.
- [x] @tdy Add `progress` websocket messages from `interface.js` to the client, so that we can show progress while pulling images, copying files, etc.
- [x] @tdy When we are adding the workspace files to the submission in `question.js`, we should check whether we have all the `gradedFiles`. If any are missing, we should add this to `data.format_errors` so the submission will be treated as invalid.
- [x] @tdy Update workspace `navbar-brand`.
- [x] @tdy Make `workspace/` question subdir optional (in which case we'll send an empty `initial.zip`).
- [ ] **(tracked in #3139)** Fix user ID mapping into containers to avoid the `I have no name!` username.
- [x] @nicknytko Autoscaling.
    - [x] Choose workspace host by most-recent-host-with-available-slots.
    - [x] Every N seconds, each workspace host should write information about itself to the `workspace_hosts` table. For example, each grading host (running PrairieGrader) writes it's load, health status, AWS [lifecycle](https://docs.aws.amazon.com/autoscaling/ec2/userguide/lifecycle-hooks.html), etc. This also updates `date` with the current timestamp, so we know which hosts are still alive. See [this code](https://github.com/PrairieLearn/PrairieGrader/blob/1579033b12ef3545dfc1e2c29aedd0d09eda8f27/lib/load.js#L64). Note that the `grader_loads` table is the equivalent of the `workspace_hosts` table. At some point we should rename `grader_loads -> grader_hosts`.
    - [x] Every M seconds, a PL cron job should:
        - [x] Aggregate the per-host information into bulk statistics. For example, for grader stats this is done [here](https://github.com/PrairieLearn/PrairieLearn/blob/6301b3fae85c3683a1cad358c983bf9d4d06a1ee/cron/externalGraderLoad.js#L25). This also generates bulk statistics about the grading jobs themselves (separate to the grading hosts stats). For workspaces we will have both workspace host statistics (e.g., number of live hosts) and workspace statistics (e.g., number of running workspaces).
        - [x] Compute number of desired workspace hosts as `number_of_running_workspaces / workspaces_per_host`, where `workspaces_per_host=20` is some config constant.
        - [x] If we have more workspace hosts than the desired number, kill any excess ones that don't have any current running workspaces on them.
        - [x] If we have less workspace hosts than the desired number, launch some more.
        - [x] Alternatively, we could use the PrairieGrader strategy of just setting the desired number of hosts for AWS autoscaling and use lifecycle actions to delay termination until all running workspaces have completed. This would also mean that we should not allocate new workspaces to hosts that are terminating.
        - [x] Mark workspaces as stopped if they don't have a live host anymore.
    - [x] Add health checks to `interface.js` (especially for docker errors).
- [ ] **(tracked in #3141)** @polohan Rewrite the file watching and upload code in `interface.js` to fix race conditions. An outline of an algorithm is:
    * Have two dictionaries of files called `update_queue` and `currently_uploading_queue`. The `update_queue` is files that need uploading, while `currently_uploading_queue` is files that are in the process of being uploaded.
    * Whenever we get a watch notification of a modified file, add it to `update_queue` (this is the current algorithm).
    * Every N seconds run `_autoUpdateJobManager()`. This takes each file that is on the `update_queue` and not on the `currently_updating_queue`, creates an upload job for that file, removes it from the `update_queue`, and adds it to the `currently_updating_queue`. Then start parallel execution of all the upload jobs. This is similar to the current algorithm, but with the addition of the `currently_updating_queue`.
    * When a file finishes uploading, we remove it from the currently_updating_queue.
    * I believe that this algorithm guarantees that we will upload all modified files eventually and we will never try and upload the same file twice at the same time.
    * For extra safety, before we start upload jobs for files on the `update_queue`, we should check that they are associated with the current `launch_uuid` for a still-running workspace.
    * Note that the current algorithm uses `[path, isDirectory_str] = key.split(',')` in `interface.js` to store both a filename and whether it's a directory. This will fail if the path contains a comma. We should convert to a more robust method of handling directories. Perhaps store the key in the format `"Dpath"` or `"Fpath"`? The branch #2834 contains a fix for this, but it introduces other problems.

---

### Future enhancements to consider:

- [ ] **(tracked in #3290)** Make a demo question to run a full graphical desktop inside a workspace. We'd need a browser-based VNC/RDP client like https://guacamole.apache.org or https://novnc.com in a workspace container that runs a linux graphical desktop and a VNC server. This would be useful for running things like Matlab, CAD programs, etc.
- [ ] **(tracked in #3169)** Add file upload/download buttons in the workspace navbar, to put or get a file directly into the workspace home directory. These would need `interface.js` to add two new functions for `putFile` and `getFile`, so that the PL web server could relay the file to the host.
- [ ] **(no longer implementing)** Integrate `workspace_host/interface.js` into the main PL `server.js`. We'll need command line args to specify whether to load (a) the "regular" PL routes, and/or (b) the "workspace host" routes.
- [ ] **(tracked in #3193)** Add option to disable outgoing network access to workspace containers. Perhaps this could be implemented like https://stackoverflow.com/a/47008421
- [x] Add workspace awareness to page titles (`pages/partials/title.ejs`).
- [ ] **(tracked in #3297)** Add `workspaceOptions.optionalFiles`.
- [ ] **(tracked in #3168)** Change `syncIgnore` and `gradedFiles` to use glob/regex patterns.
- [ ] **(tracked in #3293)** Add option to randomize starter files.
- [ ] **(tracked in #3296)** Add workspace-cloning for instructors. This copies the home directory of a student workspace into a new workspace for the instructor, so they can (for example) view the student's jupyter notebook without having to download it.
- [ ] **(tracked in #3295)** Redo `docker/start_workspace.sh` with docker-compose.
- [ ] **(tracked in #3292)** Allow instructors to specify CPU-percentage and RAM limits in workspaceOptions. Then each host will calculate its current load based on total CPU-percentage of the running containers.
- [ ] **(tracked in #3291)** Add different "pools" of workspace hosts, so we can have a pool with GPUs, for example.
- [ ] **(tracked in #3294)** Add support for workspaces that aren't inside questions. These would not be gradable, but would just provide students with access to a customized computing environment. We could host these workspaces at the course level or even at the top level of PL itself.


rfcs/003-workspaces.md
```diff
@@ -0,0 +1,70 @@
...
+* Workspaces should auto-terminate after a period of inactivity (or force-terminated by the student), but should be able to be re-launched with the persistent home-directory files restored.
+
+# Proposed solution
+
+The server architecture has three conceptual components:
+1. Main PL web servers: render questions for the student with a "launch workspace" button in them.
+2. Manager servers: coordinate the launching of workspaces and proxy all traffic from the student browser through to the host machines.
+3. Host servers: run the actual containers.
+
+These three components are implemented within the main PL executable, but for deployment we can run different fleets of servers that use a `config` option to only turn on specific functionality.
```
tdy, line 29: Wait, so the web/proxy/host servers are all the same "PL" server, just with respective settings enabled?

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,70 @@
...
+* On the host:
+    * `/editor/[uuid]/editor/*` goes to the host that’s running this container
+    *  Within the host, we’ll proxy that to the appropriate container
+    *  Each container will probably need port 80 bound to some random, unique port that we can target for forwards
+* How to map requests to hosts?
+    * Hosts table that stores current information about each host VM
+* How do we kill off old containers?
+    * Containers are killed after either:
+        * We don’t receive N heartbeats in a row
+        * The user hasn’t saved for X amount of time
```
tdy, line 64:
```suggestion
* The user clicks on that button.
* We get a request for a particular workspace instance.
* We check the authorization cookies to verify that the requesting user matches the authorized user for this workspace.
* Routes:
    * `/workspace/[uuid]` - serves outer frame
    * `/workspace/[uuid]/heartbeat`
    * `/workspace/[uuid]/container/*` - proxy `*` to inner frame
* On the host:
    * `/workspace/[uuid]/workspace/*` goes to the host that’s running this container.
    *  Within the host, we’ll proxy that to the appropriate container.
    *  Each container will probably need port 80 bound to some random, unique port that we can target for forwards.
    * The host will listen for three types of signals: launch, sync, and kill container.
* How to map requests to hosts?
    * Hosts table that stores current information about each host VM.
* How do we kill off old containers?
    * Containers are killed after either:
        * We don’t receive N heartbeats in a row.
        * The user hasn’t saved for X amount of time.
```

tdy: Seems like suggestion-commits don't highlight specific changes. Main differences aside from nits:
* auth cookies (L51)
* editor->workspace
* host signals (L60)

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,70 @@
...
+
+# Motivation and background
+
+PrairieLearn currently allows code editing using an in-page ACE Editor and can compile and test code via external graders, which run take student code and instructor-provided test code and execute them in an instructor-defined container, returning the test results back to the student. While excellent for testing small code snippets, this is not a very flexible environment for writing and debugging more complex programs.
+
+This RFC proposes to give students persistent remote containers (called _Workspaces_) to work in, configured by instructors to provide a per-question environment with a specific set of compilers, debuggers, editors, etc. This remote container would be accessed via a web-based frontend, such as VS Code or Jupyter Notebooks.
+
+Goals:
+* Workspaces should be instructor-defined on a per-question basis.
+* Workspaces should launch from a PL question with a set of initial files in the home directory, which may be dynamically generated by the question `server.py` code.
```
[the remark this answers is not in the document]
tdy: New subpoint about readonly reference copy.

rfcs/003-workspaces.md
mwest1066: Yep! This makes development easy, because you can run just one of them locally. For deployment we can choose whether to run them in separate autoscaling pools or combined in various combinations.

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,73 @@
...
+3. Host servers: run the actual containers.
+
+These three components are implemented within the main PL executable, but for deployment we can run different fleets of servers that use a `config` option to only turn on specific functionality.
+
+## How this works
+
+* When we want to launch a container, we bundle files, upload to S3, download to worker, spin up a container with files mounted to a known good location.
+* Run filesystem watcher on mounted directory.
+* When files change, we’ll do two things:
+    * We’ll upload the workspace state to S3
```
mwest1066, line 37:
```suggestion
    * We’ll upload the workspace state to S3.
```

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,149 @@
...
+#### A container has not been allocated on a host
+
+First, we'll see if we've already initialized S3 resources for this workspace in the past. If not, we'll create two files on S3:
+
+* The initial state of the workspace
+* The "working" state of the workspace that will be modified as students interact with the workspace
+
+These two will be identical initially and will diverege over time.
+
+Once we have these two S3 resources, we'll immediately respond with the "outer frame" HTML. At the same time, we'll allocate a host for this particular workspace and instruct it to begin loading the appropriate resources (Docker image, S3 resources, etc.).
```
tdy, line 142:
```suggestion
Once we have these two S3 resources, the main server will immediately respond with the "outer frame" HTML. At the same time, the host server will allocate a host for this particular workspace and instruct it to begin loading the appropriate resources (Docker image, S3 resources, etc.).
```

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,149 @@
...
+
+These three components are implemented within the main PL executable, but for deployment we can run different fleets of servers that use a `config` option to only turn on specific functionality.
+
+## How this works
+
+* When we want to launch a container, we bundle files, upload to S3, download to worker, spin up a container with files mounted to a known good location.
+* Run filesystem watcher on mounted directory.
+* When files change, we’ll do two things:
+    * We’ll upload the workspace state to S3.
+    * We’ll push the submission state into the database (somewhere, tbd) and store errors too?
```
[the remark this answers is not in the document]
mwest1066: "Submission state" here refers to the subset of files that will be graded. These will need to be specified by the instructor in the question's `info.json` file.

rfcs/003-workspaces.md
```diff
@@ -101,14 +101,31 @@ Course staff will declare workspace config per question via `workspace` in `info
...
 {
     "workspace": {
-        "image": "some-docker-image"
+        "image": "some-docker-image-name"
     }
 }
 ```
 
-The home directory in the workspace will be determined by the `workspace` directory inside a question directory. In the future, we'll add the ability to dynamically generate files via `server.py` and place them into the home directory. This is not part of the MVP.
+The home directory in the workspace will be determined by the `workspace_homedir` directory inside a question directory. In the future, we'll add the ability to dynamically generate files via `server.py` and place them into the home directory. This is not part of the MVP.
```
nwalters512, line 109: Why this change? We also don't use files/folders with underscores anywhere else in PL.

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,211 @@
...
+  * Update workspace state to `launching`
+* Commit the transaction, thereby releasing the lock
+* Send `change:state` websocket message to client
+
+#### Workspace is in state `stopped`
+
+* Respond to the request with the basic markup for the outer frame
+* [Do whole launch container thing from the above section here; let's write this out in more detail later]
+
+#### Workspace is in state `launching` or `running`k
```
tdy, line 177:
```suggestion
#### Workspace is in state `launching` or `running`
```

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,266 @@
...
+* Since PrairieLearn will be serving a bunch of different roles depending on context, PrairieLearn's `server.js` should be split up so that only code needed to serve a particular role is loaded. While we're refactoring, let's just make it better, do things async, etc.
+* There’s a distinction between workspace state and submission state - the former can include arbitrary files, the latter just includes whatever the question specifies.
+
+## Remaining work and open questions
+
+- [x] Define specifics of what happens on the client
+- [x] Define specifics of what happens on a host
+- [x] Procotol for communication between workspace hosts and web servers
+- [x] Protocol for user-initiated restart of a container
+- [ ] What does this look like when running locally?
```
tdy, line 262:
```suggestion
- [ ] What does this look like when running locally?
- [ ] How do we toggle the server mode (web vs host)?
```

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,267 @@
...
+    * When we create a variant, we (maybe) create a workspace session (if it’s enabled for that question). This is at this point just an entry in a database table somewhere.
+    * We render some kind of button to launch a workspace instance for that session.
+    * The user clicks on that button.
+    * We get a request for a particular workspace instance.
+    * We check the authorization cookies to verify that the requesting user matches the authorized user for this workspace.
+    * Routes:
+        * `/workspace/[uuid]` (referred to as `workspace_url` later in this document) - serves basic outer frame markup
+        * `/workspace/[uuid]/frame/*` - serves resources for outer frame
+        * `/workspace/[uuid]/heartbeat`
+        * `/workspace/[uuid]/container/*` - proxy `*` to inner frame
```
tdy, line 59:
```suggestion
        * `/workspace/<workspace_id>` (referred to as `workspace_url` later in this document) - serves basic outer frame markup
        * `/workspace/<workspace_id>/frame/*` - serves resources for outer frame
        * `/workspace/<workspace_id>/heartbeat`
        * `/workspace/<workspace_id>/container/*` - proxy `*` to inner frame
```

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,267 @@
...
+    * The user clicks on that button.
+    * We get a request for a particular workspace instance.
+    * We check the authorization cookies to verify that the requesting user matches the authorized user for this workspace.
+    * Routes:
+        * `/workspace/<workspace_id>` (referred to as `workspace_url` later in this document) - serves basic outer frame markup
+        * `/workspace/<workspace_id>/frame/*` - serves resources for outer frame
+        * `/workspace/<workspace_id>/heartbeat`
+        * `/workspace/<workspace_id>/container/*` - proxy `*` to inner frame
+* On the host:
+    * `/workspace/[uuid]/container/*` goes to the host that’s running this container.
```
tdy, line 61:
```suggestion
    * `/workspace/<workspace_id>/container/*` goes to the host that’s running this container.
```

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,267 @@
...
+  * Pull the Docker container
+  * Pull the workspace state from S3
+  * Place workspace state into a directory on disk
+  * Starting a container with the workspace state mounted as the home directory
+  * Change workspace state to `running`
+  * Emit `change:state` websocket event
+* `POST /workspace/<workspace_id>/stop`: Tears down any resources associated with this container.
+* `GET /workspace/<workspace_id>/graded_files`: Responds with a tarball including the set of graded files for this question.
+
+Workspace hosts will also respond to `/workspace/[uuid]/container/*`, which mirrors the route on the PrairieLearn web server. When a workspace host receives a request to that path, it forwards `*` to the workspace container for that UUID.
```
tdy, line 245:
```suggestion
Workspace hosts will also respond to `/workspace/<workspace_id>/container/*`, which mirrors the route on the PrairieLearn web server. When a workspace host receives a request to that path, it forwards `*` to the workspace container for that ID.
```

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,267 @@
...
+  * `s3_root_key`: The root "path" within the S3 bucket
+  * `workspace_host_id` (nullable): The ID of the host that this workspace container is running on, if any
+  * `state`: reflects the "state" of this workspace; one of the following:
+    * `uninitialized`: no resources have been created for this workspace yet; can transition to `initializing`
+    * `stopped`: S3 resources for the workspace exist, but it is not running on a particular machine; can transition to `launching`
+    * `launching`: we are allocating a host for this workspace and starting the appropriate container on that host; can transition to `running` or `stopped` (if launching fails)
+    * `running`: the container for this workspace is running; can transition to `stopped`
+* `workspace_logs`: notable events/messages assocaiated with a particular workspace (state transitions, errors, explicit restarts, etc.)
+  * `workspace_id`: ID of the associated workspace
+  * `timestamp`: (workshop name for consistency with rest of PL): timestamp of the event
```
tdy, line 97:
```suggestion
  * `date`: timestamp of the event
```
I'm seeing other tables use `date` instead.

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,267 @@
...
+  * `message`: string message
+  * `level`: the level of this particular log
+
+### Questions
+
+Course staff will declare workspace config per question via `workspace` in `info.json`. To begin, the only option will be an image:
+
+```json
+{
+    "workspace": {
```
tdy, line 107:
```suggestion
    "workspaceOptions": {
```
Should we mirror `externalGradingOptions`?

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,267 @@
...
+        * The user hasn’t saved for X amount of time.
+
+Need to make sure that cookies are inaccessible to client-side code (https://github.com/PrairieLearn/PrairieLearn/issues/2503) and on the server (we need to configure our proxy to strip out at least the Cookie header, if not more things).
+
+## Design
+
+### Database
+
+* `questions`
+  * Add a new `workspace_image` column
```
tdy, line 79:
```suggestion
  * Add a new `workspace_image` column
  * Add a new `workspace_gradedFiles` column
```

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,267 @@
...
+    * `running`: the container for this workspace is running; can transition to `stopped`
+* `workspace_logs`: notable events/messages assocaiated with a particular workspace (state transitions, errors, explicit restarts, etc.)
+  * `workspace_id`: ID of the associated workspace
+  * `date`: timestamp of the event
+  * `message`: string message
+  * `level`: the level of this particular log
+
+### Questions
+
+Course staff will declare workspace config per question via `workspace` in `info.json`. To begin, the only option will be an image:
```
tdy, line 103:
```suggestion
Course staff will declare workspace config per question via `workspace` in `info.json`. To begin, the only options will be an Docker image and a list of files to be graded:
```

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,267 @@
...
+    |   +-- correct_answer.c
+    |   `-- test_run.py
+    |
+    `-- workspace
+        +-- .bashrc
+        +-- starter_code.h
+        `-- starter_code.c
+```
+
+The workspace image name will need to be synced to the `questions` table via the usual syncing code. We should use a new `workspace_image` column.
```
tdy, line 136:
```suggestion
The workspace options will need to be synced to the `questions` table via the usual syncing code.
```

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,268 @@
...
+
+Need to make sure that cookies are inaccessible to client-side code (https://github.com/PrairieLearn/PrairieLearn/issues/2503) and on the server (we need to configure our proxy to strip out at least the Cookie header, if not more things).
+
+## Design
+
+### Database
+
+* `questions`
+  * Add a new `workspace_image` column
+  * Add a new `workspace_gradedFiles` column
```
tdy, line 80:
```suggestion
  * Add a new `workspace_graded_files` column
```

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,277 @@
...
+  * `message`: string message
+  * `level`: the level of this particular log
+
+### Questions
+
+Course staff will declare workspace config per question via `workspace` in `info.json`. To begin, the only options will be an Docker image and a list of files to be graded:
+
+```json
+{
+    "workspace": {
```
tdy, line 110:
```suggestion
    "workspaceOptions": {
```

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,277 @@
...
+    * `running`: the container for this workspace is running; can transition to `stopped`
+* `workspace_logs`: notable events/messages assocaiated with a particular workspace (state transitions, errors, explicit restarts, etc.)
+  * `workspace_id`: ID of the associated workspace
+  * `date`: timestamp of the event
+  * `message`: string message
+  * `level`: the level of this particular log
+
+### Questions
+
+Course staff will declare workspace config per question via `workspace` in `info.json`. To begin, the only options will be an Docker image and a list of files to be graded:
```
tdy, line 106:
```suggestion
Course staff will declare workspace config per question via `workspaceOptions` in `info.json`. To begin, the only options will be an Docker image and a list of files to be graded:
```

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,279 @@
...
+
+* `POST /workspace/<workspace_id>/launch`: Begins the asynchronous process of launching a workspace container. This entails:
+  * Pull the Docker container
+  * Pull the workspace state from S3
+  * Place workspace state into a directory on disk
+  * Starting a container with the workspace state mounted as the home directory
+  * Change workspace state to `running`
+  * Emit `change:state` websocket event
+* `POST /workspace/<workspace_id>/stop`: Tears down any resources associated with this container.
+* `GET /workspace/<workspace_id>/graded_files`: Responds with a tarball including the set of graded files for this question.
```
polohan, line 255: Or we could use WebSocket to interact with the workspace host directly, instead of POST/GET to the main server than POST/GET to the workspace host?

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,279 @@
...
+Course staff will declare workspace config per question via `workspaceOptions` in `info.json`. To begin, the only options will be a Docker image, port number, and list of files to be graded:
+
+```json
+{
+    "workspaceOptions": {
+        "image": "some-docker-image-name",
+        "port": 15000,
+        "gradedFiles": [
+            "animal.h",
+            "animal.cpp"
```
tdy, line 116:
```suggestion
            "starter_code.h",
            "starter_code.c"
```

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,279 @@
...
+
+The workspace options will need to be synced to the `questions` table via the usual syncing code.
+
+### Student-facing question interface
+
+> What happens when we render a question with an associated workspace?
+
+When a new variant of a question is created, the main server will create a corresponding workspace in the database associated with that particular variant. This database entry will contain a unique hash/id/something. However, we're not going to actually provision any containers, etc. for this workspace just yet. The `state` of the new workspace row will be `uninitialized`.
+
+We'll introduce a new `<pl-workspace>` element that renders (to start with) a "Launch workspace" button. We should introduce a new `workspace_url` to `data.options`, and this element (or potentially other elements) can use this to render a button. `workspace_url` will be something of the form `/pl/[garbage]/workspace/[workspace_id]`, where `[garbage]` corresponds to the different places that questions can be accessed from (instructor question, student variant, maybe others?).
```
tdy, line 149:
```suggestion
We'll introduce a new `<pl-workspace>` element that renders (to start with) a "Launch workspace" button. We should introduce a new `workspace_url` to `data.options`, and this element (or potentially other elements) can use this to render a button. `workspace_url` will be something of the form `/workspace/[workspace_id]`.
```

rfcs/003-workspaces.md
```diff
@@ -0,0 +1,279 @@
...
+
+`workspace_url` pages will be served by the main PrairieLearn server (someday, we could split this into a separate autoscaled component).
+
+When the main server gets a request to this url, we'll first check if we have an existing instance of a workspace by checking the `state` column of the appropriate workspace. There will be three cases here:
+
+#### Workspace is in state `uninitialized`
+
+* Respond to the request with the basic markup for the outer frame
+* Begin a transaction
+  * Obtain a lock on the row in the `workspaces` table
```
tdy, line 172: How do we lock the row? I assume it involves one of these, but not sure exactly:
* wrap the query in a `sqldb.beginTransaction`
* wrap with `namedLocks`
* create+call a `workspace_lock` sproc

tdy:
All unresolved items are now cross-referenced with their standalone issues.

PR #2507, opened 2020-05-15
Allow course-level extensions for elements

### LINKED ISSUE #3123: Element caching does not work with chunks
I believe this will lead to a bug where a PL server won't have up-to-date course element files if the elements were previously cached. In other words, a server will *never* get new course chunks once it has an entry in `courseElementsCache`. Once we throw multiple servers into the mix, and especially with syncs happening on different machines than actually serve questions, the course element cache won't work in its current form. There's a couple potential solutions here, I think:

* Key the cache on course ID *and* Git hash - this will invalidate the cache when a sync occurs, because a new hash will be associated with the course.
* Eliminate the cache entirely. Very robust, less efficient. a.k.a. "caching is hard, let's not."

The former is my preferred strategy. Anyone have any other thoughts?

_Originally posted by @nwalters512 in https://github.com/PrairieLearn/PrairieLearn/pull/2891#discussion_r487492492_

nicknytko:
Allows custom extensions for elements to be defined in a course's `elementExtensions/` folder.  This folder is structured as `elementExtensions/{element name}/{extension name}`.

Extensions may require CSS and JS dependencies, these will get autoloaded whenever the parent element is rendered on the page.

Extensions may also define custom Python functions, the parent element may manually load these with `prairielearn.load_extension()` (for one specific extension) and `prairielearn.load_all_extensions()` (for all extensions).

When elements are rendered, the `extensions` key is snuck into `data` and contains all extensions that may be loaded by that specific element.

Creates an example element that loads extensions and renders them to HTML, as seen in the question `demoExtendableElement`:

image

Motivation: have an elegant way to add course-level logic to the `pl-drawing` element, but in a general way that any other element can use.

Resolves #3123 

This also adds "elementExtension" chunks that are generated and uploaded to S3.  I can confirm them being generated and loaded locally on docker:

```sh
[root@df3bd56627ba course-1]# pwd
/tmp/pl-chunked-courses/course-1
[root@df3bd56627ba course-1]# ls elementExtensions/
extendable-element  pl-graph
[root@df3bd56627ba course-1]# 
```


question-servers/freeform.js
```diff
@@ -149,6 +151,101 @@ module.exports = {
...
+                /* Get extensions from each element folder */
+                async.map(elementFolders, (element) => {
+                    fs.readdir(path.join(sourceDir, element), (err, extensions) => {
+                        if (ERR(err, callback)) return;
+                        return callback(null, extensions.map(ext => [element, ext]));
+                    });
+                }, callback);
+            },
+            (extensions, callback) => {
+                /* Filter out non-directories */
```
mwest1066, line 176: Do we need to do this, given that we are going to skip cases where we can't open `info.json` below? I'm thinking of the way we handle this when loading the questions directory, for example.

question-servers/freeform.js
```diff
@@ -149,6 +151,101 @@ module.exports = {
...
+        ], (err, extensions) => {
+            if (ERR(err, callback)) return;
+            return callback(null, extensions);
+        });
+    },
+
+    loadExtensionsForCourse(course, callback) {
+        if (courseExtensionsCache[course.id] !== undefined) {
+            return callback(null, courseExtensionsCache[course.id]);
+        }
```
mwest1066, line 232: We should key this caching by `course.commit_hash`, so that we automatically reload whenever the `commit_hash` changes. At the moment we need to explicitly called `reloadExtensionsForCourse` in `syncFromDisk.js`, which is bad because it only happens on the server instance that actually did the sync. Alternatively, we could send a message via redis to tell the other servers to reload, but it seems simpler and more robust to just key on `commit_hash`.

Just to be clear, this is not a new problem with element extensions. The existing element cache also has this problem, as far as I'm aware.

question-servers/freeform.js
```diff
@@ -748,8 +871,9 @@ module.exports = {
                 params: _.get(variant, 'params', {}),
                 correct_answers: _.get(variant, 'true_answer', {}),
                 variant_seed: parseInt(variant.variant_seed, 36),
-                options: _.get(variant, 'options', {}),
+                options: _.get(variant, 'options',{}),
```
mwest1066, line 874: I assume this was an unintentional single-space deletion 😄

question-servers/freeform.js
```diff
@@ -812,8 +936,7 @@ module.exports = {
             data.options.base_url = locals.baseUrl;
 
             // Put key paths in data.options
-            data.options.question_path = context.question_dir;
-            data.options.client_files_question_path = path.join(context.question_dir, 'clientFilesQuestion');
+            _.extend(data.options, module.exports.getContextOptions(context));
```
mwest1066, line 939: nice!

question-servers/freeform.js
```diff
@@ -492,6 +597,11 @@ module.exports = {
                 }
 
                 const elementFile = module.exports.getElementController(elementName, context);
+                /* Populate the extensions used by this element */
+                data.extensions = [];
+                if (_.has(context.course_element_extensions, elementName)) {
+                    data.extensions = context.course_element_extensions[elementName];
+                }
```
mwest1066, line 604: I like this.

question-servers/freeform.js
```diff
@@ -501,6 +611,8 @@ module.exports = {
                         courseIssues.push(courseIssue);
                         return callback(courseIssue);
                     }
+                    delete data.extensions;
+                    delete ret_val.extensions;
```
mwest1066, line 615: This stops the extensions being passed along to `server.py`?

nicknytko: Yeah, I'm thinking only the element code should be able to access it?

mwest1066: Yep, makes sense.

question-servers/freeform.js
nicknytko: In that case, would we have to worry about periodically emptying out or pruning the caches?  Maybe this isn't that big of a deal(?) but I could see all the old commit hash cache entries wasting memory here.

question-servers/freeform.js
nicknytko: Hmm, you're right that this is probably safe to remove.  I'll try removing this from `loadElements()` as well and see if anything breaks.

question-servers/freeform.js
mwest1066: I guess one simple strategy would be to set `courseExtensionsCache[course.id] = {commit_hash: course.commit_hash, data: actualCachedData}`. Then rather than simply checking whether `courseExtensionsCache[course.id] !== undefined`, we would also check whether the `commit_hash` matched to figure out whether we need to use the cache or update it. When we do update it, we'd overwrite the old one so we don't get growth.

question-servers/freeform.js
nicknytko: It was unnecessary, removed.

question-servers/freeformPythonLib/prairielearn.py
```diff
@@ -1084,3 +1088,77 @@ def escape_invalid_string(string):
...
+
+        def wrapped_function(*args, **kwargs):
+            old_wd = os.getcwd()
+            os.chdir(ext_info['directory'])
+            ret_val = f(*args, **kwargs)
+            os.chdir(old_wd)
+            return ret_val
+        return wrapped_function
+
+    # Since extensions may have multiple scripts, load all of them
```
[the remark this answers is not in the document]
nicknytko: I was thinking you could break up your extension into multiple scripts for organization, but I guess you could just have one "controller" script that imports all the rest.  Don't really have a strong opinion here either way :shrug:

question-servers/freeform.js
```diff
@@ -130,25 +122,112 @@ module.exports = {
...
-    reloadElementsForCourse(course, callback) {
-        module.exports.loadElements(path.join(course.path, 'elements'), 'course', (err, elements) => {
+    /**
+     * Takes a directory containing an extension directory and returns a new
+     * object mapping element names to each extension, which itself an object
+     * that contains relevant extension scripts and styles. 
+     * @param  {String}   sourceDir Absolute path to the directory of extensions
+     * @param  {Function} callback  Called with any errors and the results
+     */
+    loadExtensions(sourceDir, callback) {
```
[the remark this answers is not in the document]
nicknytko: I'll try rewriting this tomorrow, haven't used the `async/await` syntax too extensively so no promises though  :crossed_fingers:

question-servers/freeformPythonLib/prairielearn.py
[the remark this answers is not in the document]
nwalters512: I think there's a lot of value in consistency, so I'd aim for as much consistency as possible with elements themselves. A single top-level "controller" sounds good to me!

exampleCourse/courseInstances/Sp15/assessments/hw04-demoQuestions/infoAssessment.json
```diff
@@ -48,7 +48,8 @@
                 {"id": "demoRandomPlot",                      "points": 1, "maxPoints": 3},
                 {"id": "demoRandomDataFrame",                 "points": 2, "maxPoints": 6},
                 {"id": "demoAnsiOutput",                      "points": 2, "maxPoints": 4},
-                {"id": "demoOverlayTriangle",                 "points": 2, "maxPoints": 4}
+                {"id": "demoOverlayTriangle",                 "points": 2, "maxPoints": 4},
+                {"id": "demoExtendableElement",               "points": 2, "maxPoints": 4}
```
coatless, line 52: Would it be possible to show how to extend an existing `pl-*` element? The `<extendable-element>` demo is useful, but feels too detached from practical usage.

docs/elementExtensions.md
```diff
@@ -0,0 +1,73 @@
...
+```python
+import prairielearn as pl
+
+def render(element_html, data):
+    extension = pl.load_element_extensions(data, 'extension_name')
+    contents = extension.my_cool_function()
+    return contents
+```
+
+This small example above will render `"hello world!"` to the question page.  Note that no order is guaranteed when loading all extensions with `load_all_extensions_for_element()`.
```
coatless, line 52: I'm not sure allowing for a random order is a good thing... Is the reason why this is the case because `prairielearn` is using a version of `Python` before 3.6 so `dict` structures are missing order? If so, could we consider adding `OrdDict` from `collections` ala `from collections import OrderedDict`?

Details: https://stackoverflow.com/a/39537308/1345455

docs/elementExtensions.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Extension Dependencies
+
+Like how questions and elements may require client-side assets, extensions may also require client-side JavaScript and CSS.  The different properties are summarized here:
```
coatless, line 56:
```suggestion
Similar to how questions and elements may require client-side assets, extensions may also require client-side JavaScript and CSS.  The different properties are summarized here:
```

docs/elementExtensions.md
```diff
@@ -0,0 +1,73 @@
...
+`elementStyles` | The styles required by this element relative to the extension's directory,`[course directory]/elementExtensions/element-name/extension-name`.
+`elementScripts` | The scripts required by this element relative to the extension's directory, `[course directory]/elementExtensions/element-name/extension-name`.
+`clientFilesCourseStyles` | The styles required by this extension relative to `[course directory]/clientFilesCourse`.
+`clientFilesCourseScripts` | The scripts required by this extension relative to `[course directory]/clientFilesCourse`.
+
+Note that all client-side extension assets are always loaded, regardless of whether their Python controller was loaded or not.
+
+### Other Client Files
+
+Other files available to the client may also be loaded, such as images or any downloadable content.  These client files should be placed in `clientFilesExtension` in the extension directory, and the full URL to that folder is given to the host extension in `data['options']['client_files_extensions_url'][extension_name]`.  If this path is needed by the extension itself, it may be passed as an argument to a defined extension function.
```
coatless, line 73: I think the documentation for developing a question element should include a reference back here or at the very least a note that extension data exists in `options`

https://github.com/PrairieLearn/PrairieLearn/blob/master/docs/question.md#question-serverpy

docs/elementExtensions.md
```diff
@@ -0,0 +1,73 @@
+# Element extensions developer guide
+
+Extensions are a way to add custom logic and extra complexity to already existing elements.  Each element has the ability to load relevant extensions and as such, extensions should be custom tailored to one specific element.
```
coatless, line 3: I think this is what I thought PR https://github.com/PrairieLearn/PrairieLearn/pull/2384#discussion_r411444013 was?

docs/elementExtensions.md
```diff
@@ -0,0 +1,73 @@
+# Element extensions developer guide
```
coatless, line 1: Add to:

https://github.com/PrairieLearn/PrairieLearn/blob/2260c3d99f0cb73f01463228b9a10f65b71804c2/mkdocs.yml#L37

tests/testCourseElementExtension.js
```diff
@@ -0,0 +1,123 @@
...
+const path = require('path');
+const freeform = require('../question-servers/freeform.js');
+const { promisify } = require('util');
+
+const helperServer = require('./helperServer');
+const helperClient = require('./helperClient');
+
+describe('Course element extensions', function() {
+    this.timeout(60000);
+    const exampleCourseDir = path.join(__dirname, '..', 'exampleCourse');
```
coatless, line 18: I'm worried that we're re-growing the "Example Course" to a "Test Course" status. Maybe guidelines should be drawn up on this? (Not a show stopper here... More thinking out loud.)

pages/elementExtensionFiles/elementExtensionFiles.js
```diff
@@ -0,0 +1,30 @@
...
+const express = require('express');
+const router = express.Router();
+const _ = require('lodash');
+
+/**
+ * Serves scripts and styles for element extensions. Only serves .js and .css files, or any
+ * static files from an extension's "clientFilesExtension" directory. 
+ */
+
+const EXTENSION_WHITELIST = ['.js', '.css'];
```
coatless, line 11: Maybe:
```suggestion
const FILE_TYPE_EXTENSION_WHITELIST = ['.js', '.css'];
```

pages/elementExtensionFiles/elementExtensionFiles.js
```diff
@@ -0,0 +1,30 @@
...
+const CLIENT_FOLDER = 'clientFilesExtension';
+
+router.get('/*', function(req, res, next) {
+    const filename = req.params[0];
+    let pathSpl = path.normalize(filename).split('/');
+    const valid = pathSpl[2] == CLIENT_FOLDER ||
+          _.some(EXTENSION_WHITELIST, (extension) => filename.endsWith(extension));
```
coatless, line 18:
```suggestion
          _.some(FILE_TYPE_EXTENSION_WHITELIST, (extension) => filename.endsWith(extension));
```

exampleCourse/questions/demoExtendableElement/question.html
```diff
@@ -0,0 +1,12 @@
+<pl-question-panel>
+
+<p>
+    This question is a demo of the element extension capabilities in PrairieLearn.  This page has one element on it: <code class="user-output">extendable-element</code>, which loads each extension that it has and renders it to a separate card.  You can see the different functionalities of extensions below.
+</p>
+<p>
+    Note that no load order is guaranteed for extensions, so the cards below may appear in differing orders on page reload.
```
coatless, line 7: See note in the doc overview.

docs/elementExtensions.md
nicknytko: Ah, let me clarify a little bit.  Course elements allow someone to replace a whole element with a local copy in their course, which allows an easy way to make changes but then you end up with duplicate copies. 
What I'm proposing here is allowing elements to almost create their own developer interfaces that courses can make extensions for, I was initially thinking this could be used for the `pl-drawing` element if people wanted to make custom drawing objects, though this can be used by any large elements in the future as needed.

exampleCourse/courseInstances/Sp15/assessments/hw04-demoQuestions/infoAssessment.json
nicknytko: See below comment, but existing elements will need to add support for extensions to be able to use them.

docs/elementExtensions.md
coatless: @nicknytko excellent. We're on the same page then! 

> but existing elements will need to add support for extensions to be able to use them.

In this case, I think prior to merging the PR, there should be an extension interface applied to at least one base element, `pl-drawing` would suffice in this case.

docs/elementExtensions.md
nicknytko: Sure, though I'm not sure the order really matters as long as it's consistent?  I can sort the extensions in alphabetical order so it's deterministic.

docs/elementExtensions.md
nicknytko: Agreed, currently working on that for `pl-drawing` in a different branch but I think that may involve quite a bit of refactoring :slightly_smiling_face: 

How about the `pl-python-variable` element?  I think one could argue for the use case of adding new renderable types to that element.

nicknytko: Discussed with @mfsilva22, added an example extension to the `pl-graph` elem

nicknytko:
@mwest1066 review bump 😀

mwest1066:
> @mwest1066 review bump 😀

Yep, sorry, I know about it. I just haven't had the mental fortitude today to tackle something so serious 😄 -- but soon! Keep hassling me about it.

question-servers/freeform.js
```diff
@@ -78,16 +81,6 @@ module.exports = {
...
-            (files, callback) => {
-                // Filter out any non-directories
-                async.filter(files, (file, callback) => fs.lstat(path.join(sourceDir, file), (err, stats) => {
-                    if (ERR(err, callback)) return;
-                    callback(null, stats.isDirectory());
-                }), (err, results) => {
-                    if (ERR(err, callback)) return;
-                    return callback(null, results);
-                });
-            },
```
mwest1066, line 90: Why don't we need this anymore? I worry that a `.DS_Store` file or similar will confuse things?

mwest1066:
I'd also like to say that I very much appreciate the image chosen for the demo question 🐈

question-servers/freeform.js
nicknytko: Good point, I'm putting the checks back in.

mwest1066 (approved):

PR #2514, opened 2020-05-21
Per-course-instance declarations for user groups 

astrosticks:
A work-in-progress implementing user groups, which are per-course-instance, staff-defined collections of users which can be referenced by name instead of individual users in assessment access rules. 

The database schema that we use for groups is borrowed from @frankyhao's [fork](https://github.com/frankyhao/PrairieLearn/commits/master) with minor modifications.

**Example: courseInstanceInfo.json**  
```json
...
    "groups": {
        "dres-2x-exams": [
            "mwest@illinois.edu",
            "zilles@illinois.edu"
        ],
        "dres-1.5x-exams": [
            "mussulma@illinois.edu",
            "nwalter2@illinois.edu"
        ]
    },
...
```
---
### Progress

- [x] Populate `groups` table from data in `courseInstanceInfo.json`
    - [x] Create new groups and user memberships on sync
    - [x] Delete unused groups and memberships on sync (if no longer present in courseInstanceInfo.json), ignoring groups with a `group_config` set (for compatibility with the @frankyhao fork)


migrations/800_groups__add.sql
```diff
@@ -0,0 +1,30 @@
...
+);
+
+CREATE TABLE IF NOT EXISTS groups (
+    id BIGSERIAL PRIMARY KEY,
+    course_instance_id BIGINT NOT NULL REFERENCES course_instances(id) ON DELETE CASCADE ON UPDATE CASCADE,
+    name TEXT, -- visible name of the group
+    group_config_id BIGINT REFERENCES group_configs(id) ON DELETE CASCADE ON UPDATE CASCADE,
+    date timestamp with time zone DEFAULT CURRENT_TIMESTAMP,
+    deleted_at timestamp with time zone,
+    UNIQUE (course_instance_id, name)
```
trombonekenny, line 22: I used a different approach in the other PR for this. Might want to keep that in sync (and since your code is the one actively using it, use your stuff to try out that PR's constraints?)

sprocs/index.js
```diff
@@ -29,134 +29,135 @@ module.exports = {
     _initWithLock(callback) {
         logger.verbose('Starting DB stored procedure initialization');
         async.eachSeries([
-            'array_dot.sql',
```
trombonekenny, line 32: Don't alphabetize this list. The load order is important for sprocs that reference other sprocs. (If you were to load PL on an empty database doing the loads, you'd see it crashes loading sprocs.) I'd suggest reverting your changes and adding your sprocs at the end of the file.

These changes are also causing a merge conflict with master, which is breaking the CI testing. You'll want to make sure you can merge `master` into your branch without challenges?

sprocs/sync_groups.sql
```diff
@@ -0,0 +1,95 @@
+-- Accepts a course instance ID and a JSON object specifying the, as an object of
```
trombonekenny, line 1: I'm going to hold off reviewing this for now. We'll want to compare it against how the other sync functions are written, and make sure we're not in any race or locking conditions with multiple course instances getting sync'd. 

Also, from our slack chats, we need to decide on a strategy for group deletion and what that looks like. It kinda relates to the group schema and how we're using `group_configs`. I'm thinking we probably want to make a group_config entry for "loaded from infoCourseInstance.json" and use that relationship to control what this process does for its groups.

astrosticks: I'm still in favor of "loaded from infoCourseInstance.json" as being implicitly associated with a null value in group_configs, since configs are built specifically for being associated with assessments. If we don't go with that, I'd still rather use an enum field in groups rather than an arbitrary text field inside of a config.

coatless:
Tied to: https://github.com/PrairieLearn/PrairieLearn/pull/2518

astrosticks:
Will wait for #2518 to merge and then fix build errors

astrosticks:
closing this as there's been no feedback or updates for quite some time

PR #2516, opened 2020-05-21
Change accuracy_critical default in Python autograder

nicknytko:


docs/python-grader/sphinx-docs.md
```diff
@@ -75,7 +75,7 @@ Check that a student list has correct length with respect to a reference list.
 * `report_failure`: If true, feedback will be given on failure.
 
 
-#### classmethod check_numpy_array_allclose(name, ref, data, accuracy_critical=True, rtol=1e-05, atol=1e-08, report_success=True, report_failure=True)
+#### classmethod check_numpy_array_allclose(name, ref, data, accuracy_critical=False, rtol=1e-05, atol=1e-08, report_success=True, report_failure=True)
```
mwest1066, line 78: Any idea why Sphinx is only listing the default args for this one function? `check_list()` above doesn't have them?

nicknytko: Yeah I noticed that too, not sure tbh.  I can investigate in a bit.

PR #2521, opened 2020-05-26
Adding check for /grade to avoid problems

echuber2:
~-p creates intermediate subdirectories if they are missing and also suppresses errors if existing. This may or may not help with the Windows deployment issue~

### DIFF
--- a/graders/python/python_autograder/run.sh
+++ b/graders/python/python_autograder/run.sh
@@ -7,9 +7,7 @@
 # the autograder directory
 AG_DIR='/python_autograder'
 
-if [ -d /grade ] ; then
-        echo "Found /grade"
-else
+if [ ! -d /grade ] ; then
         echo "ERROR: /grade not found! Mounting may have failed."
         exit 1
 fi
--- a/graders/python/python_autograder/run.sh
+++ b/graders/python/python_autograder/run.sh
@@ -7,9 +7,9 @@
 # the autograder directory
 AG_DIR='/python_autograder'
 
-if [ ! -d /grade ] ; then
-        echo "ERROR: /grade not found! Mounting may have failed."
-        exit 1
+if [[ ! -d /grade ]]; then
+  echo "ERROR: /grade not found! Mounting may have failed."
+  exit 1
 fi
 
 # the parent directory containing everything about this grading job
--- a/graders/python/python_autograder/run.sh
+++ b/graders/python/python_autograder/run.sh
@@ -18,8 +18,8 @@ OUT_DIR=$JOB_DIR'/results'
 export MERGE_DIR=$JOB_DIR'/run'
 
 # now set up the stuff so that our run.sh can work
-mkdir $MERGE_DIR
-mkdir $OUT_DIR
+mkdir -p $MERGE_DIR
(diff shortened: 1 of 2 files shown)


echuber2:
At Matt's advice, I'm going to edit this PR to just check that `/grade` was actually pre-existing when the bash script launches, because we shouldn't have to be creating it recursively. The underlying Windows problem must be related to mounting then.

graders/python/python_autograder/run.sh
```diff
@@ -7,6 +7,11 @@
 # the autograder directory
 AG_DIR='/python_autograder'
 
+if [ ! -d /grade ] ; then
+        echo "ERROR: /grade not found! Mounting may have failed."
+        exit 1
+fi
```
tdy, line 13:
```suggestion
if [[ ! -d /grade ]] ; then
  echo "ERROR: /grade not found! Mounting may have failed."
  exit 1
fi
```

mwest1066:
Why do we have the dockerfile changes here? It looks like they will add more layers (slower downloads) to the image?

echuber2:
> Why do we have the dockerfile changes here? It looks like they will add more layers (slower downloads) to the image?

@mwest1066 I think it's just the opposite; I intentionally moved the yum and pip commands *after* the updates that are particular to the autograder scripts themselves, as much as possible, so that small tweaks to the autograder script do not cause the yum/pip layers to be rebuilt too. @nicknytko do you agree? Otherwise we can scrap those changes.

echuber2:
> It looks like we'll be adding more layers because we're splitting out the pip install to another `RUN` call? I'd be in favor of keeping the yum and pip install commands in one `RUN`, and then just `COPY`ing the autograder files afterwards if that wont add too much overhead in terms of docker layers.

OK, how does it look now? My thinking before had been: pip requirements must come before pip install, but yum doesn't rely on it, so putting the requirements copy in after yum and before pip would only trigger a rebuild of the pip layer if the requirements file actually changed. I may be wrong about that, though.

nicknytko:
Sorry, I've snuck this change into #2579: https://github.com/PrairieLearn/PrairieLearn/commit/92d3c937d3e9d02bb705f4c04b5c784340524093#diff-9f2122a82dc11b72dbde9ac41e504e9d

I think we can close out this PR now?

mwest1066:
Subsumed into #2579

echuber2:
@mwest1066 @nicknytko It looks like the edit to the dockerfile was adopted but not the extra check in `run.sh`? Do we still want that or no? I'll reopen but you can close again if need be.

nicknytko:
@echuber2 I feel fine with merging the check in the `run.sh` if you undo the changes to the grader dockerfile.

echuber2:
@nicknytko It was almost the same as the version you have in production now. I've just updated the copy in this branch to match what master has.

echuber2:
Is it better to set it to the state the file had previously in this same branch?

nicknytko:
> Is it better to set it to the state the file had previously in this same branch?

Yes please :slightly_smiling_face:

echuber2:
@nicknytko Done.

PR #2523, opened 2020-05-27
Add .sql and remove .ejs from nodemon watch list

tdy:
cc @astrosticks @trombonekenny


astrosticks:
Hmm. Interesting that this didn't pass even though it's up-to-date with master. Is nodemon somehow enabled when the tests are being run?

tdy:
I think CI was just acting up.

PR #2566, opened 2020-06-02
Add "partial_scores" column to manual grading submissions CSV

### LINKED ISSUE #2565: "partial_score" attribute download
We would like to be able to see the partial_scores attribute that is visible to the grade() file within submissions_for_manual_grading.csv . I believe this should take the same form as the "submitted_answers" attribute.

nicknytko:
Resolves #2565 

The `submissions_for_manual_grading.csv` file is now structured like:

image

### DIFF
--- a/pages/instructorAssessmentDownloads/instructorAssessmentDownloads.js
+++ b/pages/instructorAssessmentDownloads/instructorAssessmentDownloads.js
@@ -185,6 +185,7 @@ router.get('/:filename', function(req, res, next) {
                 ['params', 'params'],
                 ['true_answer', 'true_answer'],
                 ['submitted_answer', 'submitted_answer'],
+                ['partial_scores', 'partial_scores'],
                 ['score_perc', null],
                 ['feedback', null],
             ];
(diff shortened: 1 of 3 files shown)


pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql
```diff
@@ -82,7 +82,8 @@ SELECT DISTINCT ON (ai.id, q.qid)
     s.id AS submission_id,
     v.params,
     v.true_answer,
-    (s.submitted_answer - '_files') AS submitted_answer
+    (s.submitted_answer - '_files') AS submitted_answer,
+    s.partial_scores
```
coatless, line 86: Should this also be added to the API for the assessment export?

cheeren:
is partial_score.name.score writeable? (and for that matter, partial_score.name.feedback...) or is the intention that we would use the partial info as INPUT, and we would only write to score_perc?

pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql
nicknytko: Yeah, might as well.  I'll add that.

pages/instructorAssessmentDownloads/instructorAssessmentDownloads.js
```diff
@@ -185,6 +185,7 @@ router.get('/:filename', function(req, res, next) {
                 ['params', 'params'],
                 ['true_answer', 'true_answer'],
                 ['submitted_answer', 'submitted_answer'],
+                ['partial_scores', 'partial_scores'],
```
mwest1066, line 188: Can we also add this to all the other `submissions` CSVs?

PR #2579, opened 2020-06-10
Switch Python autograder to use file preview element

### LINKED ISSUE #2577: Python autograder doesn't show submitted code if question is saved
The python autograder displays its source code using the `pl-external-grader-results` element, which doesn't show up during grading and when a question is only saved.

We should modify this to not display the source as a "test case", and instead use the `pl-file-preview` element.

nicknytko:
Fixes #2577 

Student submissions are no longer displayed as a test case, but instead use the `pl-file-preview` element.  This has the benefit that submissions are now displayed if a student clicks "save" instead of "save and grade".

Screenshot_2020-06-10 PrairieLearn

Screenshot_2020-06-10 PrairieLearn(1)

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -24,6 +24,7 @@ pages:
   - 'PrairieDraw graphics': 'PrairieDraw.md'
   - 'Element for drawing': 'pl-drawing/index.md'
   - 'External grading': 'externalGrading.md'
+  - 'Python autograder': 'python-grader/index.md'
   - 'Manual grading': 'manualGrading.md'
   - 'What to tell students': 'whatToTellStudents.md'
   - 'How to use PL to learn student names': 'howToLearnNames.md'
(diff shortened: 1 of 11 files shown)


mkdocs.yml
```diff
@@ -24,6 +24,7 @@ pages:
   - 'PrairieDraw graphics': 'PrairieDraw.md'
   - 'Element for drawing': 'pl-drawing/index.md'
   - 'External grading': 'externalGrading.md'
+  - 'Python autograder': 'python-grader/index.md'
```
coatless, line 27: Whoops, that should have gone in earlier. C'est la vie.

nicknytko:
@mwest1066 @coatless 
Now that `pygments` is included in `plbase`, do we want to add syntax highlighting to the file preview element itself?  I previously had it inside the autograder, but this could be a nice enhancement for everyone using the preview elem.

PR #2594, opened 2020-06-22
Correcting CR-LF line endings in mersenne, removing local copy of socket.io.js

echuber2:
Fix `CR LF` line endings in `socket.io.js` and `mersenne.js` that were causing cross-platform `git` issues sporadically (Eric Huber).

The `socket.io.js` file had a mixture of `CR LF` (Windows) and `LF` (Linux) line endings. The `mersenne.js` file had `CR LF`. I think the usual on this repo is to have `LF` line endings, and then local git clients can perform changes transparently as appropriate when pulling or pushing (should be done automatically). So, this PR edits these two files to strictly have `LF` line endings in the repo copy.

Not everyone will have noticed this issue previously. In some scenarios, git clients would get confused by the mixture of line endings and always report these two files as changed in the working directory.


echuber2:
Not sure why it's failing this time. I suppose I did the wrong thing with the changelog.

echuber2:
Nick tells me we aren't manually updating the changelog anymore.

mwest1066:
Can we delete our local copy of the `socket.io.js` file? From https://github.com/socketio/socket.io-client it sounds like the server automatically exposes a copy of the client on `/socket.io/socket.io.js`. Using that would require reconfiguring some paths. Or we could include it from the `dist` directory inside the `socket.io` directory in `node_modules`. That should just automatically work.

The change to `mersenne.js` looks fine to me.

echuber2:
> Can we delete our local copy of the `socket.io.js` file? From https://github.com/socketio/socket.io-client it sounds like the server automatically exposes a copy of the client on `/socket.io/socket.io.js`. Using that would require reconfiguring some paths. Or we could include it from the `dist` directory inside the `socket.io` directory in `node_modules`. That should just automatically work.

@mwest1066 I removed the local copies of `socket.io.js` and `socket.io.js.map` and updated internal references to `/socket.io/socket.io.js`. It seems to work fine for me now, but `jobSequenceResults.ejs` had a note at the top suggesting people to directly reference their file in some local development, so they may have a problem if they aren't aware.

echuber2:
@mwest1066 Looks like there's a conflict because someone edited the ejs file. This may happen again in the future if we go ahead with the change.

Would it make sense to recreate and preserve `/javascripts/socket.io.js` as a symlink to `/socket.io/socket.io.js`?

mwest1066:
Rather than using `/socket/socket.io`, can we directly reference the client-side library in `node_modules`? I think this is a support option as well? Using `node_modules` would avoid the need to make changes in our reverse proxy setup in prod.

mwest1066:
> @mwest1066 Looks like there's a conflict because someone edited the ejs file. This may happen again in the future if we go ahead with the change.

Yep, this PR needs a rebase (or merge).

mwest1066:
> Would it make sense to recreate and preserve `/javascripts/socket.io.js` as a symlink to `/socket.io/socket.io.js`?

The only users of `socket.io.js` should be PL itself, so I don't think we need to maintain a backwards-compatibility symlink.

echuber2:
@mwest1066 Do you want a relative reference to `/node_modules` instead or did you intend for the file to be copied out to `/javascripts/socket.io.js` upon launch? It's not clear to me which script or config would be expected to handle that.

mwest1066:
> @mwest1066 Do you want a relative reference to `/node_modules` instead or did you intend for the file to be copied out to `/javascripts/socket.io.js` upon launch? It's not clear to me which script or config would be expected to handle that.

Like this: https://github.com/PrairieLearn/PrairieLearn/blob/12e6fc74ff83eb26cda8647eb6dcffaa5fb77613/pages/instructorFileEditor/instructorFileEditor.ejs#L9

echuber2:
I made that change and it seems to work fine.

pages/partials/jobSequenceResults.ejs
```diff
@@ -8,7 +8,12 @@
 
     2) Add this line to the header of the host page:
 
-        <script src="/javascripts/socket.io.js"></script>
+        <script src="/socket.io/socket.io.js"></script>
+
+    Note: This file is not found in the PL source itself anymore, but it
+    will be made available dynamically at the relative URL above when the
+    server is running. It might also be found in the node_modules directory:
+    /node_modules/socket.io-client/dist/socket.io.js
```
mwest1066, line 16:
```suggestion
        <script src="/node_modules/socket.io-client/dist/socket.io.js"></script>

```

PR #2599, opened 2020-06-23
Add database migrations for PL Workspaces

tdy:
Reference: #2504

- [x] *.pg
- [x] *.sql
- [x] tests


database/tables/workspace_logs.pg
```diff
@@ -0,0 +1,12 @@
+columns
+    date: timestamp with time zone default now()
+    id: bigint not null default nextval('workspace_logs_id_seq'::regclass)
+    level: smallint
```
mwest1066, line 4: Let's make a `log_level` enum for this. With values of `error`, `warn`, `info`, `debug`.

database/tables/workspaces.pg
```diff
@@ -0,0 +1,17 @@
+columns
+    id: bigint not null default nextval('workspaces_id_seq'::regclass)
+    s3_bucket: text
+    s3_root_key: text
+    state: text
```
mwest1066, line 5: Maybe we actually should use an enum here.

database/tables/workspace_logs.pg
tdy: Added as `enum_workspace_log_log_level` -- name feels clunky, but I think it's the right PL convention?

database/tables/workspaces.pg
tdy: Added as `enum_workspace_state`

database/tables/workspace_logs.pg
```diff
@@ -0,0 +1,12 @@
+columns
+    date: timestamp with time zone default now()
+    id: bigint not null default nextval('workspace_logs_id_seq'::regclass)
+    log_level: enum_workspace_log_log_level
```
mwest1066, line 4: I suggest changing this to having the column name be just `level` and the enum be `log_level`.

migrations/175_workspace_logs__create.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TYPE enum_workspace_log_log_level AS ENUM ('debug', 'error', 'info', 'warn');
```
mwest1066, line 1: It's important that these should be ordered by decreasing level, so we can easily find all things with `level < warn`, for example.

migrations/174_workspaces__create.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TYPE enum_workspace_state AS ENUM ('launching', 'running', 'stopped', 'uninitialized');
```
mwest1066, line 1: The order of these doesn't really matter, but I'd be inclined to order them as we expect they will be used (that is, `uninitialized`, `stopped`, `launching`, `running`).

migrations/175_workspace_logs__create.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TYPE enum_workspace_log_level AS ENUM ('error', 'warn', 'info', 'debug');
```
mwest1066, line 1: I guess I was thinking to make this just `log_level`, so that it's usable by other log tables in the future. I think these levels are pretty generic.

PR #2601, opened 2020-06-24
show/hide button for older submissions

### LINKED ISSUE #2597: Add UI to declutter submission history by revealing it gradually/selectively
https://github.com/PrairieLearn/PrairieLearn/blob/abba6add29fe659435e5855ffcd226d957829631/pages/partials/question.ejs#L151

I'm thinking of adding something to only display the three most recent submissions (etc.), with a "Load more submissions" button to reveal three more, etc. Or, a drop-down slider could select a range of submissions to view. There are more complicated ways to do it but my main objective here is to reduce the total amount of spam shown at the bottom of the page. On coding questions where students can resubmit many times, there is already a lot of scrolling up and down to look at the editor and the test output. There shouldn't be so many ancient records shown at the bottom by default, then.

echuber2:
Add show/hide button at the bottom of an externally-graded question to hide all but the most recent three submissions.

Closes #2597

### DIFF
--- a/pages/partials/question.ejs
+++ b/pages/partials/question.ejs
@@ -146,11 +146,42 @@
   </div>
   <% } %>
 
+  <%# Submissions section %>
   <% if (showSubmissions) { %>
-  <% submissions.forEach(function(submission, idx) { %>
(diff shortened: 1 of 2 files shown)


echuber2:
Let's see if animated PNGs work here...

outpng3_apng

@nicknytko @mwest1066 How does this look?

PR #2602, opened 2020-06-24
Home page adds example course viewer perms for other course owners, editors

trombonekenny:


nicknytko:
Looks good to me, was able to confirm it works on a dev container

PR #2605, opened 2020-06-24
Change `workspaces.state` to be NOT NULL

tdy:
@mwest1066, should the state machine be `NOT NULL`?

### DIFF
--- a/database/tables/workspaces.pg
+++ b/database/tables/workspaces.pg
@@ -2,7 +2,7 @@ columns
     id: bigint not null default nextval('workspaces_id_seq'::regclass)
     s3_bucket: text
     s3_root_key: text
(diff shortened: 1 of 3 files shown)


mwest1066:
Looks good! But migrations only get run once so we can’t go back and edit them. Instead, we need a new migration that makes the change to the existing table.

tdy:
Done -- just out of curiosity, do we really need a new migration if it hasn't been deployed, or is this just for best practice?

mwest1066:
Best practice, and changing migrations would also break local dev installations if people are pulling a new PL and running it without wiping their DB (either inside docker or outside). Also, Berkeley now has a test server running and we don't control when they deploy. Our two basic rules are:
1. Don't rewrite git history.
2. Don't rewrite database migrations. 😄

PR #2610, opened 2020-06-25
Add CBTF reservation checking for mode-setting

trombonekenny:
If a student has a checked-in CBTF exam, regardless of their IP address, set their session to `"mode": "Exam"`

Is it really this simple?

mode gets set at the authzCourseInstance level (currently) but it might fit more under `middlewares/authn`? If we implemented it up a level in the express routes, we could say something about the mode in the Home page, for example. I wasn't sure if that was something worth decoupling, though, so I left it where it was.

Possible concerns: `reservations` is a big table in production - almost a million rows. I'm not sure if it's setup to do date range searches quickly.

We have tests for mode:Exam but no test for this particular patch. I wasn't sure where it was appropriate to test it - another sproc unit test?

It's not matching off of specific exams (like to constrain them to only online exams, or per course instance exams) but that seems OK. I think we want a broad coverage with this so they can't look at other courses stuff either?

We don't have a "check out" button (for students or proctors) in case a student finishes their exam early ... and then can't use anything else in PL because they're locked out. This wasn't a problem when they could use a different computer that wasn't in the IP range. We will probably have to implement something for this soon?


sprocs/ip_to_mode.sql
```diff
@@ -15,5 +18,17 @@ BEGIN
     ELSE
         mode := 'Public';
     END IF;
+
+    -- Do we have an active CBTF reservation, if so force mode to 'Exam'
+    PERFORM *
+    FROM reservations AS r
+    WHERE
+        r.user_id = ip_to_mode.authn_user_id
+        AND r.delete_date IS NULL
```
mwest1066, line 27: Do we also want `AND r.checked_in IS NOT NULL` here?

middlewares/authzCourseInstance.sql
```diff
@@ -1,6 +1,6 @@
 -- BLOCK select_authz_data
 SELECT
-    coalesce($force_mode, ip_to_mode($ip, $req_date)) AS mode,
+    coalesce($force_mode, ip_to_mode($ip, $req_date, $authn_user_id)) AS mode,
```
mwest1066, line 3: Should this be `authn_user_id` or just `user_id`?

sprocs/ip_to_mode.sql
```diff
@@ -15,5 +18,17 @@ BEGIN
...
+        AND date BETWEEN r.access_start AND r.access_end;
```
mwest1066, line 28: We don't have any good indexes on `reservations` for this sort of query, as far I can tell.

mwest1066: I think an index on `(user_id, access_start)` for `reservations` would be sufficient to make this query fast. Possibly with a `WHERE delete_date IS NULL` condition.

mwest1066:
> mode gets set at the authzCourseInstance level (currently) but it might fit more under `middlewares/authn`? If we implemented it up a level in the express routes, we could say something about the mode in the Home page, for example. I wasn't sure if that was something worth decoupling, though, so I left it where it was.

I agree that it would be sensible to shift `mode` setting up to `middlewares/authn`, but it seems safer to not make any large change at the moment. Perhaps this is something that could go into @tbretl's auth PR?

mwest1066:
> We have tests for mode:Exam but no test for this particular patch. I wasn't sure where it was appropriate to test it - another sproc unit test?

Sounds good to me! I agree that some tests would be great here to make sure we don't break this in the future.

mwest1066:
> It's not matching off of specific exams (like to constrain them to only online exams, or per course instance exams) but that seems OK. I think we want a broad coverage with this so they can't look at other courses stuff either?
> 
> We don't have a "check out" button (for students or proctors) in case a student finishes their exam early ... and then can't use anything else in PL because they're locked out. This wasn't a problem when they could use a different computer that wasn't in the IP range. We will probably have to implement something for this soon?

I think that it's sensible to make this work just like in the physical CBTF --- once you are checked in for an exam in any course, you are locked in to exam mode for all courses.

But adding a "check out" button in PrairieSchedule would be great.

I think it might also be good to add an indicator inside PL to show that the student is in Exam mode. Perhaps we could put this in the top navbar next to their name? And make it a dropdown with some explanatory text that says "You are currently checked in for an exam, so access to other assessments is restricted. Access will be re-enabled once your exam time period has ended."

mwest1066:
Should we activate the exam timer automatically for CBTF-Online exams?

middlewares/authzCourseInstance.sql
tbretl: `authn_user_id` is, indeed, what's passed to `select_authz_data` (with the "old permissions scheme") so this looks correct to me - overrides are handled with the separate SQL block `select_effective_authz_data`

sprocs/ip_to_mode.sql
tbretl: I agree - without something like that, wouldn't it be the case that users would be forced into exam mode for the duration of the reservation, regardless of whether they were actually trying to take the exam? This seems to be a problem in particular for remote users - but I don't know enough about how the CBTF is handling remote users these days. (Do they check in and out?)

middlewares/authzCourseInstance.sql
tbretl: What if an instructor chooses to make the effective user a student with an active reservation? Should the mode be forced to `Exam` as it would if the student were the authn user? (Right now, it won't be, because no changes have been made to the SQL block `select_effective_authz_data`.)

sprocs/ip_to_mode.sql
trombonekenny: It's a bit of a sticky point that there's no check out. There's only `checked_in` and `access_start` and `access_end`  (all are timestamps). A checked out session would be one where `checked_in IS NOT NULL and now() > access_end` - but we don't have an integrated way for students to "checkout", it just times out.

That didn't matter so much when the checkin was restricted to CBTF IP space. (If their reservation went longer and they left the CBTF, they were still mode:Public from other IP addresses.) Here I'm thinking we might want to at least put a notice (maybe in pages/home?) and potentially give them a button to end exam mode to get normal mode:Public back. Thoughts on that?

middlewares/authzCourseInstance.sql
trombonekenny: Because instructors can manually force emulated modes, I don't think we need to trigger mode based on reservation when an instructor emulates a student. Is that reasonable?

trombonekenny:
> Should we activate the exam timer automatically for CBTF-Online exams?

There is an old issue up for disabling exam-level time limits when in mode:Exam. Maybe that would be a nice feature to hook in `access_end` so it acts like a visible exam timer?

middlewares/authzCourseInstance.sql
mwest1066: Seems fine, at least for this first version.

trombonekenny:
> I think it might also be good to add an indicator inside PL to show that the student is in Exam mode. Perhaps we could put this in the top navbar next to their name? And make it a dropdown with some explanatory text that says "You are currently checked in for an exam, so access to other assessments is restricted. Access will be re-enabled once your exam time period has ended."

Something like this?

image

mwest1066:
> Something like this?

Perfect!

sprocs/ip_to_mode.sql
mwest1066: Also with `access_start IS NOT NULL`, I guess.

trombonekenny: Added the `access_start IS NOT NULL`

Did a query with the student with the biggest possible list of reservations (116 non-deleted) and the timing seems fine with the existing indexes (about 908k rows):

```
prairielearn=# EXPLAIN ANALYZE                                                                                          
SELECT * FROM reservations AS r
WHERE
    r.user_id = 34103
    AND r.delete_date IS NULL
    AND r.checked_in IS NOT NULL
    AND r.access_start IS NOT NULL
    AND '2019-09-21 02:02:02-00' BETWEEN r.access_start AND r.access_end;
                                                                                                     QUERY PLAN                                  
                                                                    
-------------------------------------------------------------------------------------------------------------------------------------------------
--------------------------------------------------------------------
 Index Scan using reservations_user_id_exam_id_delete_date_constraint on reservations r  (cost=0.42..76.46 rows=1 width=289) (actual time=0.248..
0.253 rows=1 loops=1)
   Index Cond: (user_id = 34103)
   Filter: ((checked_in IS NOT NULL) AND (access_start IS NOT NULL) AND ('2019-09-21 02:02:02+00'::timestamp with time zone >= access_start) AND 
('2019-09-21 02:02:02+00'::timestamp with time zone <= access_end))
   Rows Removed by Filter: 115
 Planning Time: 0.274 ms
 Execution Time: 0.295 ms
(6 rows)
```
Index-wise I don't think we need to change anything.

trombonekenny:
@mwest1066 This should be ready to review and merge. Thanks in advance!

Issue #2600, opened 2020-06-23
Add attribute for "None of the above" to <pl-multiple-choice>


bojinyao:
The attribute can be named as `enable-nota` (enable none of the above) of type: boolean, default: "false" 
When set to "true"
- "None of the above", as an option will appear at the bottom of the choices, regardless of value of `fixed-order`
- "None of the above" will be the correct answer with probability `1/(number_correct + 1)` to give all correct answer(s) equal chance of appearing
- `number-answers` still upper-bounds total number of choices, "None of the above" included
- `weight` should not be affected
- `inline` should not be affected
- `answers-name` should not be affected

nicknytko:
Resolved in #2623

PR #2611, opened 2020-06-25
Add enable-nota attribute for <pl-multiple-choice> element

bojinyao:
In response to: https://github.com/PrairieLearn/PrairieLearn/issues/2600

`enable-nota` will be an additional attribute to `<pl-multiple-choice>` element of type boolean, default to false

When set to true:
- An option `None of the above` will always appear at the *bottom* of all sampled choices
    - This option will be correct with probability `1/(number_correct + 1)`
    - If no correct answer is provided, the option will be correct with probability 1
- The total number of choices is still upper-bounded by `number-answers` attribute.
- `fixed-order` does not affect its placement
- This has no meaningful effect on `inline`, `weight`, or `answers-name` attributes


elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -11,9 +11,10 @@
 def prepare(element_html, data):
     element = lxml.html.fragment_fromstring(element_html)
     required_attribs = ['answers-name']
-    optional_attribs = ['weight', 'number-answers', 'fixed-order', 'inline']
+    optional_attribs = ['weight', 'number-answers', 'fixed-order', 'inline', 'enable-nota']
```
nicknytko, line 14: Can we name the attribute `none-of-the-above` to be more consistent with the others?

elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -11,9 +11,10 @@
...
     pl.check_attribs(element, required_attribs, optional_attribs)
     name = pl.get_string_attrib(element, 'answers-name')
+    enable_nota = pl.get_boolean_attrib(element, 'enable-nota', False)
```
nicknytko, line 17: Lets add a `NONE_OF_THE_ABOVE_DEFAULT = False` constant to the top of the file and reference it here

elements/pl-multiple-choice/pl-multiple-choice.py
bojinyao: updated 👍

elements/pl-multiple-choice/pl-multiple-choice.py
bojinyao: updated 👍

nicknytko:
Subsumed by #2623

PR #2615, opened 2020-06-26
Fix alignment of manual grading footnote

tdy:
left=old, right=new

image

### DIFF
--- a/pages/partials/questionFooter.ejs
+++ b/pages/partials/questionFooter.ejs
@@ -30,7 +30,7 @@
               <% } %>
           <% } %>
           <% if (showManualGradingMsg) { %>
-          <span class="mr-2">This question will be manually graded.</span>
+          <span class="order-3 ml-2">This question will be manually graded.</span>
           <% } %>
         </span>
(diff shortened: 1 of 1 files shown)


mwest1066:
Do we want to reverse the order of the text and button? It somehow looks weird to me having the button be in the middle of the footer.

nicknytko:
> Do we want to reverse the order of the text and button? It somehow looks weird to me having the button be in the middle of the footer.

Agreed. @tdy can we see a screenshot of this?

tdy:
> Agreed. @tdy can we see a screenshot of this?

image

nicknytko:
> > Agreed. @tdy can we see a screenshot of this?
> 
> image

That looks way better imo

tdy:
Agreed. The other thing I was considering was keeping both buttons but graying out and disabling the grade button.

mwest1066:
I think that this now looks great. I think graying out "Grade" isn't needed given that we've got the explanatory text.

trombonekenny:
lgtm

PR #2617, opened 2020-06-27
Add `questions.workspace_graded_image` column

tdy:
Looks like we also need a `questions.workspace_graded_image` column based on #2504's proposed json:
```json
{
    "workspace": {
        "image": "some-docker-image-name",
        "gradedFiles": [
            "animal.h",
            "animal.cpp"
        ]
    }
}
```

### DIFF
--- a/database/tables/questions.pg
+++ b/database/tables/questions.pg
@@ -22,6 +22,7 @@ columns
     topic_id: bigint
     type: enum_question_type
     uuid: uuid not null
+    workspace_graded_files: text[] default ARRAY[]::text[]
     workspace_image: text
 
 indexes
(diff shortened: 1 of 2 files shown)


tdy:
Also I realized that I misnamed the previously merged `176_questions__workspace_image__alter` (should be `add`), but I guess that's set in stone now?

mwest1066:
> Also I realized that I misnamed the previously merged `176_questions__workspace_image__alter` (should be `add`), but I guess that's set in stone now?

Actually the name doesn't matter, only the number, so you can rename migrations without problem. Just don't change the number!

mwest1066:
Just realized the title of this PR was wrong? Not that it really matters.

PR #2618, opened 2020-06-27
Add question setup for Workspaces

tdy:
Reference: "Questions" section in #2504 

* [x] json schema
* [x] sync input
* [x] sync query
* [x] sync tests
* [x] docs
* [x] tests


docs/workspace/index.md
```diff
@@ -0,0 +1,31 @@
+# Workspaces
+
+This file documents how to create workspace questions.
```
tdy, line 3:
```suggestion
Workspaces allow students to work in persistent remote containers via in-browser frontends such as VS Code. The remote containers are configured by instructors to provide custom, uniform environments per question. Workspace questions are integrated with the standard PrairieLearn autograding pipeline.
```

mwest1066:
Needs to be rebased on top of #2617 ?

mwest1066:
> Needs to be rebased on top of #2617 ?

I mean, needs to be merged with master now that #2617 was merged.

tdy:
> 
> 
> > Needs to be rebased on top of #2617 ?
> 
> I mean, needs to be merged with master now that #2617 was merged.

Done

docs/workspace/index.md
```diff
@@ -0,0 +1,31 @@
+# Workspaces
```
nwalters512, line 1: Since this will be shown on the live documentation site once merged, I think we should put a big "this is being built and is not yet ready for use" disclaimer at the top to prevent "why doesn't this work yet?" questions.

coatless: @nwalters512 it will be hidden from being served as it wasn't added to `mkdocs.yml`. 

https://github.com/PrairieLearn/PrairieLearn/blob/a4f35b5560305ae3b0051bfbeafcf59be170dd04/mkdocs.yml#L32

That said, @tdy you probably want to add it to the `mkdocs.yml` configuration and add a warning.

mwest1066: Good point, @coatless. Maybe we should actually leave it hidden for now?

nwalters512: > @nwalters512 it will be hidden from being served as it wasn't added to `mkdocs.yml`.

Good point!

PR #2621, opened 2020-06-29
Fix login cookie expiry

### LINKED ISSUE #1250: Logout Time
I don't know if I'm just noticing it more recently because of the pretty new login screen, but I feel like I'm being forced to log in to PL more often. Is that possible?

Regardless, it might be worth rethinking any automatic logout interval that might be being used as 2FA looms.

### LINKED ISSUE #2268: Fix login cookie expiry
We currently have a 24-hour expiry on authentication cookies. That is, we force a new authentication every 24-hours. This is bad because student work can be lost during the submission process when the authentication has expired (e.g., #1371). It would also address #1250.

To fix this, we should:

1. After we have successfully processed the authentication cookie, reset it with a new timeout from the current time. This will avoid "random" logouts in the middle of work because the cookie will keep extending while a user keeps working. To do this, we should add code just before `next()` at the end of `middlewares/authn.js`: https://github.com/PrairieLearn/PrairieLearn/blob/master/middlewares/authn.js#L125

This code should set the `pl_authn` cookie again just like this: https://github.com/PrairieLearn/PrairieLearn/blob/master/pages/authCallbackShib/authCallbackShib.js#L36

The `tokenData` for the `pl_authn` cookie needs to include `user_id: authnData.user_id` and the `authn_provider_name`, which can be taken from the previous cookie or set to `null` if we don't have one.

2. Add a config variable like `authnCookieMaxAgeMin` (in `lib/config.js`) for the `maxAge` property of the authn cookies. At the moment we hard-code 24-hours everywhere.

3. Change the default value of `authnCookieMaxAgeMin` to something like `7 * 24 * 60` so that users will stay logged in so long as they use PL at least once per week.

### LINKED ISSUE #2604: Work Lost on Forced Logout
Seemingly at random when I hit "Save and Grade" on an assignment PrairieLearn will redirect me to a login page and then lose all the work done. It does not even seem correlated to a dead session as this happened 10 minutes after moving on from a question that "save and grade"d successfully.

tdy:
* [x] reset `pl_authn` timeout on each authentication
* [x] refactor `pl_authn` max age as `config.authnCookieMaxAgeMin` (default 7 * 24 * 60)

---

* fixes #2604 
* fixes #2268 
* fixes #1250


pages/authCallbackAzure/authCallbackAzure.js
```diff
@@ -32,7 +32,7 @@ router.all('/', function(req, res, next) {
                 authn_provider_name: 'Azure',
             };
             var pl_authn = csrf.generateToken(tokenData, config.secretKey);
-            res.cookie('pl_authn', pl_authn, {maxAge: 24 * 60 * 60 * 1000});
+            res.cookie('pl_authn', pl_authn, {maxAge: config.authnCookieMaxAgeMin});
```
mwest1066, line 35: The `maxAge` parameter is in milliseconds, but we are passing a value in minutes?

tdy: Ah oops, so use something like `config.authnCookieMaxAgeMin * 60 * 1000`? Or use a millisecond var?

mwest1066: I have no preference either way

tdy: Changed to `authnCookieMaxAgeMilliseconds`

PR #2623, opened 2020-06-30
Add 'all-of-the-above' attribute to <pl-multiple-choice>

bojinyao:
Addition to #2611 
`all-of-the-above` will be an additional attribute to <pl-multiple-choice> element of type boolean, default to false

"All of the above" as an option, is considered correct only if every option above it is correct.

When set to true:

- An option "All of the above" will always appear at the bottom of all sampled choices
  - If `none-of-the-above` is also set to true, "All of the above" will always appear right above "None of the above" option, but still below all other choices, because of semantics.
  - "All of the above" option will be correct with probability 1/(number_correct + 1) when `none-of-the-above` is false
    - If `none-of-the-above` is also set to true, either option will be correct with probability 1/(number_correct + 2)
  - If no correct answer is provided, unless `none-of-the-above` is set to true, which would be guaranteed to be true; an error will be raised instead.
  - If "All of the above" is not correct, only 1 of the options above it is correct, unless "None of the above" is correct, in which case, all other options are incorrect.
- The total number of choices is still upper-bounded by `number-answers` attribute.
  - In order to be consistent in number of answers displayed (to prevent unintended hints to students who can guess correctness solely based on number of choices displayed, rather than contents of these choices), `all-of-the-above` might lower the number of answers displayed vs. when not enabled. The actual number is influenced directly by number_correct provided by the question author. For example, if number_incorrect>10, number_correct=3, `none-of-the-above`=false, `all-of-the-above`=true, `number-answers`=10; only 4 choices will be displayed because number_correct + `none-of-the-above` = 4.
- `fixed-order` does not affect its placement
- This has no meaningful effect on `inline`, `weight`, or `answers-name` attributes
- When `inline` is enabled, instead of "All of the above" or "None of the above", it will display "All of these" and "None of these" for better semantics

New Exceptions:
- when no correct choice is provided and `none-of-the-above` is not enabled, exception will be thrown
- when `all-of-the-above` is enabled, but less than 2 incorrect choice is provided, exception will be thrown to avoid confusing end user.
- when `number-answers` is set and either `none-of-the-above` inclusive-or `all-of-the-above` is enabled, but `number-answers` can't be logically met, exception will be thrown with a message telling user number of missing choices.

Please let me know if anything is confusing


bojinyao:
Fix https://github.com/ace-lab/pl-ucb-cs10/issues/42

bojinyao:
Hi @nicknytko , can you take a look that the CI checks aren't passing, doesn't seem to have anything to do with code in this PR. Thank you! 😄

tdy:
@bojinyao You can ignore all of those coveralls errors. We're hoping to fix that soon 😩

elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -5,17 +5,15 @@
...
 
-def prepare(element_html, data):
-    element = lxml.html.fragment_fromstring(element_html)
-    required_attribs = ['answers-name']
-    optional_attribs = ['weight', 'number-answers', 'fixed-order', 'inline', 'hide-letter-keys']
-    pl.check_attribs(element, required_attribs, optional_attribs)
-    name = pl.get_string_attrib(element, 'answers-name')
-
+def categorize_options(element):
+    """Get provided corret and incorrect answers"""
```
nicknytko, line 16:
```suggestion
    """Get provided correct and incorrect answers"""
```

exampleCourse/questions/element/multipleChoice/question.html
```diff
@@ -117,3 +117,33 @@
...
+<div class="card my-2">
+    <div class="card-header">
+        Part 5
+    </div>
+
+    <div class="card-body">
+
+        <pl-question-panel>
+            <p>Here, the <code>pl-multiple-choice</code> element displays "All of the above" and "None of the above" as options below the correct and incorrect options provided.
+                Either two of these options could be correct with uniform probability as all individual correct options. The number of options displayed is also consistent across
```
nicknytko, line 129:
```suggestion
                Both of these options could be correct with the same probability as all of the individual correct options. The number of options displayed is also consistent across
```

elements/pl-multiple-choice/pl-multiple-choice.py
nicknytko, line 16:
```suggestion
    """Get provided correct and incorrect answers"""
```

elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -30,34 +28,134 @@ def prepare(element_html, data):
...
+    # 2. Sample corret and incorrect choices
     sampled_correct = random.sample(correct_answers, number_correct)
     sampled_incorrect = random.sample(incorrect_answers, number_incorrect)
 
     sampled_answers = sampled_correct + sampled_incorrect
     random.shuffle(sampled_answers)
 
-    fixed_order = pl.get_boolean_attrib(element, 'fixed-order', False)
+    # 3. Modify sampled choices
+    fixed_order = pl.get_boolean_attrib(element, 'fixed-order', FIXED_ORDER_DEFAULT)
```
nicknytko, line 132: Good catch!

ChangeLog.md
```diff
@@ -99,6 +99,8 @@
 
   * Add Python library regex to plbase (Nathan Bowman).
 
+  * Add `all-of-the-above` and `none-of-the-above` attributes to `<pl-multiple-choice>` (Bojin Yao)
```
nicknytko, line 102: In the future you don't need to update the changelog, we've stopped using it in favor of Git commits when new versions are deployed

elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -30,34 +28,134 @@ def prepare(element_html, data):
...
+        number_correct = 1
+        number_incorrect = max(0, number_answers - number_correct)
 
-    number_answers = max(1, min(1 + len_incorrect, number_answers))
-    number_correct = 1
-    number_incorrect = number_answers - number_correct
     if not (0 <= number_incorrect <= len_incorrect):
         raise Exception('INTERNAL ERROR: number_incorrect: (%d, %d, %d)' % (number_incorrect, len_incorrect, number_answers))
 
+    # 2. Sample corret and incorrect choices
```
nicknytko, line 124:
```suggestion
    # 2. Sample correct and incorrect choices
```

elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -30,34 +28,134 @@ def prepare(element_html, data):
...
+            nota_text = 'None of these'
+        else:
+            nota_text = 'None of the above'
+        # Add 'None of the above' option after shuffling
+        sampled_answers.append((len_total + 1, nota_correct, nota_text))
+
+    # 4. Write to data
+    # Because 'All of the above' is below all the correct choice(s) when it's
+    # true, the variable correct_answer will save it as correct, and
+    # overwritting previous choice(s)
```
nicknytko, line 158:
```suggestion
    # overwriting previous choice(s)
```

elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -30,34 +28,134 @@ def prepare(element_html, data):
...
+
+    if enable_aota and len_correct < 2:
+        # To prevent confusion on the client side
+        raise Exception('pl-multiple-choice element must have at least 2 correct answers when all-of-the-above is set')
+
+    # 1. Pick the choice(s) to display
+    number_answers = pl.get_integer_attrib(element, 'number-answers', -1)
+    # determine if user provides number-answers
+    set_num_answers = True
+    if number_answers == -1:
```
nicknytko, line 77:
```suggestion
    number_answers = pl.get_integer_attrib(element, 'number-answers', None)
    # determine if user provides number-answers
    set_num_answers = True
    if number_answers is None:
```

let's be more explicit here and use `None` instead of `-1`

ChangeLog.md
bojinyao: Oh I see, i was wondering about that haha. I think Contributions page might need an update haha. Thank you!

PR #2625, opened 2020-06-30
Add Pl Parson element

xinhexu2:
This element using Parson method to make a shuffle list of things to display and achieve drag the things, when the strings has the correct indentation and order, the result can be submitted and get the correct answer.

- [x] Clean the code and solve the license issue

- [x] Create the new example question parsons

- [ ] handle the long line by using the break-word

- [x] use external autograder


docs/elements.md
```diff
@@ -72,6 +72,7 @@ The following **Conditional** elements are available:
 - [`pl-hide-in-panel`](#pl-hide-in-panel-element): Hides content in one or more display panels.
 - [`pl-external-grader-results`](#pl-external-grader-results-element):
   Displays results from questions that are externally graded.
+- [`pl-parsons-problem`](#pl-parsons-problem-element): Submit the result in the correct order and indentation, which by using parson method,
```
nicknytko, line 75: Lets put this in the "submission elements" section, since this is something that can be submitted for a grade.

docs/elements.md
```diff
@@ -1545,6 +1546,36 @@ It expects results to follow [the reference schema for external grading results]
...
+
+### Details
+
+This element using Parson method to make a shuffle list of things to display and achive drag the things, when the strings has the correct indentation and order, the result can be submited and get the correct answer.
+
+
+### Example Implementations
+
+- [elementParsons]
+
```
nicknytko, line 1578: ^ ditto above

elements/pl-parsons-problem/pl-parsons-problem.css
```diff
@@ -0,0 +1,73 @@
...
+    background-color: #ff0000;
+}    
+.misaligned {
+    background-color: #cccc00;
+}    
+/* .break-word {
+    overflow-wrap: break-word;
+    word-break: keep-all;
+    overflow: hidden; 
+ } */
```
nicknytko, line 73:
```suggestion
```

elements/pl-parsons-problem/pl-parsons-problem.js
```diff
@@ -0,0 +1,192 @@
...
+// //       if (this.feedback_exists) {
+// //           $("#ul-" + sortableId).removeClass("incorrect correct");
+// //           var li_elements = $("#ul-" + sortableId + " li");
+// //           $.each(this.FEEDBACK_STYLES, function(index, value) {
+// //               li_elements.removeClass(value);
+// //           });
+// //       }
+// //       this.feedback_exists = false;
+// };
+
```
nicknytko, line 68: Not just here, but in general can we remove these large blocks of commented out code?

```suggestion
```

elements/pl-parsons-problem/pl-parsons-problem.py
```diff
@@ -0,0 +1,280 @@
...
+
+        # correct_answer = data['correct_answers'].get(name, None)
+        # if correct_answer is None:
+        #     html = 'ERROR: No true answer'
+        # else:
+        #     html = '<pre>'
+        #     for piece in correct_answer:
+        #         html += ('    ' * piece.get('indent', 0)) + piece.get('html', '') + '\n'
+        #     html += '</pre>'
+
```
nicknytko, line 207:
```suggestion
```

elements/pl-parsons-problem/pl-parsons-problem.py
```diff
@@ -0,0 +1,280 @@
...
+#             # actually an invalid submission
+#             data['raw_submitted_answers'][name] = '0'
+#             data['format_errors'][name] = 'INVALID choice'
+#     elif result == 'invalid':
+#         data['raw_submitted_answers'][name] = '0'
+#         data['format_errors'][name] = 'INVALID choice'
+# 
+#         # FIXME: add more invalid choices
+#     else:
+#         raise Exception('invalid result: %s' % result)
```
nicknytko, line 280: We should eventually try to implement a `test()` function for this element.  This takes a `test_type` parameter that can be one of `correct`, `incorrect`, or `invalid` and should create a test submission &mdash; you can see an example here: https://github.com/PrairieLearn/PrairieLearn/blob/master/elements/pl-number-input/pl-number-input.py#L328.

mfsilva22:
- [ ] We should add header names to the top of each column. I think we can use as default header names like in the example below ("Drag from here" and "Construct your solution here". Instructors can customize the title using the header attribute. I am thinking something like `header_left_column` and `header_right_column`, but open for suggestions.

image

- [ ] The indentation is not reflected in the submission panel. It should show the correct indentation

elements/pl-parsons-problem/pl-parsons-problem.css
xinhexu2: I am sorry, I forget to delete it

nicknytko:
I was helping someone on Slack with using the autograder and figured this snippet could be helpful when hooking up the parsons element with the external grader:

```python
import base64
def parse(element_html, data):
    file_name = "user_code.py" # this should be pulled from an attribute
    file_data = """ ... build the file data here ... """
    data['submitted_answers']['_files'] = [{
        'name': file_name,
        'contents': base64.b64encode(file_data.encode('utf-8')).decode('utf-8')
    }]
```

Hopefully this is helpful because the base64 encodes and string encode/decodes weren't obvious to me at first.

elements/pl-parsons-problem/pl-parsons-problem.py
```diff
@@ -0,0 +1,328 @@
...
+            #     try:
+            #         score = float(score)
+            #         if score >= 1:
+            #             html = html + '&nbsp;<span class="badge badge-success"><i class="fa fa-check" aria-hidden="true"></i> 100%</span>'
+            #         elif score > 0:
+            #             html = html + '&nbsp;<span class="badge badge-warning"><i class="fa fa-circle-o" aria-hidden="true"></i> {:d}%</span>'.format(math.floor(score * 100))
+            #         else:
+            #             html = html + '&nbsp;<span class="badge badge-danger"><i class="fa fa-times" aria-hidden="true"></i> 0%</span>'
+            #     except Exception:
+            #         raise ValueError('invalid score' + score)
```
nicknytko, line 227:
```suggestion
```

elements/pl-parsons-problem/pl-parsons-problem.py
```diff
@@ -0,0 +1,328 @@
...
+        html = 'Your submission is graded from top to bottom.  Pieces shown in green are in the correct position with the correct indentation.  Yellow pieces are in the right position but with the wrong indentation.  Red pieces are in the wrong position.  Grading stops as soon as you hit a red or yellow piece.'
+        # correct_answer = data['correct_answers'].get(name, None)
+        # if correct_answer is None:
+        #     html = 'ERROR: No true answer'
+        # else:
+        #     html = '<pre>'
+        #     for piece in correct_answer:
+        #         html += ('    ' * piece.get('indent', 0)) + piece.get('html', '') + '\n'
+        #     html += '</pre>'
+
```
nicknytko, line 238:
```suggestion
```

elements/pl-parsons-problem/info.json
```diff
@@ -0,0 +1,22 @@
...
+    "dependencies": {
+        "coreScripts": [
+            "lodash.min.js"
+        ],
+        "elementStyles": [
+            "pl-parsons-problem.css"
+        ],
+        "elementScripts": [
+            "jquery-ui.min.js",
+            "jquery.ui.touch-punch.min.js",
```
nicknytko, line 12:
```suggestion
```

elements/pl-parsons-problem/pl-parsons-problem.py
```diff
@@ -0,0 +1,328 @@
...
+def getleadAnswer(submitted, pieces, leading_code):
+    if submitted == '':
+        return ''
+    answer = ''
+    for piece in submitted.split('-'):
+        piece_no, indent = piece.split(':')
+        indent = int(indent)
+        piece_no = int(piece_no)
+        answer = leading_code + '\n' + answer + ('    ' * indent) + pieces[piece_no] + '\n'
+    return answer
```
nicknytko, line 42: These three functions all look almost identical, can we refactor so that we only have one copy of the logic?

nicknytko:
@xinhexu2 it looks like when the trailing/leading code is enabled the trailing and leading code is added to *each line*, when we would only want this added once.

Current autograder submission:
```python
import numpy as np
import numpy as np
import numpy as np
import numpy as np
import numpy as np
def make_array_b(a):
print (make_array_b(a))
    b = np.copy(a)
print (make_array_b(a))
    for i in range(1,len(a)):
print (make_array_b(a))
        b[i] += a[i-1]
print (make_array_b(a))
    return b
print (make_array_b(a))
```

what we would want:
```python
import numpy as np
def make_array_b(a):
    b = np.copy(a)
    for i in range(1,len(a)):
        b[i] += a[i-1]
    return b
print (make_array_b(a))
```

i propose taking out the leading/trailing code from your `for` loops.

nicknytko:
@mfsilva22 what is the status of this PR?  Should we close it in favor of the UBC variant?

PR #2665, opened 2020-07-02
Add pl-workspace element

tdy:


lib/question.js
```diff
@@ -1117,6 +1117,8 @@ module.exports = {
             urls.baseUrl = urlPrefix;
         }
 
+        urls.workspaceUrl = urlPrefix + '/workspace/' + variant.workspace_id;
```
nwalters512, line 1120: I think we should omit `urls.workspaceUrl` if the variant doesn't have a workspace defined (that is, if `variant.workspace_id` is null). That'll keep other code simpler:

```js
// Complex
if (locals.workspaceUrl.split('/').pop() !== 'null') { }

// Simple!
if (locals.workspaceUrl != null) { }
```

pages/partials/assessmentScorePanel.ejs
```diff
@@ -6,6 +6,11 @@
     <div class="d-flex justify-content-center">
       <a class="btn btn-info" href="<%= urlPrefix %>/assessment_instance/<%= assessment_instance.id %>/">Assessment overview</a>
     </div>
+    <% if (locals.workspaceUrl.split('/').pop() !== 'null') { %>
+    <div class="d-flex justify-content-center mt-2">
+      <a class="btn btn-info" target="_blank" href="<%= locals.workspaceUrl %>">Open workspace</a>
+    </div>
+    <% } %>
```
nwalters512, line 13: Can you explain why you added this? I was thinking the link would be provided solely by the element.

elements/pl-workspace/pl-workspace.py
```diff
@@ -0,0 +1,13 @@
+import chevron
+
+
+def render(element_html, data):
+    # Get workspace url
+    workspace_url = data['options']['workspace_url']
```
nwalters512, line 6: Maybe not necessary for the MVP, but it'd be nice to do something different if `workspace_url` isn't defined (i.e., the question doesn't have a workspace set up). Either throwing an error, or rendering a red box/button/something to alert course staff to the fact that something is wrong. In general, failing loudly/brightly is better to help people catch their mistakes sooner.

tdy:
> 
> 
> When I try to access the `demo/workspace` question, I get an error. It looks like you haven't added a database migration to add `workspace_id` to the variants table yet? Blocking on that for now.

Sorry, that migration just got merged this afternoon. I merged master just now.

pages/partials/assessmentScorePanel.ejs
tdy: > 
> Can you explain why you added this? I was thinking the link would be provided solely by the element.

Yeah, I was planning to demo/pitch this at the meeting tomorrow, but I forgot you won't be able to make it. I'll try to summarize my thinking:

I like how `pl-workspace` allows instructors to place the button within the flow of a question, but I think it's also good to provide an additional predictable/reliable version of the button.

Say I'm coming back to work on a question, or I'm an instructor wanting to test the workspace, or some kind of analogous situation where I don't need/want to go through the question narrative -- it'd be preferable not having to scroll through the question every time to find `pl-workspace` again, so I thought it might be good UX to also include a standard sidebar button.

tdy:
> Sorry, that migration just got merged this afternoon. I merged master just now.

Oops this turned into a mess. I should'vve merged `master` into `ws-config` and then merged `ws-config` into this PR. Now the commits are a mess until `ws-config` gets merged into master 🤦‍♂️

lib/question.js
tdy: Changed in https://github.com/PrairieLearn/PrairieLearn/pull/2665/commits/790c99a52d62c601b1ef491a85e60c16d7fccef6

elements/pl-workspace/pl-workspace.py
tdy, line 6:
```suggestion
    # TODO: Improve UX if key undefined: https://github.com/PrairieLearn/PrairieLearn/pull/2665/files#r449319839
    workspace_url = data['options']['workspace_url']
```

elements/pl-workspace/pl-workspace.py
```diff
@@ -0,0 +1,14 @@
+import chevron
+
+
+def render(element_html, data):
+    # Get workspace url
+    # TODO: Improve UX if key undefined: https://github.com/PrairieLearn/PrairieLearn/pull/2665/files#r449319839
```
tdy, line 6:
```suggestion
    # TODO: Improve UX if key undefined (/pull/2665#discussion_r449319839)
```

elements/pl-workspace/pl-workspace.py
tdy: Added `TODO` comment

nwalters512: nit, including the full URL is preferable since that allows you to quickly cmd-click from your IDE to get to the comment.

lib/question.js
```diff
@@ -1117,6 +1117,9 @@ module.exports = {
             urls.baseUrl = urlPrefix;
         }
 
+        const workspaceUrl = urlPrefix + '/workspace/' + variant.workspace_id;
+        urls.workspaceUrl = variant.workspace_id ? workspaceUrl : null;
```
nwalters512, line 1121: Refactor for clearer logic + preference for string templates:
```suggestion
        if (variant.workspace_id) {
          urls.workspaceUrl = `${urlPrefix}/workspace/${variant.workspace_id}`;
        }
```

lib/question.js
```diff
@@ -1117,6 +1117,9 @@ module.exports = {
             urls.baseUrl = urlPrefix;
         }
 
+        const workspaceUrl = urlPrefix + '/workspace/' + variant.workspace_id;
```
nwalters512, line 1120: I don't believe we actually want `urlPrefix` to be used here. Depending on context, this can assume a lot of different values:
* `/pl/course/1`
* `/pl/course_instance/1`
* `/pl/course_instance/1/instructor`
I was thinking that workspaces would always be served from `/pl/workspace/[id]`, although that may be wrong. cc @mwest1066

pages/partials/assessmentScorePanel.ejs
nwalters512: While I agree with the motivation for a different way to access workspaces, I don't think putting this type of button in the sidebar is "standard". I think keeping it somewhere in the normal question flow is more desirable. A few options that might be better:

* The question header
* The question footer
* A dedicated panel

I mocked up the question header option; what do you think about this?

question panel with open workspaces button in the header

tdy: I like it!

elements/pl-workspace/pl-workspace.py
```diff
@@ -0,0 +1,14 @@
+import chevron
+
+
+def render(element_html, data):
+    # Get workspace url
+    # TODO: Improve UX if key undefined (/pull/2665#discussion_r449319839)
```
tdy, line 6:
```suggestion
    # TODO: Improve UX if key undefined (https://github.com/PrairieLearn/PrairieLearn/pull/2665#discussion_r449319839)
```

lib/question.js
```diff
@@ -1117,6 +1117,10 @@ module.exports = {
             urls.baseUrl = urlPrefix;
         }
 
+        if (variant.workspace_id) {
+          urls.workspaceUrl = `${urlPrefix}/workspace/${variant.workspace_id}`;
+        }
```
tdy, line 1122:
```suggestion
        if (variant.workspace_id) {
          urls.workspaceUrl = `${urlPrefix}/workspace/${variant.workspace_id}`;
        } else {
          urls.workspaceUrl = null;
        }
```
@nwalters512 I think we still need to set the `null` case, unless you prefer leaving it `undefined`?

nwalters512: Our code generally shouldn't be distinguishing between `null` and `undefined`, so I'm happy with the latter. If you feel strongly about making the `null` explicit, I won't stop you 🙂

pages/partials/question.ejs
```diff
@@ -70,13 +70,20 @@
...
+          <button class="btn btn-xs btn-light" type="button" data-toggle="modal" data-target="#iqqrcode" aria-expanded="false" aria-controls="iqqrcode">
+            Proctor <i class="fa fa-question-circle" aria-hidden="true"></i>
           </button>
         </div>
-      <% } %>
+        <% } %>
+        <% if (locals.workspaceUrl) { %>
+        <div class="float-sm-right ml-1">
+          <a class="btn btn-xs btn-light" target="_blank" href="<%= locals.workspaceUrl %>">
+            Open workspace <i class="fa fa-external-link-alt" aria-hidden="true"></i>
```
nwalters512, line 83: Nit: e should consider using the same icon in the `pl-workspace` element, and possibly using a different icon - https://fontawesome.com/icons/laptop-code?style=solid or https://fontawesome.com/icons/code?style=solid might be good options.

tdy: Good point, switched both buttons to `fa-laptop-code` in 8dad16c705a93dc1806f82063c18efda39e17c09

pages/partials/question.ejs
```diff
@@ -70,13 +70,20 @@
...
             <% throw Error('unknown question_context: ' + question_context); %>
         <% } %>
-      <% if (authz_data.mode == 'Exam') { %>
-        <div class="float-sm-right">
-          <button class="btn btn-xs btn-secondary" type="button" data-toggle="modal" data-target="#iqqrcode" aria-expanded="false" aria-controls="iqqrcode">
-            Proctor <i class="fa fa-question-circle"></i>
+        <% if (authz_data.mode == 'Exam') { %>
+        <div class="float-sm-right ml-1">
+          <button class="btn btn-xs btn-light" type="button" data-toggle="modal" data-target="#iqqrcode" aria-expanded="false" aria-controls="iqqrcode">
+            Proctor <i class="fa fa-question-circle" aria-hidden="true"></i>
```
tdy, line 76:
```suggestion
            <i class="fa fa-question-circle" aria-hidden="true"></i> Proctor
```

pages/partials/question.ejs
```diff
@@ -70,13 +70,20 @@
...
           </button>
         </div>
-      <% } %>
+        <% } %>
+        <% if (locals.workspaceUrl) { %>
+        <div class="float-sm-right ml-1">
+          <a class="btn btn-xs btn-light" target="_blank" href="<%= locals.workspaceUrl %>">
+            Open workspace <i class="fa fa-laptop-code" aria-hidden="true"></i>
```
tdy, line 83:
```suggestion
            <i class="fa fa-laptop-code" aria-hidden="true"></i> Open workspace
```

elements/pl-workspace/pl-workspace.mustache
```diff
@@ -0,0 +1,3 @@
+<a class="btn btn-info" target="_blank" href="{{workspace_url}}">
+  Open workspace <i class="fa fa-laptop-code" aria-hidden="true"></i>
```
tdy, line 2:
```suggestion
  <i class="fa fa-laptop-code" aria-hidden="true"></i> Open workspace
```

PR #2673, opened 2020-07-03
Add "request a course" page to home screen

nicknytko:
Adds a "request a course" page for instructors.

Request Page:

image
--
image
--
image

Admin page:

image

This can be reached from the home page and via the nav dropdown:
image

Resolves:
- #2772


nicknytko:
Delete "course institution field" in form and default to user's institution

pages/instructorRequestCourse/instructorRequestCourse.js
```diff
@@ -0,0 +1,78 @@
...
+        } else {
+            const sql_params = {
+                short_name,
+                title,
+                institution,
+                github_user,
+                user_id: res.locals.authn_user.user_id,
+            };
+            sqldb.query(sql.insert_request, sql_params, (err, _result) => {
+                if (ERR(err, next)) return;
```
mwest1066, line 70: Are we also going to post to slack here?

pages/partials/navbar.ejs
```diff
@@ -72,6 +72,8 @@ if (!locals.effective_user_has_no_access_to_course_instance) {
                 <span class="badge badge-pill badge-primary news-item-link-count"><%= news_item_notification_count %></span>
               <% } %>
             </a>
+            <hr>
+            <a class="dropdown-item" href="<%= plainUrlPrefix %>/request_course"> Request Course </a>
```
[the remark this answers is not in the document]
nicknytko: I'm thinking we can change this to "Course requests", and have "Request course" on the home page.

pages/instructorRequestCourse/instructorRequestCourse.js
nicknytko: Good idea! Should we reuse the existing ops-bot for this?

mwest1066: Yep!

database/tables/administrators.pg
```diff
@@ -7,4 +7,4 @@ indexes
     administrators_user_id_key: UNIQUE (user_id) USING btree (user_id)
 
 foreign-key constraints
-    administrators_user_id_fkey: FOREIGN KEY (user_id) REFERENCES users(user_id) ON UPDATE CASCADE ON DELETE CASCADE
+    administrators_user_id_fkey: FOREIGN KEY (user_id) REFERENCES users(user_id) ON UPDATE CASCADE ON DELETE CASCADE
```
mwest1066, line 10: spurious EOL deletion?

pages/instructorRequestCourse/instructorRequestCourse.js
```diff
@@ -0,0 +1,82 @@
...
+        } else {
+            const sql_params = {
+                short_name,
+                title,
+                github_user,
+                user_id: res.locals.authn_user.user_id,
+            };
+            sqldb.query(sql.insert_request, sql_params, (err, _result) => {
+                if (ERR(err, next)) return;
+                /* Ignore the callback, we don't actually care if the message gets sent before we render the page */
```
mwest1066, line 70: In this case we often use a callback like `(err) => {ERR(err, (e) => {logger.error(e);});}` so that if there is an error, we will correctly update its stack and log it. This happens in the background so doesn't stop us proceeding with page render.

pages/instructorRequestCourse/instructorRequestCourse.ejs
```diff
@@ -0,0 +1,91 @@
...
+                        <% if (!has_one_course) { %>
+                            <!-- Small blurb for people that have never requested a course before -->
+                            <p>
+                                To request a course space, you need to provide the <strong>course rubric and number</strong> (for example, XC 101) and the course's <strong>title</strong>. You may also optionally provide your GitHub username if you would like the ability to edit content offline.
+                            </p>
+                        <% } else { %>
+                            <p>
+                                Your course request is pending. Once your request is reviewed by a member of staff and your course space is created, you will receive an email with further instructions and resources to get started.
+                            </p>
+                        <% } %>
```
mwest1066, line 68: I suggest not bothering with all this conditional logic. I think the "pending" nature of existing requests is completely clear from the table above, and the "Request a new course" card is also super clear. Rather than describing the form in a paragraph of text, I suggest putting [help text](https://getbootstrap.com/docs/4.5/components/forms/#help-text) next to the various form inputs.

pages/instructorRequestCourse/instructorRequestCourse.ejs
```diff
@@ -0,0 +1,91 @@
...
+
+                        <div class="form-group">
+                            <label for="cr-shortname">Course Rubric and Number</label>
+                            <input type="text" class="form-control" name="cr-shortname" placeholder="XC 101" pattern="[a-zA-Z]+ [a-zA-Z0-9]+" title="this is a series of letters, followed by a space, followed by a series of numbers and/or letters" required>
```
mwest1066, line 72: Help text could be "Examples: MATH 101, PHYS 440". I suggest using a placeholder that is easily recognizable as subject/number. E.g., "MATH 101".

pages/instructorRequestCourse/instructorRequestCourse.ejs
```diff
@@ -0,0 +1,91 @@
...
+                        </div>
+                        <div class="form-group">
+                            <label for="cr-title">Course Title</label>
+                            <input type="text" class="form-control" name="cr-title" placeholder="Example Course" minlength="1" required>
```
mwest1066, line 76: Again, I suggest using a placeholder that is real, so that it's clear how to replace it. For example, "Introductory College Algebra" or something like that.

pages/instructorRequestCourse/instructorRequestCourse.ejs
```diff
@@ -0,0 +1,91 @@
...
+                        </div>
+                        <div class="form-group">
+                            <label for="cr-ghuser">GitHub Username (optional)</label>
+                            <input type="text" class="form-control" name="cr-ghuser">
```
mwest1066, line 80: Help text here could explain why the user might want to supply this and repeat that it's optional.

pages/instructorRequestCourse/instructorRequestCourse.sql
```diff
@@ -0,0 +1,48 @@
...
+                'status', r.approved_status,
+                'details',
+                    CASE
+                        WHEN r.approved_status = 'approved' THEN CONCAT('Approved by ', u.name)
+                        WHEN r.approved_status = 'denied'   THEN CONCAT('Denied by ', u.name)
+                        ELSE ''
+                    END,
+                'short_name', r.short_name,
+                'title', r.title)),
+        '[]'::jsonb) AS course_requests
```
mwest1066, line 16: I assume you are only using JSON here because you also want to return `has_one_course`? If you get rid of the complex form preamble then you can replace this with a simple SQL query (no JSON).

mwest1066:
Did you want to add the "make the course on GitHub" feature before merging this? If so, my suggestion would be to automatically make the course when the request is submitted, if:
1. The user already has Editor/Owner access to at least one course.
2. The have had made less than 5 course requests in the past 24 hours (or some similar rate limiter).

Whenever we create a course (manually or automatically as above) it would be great to post a message on slack, just so we stay aware of this activity. And of course whenever a non-automatic request is made, as well.

mwest1066:
Did you want to do Dave's trick for posting to slack? I'm a bit unclear about how that works but I'm happy to follow whatever you and Dave think best.

nicknytko:
> Did you want to do Dave's trick for posting to slack? I'm a bit unclear about how that works but I'm happy to follow whatever you and Dave think best.

Yeah, it looks like this is what the proctor stuff is doing so I’ll try to follow the same general flow (cc @trombonekenny )

pages/administratorOverview/administratorOverview.ejs
```diff
@@ -209,6 +209,63 @@
...
+
+      <div class="card mb-4">
+          <div class="card-header bg-primary text-white d-flex align-items-center">
+              Course Requests
+          </div>
+          <div class="table-responsive">
+              <table class="table table-sm table-hover table-striped">
+                  <thead>
+                      <tr>
+                          <th>Short name</th>
```
tdy, line 225:
```suggestion
                          <th>Short Name</th>
```
Nit (or uncapitalize the other headers?)

pages/instructorRequestCourse/instructorRequestCourse.ejs
```diff
@@ -0,0 +1,91 @@
...
+            <% if (course_requests.length > 0) { %>
+                <div class="card mb-4">
+                    <div class="card-header bg-primary text-white d-flex align-items-center">
+                        Course Requests
+                    </div>
+                    <div class="table-responsive">
+                        <table class="table table-sm table-hover table-striped">
+                            <thead>
+                                <tr>
+                                    <th>Short name</th>
```
tdy, line 32:
```suggestion
                                    <th>Short Name</th>
```

tdy:
Would also be good to update `docs/requestCourse.md`

PR #2674, opened 2020-07-03
Improve design and accessibility on the instructor file editor

### LINKED ISSUE #2575: Missing accent on nav bar over Files when in file editing mode
When overlooking the **Files** tab, there is a blue accent at the top of the tab: 

file tab accent present

After moving to "edit", the contents the accent disappears from the navigation bar. 

file tab accent mia

astrosticks:
Fixes #2575 and implements some more consistencies with button sizes and icon presence. Also, the "Show help" button wasn't displaying the correct text after showing + hiding the collapse once, but now it is.

| consistentColor |
| - |
| **Before** |
| image | 
| **After** |
| image |


pages/instructorFileEditor/instructorFileEditor.ejs
```diff
@@ -66,14 +78,24 @@
...
                         <div class="col-auto collapse <% if ((!fileEdit.failedPush) && (!fileEdit.alertChoice)) { %>show<% } %>" id="buttons-<%= fileEdit.uuid %>">
-                            <a class="btn btn-secondary" data-toggle="collapse" href="#help-<%= fileEdit.uuid %>" role="button" id="help-<%= fileEdit.uuid %>-button">Show help</a>
-                            <button id="file-editor-<%= fileEdit.uuid %>-save-button" name="__action" value="save_and_sync" class="btn btn-primary" disabled>Save and sync</button>
+                            <a id="help-<%= fileEdit.uuid %>-button"
+                               class="btn btn-light btn-sm"
+                               data-toggle="collapse" href="#help-<%= fileEdit.uuid %>"
+                               role="button" aria-labelledby="help-<%= fileEdit.uuid %>-button-label"
+                               aria-expanded="false">
+                                <i class="far fa-question-circle"></i>
+                                <span id="help-<%= fileEdit.uuid %>-button-label">Show help</span>
```
tdy, line 93:
```suggestion
                               role="button" aria-expanded="false">
                                <i class="far fa-question-circle"></i>
                                Show help
```
Actually we probably don't want `aria-labelledby` here. In general we should only use aria to encode info that's not already conveyed by native html semantics, and here the link already has a native "label" (its own link text).

pages/instructorFileEditor/instructorFileEditor.ejs
```diff
@@ -66,14 +78,24 @@
...
                         </div>
                         <div class="col-auto collapse <% if ((!fileEdit.failedPush) && (!fileEdit.alertChoice)) { %>show<% } %>" id="buttons-<%= fileEdit.uuid %>">
-                            <a class="btn btn-secondary" data-toggle="collapse" href="#help-<%= fileEdit.uuid %>" role="button" id="help-<%= fileEdit.uuid %>-button">Show help</a>
-                            <button id="file-editor-<%= fileEdit.uuid %>-save-button" name="__action" value="save_and_sync" class="btn btn-primary" disabled>Save and sync</button>
+                            <a id="help-<%= fileEdit.uuid %>-button"
+                               class="btn btn-light btn-sm"
+                               data-toggle="collapse" href="#help-<%= fileEdit.uuid %>"
+                               role="button" aria-labelledby="help-<%= fileEdit.uuid %>-button-label"
+                               aria-expanded="false">
+                                <i class="far fa-question-circle"></i>
```
tdy, line 92:
```suggestion
                                <i class="far fa-question-circle" aria-hidden="true"></i>
```

pages/instructorFileEditor/instructorFileEditor.ejs
```diff
@@ -66,14 +78,24 @@
...
+                                <span id="help-<%= fileEdit.uuid %>-button-label">Show help</span>
+                            </a>
+                            <button id="file-editor-<%= fileEdit.uuid %>-save-button" name="__action" value="save_and_sync" class="btn btn-light btn-sm" disabled>
+                                <i class="fas fa-save"></i>
```
tdy, line 96:
```suggestion
                                <i class="fas fa-save" aria-hidden="true"></i>
```

pages/instructorFileEditor/instructorFileEditor.ejs
astrosticks: Thx for the tip, testing it with NVDA, the label gets detected with or without `aria-labelledby`

pages/instructorFileEditor/instructorFileEditor.ejs
```diff
@@ -55,9 +55,16 @@
                     });
                 <% } %>
 
-                let showHelpElement = $("#help-<%= fileEdit.uuid %>-button");
-                showHelpElement.click(() => {
-                    showHelpElement.text(showHelpElement.text() == "Show help" ? "Hide help" : "Show help");
+                var helpBox = $('#help-<%= fileEdit.uuid %>');
+                var helpButton =  $('#help-<%= fileEdit.uuid %>-button');
+                var helpButtonLabel = $('#help-<%= fileEdit.uuid %>-button-label');
```
nicknytko, line 60: Tiniest of nits, but I dislike using `var` in new code

```suggestion
                const helpBox = $('#help-<%= fileEdit.uuid %>');
                const helpButton =  $('#help-<%= fileEdit.uuid %>-button');
                const helpButtonLabel = $('#help-<%= fileEdit.uuid %>-button-label');
```

tiny nit

echuber2:
Is the difference that one is the "breadcrumbs" version and one is just the path? The `/` in the breadcrumbs bar was previously a filled circle.

mwest1066:
I think with the "exploded" view that at least some of the directory names are links (maybe just the right-most ones?). I agree that it would be good to have them the same. I guess I'd go with the exploded view, because that's the one that's more functional (contains links).

lib/getPaths.js
```diff
@@ -0,0 +1,144 @@
+const path = require('path');
+const { encodePath, decodePath } = require('./uri-util');
+
+module.exports = {};
+
+function contains(parentPath, childPath) {
+    const relPath = path.relative(parentPath, childPath);
+    return (!(relPath.split(path.sep)[0] == '..' || path.isAbsolute(relPath)));
+}
+
```
nicknytko, line 10: Can we add a jsdoc comment before this function explaining what it does and what the arguments/return values are?  I think we are losing some contextual information when we move this out into another file.

astrosticks: Done: ff93240

pages/instructorFileEditor/instructorFileEditor.ejs
```diff
@@ -87,28 +94,50 @@
...
-                            You are editing the file <code><%= fileEdit.fileNameForDisplay %></code>. To save changes, click <strong>Save and sync</strong> or use <strong>Ctrl-S</strong> (Windows/Linux) or <strong>Cmd-S</strong> (Mac).
-                            <% if (config.fileEditorUseGit) { %>
-                            Doing so will write your changes to disk, will push them to the remote GitHub repository, and will sync them to the database.
-                            <% } else { %>
-                            Doing so will write your changes to disk and will sync them to your local database. You will need to push these changes to the GitHub respository manually (i.e., not in PrairieLearn), if desired.
-                            <% } %>
-                            If you reload or navigate away from this page, any unsaved changes will be lost.
-                        </p>
+                <div class="collapse" id="help-<%= fileEdit.uuid %>">
+                    <div class="card-body">
```
nwalters512, line 133: By moving the `card-body` class to a div nested within the `collapse` div, there won't be a jumpy animation as it opens and closes.

nwalters512 (approved):

PR #2698, opened 2020-07-09
Remove urlPrefix from workspaceUrl

tdy:
Removes `/pl/[garbage]` from `workspaceUrl`.

### DIFF
--- a/lib/question.js
+++ b/lib/question.js
@@ -1118,7 +1118,7 @@ module.exports = {
         }
 
         if (variant.workspace_id) {
-          urls.workspaceUrl = `${urlPrefix}/workspace/${variant.workspace_id}`;
+          urls.workspaceUrl = `/workspace/${variant.workspace_id}`;
         }
 
         return urls;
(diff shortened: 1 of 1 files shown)


lib/question.js
```diff
@@ -1118,7 +1118,7 @@ module.exports = {
         }
 
         if (variant.workspace_id) {
-          urls.workspaceUrl = `${urlPrefix}/workspace/${variant.workspace_id}`;
+          urls.workspaceUrl = `/workspace/${variant.workspace_id}`;
```
tdy, line 1121:
```suggestion
          urls.workspaceUrl = `/pl/workspace/${variant.workspace_id}`;
```
Not sure if it should also have `/pl`?

lib/question.js
```diff
@@ -1118,7 +1118,7 @@ module.exports = {
         }
 
         if (variant.workspace_id) {
-          urls.workspaceUrl = `${urlPrefix}/workspace/${variant.workspace_id}`;
+            urls.workspaceUrl = `/workspace/${variant.workspace_id}`;
```
tdy, line 1121:
```suggestion
            urls.workspaceUrl = `/pl/workspace/${variant.workspace_id}`;
```
Not sure if it should also have `/pl`?

PR #2702, opened 2020-07-09
add html display example

mfsilva22:
This illustrates two ways of displaying html tables that are generated from data in `server.py`

image

I am not a big fan of the second approach, but it seems to be useful to some instructors. I could only align the rows by removing any indentation on the `<pre>` block. Is there another way?

Open to suggestions!


mwest1066:
> I am not a big fan of the second approach, but it seems to be useful to some instructors. I could only align the rows by removing any indentation on the `<pre>` block. Is there another way?

When you say "second approach", do you mean the <pre> block?

I don't know of any way to do a <pre> block without removing the indentation on it.

exampleCourse/questions/demo/addHTMLtable/question.html
```diff
@@ -0,0 +1,36 @@
...
+</div>
+
+<div class="card my-2">
+    <div class="card-header">
+        Example Table 3: get table from console output
+    </div>
+    <div class="card-body">
+<pre>
+{{params.mytable3}}
+</pre>
```
mwest1066, line 32:
```suggestion
<pre>{{params.mytable3}}</pre>
```

This will remove the extra blank line at the end. It also means you can intent the whole thing, I think.

mfsilva22: I will try that.

mfsilva22:
> > I am not a big fan of the second approach, but it seems to be useful to some instructors. I could only align the rows by removing any indentation on the `<pre>` block. Is there another way?
> 
> When you say "second approach", do you mean the
> 
>  block?
> 
> I don't know of any way to do a 
> 
>  block without removing the indentation on it.

I just meant that I don't like the implementation using print statement. I think it can be confusing to some, and it does not display as nice, but I decided to add it to the example, to make sure others can benefit from it.

jjlh99:
Matt: I posted an example over on the Slack thread.   But in looking at some of the overall discussion, I am beginning to think that one could refine the data frame approach to make it easy for the end user AND get nice results.   I think we can assume that any user can figure out how to dump their data in a dataframe.  I would then do the following
1) implement the suggestions to optionally suppress row or column indices.

2) provide a ```df_clean``` function which
    a) if data is string, Latex etc - just leave it alone and pass through
    b) if all column data is float format (except header), check magnitude of ALL numbers in a columns and use some simple 
        logic like  (really big or really small use e format else use d format), sort of like
              ``` if  -3 < log_10 (max mag) < 3   convert to string in e format with desired # of figs```
             ```      else   convert to string in d  format  with desired # of figs```

I think this approach handles all of the issues I was thinking about, gives cleaner coding and more elegant output.  It also allows users to include nice brief Latex in their headers.

This isn't so much of a formal request as an observation.  You all have shown me enough that I now know how I can do it in my server.py files even if it's not the preferred approach for formal implementation.

Many thanks to all who have helped on this discussion.  Jon

mwest1066:
> 1. implement the suggestions to optionally suppress row or column indices.
> 2. provide a `df_clean` function which
>    a) if data is string, Latex etc - just leave it alone and pass through
>    b) if all column data is float format (except header), check magnitude of ALL numbers in a columns and use some simple
>    logic like  (really big or really small use e format else use d format), sort of like
>    ` if  -3 < log_10 (max mag) < 3   convert to string in e format with desired # of figs`
>    `      else   convert to string in d  format  with desired # of figs`

I like this! @nicknytko may we could make an issue to track this?

mfsilva22:
> > 1. implement the suggestions to optionally suppress row or column indices.
> > 2. provide a `df_clean` function which
> >    a) if data is string, Latex etc - just leave it alone and pass through
> >    b) if all column data is float format (except header), check magnitude of ALL numbers in a columns and use some simple
> >    logic like  (really big or really small use e format else use d format), sort of like
> >    ` if  -3 < log_10 (max mag) < 3   convert to string in e format with desired # of figs`
> >    `      else   convert to string in d  format  with desired # of figs`
> 
> I like this! @nicknytko may we could make an issue to track this?

Just to clarify. The comment from @jjlh99 is related to other suggestions from the slack, when using pandas and `pl-python-variable`. This example illustrates how to display html tables, not necessarily coming from pandas dataframe (because I think that is already illustrated in other examples).

exampleCourse/questions/demo/addHTMLtable/question.html
```diff
@@ -0,0 +1,34 @@
...
+          Example Table 1:
+      </div>
+      <div class="card-body">
+          {{{params.mytable}}}
+      </div>
+ </div>
+
+ <div class="card my-2">
+     <div class="card-header">
+         Example Table 2: adding table properties
```
nicknytko, line 16:
```suggestion
         Example Table 2: Adding table properties
```

exampleCourse/questions/demo/addHTMLtable/question.html
```diff
@@ -0,0 +1,34 @@
...
+     <div class="card-body">
+         <table style="width:30%" border="black">
+           {{{params.mytable2}}}
+         </table>
+     </div>
+</div>
+
+<div class="card my-2">
+    <div class="card-header">
+        Example Table 3: get table from console output
```
nicknytko, line 27:
```suggestion
        Example Table 3: Getting a table from console output
```

exampleCourse/questions/demo/addHTMLtable/question.html
```diff
@@ -0,0 +1,34 @@
...
+          {{{params.mytable}}}
+      </div>
+ </div>
+
+ <div class="card my-2">
+     <div class="card-header">
+         Example Table 2: adding table properties
+     </div>
+     <div class="card-body">
+         <table style="width:30%" border="black">
```
nicknytko, line 19: What does this look like if we use the bootstrap table themes? 

```
<table class="table table-sm table-bordered">
```

This is probably fine how it is but I prefer how the bootstrap tables look over the normal html ones.

echuber2:
With respect to the comments about suppressing headers on dataframes, I made a PR about that: https://github.com/PrairieLearn/PrairieLearn/pull/2703 It extends `pl-python-variable`. The idea could be taken further except that the dataframe does a round trip through `to_json` and `from_json` before finally invoking Pandas' own `to_html` function, so passing through even more of the options that `to_html` supports makes me think there should be an easier interface perhaps. Maybe a way to let people directly specify the arguments they want for `to_html` as another `param` that the variable element could receive.

exampleCourse/questions/demo/addHTMLtable/question.html
mfsilva22: I am okay either way. I was trying to have something very simple. Feel free to make changes for style, but keeping the simplicity.

nicknytko: You're right that this should be a simple example.  I think it is fine as-is.

exampleCourse/questions/demo/addHTMLtable/server.py
```diff
@@ -0,0 +1,29 @@
...
+def generate(data):
+
+    name_list = ["Carla", "Manoel", "Sam", "Laura", "Amanda"]
+    age_list =  [24, 18, 30, 45, 32]
+    name_greek = ["$\\gamma$", "$\\mu$", "$\\sigma$", "$\\lambda$", "$\\alpha$" ]
+
+    # Generate complete html string
+    mytable = '<table style="width:30%"><tr><th> Name </th> <th> Age </th> <th> Id </th></tr>'
+    for i in range(len(age_list)):
+        mytable += '<tr><td>'+name_list[i]+'</td><td>'+str(age_list[i])+'</td><td>'+name_greek[i]+'</td><tr>'
```
nicknytko, line 13: What do you think of using `zip()` here? i.e.
```python
for name, age, name_greek in zip(name_list, age_list, name_greek):
```
Totally minor nit, but I think it looks a little cleaner

mfsilva22: haha. Sure. Looks a lot better. Thank you :-)

Would you be able to make these changes and merge? It would help me a lot (getting behind with all the workshop stuff :-( )

PR #2703, opened 2020-07-09
New options on pl-python-variable to hide DataFrame labels

echuber2:
New options for the `pl-python-variable` element to let you hide header labels of various kinds on a Pandas DataFrame. Includes example and doc edits.

It seemed like there should also be an option to hide the overflow horizontal scrollbar if the dataframe is very small, just for aesthetics, but decided to leave that out. People can do that with a CSS hack in their question HTML if they want it.

### DIFF
--- a/exampleCourse/questions/element/pythonVariable/question.html
+++ b/exampleCourse/questions/element/pythonVariable/question.html
@@ -12,7 +12,7 @@
   <pl-code source-file-name="test.py"></pl-code>
   <pl-python-variable params-name="df"></pl-python-variable>
   <p>
-    By default, a Pandas DataFrame will have show its index column and header row, as well as a postscript label with the table dimensions. You can toggle these with attributes on the <code>pl-python-variable</code> tag. Here, we've added <code>show-header="false" show-index="false" show-dimensions="false"</code>:
+    By default, a Pandas DataFrame will display its index column and header row, as well as a postscript label with the table dimensions. You can toggle these with attributes on the <code>pl-python-variable</code> tag. Here, we've added <code>show-header="false" show-index="false" show-dimensions="false"</code>:
   </p>
   <pl-python-variable params-name="df" show-header="false" show-index="false" show-dimensions="false"></pl-python-variable>
   <p>
--- a/exampleCourse/questions/element/pythonVariable/question.html
+++ b/exampleCourse/questions/element/pythonVariable/question.html
@@ -12,7 +12,11 @@
   <pl-code source-file-name="test.py"></pl-code>
   <pl-python-variable params-name="df"></pl-python-variable>
   <p>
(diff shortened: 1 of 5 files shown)


echuber2:
Copying my rambling from the other PR thread:

> The idea could be taken further except that the dataframe does a round trip through `to_json` and `from_json` before finally invoking Pandas' own `to_html` function, so passing through even more of the options that `to_html` supports makes me think there should be an easier interface perhaps. Maybe a way to let people directly specify the arguments they want for `to_html` as another param that the variable element could receive.

exampleCourse/questions/element/pythonVariable/question.html
```diff
@@ -12,7 +12,11 @@
   <pl-code source-file-name="test.py"></pl-code>
   <pl-python-variable params-name="df"></pl-python-variable>
   <p>
-    Here is the same example with <code>text="true"</code>, forcing the variable to be rendered in its textual representation:
+    By default, a Pandas DataFrame will have show its index column and header row, as well as a postscript label with the table dimensions. You can toggle these with attributes on the <code>pl-python-variable</code> tag. Here, we've added <code>show-header="false" show-index="false" show-dimensions="false"</code>:
+  </p>
+  <pl-python-variable params-name="df" show-header="false" show-index="false" show-dimensions="false"></pl-python-variable>
+  <p>
+    Here is the same example with <code>text="true"</code>, forcing the variable to be rendered in its textual representation. (The <code>show-header</code>, <code>show-index</code>, and <code>show-dimensions</code> attributes don't affect this.)
```
nicknytko, line 19:
```suggestion
    Here is the same example with <code>text="true"</code>, forcing the variable to be rendered in its textual representation. The <code>show-header</code>, <code>show-index</code>, and <code>show-dimensions</code> attributes don't affect this.
```

exampleCourse/questions/element/pythonVariable/question.html
```diff
@@ -12,7 +12,11 @@
   <pl-code source-file-name="test.py"></pl-code>
   <pl-python-variable params-name="df"></pl-python-variable>
   <p>
-    Here is the same example with <code>text="true"</code>, forcing the variable to be rendered in its textual representation:
+    By default, a Pandas DataFrame will have show its index column and header row, as well as a postscript label with the table dimensions. You can toggle these with attributes on the <code>pl-python-variable</code> tag. Here, we've added <code>show-header="false" show-index="false" show-dimensions="false"</code>:
```
nicknytko, line 15:
```suggestion
    By default, a Pandas DataFrame will display its index column and header row, as well as a postscript label with the table dimensions. You can toggle these with attributes on the <code>pl-python-variable</code> tag. Here, we've added <code>show-header="false" show-index="false" show-dimensions="false"</code>:
```

PR #2705, opened 2020-07-10
Add option to not focus the file editor

echuber2:
I'm making a question with multiple embedded file editors that are halfway down the page. The 2nd editor was stealing focus and dragging the page scroll way down. I think it's very disorienting to the student. Also, if there are multiple editors in one question, it's confusing to focus the last one.

The culprit was:

```
    // this.editor.focus();
```

I left the line commented out with an explanatory comment to discourage someone from adding it back again. Things seem to work fine without the focus grab.

### DIFF
--- a/elements/pl-file-editor/pl-file-editor.js
+++ b/elements/pl-file-editor/pl-file-editor.js
@@ -44,6 +44,12 @@ window.PLFileEditor = function(uuid, options) {
         this.editor.setOption('maxLines', Infinity);
     }
 
+    if (options.plOptionFocus) {
+        this.plOptionFocus = true;
+    } else {
+        this.plOptionFocus = false;
+    }
+
     if (options.preview == 'markdown') {
         let renderer = new showdown.Converter();
 
@@ -108,10 +114,9 @@ window.PLFileEditor.prototype.initRestoreOriginalButton = function() {
 window.PLFileEditor.prototype.setEditorContents = function(contents) {
     this.editor.setValue(contents);
     this.editor.gotoLine(1, 0);
-    // Let's NOT set the focus here, because it can drag the window scroll
-    // down the page. Also, if there are multiple editors in one question,
-    // it's confusing to focus the last one.
-    // this.editor.focus();
+    if (this.plOptionFocus) {
+        this.editor.focus();
+    }
     this.syncFileToHiddenInput();
 };
 
--- a/elements/pl-file-editor/pl-file-editor.js
+++ b/elements/pl-file-editor/pl-file-editor.js
@@ -44,11 +44,7 @@ window.PLFileEditor = function(uuid, options) {
         this.editor.setOption('maxLines', Infinity);
     }
 
-    if (options.plOptionFocus) {
-        this.plOptionFocus = true;
-    } else {
-        this.plOptionFocus = false;
-    }
+    this.plOptionFocus = !!options.plOptionFocus;
 
     if (options.preview == 'markdown') {
         let renderer = new showdown.Converter();
--- a/elements/pl-file-editor/pl-file-editor.js
+++ b/elements/pl-file-editor/pl-file-editor.js
@@ -44,7 +44,7 @@ window.PLFileEditor = function(uuid, options) {
         this.editor.setOption('maxLines', Infinity);
     }
 
-    this.plOptionFocus = !!options.plOptionFocus;
+    this.plOptionFocus = options.plOptionFocus;
 
     if (options.preview == 'markdown') {
         let renderer = new showdown.Converter();
--- a/elements/pl-file-editor/pl-file-editor.js
+++ b/elements/pl-file-editor/pl-file-editor.js
@@ -108,7 +108,10 @@ window.PLFileEditor.prototype.initRestoreOriginalButton = function() {
 window.PLFileEditor.prototype.setEditorContents = function(contents) {
     this.editor.setValue(contents);
     this.editor.gotoLine(1, 0);
-    this.editor.focus();
+    // Let's NOT set the focus here, because it can drag the window scroll
+    // down the page. Also, if there are multiple editors in one question,
+    // it's confusing to focus the last one.
+    // this.editor.focus();
     this.syncFileToHiddenInput();
 };
 
--- a/elements/pl-file-editor/pl-file-editor.mustache
+++ b/elements/pl-file-editor/pl-file-editor.mustache
@@ -11,7 +11,8 @@ $(function() {
         minLines: {{#min_lines}}{{min_lines}}{{/min_lines}}{{^min_lines}}undefined{{/min_lines}},
         maxLines: {{#max_lines}}{{max_lines}}{{/max_lines}}{{^max_lines}}undefined{{/max_lines}},
         autoResize: {{#auto_resize}}{{auto_resize}}{{/auto_resize}}{{^auto_resize}}undefined{{/auto_resize}},
-        preview: {{#preview}}"{{preview}}"{{/preview}}{{^preview}}undefined{{/preview}}
+        preview: {{#preview}}"{{preview}}"{{/preview}}{{^preview}}undefined{{/preview}},
+        plOptionFocus: {{#focus}}{{focus}}{{/focus}}{{^focus}}undefined{{/focus}}
     });
 });
 </script>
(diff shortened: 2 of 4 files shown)


nicknytko:
Could we make this an attribute to enable the editor focus on page load (default True?)  This seems to only be an issue when there are multiple editors on the page.

Or, what if we try something "smart" and only let the first editor focus?

echuber2:
I'll work on making it an option attribute then. I think the logic to resolve which editor is "first" may be more trouble than it's worth, or there may be interesting cases where it's not desired. (In my case, the first editor appears halfway down the page after some instructions, so I'd rather not drag the page scroll down by focusing the first editor either.)

echuber2:
@nicknytko I've made it optional now with a new attribute on the tag and added documentation.

The default is `focus="true"` but I kind of feel like the default should be `false` instead. As it is currently with default being `true`, if you want to focus the first editor out of several, you have to do this:

```
<pl-file-editor ... focus="true"> ... </pl-file-editor>
<pl-file-editor ... focus="false"> ... </pl-file-editor>
<pl-file-editor ... focus="false"> ... </pl-file-editor>
```

If the default were `false`, then you could focus the first editor like this:

```
<pl-file-editor ... focus="true"> ... </pl-file-editor>
<pl-file-editor ... > ... </pl-file-editor>
<pl-file-editor ... > ... </pl-file-editor>
```

However, I don't have a strong preference either way, as long as we can merge this soon :-)

echuber2:
As an aside, it seems like it's better for accessibility to not focus the editor by default when the question loads.

elements/pl-file-editor/pl-file-editor.js
```diff
@@ -44,6 +44,12 @@ window.PLFileEditor = function(uuid, options) {
         this.editor.setOption('maxLines', Infinity);
     }
 
+    if (options.plOptionFocus) {
+        this.plOptionFocus = true;
+    } else {
+        this.plOptionFocus = false;
+    }
```
mwest1066, line 51:
```suggestion
    this.plOptionFocus = !!options.plOptionFocus;
```

The double-bang forces the variable to a boolean, but it's not actually needed here as far as I can see.

elements/pl-file-editor/pl-file-editor.mustache
```diff
@@ -11,7 +11,8 @@ $(function() {
         minLines: {{#min_lines}}{{min_lines}}{{/min_lines}}{{^min_lines}}undefined{{/min_lines}},
         maxLines: {{#max_lines}}{{max_lines}}{{/max_lines}}{{^max_lines}}undefined{{/max_lines}},
         autoResize: {{#auto_resize}}{{auto_resize}}{{/auto_resize}}{{^auto_resize}}undefined{{/auto_resize}},
-        preview: {{#preview}}"{{preview}}"{{/preview}}{{^preview}}undefined{{/preview}}
+        preview: {{#preview}}"{{preview}}"{{/preview}}{{^preview}}undefined{{/preview}},
+        plOptionFocus: {{#focus}}{{focus}}{{/focus}}{{^focus}}undefined{{/focus}}
```
mwest1066, line 15:
```suggestion
        plOptionFocus: {{focus}}
```

From the code it looks like `focus` must always be the string `"true"` or `"false"`, so we don't need to test here. The same comment actually applies above to `autoResize`.

echuber2: Would it be preferable to also simplify the `autoResize` case then? I was just trying to follow the existing style.

elements/pl-file-editor/pl-file-editor.js
echuber2: It may not be needed (I was trying to be explicit while debugging). Should we just take it out?

nicknytko: If it's not needed then lets take it out

elements/pl-file-editor/pl-file-editor.mustache
nicknytko: I'm all for cleaning up the existing code, I'm guessing that many of the attributes above can be simplified as well.

echuber2: We can remove the extra cases from the mustache but this will leave the burden on the Python side to sanitize all the variables adequately. Is this how it is for most of the other elements or would someone in the future get in trouble without these extra checks?

nicknytko: I would think anything coming from the Python side should be sanitized anyway, since we check attributes and assign a default if they aren't defined.

elements/pl-file-editor/pl-file-editor.js
echuber2: Okay, I took out the `!!` as well then.

elements/pl-file-editor/pl-file-editor.mustache
echuber2: I simplified `plOptionFocus` and `autoResize` in this way but I was hesitant for the others without having the time to test them thoroughly. Some of them have `Νone` as the default on the Python side, which would cause the quoted values to be set as `"None"` (e.g. `aceMode: "None"`) and I'm not sure how well Ace Editor even tolerates that now:

```javascript
    if (options.aceMode) {
        this.editor.getSession().setMode(options.aceMode);
    }
```

This is doing `setMode("None")` in that case. Apparently it has been working. Should this be something like:

```javascript
    if (options.aceMode && options.aceMode != "None") {
        this.editor.getSession().setMode(options.aceMode);
    }
```

echuber2: It looks like Ace Editor ends up expecting the default to be `undefined` or `false` and consequently sets the mode to `"ace/mode/text"`:

https://github.com/ajaxorg/ace/blob/5524d248e3f82a70539bc0fbcded626b1c48b7f7/lib/ace/edit_session.js#L861

Since we set it to `"None"` when not specified, it will end up trying to load the module from Ace anyway and have an error that gets discarded and the mode is set to plain text anyway. It's good that Ace handles this gracefully but I think we could just avoid passing it options that it will fail on in that case. Maybe a separate issue to consider though?

nicknytko: No worries, I think that this already looks nicer.

echuber2: OK, I'll let it be for now then.

nicknytko:
Was there some discussion on whether we should focus by default or not?  Did we ever come to a conclusion on this?

echuber2:
> Was there some discussion on whether we should focus by default or not? Did we ever come to a conclusion on this?

I didn't get to hear @tdy's thoughts about it yet, but CSID Sebastian agreed with me that it should default to not stealing focus. Maybe I can try to join the dev meeting tomorrow and bring it up. Do other input elements focus by default?

tdy:
My $0.02 -- given that `pl-file-editor` will usually have some kind of preamble, I would not autofocus by default. It's disruptive for screen reader users and keyboard users (or even people who aren't using assistive tech if the question prompt is sizable like in Eric's case).

nicknytko:
If all of the CSID's are agreeing that it is better for accessibility then I think we should just default to not focusing the file editor on page load &mdash;  this will probably be consistent with all of the other input elements as well.  Though if you want, we can discuss this at today's dev meeting.

elements/pl-file-editor/pl-file-editor.py
```diff
@@ -12,8 +12,9 @@
 SOURCE_FILE_NAME_DEFAULT = None
 MIN_LINES_DEFAULT = None
 MAX_LINES_DEFAULT = None
-AUTO_RESIZE_DEFAULT = 'true'
+AUTO_RESIZE_DEFAULT = True
 PREVIEW_DEFAULT = None
+FOCUS_DEFAULT = True
```
nicknytko, line 17: lets actually just set this to false 😀

echuber2: I will, need to edit the pending docs too

echuber2:
I changed the default to false and updated the docs. One thing that comes to mind after writing the updated docs: It's possible that someone will have a good use case for opting to steal the focus, but if it's strongly discouraged to do it, it might make sense to not expose the option. But making it optional like this is a big improvement.

PR #2710, opened 2020-07-11
Change coveralls script to not block CI

tdy:
Based on the dev meeting convo -- prevent coveralls from blocking CI.

Travis does have an [**allow_failures**](https://docs.travis-ci.com/user/customizing-the-build#jobs-that-are-allowed-to-fail) option, but I couldn't tell exactly how to use it in our config.

### DIFF
--- a/docker/coveralls.sh
+++ b/docker/coveralls.sh
@@ -2,3 +2,6 @@
 
 cd /PrairieLearn
 cat coverage/lcov.info | ./node_modules/.bin/coveralls -v
+
+# do not block CI even if coveralls fails
+exit 0
--- a/docker/coveralls.sh
+++ b/docker/coveralls.sh
@@ -1,7 +1,4 @@
 #! /bin/bash
 
 cd /PrairieLearn
-cat coverage/lcov.info | ./node_modules/.bin/coveralls -v
-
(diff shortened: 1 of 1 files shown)


echuber2:
Side note: if you need the shell script to end with code 0 and it's something that might be loaded with `source` or `.` in an existing shell session you don't want to exit, you can avoid `exit 0` by ending the script with `true` or an `echo`, etc.

mwest1066:
This PR still had a coveralls failure in CI, so it doesn't look like this fix is sufficient? Perhaps github actions is running the script with the `-e` flag in bash? Two possibilities would be to either use `|| true` after the command, or change the [workflow file](https://github.com/PrairieLearn/PrairieLearn/blob/master/.github/workflows/main.yml) to use the [`continue-on-error` flag](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error) for the coveralls job.

mwest1066:
> Travis does have an [**allow_failures**](https://docs.travis-ci.com/user/customizing-the-build#jobs-that-are-allowed-to-fail) option, but I couldn't tell exactly how to use it in our config.

Note that we are using GitHub Actions, not Travis.

PR #2715, opened 2020-07-13
Fix file upload when editing a course question

mwest1066:
At the moment we can't upload files in questions when they are viewed in the context of a course. It works fine when viewing in a course instance. This is a problem for users who have course-level Editor access but not access to any course instance.


mwest1066:
I verified that this fixes the problem on `pl-dev`.

PR #2717, opened 2020-07-13
Allow course "viewers" to copy questions

### LINKED ISSUE #2685: Course "viewers" should be able to copy questions
Needed for the exampleCourse: instructors that are only "viewers" should be able to copy questions to their own courses in which they are "editor" or above.

nicknytko:
Resolves #2685 

Turns out this was as simple as updating the interface, the server code already was allowing users to copy questions as long as they has edit permission in the destination.

I was able to test this locally by adding course permissions to the dev user, then dropping dev user from the admin table.

### DIFF
--- a/pages/instructorQuestionSettings/instructorQuestionSettings.ejs
+++ b/pages/instructorQuestionSettings/instructorQuestionSettings.ejs
@@ -102,11 +102,12 @@
           We only want to show the card footer if it will have something inside. Since it is easy
           to make a mistake when merging logical conditions, we will simply "or" them all.
         -->
-        <% if ((authz_data.has_course_permission_edit && ((! course.options.isExampleCourse) || (courses && (courses.length > 1)))) ||
+        <% if (((authz_data.has_course_permission_edit && (!course.options.isExampleCourse || (course && courses.length > 1))) || (!authz_data.has_course_permission_edit && courses && courses.length > 0)) ||
                ((authz_data.has_course_permission_edit) && (! course.options.isExampleCourse))) { %>
         <div class="card-footer">
           <div class="row">
-            <% if (authz_data.has_course_permission_edit && ((! course.options.isExampleCourse) || (courses && (courses.length > 1)))) { %>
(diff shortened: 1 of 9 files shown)


pages/instructorQuestionSettings/instructorQuestionSettings.ejs
```diff
@@ -102,11 +102,12 @@
...
           to make a mistake when merging logical conditions, we will simply "or" them all.
         -->
-        <% if ((authz_data.has_course_permission_edit && ((! course.options.isExampleCourse) || (courses && (courses.length > 1)))) ||
+        <% if (((authz_data.has_course_permission_edit && (!course.options.isExampleCourse || (courses && courses.length > 1))) || (!authz_data.has_course_permission_edit && courses && courses.length > 0)) ||
                ((authz_data.has_course_permission_edit) && (! course.options.isExampleCourse))) { %>
         <div class="card-footer">
           <div class="row">
-            <% if (authz_data.has_course_permission_edit && ((! course.options.isExampleCourse) || (courses && (courses.length > 1)))) { %>
+            <% if ((authz_data.has_course_permission_edit && (!course.options.isExampleCourse || (courses && courses.length > 1))) || (!authz_data.has_course_permission_edit && courses && courses.length > 0)) { %>
+            <!-- If we can edit this course, can we copy this question to this course or another course?  If we cannot edit this course, can we copy this question to another course? -->
```
mwest1066, line 110: Can we write all these logical conditions more like this:
```
if (
    /* Do we have permission to copy to this course? */
    (authz_data.has_course_permission_edit && ! course.options.isExampleCourse)
    ||
    /* Do we have permission to copy to some other course? */
    (courses && courses.length > 0)
) {
```

(No need to use exactly this formatting, I was just writing it out for myself).

I don't quite understand why the original code had an `||` condition in the first `if` statement -- the second clause appears to be a subset of the first?

nicknytko: I'm not sure your expression is logically equivalent -- if we can edit the current course but it is the example course, then we should only show the button if we can edit one other course apart from the example course.  The second clause should only be true if are *not* able to edit the current course.

nicknytko:
Should modify `sprocs/courses_user_can_edit.sql` to not include example courses.

PR #2722, opened 2020-07-15
Auto-expansion for Title/QID columns of instructor questions table.

lumetta:
This is the expansion I mentioned at the workshop today, Matt (14 July '20).  I debated making it case-insensitive, but the QID path names are case sensitive, so I left the expansion working the same way.  Try it and see if it feels natural to you.  (As you probably know, there's a slight delay before the filters apply, so you can type a few characters and then wait for it.)

### DIFF
--- a/pages/instructorQuestions/instructorQuestions.ejs
+++ b/pages/instructorQuestions/instructorQuestions.ejs
@@ -15,6 +15,114 @@
                 sanitize: false
             });
         });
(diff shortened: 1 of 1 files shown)


nwalters512:
This hasn't seen any activity in over two years; closing. Feel free to reopen a new rebased PR if you'd still like to contribute this!

PR #2733, opened 2020-07-16
Add attribute to <pl-multiple-choice> to pull correct and/or incorrect answers from a json file

bojinyao:
Continue from #2623 

`external-json` is an *optional* attribute to `<pl-multiple-choice>`, type string, default to `None`

User can pass a json file path to this attribute, preferably a relatively path to *where question.html lives*; ~but no check is done whether the path is absolute or relative.~ or an absolute path that might look like: `/PrairieLearn/course/etc/file.json`

Exceptions will be thrown ~by python 3 `pathlib` library~ when file is not found. And `json` library when the file is not correctly formatted.

By default, the target json file should have 2 attributes: "correct", and "incorrect", both of which are json lists containing text that will be added to poll of correct and incorrect answers already in question.html. Below is an example of what the json might look like:

```json
{
    "correct": [
        "A true statement"
    ],
    "incorrect": [
        "Incorrect statement 1",
        "Incorrect statement 2"
    ]
}
```

Alternatively, user can pass in the name of attributes representing "correct" or "incorrect" using `external-json-correct-key` and `external-json-incorrect-key` respectively to override the default attribute names.

~Another idea that's not implemented is to have the user *optionally* set the names of attributes corresponding to `"correct"` and `"incorrect"` when `"external-json"` is set. This might make it more flexible to use.~

The inspiration for this additional attribute is to make it easier for question writers to organize their bank of questions since `question.html` is not exactly portable or easily-parsed. Additionally, this allows the possibility for question writers to group banks of questions in a separate, but standalone, directory that would be easier to manage.

Fixes https://github.com/ace-lab/pl-ucb-cs10/issues/43

Attribute like this one can also be extended to `<pl-checkbox>`


bojinyao:
Resolve https://github.com/ace-lab/pl-ucb-cs10/issues/43

nicknytko:
Does this also include the changes from #2623 ? Should we close out that PR in favor of this one?

bojinyao:
It does include the changes from the other PR, but I'm still developing this one, and would like to hear some inputs from other pl devs. I think it'd be great if we can close out #2623 first. This one doesn't involve significant code changes.

docs/elements.md
```diff
@@ -116,6 +116,9 @@ Attribute | Type | Default | Description
 `hide-letter-keys` | boolean | false | Hide the letter keys in the answer list, i.e., (a), (b), (c), etc.
 `all-of-the-above` | boolean | false | Add "All of the above" choice below all answer choices, but above "None of the above" if enabled. Bounded by `number-answers` and not affected by `fixed-order`.
 `none-of-the-above` | boolean | false | Add "None of the above" choice below all answer choices regardless of `fixed-order`, and is bounded by `number-answers`.
+`external-json` | string | special | Optional [abs,rel]path to a json file that stores choices as lists under "correct" and "incorrect" attribute names.
+`external-json-correct` | string | special | Optionally override default json "correct" attribute name when using `external-json` file.
+`external-json-incorrect` | string | special | Optionally override default json "incorrect" attribute name when using `external-json` file.
```
nicknytko, line 121: Can we instead name these `external-json-correct-key` and `external-json-incorrect-key` so it's more obvious from a glance what they are?

docs/elements.md
```diff
@@ -116,6 +116,9 @@ Attribute | Type | Default | Description
 `hide-letter-keys` | boolean | false | Hide the letter keys in the answer list, i.e., (a), (b), (c), etc.
 `all-of-the-above` | boolean | false | Add "All of the above" choice below all answer choices, but above "None of the above" if enabled. Bounded by `number-answers` and not affected by `fixed-order`.
 `none-of-the-above` | boolean | false | Add "None of the above" choice below all answer choices regardless of `fixed-order`, and is bounded by `number-answers`.
+`external-json` | string | special | Optional [abs,rel]path to a json file that stores choices as lists under "correct" and "incorrect" attribute names.
```
nicknytko, line 119:
```suggestion
`external-json` | string | special | Optional path to a JSON file to load external answer choices from.  Answer choices are stored as lists under "correct" and "incorrect" key names.
```

elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -28,17 +33,36 @@ def categorize_options(element):
...
+    file_path = pl.get_string_attrib(element, 'external-json', EXTERNAL_JSON_DEFAULT)
+    if file_path is not EXTERNAL_JSON_DEFAULT:
+        correct_attrib = pl.get_string_attrib(element, 'external-json-correct', EXTERNAL_JSON_DEFAULT_CORRECT)
+        incorrect_attrib = pl.get_string_attrib(element, 'external-json-incorrect', EXTERNAL_JSON_DEFAULT_INCORRECT)
+        if pathlib.PurePath(file_path).is_absolute():
+            json_file = file_path
+        else:
+            json_file = pathlib.PurePath(data['options']['question_path']).joinpath(file_path)
+        with open(json_file, mode='r', encoding='utf-8') as f:
+            obj = json.load(f)
```
nicknytko, line 46: Maybe lets be very explicit here and throw our own error if the file is not found.  Something like `external JSON answer file could not be found` so that the user knows exactly what went wrong, instead of thinking that something broke internally.

bojinyao: Implemented 👍

docs/elements.md
bojinyao: Updated 👍

bojinyao:
Thanks for fixing up the merge 🙏!! It was a mess haha 😅

elements/pl-multiple-choice/pl-multiple-choice.py
bojinyao: I also display the resolved file path to make it easier for user to debug 👍

exampleCourse/questions/element/multipleChoice/question.html
```diff
@@ -147,3 +147,25 @@
...
+
+<div class="card my-2">
+    <div class="card-header">
+        Part 6
+    </div>
+
+    <div class="card-body">
+
+        <pl-question-panel>
+            <p>Here, the <code>pl-multiple-choice</code> element does not contain any <code>pl-answer</code> inner tag; instead, it pulls the correct and incorrect choices, if exist,
```
nicknytko, line 159:
```suggestion
            <p>Here, the <code>pl-multiple-choice</code> element does not contain any <code>pl-answer</code> inner tag; instead, it pulls the correct and incorrect choices, if they exist,
```

elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -28,17 +33,39 @@ def categorize_options(element):
...
+            with open(json_file, mode='r', encoding='utf-8') as f:
+                obj = json.load(f)
+                for text in obj.get(correct_attrib, []):
+                    correct_answers.append((index, True, text))
+                    index += 1
+                for text in obj.get(incorrect_attrib, []):
+                    incorrect_answers.append((index, False, text))
+                    index += 1
+        except FileNotFoundError:
+            raise Exception(f'JSON answer file: "{json_file}" could not be found')
```
nicknytko, line 55:
```suggestion
            raise Exception(f'JSON answer file: "{file_path} could not be found')
```

We want to display the path here and not the actual file object?

bojinyao: Oh, this is to also accommodate relative paths. If `file_path` is abs, then `json_file` is the same.
But when a relative path is given, I think it will be more helpful to show the user what the resolved `data['options']['question_path'] + file_path` looks like for easier debug.

nicknytko: Ah, okay, nevermind then.  This is fine.

PR #2735, opened 2020-07-16
Change coveralls CI step to continue on error

tdy:
#2710 changed `coveralls.sh`, but it seems that was only used by the old travis CI config. This PR changes the current CI workflow to `continue-on-error` for the coveralls step.

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

Side note: should we delete the deprecated travis config/scripts?


mwest1066:
Yep, deleting Travis configs is probably a good idea.

Issue #2506, opened 2020-05-14
Need to validate time limit expiry on the server


mwest1066:
At the moment, when a student's browser sends a `__timeLimitExpired` POST request, we unconditionally close the assessment. This can cause a problem when we manually extend a time limit, but the student has an old tab open with a countdown timer that's using the old time limit.

We should check that we are indeed at the time limit on the server before closing the assessment. If not, we can just redirect back to the page with no changes.

This issue will be important when we implement manual time limit changes in #2293

tdy:
> We should check that we are indeed at the time limit on the server before closing the assessment. If not, we can just redirect back to the page with no changes.

If I understand correctly, this sounds like a bandaid for a bigger issue -- students won't see updated timers unless they personally refresh their page? Can we resolve both issues by somehow force-refreshing existing assessment timers?

mwest1066:
The places where we process time limit expirations is: https://github.com/PrairieLearn/PrairieLearn/blob/bf2f3c3a6b618f79a426a069b61b4fb5553aabaa/pages/studentAssessmentInstanceExam/studentAssessmentInstanceExam.js#L34 and: https://github.com/PrairieLearn/PrairieLearn/blob/bf2f3c3a6b618f79a426a069b61b4fb5553aabaa/pages/studentInstanceQuestionExam/studentInstanceQuestionExam.js#L73

mwest1066:
Here's how we test whether the time limit has actually expired: https://github.com/PrairieLearn/PrairieLearn/blob/bf2f3c3a6b618f79a426a069b61b4fb5553aabaa/pages/studentInstanceQuestionExam/studentInstanceQuestionExam.js#L62

Issue #1376, opened 2018-12-06
Optimize course gradebook page HTML


nwalters512:
On the course gradebook page, we currently spend a large amount of time and bytes rendering the `editTotalScorePercentButton` and its associated popover for each and every cell in the table. This makes up a majority of the bytes sent over the network, contributing to the 120MB page size for courses like CS125. There's no good reason to render all of those on the server; we should take advantage of JS on the client to generate the popover dynamically when the edit button is clicked.

nwalters512:
As a possible followup: we could send an optimized representation of the table (CSV?) with the initial response and then render the table solely on the client.

PR #2750, opened 2020-07-20
Add instructor buttons to live-edit exam time limits

### LINKED ISSUE #2293: Add instructor button to extend timeLimitMin for in-progress exams
At the moment, once an exam is started with a time limit there is no way for the instructor to change the active timer. It would be great to add both per-instance buttons and a for-all-students button to add (or subtract) time.

The current time limit is stored in `instance_questions.date_limit`, so we need to add to that value (presumably only for open exams? or do we re-open closed ones and also give them more time).

See also #611, #2275

tdy:
Another possibility is to turn the time limits into editable inline form fields (instead of these +/- buttons).

image

image

---

* [x] Resolve #2293 
    * [x] per-instance buttons
    * [x] per-assessment buttons
 #2506 moved to a different PR (#3386 )


tdy:
@mwest1066 @cheeren I've merged this with the latest `master`. Just let me know if you want to hand this off, or you have other plans for this feature, etc. I'm fine with whatever.

jonatanschroeder:
A few suggestions:

- Instead of a fixed -5/+5 button, I think it may be more appropriate to have a small modal with an `input type=number` with the number of minutes to increase/decrease, similar to the "Actions" menu right beside it. Not make the time remaining itself editable, as those can get confusing with a time that keeps running.

- The form for this setting should probably update the data via Ajax.  The rationale is that it will be unlikely that such a change would involve a single student, so updating the entire page would just lose the position of where that student was. The Ajax return could give a new time limit for that student, which would update only the corresponding cell with the proper time limit.

- It probably makes sense to couple this with live updates on the available time (which would address #3340 ). An ajax update in the previous item would probably make this live update easier to do. Maybe do a periodic update on all cells (e.g., every 10 seconds, or based on how many seconds are left on the next update; and more often when timer for at least one student is running out). To be decided: how to handle students that start after the page was loaded, are they added too? Maybe a warning asking instructor to refresh?

- For updating all students I think it makes more sense to have this be in the same page as the per-instance update, just to avoid confusion. Something like a button "add more time to all students", or something like that.

- For time limit expiration on the server, again this could be done with the student code periodically asking the server how many seconds are left in the exam.

I'm happy to take over and work with these options.

jonatanschroeder:
We can move a copy of the "Actions" menu to the top, with the same options (including more time) added there to be applied to all students.

mfsilva22:
Hi @jonatanschroeder do you think you can add text to the documentation about this new feature?

jonatanschroeder:
From https://github.com/PrairieLearn/PrairieLearn/pull/3600#issuecomment-752321506: add Close all button, as well as Grade/Grade all.

jonatanschroeder:
@mwest1066 I was thinking about this better, and I'll implement Close all/Grade all in #3600 instead. I have made some changes to the library there, and I'd need to adapt those changes to the new interface anyway, so easier to do it there. This means this PR is ready for review.

mwest1066:
> @mwest1066 I was thinking about this better, and I'll implement Close all/Grade all in #3600 instead. I have made some changes to the library there, and I'd need to adapt those changes to the new interface anyway, so easier to do it there. This means this PR is ready for review.

Excellent! Sounds good.

pages/instructorAssessmentInstances/instructorAssessmentInstances.sql
```diff
@@ -52,3 +68,95 @@ INSERT INTO assessment_state_logs AS asl
...
+                                    WHEN $base_time = 'current_date' THEN current_timestamp
+                                    ELSE ai.date_limit
+                                    END) +
+                                   $time_add * INTERVAL '1 sec')
+                     END,
+        modified_at = now()
+    WHERE
+        ai.open
+        AND ai.assessment_id = $assessment_id
+        AND (ai.date_limit IS NOT NULL OR $base_time <> 'date_limit')
```
mwest1066, line 148: Looks like Pascal! I much prefer C 😄
```suggestion
        AND (ai.date_limit IS NOT NULL OR $base_time != 'date_limit')
```

pages/instructorAssessmentInstances/instructorAssessmentInstances.sql
```diff
@@ -52,3 +68,95 @@ INSERT INTO assessment_state_logs AS asl
...
+    SET
+        date_limit = CASE
+                     WHEN $base_time = 'null' THEN NULL
+                     ELSE GREATEST(current_timestamp,
+                                   (CASE
+                                    WHEN $base_time = 'start_date' THEN ai.date
+                                    WHEN $base_time = 'current_date' THEN current_timestamp
+                                    ELSE ai.date_limit
+                                    END) +
+                                   $time_add * INTERVAL '1 sec')
```
mwest1066, line 142:
```suggestion
                                   make_interval(sec => $time_add)
```

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -129,29 +212,63 @@
...
                   <div class="dropdown">
                     <button type="button" class="btn btn-xs dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
                       Action <span class="caret"></span>
                     </button>
+                    <div id="row<%= iRow %>PopoverSetTimeLimit"
+                         tabindex="0" data-toggle="popover" data-trigger="manual" data-container="body"
+                         data-html="true" data-placement="auto" title="Set Time Limit"
+                         data-content="
+                                       <p>
+                                         Time limit: <%= row.total_time %><br/>
```
mwest1066, line 237: Maybe emphasize that this is the total time limit that the student has? I'm not sure how to best word this.
```suggestion
                                         Total time limit: <%= row.total_time %><br/>
```

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -129,29 +212,63 @@
...
+                                           <input class=&quot;form-control time-limit-field&quot; type=&quot;number&quot; name=&quot;time_add&quot; style=&quot;width: 5em&quot; value=&quot;5&quot;>
+                                           <select class=&quot;form-control time-limit-field&quot; name=time_ref><option value=60>minutes</option><option value=1>seconds</option></select>
+                                         </p>
+                                         <button type=&quot;button&quot; class=&quot;btn btn-secondary&quot; onclick=&quot;$('#row<%= iRow %>PopoverSetTimeLimit').popover('hide')&quot;>
+                                           Cancel
+                                         </button>
+                                         <button type=&quot;submit&quot; class=&quot;btn btn-success&quot;>Set</button>
+                                       </form>
+                                       ">
+                    </div>
```
mwest1066, line 265: Rather than putting the "Set time limit" in "Actions", could we instead put an ["edit" button](https://github.com/PrairieLearn/PrairieLearn/blob/master/pages/instructorAssessmentInstance/instructorAssessmentInstance.ejs#L64) next to the current time in the "Remaining" column? I like generally having the "edit" functionality directly connected to the thing being edited.

Given that the "Remaining" column is were we currently put "Open/Closed" (and where we will put "Expired" or whatever once #3622 is implemented), I'm thinking that it would be more logical to put all of the close/re-open/timer functionality into an edit button within the "Remaining" column. This edit button could either: (1) pop up a menu with items "Set Time Limit", "Re-open", and "Close", or (2) open a single dialog with buttons to adjust the time or re-open/close. I have a slight preference for immediately opening a single dialog, if it can be made to look ok.

If we keep the menu items, can we change "Set Time Limit" to "Change Time Limit"? I was surprised when I clicked "Set Time Limit" but the actual dialog defaulted to "Add". I think that defaulting to "Add" (when there is an existing time limit) is the right behavior, but the menu item being called "Set" made me think it was going to default to "Setting" the time limit.

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -93,10 +100,86 @@
...
+              <button type="button" class="btn btn-light dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                Action for all instances <span class="caret"></span>
+              </button>
+              <div id="PopoverSetTimeLimitAll"
+                   tabindex="0" data-toggle="popover" data-trigger="manual" data-container="body"
+                   data-html="true" data-placement="auto" title="Set Time Limit"
+                   data-content="
+                                 <p>
+                                   Time limit: <%= time_limit_list %><br/>
+                                   Remaining time: <%= remaining_time_range %>
```
mwest1066, line 113: I found these fields a bit confusing (it was only after reading the code that I understood them). Maybe they aren't really needed? I feel like if I'm proctoring an exam and want to add 5 min to everyone, I already have a sense of what's going on.

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -93,10 +100,86 @@
...
+                                        } %>
+                                     <option value=set_total>Set time limit to</option>
+                                     <option value=set_rem>Set remaining time to</option>
+                                     <% if (remaining_time_min !== null) { %>
+                                     <option value=unlimited>Set as unlimited</option>
+                                     <% } %>
+                                   </select>
+                                   <p class=&quot;form-inline&quot;>
+                                     <input class=&quot;form-control time-limit-field&quot; type=&quot;number&quot; name=&quot;time_add&quot; style=&quot;width: 5em&quot; value=&quot;5&quot;>
+                                     <select class=&quot;form-control time-limit-field&quot; name=time_ref><option value=60>minutes</option><option value=1>seconds</option></select>
```
mwest1066, line 136: Ideally I'd like a third option of `percent of total time` here, so that I could easily update everyone's times while preserving the extended-duration for disability accommodations.

Personally, I'm not sure that it's worth bothering with `seconds` as an option. Or is there a use case where `minutes` wouldn't be sufficient?

docs/accessControl.md
```diff
@@ -108,6 +108,30 @@ The above example will give students 90 minutes for this exam, and they must sta
...
+
+On the Instructor View for the assessment, select the Students tab. This tab will show a list of all students that have started the assessment. The "Remaining" column will include one of the following values:
+
+* The value "Closed" indicates that the student has already finished the assessment, that their time ran out, or ([for exam assessments with `autoClose` set as `true`](assessment.md#auto-closing-exam-assessments)) that the student was inactive for over six hours. Students do not have the ability to submit answers for these assessments.
+* The value "Unlimited" indicates that the assessment is open and has no time limit (other than the `autoClose` setting listed above). Students are able to submit answers for these assessments.
+* A value indicating a number of minutes indicates the remaining time until the exam is automatically closed. Students are able to submit answers until this time expires. Note that this time is computed when the page is loaded, so to get an updated time, the page must be refreshed.
+
+To change the remaining time, or to close or re-open an assessment, click on the "Actions" button, which will open the following options:
```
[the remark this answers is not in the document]
jonatanschroeder: Good point, will do.

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
jonatanschroeder: I think they are useful when setting a fixed time instead of adding/subtracting. What if I use "Total time allotted" instead of "Time limit"?

pages/instructorAssessmentInstances/instructorAssessmentInstances.sql
```diff
@@ -52,3 +68,95 @@ INSERT INTO assessment_state_logs AS asl
...
+                     END,
+        modified_at = now()
+    WHERE
+        ai.open
+        AND ai.id = $assessment_instance_id
```
mwest1066, line 114: This `$assessment_instance_id` came from `req.body.assessment_instance_id` so it's untrusted. We need to check that the current user has access to modify it. The easiest way to do this is to pass in `locals.assessment.id` as `assessment_id` and add `AND ai.assessment_id = $assessment_id`. This is safe because all of the values in `res.locals` have been authorized and we check `res.locals.authz_data.has_instructor_edit` in the POST handler.

docs/accessControl.md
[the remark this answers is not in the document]
jonatanschroeder: Sorry, I thought you meant a different doc page, you meant the actual page. Since instructors won't be able to change the timeLimitMin there, I'm not sure how to best describe this option...

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -93,10 +100,86 @@
...
+                                   <button type=&quot;button&quot; class=&quot;btn btn-secondary&quot; onclick=&quot;$('#PopoverOpenAll').popover('hide')&quot;>
+                                     Cancel
+                                   </button>
+                                   <button type=&quot;submit&quot; class=&quot;btn btn-success&quot;>Open all instances</button>
+                                 </form>
+                                 ">
+              </div>
+              
+              <!-- Capture all clicks to dropdown items to prevent scrolling to the top of the page -->
+              <div class="dropdown-menu dropdown-menu-right" onclick="window.event.preventDefault();">
```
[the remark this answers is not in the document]
jonatanschroeder: This is copied from the Actions dropdown (line 316).

pages/instructorAssessmentInstances/instructorAssessmentInstances.js
```diff
@@ -21,6 +21,40 @@ router.get('/', function(req, res, next) {
...
+                if (res.locals.remaining_time_min === null ||
+                    res.locals.remaining_time_min > row.time_remaining_sec)
+                    res.locals.remaining_time_min = row.time_remaining_sec;
+                if (res.locals.remaining_time_max === null ||
+                    res.locals.remaining_time_max < row.time_remaining_sec)
+                    res.locals.remaining_time_max = row.time_remaining_sec;
+            }
+        });
+        res.locals.time_limit_list = Object.values(res.locals.time_limit_list);
+        res.locals.time_limit_list = res.locals.time_limit_list.length > 0 ? res.locals.time_limit_list.join(', ') : 'No time limits';
```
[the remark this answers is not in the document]
jonatanschroeder: I can add some limitation of the size of this list, but I expect that most cases will have a limited number of different time limits at a time. So even if you have 600 students, 590 of them will have the same time limit. Note that the list only adds limits if they're not in the list already.

docs/accessControl.md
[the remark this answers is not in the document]
mwest1066: I agree that it's a bit weird because you can't edit on the Access page. I just worry that people will think that editing an access rule to increase the time limit will cause that time limit to be extended for exams that are in progress. I think we saw this with some UBC exams in Fall, where instructors tried to extend time limits in this way.

I guess one possibility would be to actually support this, although it's a bit clunky. We wouldn't want to overwrite manual adjustments, so we'd probably need to store a flag like `time_limit_set_manually` so we would know which ones to extend on a sync. Personally I'm not very enthusiastic about doing this, as I think it would be confusing.

Maybe at the bottom of the Access page we could say "To change the access rules, [Edit `infoAssessment.json]" where we link to the editing page. And then underneath that we could say "Note that editing time limits in the access rules will not change the time limits for in-progress assessments. To change these, go to the [Students] tab."

Or maybe we just forget about all of this idea. Instructors will probably figure it out and we can always put a FAQ entry about it!

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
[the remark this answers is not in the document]
mwest1066: Huh, good point. I don't know what it's doing but I agree we should keep it 😄

pages/instructorAssessmentInstances/instructorAssessmentInstances.sql
```diff
@@ -9,9 +9,25 @@ SELECT
     CASE
         WHEN ai.open AND ai.date_limit IS NOT NULL
             THEN greatest(0, floor(extract(epoch from (ai.date_limit - current_timestamp)) / 60))::text || ' min'
-        WHEN ai.open THEN 'Open'
+        WHEN ai.open THEN 'Unlimited'
```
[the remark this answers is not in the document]
jonatanschroeder: There are two different concepts here, close vs open and time limits. Re-opening an exam will set the time limit to unlimited (this is sorta mentioned in the popover for re-open as well, though I can make it clearer).

docs/accessControl.md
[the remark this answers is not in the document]
jonatanschroeder: Would a link to the access rules docs be an option?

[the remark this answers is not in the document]
mwest1066: Absolutely, that's a good idea.

pages/instructorAssessmentInstances/instructorAssessmentInstances.sql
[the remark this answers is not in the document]
jonatanschroeder: An alternative is to have "Re-open" open the time limit popover and then update the time limit in the same operation as re-open. Then a closed assessment would show "Re-open", while an open assessment would should "Close" or "Change Time Limit".

sprocs/assessment_instances_select_log.sql
```diff
@@ -279,9 +279,20 @@ BEGIN
                     NULL::INTEGER AS variant_id,
                     NULL::INTEGER AS variant_number,
                     NULL::INTEGER AS submission_id,
-                    NULL::JSONB AS data
+                    CASE
+                    WHEN asl.open THEN jsonb_build_object(
+                         'date_limit',
+                         CASE WHEN asl.date_limit IS NULL THEN 'Unlimited'
+                         ELSE format_date_full_compact(asl.date_limit, ci.display_timezone)
+                         END
```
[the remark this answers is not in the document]
jonatanschroeder: Done, I also added the remaining time at the time of the log.

pages/instructorAssessmentInstances/instructorAssessmentInstances.sql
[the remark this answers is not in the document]
mwest1066: To make sure I'm understanding, at the moment we have both `open` and `date_limit` fields. The useful combinations of these are:
1. `open=true` and `date_limit=null`: Assessment is open without a timer (or with an unlimited timer, to say it a different way)
2. `open=true` and `date_limit > now()`: Timer is counting down
3. `open=true` and `date_limit < now()`: Timer has expired, the assessment will close shortly
4. `open=false`: Assessment is closed (timer is irrelevant)

I think we are talking about what to call case (1) above?

Until now we've been saying that case (1) is "Open" and case (2) is "Closed". I agree that, especially once #3622 is implemented, we'll have to change these terms.

Could we wait until #3622 before making the change "Open" -> "Unlimited"? When #3622 is implemented we'll also change the "Re-open"/"Close" buttons to better names, and we can make sure everything is consistent. Until we do that I'd prefer to avoid terminology/UI changes.

sprocs/assessment_instances_select_log.sql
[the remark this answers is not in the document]
mwest1066: Nice!

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
jonatanschroeder: New option (still some tests pending) lets you do weird things like "Set remaining time to 30% of total time limit"...

pages/instructorAssessmentInstances/instructorAssessmentInstances.sql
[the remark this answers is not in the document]
jonatanschroeder: I think making the distinction between Open and Unlimited is probably a good idea here, to make it clear that the exam is not only open, but also has no time limit. An alternative is to list this as "No time limit".

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
jonatanschroeder: Like this?
image

jonatanschroeder: I kept the Delete/Regrade/Re-open/Close in the Actions menu (where they were before), but moved the Change action to a button in the Remaining column. This button is also now Enabled when the setting is "Closed", in that case the exam is re-opened and the time limit is set as needed.

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -129,29 +212,63 @@
...
-                <td class="text-center">
+                <td class="text-center align-middle"><%= row.date_formatted %></td>
+                <td class="text-center align-middle" data-text="<%= row.duration_secs %>"><%= row.duration %></td>
+                <td class="text-center align-middle"><%= row.time_remaining %></td>
+                <td class="text-center align-middle">
                   <div class="dropdown">
                     <button type="button" class="btn btn-xs dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
                       Action <span class="caret"></span>
                     </button>
+                    <div id="row<%= iRow %>PopoverSetTimeLimit"
```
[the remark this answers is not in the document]
jonatanschroeder: It may be long in terms of bytes, but it's an attribute, so it's not DOM heavy. Are you concerned about page size/network or load time?

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
jonatanschroeder: I changed it to Total time limit. I'd prefer to keep the list of possible time limits, as that would remind proctors if there is a list of students with different total time, and that they need to consider those when computing the remaining time.

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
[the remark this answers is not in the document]
jonatanschroeder: The way that popover works makes it a bit hard to work with this kind of setting in a separate element. So if you're concerned about page size, it'd probably be better to switch from using popovers to using modals.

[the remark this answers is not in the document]
mwest1066: I was thinking about #1430 that was in response to #1376. The Gradebook page was over 100 MB for a large course! I haven't measured the size of `instructorAssessmentInstances` for a large course, but I'm worried about network size. Could we use the same dynamic-popover technique that #1430 used?

pages/instructorAssessmentInstances/instructorAssessmentInstances.sql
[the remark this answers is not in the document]
mwest1066: My concern is how a user is supposed to know that "Unlimited" also means "Open".

I could live with "Open (no time limit)" as a temporary fix. What do you think? I agree that it's basically all very confusing until we get #3622 done.

Once we do something about #3622 then I guess we'll have this column show "No time limit", "5 min", or "Expired"? With actions like "Remove time limit", "Change time limit", "Expire time limit"? And no mention of "open"/"closed" at all?

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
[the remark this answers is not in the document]
jonatanschroeder: Should I do that for reopen/regrade/close?

pages/instructorAssessmentInstances/instructorAssessmentInstances.sql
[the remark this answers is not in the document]
jonatanschroeder: One option, in preparation for #3622, is to include the "Expired" option if the time expired but the exam is not closed yet (possibly because the student's browser didn't trigger the close yet, or the student closed the tab without closing the assessment). I'm also fine with renaming "Set to unlimited" to "Remove time limit", and add an "Expire time limit" option to the dropdown. Then remove the "Re-open" option (since everything that needs to be done in that button is already possible in the Change time limit option), so that all actions in the "Actions" dropdown are related to grading (for now Close & Grade and Regrade, but based on #3644 could also include an option to grade saved answers without closing or affecting the time limit).

This would also affect the "Re-open all" option, so for consistency I can add the same options in the "Change all time limits", and add a flag there to indicate if it should also be applied to closed and/or expired assessments.

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -93,10 +173,26 @@
...
+                
+                <% if (authz_data.has_instructor_edit) { %>
+                <a class="dropdown-item"
+                   href="#"
+                   data-toggle="modal" data-target="#deleteAllAssessmentInstancesModal">
+                  <i class="fa fa-times" aria-hidden="true"></i> Delete all instances
+                </a>
+                <a class="dropdown-item time-limit-edit-button" data-row='<%= JSON.stringify(time_limit_totals) %>'>
+                  <i class="far fa-clock" aria-hidden="true"></i> Change time limit for all instances
+                </a>
```
nwalters512, line 191: For accessibility purposes, these should be `<button type="button">` and not anchors.

```suggestion
                <button type="button" class="dropdown-item"
                   data-toggle="modal" data-target="#deleteAllAssessmentInstancesModal">
                  <i class="fa fa-times" aria-hidden="true"></i> Delete all instances
                </button>
                <button type="button" class="dropdown-item time-limit-edit-button" data-row='<%= JSON.stringify(time_limit_totals) %>'>
                  <i class="far fa-clock" aria-hidden="true"></i> Change time limit for all instances
                </button>
```

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -129,41 +225,30 @@
...
-                <td class="text-center"><%= row.time_remaining %></td>
-                <td class="text-center">
+                <td class="text-center align-middle"><%= row.date_formatted %></td>
+                <td class="text-center align-middle" data-text="<%= row.duration_secs %>"><%= row.duration %></td>
+                <td class="text-center align-middle">
+                  <%= row.time_remaining %>
+                  <% if (authz_data.has_instructor_edit) { %>
+                  <a class="btn btn-xs time-limit-edit-button" data-row='<%= JSON.stringify(row) %>' >
+                    <i class="fas fa-edit" aria-hidden="true"></i>
+                  </a>
```
nwalters512, line 244: Ditto re: accessibility concerns.

```suggestion
                  <button type="button" class="btn btn-xs time-limit-edit-button" data-row='<%= JSON.stringify(row) %>' >
                    <i class="fas fa-edit" aria-hidden="true"></i>
                  </button>
```

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -9,7 +9,80 @@
...
+                      time_ref_select.append('<option value=percent>% total limit</option>');
+                  new_time.append(time_ref_select);
+                  form.append(new_time);
+                  if (row.has_closed_instance) {
+                      let checkbox = $('<div class="form-check mb-2 reopen-closed-field"><input class="form-check-input" type="checkbox" name="reopen_closed" value="true" id="reopen-closed"><label class="form-check-label" for="reopen-closed">Also re-open closed instances</label></div>');
+                      checkbox.toggle(row.time_remaining_sec === null);
+                      form.append(checkbox);
+                  }
+                  let buttons = $('<div class="btn-toolbar pull-right">');
+                  let cancel_button = $('<button type="button" class="btn btn-secondary mr-2" onclick="">Cancel</button>');
```
nwalters512, line 73: Nit: is the empty `onclick` necessary?

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -129,41 +225,30 @@
...
-                <td class="text-center"><%= row.date_formatted %></td>
-                <td class="text-center" data-text="<%= row.duration_secs %>"><%= row.duration %></td>
-                <td class="text-center"><%= row.time_remaining %></td>
-                <td class="text-center">
+                <td class="text-center align-middle"><%= row.date_formatted %></td>
+                <td class="text-center align-middle" data-text="<%= row.duration_secs %>"><%= row.duration %></td>
+                <td class="text-center align-middle">
+                  <%= row.time_remaining %>
+                  <% if (authz_data.has_instructor_edit) { %>
+                  <a class="btn btn-xs time-limit-edit-button" data-row='<%= JSON.stringify(row) %>' >
```
nwalters512, line 242: Nit: single quotes aren't necessary since EJS will escape the JSON string for you.

```suggestion
                  <a class="btn btn-xs time-limit-edit-button" data-row="<%= JSON.stringify(row) %>">
```

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -93,10 +173,26 @@
...
+              <!-- Capture all clicks to dropdown items to prevent scrolling to the top of the page -->
+              <div class="dropdown-menu dropdown-menu-right" onclick="window.event.preventDefault();">
+                
+                <% if (authz_data.has_instructor_edit) { %>
+                <a class="dropdown-item"
+                   href="#"
+                   data-toggle="modal" data-target="#deleteAllAssessmentInstancesModal">
+                  <i class="fa fa-times" aria-hidden="true"></i> Delete all instances
+                </a>
+                <a class="dropdown-item time-limit-edit-button" data-row='<%= JSON.stringify(time_limit_totals) %>'>
```
nwalters512, line 189: Nit: single quotes aren't necessary since EJS will escape the JSON string for you.

```suggestion
                <a class="dropdown-item time-limit-edit-button" data-row="<%= JSON.stringify(time_limit_totals) %>">
```

pages/instructorAssessmentInstances/instructorAssessmentInstances.js
```diff
@@ -21,30 +21,58 @@ router.get('/', function(req, res, next) {
...
+            res.locals.time_limit_totals.time_remaining = 'No time limits';
+        else if (remaining_time_max < 60)
+            res.locals.time_limit_totals.time_remaining = 'Less than a minute';
+        else if (remaining_time_min < 60)
+            res.locals.time_limit_totals.time_remaining = 'up to ' + Math.floor(remaining_time_max / 60) + ' min';
+        else if (Math.floor(remaining_time_min / 60) == Math.floor(remaining_time_max / 60))
+            res.locals.time_limit_totals.time_remaining = Math.floor(remaining_time_min / 60) + ' min';
+        else
+            res.locals.time_limit_totals.time_remaining = 'between ' + Math.floor(remaining_time_min / 60) + ' and ' + Math.floor(remaining_time_max / 60) + ' min';
+
```
nwalters512, line 67: Nit: please include curly braces in `if` blocks.

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -9,7 +9,80 @@
...
+                               </p>`);
+                  form.append(`<input type="hidden" name="__action" value="${action}">`);
+                  form.append('<input type="hidden" name="__csrf_token" value="<%= __csrf_token %>">');
+                  if (row.assessment_instance_id)
+                      form.append(`<input type="hidden" name="assessment_instance_id" value="${row.assessment_instance_id}">`);
+
+                  let select = $('<select class="form-control select-time-limit" name=plus_minus>');
+                  if (row.time_remaining_sec !== null) {
+                      if (row.has_open_instance) {
+                          select.append('<option value=+1>Add to instances with time limit</option>');
```
nwalters512, line 40: Nit: I understand that omitting the `"` in attribute values is technically valid HTML but I'd rather we use it for consistency.

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -217,33 +302,23 @@
...
-                        <i class="fa fa-sync" aria-hidden="true"></i> Regrade
-                      </a>
-                      <% } %>
-
-                      <!-- ################################################## -->
-
-                      <% if (authz_data.has_instructor_edit) { %>
-                      <a class="dropdown-item <% if (row.open) { %>disabled<% } %>" <% if (!row.open) { %>onclick="$('#row<%= iRow %>PopoverOpen').popover('show')"<% } %> href="#">
-                        <i class="far fa-circle" aria-hidden="true"></i> Re-open
+                      <a class="dropdown-item <% if (!row.open) { %>disabled<% } %>" <% if (row.open) { %>onclick="$('#row<%= iRow %>PopoverClose').popover('show')"<% } %> href="#">
```
nwalters512, line 305: Ditto for accessibility concerns - this should be a `<button type="button">` (apologies, but GitHub won't let me leave a code suggestion here).

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -217,33 +302,23 @@
...
+                        <i class="fa fa-ban" aria-hidden="true"></i> Grade & Close
                       </a>
                       <% } %>
 
-                      <!-- ################################################## -->
-
                       <% if (authz_data.has_instructor_edit) { %>
-                      <a class="dropdown-item <% if (!row.open) { %>disabled<% } %>" <% if (row.open) { %>onclick="$('#row<%= iRow %>PopoverClose').popover('show')"<% } %> href="#">
-                        <i class="fa fa-ban" aria-hidden="true"></i> Close
+                      <a class="dropdown-item" onclick="$('#row<%= iRow %>PopoverRegrade').popover('show')" href="#">
```
nwalters512, line 311: Ditto - should be a `<button type="button">`.

pages/instructorAssessmentInstances/instructorAssessmentInstances.js
```diff
@@ -91,6 +119,56 @@ router.post('/', function(req, res, next) {
...
+            params.base_time = 'current_date';
+            params.time_add = 0;
+            params.time_ref = 'minutes';
+        } else if (req.body.plus_minus == 'set_total')
+            params.base_time = 'start_date';
+        else if (req.body.plus_minus == 'set_rem')
+            params.base_time = 'current_date';
+        else
+            params.time_add *= req.body.plus_minus;
+        sqldb.query(sql.set_time_limit, params, function(err, _result) {
```
nwalters512, line 143: Nit:
```suggestion
        sqldb.query(sql.set_time_limit, params, function(err) {
```

pages/instructorAssessmentInstances/instructorAssessmentInstances.js
```diff
@@ -91,6 +119,56 @@ router.post('/', function(req, res, next) {
...
+        else if (req.body.plus_minus == 'expire') {
+            params.base_time = 'current_date';
+            params.time_add = 0;
+            params.time_ref = 'minutes';
+        } else if (req.body.plus_minus == 'set_total')
+            params.base_time = 'start_date';
+        else if (req.body.plus_minus == 'set_rem')
+            params.base_time = 'current_date';
+        else
+            params.time_add *= req.body.plus_minus;
```
nwalters512, line 142: Nit: please use curly braces for `if` blocks.

pages/instructorAssessmentInstances/instructorAssessmentInstances.js
```diff
@@ -91,6 +119,56 @@ router.post('/', function(req, res, next) {
...
```
nwalters512, line 167: Nit: please use curly braces for if blocks.

pages/instructorAssessmentAccess/instructorAssessmentAccess.ejs
```diff
@@ -45,6 +45,9 @@
             </tbody>
           </table>
         </div>
+        <div class="card-footer">
+          <small>Instructions on how to change the access rules can be found in the <a href="https://prairielearn.readthedocs.io/en/latest/accessControl/" target="_blank">PrairieLearn documentation</a>. Note that changing some of these rules does not affect assessments in progress; to change the time limit for these exams please visit the <a href="instances">Students tab</a>.</small>
```
nwalters512, line 49: Can we be more precise about which rules the "some of" applies to?

nwalters512 (commented):
Left a bunch of opinionated/preference style nits. My main concern is the accessibility of `<a>` vs `<button>` - not going to personally approve until that is resolved.

jonatanschroeder:
> Left a bunch of opinionated/preference style nits. My main concern is the accessibility of `<a>` vs `<button>` - not going to personally approve until that is resolved.

Check #3610 and https://getbootstrap.com/docs/4.5/components/popovers/#dismiss-on-next-click. @mwest1066 has asked me to use `a` instead of `button` on some circumstances, I believe they apply here.

nwalters512:
> > Left a bunch of opinionated/preference style nits. My main concern is the accessibility of `<a>` vs `<button>` - not going to personally approve until that is resolved.
> 
> Check #3610 and https://getbootstrap.com/docs/4.5/components/popovers/#dismiss-on-next-click. @mwest1066 has asked me to use `a` instead of `button` on some circumstances, I believe they apply here.

Ah! Good callout. `<a tabindex="0" role="button">` would be fine too from an accessibility perspective. The important thing is that we don't use `<a href="#">`.

jonatanschroeder:
Changes applied and docs updated.

pages/instructorAssessmentInstances/instructorAssessmentInstances.js
```diff
@@ -21,30 +21,62 @@ router.get('/', function(req, res, next) {
...
+        if (remaining_time_min === null)
+            res.locals.time_limit_totals.time_remaining = 'No time limits';
+        else if (remaining_time_max < 60)
+            res.locals.time_limit_totals.time_remaining = 'Less than a minute';
+        else if (remaining_time_min < 60)
+            res.locals.time_limit_totals.time_remaining = 'up to ' + Math.floor(remaining_time_max / 60) + ' min';
+        else if (Math.floor(remaining_time_min / 60) == Math.floor(remaining_time_max / 60))
+            res.locals.time_limit_totals.time_remaining = Math.floor(remaining_time_min / 60) + ' min';
+        else
+            res.locals.time_limit_totals.time_remaining = 'between ' + Math.floor(remaining_time_min / 60) + ' and ' + Math.floor(remaining_time_max / 60) + ' min';
```
nwalters512, line 70: Nit: curly braces.
```suggestion
        if (remaining_time_min === null) {
            res.locals.time_limit_totals.time_remaining = 'No time limits';
        } else if (remaining_time_max < 60) {
            res.locals.time_limit_totals.time_remaining = 'Less than a minute';
        } else if (remaining_time_min < 60) {
            res.locals.time_limit_totals.time_remaining = 'up to ' + Math.floor(remaining_time_max / 60) + ' min';
        } else if (Math.floor(remaining_time_min / 60) == Math.floor(remaining_time_max / 60)) {
            res.locals.time_limit_totals.time_remaining = Math.floor(remaining_time_min / 60) + ' min';
        } else {
            res.locals.time_limit_totals.time_remaining = 'between ' + Math.floor(remaining_time_min / 60) + ' and ' + Math.floor(remaining_time_max / 60) + ' min';
        }
```

nwalters512 (approved):
LGTM! Will leave the final approval to @mwest1066.

pages/instructorAssessmentInstances/instructorAssessmentInstances.js
jonatanschroeder: I'd prefer not to use braces here, since there's a single variable being modified based on these rules, but I don't have a strong preference.

pages/instructorAssessmentAccess/instructorAssessmentAccess.ejs
```diff
@@ -46,7 +46,7 @@
           </table>
         </div>
         <div class="card-footer">
-          <small>Instructions on how to change the access rules can be found in the <a href="https://prairielearn.readthedocs.io/en/latest/accessControl/" target="_blank">PrairieLearn documentation</a>. Note that changing some of these rules does not affect assessments in progress; to change the time limit for these exams please visit the <a href="instances">Students tab</a>.</small>
+          <small>Instructions on how to change the access rules can be found in the <a href="https://prairielearn.readthedocs.io/en/latest/accessControl/" target="_blank">PrairieLearn documentation</a>. Note that changing time limit rules does not affect assessments in progress; to change the time limit for these exams please visit the <a href="instances">Students tab</a>.</small>
```
mwest1066, line 49: (Warning! untested suggested change)

Nit: prefer explicit URLs. This is because relative URLs mean that we have to be extremely careful with trailing slashes on URLs. That is, if we are on a current page with `.../access/` versus `.../access` then the current page will display the same, but a relative URL of `instances` will go to either `.../access/instances` (incorrect) or `.../instances` (correct).

```suggestion
          <small>Instructions on how to change the access rules can be found in the <a href="https://prairielearn.readthedocs.io/en/latest/accessControl/" target="_blank">PrairieLearn documentation</a>. Note that changing time limit rules does not affect assessments in progress; to change the time limit for these exams please visit the <a href="<%= urlPrefix %>/assessment/<%= assessment.id %>/instances">Students tab</a>.</small>
```

mwest1066 (commented):
One last tiny change, sorry!

pages/instructorAssessmentAccess/instructorAssessmentAccess.ejs
jonatanschroeder: Tested and works.

mwest1066 (approved):

PR #2751, opened 2020-07-20
Change pl-prairiedraw-figure to the deprecated elements docs

mwest1066:
This PR officially deprecates the `pl-prairiedraw-figure` element and the `PrairieDraw.js` library. These are now quite old and a bit clunky (especially in the way they use `Sylvester.js` for vectors). I don't think we should stop supporting these anytime soon, but new questions should be encouraged to use the `pl-drawing` element instead. This PR hopefully makes that clear in the docs, and it also shifts the exampleCourse question to the testCourse to hide it from regular users.


docs/PrairieDraw.md
```diff
@@ -1,6 +1,10 @@
 
 # PrairieDraw
 
+**WARNING**: This page describes the old PrairieDraw library. This library will be supported indefinitely, but it is strongly recommended that all new questions use the [`pl-drawing`](elements/#pl-prairiedraw-figure-element) element instead.
```
nicknytko, line 4:
```suggestion
**WARNING**: This page describes the old PrairieDraw library. This library will be supported indefinitely, but it is strongly recommended that all new questions use the [`pl-drawing`](elements/#pl-drawing-element) element instead.
```

mwest1066: Haha, oops!

PR #2757, opened 2020-07-22
Add attribute to <pl-checkbox> to pull correct and/or incorrect answers from a json file

bojinyao:
Similar to #2733 

`external-json` is an *optional* attribute to `<pl-checkbox>`, type string, default to `None`

User can pass a json file path to this attribute, preferably a relatively path to *where question.html lives*; or an absolute path that might look like: `/PrairieLearn/course/etc/file.json`

Exceptions will be thrown when file is not found, and `json` library when the file is not correctly formatted.

By default, the target json file should have 2 attributes: "correct", and "incorrect", both of which are json lists containing text that will be added to poll of correct and incorrect answers already in question.html. Below is an example of what the json might look like:

```json
{
    "correct": [
        "A true statement"
    ],
    "incorrect": [
        "Incorrect statement 1",
        "Incorrect statement 2"
    ]
}
```

Alternatively, user can pass in the name of attributes representing "correct" or "incorrect" using `external-json-correct-key` and `external-json-incorrect-key` respectively to override the default attribute names.

Resolve https://github.com/ace-lab/pl-ucb-cs10/issues/47

### DIFF
--- a/exampleCourse/questions/element/checkbox/question.html
+++ b/exampleCourse/questions/element/checkbox/question.html
@@ -203,3 +203,25 @@
         </pl-checkbox>
     </div>
 </div>
+
+<div class="card my-2">
+    <div class="card-header">
+        Part 8
+    </div>
+
+    <div class="card-body">
+
+        <pl-question-panel>
+            <p>Here, the <code>pl-checkbox</code> element does not contain any <code>pl-answer</code> inner tag; instead, it pulls the correct and incorrect choices, if exist,
+                from <code>group-work.json</code> file inside <code>serverFilesQuestion/</code> directory. By default, the external json file should have <code>"correct"</code>
+                and <code>"incorrect"</code> attributes of type list to store choices, these attribute names can be overridden with <code>"external-json-correct-key"</code>
+                and <code>"external-json-incorrect-key"</code>.</p>
+
+            <p>Group work provides which of the following benefits:</p>
+        </pl-question-panel>
+
+        <pl-checkbox answers-name="group_ans_v7" external-json="serverFilesQuestion/group-work.json" external-json-incorrect-key="incorrect_choices">
+        </pl-checkbox>
+
+    </div>
+</div>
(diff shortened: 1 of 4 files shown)


exampleCourse/questions/element/checkbox/question.html
```diff
@@ -203,3 +203,25 @@
...
+        Part 8
+    </div>
+
+    <div class="card-body">
+
+        <pl-question-panel>
+            <p>Here, the <code>pl-checkbox</code> element does not contain any <code>pl-answer</code> inner tag; instead, it pulls the correct and incorrect choices, if exist,
+                from <code>group-work.json</code> file inside <code>serverFilesQuestion/</code> directory. By default, the external json file should have <code>"correct"</code>
+                and <code>"incorrect"</code> attributes of type list to store choices, these attribute names can be overridden with <code>"external-json-correct-key"</code>
+                and <code>"external-json-incorrect-key"</code>.</p>
```
nicknytko, line 218:
```suggestion
            <p>Here, the <code>pl-checkbox</code> element does not contain any <code>pl-answer</code> inner tag; instead, it pulls the correct and incorrect choices, if they exist,
                from the <code>group-work.json</code> file inside of the <code>serverFilesQuestion/</code> directory. By default, the external json file should have <code>"correct"</code>
                and <code>"incorrect"</code> attributes of type list to store choices, these attribute names can be overridden with <code>"external-json-correct-key"</code>
                and <code>"external-json-incorrect-key"</code>.</p>
```

bojinyao:
Converted to Draft because code involved with this is nearly identical to the same attribute in `<pl-multiple-choice>` #2733 
And will be nearly identical once again if the same functionality is added to `<p-dropdown>` 
I want to write a helper function and place inside `freeformPythonLib/prairielearn.py` to reduce code duplication.
But i need to put this aside this week because I'm busy with Berkeley's SnapCon this whole week 😅

bojinyao:
AH, can't find a nice and reasonable place to create a function... I think it's not too bad to copy-paste the same function 3 times.... unless there's a nice place to put this helper function 😕

nwalters512:
@bojinyao I'm running through old PRs right now - would you still like to see this through?

nwalters512:
Going to close this to clean up stale issues; please reopen in a new PR if you'd still like to contribute this!

PR #2758, opened 2020-07-22
Add "inline" attribute to <pl-figure>

bojinyao:
Resolve https://github.com/ace-lab/pl-ucb-cs10/issues/50

attribute `inline` type `boolean`, default `false`

When enabled:
- Override the default image display where `<pl-figure>` will render `` tag inside `<div class="container-fluid">` making it difficult to make inline with other html elements like text etc.
- Only `` will be rendered instead


docs/elements.md
```diff
@@ -957,6 +957,7 @@ Attribute | Type | Default | Description
 `type` | text | 'static' | Type of file, either 'static' (an existing file) or 'dynamic' (a file generated by element or server code).
 `directory` | text | "clientFilesQuestion" | The directory that contains the file, either 'clientFilesQuestion' or 'clientFilesCourse' (see [client and server files](clientServerFiles.md)). A directory cannot be specified if `type='dynamic'`.
 `width` | number | `None` | Width of image (e.g., '250px').
+`inline` | boolean | false | Display figure inline.
```
nicknytko, line 960:
```suggestion
`inline` | boolean | false | Display figure inline with text (true) or on a separate line (false).
```

bojinyao:
The mustache code was unnecessarily difficult to read, and had 3 different sections to evaluate and render. 
The refactored one is much easier to read and has 2 sections to evaluate and render.

PR #2762, opened 2020-07-25
Add database support to workspace host server

tdy:
* [x] connect/initialize database
* [x] query container settings
* [x] add debugging code (from #2763)


workspace_host/interface.js
```diff
@@ -106,6 +152,29 @@ function _getAvailablePort(workspace_id, curPort, callback) {
...
 };
 
+function _queryContainerSettings(workspace_id, callback) {
+    sqldb.queryOneRow(sql.select_workspace_settings, {workspace_id}, function(err, result) {
+        if (err) return;
+        logger.info(`Query results: ${JSON.stringify(result.rows[0])}`);
+        const settings = {
+            workspace_image: result.rows[0].workspace_image || 'codercom/code-server',
+            workspace_port: result.rows[0].workspace_port || 8080,
+            workspace_args: result.rows[0].workspace_args || '--auth none',
```
mwest1066, line 162: Do we need these defaults here? They feel a bit hacky.

tdy: removed the defaults except for `workspace_args`, which now has a default of `|| ''` because it's optional (not sure if there's a better way to handle this)

mwest1066: That seems fine

PR #2769, opened 2020-07-27
Add workspace navbar with status badge and action buttons

tdy:
* [x] Install placeholder status badge and action buttons
* [x] Convert socket actions to route handlers

---

* `Close` button does not currently kill container -- had some issues converting this from socket to route, so punting for now
* `State` div is still heartbeating from `workspaceSockets.js` (not watching `change:state` event) -- will clean this up later when `workspaces.state` updates/emits properly


pages/partials/navbarWorkspace.ejs
```diff
@@ -0,0 +1,41 @@
+<nav class="navbar navbar-expand-md navbar-dark bg-info align-items-center" style="height:55px">
```
mwest1066, line 1: Rather than making this a partial, I suggest just putting it all directly into `pages/workspace/workspace.ejs`. It's only going to be used in that single file, so having a partial doesn't really make sense?

mwest1066:
The new nav on the workspace page looks great. Tell me when you've converted to use routes rather than websockets and we can get this merged.

tdy:
@mwest1066 I guess it's fine to merge now, unless you see glaring issues. Code is ugly/unfinished but should be fixed as we iterate through the to-dos.

pages/workspace/workspace.js
```diff
@@ -82,4 +126,15 @@ router.get('/:workspace_id', (req, res, next) => {
...
 
+router.get('/:workspace_id/:action', (req, res, _next) => {
+    const workspace_id = req.params.workspace_id;
+    const action = req.params.action;
+
+    if (action === 'reboot') {
+        logger.info(`[workspace.js] Rebooting workspace ${workspace_id}.`);
+        controlContainer(workspace_id, 'destroy');
+        res.redirect(`/workspace/${workspace_id}`);
+    }
```
mwest1066, line 137: Add an else that goes to an error page like this: https://github.com/PrairieLearn/PrairieLearn/blob/1b49e1dd1171d745f3d6b4379f3efc64c4c0a266/pages/instructorAssessmentInstances/instructorAssessmentInstances.js#L91

pages/workspace/workspace.js
```diff
@@ -1,18 +1,62 @@
...
 const async = require('async');
 
+const logger = require('../../lib/logger');
 const config = require('../../lib/config');
 const sqldb = require('@prairielearn/prairielib/sql-db');
 const sqlLoader = require('@prairielearn/prairielib/sql-loader');
 
 const sql = sqlLoader.loadSqlEquiv(__filename);
 
+// FIXME: duped from lib/workspaceSocket.js
```
mwest1066, line 18: Can we put this in a helper library somewhere?

PR #2770, opened 2020-07-27
Add workspace home directory option

tdy:
* [x] add `home` directory option to `questions` table/json
* [x] add `port` option to `workspaces` table
* [x] integrate into `interface.js`


migrations/188_workspaces__port__add.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE workspaces ADD COLUMN IF NOT EXISTS port smallint;
```
mwest1066, line 1: `smallint` is signed, so storing ports above 32K is going to be weird. I suggest just using `integer`.

exampleCourse/questions/demo/workspace/vscode/info.json
```diff
@@ -11,6 +11,7 @@
         "image": "codercom/code-server",
         "port": 8080,
         "args": "--auth none",
+        "home": "/home/coder/project",
```
mwest1066, line 14: I think in this example we should actually make this `/home/coder`, so that it's the real home directory?

mwest1066:
@tdy looks great! Let me know when you've added the `interface.js` changes and I'll merge it.

exampleCourse/questions/demo/workspace/vscode/info.json
tdy: Hmm should we just rename the option then? Because I think we still need the `.../project` part to be configurable?

mwest1066: Why do we need that to be configurable? Can't you just make the `project` directory inside `questions/XXX/workspace/project` if you want it?

tdy: Ah ok

migrations/188_workspaces__port__add.sql
tdy: fixed in defff526a67fe0ef819f92991817ee3335b187e9

tdy:
I think it's working now, but for some reason `.bashrc` `.profile` `.bash_logout` aren't there anymore... not sure why.

tdy:
Previously code-server was expecting:
```javascript
        HostConfig: {
            ...
            Binds: [`${workspacePath}:'/home/coder/project`],
            ...
        },
        ...
        Volumes: {
            '/home/coder/project': {}
          },
```
Now it receives this (with `/project` inside the question workspace folder):
```javascript
        HostConfig: {
            ...
            Binds: [`${workspacePath}:'/home/coder`],
            ...
        },
        ...
        Volumes: {
            '/home/coder': {}
          },
```

mwest1066:
> I think it's working now, but for some reason `.bashrc` `.profile` `.bash_logout` aren't there anymore... not sure why.

Because we are mounting the entire home directory into the container, replacing whatever was there before. I think this is ok?

tdy:
Ah makes sense, then I think it's good to go from my end

PR #2773, opened 2020-07-27
Revert workspace controlContainer() refactor

tdy:
Plus a couple `interface.js` logging clarifications.


tdy:
fixed properly in #2774

PR #2774, opened 2020-07-27
Fix workspaces by emitting 'host-ready' when ready

mwest1066:


lib/workspace.js
```diff
@@ -1,44 +1,38 @@
...
+            } else {
+                let server_error = 'unknown error';
+                if ('json' in body) {
+                    if ('data' in body.json) {
+                        // need to decode the `data` array to be human readable
+                        server_error = new Buffer.from(body.json.data).toString();
+                    } else if ('message' in body.json) {
+                        server_error = body.json.message;
                     }
-                    logger.error(`controlContainer could not connect to workspace host: ${server_error}`);
```
tdy, line 30: Not sure if this was reverted intentionally or accidentally. This was @nicknytko's solution for making the json message human-readable.

PR #2777, opened 2020-07-28
Change workspace docker/directory names to be unique

tdy:
* [x] change to unique container/directory `localName=workspace-<UUID>` mapped to `s3Name=workspace-<ID>`
* [x] fix minor unrelated logging error in `lib/workspace.js`


workspace_host/interface.js
```diff
@@ -335,17 +336,48 @@ function _recursiveUploadJobManager(curDirPath, S3curDirPath) {
...
+        logger.info(`watch: workspace_id=${workspace_id}, localPath=${localPath}`);
+        let s3Path;
+        if (!workspace_id) {
+            logger.info(`watch return: workspace_id not mapped yet`);
+            return;
+        } else if (localPath === '') {
+            logger.info(`watch continue: empty (root) path`);
+            continue;
+        } else {
+            s3Path = `workspace-${workspace_id}/${localPath}`;
```
mwest1066, line 370: Should this be using `s3Name`?

tdy: I think there's a race condition where fswatcher instantly sees the dirs but `s3Name` might not have been assigned yet by the workspace initializer.

mwest1066: As we discussed in the meeting, I think we should be checking that the `id_workspace_mapper` is correctly initialized for this workspace and returning an error if it isn't. An error here probably means there is an old leftover workspace still running that we should be ignoring.

tdy: Handled in `_getWorkspaceByPath()` now:

https://github.com/PrairieLearn/PrairieLearn/pull/2777/files#diff-e7517976b89db12af19f23b917645289R422-R427

PR #2778, opened 2020-07-28
Update workspace hosts to use database

tdy:
For new `interface` servers:
* [x] create `workspace_hosts` row
* [x] set `workspace_hosts.hostname` to `config.workspaceContainerLocalhost`

For new containers:
* [x] query `workspace_hosts`
* [x] select host (currently `Math.random()`)
* [x] update `workspaces.workspace_host_id`

### DIFF
--- a/lib/workspace.js
+++ b/lib/workspace.js
@@ -1,38 +1,81 @@
 const ERR = require('async-stacktrace');
+const async = require('async');
 const request = require('request');
 
 const config = require('./config.js');
 const logger = require('./logger');
 
+const sqldb = require('@prairielearn/prairielib/sql-db');
+const sqlLoader = require('@prairielearn/prairielib/sql-loader');
+const sql = sqlLoader.loadSqlEquiv(__filename);
+
 module.exports = {
     controlContainer(workspace_id, action, callback) {
-        const options = {
-            json: {
-                workspace_id: workspace_id,
-                action: action,
-            },
-        };
-        request.post(`http://${config.workspaceContainerLocalhost}:${config.workspaceContainerPort}/`, options, (err, res, body) => {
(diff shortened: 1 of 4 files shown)


lib/workspace.js
```diff
@@ -1,38 +1,84 @@
...
-                workspace_id: workspace_id,
-                action: action,
+        async.waterfall([
+            // query available hosts
+            (callback) => {
+                sqldb.query(sql.select_workspace_hosts, {}, function(err, result) {
+                    if (err) {
+                        logger.error('Error querying workspace hosts', err);
+                        return;
+                    }
```
mwest1066, line 21: Should be `if (ERR(err, callback)) return;`

lib/workspace.js
```diff
@@ -1,38 +1,84 @@
...
+                            }
+                        }
+                        callback(new Error(`controlContainer could not connect to workspace host: ${server_error}`));
+                    }
+                });
+            },
+        ], function(err) {
+            if (err) {
+                logger.error('controlContainer error:', err);
             }
```
mwest1066, line 80: Should be `if (ERR(err, callback)) return;`

Error handling should almost always look like this ^^^

lib/workspace.sql
```diff
@@ -0,0 +1,10 @@
+-- BLOCK select_workspace_hosts
+SELECT * FROM workspace_hosts;
+
+-- BLOCK update_workspaces_workspace_host_id
+UPDATE
+    workspaces as w
+SET
+    workspace_host_id = $workspace_host_id
+WHERE
+    w.id = $workspace_id
```
mwest1066, line 10: Prefer to end with semicolon.

PR #2779, opened 2020-07-28
Fix workspace state machine

tdy:
* [x] create shared `updateState()` function in `lib/workspace`
* [x] call `updateState()` at state transitions
* [x] fix socket server in `lib/workspace`
* [x] write state transitions to `workspace_logs`


sprocs/workspaces_state_update.sql
```diff
@@ -0,0 +1,15 @@
...
+AS $$
+BEGIN
+    UPDATE
+        workspaces as w
+    SET
+        state = workspace_state::enum_workspace_state
+    WHERE
+        w.id = workspace_id;
+END;
+$$ LANGUAGE plpgsql VOLATILE;
```
tdy, line 15: Should we be writing to `workspace_logs` here in this sproc?

mwest1066: Good idea!

lib/workspaceSocket.js
```diff
@@ -27,24 +23,28 @@ module.exports.connection = function (socket) {
         console.log('Initializing remote workspace host.');
         workspace.controlContainer(workspace_id, 'init', (err) => {
             if (ERR(err, e => logger.error(e))) return;
-            module.exports._namespace.to(workspace_id).emit('host-ready');
+            
+            const state = 'running';
+            workspace.updateState(workspace_id, state, () => {
+                module.exports._namespace.to(workspace_id).emit('change:state', {workspace_id, state});
```
tdy, line 29: this manual callback is only here right now because the `lib/workspace#updateState` sockets are currently broken

lib/workspace.js
```diff
@@ -3,8 +3,25 @@ const request = require('request');
 
 const config = require('./config.js');
 const logger = require('./logger');
+const socketServer = require('./socket-server');
+
+const sqldb = require('@prairielearn/prairielib/sql-db');
+
+module.exports.init = function(callback) {
```
mwest1066, line 10: We are setting `module.exports.init` here, but then below we are setting `module.exports = {...}` so we will lose the init function. Include `init` below if you want it. But maybe we are never calling it at the moment?

lib/workspace.js
```diff
@@ -3,8 +3,25 @@ const request = require('request');
...
+    socketServer.io.on('connection', this.connection);
```
mwest1066, line 11: We are referencing a function called `this.connection` but it doesn't exist?

lib/workspace.js
```diff
@@ -3,8 +3,25 @@ const request = require('request');
...
+    callback(null);
+};
 
 module.exports = {
+    updateState(workspace_id, state, callback) {
+        // TODO: add locking
+        sqldb.call('workspaces_state_update', [workspace_id, state], function(err, _result) {
+            if (ERR(err, callback)) return;
+            logger.info(`[workspace.js] set workspaces.state to '${state}'`);
+            socketServer.io.of('/workspace').to(workspace_id).emit('change:state', {workspace_id, state});
```
mwest1066, line 21: Here we are using socket namespaces (the `of('/workspace')` part). But I don't think this is needed? Can't we just delete this everywhere and do `socketServer.io.of('/workspace').to(workspace_id).emit(...)`?

sprocs/workspaces_state_update.sql
tdy: Added in c51024b

tdy:
@mwest1066 I merged `lib/workspaceSocket` into `lib/workspace`.

Still getting the issue where `running` is being detected properly:
https://github.com/PrairieLearn/PrairieLearn/blob/c51024bd1a1c02d3e2946a1520651ff841711dc4/lib/workspace.js#L74-L77

but `launching` and `stopped` are not:
https://github.com/PrairieLearn/PrairieLearn/blob/c51024bd1a1c02d3e2946a1520651ff841711dc4/pages/workspace/workspace.js#L74-L77

PR #2791, opened 2020-07-30
News items and navbar link improvements

### LINKED ISSUE #2170: Add to news item footer contribute option
For each news item, PL is currently using hard coded HTML at the end of the item.

c.f. 
https://github.com/PrairieLearn/PrairieLearn/pull/2065/files/f0480304c5cc9f6d8a699137cd52d352e1d5c9cc..afa4c4b2e09426afa7efdaf9d79f7f2dfd5e1e60#diff-27f2d559ef70a82b9e71a4f609384623R78-R81

Instead, this should be included with the HTML view for displaying news items.

astrosticks:
This PR has some visual/maintenance improvements related to news items.

1. Gives a proper ARIA label to the news link in the user dropdown. A link that reads "News [3]" on-screen will be read as "News (3 unread)"

2. There was a call to action for open-source contributions that was copied throughout various news items, I've now moved that into the template so that it appears on all news items. This change also guarantees that all posts have some extra whitespace at the the bottoms of posts before the document closes. Resolves #2170.

3. Adding some semantic HTML5 elements to the news item page template.


pages/news_item/news_item.ejs
```diff
@@ -6,31 +6,38 @@
...
+        </p>
+      </header>
 
       <%- news_item_html %>
-
-    </div>
+    </article>
+    <aside class="container">
+      <hr class="mt-5">
+      <p class="text-right small">
```
tdy, line 38:
```suggestion
      <p class="text-right small border-top mt-5 pt-3">
```
Preferable to avoid styling html (`<hr>`, `&nbsp;`, empty `<p></p>`, etc). Alternatively, we can make the horizontal rule `aria-hidden`.

eliotwrobson:
Looks like this would resolve https://github.com/PrairieLearn/PrairieLearn/issues/2170

nwalters512 (approved):

news_items/028_java_grader/index.html
```diff
@@ -62,9 +62,4 @@
   To read more about new grader features you can check
   out <a href="https://prairielearn.readthedocs.io/en/latest/java-grader/">the
   documentation</a>.
-</p>
-
-<hr class="mt-5">
-<p class="text-right small">
-    Want to help make PrairieLearn better? It's open source and <a href="https://github.com/PrairieLearn/PrairieLearn">contributions are welcome!</a>
-</p>
+</p>
```
eliotwrobson, line 65: There's a few files without trailing newlines here.

eliotwrobson (commented):
One small change but everything else looks reasonable. I'm not super familiar with how the news items are set up, however.

news_items/028_java_grader/index.html
nwalters512: Should be fixed by 845bfdc.

eliotwrobson (approved):
Lgtm!

PR #2833, opened 2020-08-02
Add xtermjs workspace enhancements

nicknytko:
Adds:
- The host now only forks one pseudoterminal and each connection shares the same terminal.
- `LC_CTYPE` is set by default so `tmux` will work out of the box
- Copy + paste works correctly
- The terminal process will automatically restart when the running process dies


mwest1066:
Works for me!

PR #2834, opened 2020-08-03
Workspace graceful shutdown

polohan:


lib/workspace.js
```diff
@@ -123,6 +123,9 @@ module.exports = {
             if (state == 'uninitialized') {
                 await module.exports.initialize(workspace_id);
                 await module.exports.updateState(workspace_id, 'stopped');
+            } else {
+                await module.exports.controlContainer(workspace_id, 'destroy');
```
mwest1066, line 127: I think we can't send this message here, because the host might no longer exist at this point (it could be an old host). I think it would be more robust to have the running hosts check the DB to see whether any of their containers are `stopped` and delete them if so.

workspace_host/interface.js
```diff
@@ -116,54 +116,90 @@ app.post('/', function(req, res) {
 app.listen(config.workspaceDevHostPort, () => console.log(`Listening on port ${config.workspaceDevHostPort}`));
 
 // For detecting file changes
-var update_queue = {};  // key: path of file on local, value: action ('update' or 'remove').
+class jobQueue {
+    constructor(skip_ttl=5) {
+        this.ttl = skip_ttl;     // ttl for item on skip_queue
```
mwest1066, line 121: If I'm understanding correctly, this `ttl`/`skip_queue` mechanism is designed to prevent simultaneous uploads of the same file? But what happens if it's a big file and it takes longer than 5 update cycles to send it? Won't we try and upload it twice simultaneously in this case?

I'm wondering if it would be more robust to remove files from the `update_queue` one at a time when we begin uploading them, and have a separate `currently_uploading_queue` where we store which files are being uploaded at present. Then the algorithm would be:
1. When we get a `watch` update that a file is modified, add it to the `update_queue`.
2. When we run `_autoUpdateJobManager`, for each file that is on the `update_queue` and not on the `currently_updating_queue`, we create an upload job for that file, remove it from the `update_queue`, and add it to the `currently_updating_queue`.
3. When a file finishes uploading, we remove it from the `currently_updating_queue`.

I think that this algorithm guarantees that we will upload all modified files eventually and we will never try and upload the same file twice at the same time. I think it's also a lot simpler than the `skip_queue` logic.

workspace_host/interface.js
```diff
@@ -431,7 +470,7 @@ function _autoUpdateJobManager() {
             });
         }
     }
-    update_queue = {};
+    queue.resetUpdateQueue;
```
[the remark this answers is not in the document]
polohan: Oh yeah, good catch.

workspace_host/interface.js
```diff
@@ -503,38 +542,38 @@ function _syncPullContainer(workspace_id, callback) {
...
+//                 status.push(res[2]);
+//             }
+//         });
+//         if (status.length != 0) {
+//             callback(status);
+//         } else {
+//             callback(null, workspace_id);
+//         }
+//     });
+// }
```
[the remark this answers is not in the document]
polohan: Will do.

workspace_host/interface.js
polohan: Oh, skip_queue is for skipping the update of a file when the file was just downloaded or when we delete it because the host isn't running anymore.

polohan: And I think for your algorithm makes sense, but since our `interface.js` is going to run on EC2, I don't think it will be such a big issue right away(?).  Maybe add this to TODO?

lib/workspace.js
polohan: So the concern here is that there might be nothing to destroy?

polohan: I mean, I can totally remove these lines and rely on whatever method we come up with later to recycle those resources. I just put it there so whenever a user clicks the upload destroy the container and delete the folder on the host.

mwest1066:
Implemented in #2866

PR #2837, opened 2020-08-04
Add workspace host endpoint to get graded files

polohan:


workspace_host/interface.js
```diff
@@ -719,3 +734,51 @@ function destroySequence(workspace_id, res) {
         }
     });
 }
+
+function gradeSequence(workspace_id, res) {
+    var workspace_dir = `${workspacePrefix}/${id_workspace_mapper[workspace_id].localName}/project`;
+    var graded_file_list = id_workspace_mapper[workspace_id].settings.workspace_graded_files.map((file) => {return path.join(workspace_dir, file);});
+    graded_file_list = ['starter_code.c', 'starter_code.h'].map((file) => {return path.join(workspace_dir, file);});    // since in the DB the workspace_graded_file col is empty, hardcode file list
```
tdy, line 741: > // since in the DB the workspace_graded_file col is empty

Hmm not sure what this means? `demo/workspace/{vscode,jupyterlab,xtermjs}` all have `gradedFiles` synced to the DB.

pages/workspace/workspace.js
```diff
@@ -22,6 +22,9 @@ router.get('/:workspace_id/:action', asyncHandler(async (req, res, next) => {
         const state = 'stopped';
         await workspace.updateState(workspace_id, state);
         res.redirect(`/workspace/${workspace_id}`);
+    } else if (action === 'grade') {
+        await workspace.controlContainer(workspace_id, 'grade');
+        res.redirect(`/workspace/${workspace_id}`);     // refresh anyway
```
mwest1066, line 27: We don't need an action here --- the `grade/` endpoint will only be used by the PL servers themselves during the grading process.

lib/workspace.js
```diff
@@ -106,11 +106,17 @@ module.exports = {
...
             action: action,
         };
+
         const res = await fetch(`http://${workspace_host.hostname}/`, {
             method: 'post',
             body: JSON.stringify(postJson),
             headers: {'Content-Type': 'application/json'},
         });
+        if (action === 'grade') {
+            const dest = fs.createWriteStream('test.tgz');
```
mwest1066, line 116: Maybe `test.tgz` is not the best name here? Could we call this `gradedFiles.tgz` to fit the purpose better?

workspace_host/interface.js
```diff
@@ -106,64 +108,104 @@ app.post('/', function(req, res) {
...
     }
 });
 
+
+
 app.listen(config.workspaceDevHostPort, () => console.log(`Listening on port ${config.workspaceDevHostPort}`));
 
 // For detecting file changes
-var update_queue = {};  // key: path of file on local, value: action ('update' or 'remove').
+class jobQueue {
```
mwest1066, line 125: It looks like this PR includes the changes from #2834. I don't want to merge #2834 yet, so can you remove these changes from this PR so that this one only includes the `grade/` endpoint addition?

workspace_host/interface.js
```diff
@@ -719,3 +734,51 @@ function destroySequence(workspace_id, res) {
...
+                }
+            }
+            return null;
+        },
+        (callback) => {
+            tar.create(
+                {
+                  gzip: true,
+                  file: zip_file_name,
+                  sync: false,
```
mwest1066, line 762: nit: Indentation looks wrong here.

workspace_host/interface.js
```diff
@@ -719,3 +734,51 @@ function destroySequence(workspace_id, res) {
...
+    var timestamp = new Date().toISOString().replace(/T/, '*').replace(/\..+/, '');
+    var zip_file_name = `workspace-${workspace_id}-${timestamp}.tgz`;
+
+    async.series([
+        async () => {
+            for (var file_path in graded_file_list) {
+                try {
+                    await fsPromises.lstat(file_path);
+                } catch(err) {
+                    return(`Required graded file ${path.basename(file_path)} does not exist.`);
```
mwest1066, line 752: Rather than returning an error if the file is missing, I think it would be better to return all the files we do have (possibly zero) and let the PL server figure out what to do with errors.

workspace_host/interface.js
polohan: Oh, that's just for my own testing purposes. I forgot to remove it.

lib/workspace.js
polohan: Changed.

workspace_host/interface.js
polohan: Removed.

workspace_host/interface.js
polohan: Done. And I modified the code so it doesn't contain the unnecessary folder in the zip file.

mwest1066:
Incorporated into #2841

PR #2840, opened 2020-08-05
Fix workspace proxying for SocketIO

nicknytko:


mwest1066:
This is awesome! Completely fixes the Safari issues and also eliminates those confusing `/socket-io` path mismatches.

PR #2855, opened 2020-08-09
Checkbox allow none correct

bojinyao:
Fixes https://github.com/ace-lab/pl-ucb-cs10/issues/61

Add `allow-none-correct` attribute to `<pl-checkbox>` of type: boolean, default False

When set to True:
- Checkbox question will have "None" as the correct answer (i.e display all incorrect options) with probability `1/(len_correct + 1)`, so that when no correct option is provided, the question will default to having "None" as correct all the time.
- Student can click `Save & Grade` when no option is selected, and no format error will be raised.
- Does not affect grading options such as partial credit calculations.
- Similar to `none-of-the-above` and `all-of-the-above` in `<pl-multipe-choice>`, enabling this element could reduce the number of options displayed. This is to maintain consistent number of displayed options across all variants. The maximum number of options it is capable of displaying is the same as *number of incorrect options provided* because when "None" is correct, no correct option can be displayed

New Exceptions:
- when `allow-none-correct` is enabled; but the specified `number-answers` cannot be met (due to insufficient number of incorrect options), an error will be raised that tells the user how many more incorrect options are needed.


bojinyao:
@nicknytko Can you take a look at the CI tests? I can't figure out which 2 test cases are causing errors 😅 Thank you!!!

nwalters512:
@bojinyao I'm running through old PRs right now, do you still want to see this through?

nwalters512:
Going to close this to clean up stale PRs; please reopen in a new PR if you'd still like to contribute this!

PR #2860, opened 2020-08-11
Refactor S3 helpers to lib/aws

tdy:
Subsumed by #2870 and #2863 

---

I needed these helpers, so refactoring for general reuse.

---

Converting to draft now that this is a code refactor instead of just location refactor.

* [x] fix dir trailing slash
* [ ] convert fully to async style

### DIFF
--- a/lib/aws.js
+++ b/lib/aws.js
@@ -36,7 +36,7 @@ module.exports.getS3RVERConfiguration = function() {
     });
 };
 
-module.exports.uploadToS3 = async function (filePath, isDirectory, bucket, S3FilePath, localPath, callback) {
+module.exports.uploadToS3 = async function (filePath, isDirectory, bucket, S3FilePath, localPath) {
     const s3 = new AWS.S3();
 
     let body;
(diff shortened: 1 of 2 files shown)


nicknytko:
While we're refactoring stuff, can we convert the functions to be one of async/await or callback style only?  It seems we're currently mixing the two and I'm worried this will lead to more confusion in the future.

lib/aws.js
```diff
@@ -35,3 +35,53 @@ module.exports.getS3RVERConfiguration = function() {
         endpoint: new AWS.Endpoint('http://localhost:5000'),
     });
 };
+
+module.exports.uploadToS3 = async function (filePath, isDirectory, bucket, S3FilePath, localPath, callback) {
```
mwest1066, line 39: Following up on @nicknytko’s comment, `async` functions generally should not take `callback` parameters. I don’t mind whether this is `async`-style or `callback`-style, but it really shouldn’t be both at once.

lib/aws.js
```diff
@@ -35,3 +35,53 @@ module.exports.getS3RVERConfiguration = function() {
...
+    const s3 = new AWS.S3();
+
+    let body;
+    if (isDirectory) {
+        body = '';
+        S3FilePath += '/';
```
mwest1066, line 45: What if the file already ends in `/`?

lib/aws.js
```diff
@@ -35,3 +35,53 @@ module.exports.getS3RVERConfiguration = function() {
...
+    } else {
+        try {
+            body = await fs.promises.readFile(filePath);
+        } catch (err) {
+            callback(null, [filePath, S3FilePath, err]);
```
mwest1066, line 50: I don’t like the way we return a length-3 array on error and a string “OK” on success. We should always return the same type. Also we should return errors normally as actual errors. We can capture the error in the caller if we want to suppress it.

lib/aws.js
tdy: Changed to `s3FilePath += s3FilePath.endsWith('/') ? '' : '/';`

PR #2861, opened 2020-08-11
Fix workspace dirs when running without external grader

tdy:


tdy:
I believe this part should actually stay the same?

https://github.com/PrairieLearn/PrairieLearn/blob/648490c8f547c728384db0c2d5ff5ed3b5122894/workspace_host/interface.js#L662-L664

workspace_host/interface.js
```diff
@@ -187,7 +187,7 @@ async.series([
 
 // For detecting file changes
 var update_queue = {};  // key: path of file on local, value: action ('update' or 'remove').
-const workspacePrefix = process.env.HOST_JOBS_DIR ? '/jobs/workspaces' : config.workspaceJobsDirectory;
+const workspacePrefix = config.workspaceJobsDirectory;
```
mwest1066, line 190: Can we do something similar for: https://github.com/PrairieLearn/PrairieLearn/blob/b6cb352938d23f22dd14df5ea9e12605364d839d/workspace_host/interface.js#L34

tdy: Done. Working on my end (dockerized).

docs/workspaces/index.md
```diff
@@ -96,19 +96,25 @@ docker run -it --rm -p 3000:3000 `
...
 
 For development, run the docker command with a final extra argument of `/PrairieLearn/docker/start_workspace.sh` to load PL and the workspace host interface in separate tmux panes.
 
+## Running locally (natively, not on Docker)
+
+Set these variables in your `config.json`:
+
+* `workspaceJobsDir` 
+* `workspaceGradedFilesReceiveDirectory` 
+* `workspaceGradedFilesSendDirectory`
```
tdy, line 120: Maybe could be fleshed out more

PR #2863, opened 2020-08-11
Refactor asynchronous workspaces code

nicknytko:
`admZip` has been replaced with `archiver` and `unzipper` stream based libraries
cleaned up async/callback hybrid code to be async only
refactored `s3Sync` to `uploadDirectoryToS3Async`
refactored the functions that return `'OK'` or an error array to actually throw errors now

confirmed works on pl-dev


lib/workspace.js
```diff
@@ -18,38 +19,37 @@ const sql = sqlLoader.loadSqlEquiv(__filename);
 
 const zipPrefix = process.env.HOST_JOBS_DIR ? '/jobs/workspace_receive_zips' : config.workspaceGradedFilesReceiveDirectory;
 
-// https://stackoverflow.com/a/46213474/13138364
-const s3Sync = function (s3Path, bucketName) {
+async function uploadDirectoryToS3Async(s3Path, bucket_name) {
```
tdy, line 22:
```suggestion
async function uploadDirectoryToS3Async(s3Path, bucketName) {
```

lib/workspace.js
```diff
@@ -18,38 +19,37 @@ const sql = sqlLoader.loadSqlEquiv(__filename);
...
 
-    function walkSync(currentDirPath, callback) {
-        fs.readdirSync(currentDirPath).forEach(function (name) {
-            const filePath = path.join(currentDirPath, name);
-            const stat = fs.statSync(filePath);
+    async function walk_directory(directory) {
+        const files = await fsPromises.readdir(directory);
+        await async.each(files, async (file) => {
+            const file_path = path.join(directory, file);
+            const stat = await fsPromises.stat(file_path);
```
tdy, line 29:
```suggestion
    async function walkDirectory(directory) {
        const files = await fsPromises.readdir(directory);
        await async.each(files, async (file) => {
            const filePath = path.join(directory, file);
            const stat = await fsPromises.stat(filePath);
```

lib/workspace.js
```diff
@@ -18,38 +19,37 @@ const sql = sqlLoader.loadSqlEquiv(__filename);
...
+                    Body: file_body,
+                };
+                try {
+                    await s3.putObject(params).promise();
+                    logger.info(`[workspace.js] synced ${bucket_path} to ${bucket_name}`);
+                } catch (err) {
+                    logger.error(`[workspace.js] error syncing ${bucket_path}:\n ${err}`);
+                }
+            } else {
+                return await walk_directory(file_path);
```
tdy, line 45:
```suggestion
                const fileBody = await fsPromises.readFile(filePath);
                const bucketPath = filePath.substring(s3Path.length + 1);
                const params = {
                    Bucket: bucketName,
                    Key: bucketPath,
                    Body: fileBody,
                };
                try {
                    await s3.putObject(params).promise();
                    logger.info(`[workspace.js] synced ${bucketPath} to ${bucketName}`);
                } catch (err) {
                    logger.error(`[workspace.js] error syncing ${bucketPath}:\n ${err}`);
                }
            } else {
                return await walkDirectory(filePath);
```

lib/workspace.js
```diff
@@ -18,38 +19,37 @@ const sql = sqlLoader.loadSqlEquiv(__filename);
...
-            if (err) {
-                console.log(err);
-            } else {
-                console.log(`[workspace.js] synced ${bucketPath} to ${bucketName}`);
-            }
-        });
-    });
-};
+    console.log(s3Path);
+    await walk_directory(s3Path);
```
tdy, line 51:
```suggestion
    await walkDirectory(s3Path);
```

workspace_host/interface.js
```diff
@@ -331,44 +327,34 @@ async function _uploadToS3(filePath, isDirectory, S3FilePath, localPath, callbac
...
-    s3.deleteObject(deleteParams, function(err, _data) {
-        if (err) {
-            callback(null, [filePath, S3FilePath, err]);
-            return;
-        }
-        logger.info(`Deleted ${localPath}`);
-        callback(null, 'OK');
-    });
+    await s3.deleteObject(deleteParams).promise();
+    logger.info(`Deleted ${localPath}`);
```
tdy, line 355:
```suggestion
    await s3.deleteObject(deleteParams).promise();
    logger.info(`Deleted s3://${config.workspaceS3Bucket}/${S3FilePath}`);
```

lib/workspace.js
```diff
@@ -18,38 +19,36 @@ const sql = sqlLoader.loadSqlEquiv(__filename);
...
+                    Key: bucketPath,
+                    Body: fileBody,
+                };
+                try {
+                    await s3.putObject(params).promise();
+                    logger.info(`[workspace.js] synced ${bucketPath} to ${bucketName}`);
+                } catch (err) {
+                    logger.error(`[workspace.js] error syncing ${bucketPath}:\n ${err}`);
+                }
+            } else {
```
mwest1066, line 44: Should we check whether this is a directory? As opposed to some other kind of weird file?

lib/workspace.js
```diff
@@ -18,38 +19,36 @@ const sql = sqlLoader.loadSqlEquiv(__filename);
...
+                return await walkDirectory(filePath);
```
mwest1066, line 45: Either upload the directory itself to capture empty directories, or make a code comment that we don't do this.

workspace_host/interface.js
```diff
@@ -57,7 +57,8 @@ app.post('/', function(req, res) {
     } else if (action == 'reset') {
         resetSequence(workspace_id, res);
     } else if (action == 'destroy') {
-        destroySequence(workspace_id, res);
+        /* TODO: Add an actual destroy sequence? */
+        res.status(500).send('Action "destroy" unimplemented');
```
mwest1066, line 61: I think we can just get rid of `destroy`.

workspace_host/interface.js
```diff
@@ -186,9 +187,9 @@ async.series([
 });
 
 // For detecting file changes
-var update_queue = {};  // key: path of file on local, value: action ('update' or 'remove').
+let update_queue = {};  // key: path of file on local, value: action ('update' or 'remove').
 const workspacePrefix = config.workspaceJobsDirectory;
-var interval = 5; // the base interval of pushing file to S3 and scanning for file change in second
+const interval = 5; // the base interval of pushing file to S3 and scanning for file change in second
```
mwest1066, line 192: Make it a `config` variable?

workspace_host/interface.js
```diff
@@ -250,7 +250,7 @@ async function _getAvailablePort(workspace_id, lowest_usable_port, callback) {
         });
     }
 
-    for (var i = lowest_usable_port; i < 65535; i++) {
+    for (let i = lowest_usable_port; i < 65535; i++) {
```
mwest1066, line 253: Should we randomly choose ports? Or skip ports we already know are allocated in the DB? While still doing the real check, I mean.

What about locking here? Two `init` requests could conflict and choose the same port. One way to handle this is to have a global variable for available ports, and we take them off before trying them.

workspace_host/interface.js
```diff
@@ -516,18 +479,8 @@ function _autoUpdateJobManager() {
...
-            if (res != 'OK') {
-                res[2].fileLocalPath = res[0];
-                res[2].fileS3Path = res[1];
-                status.push(res[2]);
-            }
-        });
-        if (status.length != 0) {
-            logger.error(`Error during file sync: ${status}`);
-        }
+    async.parallel(jobs, function(err) {
```
mwest1066, line 482: Use `parallelLimit()`?

workspace_host/interface.js
```diff
@@ -562,61 +515,17 @@ function _syncPullContainer(workspace_id, callback) {
...
-            if (res != 'OK') {
-                res[2].fileLocalPath = res[0];
-                res[2].fileS3Path = res[0];
-                status.push(res[2]);
-            }
-        });
-        if (status.length != 0) {
-            callback(status);
-        } else {
+        async.parallel(jobs, function(err) {
```
mwest1066, line 525: `parallelLimit()`?

workspace_host/interface.js
nicknytko: Working on a better implementation here -> https://github.com/PrairieLearn/PrairieLearn/blob/da422dfe8b853202618931a4d0274544e13a5e37/sprocs/workspace_host_allocate_port.sql

But yes, I think we should consider locking or something to prevent concurrency issues.

PR #2866, opened 2020-08-12
Allow workspace host to be restartable

nicknytko:
- Adds `launch_uuid` and `launch_port` columns to the workspace table.
- On startup and every 10 minutes all running containers are automatically synced to S3
- Port allocation is now a two step process where both the db and the actual port are checked.  This is (presumably) safe from concurrency issues.
- All workspace_id_mapper etc stuff has been rewritten to use db queries
- Every minute recently stopped containers and runaway containers (those without matching launch_uuid in db) are killed and removed

Note you may get an error while running this locally on docker.  This is because the workspace server is run before the migration finishes -- I just run `init.sh`, ctrl-c, then run it again and it works fine.


lib/workspace.js
```diff
@@ -91,6 +60,60 @@ module.exports = {
...
+                            logger.info(`synced ${bucketPath} to ${bucketName}`);
+                        } catch (err) {
+                            logger.error(`error syncing ${bucketPath}:\n ${err}`);
+                        }
+                    }
+                } else if (stat.isDirectory()) {
+                    try {
+                        await s3.putObject({
+                            Bucket: bucketName,
+                            Key: bucketPath + '/', /* Put a forward slash at the end to create a directory */
```
tdy, line 102:
```suggestion
                            Key: bucketPath + (bucketPath.endsWith('/') ? '' : '/'), /* Put a forward slash at the end to create a directory */
```
Not sure if this is the cleanest way to do this, but I think Matt was concerned about appending an extra `'/'` if there was already a slash.

workspace_host/interface.js
```diff
@@ -235,11 +252,173 @@ watcher.on('unlinkDir', filename => {
...
+        try {
+            /* Try to kill, but don't care if it's already dead */
+            const container = await _getDockerContainerByLaunchUuid(ws.launch_uuid);
+            await container.kill();
+            await container.remove();
+        } catch (err) {
+            logger.error(err);
+        }
+        await sqldb.queryAsync(sql.set_workspace_launch_uuid, { workspace_id: ws.id, launch_uuid: null });
+        await sqldb.queryAsync(sql.set_workspace_launch_port, { workspace_id: ws.id, launch_port: null });
```
mwest1066, line 307: We could also set `workspace_host_id = null`. That way we are saying this WS is not associated with us in any way anymore. Then it won't show up the next time we check for stopped containers, and we can change `get_recently_stopped_workspaces` to get all stopped workspaces, not just recently stopped ones.

lib/workspace.js
```diff
@@ -91,6 +60,60 @@ module.exports = {
...
+                if (stat.isFile()) {
+                    if (!ignoreList.has(bucketPath)) {
+                        const fileBody = await fsPromises.readFile(filePath);
+                        try {
+                            await s3.putObject({
+                                Bucket: bucketName,
+                                Key: bucketPath,
+                                Body: fileBody,
+                            }).promise();
+                            logger.info(`synced ${bucketPath} to ${bucketName}`);
```
mwest1066, line 93:
```suggestion
                            debug(`synced ${bucketPath} to ${bucketName}`);
```

lib/workspace.js
```diff
@@ -91,6 +60,60 @@ module.exports = {
...
+                        }
+                    }
+                } else if (stat.isDirectory()) {
+                    try {
+                        await s3.putObject({
+                            Bucket: bucketName,
+                            Key: bucketPath + '/', /* Put a forward slash at the end to create a directory */
+                            Body: '',
+                        }).promise();
+                        logger.info(`synced directory ${bucketPath} to ${bucketName}`);
```
mwest1066, line 105:
```suggestion
                        debug(`synced directory ${bucketPath} to ${bucketName}`);
```

workspace_host/interface.js
```diff
@@ -163,6 +157,29 @@ async.series([
...
+        await async.eachSeries(result.rows, async (ws) => {
+            if (ws.state == 'launching') {
+                /* We don't know what state the container is in, kill it and let the user
+                   retry initializing it */
+                try {
+                    const container = await _getDockerContainerByLaunchUuid(ws.launch_uuid);
+                    await container.kill();
+                    await container.remove();
+                } catch (err) {
+                    logger.info(`Couldn't kill container ${ws.launch_uuid}: ${err}`);
```
mwest1066, line 173:
```suggestion
                    logger.error(`Couldn't kill container ${ws.launch_uuid}: ${err}`);
```

workspace_host/interface.js
```diff
@@ -235,11 +252,173 @@ watcher.on('unlinkDir', filename => {
...
+ * Allocates and returns an unused port for a workspace.  This will insert the new port into the workspace table.
+ * @param {object} workspace Workspace object, should at least contain an id.
+ * @return {integer} Port that was allocated to the workspace.
+ */
+let _allocateContainerPortLock = false;
+let _allocateContainerPortLockWaiting = [];
+async function _allocateContainerPort(workspace) {
+    /* Lock and unlock functions.  If the lock is free then this will return immediately.
+       If someone already has the lock then this will wait until someone calls unlock() */
+    async function lock() {
```
mwest1066, line 391: At some point it would be nice to make a `lib/local-lock.js` with these functions in it, maybe in an object so we can do `let _allocateContainerPortLock = new LocalLock()` and then `_allocateContainerPortLock.lock()`, etc.

workspace_host/interface.sql
```diff
@@ -37,3 +37,68 @@ FROM
...
     w.id = $workspace_id
     AND w.workspace_host_id = wh.id;
+
+-- BLOCK get_workspace
+SELECT
+    w.*
+FROM
+    workspaces AS w
+WHERE
+    w.id = $workspace_id;
```
mwest1066, line 47: For safety, we could also pass in `$instance_id` here and make sure that matches. Of course this shouldn't be needed, but if PL ever gets confused about which host gets which message (maybe a race condition?) then that would help contain the damage.

mwest1066: Actually, we may as well change all the queries here (including the UPDATEs) to restrict to `$instance_id`

workspace_host/interface.js
```diff
@@ -627,47 +803,56 @@ function _createContainer(workspace_id, port, settings, callback) {
...
 
-    id_workspace_mapper[workspace_id] = {};
-    id_workspace_mapper[workspace_id].localName = `workspace-${uuidv4()}`;
-    id_workspace_mapper[workspace_id].s3Name = `workspace-${workspace_id}/current`;
+    const uuid = uuidv4();
+    const workspace = {
+        'id': workspace_id,
+        'launch_uuid': uuid,
+        'local_name': `workspace-${uuid}`,
+        's3_name': `workspace-${workspace_id}`,
```
tdy, line 833:
```suggestion
        's3_name': `workspace-${workspace_id}/current`,
```
Fixes the container-s3 syncing regression.

lib/local-locks.js
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+class LocalLock {
+    constructor() {
+        this.lock = false;
+        this.waiting = [];
+    }
+
+    /**
+     * Trys to acquire the lock.  If it is free this will return immediately.
```
tdy, line 14:
```suggestion
     * Tries to acquire the lock.  If it is free this will return immediately.
```

lib/local-locks.js
```diff
@@ -0,0 +1,72 @@
...
+                });
+                await promise;
+            } else {
+                throw new Error("Couldn't obtain lock!");
+            }
+        }
+    }
+
+    /**
+     * Trys to acquire the lock.  If it is free this will return immediately.
```
tdy, line 36:
```suggestion
     * Tries to acquire the lock.  If it is free this will return immediately.
```

workspace_host/interface.js
```diff
@@ -830,7 +806,7 @@ function initSequence(workspace_id, useInitialZip, res) {
         'id': workspace_id,
         'launch_uuid': uuid,
         'local_name': `workspace-${uuid}`,
-        's3_name': `workspace-${workspace_id}`,
+        's3_name': `workspace-${workspace_id}/current`,
```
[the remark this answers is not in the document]
nicknytko: I suppose it's on purpose since we shouldn't have a `launch_uuid` or `launch_port` yet at this point, and we're going through and setting these fields up.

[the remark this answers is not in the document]
tdy: ah nevermind, `launch_uuid` hasn't been set by the `set_workspace_launch_uuid` yet

lib/local-locks.js
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+class LocalLock {
+    constructor() {
+        this.lock = false;
+        this.waiting = [];
+    }
+
+    /**
+     * Tries to acquire the lock.  If it is free this will return immediately.
```
nicknytko, line 14: Would ideally like another set of eyes on this locking stuff to make sure something isn't (blatantly) wrong here

lib/local-locks.js
```diff
@@ -0,0 +1,72 @@
...
+            const next = this.waiting.shift();
+            next();
+        } else {
+            /* Only unlock if we're the last person in the queue to use this */
+            this.lock = false;
+        }
+    }
+}
+
+module.exports.LocalLock = LocalLock;
```
mwest1066, line 72: A common pattern here is to do `module.exports = LocalLock;` and name the file `local-lock.js` (no `s`).

Then in callers you can do:
```
const LocalLock = require(`local-lock`);

const myLock = new LocalLock();
```

An example of this style is `archiver`: https://github.com/PrairieLearn/PrairieLearn/blob/95a25e75248f5eb6fd0ba1e31a284765fb7b1bac/lib/workspace.js#L25

I have no particular opinion about whether you should change to this style, I just wanted to point it out as an option.

lib/local-locks.js
```diff
@@ -0,0 +1,72 @@
+/**
+ * Implementation of a local JavaScript lock.
+ * Waiting threads (?) are placed in a queue and locks are obtained in the order
```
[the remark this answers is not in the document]
nicknytko: thank you i was struggling with what the right word to put here was :slightly_smiling_face:

workspace_host/interface.js
```diff
@@ -235,11 +258,193 @@ watcher.on('unlinkDir', filename => {
...
+        const name = container_info.Names[0].substring(1); /* Remove the preceding forward slash */
+        if (!name.startsWith('workspace-') || db_workspaces_uuid_set.has(name)) return;
+        await dockerAttemptKillAndRemove(container_info.Id);
+    });
+}
+
+setInterval(async () => {
+    await pruneStoppedContainers();
+    await pruneRunawayContainers();
+}, config.workspaceHostPruneContainersSec * 1000);
```
mwest1066, line 345: Technically we could have multiple copies of this running simultaneously?

nicknytko: use a lock? :grinning:

PR #2870, opened 2020-08-12
Fix workspace S3 race condition using initial/current sync scheme

tdy:
* [x] Change main server to copy initial files to `s3://workspaces/workspace-[N]/initial.zip`
* [x] Change main server to also copy initial files to `s3://workspaces/workspace-[N]/current/` as individual files
* [x] Change `interface.init` webhook to receive `req.body.state` and use it to set/pass `useInitialZip` to `initSequence()`
* [x] If `useInitialZip` is `true`, change `initSequence()` to copy/unpack `initial.zip` locally rather than syncing from S3 as usual
* [x] Change `interface.js` file watching/copying code to use `workspace-[N]/current/` rather than just `workspace-[N]`


lib/workspace.js
```diff
@@ -17,7 +18,49 @@ const sqlLoader = require('@prairielearn/prairielib/sql-loader');
...
+            resolve(zip);
+        });
+    });
+}
+
+const uploadFileToS3Async = async function (localPath, s3Path) {
+    const s3 = new AWS.S3();
+    const bucketName = config.workspaceS3Bucket;
+    const bucketPath = s3Path.substring(bucketName.length + 1);
+    const body = await fs.promises.readFile(localPath);
```
mwest1066, line 46: Isn't this just:
```
    const params = {
         Bucket: config.workspaceS3Bucket,
         Key: s3Path,
         Body: body,
     };
     await s3.upload(params).promise();
```

(you have this code earlier in the PR)

workspace_host/interface.js
```diff
@@ -46,14 +48,16 @@ app.use(bodyParser.json());
 
 // TODO: refactor into RESTful endpoints (https://github.com/PrairieLearn/PrairieLearn/pull/2841#discussion_r467245108)
 app.post('/', function(req, res) {
-    var workspace_id = req.body.workspace_id;
-    var action = req.body.action;
+    const workspace_id = req.body.workspace_id;
+    const state = req.body.state;
```
mwest1066, line 52: Rather than sending the current state, let's send a parameter called `useInitialZip` that defaults to `false`.

workspace_host/interface.js
```diff
@@ -664,18 +645,26 @@ function _startContainer(workspace_id, container, callback) {
...
 
     // send 200 immediately to prevent socket hang up from _pullImage()
     res.status(200).send(`Container for workspace ${workspace_id} initialized.`);
 
     async.waterfall([
-        (callback) => {_syncPullContainer(workspace_id, callback);},
+        (callback) => {
+            if (useInitialZip) {
+                logger.info(`Bootstrapping workspace with initial.zip`);
+                _syncInitialZip(workspace_id, callback);
```
mwest1066, line 662:
```suggestion
                _syncInitialZip(workspace_id, (err) => {
                    if (ERR(err, callback)) return;
                    callback(null);
                });
```

workspace_host/interface.js
```diff
@@ -664,18 +645,26 @@ function _startContainer(workspace_id, container, callback) {
...
+            } else {
+                logger.info(`Syncing workspace from S3`);
+                _syncPullContainer(workspace_id, callback);
```
mwest1066, line 665: same change here

lib/workspace.js
```diff
@@ -124,6 +144,7 @@ module.exports = {
         const postJson = {
             workspace_id: workspace_id,
             action: action,
+            useInitialZip: state == 'uninitialized',
```
tdy, line 147: @mwest1066 Or should I be passing in `useInitialZip` on each `controlContainer` call (that also feels a bit odd since `useInitialZip` only applies to the `init` action)? Can we determine `useInitialZip` using on the `state` here, even for to-be-implemented actions?

mwest1066: I think we should not pass `state` to `controlContainer`. Instead I suggest adding an `options` argument like this: `controlContainer(workspace_id, action, options={})`. The `options` value should be an object. Inside `controlContainer` we can do:
```
        const postJson = {
            workspace_id: workspace_id,
            action: action,
            ...options,
        };
```

mwest1066: This means that we can either do `controlContainer(workspace_id, 'getGradedFiles')` or `controlContainer(workspace_id, 'init', {useInitialZip: true})`.

mwest1066: And in `interface.js` for the `init` handler we can do `const useInitialZip = _.get(options, 'useInitialZip', false)` to get it with a default value.

PR #2876, opened 2020-08-13
Add created_at and state_updated_at timestamps for workspaces

tdy:
* [x] `created_at` default `now()`
* [x] `state_updated_at` on each update
* [x] snuck in silencing of the proxy logger


nicknytko:
Can you move your migrations to start at 198?  I'm using 197 in https://github.com/PrairieLearn/PrairieLearn/pull/2866 :grinning:

tdy:
> 
> 
> Can you move your migrations to start at 198? I'm using 197 in #2866 😀

Done 😄

database/tables/workspaces.pg
```diff
@@ -1,9 +1,11 @@
 columns
+    created_at: timestamp with time zone not null default CURRENT_TIMESTAMP
     heartbeat_at: timestamp with time zone
     hostname: text
     id: bigint not null default nextval('workspaces_id_seq'::regclass)
     launched_at: timestamp with time zone
     state: enum_workspace_state not null default 'uninitialized'::enum_workspace_state
+    state_updated_at: timestamp with time zone
```
mwest1066, line 8: I suggest making this a default current_timestamp as well (and maybe not null) so that it will be set to be not-null when the row is first created. As we saw with the heartbeat, this is generally more useful to have non-nulls.

PR #2879, opened 2020-08-14
Add workspace reboot confirmation dialog

tdy:
pl-workspace-reboot-confirm

### DIFF
--- a/pages/workspace/workspace.ejs
+++ b/pages/workspace/workspace.ejs
@@ -24,7 +24,7 @@
           </ul>
         </div>
         <div class="modal-footer">
-          <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
+          <button type="button" class="btn btn-secondary" data-dismiss="modal">Cancel</button>
           <a class="btn btn-info" href="/pl/workspace/<%= workspace_id %>/reboot">
             <i class="fas fa-sync text-white" aria-hidden="true"></i>
             Reboot
--- a/pages/workspace/workspace.ejs
+++ b/pages/workspace/workspace.ejs
@@ -65,7 +65,7 @@
               data-content='<ul class="list-group list-group-flush">
                               <li class="list-group-item p-1"><span class="badge badge-secondary mr-1"><i class="fas fa-fw fa-redo"></i> Refresh</span>Reload this tab at any time, or close and re-open it.</li>
                               <li class="list-group-item p-1"><span class="badge badge-secondary mr-1"><i class="fas fa-fw fa-sync"></i> Reboot</span>Reboot the virtual machine.</li>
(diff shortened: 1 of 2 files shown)


pages/workspace/workspace.ejs
```diff
@@ -7,6 +7,32 @@
...
+        </div>
+        <div class="modal-body">
+          <p>Are you sure you want to reboot the virtual machine?<p>
+          <ul class="fa-ul">
+            <li><span class="fa-li"><i class="fas fa-window-restore" aria-hidden="true"></i></span>Your workspace will reboot into this same tab.</li>
+            <li><span class="fa-li"><i class="fas fa-save" aria-hidden="true"></i></span>Your files will remain intact through the reboot.</li>
+          </ul>
+        </div>
+        <div class="modal-footer">
+          <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
```
mwest1066, line 27: I think we are generally using `Cancel`? https://github.com/PrairieLearn/PrairieLearn/blob/d59233ec111d0b0daff2c06574d40acdafaf8e20/pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs#L40

pages/workspace/workspace.ejs
```diff
@@ -39,11 +65,10 @@
               data-content='<ul class="list-group list-group-flush">
                               <li class="list-group-item p-1"><span class="badge badge-secondary mr-1"><i class="fas fa-fw fa-redo"></i> Refresh</span>Reload this tab at any time, or close and re-open it.</li>
                               <li class="list-group-item p-1"><span class="badge badge-secondary mr-1"><i class="fas fa-fw fa-sync"></i> Reboot</span>Reboot the virtual machine.</li>
-                              <li class="list-group-item p-1"><span class="badge badge-secondary mr-1"><i class="fas fa-fw fa-save"></i> Save</span>Your files are automatically saved.</li>
+                              <li class="list-group-item p-1"><span class="badge badge-secondary mr-1"><i class="fas fa-fw fa-save"></i> Save</span>Your files are automatically saved every <%= workspaceHostFileWatchIntervalSec %> seconds.</li>
```
mwest1066, line 68: I worry about this a bit because from the user's point of view this may depend on how the image is configured. For example, the Jupyter image defaults to a 120-second autosave interval. Maybe it's better to not specify the time interval?

pages/workspace/workspace.ejs
tdy: fixed in 25f9be8c3f5be30d776d8ca12ebf13535aa037cc

pages/workspace/workspace.ejs
tdy: Ok, reverted in 88a3417e61dcf2c70890aafbc777a137721b8b4c. I had added it after losing file changes when testing `Reboot`, so I was again feeling uneasy about the opacity of our autosaving as you're aware.

mwest1066: Oh! How did you lose changes? Did you reboot very soon after a change?

PR #2880, opened 2020-08-14
Update workspace status indicator with more detailed info

tdy:
* [x] refactor `updateState` for sql atomicity
* [x] redesign badge aesthetics

pl-workspace-status

### DIFF
--- a/lib/workspace.js
+++ b/lib/workspace.js
@@ -148,11 +148,18 @@ module.exports = {
         });
     },
 
-    async updateState(workspace_id, state) {
+    async updateMessage(workspace_id, message) {
+        debug(`Setting workspaces.message to '${message}'`);
+        await sqldb.callAsync('workspaces_message_update', [workspace_id, message]);
+        module.exports._namespace.to(workspace_id).emit('change:message', {workspace_id, message});
+    },
+
+    async updateState(workspace_id, state, message='') {
         // TODO: add locking
+        debug(`Setting workspaces.state='${state}'`);
         await sqldb.callAsync('workspaces_state_update', [workspace_id, state]);
-        debug(`Set workspaces.state to '${state}'`);
         module.exports._namespace.to(workspace_id).emit('change:state', {workspace_id, state});
+        await module.exports.updateMessage(workspace_id, message);
     },
 
     async controlContainer(workspace_id, action, options={}) {
--- a/lib/workspace.js
+++ b/lib/workspace.js
@@ -156,10 +156,9 @@ module.exports = {
 
     async updateState(workspace_id, state, message='') {
         // TODO: add locking
-        debug(`Setting workspaces.state='${state}'`);
-        await sqldb.callAsync('workspaces_state_update', [workspace_id, state]);
-        module.exports._namespace.to(workspace_id).emit('change:state', {workspace_id, state});
-        await module.exports.updateMessage(workspace_id, message);
+        debug(`Setting workspaces.state='${state}', workspaces.message='${message}'`);
+        await sqldb.callAsync('workspaces_state_update', [workspace_id, state, message]);
+        module.exports._namespace.to(workspace_id).emit('change:state', {workspace_id, state, message});
     },
 
     async controlContainer(workspace_id, action, options={}) {
@@ -210,10 +209,11 @@ module.exports = {
         if (state == 'uninitialized' || state == 'stopped') {
             if (state == 'uninitialized') {
                 await module.exports.initialize(workspace_id);
-                await module.exports.updateState(workspace_id, 'stopped');
+                await module.exports.updateState(workspace_id, 'stopped', 'Initialization complete');
             }
-            await module.exports.updateState(workspace_id, 'launching');
+            await module.exports.updateState(workspace_id, 'launching', 'Assigning workspace host');
             await module.exports.assignHost(workspace_id);
+            await module.exports.updateMessage(workspace_id, 'Sending launch command to host');
             await module.exports.controlContainer(workspace_id, 'init', {useInitialZip: true});
         }
     },
--- a/pages/workspace/workspace.ejs
+++ b/pages/workspace/workspace.ejs
@@ -3,6 +3,7 @@
 
 <head>
   <%- include('../partials/head'); %>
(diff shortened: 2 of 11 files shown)


lib/workspace.js
```diff
@@ -148,11 +148,18 @@ module.exports = {
...
+        module.exports._namespace.to(workspace_id).emit('change:message', {workspace_id, message});
+    },
+
+    async updateState(workspace_id, state, message='') {
         // TODO: add locking
+        debug(`Setting workspaces.state='${state}'`);
         await sqldb.callAsync('workspaces_state_update', [workspace_id, state]);
-        debug(`Set workspaces.state to '${state}'`);
         module.exports._namespace.to(workspace_id).emit('change:state', {workspace_id, state});
+        await module.exports.updateMessage(workspace_id, message);
```
mwest1066, line 162: I know the current code is nice and clean, but I'd rather update state+message atomically as a pair in this case. That is, have the SQL call update both of them, and send the new message as part of the `change:state` websocket message. This will help to avoid any potential race conditions.

Of course we should still have a separate `updateMessage()` just like you have it now.

workspace_host/interface.js
```diff
@@ -693,6 +693,7 @@ function _recursiveDownloadJobManager(curDirPath, S3curDirPath, callback) {
 }
 
 async function _syncInitialZipAsync(workspace) {
+    workspaceHelper.updateMessage(workspace.id, 'Preparing initial files');
```
mwest1066, line 696:
```suggestion
    workspaceHelper.updateMessage(workspace.id, 'Loading initial files');
```

workspace_host/interface.js
```diff
@@ -710,6 +711,7 @@ async function _syncInitialZipAsync(workspace) {
 const _syncInitialZip = util.callbackify(_syncInitialZipAsync);
 
 function _syncPullContainer(workspace, callback) {
```
mwest1066, line 713:
```suggestion
function _getInitialFiles(workspace, callback) {
```

I think this is getting the files, not the container?

workspace_host/interface.js
```diff
@@ -693,6 +693,7 @@ function _recursiveDownloadJobManager(curDirPath, S3curDirPath, callback) {
 }
 
 async function _syncInitialZipAsync(workspace) {
```
mwest1066, line 695:
```suggestion
async function _getInitialZipAsync(workspace) {
```

I think using "sync" to mean "get"/"fetch" is confusing.

mwest1066:
In `lib/workspace.js` we could also add messages, such as:

```
    async startup(workspace_id, state) {
        if (state == 'uninitialized' || state == 'stopped') {
            if (state == 'uninitialized') {
                await module.exports.initialize(workspace_id);
                await module.exports.updateState(workspace_id, 'stopped', 'Initialization complete');
            }
            await module.exports.updateState(workspace_id, 'launching', 'Assigning workspace host');
            await module.exports.assignHost(workspace_id);
            await module.exports.updateMessage(workspace_id, 'launching', 'Sending launch command to host');
            await module.exports.controlContainer(workspace_id, 'init', {useInitialZip: true});
        }
    },
```

pages/workspace/workspace.ejs
```diff
@@ -18,7 +18,10 @@
     <div class="d-flex flex-row ml-auto align-items-center">
       <ul class="navbar-nav flex-row">
         <li class="mr-2">
-          <span id="state" class="badge badge-dark text-uppercase"></span>
+          <div class="badge badge-dark">
+            <span id="state" class="text-uppercase"><i class="fas fa-spinner fa-pulse"></i></span>
+            <span id="message" class="font-weight-normal"></span>
```
mwest1066, line 23: It looks to me like the state and message are too close together?

tdy: You mean the visual rendering? As in you want more separation between the state and message?

mwest1066: Yep, or an emdash or colon or something? You are better at visual design than me, but it just looked a bit squashed.

lib/workspace.js
```diff
@@ -203,10 +209,11 @@ module.exports = {
         if (state == 'uninitialized' || state == 'stopped') {
             if (state == 'uninitialized') {
                 await module.exports.initialize(workspace_id);
-                await module.exports.updateState(workspace_id, 'stopped');
+                await module.exports.updateState(workspace_id, 'stopped', 'Initialization complete');
             }
-            await module.exports.updateState(workspace_id, 'launching');
+            await module.exports.updateState(workspace_id, 'launching', 'Assigning workspace host');
             await module.exports.assignHost(workspace_id);
+            await module.exports.updateMessage(workspace_id, 'Sending launch command to host');
```
tdy, line 216: Or should we set `launching` state again with a new `message`?

mwest1066: From a state-machine perspective, we aren't changing state. I think it's best to conceptually separate state changes (which mean something to the code execution) and message updates (purely for display).

lib/workspace.js
tdy: `sprocs/workspaces_state_update.sql` now takes both `state` and `message`

pages/workspace/workspace.ejs
tdy: Tried a slight redesign (updated OP screenshot)

mwest1066: I really like the new status/message badge! Looks great.

When the display width is narrow the status badge is right aligned, which causes the status (e.g., "launching") to jump around a lot as the message changes. It also makes a hamburger menu that displays the reboot/help buttons in a weird overlay way (doesn't look like a menu). I'm not sure whether this is all intentional?

tdy: Not intentional, but I haven't figured out how to fix it. I was kinda just hoping no one would be using workspaces in a mobile viewport 😆 but I can dig into it more if you want.

tdy: One workaround could be to just hide the `message` part in the `xs` breakpoint.

mwest1066: I think it's fine, I just wasn't sure whether it's intentional. Your screenshot in the PR shows the `xs` view, which is why I thought you might have intended this 😄

PR #2881, opened 2020-08-14
adding option to grade specific cells in jupyter notebook

mfsilva22:
When using JupyterLab as workspace, we can use the python autograder to grade parts of the notebook. Instead of converting the entire notebook to .py as student code, this PR adds the option to search for specific cells based on a keyword (default is #keep), and grade the content of these cells only. This will be helpful when the notebook contains lots of code with intermediate steps, plots, print statements, additional libraries, etc, that we don't want to send to the autograder.

### DIFF
--- a/exampleCourse/questions/demo/workspace/jupyterlab/info.json
+++ b/exampleCourse/questions/demo/workspace/jupyterlab/info.json
@@ -1,19 +1,27 @@
 {
-    "uuid": "d9046679-ca5d-496d-8120-98aed21281ad",
+    "uuid": "F158CCF4-48F8-42F9-BBE6-C761210EA540",
     "title": "Workspace demo: JupyterLab",
     "topic": "Workspace",
     "tags": [
         "su20",
-        "nnytko2"
+        "tyang15", "mfsilva", "nntyko2"
     ],
     "type": "v3",
     "workspaceOptions": {
-        "image": "prairielearn/workspace-jupyterlab",
-        "port": 8080,
-        "home": "/home/jovyan",
-        "rewriteUrl": false,
-        "gradedFiles": [
-            "Workbook.ipynb"
-        ]
+      "image": "prairielearn/workspace-jupyterlab",
+      "port": 8080,
+      "home": "/home/jovyan",
+      "rewriteUrl": false,
+      "gradedFiles": [
+          "Workbook.ipynb"
+      ]
+    },
+    "singleVariant": true,
+    "gradingMethod": "External",
+    "externalGradingOptions": {
+        "enabled": true,
+        "image": "prairielearn/grader-python",
+        "entrypoint": "/python_autograder/run.sh",
+        "timeout": 20
     }
 }
--- a/exampleCourse/questions/demo/workspace/jupyterlab/info.json
+++ b/exampleCourse/questions/demo/workspace/jupyterlab/info.json
@@ -4,7 +4,7 @@
     "topic": "Workspace",
     "tags": [
         "su20",
-        "tyang15", "mfsilva", "nntyko2"
+        "tyang15", "mfsilva", "nnytko2"
     ],
     "type": "v3",
     "workspaceOptions": {
--- a/exampleCourse/questions/demo/workspace/jupyterlab/question.html
+++ b/exampleCourse/questions/demo/workspace/jupyterlab/question.html
@@ -4,7 +4,7 @@
 
   <p>
     The Fibonacci numbers are 1, 1, 2, 3, 5, 8, ..., where each
-    number is the sum of the two before it.
+    number is the sum of the two before it. These are numbered $F_1, F_2, ...$, and
     \[
     F_n = F_{n-1} + F_{n-2}
     \]
--- a/exampleCourse/questions/demo/workspace/jupyterlab/question.html
+++ b/exampleCourse/questions/demo/workspace/jupyterlab/question.html
@@ -1,4 +1,26 @@
 <pl-question-panel>
-  <p>This is a workspace question with an in-browser <a href="https://jupyter.org">JupyterLab</a>.</p>
+  <p>This is a workspace question with an in-browser <a href="https://jupyter.org">JupyterLab</a>.
+  </p>
+
+  <p>
+    The Fibonacci numbers are 1, 1, 2, 3, 5, 8, ..., where each
+    number is the sum of the two before it.
+    \[
+    F_n = F_{n-1} + F_{n-2}
+    \]
+  </p>
+
+  <p>
+    In the notebook, you will be prompted to write a Python function <tt>fib</tt> that takes a number <tt>n</tt>
+    and returns the n<sup>th</sup> Fibonacci number.
+  </p>
+
+  <pl-external-grader-variables params-name="names_from_user"></pl-external-grader-variables>
+
   <pl-workspace></pl-workspace>
 </pl-question-panel>
+
+<pl-submission-panel>
+  <pl-external-grader-results></pl-external-grader-results>
+  <pl-file-preview></pl-file-preview>
+</pl-submission-panel>
--- a/exampleCourse/questions/demo/workspace/jupyterlab/tests/test.py
+++ b/exampleCourse/questions/demo/workspace/jupyterlab/tests/test.py
@@ -1,0 +1,51 @@
+from pl_helpers import name, points, not_repeated
+from pl_unit_test import PLTestCaseWithPlot, PLTestCase
+from code_feedback import Feedback
+from functools import wraps
+import numpy as np
+import numpy.random
+
+
+class Test(PLTestCase):
+
+    student_code_file = 'Workbook.ipynb'
+
+    @points(1)
+    @name('Check fib(0)')
+    def test_0(self):
+        user_val = Feedback.call_user(self.st.fib, 0)
+        if Feedback.check_scalar("fib(0)", self.ref.fib(0), user_val):
+            Feedback.set_score(1)
+        else:
+            Feedback.set_score(0)
+
+    @points(1)
+    @name('Check fib(1)')
+    def test_1(self):
+        user_val = Feedback.call_user(self.st.fib, 1)
+        if Feedback.check_scalar("fib(1)", self.ref.fib(1), user_val):
+            Feedback.set_score(1)
+        else:
+            Feedback.set_score(0)
+
+    @points(2)
+    @name('Check fib(7)')
+    def test_2(self):
+        user_val = Feedback.call_user(self.st.fib, 7)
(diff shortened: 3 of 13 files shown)


graders/python/python_autograder/pl_unit_test.py
```diff
@@ -26,6 +26,7 @@ class PLTestCase(unittest.TestCase):
     student_code_file = 'user_code.py'
     iter_num = 0
     total_iters = 1
+    ipynb_key = '#keep'
```
mwest1066, line 29:
```suggestion
    ipynb_key = '#grade'
```

I know this is bikeshedding, but would `#grade` make the purpose clearer here?

exampleCourse/questions/demo/workspace/jupyterlab/workspace/Workbook.ipynb
```diff
@@ -1,14 +1,26 @@
...
+   "metadata": {},
+   "source": [
+    "# Fibonacci numbers"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The Fibonacci numbers are 1, 1, 2, 3, 5, 8, ..., where each number is the sum of the two before it.\n",
```
mwest1066, line 14:
```suggestion
    "The Fibonacci numbers are 1, 1, 2, 3, 5, 8, ..., where each number is the sum of the two before it. These are numbered $F_1, F_2, ...$\n",
```

It's always bothered me that this question doesn't specify whether the sequence starts from `n=0` or `n=1` 😄 (the grader assumes that it starts from `n=1`, I believe).

exampleCourse/questions/demo/workspace/jupyterlab/tests/test.py
```diff
@@ -0,0 +1,51 @@
...
+
+    @points(1)
+    @name('Check fib(0)')
+    def test_0(self):
+        user_val = Feedback.call_user(self.st.fib, 0)
+        if Feedback.check_scalar("fib(0)", self.ref.fib(0), user_val):
+            Feedback.set_score(1)
+        else:
+            Feedback.set_score(0)
+
```
mwest1066, line 21:
```suggestion
```

As I mentioned in my other comment, the sequence starts at `n=1`.

graders/python/python_autograder/pl_unit_test.py
mfsilva22: Sure, we can do that :-)

My script currently change `#clear` to `#keep`, and that is an easy change!

exampleCourse/questions/demo/workspace/jupyterlab/info.json
```diff
@@ -1,19 +1,27 @@
 {
-    "uuid": "d9046679-ca5d-496d-8120-98aed21281ad",
+    "uuid": "F158CCF4-48F8-42F9-BBE6-C761210EA540",
     "title": "Workspace demo: JupyterLab",
     "topic": "Workspace",
     "tags": [
         "su20",
-        "nnytko2"
+        "tyang15", "mfsilva", "nntyko2"
```
nicknytko, line 7:
```suggestion
        "tyang15", "mfsilva", "nnytko2"
```

exampleCourse/questions/demo/workspace/jupyterlab/workspace/Workbook.ipynb
mfsilva22: Do you want me to make the change in the vs question as well?

nicknytko: Should we also change the two other copies of this question as well? :slightly_smiling_face:

graders/python/python_autograder/pl_execute.py
```diff
@@ -50,7 +65,10 @@ def execute_code(fname_ref, fname_student, include_plt=False,
     with open(fname_ref, 'r', encoding='utf-8') as f:
         str_ref = f.read()
     with open(fname_student, 'r', encoding='utf-8') as f:
-        str_student = f.read()
+        if fname_student[-6:] == '.ipynb':
```
nicknytko, line 68: Could we use os.path.splitext (https://stackoverflow.com/a/37896418)?  This is probably unlikely, but I could see this breaking if we have a file name that is less than 6 characters.

exampleCourse/questions/demo/workspace/jupyterlab/info.json
mfsilva22: sorry!!

exampleCourse/questions/demo/workspace/jupyterlab/workspace/Workbook.ipynb
mwest1066: Yes please! It would be great to fix this everywhere.

mwest1066: Where "everywhere" = "all files in this question and all questions that refer to the Fibonacci sequence" 😄 -- `demo/autograder/codeEditor`, `demo/autograder/codeUpload`, `demo/manualGrade/codeUpload`.

graders/python/python_autograder/pl_unit_test.py
mfsilva22: done!

exampleCourse/questions/demo/workspace/jupyterlab/workspace/Workbook.ipynb
mfsilva22: oh... missed that. I will change it

exampleCourse/questions/demo/workspace/jupyterlab/question.html
```diff
@@ -1,4 +1,26 @@
...
+  <p>
+    The Fibonacci numbers are 1, 1, 2, 3, 5, 8, ..., where each
+    number is the sum of the two before it. These are numbered $F_1, F_2, ...$, and
+    \[
+    F_n = F_{n-1} + F_{n-2}
+    \]
+  </p>
+
+  <p>
+    In the notebook, you will be prompted to write a Python function <tt>fib</tt> that takes a number <tt>n</tt>
```
tdy, line 14:
```suggestion
    In the notebook <tt>Workbook.ipynb</tt>, you will be prompted to write a Python function <tt>fib</tt> that takes a number <tt>n</tt>
```
Should we be more explicit?

exampleCourse/questions/demo/workspace/jupyterlab/workspace/Workbook.ipynb
```diff
@@ -17,18 +29,25 @@
...
-    "y = np.sin(x)"
+    "#grade (write your code in this cell and DO NOT DELETE THIS LINE)\n",
+    "\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The autograder will search for cells that start with a keyword (in this example \"#keep\"), and only use the contents from this cell for grading. \n",
```
tdy, line 40:
```suggestion
    "The autograder will search for cells that start with a keyword (in this example \"#grade\"), and only use the contents from this cell for grading. \n",
```

mfsilva22: great catch! thanks. I will change it.

graders/python/python_autograder/pl_execute.py
```diff
@@ -5,9 +5,25 @@
...
+from IPython.core.interactiveshell import InteractiveShell
+
+def extract_cell_content(f, ipynb_key):
+    nb = read(f, 4)
+    shell = InteractiveShell.instance()
+    content = ""
+    for cell in nb.cells:
+        if cell["cell_type"] == "code":
+            code = shell.input_transformer_manager.transform_cell(cell.source)
+            if ipynb_key in code:
```
tdy, line 24: This will pass if `#grade` is _anywhere_ in the whole cell, I think? Or is that what you're wanting?

mfsilva22: originally I had only checking the first line, but then decided not to worry about it? I was getting the source code in a different way, which gave me lines, and when I changed I guess I didn't think of that anymore. I am not attached to any implementation, so I am open to suggestions

mwest1066: It does feel like it would be safer to only look in the first line of the cell?
```
lines = code.splitlines()
first_line = lines[0] if len(lines) > 0 else ''
if ipynb_key in first_line:
```

mwest1066: Scrap that, I think it would be better to do:
```
if code.strip().startswith(ipynb_key):
```

The current notebook actually says this ("The autograder will search for cells that **_start with_** a keyword").

exampleCourse/questions/demo/workspace/jupyterlab/workspace/Workbook.ipynb
```diff
@@ -17,18 +29,25 @@
...
-    "y = np.sin(x)"
+    "#grade (write your code in this cell and DO NOT DELETE THIS LINE)\n",
+    "\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The autograder will search for cells that start with a keyword (in this example `#grade`), and only use the contents from this cell for grading. \n",
```
tdy, line 40:
```suggestion
    "The autograder will search for cells that start with a configurable keyword (in this example `#grade`), and only use the contents from those cells for grading. This keyword is configured using the test suite's `ipynb_key` variable and has been defined in this demo question as `ipynb_key = '#grade'`. \n",
```

exampleCourse/questions/demo/workspace/jupyterlab/workspace/Workbook.ipynb
```diff
@@ -17,18 +29,25 @@
...
+    "\n",
+    "This will avoid the autograder tring to execute additional lines of code that include several print statements, plots, and use of libraries that are not installed in PL"
```
tdy, line 42:
```suggestion
    "This will avoid the autograder trying to execute additional lines of code that include several print statements, plots, and use of libraries that are not installed in PL."
```

mwest1066:
I got an error on this question. Steps to reproduce:

1. Go to the `demo/workspace/jupyterlab` question.
2. Open the workspace.
3. Type a valid `fib(n)` function into the correct cell.
4. Click "Save" in the notebook.
5. Go back to the question and click "Save & Grade".
6. The following error occurs. Note that there is no error in the `fib(n)` function (or anywhere else in the notebook).

error

mwest1066:
Argh, sorry, wrong button!

PR #2886, opened 2020-08-16
Add workspace grading validation for required files

tdy:
* [x] Save `question.workspace_graded_files` into `variant.params._workspace_required_file_names` in `lib/question.js`
* [x] Validate `_files` against `_workspace_required_file_names` in `pl-workspace.parse()`
* [x] Limit `pl-workspace` to question panel
* [x] Update docs to note `pl-file-preview` and `pl-workspace`

image

### DIFF
--- a/elements/pl-workspace/pl-workspace.py
+++ b/elements/pl-workspace/pl-workspace.py
@@ -1,6 +1,12 @@
 import chevron
 
 
+def add_format_error(data, error_string):
+    if '_files' not in data['format_errors']:
+        data['format_errors']['_files'] = []
+    data['format_errors']['_files'].append(error_string)
+
+
 def render(element_html, data):
     # Get workspace url
     # TODO: Improve UX if key undefined (https://github.com/PrairieLearn/PrairieLearn/pull/2665#discussion_r449319839)
@@ -12,3 +18,24 @@ def render(element_html, data):
         html = chevron.render(f, html_params).strip()
 
     return html
+
+
+def parse(element_html, data):
+    required_file_names = data['submitted_answers'].get('_required_file_names', [])
+
+    # Get submitted files or return parse_error if it does not exist
+    files = data['submitted_answers'].get('_files', None)
+    if not files:
+        add_format_error(data, 'No submitted answer for workspace.')
+        return
+
+    # Filter out any files that were not listed in workspaceOptions.gradedFiles
+    parsed_files = [x for x in files if x.get('name', '') in required_file_names]
+
+    # Validate that all required files are present
+    if parsed_files is not None:
+        submitted_file_names = [x.get('name', '') for x in parsed_files]
+        missing_files = [x for x in required_file_names if x not in submitted_file_names]
+
+        if len(missing_files) > 0:
+            add_format_error(data, f'The following required files were missing: {", ".join(missing_files)}')
--- a/elements/pl-workspace/pl-workspace.py
+++ b/elements/pl-workspace/pl-workspace.py
@@ -24,21 +24,8 @@ def render(element_html, data):
 
 
 def parse(element_html, data):
-    required_file_names = data['submitted_answers'].get('_required_file_names', [])
-
-    # Get submitted files or return parse_error if it does not exist
-    files = data['submitted_answers'].get('_files', None)
-    if not files:
-        add_format_error(data, 'No submitted answer for workspace.')
-        return
-
-    # Filter out any files that were not listed in workspaceOptions.gradedFiles
-    parsed_files = [x for x in files if x.get('name', '') in required_file_names]
-
-    # Validate that all required files are present
-    if parsed_files is not None:
-        submitted_file_names = [x.get('name', '') for x in parsed_files]
-        missing_files = [x for x in required_file_names if x not in submitted_file_names]
-
-        if len(missing_files) > 0:
-            add_format_error(data, f'The following required files were missing: {", ".join(missing_files)}')
+    workspace_required_file_names = data['params'].get('_workspace_required_file_names', [])
+    submitted_file_names = [f.get('name', '') for f in data['submitted_answers'].get('_files', [])]
+    missing_files = [r for r in workspace_required_file_names if r not in submitted_file_names]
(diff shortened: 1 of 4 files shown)


lib/question.js
```diff
@@ -349,10 +349,19 @@ module.exports = {
...
                 if (workspace_id == null) return callback(null);
-                util.callbackify(workspace.getGradedFiles)(workspace_id, (err, result) => {
+                sqldb.queryZeroOrOneRow(sql.select_workspace_graded_files, {workspace_id}, (err, result) => {
                     if (ERR(err, callback)) return;
-                    zipPath = result;
-                    callback(null);
+                    let workspace_graded_files = [];
+                    if (result.rowCount > 0) {
+                        workspace_graded_files = result.rows[0].workspace_graded_files;
+                    }
```
mwest1066, line 357: I think it's safer to do the following, so that we will get `[]` even if we have a row but `gradedFiles` is `NULL`.

```suggestion
                    const workspace_graded_files = result.rows[0].workspace_graded_files || [];
```

lib/question.js
```diff
@@ -349,10 +349,19 @@ module.exports = {
...
+                    submission.submitted_answer['_required_file_names'] = workspace_graded_files;
```
mwest1066, line 358: I think we should:
1. Also set `submission.submitted_answer['_workspace_required_file_names']` to the same value, and use that in the `parse()` function. This will let us distinguish files required by the workspace versus files required by other elements.
2. Don't overwrite `_required_file_names` if it already exists, but instead append to it.

elements/pl-workspace/pl-workspace.py
```diff
@@ -12,3 +18,24 @@ def render(element_html, data):
         html = chevron.render(f, html_params).strip()
 
     return html
+
+
+def parse(element_html, data):
```
mwest1066, line 23: This function seems to have a lot of checks that aren't needed. I think we could just do something like:
```
workspace_required_file_names = data['submitted_answers'].get('_workspace_required_file_names', [])
submitted_file_names = [f.get('name', '') for f in data['submitted_answers'].get('_files', [])]
missing_files = [n in workspace_required_file_names if n not in submitted_file_names]
if (missing_files):
    add_format_error(data, f'The following required files were missing: {", ".join(missing_files)}')
```

(see comment below about `_workspace_required_file_names`.

elements/pl-workspace/pl-workspace.py
```diff
@@ -1,6 +1,12 @@
 import chevron
 
 
+def add_format_error(data, error_string):
+    if '_files' not in data['format_errors']:
+        data['format_errors']['_files'] = []
+    data['format_errors']['_files'].append(error_string)
+
+
 def render(element_html, data):
```
mwest1066, line 10: As @nicknytko pointed out, we should render differently depending on which panel we are in.

mwest1066, line 10: Add a check to only render in the question panel, like this: https://github.com/PrairieLearn/PrairieLearn/blob/d4306520a7c9cff858aea21b4298200baeec00dd/elements/pl-file-editor/pl-file-editor.py#L47

lib/question.js
tdy: Done. I also duped `_workspace_required_file_names` into `_required_file_names`. Not sure if you still wanted this (I guess that was pre-edit?).

elements/pl-workspace/pl-workspace.py
tdy: Done. Originally I had more or less copied the `file-*` checks and errors. I thought that we'd get more consistent errors that way (e.g., `No submitted answer for _____`).

lib/question.js
```diff
@@ -111,6 +111,13 @@ module.exports = {
                     options: data.options || {},
                     broken: hasFatalIssue,
                 };
+                if (question.workspace_image !== null) { // if workspace, add graded files to params
+                    variant.params['_workspace_required_file_names'] = question.workspace_graded_files;
```
mwest1066, line 115: When possible, we should always assume that data from DB might be `NULL`, unless we have an explicit non-NULL constraint. In this case we don't, so we should treat `question.workspace_graded_files` as potentially NULL and do something like:
```suggestion
                    variant.params['_workspace_required_file_names'] = question.workspace_graded_files || [];
```

lib/question.js
```diff
@@ -111,6 +111,13 @@ module.exports = {
...
+                    if(!('_required_file_names' in variant.params)) {
+                        variant.params['_required_file_names'] = [];
+                    }
+                    variant.params['_required_file_names'].push(...question.workspace_graded_files);
```
mwest1066, line 119: It's generally safer to use `concat()` so that we don't run into JS limits on number of function arguments (the `...` syntax expands into an argument list). We won't hit that limit here in practice (it's at least 32k in modern browsers) but I prefer to not get in the habit of assuming that arbitrary arrays can expand to argument lists.

lib/question.js
mwest1066: It doesn't really matter either way at this point (we never use it, but all file creation elements consistently make it). I'm fine to have it there.

elements/pl-workspace/pl-workspace.py
mwest1066: I guess I was worried about the use of the phrase "submitted answer" there. For workspaces I think students might not be thinking about a "submitted answer" as such. But if we say "these files were missing" then hopefully it's clear that those files needed to be present?

mwest1066:
@tdy note that this PR is still in "Draft" mode

lib/question.js
tdy: I see, so even though it's required by the JSON schema, something still might have gone wrong in the JSON sync?

lib/question.js
tdy: Done

elements/pl-workspace/pl-workspace.py
tdy: Sure, sounds good

tdy:
@mwest1066 Should be done with all the changes on my end

PR #2888, opened 2020-08-17
Add new workspace zip dirs to Dockerfile

tdy:
Added new zip dirs to the `Dockerfile`. Leaving the rest of the transition for part 2.


tdy:
Oops don't merge yet

tdy:
Ok, should be good to go now

PR #2889, opened 2020-08-17
Change workspace zipping code to the new zip dirs

tdy:
Final part of the transition to new zip dirs.


Dockerfile
```diff
@@ -13,7 +13,8 @@ COPY . /PrairieLearn/
 # set up PrairieLearn and run migrations to initialize the DB
 RUN chmod +x /PrairieLearn/docker/init.sh \
     && mkdir /course{,{2..9}} \
-    && mkdir -p /jobs/workspace{s,_send_zips,_receive_zips} \
+    && mkdir -p /workspace_{main,host}_zips \
+    && mkdir -p /jobs/workspaces \
```
mwest1066, line 17: While we are cleaning this up, it doesn’t actually make sense to create `workspaces` inside `/jobs`, because we are going to mount a directory over `/jobs` using something like `-v ~/pl_ag_jobs:/jobs` when we run docker (as @tdy pointed out to me on slack). This will hide the copy of `/jobs/workspaces` that we made with this line. When we make the first `/jobs/workspaces/workspace-NN` directory, that will actually create the `/jobs/workspaces` directory as `~/pl_ag_jobs/workspaces`. For this reason, I think it would be clearer to only make `/jobs` here, so we don’t get confused about what the `workspaces` subdirectory is doing.

```suggestion
    && mkdir -p /jobs \
```

PR #2891, opened 2020-08-17
Generate and consume course chunks

nwalters512:
Continuation of #2851 and #2852; this should be the culmination of the chunks work.


pages/instructorLoadFromDisk/instructorLoadFromDisk.js
```diff
@@ -53,7 +55,24 @@ var update = function(locals, callback) {
...
+                                util.callbackify(chunks.createChunksSymlinks)({ coursePath: courseDir, courseId: result.courseId, courseData: result.courseData }, (err) => {
+                                    if (ERR(err, callback)) return;
+                                    // NOTE: Just for testing
+                                    util.callbackify(chunks.updateChunksForCourse)({
+                                        coursePath: courseDir,
+                                        courseId: result.courseId,
+                                        courseData: result.courseData,
+                                        oldHash: 'HEAD~1',
+                                        newHash: 'HEAD',
+                                    }, (err) => {
```
nicknytko, line 68: Do we need this line?  I would think that we should always try to update chunks whenever we sync

mwest1066 (approved):

pages/instructorFileEditor/instructorFileEditor.js
```diff
@@ -932,7 +951,23 @@ function saveAndSync(fileEdit, locals, callback) {
                     } else if (result.hadJsonErrors) {
                         job.fail('One or more JSON files contained errors and were unable to be synced');
                     } else {
-                        job.succeed();
+                        if (config.chunksGenerator) {
```
nwalters512, line 954: Nit/question, for my own understanding: the way I was thinking about things, all file edits should be served by servers that should unconditionally do chunk generation. So is this flag really needed? Or is it just for safety + explicitness?

question-servers/freeform.js
```diff
@@ -134,10 +135,14 @@ module.exports = {
         if (courseElementsCache[course.id] !== undefined) {
             return callback(null, courseElementsCache[course.id]);
         }
-        module.exports.loadElements(path.join(course.path, 'elements'), 'course', (err, elements) => {
+        const coursePath = chunks.getRuntimeDirectoryForCourse(course);
+        chunks.ensureChunksForCourse(course.id, {'type': 'elements'}, (err) => {
```
nwalters512, line 139: I believe this will lead to a bug where a PL server won't have up-to-date course element files if the elements were previously cached. In other words, a server will *never* get new course chunks once it has an entry in `courseElementsCache`. Once we throw multiple servers into the mix, and especially with syncs happening on different machines than actually serve questions, the course element cache won't work in its current form. There's a couple potential solutions here, I think:

* Key the cache on course ID *and* Git hash - this will invalidate the cache when a sync occurs, because a new hash will be associated with the course.
* Eliminate the cache entirely. Very robust, less efficient. a.k.a. "caching is hard, let's not."

The former is my preferred strategy. Anyone have any other thoughts?

sprocs/chunks_insert.sql
```diff
@@ -0,0 +1,28 @@
...
+        (chunk->>'type')::enum_chunk_type,
+        in_course_id,
+        q.id,
+        ci.id,
+        a.id,
+        (chunk->>'uuid')::uuid
+    FROM unnest(generated_chunks) AS chunk(json)
+    LEFT JOIN questions AS q ON (q.uuid = (chunk->>'questionUuid')::uuid AND q.qid = (chunk->>'questionName') AND q.course_id = in_course_id)
+    LEFT JOIN course_instances AS ci ON (ci.uuid = (chunk->>'courseInstanceUuid')::uuid AND ci.short_name = (chunk->>'courseInstanceName') AND ci.course_id = in_course_id)
+    LEFT JOIN assessments AS a ON (a.uuid = (chunk->>'assessmentUuid')::uuid AND a.tid = (chunk->>'assessmentName') AND a.course_instance_id = ci.id)
```
nwalters512, line 19: Are there any concerns about query performance when querying `uuid`? The only table with an index that *might* help here is `questions`, but even that has a `WHERE deleted_at_ IS NULL`, so I'm not sure the query engine will decide to use it.

Aside: do we need/want `deleted_at` checks here? It's late and I can't think of the answer myself, but wanted to raise it for discussion.

nwalters512 (commented):
Overall looks good! A few of these things are bugs and should be addressed before merging, the rest are nits/suggestions/questions. Thanks for all the hard work @nicknytko and @mwest1066!

pages/instructorFileEditor/instructorFileEditor.js
mwest1066: It's because some people might want to run a PL instance without using chunks at all, in a simple single-server configuration. The default is `chunksGenerator=false` and `chunksConsumer=false` so that things are easy for other people to set up (and we won't break existing installations at Berkeley, Maryland, etc). In the future we'll need to add admin docs about "how to set up a multi-server install".

question-servers/freeform.js
mwest1066: @nicknytko apparently already fixed this in #2507

pages/instructorFileEditor/instructorFileEditor.js
nwalters512: Gotcha! Related to this: I was thinking we'd want the default to actually be with chunks (at least for local dev) in the interest of trying to make things behave as similarly as possible between dev and prod. This is what we built the symlinking chunks thing for. Was that reconsidered?

mwest1066: That's a great point. I think it's actually more that anything not on the critical path to deployment got ignored 😄.

Actually the `chunksGenerator`/`chunksConsumer` options were renamed when we were thinking about a different deployment strategy, which we discarded in the end. I think the justification above about single-server defaults is something of an after-the-fact justification that I just made up 😄.

Our current deployment strategy is to only opt-in courses one by one via the load balancer. That is, we only get the load balancer to redirect `/pl/course_instance/23423/*` to the chunk servers, for a couple of specific `course_instance_ids`. Initially we'll only redirect `pl-dev101` and `pl-test101`, and then we'll redirect the Calc 3 class that's running an exam on Tuesday. After we get through that we'll re-consider local dev.

I do agree that we want local dev to mimic the prod setup as much as possible.

We also don't have the test suite set up yet to use chunks.

sprocs/chunks_insert.sql
mwest1066: This is excellent! I was looking at the DB table scan statistics the other day and I noticed that by far the worst table is `questions`, but I didn't have time to explore why. I think it's because we have an index on `(course_id, uuid) WHERE deleted_at IS NULL` but we should _not_ have the `deleted_at IS NULL` condition there. In fact UUIDs are globally unique (within a course), even among deleted rows. This is what we had in the `paper_sync.pdf` document describing the sync algorithm.

We do have the correct indexes on `course_instances` and `assessments`:
```
course_instances_course_id_uuid_key: UNIQUE (course_id, uuid) USING btree (course_id, uuid)
assessments_course_instance_id_uuid_key: UNIQUE (course_instance_id, uuid) USING btree (course_instance_id, uuid)
```

Considering `course_instances`, the LEFT JOIN condition is `ci.uuid = THE_UUID AND ci.course_id = in_course_id` so it will use the index above.

Considering `questions`, the LEFT JOIN condition is basically the same, so we would be ok if we had an index without `deleted_at IS NULL`.

I'm less certain about `assessments`. This would be fine if we were using regular JOIN rather than LEFT JOIN for `course_instances`, so we would definitely have a `ci.id`. Fortunately there are only 9000 assessments in the DB so even if it falls back to a sequential scan it will still be fast enough at the moment. I'll make an issue to check the query planner for this after we've deployed.

question-servers/freeform.js
mwest1066: I'm filing this as a separate issue to be resolved after deployment. The initial courses we'll use chunks for don't have course elements.

pages/instructorFileEditor/instructorFileEditor.js
mwest1066: I've filed issues for the above items.

pages/elementFiles/elementFiles.js
```diff
@@ -26,12 +29,16 @@ router.get('/*', function(req, res, next) {
     let elementFilesDir;
     if (res.locals.course) {
         // Files should be served from the course directory
-        elementFilesDir = path.join(res.locals.course.path, 'elements');
+        const coursePath = chunks.getRuntimeDirectoryForCourse(res.locals.course);
+        chunks.ensureChunksForCourse(res.locals.course.id, {'type': 'elements'}, (err) => {
```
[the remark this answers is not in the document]
mwest1066: Newer pages are using https://www.npmjs.com/package/express-async-handler to do this. For example: https://github.com/PrairieLearn/PrairieLearn/blob/57a2255d372f457bc6655b37b60e40bf1458c240/pages/workspace/workspace.js#L30

I like your `safeAsync` function for being easy to drop in for a short segment of code, rather than writing the whole route handler as async/await with `express-async-handler`.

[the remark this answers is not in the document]
mwest1066: Filed an issue for this.

sprocs/chunks_insert.sql
mwest1066: I pushed a new commit to fix the index on `questions` and I added an issue to investigate the query planning after deploy.

pages/elementFiles/elementFiles.js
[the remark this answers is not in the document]
nwalters512: Ah! TIL about `express-async-handler`! The implementation seems to be practically identical to my own, except it correctly handles the 4-argument form of an express handler. I think we should prefer that over my suggestion.

lib/chunks.js
```diff
@@ -71,19 +155,600 @@ module.exports.identifyChunksFromChangedFiles = (changedFiles, course) => {
...
+};
+
+/**
+ *
+ * @param {Object} options
+ * @param {string} options.coursePath
+ * @param {string} options.courseId
+ * @param {import('../sync/course-db').CourseData} options.courseData
+ * @param {?string} options.oldHash
+ * @param {?string} options.newHash
```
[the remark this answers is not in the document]
mwest1066: Done

lib/chunks.js
```diff
@@ -71,19 +155,600 @@ module.exports.identifyChunksFromChangedFiles = (changedFiles, course) => {
...
+                await module.exports._generateAllChunksForCourseWithJob(course_id, authn_user_id, job);
+            } catch (err) {
+                job.error(chalk.red(JSON.stringify(err)));
+                await job.failAsync(String(err));
+                throw err;
+            }
+        }
+    } catch (err) {
+        try {
+            await serverJobs.failJobSequence(job_sequence_id);
```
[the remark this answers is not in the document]
mwest1066: Done

lib/chunks.js
```diff
@@ -71,19 +155,600 @@ module.exports.identifyChunksFromChangedFiles = (changedFiles, course) => {
...
+                last_in_sequence: (i == course_ids.length - 1),
+            };
+            const job = await serverJobs.createJobAsync(jobOptions);
+            job.info(chalkDim(`Course ID = ${course_id}`));
+
+            try {
+                await module.exports._generateAllChunksForCourseWithJob(course_id, authn_user_id, job);
+            } catch (err) {
+                job.error(chalk.red(JSON.stringify(err)));
+                await job.failAsync(String(err));
```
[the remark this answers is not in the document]
mwest1066: Removed

lib/chunks.js
```diff
@@ -71,19 +155,600 @@ module.exports.identifyChunksFromChangedFiles = (changedFiles, course) => {
...
+ * @property {string} question_name
+ */
+
+/**
+ *
+ * @param {string} courseId
+ * @param {DatabaseChunk} chunk
+ */
+const ensureChunk = async (courseId, chunk) => {
+    const courseDir = module.exports.getChunksDirectoryForCourseId(courseId);
```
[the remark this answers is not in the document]
mwest1066: Done

lib/chunks.js
```diff
@@ -71,19 +155,600 @@ module.exports.identifyChunksFromChangedFiles = (changedFiles, course) => {
...
+
+    // Finally, rename to remove the `-staging` from the name.
+    await fs.move(chunkStagingPath, chunkPath);
+};
+
+/** @type {Map<string, Promise>} */
+const pendingChunksMap = new Map();
+
+/**
+ * Ensures that specific chunks for a course are loaded.  These chunks will either be pulled
```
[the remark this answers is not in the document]
mwest1066: Done

lib/github.js
```diff
@@ -326,17 +327,35 @@ module.exports = {
...
+                    'type': 'load_chunks',
+                    'description': 'Create course chunks',
+                    'last_in_sequence': true,
+                }, job_sequence_id, authn_user.user_id, async () => {
+                    await chunks.updateChunksForCourse({
+                        'coursePath': inserted_course.path,
+                        'courseId': inserted_course.id,
+                        'courseData': sync_result.courseData,
+                        'oldHash': 'HEAD~1',
+                        'newHash': 'HEAD',
```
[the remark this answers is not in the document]
mwest1066: Removed these variables, which will not find changed files and will trigger a complete generate.

PR #2893, opened 2020-08-17
Add workspace versioning to support resets

tdy:
- [x] add a `workspaces.version` column that defaults to 1
- [x] the S3 directory name is now of the format `workspace-ID-VERSION` (we'll need to change all the places in the code that compute the S3 workspace directory name)
- [x] to reboot a workspace we set it to `stopped`, leave the `version` alone, and then redirect to the workspace URL
- [x] to reset a workspace we set it to `uninitialized`, increment `version`, and then redirect to the workspace URL
    * the launch code will not require any changes at all, because it will see that the workspace is `uninitialized`, it will make the new `workspace-ID-VERSION` (with the new version), and everything proceeds as normal
- [x] `workspace_logs` should also have a version column that we set to the current version whenever we write a log entry


lib/workspace.js
```diff
@@ -217,20 +217,25 @@ module.exports = {
         // FIXME: Check we are in "uninitialized"
         let result = await sqldb.queryOneRowAsync(sql.select_workspace_paths, {workspace_id});
         const {course_path, qid} = result.rows[0];
+        result = await sqldb.queryOneRowAsync(sql.select_workspace_version, {workspace_id});
+        const { workspace_version } = result.rows[0];
         const localPath = `${course_path}/questions/${qid}/workspace`;
-        const s3Path = `${config.workspaceS3Bucket}/workspace-${workspace_id}`;
+
+        const s3Name = `workspace-${workspace_id}-${workspace_version}/current`;
+        const s3Path = `${config.workspaceS3Bucket}/${s3Name}`;
```
mwest1066, line 225: Very minor nit, but I'd rather do:
```suggestion
        const s3Name = `workspace-${workspace_id}-${workspace_version}`;
        const s3Path = `${config.workspaceS3Bucket}/${s3Name}/current`;
```

This way we can avoid the `replace()` call below to make `s3ZipPath`. I agree that in this case the use of `replace()` is safe, but I'm scared to use `replace()` on path-like strings because we run the risk that the replacement word appears more than once in the string.

lib/workspace.js
```diff
@@ -217,20 +217,25 @@ module.exports = {
...
+
         const now = new Date().toISOString().replace(/[-T:.]/g, '-');
         const zipName = `workspace-${workspace_id}-${now}.zip`;
         const zipPath = path.join(config.workspaceMainZipsDirectory, zipName);
 
         debug(`Zipping ${localPath} as ${zipPath}`);
         await zipDirectory(localPath, zipPath);
         const isDirectory = false;
-        const s3ZipPath = `workspace-${workspace_id}/initial.zip`;
+        const s3ZipPath = s3Name.replace('current', 'initial.zip');
```
mwest1066, line 234:
```suggestion
        const s3ZipPath = `${s3Name}/initial.zip`;
```

lib/workspace.js
```diff
@@ -217,20 +217,25 @@ module.exports = {
...
         await awsHelper.uploadToS3Async(zipPath, isDirectory, s3ZipPath, zipPath);
 
         debug(`Syncing ${localPath} to ${s3Path}`);
-        await module.exports.uploadDirectoryToS3Async(localPath, `${s3Path}/current`);
+        await module.exports.uploadDirectoryToS3Async(localPath, `${s3Path}`);
```
mwest1066, line 238: Nit: remove unnecessary formatter:
```suggestion
        await module.exports.uploadDirectoryToS3Async(localPath, s3Path);
```

migrations/201_workspaces__version__add.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE workspaces ADD COLUMN IF NOT EXISTS version bigint NOT NULL DEFAULT 1;
+ALTER TABLE workspace_logs ADD COLUMN IF NOT EXISTS workspace_version bigint NOT NULL;
```
mwest1066, line 2: This will fail if we run it on production, because it won't know what to set `workspace_version` to for existing rows in `workspace_logs`. It either needs to allow NULL or it needs a DEFAULT.

pages/workspace/workspace.js
```diff
@@ -22,6 +26,11 @@ router.get('/:workspace_id/:action', asyncHandler(async (req, res, next) => {
         logger.info(`Rebooting workspace ${workspace_id}.`);
         await workspaceHelper.updateState(workspace_id, 'stopped', 'Rebooting container');
         res.redirect(`/pl/workspace/${workspace_id}`);
+    } else if (action === 'reset') {
+        logger.info(`Resetting workspace ${workspace_id}.`);
```
mwest1066, line 30: I think we should change all `logger.info()` calls in this file to `debug()`, to avoid spamming the logs in production.

workspace_host/interface.js
```diff
@@ -49,9 +49,9 @@ app.post('/', function(req, res) {
     const workspace_id = req.body.workspace_id;
     const action = req.body.action;
     const useInitialZip = _.get(req.body.options, 'useInitialZip', false);
-    if (workspace_id == undefined) {
+    if (typeof workspace_id === 'undefined') {
         res.status(500).send('Missing workspace_id');
-    } else if (action == undefined) {
+    } else if (typeof action === 'undefined') {
```
mwest1066, line 54: A good JS pattern I often use in situations like this is `if (workspace_id == null)`. Using `== null` (two equal signs) will give `true` if `workspace_id` is either `undefined` or `null`, both of which it would be good to detect.

pages/workspace/workspace.ejs
```diff
@@ -16,24 +16,51 @@
...
+        <div class="modal-header">
+          <h1 class="h5 modal-title" id="resetModalLabel">Reset the workspace</h1>
+          <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+            <span aria-hidden="true">&times;</span>
+          </button>
+        </div><!-- .modal-header -->
+        <div class="modal-body">
+          <p>Are you sure you want to reset the virtual machine?<p>
+          <ul class="fa-ul">
+            <li><span class="fa-li"><i class="fas fa-exclamation-triangle" aria-hidden="true"></i></span>Your file changes <em>will be lost</em>. The workspace will return to its original state upon reset.</li>
```
mwest1066, line 50: Can we make this more noticeable? Maybe make the fa-exclamation-triangle red? Or the whole sentence above red? Or something else?

migrations/201_workspaces__version__add.sql
tdy: Changed to default=1

pages/workspace/workspace.ejs
tdy: Tweaked to this:
image

mwest1066: I love it! This seems impossible to overlook (although I'm sure some student will manage it 😄).

lib/workspace.js
tdy: Went ahead and changed this in `interface` as well (this `replace` scheme started there as a lazy way to propagate the `*/current` change from 1 spot).

PR #2898, opened 2020-08-19
Add more detailed JSON validation errors

mwest1066:
Rather than displaying only `data should NOT have additional properties`, which doesn't tell you which property is the problem, we will now display additional details:
```
Error: data should NOT have additional properties
Error details:
[
  {
    "keyword": "additionalProperties",
    "dataPath": "",
    "schemaPath": "#/additionalProperties",
    "params": {
      "additionalProperty": "badKey"
    },
    "message": "should NOT have additional properties"
  }
]
```

This extra detail is shown for all JSON validation errors, not just "additional properties".

### DIFF
--- a/lib/json-load.js
+++ b/lib/json-load.js
@@ -52,7 +52,7 @@ module.exports.validateJSON = function(json, schema, callback) {
         return;
     }
     if (!valid) {
-        callback(new Error(ajv.errorsText(validate.errors)));
+        callback(new Error(`${ajv.errorsText(validate.errors)}\nError details:\n${JSON.stringify(validate.errors, null, 2)}`));
     } else {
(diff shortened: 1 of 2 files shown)


mwest1066:
I found [better-ajv-errors](https://github.com/atlassian/better-ajv-errors) which works well with the [ansi_up](https://github.com/drudru/ansi_up) library we were already using. Now we get:

new_error2

nicknytko:
> I found [better-ajv-errors](https://github.com/atlassian/better-ajv-errors) which works well with the [ansi_up](https://github.com/drudru/ansi_up) library we were already using. Now we get:
> 
> new_error2

that looks awesome !!

PR #2901, opened 2020-08-19
Documentation: dates based on instance timezone

jonatanschroeder:
Updated documentation for access control to refer to the course instance timezone instead of the default.


mwest1066:
Thanks!

PR #2902, opened 2020-08-19
Add workspace autoscaling

nicknytko:
todo:

- [ ] Allow hosts to be launched into any subnet/az &mdash; they are all currently launched into one.  Not sure if this is something that can be done with launch templates or we'll have to custom write something. (Can this be left for the future?)
- [x] Fix permissions error on new hosts


cron/workspacesLoad.js
```diff
@@ -0,0 +1,146 @@
...
+    let desired_hosts = await config.getDBConfigValueAsync('workspaceDesiredHostCount');
+    if (!desired_hosts) {
+        desired_hosts = stats.workspace_hosts_desired;
+    }
+    let launch_template_id = await config.getDBConfigValueAsync('workspaceLaunchTemplateId');
+    if (!launch_template_id) {
+        launch_template_id = config.workspaceLoadLaunchTemplateId;
+    }
+
+    const running_hosts = stats.workspace_hosts_ready_count;
```
mwest1066, line 122: I think we want the sum of `ready_count` and `launching_count`? Otherwise we will keep thinking we need to launch more, even though some are launching?

database/tables/workspace_hosts.pg
```diff
@@ -1,8 +1,11 @@
 columns
+    became_unhealthy_at: timestamp with time zone
     hostname: text
     id: bigint not null default nextval('workspace_hosts_id_seq'::regclass)
     instance_id: text
+    launched_at: timestamp with time zone
     load_count: integer default 0
+    state: enum_workspace_host_state
```
mwest1066, line 8: Maybe also columns for:
* `state_changed_at` -- last time we changed `state`
* `ready_at` -- first time we entered state "ready", so `ready_at - launched_at` tells us how long it took to launch
* `terminated_at`

database/tables/workspace_hosts.pg
```diff
@@ -1,8 +1,11 @@
 columns
+    became_unhealthy_at: timestamp with time zone
```
mwest1066, line 2: Maybe use `unhealthy_at` to keep column names short?

lib/config.js
```diff
@@ -227,3 +242,32 @@ config.setLocals = (locals) => {
...
+/* Some helper functions to access the 'config' table in the database */
+config.getDBConfigValueAsync = async function(key, defaultVal = null) {
+    const result = await sqldb.call('config_select', {key});
+    if (result.rows.length == 0) {
+        return defaultVal;
+    } else {
+        return result.rows[0].value;
+    }
+};
+config.getDBConfigValue = function(key, callback) {
```
mwest1066, line 255: This function doesn't seem to be quite the sync version of the async function above?

sprocs/workspace_hosts_drain_extra.sql
```diff
@@ -0,0 +1,23 @@
+CREATE OR REPLACE FUNCTION
+    workspace_hosts_drain_extra(surplus integer) RETURNS void
+AS $$
+BEGIN
+    -- Grab a random assortment of extra hosts
+    CREATE TEMPORARY TABLE extra ON COMMIT DROP AS (
+        SELECT *
+        FROM workspace_hosts AS wh
+        WHERE wh.state = 'ready'
+        ORDER BY random()
```
mwest1066, line 10: I suggest removing the oldest ones first, so that we tend to cycle in newer hosts. This can provide us with a way to upgrade workspace hosts by launching more with a new AMI version, then scaling down again to kill off the old ones.

cron/workspacesLoad.js
mwest1066: Actually, I think this is a bit more subtle. I think that we want:
* if `desired_hosts > ready_hosts + launching_hosts` then start some more (or un-drain some)
* if `desired_hosts < ready_hosts` then set some running hosts to drain

I think we can't make this symmetric because:
* we don't want to use `desired_hosts > ready_hosts` for the scale-up check, because we will keep launching new hosts all the time while waiting for the launching ones to become ready.
* we don't want to use `desired_hosts < ready_hosts + launching_hosts`, because the launching ones don't really count yet for actual useable capacity

cron/workspacesLoad.js
```diff
@@ -0,0 +1,169 @@
...
+            if (!(key in cloudwatch_definitions)) {
+                throw new Error(`Unknown datapoint ${key}!`);
+            }
+            const def = cloudwatch_definitions[key];
+            return {
+                MetricName: def.name,
+                Dimensions: dimensions,
+                Timestamp: stats.timestamp_formatted,
+                Unit: def.unit,
+                Value: value,
```
mwest1066, line 123: If you add `StorageResolution: 1` then I think you get hires data: https://github.com/PrairieLearn/PrairieLearn/blob/2843028d43de1a4f4197bd448e2632ca9cf06f8f/cron/externalGraderLoad.js#L49

sprocs/workspace_loads_current.sql
```diff
@@ -0,0 +1,120 @@
+CREATE OR REPLACE FUNCTION
+    workspace_loads_current(
+        IN workspace_capacity_factor double precision,
+        IN workspace_host_capacity integer,
```
mwest1066, line 4: I suggest making this a double -- not really any reason not to and over time I've found it's sometimes painful to lock ourselves in to integers.

workspace_host/interface.sql
```diff
@@ -29,20 +29,24 @@ WHERE
...
-        (instance_id,  hostname)
-VALUES ($instance_id, $hostname)
+        (instance_id,  hostname, state, state_changed_at, ready_at)
+VALUES ($instance_id, $hostname, 'ready', NOW(), NOW())
 ON CONFLICT (instance_id) DO UPDATE
-SET hostname = EXCLUDED.hostname;
+SET hostname = EXCLUDED.hostname,
+    state = EXCLUDED.state,
+    state_changed_at = EXCLUDED.state_changed_at,
+    ready_at = EXCLUDED.ready_at;
```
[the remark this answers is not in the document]
nicknytko: Could be beneficial to leave it at the upsert in case we manually are launching more hosts without having them in the db.  Dunno, don't feel very strongly either way here.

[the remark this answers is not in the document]
mwest1066: Good point, I like the ability to keep manual interventions working.

sprocs/workspace_loads_current.sql
```diff
@@ -0,0 +1,120 @@
...
+        workspace_hosts_longest_launching_sec,
+        workspace_hosts_longest_ready_sec,
+        workspace_hosts_longest_draining_sec,
+        workspace_hosts_longest_unhealthy_sec,
+        workspace_hosts_longest_terminating_sec
+    FROM
+        workspace_hosts AS wh;
+
+    -- Compute desired number of workspace hosts
+    workspace_jobs_capacity_desired := (workspace_running_count * workspace_capacity_factor) + 2 * SQRT(workspace_running_count * workspace_capacity_factor);
```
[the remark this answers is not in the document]
mwest1066: I think we could go with just a simple safety factor for now

cron/workspacesTransition.js
```diff
@@ -0,0 +1,111 @@
...
+        MaxResults: 500,
+    }).promise()).Reservations;
+    for (const reservation of reservations) {
+        for (const instance of Object.values(reservation.Instances)) {
+            running_host_set.add(instance.InstanceId);
+        }
+    }
+
+    const db_hosts = new Set((await sqldb.queryAsync(sql.select_nonterminated_or_nonterminating_workspace_hosts, [])).rows.map(instance => instance.instance_id));
+    const db_hosts_terminating = new Set((await sqldb.queryAsync(sql.select_running_or_terminating_workspace_hosts, [])).rows.map(instance => instance.instance_id));
```
mwest1066, line 49: For consistency with the query name:

```suggestion
    const db_hosts_running_or_terminating = new Set((await sqldb.queryAsync(sql.select_running_or_terminating_workspace_hosts, [])).rows.map(instance => instance.instance_id));
```

PR #2903, opened 2020-08-20
Update workspace logging functions

tdy:
* [x] parameterize log level instead of hardcoding all entries as `'info'`
* [x] convert log message into dict with `state` and `message` keys (or should this be a proper `jsonb` column?)

### DIFF
--- a/database/tables/workspace_logs.pg
+++ b/database/tables/workspace_logs.pg
@@ -3,8 +3,8 @@ columns
     id: bigint not null default nextval('workspace_logs_id_seq'::regclass)
     level: enum_log_level
     message: text
+    version: bigint not null default 1
     workspace_id: bigint not null
-    workspace_version: bigint not null default 1
 
 indexes
     workspace_logs_pkey: PRIMARY KEY (id) USING btree (id)
--- a/sprocs/workspaces_message_update.sql
+++ b/sprocs/workspaces_message_update.sql
@@ -1,6 +1,9 @@
+DROP FUNCTION IF EXISTS workspaces_message_update(bigint, text);
+
 CREATE OR REPLACE FUNCTION
     workspaces_message_update(
         workspace_id bigint,
+        workspace_level text,
         workspace_message text
     ) returns void
 AS $$
@@ -18,8 +21,8 @@ BEGIN
         w.version INTO workspace_version;
 
     INSERT INTO workspace_logs
-        (date, level, message, workspace_id, workspace_version)
+        (date, level, message, workspace_id, version)
     VALUES
-        (now(), 'info'::enum_log_level, 'message:' || workspace_message, workspace_id, workspace_version);
+        (now(), workspace_level::enum_log_level, workspace_message, workspace_id, workspace_version);
 END;
 $$ LANGUAGE plpgsql VOLATILE;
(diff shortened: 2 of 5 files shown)


sprocs/workspaces_state_update.sql
```diff
@@ -1,6 +1,9 @@
+DROP FUNCTION IF EXISTS workspaces_state_update(bigint, text, text);
+
 CREATE OR REPLACE FUNCTION
     workspaces_state_update(
         workspace_id bigint,
+        workspace_level text,
```
mwest1066, line 6: Why is this not of type `enum_log_level`?

sprocs/workspaces_state_update.sql
```diff
@@ -21,16 +24,8 @@ BEGIN
...
-        (now(),
-         'info'::enum_log_level,
-         CASE
-             WHEN (workspace_message = '') IS NOT FALSE -- either empty or null
-             THEN 'state:' || workspace_state
-             ELSE 'state:' || workspace_state || ', message:' || workspace_message
-         END,
-         workspace_id,
-         workspace_version);
+        (now(), workspace_level::enum_log_level, workspace_message, workspace_id, workspace_version);
```
mwest1066, line 29: I think we shouldn't need `::enum_log_level` here if the type is correct in the function?

sprocs/workspaces_state_update.sql
```diff
@@ -21,16 +24,8 @@ BEGIN
         w.version INTO workspace_version;
 
     INSERT INTO workspace_logs
-        (date, level, message, workspace_id, workspace_version)
+        (date, level, message, workspace_id, version)
```
mwest1066, line 27: We don't seem to be logging the state change?

sprocs/workspaces_message_update.sql
```diff
@@ -1,6 +1,9 @@
+DROP FUNCTION IF EXISTS workspaces_message_update(bigint, text);
+
 CREATE OR REPLACE FUNCTION
     workspaces_message_update(
         workspace_id bigint,
+        workspace_level text,
```
mwest1066, line 6: Type enum_log_level?

database/tables/workspace_logs.pg
```diff
@@ -3,8 +3,8 @@ columns
     id: bigint not null default nextval('workspace_logs_id_seq'::regclass)
     level: enum_log_level
     message: text
```
mwest1066, line 5: Where is the state logged?

database/tables/workspace_logs.pg
```diff
@@ -3,8 +3,8 @@ columns
     id: bigint not null default nextval('workspace_logs_id_seq'::regclass)
     level: enum_log_level
```
mwest1066, line 4: Are we ever using a log level other than `info`? If not, I'm not sure we need a column for `level` at all?

sprocs/workspaces_state_update.sql
tdy: State changes are now stored like:

```
{"state": "launching", "message": "Pulling container"}
```

Previously I had stitched it together as a string like:

```
"state=launching, message=container"
```

database/tables/workspace_logs.pg
tdy: State changes were always logged in `workspace_logs.message`. We created that column before we actually had a "message" which is now a bit confusing.

database/tables/workspace_logs.pg
tdy: Currently we're not, but I thought we were planning to log errors eventually?

PR #2912, opened 2020-08-22
Fix assessment syncing with deleted assessments having the same TID

mwest1066:
We forgot some `WHERE deleted_at IS NULL` checks, so we were picking up deleted rows and updating more than assessments row in this UPDATE: https://github.com/PrairieLearn/PrairieLearn/blob/eac3eb5ba50654be86e274119be499b7c3d01948/sprocs/sync_assessments.sql#L134

This would then fail with `query returned more than one row` because it was RETURNING into a scalar value here: https://github.com/PrairieLearn/PrairieLearn/blob/eac3eb5ba50654be86e274119be499b7c3d01948/sprocs/sync_assessments.sql#L164

The actual fix is just the one for this query (plus a test case that used to fail and now doesn't). I also added `WHERE deleted_at IS NULL` to a few other places where it seemed necessary (anywhere we are looking up an assessment by its `tid`), but I didn't write tests beforehand to see whether those were actually causing bugs.

I haven't checked the sync sprocs for course_instances, courses, and questions to see whether they have similar issues.

### DIFF
--- a/sprocs/sync_assessments.sql
+++ b/sprocs/sync_assessments.sql
@@ -159,6 +159,7 @@ BEGIN
             ) AS aggregates
         WHERE
             a.tid = valid_assessment.tid
+            AND a.deleted_at IS NULL
             AND a.tid = aggregates.tid
             AND a.course_instance_id = syncing_course_instance_id
         RETURNING id INTO new_assessment_id;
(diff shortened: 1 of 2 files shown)


mwest1066:
I restructured this PR in the style of #2917 and #2918, with commits that add a failing test follow by commits to fix the problem. I didn't add a test of the `order_by` change, because doing the ordering over all assessments (deleted and non-deleted) gives exactly the same order as just doing it over non-deleted assessments, so it wasn't really a bug at all. I only made the change for consistency with the rest of the code.

PR #2913, opened 2020-08-22
Fix group work flag bug

tbretl:
The last input argument of the function `authz_assessment_instance` is the boolean `group_work`, which is meant to indicate whether or not the corresponding assessment is for group work. This argument was hard-coded as `TRUE` in the two places where `authz_assessment_instance` was being called, when it should have been `a.group_work`.

This mistake does not seem to have any negative consequences right now, but that is just a lucky chance.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -261,6 +261,8 @@
 
   * Fix default institution in course instance access rules (Tim Bretl).
 
+  * Fix `group_work` flag when calling `authz_assessment_instance` (Tim Bretl).
+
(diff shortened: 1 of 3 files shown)


ChangeLog.md
```diff
@@ -261,6 +261,8 @@
 
   * Fix default institution in course instance access rules (Tim Bretl).
 
+  * Fix `group_work` flag when calling `authz_assessment_instance` (Tim Bretl).
```
mwest1066, line 264: Minor note that we aren't adding lines to the ChangeLog anymore

PR #2914, opened 2020-08-23
Group work for exam mode

frankyhao:
groupWork support for exam mode
- [x] add a copy of exam 3 to be group work exam example
- [x] start assessment (create a new group instance)
- [x] display group options and info in both pages (studentAssessmentExam, studentAssessmentInstanceExam)

Fixed Issues:
- [x] instructorAssessmentGroups - displaying not assigned students from other courses' instances(pointed out by Mariana)
- [x] edit assessment_instances_insert - to solve race conditions when 2 students want to create new instance simultaneously

### DIFF
--- a/pages/studentAssessmentExam/studentAssessmentExam.ejs
+++ b/pages/studentAssessmentExam/studentAssessmentExam.ejs
@@ -21,39 +21,43 @@
           <p class="lead text-center">
             This is <strong><%= assessment_set.name %> <%= assessment.number %>: <%= assessment.title %></strong> for <strong><%= course.short_name %></strong>
           </p>
+          <% if (!start) { %>
+            <p class="text-center">
+              This is a group homework. Please wait for the instructor to assign groups.
+            </p>
+          <%} else {%>
+            <% if (authz_result.time_limit_min != undefined) { %>
+            <p class="lead text-center">
+              The time limit for this assessment is
+              <strong>
+                <%= authz_result.time_limit_min %>
+                <%= (authz_result.time_limit_min == 1) ? 'minute' : 'minutes' %>
+              </strong>
+            </p>
+            <% } %>
+            
+            <% if (locals.assessment.require_honor_code) { %>
+            <div class="card card-secondary mb-4 test-class-honor-code">
+              <ul class="list-group list-group-flush">
+                <li class="list-group-item py-2">I certify that I am <%= user.name %> and I am allowed to take this assessment.</li>
+                <li class="list-group-item py-2">I pledge on my honor that I will not give or receive any unauthorized assistance on this assessment and that all work will be my own.</li>
+              </ul>
 
-          <% if (authz_result.time_limit_min != undefined) { %>
-          <p class="lead text-center">
-            The time limit for this assessment is
-            <strong>
-              <%= authz_result.time_limit_min %>
-              <%= (authz_result.time_limit_min == 1) ? 'minute' : 'minutes' %>
-            </strong>
-          </p>
-          <% } %>
-
-          <% if (locals.assessment.require_honor_code) { %>
-          <div class="card card-secondary mb-4 test-class-honor-code">
-            <ul class="list-group list-group-flush">
-              <li class="list-group-item py-2">I certify that I am <%= user.name %> and I am allowed to take this assessment.</li>
-              <li class="list-group-item py-2">I pledge on my honor that I will not give or receive any unauthorized assistance on this assessment and that all work will be my own.</li>
-            </ul>
-
-            <div class="card-footer text-center border-top-0 py-2">
-              <span class="form-check d-inline">
-                <input type="checkbox" class="form-check-input" id="certify-pledge">
-                <label class="form-check-label font-weight-bold" for="certify-pledge">I certify and pledge the above.</label>
-              </span>
+              <div class="card-footer text-center border-top-0 py-2">
+                <span class="form-check d-inline">
+                  <input type="checkbox" class="form-check-input" id="certify-pledge">
+                  <label class="form-check-label font-weight-bold" for="certify-pledge">I certify and pledge the above.</label>
+                </span>
+              </div>
             </div>
-          </div>
+            <% } %>
+
+            <form id="confirm-form" name="confirm-form" method="POST" class="d-flex justify-content-center">
+              <input type="hidden" name="__action" value="newInstance">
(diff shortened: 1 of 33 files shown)


pages/studentAssessmentExam/studentAssessmentExam.sql
```diff
@@ -4,9 +4,119 @@ SELECT
 FROM
     assessment_instances AS ai,
```
tbretl, line 5: This query is incorrect. See https://github.com/PrairieLearn/PrairieLearn/pull/2919.

I suggest you avoid all comma-joins in future.

pages/studentAssessmentExam/studentAssessmentExam.js
```diff
@@ -33,9 +56,127 @@ router.post('/', function(req, res, next) {
     if (res.locals.assessment.type !== 'Exam') return next();
     if (!res.locals.authz_result.authorized_edit) return next(error.make(403, 'Not authorized', res.locals));
     if (req.body.__action == 'newInstance') {
-        assessment.makeAssessmentInstance(res.locals.assessment.id, res.locals.user.user_id, res.locals.assessment.group_work, res.locals.authn_user.user_id, res.locals.authz_data.mode, res.locals.authz_result.time_limit_min, res.locals.req_date, (err, assessment_instance_id) => {
+        //make sure there is no new instance created by student's teammate simultaneously
```
tbretl, line 59: There is a possible race condition here:

* Student A calls `select_single_assessment_instance` and finds nothing.
* Student B calls `select_single_assessment_instance` and finds nothing.
* Student A calls `assessment.makeAssessmentInstance`.
* Student B calls `assessment.makeAssessmentInstance` => ERROR.

This will definitely happen in practice.

One way to avoid this condition might be to both check for and make an assessment instance in the same SQL query.

pages/studentAssessmentExam/studentAssessmentExam.sql
frankyhao: Thanks for pointing out. We will use JOIN ON instead of comma-join in future

pages/studentAssessmentExam/studentAssessmentExam.js
frankyhao: Thank you! I just moved this race condition check into assessment.makeAssessmentInstance

pages/studentAssessmentExam/studentAssessmentExam.ejs
```diff
@@ -1,75 +1,180 @@
...
-            Please wait until instructed to start by a proctor
-          </p>
-          <% } %>
 
-          <p class="lead text-center">
-            This is <strong><%= assessment_set.name %> <%= assessment.number %>: <%= assessment.title %></strong> for <strong><%= course.short_name %></strong>
-          </p>
+<head>
+  <%- include('../partials/head'); %>
+</head>
```
mwest1066, line 6: Why is the indentation changed here? It makes it very hard to review where there are spurious changes showing up in the diff.

pages/studentAssessmentExam/studentAssessmentExam.ejs
```diff
@@ -1,75 +1,180 @@
...
+        <% if (authz_result.time_limit_min != undefined) { %>
+        <p class="lead text-center">
+          The time limit for this assessment is
+          <strong>
+            <%= authz_result.time_limit_min %>
+            <%= (authz_result.time_limit_min == 1) ? 'minute' : 'minutes' %>
+          </strong>
+        </p>
+        <% } %>
+        <% if (assessment.group_work) { %>
```
mwest1066, line 36: All this code looks essentially identical to the copy in `pages/studentAssessmentHomework`. Can we use a partial to avoid duplicating it?

pages/studentAssessmentExam/studentAssessmentExam.js
```diff
@@ -21,7 +21,30 @@ router.get('/', function(req, res, next) {
         sqldb.query(sql.select_single_assessment_instance, params, function(err, result) {
             if (ERR(err, next)) return;
             if (result.rowCount == 0) {
-                res.render(__filename.replace(/\.js$/, '.ejs'), res.locals);
+                if (res.locals.assessment.group_work) {
```
mwest1066, line 24: Duplicate code from homework.

pages/studentAssessmentExam/studentAssessmentExam.js
```diff
@@ -37,6 +60,104 @@ router.post('/', function(req, res, next) {
             if (ERR(err, next)) return;
             res.redirect(res.locals.urlPrefix + '/assessment_instance/' + assessment_instance_id);
         });
+    } else if (req.body.__action == 'joinGroup') {
```
mwest1066, line 63: Duplicate code from homework.

pages/studentAssessmentExam/studentAssessmentExam.sql
```diff
@@ -9,3 +9,112 @@ WHERE
     ai.assessment_id = $assessment_id
     AND ai.number = 1
     AND ((gu.user_id = $user_id) OR (ai.user_id = $user_id));
+
+-- BLOCK get_config_info
```
mwest1066, line 13: I think this is basically all duplicated code?

pages/studentAssessmentInstanceExam/studentAssessmentInstanceExam.sql
```diff
@@ -26,3 +26,39 @@ WHERE
 WINDOW
     w AS (ORDER BY qo.row_order)
 ORDER BY qo.row_order;
+
+-- BLOCK get_group_info
```
mwest1066, line 30: Duplicated code.

mwest1066:
I didn't go through and carefully compare all the code, but it looks like this PR is duplicating huge amounts of code (in `.js`, `.ejs`, and `.sql` files). Can we refactor this so we only have one copy of this code?

pages/studentAssessmentExam/studentAssessmentExam.ejs
frankyhao: Yes, so I created an ejs template at pages/shared/ for the "group section" for both homework and exam. If you could review #3023 first, it would be great.

mwest1066: Oh right, I think I remember you telling me that #3023 needed to go in first. I've reviewed that now.

nwalters512:
This was made obsolete by #5290.

PR #2916, opened 2020-08-23
Fix Group Work - displaying not assigned students from other courses' instances

frankyhao:
Thanks @mfsilva22 for pointing out the bug. It was a SQL mistake that did not check course_instance_id.

### DIFF
--- a/pages/instructorAssessmentGroups/instructorAssessmentGroups.sql
+++ b/pages/instructorAssessmentGroups/instructorAssessmentGroups.sql
@@ -40,11 +40,12 @@ SELECT
     u.uid
 FROM
     groups AS gr
-    JOIN group_users AS gu ON gr.group_config_id = $group_config_id AND gu.group_id = gr.id AND gr.deleted_at IS NULL
-    RIGHT JOIN enrollments AS e ON e.user_id = gu.user_id AND e.course_instance_id = $course_instance_id
+    JOIN group_users AS gu ON gu.group_id = gr.id AND gr.group_config_id = $group_config_id AND gr.deleted_at IS NULL
+    RIGHT JOIN enrollments AS e ON e.user_id = gu.user_id
     JOIN users AS u ON u.user_id = e.user_id
 WHERE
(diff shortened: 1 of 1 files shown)


pages/instructorAssessmentGroups/instructorAssessmentGroups.sql
```diff
@@ -40,11 +40,12 @@ SELECT
     u.uid
 FROM
     groups AS gr
-    JOIN group_users AS gu ON gr.group_config_id = $group_config_id AND gu.group_id = gr.id AND gr.deleted_at IS NULL
-    RIGHT JOIN enrollments AS e ON e.user_id = gu.user_id AND e.course_instance_id = $course_instance_id
+    JOIN group_users AS gu ON gu.group_id = gr.id AND gr.group_config_id = $group_config_id AND gr.deleted_at IS NULL
```
mwest1066, line 43: It looks a bit weird to me to have `gr.group_config_id = $group_config_id AND gr.deleted_at IS NULL` in this join condition, given that this is a JOIN with `group_users`. Putting these things into join conditions looks like the same sort of bug that this PR is fixing.

Would it be cleaner to shift `gr.group_config_id = $group_config_id AND gr.deleted_at IS NULL` into the WHERE clause?

frankyhao: I tried this earlier, but it would give a different result as I am using a RIGHT JOIN here.
An alternative way would be: 
leave gr.deleted_at at the same line and move `gr.group_config_id = $group_config_id` to WHERE:
```
AND (gr.group_config_id IS NULL
OR gr.group_config_id != $group_config_id)
```
This way is not straightforward. Which do you prefer?

mwest1066: Let's leave it as is for now. I actually think that this might be much clearer if we just use an EXCEPT to compute the set difference between all enrolled students and all students in groups.

PR #2918, opened 2020-08-23
Fix course instance syncing with deleted course instances

mwest1066:
This is more of the same bug class as #2917 and #2912 (not checking for `WHERE deleted_at IS NULL` during sync).

In this case two of the bugs were serious (would sync data to the deleted course instance rather than to the new non-deleted one) and one of the bugs was harmless for end users.

### DIFF
--- a/sprocs/sync_course_instances.sql
+++ b/sprocs/sync_course_instances.sql
@@ -142,6 +142,7 @@ BEGIN
         FROM course_instances AS ci
         WHERE
             ci.short_name = valid_course_instance.short_name
+            AND ci.deleted_at IS NULL
             AND ci.course_id = syncing_course_id;
 
         INSERT INTO course_instance_access_rules (
--- a/sprocs/sync_course_instances.sql
+++ b/sprocs/sync_course_instances.sql
@@ -229,6 +229,7 @@ BEGIN
     FROM disk_course_instances AS src
     WHERE
         dest.short_name = src.short_name
+        AND dest.deleted_at IS NULL
         AND dest.course_id = syncing_course_id
         AND (src.errors IS NOT NULL AND src.errors != '');
 END;
--- a/sprocs/sync_course_instances.sql
+++ b/sprocs/sync_course_instances.sql
@@ -128,6 +128,7 @@ BEGIN
     FROM disk_course_instances AS src
     WHERE
         dest.short_name = src.short_name
+        AND dest.deleted_at IS NULL
         AND dest.course_id = syncing_course_id
         AND (src.errors IS NULL OR src.errors = '');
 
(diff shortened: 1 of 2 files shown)


tests/sync/courseInstancesSync.js
```diff
@@ -179,4 +179,63 @@ describe('Course instance syncing', () => {
...
+    courseInstance.courseInstance.uuid = '276eeddb-74e1-44e5-bfc5-3c39d79afa85';
+    courseInstance.courseInstance.longName = 'test new long name';
+    await util.overwriteAndSyncCourseData(courseData, courseDir);
+    const syncedCourseInstances = await util.dumpTable('course_instances');
+    const syncedCourseInstance = syncedCourseInstances.find(ci => ci.short_name === 'repeatedCourseInstance' && ci.deleted_at == null);
+    assert.equal(syncedCourseInstance.uuid, courseInstance.courseInstance.uuid);
+    assert.equal(syncedCourseInstance.long_name, courseInstance.courseInstance.longName);
+  });
+
+  it('does not modify deleted course instance long names', async () => {
```
nwalters512, line 199: Do you think it's worth adding similar tests for assessments in https://github.com/PrairieLearn/PrairieLearn/pull/2912?

sprocs/sync_course_instances.sql
```diff
@@ -137,10 +138,11 @@ BEGIN
         FROM disk_course_instances AS src
         WHERE (src.errors IS NULL OR src.errors = '')
     ) LOOP
-        SELECT ci.id INTO syncing_course_instance_id
+        SELECT ci.id INTO STRICT syncing_course_instance_id
```
nwalters512, line 141: Whoa, what does the `STRICT` keyword do?

mwest1066: It ensures that the SELECT query returns exactly one row, which is what goes into `syncing_course_instance_id`. Without STRICT it allows the SELECT to return no rows (in which case `syncing_course_instance_id` is NULL) or multiple rows (in which case it picks the first one at random to put into `syncing_course_instance_id`). This last case was what was happening for us when we didn't have `WHERE deleted_at IS NULL`, and we were sometimes getting the "correct" row that we wanted, and sometimes we were getting one of the old rows.

When we use INTO with an INSERT or UPDATE then it's automatically STRICT. It's only SELECT INTO that needs this. I went through all other cases in the code base where we use SELECT INTO. They can all only return a single row by design, so I didn't bother adding STRICT to them.

mwest1066: I also only learned about STRICT today 😄. I had thought that it was automatically applied (but it turns out it's only automatic for INSERT/UPDATE). See https://www.postgresql.org/docs/current/plpgsql-statements.html#PLPGSQL-STATEMENTS-SQL-ONEROW

PR #2919, opened 2020-08-23
Fix group users comma joins

tbretl:
The block `find_single_assessment_instance` appears in both of these SQL files:
* `pages/studentAssessmentHomework/studentAssessmentHomework.sql`
* `pages/studentAssessmentExam/studentAssessmentExam.sql`

As its name indicates, this block returns an assessment instance, if one exists. Both blocks used comma-joins like this:
```
SELECT
    ai.*
FROM
    assessment_instances AS ai,
    (SELECT *
     FROM group_users AS gi
     JOIN groups AS gr ON gi.group_id = gr.id
     WHERE $user_id = gi.user_id AND gr.deleted_at IS NULL) AS gid
WHERE
    ai.assessment_id = $assessment_id
    AND ai.number = 1
    AND ((ai.group_id = gid.group_id) OR (ai.user_id = $user_id));
```

Unfortunately, these comma-joins cause the block to return nothing (always) if the assessment instance is not associated with group work. So, `studentAssessment*.js` will try to create a new assessment instance, which will fail.

The ultimate consequence of this mistake is that no student will be able to access any assessment instance after the very first time it is created.

This PR eliminates the comma-joins and fixes the bug, restoring access to assessment instances.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -261,6 +261,8 @@
 
   * Fix default institution in course instance access rules (Tim Bretl).
 
+  * Fix bad comma joins in select_single_assessment_instance block queries (Tim Bretl).
+
   * Remove `number` column from `course_instances` table and `number` property from `infoCourseInstance.json` schema (Tim Bretl).
 
   * Remove introduction alert at the top of `homework` assessments (Tim Yang).
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -261,7 +261,6 @@
 
   * Fix default institution in course instance access rules (Tim Bretl).
 
-  * Fix bad comma joins in select_single_assessment_instance block queries (Tim Bretl).
   * Fix `group_work` flag when calling `authz_assessment_instance` (Tim Bretl).
 
   * Remove `number` column from `course_instances` table and `number` property from `infoCourseInstance.json` schema (Tim Bretl).
--- a/pages/studentAssessmentExam/studentAssessmentExam.sql
+++ b/pages/studentAssessmentExam/studentAssessmentExam.sql
@@ -3,10 +3,9 @@ SELECT
     ai.*
 FROM
     assessment_instances AS ai
-    LEFT JOIN (SELECT *
-     FROM group_users AS gi
-     WHERE $user_id = gi.user_id) AS gid ON TRUE
+    LEFT JOIN groups AS g ON (g.deleted_at IS NULL)
+    LEFT JOIN group_users AS gu ON (gu.group_id = ai.group_id AND gu.group_id = g.id)
 WHERE
     ai.assessment_id = $assessment_id
     AND ai.number = 1
-    AND ((ai.group_id = gid.group_id) OR (ai.user_id = $user_id));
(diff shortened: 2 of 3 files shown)


ChangeLog.md
```diff
@@ -261,6 +261,7 @@
 
   * Fix default institution in course instance access rules (Tim Bretl).
 
+  * Fix bad comma joins in select_single_assessment_instance block queries (Tim Bretl).
```
mwest1066, line 264: Don't need ChangeLog entries anymore

pages/studentAssessmentExam/studentAssessmentExam.sql
```diff
@@ -2,11 +2,10 @@
 SELECT
     ai.*
 FROM
-    assessment_instances AS ai,
-    (SELECT *
-     FROM group_users AS gi 
-     WHERE $user_id = gi.user_id) AS gid
+    assessment_instances AS ai
+    LEFT JOIN groups AS g ON (g.deleted_at IS NULL)
+    LEFT JOIN group_users AS gu ON (gu.group_id = ai.group_id AND gu.group_id = g.id)
```
mwest1066, line 7: I think the following is equivalent and it feels a bit more natural to me:

```suggestion
    assessment_instances AS ai
    LEFT JOIN groups AS g ON (g.id = ai.group_id AND g.deleted_at IS NULL)
    LEFT JOIN group_users AS gu ON (gu.group_id = g.id)
```

pages/studentAssessmentHomework/studentAssessmentHomework.sql
```diff
@@ -2,111 +2,109 @@
...
     ai.*
 FROM
-    assessment_instances AS ai,
-    (SELECT *
-     FROM group_users AS gi
-     JOIN groups AS gr ON gi.group_id = gr.id
-     WHERE $user_id = gi.user_id AND gr.deleted_at IS NULL) AS gid
+    assessment_instances AS ai
+    LEFT JOIN groups AS g ON (g.deleted_at IS NULL)
+    LEFT JOIN group_users AS gu ON (gu.group_id = ai.group_id AND gu.group_id = g.id)
```
mwest1066, line 7: Same suggestion:

```suggestion
    assessment_instances AS ai
    LEFT JOIN groups AS g ON (g.id = ai.group_id AND g.deleted_at IS NULL)
    LEFT JOIN group_users AS gu ON (gu.group_id = g.id)
```

ChangeLog.md
tbretl: haha, missed the memo 😄 - reverted

pages/studentAssessmentExam/studentAssessmentExam.sql
tbretl: agreed, done, thanks!

pages/studentAssessmentHomework/studentAssessmentHomework.sql
tbretl: agreed, done, thanks!

PR #2938, opened 2020-08-26
Fix workspace S3 path

tdy:
Not sure this will address all the weirdness on prod, but this bug is at least part of it.

### DIFF
--- a/workspace_host/interface.js
+++ b/workspace_host/interface.js
@@ -323,9 +323,10 @@ async.series([
  */
 async function pushContainerContentsToS3(workspace) {
     const workspacePath = path.join(workspacePrefix, `workspace-${workspace.launch_uuid}`);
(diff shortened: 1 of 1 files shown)


workspace_host/interface.js
```diff
@@ -323,9 +323,10 @@ async.series([
  */
 async function pushContainerContentsToS3(workspace) {
     const workspacePath = path.join(workspacePrefix, `workspace-${workspace.launch_uuid}`);
+    const s3Path = path.join(config.workspaceS3Bucket, `workspace-${workspace.id}-${workspace.version}`);
```
mwest1066, line 326: I think this should have `current/` on the end, right?

tdy: 🤦‍♂️ I think so

mwest1066: See #2975

PR #2941, opened 2020-08-26
Fix MathJax rendering in pl-drawing

nicknytko:


elements/pl-drawing/mechanicsObjects.js
```diff
@@ -771,26 +771,33 @@ mechanicsObjects.LatexText = fabric.util.createClass(fabric.Object, {
         let ref = this;
```
mwest1066, line 771: Add `let retryCount = 0;` here.

elements/pl-drawing/mechanicsObjects.js
```diff
@@ -771,26 +771,33 @@ mechanicsObjects.LatexText = fabric.util.createClass(fabric.Object, {
...
-                /* Replace the reference with the actual value */
-                use.parentNode.replaceChild(replacement, use);
-            });
+                    /* Replace the reference with the actual value */
+                    use.parentNode.replaceChild(replacement, use);
+                });
+            } catch (_err) {
+                /* If we're erroring, it's because the MathJax glyph cache hasn't loaded yet.
+                   We can retry again in a little bit to check if it's ready. */
+                setTimeout(callback, 100);
```
mwest1066, line 798: I'm a bit worried that an actual error in the above code is going to result in an infinite loop and we'll never find out what happened. Perhaps we could do something like:
```suggestion
                if (retryCount++ < 100) {
                    setTimeout(callback, 100);
                } else {
                    throw err;
                }
```

PR #2945, opened 2020-08-27
Add overall timeout on workspace launching

mwest1066:
Also separate out launched-timeout and heartbeat-timeout so we can add better messages to the workspaces in these cases.

### DIFF
--- a/cron/stopStaleWorkspaces.js
+++ b/cron/stopStaleWorkspaces.js
@@ -53,5 +53,5 @@ module.exports.run = function(callback) {
         await stopLaunchedTimeoutWorkspaces();
         await stopHeartbeatTimeoutWorkspaces();
         await stopInLaunchingTimeoutWorkspaces();
-})(callback);
+    })(callback);
 };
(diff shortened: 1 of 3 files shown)


cron/stopStaleWorkspaces.js
```diff
@@ -10,17 +10,48 @@ const sql = sqlLoader.loadSqlEquiv(__filename);
...
-        for (const staleWorkspace of staleWorkspaces) {
-            logger.verbose(`stopStaleWorkspaces: stopping workspace_id = ${staleWorkspace.id}`);
-            await workspaceHelper.updateState(staleWorkspace.id, 'stopped', 'Cron job');
-        }
-    })(callback);
+        await stopLaunchedTimeoutWorkspaces();
+        await stopHeartbeatTimeoutWorkspaces();
+        await stopInLaunchingTimeoutWorkspaces();
+})(callback);
 };
```
tdy, line 57:
```suggestion
    })(callback);
};
```

PR #2947, opened 2020-08-27
Add chown hook to workspace host initial unzip

### LINKED ISSUE #2946: Workspace hosts don't chown after initial unzip
Fixing #2937 revealed that `_getInitialZipAsync()` doesn't call `_workspaceFileChangeOwner()` so all the files end up owned by root and can't be edited. @nicknytko suggests hooking into the extract and calling the chown after we’re done extracting each file https://github.com/ZJONSSON/node-unzipper#parse-zip-file-contents

tdy:
Fixes #2946 

* [x] `chown` extracted files
* [x] support nested dirs

### DIFF
--- a/workspace_host/interface.js
+++ b/workspace_host/interface.js
@@ -598,6 +598,7 @@ function _getSettingsWrapper(workspace, callback) {
 }
 
 function _workspaceFileChangeOwner(filepath, callback) {
+    debug(`Enforcing file ownership for ${filepath}`);
     if (config.workspaceJobsDirectoryOwnerUid == 0 ||
         config.workspaceJobsDirectoryOwnerGid == 0) {
         /* No-op if there's nothing to do */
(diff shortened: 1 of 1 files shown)


nicknytko:
At the point when we call `_getInitialZip()` is the `/jobs/workspaces/workspace-*` home folder created already?  I think we should check that we're also chown'ing that.

tdy:
> At the point when we call `_getInitialZip()` is the `/jobs/workspaces/workspace-*` home folder created already? I think we should check that we're also chown'ing that.

Updated the code to `chown` this base dir as well as extracted subdirs.

workspace_host/interface.js
```diff
@@ -764,8 +765,27 @@ async function _getInitialZipAsync(workspace) {
     debug(`Downloading s3Path=${s3Path} to zipPath=${zipPath}`);
     await _downloadFromS3Async(zipPath, s3Path);
 
+    try {
+        await fsPromises.lstat(localPath);
```
mwest1066, line 769: Can we skip this `lstat()` call and just always call `mkdir()` and `_workspaceFileChangeOwnerAsync()`?

PR #2952, opened 2020-08-27
Add percentage indicator when pulling workspace images

tdy:
~~Currently this writes a new `workspace_logs.message` for each percentage update, which probably isn't ideal. Maybe we could add an `updateMessage` flag about whether to write to the DB?~~

* [x] update `message` badge with percentage estimate when pulling image
* [x] bypass db logging

pl-pull-layer-progress


mwest1066:
I love this! I agree we should add a flag to `updateMessage()` to tell it to skip writing this message to the DB.

lib/workspace.js
```diff
@@ -143,12 +143,26 @@ module.exports = {
...
+     */
+    async updateMessage(workspace_id, message, toDatabase=true) {
         debug(`Setting workspaces.message to '${message}'`);
-        await sqldb.callAsync('workspaces_message_update', [workspace_id, message]);
+        if (toDatabase) await sqldb.callAsync('workspaces_message_update', [workspace_id, message]);
         module.exports._namespace.to(workspace_id).emit('change:message', {workspace_id, message});
     },
 
+    /**
+     * Updates a workspace's current state.
```
tdy, line 160:
```suggestion
     * Updates a workspace's current state and message.
```

workspace_host/interface.js
```diff
@@ -813,11 +813,24 @@ function _pullImage(workspace, callback) {
...
                 debug('Docker pull output: ', output);
+                if ('progressDetail' in output && output.progressDetail.total) {
+                    progressDetails[output.id] = output.progressDetail;
+                }
+                layers = Object.keys(progressDetails).length;
+                current = Object.values(progressDetails).reduce((current, detail) => detail.current + current, 0);
+                total = Object.values(progressDetails).reduce((total, detail) => detail.total + total, 0);
+                if (total) {
+                    percent = parseInt(100 * current / total);
+                    workspaceHelper.updateMessage(workspace.id, `Pulling ${layers} layers (≈${percent}%)`, toDatabase);
```
nicknytko, line 832: Random thought - I imagine some end users have no idea what a docker layer is.  Maybe we should have this message be something like `Pulling ${layers} parts (${percent}%)`, or even just have the percentage?  Would like to know what you think, though.

mwest1066: I'd be personally happy to just see a percentage, and I wouldn't bother with the approximately equals?

mwest1066: I'd suggest something like `Pulling image (13%)`

tdy: Ah good point, I forgot to mention a couple things:

**Layer count:** Every time another layer gets added, the percentage drops. Sometimes this is a significant drop, like maybe 90% back down to 15% if it's a large layer. I was just worried that (seemingly) random drops would be frustrating for end users, but I can drop the layer count if you guys don't think this matters much.

**Approximately equals:** Towards the end of the pull, the percentage estimate fluctuates quite a bit for some reason (even after layers have stabilized). Like it'll reach 99%, then back down to 97%, then get back up to 99% but drop down again, and so on for like the last 5 seconds. I'm not sure if dockerode's own `total` values are just estimates that fluctuate as the downloads finalize, or if it's a bug in my code, or what.

nicknytko: could we instead just display the number of downloaded layers?

tdy: Hmm then I think the large layers would just stall out the percentage for long periods?

Also I don't think we can get the total layer count from dockerode, so the downloaded-layer-percent would also fluctuate a lot. It just starts pulling some layers in parallel and adds more layers randomly. Like xterm starts with a few layers, then adds a couple more layers while the first few are still going, then eventually adds a few more, etc.

mwest1066:
I changed the calculation to compute the percentage incrementally, so now it never goes backwards.

PR #2955, opened 2020-08-27
Make pl-file-upload do fuzzy case match and rename silently

echuber2:
When the user uploads a file that differs from an accepted filename only by case, it considers it a match and renames the uploaded file to the expected case version.

My implementation may be a little inelegant but I did test it. Could make it optional. Please point out any huge security issues this may raise...


echuber2:
I'm assuming this will need at least a rewrite for cleanness and to be made optional, but I'll have to come back to it on another day in that case

elements/pl-file-upload/pl-file-upload.js
```diff
@@ -11,6 +11,11 @@ window.PLFileUpload = function(uuid, options) {
     this.uuid = uuid;
     this.files = options.files || [];
     this.acceptedFiles = options.acceptedFiles || [];
+    this.acceptedFilesLowerCase = [];
+    var i;
+    for (i = 0; i < this.acceptedFiles.length; i++) {
```
mwest1066, line 16:
```suggestion
    for (let i = 0; i < this.acceptedFiles.length; i++) {
```

echuber2: I'm fine with this but I'd thought maybe it was written in a style for pre-ES2015 compatibility.

elements/pl-file-upload/pl-file-upload.js
```diff
@@ -34,16 +38,27 @@ window.PLFileUpload.prototype.initializeTemplate = function() {
...
+            var fileNameLowerCase = file.name.toLowerCase();
+            if (_.includes(that.acceptedFilesLowerCase, fileNameLowerCase)) {
                 return done();
             }
             return done('invalid file');
         },
         addedfile: function(file) {
-            if (!_.includes(that.acceptedFiles, file.name)) {
+            // fuzzy case match
+            var fileNameLowerCase = file.name.toLowerCase();
```
tdy, line 50:
```suggestion
            const fileNameLowerCase = file.name.toLowerCase();
```

elements/pl-file-upload/pl-file-upload.js
```diff
@@ -34,16 +38,27 @@ window.PLFileUpload.prototype.initializeTemplate = function() {
         url: '/none',
         autoProcessQueue: false,
         accept: function(file, done) {
-            if (_.includes(that.acceptedFiles, file.name)) {
+            // fuzzy case:
+            var fileNameLowerCase = file.name.toLowerCase();
```
tdy, line 42:
```suggestion
            const fileNameLowerCase = file.name.toLowerCase();
```

elements/pl-file-upload/pl-file-upload.js
```diff
@@ -34,16 +38,27 @@ window.PLFileUpload.prototype.initializeTemplate = function() {
...
                 that.addWarningMessage('<strong>' + file.name + '</strong>' + ' did not match any accepted file for this question.');
                 return;
             }
+            var matchIdx;
+            for (matchIdx = 0; matchIdx < that.acceptedFilesLowerCase.length; matchIdx++) {
+                if (that.acceptedFilesLowerCase[matchIdx] === fileNameLowerCase) {
+                    break;
+                }
+            }
+            var acceptedName = that.acceptedFiles[matchIdx];
```
tdy, line 61:
```suggestion
            const acceptedFilesIdx = that.acceptedFilesLowerCase.indexOf(fileNameLowerCase);
            const acceptedName = that.acceptedFiles[acceptedFilesIdx];
```

elements/pl-file-upload/pl-file-upload.js
```diff
@@ -52,10 +59,10 @@ window.PLFileUpload.prototype.initializeTemplate = function() {
 
                 // Store the file as base-64 encoded data
                 var base64FileData = dataUrl.substring(commaSplitIdx + 1);
-                that.saveSubmittedFile(file.name, base64FileData);
+                that.saveSubmittedFile(acceptedName, base64FileData);
                 that.renderFileList();
                 // Show the preview for the newly-uploaded file
-                that.element.find('li[data-file="' + file.name + '"] .file-preview').addClass('in');
+                that.element.find('li[data-file="' + acceptedName + '"] .file-preview').addClass('in');
```
tdy, line 65:
```suggestion
                that.element.find(`li[data-file="${acceptedName}"] .file-preview`).addClass('in');
```

tdy: one more nit

elements/pl-file-upload/pl-file-upload.js
```diff
@@ -34,16 +35,22 @@ window.PLFileUpload.prototype.initializeTemplate = function() {
...
             return done('invalid file');
         },
         addedfile: function(file) {
-            if (!_.includes(that.acceptedFiles, file.name)) {
+            // fuzzy case match
+            const fileNameLowerCase = file.name.toLowerCase();
+            if (!_.includes(that.acceptedFilesLowerCase, fileNameLowerCase)) {
                 that.addWarningMessage('<strong>' + file.name + '</strong>' + ' did not match any accepted file for this question.');
                 return;
             }
```
tdy, line 51:
```suggestion
            }
            if (file.name !== fileNameLowerCase) {
                that.addWarningMessage(`<strong>${file.name}</strong> was automatically renamed to <strong>${fileNameLowerCase}</strong>.`);
            }
```
I know you explicitly said "rename silently" but maybe it'd be nice to show this message? Not sure, just an idea.

echuber2: We rename it to `acceptedName` when we save it, not the lowercase name. The lowercase version is only for finding the match.

elements/pl-file-upload/pl-file-upload.js
```diff
@@ -34,16 +35,22 @@ window.PLFileUpload.prototype.initializeTemplate = function() {
...
+            const acceptedFilesIdx = that.acceptedFilesLowerCase.indexOf(fileNameLowerCase);
+            const acceptedName = that.acceptedFiles[acceptedFilesIdx];
```
tdy, line 53:
```suggestion
            const acceptedName = that.acceptedFiles[acceptedFilesIdx];
            if (file.name !== acceptedName) {
                that.addWarningMessage(`<strong>${file.name}</strong> was automatically renamed to <strong>${acceptedName}</strong>.`);
            }
```
Ah right, so it would be something like this if we wanted to give a renaming message. Not sure if we want to do this, just a thought.

mwest1066: I'm inclined to not show any warnings. Students using Windows or MacOS are probably unaware of the case of their filename extensions, for example, and so they are going to get these warnings that won't make any sense to them.

elements/pl-file-upload/pl-file-upload.js
echuber2: @tdy I'm not against the message but can we move it to target `acceptedName`?

echuber2: Sorry, just now my page refreshed. Heh.

elements/pl-file-upload/pl-file-upload.js
echuber2: I guess that's reasonable. If a warning implies that the user should take action, this may be unnecessarily confusing now.

PR #2960, opened 2020-08-28
Add workspace timeout detect for update manager

nicknytko:


workspace_host/interface.js
```diff
@@ -207,6 +207,7 @@ async.series([
         });
         watcher.on('add', filename => {
             // Handle new files
+            logger.info(`Watching file ${filename}`);
```
mwest1066, line 210: Should we also log `change` events? Those are the ones I've been testing by editing a file.

workspace_host/interface.js
```diff
@@ -243,14 +245,19 @@ async.series([
             update_queue[key] = {action: 'delete'};
         });
         async function autoUpdateJobManagerTimeout() {
+            const timeout_id = setTimeout(() => {
+                logger.info(`_autoUpdateJobManager() timed out, update queue:\n${update_queue}`);
+            }, config.workspaceHostFileWatchIntervalSec * 1000);
             await _autoUpdateJobManager();
+            clearTimeout(timeout_id);
```
mwest1066, line 252: Another approach to make sure `_autoUpdateJobManager()` is running is to do something like:
```
let lastRunTime = Date.now();

setInterval(() => {
    const elapsedSec = (Date.now() - lastRunTime) / 1000;
    if (elapsedSec > 30) {
        logger.error(`_autoUpdateJobManager() has not run for ${elapsedSec} seconds`);
    }
}, 1000);

# in _autoUpdateJobManager:
    lastRunTime = Date.now();
```

PR #2961, opened 2020-08-28
Student Assessment Panel Redesign

astrosticks:
A graphical redesign of the card body in `pages/studentAssessmentInstanceHomework`.

- Sections now have headers to indicate what they are
- Group section uses horizontal space more wisely
- Emphasizing relevant information
- Responds appropriately to screen resizing

(click on images to see them in full size)

### Invites enabled
| New | Old |
| --- | --- | 
| image | image |

### Invites disabled
| New | Old |
| --- | --- | 
| image | image |

---

Once groupwork is supported for exams, these changes should be easy to apply to `pages/studentAssessmentInstanceExam`.

### DIFF
--- a/pages/studentAssessmentInstanceHomework/studentAssessmentInstanceHomework.ejs
+++ b/pages/studentAssessmentInstanceHomework/studentAssessmentInstanceHomework.ejs
@@ -79,11 +79,11 @@
                 </div>
               <% } %>
               <% if (groupinfo[0].student_authz_leave) { %>
-              <form id="confirm-form" name="confirm-form" method="POST" class="mt-2">
-                <input type="hidden" name="__action" value="leaveGroup">
-                <input type="hidden" name="__csrf_token" value="<%= __csrf_token %>">
-                <button id="leave-group" type="submit" class="btn btn-sm btn-danger">Leave group</button>
-              </form>
+                <form id="confirm-form" name="confirm-form" method="POST" class="mt-2">
+                  <input type="hidden" name="__action" value="leaveGroup">
+                  <input type="hidden" name="__csrf_token" value="<%= __csrf_token %>">
+                  <button id="leave-group" type="submit" class="btn btn-sm btn-danger">Leave group</button>
+                </form>
               <% } %>
             </div>
           <% } %>
--- a/pages/studentAssessmentInstanceHomework/studentAssessmentInstanceHomework.ejs
+++ b/pages/studentAssessmentInstanceHomework/studentAssessmentInstanceHomework.ejs
@@ -51,7 +51,7 @@
         <% if (assessment.group_work) { %>
           <div class="card-body border-top bg-light">
             <div>
-              <p class="font-weight-bold mb-0">Group assessment</p>
+              <p class="font-weight-bold mb-2">Group assessment</p>
               <span>
                 <% if (groupinfo[0].student_authz_join) { %>
                   You are in 
@@ -91,7 +91,7 @@
 
         <% if (assessment_text_templated) { %>
           <div class="card-body border-top">
-            <p class="font-weight-bold mb-0">Instructor's note</p>
+            <p class="font-weight-bold mb-2">Instructor's note</p>
             <%- assessment_text_templated %>
           </div>
         <% } %>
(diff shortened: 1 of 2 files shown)


pages/studentAssessmentInstanceHomework/studentAssessmentInstanceHomework.ejs
```diff
@@ -47,41 +47,54 @@
...
                 <% } %>
-                <span class="badge badge-secondary">Group Members</span>
-                <% groupinfo.forEach(function(row) {%>
-                <li><%= row.uid %></li>
-                <% }); %>
-              </div>
+              </span>
+              <%# Needs to be without whitespace (newlines) to avoid automatic spaces between list elements %>
+              <ul class="d-inline list-inline mb-0">
+                <% groupinfo.forEach(function(row) {%><li class="list-inline-item badge border"><%= row.uid %></li><% }); %>
```
tdy, line 68:
```suggestion
                <% groupinfo.forEach(function(row) {%><li class="list-inline-item"><a class="badge border" href="mailto:<%= row.uid %>"><%= row.uid %></a></li><% }); %>
```
What about making these badges clickable (`mailto`)? Just a thought, not particularly attached to the idea either way.

astrosticks: Great idea!

astrosticks: Are all uids email addresses though?

pages/studentAssessmentInstanceHomework/studentAssessmentInstanceHomework.ejs
```diff
@@ -47,41 +47,54 @@
...
-                <% if (groupinfo[0].student_authz_leave) { %>
-                <form id="confirm-form" name="confirm-form" method="POST" class="d-flex justify-content-end">
-                  <input type="hidden" name="__action" value="leaveGroup">
-                  <input type="hidden" name="__csrf_token" value="<%= __csrf_token %>">
-                  <button id="leave-group" type="submit" class="btn btn-danger">Leave the Group</button>
-                </form>
+        </div>
+        <% if (assessment.group_work) { %>
+          <div class="card-body border-top bg-light">
+            <div>
```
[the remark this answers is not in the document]
astrosticks: Nope, deleted!

pages/studentAssessmentInstanceHomework/studentAssessmentInstanceHomework.ejs
tdy: Ah good point, not sure about that.

astrosticks: Can always add it later if it's requested, but since we don't have a guarantee, we should probably keep it simple

mwest1066: They aren’t necessarily email addresses. Illinois does use NetIDs that are identical to email addresses, but other schools don’t (e.g., UBC).  Also LTI users have non-email UIDs.

tdy:
The more I look at this, the more I'm thinking the original version is actually much easier to scan/read. I do see your point about the big horizontal gap when invites are disabled though, so I dunno.

I guess one possibility would be to keep the original layout but just polish it a bit? Like restyling with `.list-group`, better button positioning, etc. Or do you feel strongly about this inline/horizontal layout?

mwest1066:
Why no team name when invites are disabled?

mwest1066:
Minor suggestion: I'd be inclined to do:
```
Group assessment: You are in (team0). [Leave group]
Your teammates are: ...
---
Use join ...
```

I think the [Leave group] button should logically go close to the team name.

frankyhao:
@astrosticks Thank you so much for helping with the UI.
Please see [PR#3023](https://github.com/PrairieLearn/PrairieLearn/pull/3023) as well. This section appears many times in different pages, so I just create a template. You only need to modify pages/share/groupWorkInfoContainer.ejs.
I agree with Matt that it is logical to put Leave next to the group information. And I like your way of displaying each group member within a [box].
However, personally, in the computer view, I prefer a multi-column display. And maybe you can also polish the grey-box group name a little bit, which does not look formal to me.

astrosticks:
Thanks for the feedback everyone! I'll close this PR and wait until Franky's gets merged before submitting any more changes.

Issue #2927, opened 2020-08-24
Gradebook duplicating entries for some users


daviddalpiaz:
Screenshot shows two rows of the a gradebook. First row is normal. (Number of entries matches number of assignments.) All other rows in the gradebook appear normal like this. Second row is the outlier. It has double the spaces expected.

![Screen Shot 2020-08-24 at 11 06 18 AM](https://user-images.githubusercontent.com/9003346/91069400-e1d0f380-e624-11ea-8220-9858d8b06863.png)

tdy:
The duping is occurring like this:

Q1 | Q2 | E1 | E2 | | | | |
----- | ----- | ----- | ----- | ----- | ----- | ----- | ----- 
q1_score | q2_score | e1_score | e2_score
q1_score | q1_score | q2_score | q2_score | e1_score | e1_score | e2_score | e2_score
q1_score | q2_score | e1_score | e2_score
q1_score | q2_score | e1_score | e2_score

So far I've noticed that the dupes are students who have taken your course previously (though possibly just a red herring). I'm guessing this has something to do with the group changes in `instructorGradebook.sql`.

mwest1066:
Yep, I bet the group changes are responsible.

zhining4:
I will fix that soon.

daviddalpiaz:
Update: The number of users with this duplication is growing. Also, I have now seen at least one user that has each record tripled.

zhining4:
I have fixed and tested that, it should work now. Just wait for the merge!

PR #2964, opened 2020-08-28
Group work fix on instructorGradebook

zhining4:
That pull request is to address the issue in #2927  and I will use this branch for later fix based on Matt's comments.

### DIFF
--- a/pages/instructorGradebook/instructorGradebook.sql
+++ b/pages/instructorGradebook/instructorGradebook.sql
@@ -1,26 +1,38 @@
 -- BLOCK course_assessments
 SELECT
-    a.id AS assessment_id,a.number AS assessment_number,
+    a.id AS assessment_id,
+    a.number AS assessment_number,
     aset.number AS assessment_set_number,aset.color,
     (aset.abbreviation || a.number) AS label
-FROM assessments AS a
-JOIN assessment_sets AS aset ON (aset.id = a.assessment_set_id)
-WHERE a.deleted_at IS NULL
-AND a.course_instance_id = $course_instance_id
+FROM 
+    assessments AS a
+    JOIN assessment_sets AS aset ON (aset.id = a.assessment_set_id)
+WHERE 
+    a.deleted_at IS NULL
+    AND a.course_instance_id = $course_instance_id
 ORDER BY (aset.number, a.order_by, a.id);
 
 -- BLOCK user_scores
 WITH
 course_assessments AS (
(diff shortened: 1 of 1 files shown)


pages/instructorGradebook/instructorGradebook.sql
```diff
@@ -49,7 +61,10 @@ scores AS (
         course_users AS u
         CROSS JOIN course_assessments AS a
         LEFT JOIN group_users AS gu ON (u.user_id = gu.user_id)
-        LEFT JOIN course_scores AS s ON (((s.user_id = u.user_id)OR(s.group_id = gu.group_id)) AND s.assessment_id = a.id)
+        LEFT JOIN groups AS gr ON (gr.id = gu.group_id)
+        LEFT JOIN course_scores AS s ON (((s.user_id = u.user_id) OR (s.group_id = gu.group_id)) AND s.assessment_id = a.id)
+     WHERE 
+         gr.deleted_at IS NULL
```
tdy, line 67: Just to confirm, this part is the main fix for the bug, right? And the rest of the changes are purely for reformatting?

zhining4: Yeah. That bug occurs when old groups deleted and the student join a new group, I did not check old groups' delete_at.

pages/instructorGradebook/instructorGradebook.sql
```diff
@@ -1,26 +1,38 @@
 -- BLOCK course_assessments
 SELECT
-    a.id AS assessment_id,a.number AS assessment_number,
+    a.id AS assessment_id,
+    a.number AS assessment_number,
```
mwest1066, line 4: While I appreciate the effort to clean up code, please don't introduce additional unrelated changes (such as code reformatting) into bug fix PRs. It makes it hard for reviewers to see what's actually changed in the PR.

pages/instructorGradebook/instructorGradebook.sql
mwest1066: While this is a bug, I don't think it's actually the bug here. The problems with the gradebook are showing up for courses that aren't even using groups at all. I think the problem is that you are JOINing over all `group_users`, even if they are unrelated to the current course_instance or assessment.

mwest1066: I submitted PR #2965 which I think is a fix for this bug. But please check that PR!

mwest1066:
Closing in favor of #2965

PR #2967, opened 2020-08-28
Add admin queries for workspaces

mwest1066:
I might need to add a couple of indexes to speed these queries up.

### DIFF
--- a/admin_queries/workspace_hosts_all_unhealthy.json
+++ b/admin_queries/workspace_hosts_all_unhealthy.json
@@ -1,0 +1,3 @@
+{
+    "description": "Set all workspace hosts to unheathly."
+}
--- a/admin_queries/workspace_hosts_all_unhealthy.json
+++ b/admin_queries/workspace_hosts_all_unhealthy.json
@@ -1,3 +1,3 @@
 {
-    "description": "Set all workspace hosts to unheathly."
+    "description": "Set all workspace hosts to unhealthy."
 }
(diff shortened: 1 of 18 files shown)


nicknytko:
Maybe a "drain current workspace hosts" (set to unhealthy) when deploying a new host image?  And maybe one to set values in the config table.

admin_queries/workspace_hosts_all_unhealthy.json
```diff
@@ -0,0 +1,3 @@
+{
+    "description": "Set all workspace hosts to unheathly."
```
nicknytko, line 2:
```suggestion
    "description": "Set all workspace hosts to unhealthy."
```

admin_queries/workspaces.sql
```diff
@@ -0,0 +1,49 @@
...
+        a.id AS assessment_id,
+        v.authn_user_id,
+        count(*) AS workspace_count
+    FROM
+        workspaces AS w
+        JOIN variants AS v ON (v.workspace_id = w.id)
+        JOIN instance_questions AS iq ON (iq.id = v.instance_question_id)
+        JOIN assessment_instances AS ai ON (ai.id = iq.assessment_instance_id)
+        JOIN assessments AS a ON (a.id = ai.assessment_id)
+    WHERE w.state != 'stopped'
```
nicknytko, line 13: Will this query grab recently stopped workspaces as well?

mwest1066: No, but I think that's ok because we aren't displaying any details about individual workspaces. This query is more to find out which courses/assessments are causing all the workspace load (equivalent to `current_submission_activity.sql`). I find this useful after looking out CloudWatch and thinking "wow, there is huge load spike, what's causing that". Then this query can tell me "it's CS 357, they just started MP2".

Does that sound reasonable, or do you think we should include recently-stopped workspaces in this?

However, your question did make me realize that the above query was including `uninitialized` workspaces, which is wrong. I fixed it just now.

nicknytko: Right that seems reasonable, I only ask because the query description includes recently stopped workspaces.  I'm fine with not including recently stopped workspaces.

mwest1066: Ooops. Good call. Description is fixed.

Issue #2935, opened 2020-08-26
Notes about workspace hosts


mwest1066:
* [x] How do hosts in state `terminating` eventually get to state `terminated`? If I'm understanding correctly, this will be done by `sql.set_terminated_hosts_if_not_launching` in `cron/workspacesTransition.js`. But this uses `select_nonterminated_or_nonterminating_workspace_hosts` to select the hosts to be terminated, so I don't see how we are ever going to select a `terminating` host to be able to change its state to `terminated`?

* [x] @tdy I think this LEFT JOIN on `workspaces` is not needed: https://github.com/PrairieLearn/PrairieLearn/blob/fed9309f4c17bd527ae0578d5919c26ac7478c08/sprocs/workspace_hosts_find_terminable.sql#L18

* [x] @tdy I think this `WHERE` clause: https://github.com/PrairieLearn/PrairieLearn/blob/fed9309f4c17bd527ae0578d5919c26ac7478c08/sprocs/workspace_hosts_find_terminable.sql#L32
Can be more succinctly written as:
```
    UPDATE workspace_hosts AS wh
    SET state = 'terminating',
        state_changed_at = NOW()
    FROM terminable_hosts AS th
    WHERE wh.id = th.id;
```

* [x] @tdy Same `WHERE` change can be applied here: https://github.com/PrairieLearn/PrairieLearn/blob/fed9309f4c17bd527ae0578d5919c26ac7478c08/sprocs/workspace_hosts_recapture_draining.sql#L22
and here:
https://github.com/PrairieLearn/PrairieLearn/blob/fed9309f4c17bd527ae0578d5919c26ac7478c08/sprocs/workspace_hosts_drain_extra.sql#L18

* [x] @tdy I think it would be more accurate to shift the `recaptured_hosts` count to be the output of the UPDATE (not the SELECT) because it's not guaranteed that we actually make the same number of UPDATEs (for example, someone else might terminate one of the `found_draining_hosts` before we convert it back to `ready`): https://github.com/PrairieLearn/PrairieLearn/blob/fed9309f4c17bd527ae0578d5919c26ac7478c08/sprocs/workspace_hosts_recapture_draining.sql#L16
To return a count from an UPDATE we need to put the UPDATE into a CTE (WITH clause). Given this, it might be simpler to also make `found_draining_hosts` another CTE before the UPDATE, instead of using temporary tables.

[...]

PR #2969, opened 2020-08-29
Update workspace host sprocs/crons

tdy:
Described in #2935


sprocs/workspace_hosts_find_terminable.sql
```diff
@@ -15,7 +15,7 @@ BEGIN
             wh.*
         FROM
```
nicknytko, line 16: we can probably remove the join and group entirely from this query, I think I was originally trying to calculate the number of running workspaces before realizing that there was a `load_count` column

PR #2974, opened 2020-08-29
Fix workspace host S3 path

tdy:


mwest1066:
Oh sorry! I didn't notice this PR. Could you take a look at #2975 where I also cleaned up the directory sync function and moved it to `lib/aws.js`?

tdy:
Subsumed by #2975

PR #2983, opened 2020-08-31
Add strict load limits for workspace hosts

mwest1066:


nicknytko:
This looks good, the 60*10 = 600 seconds *should* be enough wait time for a new host to come online.

mwest1066:
> This looks good, the 60*10 = 600 seconds _should_ be enough wait time for a new host to come online.

Yep, we can play with this in practice. From what I saw so far, new hosts are coming up within a minute or two.

PR #2988, opened 2020-08-31
Refactor recaptured_hosts count using CTEs

tdy:
From #2935:

* [x] I think it would be more accurate to shift the `recaptured_hosts` count to be the output of the UPDATE (not the SELECT) because it's not guaranteed that we actually make the same number of UPDATEs (for example, someone else might terminate one of the `found_draining_hosts` before we convert it back to `ready`): https://github.com/PrairieLearn/PrairieLearn/blob/fed9309f4c17bd527ae0578d5919c26ac7478c08/sprocs/workspace_hosts_recapture_draining.sql#L16
To return a count from an UPDATE we need to put the UPDATE into a CTE (WITH clause). Given this, it might be simpler to also make `found_draining_hosts` another CTE before the UPDATE, instead of using temporary tables.

### DIFF
--- a/sprocs/workspace_hosts_recapture_draining.sql
+++ b/sprocs/workspace_hosts_recapture_draining.sql
@@ -5,21 +5,27 @@ CREATE OR REPLACE FUNCTION
     )
 AS $$
 BEGIN
+    WITH
     -- Find ids of at most `needed_hosts` hosts that are currently draining
-    CREATE TEMPORARY TABLE found_draining_hosts ON COMMIT DROP AS (
+    found_draining_hosts AS (
         SELECT *
         FROM workspace_hosts AS wh
         WHERE wh.state = 'draining'
         ORDER BY launched_at DESC
         LIMIT needed_hosts
-    );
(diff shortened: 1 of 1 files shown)


sprocs/workspace_hosts_recapture_draining.sql
```diff
@@ -5,21 +5,27 @@ CREATE OR REPLACE FUNCTION
...
+    -- Update found hosts to be ready if still draining
+    updated_draining_hosts AS (
+        UPDATE workspace_hosts AS wh
+        SET
+            state = 'ready',
+            state_changed_at = NOW()
+        FROM found_draining_hosts AS fdh
+        WHERE
+            wh.id = fdh.id
+            AND wh.state = 'draining'
```
tdy, line 26: Added this check for `draining`, but not sure if it's still needed after converting to CTEs?

mwest1066: Good question, I'm actually not certain either, but I think it can't hurt.

PR #3018, opened 2020-09-01
Add min/max number of ready hosts to autoscaling code

### LINKED ISSUE #2992: Add min and max workspace hosts for autoscaling
These should be values in the config table

nicknytko:
imposes limits on max/min *ready* hosts, but this can be changed so that the max applies to hosts in *all* states

fixes #2992


mwest1066:
> imposes limits on max/min _ready_ hosts, but this can be changed so that the max applies to hosts in _all_ states

I think we should clamp the max for all hosts. I'm worried about unhealthy hosts. Say we have a bug that starts marking hosts as unhealthy at a high rate, we are going to enter into a runaway loop where we spin up a new ready host, quickly mark it as unhealthy, spin up a new one, etc.

Clamping the max for all hosts will mean that we might get stuck with all our hosts unhealthy, which would be bad, but at least we won't runaway and spin up thousands of hosts.

mwest1066 (commented):
See previous comment.

nicknytko:
> See previous comment.

Thanks for the poke, I definitely need to wrap this PR up!

nwalters512:
I'm going to close this; workspaces have been running fine for almost three years now, and I'm pretty sure *not* having an explicit cap has saved us more than once.

PR #3035, opened 2020-09-02
File upload s3

andrewstec:
The `pl-file-upload` currently base64 encodes student file upload data and posts the data as part of a Submission that ends up in the database. We would like the data to be posted as a submission that can be parsed and graded, but stored in S3 instead of the database.

Implemented thus far: 

- S3 upload and read operations added to file-store API
- pl-file-upload element has been modified to validate s3 file storage data in addition to base64 encoded data submitted to the database (backwards compatible for data that already exists in DB)
- file download from S3 works.

Loose ends:

- Students currently cannot download Base64 submissions or S3 submissions. Do we want them to have this ability?
- Side-panel student file upload panel displays S3 uploads from pl-file-upload due to duplicate query.
- pl-file-upload element has to be made more robust after recent changes (reload 
- After a file is submitted, it appears in the Submitted panel. It also however appears in the new Question panel. The data is encoded and ready to re-submit, which is not something we can easily do with S3 unless we want to re-download the file each time the user loads the view: 
image
- Need to produce the right download link in the pl-file-upload element based on an Instructor/Student role with different available parameters.

Questions: 

- In terms of security, we have a query param that includes an assessment_instance_id. I assume that the assessment_instance_id is unique to a student and always refers to the students work in the query. However, what is to stop the malicious submission of an incrementing `assessment_instance_id`, `file_id` and `display_filename` (assumed to be fib.py etc. based on assignment) to collect data? A random Google authenticated user may have this ability. I have to look at the authentication to understand the security better.
- We have different functionality when the Instructor uses the pl-file-upload element. As an assessment and question ID will be stored in the Files data, other instructors will be able to see previously uploaded files for that assessment/question combination. Is this an acceptable approach? We can modify the query to use the user ID under the created_by Files row.


lib/file-store.js
```diff
@@ -74,10 +105,9 @@ module.exports.delete = async (file_id, authn_user_id) => {
  * @param {number} file_id - The file to get.
  * @return {object} An object with a buffer (of the file contents) and a file object.
  */
-module.exports.get = async (file_id) => {
-    debug(`get(): file_id=${file_id}`);
-
-    const params = { file_id };
+module.exports.get = async (file_id, assessment_instance_id, display_filename) => {
+    debug(`get(): file_id=${file_id} storage_type=${assessment_instance_id} display_filename=${display_filename}`);
```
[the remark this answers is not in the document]
andrewstec: The only problem with the approach above is that we get a list of Assessments to pull IDs from res.locals, so I am not too sure if it is safe to pick the first item in the list. That would allow us to change query approaches to pull the right file data.

image

lib/config.js
```diff
@@ -43,8 +43,10 @@ config.serverPort = '3000';
 config.sslCertificateFile = '/etc/pki/tls/certs/localhost.crt';
 config.sslKeyFile = '/etc/pki/tls/private/localhost.key';
 config.sslCAFile = '/etc/pki/tls/certs/server-chain.crt';
-config.fileUploadMaxBytes = 1e7;
+config.fileUploadMaxBytes = 1e8;
 config.fileUploadMaxParts = 1000;
+config.fileUploadS3Bucket = 'prairielearn-dev';
+config.fileUploadS3MinBytes = 1e7;
```
andrewstec, line 49: Small files can currently go in the database, as PL already demonstrates. 

We can decide to support small files (ie. < 200kb) or we can upload all files in S3 if we would like. Any thoughts on which approach is preferable?

lib/question.js
```diff
@@ -1337,6 +1408,32 @@ module.exports = {
...
+
+                if (locals.submissions[0] && locals.submissions[0].submitted_answer['_file_upload_storage']) {
+                    var params = {
+                        file_ids: locals.submissions[0].submitted_answer['_file_upload_storage']['file_ids'],
+                    };
+
+                    sqldb.query(sql.select_files, params, (err, result) => {
+                        if (ERR(err, callback)) return;
+                        if (result.rowCount >= 1) {
+                            // Do I want to fetch the file through S3 or send hyperlinks back?
```
andrewstec, line 1424: Is there an easy way to assemble the hyperlink to the API in *.js files? I have not found one yet. We have the assessment_instance_id, question_instance, etc. to form any hyperlink on the back-end, but we may not have that on the front-end.

I will figure out this logic soon, but open to input!

pages/shared/studentAssessmentInstance.js
```diff
@@ -4,12 +4,12 @@ const fileStore = require('../../lib/file-store');
 
 module.exports.processFileUpload = async (req, res) => {
     if (!res.locals.assessment_instance.open) throw new Error(`Assessment is not open`);
-    await fileStore.upload(req.file.originalname, req.file.buffer, 'student_upload', res.locals.assessment_instance.id, null, res.locals.user.user_id, res.locals.authn_user.user_id);
+    await fileStore.upload(req.file.originalname, req.file.buffer, 'student_upload', res.locals.assessment_instance.id, null, res.locals.user.user_id, res.locals.authn_user.user_id, fileStore.storageTypes.FileSystem);
```
andrewstec, line 7: Need to finalize these parameters and update them before being done.

pages/studentInstanceQuestionHomework/studentInstanceQuestionHomework.js
```diff
@@ -32,10 +32,19 @@ function processSubmission(req, res, callback) {
...
         mode: res.locals.authz_data.mode,
+        assessment_instance_id: res.locals.assessment_instance.id,
+        assessment_id: res.locals.assessment.id,
+        instance_question_id: res.locals.instance_question.id,
     };
     sqldb.callOneRow('variants_ensure_instance_question', [submission.variant_id, res.locals.instance_question.id], (err, result) => {
         if (ERR(err, callback)) return;
         const variant = result.rows[0];
+        // There is going to be a situation where the file is needed for grading (likely!) so we would
+        // have to go inside the save and grade submission logic to let it in but not save it.
```
andrewstec, line 43: In the first callback before a submission is saved, we remove the file so it does not enter the database. We can cache the data in a variable here and put it back in the grader for later.

OR, we can start parsing out the file data in the SQL. Any preference? It seems like a much easier problem to deal with at this scope. 

We also, likely, have to submit this data for External Grading. I have to start analyzing that logic to understand our options.

migrations/215_files__add_storage_type.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE files 
+    ADD COLUMN storage_type text DEFAULT 'fileSystem',
```
tdy, line 2:
```suggestion
    ADD COLUMN storage_type enum_file_storage_type DEFAULT 'fileSystem',
```
Would this make sense as a formal enum? (defined in `database/enums/enum_file_storage_type.pg`)

andrewstec: I think that is way better too!  I'll do that.

nwalters512:
@andrewstec do you think you can get this PR up-to-date with master and take it out of draft?

nwalters512:
This PR is very out of date, and we're likely to go in a different direction when we ultimately work to get large files out of the database.

PR #3066, opened 2020-09-03
Add local cache for workspace urlRewrite query

tdy:
Caches `questions.workspace_url_rewrite` to address DB load issues.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -155,6 +155,7 @@ config.workspaceLoadLaunchTemplateId = null;
 config.workspaceLoadLaunchTag = 'workspace-host';
 config.workspaceHostUnhealthyTimeoutSec = 12 * 60 * 60;
 config.workspaceHostLaunchtimeoutSec = 10 * 60;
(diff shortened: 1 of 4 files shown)


nicknytko:
Do we ever want to dump this cache? I don’t expect this value to change very often (if at all), but maybe we can flush on course sync just to be safe?

mwest1066:
> Do we ever want to dump this cache? I don’t expect this value to change very often (if at all), but maybe we can flush ~on course sync~ after some interval just to be safe?

I pushed an update to age out old cached values. I actually don't think it's needed in this case because we aren't going to have more than a million workspaces or something, but it's good hygiene to avoid slowly growing memory without limits 😄 

@tdy does the update look ok to you?

tdy:
> @tdy does the update look ok to you?

Yep looks good! I just wasn't sure what happened with the CI.

mwest1066:
> Yep looks good (and tested locally)! I just wasn't sure what happened with the CI.

I fixed the CI now (need to clear the setIntervals).

Issue #2924, opened 2020-08-24
Groupwork permission checkboxes


glherman:
When I check Join Groups, Create Groups, and Leave Groups, the checkboxes don't stay checked when re-opening config. Student Permissions at the top indicates that those settings have been applied, but it would be nice to have the checkboxes reflect the current status of the permissions when opening

glherman:
Actually, it looks like the permissions might be getting lost sometimes. After I switched my effective user mode to student and then returned, my permissions got lost.

glherman:
Yeah, already accidentally deleted my permissions by not re-checking the boxes.

frankyhao:
I'll edit this part later, so the checkboxes will be shown according to the current status. Instructors will not need to re-checking the boxes everytime.

PR #3067, opened 2020-09-04
[Group Work] Group Config - Student Permission Checkboxes

frankyhao:
Fix [Issue#2924](https://github.com/PrairieLearn/PrairieLearn/issues/2924) "When I check Join Groups, Create Groups, and Leave Groups, the checkboxes don't stay checked when re-opening config. Student Permissions at the top indicates that those settings have been applied, but it would be nice to have the checkboxes reflect the current status of the permissions when opening" @glherman
- [x] set input type for group size to be number
- [x] display checked checkboxes according to the current student permissions (instructors do not need to recheck them again)
- [x] delete redundant display
image

### DIFF
--- a/pages/instructorAssessmentGroups/instructorAssessmentGroups.ejs
+++ b/pages/instructorAssessmentGroups/instructorAssessmentGroups.ejs
@@ -213,7 +213,7 @@
                   <br>
                   <label for="formman">Student Permissions</label> <br>
                   <div class="form-check form-check-inline">
-                    <input class="form-check-input" type="checkbox" id = 'joincheck' name="joincheck" value="true">
+                    <input class="form-check-input" type="checkbox" name="joincheck" value="<%- config_info.student_authz_join ? 'true' : 'false' %>">
                     <label class="form-check-label" for="inlineCheckbox1">Join Groups</label>
                   </div>
                   <div class="form-check form-check-inline">
--- a/pages/instructorAssessmentGroups/instructorAssessmentGroups.ejs
+++ b/pages/instructorAssessmentGroups/instructorAssessmentGroups.ejs
@@ -424,16 +424,6 @@
         <% } %>
       </div>
       <script>
-        if (<%- config_info.student_authz_join %>) {
-          document.getElementById("joincheck").checked = true;
-        }
-        if (<%- config_info.student_authz_create %>) {
-          document.getElementById("createcheck").checked = true;
-        }
-
-        if (<%- config_info.student_authz_leave %>) {
-          document.getElementById("leavecheck").checked = true;
-        }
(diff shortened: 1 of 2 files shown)


pages/instructorAssessmentGroups/instructorAssessmentGroups.ejs
```diff
@@ -201,29 +201,27 @@
...
                   <div class="form-max">
                     <label for="formman">Maximum group size</label>
-                    <input type="text" value=<%= config_info.defaultMax %> class="form-control" id="formmax" name="maxsize">
+                    <input type="number" value=<%= config_info.defaultMax %> class="form-control" id="formmax" name="maxsize">
                   </div>
+                  <br>
                   <label for="formman">Student Permissions</label> <br>
                   <div class="form-check form-check-inline">
-                    <input class="form-check-input" type="checkbox" name="joincheck" value="true">
+                    <input class="form-check-input" type="checkbox" id = 'joincheck' name="joincheck" value="true">
```
mwest1066, line 216: The new `id` attribute is not valid HTML. You have to use `"` quotes, not `'`. Also there should not be spaces around the `=`. However, I think we actually don't need this `id` at all (see below).

The right way to fix this is to set the `value` to be correct in the EJS template, based on the current database information.
```suggestion
                    <input class="form-check-input" type="checkbox" name="joincheck" value="<%- config_info.student_authz_join ? 'true' : 'false' %>">
```

pages/instructorAssessmentGroups/instructorAssessmentGroups.ejs
```diff
@@ -426,6 +424,16 @@
...
+        if (<%- config_info.student_authz_join %>) {
+          document.getElementById("joincheck").checked = true;
+        }
+        if (<%- config_info.student_authz_create %>) {
+          document.getElementById("createcheck").checked = true;
+        }
+
+        if (<%- config_info.student_authz_leave %>) {
+          document.getElementById("leavecheck").checked = true;
+        }
```
mwest1066, line 436: All unnecessary, I think.

mwest1066:
@frankyhao this PR was really easy to review! 🚀 Thanks!

pages/instructorAssessmentGroups/instructorAssessmentGroups.ejs
frankyhao: for the checkboxes, the value is actually not useful. I replaced the value with checked:
<%= config_info.student_authz_join ? "checked" : "" %>

mwest1066: Oh yes, you are correct.

PR #3069, opened 2020-09-05
Fix double header send on workspace proxy error

nicknytko:


server.js
```diff
@@ -215,11 +215,11 @@ const workspaceProxyOptions = {
     },
     onError: (err, req, res) => {
         logger.error(`Error proxying workspace request: ${err}`);
-        try {
+        /* Check to make sure we weren't already in the middle of sending a response
+           before replying with an error 500 */
+        if (!res.headersSent) {
             res.writeHead(500, { 'Content-Type': 'text/plain' });
             res.send('There was an error proxying this workspace request');
```
mwest1066, line 222:
```suggestion
            res.status(500).send('Error proxying workspace request');
```

PR #3071, opened 2020-09-05
Add workspace fallback function to get graded files from S3

tdy:
* [x] refactor S3 downloader to `aws.downloadFromS3()` (used by `getGradedFilesFromS3()`)
    * a bit unwieldy since `update_queue` and `chown`ing are workspace specific, but it felt ugly to dupe another download function
* [x] refactor workspace chown'er to `workspace.changeWorkspaceFileOwner()`
* [x] implement `workspace.getGradedFilesFromS3()`
* [x] clean out intermediate zip dir after zipping
    * ideally bypass zip dir and stream them straight from s3 to zip file if possible (but not sure how)
* [x] ~~move chowning to `lib/aws`~~
* [x] remove lstats
* [x] bundle version+graded_files queries
* [x] remove chowning helper
* [x] parameterize `owner` and `group` in `downloadFromS3()`
* [x] factor out watcher code from `downloadFromS3()`
* [x] parallelize zip jobs
* [x] parallelize unzip jobs
* [x] fix unzip pipe closing before `finish`

---

cc @mfsilva22


lib/workspace.js
```diff
@@ -279,15 +279,65 @@ module.exports = {
     },
 
     async getGradedFilesFromS3(workspace_id) {
+        debug('Getting graded files from S3');
+        let result = await sqldb.queryOneRowAsync(sql.select_workspace_graded_files, { workspace_id });
+        const { workspace_graded_files } = result.rows[0];
+        result = await sqldb.queryOneRowAsync(sql.select_workspace_version, { workspace_id });
+        const { workspace_version } = result.rows[0];
```
mwest1066, line 286: Repeated querying of the DB is very expensive. Can we bundle these queries?

lib/aws.js
```diff
@@ -133,6 +134,65 @@ module.exports.uploadDirectoryToS3Async = async function(s3Bucket, s3Path, local
...
+    return new Promise((resolve, reject) => {
+        s3Stream.on('error', function(err) {
+            debug(`downloadFromS3: missing s3://${s3Bucket}/${s3Path}`);
+            reject(err);
+        });
+        s3Stream.pipe(fileStream).on('error', function(err) {
+            debug(`downloadFromS3: connection lost`);
+            reject(err);
+        }).on('close', function() {
+            if (watcherUpdateQueue != null) watcherUpdateQueue[[localPath, false]] = {action: 'skip'};
```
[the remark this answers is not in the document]
tdy: You mean by delaying it to the callback? I was thinking that would cause a race condition where chokidar might have already added it to the `update_queue` before we've added a `skip` action here. Does that make sense, or would it be fine to just add this `skip` action in the callback?

lib/workspace.js
```diff
@@ -279,15 +279,70 @@ module.exports = {
...
+        const zipName = `workspace-${workspace_id}-${workspace_version}-${timestamp}`;
+        const zipDir = path.join(config.workspaceMainZipsDirectory, zipName);
+        const zipPath = `${zipDir}.zip`;
 
-        // we should get the gradedFiles from S3 and zip them into zipPath
+        // download graded files from S3 -> zip dir -> zip file
+        // FIXME: stream straight from S3 -> zip file
+        const archive = archiver('zip');
+        const s3Name = `workspace-${workspace_id}-${workspace_version}`;
+        for (const file of workspace_graded_files) {
```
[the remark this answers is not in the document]
tdy: Parallelized with `config.workspaceJobsParallelLimit = 50;` (wasn't sure on the value here).

Also, when I added a bunch of initial/graded files for testing, I ran into some bug where the host wasn't consistently unzipping all initial files. I went ahead and parallelized the unzip jobs as well, and that seems to have fixed it (?) but I never did figure out what the actual original bug was.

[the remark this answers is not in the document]
mwest1066: Weird!

lib/aws.js
```diff
@@ -133,6 +133,54 @@ module.exports.uploadDirectoryToS3Async = async function(s3Bucket, s3Path, local
...
+ * @param {string} s3Bucket - The S3 bucket name.
+ * @param {string} s3Path - The S3 source path.
+ * @param {string} localPath - The local target path.
+ * @param {object?} options - Optional parameters, including owner and group (optional, defaults to {}).
+ */
+module.exports.downloadFromS3Async = async function (s3Bucket, s3Path, localPath, options={}) {
+    if (localPath.endsWith('/')) {
+        debug(`downloadFromS3: bypassing S3 and creating directory localPath=${localPath}`);
+        await fs.promises.mkdir(localPath, { recursive: true });
+        if (!isNaN(options.owner) && !isNaN(options.group)) await fs.promises.chown(localPath, options.owner, options.group);
```
mwest1066, line 148:
```suggestion
        if (options.owner != null && options.group != null) await fs.promises.chown(localPath, options.owner, options.group);
```

lib/aws.js
```diff
@@ -133,6 +133,54 @@ module.exports.uploadDirectoryToS3Async = async function(s3Bucket, s3Path, local
...
+        return;
+    }
+
+    debug(`downloadFromS3: creating containing directory for file localPath=${localPath}`);
+    await fs.promises.mkdir(path.dirname(localPath), { recursive: true });
+    if (!isNaN(options.owner) && !isNaN(options.group)) await fs.promises.chown(path.dirname(localPath), options.owner, options.group);
```
mwest1066, line 154: Same `isNaN()` change. Do we need to worry about chowning intermediate paths? For example, if we make `a/b/c/d`, do we just need to chown `d` or should we also chown `c`, `b`, etc?

lib/aws.js
```diff
@@ -133,6 +133,54 @@ module.exports.uploadDirectoryToS3Async = async function(s3Bucket, s3Path, local
...
+    return new Promise((resolve, reject) => {
+        s3Stream.on('error', function(err) {
+            debug(`downloadFromS3: missing s3://${s3Bucket}/${s3Path}`);
+            reject(err);
+        });
+        s3Stream.pipe(fileStream).on('error', function(err) {
+            debug(`downloadFromS3: connection lost`);
+            reject(err);
+        }).on('close', function() {
+            if (!isNaN(options.owner) && !isNaN(options.group)) {
```
mwest1066, line 173: `isNaN()`

lib/config.js
```diff
@@ -133,6 +133,7 @@ config.workspaceDevContainerHostname = 'host.docker.internal';
 config.workspaceJobsDirectory = '/jobs/workspaces';
 config.workspaceJobsDirectoryOwnerUid = 0;
 config.workspaceJobsDirectoryOwnerGid = 0;
+config.workspaceJobsParallelLimit = 50;
```
mwest1066, line 136:
```suggestion
config.workspaceJobsParallelLimit = 5;
```

lib/workspace.js
```diff
@@ -279,15 +280,58 @@ module.exports = {
...
+        const zipName = `workspace-${workspace_id}-${workspace_version}-${timestamp}`;
+        const zipDir = path.join(config.workspaceMainZipsDirectory, zipName);
+        const zipPath = `${zipDir}.zip`;
 
-        throw new Error('getGradedFilesFromS3() is not yet implemented');
+        // download graded files from S3 -> zip dir -> zip file
+        // FIXME: stream straight from S3 -> zip file
+        const archive = archiver('zip');
+        const s3Name = `workspace-${workspace_id}-${workspace_version}`;
+        const zipJobs = workspace_graded_files.map(file => (
```
mwest1066, line 295: You can just use https://caolan.github.io/async/v3/docs.html#mapLimit directly. Also the `async` lib can directly handle `async` functions (see https://caolan.github.io/async/v3/global.html) so there is no need to use `promisify()`.

lib/workspace.js
```diff
@@ -279,15 +280,58 @@ module.exports = {
...
+                    const options = {
+                        owner: config.workspaceJobsDirectoryOwnerUid,
+                        group: config.workspaceJobsDirectoryOwnerGid,
+                    };
+                    await awsHelper.downloadFromS3Async(config.workspaceS3Bucket, s3Path, localPath, options);
+                    await fsPromises.lstat(localPath);
+                    debug(`Zipping graded file ${localPath} into ${zipPath}`);
+                    archive.file(localPath, { name: file });
+                } catch (err) {
+                    logger.warn(`Graded file ${file} does not exist`);
```
mwest1066, line 309: This is a user error so let's not warn. Maybe a `debug()`?

workspace_host/interface.js
```diff
@@ -54,6 +54,9 @@ if ('config' in argv) {
 }
 config.loadConfig(configFilename);
 
+const workspaceUid = config.workspaceJobsDirectoryOwnerUid;
+const workspaceGid = config.workspaceJobsDirectoryOwnerGid;
```
mwest1066, line 58: This will cause this bug again: https://github.com/PrairieLearn/PrairieLearn/pull/2887

lib/aws.js
tdy: It turns out nested dirs will currently break the workspace launch (usually). Nested dirs used to work at some point, so I'm not sure when exactly this regressed.

```
info: Launching workspace-1-1 (useInitialZip=true)
  prairielearn:interface Bootstrapping workspace with initial.zip +0ms
  prairielearn:workspace Setting workspaces.message to 'Loading initial files' +0ms
  prairielearn:interface Downloading s3Path=workspace-1-1/initial.zip to zipPath=/workspace_host_zips/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba-initial.zip +2ms
  prairielearn:aws downloadFromS3: creating containing directory for file localPath=/workspace_host_zips/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba-initial.zip +0ms
  prairielearn:interface Making directory /jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba +50ms
  prairielearn:interface Preparing unzip jobs +13ms
info: Watching directory add /jobs/workspaces
  prairielearn:interface Unzipping /workspace_host_zips/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba-initial.zip to /jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba +20ms
  prairielearn:interface Making directory /jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba/a/ +1ms
  prairielearn:interface Making directory /jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba/a/b/ +1ms
  prairielearn:interface Making directory /jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba/a/b/c/ +0ms
  prairielearn:interface Extracting file /jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba/a/b/c/data.csv +1ms
  prairielearn:interface Chowning entryPath=/jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba/a/b/c/data.csv +1ms
  prairielearn:interface Extracting file /jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba/fibonacci.py +1ms
  prairielearn:interface Chowning entryPath=/jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba/fibonacci.py +1ms
  prairielearn:interface Chowning entryPath=/jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba/a/ +2ms
  prairielearn:interface Chowning entryPath=/jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba/a/b/ +3ms
info: Watching directory add /jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba
  prairielearn:workspace Setting workspaces.state='stopped', workspaces.message='Error! Click "Reboot" to try again. Detail: Error: ENOENT: no such file or directory, chown '/jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba/a/b/c/data.csv'' +101ms
events.js:292
      throw er; // Unhandled 'error' event
      ^

Error: ENOENT: no such file or directory, open '/jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba/a/b/c/data.csv'
Emitted 'error' event on WriteStream instance at:
    at errorOrDestroy (internal/streams/destroy.js:108:12)
    at WriteStream.onerror (_stream_readable.js:752:7)
    at WriteStream.emit (events.js:315:20)
    at WriteStream.EventEmitter.emit (domain.js:483:12)
    at /PrairieLearn/node_modules/graceful-fs/graceful-fs.js:211:14
    at /PrairieLearn/node_modules/graceful-fs/graceful-fs.js:241:16
    at FSReqCallback.oncomplete (fs.js:156:23) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba/a/b/c/data.csv'
}
```
```
[root@81dbee9b9013 PrairieLearn]# find /jobs/workspaces/ -print
/jobs/workspaces/
/jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba
/jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba/a
/jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba/a/b
/jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba/a/b/c
/jobs/workspaces/workspace-9220926b-6e02-45c7-9d40-bfbc68f451ba/fibonacci.py
```

tdy: Actually I'm not sure it's related to nested dirs. It seems to be an unzipping issue overall.

tdy: It seems to have something to do with the parallelized unzip jobs (for `initial.zip`), so I reverted that change. But with the original unzip code, I'm still getting inconsistent unzipping of `initial.zip`.

mwest1066: I'm happy to go back to serial zip/unzip. It's only a small speedup anyway and it's possible that the `unzipper` library doesn't play well with parallel unzip.

mwest1066: I haven't investigated in detail, but in the [Quick Examples](https://www.npmjs.com/package/unzipper#quick-examples) section of the `unzipper` docs they say "Extract emits the 'close' event once the zip's contents have been fully extracted to disk." We don't currently wait for this, so I suspect we are returning before the individual file unzips have actually occurred.

mwest1066: I think it might be as simple as doing:
```
await entry.pipe(fs.createWriteStream(entryPath)).promise();
```

tdy:
```diff
-             entry.pipe(fs.createWriteStream(entryPath));
+             await entry.pipe(fs.createWriteStream(entryPath)).promise();
```

^ that still didn't work on my end, but this `stream.Transform` pipe is working consistently so far on my end:

https://www.npmjs.com/package/unzipper#parse-zip-by-piping-entries-downstream

Not sure if there's a cleaner way, but I pushed this piping/transform version for now.

tdy: I believe this comment from the `unzipper` author confirms that the basic async iterator won't work for our use case (parallel or not). From https://github.com/ZJONSSON/node-unzipper/issues/138#issuecomment-506376795:

> The problem is that the pipe chain finishes when the zip file has been read, not when the output files have been written. For small files this is roughly the same thing, but for large files there can be a some lag.

He suggests:

> If you are using unzipper.Parse() one way to tackle this is to use .pipe into a transform stream that only performs callback when each file has been written. You can see some examples in the readme here: https://www.npmjs.com/package/unzipper#parse-zip-by-piping-entries-downstream

So this code is working on my end and pushed upstream. I don't know how to get rid of the ugly callback+async mixing here.

https://github.com/PrairieLearn/PrairieLearn/blob/2a700d7c4a61ff51c9367077f79c76eada98b625/workspace_host/interface.js#L795-L817

lib/workspace.js
tdy: Combined into a single `eachLimit()`

lib/aws.js
tdy: Reverted and moved to #3146

mwest1066 (approved):

PR #3072, opened 2020-09-05
[Group Work] display empty groups on instances and instance detail page

### LINKED ISSUE #2990: Groups with 0 members disappear
Groups with 0 members disappear, which is reasonable but is problematic because if a student accidentally leaves a group of 1, they will lose all of their work.

Steps to reproduce.
1) Create a new student group
2) Submit anything in any problem
3) Leave the group
4) Look at list of students/teams and the group will no longer be there.

frankyhao:
Fixes #2990
Some students created a group, started an instance, and left the group. These "empty groups" are displayed on the Instructor Group page but not the Instructor Student page (instructorAssessmentInstances.ejs)

- [x] Modified the select rule to LEFT JOIN groups with no users -> instructor could now check the instance details for empty groups

### DIFF
--- a/pages/instructorAssessmentInstance/instructorAssessmentInstance.js
+++ b/pages/instructorAssessmentInstance/instructorAssessmentInstance.js
@@ -12,7 +12,7 @@ const sql = sqlLoader.loadSqlEquiv(__filename);
 
 const logCsvFilename = (locals) => {
     return sanitizeName.assessmentFilenamePrefix(locals.assessment, locals.assessment_set, locals.course_instance, locals.course)
-        + sanitizeName.sanitizeString(locals.assessment.group_work ? locals.group_info.name : locals.instance_user.uid)
+        + sanitizeName.sanitizeString(locals.assessment.group_work ? locals.group.name : locals.instance_user.uid)
         + '_'
         + locals.assessment_instance.number
         + '_'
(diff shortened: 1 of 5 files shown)


pages/instructorAssessmentInstance/instructorAssessmentInstance.js
```diff
@@ -12,7 +12,7 @@ const sql = sqlLoader.loadSqlEquiv(__filename);
 
 const logCsvFilename = (locals) => {
     return sanitizeName.assessmentFilenamePrefix(locals.assessment, locals.assessment_set, locals.course_instance, locals.course)
-        + sanitizeName.sanitizeString(locals.instance_user.uid)
+        + ((locals.assessment.group_work) ? (locals.group_info.name) : (sanitizeName.sanitizeString(locals.instance_user.uid)))
```
mwest1066, line 15: Should we sanitize the group name as well? At the moment I think it won't matter, but in the future we might allow extra characters and then this would be a good idea?
```suggestion
        + sanitizeName.sanitizeString(locals.assessment.group_work ? locals.group_info.name : locals.instance_user.uid)
```

mwest1066 (commented):
Very nice PR to review! 🎉 Thanks for keeping it short and clean.

PR #3083, opened 2020-09-08
add script verify_course_repo.js

1733116199:
There is currently no built-in script aiming to verify course repositories on an CI machine. This commit adds a new script that can be execute directly in docker with the following command:
docker run -it -v $(pwd):/course prairielearn/prairielearn node /PrairieLearn/tools/verify_course_repo.js
The script prints nothing if the course repo is valid, and prints errors and returns with an exit code of 1 if there are any errors.

To setup CI for a course repo:
1. Make install docker is installed on the CI machine.
2. Execute command:
docker run -it -v $(pwd):/course prairielearn/prairielearn node /PrairieLearn/tools/verify_course_repo.js

### DIFF
--- a/tools/verify_course_repo.js
+++ b/tools/verify_course_repo.js
@@ -1,0 +1,10 @@
+const courseDB = require('../sync/course-db');
+(async () => {
+    const courseData = await courseDB.loadFullCourseNew("/course");
+    const errors = [];
+    courseDB.writeErrorsAndWarningsForCourseData("fakeCourseId", courseData, line => line ? errors.push(line) : null);
+    if(errors.length !== 0){
+        errors.forEach(line => console.error(line));
+        process.exit(1);
+    }
+})();
--- a/tools/verify_course_repo.js
+++ b/tools/verify_course_repo.js
@@ -1,8 +1,8 @@
 const courseDB = require('../sync/course-db');
 (async () => {
-    const courseData = await courseDB.loadFullCourseNew("/course");
+    const courseData = await courseDB.loadFullCourseNew('/course');
     const errors = [];
-    courseDB.writeErrorsAndWarningsForCourseData("fakeCourseId", courseData, line => line ? errors.push(line) : null);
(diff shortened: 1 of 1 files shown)


tools/verify_course_repo.js
```diff
@@ -0,0 +1,10 @@
+const courseDB = require('../sync/course-db');
+(async () => {
+    const courseData = await courseDB.loadFullCourseNew('/course');
+    const errors = [];
+    courseDB.writeErrorsAndWarningsForCourseData('fakeCourseId', courseData, line => line ? errors.push(line) : null);
+    if(errors.length !== 0){
```
nwalters512, line 6: Nit: formatting
```suggestion
    if (errors.length !== 0) {
```

tools/verify_course_repo.js
```diff
@@ -0,0 +1,10 @@
+const courseDB = require('../sync/course-db');
+(async () => {
+    const courseData = await courseDB.loadFullCourseNew('/course');
+    const errors = [];
+    courseDB.writeErrorsAndWarningsForCourseData('fakeCourseId', courseData, line => line ? errors.push(line) : null);
```
nwalters512, line 5: Nit: the course directory argument is currently unused and should probably be removed from the function. In the meantime, let's just pass null.

```suggestion
    courseDB.writeErrorsAndWarningsForCourseData(null, courseData, line => line ? errors.push(line) : null);
```

Aside: is there a case when `line` is null? I looked at the code again and couldn't see why this check would be necessary.

nwalters512 (approved):
This is awesome! Been thinking about something like this for a while and glad someone finally took the initiative to write it. Also pleasantly surprised to see how easy the new syncing code makes this script.

Approving, but would appreciate reviewing my comments before merging.

mwest1066:
@nwalters512 I went ahead and applied your changes. I'm inclined to merge this as is. Thoughts?

nwalters512:
@mwest1066 go ahead, my approval stands!

PR #3091, opened 2020-09-08
File store s3 extension

### LINKED ISSUE #3090: Default file storage upload should become S3
Files uploaded through `attachFilePanel` are currently uploaded to the filesystem on AWS. The default should become S3 due to less operations complexity. All previously uploaded file system uploads should still be supported.

andrewstec:
This PR introduces an S3 upload option to the file-store API.

Closes #3090


andrewstec:
I think I will be taking a look at this again on Friday. I seem to have a good hypothesis that the problem is an S3 mock integration failure, but I do not know that yet. Any basic pointers about the S3 mocking in the test environment would be helpful.

andrewstec:
https://github.com/PrairieLearn/PrairieLearn/pull/3114 is a dependency for this PR. Do *NOT* merge this PR before https://github.com/PrairieLearn/PrairieLearn/pull/3114.

andrewstec:
Just a note, there was a lot of debugging while I was figuring out why some tests were failing. I would be happy to squash this in with some cleaner commits.

pages/studentAssessmentInstanceFile/studentAssessmentInstanceFile.js
```diff
@@ -1,28 +1,17 @@
-const ERR = require('async-stacktrace');
+// const ERR = require('async-stacktrace');
```
nicknytko, line 1:
```suggestion
```

lib/file-store.js
```diff
@@ -43,7 +42,40 @@ module.exports.upload = async (display_filename, contents, type, assessment_inst
...
+        // and put it in two levels of directories corresponding to the first-3
+        // and second-3 characters of the filename.
+
+        if (config.filesRoot == null) throw new Error('config.filesRoot is null, which does not allow uploads');
+
+        debug(`upload() : mkdir ${dir}`);
+        await fsPromises.mkdir(dir, {recursive: true, mode: 0o700});
+        debug(`upload(): writeFile ${filename}`);
+        await fsPromises.writeFile(filename, buffer, {mode: 0o600});
+    }
```
nicknytko, line 61: Is this for the "attach a file" dialog?  I'm actually thinking we can remove local EC2 file storage as that will just cause problems when we're running on multiple servers.  On local dev, we can probably just use s3rver as well. (Thoughts, @mwest1066 ?)

database/enums/enum_file_storage.pg
```diff
@@ -0,0 +1 @@
+FileSystem, S3
```
nicknytko, line 1: add a newline?

nicknytko:
> Just a note, there was a lot of debugging while I was figuring out why some tests were failing. I would be happy to squash this in with some cleaner commits.

No worries, default behavior is to squash all commits on merge.

lib/file-store.js
andrewstec: @nicknytko "Attach a file" dialog is right. Keep in mind that the instructor file editor also goes through this path, so the files get uploaded to S3. That may be something you would prefer the FileSystem for. In my tests, S3 was quite fast though.

For something like the instructor file editor, are we happy with the history of the files being left on S3? I cannot help but notice the soft deletes.

Whichever approach you want to take should work. They both work in the manual and CI tests.

andrewstec: The get() function will need to continue retrieving FileSystem files that already exist.

mwest1066: I think sending everything to S3 is fine. Keeping a history in S3 is ok (we can always clean these up later if necessary).

@nicknytko I agree that we want to switch everything to S3, but we have a large set of files that need to be migrated from the existing disk store. My thinking was to get this PR merged and default to putting new files in S3, then at some point we can migrate the existing files off the disk.

lib/file-store.js
```diff
@@ -43,7 +42,39 @@ module.exports.upload = async (display_filename, contents, type, assessment_inst
...
+
+    // uploads to s3 bucket
+    if (storage_type === this.storageTypes.S3) {
```
mwest1066, line 64: Can we make this an `else if` from the previous `if (storage_type === this.storageTypes.FileSystem)` and add a final `else` clause that throws an exception if we have an unknown storage type?

lib/file-store.js
```diff
@@ -43,7 +42,39 @@ module.exports.upload = async (display_filename, contents, type, assessment_inst
...
+        debug('upload() : setting s3 params');
+
+        if (config.fileUploadS3Bucket == null) throw new Error('config.fileUploadS3Bucket is null, which does not allow uploads');
+
+        const options = {
+            bucket: config.fileUploadS3Bucket,
+            key: filename.replace(config.filesRoot + '/', ''),
+            body: buffer,
+        };
+        const res = await uploadToS3Async(options.bucket, options.key, null, null, options.body);
```
mwest1066, line 74: I'm not quite clear why we have an `options` object here. Would it perhaps be simpler to just call the function with the arguments directly? Like this:
```suggestion
        const res = await uploadToS3Async(config.fileUploadS3Bucket, storage_filename, null, null, buffer);
```

tests/testHomework.sql
```diff
@@ -38,3 +38,21 @@ SET
...
+-- BLOCK insert_file_fs_ai
+INSERT INTO files (user_id, assessment_id, assessment_instance_id, created_at, created_by, display_filename, storage_filename, type, storage_type)
+VALUES(1, $assessment_id, $assessment_instance_id, current_timestamp, 1, $filename, $filepath, 'student_upload', 'FileSystem');
+
+
+-- BLOCK delete_test_file_fs
+DELETE FROM files 
+WHERE $filename = display_filename;
+
+-- BLOCK select_first_ai
```
mwest1066, line 55: This block doesn't seem to be used anymore?

lib/file-store.js
```diff
@@ -17,24 +18,22 @@ const sql = sqlLoader.loadSqlEquiv(__filename);
...
-    debug('upload()');
-    // Make a filename to store the file. We use a UUIDv4 as the filename,
-    // and put it in two levels of directories corresponding to the first-3
-    // and second-3 characters of the filename.
+
+module.exports.storageTypes = Object.freeze({S3: 'S3', FileSystem: 'FileSystem'});
+
+module.exports.upload = async (display_filename, buffer, type, assessment_instance_id, instance_question_id, user_id, authn_user_id, storage_type) => {
+    // default S3
+    if (!storage_type) storage_type = this.storageTypes.S3;
```
mwest1066, line 29: Could we add a config variable for this default? Maybe `config.fileStoreStorageDefault = 'S3'`?

lib/file-store.js
```diff
@@ -17,24 +18,22 @@ const sql = sqlLoader.loadSqlEquiv(__filename);
  * @param {number|null} instance_question_id - The instance question for the file.
  * @param {number} user_id - The current user performing the update.
  * @param {number} authn_user_id - The current authenticated user.
+ * @param {string} storage_type - AWS 's3' or 'filesystem' storage options.
```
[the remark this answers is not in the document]
andrewstec: It should have been changed to match a convention I later saw. Thanks for catching this.

lib/file-store.js
```diff
@@ -17,24 +18,22 @@ const sql = sqlLoader.loadSqlEquiv(__filename);
...
  */
-module.exports.upload = async (display_filename, contents, type, assessment_instance_id, instance_question_id, user_id, authn_user_id) => {
-    debug('upload()');
-    // Make a filename to store the file. We use a UUIDv4 as the filename,
-    // and put it in two levels of directories corresponding to the first-3
-    // and second-3 characters of the filename.
+
+module.exports.storageTypes = Object.freeze({S3: 'S3', FileSystem: 'FileSystem'});
+
+module.exports.upload = async (display_filename, buffer, type, assessment_instance_id, instance_question_id, user_id, authn_user_id, storage_type) => {
```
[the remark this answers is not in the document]
andrewstec: Ah, indeed it did. Fixed: d623b84d1.

lib/file-store.js
andrewstec: Most certainly.

lib/file-store.js
andrewstec: Old linting habit for shorter lines. Removed.

lib/file-store.js
andrewstec: Added here: dfc65ae9b6d5bf843d6aa0d093620ff675836e7a.

lib/workspace.js
```diff
@@ -211,7 +198,7 @@ module.exports = {
         const s3Path = `${s3Name}/current`;
 
         const now = new Date().toISOString().replace(/[-T:.]/g, '-');
-        const zipName = `${s3Name}-${now}.zip`;
+        const zipName = `workspace-${workspace_id}-${now}.zip`;
```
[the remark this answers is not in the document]
andrewstec: Done!

tests/testHomework.js
```diff
@@ -327,7 +329,32 @@ describe('Homework assessment', function() {
...
     });
 
-    describe('6. instance_question: attach files setup', function() {
+    describe('6. assessment_instance: download deprecated text file from filesystem ', function() {
+        const filepath = 'testpath/';
+        const filename = 'test.zyx';
+
+        before('should insert deprecated file', async function() {
+            const content = 'This is the test text';
+            fs.mkdirSync(path.join(config.filesRoot, filepath));
```
tdy, line 338: Are all these `fs` calls intentionally *Sync?

tests/testHomework.js
```diff
@@ -342,29 +369,53 @@ describe('Homework assessment', function() {
...
     });
 
-    describe('11. submit correct answer to question addVectors', function() {
+    describe('12. instance_question: download deprecated filesystem', function() {
+        const filepath = 'testpath/';
+        const filename = 'test.zyx';
+
+        before('should insert deprecated file', async function() {
+            const content = 'This is the test text';
+            fs.mkdirSync(path.join(config.filesRoot, filepath));
```
tdy, line 400: Same here, intentionally *Sync?

lib/workspace.js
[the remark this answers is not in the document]
tdy: Maybe the merge hasn't been pushed? I'm still seeing this `zipName` line change as of the latest upstream commit 5dbd4734ba01adcbae6f73bdf4c1bca8ed49a7c8.

tests/testHomework.js
mwest1066:
```suggestion
            await fs.mkdir(path.join(config.filesRoot, filepath));
```

andrewstec: They were, but do not have to be at all. I will change that now.

andrewstec: image

I didn't know this, but we can return a promise in a slightly different way:

ie. fs.promises.mkdir('whatever/path')

tests/testHomework.js
andrewstec: Fixed here: 54d589c15e5cd322325a04232bef3f4355c7cad1

tests/testHomework.js
andrewstec: Fixed here: 54d589c15e5cd322325a04232bef3f4355c7cad1

pages/studentAssessmentInstanceFile/studentAssessmentInstanceFile.js
```diff
@@ -1,28 +1,16 @@
...
 
-const config = require('../../lib/config');
-const sqldb = require('@prairielearn/prairielib/sql-db');
-const sqlLoader = require('@prairielearn/prairielib/sql-loader');
-
-const sql = sqlLoader.loadSqlEquiv(__filename);
-
-router.get('/:file_id/:display_filename', function(req, res, next) {
-    let params = {
+router.get('/:file_id/:display_filename', async function(req, res, next) {
```
mwest1066, line 5: Express route handlers cannot be `async` functions. Instead we need to do this: https://github.com/PrairieLearn/PrairieLearn/blob/e78a6d01b39d88602ab7ea76088f5ecb86b6f21a/pages/workspace/workspace.js#L30

I'll fix this one.

lib/workspace.js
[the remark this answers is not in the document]
mwest1066: I fixed this.

nicknytko (approved):
Looks good, I was trying this out with the local S3 mocking. Next step would be to try this out on pl-dev with actual buckets?

lib/aws.js
```diff
@@ -99,19 +99,22 @@ module.exports.getS3RVERConfiguration = function() {
...
     let body;
     if (isDirectory) {
         body = '';
         s3Path += s3Path.endsWith('/') ? '' : '/';
     } else {
-        try {
+        if (localPath) {
             body = await fs.promises.readFile(localPath);
-        } catch(err) {
-            return [localPath, s3Path, err];
```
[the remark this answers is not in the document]
mwest1066: No kidding! 😄

mwest1066:
> Looks good, I was trying this out with the local S3 mocking. Next step would be to try this out on pl-dev with actual buckets?

Yep. I actually went ahead and merged this (perhaps unwisely). I'll try it out on pl-dev.

mwest1066:
> > Looks good, I was trying this out with the local S3 mocking. Next step would be to try this out on pl-dev with actual buckets?
> 
> Yep. I actually went ahead and merged this (perhaps unwisely). I'll try it out on pl-dev.

Now I'm just confusing myself -- I did NOT merge this yet 😄 

Testing on pl-dev is indeed the next step. If you want to do it then go right ahead, otherwise I'll give it a spin at some point.

nicknytko:
> > > Looks good, I was trying this out with the local S3 mocking. Next step would be to try this out on pl-dev with actual buckets?
> > 
> > 
> > Yep. I actually went ahead and merged this (perhaps unwisely). I'll try it out on pl-dev.
> 
> Now I'm just confusing myself -- I did NOT merge this yet smile
> 
> Testing on pl-dev is indeed the next step. If you want to do it then go right ahead, otherwise I'll give it a spin at some point.

Just checked, working on pl-dev!  Can confirm that uploaded files are getting sent to s3.

mwest1066:
> Just checked, working on pl-dev! Can confirm that uploaded files are getting sent to s3.

Awesome! I hope they are also coming back again?

nicknytko:
> > Just checked, working on pl-dev! Can confirm that uploaded files are getting sent to s3.
> 
> Awesome! I hope they are also coming back again?

:joy: yes!

nicknytko:
How are the file names getting stored?  I'm trying this out on `staging.prairielearn.org` and they seem to be getting replaced with UUID's:

image

andrewstec:
> ~How are the file names getting stored? I'm trying this out on `staging.prairielearn.org` and they seem to be getting replaced with UUID's:~
> 
> ~image~
> 
> Disregard the above, `bedc4d11-4510-4038-87b5-bd2b46e1562f` was the actual filename that I uploaded 🙃

Thanks for doing this QA work, @nicknytko! I hope there are not any surprises. @mwest1066 Thanks for fixing up those tests!

mwest1066 (approved):

PR #3096, opened 2020-09-10
Fix broken assessment timer

tdy:
`date_limit` was being inserted before it was actually calculated, so it was always null, which meant `assessment_instance_remaining_ms` was also null.

### DIFF
--- a/sprocs/assessment_instances_insert.sql
+++ b/sprocs/assessment_instances_insert.sql
@@ -63,17 +63,6 @@ BEGIN
         -- number = 1 so we will error on INSERT if there
         -- are existing assessment_instances
         
-        -- ######################################################################
-        -- do the actual insert
-
-    INSERT INTO assessment_instances
-            ( auth_user_id, assessment_id, user_id,     group_id, mode, auto_close, date_limit, number)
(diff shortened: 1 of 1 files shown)


mwest1066:
@frankyhao Can you remember why you moved this INSERT query higher in the sproc? It triggered a bug because the INSERT needs `auto_close` and `date_limit`, but the INSERT was moved before these variables were computed.

frankyhao:
This was a very early edit by Zhining. I guess he just mistakenly lift it up to apply "IF groupwork THEN" style, and we did not pay attention to the auto_close variable.

https://github.com/frankyhao/PrairieLearn/commit/59b5eaa874fc00d3011fb3c01cf3a0f2bc045e2b#diff-5d8cf841d8e00cfbba1c6b4c0f1c5931

mwest1066:
> This was a very early edit by Zhining. I guess he just mistakenly lift it up to apply "IF groupwork THEN" style, and we did not pay attention to the auto_close variable.

Ah, no worries, I was just curious whether there was some other issue that I missed. Sounds like this is an easy fix!

Thanks for the fast response!

PR #3114, opened 2020-09-11
Unify AWS config loading

tdy:


lib/aws.js
```diff
@@ -10,15 +10,17 @@ const logger = require('./logger');
 const config = require('./config');
 
 module.exports.init = function(callback) {
-    // AWS will look relative to the Node working directory, not the current
```
tdy, line 13: The `lib/workspace` version used `process.cwd()`, so I ported that here and removed this comment. Should this be switched back to `./aws-config.json`?

lib/aws.js
```diff
@@ -10,15 +10,17 @@ const logger = require('./logger');
...
-    // If it exists, load it. If it does not exist, then we are either (1)
-    // running locally, and so probably don't care, or (2) running on prod,
+    // If we are runningInEc2, then we are running on prod,
     // and so AWS will have already been configured.
-    if (fs.existsSync('./aws-config.json')) {
-        logger.info('Loading AWS credentials for external grading and workspace S3 resources');
-        AWS.config.loadFromPath('./aws-config.json');
+    if (!config.runningInEc2) {
+        const configFile = path.join(process.cwd(), 'aws-config.json');
+        if (fs.existsSync(configFile)) {
```
tdy, line 17: Was this `*Sync` call intentional or just old code? I read that sometimes it's good to load configs using `*Sync`? The `lib/workspace` version used async.

mwest1066 (approved):
One other tiny point: there is an unnecessary extra load of `aws.js` here: https://github.com/PrairieLearn/PrairieLearn/blob/43a2df3f52e31e63d72ac3da70ef2f61b37503f1/workspace_host/interface.js#L49
(we have both `aws` and `awsHelper` as the same lib).

lib/aws.js
mwest1066: I think async would be better. We correctly sequence all the initialization code when we call it using `async.series()`, so async will work fine and I think it's cleaner to always use async.

lib/aws.js
mwest1066: I'm honestly not really sure about when it would matter either way. I guess it's fine as is? But I don't have a strong preference.

lib/aws.js
tdy: Ok, `lib/aws.init()` should be async now: d275290

andrewstec:
I do not believe the server initialization is having the intended effect that we are looking for. The AWS configuration is not being configured globally. I had to add `await init()` to each function in `lib/aws.js` before S3RVER would run effectively in the testing environment.

It may be possible that the AWS SDK configuration is not global and the AWS SDK object instantiated in the `server.js` file is not using the same configuration as the AWS SDK object instantiated in `lib/aws.js`.

tdy:
> The AWS configuration is not being configured globally.

I'm not sure what you mean here. Can you clarify the specific symptoms? Is it just that the bucket isn't filling as you expect? Are you getting stack traces? Etc.

Just trying to narrow down what's happening since the workspace code is still successfully using `require('lib/aws')`.

andrewstec:
> > The AWS configuration is not being configured globally.
> 
> I'm not sure what you mean here. Can you clarify the specific symptoms? Is it just that the bucket isn't filling as you expect? Are you getting stack traces? Etc.
> 
> Just trying to narrow down what's happening since the workspace code is still successfully using `require('lib/aws')`.

Totally understandable @tdy. I am just narrowing down on the specifics and do not have more information at this moment. I have had trouble getting an error from the tests unfortunately, but I know that when I init() and run the S3RVER configuration, the tests pass, so I am making maybe a bit too big of an assumption here.

I was wrong about one thing being that this.init() will not work in the aws.js file. I think the tests passed when the init() was in the file-store.js file. Just refining my changes and I will hopefully have more information for you. I maybe have some logging information in #pl-dev shortly. Thanks for checking in so quickly!

andrewstec:
@tdy I merged this branch into my branch and integrated the changes. They work. Thank you a ton! I do not have a stacktrace to elaborate on the prior situation, but I can show you this: 

Tests passing -> https://github.com/PrairieLearn/PrairieLearn/runs/1124161552
Tests failing after taking out init() -> https://github.com/PrairieLearn/PrairieLearn/runs/1124307851

If the init() is not having an application wide effect on the configuration, the server init() might not be serving its purpose. I took a look at the AWS docs: https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/global-config-object.html. It looks like the `getS3RVERConfiguration()` function successfully returns the AWS Singleton Configuration object. But I suspect that we are making multiple new AWS objects each time they are imported into a *.js file and each imported AWS object cannot see the counterpart object's singleton configuration object.

I don't think this is an issue right now in all practical matters. I don't mind using the init(). Did you have a preference?

I would ask that you can merge your code first before I merge mine in. @mwest1066 Just a heads-up that this is a dependency for

lib/aws.js
mwest1066: Excellent!

mwest1066 (approved):

PR #3129, opened 2020-09-13
Add API endpoint for assessment instance event log

CoffeePoweredComputers:


api/v1/endpoints/queries.sql
```diff
@@ -208,3 +208,343 @@ SELECT
     ), '[]'::jsonb) AS item
 FROM
     object_data;
+
+-- BLOCK select_log
```
mwest1066, line 212: This looks like it was duplicated from `instructorAssessmentInstance.sql`. We try to avoid code duplication because it's hard to remember to update the two blocks of code separately, or to fix bugs in two places.

When we have repeated SQL code in PL, one strategy is to put it in a stored procedure in `sprocs/`. Would that work for you? Feel free to ask questions if it's not clear how to do that.

CoffeePoweredComputers: There was one small modification I made which was to include a few more fields in the "data" json field. There is also the potential that this query might change as the research progresses and I find there is more data I can take advantage of or require. I wasn't sure if you would want these changes, current and potential, mixed in with the existing query used in instructorAssessmentInstance.sql. However, if having these changes present across both use cases sounds okay moving it to `sprocs/` would work and be easy enough.

mwest1066: I'd rather avoid the code duplication, and I don't mind a bit more data being available in the instructor log.

CoffeePoweredComputers: After reviewing documentation and existing code I think I have correctly moved the `select_log` query to `sprocs/`.

sprocs/index.js
```diff
@@ -175,6 +175,7 @@ module.exports = {
             'workspace_hosts_recapture_draining.sql',
             'workspace_hosts_drain_extra.sql',
             'workspace_hosts_find_terminable.sql',
+            'select_log.sql',
```
mwest1066, line 178: Can we call this `assessment_instances_select_log`? The PL naming convention for sprocs is to start with the table name they are most closely associated with.

mwest1066: And in `sprocs/index.js` list it with the other `assessment_instances` sprocs.

PR #3131, opened 2020-09-14
Change workspace progress indicator to be rate limited

### LINKED ISSUE #3027: Rate-limit updateMessage for workspace percent progress
To reduce websocket traffic on the PL servers (all going via redis), we should only call `updateMessage()` here:
https://github.com/PrairieLearn/PrairieLearn/blob/e81745aaecb1fe52e1bf4b14f10ad00e1bf016f7/workspace_host/interface.js#L961 if two conditions are met:
* [x] The value of `percent` is different to when we last called `updateMessage()`
* [x] It's at least 1 second (or some other appropriate value) since we last called `updateMessage()`

tdy:
Resolves #3027:

Only update the percent message if these two conditions are met:

* [x] The value of `percent` is different to when we last called `updateMessage()`
* [x] It's at least `config.workspacePercentMessageRateLimitSec` since we last called `updateMessage()`

### DIFF
--- a/workspace_host/interface.js
+++ b/workspace_host/interface.js
@@ -960,7 +960,7 @@ function _pullImage(workspace, callback) {
                     const date = Date.now();
                     const percentDelta = percent - percentCache;
                     const dateDeltaSec = (date - dateCache) / 1000;
-                    if ((percentDelta > 0) && (dateDeltaSec > 1)) {
+                    if ((percentDelta > 0) && (dateDeltaSec >= config.workspacePercentMessageRateLimitSec)) {
                         percentCache = percent;
                         dateCache = date;
                         percentDisplayed = true;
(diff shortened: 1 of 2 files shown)


workspace_host/interface.js
```diff
@@ -925,6 +925,7 @@ function _pullImage(workspace, callback) {
             let current, total = 0, fraction = 0;
             let currentBase, fractionBase;
             let outputCount = 0;
+            let percentCache = 0, dateCache = Date.now();
```
mwest1066, line 928: Maybe initialize these at lower values so we will always send the very first percent that we calculate?
```suggestion
            let percentCache = -1, dateCache = Date.now() - 1e6;
```

tdy: done

PR #3132, opened 2020-09-14
Add config option to disable attach file dialog

nicknytko:


pages/studentInstanceQuestionExam/studentInstanceQuestionExam.ejs
```diff
@@ -75,7 +75,9 @@
             <% } %>
           </p>
 
-          <%- include('../partials/attachFilePanel') %>
+          <% if (show_attached_files) { %>
```
mwest1066, line 78: I think we can directly reference `config.attachedFilesDialogEnabled` here: https://github.com/PrairieLearn/PrairieLearn/blob/a44a7d0c7288e3aaf33b5d5ba438ccef7a954c6d/server.js#L261

mwest1066 (approved):

Issue #3026, opened 2020-09-02
Set workspace "launched_at" when going to state "launching"


mwest1066:
At the moment we transition to `launching` here: https://github.com/PrairieLearn/PrairieLearn/blob/e81745aaecb1fe52e1bf4b14f10ad00e1bf016f7/lib/workspace.js#L180

But we don't set `launched_at` until later, here: https://github.com/PrairieLearn/PrairieLearn/blob/e81745aaecb1fe52e1bf4b14f10ad00e1bf016f7/workspace_host/interface.js#L1130

We are actually setting `launched_at = now()` when the launching phase _finishes_. Instead we should set it when launching _begins_. To do this, we should:

* [x] remove the `update_workspace_launched_at_now` query from `interface.js`
* [x] modify [`sprocs/workspaces_state_update.sql`](https://github.com/PrairieLearn/PrairieLearn/blob/master/sprocs/workspaces_state_update.sql) to set `launching_at = now()` whenever we change to state `launching`
* [x] add two new columns `workspaces.running_at` and `workspaces.stopped_at` and set these to `now()` in `sprocs/workspaces_state_update.sql` when we transition to those states
* [x] add two new columns `workspaces.rebooted_at` and `workspaces.reset_at` and set those to `now()` in the routes that handle reboot and reset in `pages/workspace/workspace.js`
* [x] add a new metric to `sprocs/workspace_loads_current.sql` (and `cron/workspaceHostLoads.js`) called `workspace_relaunching_count` which is the count of workspaces in state `launching` which have non-NULL values of `rebooted_at` or `reset_at`

PR #3134, opened 2020-09-14
Change workspace "launched_at" to denote start rather than end of launch

tdy:
Per #3026:

* [x] remove the `update_workspace_launched_at_now` query from `interface.js`
* [x] modify [`sprocs/workspaces_state_update.sql`](https://github.com/PrairieLearn/PrairieLearn/blob/master/sprocs/workspaces_state_update.sql) to set `launching_at = now()` whenever we change to state `launching`

### DIFF
--- a/workspace_host/interface.js
+++ b/workspace_host/interface.js
@@ -1128,11 +1128,8 @@ function initSequence(workspace_id, useInitialZip, res) {
                 logger.error(`Error for workspace_id=${workspace_id}: ${err}\n${err.stack}`);
             });
         } else {
-            sqldb.query(sql.update_workspace_launched_at_now, {workspace_id}, (err) => {
-                if (ERR(err)) return;
-                debug(`Container initialized for workspace_id=${workspace_id}`);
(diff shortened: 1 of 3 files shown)


workspace_host/interface.js
```diff
@@ -1128,11 +1128,8 @@ function initSequence(workspace_id, useInitialZip, res) {
                 logger.error(`Error for workspace_id=${workspace_id}: ${err}\n${err.stack}`);
             });
         } else {
-            sqldb.query(sql.update_workspace_launched_at_now, {workspace_id}, (err) => {
-                if (ERR(err)) return;
-                debug(`Container initialized for workspace_id=${workspace_id}`);
-                workspaceHelper.updateState(workspace_id, 'running', null);
-            });
+            debug(`Container initialized for workspace_id=${workspace_id}`);
+            workspaceHelper.updateState(workspace_id, 'running', null);
```
mwest1066, line 1132: Can we also delete the `sql.update_workspace_launched_at_now` query?

tdy: done

PR #3138, opened 2020-09-15
Change workspace init to keep hosts healthy on container errors

### LINKED ISSUE #3081: Don't set workspace hosts unhealthy if the error was due to the container
As @nicknytko pointed out, we currently set the workspace host to unhealthy if there is any error during launch: https://github.com/PrairieLearn/PrairieLearn/blob/05e00bfa577ec2d4edd0d40085ecb56419acc7c2/workspace_host/interface.js#L1124

However, these errors might be caused by problems with the workspace itself. Maybe the instructor set an invalid image name. Maybe the image has an error inside it that means it can't start up correctly.

Instead of always setting the host to unhealthy, we should just set the workspace to `stopped` (with an appropriate error message) and return. This would be probably be easier to implement if we rewrite this function as async/await rather than using `waterfall()`. It could be something like:

```
try {
    ...
    await _getInitialZipAsync(); // errors from this will be caught by the top-level catch which will set host to unhealthy
    ...
    try {
        await _pullImage(); // errors from this will be caught immediately
    } catch (err) {
        workspaceHelper.updateState(workspace_id, 'stopped', `Error! Unable to pull image. Click "Reboot" to try again. Detail: ${err}`);
        return; // immediately exit from the initSequence(); don't set host to unhealthy
    }
    ...
} catch (err) {
    workspaceHelper.updateState(workspace_id, 'stopped', `Error! Click "Reboot" to try again. Detail: ${err}`);
    markSelfUnhealthy((err2) => {...});
    ...
}
```

tdy:
Resolves #3081:

* [x] Rewrite `interface.initSequence()` as async/await rather than waterfall
* [x] Instead of always setting the host to unhealthy, just set workspace to `stopped` (with an appropriate error message) and return

### DIFF
--- a/workspace_host/interface.js
+++ b/workspace_host/interface.js
@@ -89,7 +89,7 @@ app.get('/status', asyncHandler(async (req, res) => {
 }));
 
 // TODO: refactor into RESTful endpoints (https://github.com/PrairieLearn/PrairieLearn/pull/2841#discussion_r467245108)
-app.post('/', function(req, res) {
+app.post('/', asyncHandler(async (req, res) => {
     const workspace_id = req.body.workspace_id;
     const action = req.body.action;
     const useInitialZip = _.get(req.body.options, 'useInitialZip', false);
@@ -98,7 +98,7 @@ app.post('/', function(req, res) {
     } else if (action == null) {
         res.status(500).send('Missing action');
     } else if (action == 'init') {
-        initSequence(workspace_id, useInitialZip, res);
+        await initSequenceAsync(workspace_id, useInitialZip, res);
     } else if (action == 'reset') {
         resetSequence(workspace_id, res);
     } else if (action == 'getGradedFiles') {
@@ -106,7 +106,7 @@ app.post('/', function(req, res) {
     } else {
         res.status(500).send(`Action '${action}' undefined`);
     }
-});
+}));
 
 let server;
 let workspace_server_settings = {};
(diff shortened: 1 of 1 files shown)


workspace_host/interface.js
```diff
@@ -1070,71 +1050,78 @@ function _startContainer(workspace, callback) {
         callback(null, workspace);
     });
 }
+const _startContainerAsync = util.promisify(_startContainer);
 
 // Called by the main server the first time a workspace is used by a user
-function initSequence(workspace_id, useInitialZip, res) {
+async function initSequenceAsync(workspace_id, useInitialZip, res) {
```
mwest1066, line 1056: I love the new `async` version of this function! So much easier to read.

mwest1066 (approved):

tdy:
* merged `master`
* also moved the hostname query inline: https://github.com/PrairieLearn/PrairieLearn/pull/3138/commits/d9f1e3cee7730c1e9dc8974639acea49bb98c7e9

mwest1066:
Another conflict, I'm afraid.

tdy:
> 
> 
> Another conflict, I'm afraid.

* conflict resolved
* also fixed a `launched_at` oversight from #3134 (could've sworn I tested it, so not sure how that slipped by)

mwest1066:
> * added `logger.warn` when marking host unhealthy

Good idea.

PR #3151, opened 2020-09-18
adding retry for 502 bad gateway error for api_download.py

chen386:


tools/api_download.py
```diff
@@ -42,8 +42,16 @@ def get_and_save_json(path, filename, args, logfile):
...
     r = requests.get(url, headers=headers)
-    if r.status_code != 200:
-        raise Exception(f'Invalid status returned for {url}: {r.status_code}')
+    while True:
+        r = requests.get(url, headers=headers)
+        if r.status_code == 200:
+            break
+        elif r.status_code == 502:
+            log(logfile, f'Bad Gateway Error encountered for {url}, retrying in 1 second, please check your network if it persists.')
+            time.sleep(1)
```
mwest1066, line 51: Can we add a limit to the number of retries, past which we will throw an exception to terminate?

tools/api_download.py
```diff
@@ -42,8 +42,22 @@ def get_and_save_json(path, filename, args, logfile):
     log(logfile, f'downloading {url} ...')
     start_time = time.time()
     r = requests.get(url, headers=headers)
-    if r.status_code != 200:
-        raise Exception(f'Invalid status returned for {url}: {r.status_code}')
+    retry_502_max = 300
```
mwest1066, line 45:
```suggestion
    retry_502_max = 30
```

tools/api_download.py
```diff
@@ -42,8 +42,22 @@ def get_and_save_json(path, filename, args, logfile):
...
+    retry_502_i = 0
+    while True:
+        r = requests.get(url, headers=headers)
+        if r.status_code == 200:
+            break
+        elif r.status_code == 502:
+            retry_502_i += 1
+            if retry_502_i == retry_502_max:
```
mwest1066, line 53:
```suggestion
            if retry_502_i >= retry_502_max:
```

tools/api_download.py
```diff
@@ -42,8 +42,22 @@ def get_and_save_json(path, filename, args, logfile):
...
+                raise Exception(f'Maximum number of retries reached on 502 Bad Gateway Error for {url})
+            else:
+                log(logfile, f'Bad Gateway Error encountered for {url}, retrying in 1 second')
+                time.sleep(1)
```
mwest1066, line 57:
```suggestion
                log(logfile, f'Bad Gateway Error encountered for {url}, retrying in 10 seconds')
                time.sleep(10)
```

PR #3154, opened 2020-09-20
Fix unload confirmation by filtering question form serialization

### LINKED ISSUE #3153: Confirmation dialog for unsaved file even though file was submitted
I'm running an exam in TA role and student view\. One of the questions is a file-upload question with external grading. Once the grading finishes and I get 100%, the question is closed and I can't submit any longer. But if I then click on "Next question" or "Assessment Overview", a warning is given: "Leave site? Changes you made may not be saved." Even though the file was already submitted, and wasn't changed again. This was tested on Chrome and Edge Chromium.

This was discussed on Slack: https://prairielearn.slack.com/archives/C266KEH9A/p1600474411047100

I'm not posting the question given it's being delivered to students this week, but can provide in private if required.

tdy:
With Jonatan's question:

* On submission, the question form gets serialized with `__variant_id` and `__csrf_token` query params
* On reload, those params are gone (thus the confirmation dialog)

---

* [x] Fix #3153 by stripping `__variant_id` and `__csrf_token` when checking initial/current form states

### DIFF
--- a/public/localscripts/questionPageScripts.js
+++ b/public/localscripts/questionPageScripts.js
@@ -1,6 +1,6 @@
 function confirmOnUnload() {
     const form = $('form.question-form');
-    const getForm = () => form.serialize().split('&')[0];
+    const getForm = () => form.find(':not([name="__variant_id"]):not([name="__csrf_token"])').serialize();
 
     // Set form state on load and submit
     var initialForm = getForm();
--- a/public/localscripts/questionPageScripts.js
+++ b/public/localscripts/questionPageScripts.js
@@ -1,6 +1,6 @@
 function confirmOnUnload() {
     const form = $('form.question-form');
-    const getForm = () => form.serialize();
+    const getForm = () => form.serialize().split('&')[0];
 
     // Set form state on load and submit
     var initialForm = getForm();


public/localscripts/questionPageScripts.js
```diff
@@ -1,6 +1,6 @@
 function confirmOnUnload() {
     const form = $('form.question-form');
-    const getForm = () => form.serialize();
+    const getForm = () => form.serialize().split('&')[0];
```
jonatanschroeder, line 3: Wouldn't this affect questions with multiple components? Since (I believe) in that case this would only consider the first field.

jonatanschroeder: Would something like this be more appropriate? https://stackoverflow.com/questions/4556172/excluding-certain-inputs-on-serialize

tdy: Switched to something more robust. Tested/working on my end.

https://github.com/PrairieLearn/PrairieLearn/blob/d56ceaf1bc14e277bfe3757a434e31e9d8a101a5/public/localscripts/questionPageScripts.js#L3

PR #3155, opened 2020-09-20
Use ECR image cache for workspaces

mwest1066:


workspace_host/interface.js
```diff
@@ -624,14 +625,24 @@ const _checkServerAsync = util.promisify(_checkServer);
...
         workspace_image: result.rows[0].workspace_image,
         workspace_port: result.rows[0].workspace_port,
         workspace_home: result.rows[0].workspace_home,
         workspace_graded_files: result.rows[0].workspace_graded_files,
         workspace_args: result.rows[0].workspace_args || '',
         workspace_sync_ignore: result.rows[0].workspace_sync_ignore || [],
     };
+
+    if (config.externalGradingImageRepository) {
+        var repository = new dockerUtil.DockerName(settings.workspace_image);
```
tdy, line 638:
```suggestion
        let repository = new dockerUtil.DockerName(settings.workspace_image);
```
or even `const`?

mwest1066: Done

tdy (approved):
Don't see any glaring issues by eye test. Is there some way I should be testing locally, or not needed?

mwest1066:
> Don't see any glaring issues by eye test. Is there some way I should be testing locally, or not needed?

I've been testing on pl-dev, so I think we should be ok. A local testing setup would be awesome but I don't know how to do it without a huge amount of setup.

PR #2518, opened 2020-05-23
Groupwork project

frankyhao:
Implementing group work functions for Prairelearn so that students could form groups and collaborate together.
[New DB schema design and screenshots
](https://docs.google.com/document/d/14LjsCisZy9hc2fL1GdeyvyqOIVJMwhVUGG9IEAWcdS4/edit?usp=sharing)

---
### Progress
- [x] new DB migration 
- [x] sync (load from disk)
    - [x] create group_configs if instructor set "groupwork = true" in an assessmentInfo.json
    - [x] [stage 2] enforce max and min requirement for groups
- [x] group management admin page
    - [x] UI design
    - [x] add a group
    - [x] delete all groups
    - [x] upload a csv format file for group setting (manually assigned by instructor)
    - [x] auto generate groups (randomly assigned by system)
    - [ ] [stage 2] more smart options to auto generate groups
    - [x] copy the group setting from another assessment
    - [x] actions for a group (add/delete members & delete a group)
    - [x] display students who do not have a group
    - [x] English language check
- [x] downloads
    - [x] download a csv file containing all groups info
    - [ ] [stage 2] include GID inside csv, so it could be easily recover
- [x] instance management (instructor perspective)
    - [x] display instances based on GID instead of UID(original students admin page)
    - [x] show submission history in details page (instructor could track which team member does not contribute)
    - [ ] [stage 2] contribution statistics
- [x] instance (student perspective)
    - [x] share same instance within the group
    - [x] a "middle page" for students who are not in group to create/join groups
    - [ ] [stage 2] group chat box
    - [ ] [stage 2] sync code box (similar to google docs / VScode online)

---
### Testing
- [x] sync (students assessment instance progress) integration tests: generate 3 students -> put them into a group -> start instance -> they are in the same instance -> when a student have a new submission, instance grade will be updated and synchronized
- [x] instructor group management: use auto generate to put 500 students into 50 groups -> check whether everyone is in group & whether each group has 10 students -> delete all groups -> check whether all groups are soft deleted
- [x] student group management: put 5 students into a group -> set minimum group size to be 3 in group config & group access control -> lead student create a group -> cannot start assessment because min group size is not reached -> others could join the same instance via friendcode -> they share the same instance -> students cannot join via friendcode if the group exceeds the maximum size -> someone left the group and do not have permission to the instance anymore

---
### Known bugs
- [x] groupname should be unique for each group (Frankie)
- [x] question sql sharing by groupwork (Zhining)
[...]

PR #3157, opened 2020-09-21
Group support

### LINKED ISSUE #2894: Updates to group work support
@frankyhao @zhining4 I merged #2518 so that faculty can get ready for Fall semester, but I also added a number of comments on the PR. Could you please address these in the (closed) PR, and open a new PR to make requested changes?

frankyhao:
fixes #2894
- [x] code styles - JS and SQL naming standards such as snake_case for _actions
- [x] error handling
- [x] passing res.locals.authn_user.user_id  into group_logs
- [x] race condition when sql.check_group_size
- [x] verify untrusted user input for copying assessment by making sure they belong to the same course instance
- [x] verify untrusted user input by checking users enrollments before add/remove members
- [x] verify untrusted user input by checking whether the group belong to the assessment

There are 5 more comments that need to be resolved in #2518. I am asking @zhining4 to help with that, and he will create another PR for them.


sprocs/check_join_group.sql
```diff
@@ -0,0 +1,70 @@
...
+        JOIN group_configs AS gc ON g.group_config_id = gc.id
+    WHERE 
+        g.name = check_join_group.arg_group_name 
+        AND g.join_code = check_join_group.arg_join_code
+        AND gc.assessment_id = check_join_group.arg_assessment_id
+        AND g.deleted_at IS NULL
+        AND gc.deleted_at IS NULL;
+
+    IF NOT FOUND THEN
+        RAISE EXCEPTION 'Cannot find the group by provided join code';
```
mwest1066, line 31:
```suggestion
        RAISE EXCEPTION 'Cannot find the group by provided join code: %', arg_join_code;
```

sprocs/check_join_group.sql
```diff
@@ -0,0 +1,70 @@
...
+        g.id
+    INTO 
+        arg_group_id
+    FROM 
+        groups AS g
+        JOIN group_configs AS gc ON g.group_config_id = gc.id
+    WHERE 
+        g.name = check_join_group.arg_group_name 
+        AND g.join_code = check_join_group.arg_join_code
+        AND gc.assessment_id = check_join_group.arg_assessment_id
```
mwest1066, line 26: If we are using unique argument names that start with `arg_` then we don't need to add the `check_join_group.` at the start.

sprocs/check_join_group.sql
```diff
@@ -0,0 +1,70 @@
...
+        groups g
+        JOIN group_configs AS gc ON g.group_config_id = gc.id
+        LEFT JOIN group_users AS gu ON gu.group_id = g.id
+    WHERE
+        g.id = arg_group_id
+    GROUP BY
+        g.id;
+
+    IF arg_cur_size >= arg_max_size THEN
+        RAISE EXCEPTION 'The group is full';
```
mwest1066, line 55:
```suggestion
        RAISE EXCEPTION 'The group is full for join code: %', arg_join_code;
```

sprocs/check_join_group.sql
```diff
@@ -0,0 +1,70 @@
...
+        AND g.join_code = check_join_group.arg_join_code
+        AND gc.assessment_id = check_join_group.arg_assessment_id
+        AND g.deleted_at IS NULL
+        AND gc.deleted_at IS NULL;
+
+    IF NOT FOUND THEN
+        RAISE EXCEPTION 'Cannot find the group by provided join code';
+    END IF;
+
+    -- lock the group_users
```
mwest1066, line 34: I don't think that locking the group users is what we want here. Locking these will stop them being deleted, but it won't stop the addition of a new row in `group_users` for this group. Instead I think we should lock the `groups` row. If all functions which are going to modify group membership lock the corresponding `groups` row before they add/remove group users then we will be safe from race conditions. Please check that this makes sense! I'm just thinking out loud here.

sprocs/check_join_group.sql
```diff
@@ -0,0 +1,70 @@
...
+    PERFORM gu.group_id
+    FROM group_users AS gu
+    WHERE gu.group_id = arg_group_id
+    FOR UPDATE OF gu;
+
+    -- count the group size and compare with the max size
+    SELECT
+        COUNT(gu) AS cur_size, AVG(gc.maximum) AS maximum
```
mwest1066, line 42: No need for `AS` names here.
```suggestion
        COUNT(gu), AVG(gc.maximum)
```

sprocs/check_join_group.sql
```diff
@@ -0,0 +1,70 @@
+CREATE OR REPLACE FUNCTION
+    check_join_group (
```
mwest1066, line 2: For sprocs we are trying to name them starting with the table they mainly affect. Based on that, could we call this `group_users_insert` or something like that?

pages/instructorAssessmentGroups/instructorAssessmentGroups.js
```diff
@@ -146,67 +158,112 @@ router.post('/', function(req, res, next) {
...
-                    failedUids += '[' + uid + '] ';
-                }
-            }
-            if (failedUids.length > 0) {
-                res.locals.errormsg += 'Failed to add ' + failedUids + 'to Group No.' + gid + '. Please check if the uid exist.\n';
+
+        //verify the group belong to the assessment first
+        let params = {
+            assessment_id,
+            gid,
```
mwest1066, line 172: In PL we use `TABLENAME_id` to refer to the `id` column in tables. So I think this should be `group_id`. Names like `qid` for questions refer to the string name of the question, not the `question_id`. Sorry! I know this is confusing.

pages/instructorAssessmentGroups/instructorAssessmentGroups.js
```diff
@@ -146,67 +158,112 @@ router.post('/', function(req, res, next) {
...
+        };
+        sqldb.query(sql.verify_group, params, function(err, result) {
```
mwest1066, line 174: While it might be ok here, in general I think this pattern of "verify and then change" is risky because it tends to introduce [TOCTAU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) race conditions.

Instead, here I think it would be better to send the `assessment_id` to `assessment_groups_add_member()` so that we can check at the same time as we make the change. The checking should happen either as we lock the `groups` row, or _after_ we've locked the `groups` row.

(and similarly for the other functions like `_delete_member` and `_delete_group`)

pages/instructorAssessmentGroups/instructorAssessmentGroups.js
```diff
@@ -92,10 +104,10 @@ router.post('/', function(req, res, next) {
             if (ERR(err, next)) return;
             res.redirect(req.originalUrl);
         });
-    } else if (req.body.__action == 'addGroup') {
+    } else if (req.body.__action == 'add_group') {
         const assessment_id = res.locals.assessment.id;
-        const groupname = req.body.groupname;
-        if (String(groupname).length < 1) {
+        const group_name = req.body.group_name;
+        if (String(group_name).length < 1) {
```
mwest1066, line 110: If `group_name` is `undefined` then I think this will change it to the string `"undefined"` and pass this check?

pages/instructorAssessmentGroups/instructorAssessmentGroups.js
```diff
@@ -74,9 +74,21 @@ router.post('/', function(req, res, next) {
...
+        res.locals.assessments.forEach(element => {
+            if (element.id == req.body.copyAssessmentId) {
+                sameInstance = true;
+            }
+        });
+        if (!sameInstance) {
+            res.locals.errormsg = 'The assessment that you are copying does not belong to the same course instance. Please select another one.';
+            obtainInfo(req, res, next);
+            return;
+        }
```
mwest1066, line 88: I would suggest deleting all these checks in the JS code and instead add a check in the `assessment_groups_copy` sproc. When you look up the `temp_copy_group_config_id` you can do something like:
```
SELECT gc.id INTO temp_copy_group_config_id
FROM
    assessments AS src_a
    JOIN group_configs AS src_gc ON (src_gc.assessment_id = src_a.id)
    JOIN assessments AS dest_a ON (dest_a.id = assessment_id)
WHERE
    src_a.id = copying_assessment_id
    AND src_a.course_instance_id = dest_a.course_instance_id   -- check they are in the same course instance
    AND src_a.deleted_at IS NULL
    AND gc.deleted_at IS NULL;
```

After this you can do:
```
IF NOT FOUND THEN
    RAISE EXCEPTION 'Could not find source assessment';
END IF;
```

mwest1066:
What does the `@` symbol mean here? https://github.com/PrairieLearn/PrairieLearn/blob/79560ac856e2dee7123e7278a326da231774ecd9/sprocs/assessment_groups_copy.sql#L39

sprocs/assessment_groups_delete_group.sql
```diff
@@ -1,21 +1,38 @@
...
+    -- then lock the group row
+    PERFORM 1
+    FROM
+        group_configs AS gc
+        JOIN groups AS g ON gc.id = g.group_config_id
+    WHERE
+        gc.assessment_id = assessment_groups_delete_group.assessment_id
+        AND g.id = arg_group_id
+        AND g.deleted_at IS NULL
+        FOR UPDATE of g;
```
mwest1066, line 21:
```suggestion
    FOR UPDATE of g;

```

sprocs/assessment_groups_add_member.sql
```diff
@@ -10,17 +11,40 @@ DECLARE
...
+    -- then lock the group row
+    PERFORM 1
+    FROM
+        group_configs AS gc
+        JOIN groups AS g ON gc.id = g.group_config_id
+    WHERE
+        gc.assessment_id = assessment_groups_add_member.assessment_id
+        AND g.id = arg_group_id
+        AND g.deleted_at IS NULL
+        FOR UPDATE of g;
```
mwest1066, line 24:
```suggestion
    FOR UPDATE of g;

```

sprocs/assessment_groups_delete_member.sql
```diff
@@ -10,21 +11,44 @@ DECLARE
...
+    -- then lock the group row
+    PERFORM 1
+    FROM
+        group_configs AS gc
+        JOIN groups AS g ON gc.id = g.group_config_id
+    WHERE
+        gc.assessment_id = assessment_groups_delete_member.assessment_id
+        AND g.id = arg_group_id
+        AND g.deleted_at IS NULL
+        FOR UPDATE of g;
```
mwest1066, line 24:
```suggestion
    FOR UPDATE of g;

```

sprocs/group_users_insert.sql
```diff
@@ -0,0 +1,77 @@
...
+        AND g.join_code = arg_join_code
+        AND gc.assessment_id = arg_assessment_id
+        AND g.deleted_at IS NULL
+        AND gc.deleted_at IS NULL;
+
+    IF NOT FOUND THEN
+        RAISE EXCEPTION 'Cannot find the group by provided join code: %', arg_join_code;
+    END IF;
+
+    -- lock the group
```
mwest1066, line 35: This is great! I really like the clear pattern of "find the group, lock it, check the size constraint, modify the group". This is very safe and easy to understand.

mwest1066 (approved):

Issue #3171, opened 2020-09-22
Add a Manual Grading "Element" to the Element Library


bevansm:
**Adding Manual Grade Flag to Elements**

As an instructor who wants to use manual grading, I want to add a "manual=true" flag in an HTML element, so that I can specify that the element is for manual grading.

- Can all elements support the "manual=true" flag? ie. pl-file-upload
- Is it possible to implement this for all elements given time constraints? (In other words, is there a lot of programming for each element to implement this). Is there a workaround?
- What is the ideal JSON data structure produced by an element that is flagged "manual=true"? (the data exported to CSV)

Blocked by:
- [Manual Grading "Element" Spike](https://github.com/PrairieLearn/PrairieLearn/projects/9#card-45195508)

mebird:
- We don't have the endpoint set up to submit the form, but this is a good first pass of things

jonatanschroeder:
All the rationale for this element has been replaced with other implementation strategies. Closing.

PR #3174, opened 2020-09-22
Manual Grading Element

mebird:
From https://github.com/PrairieLearn/PrairieLearn/issues/3171

- Creates a basic manual grading overlay, with the endpoint aggregating scores
   - Creates an overlay where students can see what partials they need graded
   - Creates an instructor overlay in order to manipulate partials
- Stubs out functions and endpoints for future functionality (i.e. admin page, "next question" endpoint)
- Needs to deal with question & partial statuses
- Need to send data off to whatever internal grading workflow we have

How to use:
- Run the PL repo
- Navigate to the first test course. Open up the manual grading assignment.
- Create a question
- In the resulting page, change the url from `/instructor/question/x/preview/?variant_id=x` to  `/instructor/question/x/manual_grading/?variant_id=x` (move to the manual grading endpoint)
- Click the edit button in the upper right corner of the element, and update score/feedback. Close the popup.
- Click "submit grades".
- View your terminal. You should see the JSON object we'll pass on printed there.


docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+      context: .
+    ports:
+      - 3000:3000
+    volumes:
+      - ./testCourse:/course
+      - /var/run/docker.sock:/var/run/docker.sock
+      - ${HOME}/pl_ag_jobs:/jobs
+    container_name: pl
+    environment:
+      - HOST_JOBS_DIR=${HOME}/pl_ag_jobs
```
tdy, line 14: Does this replace the huge `docker run` command with something like `docker-compose -p pl up`?

mebird: This should. If you just run `docker-compose up`, then the entire stack should go up, with the container named `pl`.

Before this comes along for the ride on my PR (or i pull it out into another PR), I did want to check if this had all the stuff we actually need to run PL in a docker container with someone else from the team.

tdy: This is awesome! Is there a way to replicate something like this where we append a shell command:
`docker run ... prairielearn/prairielearn /PrairieLearn/docker/start_workspace.sh`

tdy: > or i pull it out into another PR

Also imo this is worth splitting out into its own PR.

mebird: Yup. You can either do:
`docker-compose run /PrairieLearn/docker/start_workspace.sh` (one time command in a running container)
or, alternatively, feed it in as a [command to the container](https://docs.docker.com/compose/compose-file/#command), overriding the `/PrairieLearn/docker/init.sh
` at the end of the Dockerfile.
```yml
version: '3.2'
services:
  pl:
    command: /PrairieLearn/docker/start_workspace.sh
```
We could add this as a commented out line in the docker compose, so people can play with it as they need. If it's accidently committed, then we can just remind someone to comment it back out.

I'm not 100% sure what `start_workspace.sh` does--is it a wrapper of `init.sh`? Does it spin up other services or containers?

Since this sounds like something that's more generally useful (and this is really a draft-draft for manual grading rn), should I pull this out into a separate PR and request you and matt on it? We can always add more stuff to the file after, but this might be a good first step to kill the magical bash script floating around.

tdy: > I'm not 100% sure what start_workspace.sh does--is it a wrapper of init.sh? Does it spin up other services or containers?

Yeah it's just a wrapper that splits the main PL server and workspace host server into separate tmux panes. The main `init` script lumps output from both servers together, but when dev'ing workspaces, it's useful to view the servers separately.

mebird: Out of curiosity, do we run these as separate servers in prod.

tdy: Yeah they are separate. Based on https://github.com/PrairieLearn/PrairieLearn/issues/3178#issuecomment-698030570, it looks like these are also separate:

> * Main PL server
> * Chunk servers
> * PrairieGrader hosts
> * Workspace hosts

mebird: If they're separate servers, it may be worth looking into pulling them apart into different services (i.e. containers) in the compose file.

For now though, I'll shift the basic docker compose to a different pr and update docs, making notes about running a different entry command, and commenting out the command line for now.

mebird:
Todos moving forward:

- Question statuses
 - Need to be able to mark an entire question as grading/pending graded
 - Need to be able to "intercept" partials in the grade step so we can set a status as graded/needs grading. We need the grading stage separation to do this. This is because some element code acts on partials before some questions even build their partials.
 - Need to be able to retrieve the next variant with the matching question ID in the given assignment so we can retrieve "next to grade". Will likely be modifying the query to return all with the given variant id and ungraded status, but take top 1.
 - Need to be able to manage statuses on a partial and a question level.
- UI/Interface
 - Need the bar to navigate between/to questions at the bottom of the page
 - Need the tab and corresponding page on the assignments tab
 - Need to brainstorm stats & the like

Issue #2576, opened 2020-06-08
Use tabulator for large tables


mwest1066:
For the Gradebook page and similar large tables it would be good to use something like http://www.tabulator.info

_Originally posted by @mwest1066 in https://github.com/PrairieLearn/PrairieLearn/pull/2479#issuecomment-639765218_

mwest1066:
Implemented in #4543

Issue #1297, opened 2018-09-28
Investigate Tabulator for our large tables


mwest1066:
See http://tabulator.info -- MIT licensed, has Bootstrap 4 styling available.

nwalters512:
We now use `bootstrap-table` for many of our large tables, which has been working well for us.

PR #3177, opened 2020-09-23
Add sticky headers to Gradebook table

jonatanschroeder:
The gradebook table, on large classes with large number of assessments, gets pretty big and hard to follow. This PR addresses it with the CSS `position: sticky` option which is supported by most modern browsers.


tdy:
Awesome. This can either replace or at least be a stopgap for #2576 #1297

jonatanschroeder:
The issue with this solution is that it only works if we scroll the table as a separate element. This is because `position: sticky` will be based on the closest parent with an `overflow` setting, which in this case is the `table-responsive` used to allow horizontal scrolling. So in order for this to work, either `overflow-x` needs to be disabled (horizontal scrolling is gone, which I imagine is not desired), or a fixed height needs to be set for the table and `overflow-y` set to auto, allowing the table to scroll separately from the page. Deciding on that height is the issue.

mfsilva22:
Just tested with 300 students in my course, and the gradebook looks awesome with the frozen headers. 

Can we do the same for the first two columns as well (UID and UIN)? I have lots of assessment, and it would be great to scroll to the right, and still see the corresponding names.

jonatanschroeder:
The library I used doesn't have the same for columns, but since there is no overlay problem in the columns I can use simple CSS. That only works for one column, though. Once side-effect is that I have to set the background to white to avoid problems with text on top of text, but that means the row hover feature doesn't work as well.

mfsilva22:
> The library I used doesn't have the same for columns, but since there is no overlay problem in the columns I can use simple CSS. That only works for one column, though. Once side-effect is that I have to set the background to white to avoid problems with text on top of text, but that means the row hover feature doesn't work as well.

if we have a trade-off, which feature do you think is more important as a starting point? I am okay either way, just thought fixing the user would make it easier to navigate (and it is okay to just freeze the first column, we usually only search for netids anyway.

jonatanschroeder:
I'd rather have the sticky column than the row hover (which is what I just pushed).

mfsilva22:
I think I am getting close to completely abandon compass next year! It worked very well running with my own course (and lots of assessments).

mwest1066:
@jonatanschroeder I’m afraid my package updates have caused a conflict with this PR, sorry!

jonatanschroeder:
> @jonatanschroeder I’m afraid my package updates have caused a conflict with this PR, sorry!

Resolved. This was an easy fix.

mwest1066:
I've been wondering for a while whether we should start using a more serious table UI library. At the moment we use https://github.com/Mottie/tablesorter, but maybe we should move to https://datatables.net or something like that? This PR is a partial move in that direction, so this is a good point to make a decision.

@nwalters512 what are your thoughts?

jonatanschroeder:
We've had a student playing around with tabulator in the students tab of an assessment.

mwest1066:
> We've had a student playing around with tabulator in the students tab of an assessment.

Excellent! Is tabulator your current best pick? I've struggled to find something that allows freezing of the first few columns, as well as the first (header) row.

jonatanschroeder:
I couldn't get the freeze option working for either datatables or tabulator yet. I can play around a bit later.

mwest1066:
> I couldn't get the freeze option working for either datatables or tabulator yet. I can play around a bit later.

Cool! I see that DataTables claims to support it (https://datatables.net/extensions/fixedcolumns/). I'd also be interested to know whether any of the pure-CSS solutions work. E.g., https://stackoverflow.com/questions/1312236/how-do-i-create-an-html-table-with-a-fixed-frozen-left-column-and-a-scrollable-b

jonatanschroeder:
> Cool! I see that DataTables claims to support it (https://datatables.net/extensions/fixedcolumns/).

The support in DataTables doesn't play nicely with scrolling, and fixedRows and fixedColumns are not compatible with each other... :-/

> I'd also be interested to know whether any of the pure-CSS solutions work. E.g., https://stackoverflow.com/questions/1312236/how-do-i-create-an-html-table-with-a-fixed-frozen-left-column-and-a-scrollable-b

This is roughly what this PR is using for fixed columns, the problem is fixed header rows, which the pure CSS solution doesn't properly solve if there is horizontal scrolling involved.

jonatanschroeder:
Bootstrap-table looks promising: https://bootstrap-table.com/

They have an extension for sticky header: https://examples.bootstrap-table.com/#extensions/sticky-header.html

I can't find an official way to to sticky columns, but I can use CSS for that.

mwest1066:
> The support in DataTables doesn't play nicely with scrolling, and fixedRows and fixedColumns are not compatible with each other... :-/

Oh yuck! I never imagined that they would be incompatible! 😄

> This is roughly what this PR is using for fixed columns, the problem is fixed header rows, which the pure CSS solution doesn't properly solve if there is horizontal scrolling involved.

Ah, very good, I had not read it carefully enough to realize that.

> Bootstrap-table looks promising: https://bootstrap-table.com/

Oh nice!

Ok, I'll let you figure out the best all round solution 😄

jonatanschroeder:
@mwest1066 The implementation with bootstrap-table looks **way** better than the other one. I'm still doing some tests, though, to ensure the sorting of grades is done with numeric values (and role has appropriate order), but feel free to take a look. I also added a search button, which I'm trying to tweak to remove the extra vertical space.

mwest1066:
> @mwest1066 The implementation with bootstrap-table looks **way** better than the other one. I'm still doing some tests, though, to ensure the sorting of grades is done with numeric values (and role has appropriate order), but feel free to take a look. I also added a search button, which I'm trying to tweak to remove the extra vertical space.

Sweet! Any chance you could move this to a branch in the main PL repo, rather than a fork? Might need a new PR?

PR #3180, opened 2020-09-23
Add wh.unhealthy_reason when workspace hosts go unhealthy

tdy:
Per #2935:

* [x] Add `workspace_hosts.unhealthy_reason` column and set it
* [x] Make sure we only transition to `unhealthy` if we aren't already unhealthy, so the `unhealthy_reason` will be the very first time we detect unhealthy
* [x] Update admin queries (https://github.com/PrairieLearn/PrairieLearn/pull/3180/commits/cc8421016538e4a2003754349e23af169233bd38)
    * `unhealthy_reason` input field in `admin_queries/workspace_hosts_set_all_unhealthy` (default reason: "admin query")
    * `unhealthy_reason` column in `admin_queries/workspace_hosts`


workspace_host/interface.sql
```diff
@@ -145,7 +145,8 @@ UPDATE
 SET
     state = 'unhealthy',
     state_changed_at = NOW(),
-    unhealthy_at = NOW()
+    unhealthy_at = NOW(),
+    unhealthy_reason = $unhealthy_reason
 WHERE
     wh.instance_id = $instance_id
     AND wh.state IN ('launching', 'ready', 'draining');
```
tdy, line 152:
```suggestion
    AND wh.state != 'unhealthy';
```
Not sure which one we want here (current code ignores `terminating` and `terminated`).

mwest1066: I think we want to stick with the current version. The `terminating` and `terminated` states should be a one-way trip to oblivion 😄

mwest1066 (commented):
Could we also set `unhealthy_reason` here: https://github.com/PrairieLearn/PrairieLearn/blob/4fb8e0c5f31f69cfce4b1a6271b63409d59d45cd/admin_queries/workspace_hosts_all_unhealthy.sql#L6

Maybe use "admin query" as the reason, or something like that.

Could we also add a column for `unhealthy_reason` here: https://github.com/PrairieLearn/PrairieLearn/blob/4fb8e0c5f31f69cfce4b1a6271b63409d59d45cd/admin_queries/workspace_hosts.sql#L7

mwest1066:
@tdy I removed the "Draft" flag (hope that's ok!)

PR #3189, opened 2020-09-24
Add cron job to warn workspaces nearing timeout

tdy:
Per #2935


lib/config.js
```diff
@@ -53,7 +53,8 @@ config.cronIntervalExternalGraderLoadSec = 8;
 config.cronIntervalServerLoadSec = 8;
 config.cronIntervalServerUsageSec = 8;
 config.cronIntervalCalculateAssessmentQuestionStatsSec = 10 * 60;
-config.cronIntervalStopStaleWorkspacesSec = 10 * 60;
+config.cronIntervalStopStaleWorkspacesSec = 60;
+config.cronIntervalWarnStaleWorkspacesSec = 60;
```
tdy, line 57: Are these crons guaranteed to be in sync?

mwest1066: Not guaranteed. Multiple servers can run cron jobs and they can diverge over time. Is it important that these stay in sync?

cron/warnStaleWorkspaces.js
```diff
@@ -0,0 +1,29 @@
...
+        const params = {
+            launched_timeout_sec: config.workspaceLaunchedTimeoutSec,
+            launched_timeout_warn_sec: config.workspaceLaunchedTimeoutWarnSec,
+        };
+        const result = await sqldb.queryAsync(sql.select_almost_stale_workspaces, params);
+        const workspaces = result.rows;
+        for (const workspace of workspaces) {
+            const launched_at = new Date(workspace.launched_at);
+            const timeout_at = new Date(launched_at.getTime() + config.workspaceLaunchedTimeoutSec * 1000);
+            const time_to_timeout_min = (timeout_at - Date.now()) / 1000 / 60;
```
mwest1066, line 24: Could we shift the calculation of `time_to_timeout_min` into the SQL? I think that the JS is completely correct here, but I generally prefer to keep time/date calculations in the SQL to avoid surprises with JS date handling.

lib/config.js
tdy: Both `stop` and `warn` do their calculations based on `launched_at`, but we don't actually stop those workspaces at exactly `launched_at + config.launchedTimeout`. I think the actual stopping occurs whenever the next `stop` cron happens to fire, so basically `warn`'s estimate is always off -- first because `stop` only occurs at given intervals and further if `stop`/`warn` get desync'ed.

I lowered the sproc intervals to 60 sec, so I think it will be off by at most 59 sec. Maybe that's good enough given there should be no data loss here, just interruption of workflow?

mwest1066: Yep, I think this is fine. I like the way you said "_less than_ XX min" in the warning message.

cron/warnStaleWorkspaces.sql
```diff
@@ -0,0 +1,6 @@
+-- BLOCK select_almost_stale_workspaces
+SELECT *
+FROM workspaces AS w
+WHERE
+    w.state = 'running'::enum_workspace_state
+    AND w.launched_at < now() - make_interval(secs => $launched_timeout_sec::int - $launched_timeout_warn_sec::int);
```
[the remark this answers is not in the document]
mwest1066: To speed up other queries it might actually be a good idea to have an index on `workspaces` for `(state, launched_at)`.

cron/index.js
```diff
@@ -90,6 +90,11 @@ module.exports = {
                 module: require('./stopStaleWorkspaces'),
                 intervalSec: config.cronOverrideAllIntervalsSec || config.cronIntervalStopStaleWorkspacesSec,
             },
+            {
+                name: 'warnStaleWorkspaces',
```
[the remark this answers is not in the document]
tdy: done in bbac2a018ee9b3f32b2cf650c9057d126431a0ad (also reworded stale->timeout: `workspaceTimeout{Stop,Warn}`)

cron/warnStaleWorkspaces.js
```diff
@@ -0,0 +1,29 @@
...
+const logger = require('../lib/logger');
+const workspaceHelper = require('../lib/workspace');
+const sqldb = require('@prairielearn/prairielib/sql-db');
+const sqlLoader = require('@prairielearn/prairielib/sql-loader');
+
+const sql = sqlLoader.loadSqlEquiv(__filename);
+
+module.exports = {};
+
+module.exports.run = function(callback) {
```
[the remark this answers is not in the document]
tdy: done in 42be2270ed015835c4852620d9ccfadadbf1f6e5

cron/warnStaleWorkspaces.js
tdy: done in 89c13c3830c4998c6b926ed16f18c3b31dfdb1f0

cron/warnStaleWorkspaces.sql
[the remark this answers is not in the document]
tdy: added in 330afea90fd8f223209de2ff1773e98dd1e6899a (or did you mean create index for both `launched_at` and `(state, launched_at)`?)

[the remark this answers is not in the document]
mwest1066: Sorry for the two-part comment above. You did exactly what I was thinking!

PR #3190, opened 2020-09-24
Disable timeLimitMin for CBTF assessments

### LINKED ISSUE #2054: Disable `timeLimitMin` rule on Exam mode assessments
Since `"mode": "Exam"` assessments have their timing managed by the CBTF and proctors, we should reconsider 1) if assessment time limits should apply in that environment, or 2) if we should try to make the visible time limits sourced from the scheduler instead of the assessment.

tdy:
Resolves #2054:

* [x] Add a second CASE option like `WHEN aar.mode = 'Exam' THEN NULL` here: https://github.com/PrairieLearn/PrairieLearn/blob/4fb8e0c5f31f69cfce4b1a6271b63409d59d45cd/sprocs/check_assessment_access.sql#L44
* [x] Add a documentation note near "Note that time limits should not be set for exams" to say that time limits will not activate in Exam mode: https://github.com/PrairieLearn/PrairieLearn/blob/master/docs/accessControl.md
* [x] Add test to check for `#countdownDisplay`

### DIFF
--- a/tests/testAccess.js
+++ b/tests/testAccess.js
@@ -305,26 +305,6 @@ describe('Access control', function() {
         });
     };
 
-    const getAssessmentInstanceCountdown = function(cookies, expectedStatusCode, callback) {
-        request({url: assessmentInstanceUrl, jar: cookies}, function (error, response, body) {
-            if (error) {
-                return callback(error);
-            }
-            if (response.statusCode != expectedStatusCode) {
-                return callback(new Error('bad status: ' + response.statusCode));
-            }
-            page = body;
-            try {
-                $ = cheerio.load(page);
-                elemList = $('#countdownDisplay');
-                assert.lengthOf(elemList, 0);
-            } catch (err) {
-                return callback(err);
-            }
-            callback(null);
-        });
-    };
-
     describe('9. GET to assessment_instance URL', function() {
         it('as student should return 403', function(callback) {
             getAssessmentInstance(cookiesStudent(), 403, callback);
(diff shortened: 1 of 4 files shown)


tests/testAccess.js
```diff
@@ -305,6 +305,26 @@ describe('Access control', function() {
         });
     };
 
+    const getAssessmentInstanceCountdown = function(cookies, expectedStatusCode, callback) {
```
tdy, line 308: I have a stashed version that merges this functionality into `getAssessmentInstance()` via `options` if that's preferred.

mwest1066: I guess I'd be inclined to change `getAssessmentInstance()` to return the page via `callback(null, page)`. Then the `should not contain countdown timer` test could call the regular `getAssessmentInstance()` function and do the `elemList = $('#countdownDisplay');` testing inside that particular test. This would remove the need for the special `getAssessmentInstanceCountdown()` function.

tdy: Not sure if this is what you had in mind: 1706b2f0d0d0feaeb7bf9e1ccdcb84f304789354

tdy: Actually I guess you were thinking to leave the `statusCode` check inside `getAssessmentInstance()` 🤦

tdy: Take 2: 0792d77264fa6832170e1190512199cadbcb5757 (I realized `page` is actually global, so we can just do the timer check via callback)

tests/testAccess.js
```diff
@@ -565,6 +565,18 @@ describe('Access control', function() {
         it('should enable access to the assessment_instance', function(callback) {
             getAssessmentInstance(cookiesStudentExam(), 200, callback);
         });
+        it('should not contain countdown timer', function(callback) {
+            getAssessmentInstance(cookiesStudentExam(), 200, function() {
```
mwest1066, line 569:
```suggestion
            getAssessmentInstance(cookiesStudentExam(), 200, function(err) {
                if (ERR(err, callback)) return;
```

tests/testAccess.js
mwest1066: Looks great! One tiny change below. I also hadn't realized that `page` was global. That makes things easy!

tdy: Done

PR #3205, opened 2020-09-27
Implementation of C autograder

jonatanschroeder:
This PR replaces #3082.


exampleCourse/questions/demo/cGrader/info.json
```diff
@@ -0,0 +1,17 @@
...
+    "tags": [
+        "jonatan", "demo", "autograder", "c"
+    ],
+    "type": "v3",
+    "singleVariant": true,
+    "gradingMethod": "External",
+    "externalGradingOptions": {
+        "enabled": true,
+        "image": "prairielearn/grader-c",
+        "timeout": 1000,
```
mwest1066, line 14: This is in seconds, so I worry that 1000 is rather long to detect that student code is spinning in an infinite loop. Maybe we could shorten this to 30 or something like that?

exampleCourse/questions/demo/cGrader/info.json
```diff
@@ -0,0 +1,17 @@
+{
```
mwest1066, line 1: Maybe shift this question to `demo/autograder/c/square` to mirror `demo/autograder/python/square`?

graders/c/cgrader/cgrader.py
```diff
@@ -0,0 +1,217 @@
...
+                tostr = '\n\nTIMEOUT! Typically this means the program took too long,' + \
+                        '\nrequested more inputs than provided, or an infinite loop was found.' + \
+                        '\nIf your program is reading data using scanf inside a loop, this ' + \
+                        '\ncould also mean that scanf does not support the input provided ' + \
+                        '\n(e.g., reading an int if the input is a double).'
+            finally:
+                out = (out1 if out1 else '') + (out2 if out2 else '') + tostr
+                return out
+
+    def test_compile_file(self, c_file, exec_file, main_file=None, points=1, field=None, name='Compilation'):
```
mwest1066, line 46: It's possible to have autograders return an invalid/non-gradable score for the grading, which doesn't use up retry-attempts on a question. Some people have been using this for cases when the code doesn't compile, so students don't get penalized in this case. I'm not sure whether that's something you want to consider here? I don't have a strong opinion either way, I'm just raising it as a possibility.

If you want to do this, you need to set `"gradable": false` in `results.json`. See https://prairielearn.readthedocs.io/en/latest/externalGrading/#grading-results

exampleCourse/questions/demo/cGrader/info.json
jonatanschroeder: I changed it to 100. The tests themselves have a timeout as well (default to 1 second each), but you're right, it's good to have another external control on the timeout.

graders/c/cgrader/cgrader.py
jonatanschroeder: Great suggestion, and I even set it as default, and caused an exception to stop other tests from running. Students with code that doesn't compile were focusing on the fact that tests failed instead of noticing the errors.

jonatanschroeder:
@mwest1066 is there anything that needs to be done for the docker container to be built and pushed to docker hub? Or should I assume this is done by script when this PR is pushed to master?

jonatanschroeder:
Still pending: new questions in example course for arguments like `add_c_file` or explicit `run_command`/`add_test_result` cases.

mwest1066:
> @mwest1066 is there anything that needs to be done for the docker container to be built and pushed to docker hub? Or should I assume this is done by script when this PR is pushed to master?

I need to add a build rule on docker hub once the PR is merged (this is a manual step).

graders/c/cgrader/cgrader.py
```diff
@@ -0,0 +1,266 @@
...
+    def __init__(self):
+        pass
+
+class CGrader:
+
+    def __init__(self):
+        with open(DATAFILE) as file:
+            self.data = json.load(file)
+
+    def run_command(self, command, input=None, sandboxed=False, timeout=None):
```
nicknytko, line 26: Any objection to making `sandboxed` default to `True`, and letting users turn it off if needed?  I'd think for most cases we wouldn't need to run the student's code as root?

jonatanschroeder: My initial thought was that `test_run` calls `run_command` with `sandboxed` set to true, and most users are expected to run it from there. But you're right, if the question calls it directly, it probably makes more sense to set it to `False` by default, and change the `cgrader` internal calls (e.g., compilation) to run as root. I'll review this.

jonatanschroeder:
I think this is good to go, any new changes can be done via new issues/PRs, unless @mwest1066 has anything to add.

Issue #3141, opened 2020-09-16
Refactor workspace file watcher


tdy:
Pulled from #2504

> * Rewrite the file watching and upload code in `interface.js` to fix race conditions. An outline of an algorithm is:
>     * Have two dictionaries of files called `update_queue` and `currently_uploading_queue`. The `update_queue` is files that need uploading, while `currently_uploading_queue` is files that are in the process of being uploaded.
>     * Whenever we get a watch notification of a modified file, add it to `update_queue` (this is the current algorithm).
>     * Every N seconds run `_autoUpdateJobManager()`. This takes each file that is on the `update_queue` and not on the `currently_updating_queue`, creates an upload job for that file, removes it from the `update_queue`, and adds it to the `currently_updating_queue`. Then start parallel execution of all the upload jobs. This is similar to the current algorithm, but with the addition of the `currently_updating_queue`.
>     * When a file finishes uploading, we remove it from the currently_updating_queue.
>     * I believe that this algorithm guarantees that we will upload all modified files eventually and we will never try and upload the same file twice at the same time.
>     * For extra safety, before we start upload jobs for files on the `update_queue`, we should check that they are associated with the current `launch_uuid` for a still-running workspace.
>     * Note that the current algorithm uses `[path, isDirectory_str] = key.split(',')` in `interface.js` to store both a filename and whether it's a directory. This will fail if the path contains a comma. We should convert to a more robust method of handling directories. Perhaps store the key in the format `"Dpath"` or `"Fpath"`? The branch #2834 contains a fix for this, but it introduces other problems.

mwest1066:
Superseded by the EFS storage rewrite in #3728

PR #3207, opened 2020-09-28
Refactor the workspace file watcher

polohan:
PR for issue [3141](https://github.com/PrairieLearn/PrairieLearn/issues/3141).

> For extra safety, before we start upload jobs for files on the update_queue, we should check that they are associated with the current launch_uuid for a still-running workspace.

Is not implemented because I believe the overhead is too much.


mwest1066:
This PR was superseded by the EFS storage rewrite in #3728

PR #3210, opened 2020-09-28
Allow config loading from AWS secrets

### LINKED ISSUE #3188: Chunk servers should pull config from AWS secrets
Use AWS secrets for config overrides, like the external grader and workspace hosts.

nicknytko:
This attempts to load config from a secret right after we load the main config file in `server.js`, so we can set almost any value with secrets.

This is an almost 100% copy-paste of the code in the workspace host.  I didn't want to touch that code in this PR, but we should update that to reference the new method in `lib/aws.js`.

Resolves #3188


server.js
```diff
@@ -78,6 +79,16 @@ if (config.startServer) {
         logger.addFileLogging(config.logFilename);
         logger.verbose('activated file logging: ' + config.logFilename);
     }
+
+    /* This synchronous code may seem gross, but we really do want to block
+       the below from running until we are able to grab config values from
+       AWS. */
+    try {
+        deasync(util.callbackify(awsHelper.init))();
+        deasync(util.callbackify(awsHelper.loadConfigSecrets))();
```
mwest1066, line 88: I appreciate the minimally-invasive code change.

As an alternative, could we put all other top-level code below this into a function (placeholder name `topLevelInit()`) and re-arrange the `async.series()` call at the end of `server.js` to do something like:
```
const app = express();
module.exports.app = app;

if (config.startServer) {
    async.series([
        async () => {
            await awsHelper.init();
            await awsHelper.loadConfigSecrets();
        },
        async () => {
            topLevelInit();
        },
        ... all the other existing async.series() blocks
    ]);
}
```

We would need to call `server.topLevelInit()` in `tests/helperServer.js`.

nicknytko: There's also https://itnext.io/node-js-script-writers-top-level-async-await-now-available-a1e86566d5f6 (we may have to update node, though).

I like the idea of cleaning up the main server file, and while we're at it we could change out all the sync code to be async/callback.

nicknytko:
Looks good on staging: https://staging.prairielearn.org/

```
Sep 28 18:03:46 ip-10-0-0-104.us-east-2.compute.internal node[7554]: info: Secret ID: /pl/staging/test
Sep 28 18:03:46 ip-10-0-0-104.us-east-2.compute.internal node[7554]: info: Secret value: {"test":"hello, world!"}
Sep 28 18:03:46 ip-10-0-0-104.us-east-2.compute.internal node[7554]: info: Parsed secret config {"test":"hello, world!"}
```

PR #3212, opened 2020-09-28
Fix missing launched_at in workspace_state_update sproc

### LINKED ISSUE #3211: Workspace 12-hour limits do not reset after reboot
This workspace has a `launched_at` that was never reset after the first time. This causes it to be stopped immediately on re-launch because it looks like it has exceeded the 12-hour limit.

```
prairielearn=> select * from workspaces where id = 5542;
-[ RECORD 1 ]------+-----------------------------------------------------------------------
id                 | 5542
state              | stopped
workspace_host_id  |
hostname           | ip-10-0-0-184.us-east-2.compute.internal:3321
launched_at        | 2020-09-15 21:02:46.795598+00
heartbeat_at       | 2020-09-28 02:19:53.520341+00
launch_uuid        |
launch_port        |
created_at         | 2020-09-15 21:02:21.527893+00
state_updated_at   | 2020-09-28 02:19:24.074981+00
message            | Maximum run time of 12 hours exceeded. Click "Reboot" to keep working.
message_updated_at | 2020-09-28 02:19:24.074981+00
version            | 1
running_at         | 2020-09-28 02:18:58.928638+00
stopped_at         | 2020-09-28 02:19:24.074981+00
rebooted_at        | 2020-09-28 00:54:30.109198+00
reset_at           |
```

tdy:
Fixes #3211 (`launched_at` was mistakenly removed when merging #3135 with master)

### DIFF
--- a/sprocs/workspaces_state_update.sql
+++ b/sprocs/workspaces_state_update.sql
@@ -17,6 +17,7 @@ BEGIN
         state_updated_at = now(),
         message = workspace_message,
         message_updated_at = now(),
+        launched_at = CASE WHEN workspace_state = 'launching' THEN now() ELSE launched_at END,
         running_at = CASE WHEN workspace_state = 'running' THEN now() ELSE running_at END,
         stopped_at = CASE WHEN workspace_state = 'stopped' THEN now() ELSE stopped_at END
     WHERE
--- a/sprocs/workspaces_state_update.sql
+++ b/sprocs/workspaces_state_update.sql
@@ -1,4 +1,5 @@
-DROP FUNCTION IF EXISTS workspaces_state_update(bigint, text, text);
+DROP FUNCTION IF EXISTS workspaces_state_update(bigint, text);
+DROP FUNCTION IF EXISTS workspaces_state_update(bigint, enum_workspace_state, text);
 
 CREATE OR REPLACE FUNCTION
     workspaces_state_update(


mwest1066 (approved):
I think that this:
```DROP FUNCTION IF EXISTS workspaces_state_update(bigint, text, text);```
should actually be:
```
DROP FUNCTION IF EXISTS workspaces_state_update(bigint, enum_workspace_state, text);
```

While we are there, could we add an extra line at the very top like:
```
DROP FUNCTION IF EXISTS workspaces_state_update(bigint, text);
```
(I noticed in the DB that we have an old copy of this sproc lying around with a previous set of arguments)

tdy:
> 
> 
> I think that this:
> `DROP FUNCTION IF EXISTS workspaces_state_update(bigint, text, text);`
> should actually be:
> `DROP FUNCTION IF EXISTS workspaces_state_update(bigint, enum_workspace_state, text);`
> 
> While we are there, could we add an extra line at the very top like:
> `DROP FUNCTION IF EXISTS workspaces_state_update(bigint, text);`
> (I noticed in the DB that we have an old copy of this sproc lying around with a previous set of arguments)

done https://github.com/PrairieLearn/PrairieLearn/pull/3212/commits/80b75bb7516a37abdb78e3b4952a2e462040c684

PR #3214, opened 2020-09-28
Add a Docker Compose File

mebird:
In order to make it easier to pop up a local build of PraireLearn, I've created a docker compose file that emulates the ["running locally with docker"](https://prairielearn.readthedocs.io/en/latest/externalGrading/#running-locally-on-docker) command from the ReadTheDocs.

I've had some issues trying to get alternative start commands to run on windows due to the split terminal--I don't know if this works on other OSes.

There's more comments on `docker-compose` in the comments of #3174.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,19 +1,17 @@
-version: '3.2'
+version: '3.8'
 services:
   pl:
     build:
       context: .
+    image: prairielearn/prairielearn:local
     ports:
       - 3000:3000
     volumes:
       - ./testCourse:/course
       - /var/run/docker.sock:/var/run/docker.sock
       - ${HOME}/pl_ag_jobs:/jobs
+      - .:/PrairieLearn
     container_name: pl
     environment:
       - HOST_JOBS_DIR=${HOME}/pl_ag_jobs
-
-    # Starts the server with a different starting command
-    # Replaces /PrairieLearn/docker/init.sh
-    # This isn't functional on windows machines
-    # command: /PrairieLearn/docker/start_workspace.sh
+      - NODEMON=true
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,19 @@
+version: '3.2'
+services:
+  pl:
+    build:
+      context: .
+    ports:
+      - 3000:3000
+    volumes:
+      - ./testCourse:/course
+      - /var/run/docker.sock:/var/run/docker.sock
+      - ${HOME}/pl_ag_jobs:/jobs
+    container_name: pl
+    environment:
+      - HOST_JOBS_DIR=${HOME}/pl_ag_jobs
+
+    # Starts the server with a different starting command
+    # Replaces /PrairieLearn/docker/init.sh
+    # This isn't functional on windows machines
+    # command: /PrairieLearn/docker/start_workspace.sh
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+      - /var/run/docker.sock:/var/run/docker.sock
+      - ${HOME}/pl_ag_jobs:/jobs
+    container_name: pl
+    environment:
+      - HOST_JOBS_DIR=${HOME}/pl_ag_jobs
+
+    # Starts the server with a different starting command
+    # Replaces /PrairieLearn/docker/init.sh
+    # This isn't functional on windows machines
+    # command: /PrairieLearn/docker/start_workspace.sh
```
tdy, line 19:
```suggestion
```

Hmm I can't get this to work on linux either: `open terminal failed: not a terminal`. Maybe some issue with the script using tmux?

Also now that I see the docs below, I'm realizing this is more for instructors doing course dev, so maybe we should remove this anyway. It might make sense to have another yml for PL dev that includes `command`, `/PrairieLearn` mount, etc.

mebird: It might be. I'm not the most familiar with tmux, so I wasn't sure if it was just a windows config issue.

It it still useful to add this PR in (as in this `docker-compose`, albiet with the command removed) or not?

tdy: Personally I still think this docker-compose (sans `command`) is useful for instructors, but we can see what others think in the dev meeting. Just seems like a simple `docker-compose up` with an editable yml is much easier to digest than a huge shell command.

Maybe we could have a separate yml for PL dev work with stuff like `command`, `-v ./:/PrairieLearn`, etc.

mebird: This docker file builds the pl server from the given dockerfile, so we shouldn't have to copy over the entire root dir into `/PrairieLearn`, but agreed that this is easier to understand than the shell.

I unfortunately won't be able to make the dev meeting today, but I'll be here and in the dev channel, so keep me up to date.

tdy: > This docker file builds the pl server from the given dockerfile, so we shouldn't have to copy over the entire root dir into `/PrairieLearn`

I usually mount `./:/PrairieLearn` so I can hot-reload the node server itself (not the whole container) when the source changes, but maybe there's a better way to do that with docker-compose?

mebird:
@mwest1066 As requested!

mwest1066:
> @mwest1066 As requested!

Thanks!

mwest1066 (approved):

astrosticks:
Is there a way to write our own docker compose file that will work alongside this, in case we want to override something w/o editing a file in the git history? Like an optional .docker-compose-custom.yml that's also gitignored?

mebird:
> Is there a way to write our own docker compose file that will work alongside this, in case we want to override something w/o editing a file in the git history? Like an optional .docker-compose-custom.yml that's also gitignored?

Yup!

https://prairielearn.readthedocs.io/en/latest/docker/#multiple-compose-files
https://docs.docker.com/compose/extends/#understanding-multiple-compose-files

astrosticks:
Looks like the officially supported name would be `docker-compose.override.yml`, and you wouldn't even need to specify it in your command!

PR #3023, opened 2020-09-01
[Group Work] Create template and Leave confirm

frankyhao:
image
- [x] The above section is shared by studentAssessmentHomework and studentAssessmentExam
Putting the code into pages/share/groupWorkInitial.ejs
---
image
- [x] The above section is shared by studentAssessmentHomework, studentAssessmentInstanceHomework, studentAssessmentExam, studentAssessmentInstanceExam
Putting the code into pages/share/groupWorkInfoContainer.ejs
---
image
- [x] Reflected by some professors: there are students left the groups mistakenly. 
Suggested by Issue #2989 adding a confirm message modal when user click "leave the group" at pages/share/groupWorkInfoContainer.ejs
(Please do the language check)

PR #3215, opened 2020-09-28
Fix join code naming bug

tdy:
Tracked down by @andrewstec:

#3023 renamed `joincode` to `joinCode` on the EJS side but not JS.


frankyhao (commented):
Sorry for the typo. Thank you!

andrewstec:
Brilliant. Thank you so much for this quick fix. LGTM!

Issue #3159, opened 2020-09-21
Allow file download and non-ASCII characters in file preview


jonatanschroeder:
In file preview (either `pl-file-preview` or a submitted file in `pl-file-upload`), it is possible to show the text of the file when expanding the list, but that isn't available when the file isn't pure ASCII text. In those cases, there is an error "Unable to decode file." Could you please allow UTF-8 characters to be presented in the preview, when appropriate?

Also, please create a download button that allows the file to be downloaded and checked outside of the interface. This is particularly useful for code files, but can be helpful when other kinds of files are submitted (e.g., PDF or images).

Probably related to #1154 and #1133.

jonatanschroeder:
It seems this is resolved in #6995 by using mime types instead of binary checks. Closing.

PR #3217, opened 2020-09-28
Added download button to pl-file-preview.

jonatanschroeder:
One of the issues in #3159.

The button is currently created with base64 blob in the link itself, which works fine for small enough files, but may cause the page to take longer to load if the file is too big. Alternative would be to create a new URL entrypoint for downloading, but that would probably take more work.

### DIFF
--- a/elements/pl-file-preview/pl-file-preview.mustache
+++ b/elements/pl-file-preview/pl-file-preview.mustache
@@ -16,6 +16,11 @@
               <p class="file-status">uploaded</p>
             </div>
             <div class="file-status-container-right">
+              <a download="{{name}}" class="btn btn-outline-secondary btn-sm"
+                 onclick="event.stopPropagation();"
+                 href="data:application/octet-stream;base64,{{contentsb64}}">
+                 Download
+              </a>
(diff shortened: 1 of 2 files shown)


elements/pl-file-preview/pl-file-preview.mustache
```diff
@@ -16,6 +16,11 @@
               <p class="file-status">uploaded</p>
             </div>
             <div class="file-status-container-right">
+              <a download="{{name}}" class="btn btn-outline-secondary btn-sm"
+                 onclick="event.stopPropagation();"
```
tdy, line 20: Is this `onclick` meant to prevent redirects? Seems to have no effect on my end whether it's there or not (firefox, chrome, edge).

jonatanschroeder: It is to prevent the preview to open or close when download is pressed.

tdy: Ah nice 👍

tdy (approved):
LGTM, tested and working on my end in firefox/chrome/edge.

PR #3229, opened 2020-10-01
[Group Work] Config in .JSON

frankyhao:
Instructor should set information about group size constraints and student permissions in the json format. Please also notice that the group-config at the admin page is no longer available. Example:
```json
{
        "groupWork": true,
        "groupMaxSize": 6,
        "groupMinSize": 2,
        "studentGroupCreate": true,
        "studentGroupJoin": true,
        "studentGroupLeave": true,
}

### CODE REVIEW

**Comment by jonatanschroeder** [comment]
There doesn't seem to be any migration strategy for anyone already using it in production. I will test it later on, but it might be advisable to, for example, update the JSONs if the database already has some configuration for these fields.

Also, since the config page is already there anyway, wouldn't it be better to keep it, and then update the JSON instead of the DB?

**Comment by frankyhao** [comment]
Thanks, @jonatanschroeder 
Yes, you are right, I did not consider database migration for the existing assessments. @mwest1066 Which one do you prefer? Or any other good suggestions?
1. Have a one-time migration reading through the existing databases and update JSON. I am not sure whether we have permission to do that as the JSON is owned by each professor in their repo.
2. Keep the change config function in the user interface. Only synchronized the config info for the first time we create the assessment into the database. In this way, we don't need to worry about the existing assessments.

**Comment by jonatanschroeder** [comment]
> 2. Keep the change config function in the user interface. Only synchronized the config info for the first time we create the assessment into the database. In this way, we don't need to worry about the existing assessments.

I personally like this option better, but I'd suggest having the config update the JSON on modification, so we don't have two conflicting pieces of data (JSON and database) and the user doesn't know which one is used. This could potentially be implemented by having the JSON default, instead of being null/false, be the database setting.

File: sprocs/sync_assessments.sql [hunk 1]
```diff
@@ -174,6 +175,16 @@ BEGIN
...
+            UPDATE group_configs
+            SET 
+                maximum = (valid_assessment.data->>'group_max_size')::bigint,
+                minimum = (valid_assessment.data->>'group_min_size')::bigint,
+                student_authz_create = (valid_assessment.data->>'group_create')::boolean,
+                student_authz_join = (valid_assessment.data->>'group_join')::boolean,
+                student_authz_leave = (valid_assessment.data->>'group_leave')::boolean
+            WHERE
+                course_instance_id = syncing_course_instance_id
+                AND assessment_id = new_assessment_id;
```
mwest1066, line 187: The preceding INSERT and this UPDATE would be better as a single UPSERT, I think? In postgres, this is written as
```
INSERT ...
ON CONFLICT DO UPDATE
...
```

schemas/schemas/infoAssessment.json
```diff
@@ -97,6 +97,31 @@
             "description": "Whether the assessment will support group work.",
             "type": "boolean",
             "default": false
+        },
+        "groupMaxSize": {
+            "description": "Maximum size of a group.",
+            "type": "number",
+            "default": 5
```
mwest1066, line 104: This doesn't seem to be true. It looks like the default is actually `null`? (And similarly for `groupMinSize`.)

mwest1066:
> > 1. Keep the change config function in the user interface. Only synchronized the config info for the first time we create the assessment into the database. In this way, we don't need to worry about the existing assessments.
> 
> I personally like this option better, but I'd suggest having the config update the JSON on modification, so we don't have two conflicting pieces of data (JSON and database) and the user doesn't know which one is used. This could potentially be implemented by having the JSON default, instead of being null/false, be the database setting.

@jonatanschroeder raises an excellent question about the migration plan. In principle I like the idea of having in-browser editors for config options like this, which would then update the JSON to keep everything in sync. However this is quite a lot more complex to implement (see `lib/editor.js` for the current code to do this) and I don't want to hold up the change for this.

@frankyhao I would be happy to merge the PR as is (removing the UI functionality) if you contact everyone using groups and give them a heads up that the change is coming so they can be ready for it. The current list of classes using groups is:
* UIUC: CEE 202, CS 126, CS 233, CS 357, CS 411, CS 421, MSE 404MIC, MSE 584
* GVSU: CIS 452
* UBC: CPSC 313, CPSC 417

This might be something to discuss with @mfsilva22 and the rest of the team before we proceed. I think the basic idea would be that we agree on a date when this change will go live, you contact all the classes and tell them about it, and check whether they will be able to manage this change, and then we merge/deploy the PR on the agreed date.

What do you think?

jonatanschroeder:
> @frankyhao I would be happy to merge the PR as is (removing the UI functionality) if you contact everyone using groups and give them a heads up that the change is coming so they can be ready for it. The current list of classes using groups is:
> 
> * UIUC: CEE 202, CS 126, CS 233, CS 357, CS 411, CS 421, MSE 404MIC, MSE 584
> * GVSU: CIS 452
> * UBC: CPSC 313, CPSC 417
> 
> This might be something to discuss with @mfsilva22 and the rest of the team before we proceed. I think the basic idea would be that we agree on a date when this change will go live, you contact all the classes and tell them about it, and check whether they will be able to manage this change, and then we merge/deploy the PR on the agreed date.
> 
> What do you think?

I like that idea. As for CPSC 417 at UBC, that's mine, and I'm happy to have the settings duplicated (DB and JSON) until this PR takes effect. CPSC 313 is Margo/Don's, and I can help them with the migration if needed.

jonatanschroeder:
Would it be possible to enable to JSON schemas first, so we can change the JSON files, and then in a next phase release the updates?

mwest1066:
> Would it be possible to enable to JSON schemas first, so we can change the JSON files, and then in a next phase release the updates?

Excellent idea! I love it. @frankyhao can you make a separate PR with only the JSON schema changes?

frankyhao:
@mwest1066 I made a few changes to this PR, so it achieves the following functions:
1. remove the interface of group_config button. All instructors should change JSON for group_config for the future.
2. add JSON property to allow the changes 
3. change the corresponding examples & tests
4. the sync now looks like:
`IF (valid_assessment.data->>'group_work')::boolean THEN
            INSERT INTO group_configs
                (course_instance_id,
                assessment_id,
                maximum,
                minimum,
                student_authz_create,
                student_authz_join,
                student_authz_leave
            ) VALUES (
                syncing_course_instance_id,
                new_assessment_id,
                (valid_assessment.data->>'group_max_size')::bigint,
                (valid_assessment.data->>'group_min_size')::bigint,
                (valid_assessment.data->>'student_group_create')::boolean,
                (valid_assessment.data->>'student_group_join')::boolean,
                (valid_assessment.data->>'student_group_leave')::boolean
            ) ON CONFLICT DO NOTHING;`
It will only sync newly added JSON for group_config but will not change the current group_configs in the database. After this PR is merged, we make announcement and give ~7-day period for instructors to change current assessment JSONs (no need to change old JSONs if assessments are no longer active).
After that, I'll make another PR to change the `ON CONFLICT DO UPDATE`, so the JSON will sync with group_configs in the database all the time.

jonatanschroeder:
The issue is that during this seven-day period we can't change the settings for a group exam.

jonatanschroeder:
I suggest moving at least the first item (removing the interface) to the second stage, or do this change during the Winter break. Knowing the people involved in this at UBC, the latter may be preferred.

docs/assessment.md
```diff
@@ -106,6 +106,17 @@ For practice exams it is often desirable to make a *multiple instance* assessmen
 ## Enabling group work for collaborative assessments
 
 By default, assessment instances are tied to only one user. By setting `groupWork: true`, multiple students will be able to work on the same assessment instance.
+Instructor should also set information about group size constraints and student permissions in the json format. Example:
```
mwest1066, line 109:
```suggestion
Information about the group configuration can be set in the `infoAssessment.json` file. For example:
```

docs/assessment.md
```diff
@@ -106,6 +106,17 @@ For practice exams it is often desirable to make a *multiple instance* assessmen
...
+```json
+{
+        "groupWork": true,
+        "groupMaxSize": 6,
+        "groupMinSize": 2,
+        "studentGroupCreate": true,
+        "studentGroupJoin": true,
+        "studentGroupLeave": true,
+}
+```
```
mwest1066, line 119: Can we add a table listing all the options here, their types, default values, and descriptions? This would be similar to the tables of options here: https://prairielearn.readthedocs.io/en/latest/elements/#customizations

sprocs/sync_assessments.sql
```diff
@@ -166,13 +166,24 @@ BEGIN
...
+                student_authz_leave
             ) VALUES (
                 syncing_course_instance_id,
-                new_assessment_id
+                new_assessment_id,
+                (valid_assessment.data->>'group_max_size')::bigint,
+                (valid_assessment.data->>'group_min_size')::bigint,
+                (valid_assessment.data->>'student_group_create')::boolean,
+                (valid_assessment.data->>'student_group_join')::boolean,
+                (valid_assessment.data->>'student_group_leave')::boolean
```
mwest1066, line 186: Can we change this back to always sync from the JSON files to the DB? We want the DB to always be exactly the same as what is in the JSON.

schemas/schemas/infoAssessment.json
```diff
@@ -97,6 +97,31 @@
             "description": "Whether the assessment will support group work.",
             "type": "boolean",
             "default": false
+        },
+        "groupMaxSize": {
+            "description": "Maximum size of a group.",
```
mwest1066, line 102:
```suggestion
            "description": "Maximum number of students in a group.",
```

schemas/schemas/infoAssessment.json
```diff
@@ -97,6 +97,31 @@
...
+            "type": "number",
+            "default": null
+        },
+        "groupMinSize": {
+            "description": "Minimum size of a group.",
```
mwest1066, line 107:
```suggestion
            "description": "Minimum number of students in a group.",
```

mwest1066:
@frankyhao we are now done with semester so it's safe to merge this! Finally! 😄 

I added a few more comments above that should be fixed before merging. Thanks!

docs/assessment.md
```diff
@@ -109,6 +109,25 @@ For practice exams it is often desirable to make a *multiple instance* assessmen
...
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`groupWork` | boolean | false | Enable the group work for the assessment.
+`groupMaxSize` | integer | - | The maximum size of a group.
+`groupMinSize` | integer | - | The minimum size of a group.
```
mwest1066, line 127:
```suggestion
`groupMaxSize` | integer | - | The maximum size of a group (default: no minimum).
`groupMinSize` | integer | - | The minimum size of a group (default: no maximum).
```

pages/studentAssessmentExam/studentAssessmentExam.ejs
```diff
@@ -62,7 +62,7 @@
             certify.onchange = () => {
               start.disabled = !certify.checked;
             };
-
+
```
mwest1066, line 65:
```suggestion

```

pages/studentAssessmentInstanceExam/studentAssessmentInstanceExam.ejs
```diff
@@ -63,7 +63,7 @@
         $(function() {$('#timeLimitExpiredModal').modal({})})
       </script>
     <% } // showTimeLimitExpiredModal %>
-
+
```
mwest1066, line 66:
```suggestion

```

schemas/schemas/infoAssessment.json
```diff
@@ -102,6 +102,31 @@
             "type": "boolean",
             "default": false
         },
+        "groupMaxSize": {
+            "description": "Maximum number of students in a group.",
+            "type": "number",
+            "default": null
```
mwest1066, line 108:
```suggestion
```

schemas/schemas/infoAssessment.json
```diff
@@ -102,6 +102,31 @@
...
+        },
+        "groupMinSize": {
+            "description": "Minimum number of students in a group.",
+            "type": "number",
+            "default": null
```
mwest1066, line 113:
```suggestion
```

sprocs/sync_assessments.sql
```diff
@@ -174,6 +174,17 @@ BEGIN
                 syncing_course_instance_id,
                 new_assessment_id
             ) ON CONFLICT DO NOTHING;
+            -- keep the database same as local
```
mwest1066, line 177: Rather than doing two steps here (an INSERT and then an UPDATE) they should be combined in the UPSERT pattern like this:
```
INSERT INTO group_configs
    (course_instance_id, assessment_id, maximum, minimum, student_authz_create, ...etc...)
VALUES
    (syncing_course_instance_id, new_assessment_id,
    (valid_assessment.data->>'group_max_size')::integer,
    ...etc...
ON CONFLICT (assessment_id) UPDATE
SET
    maximum = EXCLUDED.maximum,
    minimum = EXCLUDED.minimum,
    ...etc...
```

For an example, see https://github.com/PrairieLearn/PrairieLearn/blob/master/middlewares/logPageView.sql#L13

sprocs/sync_assessments.sql
```diff
@@ -167,13 +167,30 @@ BEGIN
...
                 syncing_course_instance_id,
-                new_assessment_id
-            ) ON CONFLICT DO NOTHING;
+                new_assessment_id,
+                (valid_assessment.data->>'group_max_size')::bigint,
+                (valid_assessment.data->>'group_min_size')::bigint,
+                (valid_assessment.data->>'student_group_create')::boolean,
+                (valid_assessment.data->>'student_group_join')::boolean,
+                (valid_assessment.data->>'student_group_leave')::boolean
+            ) ON CONFLICT (assessment_id) WHERE deleted_at IS NULL
```
mwest1066, line 186: Do we need to be more careful with setting and clearing the `deleted_at` flag on the `group_configs` row? I think that if we UPDATE then we should set `deleted_at = NULL`, and if we don't have `group_work` then we should set `deleted_at = now()`?

mwest1066: That is, I think we should remove the `WHERE deleted_at IS NULL` from the line above.

sprocs/sync_assessments.sql
```diff
@@ -167,13 +167,30 @@ BEGIN
...
+            DO UPDATE
+            SET 
+                maximum = EXCLUDED.maximum,
+                minimum = EXCLUDED.minimum,
+                student_authz_create = EXCLUDED.student_authz_create,
+                student_authz_join = EXCLUDED.student_authz_join,
+                student_authz_leave = EXCLUDED.student_authz_leave;
         END IF;
```
mwest1066, line 194: Here I think we need an ELSE clause with `UPDATE group_configs SET delete_at = now() where assessment_id = new_assessment_id`?

sprocs/sync_assessments.sql
```diff
@@ -167,13 +167,30 @@ BEGIN
...
+                (valid_assessment.data->>'student_group_join')::boolean,
+                (valid_assessment.data->>'student_group_leave')::boolean
+            ) ON CONFLICT (assessment_id) WHERE deleted_at IS NULL
+            DO UPDATE
+            SET 
+                maximum = EXCLUDED.maximum,
+                minimum = EXCLUDED.minimum,
+                student_authz_create = EXCLUDED.student_authz_create,
+                student_authz_join = EXCLUDED.student_authz_join,
+                student_authz_leave = EXCLUDED.student_authz_leave;
```
mwest1066, line 193: I think we need to set `deleted_at = NULL` here to make sure we have "undeleted" the group config if it was previously deleted?

sprocs/sync_assessments.sql
frankyhao: For this part, I have to write "WHERE deleted_at IS NULL" here because our UNIQUE_INDEX is on `assessment_id WHERE deleted_at IS NULL`, which allows multiple deleted configs but only one unique current config. It won't let me complier if I don't put the statement here.

mwest1066: Why do we have this constraint (`WHERE deleted_at IS NULL`) on the UNIQUE_INDEX? Could we drop this constraint and have the unique index be directly on `assessment_id`? This would allow us to delete and undelete group_configs as I suggested in the other comments.

mwest1066:
@frankyhao I'm replying to your offline comment here:

> I don't think we could currently support create individual assessments -> switch them to group assessment or vice versa. To achieve that, it requires a lot of extra work on handling already created instances (changing them to untie uid and tie group_id). This is a problem Abdu faced last semester, and he cannot fix it until removed the entire assessment and copy a new one with a new uuid.
> I think currently we should notice professors that they have to decide whether that's individual or group work before creating them. They are allowed to change the min/max number of a group and also student permissions on creating/joining/leaving the group.

This is a good point that we need to be careful with switching an assessment between group and individual modes. I don't think we need to do anything clever here (no need to convert assessment instances) but the sync process should work even if an instructor makes a change.

I think we should:
* Change the unique constraint on `group_configs` to just be on `assessment_id`.
* Make the changes I suggested above so that `group_configs` will be deleted (that is, `SET deleted_at = now()`) or undeleted (that is, set `deleted_at = NULL`) during the sync process, as dictated by the value of `valid_assessment.data->>'group_work'`.
* Don't do anything to `assessment_instances` during sync.
* Add a note to the documentation that changing an assessment from group -> individual or vice versa after students have started working on it will cause student work to be lost.

Questions for you:
* Why do we currently have the `WHERE deleted_at IS NULL` constraint on the unique index for `group_configs.assessment_id`?
* If we make the changes I suggest above, what will happen if an instructor does switch an assessment from group -> individual or vice versa, when there are existing assessment instances?
* Could we improve robustness to switching by finding the places in the code where we have this line: https://github.com/PrairieLearn/PrairieLearn/blob/279b324879eb8928ce89c0a49752927533280365/pages/studentAssessmentHomework/studentAssessmentHomework.sql#L11 and change it to:
```
AND ((gu.user_id = $user_id and NOT a.group_work) OR (ai.user_id = $user_id AND a.group_work));
```

I'm happy to defer the last item above to a future PR if necessary, but I'd like to get the sync code (and the unique index) resolved now.

PR #3231, opened 2020-10-01
[Group Work] Enforce g.deleted_at

### LINKED ISSUE #3077: Access is not correctly blocked with deleted groups
Reported by @DallasTrinkle 

Dallas reported that he made new groups for students but they were able to continue working on their existing assessment instances that were associated with the old (deleted) groups.

If a student starts a group assessment and the instructor deletes their group, they should be blocked from further access to either the assessment instance or the instance questions (or to anything else associated with them).

frankyhao:
fix #3077 

- [x] An order issue at selectAndAuthzInstanceQuestion.sql
- [x] Go over all files and find another potential issue at instructorAssessmentInstance.sql

@mwest1066 Please check whether my understanding is correct:
1. When using LEFT JOIN, the order matters. We should LEFT JOIN group and confirm its deleted_at first, then group_users. This order is critical if I am putting the check on the same line with LEFT JOIN
```
LEFT JOIN groups AS gr ON (gr.id = ai.group_id AND gr.deleted_at IS NULL)
LEFT JOIN group_users AS gu ON (gu.group_id = gr.id)
```
2. The order will not matter when I put the deleted_at check in the WHERE clause
```
LEFT JOIN groups AS gr ON (gr.id = ai.group_id)
LEFT JOIN group_users AS gu ON (gu.group_id = gr.id)
WHERE gr.deleted_at IS NULL
```

### DIFF
--- a/middlewares/selectAndAuthzInstanceQuestion.sql
+++ b/middlewares/selectAndAuthzInstanceQuestion.sql
@@ -65,8 +65,8 @@ FROM
     JOIN course_instances AS ci ON (ci.id = a.course_instance_id)
     JOIN pl_courses AS c ON (c.id = ci.course_id)
     JOIN assessment_sets AS aset ON (aset.id = a.assessment_set_id)
-    LEFT JOIN groups AS gr ON (gr.id = ai.group_id AND gr.deleted_at IS NULL)
-    LEFT JOIN group_users AS gu ON (gu.group_id = gr.id)
+    LEFT JOIN groups AS g ON (g.id = ai.group_id AND g.deleted_at IS NULL)
+    LEFT JOIN group_users AS gu ON (gu.group_id = g.id)
     JOIN users AS u ON (u.user_id = gu.user_id OR u.user_id = ai.user_id)
     LEFT JOIN enrollments AS e ON (e.user_id = u.user_id AND e.course_instance_id = ci.id)
     JOIN LATERAL authz_assessment_instance(ai.id, $authz_data, $req_date, ci.display_timezone, a.group_work) AS aai ON TRUE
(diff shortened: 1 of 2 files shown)


middlewares/selectAndAuthzInstanceQuestion.sql
```diff
@@ -65,8 +65,8 @@ FROM
     JOIN course_instances AS ci ON (ci.id = a.course_instance_id)
     JOIN pl_courses AS c ON (c.id = ci.course_id)
     JOIN assessment_sets AS aset ON (aset.id = a.assessment_set_id)
-    LEFT JOIN group_users AS gu ON (gu.group_id = ai.group_id)
-    LEFT JOIN groups AS gr ON (gr.id = gu.group_id AND gr.deleted_at IS NULL)
+    LEFT JOIN groups AS gr ON (gr.id = ai.group_id AND gr.deleted_at IS NULL)
+    LEFT JOIN group_users AS gu ON (gu.group_id = gr.id)
```
mwest1066, line 69: While we are making this change, can we switch to `g` as the abbreviation for `groups`, to match https://github.com/PrairieLearn/PrairieLearn/blob/5dfb7b3ab2f825f76dfb05aed84a36c3c5598e5c/middlewares/selectAndAuthzAssessmentInstance.sql#L36

frankyhao:
Yeah, I agree with you. The key problem of the original code is that it joins groups depending group_users, so actually, the group_users' rows will not be affected by groups' deleted_at.

middlewares/selectAndAuthzInstanceQuestion.sql
frankyhao: Sure, I am also making all the gr -> g changes in #3157

mwest1066:
> Yeah, I agree with you. The key problem of the original code is that it joins groups depending group_users, so actually, the group_users' rows will not be affected by groups' deleted_at.

Excellent, thanks for confirming!

Issue #3246, opened 2020-10-01
Distinction in tab title/icon between local test and production environment


jonatanschroeder:
It would be great to have an easy way to distinguish, in the tab title, the local testing environment from the production environment. Something as simple as adding `[TEST]` to the start of the title of the page would probably suffice. A different favicon would also be an alternative.

mwest1066:
Great idea! In `pages/partials/navbar.ejs` and children we can check for `config.devMode` which is `true` when running locally and `false` in production. I'm not sure exactly what to put in the navbar, but I'd be happy to merge anything that looks reasonable.

PR #3252, opened 2020-10-01
Separate favicon and title for development mode.

jonatanschroeder:
Solves #3246. It currently works with both a separate favicon and a specific title, if only one is needed it's easy to reverse.


pages/partials/title.ejs
```diff
@@ -60,6 +60,7 @@
     }
 %>
 <title>
+    <% if (config.devMode) { %>[DEVMODE]<% } %>
```
tdy, line 63:
```suggestion
    <% if (config.devMode) { %>[DEV]<% } %>    
```
Or anything shorter, just to preserve as much space as possible for the real title content.

server.js
```diff
@@ -189,7 +189,10 @@ module.exports.initExpress = function() {
     app.use(bodyParser.urlencoded({extended: false, limit: 5 * 1536 * 1024}));
     app.use(cookieParser());
     app.use(passport.initialize());
-    app.use(favicon(path.join(__dirname, 'public', 'favicon.ico')));
+    if (config.devMode)
+        app.use(favicon(path.join(__dirname, 'public', 'favicon-dev.ico')));
```
mwest1066, line 193: Love the icon! 😄

mwest1066 (approved):

Issue #1915, opened 2019-12-09
Allow change of remote repo URL for a course


mwest1066:
At the moment we only set the git remote implicitly when cloning a course repo for the first time. This means that changing a course repo doesn't actually do anything. We should probably force-set the remote URL to the current value before each sync.

PR #3262, opened 2020-10-02
Git remote update

andrewstec:
Closes: #1915, Closes: #3264

- Adds one sync step to the pullAndUpdate() Github sync function to update the Git remote origin address if the Github repo is already cloned on the filesystem. 
- Allows customizability of the default working branch in Prairie Learn (main, master, alternativeName, etc.)

Note: Master/main workflow options can be chosen within the user settings and organization settings.

### New User Level Repository Default Branch Settings
image

### New Organization Level Repository Default Branch Settings:
image

### Default Placeholder on 'branch' columns:
image

While one is not forced to update Prairie Learn with these new features because of the master/main branch defaults can be modified in Github, the option of loading code from any branch and updating the remote repository address is likely still desirable. More detailed main/master change information: https://github.com/github/renaming


andrewstec:
Note: Needs front-end and database entry for default branch (defaults to `master` branch for now and can be updated to `main` if we'd like later)

trombonekenny:
I just force pushed an update because (out of habit) I did a `git rebase master` when starting to look at it normal. Nothing was functionally changed besides the rebase.

andrewstec:
> I just force pushed an update because (out of habit) I did a `git rebase master` when starting to look at it normal. Nothing was functionally changed besides the rebase.

Thank you for the heads-up!

trombonekenny:
Two things from a quick review: It's mixing two features, the "reset remote on sync" (which I know we need) and the larger "be able to specify a branch" feature (which I know is in reaction to master->main in newer repos).

The branch functionality is the bulk of the code (update the editing, schemas, DB, etc.) but I'm not sure it's needed since we never specified master anywhere in the first place. https://git-scm.com/docs/git-clone suggests a clone gets the HEAD pointer from the cloned repository, so as long as the remote was set properly, our clone should reflect that without having to set it?

We do reference the branch name at https://github.com/PrairieLearn/PrairieLearn/blob/808dc9abaa8678f941899ebdc79dbd565a623065/pages/shared/syncHelpers.js#L131 but I suspect we don't have to, if we indeed trust origin/HEAD. I think we can remove the branch from that command and get the same effect.

The only "new" thing we might want to do is make the sync get the HEAD reference from remote (maybe it does this in a fetch, I'm not sure) and make sure we follow it. If so, then we don't need to do anything (new) in PL to configure the branch, we just use what git tells us?

trombonekenny:
I'm not sure if we want to test more around the branch specifying outside of git/HEAD is necessary and trim the PR back, or move the "reset origin" stuff to a different branch/PR.

andrewstec:
@trombonekenny I think that is a really good argument and you are absolutely right. We could indeed go that route and take out some of this code. The branch work was additional value-added work that @mwest1066 requested while I was working in that area. 

I think it comes down to how UIUC envisions this part of the Git repo sync feature. @mwest1066 Were you aware of the git/HEAD possibly being set as another branch? I honestly was not, so it was not something that I had considered.

pages/administratorOverview/administratorOverview.ejs
```diff
@@ -276,6 +277,7 @@
                 <%- include('courseUpdateColumn', {column_name: 'display_timezone', label: 'timezone',   course: course, i: i}) %>
                 <%- include('courseUpdateColumn', {column_name: 'path',             label: 'path',       course: course, i: i}) %>
                 <%- include('courseUpdateColumn', {column_name: 'repository',       label: 'repository', course: course, i: i}) %>
+                <%- include('courseUpdateColumn', {column_name: 'branch',           label: 'branch',     course: course, i: i,    placeholder: "Defaults to '" + config.cloneCourseDefaultBranch + "'"}) %>
```
mwest1066, line 280: I like the way you implemented `placeholder` logic in these forms. However, I'd rather just treat `branch` like all the other fields and require explicit text. Could we remove all the `placeholder` logic?

pages/administratorOverview/courseInsertForm.ejs
```diff
@@ -29,6 +29,10 @@
     <label for="courseAddInputRepository">Repository:</label>
     <input type="text" class="form-control" id="courseAddInputRepository" name="repository" value="git@github.com:PrairieLearn/pl-XXX.git">
   </div>
+  <div class="form-group">
+    <label for="courseAddInputBranch">Branch:</label>
+    <input type="text" class="form-control" id="courseAddInputBranch" name="branch" placeholder="Defaults to '<%= config.cloneCourseDefaultBranch %>'">
```
mwest1066, line 34: Rather than using a `placeholder`, this can just be a `value`.

pages/shared/syncHelpers.js
```diff
@@ -43,6 +43,7 @@ module.exports.pullAndUpdate = function(locals, callback) {
         // update the content.
 
         const syncStage1A = function() {
+            const branch = locals.course.branch || config.cloneCourseDefaultBranch;
```
mwest1066, line 46: In all of this code, we can assume that `locals.course.branch` is valid, so we can drop all the `||` logic.

pages/administratorOverview/administratorOverview.ejs
andrewstec: Sounds good. Done and tested. No issues.

pages/shared/syncHelpers.js
andrewstec: Done. Tested. No issues.

andrewstec:
@mwest1066 The placeholder logic has been removed. The fallback to the config default if locals.course.branch is not defined is removed-- except for one instance in the sync file.

If we are happy with the syncing failing if the branch is not filled out by the administrator, then we can remove the last instance of the `||` logic. Also, we can remove `cloningDefaultBranch` from the `config.js` file.

andrewstec:
> @mwest1066 The placeholder logic has been removed. The fallback to the config default if locals.course.branch is not defined is removed-- except for one instance in the sync file.
> 
> If we are happy with the syncing failing if the branch is not filled out by the administrator, then we can remove the last instance of the `||` logic. Also, we can remove `cloningDefaultBranch` from the `config.js` file.

I just took it out.

pages/administratorOverview/courseInsertForm.ejs
```diff
@@ -29,6 +29,10 @@
     <label for="courseAddInputRepository">Repository:</label>
     <input type="text" class="form-control" id="courseAddInputRepository" name="repository" value="git@github.com:PrairieLearn/pl-XXX.git">
   </div>
+  <div class="form-group">
+    <label for="courseAddInputBranch">Branch:</label>
+    <input type="text" class="form-control" id="courseAddInputBranch" name="branch">
```
mwest1066, line 34:
```suggestion
    <input type="text" class="form-control" id="courseAddInputBranch" name="branch" value="<%= config.courseRepoDefaultBranch %>>
```

mwest1066:
> > @mwest1066 The placeholder logic has been removed. The fallback to the config default if locals.course.branch is not defined is removed-- except for one instance in the sync file.
> > If we are happy with the syncing failing if the branch is not filled out by the administrator, then we can remove the last instance of the `||` logic. Also, we can remove `cloningDefaultBranch` from the `config.js` file.
> 
> I just took it out.

Oh sorry, I meant that we should have `config.courseRepoDefaultBranch` and use it as the `value` in `courseInsertForm.ejs`. I only meant that we should not use it as a `placeholder`. That is, I think we should now make two more changes:

1. Add `config.courseRepoDefaultBranch` to `config.js` (this used to be called `config.cloneCourseDefaultBranch`)
2. Add the `value` default as in my suggested code edit above.

All of the other placeholder logic can stay removed.

mwest1066:
Super minor nit: GitHub doesn't support "closes" with multiple issue numbers (like `Closes: #1915, #3264`). Instead you need to write `Closes: #1915, Closes: #3264`.

mwest1066:
> Super minor nit: GitHub doesn't support "closes" with multiple issue numbers (like `Closes: #1915, #3264`). Instead you need to write `Closes: #1915, Closes: #3264`.

After you've edited the PR description, you can check that GitHub detected both issues by looking in the right-hand column under "Linked Issues".

mwest1066 (approved):

PR #3277, opened 2020-10-04
Fix headers structure in docs/pl-drawing

mwest1066:


tdy (approved):
Depending on how nitpicky we want to be:

* `Example Implementations` and `Sample Element` are currently title case
* Technically I think `Sample Element` and `Customizations` should be bumped to h3 instead of dropping `Example Implementations` to h4

mwest1066:
> * `Example Implementations` and `Sample Element` are currently title case

Please push a fix for this. Thanks!

> * Technically I think `Sample Element` and `Customizations` should be bumped to h3 instead of dropping `Example Implementations` to h4

I used h4 only because the main `elements.md` uses that structure and I wanted to bring these two into line. We should either fix both or neither, and I don't have a strong preference either way.

docs/pl-drawing/index.md
```diff
@@ -1360,6 +1366,8 @@ Attribute | Type | Default | Description
 `color` | string | red | Set the color of the coordinate system ( [PL colors](https://prairielearn.readthedocs.io/en/latest/course/#colors) or [HTML colors](https://htmlcolorcodes.com/color-chart/) ).
 `stroke-width` | float | 4 | Set the width of the stroke.
 
+## Example drawing questions
+
```
tdy, line 1370:
```suggestion
```
I think these are just named reference links, not user-visible links.

tdy:
> Please push a fix for this. Thanks!

Done

> I used h4 only because the main `elements.md` uses that structure and I wanted to bring these two into line. We should either fix both or neither, and I don't have a strong preference either way.

I touched up the hierarchy in both `elements.md` and `pl-drawing/index.md`, so they should both be consistent and (more) accurate now.

Issue #3191, opened 2020-09-24
Continue to allow resubmissions after scoring 100


echuber2:
It seems that when students get 100 on a coding problem, they can't submit anymore. This has caused various issues for students psychologically or for staff logistically (students can't resubmit their _preferred_ version, staff can't refer to the version the student _meant_ to submit, etc.). If a question has manual review involved, this can be a problem.

Can we add an option to allow continued saving and submitting even when 100 is achieved?

(I thought there was already a feature request open for this, but I can't find it now.)

jonatanschroeder:
Would that be #2083?

mwest1066:
> Would that be #2083?

Oh wow! Nice find of a closed issue. I'm impressed!

echuber2:
So it sounds like maybe there is already a workaround for this, but probably not documented very clearly.

mwest1066:
> So it sounds like maybe there is already a workaround for this, but probably not documented very clearly.

Does that workaround satisfy the needs of the courses that you have encountered? If so, perhaps you could make a docs PR to describe this?

echuber2:
I'm not sure; I'm just referring to the trick you had mentioned on https://github.com/PrairieLearn/PrairieLearn/issues/2083#issuecomment-605526479 (also discussed on Slack a few weeks ago):

> If the instructor does not want to award max points for a particular answer, they can simply change the grader to give less than 100%.

This _can_ work, but I think Jonatan and others pointed out that it's not great psychologically for the students (they get upset not seeing 100%, and they file issues or support requests, etc.). So maybe we could still use an option to just let students resubmit even after getting full score.

I'll make a note in the docs about the trick, though.

echuber2:
@frankyhao Hi, it looks like you has some merges that referenced this. Is the issue resolved in a planned update?

jonatanschroeder:
> @frankyhao Hi, it looks like you had some merges that referenced this. Is the issue resolved in a planned update?

@echuber2 if you check the history it looks like the references are only links to your documentation in merges, not an implementation.

echuber2:
Ah, you're right. Thanks.

joelostblom:
[...]

PR #3285, opened 2020-10-06
Add doc note about preventing 100 credit from locking question

echuber2:
As noted in https://github.com/PrairieLearn/PrairieLearn/issues/3191 it would be ideal to get a real option for this rather than a workaround...


jonatanschroeder:
Wouldn't it be better to include this comment in the external grader docs? I don't think this kind of issue would arise often in a question with internal grading.

echuber2:
It could happen on an internal question if `server.py` is customized and the input style of the question invites experimentation or has a style component. Certainly the note could be placed in both places or linked from the other place.

PR #3287, opened 2020-10-06
Add glob support for workspace file options

tdy:
* [x] add `questions.workspace_required_file_names`
    * this stores the results from globbing `gradedFiles` to avoid re-globbing at multiple points in the grading pipeline
* [x] update `syncIgnore` and `gradedFiles` to accept globs
* [x] update docs


echuber2:
A few things that probably have to be considered, maybe as part of a redesign of the upload widgets on the question page: currently the upload widgets require/preview individual files. If staff explicitly enumerate only a few required files they really want to collect, but also use a glob to collect more stuff than that, are the extra files meant to be collected anyway? There's some convenience in being able to specify a directory name or a glob to pick up multiple files, but if staff need to worry about vacuuming up unwanted files and then filtering them in the grader, is it an antipattern? If staff use globs indiscriminately for unneeded files (maybe preferring to filter in the autograder rather than edit the json separately), would there be significant waste of storage/bandwidth on the server?

PR #3303, opened 2020-10-07
Fix allowRealTimeGrading:false bug to allow 1-element points arrays

### LINKED ISSUE #2211: allowRealTimeGrading:false rejects questions with length-1 points arrays
When we sync an assessment with `"allowRealTimeGrading": false`, we check to make sure it doesn't contain questions with arrays of points for multiple attempts, like this:
```
"points": [10, 8, 6, 4]
```

We currently check for this by insisting that `points` should be a scalar (non-array) value, like `"points": 10`. But some instructors are leaving the brackets, like `"points": [10]`, and are confused about why this also triggers an error.

We should fix this by allowing a length-1 array as well, and improve the error message to only talk about arrays with more than one points value (or something like that).

echau01:
Closes #2211 

* Add unit tests and edit titles of tests for allowRealTimeGrading:false used with arrays

* Update error messages when a points array with multiple elements is used in this case

### DIFF
--- a/sync/course-db.js
+++ b/sync/course-db.js
@@ -895,8 +895,8 @@ async function validateAssessment(assessment, questions) {
     };
     (assessment.zones || []).forEach(zone => {
         (zone.questions || []).map(zoneQuestion => {
-            if (!allowRealTimeGrading && Array.isArray(zoneQuestion.points)) {
-                errors.push(`Cannot specify an array of points for a question if real-time grading is disabled`);
+            if (!allowRealTimeGrading && Array.isArray(zoneQuestion.points) && zoneQuestion.points.length > 1) {
+                errors.push(`Cannot specify more than 1 point value for a question if real-time grading is disabled`);
             }
             // We'll normalize either single questions or alternative groups
             // to make validation easier
@@ -907,8 +907,8 @@ async function validateAssessment(assessment, questions) {
             } else if ('alternatives' in zoneQuestion) {
                 zoneQuestion.alternatives.forEach(alternative => checkAndRecordQid(alternative.id));
                 alternatives = zoneQuestion.alternatives.map(alternative => {
-                    if (!allowRealTimeGrading && Array.isArray(alternative.points)) {
-                        errors.push(`Cannot specify an array of points for an alternative if real-time grading is disabled`);
+                    if (!allowRealTimeGrading && Array.isArray(alternative.points) && alternative.points.length > 1) {
+                        errors.push(`Cannot specify more than 1 point value for an alternative if real-time grading is disabled`);
                     }
                     return {
                         points: alternative.points || zoneQuestion.points,
--- a/sync/course-db.js
+++ b/sync/course-db.js
@@ -896,7 +896,7 @@ async function validateAssessment(assessment, questions) {
     (assessment.zones || []).forEach(zone => {
         (zone.questions || []).map(zoneQuestion => {
             if (!allowRealTimeGrading && Array.isArray(zoneQuestion.points) && zoneQuestion.points.length > 1) {
-                errors.push(`Cannot specify an an array of multiple point values for a question if real-time grading is disabled`);
+                errors.push(`Cannot specify an array of multiple point values for a question if real-time grading is disabled`);
             }
             // We'll normalize either single questions or alternative groups
             // to make validation easier
--- a/sync/course-db.js
+++ b/sync/course-db.js
@@ -896,7 +896,7 @@ async function validateAssessment(assessment, questions) {
     (assessment.zones || []).forEach(zone => {
         (zone.questions || []).map(zoneQuestion => {
             if (!allowRealTimeGrading && Array.isArray(zoneQuestion.points) && zoneQuestion.points.length > 1) {
-                errors.push(`Cannot specify more than 1 point value for a question if real-time grading is disabled`);
+                errors.push(`Cannot specify an an array of multiple point values for a question if real-time grading is disabled`);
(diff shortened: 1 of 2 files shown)


sync/course-db.js
```diff
@@ -895,8 +895,8 @@ async function validateAssessment(assessment, questions) {
     };
     (assessment.zones || []).forEach(zone => {
         (zone.questions || []).map(zoneQuestion => {
-            if (!allowRealTimeGrading && Array.isArray(zoneQuestion.points)) {
-                errors.push(`Cannot specify an array of points for a question if real-time grading is disabled`);
+            if (!allowRealTimeGrading && Array.isArray(zoneQuestion.points) && zoneQuestion.points.length > 1) {
+                errors.push(`Cannot specify more than 1 point value for a question if real-time grading is disabled`);
```
mwest1066, line 899: I'm slightly worried about the confusion between "1 point" meaning "this question is worth 1 point" and "1 point value" meaning "not an array with length more than one". Maybe it would be clearer to use:
```suggestion
                errors.push(`Cannot specify an array of multiple point values for a question if real-time grading is disabled`);
```

tests/sync/assessmentsSync.js
```diff
@@ -485,10 +485,34 @@ describe('Assessment syncing', () => {
...
+        id: util.QUESTION_ID,
+        points: [5],
+      }, {
+        id: util.ALTERNATIVE_QUESTION_ID,
+        points: [10],
+      }],
+    }];
+    courseData.courseInstances[util.COURSE_INSTANCE_ID].assessments['points_array_size_one'] = assessment;
+    await util.writeAndSyncCourseData(courseData);
+    const syncedAssessment = await getSyncedAssessmentData('points_array_size_one');
```
mwest1066, line 507: Maybe we should explicitly check that there weren't any sync errors for this assessment, like this: https://github.com/PrairieLearn/PrairieLearn/blob/400677ab19fa544354138862ab00124369b52d3d/tests/sync/assessmentsSync.js#L646

tests/sync/assessmentsSync.js
```diff
@@ -507,7 +531,34 @@ describe('Assessment syncing', () => {
...
+          id: util.QUESTION_ID,
+        }, {
+          id: util.ALTERNATIVE_QUESTION_ID,
+          points: [5],
+        }],
+      }],
+    }];
+    courseData.courseInstances[util.COURSE_INSTANCE_ID].assessments['points_array_size_one'] = assessment;
+    await util.writeAndSyncCourseData(courseData);
+    const syncedAssessment = await getSyncedAssessmentData('points_array_size_one');
```
mwest1066, line 555: Also check for no sync errors?

sync/course-db.js
```diff
@@ -907,8 +907,8 @@ async function validateAssessment(assessment, questions) {
             } else if ('alternatives' in zoneQuestion) {
                 zoneQuestion.alternatives.forEach(alternative => checkAndRecordQid(alternative.id));
                 alternatives = zoneQuestion.alternatives.map(alternative => {
-                    if (!allowRealTimeGrading && Array.isArray(alternative.points)) {
-                        errors.push(`Cannot specify an array of points for an alternative if real-time grading is disabled`);
+                    if (!allowRealTimeGrading && Array.isArray(alternative.points) && alternative.points.length > 1) {
+                        errors.push(`Cannot specify more than 1 point value for an alternative if real-time grading is disabled`);
```
mwest1066, line 911: Same wording comment as above.

mwest1066 (commented):
Thanks @echau01 for a really nice PR! I very much appreciate that you figured out the unit tests and added tests for the new cases.

tests/sync/assessmentsSync.js
```diff
@@ -507,7 +535,37 @@ describe('Assessment syncing', () => {
     courseData.courseInstances[util.COURSE_INSTANCE_ID].assessments['fail'] = assessment;
     await util.writeAndSyncCourseData(courseData);
     const syncedAssessment = await findSyncedAssessment('fail');
-    assert.match(syncedAssessment.sync_errors, /Cannot specify an array of points for an alternative/);
+    assert.match(syncedAssessment.sync_errors, /Cannot specify an array of multiple point values for an alternative/);
+  });
+
+  it('accepts a points array of size 1 being specified for an alternative when real-time grading is disallowed', async () => {
```
tdy, line 541:
```suggestion
  it('accepts a single-element points array being specified for an alternative when real-time grading is disallowed', async () => {
```
nit: mirroring the multiple-element wording

tests/sync/assessmentsSync.js
```diff
@@ -485,10 +485,38 @@ describe('Assessment syncing', () => {
     courseData.courseInstances[util.COURSE_INSTANCE_ID].assessments['fail'] = assessment;
     await util.writeAndSyncCourseData(courseData);
     const syncedAssessment = await findSyncedAssessment('fail');
-    assert.match(syncedAssessment.sync_errors, /Cannot specify an array of points for a question/);
+    assert.match(syncedAssessment.sync_errors, /Cannot specify an array of multiple point values for a question/);
   });
 
-  it('records an error if points array is specified for an alternative when real-time grading is disallowed', async () => {
+  it('accepts a points array of size 1 being specified for a question when real-time grading is disallowed', async () => {
```
tdy, line 491:
```suggestion
  it('accepts a single-element points array being specified for a question when real-time grading is disallowed', async () => {
```
nit: mirroring the multiple-element wording

PR #3315, opened 2020-10-10
Add config.instanceIdEc2Override to control EC2 server naming

mwest1066:


nicknytko (approved):
Looks good.  I'm debating if we should have it after we load the EC2 secret, but that will override the config value anyway (but with the `instanceId` key instead of `instanceIdEc2Override`).

nicknytko:
Actually I don't think we need to worry about that.  Lets merge it.

PR #3321, opened 2020-10-10
Fix more editor crashes where existing objects don't have long names

mwest1066:


nicknytko (approved):
Looks good, this seems to cover all the other long names in the editor code.  I'm not overly familiar with the changes the new sync code brought, is there anything else we should be concerned about being null?

mwest1066:
> Looks good, this seems to cover all the other long names in the editor code. I'm not overly familiar with the changes the new sync code brought, is there anything else we should be concerned about being null?

Good question. The new sync code ("robust" syncing) means that we will sync basically anything, so the only thing that guaranteed to be non-NULL are the "short name" identifiers for objects (that is, QIDs for questions, AIDs/TIDs for assessments, short_names for course_instances). The short names must exist and be unique because these are the directories on disk, so we definitely have them and they definitely don't conflict 😄 

I don't really have a good idea about how to (easily) audit the rest of the code for non-NULL assumptions. I'll make an issue anyway.

PR #3323, opened 2020-10-11
Fix flapping during external grader autoscaling

mwest1066:
The external grader fleet autoscaling code using multiple metrics to
estimate the desired number of graders. Most of these take a `max()`
over recent historical data and so are not subject to "flapping" (rapid
upscaling followed by immediate downscaling). However, we also use the
instantaneous number of ungraded jobs to compute the desired number of
graders, which can rapidly fluctuate. To avoid the flapping that this
can cause, this PR unconditionally latches the desired number of graders
to a `max()` over the recent past. This will avoid flapping due to
upscaling-then-downscaling. We can still have downscaling-then-upscaling
flapping, but we want that because it's caused by the need to respond to
actual load increases.

### DIFF
--- a/sprocs/grader_loads_current.sql
+++ b/sprocs/grader_loads_current.sql
@@ -232,8 +232,8 @@ BEGIN
         ('desired_instances_by_history_jobs', desired_instances_by_history_jobs),
         ('desired_instances_by_current_users', desired_instances_by_current_users),
         ('desired_instances_by_history_users', desired_instances_by_history_users),
-        ('desired_instances_current', desired_instances_current);
-        ('desired_instances_history', desired_instances_history);
+        ('desired_instances_current', desired_instances_current),
+        ('desired_instances_history', desired_instances_history),
         ('desired_instances', desired_instances);
 
     -- ######################################################################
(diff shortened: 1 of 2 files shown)


sprocs/grader_loads_current.sql
```diff
@@ -180,6 +182,18 @@ BEGIN
...
 
+    -- ######################################################################
+    -- anti-flapping using historical information
+
+    SELECT coalesce(max(value), 0)
+    INTO desired_instances_history
+    FROM time_series
+    WHERE
+        name = 'desired_instances'
+        AND date > now() - history_interval;
```
nicknytko, line 193: How big will this `history_interval` be in practice?  Something in the range of a few seconds?

mwest1066: It's the existing value which is currently https://github.com/PrairieLearn/PrairieLearn/blob/19c59baa89e3786b7cc02fc7db3a1a26207d5398/lib/config.js#L84

This is too long, but at least it will reduce the flapping. We should change the grader autoscaler algorithm more, so this PR is just a stop-gap measure.

There are a number of other changes that I want to make to the grader hosts to bring them into line with workspace hosts. Things like clearer state tracking, better websocket communication back to the end user, better logging, etc. I need to file issues for all these things.

PR #3328, opened 2020-10-12
Allow admins to force generate course chunks

### LINKED ISSUE #3280: Chunk creation from admin panel
The admin panel has code to force-generate chunks for a list of courses. This has never been tested, or even executed 😄

nicknytko:
Fixes up the chunk generation code that @mwest1066 wrote into a working state on staging (which basically worked as-is, so kudos)

Also adds some chunk admin queries so that we can see which courses have chunks generated.

Resolves #3280

### DIFF
--- a/lib/chunks.sql
+++ b/lib/chunks.sql
@@ -37,4 +37,4 @@ WHERE
 -- BLOCK select_course_dir
 SELECT c.path
 FROM pl_courses AS c
-WHERE c.id = ${course_id};
+WHERE c.id = $course_id;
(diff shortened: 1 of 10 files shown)


lib/chunks.sql
```diff
@@ -37,4 +37,4 @@ WHERE
 -- BLOCK select_course_dir
 SELECT c.path
 FROM pl_courses AS c
-WHERE c.id = ${course_id};
```
mwest1066, line 40: Haha, too many backtick string templates in JS 😄

mwest1066 (approved):

PR #3329, opened 2020-10-12
Add RFC for question regrading

polohan:


rfcs/006-regrading.md
```diff
@@ -0,0 +1,39 @@
...
+# Proposed solution
+
+## Backend
+
+### Overview
+We can modify the current [regrading](../lib/regrading.js) system and add the option to rerun the internal/external grader. The code in the backend that runs the grader can be found in [question.js](../lib/question.js). We can also add a new method in this file to handle the regrade functionality.
+
+### Workflow (for regrading a question for one student)
+1.  Get all the submissions and variants of a student of a question
+2.  Rerun the grader and get the new score (and stop early if the score reach 100%)
```
cheeren, line 26: All this time, I haven't been thinking about the ASSESSMENT scoring function. At what point of grading do the "multiple attempts" come into play? 

Behavior:
1) orig early attempt was marked correct, but was actually incorrect? (ans: leave it as correct -- don't decrease points) 
2) orig early attempt was marked incorrect, but it was actually correct?

rfcs/006-regrading.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Frontend
+We want to provide more flexibilities to the instructors by allowing them to regrade a question for a single student and whole class. However, since we already had the "Regrading" tab in instructor assessment page, there is more than one way to update the UI.
+* Regrade a question for whole class
+    *   We could keep the "Regrading" tab and remake the regrading page by adding a new section that contains a list of every questions in this assessment (pretty much like the assessment questions page). And the instructor can choose to regrade a question for the whole class using a button.
+    *   We could remove the "Regrading" tab and add the option to regrade at assessment questions page by adding a new column that trigger the regrade options.
+* Regrade a question for a student (remove the "Regrade" option from "Action" menu in assessment instance page)
+    *   In the assessment_instance, add a new column to "Questions" table that contains a regrade button.
+    *   Move it to the "Regrading" page by adding a new section.
+
```
cheeren, line 36: Also want some indication of *when* a regrade was run on a question. Do we also need a grading status -- "pending regrade", or some such, while awaiting the external grader (or manual grading??)?

rfcs/006-regrading.md
```diff
@@ -0,0 +1,39 @@
...
+* Instructor can go to the assessment questions page to regrade a question for the whole class.
+* The regrading job should take all the user submissions and rerun grader on all of them in the order as they come in to calculate the new score.
+* The instructor can choose between whether to replace the old score with a lower new score or not. 
+
+# Proposed solution
+
+## Backend
+
+### Overview
+We can modify the current [regrading](../lib/regrading.js) system and add the option to rerun the internal/external grader. The code in the backend that runs the grader can be found in [question.js](../lib/question.js). We can also add a new method in this file to handle the regrade functionality.
```
cheeren, line 22: From an instructor's perspective, the "regrade" means that the server.py grade(data) function is run on all submitted answers. Setting data["correct_answers"] occurs in generate(data), so "regrade" does *not* mean resetting data["correct_answers"], nor making changes to html. I think we should make grade(data) templates for fixing MC and checkbox questions.

rfcs/006-regrading.md
mwest1066: The current grading setup has three levels to it, which I think is good:
1. Grade the `submission`: This runs external or internal grading code and computes a score between 0 and 1. This is idempotent, assuming the question grading code hasn't changed.
2. Grade the `instance_question`: This runs https://github.com/PrairieLearn/PrairieLearn/blob/master/sprocs/instance_questions_grade.sql which takes `submission_score` (0 to 1) and updates `instance_question.points` based on the current number of attempts and the multiple-attempt score values. At the moment this is _not_ idempotent, because it assumes the `submission_score` is a new attempt. For example, it will always increment `number_attempts`.
3. Grade the `assessment_instance`: This runs https://github.com/PrairieLearn/PrairieLearn/blob/master/sprocs/assessment_instances_grade.sql and basically just adds up all the question points to get a total score (taking into account things like best-M-of-N-questions and max_points limits). This is idempotent.

To regrade, I think we want to re-run step 1 on each submission from the history. Step 3 above is straightforward, because it's idempotent and we just need to re-run this every time we change anything about the grading. The tricky bit is all about step 2 (updating `instance_question` points).

It seems to me that there are (at least) two strategies we could use for regrading:
1. Reset the instance question grading information to its initial state (e.g., set `number_attempts = 0`). Regrade all the submissions sequentially in strict historical order, and call `instance_questions_grade()` once per graded submission in order. This will require being careful about the order that we grade submissions in.
2. Change `instance_questions_grade()` to accept a `(submission_score, attempt_number)` pair, where `attempt_number` doesn't have to increase on each call. For example, during a regrade we'll call this function again with `attempt_number = 1` again, etc, and these calls might come out of order. This simplifies the regrading order (we don't care about the order) but will make `instance_questions_grade()` more complex.

@cheeren to address the "stinks to be ... someone who stopped because they thought they got it right" issue: I think this is why we should default to "don't decrease existing points" during a regrade. It's really unfair (and _extremely_ unpopular) to take points away from students after they've been given.

To address the "stinks to be ... someone who kept trying (and using time) because the system said they were wrong" issue: I think this is really hard. I don't know of any way to fairly address this other than "don't write broken questions in the first place" 😄. Once there is a broken question on an exam, we are firmly in the realm of damage mitigation and so all we can do is provide controls for instructors to mitigate as best they can. At least that's my current feeling. I'd love to know if there are other things that can be done in this situation.

rfcs/006-regrading.md
mwest1066: As an alternative for MC/checkbox questions, we could consider re-reading the `correct` status for each answer from `question.html` in the `grade()` function. At the moment we store this information in `params` during the `generate()` phase, which is why instructors can't change it later.

The main issue with re-reading `correct` status from the HTML during grading is how to match answers from `params` to those in the HTML. We could do this by index (answer-1, answer-2, etc) or by fuzzy-string match?

nwalters512:
This hasn't seen any activity in years, and it looks like @polohan is no longer active in PrairieLearn development. We'll most likely start over with a new RFC when we're ready to implement regrading. I'm going to close this PR.

Issue #2266, opened 2020-03-29
Add `comments` attribute to all objects in all JSON schemas


mwest1066:
JSON doesn't allow comments, which stops people making notes in their configuration files.

As suggested by @zillesc, we can mitigate this by adding a `comments` property to all objects in all of our schemas.

See https://stackoverflow.com/a/244858

PR #3338, opened 2020-10-13
PR for json-schema-comments

xinhexu2:
Adding the property `comment` to json files. This PR closes issue #2266


docs/course.md
```diff
@@ -77,6 +79,7 @@ Each assessment belongs to an *assessment set*. Each assessment set must have th
 
 Property | Description
 --- | ---
+`comment` | The comments for each property.
```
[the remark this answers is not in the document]
mfsilva22: Instead, do the following:

- move the section `##Timezone` before `## Colors`
- between `Timezone` and `Colors`, add a section named `## Comment`

## Comment

You can add comments to `json` files using the property `comment`. You can use this property more than once in the same file. For example:

```json
{
    "comment": "select the timezone for the course",
    "timezone": "America/New_York"
}
```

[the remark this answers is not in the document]
mfsilva22: I fixed the comment above

schemas/schemas/adminQuery.json
```diff
@@ -8,6 +8,10 @@
         "description"
     ],
     "properties": {
+        "comment": {
+            "description": "Add the comment",
+            "type": "string"
```
mwest1066, line 13: I think we should allow comments to also be arrays or objects:
```suggestion
            "type": ["string", "array", "object"]
```

schemas/schemas/adminQuery.json
```diff
@@ -8,6 +8,10 @@
         "description"
     ],
     "properties": {
+        "comment": {
+            "description": "Add the comment",
```
mwest1066, line 12:
```suggestion
            "description": "Arbitrary comment for reference purposes.",
```

rfcs/005-json-comments-schemas.md
```diff
@@ -0,0 +1,13 @@
+# Summary
```
mwest1066, line 1: We can delete this file -- it was helpful while writing the PR but I think it's no longer necessary.

news_items/021_Json_schema_comment/index.html
```diff
@@ -0,0 +1,41 @@
+<style>
```
mwest1066, line 1: Please rename the news item directory to use a lower-case `j`

docs/question.md
```diff
@@ -54,7 +54,8 @@ The `info.json` file for each question defines properties of the question. For e
     "title": "Newton's third law",
     "topic": "Forces",
     "tags": ["secret", "Fa18"],
-    "type": "v3"
+    "type": "v3",
+    "comment": "You can add comments to json files using this property."
```
mwest1066, line 58:
```suggestion
    "comment": "You can add comments to JSON files using this property."
```

docs/course.md
```diff
@@ -217,3 +213,24 @@ The default timezone for courses is `America/Chicago` (U.S. Central Time). This
 ```
 
 Allowable timezones are those in the TZ column in the [list of tz database time zones](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones), which is a display version of the [IANA Time Zone Database](https://www.iana.org/time-zones).
+
+## Comment
```
mwest1066, line 217:
```suggestion
## Comments in JSON files
```

docs/course.md
```diff
@@ -217,3 +213,24 @@ The default timezone for courses is `America/Chicago` (U.S. Central Time). This
...
+
+You can add comments to json files using the property `comment`. You can use this property more than once in the same file. For example:
```
mwest1066, line 219:
```suggestion
You can add comments to JSON files using the property `comment` on any object. You can only use this property once for each object. For example:
```

docs/course.md
```diff
@@ -217,3 +213,24 @@ The default timezone for courses is `America/Chicago` (U.S. Central Time). This
...
+
+```json
+{
+    "comment": "assessments that are shared among a group of students",
+    "assessmentSets": [
+      {"abbreviation": "GA", "name": "Group Activity", "heading": "Group Activities", "color": "green1"}
+    ],
+    "comment": "timezone for the course",
```
mwest1066, line 227: We should delete this the second comment here. Although PL currently allows more than one comment per object, it actually silently drops all the comments except the last one if the file is automatically edited.

I don't currently have a good way to catch JSON files with repeated keys (like more than one comment in the same object). `JSON.parse()` silently drops all earlier copies of a key. As does `jju` that we use.

docs/course.md
```diff
@@ -217,3 +213,24 @@ The default timezone for courses is `America/Chicago` (U.S. Central Time). This
...
+    "timezone": "America/New_York"
+}
+```
+
+## Colors
```
mwest1066, line 232: Why was this shifted?

docs/course.md
```diff
@@ -52,10 +52,12 @@ This file specifies basic information about the course:
     "uuid": "cef0cbf3-6458-4f13-a418-ee4d7e7505dd",
     "name": "TAM 212",
     "title": "Introductory Dynamics",
+    "comment": "The assessment set order used here will be the one shown within PrairieLearn",
     "assessmentSets": [
         {"abbreviation": "HW", "name": "Homework", "heading": "Homeworks", "color": "green1"},
         {"abbreviation": "E", "name": "Exam", "heading": "Exams", "color": "red1"}
     ],
+    "comment": "Topics should be thought of as chapters or sections in a textbook",
```
mwest1066, line 60: While I like these examples, I don't think we can support multiple comments within the same object.

docs/assessment.md
```diff
@@ -15,7 +15,8 @@ Each assessment is a single directory in the `assessments` folder or any subfold
     "set": "Quiz",
     "number": "2",
     "allowAccess": [],
-    "zones": []
+    "zones": [],
+    "comment": "You can add comments to json files using this property."
```
mwest1066, line 19:
```suggestion
    "comment": "You can add comments to JSON files using this property."
```

schemas/schemas/questionOptionsv3.json
```diff
@@ -3,5 +3,14 @@
     "title": "v3 question options",
     "description": "Options for a v3 question.",
     "type": "object",
-    "properties": {}
+    "comment": {
+        "description": "Add the comment",
+        "type": "string"
+    },
```
mwest1066, line 9: I think these lines are incorrect?

news_items/021_Json_schema_comment/info.json
```diff
@@ -0,0 +1,6 @@
+{
+  "uuid": "73dab6c2-5988-4324-a5b4-01728cde7d26",
+  "title": "Comment attribute for JSON file",
```
mwest1066, line 3:
```suggestion
  "title": "Comment properties in JSON files",
```

news_items/021_Json_schema_comment/index.html
```diff
@@ -0,0 +1,41 @@
...
+ .code-block div {
+     padding: 0.5rem;
+     display: block;
+     margin-bottom: -1em;
+     line-height: 125%;
+ }
+</style>
+
+<p>
+    PrairieLearn now add a new attribute <code class="user-output">comment</code>, in order to add comment for the code in JSON file. Here is an example to imply this attribute:
```
mwest1066, line 29: This needs to be edited. I will do that before merging. Please leave this comment unresolved until I do it.

mwest1066 (changes requested):

nwalters512:
Drive-by comment: apologies if this has been discussed before, but why not switch to using a JSON parser that allows actual comments instead of adding `comment` keys everywhere? https://www.npmjs.com/package/json5 and https://www.npmjs.com/package/comment-json are two such parsers. Adding comments in JSON is increasingly common - see TypeScript and Dealing, two very widely-users packages that allow JS-style comments in their JSON config files.

mwest1066:
> Drive-by comment: apologies if this has been discussed before, but why not switch to using a JSON parser that allows actual comments instead of adding `comment` keys everywhere? https://www.npmjs.com/package/json5 and https://www.npmjs.com/package/comment-json are two such parsers. Adding comments in JSON is increasingly common - see TypeScript and ESLint, two very widely-users packages that allow JS-style comments in their JSON config files.

I would like to do this, but we need to be able to modify the JSON programmatically while preserving the comments. This is necessary for UI controls that edit JSON configs (e.g., the "copy question" button needs to assign a new UUID). It looks like [comment-json](https://www.npmjs.com/package/comment-json) can roundtrip (parse-edit-stringify) while preserving comments, but I dislike the fact that this is an ad hoc file format (no spec). JSON5 is much more appealing, but I can't find a parser with roundtrip capabilities.

echuber2:
If multiple comments are needed within one object, would it make sense to have the `comment` property be completely ignored, in which case it could contain anything (a list of strings, etc.). This would make it harder to try to display the comment in the UI somewhere though. Alternatively, cramming escaped markdown into the comment might allow more documentation.

(I haven't read through the entire PR so I'm not sure if this has already been raised...)

nwalters512:
> I would like to do this, but we need to be able to modify the JSON programmatically while preserving the comments. This is necessary for UI controls that edit JSON configs (e.g., the "copy question" button needs to assign a new UUID). It looks like [comment-json](https://www.npmjs.com/package/comment-json) can roundtrip (parse-edit-stringify) while preserving comments, but I dislike the fact that this is an ad hoc file format (no spec). JSON5 is much more appealing, but I can't find a parser with roundtrip capabilities.

Excellent point about needing these files to be machine-editable. https://github.com/noahsug/json5-writer looks like it would work well for our use case! And if not that, I understand your concerns about `comment-json`, but it's used by a variety of well-known JS packages, so I wouldn't write it off quite so quickly.

docs/course.md
zillesc: Applied.

schemas/schemas/adminQuery.json
zillesc: Did you want this for all comments?

schemas/schemas/questionOptionsv3.json
zillesc: Can you explain?  Also, is this questionOptionsv3 deprecated or is it somehow included from another schema?

docs/course.md
zillesc: included in 502ad59f3

rfcs/005-json-comments-schemas.md
zillesc: resolved in commit 502ad59f3

news_items/021_Json_schema_comment/index.html
zillesc: resolved in 502ad59f3

docs/course.md
zillesc: resolved in 502ad59f3

schemas/schemas/adminQuery.json
mwest1066: What do you think? We can leave comments as just simple strings, which might be convenient in the future if we want to edit them programmatically, or we can allow comments to be arbitrary data (strings, objects, or arrays). The benefit of allowing them to be arbitrary data would be to allow flexibility for people to store other metadata into their JSON files (without having to encode it as a string). I don't really have a good use case for this at the moment, so it's more of a "why not allow it" sort of thing. What do you think? @nwalters512 @nicknytko any thoughts?

schemas/schemas/questionOptionsv3.json
mwest1066: There shouldn't be a `"comment"` at the top level of the schema file (it doesn't do anything). The `"comment"` property should only be added under `"properties"`.

schemas/schemas/adminQuery.json
```diff
@@ -8,6 +8,10 @@
         "description"
     ],
     "properties": {
+        "comment": {
+            "description": "Arbitrary comment for reference purposes.",
```
mwest1066, line 12: You are probably intending to do this anyway, but this description should be used for all `"comment"` properties in all schemas.

schemas/schemas/adminQuery.json
```diff
@@ -8,6 +8,10 @@
...
+            "type": ["string", "array", "object"]
```
mwest1066, line 13: See discussion thread above for whether we should use `"type": "string"` or this more complicated type everywhere.

schemas/schemas/adminQuery.json
mwest1066: I guess I lean towards allowing arbitrary data here, but I don't feel strongly and I'm happy to be swayed to just "string".

mwest1066 (approved):

schemas/schemas/adminQuery.json
nicknytko: I personally don't see why we shouldn't allow any data type here &mdash; one use case could be to have an array of string comments maybe?  Or for formatting purposes:

```
"comment": [
    "bla bla here are some bullet points",
    [
        "1. asdf",
        "2. qwerty",
        "3. yuiop"
    ]
]
```

nicknytko: In any case I agree that this is sort of a "why not" change, and it seems like it would be trivial enough to add as well.

mwest1066: Yeah, agreed. We don't need to emphasize this in the docs or examples. Just a brief note in the docs would be sufficient, I think.

schemas/schemas/adminQuery.json
zillesc: I'm embarrassed to say that hadn't occurred to me, but it does now... :)

news_items/021_Json_schema_comment/index.html
zillesc: I took a swing at it, but you can look.  (I did it before I realized this comment was for that file... :)

mwest1066: Most appreciated! The less editing I need to do the better 😄

schemas/schemas/adminQuery.json
zillesc: propagated to all in d80069e1c

schemas/schemas/adminQuery.json
zillesc: Added brief mention in docs in d80069e1c

zillesc:
I think I've addressed the reviews.  I've tested locally.  It previously failed in CI / build (pull_request) for a reason I don't understand, but if that is fine, it could be merged after Matt reviews the news item.

mwest1066:
> I think I've addressed the reviews. I've tested locally. It previously failed in CI / build (pull_request) for a reason I don't understand, but if that is fine, it could be merged after Matt reviews the news item.

Excellent, thanks!

schemas/schemas/adminQuery.json
mwest1066: Looks good!

PR #3348, opened 2020-10-15
Extend the Assessment Access page to display more relevant information

jonatanschroeder:
This PR intends to provide a plain-English explanation of individual settings for a specific assessment. Instructors should be able to see what rules will be applied to each set of students, with a detailed explanation of what each student will see at each time.


jonatanschroeder:
Adding @echuber2 suggestion as a comment: consider a wizard where instructors can choose options and pick the right settings.

jonatanschroeder:
image

image

image

jonatanschroeder:
image

pages/instructorAssessmentAccess/instructorAssessmentAccess.ejs
```diff
@@ -7,21 +7,23 @@
...
       <%- include('../partials/assessmentSyncErrorsAndWarnings'); %>
 
-      <div class="card mb-4">
-        <div class="card-header bg-primary text-white d-flex align-items-center">
-          <%= assessment_set.name %> <%= assessment.number %>: Access
-        </div>
+      <details open class="card mb-4">
+        <summary class="card-header bg-primary text-white d-flex align-items-center">
+          <%= assessment_set.name %> <%= assessment.number %>: Defined Access rules
+        </summary>
```
nwalters512, line 20: This would be a brand-new pattern for PL. Without any visual indications that this can be collapsed, I'm a bit hesitant to thumbs-up this. I'd recommend adding something like the down/left chevron that GitHub uses on the "File changed" page of a PR:

Screen Shot 2022-04-07 at 15 04 10

Screen Shot 2022-04-07 at 15 05 07

pages/instructorAssessmentAccess/instructorAssessmentAccess.ejs
```diff
@@ -7,21 +7,23 @@
...
     <div id="content" class="container-fluid">
       <%- include('../partials/assessmentSyncErrorsAndWarnings'); %>
 
-      <div class="card mb-4">
-        <div class="card-header bg-primary text-white d-flex align-items-center">
-          <%= assessment_set.name %> <%= assessment.number %>: Access
-        </div>
+      <details open class="card mb-4">
+        <summary class="card-header bg-primary text-white d-flex align-items-center">
+          <%= assessment_set.name %> <%= assessment.number %>: Defined Access rules
```
nwalters512, line 19:
```suggestion
          <%= assessment_set.name %> <%= assessment.number %>: Defined access rules
```

pages/instructorAssessmentAccess/instructorAssessmentAccess.ejs
```diff
@@ -83,8 +92,129 @@
         <div class="card-footer">
           <small>Instructions on how to change the access rules can be found in the <a href="https://prairielearn.readthedocs.io/en/latest/accessControl/" target="_blank">PrairieLearn documentation</a>. Note that changing time limit rules does not affect assessments in progress; to change the time limit for these exams please visit the <a href="<%= urlPrefix %>/assessment/<%= assessment.id %>/instances">Students tab</a>.</small>
         </div>
-      </div>
+      </details>
+      
+      <details class="card mb-4" open>
+        <summary class="card-header bg-primary text-white d-flex align-items-center">
+          <%= assessment_set.name %> <%= assessment.number %>: Interpreted Access rules
```
nwalters512, line 99:
```suggestion
          <%= assessment_set.name %> <%= assessment.number %>: Interpreted access rules
```

pages/instructorAssessmentAccess/instructorAssessmentAccess.ejs
```diff
@@ -83,8 +92,129 @@
...
+                  <% } else { %>
+                  <a role="button" class="btn btn-xs btn-warning" tabindex="0" data-toggle="popover" data-trigger="focus" data-container="body" data-placement="auto" title="Hidden UIDs" data-content="This access rule is specific to individual students. You need permission to view student data in order to see which ones.">Hidden</a>
+                  <% } %>
+                </th>
+              </tr>
+              <% set.rules.forEach(function(access_rule) { %>
+              <tr>
+                <td>
+                  <% if (access_rule.valid_now_public) { %>
+                  <span data-toggle="tooltip" title="This rule is currently in place in public mode">
```
nwalters512, line 139: Is this a better phrasing?
```suggestion
                  <span data-toggle="tooltip" title="This rule currently applies in public mode">
```

pages/instructorAssessmentAccess/instructorAssessmentAccess.ejs
```diff
@@ -83,8 +92,129 @@
...
+              <% }); %>
+              <% }); %>
+            </tbody>
+          </table>
+        </div>
+      </details>
+
+      <details open class="card mb-4">
+        <summary class="card-header bg-primary text-white d-flex align-items-center">
+          <%= assessment_set.name %> <%= assessment.number %>: Other Access-related Settings
```
nwalters512, line 194:
```suggestion
          <%= assessment_set.name %> <%= assessment.number %>: Other access-related settings
```

pages/instructorAssessmentAccess/instructorAssessmentAccess.js
```diff
@@ -10,18 +11,164 @@ const sqlLoader = require('../../prairielib/lib/sql-loader');
...
-router.get('/', function (req, res, next) {
-  debug('GET /');
-  var params = {
-    assessment_id: res.locals.assessment.id,
-    link_exam_id: config.syncExamIdAccessRules,
-  };
-  sqldb.query(sql.assessment_access_rules, params, function (err, result) {
-    if (ERR(err, next)) return;
-    res.locals.access_rules = result.rows;
+let compare_date = function (old_date, new_date, old_is_null, new_is_null, both_are_null) {
```
nwalters512, line 14: Nit: we apply this rule inconsistently (and not at all for data coming from database queries), but for your consideration, we typically use `camelCase` for function names and parameters.

pages/instructorAssessmentAccess/instructorAssessmentAccess.js
```diff
@@ -10,18 +11,164 @@ const sqlLoader = require('../../prairielib/lib/sql-loader');
...
+  if (old_date === null) return new_date === null ? both_are_null : old_is_null;
+  else if (new_date === null) return new_is_null;
+  else return old_date - new_date;
+};
+
+let apply_rule = function (list, formal_rule) {
+  let new_rule = Object.assign({}, formal_rule);
```
nwalters512, line 21: Nit, modern JS:
```suggestion
  let new_rule = { ...formal_rule };
```

pages/instructorAssessmentAccess/instructorAssessmentAccess.js
```diff
@@ -10,18 +11,164 @@ const sqlLoader = require('../../prairielib/lib/sql-loader');
...
+        link_exam_id: config.syncExamIdAccessRules,
+      })
+    ).rows;
+
+    debug('building user-friendly description');
+    let student_rules = [];
+    let user_spec_rules = [];
+
+    // Creates sets of unique user lists
+    res.locals.access_rules.forEach((formal) => {
```
nwalters512, line 112: It's not clear to me what "formal" means throughout this file.

pages/instructorAssessmentAccess/instructorAssessmentAccess.js
```diff
@@ -10,18 +11,164 @@ const sqlLoader = require('../../prairielib/lib/sql-loader');
...
+    }
+  });
+  if (valid) {
+    list.push(new_rule);
+  }
+};
+
+router.get(
+  '/',
+  asyncHandler(async (req, res, _next) => {
```
nwalters512, line 93: There's a lot of semi-complex logic here, and this kind of pure-data-in, pure-data-out thing is a fantastic candidate for a unit test.

pages/instructorAssessmentAccess/instructorAssessmentAccess.sql
```diff
@@ -57,6 +79,17 @@ FROM
...
+SELECT
+    a.type AS type,
+    a.auto_close AS auto_close,
+    a.allow_real_time_grading AS allow_real_time_grading,
+    a.multiple_instance AS multiple_instance
+FROM
+    assessments AS a
+WHERE
+    a.id = $assessment_id
+LIMIT 1;
```
nwalters512, line 95: We should never have multiple assessments with the same ID.

```suggestion
    a.id = $assessment_id;
```

pages/instructorAssessmentAccess/instructorAssessmentAccess.ejs
echuber2: I think the Bootstrap theme already has some styles for collapsible sections too. The grading results divs on question pages use them.

jonatanschroeder: This was something I've originally added when I first created the PR in 2020, but I don't feel strongly about. I'm happy to replace it with regular divs instead. That said, removing `d-flex` from the summary element causes the browser's default chevron to show up:

image

Any preference?

pages/instructorAssessmentAccess/instructorAssessmentAccess.sql
jonatanschroeder: This query is actually leftover from a time when I was a bit naive about the system, it is not needed at all...

pages/instructorAssessmentAccess/instructorAssessmentAccess.js
jonatanschroeder: It refers to the original rule from the database. I renamed it for clarity.

mwest1066:
It wasn't immediately obvious to me that the calendar icon on the far left indicated the active rule. Perhaps a green badge with the word `active` in it?

pages/instructorAssessmentAccess/instructorAssessmentAccess.js
jonatanschroeder: Done in c07247c90.

trombonekenny:
Functionality-wise (I haven't code reviewed), it looks good. Shows what's there in the traditional "here's what's in the file" view but gives different parsing of it to show how the configuration defines the functionality. This will be very helpful, thank you for your work.

Three small thoughts if you're still entertaining feature ideas:
When an exam is `multipleInstance: true`, add that statement to the "Other access-related settings" and change the language of "Can start the assessment" to something like "Can start multiple instances"...

I like that it hints that an access rule will never be applied because of the order of precedence makes it impossible:
image

I'd say you could go even further to make a whole section for "Access control lists that will never trigger", make it's top banner red, maybe put it second after the "Defined access rules"? This is kind of an FAQ for access problems and this would give us a space to make it much more visible with the space to describe it (besides a hover which might get overlooked on a busy access page). I bet we could come up with other "access rules that are impossible" to add more matches to that section?

Finally, I've started thinking about how Canvas does UX as a kind of "standard" -- more about "this is what the masses are already used to seeing" rather than any commentary about the quality of Instructure's design. If I was getting started on this feature, I would give a think on how access rules are edited and displayed in other systems. That might help guide where we'd mimic another known pattern or when we need to go off on our own. Perhaps you've already considered that for this? For example, I think this PR shares a lot of new information in a few densely packed tables but more can also be distracting to newbies.

jonatanschroeder:
> Three small thoughts if you're still entertaining feature ideas: When an exam is `multipleInstance: true`, add that statement to the "Other access-related settings" and change the language of "Can start the assessment" to something like "Can start multiple instances"...

Good point. Will do.

> I like that it hints that an access rule will never be applied because of the order of precedence makes it impossible: 
> I'd say you could go even further to make a whole section for "Access control lists that will never trigger", make it's top banner red, maybe put it second after the "Defined access rules"? This is kind of an FAQ for access problems and this would give us a space to make it much more visible with the space to describe it (besides a hover which might get overlooked on a busy access page). I bet we could come up with other "access rules that are impossible" to add more matches to that section?

This is a good point. What do you think of an alert at the top instead?

> Finally, I've started thinking about how Canvas does UX as a kind of "standard" -- more about "this is what the masses are already used to seeing" rather than any commentary about the quality of Instructure's design. If I was getting started on this feature, I would give a think on how access rules are edited and displayed in other systems. That might help guide where we'd mimic another known pattern or when we need to go off on our own. Perhaps you've already considered that for this? For example, I think this PR shares a lot of new information in a few densely packed tables but more can also be distracting to newbies.

One way this could be achieved to some extent is to not show the "Defined access rules" by default (e.g., with a summary/details), or move it to the bottom with smaller font, or even show it as a modal with a button like "Original rules". In other words, the purpose of this page would be to show how access is interpreted, with the list of defined rules as a side information. In that case, the alert mentioned above would have a link to open the rules. What do you think? @mwest1066 comments?

trombonekenny:
> > I like that it hints that an access rule will never be applied because of the order of precedence makes it impossible:
> > I'd say you could go even further to make a whole section for "Access control lists that will never trigger", make it's top banner red, maybe put it second after the "Defined access rules"? This is kind of an FAQ for access problems and this would give us a space to make it much more visible with the space to describe it (besides a hover which might get overlooked on a busy access page). I bet we could come up with other "access rules that are impossible" to add more matches to that section?
> 
> This is a good point. What do you think of an alert at the top instead?

That's fine - anything to draw attention to it as that's a frequently overlooked thing. I see your other issue on errors on sync as well. I'm not sure our precedent on sync errors being grammar or breaking only, or if we're doing logic checks too? Or what our instructors would expect.

> One way this could be achieved to some extent is to not show the "Defined access rules" by default (e.g., with a summary/details), or move it to the bottom with smaller font, or even show it as a modal with a button like "Original rules". In other words, the purpose of this page would be to show how access is interpreted, with the list of defined rules as a side information. In that case, the alert mentioned above would have a link to open the rules. What do you think? @mwest1066 comments?

I like the idea of showing the rules "as is" at the top (for prior use of the page if nothing else), but also expanding on it deeper in the page. I was thinking more big picture where we show alternative views of the access rules (maybe not tables; something else?), if that's not too confusing (or gives an option to enlarge the UX via web editing, like we've done in a few other places. Example: copy this access rule, validation while editing, etc.)

trombonekenny:
Oh, I just remembered another error case it would be nice to catch here. If the assessment is configured for access outside of the scope of the course instance. For example, giving a makeup test the next semester when the course instance access turned off months ago. Even something like a warning at the top that says if the course instance is not currently available (with a link to the view/edit page there)?

jonatanschroeder:
> Oh, I just remembered another error case it would be nice to catch here. If the assessment is configured for access outside of the scope of the course instance. For example, giving a makeup test the next semester when the course instance access turned off months ago. Even something like a warning at the top that says if the course instance is not currently available (with a link to the view/edit page there)? Please and thank you.

That's a great point. I'll add some information in this regard, though it may be tricky to do it for exam-mode assessments.

trombonekenny:
UX-wise I've even thought about putting an alert bar under the navbar for instructor pages when in a course instance that is not available to students (or perhaps warning if it will go unavailable in 3 weeks or something). I could see where that's helpful for beginners as a reminder for them to configure, and also to know that the toy assessments they're starting to assemble won't be seen by students because access is not open.

trombonekenny:
I wonder if this PR is appropriate to warn about `allowAccess: []`, or even `allowAccess: [ {} ]`? This came up in #pl-help today, and reinforced that no access lists on an assessment means no access.

PR #3353, opened 2020-10-16
Grading pre post split

andrewstec:
## To do: 

- Must implement statuses in `instance_question` that determine if an item has been pre-graded or post-graded. (Old convention is to use the graded_at timestamp to know if something has been graded).
- Pre-graded `instance questions` are anything that have finished the grading run but has unmarked manual grading objects. 
- Figure out a manual grading flag. Perhaps just default to manual graded = false and the UI can use JavaScript/Locals to determine if manual grading element exists on front-end. 
- 

## Questions for @cheeren and @mebird (for meeting): 
--------------------------------
- It seems like something is ready to be post-graded if it has completed pre-grading by an internal or external grading operation and has either been manually graded or is precluded from manual grading. Can we agree?
- It seems like a pre-grade operation should return the 
- What is the criteria to be precluded OR included for manual grading, other than having a pre-grade? (@mebird, what data do you need sent to the front-end here to determine if an item is manually gradable on the front-end? status, and is there something in the question that labels something for later manual grading?)

### Instance Question Ready to be Post-Graded:

- Has been pre-graded by internal OR external grader OR both grader types
   AND
- IF AND ONLY IF all manual grade objects that exist on an `instance question` have been manually graded.

To implement the above, we can add a hook after the end of the grading cycle to always check if manual grading exists on the object. If it does, then label the instance question as pre-graded and await for manual grading. 

@mebird We will send back information about an `instance question` that is loaded on the instructor UI with a pre-graded or post-graded status. Your logic should run when something is in pre-graded state, but not sure about post-graded state, eh?
@cheeren We have to assess whether we want this UI logic to be available for the post-graded `instance questions`. Perhaps we are blocked by the `regrading` topic? Or perhaps, as manual grading only occurs after internal and external grading, then we can always re-submit the manual grade scores and re-tabulate the final score? Post-graded items can still be manually re-graded as many times as you want, in other words, as it doesn't have too many moving parts, I beileve. 

### Internal Grader:
-------------------

Summary: The internal grader is already used by instructors' question banks. We should leave this logic intact. It 1) produces a partial score and 2) a tabulated score with the partial score data that, so far exists. This data can be sent to the external grader operation, returned, and the concatenated data from internal and external grading can be available for the manual grader operation.

- [ ] Occurs before any external grading to ensure that external grading container receives pre-tabulated `partial scores` and pre-grade `scores` that are finalized after external grading and/or manual grading have completed.
- [ ] Does not stamp a pre-graded or post-graded status.

### External Grader:
-------------------

Summary: The external grader is already used by instructors' grading containers. Backwards compatibility must be maintained for external grading containers that already exist and produce a score from partial scores (post-grading).  The external grading containers that already exist return a `grade` instead of `partial_scores`. 

Hence, the external grader will have to stamp a:
- [ ] Does NOT stamp a pre-graded or post-graded status.
- [ ] Needs to query database to get `partial_scores` from internal grader as input for its own internal sake.
- [ ] Can produce the score for a question (if it does, then we do not partial grade)
- [ ] Can produce partial_scores for a question (if it does, then we partial grade to produce a score)

### Manual Grading:

- [ ] Receives `[questionName]`: {score: number, weight: number} type Array object from an endpoint.
- [ ] Manual Grading Controller takes pre-graded scores (either an internal or external container run, or both) and re-calculates with the prior `partial score` and new `manual scores` to produce the final `score` object. 
- [ ] Does not make sense to stamp a manual graded status here, as condition for post-grading is that [Instance Question Ready to be Post Graded](#instance-question-ready-to-be-post-graded)

### Test Acceptance Criteria:

- [ ] An `instance question` can be labelled as pre-graded, which enables the manual grading UI.
  - [ ] The `instance question` is ONLY labelled as pre-graded, even if it returns a tabulated `score` from `partial scores` because the grade could potentially be changed at the mid-MANUAL step.
- [ ] Every `instance question` that finishes the pre-grading steps is sent for post-grading IF AND ONLY IF there are no manual grading questions to be submitted.
- [ ] Backwards compatibility is provided for old external grading containers that simply produce a score and reach POST-grade state. @cheeren How do we know if this is an old container?

  ::: BUT perhaps this summary makes the most sense:

  - IF pre-grade has run, and no manual grading to do, then runs post-grading operation.
  - IF pre-grade has run, and manual grading to do, then still in pre-grading state.
  - IF post-grade state has been reached, then pre-grade necessarily is complete and manual grading has OR has not run.

### DIFF
--- a/database/tables/grading_jobs.pg
+++ b/database/tables/grading_jobs.pg
@@ -15,6 +15,7 @@ columns
     grading_started_at: timestamp with time zone
     grading_submitted_at: timestamp with time zone
     id: bigint not null default nextval('grading_jobs_id_seq'::regclass)
+    manual_score: double precision
     output: text
     partial_scores: jsonb
     s3_bucket: text
@@ -22,6 +23,7 @@ columns
     score: double precision
     submission_id: bigint not null
     v2_score: double precision
+    weight_score: double precision
 
 indexes
     grading_jobs_pkey: PRIMARY KEY (id) USING btree (id)
--- a/database/tables/questions.pg
+++ b/database/tables/questions.pg
@@ -12,6 +12,7 @@ columns
     external_grading_timeout: integer
     grading_method: enum_grading_method not null default 'Internal'::enum_grading_method
     id: bigint not null default nextval('questions_id_seq'::regclass)
+    manual_grading_score_limit: double precision
     number: integer
     options: jsonb
     partial_credit: boolean default true
--- a/exampleCourse/questions/demo/autograder/codeUpload/question.html
+++ b/exampleCourse/questions/demo/autograder/codeUpload/question.html
@@ -18,6 +18,12 @@
   <pl-external-grader-variables params-name="names_from_user"></pl-external-grader-variables>
 
   <pl-file-upload file-names="fib.py"></pl-file-upload>
+
+  A <pl-dropdown sort="ascend" answers-name="hume">
+    <pl-answer correct="true">wise</pl-answer>
+    <pl-answer correct="false">clumsy</pl-answer>
+    <pl-answer correct="false">reckless</pl-answer>
+  </pl-dropdown> man proportions his belief to the evidence. <p></p>
 </pl-question-panel>
 
 <pl-submission-panel>
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -327,7 +327,7 @@ module.exports = {
                 }
                 // 0 is valid score
                 if (content.grading.score === 0 || content.grading.score) {
-                    if (content.grading.partial_scores && content.grading.partial_scores.length > 0) {
+                    if (content.grading.partial_scores && Object.keys(content.grading.partial_scores).length) {
                         content.grading['status'] = 'pre-graded';
                         content.grading['score'] = formulas.calcQuestionScore(content.grading.partialScores, content.grading.partialCredit);
                     }
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -10,6 +10,7 @@ const logger = require('../lib/logger');
 const question = require('../lib/question');
 const externalGrader = require('./externalGrader');
 const externalGradingSocket = require('../lib/externalGradingSocket');
+const formulas = require('../lib/formulas');
 const sqldb = require('@prairielearn/prairielib/sql-db');
 const sqlLoader = require('@prairielearn/prairielib/sql-loader');
 const ltiOutcomes = require('../lib/ltiOutcomes');
@@ -282,11 +283,11 @@ module.exports = {
     /**
      * Process the result of an external grading job.
      *
-     * @param {Obect} content - The grading job data to process.
+     * @param {Object} content - The grading job data to process.
      */
     processGradingResult(content) {
         async.series([
-            (callback) => {
+            async (callback) => {
                 if (!_(content.grading).isObject()) {
                     return callback(error.makeWithData('invalid grading', {content: content}));
                 }
@@ -308,6 +309,17 @@ module.exports = {
                     if (content.grading.score < 0 || content.grading.score > 1) {
                         return callback(error.makeWithData('grading.score out of range', {content: content}));
                     }
+                    if (content.grading.score && content.grading.partialScores && content.grading.partialScores.length > 0) {
+                        return callback(error.makeWithData('cannot submit both a grading.score and a grading.partial_scores.'));
+                    }
+                    // if (content.grading.score.isNumber() && content.grading.partial_scores) {
+                    //     return callback(error.makeWithData('cannot submit partial grades and grade score', {content: content}));
+                    // }
+                }
+
+                if (content.grading.partial_scores && content.grading.partial_scores.length > 0) {
+
+                    content.grading['score'] = formulas.calcQuestionScore(content.grading.partialScores, content.grading.partialCredit);
                 }
                 
                 const params = [
@@ -320,6 +332,7 @@ module.exports = {
                     content.grading.endTime,
                     gradable,
                 ];
+
                 sqldb.call('grading_jobs_process_external', params, (err) => {
                     if (ERR(err, callback)) return;
                     callback(null);
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -287,7 +287,15 @@ module.exports = {
      */
     processGradingResult(content) {
         async.series([
-            async (callback) => {
+            (callback) => {
+                // We need to know partial_credit bool to grade partial_scores
+                sqldb.queryOneRow(sql.select_question_for_grading_job, {grading_job_id: content.gradingId}, (err, result) => {
+                    if (ERR(err, callback)) return;
+                    if (content && content.grading) content.grading['partialCredit'] = result.rows[0].partial_credit;
+                    callback(null);
+                });
+            },
+            (callback) => {
                 if (!_(content.grading).isObject()) {
                     return callback(error.makeWithData('invalid grading', {content: content}));
                 }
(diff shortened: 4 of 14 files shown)


andrewstec:
I just wrote down some thoughts here for everyone. I have to run right now, or else I would finish my thoughts... but just in case you are curious of where things are going, I am writing down some cases that I am trying to fulfill as part of the requirements for this task. I think these logical requirements are steering us towards some conclusions. 

I will have to compare this to your notes, @cheeren .

Feel free to discuss/ask anything.

mebird:
## Statuses
Here's my initial take on how to handle statuses. Andrew, I think this aligns with how you're thinking of things, but I'm not 100% sure.

### Core Ideas
- A question should be `pending manual grading`/`manually graded`.
- The status of a question should be the aggregate of its partial statuses.

### When Do We Mark Something For Grading?
Currently, we take a very granular approach to "manually grading" a given question. We have an element that wraps another PL element, which will give instructors grading toggles in the manual grading view (#3174).

As we configure the toggles at the element HTML level, we have three main questions.
- When do we know parts of a question must be manually graded?
- When do we mark a partial score as needs grading/graded, and when do we bubble this up to the top level question?
- How do we handle instructors who change which partials should be manually graded?

#### The Lazy Approach
Usually, an element doesn't build partials until a student submits their answers. Hence, we can be "lazy", and only add grading statuses to a question when someone submits. A few issues come to mind.

- We can't append these statuses onto `data["partials"]`, since we can't ensure our "manual grading element" grade code will be run _before_ an element's grading code (where it builds these partials). Hence, we should implement a new field for `data`, `data["statuses"]`, which is a JSON dict from the partial name to a text status.
- We need to bubble the partial statuses up to the parent question in the database.
  - We should do this on question submit/as a part of the postgrade hook.
  - If all partial statuses are set to `graded`, the question as a whole should be set to graded.
  - If there are no partial statuses, the question should be set to graded.
  - If there is at least one partial marked as pending, then the question should be set to pending grading.
  - We should keep in mind how to conflate this with external grading, and how this may display on the manual grading menu.
- If an instructor changes the element html, the manual grading overlay may change.
   - There is no way to automatically check this within the element code.
   - However, we can use the idea of an "invalidated" grading state in the parent question to help wipe partials.
      - If an instructor changes the HTML, they should reset the grading state of all questions of that type that may be pending grading.
      - However, we don't have an easy way to trickle this back to the partial level, since this logic is embedded in the element code. 
      - We'd have to re-run the code from the pre-grade step to correctly re-mark statuses for the question.
   - We'd still have to conflate this with the external grading element.

#### TLDR: We Marking What When?
The short and sweet:
- We mark a question for manual grading on submit, by filling out `data["statuses"]` with statuses for corresponding partials.
- If someone changes the question HTML, it is their responsibility to re-run the pre-grading step to correctly reset partials (set all `data["status"]` fields to an empty json object, so we can rebuild the statuses object).
- The grading step should aggregate `data["statuses"]` to mark the parent question status correctly.

andrewstec:
Great info, @mebird!

Take-aways: 

Instead of pre-graded and post-graded, we will give the statuses more user-friendly semantics, as you suggested:

- `pending manual grading`
- `manually graded`

Those seem more reasonable than pre and post graded, etc., which are abstractions to help get us to these manually graded status points in tech-speak.

- **DO NOT** append statuses on partials, as they do not exist until internal grader is run on back-end.
- **DO** append statuses to a new data['statuses'] dictionary.

 I am just copy and pasting some of your logic here, and amending it, as it looks perfect:

- We need to bubble the partial statuses up to the parent question in the database.
  - We should do this on question submit/as a part of the post-grade hook.
  - If all partial statuses are set to graded, the question as a whole should be set to graded.
  - If there are no partial statuses, the question should be set to graded.
  - If there is at least one partial marked as pending, then the question should be set to pending grading.
  - We should keep in mind how to conflate this with external grading, and how this may display on the manual grading menu. **Regarding this last point, we can add in a pre-finalized ```score``` field to create the latest tallied score from partial_scores, but when a `score` field is submitted from the external grader, it will automatically become finalized. Any manual grading will be precluded with the old logic. I think we have to do this or else all of the score and grade views, and underlying, would have to be re-considered in combining our manual grading logic since they all use the `score` value**
 - If we accept the bold texted amendment in the last point
   - then we calculate the `score` field _if and only if_ all of the partial statuses are set to `graded` or `manually graded`.
    - then, since all internal and external grading would be completed, an instructor can repeatedly update the manual grade points, which triggers the final score calculation any number of 'x' times for a manual score re-grade operation.

As for what happens when the instructor changes the HTML, please let me know if we need to plan out the solution thoroughly now. Your approach seems sound, and I will gladly factor it in, but I do not have enough information/knowledge to properly assess it at this time. I will be keeping my eye out on how to implement it though.

I will start development on this plan. I think this is really good! 

@mwest1066 Is it okay to start development on this, and see what happens with the sub-question / question plan at a later time? Let us suppose that we were to write strong tests to support our manual grading features. Based on the sub-question / question manual work that you want, would it be preferable if we waited with our work or started. We could re-configure some of our logic to work with the new sub-question / question approach. Perhaps this would be easier too, as we would have more tangible work completed that shows our implementation clearly.

andrewstec:
Cinda: 

OK!! general recap. IMO this is all just a re-ordering of what we want to do. We need to pivot away from thinking about partials, for the moment.  Assume, for the moment, that there are few elements on a page. In the question’s info.json, we specify a manualgrading flag, and a manual grading mode. The mode options are
1) override prior grade, which allows for 100% of the question to be determined by the value entered manually. The prior grading info must be seen here (we have to know what we’re overwriting.
2) weighted grading, which sets aside x% of the final mark for manual score, and (100-x)% for AG score. (is 1) just 2) but with x == 100?)
3) filter by score: only MG submissions whose original scores are below a threshold.
The order in which the grading is done is: AG(+postgrade), EG(+postgrade), MG(+postgrade).

andrewstec:
@cheeren Could you please elaborate on `filter by score: only MG submissions whose original scores are below a threshold.`? I am assuming that this is per student. We can send all the MG submissions to the front-end and use a jQuery filter already implemented to filter the score.

mebird:
> @cheeren Could you please elaborate on `filter by score: only MG submissions whose original scores are below a threshold.`? I am assuming that this is per student. We can send all the MG submissions to the front-end and use a jQuery filter already implemented to filter the score.

I agree that all of this filtering can and should be done on the frontend. I think Cinda means that she wants to have a feature that _conditionally_ triggers "manual grading", i.e. if a score is below a certain threshold, we reserve some percentage of points for manual grading, else, we give the student full marks.

I can see how to implement an "override prior grade" and "calculate x% manual grade" on the frontend (just show different menus). My only concern is if an instructor changes this flag, and how this interface should interact with the "conditional manual grading". @andrewstec any opinions? In regards to the flag changing, we could require instructors to re-run the "regrade" workflow (I think this is what @astrosticks is working on, irrc), and then go from there.

andrewstec:
> > @cheeren Could you please elaborate on `filter by score: only MG submissions whose original scores are below a threshold.`? I am assuming that this is per student. We can send all the MG submissions to the front-end and use a jQuery filter already implemented to filter the score.
> 
> I agree that, as stated, this sort of filtering can and should be done on the frontend. However, I think Cinda means that she wants to have a feature that _conditionally_ triggers "manual grading", i.e. if a score is below a certain threshold, we reserve some percentage of points for manual grading, else, we give the student full marks.
> 
> I can see how to implement an "override prior grade" and "calculate x% manual grade" on the frontend (just show different menus). My only concern is if an instructor changes this flag, and how this interface should interact with the "conditional manual grading". @andrewstec any opinions? In regards to the flag changing, we could require instructors to re-run the "regrade" workflow (I think this is what @astrosticks is working on, irrc), and then go from there.

Hmmmmm... I am thinking this over. I have to delve into this logic a bit further. At first glance, it looks we will go into a manual grading mode that is configured at the question level.

Next, the mode options would dictate whether the manual grading is conditionally triggered and with particular point options. 

From a front-end development perspective, particular components can load if manual grading options are enabled. It may not make sense to combine all of these options from a UI/UX perspective though. We haven't delved into this logic enough to really know how this logic will look and what option combinations are logically possible together and work from a UX/UI perspective. 

I hope this helps a little... I will try to think of an answer to your question while zoning in on this.

andrewstec:
- Backend: Create a manual grading flag in the config. Give a percentage towards the manual graded score. Dictate any DB changes/workflow changes.
- UI: Load any question with this view. Do not allow students to access this page. Only render for pages with a grading config. Get graded score data (manual & automatic) and hit an endpoint statically.
- JSON structure in config file.
- How to resolve changes to the config file (do we have a reset/make happy button)?
- Future: building instructor menus, updating grading toggles. This menu will likely have a button to indicate if the config file changes.

andrewstec:
Closing, and I don't think it is relevant to future work.

PR #3354, opened 2020-10-17
Change chunk consumers to use symlinks to unpacked chunks

### LINKED ISSUE #3124: Test suite needs to run with both chunks enabled and with local dev chunks
We should have the test suite run partially with chunks fully enabled (that is, `chunksGenerator=true` and `chunksConsumer=true` using s3rver) and partially with whatever the local dev setup will be.

### LINKED ISSUE #3208: Directory delete/unpack is a race condition while unpacking chunks
When unpacking a chunk we currently delete and then recreate the targetDirectory: https://github.com/PrairieLearn/PrairieLearn/blob/b11c990682600c9e3b0b626747ad959eb6e338d5/lib/chunks.js#L699

This will cause an error for any other code that's currently accessing the question (for example).

To fix this, we could either add locking code or unpack into a temp directory and then move the temp directory over the old one atomically. See https://stackoverflow.com/questions/307437/moving-a-directory-atomically

### LINKED ISSUE #3276: Ordering of chunk downloads is not guaranteed
There is a race condition in the chunk download code. The scenario is:

1. Chunk consumer starts downloading chunk X for question Q.
2. New version of question Q is created and pushed as chunk Y.
3. Chunk consumer starts downloading chunk Y.
4. Chunk Y download completes and is unpacked into Q.
5. Chunk X download completes and is unpacked into Q.
6. Result is that we have the incorrect version of question Q on disk.

Note that this doesn't require two changes to question Q in a short time period (chunk X could have been generated a long time ago). It only requires that X is requested for the first time by a chunk consumer shortly before Y is pushed. The impact is especially bad because we won't fix the problem until a new chunk Z is generated for Q.

To fix this, one possibility would be to:
* [ ] Add a column `chunks.seq bigint` and assign this as `chunks_seq.nextval()` from a new `chunk_seq` sequence. This will let us tell which of chunks X and Y is the most recent (has the highest seq value).
* [ ] In `chunks.js`, add an in-memory Map called `currentTargetSeq` that maps from `targetDirectory` to the latest unpacked chunk `seq` value.
* [ ] When we are about to change `targetDirectory` to a new chunk: https://github.com/PrairieLearn/PrairieLearn/blob/05f78f0eb658c659667cab5c654834f088b7922a/lib/chunks.js#L697 we should:
    * [ ] Add a [LocalLock](https://github.com/PrairieLearn/PrairieLearn/blob/master/lib/local-lock.js) on `targetDirectory` around the change to it.
    * [ ] Get the `currentTargetSeq` value for the current `targetDirectory`.
    * [ ] Only change `targetDirectory` if the old `seq` number for `targetDirectory` was less than `chunk.seq`.
    * [ ] Update `currentTargetSeq` for this `targetDirectory` to `chunk.seq`.
    * [ ] Only do the final `move()` if we actually changed `targetDirectory`: https://github.com/PrairieLearn/PrairieLearn/blob/05f78f0eb658c659667cab5c654834f088b7922a/lib/chunks.js#L707

A completely different possibility would be:
* [ ] Implement #3208 using symlinks, so `targetDirectory` is actually a symlink to `/unpacked-chunks/course-<ID>-<UUID>`.
* [ ] Change this check: https://github.com/PrairieLearn/PrairieLearn/blob/05f78f0eb658c659667cab5c654834f088b7922a/lib/chunks.js#L643 to instead check whether `targetDirectory` points to the UUID of the chunk we are currently processing.
* This strategy won't eliminate the race condition, so it's still possible that we'll unpack an older chunk over a newer one, but it will mean that we are always driving towards eventual consistency. That is, even if we make a mistake, we'll fix it on the next access to that question.

mwest1066:
Rather than unpacking chunks directly into the local course directory, this now makes a directory structure like:
```
/chunks/course-1/unpacked/0e62f70a-e0b1-49b1-90a2-ea0a7d5211f3/info.json
/chunks/course-1/unpacked/0e62f70a-e0b1-49b1-90a2-ea0a7d5211f3/question.html
/chunks/course-1/unpacked/0e62f70a-e0b1-49b1-90a2-ea0a7d5211f3/server.py
/chunks/course-1/course/questions/addNumbers -> ../../unpacked/0e62f70a-e0b1-49b1-90a2-ea0a7d5211f3
```

Resolves #3208, resolves #3124, resolves #3276


lib/chunks.js
```diff
@@ -428,8 +436,11 @@ module.exports.getRuntimeDirectoryForCourse = (course) => {
  * @param {import('../sync/course-db').CourseData} options.courseData
  */
 module.exports.createChunksSymlinks = async ({ coursePath, courseId, courseData }) =>  {
+    /* eslint-disable no-unreachable */
```
nicknytko, line 439: Are we still thinking of doing symlinks on local dev?  The error is getting thrown for me on sync so we could probably remove this until it's implemented.

mwest1066: Good catch! I guess I never actually tested local sync while a `chunksGenerator` 😄 -- just to make sure I understand, this is only a problem on local dev with `chunksGenerator` set to `true`, right?

And good question about the local dev plan. Doing some kind of symlinking to make local dev look more like prod is a good idea. At the moment `createChunksSymlinks()` unconditionally symlinks everything, so I worry that things like reading a file from another question will still work. I was wondering whether we should try and set up symlinks on the fly for chunks, so that when you access a question you get a unique temporary "course directory" that only has the minimal set of symlinks (the specific question, serverFilesCourse, course elements). This would involve:

* [ ] Change `ensureChunksForCourse()` to return the newly-created `coursePath` and eliminate `getRuntimeDirectoryForCourse()`.
* [ ] In `ensureChunksForCourse()`, if we are not a `chunksConsumer` then make a new temporary `coursePath` and put symlinks inside it for the list of `chunks` passed as an argument.

What do you think?

nicknytko: I'm on board with the idea of having questions in separate "course" directories like this.  I'm not sure why we should be symlinking only if we are not a consumer?  Maybe we should be more explicit and check if we're generating chunks and not running in ec2?

At the moment, I'm okay with spinning this out to a separate issue (since I wouldn't consider it pressing atm) and we can just comment out this call or something.

mwest1066: After discussion with @nicknytko we decided to just delete `createChunksSymlinks()` completely. For local dev (where this was used) we can run in one of two setups:
1. (Default) Both `chunksGenerator` and `chunksConsumer` are false. This uses course content straight out of the on-disk course directly, just like we've always done.
2. Set both `chunksGenerator` and `chunksConsumer` to true. This will push/pull chunks to the local mock S3 for testing, just like in production.

Issue #3164, opened 2020-09-22
Latex rendering error


BenjaminCosman:
![image](https://user-images.githubusercontent.com/4876098/93893278-4e7df300-fca2-11ea-89ed-a05c653fb3af.png)
![image](https://user-images.githubusercontent.com/4876098/93893677-c3e9c380-fca2-11ea-92d2-305189511d43.png)

Latex Markdown involving two or more underscores does not render properly - it looks like the Markdown rule treating them as italics is taking precedence over the Latex rule treating them as subscript indicators.

mwest1066:
The markdown rendering happens first (the latex is rendered client-side by mathjax), so I guess this isn't too surprising. Does it work if you write `<span>$a_2 b_8$</span>` in the markdown, to protect the math?

BenjaminCosman:
No, <span>$a_2 b_8$</span> identically to $a_2 b_8$

mwest1066:
It looks like PL is using [showdown](https://github.com/showdownjs/showdown) to render the markdown: https://github.com/PrairieLearn/PrairieLearn/blob/b7c3105ccd65bbb6316ccf5bf13cac26f235cc68/elements/pl-file-editor/pl-file-editor.js#L50

Maybe we should turn on the `literalMidWordUnderscores` option? Docs at https://github.com/showdownjs/showdown

mwest1066:
@BenjaminCosman if you could make & test a patch for this and submit a PR then that would be awesome!

BenjaminCosman:
Fixed by #3355

PR #3355, opened 2020-10-17
pl-file-editor markdown rendering: treat mid-word * and _ as literal, not bold/italics

BenjaminCosman:
Addresses #3164. Two downsides to this solution are

- It doesn't solve #3164 in 100% of cases, though the holdouts don't seem likely to be naturally occurring - e.g. `$3 *4 5* 6$` (with exactly that weird spacing) would still break the Latex rendering
- It does change how markdown is rendered even for folks not using Latex, e.g. 'foo_bar_baz' will now be rendered unchanged instead of italicizing the 'bar' (but how often does someone _actually_ want to italicize/bold in the middle of a word?)

If anyone objects to the second issue, I could create a new option for options.preview so the new rendering mode would be entirely opt-in.

### DIFF
--- a/elements/pl-file-editor/pl-file-editor.js
+++ b/elements/pl-file-editor/pl-file-editor.js
@@ -33,11 +33,11 @@ window.PLFileEditor = function(uuid, options) {
 
     if (options.minLines) {
         this.editor.setOption('minLines', options.minLines);
-    } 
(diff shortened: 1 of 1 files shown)


mwest1066:
Looks great, thanks!

PR #3363, opened 2020-10-19
Automatic save of student answer.

### LINKED ISSUE #3358: Add option to auto-save answer
Students writing long answers to exam questions may be unable to save due to running out of time or network connectivity issues. There should be an option to automatically save the submission periodically via AJAX.

Workaround is to add to top of question:

```
<script>
  let serial = false;
  function saveSubmission() {
    let new_serial = $('.question-form').serialize();
    if (serial !== new_serial) {
      serial = new_serial;
      $.post(window.location.href, serial + '&__action=save');
    }
  }
  setInterval(saveSubmission, 10000);
</script>
```

Which will save every 10 seconds if something changed, but it will create several submissions. Ideally this should replace an existing submission instead.

jonatanschroeder:
Resolves #3358 . Still pending: option to enable/disable the feature (TBD: who decides? student or instructor?).


jonatanschroeder:
Some options for configuration:

- instructor enables the feature on a course-level.
- instructor enables the feature on an assessment-level.
- instructor enables the feature on a question-level.
- student setting stored as a cookie (or local storage, or something similar), with checkbox on the right panel.
- others?

jonatanschroeder:
Adding a note so I don't forget: this still needs to take care of race conditions: what happens if a save happens when an autosave is still in progress? What happens if an autosave takes longer than 2 seconds to run (the next autosave should probably not run, or the interval should be cancelled). It might also be a good idea to review the process of what happens in case of error.

jonatanschroeder:
Just adding a note since this recently came into discussion: it is probably a good idea to review this process and make it based on web sockets instead of ajax. Also, for groupwork, it may make sense to have a way to automatically load saved answers from other group members. This may involve some tweaking in element development, and some elements may not be saved with the same frequency or quality as others.

PR #3378, opened 2020-10-26
Add subdirectory reassignment.

lkwagner:
Add a script that updates to the subdirectory structure based on topic.


mwest1066:
Looks great! Maybe rename to `reassign_questions_to_subdirs.py` or something like that?

lkwagner:
OK, done!

PR #3381, opened 2020-10-29
Add access control example for CBTF exams with outside students

mwest1066:


docs/remoteExams.md
```diff
@@ -6,6 +6,55 @@ This page lists sample assessment configurations for remote exams, where student
...
+        "showClosedAssessment": false
+    },
+    {
+        "mode": "Exam",
+        "examUuid": "c48e40db-258d-43c8-bb26-1f559ffe2228",
+        "credit": 100
+    }
+],
+```
+
```
mfsilva22, line 56: Should we add a note saying that "the additional access rule for specific students can be added after other students already completed the exam using CBTf. This can facilitate setting up later accommodations for students that missed the exam."

mfsilva22: Should we also say that in this case the order does not matter?

mfsilva22 (approved):

docs/remoteExams.md
mwest1066: done and done

PR #3386, opened 2020-10-30
Server-side time limit validation

### LINKED ISSUE #2506: Need to validate time limit expiry on the server

jonatanschroeder:
Fixes #2506 . These commits were originally part of #2750 , but since these changes don't depend on the other and this fixes a potential bug, I move it into a separate PR.

Note: I removed the use of window.performance for timing issues, since it complicates the code and potentially introduces problems, without significant benefits. Also the fields previously used are deprecated.


jonatanschroeder:
To do:

- Update comments in source code to apply more closely to the actual functionality.
- Check if any documentation needs to be updated based on this functionality.
- Update interval should be every 60 seconds (using setInterval instead of setTimeout).
- Use GET request instead of POST. Consider web sockets in the future?

pages/partials/countdown.ejs
```diff
@@ -49,23 +52,43 @@ process) in milliseconds as the variable "serverRemainingMS".
...
+      window.allowQuestionUnload = true;
+      document.forms['time-limit-finish-form'].submit();
+  }
+
+  var handleServerResponseRemainingMS = function(data, textStatus, jqxhr) {
+      serverRemainingMS = data.serverRemainingMS;
+      serverTimeLimitMS = data.serverTimeLimitMS;
+      console.log('Time remaining: ' + serverRemainingMS + ' ms');
+      clientStart = Date.now();
+      if (serverRemainingMS <= 500) {
```
mwest1066, line 75: Should we change this to give them slightly more time? We'd also need to change `remainingMS > 500` below to `remainingMS > 0`.
```suggestion
      if (serverRemainingMS <= 0) {
```

server.js
```diff
@@ -647,6 +647,12 @@ module.exports.initExpress = function() {
         require('./middlewares/studentAssessmentAccess'),
         require('./pages/studentAssessmentInstanceFile/studentAssessmentInstanceFile'),
     ]);
+    app.use('/pl/course_instance/:course_instance_id/assessment_instance/:assessment_instance_id/time_left', [
```
mwest1066, line 650: All of the other timer code refers to this as remaining rather than left. I don't really mind either way, but we should be consistent.
```suggestion
    app.use('/pl/course_instance/:course_instance_id/assessment_instance/:assessment_instance_id/time_remaining', [
```

server.js
```diff
@@ -647,6 +647,12 @@ module.exports.initExpress = function() {
...
+        require('./middlewares/selectAndAuthzAssessmentInstance'),
+        // require('./middlewares/logPageView')('studentAssessmentInstanceTimeLeft'),
```
mwest1066, line 652: I agree that we don't want to log this.

pages/studentAssessmentInstanceExam/studentAssessmentInstanceExam.ejs
```diff
@@ -135,6 +135,8 @@
             <%- include('../partials/countdown', {
               serverTimeLimitMS: assessment_instance_time_limit_ms,
               serverRemainingMS: assessment_instance_remaining_ms,
+              serverUpdateURL: urlPrefix + '/assessment_instance/' +
+                               assessment_instance.id + '/time_left',
```
jonatanschroeder, line 139:
```suggestion
                               assessment_instance.id + '/time_remaining',
```

mwest1066 (approved):

PR #3453, opened 2020-11-09
Update deprecated mdast option

nicknytko:
This was changed in https://github.com/syntax-tree/mdast-util-to-hast/pull/39

Currently emits a harmless warning, but the option has been renamed and the old name is deprecated.

### DIFF
--- a/lib/markdown.js
+++ b/lib/markdown.js
@@ -85,7 +85,7 @@ const processor = unified()
     .use(math)
     .use(handleCode)
     .use(handleMath)
-    .use(remark2rehype, { allowDangerousHTML: true })
+    .use(remark2rehype, { allowDangerousHtml: true })
     .use(raw)
     .use(stringify);
 


mwest1066:
When I first saw this warning some time ago I tried to make this change, but it turned out to actually not work. I'm afraid I can't remember in what way it failed. But just to check, have you confirmed that this does in fact still work?

nicknytko:
> When I first saw this warning some time ago I tried to make this change, but it turned out to actually not work. I'm afraid I can't remember in what way it failed. But just to check, have you confirmed that this does in fact still work?

I was finally noticing the deprecation error when I was running unit tests, but er actually testing it out is probably a good idea. It could have been that earlier the change was made before the package had been updated.

nicknytko:
This looks good on my end.  Running without the flag changed will actually display a ```mdast-util-to-hast: deprecation: `allowDangerousHTML` is nonstandard, use `allowDangerousHtml` instead``` message in the terminal whenever I view a question with markdown.

mwest1066:
> This looks good on my end. Running without the flag changed will actually display a `` mdast-util-to-hast: deprecation: `allowDangerousHTML` is nonstandard, use `allowDangerousHtml` instead `` message in the terminal whenever I view a question with markdown.

Excellent, thanks for checking!

PR #3466, opened 2020-11-14
Add contrib nbextensions for export_embedded

echuber2:
This allows you to do: *(note the `_embed` part!)*

```
!jupyter nbconvert *.ipynb --to html_embed
```

from a notebook cell, which is helpful for exporting notebooks as PDF. (It's also possible for students to pip install this themselves, but that's going to waste a lot of bandwidth over time.)

### DIFF
--- a/workspaces/jupyterlab/Dockerfile
+++ b/workspaces/jupyterlab/Dockerfile
@@ -12,4 +12,6 @@ ENV NO_UPDATE_NOTIFIER=true
 ENV IPYTHONDIR=/tmp/ipython
 ENV YARN_CACHE_FOLDER=/tmp/yarn_cache
 
+RUN pip install --no-cache-dir jupyter_contrib_nbextensions
+
 CMD ["start.sh", "jupyter", "lab"]


nicknytko (approved):

mwest1066:
@echuber2 this looks great! Could we document the "save to PDF" command somewhere? Perhaps in https://github.com/PrairieLearn/PrairieLearn/blob/master/exampleCourse/questions/demo/workspace/jupyterlab/question.html ?

echuber2:
I can document how I meant this trick to be used, but it's still a little awkward as students (or their instructors) have to write an export command in a notebook cell. The built-in "save to PDF" feature in the File menu doesn't work reliably by itself, so the workaround is needed.

I think a better solution overall would be for us to tweak the export options that appear in the File menu so that the save to PDF option is replaced by a custom command that does the embedded HTML export and pops open the generated HTML in a new tab automatically.

nwalters512:
@echuber2 I'm going through old PRs - would you still like to see this merged?

echuber2:
I'm not sure if we should now. It adds a lot of third-party extensions that I haven't vetted. The one I needed at the time was this:

https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/export_embedded/readme.html

This extension made it a bit easier to write a PDF output script that works, but the notebook still has to be prepared in a certain way to go along with it, so it's not a silver bullet for the printing issues before. I haven't checked on whether converting to PDF works more reliably out of the box in the latest releases of Jupyter Lab though.

nwalters512:
@echuber2 given that uncertainty, and the fact that no one has come asking for the feature yet, I'm going to go ahead and close this out. Please open a new PR if you still need this!

PR #3470, opened 2020-11-17
Add group contribution to assessment instances page as a new column

yScribblez:
This allows you to view individual member's group contributions in the assessment instances page as a mini histogram. This is my first PR, so looking for any and every form of feedback, as well as additional guidance on testing. I added the contribution query as a new stored procedure that also merges the results of the query that was previously used in assessment instances, but if there is some other way this should be done that'd be great to know as well.


yScribblez:
closing pr to combine it soon with submissions migration + submissions group functionality

PR #3472, opened 2020-11-18
Add option to allow number inputs to be answered with a blank answer

### LINKED ISSUE #2716: Add blank-is-invalid, gradable attribute to elements
We should do these elements, at least:

- `pl-number-input`
- `pl-integer-input`
- `pl-component-matrix-input`

jonatanschroeder:
Added option `blank-value` in `pl-integer-input` for a value to be used if the answer is blank. Useful in cases where students may leave a field blank, such as for invalid values or in case of multiple integer values where only some need to be answered.

(requested by Margo).

Resolves #2716 

- [x] pl-integer-input
- [x] pl-number-input
- [x] pl-component-matrix-input
- [x] pl-symbolic-input


mfsilva22:
This is a feature improvement we have been planning to make for all elements, as also pointed out in #2716 . To be consistent with the `pl-string` element, we should consider setting this up using `allow-blank` attribute, following similar design.

jonatanschroeder:
I thought of using `allow-blank`, but since we expect the value of `data['submitted_answers'][field]` to be an integer in this case, there would need to be a default value, and I don't think 0 would work in this case. Maybe a combination, with both attributes?

mfsilva22:
The combo `allow-blank` and `blank-value` makes a lot of sense! I like this.

jonatanschroeder:
For pl-component-matrix-input there isn't a simple way to determine how `blank-value` should be treated. Should it be the default for each component that is left blank? Should it include enough elements to fill all values?

docs/elements.md
```diff
@@ -440,6 +444,8 @@ Attribute | Type | Default | Description
 `variables` | string | — | A comma-delimited list of symbols that can be used in the symbolic expression.
 `allow-complex` | boolean | false | Whether complex numbers (expressions with `i` or `j` as the imaginary unit) are allowed.
 `imaginary-unit-for-display` | string | `i` | The imaginary unit that is used for display. It must be either `i` or `j`. Again, this is *only* for display. Both `i` and `j` can be used by the student in their submitted answer, when `allow-complex="true"`.
+`allow-blank` | boolean | false |  Whether or not an empty input box is allowed. By default, an empty input box will not be graded (invalid format).
+`blank-answer` | string | 0 (zero) | Expression to be used as an answer if the answer is left blank. Only applied if `allow-blank` is `true`. Must follow the same format as an expected user input (e.g., same variables, etc.).
```
mfsilva22, line 448: It appears the element has the variable named `blank-value` instead of `blank-answer`

docs/elements.md
```diff
@@ -558,10 +564,12 @@ Attribute | Type | Default | Description
 `allow-partial-credit` | boolean | false | Whether or not to allow credit for each correct matrix component. By default, the variable is graded as correct only when all matrix components are correct.
 `allow-feedback` | boolean | `allow-partial-credit` | Whether or not to allow feedback indicating which matrix components are incorrect. The default value of `allow-feedback` is the value of `allow-partial-credit`.
 `allow-fractions` | boolean | true | Whether to allow answers expressed as a rational number of the format `a/b`.
+`allow-blank` | boolean | false |  Whether or not empty input boxes are allowed. By default, matrices with at least one empty input box will not be graded (invalid format).
+`blank-answer` | string | 0 (zero) | Value to be used as an answer for each individual component if the component is left blank. Only applied if `allow-blank` is `true`. Must follow the same format as an expected user input (e.g., fractions if allowed, etc.).
```
mfsilva22, line 568: It appears the element has the variable named `blank-value` instead of `blank-answer`

docs/elements.md
```diff
@@ -380,6 +382,8 @@ Attribute | Type | Default | Description
 `answers-name` | string | — | Variable name to store data in.
 `weight` | integer | 1 | Weight to use when computing a weighted average score over elements.
 `correct-answer` | float | special | Correct answer for grading. Defaults to `data["correct_answers"][answers-name]`.
+`allow-blank` | boolean | false |  Whether or not an empty input box is allowed. By default, empty input boxes will not be graded (invalid format).
+`blank-answer` | float | 0 (zero) | Value to be used as an answer if element is left blank. Only applied if `allow-blank` is `true`.
```
mfsilva22, line 386: It appears the element has the variable named `blank-value` instead of `blank-answer`

docs/elements.md
```diff
@@ -268,6 +268,8 @@ Attribute | Type | Default | Description
 `atol` | number | 1e-8 | Absolute tolerance for `comparison="relabs"`.
 `digits` | integer | 2 | number of digits that must be correct for `comparison="sigfig"` or `comparison="decdig"`.
 `allow-complex` | boolean | false | Whether or not to allow complex numbers as answers. If the correct answer `ans` is a complex object, you should use `import prairielearn as pl` and `data['correct_answer'][answers-name] = pl.to_json(ans)`.
+`allow-blank` | boolean | false |  Whether or not an empty input box is allowed. By default, empty input boxes will not be graded (invalid format).
+`blank-answer` | string | 0 (zero) | Value to be used as an answer if element is left blank. Only applied if `allow-blank` is `true`. Must follow the same format as an expected user input (e.g., fractions if allowed, complex numbers if allowed, etc.).
```
mfsilva22, line 272: It appears the element has the variable named `blank-value` instead of `blank-answer`

mfsilva22 (commented):
This new feature is awesome! I could not find any errors when checking the examples. I want to update one of the questions from the workshop that needs this feature (but instead, we implemented a parse function), but I will do this after this PR is merged.

elements/pl-integer-input/pl-integer-input.py
```diff
@@ -131,6 +134,9 @@ def render(element_html, data):
 
             html_params['suffix'] = suffix
             html_params['a_sub'] = numpy.base_repr(a_sub, base) if base > 0 else data['raw_submitted_answers'].get(name, str(a_sub))
+            if data['raw_submitted_answers'].get(name, 'X').strip() == '':
+                html_params['a_sub'] = '(blank)'
+
```
mwest1066, line 139: If I'm understanding correctly, this behaves differently than pl-matrix-input? That is, pl-matrix-input will convert a blank to 0 (by default) and display this 0 value, so the student sees it. But pl-integer-input will treat the blank value as 0 (by default) but display `(blank)`. Would it be better to follow the pl-matrix-input convention, both for consistency and so the student actually sees the value that's being used? Especially if the `blank-value` is not zero, I think it might be helpful to show it.

elements/pl-symbolic-input/pl-symbolic-input.py
```diff
@@ -134,7 +136,10 @@ def render(element_html, data):
             else:
                 a_sub = phs.json_to_sympy(a_sub, allow_complex=allow_complex)
             a_sub = a_sub.subs(sympy.I, sympy.Symbol(imaginary_unit))
-            html_params['a_sub'] = sympy.latex(a_sub)
+            if data['raw_submitted_answers'].get(name, '').strip() == '':
+                html_params['a_sub'] = '\\texttt{(blank)}'
```
mwest1066, line 140: Same comment here. I might be misunderstanding, but I think that pl-number-input and pl-matrix-input will both show the `blank-value` in the submission (if it's used), whereas pl-integer-input and pl-symbolic-input will show `blank`. I prefer the former behavior, because it lets the student see what's happening.

elements/pl-integer-input/pl-integer-input.py
jonatanschroeder: Done. I also changed the example to make this clearer.

mwest1066 (approved):

PR #3474, opened 2020-11-20
Fix client files access when no filename provided

### LINKED ISSUE #3349: clientFilesQuestion.js did not provide path argument
Stack trace:

```
node: /PrairieLearn/node_modules/express/lib/response.js:411
node: throw new TypeError('path argument is required to res.sendFile');
node: ^
node: TypeError: path argument is required to res.sendFile
node: at ServerResponse.sendFile (/PrairieLearn/node_modules/express/lib/response.js:411:11)
node: at Object.<anonymous> (/PrairieLearn/pages/clientFilesQuestion/clientFilesQuestion.js:24:13)
node: at cb (util.js:207:31)
node: at processTicksAndRejections (internal/process/task_queues.js:84:21)
systemd: prairielearn4.service: main process exited, code=exited, status=1/FAILURE
```

mwest1066:
Fixes #3349

### DIFF
--- a/pages/clientFilesAssessment/clientFilesAssessment.js
+++ b/pages/clientFilesAssessment/clientFilesAssessment.js
@@ -1,7 +1,8 @@
-var path = require('path');
-var express = require('express');
-var router = express.Router();
+const path = require('path');
+const express = require('express');
+const router = express.Router();
 
+const error = require('@prairielearn/prairielib/error');
 const chunks = require('../../lib/chunks');
 const ERR = require('async-stacktrace');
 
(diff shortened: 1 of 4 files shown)


nicknytko (approved):
Grepped around in the source, only other file that looks like it would need checks is `pages/elementFiles/elementFiles.js`.  Otherwise, looks good to me.

mwest1066:
> Grepped around in the source, only other file that looks like it would need checks is `pages/elementFiles/elementFiles.js`. Otherwise, looks good to me.

Thanks!

PR #3475, opened 2020-11-20
New element pl-xss-safe

jonatanschroeder:
Can be used when showing HTML or Markdown provided by the student.


mfsilva22:
This element will be helpful especially when the manual grader is ready. In general, I think elements should be packaged in a way that eliminates the need for an average user to implement additional code, such as the parse function in the example. Wondering if we can make changes to this element, so that an example question like the one you added (a simple case) would require very little coding in server.py.

mfsilva22 (commented):
This is so much cleaner! Love it :-)

Can you please:
- add your question (`demo/markdownEditorLivePreview`) to the assessment `hw04-demoQuestions`
- add your netid to the question

As I was reviewing this PR, I noticed that another markdown question `element/markdown` was not included in any assessment. Can you please add that question to the assessment `hw03-elementExamples`

docs/elements.md
```diff
@@ -1577,6 +1578,36 @@ Attribute | Type | Default | Description
...
 
+### `pl-xss-safe` element
+
+Removes potentially dangerous scripts from an HTML code. This is recommended when parsing and displaying student-provided code. Note that any code parsed by this element must be supported directly by the browser, i.e., it cannot include PrairieLearn elements or special tags.
+
+#### Sample element
+
+```html
+<pl-xss-safe>
+{{{submitted_answers.answer}}}
```
mwest1066, line 1589: What will happen if `submitted_answers.answer` contains an unmatched `</pl-xss-safe>` followed by some dangerous code?

elements/pl-xss-safe/pl-xss-safe.mustache
```diff
@@ -0,0 +1,14 @@
+<script>
+(function() {
+    let code = {{{quoted_code}}};
+    let language = "{{language}}";
+    if (language == "markdown") {
+        let renderer = new showdown.Converter({
```
mwest1066, line 6: Where does `showdown` get included on the page? Does this need to be listed as a client-side dependency?

elements/pl-xss-safe/pl-xss-safe.mustache
```diff
@@ -0,0 +1,14 @@
...
+            'literalMidWordUnderscores':true,
+            'literalMidWordAsterisks':true,
+        });
+        code = renderer.makeHtml(code);
+    }
+    document.write(filterXSS(code));
```
mwest1066, line 12: Where is the `filterXSS` function coming from?

mwest1066: Can you help me understand this `document.write()` here? Does that replace the entire document? Or how does it know where to insert/replace?

exampleCourse/questions/demo/markdownEditorLivePreview/question.html
```diff
@@ -9,5 +9,6 @@
 </pl-question-panel>
 
 <pl-submission-panel>
+  <pl-xss-safe submitted-file-name="ans.md" language="markdown"></pl-xss-safe>
```
mwest1066, line 12: In addition to this demo question, should we also add `questions/element/xssSafe` to demonstrate it in isolation?

docs/elements.md
jonatanschroeder: That's a great point. Probably good to only get input from files or feedback fields directly. I'll change it to account for that.

elements/pl-xss-safe/pl-xss-safe.mustache
jonatanschroeder: This was an oversight on my part, these come from dependencies in `pl-file-editor` that should have been added as dependencies here as well. I just hadn't tested it in a question without this element. I'll review it.

elements/pl-xss-safe/pl-xss-safe.mustache
jonatanschroeder: > Where is the `filterXSS` function coming from?

Same as `showdown` comment above.

> Can you help me understand this `document.write()` here? Does that replace the entire document? Or how does it know where to insert/replace?

`document.write()` inserts the HTML in the current position in the file. Note that the function is called immediately while page loading is still in progress, not after loading is complete.

exampleCourse/questions/demo/markdownEditorLivePreview/question.html
jonatanschroeder: Probably a good idea, maybe in combination with a `pl-file-upload` and with different options.

docs/elements.md
jonatanschroeder: Maybe allow for something like `<pl-xss-safe content="{{submitted_answers.answer}}"></pl-xss-safe>`, since that would ensure the case above never happens?

elements/pl-xss-safe/pl-xss-safe.mustache
mwest1066: Makes sense!

docs/elements.md
mwest1066: Is mustache quoting guaranteed to be safe for use in an attribute value? I thought it's quoting rules were only safe for HTML? I'm thinking about someone putting `"` characters in `submitted_answers.answer`, for example.

elements/pl-xss-safe/pl-xss-safe.mustache
mwest1066: Great! Thanks for the explanation.

exampleCourse/questions/demo/markdownEditorLivePreview/question.html
mwest1066: That would be great.

docs/elements.md
jonatanschroeder: Yes, it replaced that with &quot;

mwest1066: Ok, sounds good to me then.

docs/elements.md
```diff
@@ -1615,6 +1616,42 @@ Attribute | Type | Default | Description
 - [demo/autograder/python/plots]
 - [demo/autograder/python/random]
 
+### `pl-xss-safe` element
+
+Removes potentially dangerous scripts from an HTML code. This is recommended when parsing and displaying student-provided code. The element will remove some elements like scripts and triggers that may have been maliciously inserted by the student. Note that any code parsed by this element must be supported directly by the browser, i.e., it cannot include PrairieLearn elements or special tags.
```
mwest1066, line 1621:
```suggestion
Removes potentially dangerous scripts from HTML. This is recommended when parsing and displaying student-provided content. The element will remove some elements like scripts and triggers that may have been maliciously inserted by the student. Note that any code parsed by this element must be supported directly by the browser, i.e., it cannot include PrairieLearn elements or special tags.
```

exampleCourse/questions/element/xssSafe/question.html
```diff
@@ -0,0 +1,44 @@
+<pl-question-panel>
+  <p>
+    Questions below were designed to showcase the different features of 
+    <a href="http://prairielearn.readthedocs.io/en/latest/elements/#pl-xss-safe-element"><code>pl-xss-safe</code></a>. 
+    The goal of <code>pl-xss-safe</code>
+    is to allow instructors to display to students formatted HTML code with potentially malicious source, such as student-submitted code.
```
mwest1066, line 6:
```suggestion
    is to allow instructors to display to students formatted HTML with potentially malicious source, such as student-submitted content.
```

mwest1066 (approved):

mwest1066:
> Ensuring HTML isn't erroneously considered code :-D

😄

PR #3518, opened 2020-12-03
Add Group Contributions to Assessment Instances page

yScribblez:
* submission migration to add user id column to submissions table
* moved assessment instances page query to a stored procedure
* added logic to compute contribution for each group member, display as mini histogram
Screen Shot 2020-12-02 at 3 49 16 PM


pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -270,8 +293,8 @@
                   widgets: ["uitheme"],
                   headers: {
                     <% if (assessment.group_work) { %>
-                      7: {sorter: false},
                       8: {sorter: false},
+                      9: {sorter: false},
```
tdy, line 297: Does the `else` block need to be bumped to 9 and 10? Otherwise it looks like we can just ditch the `if`/`else`.

jonatanschroeder:
Note that the assessment instances pages was recently revamped, so this PR needs some rework. Are you still interested in moving forward with it?

PR #3520, opened 2020-12-04
Fix thin lines in math on Chrome/Windows

### LINKED ISSUE #3086: MathJax SVG rendering does not display correctly on low-dpi displays
Reported by Sarka. Thin lines in math SVGs vanish. For example, some of the lines in `+` signs are missing, especially in fractions.

@echuber2 reports this is a known issue in Chrome: https://techcommunity.microsoft.com/t5/discussions/svg-antialiasing-issues/m-p/788141

mwest1066:
This is a hack to default to the CommonHTML MathJax renderer (which works on lo-dpi displays with Chrome/Windows) and only use the SVG MathJax renderer for elements that need it (at the moment only `pl-drawing`). Elements can choose to use SVG by adding the class `pl-requires-svg-mathjax` to their output HTML.

Fixes #3086

### DIFF
--- a/elements/pl-drawing/pl-drawing.mustache
+++ b/elements/pl-drawing/pl-drawing.mustache
@@ -20,7 +20,7 @@
 </span>
 {{/parse_error}}
 {{^parse_error}}
-<div id="drawing-interface-{{uuid}}" class="pl-drawing-container">
+<div id="drawing-interface-{{uuid}}" class="pl-drawing-container pl-requires-svg-mathjax">
     <div>
         <canvas width="{{width}}" height="{{height}}"> </canvas>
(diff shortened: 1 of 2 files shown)


nicknytko (approved):
Can confirm this fixes rendering on chrome/windows

before:
image

after:
image


Strangely enough, I couldn't get the rendering issue to show on Linux+Chromium or Firefox on any platform.

Also, I took out that small hack for SVG rendering since it seems to have been fixed in MathJax.

mwest1066:
> Also, I took out that small hack for SVG rendering since it seems to have been fixed in MathJax.

Awesome! Thanks!

PR #3521, opened 2020-12-04
Set submission to gradable if grade uploaded.

jonatanschroeder:
In file upload, if new grade is uploaded, set the submission to gradable.

### DIFF
--- a/sprocs/instance_questions_update_score.sql
+++ b/sprocs/instance_questions_update_score.sql
@@ -112,7 +112,8 @@ BEGIN
             grading_method = 'External',
             override_score = new_score,
             score = COALESCE(new_score, score),
-            correct = COALESCE(new_correct, correct)
+            correct = COALESCE(new_correct, correct),
+            gradable = CASE WHEN new_score IS NULL THEN gradable ELSE TRUE END
         WHERE s.id = submission_id;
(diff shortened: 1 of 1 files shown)


mwest1066:
I think we also need to update `sprocs/instance_questions_update_points.sql`? We seem to use one or the other: https://github.com/PrairieLearn/PrairieLearn/blob/93b4214a4d92f550c8bdc54aa579934122f0655a/lib/score-upload.js#L355 and https://github.com/PrairieLearn/PrairieLearn/blob/93b4214a4d92f550c8bdc54aa579934122f0655a/lib/score-upload.js#L365

jonatanschroeder:
> I think we also need to update `sprocs/instance_questions_update_points.sql`? We seem to use one or the other:
> 
> https://github.com/PrairieLearn/PrairieLearn/blob/93b4214a4d92f550c8bdc54aa579934122f0655a/lib/score-upload.js#L355
> 
> and
> https://github.com/PrairieLearn/PrairieLearn/blob/93b4214a4d92f550c8bdc54aa579934122f0655a/lib/score-upload.js#L365

Those are for instance score, not question score. There is no `instance_questions_update_points`, the points update uses the same sproc as percent.

PR #3523, opened 2020-12-07
Fix unicode characters in server.py

mwest1066:


nicknytko:
I wasn't able to get unicode in `server.py` working until I added that encoding flag to the python trampoline.  I'm not familiar with the code paths for the non forking version, do you think we'd need to do something similar?

mwest1066:
> I wasn't able to get unicode in `server.py` working until I added that encoding flag to the python trampoline.  I'm not familiar with the code paths for the non forking version, do you think we'd need to do something similar?

Do we still use the non-forking version for anything? Can that even be activated?

PR #3524, opened 2020-12-07
Fix pl-dropdown to allow unescaped values

mwest1066:


elements/pl-dropdown/pl-dropdown.mustache
```diff
@@ -12,7 +12,7 @@
     <div id="pl-dropdown-{{uuid}}" class="d-inline-block form-group form-inline">
         <select name="{{answers-name}}" class="form-control">
             {{#options}}
-                <option value="{{value}}" {{#selected}}selected{{/selected}}>{{value}}</option>
+                <option value="{{value}}" {{#selected}}selected{{/selected}}>{{{value}}}</option>
```
echuber2, line 15: Does the value attribute need triple braces also?

jonatanschroeder: That's not even enough. There are some problems when the value submitted in the form includes the &. Reviewing now.

jonatanschroeder: It works properly without unescaping value. I added the change in submitted and correct panels as well.

jonatanschroeder (approved):

mwest1066:
@jonatanschroeder Good call on the submitted and correct answers!

PR #3465, opened 2020-11-13
Manual question config

andrewstec:
@mebird Copy and paste of our plan from Slack:

- [ ] Backend: Create a manual grading flag in the config. Give a percentage towards the manual graded score. Dictate any DB changes/workflow changes.
- [ ] UI: Load any question with this view. Do not allow students to access this page. Only render for pages with a grading config. Get graded score data (manual & automatic) and hit an endpoint statically.
JSON structure in config file.
- [ ] How to resolve changes to the config file (do we have a reset/make happy button)?
- [ ] Future: building instructor menus, updating grading toggles. This menu will likely have a button to indicate if the config file changes.

PR #3526, opened 2020-12-09
Manual grading submission config

andrewstec:
Merged from https://github.com/PrairieLearn/PrairieLearn/pull/3465 and https://github.com/PrairieLearn/PrairieLearn/pull/3477.


mwest1066 (commented):
Can we make two main changes to this PR:
1. Remove all the changes that are also made in #3477
2. Remove the "reserved points" functionality for now.

I'd like to get an initial version of manual grading merged that is "all or nothing". That is, if a question is manually graded then it can't have any other kind of grading (no internal or external) and it is 100% manually graded.

As soon as we've got this basic implementation working then we can add all the other cool features 😄

andrewstec:
Closing this, but keeping the branch as an artifact to base future work on.

PR #3541, opened 2020-12-13
Upgrade to Python 3.8, NodeJS 14, and R 4.0

mwest1066:
This changes the base image from CentOS (discontinued) to Amazon Linux.

Python is now being installed with https://github.com/conda-forge/miniforge

Still need to:
* [x] Add a news item
* [x] Upgrade `graders/python` the same way


mwest1066:
The `ci/dockercloud-stage` failure is expected because this PR changes the name of the base image to `prairielearn/plbase` (from centos7-plbase) and so the images won't autobuild anymore. I'll fix this after we've merged this PR.

environments/plbase/plbase-install.sh
```diff
@@ -1,35 +1,33 @@
...
-    https://repo.ius.io/ius-release-el7.rpm
+yum update -y
 
-yum -y update
+amazon-linux-extras install -y \
+    vim \
+    docker \
+    postgresql11 \
+    redis4.0 \
+    R3.4
```
coatless, line 10:
```suggestion
    R3.4
```

Unfortunately, this needs to get bumped to R3.6 or R 4.0. The bad part: Amazon Linux Extras doesn't support higher _R_ versions. The good part, EPEL does. 

```bash
yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
yum -y install R
```

per: https://stackoverflow.com/questions/59415973/how-to-install-newer-version-of-r-on-amazon-linux-2

mwest1066: Sounds good. I switched to 3.6 from EPEL. I couldn't see an easy way to get 4.0?

coatless: Two options for installing _R_ 4.0: 

- Upgrade to EPEL-8 instead of 7 (which was used above)
  - EPEL-8 ships with **R 4.0.3** per http://mirrors.sonic.net/epel/8/Everything/SRPMS/Packages/r/ 
  - EPEL-7 ships with **R 3.6.0** per https://mirror.lax.genesisadaptive.com/epel/7/SRPMS/Packages/r/
- Install R using `conda` via `conda-forge`. 
  - I think this approach also would provide binary copies of R packages and, thus, cut down on image builds as the packages wouldn't be compiled from source. 

```bash
conda install -c conda-forge r-essentials
```

For more details, see the formula at: https://anaconda.org/conda-forge/r-essentials

With this being said, deploying with conda has the added benefit of using an `environment.yml` file storing package information in a bit more structured way than `requirements.txt`. Take for example: 

```yaml
name: plbase
channels:
  - defaults
  - conda-forge
dependencies:
  - python=3.8   # or 2.7
  - r-essentials=4.0.*
  - numpy=1.19.*
  - matplotlib
  - package-name-here # more packages from conda ... 
  - pip: # download using pip
    - package-name-here
```

c.f. https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#create-env-file-manually

nicknytko (commented):
I don’t see anything immediately wrong just looking at the files — will try running and report back.

environments/plbase/plbase-install.sh
mwest1066: Conda is a great idea for installing R! I switched to that, so now we have R 4.0.3.

I can't figure out how to use an `environment.yml` file with the root conda environment. I'm avoiding using a new conda environment because I don't have a good way to activate it by default when starting the container (because we don't seem to run startup scripts when starting the container). This can be worked around using explicit bash calls for CMD in the container, but that's a bit ugly. Also I don't think dependabot supports `environment.yml`, which is another reason to avoid it.

nicknytko:
Python stuff looks to be running well for me, I wasn't able to break anything in `plbase` or the python grader.

mwest1066:
> Python stuff looks to be running well for me, I wasn't able to break anything in `plbase` or the python grader.

Awesome, thanks!

environments/plbase/plbase-install.sh
coatless: > Conda is a great idea for installing R! I switched to that, so now we have R 4.0.3.

👍 

>  Also I don't think dependabot supports environment.yml, which is another reason to avoid it.

@mwest1066 good point on the dependabot. Let's just forget it then.

> I can't figure out how to use an environment.yml file with the root conda environment

Okay, no worries. Instead, let me just remake the `r-requirements.txt` to be based off of conda packages. This will simplify the build still (no compiling, plus avoiding hiccups.)

I think I can branch off of this PR or I'll just post a diff here.

mwest1066: > Okay, no worries. Instead, let me just remake the `r-requirements.txt` to be based off of conda packages. This will simplify the build still (no compiling, plus avoiding hiccups.)
> 
> I think I can branch off of this PR or I'll just post a diff here.

Sounds great!

I'll go ahead and merge this PR once CI passes, but I won't upgrade production until sometime in the coming days. That way you can just branch off the new master.

PR #3543, opened 2020-12-14
Added old feedback to manual grading download file.

### LINKED ISSUE #3387: Include old feedback in manual grading CSV
The current manual grading CSV includes the old score and a column for new score and feedback, but not the old feedback. If we were to update the feedback we'd need to keep track of the old CSV file. Please include the old feedback there so we can update grades after uploading them the first time.

It would also be great to have a way to change the feedback without involving a CSV file.

jonatanschroeder:
Resolves #3387


mwest1066:
@jonatanschroeder this is the most minor nitpick of all, but we've been writing PR titles with "Add" rather than "Added", and without the period on the end.

PR #3548, opened 2020-12-15
Use new question HTML renderer by default

mwest1066:


mwest1066:
This will need to be updated in light of #4847

nwalters512:
I'm taking a closer look at this now. A useful query to run against production databases:

```
select id, short_name, title, options, repository from pl_courses where options->>'useNewQuestionRenderer' != 'true';
```

This yields 4 courses for the Canada instance and (gulp) 84 courses for the US instance.

I think it's time to get serious about making this change.

- Announce in #announce on Slack that this change will be merged by XX/YY/ZZZZ (I propose we land this by the end of 2021.
- Use https://github.com/nwalters512/pl-question-migration to automatically migrate existing repos; include that in the above announcement.
- Follow up with courses that aren't compliant by Dec 1, 2021.
- Merge and deploy this around Dec 30, 2021.

nwalters512:
Going to close this out. While we still want to do this work, we aren't immediately able to devote the planning/resources necessary to make this happen.

PR #3568, opened 2020-12-21
Link assessment instance from question preview when showing variant linked to instance question

### LINKED ISSUE #1350: Issues pagination has incorrect number of pages
When we paginate the issues list, we compute the number of pages (and whether we should paginate) based on the total issue count: https://github.com/PrairieLearn/PrairieLearn/blob/master/pages/instructorIssues/instructorIssues.js#L113

We should instead apply the filters when doing this, to obtain a count of the total number of issues with the current filter applied, and use this for pagination.

The change will require using `issues_select_with_filter` in `BLOCK issues_count`, in the same way as we do in `BLOCK select_issues`.

### LINKED ISSUE #3304: Pagination problem on issues page
On the issues listing, if there are more than 100 issues, pagination takes place. But if the current filter includes less than 100 items, the pagination still shows, even though a second page wouldn't have anything.

### LINKED ISSUE #3549: From Issues page add link to student's assessment instance
The issues page links to the student's assessment instance question. It would be really helpful if it also had a link to the student's assessment instance itself. The reasoning is that sometimes an issue requires us to review grades manually, and currently we need to navigate to the assessment, find the student, go to Details and then change it.

The link could come from the assessment instance question page itself.

jonatanschroeder:
Fixes #3549 
Also fixes #3304 
Also fixes #1350


echuber2:
If you're not checking the `assessment` local anymore, does this part need to change too?

https://github.com/PrairieLearn/PrairieLearn/blob/b494a32772e92139595bd62ba52f3b7bdcee654c/lib/question.js#L1334-L1337

I'm not sure I'm the best one to comment on the implementation of this one, but I could test it out later.

pages/instructorIssues/instructorIssues.ejs
```diff
@@ -170,9 +170,15 @@
               <% } else { %>
               <i class="fa fa-check-circle text-success issue-status-icon"></i>
               <% } %>
+              <% if (row.instance_question_id) { %>
+              <a class="d-block" href="/pl/course_instance/<%= row.course_instance_id %>/instance_question/<%= row.instance_question_id %>/?variant_id=<%= row.variant_id %>">
```
jonatanschroeder, line 174: @mwest1066 I changed the issues page to link to instance_question when possible. I believe the `urlPrefix` should not be used in this context, since you may have the Issues page from the course context instead of the instance context, can you check if that is indeed the case? Is there a better way to handle this link?

pages/partials/instructorInfoPanel.ejs
```diff
@@ -69,26 +69,38 @@
...
   <% } %>
 
-  <% if (typeof assessment !== "undefined") { %>
+  <% let instructorUrlPrefix = urlPrefix;
+     let ip_assessment = null;
+     let ip_assessment_instance = null;
+     if (typeof assessment !== "undefined") {
+        ip_assessment = assessment;
+        ip_assessment_instance = assessment_instance;
+        instructorUrlPrefix = urlPrefix + '/instructor';
```
mwest1066, line 78: This feels a bit fragile, because it's relying on `assessment` only being defined on student pages (which I agree is currently true, but might not be true in the future). To make it clearer what's going on, could we construct everything from `plainUrlPrefix`? And test for `assessment_instance !== "undefined"` rather than `assessment`? It feels to me logical that having an `assessment_instance` implies having an `assessment`, but at the moment we are doing it the other way around.

That is, I'm thinking about something like:
* If we have an `assessment_instance`, assume we also have an `assessment` and `course_instance` and explicitly construct the URL starting from `plainUrlPrefix`.

pages/partials/instructorInfoPanel.ejs
```diff
@@ -69,26 +69,38 @@
...
+     }
+     if (typeof variant !== "undefined" && variant.assessment) {
+        ip_assessment = variant.assessment;
+        ip_assessment_instance = variant.assessment_instance;
```
mwest1066, line 82: Could we separately test for `variant.assessment_instance` to be robust in the future if we have variants associated with assessments but not assessment instances?

pages/instructorIssues/instructorIssues.ejs
```diff
@@ -170,9 +170,15 @@
               <% } else { %>
               <i class="fa fa-check-circle text-success issue-status-icon"></i>
               <% } %>
+              <% if (row.instance_question_id) { %>
+              <a class="d-block" href="<%= plainUrlPrefix %>/course_instance/<%= row.course_instance_id %>/instance_question/<%= row.instance_question_id %>/?variant_id=<%= row.variant_id %>">
```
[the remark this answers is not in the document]
jonatanschroeder: https://github.com/PrairieLearn/PrairieLearn/issues/3549#issuecomment-750345502

> I'm sorry to only now comment on this after you already implemented the other option in #3568, but I actually think it would be conceptually cleaner to fix the issue with the idea above. **That is, link to the student view of the question**. This way we can keep "instructor view of question" and "student view of question" completely separate. I worry that it's confusing to show the instructor view of a question but with a link to a student's assessment instance. I also think it's conceptually cleaner to link from the issue to the page that the student was on when they filed the issue.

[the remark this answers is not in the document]
jonatanschroeder: Emphasis added.

pages/partials/instructorInfoPanel.ejs
jonatanschroeder: > Could we separately test for `variant.assessment_instance` to be robust in the future if we have variants associated with assessments but not assessment instances?

That would probably rely on a different view anyway. I'll change the logic to use `assessment_instance` (the other conversion) instead of assessment, and leave the view of having variants linked to assessments when they come.

pages/partials/instructorInfoPanel.ejs
jonatanschroeder: Done.

pages/instructorIssues/instructorIssues.ejs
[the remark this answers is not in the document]
jonatanschroeder: And yes, it supports variant_id (this is linked to from the assessment log, when selecting a specific instance (e.g., S-A1.5#1).

jonatanschroeder:
Another option: in the Issues list, we can have three buttons: instructor view of the question, student view of the question, or direct to assessment log. They could be in a dropdown, if clutter is a concern. Future versions could then also have a button that points straight to the manual grading page as well, as @cheeren has pointed out somewhere (I know remember where).

pages/instructorIssues/instructorIssues.ejs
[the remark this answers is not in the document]
jonatanschroeder: In any case, switched back to the instructor view, so the link has a consistent destination. The student view is now added to the dropdown on the side.

pages/instructorIssues/instructorIssues.ejs
```diff
@@ -170,9 +170,23 @@
...
               <i class="fa fa-check-circle text-success issue-status-icon"></i>
               <% } %>
-              <a class="d-block" href="<%= urlPrefix %>/question/<%= row.question_id %>/?variant_id=<%= row.variant_id %>">
-                <%= row.question_qid %>
-              </a>
+              <div class="d-block">
+                <a href="<%= urlPrefix %>/question/<%= row.question_id %>/?variant_id=<%= row.variant_id %>">
+                  <%= row.question_qid %>
+                </a>
+                <button class="btn btn-xs dropdown-toggle" type="button" id="issue-link-menu-<%= row.issue_id %>"
```
mwest1066, line 177: I didn't notice this dropdown at first in the UI. It was only because I'd seen the code that I knew where to find it 😄

To make this more visible, how about making the `question_qid` not be a link and adding links afterwards. For example:

**demo/calculation** ([instructor view](#), [student view](#), [assessment details](#))

sprocs/variants_select.sql
```diff
@@ -10,9 +10,16 @@ DECLARE
...
+    SELECT
+        v.*,
+        to_jsonb(a.*) AS assessment,
+        to_jsonb(ai.*) AS assessment_instance,
+        ai.date AS assessment_instance_date
     INTO variant_with_id
     FROM variants as v
+    LEFT JOIN instance_questions AS iq ON (iq.id = v.instance_question_id)
+    LEFT JOIN assessment_instances AS ai ON (ai.id = iq.assessment_instance_id)
+    LEFT JOIN assessments AS a ON (a.id = ai.assessment_id)
```
mwest1066, line 22: Nit: indentation.
```suggestion
    FROM
        variants as v
        LEFT JOIN instance_questions AS iq ON (iq.id = v.instance_question_id)
        LEFT JOIN assessment_instances AS ai ON (ai.id = iq.assessment_instance_id)
        LEFT JOIN assessments AS a ON (a.id = ai.assessment_id)
```

sprocs/variants_select.sql
```diff
@@ -34,6 +41,11 @@ BEGIN
     WHERE
         q.id = variant_with_id.question_id;
 
+    IF variant_with_id.instance_question_id IS NOT NULL THEN
```
mwest1066, line 44: How can `variant_with_id.instance_question_id` ever be NOT NULL? I can't see where it would get set above.

I also can't see where `variant_with_id.course_instance_id` (line 27) would be set?

pages/instructorIssues/instructorIssues.ejs
```diff
@@ -170,9 +170,23 @@
...
+                        data-toggle="dropdown" aria-haspopup="true" aria-expanded="false"></button>
+                <div class="dropdown-menu" aria-labelledby="issue-link-menu-<%= row.issue_id %>">
+                  <a class="dropdown-item" href="<%= urlPrefix %>/question/<%= row.question_id %>/?variant_id=<%= row.variant_id %>">Instructor view</a>
+                  <% if (row.instance_question_id) { %>
+                  <a class="dropdown-item" href="<%= plainUrlPrefix %>/course_instance/<%= row.course_instance_id %>/instance_question/<%= row.instance_question_id %>/?variant_id=<%= row.variant_id %>">Student view</a>
```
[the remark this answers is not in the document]
jonatanschroeder: It can for Homework questions, this is what's done in the assessment log when you click on the student variant.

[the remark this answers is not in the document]
mwest1066: Ah! Very nice! Good call.

pages/instructorIssues/instructorIssues.ejs
jonatanschroeder: That can work, though I can also make the whole thing a button with the dropdown.

mwest1066: It seems that we've got lots of horizontal space available, so I'm inclined to make the links directly clickable without a dropdown?

sprocs/variants_select.sql
jonatanschroeder: instance_question_id and course_instance_id come from v.* in the query. They are null when the issue is caused by an instructor in question preview (e.g., a syntax error in server.py).

pages/instructorIssues/instructorIssues.ejs
jonatanschroeder: sure, I can do that

jonatanschroeder: done.

sprocs/variants_select.sql
mwest1066: D'oh! Thank you for explaining to those of us who are a bit slow 😄

PR #3597, opened 2020-12-21
UI Work for First Pass of Manual Grading

mebird:
- Refactors GET endpoint to use the `async.series` pattern so it's easier to rearrange steps
- Moves the grading interface into a side panel, which acts as an overlay on top of the question interface


mwest1066 (approved):
I'm afraid that this will need to be rebased on #3477 once that is edited in response to my recent review. Other than the changes to the underlying #3477, everything in this PR looks good to me.

mwest1066:
@mebird how do you want to proceed with this PR? Do you want to wait updates to the `manual-grading-submission` branch as part of #3477 and then rebase this PR onto the new version, or do you want to merge this into the `manual-grading-submission` branch now? Either is fine with me.

mebird:
> @mebird how do you want to proceed with this PR? Do you want to wait updates to the `manual-grading-submission` branch as part of #3477 and then rebase this PR onto the new version, or do you want to merge this into the `manual-grading-submission` branch now? Either is fine with me.

There's one or two issues I'm debugging locally with missing submission IDs, so this PR is going to need another day and a bit. I rebase this PR onto the new version, since it's likely some of the async code I touched is going to change.

mwest1066:
> There's one or two issues I'm debugging locally with missing submission IDs, so this PR is going to need another day and a bit. I rebase this PR onto the new version, since it's likely some of the async code I touched is going to change.

Sounds great! I should be able to turn around code reviews and merges very rapidly while we are in winter break (until the end of January), so keep them coming!

docs/docker.md
```diff
@@ -72,6 +72,51 @@ E.g., to start a shell in a container started with `--name pl`:
 docker exec -it pl /bin/bash
 ```
 
+## Docker-Compose
+
+### Basics
+
+A docker-compose file describes the services an application needs to run. In our case, we use `docker-compose` to configure and run the PrairieLearn docker container locally.
+
+To run PrairieLearn with `docker-compose`, run `docker-compose up pl`. This will, in order:
```
andrewstec, line 81: You can use a numbered list to denote the sequential nature of the steps.

andrewstec: I don't know if search engines read markdown syntax for semantics yet, but I wouldn't be surprised if they do. They do read semantics in HTML.

mebird:
Pointing towards master instead.

Issue #3347, opened 2020-10-15
Unify homework and exam assessments


mwest1066:
### What's the problem?

At the moment we have two "types" of assessment, namely `type: Homework` and `type: Exam`. These have different default behaviors. This division results in lots of duplicated code internally and makes it harder to explain what's going on.

### Unification strategy

As @mfsilva22 has suggested, we should unify these types with the following strategy:

1. Add options to `type: Exam` assessments so that they can exactly mimic the behavior of `type: Homework`.
2. Make `type` an optional parameter in `infoAssessment.json`, defaulting to `Exam`.
3. Change the sync code so that assessments with `type: Homework` in JSON are synced to the DB as `type: Exam`, but with other properties set to mimic the traditional `Homework` behavior.
4. Run a migration on the DB to change existing `Homework` assessments to `Exam`, adding any additional properties to mimic existing behavior.
5. Remove all code for `type: Homework`.
6. Change the documentation to remove all reference to `type` and only discuss assessments in general.

### What specific changes are needed?

* Homeworks don't auto-close after 6 hours of inactivity, unlike exams. We can either default `autoClose: false` for homeworks, or we can implement the changes to auto-close described in #3346 to remove the auto-close setting entirely and only auto-close after 1-hour of non-access (which shouldn't happen for homeworks).
* Question scoring on homeworks allows unlimited attempts, with new variants being generated each time. We will need to add a column to `assessment_questions` for whether new variants should be generated.
* Add a setting for `assessment_questions` to mimic the behavior of `questions.singleVariant`. This would also allow exams to have "infinite retry" questions. We can then deprecate `questions.singleVariant` (see #2357).
* What else?

trombonekenny:
This came up in office hours this week too... the desire to set attributes for assessments that might not exist in type: homework or type: exam

nwalters512:
Small detail from a conversation with @mwest1066 today: to differentiate from Exam-style "fixed number of attempts" and Homework-style "unlimited new variants allowed" via if points is an array or not:

- `"points": 1`: student can achieve full credit with a single 100% submission, but may try an unlimited number of times to get 1 point.
- `"points": 1, "maxPoints": 5`: student must try the question multiple times to achieve full credit.
- `"points": [1]`: student has a single attempt and will earn up to one point; no retries.
- `"points": [3, 2, 1]`: student can earn up to three points on their first submission, and so on.

The key here is that we use an array to differentiate between the second and third cases above.

[...]

PR #3600, opened 2020-12-22
Limit submission rate for a question

### LINKED ISSUE #1195: Rate limit submissions
Add an option to assessment questions to rate limit submissions. It's probably better to have it set per-question and per-assessment, not just in the question itself.

jonatanschroeder:
Resolves #1195 

- [x] Schema updates
- [x] Database updates for configuration
- [x] Question instance UI updates
- [X] Lib updates that block open instance questions from grading a submission based on rate
- [x] Assessment overview UI updates
- [X] Handling of assessment close (grade even if faster than rate?)
- *** Add Grade all/Close all button on assessment instance list
   - will be added in a separate PR
- [x] Create tests
- [x] Documentation


jonatanschroeder:
@mwest1066  Lib is updated and does the necessary checks in the background, so now it's a matter of updating UI, documentation and tests. Regrading is not affected, since it uses a different lib altogether. For UI:

* The instance question UI has a preliminary update, but there will be tons of conflicts with #3601 so I'll wait until you merge that before moving on with that one.
* I like the idea of having a countdown-like interface, but I believe it would be a stretch to use the same partial as the time limit one, since that one has several processes that are specific to that use-case (use of "danger" colors closer to the expiry of the time, submitting a form once the time expires, etc.). I might separate the UI aspect from the functionality aspect of that partial, then use that instead (I need to experiment before committing to something).
* Should there be some indication in the UI for the user, before the submission, that there is a submission rate in effect? Maybe similar to the "additional attempts" note used in #3601? We can always send the responsibility of explaining that to the instructor, in question.html, but from experience this doesn't work well.
* For the assessment overview, right now it properly skips questions blocked due to submission rate from being graded (since it uses the same lib function), but the interface doesn't provide any indication that there is a question that can't be graded due to rate. Any ideas on how to present that?
  * The easiest solution would be to add a note after the "Grade XX answers" button that says something to the effect "at least one saved submission can't be graded because its submission rate has not been reached yet."
  * Another option is to change the counter in this button to only include gradable questions, but that would also need to include a statement on why a question with status "saved" cannot be graded.
  * We can also include an icon next to "saved" status in the question list that opens a popover explaining the submission rate.
* Do we need to provide instructors with some view of the blocked rate in effect for a student? Maybe in the assessment instance log?

mwest1066:
> * I like the idea of having a countdown-like interface, but I believe it would be a stretch to use the same partial as the time limit one, since that one has several processes that are specific to that use-case (use of "danger" colors closer to the expiry of the time, submitting a form once the time expires, etc.). I might separate the UI aspect from the functionality aspect of that partial, then use that instead (I need to experiment before committing to something).

Makes sense. If it's possible to abstract the actual countdown code then that would be great, because this code is slightly subtle and I'd rather not have multiple versions of it. But do whatever you think is best.

> * Should there be some indication in the UI for the user, before the submission, that there is a submission rate in effect? Maybe similar to the "additional attempts" note used in #3601? We can always send the responsibility of explaining that to the instructor, in question.html, but from experience this doesn't work well.

I agree that we should never leave this sort of thing up to the question author. They will always forget or write something unclear 😄. Perhaps we could add a note to the question footer like "Can only be graded once every X minutes"?

> * For the assessment overview, right now it properly skips questions blocked due to submission rate from being graded (since it uses the same lib function), but the interface doesn't provide any indication that there is a question that can't be graded due to rate. Any ideas on how to present that?
>   * The easiest solution would be to add a note after the "Grade XX answers" button that says something to the effect "at least one saved submission can't be graded because its submission rate has not been reached yet."
>   * Another option is to change the counter in this button to only include gradable questions, but that would also need to include a statement on why a question with status "saved" cannot be graded.
>   * We can also include an icon next to "saved" status in the question list that opens a popover explaining the submission rate.

I like the idea of putting an icon (maybe a clock?) next to the "saved" status for questions with a limit in place. And then we can also include the statement about "at least one ... reached yet (marked with [clock-icon] above). Reload the page to update.". By explicitly referencing [clock-icon] I hope it will be clear what is happening. To be clear, I don't think we should live-update these icons. Or should we? Seems like lots of work for minimal gain.

> * Do we need to provide instructors with some view of the blocked rate in effect for a student? Maybe in the assessment instance log?

You mean a real-time view? Or a log for after the assessment is over? I'm not understanding what we would show in either case?

jonatanschroeder:
> > * Do we need to provide instructors with some view of the blocked rate in effect for a student? Maybe in the assessment instance log?
> 
> You mean a real-time view? Or a log for after the assessment is over? I'm not understanding what we would show in either case?

I was thinking more in a sense of helping students figure out why they can't submit for grading, but I guess that's not a big deal now. Instructors can check the student view if needed, and if something else comes up after implementation we can have it as a separate PR.

> I like the idea of putting an icon (maybe a clock?) next to the "saved" status for questions with a limit in place.

I have a stopwatch right now, which looks cute, but is possibly too discrete. Should I maybe add some color or increase the size?

Another question: if a student saves the answer when credit is 100, but then grades when credit is 0, which credit will apply? Which credit should apply? What if they save but never grade (thinking of Homework-style, where you could potentially never close the assessment)? Could there be a mechanism to maybe grade all saved assessments at the deadline?

jonatanschroeder:
image

Any suggestions? Bigger, different color, etc.?

mwest1066:
> Any suggestions? Bigger, different color, etc.?

Somehow that icon looks more like a hand grenade than a clock 😄 

The vertical alignment looks off? Perhaps an `align-middle` would help? https://getbootstrap.com/docs/4.5/utilities/vertical-align/

mwest1066:
Ah, I think you are using `fa-stopwatch`. I have a slight preference for `<i class="far fa-clock"></i>` because it looks more like a clock to me. I don't think I would have guessed what the stopwatch icon is!

jonatanschroeder:
I don't like clocks for countdowns, I replaced it with an hourglass (`fa fa-hourglass-half`)...

For the Homework overview page, it doesn't have a "Grade XX saved answers", nor does it show "saved" status. I imagine keeping that information in the question only, and not messing with the overview page, may be the way to go.

Repeating it here in case you missed it: 
> if a student saves the answer when credit is 100, but then grades when credit is 0, which credit will apply? Which credit should apply? What if they save but never grade (thinking of Homework-style, where you could potentially never close the assessment)? Could there be a mechanism to maybe grade all saved assessments at the deadline?

Would it be viable/acceptable to have some way to, from the list of submissions under a question/instance question, have some button that grades that specific submission (could be limited to the last one) instead of requiring a "Save & Grade" to create a new submission? If such a button exists, we could use the credit at the time of saving instead of grading, if you think that's appropriate.

jonatanschroeder:
I'm working on the countdown partial assuming #3386 will be approved. I'll keep an eye for changes there so they're properly applied in this code.

mwest1066:
> I don't like clocks for countdowns, I replaced it with an hourglass (`fa fa-hourglass-half`)...

Looks great!

> For the Homework overview page, it doesn't have a "Grade XX saved answers", nor does it show "saved" status. I imagine keeping that information in the question only, and not messing with the overview page, may be the way to go.

Perfect.

> Repeating it here in case you missed it:
> 
> > if a student saves the answer when credit is 100, but then grades when credit is 0, which credit will apply? Which credit should apply? What if they save but never grade (thinking of Homework-style, where you could potentially never close the assessment)? Could there be a mechanism to maybe grade all saved assessments at the deadline?

I believe (but I haven't checked) that when we create the submission we save the current credit as `submissions.credit` and this is what is used for any grading of that submission. This applies even if the grading occurs much later (e.g., a slow external grader, automated grading on assessment close, etc).

If the student saves but never grades on a homework then they never get the points. In practice we've never found this to be a problem. Are you thinking that this could be a problem with rate-limited grading? In general I'm not very enthusiastic about delayed grading. It makes everything so much more complicated and I feel that it's generally hostile to students. What's the use case you are thinking about here?

> Would it be viable/acceptable to have some way to, from the list of submissions under a question/instance question, have some button that grades that specific submission (could be limited to the last one) instead of requiring a "Save & Grade" to create a new submission? If such a button exists, we could use the credit at the time of saving instead of grading, if you think that's appropriate.

I guess we could do that? But I'm not understanding what's the advantage of this?

mwest1066:
> I'm working on the countdown partial assuming #3386 will be approved. I'll keep an eye for changes there so they're properly applied in this code.

I just shifted this to "Review in progress" to get it on the queue.

jonatanschroeder:
> > Would it be viable/acceptable to have some way to, from the list of submissions under a question/instance question, have some button that grades that specific submission (could be limited to the last one) instead of requiring a "Save & Grade" to create a new submission? If such a button exists, we could use the credit at the time of saving instead of grading, if you think that's appropriate.
> 
> I guess we could do that? But I'm not understanding what's the advantage of this?

Use case (which I've seen in the past in courses where we had a rate limit): Homework-type assessment, deadline is 11:59:59, with a limit of one submission per 30 minutes. At 11:45, student submits what they expect to be their final version, but the autograder finds a bug that is easy to fix. They can save at 11:55, but will only be able to grade at 12:15, which is after the deadline. So if they can grade the 11:55 submission, they get the credit, but if they save & grade after that, they don't get the credit.

An alternative is to add a Close button in the Homework assessment overview that would grade all assessments and block future submissions. Another option is to add (in a future PR) the "Close All" button in the #2750 interface, and instructors select it after assessments are done. This would trigger a grade in all assessments. It is necessary because it's a Homework assessment, which doesn't typically include a Close button.

mwest1066:
> Use case (which I've seen in the past in courses where we had a rate limit): Homework-type assessment, deadline is 11:59:59, with a limit of one submission per 30 minutes. At 11:45, student submits what they expect to be their final version, but the autograder finds a bug that is easy to fix. They can save at 11:55, but will only be able to grade at 12:15, which is after the deadline. So if they can grade the 11:55 submission, they get the credit, but if they save & grade after that, they don't get the credit.

Ok, thanks for reminding me about this! I still think rate-limiting is a bad idea, but I know lots of people want to do it so we should support it in good way 😄

> An alternative is to add a Close button in the Homework assessment overview that would grade all assessments and block future submissions. Another option is to add (in a future PR) the "Close All" button in the #2750 interface, and instructors select it after assessments are done. This would trigger a grade in all assessments. It is necessary because it's a Homework assessment, which doesn't typically include a Close button.

I like all these ideas! I'd be very happy to have instructor buttons for "Grade any outstanding submissions" and "Close and grade". These seem clearly good.

I'm also thinking about #3347 and how we can make Exams and Homeworks behave more the same.

I propose:

* We try and eliminate "closing" of assessments as much as possible. Maybe we can completely get rid of it. The only time we really need it at the moment is when a timer expires, and we could just check explicitly for an expired timer (and block access) rather than separately closing the assessment instance. We eliminate the `autoClose` option for Exams.
* For all assessments we automatically grade any outstanding submissions after the deadline/timer has expired. To do this, we add a flag `assessment_instances.has_ungraded_submissions BOOLEAN`. Every time we save a submission for this assessment instance, we set `has_ungraded_submissions := true`. We repurpose `cron/autoFinishExams` to search for any assessment instances that have `has_ungraded_submissions = true` and which are not accessible to the student, grade them, and set `has_ungraded_submissions := false`.
* On the instructorAssessmentInstances page we add a column to show `has_ungraded_submissions`. We also add a button to "Grade all ungraded submissions".
* Getting rid of "closing" would require implementing #3616 so that instructors can give students access to view (but not edit) their exams after they are over, even without the concept of "closing".

What do you think? Can we get rid of closing and move to a world where we always grade last-minute submissions?

jonatanschroeder:
These sounds like great ideas, but I'd propose moving the elimination of "Close" to a separate PR/Issue. As for the implementation, I'd be in favour of having the autoFinishExams grade submissions/instance questions, not assessments. That would mean we don't need to keep the flag with cached information, as caching is prone to errors. The autoFinishExams would grade instance questions with status "saved" and whose assessment has expired (go up instead of down).

For now, I'll try to add "Close all" to #2750, and maybe add a "Grade"/"Grade all" option as well (these are exactly the same library function, only with a `false` instead of a `true` in one of the arguments).

mwest1066:
> These sounds like great ideas, but I'd propose moving the elimination of "Close" to a separate PR/Issue. As for the implementation, I'd be in favour of having the autoFinishExams grade submissions/instance questions, not assessments. That would mean we don't need to keep the flag with cached information, as caching is prone to errors. The autoFinishExams would grade instance questions with status "saved" and whose assessment has expired (go up instead of down).

Oh yes, sorry, I didn't mean to imply that we should make any of these changes as part of this PR! 😄

I'll add a new Issue to document this.

> For now, I'll try to add "Close all" to #2750, and maybe add a "Grade"/"Grade all" option as well (these are exactly the same library function, only with a `false` instead of a `true` in one of the arguments).

Excellent!

mwest1066:
> I'll add a new Issue to document this.

See #3622

jonatanschroeder:
image

Thoughts on UI?

mwest1066:
> Thoughts on UI?

Nice! Maybe "Grading possible in XX min", to sound more positive? 😄

Add a bit more vertical space below the Save&Grade button?

Add a question-mark icon next to "Can only be graded..." to explain that saving is always possible, and to explain what will happen to saved answers after the deadline passes (currently nothing, but we will autograde according to #3622).

jonatanschroeder:
@mwest1066 before I move on to documentation and tests: I added the setting as `submissionRateLimitMin`, and can be set at the assessment top level, in a zone, in a question, or as an alternative for a question set. All 4 set the same field `submission_rate_limit_min` in `assessment_question`, and the first 3 are just defaults for the fourth. Is the name/structure ok or do you want something shorter/more limited?

mwest1066:
> @mwest1066 before I move on to documentation and tests: I added the setting as `submissionRateLimitMin`, and can be set at the assessment top level, in a zone, in a question, or as an alternative for a question set. All 4 set the same field `submission_rate_limit_min` in `assessment_question`, and the first 3 are just defaults for the fourth. Is the name/structure ok or do you want something shorter/more limited?

Based on our experiences with `timeLimitMin`, I actually think we should use `Minutes` rather than `Min`. Many people thought that `Min` was short for `Minimum` and didn't understand what the units were. I don't like the fact that `submissionRateLimitMinutes` is so long, but I'd rather be long than unclear.

jonatanschroeder:
> > @mwest1066 before I move on to documentation and tests: I added the setting as `submissionRateLimitMin`, and can be set at the assessment top level, in a zone, in a question, or as an alternative for a question set. All 4 set the same field `submission_rate_limit_min` in `assessment_question`, and the first 3 are just defaults for the fourth. Is the name/structure ok or do you want something shorter/more limited?
> 
> Based on our experiences with `timeLimitMin`, I actually think we should use `Minutes` rather than `Min`. Many people thought that `Min` was short for `Minimum` and didn't understand what the units were. I don't like the fact that `submissionRateLimitMinutes` is so long, but I'd rather be long than unclear.

Maybe submissionRateMinutes? Or gradeRateMinutes?

mwest1066:
> Maybe submissionRateMinutes? Or gradeRateMinutes?

Oh yes, good point, it's actually grading that we are rate limiting. `gradeRateMinutes` sounds good. Or how about `gradeWaitMinutes` to emphasize that students need to wait this time before they can grade again?

jonatanschroeder:
> > Maybe submissionRateMinutes? Or gradeRateMinutes?
> 
> Oh yes, good point, it's actually grading that we are rate limiting. `gradeRateMinutes` sounds good. Or how about `gradeWaitMinutes` to emphasize that students need to wait this time before they can grade again?

I don't like wait, it sounds like they have to submit then wait to grade. Rate sounds more like it's related to the previous occurrence.

mwest1066:
> I don't like wait, it sounds like they have to submit then wait to grade. Rate sounds more like it's related to the previous occurrence.

That's fair, I'm fine with `gradeRateMinutes`.

sprocs/instance_questions_next_allowed_grade.sql
```diff
@@ -0,0 +1,63 @@
+DROP FUNCTION IF EXISTS instance_questions_next_allowed_grade(bigint);
+CREATE OR REPLACE FUNCTION
+    instance_questions_next_allowed_grade (
+        IN iq_id BIGINT,
+        OUT allow_grade_date TIMESTAMPTZ,
+        OUT allow_grade_formatted_date TEXT,
```
jonatanschroeder, line 6: This will be removed, as it's only used in one place and requires additional queries to get the timezone. The computation of this date may be replaced by the interval as a text (e.g., 5 minutes, or 30 seconds).

pages/partials/examQuestionScore.ejs
```diff
@@ -15,3 +16,10 @@
         <span class="badge badge-warning"><%= Math.floor(instance_question.highest_submission_score * 100) %>%</span>
     <% } %>
 <% } %>
+
+<% if ("allow_grade_left_ms" in instance_question && instance_question.allow_grade_left_ms > 0) { %>
+<a class="btn btn-xs" data-toggle="popover" data-html="true" data-content="This question limits the rate of submissions. No further submissions will be processed until: <%= instance_question.allow_grade_formatted_date %>" data-placement="auto" data-trigger="focus" data-container="body" tabindex="0">
```
jonatanschroeder, line 21: Replace the formatted date with the interval.

jonatanschroeder:
@mwest1066 @nwalters512 it'd be great to have this working by the time we release the first assignment, mid-next week. Do you think this is possible?

pages/partials/examQuestionScore.ejs
```diff
@@ -15,3 +16,15 @@
...
+
+<% if ("allow_grade_left_ms" in instance_question && instance_question.allow_grade_left_ms > 0) { %>
+<a class="grade-rate-limit-popover btn btn-xs" data-toggle="popover" data-html="true" data-content="This question limits the rate of submissions. Further grade allowed <%= instance_question.allow_grade_interval %>" data-placement="auto" data-trigger="focus" data-container="body" tabindex="0">
+    <i class="fa fa-hourglass-half" aria-hidden="true"></i>
+</a>
+<script>
+  window.setTimeout(function() {
+    $('.grade-rate-limit-popover').hide();
+  }, <%= instance_question.allow_grade_left_ms %>);
+</script>
```
mwest1066, line 28: How does this work if there are multiple questions with different timeouts? I think we should either do nothing (that is, remove this script) or do it properly (with unique IDs for each question, etc).

pages/partials/submitRateFooter.ejs
```diff
@@ -0,0 +1,26 @@
...
+  </span>
+  <span class="d-flex align-self-center">
+    <small class="font-italic">Can only be graded once every <%= assessment_question.grade_rate_minutes %> minute<% if (assessment_question.grade_rate_minutes > 1) { %>s<% } %></small>
+    <a class="btn btn-xs" data-toggle="popover" data-html="true" data-content="<p>This question has been set up to limit the rate at which you receive feedback for your work to no more than once every <%= assessment_question.grade_rate_minutes %> minute<% if (assessment_question.grade_rate_minutes > 1) { %>s<% } %>. This means that, once you select <strong>Save &amp; Grade</strong>, you will be unable to use the same option again for this amount of time.</p><p>This limit does not affect your ability to Save your submission. You may save a submission at any time. Instructors have the ability to grade your saved submission, which will take the time you saved your answer into account to account for deadlines and credit. <% if (question_context == "student_exam") { %>A saved submission will also be graded automatically once you close the assessment. <% } %></p>" data-placement="auto" data-trigger="focus" data-container="body" tabindex="0">
```
mwest1066, line 20: I'm trying to use "your answer" rather than "submission" when talking to students. I felt like the details about what instructors can do are not really helpful for students?
```suggestion
    <a class="btn btn-xs" data-toggle="popover" data-html="true" data-content="<p>Once you have clicked <strong>Save &amp; Grade</strong>, you will need to wait <%= assessment_question.grade_rate_minutes %> minute<% if (assessment_question.grade_rate_minutes > 1) { %>s<% } %> before you can grade again.</p><p>You can still save your answer as frequently as you like. <% if (question_context == "student_exam") { %>If your assessment ends before your last saved answer is graded, it will be automatically graded for you.<% } %></p>" data-placement="auto" data-trigger="focus" data-container="body" tabindex="0">
```

pages/studentAssessmentInstanceExam/studentAssessmentInstanceExam.ejs
```diff
@@ -237,6 +237,11 @@
               <% } %>
             </form>
             <ul class="my-1">
+              <% if (suspendedSavedAnswers > 1) { %>
+              <li>There are <%= suspendedSavedAnswers %> saved answers that cannot be graded yet because their grade rate has not been reached. They are marked with the <i class="fa fa-hourglass-half"></i> icon above.</li>
+              <% } else if (suspendedSavedAnswers == 1) { %>
+              <li>There is one saved answer that cannot be graded yet because its grade rate has not been reached. It is marked with the <i class="fa fa-hourglass-half"></i> icon above.</li>
```
mwest1066, line 243:
```suggestion
              <li>There are <%= suspendedSavedAnswers %> saved answers that cannot be graded yet due to grading rate limits. They are marked with the <i class="fa fa-hourglass-half"></i> icon above. Reload this page to update this information.</li>
              <% } else if (suspendedSavedAnswers == 1) { %>
              <li>There is one saved answer that cannot be graded yet due to a grading rate limit. It is marked with the <i class="fa fa-hourglass-half"></i> icon above. Reload this page to update this information.</li>
```

pages/partials/examQuestionScore.ejs
mwest1066: Thinking about it more, I think we should just remove this. We aren't going to update the text that says "XX questions are rate limited", so things will get confusing in any case. I suggested a change to the information text to explicitly say that a page reload is needed for updates.

At some point we can rewrite this page with react and have everything update in real time 😄

pages/studentAssessmentInstanceExam/studentAssessmentInstanceExam.js
```diff
@@ -72,11 +72,17 @@ router.get('/', function(req, res, next) {
...
-            }, 0);
-
+            res.locals.savedAnswers = 0;
+            res.locals.suspendedSavedAnswers = 0;
+            res.locals.instance_questions.forEach((question) => {
+                if (question.status == 'saved') {
+                    if (question.allow_grade_left_ms > 0)
+                        res.locals.suspendedSavedAnswers++;
+                    else
+                        res.locals.savedAnswers++;
```
mwest1066, line 82: Nit: I'll preempt @nwalters512 by requesting braces 😄

schemas/schemas/infoAssessment.json
```diff
@@ -101,6 +101,11 @@
             "description": "Whether the assessment will support group work.",
             "type": "boolean",
             "default": false
+        },
+        "gradeRateMinutes": {
+            "description": "Minimum amount of time (in minutes) between graded submissions of the same question.",
```
mwest1066, line 106:
```suggestion
            "description": "Minimum amount of time (in minutes) between graded submissions to the same question.",
```

schemas/schemas/infoAssessment.json
```diff
@@ -233,6 +238,11 @@
                     "items": {
                         "$ref": "#/definitions/zoneQuestion"
                     }
+                },
+                "gradeRateMinutes": {
+                    "description": "Minimum amount of time (in minutes) between graded submissions of the same question.",
```
mwest1066, line 243:
```suggestion
                    "description": "Minimum amount of time (in minutes) between graded submissions to the same question.",
```

schemas/schemas/infoAssessment.json
```diff
@@ -276,6 +286,11 @@
                     "description": "The maximum number of graded submissions allowed for each question instance.",
                     "type": "number",
                     "default": 1
+                },
+                "gradeRateMinutes": {
+                    "description": "Minimum amount of time (in minutes) between graded submissions of the same question.",
```
mwest1066, line 291:
```suggestion
                    "description": "Minimum amount of time (in minutes) between graded submissions to the same question.",
```

schemas/schemas/infoAssessment.json
```diff
@@ -302,6 +317,11 @@
...
```
mwest1066, line 322:
```suggestion
                    "description": "Minimum amount of time (in minutes) between graded submissions to the same question.",
```

sprocs/instance_questions_next_allowed_grade.sql
```diff
@@ -0,0 +1,56 @@
+DROP FUNCTION IF EXISTS instance_questions_next_allowed_grade(bigint);
+CREATE OR REPLACE FUNCTION
+    instance_questions_next_allowed_grade (
+        IN iq_id BIGINT,
```
mwest1066, line 4: We conventionally call these by their full names in argument lists.
```suggestion
        IN instance_question_id BIGINT,
```

sprocs/instance_questions_next_allowed_grade.sql
```diff
@@ -0,0 +1,56 @@
...
+        OUT allow_grade_date TIMESTAMPTZ,
+        OUT allow_grade_left_ms BIGINT,
+        OUT allow_grade_interval TEXT
+    )
+AS $$
+BEGIN
+    SELECT
+        MAX(gj.date + aq.grade_rate_minutes * INTERVAL '1 min')
```
mwest1066, line 12:
```suggestion
        MAX(gj.date + make_interval(mins => aq.grade_rate_minutes))
```

mwest1066:
> @mwest1066 @nwalters512 it'd be great to have this working by the time we release the first assignment, mid-next week. Do you think this is possible?

Should be fine. I need to do the server-python update first. Don't hesitate to remind me about this.

When exactly do you need it deployed?

mwest1066:
> Preferably by Tuesday next week (Jan 12), though earlier is better, gives us time to test it properly.

Ok, I'll try and get a deploy later this week. Don't hesitate to remind me.

pages/partials/examQuestionScore.ejs
jonatanschroeder: The issue is that this partial is also used in the instance question page, though it's easy enough to add this update in the function that is called when the countdown ends.

database/tables/assessment_questions.pg
```diff
@@ -13,6 +13,7 @@ columns
     first_submission_score_hist: double precision[]
     first_submission_score_variance: double precision
     force_max_points: boolean default false
+    grade_rate_minutes: integer default 0
```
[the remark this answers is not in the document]
jonatanschroeder: No particular reason, should we allow half a minute? Autograders sometimes take longer than that to complete...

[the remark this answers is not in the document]
jonatanschroeder: I changed it to allow double precision. This is the last change, everything else is resolved.

sprocs/instance_questions_next_allowed_grade.sql
jonatanschroeder: I had to switch back, `make_interval` doesn't support double precision.

mwest1066:
And I forgot to ask about a news item. Want to add one into this PR?

jonatanschroeder:
> And I forgot to ask about a news item. Want to add one into this PR?

I can add one. Do you want screenshots or just the explanation?

mwest1066:
> I can add one. Do you want screenshots or just the explanation?

Screenshots! Everyone loves an image 😄

mwest1066:
@jonatanschroeder I edited the news item so that it's now in a more "action oriented" order:

1. What is this feature?
2. How do we configure it?
3. Why do we want it?

Please feel free to edit anything as always!

mwest1066 (approved):

mwest1066:
> Please feel free to edit anything as always!

And tell me when you are happy with the final state for merging.

jonatanschroeder:
> @jonatanschroeder I edited the news item so that it's now in a more "action oriented" order:
> 
> 1. What is this feature?
> 2. How do we configure it?
> 3. Why do we want it?
> 
> Please feel free to edit anything as always!

I added testing as an additional motivation. Otherwise looks good. And I don't mind if you change the news item, since you're listed as author anyway...

mwest1066:
> I added testing as an additional motivation. Otherwise looks good. And I don't mind if you change the news item, since you're listed as author anyway...

Haha, so true 😄

Thanks!

Issue #3551, opened 2020-12-16
Clearer number of attempts left in an exam question


jonatanschroeder:
Exam questions with a limited number of attempts should show the number of attempts left (either based on triesPerVariant or on the number of points entries left), ideally right beside the Save & Grade button.

PR #3601, opened 2020-12-23
Add tries left in Grade button for exams

jonatanschroeder:
Doesn't solve fully #3551 but is a workaround for questions with single variant.

### DIFF
--- a/pages/partials/newVariantInfo.ejs
+++ b/pages/partials/newVariantInfo.ejs
@@ -1,0 +1,13 @@
+<p>This assessment is set up to allow you to try multiple variants of
+the same question. Each of these variants is equivalent to the
+question you have been presented with, but may include changes in
+input values, parameters and other settings. Although
+<% if (variantAttemptsLeft > 1) { %>
+you have <%= variantAttemptsLeft %> attempts left
+<% } else if (variantAttemptsTotal > 1) { %>
+this is your last attempt
+<% } else { %>
+you have a single attempt
+<% } %>
+with the current variant, you are allowed to try unlimited other
+variants.</p>
--- a/pages/partials/newVariantInfo.ejs
+++ b/pages/partials/newVariantInfo.ejs
@@ -1,7 +1,7 @@
-<p>This assessment is set up to allow you to try multiple variants of
-the same question. Each of these variants is equivalent to the
+<p>This question allows you to try multiple variants.
+Each of these variants is equivalent to the
 question you have been presented with, but may include changes in
-input values, parameters and other settings. Although
+input values, parameters, and other settings. Although
 <% if (variantAttemptsLeft > 1) { %>
 you have <%= variantAttemptsLeft %> attempts left
 <% } else if (variantAttemptsTotal > 1) { %>
@@ -9,5 +9,5 @@ this is your last attempt
 <% } else { %>
 you have a single attempt
 <% } %>
-with the current variant, you are allowed to try unlimited other
+with the current variant, you are allowed to try an unlimited number of other
 variants.</p>
--- a/pages/partials/questionFooter.ejs
+++ b/pages/partials/questionFooter.ejs
@@ -78,6 +78,8 @@
                <% } else if (variantAttemptsLeft == 1) { %>
                <small class="font-italic ml-2">Single attempt</small>
                <% } %>
+            <% } else if (question_context != 'instructor') { %>
+               <small class="font-italic ml-2">Unlimited attempts</small>
             <% } %>
             </button>
               <% } %>
--- a/pages/partials/questionFooter.ejs
+++ b/pages/partials/questionFooter.ejs
@@ -16,7 +16,7 @@
             </button>
           <% } %>
           <% if (showGradeButton) { %>
-              <% if (question_context == 'student_homework') { %>
+              <% if (question_context == 'student_homework' || question_context == 'student_exam') { %>
                   <button class="btn btn-primary question-grade disable-on-submit order-1 mr-1" name="__action" value="grade">Save & Grade
                       <% if ((assessment_question.tries_per_variant - variant.num_tries) > 1) { %>
                           <span class="badge">
@@ -62,7 +62,7 @@
                 </button>
               <% } %>
               <% if (showGradeButton) { %>
-                  <% if (question_context == 'student_homework') { %>
+                  <% if (question_context == 'student_homework' || question_context == 'student_exam') { %>
                       <button class="btn btn-primary question-grade disable-on-submit order-1 mr-1">Save & Grade
                           <% if ((assessment_question.tries_per_variant - variant.num_tries) > 1) { %>
                               <span class="badge">
--- a/pages/partials/questionFooter.ejs
+++ b/pages/partials/questionFooter.ejs
@@ -39,6 +39,9 @@
           <% } else if (showTryAgainButton) { %><a href="<%= tryAgainUrl %>" class="btn btn-primary disable-on-click ml-1">Try a new variant</a>
           <% } else if (hasAttemptsOtherVariants) { %>
           <small class="font-italic align-self-center">Additional attempts available with new variants</small>
+          <button type="button" class="btn btn-xs" data-toggle="popover" data-html="true" data-content="<%= include('newVariantInfo') %>" data-placement="auto" data-trigger="focus" data-container="body" tabindex="0">
+              <i class="fa fa-question-circle" aria-hidden="true"></i>
+          </button>
           <% } %>
         </span>
       </div>
@@ -88,6 +91,9 @@
             <% } else if (showTryAgainButton) { %><a href="<%= tryAgainUrl %>" class="btn btn-primary disable-on-click ml-1">Try a new variant</a>
             <% } else if (hasAttemptsOtherVariants) { %>
             <small class="font-italic align-self-center">Additional attempts available with new variants</small>
+            <button type="button" class="btn btn-xs" data-toggle="popover" data-html="true" data-content="<%= include('newVariantInfo') %>" data-placement="auto" data-trigger="focus" data-container="body" tabindex="0">
+                <i class="fa fa-question-circle" aria-hidden="true"></i>
+            </button>
             <% } %>
           </span>
         </div>
@@ -100,5 +106,9 @@
   // Install on page load and reinstall on websocket re-render
   $('a.disable-on-click').on('click', function() {
     $(this).addClass('disabled');
+  });
+  // Enable popover
+  $(document).ready(function(){
+    $('[data-toggle="popover"]').popover({sanitize: false, container: 'body'});
   });
 </script>
--- a/pages/partials/questionFooter.ejs
+++ b/pages/partials/questionFooter.ejs
@@ -16,18 +16,27 @@
             </button>
           <% } %>
           <% if (showGradeButton) { %>
-              <% if (question_context == 'student_homework' || question_context == 'student_exam') { %>
-                  <button class="btn btn-primary question-grade disable-on-submit order-1 mr-1" name="__action" value="grade">Save & Grade
-                      <% if ((assessment_question.tries_per_variant - variant.num_tries) > 1) { %>
-                          <span class="badge">
-                              <%= String(assessment_question.tries_per_variant - variant.num_tries) %>
-                              tries left
-                          </span>
-                      <% } %>
-                  </button>
-              <% } else { %>
-                  <button class="btn btn-primary question-grade disable-on-submit order-1 mr-1" name="__action" value="grade">Save & Grade</button>
-              <% } %>
+            <button class="btn btn-primary question-grade disable-on-submit order-1 mr-1" name="__action" value="grade">Save & Grade
+            <%  var attemptsLeft = 0, attemptsTotal = 0;
+                if (question_context == 'student_homework') {
+                    attemptsLeft = assessment_question.tries_per_variant - variant.num_tries;
+                    attemptsTotal = assessment_question.tries_per_variant;
+                } else if (question_context == 'student_exam') {
+                    attemptsLeft = instance_question.points_list.length;
+                    attemptsTotal = instance_question.points_list_original.length;
+                }
+                if (attemptsTotal > 0) { %>
+                <span class="badge">
+                    <% if (attemptsLeft > 1) { %>
+                    <%= attemptsLeft %> attempts left
+                    <% } else if (attemptsLeft == 1 && attemptsTotal > 1) { %>
+                    Last attempt
+                    <% } else if (attemptsLeft == 1 && attemptsTotal == 1) { %>
+                    Single attempt
+                    <% } %>
+                </span>
+                <% } %>
+            </button>
           <% } %>
           <% if (showManualGradingMsg) { %>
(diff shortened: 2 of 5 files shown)


mwest1066:
If I'm reading the sync code correctly, I think exam questions (which don't have `triesPerVariant` in JSON) will have `assessment_questions.tries_per_variant = 1`. Won't this mean that on exams we'll always display "1 tries left"?

jonatanschroeder:
> If I'm reading the sync code correctly, I think exam questions (which don't have `triesPerVariant` in JSON) will have `assessment_questions.tries_per_variant = 1`. Won't this mean that on exams we'll always display "1 tries left"?

Exam questions can have `triesPerVariant`, check https://ca.prairielearn.org/pl/course_instance/1678/instructor/assessment/15947/file_view. Also, the tries left only shows if it's higher than 1.

mwest1066:
> Exam questions can have `triesPerVariant`, check https://ca.prairielearn.org/pl/course_instance/1678/instructor/assessment/15947/file_view.

Ah, ok, yes, but we need to set it manually in the JSON file.

I think the right thing to do for exams is to ignore `tries_per_variant` and instead use the length of `instance_questions.points_list`. This is the variable that we are actually using to compute whether there are tries remaining.

Something like:
```
                  <button class="btn btn-primary question-grade disable-on-submit order-1 mr-1" name="__action" value="grade">Save & Grade
                      <%
                          var triesLeft = 0;
                          if (question_context == 'student_homework') {
                              triesLeft = assessment_question.tries_per_variant - variant.num_tries;
                          } else if (question_context == 'student_exam') {
                              triesLeft = instance_question.points_list.length;
                          }
                      %>
                      <% if (triesLeft > 1) { %>
                          <span class="badge">
                              <%= String(triesLeft) %>
                              tries left
                          </span>
                      <% } %>
                  </button>
```

pages/partials/questionFooter.ejs
```diff
@@ -16,18 +16,27 @@
...
-                              <%= String(assessment_question.tries_per_variant - variant.num_tries) %>
-                              tries left
-                          </span>
-                      <% } %>
-                  </button>
-              <% } else { %>
-                  <button class="btn btn-primary question-grade disable-on-submit order-1 mr-1" name="__action" value="grade">Save & Grade</button>
-              <% } %>
+            <button class="btn btn-primary question-grade disable-on-submit order-1 mr-1" name="__action" value="grade">Save & Grade
+            <%  var attemptsLeft = 0, attemptsTotal = 0;
```
mwest1066, line 20: Server-side code should use `let` (this was my error in the code I posted to slack code, sorry!)

```suggestion
            <%  let attemptsLeft = 0, attemptsTotal = 0;
```

pages/partials/questionFooter.ejs
```diff
@@ -62,18 +71,28 @@
...
-                                  tries left
-                              </span>
-                          <% } %>
-                      </button>
-                  <% } else { %>
-                      <button class="btn btn-primary question-grade disable-on-submit order-1 mr-1">Save & Grade</button>
-                  <% } %>
+                <button class="btn btn-primary question-grade disable-on-submit order-1 mr-1">Save & Grade
+                <%
+                var attemptsLeft = 0, attemptsTotal = 0;
```
mwest1066, line 76:
```suggestion
                let attemptsLeft = 0, attemptsTotal = 0;
```

mwest1066 (approved):
Looks good to me!

jonatanschroeder:
And now it actually resolves #3551 .

mwest1066:
> And now it actually resolves #3551 .

Yay! 🎉

pages/partials/questionFooter.ejs
```diff
@@ -16,18 +16,27 @@
...
+                    attemptsLeft = assessment_question.tries_per_variant - variant.num_tries;
+                    attemptsTotal = assessment_question.tries_per_variant;
+                } else if (question_context == 'student_exam') {
+                    attemptsLeft = instance_question.points_list.length;
+                    attemptsTotal = instance_question.points_list_original.length;
+                }
+                if (attemptsTotal > 0) { %>
+                <span class="badge">
+                    <% if (attemptsLeft > 1) { %>
+                    <%= attemptsLeft %> attempts left
```
mwest1066, line 31: I think it might look better with parentheses around the "attempts" text?

```suggestion
                    (<%= attemptsLeft %> attempts left)
```

pages/partials/questionFooter.ejs
```diff
@@ -16,18 +16,27 @@
...
-                              tries left
-                          </span>
-                      <% } %>
-                  </button>
-              <% } else { %>
-                  <button class="btn btn-primary question-grade disable-on-submit order-1 mr-1" name="__action" value="grade">Save & Grade</button>
-              <% } %>
+            <button class="btn btn-primary question-grade disable-on-submit order-1 mr-1" name="__action" value="grade">Save & Grade
+            <%  let attemptsLeft = 0, attemptsTotal = 0;
+                if (question_context == 'student_homework') {
```
mwest1066, line 21: I think the exam case works really well, but I actually think it's a bit confusing for homeworks. HW questions allow unlimited attempts, just at different variants. I think students will be confused when we say "single attempt" but then they can try the "same" question again (but with a different variant).

Maybe for homeworks we can do:
* if tries_per_variant > 1, then in the Grade&Submit button say "(2 attempts left at this version)" or "(Last attempt at this version)"
* for any homework question, put some extra text on the right side of the footer (not inside the Save&Grade button) that says something like "Unlimited question versions can be attempted"

That is, we won't display "Single attempt" for homework questions.

jonatanschroeder: May be a better idea to move the whole text regardless to outside of the button, using a badge instead? Having long text inside the button seems a bit crowded.

jonatanschroeder: I moved the badge outside the button. An alternative would be to move the badge back into the button and use badge-light to highlight it.

mwest1066: I guess I think the clearest thing is to put the badge inside the button. Otherwise the student needs to make the association between the "Save&Grade" button and the text "XX grade attempts left", and they have to realize the badge doesn't apply to "Save", which all feels a bit complicated. Putting the badge inside the button makes it very obvious what's going on. I agree that the large button is a bit ugly, but I prefer clarity over beauty 😄

In terms of styling, I played around with some options below. I think that in isolation (like in this comment) option (2) looks best. But on the actual page the light badge inside the button looks a bit cluttered to me, and I think I actually prefer option (5) with parentheses, or possibly option (6). What do you think?

(1) `<span class="badge badge-light">`
image

(2) `<span class="badge badge-light ml-2">`
image

(3) `<span class="badge">`
image

(4) `<small class="font-weight-bold ml-1">`
image

(5) `<small class="ml-1">`
image

(6) `<small class="font-italic ml-2">`
image

jonatanschroeder: Option 6 actually looks pretty good.

mwest1066: And I still think that we should add something like `<span class="d-flex align-self-center"> Unlimited question versions can be attempted</span>` to the end of the footer for homework questions. This will help us with #3347, where we want to add the ability to mix and match question grading schemes. It would be great if the question footer could clearly indicate both:
1. How many attempts are left at this variant.
2. Are new variants going to be allowed?

mwest1066: > Option 6 actually looks pretty good.

Yeah, I'm definitely not opposed to that.

mwest1066: More playing with styling: For homework questions with unlimited variants I like:

`<span class="d-flex align-self-center"><small class="font-italic">Unlimited question versions can be attempted<small></small></small></span>`
image

Thinking about getting ready for #3347 I propose:
1. The button text always shows how many attempts are remaining for the current variant (e.g., "6 attempts left", "Last attempt", "Single attempt", "unlimited attempts", "blocked for 5 min").
2. We add extra text like above to the right-hand side of the footer to explain whether new variants will be allowed. If new variants are not allowed (like on exams or for single-variant questions) then we don't say anything. If new variants are allowed then we say something.

What do you think about this?

I'm not sure whether we should use the work "variant" or "version" (or something else?) to refer to the thing that we internally call a `variant`. I generally prefer to keep internal and external terminology the same whenever possible, but sometimes it's clearer to use different words. What do you think?

In any case, maybe we should change the button that says "Try question again" to "Try another question variant" or something like that?

jonatanschroeder: > And I still think that we should add something like <span class="d-flex align-self-center"> Unlimited question versions can be attempted</span> to the end of the footer for homework questions. 

I'm okay with that, just not for single variant questions.

jonatanschroeder: > `<span class="d-flex align-self-center"><small class="font-italic">Unlimited question versions can be attempted<small></small></small></span>`

Wording sounds weird in juxtaposition with the button text. Maybe "Additional attempts available with different variants"? Just to make it clear the difference between attempts in the button and attempts in other variants.

mwest1066: > I'm okay with that, just not for single variant questions.

Yep, absolutely, we already have the distinction between "new variants allowed" and "no new variants" (singleVariant) for homeworks. It would be great if the question footer made this clear.

mwest1066: > Wording sounds weird in juxtaposition with the button text. Maybe "Additional attempts available with different variants"? Just to make it clear the difference between attempts in the button and attempts in other variants.

I like it! Maybe "... with new variants", and we could change "Try question again" to "Try a new question variant" to link together the ideas? Or use "different" rather than "new" in both cases.

We could also add a question-mark icon next to "Additional attempts available..." with a popover that more clearly explains what variants are and the idea of trying new variants. This is a somewhat unusual feature of PL, so it wouldn't hurt to explain it more for students. Although it seems like they mainly get the idea very quickly when they just use the system.

jonatanschroeder:
@mwest1066 I believe that, assuming all tests pass, this should now be good to go, though I'd appreciate some feedback on the text for the info button.

pages/partials/questionFooter.ejs
```diff
@@ -71,27 +67,17 @@
...
-                <% } %>
+                <button class="btn btn-primary question-grade disable-on-submit order-1 mr-1">Save & Grade
+            <% if (variantAttemptsTotal > 0) { %>
+               <% if (variantAttemptsLeft > 1) { %>
+               <small class="font-italic ml-2"><%= variantAttemptsLeft %> attempts left</small>
+               <% } else if (variantAttemptsLeft == 1 && variantAttemptsTotal > 1) { %>
+               <small class="font-italic ml-2">Last attempt</small>
+               <% } else if (variantAttemptsLeft == 1) { %>
+               <small class="font-italic ml-2">Single attempt</small>
+               <% } %>
```
mwest1066, line 78: If this is a `singleVariant` question, should we add text here to say `Unlimited attempts`? Otherwise I think it's not entirely clear what will happen for `singleVariant` questions, because we don't display any guidance at all.

pages/partials/newVariantInfo.ejs
```diff
@@ -0,0 +1,13 @@
+<p>This assessment is set up to allow you to try multiple variants of
+the same question. Each of these variants is equivalent to the
```
mwest1066, line 2: I think it's best to make this per-question, because even on a homework some other questions might be `singleVariant`.
```suggestion
<p>This question allows you to try multiple variants.
Each of these variants is equivalent to the
```

pages/partials/newVariantInfo.ejs
```diff
@@ -0,0 +1,13 @@
...
+question you have been presented with, but may include changes in
+input values, parameters and other settings. Although
```
mwest1066, line 4: I have a slight preference for the Oxford comma, but I don't insist on it.
```suggestion
input values, parameters, and other settings. Although
```

pages/partials/newVariantInfo.ejs
```diff
@@ -0,0 +1,13 @@
...
+<% if (variantAttemptsLeft > 1) { %>
+you have <%= variantAttemptsLeft %> attempts left
+<% } else if (variantAttemptsTotal > 1) { %>
+this is your last attempt
+<% } else { %>
+you have a single attempt
+<% } %>
+with the current variant, you are allowed to try unlimited other
```
mwest1066, line 12:
```suggestion
with the current variant, you are allowed to try an unlimited number of other
```

pages/partials/questionFooter.ejs
```diff
@@ -45,8 +35,14 @@
         <span class="d-flex">
           <input type="hidden" name="__variant_id" value="<%= variant.id %>">
           <input type="hidden" name="__csrf_token" value="<%= __csrf_token %>">
-          <% if (showNewVariantButton) { %><a href="<%= newVariantUrl %>" class="btn btn-primary disable-on-click ml-1">New variant</a><% } %>
-          <% if (showTryAgainButton) { %><a href="<%= tryAgainUrl %>" class="btn btn-primary disable-on-click ml-1">Try question again</a><% } %>
+          <% if (showNewVariantButton) { %><a href="<%= newVariantUrl %>" class="btn btn-primary disable-on-click ml-1">New variant</a>
+          <% } else if (showTryAgainButton) { %><a href="<%= tryAgainUrl %>" class="btn btn-primary disable-on-click ml-1">Try a new variant</a>
+          <% } else if (hasAttemptsOtherVariants) { %>
+          <small class="font-italic align-self-center">Additional attempts available with new variants</small>
+          <button type="button" class="btn btn-xs" data-toggle="popover" data-html="true" data-content="<%= include('newVariantInfo') %>" data-placement="auto" data-trigger="focus" data-container="body" tabindex="0">
```
mwest1066, line 42: When `data-trigger="focus"` it's necessary to use `<a>` rather than `<button>`: https://getbootstrap.com/docs/4.5/components/popovers/#dismiss-on-next-click

If we don't do this then the popovers break on Safari (see #3610).

pages/partials/questionFooter.ejs
mwest1066: I think our principle should be:
1. _always_ have help text inside the button saying how many attempts are allowed (at the current variant)
2. If the question allows multiple variants, have the extra text inside the question footer explaining this.

I think the only case we are missing at the moment is `singleVariant` questions on Homeworks?

jonatanschroeder: What about instructor view?

pages/partials/questionFooter.ejs
```diff
@@ -25,6 +25,8 @@
                <% } else if (variantAttemptsLeft == 1) { %>
                <small class="font-italic ml-2">Single attempt</small>
                <% } %>
+            <% } else if (question_context != 'instructor') { %>
+               <small class="font-italic ml-2">Unlimited attempts</small>
```
mwest1066, line 29: I think we also need this change in the non-Freeform question section down below (yuck! I'm sorry! 😢).

jonatanschroeder: Done, though I don't have these kinds of questions to test them. The freeform ones are not broken, though, so that's good :-D

PR #3603, opened 2020-12-23
Make startup messages less noisy

mwest1066:


workspace_host/interface.js
```diff
@@ -212,7 +212,7 @@ async.series([
     (callback) => {
         server = http.createServer(app);
         server.listen(workspace_server_settings.port);
-        logger.info(`Listening on port ${workspace_server_settings.port}`);
+        logger.verbose(`Listening on port ${workspace_server_settings.port}`);
```
jonatanschroeder, line 215:
```suggestion
        logger.verbose(`Workspace server listening on port ${workspace_server_settings.port}`);
```

jonatanschroeder (approved):
I don't necessarily have an issue with it (though I agree it's less confusing without the extra port numbers for s3rver and workspaces), but I'm fine with the changes.

nicknytko (approved):
Looks good — I see the workspace host is littered with `logger.info` calls but we can probably  address those in another PR.

mwest1066:
> Looks good — I see the workspace host is littered with `logger.info` calls but we can probably address those in another PR.

Yep, good point, that will be very noisy for anyone using workspaces locally. We should definitely clean that up at some point.

PR #3608, opened 2020-12-24
Install latest socket.io dependencies

nwalters512:
This PR is a coordinated bump of all socket.io-related dependencies. I tested the syncing and question pages and confirmed that websocket functionality looks good. I was unable to test workspaces locally - would appreciate some help with that.

**NOTE**: the new server is incompatible with the old client. It would be best to deploy this over break when we're unlikely to interrupt any live connections. If we do interrupt any connections, a page refresh should fix things up.

closes #3512 
closes #3533 
closes #3539


mwest1066:
This breaks workspaces for me. I'm testing with:
```
docker run -it --rm -p 3000:3000 -v "$HOME/pl_ag_jobs:/jobs" -e HOST_JOBS_DIR="$HOME/pl_ag_jobs"v /var/run/docker.sock:/var/run/docker.sock -v ${PWD}:/PrairieLearn prairielearn/prairielearn
```

When using `master`, workspaces work fine.

But when using this branch, workspaces don't work. I just get a status bar with a spinner, but no feedback (normally it shows messages like "pulling container"):

image

mwest1066:
However, external grading does work for me with this PR, so websockets are working in general.

nwalters512:
The above issue was resolved in #e373b69 - per the list of breaking changes for Socket.io v3, [`join()` is now synchronous](https://socket.io/docs/v3/migrating-from-2-x-to-3-0/#Socket-join-and-Socket-leave-are-now-synchronous), which means it no longer supports a callback. Only workspaces relied on the callback of that API, which is why only workspaces broke!

mwest1066:
When I test locally it looks like there is still an issue with workspaces. When I launch a new workspace, the initial websocket messages are displayed in the top status area, but then it gets stuck on `LAUNCHING | Sending launch command to host`. If I wait until the workspace is actually running in the background and reload the page, then the workspace is loaded correctly.

In particular, I don't get any of the websocket status messages showing the progress of the docker pull (it should count from 0% to 100% as it pulls the image).

One possible explanation might be that the redis adaptor isn't working. The workspace status messages come partially from the main PL server process, and mainly from the workspace host process (`workspace_host/interface.js`). The client makes a websocket connection to the PL server, and the PL server is connected to the workspace host via the socker.io-redis adaptor. Messages such as the pull progress percentage are sent from the workspace host, and should be relayed via the PL server to the client. It looks like this might not be happening.

nwalters512:
Ah, I actually noticed that too, figured it was just an issue with running locally 😂 let me poke at it some more.

mwest1066:
> Ah, I actually noticed that too, figured it was just an issue with running locally 😂 let me poke at it some more.

Ha! No, unfortunately everything should work correctly locally 😄 --- we have redis running inside the container for exactly this reason.

mwest1066 (approved):

PR #3614, opened 2020-12-27
Add support for leading and trailing code in Python autograder

### LINKED ISSUE #3613: Leading/trailing code are not documented for Python grader
The Python grader docs don't mention `leading_code.py` and `trailing_code.py`.

nicknytko:
Adds support for `tests/trailing_code.py` for python autograder and a small blurb to the documentation.

Adds example question `demo/autograder/python/leadingTrailing`:

image

Closes #3613

### DIFF
--- a/graders/python/python_autograder/pl_execute.py
+++ b/graders/python/python_autograder/pl_execute.py
@@ -150,7 +150,7 @@ def execute_code(fname_ref, fname_student, include_plt=False,
     except Exception:
         err = sys.exc_info()
 
-    # Now that user code has been run, replace deleted files incase we are to run the tests again.
+    # Now that user code has been run, replace deleted files in case we are to run the tests again.
     with open(join(filenames_dir, 'data.json'), 'w', encoding='utf-8') as f:
         json.dump(data, f)
     with open(fname_ref, 'w', encoding='utf-8') as f:
--- a/graders/python/python_autograder/pl_execute.py
+++ b/graders/python/python_autograder/pl_execute.py
@@ -81,6 +81,8 @@ def execute_code(fname_ref, fname_student, include_plt=False,
     os.remove(join(filenames_dir, 'data.json'))
     os.remove(fname_ref)
     os.remove(join(filenames_dir, 'setup_code.py'))
+    os.remove(join(filenames_dir, 'leading_code.py'))
+    os.remove(join(filenames_dir, 'trailing_code.py'))
     os.remove(join(filenames_dir, 'test.py'))
 
     repeated_setup_name = 'repeated_setup()'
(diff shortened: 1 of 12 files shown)


mfsilva22:
This is really great! I tested the example above, and one of my course questions that use Jupyter notebooks and the external grader (with leading code). 

We should merge this soon, and use this feature for the `pl-order-blocks`.

This example makes me want even more a `pl-text` that can be used for survey questions, coding, future manual graded questions (so we can avoid the encoding part needed for this example).

mfsilva22 (commented):
This is really great! I tested the example above, and one of my course questions that use Jupyter notebooks and the external grader (with leading code).

We should merge this soon, and use this feature for the pl-order-blocks.

This example makes me want even more a pl-text that can be used for survey questions, coding, future manual graded questions (so we can avoid the encoding part needed for this example).

graders/python/python_autograder/pl_execute.py
```diff
@@ -51,20 +60,24 @@ def execute_code(fname_ref, fname_student, include_plt=False,
...
-            str_student = str_leading + pl_helpers.extract_ipynb_contents(f, ipynb_key)
+            str_student = pl_helpers.extract_ipynb_contents(f, ipynb_key)
         else:
             str_student = f.read()
+    str_student = str_leading + str_student + str_trailing
+
     with open(join(filenames_dir, 'test.py'), encoding='utf-8') as f:
         str_test = f.read()
 
+    # Delete sensitive code so students can't read e.g. test cases or setup code
```
mwest1066, line 80: Should we also delete leading/trailing code? I know that we will often want to show then to students, by maybe we want to hide them sometimes? In any case, I don't think that deleting them can do any harm?

mwest1066:
> This example makes me want even more a pl-text that can be used for survey questions, coding, future manual graded questions (so we can avoid the encoding part needed for this example).

Yes! We need a `pl-text` so bad! (or a `paragraph=true` attribute for `pl-string`)

mwest1066:
Do we currently have workspace example questions where we could use this? I'm thinking about @mfsilva22's comments about using `input-code` for workspace questions. I'm not sure whether that applied to XC 101 questions, or just questions in a course.

graders/python/python_autograder/pl_execute.py
mfsilva22: I agree that deleting makes sense. The code can always be shown like in the example, using pl-code

mfsilva22:
I don't think we need to modify our examples at this time. Questions can still use code using the input-code file, or leading/trailing. They are two different features, that should co-exist nicely and give flexibility to question authors.

graders/python/python_autograder/pl_execute.py
```diff
@@ -147,12 +149,20 @@ def execute_code(fname_ref, fname_student, include_plt=False,
         err = None
     except Exception:
         err = sys.exc_info()
+
+    # Now that user code has been run, replace deleted files incase we are to run the tests again.
```
mwest1066, line 153:
```suggestion
    # Now that user code has been run, replace deleted files in case we are to run the tests again.
```

mwest1066 (approved):

PR #3620, opened 2020-12-29
Bugfix: score is shown if showClosedAssessmentScore is false but password is set

jonatanschroeder:


middlewares/studentAssessmentAccess.js
```diff
@@ -35,11 +35,6 @@ router.all('/', function(req, res, next) {
         'password' in res.locals.authz_result &&
         res.locals.authz_result.password) {
 
-        // If the assessment is complete, use this middleware to show the logout page
-        if ('assessment_instance' in res.locals && res.locals.assessment_instance.open == false) {
-            return badSEB(req, res);
-        }
-
```
mwest1066, line 42: Why do we want to delete this block of code? Won't that potentially interfere with SEB logouts? (I know that SEB is not currently in use, but I'd rather not break it unnecessarily.)

Instead, could we move this block: https://github.com/PrairieLearn/PrairieLearn/blob/e6b5eabfb84c79ca3d0e4eb0b7628b8fd40accbb/middlewares/studentAssessmentAccess.js#L59 above this block: https://github.com/PrairieLearn/PrairieLearn/blob/e6b5eabfb84c79ca3d0e4eb0b7628b8fd40accbb/middlewares/studentAssessmentAccess.js#L33

That is, we would first check for a closed assessment with `show_closed_assessment=false`, and only if we get past that will we worry about passwords.

I might have completely misunderstood the logical flow here, so please feel free to disagree with my suggestion!

jonatanschroeder: The test that I deleted is already done in lines 22-25, which means that these lines would only run if the mode is not SEB. Getting a SEB error in a non-SEB context is, in my view, incorrect.

jonatanschroeder: But I wouldn't be opposed to moving the blocks anyway...

mwest1066: Ha! I was just coming to the same conclusion. I agree that the block you are deleting seems like it's obviously in the wrong place and should be deleted. But I think we should also move the line-59 block (the test for `authz_result.show_closed_assessment`) to the very top of the `router.all()` function.

jonatanschroeder: Ok, it seems the difference, after moving, is what happens if a closed assessment with `showClosedAssessment` set to the default `true`. I'll do some testing on these cases and commit the move.

jonatanschroeder: Yeah, that is there to ensure that a password isn't required to show a closed assessment.

mwest1066: Ah, ok, makes sense. I'll leave it to you to fix up the logic somehow 😄

jonatanschroeder: Ok, fixed it.

mwest1066 (approved):

PR #3643, opened 2021-01-01
install dvipng library

mfsilva22:
Adding library `dvipng` (this will allow using latex fonts when creating dynamic plots using pl-figure

### DIFF
--- a/environments/plbase/plbase-install.sh
+++ b/environments/plbase/plbase-install.sh
@@ -21,6 +21,7 @@ yum -y install \
     ImageMagick           `# for PrairieDraw label images` \
     texlive               `# for PrairieDraw label images` \
     git \
+    texlive-dvipng \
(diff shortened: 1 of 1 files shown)


environments/plbase/plbase-install.sh
```diff
@@ -21,6 +21,7 @@ yum -y install \
     ImageMagick           `# for PrairieDraw label images` \
     texlive               `# for PrairieDraw label images` \
     git \
+    texlive-dvipng \
```
mwest1066, line 24: Those comments are actually no longer valid, so let's get rid of them. And it's a bit tidier to keep the texlive entries together.
```suggestion
    ImageMagick \
    texlive \
    texlive-dvipng \
    git \
```

mwest1066 (approved):

PR #3648, opened 2021-01-02
Fix leading, trailing code in Python grader

nicknytko:
Fixes the grader to check if leading and trailing code files exist before we delete them


graders/python/python_autograder/pl_execute.py
```diff
@@ -81,8 +80,10 @@ def execute_code(fname_ref, fname_student, include_plt=False,
     os.remove(join(filenames_dir, 'data.json'))
     os.remove(fname_ref)
     os.remove(join(filenames_dir, 'setup_code.py'))
-    os.remove(join(filenames_dir, 'leading_code.py'))
-    os.remove(join(filenames_dir, 'trailing_code.py'))
+    if exists(join(filenames_dir, 'leading_code.py')):
+        os.remove(join(filenames_dir, 'leading_code.py'))
+    if exists(join(filenames_dir, 'trailing_code.py')):
+        os.remove(join(filenames_dir, 'trailing_code.py'))
```
mwest1066, line 86: Nit: It's undoubtedly fine in this case because we are the only code running, but in general I worry about [TOCTAU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) bugs with this pattern. In general I prefer `try: XXX catch FileNotFoundError: pass`

mwest1066 (approved):
@nicknytko you are very kind to humor my preferences! ❤️

Issue #3625, opened 2020-12-30
Enable caching of client-side resources


nwalters512:
Currently, on each page request, all client-side scripts and styles (Bootstrap CSS + JS, MathJax, lodash, etc.) is not cached on the client and must be loaded in their entirety on each page reload. This hurts the user experience for folks who are located a long way from our origin servers (e.g., anywhere outside the US) or who have slow network connections, or both.

This gets worse as we move towards running each question panel in a separate iframe. Without caching, we must load each resource once per question panel. By enabling caching, we make it possible to reasonably render a page with many submissions.

We should update PrairieLearn to allow many client-side resources to be cached.

## Resource types

There are several categories of resources we could conceivably handle, each of which must be handled differently:

* resources served directly out of `node_modules`
* resources served from PrairieLearn source (think local scripts and stylesheets)
* element resources, both PL-level and course-level
* files from `clientFilesCourse`, `clientFilesQuestion`, etc.

### `node_modules` resources

This includes things like `socket.io-client`, `jquery`, `fontawesome`, etc. - think things that individual pages would declare in their page `<head>`.

For these, we should append a `?cache=[key]` query param to all resource URLs, where `[key]` is some stable identifier for a particular version of the dependency, likely the dependency version (or the hash of the version). If that query param is present on a request, we can respond with `Cache-Control: public, max-age=31536000` to cache the resource essentially indefinitely. If the query param is not included, we'll continue to behave as we do today (don't send any `Cache-Control` header). When we install a new version of a dependency, the version will be different, and thus the `cache` query param will be different as well, so clients will request the new version and cache it instead.

Note that the value of the `cache` query param doesn't actually mean anything - if a client requests `cache=1.2.3` but version `2.3.4` is present on the server, version `2.3.4` is what will be served.

### Resources served by PrairieLearn

Caching resources served from the PL source code is slightly tricker, as we don't have a formal version to use as a cache buster. There are a few things we could use instead:

* The git commit hash of the version of PrairieLearn that is running.
* The hash of the file contents. This might be preferable to the git commit hash in case we start deploying PL without a `.git` directory present, such as in a Docker image.

As above, the cache buster would be added with `?cache=...`.

### Element resources

Elements *declare* the resources they depend on, which makes it easy to generate appropriate URLs for them on the fly. We can use a combination of the "`node_modules`" and "Resources served by PrairieLearn" resources outlined above to handle these assets.

[...]

PR #3649, opened 2021-01-03
Enable caching of files served from public/ and node_modules/

nwalters512:
This PR is a partial implementation of the caching strategies outlined by #3625 - it enables caching for files served from `public/` and `node_modules/`. It achieves this by introducing two new routes:

- `/assets/:cachebuster/*`, which will serve `*` from `/public`. In dev mode, it sets `max-age` of the `Cache-Control` header to `0` to ensure that assets aren't cached, as they're likely to change on disk while the server is running. In prod, `max-age` is set to `31557600` (one year).
- `/cacheable_node_modules/:cachebuster/*`, which will serve `*` from `/node_modules`. `max-age` is set to `31557600` (one year) in both dev and prod environments.

I've left the legacy `/node_modules` and `/localscripts`/`/javascripts`/`/stylesheets/`/etc. routes in place for now to handle legacy code that might expect to access files from those routes. Ideally we can look at traffic on the prod server to identify which files are still being accessed from those routes and figure out how to shift them to the new, cacheable routes.

Before deploying this, we should probably configure Apache to handle these routes directly and serve them without going through our application server.

On the server, URLs for resources are computed with the help of two functions:
- `assetPath` takes a path to an asset in `/public` (like `stylesheets/colors.css`) and returns a path like `/assets/HASH/stylesheets/colors.css`. `HASH` is computed as the hash of the `/public` directory as determined by [`folder-hash`](https://www.npmjs.com/package/folder-hash). This hash is determined at server startup.
- `nodeModulesAssetPath` takes a path to an asset in `/public` (like `@fortawesome/fontawesome/js/all.min.js`) and returns a path like `/cacheable_node_modules/HASH/@fortawesome/fontawesome/js/all.min.js`. `HASH` is computed as the SHA-256 hash of the given module's version. These paths are used in dev mode too, since if a package changes, its version should change too, and it would be served with a different hash.

I've left the serving of all element files, course files, and dynamic question files for a later PR.


pages/instructorCourseAdminGrading/instructorCourseAdminGrading.ejs
```diff
@@ -2,9 +2,6 @@
 <html lang="en">
     <head>
         <%- include('../partials/head', {pageTitle: 'External Grading'}); %>
-        <script src="/javascripts/lodash.min.js"></script>
-        <script src="/javascripts/d3.min.js"></script>
-        <script src="/localscripts/histmini.js"></script>
```
mwest1066, line 7: Wow, so many pages including so much unnecessary JS!

pages/partials/mathjax.ejs
```diff
@@ -41,4 +41,4 @@
      }
  };
 </script>
-<script type="text/javascript" id="MathJax-script" src="/MathJax/startup.js"></script>
+<script type="text/javascript" id="MathJax-script" src="<%= node_modules_asset_path('mathjax/es5/startup.js') %>"></script>
```
mwest1066, line 44: Yay! Direct loading of mathjax without a special route!

server.js
```diff
@@ -210,10 +211,41 @@ module.exports.initExpress = function() {
         logger.info(`localRootFilesDir: Mapping ${config.localRootFilesDir} into /`);
         app.use(express.static(config.localRootFilesDir));
     }
+
+    // To allow for more aggressive caching of static files served from public/,
+    // we introduce a new `assets/` path that includes a cachebuster in the path.
```
mwest1066, line 216: This comment won't always be so "new" 😄
```suggestion
    // we use the `assets/` path that includes a cachebuster in the path.
```

server.js
```diff
@@ -210,10 +211,41 @@ module.exports.initExpress = function() {
...
+    // In requests for resources, the cachebuster will be a hash of the contents
+    // of `/public`, which we will compute at startup. See `lib/assets.js` for
+    // implementation details.
+    app.use((req, res, next) => {
+        res.locals.asset_path = assets.assetPath;
+        next();
+    });
+    app.use('/assets/:cachebuster', express.static(path.join(__dirname, 'public'), {
+        maxAge: '31557600',
```
mwest1066, line 225: Should we also set `immutable=true` and `etags=false`? See http://expressjs.com/en/resources/middleware/serve-static.html

server.js
```diff
@@ -1115,6 +1147,12 @@ if (config.startServer) {
                 callback(null);
             });
         },
+        (callback) => {
+            util.callbackify(assets.init)(err => {
+                if (ERR(err, callback)) return;
+                callback(null);
+            });
+        },
```
mwest1066, line 1155: This works!
```suggestion
        async () => {
            await assets.init();
        },
```

pages/instructorCourseAdminGrading/instructorCourseAdminGrading.ejs
nwalters512: Right! Blind copy-pasting coming back to bite us 😛

server.js
nwalters512: 🤯 that's awesome! Was just pattern-matching off some other stuff in this section, should probably go back and "modernize" those too. I bet this could be even simpler:
```suggestion
        () => assets.init(),
```
since it's all just Promises under the hood.

server.js
nwalters512: I think we should definitely set `immutable=true`. For browsers that don't understand the `immutable` header, it miiiight still be a useful optimization to leave Etags enabled? At the very least, I don't think it'd hurt. For instance, at NerdWallet, we use `max-age=1year` but still leave Etags enabled (and don't even set `immutable`!).

server.js
nwalters512: Update: you definitely cannot do that! The `async` library internally has a check like `fn.constructor.name === 'AsyncFunction';` - it doesn't just check if a return value is a Thenable. I'll update to use your suggestion.

server.js
mwest1066: I agree that there doesn't seem to be much downside to leaving `etags` enabled.

server.js
mwest1066: Interesting!

Of course I'd also be happy to converted the entire sequence to a/a and get rid of `async.series()` completely.

mwest1066 (approved):

PR #3650, opened 2021-01-03
Added download button to file upload element

zillesc:
The existing file upload element didn't allow download.  (Download buttons appeared only in the corresponding pl_file_preview element, which requires the user to first save or save & grade.  This is non-intuitive.) 

I basically copied the code from: https://github.com/PrairieLearn/PrairieLearn/pull/3217/files

Currently the download button is too close to the 'Show Preview' button.  I don't know the preferred way to add some space between them. 

(Also, I apologize for not making this branch in my own repo... I didn't see the contributing guidelines until the patch was complete.)

### DIFF
--- a/elements/pl-file-upload/pl-file-upload.js
+++ b/elements/pl-file-upload/pl-file-upload.js
@@ -168,6 +168,8 @@ window.PLFileUpload.prototype.renderFileList = function() {
             $fileStatusContainerLeft.append('<p class="file-status">uploaded</p>');
         }
         if (fileData) {
+            var download = '<a download="' + fileName + '" class="btn btn-outline-secondary btn-sm" id="file-download-' + uuid + '-' + index + '" onclick="event.stopPropagation();" href="data:application/octet-stream;base64,' + fileData + '">Download</a>'
+            
             var $preview = $('<div class="file-preview collapse" id="file-preview-' + uuid + '-' + index + '"><pre class="bg-dark text-white rounded p-3 mb-0"><code></code></pre></div>');
             if (isExpanded) {
                 $preview.addClass('in');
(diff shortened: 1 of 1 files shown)


elements/pl-file-upload/pl-file-upload.js
```diff
@@ -168,6 +168,8 @@ window.PLFileUpload.prototype.renderFileList = function() {
             $fileStatusContainerLeft.append('<p class="file-status">uploaded</p>');
         }
         if (fileData) {
+            var download = '<a download="' + fileName + '" class="btn btn-outline-secondary btn-sm" id="file-download-' + uuid + '-' + index + '" onclick="event.stopPropagation();" href="data:application/octet-stream;base64,' + fileData + '">Download</a>';
```
nwalters512, line 171: Is the `id` attribute necessary?

Separately, this should be all you need for proper spacing:

```suggestion
            var download = '<a download="' + fileName + '" class="btn btn-outline-secondary btn-sm mr-2" id="file-download-' + uuid + '-' + index + '" onclick="event.stopPropagation();" href="data:application/octet-stream;base64,' + fileData + '">Download</a>';
```

nwalters512 (commented):
Left a suggestion to achieve proper spacing. If the `id` attribute isn't necessary, I'd prefer to remove it. After that, should be good to merge!

elements/pl-file-upload/pl-file-upload.js
jonatanschroeder: I'd suggest `mr-1`, for consistency with the spacing used in `pl-file-preview`.

zillesc: @jonatanschroeder is the onclick necessary here?  (I copied your code, but wasn't sure.)

jonatanschroeder: The `onclick` is to keep the preview from opening. It is to avoid the button click to trigger the opening of the preview, which is caused by clicking anywhere on the container of this link, including the download button.

zillesc: (answered my own question about the onclick...)

zillesc: da39769 adds mr-1 and removes the id.

jonatanschroeder (approved):

nwalters512 (approved):

PR #3663, opened 2021-01-06
News item for time limit change

jonatanschroeder:
News for #2750 .


mwest1066 (approved):

mwest1066:
@jonatanschroeder can you quickly check my changes to make sure I didn't say anything too incorrect? Our general style is to have the news item posted by a different person to the one who implemented the feature, so that we can praise the feature author 😄

jonatanschroeder:
> @jonatanschroeder can you quickly check my changes to make sure I didn't say anything too incorrect? Our general style is to have the news item posted by a different person to the one who implemented the feature, so that we can praise the feature author 😄

Looks good. This change of author needs to be done in 3600 too, then.

PR #3668, opened 2021-01-07
Fix blank github user breaking auto course creation

nicknytko:


pages/instructorRequestCourse/instructorRequestCourse.js
```diff
@@ -96,7 +96,7 @@ router.post('/', function(req, res, next) {
                                 display_timezone: result.rows[0].display_timezone,
                                 path: path.join(config.coursesRoot, repo_short_name),
                                 repo_short_name: repo_short_name,
-                                github_user: github_user.length > 0 ? github_user : null,
+                                github_user: (github_user && github_user.length > 0) ? github_user : null,
```
mwest1066, line 99: From the line `const github_user = req.body['cr-ghuser'] || null;` I think we will already have `github_user == null` if it was a string of length zero? In which case we could just do:
```suggestion
                                github_user,
```

nicknytko: ah, brilliant

mwest1066 (approved):

PR #3670, opened 2021-01-07
New pl-matching element and docs

nealterrell:
This is a first pass on a `pl-matching` element, inspired by the "Matching" question type on my school's LMS (Desire2Learn). 

localhost_3000_pl_course_2_question_90_preview_

```html

<pl-matching answers-name="capitals" number-statements="3" number-options="5" blank="true">
    <pl-statement match="Ottawa">Canada</pl-statement>
    <pl-statement match="d.c.">United States</pl-statement>
    <pl-statement match="Mexico City">Mexico</pl-statement>
    <pl-statement match="Paris">France</pl-statement>
    
    <pl-option>Ottawa</pl-option>
    <pl-option name="d.c.">Washington, D.C.</pl-option>
    <pl-option>Mexico City</pl-option>
    <pl-option>Tijuana</pl-option>
    <pl-option>Calgary</pl-option>
    <pl-option>Toronto</pl-option>
    <pl-option>New York City</pl-option>
    <pl-option>Paris</pl-option>
</pl-matching>
```

The left column of elements are generated from the `pl-answer` children of `pl-matching`. The elements can be any HTML, including other PL elements like drawings or figures. 

The right column of alternatives/options/potential matches are generated from `pl-option` children; they too can show any HTML. Each option specifies a `name` that is used in the `pl-answer` elements to indicate each answer's correct match. Not every option needs to be a match for an answer, but every answer must have a matching option. In the absence of a `name` attribute, a `pl-option`'s name is set to its content (so above, `<pl-option>Ottawa</pl-option>` is assigned the name `Ottawa`).

I mimicked other elements as best I could, including optional attributes that made sense to me:

* `number-answers`, to show fewer than all the answer elements
* `none-of-the-above`, to add a NOTA option to the end of the options list.
* `number-options`, to show fewer than all the option elements, and automatically turns on NOTA.
* `fixed-order`, to show the pl-answers in the same order as written (otherwise they are randomized).
* `blank`, to insert a blank into the drop-down matching list instead of defaulting to the number "1".

Could probably use a little cleanup, and I have some questions if you don't mind:

* Does my pl-matching.mustache template need the "score badge" icons in the "question" panel rendering, or is that only in "answer" and "submission" panels?
* What do I put for tags in my demo question in the example course? My GitHub username, or some other pattern?
* ~~I can't require my brain to use single quotes for strings 😄. But I see that's the general style of your .py files. Shall I replace my double-quotes?~~ answer is certainly yes.


mwest1066:
> This is a first pass on a `pl-matching` element, inspired by the "Matching" question type on my school's LMS (Desire2Learn).

Looks great! I'll review in more detail later.

> * Does my pl-matching.mustache template need the "score badge" icons in the "question" panel rendering, or is that only in "answer" and "submission" panels?

The score badges should be shown in the question and submission panels, and not in the answer panel. The logic is that the question and submission panels show the student's answer, which is the thing that has been scored. The answer panel shows the correct answer, so the scores don't really have meaning there.

> * What do I put for tags in my demo question in the example course? My GitHub username, or some other pattern?

Really any ID of yours that you want. GH username is fine.

> * ~I can't require my brain to use single quotes for strings 😄. But I see that's the general style of your .py files. Shall I replace my double-quotes?~ answer is certainly yes.

Yes, I'm afraid so. We picked a single quote style (because editing code with mixed quote styles is horrible!) and we somewhat arbitrarily picked single quotes.

nealterrell:
I updated the score badges to match your recommendation: visible on the question and submission panels, not on answer panel.

nealterrell:
I added a counter-type attribute to change from numeric numbering of the options to lower-alpha or upper-alpha. I believe this is ready for review.

nealterrell:
I have a few changes to merge from my own personal use of this element for my summer course, then I would love a review and pathway toward merging.

nealterrell:

docs/elements.md
```diff
@@ -608,6 +610,61 @@ Attribute | Type | Default | Description
...
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`answers-name` | string | — | Variable name to store data in.
+`weight` | integer | 1 | Weight to use when computing a weighted average score over elements.
+`fixed-order` | boolean | False | Whether or not to display the answers in a fixed order; otherwise they are shuffled. Options are always shuffled.
+`number-answers` | integer | special | The number of answers to display. Defaults to all answers.
+`number-options` | integer | special | The number of options to display. Defaults to all options.
+`none-of-the-above` | boolean  | false | Whether or not to add a "None of the above" to the end of the options. Automatically set to true if number-options is less than the total number of options.
+`blank` | boolean | False | Option to add blank dropdown entry as the default selection in each drop-down list..
+`counter-type` | "decimal" or "lower-alpha" or "upper-alpha" | "decimal" | The type of counter to use when enumerating the options.
```
jonatanschroeder, line 645: I'd love to see an option where what's shown is the option text itself (either the name or the text). This can be useful when there is a limited number of options. For example, I have a question where students match to "static" or "dynamic", and listing these instead of 1/2 or a/b is probably less error-prone to students. I don't know if this would be a separate option (e.g., `include-option-text` or something like that) or replace the counter in that case, let me know what you think.

elements/pl-matching/pl-matching.mustache
```diff
@@ -0,0 +1,110 @@
...
+            container: 'body',
+            template: '<div class="popover pl-dropdown-popover" role="tooltip"><div class="arrow"></div><h3 class="popover-header"></h3><div class="popover-body"></div></div>',
+        });
+    });
+</script>
+<style>
+.matching-options ol > li:before {
+    content: counter(item, {{counter_type}}) ".";
+}
+</style>
```
jonatanschroeder, line 14: This could be a minor issue if there is more than one `pl-matching` element in the same question, particularly if the counter type is different. Suggestion:
```suggestion
#pl-matching-{{uuid}} .matching-options ol > li:before {
    content: counter(item, {{counter_type}}) ".";
}
</style>
```

elements/pl-matching/pl-matching.mustache
```diff
@@ -0,0 +1,110 @@
...
+<div id="pl-matching-{{uuid}}" style="display: flex;">
+<div class="matching-answers">
+{{#answers}}
+    <div>
+        <div class="matching-answer-select">
+            <select disabled class="form-control">
+                {{#options}}
+                    <option {{selected}} value="{{value}}">{{value}}</option>
+                {{/options}}
+            </select>
```
jonatanschroeder, line 66: Could this be just [formatted] text instead of a disabled select? This would make it more consistent with other elements.

elements/pl-matching/pl-matching.mustache
```diff
@@ -0,0 +1,110 @@
...
+<div id="pl-matching-{{uuid}}" style="display: flex;">
+<div class="matching-answers">
+{{#answers}}
+    <div>
+        <div class="matching-answer-select">
+            <select disabled name="{{name}}" class="form-control">
+                {{#options}}
+                    <option {{selected}} value="{{value}}">{{value}}</option>
+                {{/options}}
+            </select>
```
jonatanschroeder, line 101: Same comment as above.

elements/pl-matching/pl-matching.py
```diff
@@ -0,0 +1,399 @@
...
+PARTIAL_CREDIT_DEFAULT = True
+HIDE_ANSWER_PANEL_DEFAULT = False
+HIDE_HELP_TEXT_DEFAULT = False
+DETAILED_HELP_TEXT_DEFAULT = False
+HIDE_LETTER_KEYS_DEFAULT = False
+HIDE_SCORE_BADGE_DEFAULT = False
+BLANK_DEFAULT = False
+BLANK_ANSWER = ' '
+NOTA_DEFAULT = False
+COUNTER_TYPE_DEFAULT = 'lower-alpha'
```
jonatanschroeder, line 19: Docs say the default is decimal.

elements/pl-matching/pl-matching.py
```diff
@@ -0,0 +1,399 @@
...
+def get_form_name(answers_name, index):
+    return f'{answers_name}-dropdown-{index}'
+
+
+def get_counter(i, counter_type):
+    """Converts an integer counter to the specified CSS counter type"""
+    if counter_type == 'lower-alpha':
+        return chr(ord('a') + i - 1)
+    elif counter_type == 'upper-alpha':
+        return chr(ord('A') + i - 1)
```
jonatanschroeder, line 31: A bit of a pet-peeve, but this could be a problem if there are more than 26 options...

elements/pl-matching/pl-matching.py
```diff
@@ -0,0 +1,399 @@
...
+    elif counter_type == 'decimal':
+        return str(i)
+    else:
+        raise Exception('Illegal counter_type in pl-matching element.')
```
jonatanschroeder, line 35:
```suggestion
        raise Exception('Illegal counter-type in pl-matching element: "' + counter_type + '" should be "decimal", "lower-alpha" or "upper-alpha".')
```

elements/pl-matching/pl-matching.py
```diff
@@ -0,0 +1,399 @@
...
+def legal_answer(answer, counter_type, options):
+    """Checks that the given answer is within the range of the given counter type."""
+    if counter_type == 'lower-alpha':
+        return 'a' <= answer <= chr(ord('a') + len(options))
+    elif counter_type == 'upper-alpha':
+        return 'A' <= answer <= chr(ord('A') + len(options))
+    elif counter_type == 'decimal':
+        return 1 <= answer <= len(options)
+    else:
+        raise Exception('Illegal counter_type in pl-matching element.')
```
jonatanschroeder, line 47: Based on how this function is used, I believe it makes more sense to return false in this case.
```suggestion
        return False
```

elements/pl-matching/pl-matching.py
```diff
@@ -0,0 +1,399 @@
...
+
+    # Shuffle or sample the options.
+    if number_options == len(options):
+        random.shuffle(options)
+    else:
+        options = random.sample(options, number_options)
+        # If we aren't using all options, there is a chance that one answer will be missing its option,
+        # so we force none-of-the-above.
+        nota = True
+    # Then append None of the above if requested.
```
jonatanschroeder, line 120: Wouldn't it make more sense to select the options associated to the selected answers, plus a random sample of distractor options to complete `number_options`? In that case, if `number_options` is less than the number of options required to fulfill the selected answers, I believe it makes sense to set `nota` to True.

elements/pl-matching/pl-matching.py
```diff
@@ -0,0 +1,399 @@
...
+    if nota:
+        options.append((len(options), '__nota__', 'None of the above'))
+
+    # Answers are always sampled/shuffled unless a fixed order is required.
+    if number_answers < len(answers):
+        if not fixed_answer_order:
+            answers = random.sample(answers, number_answers)
+        else:
+            answers = answers[:number_answers]  # Does this configuration even make sense?
```
jonatanschroeder, line 129: How about a random set of answers, but always in the same order as the original list?
```suggestion
            answers = [answer[i] for i in sorted(random.sample(range(len(answers)), number_answers))]
```

elements/pl-matching/pl-matching.py
```diff
@@ -0,0 +1,399 @@
...
+        if student_answer is BLANK_ANSWER:
+            data['format_errors'][expected_html_name] = 'The submitted answer was left blank.'
+        elif student_answer is None:
+            data['format_errors'][expected_html_name] = 'No answer was submitted.'
+        else:
+            try:
+                if not legal_answer(student_answer, counter_type, display_options):
+                    data['format_errors'][expected_html_name] = 'The submitted answer is invalid.'
+            except Exception:
+                data['format_errors'][expected_html_name] = 'The submitted answer is invalid.'
```
jonatanschroeder, line 181: If you change `legal_answer` to return False, this should not be necessary.

elements/pl-matching/pl-matching.py
```diff
@@ -0,0 +1,399 @@
...
+                    'html': option['html'].strip()
+                }
+                optionset.append(option_html)
+
+            html_params = {
+                'submission': True,
+                'answers': answerset,
+                'options': optionset,
+                'display_score_badge': score is not None,
+                'hide_letter_keys': pl.get_boolean_attrib(element, 'hide-letter-keys', HIDE_LETTER_KEYS_DEFAULT),
```
jonatanschroeder, line 287: I don't think this line is necessary.

elements/pl-matching/pl-matching.py
```diff
@@ -0,0 +1,399 @@
...
+                    'key': option['key'],
+                    'html': option['html'].strip()
+                }
+                optionset.append(option_html)
+
+            html_params = {
+                'answer': True,
+                'answers': answerset,
+                'options': optionset,
+                'hide_letter_keys': pl.get_boolean_attrib(element, 'hide-letter-keys', HIDE_LETTER_KEYS_DEFAULT),
```
jonatanschroeder, line 332: I don't think this line is necessary.

elements/pl-matching/pl-matching.py
```diff
@@ -0,0 +1,399 @@
...
+        if student_answer == correct_answer:
+            num_correct += 1
+
+    score = 0
+    if not partial_credit and num_correct == number_answers:
+        score = 1
+    elif partial_credit:
+        # EDC grading
+        number_wrong = number_answers - num_correct
+        score = 1 - 1.0 * number_wrong / number_answers
```
jonatanschroeder, line 368: Why not `num_correct / number_answers`? Note that in Python `/` is floating point division and `//` is integer division, so the `1.0 *` is not necessary.

jonatanschroeder (commented):
Looks great! Looking forward to using this in my courses.

elements/pl-matching/pl-matching.mustache
nealterrell: How would you prefer it to be formatted? Maybe bold and in parentheses, like **(a)** ?

elements/pl-matching/pl-matching.py
nealterrell: Definitely hacky. I wasn't aware that CSS goes from `z` to `aa`, `ab`, etc; will change to handle this.

elements/pl-matching/pl-matching.py
nealterrell: Copied this from pl-checkbox and turned off my brain 😳.

elements/pl-matching/pl-matching.py
nealterrell: Wrote a decimal-to-base-26 helper.

nealterrell:
I slightly reworked the selection of statements and options based on @jonatanschroeder ideas.

First we sample `number-statements` statements to display. We gather the set of correct matching options to display as well.

If the number of matching options equals `number-options`, great!

If it is too small, then we add random distractor options and shuffle the result. 

If the correct options are more than `number-options`, we randomly sample them and add None of the Above as the correct answer for the options that were not sampled.

elements/pl-matching/pl-matching.py
nealterrell: Resolved, see comment on PR.

elements/pl-matching/pl-matching.py
nealterrell: Resolved, see comment on PR.

elements/pl-matching/pl-matching.mustache
jonatanschroeder: Whatever you think looks better. Bold with either parenthesis or followed by a period and space should be fine too.

elements/pl-matching/pl-matching.css
```diff
@@ -0,0 +1,35 @@
+
+.matching-statements > div {
+    display: flex;
+    align-items: baseline;
+    margin-top: 1ex;
+}
+
+.matching-statements > div > div {
+    margin-right: 1ex;
+}
```
jonatanschroeder, line 10: These seem like pretty generic, and I'm afraid a future change may inadvertently add a div that triggers (or not) this formatting in a way that is unintended. I'd be more in favour of adding classes to these divs, and using those in the CSS blocks. Also, nitpick, I think it may be better to have a `pl-` prefix (e.g., `pl-matching-statements` to avoid potential conflict with other elements outside the `pl-matching` element itself.

elements/pl-matching/pl-matching.py
```diff
@@ -0,0 +1,440 @@
...
+
+def alpha_counter(i, letter_a):
+    """Converts a base-10 integer to base 26, using letter_a as the "0" digit."""
+    result = ''
+    while i > 0:
+        i -= 1
+        result = chr(ord(letter_a) + i % 26) + result
+        i //= 26
+
+    return result
```
jonatanschroeder, line 35: I believe `pl.index2key` is equivalent.

elements/pl-matching/pl-matching.py
```diff
@@ -0,0 +1,440 @@
...
+    else:
+        raise Exception('Illegal counter-type in pl-matching element: "' + counter_type + '" should be "decimal", "lower-alpha" or "upper-alpha".')
+
+
+def legal_answer(answer, counter_type, options):
+    """Checks that the given answer is within the range of the given counter type."""
+    if counter_type == 'lower-alpha':
+        return 'a' <= answer <= chr(ord('a') + len(options))
+    elif counter_type == 'upper-alpha':
+        return 'A' <= answer <= chr(ord('A') + len(options))
```
jonatanschroeder, line 55: These don't work with `aa`, `ab`, etc. I'm wondering if it isn't easier to have a check if the answer is in the valid keys? In other words, if there is a `pl-option` with that key?

elements/pl-matching/pl-matching.py
```diff
@@ -0,0 +1,440 @@
...
+            option = {'index': index, 'name': option_name, 'html': child_html}
+            options[option_name] = option
+            index += 1
+
+        elif child.tag in ['pl-statement', 'pl_statement']:
+            pl.check_attribs(child, required_attribs=['match'], optional_attribs=[])
+            child_html = pl.inner_html(child)
+            match_name = pl.get_string_attrib(child, 'match')
+            if match_name not in options:
+                raise Exception(f'pl-statement "match" attribute of {match_name} does not match any of the pl-option elements.')
```
jonatanschroeder, line 107: Not saying this must be done, just raising it as an option: what if a `pl-option` could be implied if it doesn't exist? In that case, `pl-option` would be optional, only used for (1) names that link to longer html, or (2) distractors. This could make the most common cases (where the option is a simple string) easier to create. In your example question, options like `Paris` and `Mexico City` wouldn't need to be added, but options like `d.c.` would be added to change the html of the option, and `New York City` would be a distractor. Thoughts?

elements/pl-matching/pl-matching.py
jonatanschroeder: This would imply changing the options in params to use the converted key (e.g., `a`) instead of the index.

exampleCourse/questions/element/matching/question.html
```diff
@@ -0,0 +1,21 @@
+<pl-question-panel>
```
jonatanschroeder, line 1: The example course question for an element typically demonstrates several options for the element, from the most basic (only `answers-name` set) to each individual argument. The same matching can be used in all cases, the idea is to demonstrate how to use the element with multiple different options.

Also, can you add it to the demo homework?

elements/pl-matching/pl-matching.py
nealterrell: That's a lovely idea.

elements/pl-matching/pl-matching.py
```diff
@@ -95,7 +88,7 @@ def categorize_matches(element, data):
 
             # An option object has: index of appearance in the pl-matching element;
             # the name attribute; and the html content.
-            option = {'index': index, 'name': option_name, 'html': child_html}
+            option = make_option(option_name, child_html)
             options[option_name] = option
             index += 1
```
jonatanschroeder, line 93: shouldn't this be removed, given line 80?

elements/pl-matching/pl-matching.py
```diff
@@ -77,6 +60,10 @@ def partition(data, pred):
     return (yes, no)
 
 
+def get_chosen_option(option_key, display_options):
+    return display_options[option_key]
+
+
```
jonatanschroeder, line 66: I'm not sure I see the merit of having a function for this operation.

exampleCourse/questions/element/matching/question.html
```diff
@@ -1,21 +1,113 @@
...
+    
+    <pl-option>Tijuana</pl-option>
+    <pl-option>Toronto</pl-option>
+    <pl-option>New York City</pl-option>
+</pl-matching>
+
+<pl-question-panel>
+    <p>
+        Example 5: the <code>counter-type</code> can be changed to lower-alpha, upper-alpha, or
+        decimal.
```
jonatanschroeder, line 85:
```suggestion
        Example 5: the <code>counter-type</code> option indicates how options are identified. It can be set to <code>lower-alpha</code> (a, b, c, ...), <code>upper-alpha</code> (A, B, C, ...), or <code>decimal</code> (1, 2, 3).
```

exampleCourse/questions/element/matching/question.html
```diff
@@ -1,21 +1,113 @@
...
+        All PrairieLearn and HTML elements are supported in options and statements.
+    </p>
+</pl-question-panel>
+    
+<pl-matching answers-name="ex6">
+    <pl-statement match="USA">United States</pl-statement>
+    <pl-statement match="France">France</pl-statement>
+    
+    <pl-option name="USA"><img src="https://upload.wikimedia.org/wikipedia/en/thumb/a/a4/Flag_of_the_United_States.svg/320px-Flag_of_the_United_States.svg.png"/></pl-option>
+    <pl-option name="France"><img src="https://upload.wikimedia.org/wikipedia/en/thumb/c/c3/Flag_of_France.svg/320px-Flag_of_France.svg.png"/></pl-option>
```
jonatanschroeder, line 112: This is a great example of using custom HTML, but I don't like relying on external resources. Can you copy these images to the `questionFilesCourse` folder? Alternatively, can you make a question that uses the files in `exampleCourse/clientFilesCourse/student_images`? Questions like `demo/studentFaces` or `demo/studentNames` may be a starting point for what to expect.

docs/elements.md
```diff
@@ -642,6 +644,63 @@ Attribute | Type | Default | Description
...
+
+A total of `number-statements` statements will be randomly selected and displayed to the student. The corresponding matching options will be gathered; if `number_options` requires more entries, then distractors will be selected from the remaining unused options. If the gathered options are more numerous than `number_options`, they will be randomly sampled and `none-of-the-above` will automatically set to true.
+
+The content of a `pl-statement` can be any HTML element, including other PrairieLearn elements. A `pl-statement` must be specified with these attributes:
+
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`match` | string | — | Must be equal to the `name` of one of the `pl-option` elements, and identifies that option as the correct response for this `pl-answer.`
+
+The content of a `pl-option` can be any HTML element, including other PrairieLearn eleemnts. `A pl-option` must be specified with these attributes:
```
jonatanschroeder, line 692:
```suggestion
The content of a `pl-option` can be any HTML element, including other PrairieLearn elements. Including `pl-option` tags is optional, since options are created by default based on the `match` setting of each `pl-statement`;   additional `pl-option` tags can be added to serve as distractors (an option that is always incorrect, such as "New York City" in the example above), or to provide a more detailed description of the option (such as "d.c." in the example above). A `pl-option` may specified with these attributes:
```

docs/elements.md
```diff
@@ -642,6 +644,63 @@ Attribute | Type | Default | Description
...
+
+Inside the `pl-matching` element, a series of `pl-statement` and `pl-option` elements specify the questions the student must answer and the options to which they can be matched, respectively. Statements are displayed in the left column, and options in the right.
+
+A total of `number-statements` statements will be randomly selected and displayed to the student. The corresponding matching options will be gathered; if `number_options` requires more entries, then distractors will be selected from the remaining unused options. If the gathered options are more numerous than `number_options`, they will be randomly sampled and `none-of-the-above` will automatically set to true.
+
+The content of a `pl-statement` can be any HTML element, including other PrairieLearn elements. A `pl-statement` must be specified with these attributes:
+
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`match` | string | — | Must be equal to the `name` of one of the `pl-option` elements, and identifies that option as the correct response for this `pl-answer.`
```
jonatanschroeder, line 690:
```suggestion
`match` | string | — | Identifies that option as the correct response for this `pl-answer`. If `match` corresponds to the `name` of any `pl-option` element, the statement will be linked to that `pl-option`, otherwise a new option is implicitly created based on this `match` value.
```

docs/elements.md
```diff
@@ -642,6 +644,63 @@ Attribute | Type | Default | Description
...
+**question.html**
+```html
+<pl-matching answers-name="string_value">
+  <pl-statement match="d.c.">United States</pl-statement>
+  <pl-statement match="Mexico City">Mexico</pl-statement>
+  <pl-statement match="Paris">France</pl-statement>
+
+  <pl-option name="d.c.">Washington, D.C.</pl-option>
+  <pl-option>Mexico City</pl-option>
+  <pl-option>Paris</pl-option>
```
jonatanschroeder, line 664:
```suggestion
  <pl-option>New York City</pl-option>
```

docs/elements.md
```diff
@@ -642,6 +644,63 @@ Attribute | Type | Default | Description
...
+
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`answers-name` | string | — | Variable name to store data in.
+`weight` | integer | 1 | Weight to use when computing a weighted average score over elements.
+`fixed-order` | boolean | False | Whether or not to display the statements in a fixed order; otherwise they are shuffled. Options are always shuffled.
+`number-statements` | integer | special | The number of statements to display. Defaults to all statements.
+`number-options` | integer | special | The number of options to display. Defaults to all options.
+`none-of-the-above` | boolean  | false | Whether or not to add a "None of the above" to the end of the options.
+`blank` | boolean | False | Option to add blank dropdown entry as the default selection in each drop-down list..
```
jonatanschroeder, line 678:
```suggestion
`blank` | boolean | False | Option to add blank dropdown entry as the default selection in each drop-down list.
```

I'm wondering also if it should default to True instead, which is the default behaviour of `pl-dropdown`. What do you think?

exampleCourse/questions/element/matching/question.html
```diff
@@ -0,0 +1,113 @@
+<pl-question-panel>
+    <p>
+        This question illustrates different use cases for the
+        <a href="http://prairielearn.readthedocs.io/en/latest/elements/#pl-matching-element"><code>pl-matching</code></a>
+        element.
+        </p>
+    <p>
```
jonatanschroeder, line 7: For separating the examples, can you use boxes (cards) like those in `elements/integerInput`?

jonatanschroeder:
@mwest1066 do you want a news item for this one?

elements/pl-matching/pl-matching.py
nealterrell: A victim of refactoring... :)

docs/elements.md
nealterrell: I'm fine with that.

docs/elements.md
```diff
@@ -642,6 +644,63 @@ Attribute | Type | Default | Description
...
+`counter-type` | "decimal" or "lower-alpha" or "upper-alpha" or "full-text" | "lower-alpha" | The type of counter to use when enumerating the options. If set to "full-text", the column of options will be hidden, and the text of each option will be used in the statements' dropdown lists, instead of counters.
+`hide-score-badge` | boolean | false | Whether or not to hide the correct/incorrect score badge next to each graded answer choice.
+
+Inside the `pl-matching` element, a series of `pl-statement` and `pl-option` elements specify the questions the student must answer and the options to which they can be matched, respectively. Statements are displayed in the left column, and options in the right.
+
+A total of `number-statements` statements will be randomly selected and displayed to the student. The corresponding matching options will be gathered; if `number_options` requires more entries, then distractors will be selected from the remaining unused options. If the gathered options are more numerous than `number_options`, they will be randomly sampled and `none-of-the-above` will automatically set to true.
```
jonatanschroeder, line 682:
```suggestion
A total of `number-statements` statements will be randomly selected and displayed to the student. The corresponding matching options will be gathered; if `number-options` is larger than the number of options used by the selected statements, then random distractors will be selected from the remaining unused options. If the selected statements require more options than `number-options`, then `none-of-the-above` will automatically set to true.
```

docs/elements.md
```diff
@@ -642,6 +644,63 @@ Attribute | Type | Default | Description
...
+
+#### Customizations
+
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`answers-name` | string | — | Variable name to store data in.
+`weight` | integer | 1 | Weight to use when computing a weighted average score over elements.
+`fixed-order` | boolean | False | Whether or not to display the statements in a fixed order; otherwise they are shuffled. Options are always shuffled.
+`number-statements` | integer | special | The number of statements to display. Defaults to all statements.
+`number-options` | integer | special | The number of options to display. Defaults to all options.
```
jonatanschroeder, line 674:
```suggestion
`number-options` | integer | special | The number of options to display. Defaults to all options. The `none-of-the-above` option does not count towards this number.
```

That is true, right?

docs/elements.md
```diff
@@ -642,6 +644,63 @@ Attribute | Type | Default | Description
...
+
+The content of a `pl-statement` can be any HTML element, including other PrairieLearn elements. A `pl-statement` must be specified with these attributes:
+
+Attribute | Type | Default | Description
+--- | --- | --- | ---
+`match` | string | — | Identifies the option as the correct response for this `pl-statement`. If `match` corresponds to the `name` of any `pl-option` element, the statement will be linked to that `pl-option`, otherwise a new option is implicitly created based on this `match` value.
+
+The content of a `pl-option` can be any HTML element, including other PrairieLearn elements. `pl-option` elements are optional; options are created by default based on the `match` attribute of each `pl-statement`. Additional `pl-option` elements can be added to serve as distractors (an option that is always incorrect, such as "New York City" in the example above), or to render formatted HTML/PrairieLearn elements instead of plain text (see the last question in the demo problem linked in the "Example implementations" below).
+
+`A pl-option` must be specified with these attributes:
```
jonatanschroeder, line 692:
```suggestion
A `pl-option` must be specified with these attributes:
```

elements/pl-matching/pl-matching.mustache
```diff
@@ -0,0 +1,108 @@
...
+{{/question}}
+
+{{#submission}}
+<div id="pl-matching-{{uuid}}" style="display: flex;">
+<div class="pl-matching-statements">
+{{#statements}}
+    <div class="pl-matching-statement">
+        {{^no_counters}}
+        <div class="pl-matching-statement-submission">
+            <b>{{option}}.</b>
```
jonatanschroeder, line 65:
```suggestion
            <strong>{{option}}.</strong> 
```
Better for accessibility.

elements/pl-matching/pl-matching.mustache
```diff
@@ -0,0 +1,108 @@
...
+{{/submission}}
+
+{{#answer}}
+<div id="pl-matching-{{uuid}}" style="display: flex;">
+<div class="pl-matching-statements">
+{{#statements}}
+    <div class="pl-matching-statement">
+        {{^no_counters}}
+        <div class="pl-matching-statement-submission">
+            <b>{{option}}.</b>
```
jonatanschroeder, line 98:
```suggestion
            <strong>{{option}}.</strong>
```

elements/pl-matching/pl-matching.py
```diff
@@ -0,0 +1,438 @@
...
+            statements.append(statement)
+
+    return list(options.values()), statements
+
+
+def prepare(element_html, data):
+    element = lxml.html.fragment_fromstring(element_html)
+
+    required_attribs = ['answers-name']
+    optional_attribs = ['fixed-order', 'number-statements', 'number-options', 'none-of-the-above', 'blank', 'counter-type', 'no-counters']
```
jonatanschroeder, line 110:
```suggestion
    optional_attribs = ['fixed-order', 'number-statements', 'number-options', 'none-of-the-above', 'blank', 'counter-type']
```

elements/pl-matching/pl-matching.py
```diff
@@ -0,0 +1,438 @@
...
+def get_counter(i, counter_type):
+    """Converts an integer counter to the specified CSS counter type"""
+    if counter_type == 'lower-alpha':
+        return pl.index2key(i - 1)
+    elif counter_type == 'upper-alpha':
+        return pl.index2key(i - 1).upper()
+    elif counter_type == 'decimal':
+        return str(i)
+    else:
+        raise Exception('Illegal counter-type in pl-matching element: "' + counter_type + '" should be "decimal", "lower-alpha" or "upper-alpha".')
```
jonatanschroeder, line 36: `full-text` support is missing here.

elements/pl-matching/pl-matching.py
```diff
@@ -0,0 +1,438 @@
...
+FIXED_STATEMENTS_ORDER_DEFAULT = False
+FIXED_OPTIONS_ORDER_DEFAULT = False
+INLINE_DEFAULT = False
+PARTIAL_CREDIT_DEFAULT = True
+HIDE_ANSWER_PANEL_DEFAULT = False
+HIDE_HELP_TEXT_DEFAULT = False
+DETAILED_HELP_TEXT_DEFAULT = False
+HIDE_SCORE_BADGE_DEFAULT = False
+BLANK_DEFAULT = True
+BLANK_ANSWER = ' '
```
jonatanschroeder, line 18: I believe this is now -1, right?

nealterrell: This is the literal string shown in the `select` list.

elements/pl-matching/pl-matching.py
```diff
@@ -0,0 +1,438 @@
...
+    for i in range(len(display_statements)):
+        expected_html_name = get_form_name(name, i)
+        try:
+            student_answer = int(submitted_answers.get(expected_html_name, None))
+        except (ValueError, TypeError):
+            data['format_errors'][expected_html_name] = 'The submitted answer is not a legal option.'
+            continue
+
+        # A blank is a valid submission from the HTML, but causes a format error.
+        if student_answer is BLANK_ANSWER:
```
jonatanschroeder, line 211:
```suggestion
        if student_answer == -1:
```

jonatanschroeder:
I've done a few more tests, and one concern I have looking now at the examples is that the answer and submission panels show the option name, but not the statement, which looks a bit confusing, since you have to scroll back to the question to find what each option refers to.

image

Can you think of a better way to show both the matching option and the statement? I'm not sure what the best way is (feel free to brainstorm or suggest ideas). One option is to have a box around the option counter + html (similar to bootstrap badges), followed by a small margin and the statement itself.

Also, for full-text selects the "blank" answer is kept blank in answer/submission, rather than the `[blank].` text used in the other types.

About the news item, it's tradition to have the official author be someone other than the one that implemented. I'll build on top of your news item and send a commit.

nealterrell:
> I've done a few more tests, and one concern I have looking now at the examples is that the answer and submission panels show the option name, but not the statement, which looks a bit confusing, since you have to scroll back to the question to find what each option refers to.
> 
> image
> 
> Can you think of a better way to show both the matching option and the statement? I'm not sure what the best way is (feel free to brainstorm or suggest ideas). One option is to have a box around the option counter + html (similar to bootstrap badges), followed by a small margin and the statement itself.
> 
> Also, for full-text selects the "blank" answer is kept blank in answer/submission, rather than the `[blank].` text used in the other types.
> 
> About the news item, it's tradition to have the official author be someone other than the one that implemented. I'll build on top of your news item and send a commit.

Yes, I feel the same about the answers and submission views. Originally those views were the same as the question view, except the select list was disabled on the correct or submitted answers. I can try to mimic that look but with divs instead? But that kind of begs the question of why not just go back to a disabled select.

jonatanschroeder (approved):

news_items/034_matching_element/index.html
```diff
@@ -0,0 +1,65 @@
...
+    <a href="https://github.com/nealterrell">Neal Terrell</a> (California State University, Long Beach), with input from
+    <a href="https://github.com/jonatanschroeder">Jonatan Schroeder</a>.
+</p>
+
+<p>
+    This element can be used to create two-column &quot;matching&quot; problems, where students
+    must select an option that best matches each entry in a list of statements.
+</p>
+
+<img class="border rounded" src="pl-matching.png" width="600px">
```
mwest1066, line 39: I'm seeing a double-border on the left of this image in the news item:
image

exampleCourse/questions/element/matching/question.html
```diff
@@ -0,0 +1,154 @@
...
+        </pl-matching>
+    </div>
+</div>
+
+
+<div class="card my-2">
+    <div class="card-body">
+        <pl-question-panel>
+            <p>
+                Example 7: Setting <code>counter-type="full-text"</code> will hide the column of options. The dropdown list
```
mwest1066, line 144: This example 7 using `counter-type="full-text"` is great! I find it much less error prone than using letters as keys to match between statements and options.

I'm wondering why this isn't the default mode? From what I can see the only downside is if the options can't be displayed in a dropdown (e.g., equations, images, other complex HTML). If we have text options then I think I'd always want to use `full-text`.

As I was looking at this element, I was thinking to myself that I didn't like the matching via letter keys, and I was wishing that I could just directly match options to statements. For example, I was thinking about something like `<pl-order-blocks>` where the student could drag options next to matching statements. But then I saw Example 7 and I was amazed! This was exactly what I wanted! So my only question is why it was hidden as the 7th option rather than being the primary mode of use? 😄

news_items/034_matching_element/index.html
```diff
@@ -0,0 +1,65 @@
...
+     display: block;
+     margin-bottom: -1em;
+     line-height: 125%;
+ }
+</style>
+
+<p>
+    PrairieLearn now has a new <code class="user-output">pl-matching</code> element, thanks to the efforts of 
+    <a href="https://github.com/nealterrell">Neal Terrell</a> (California State University, Long Beach), with input from
+    <a href="https://github.com/jonatanschroeder">Jonatan Schroeder</a>.
```
mwest1066, line 31: Does @jonatanschroeder want an institutional association?

news_items/034_matching_element/info.json
```diff
@@ -0,0 +1,6 @@
+{
+    "uuid": "bed17803-fce6-4732-96e1-4b91e30a907b",
+    "title": "New pl-matching element",
+    "author": "<a href=\"https://github.com/nealterrell\">Neal Terrell</a>",
```
mwest1066, line 4: We generally have someone other than the PR author listed as the news item author, to avoid having it look like someone is praising themselves in the news item 😄. Maybe @jonatanschroeder could be the news item author?

jonatanschroeder: Do you want a joint news item that includes the other elements? In that case it may need to be you...

exampleCourse/questions/element/matching/question.html
nealterrell: My initial design was a straightforward reimplementation of my LMS's matching quiz element, which uses numeric counters; full-text counters weren't even on my radar until Jonatan's suggestion. 

I personally feel that full-text is quite restrictive -- only being applicable for very short option texts, which in my personal use history is somewhat uncommon. But I am happy to make that the default and/or rearrange the examples if you want to promote that style as the default.

news_items/034_matching_element/index.html
jonatanschroeder: I removed my name since I'm the author. No need to mention me, I'm getting enough attention as is 😉

exampleCourse/questions/element/matching/question.html
mwest1066: I agree that it's probably fine to leave the default as is. But maybe we could elevate Example 7 up to Example 2?

mwest1066:
Related to this, the Calc 3 teaching team at Illinois came to our PL office hours today and asked how they can do matching-type problems. I was very happy to tell them they they should just wait a day or two and there will be a brand new element for this!

exampleCourse/questions/element/matching/question.html
nealterrell: Done!

mwest1066: Awesome, thanks!

mwest1066 (approved):

mwest1066:
@jonatanschroeder I'm good with this PR -- merge whenever you like.

news_items/034_matching_element/index.html
jonatanschroeder:
```suggestion
<img src="pl-matching.png" width="600px">
```

PR #3074, opened 2020-09-06
Add code to redirect instructor question page for student

polohan:

PR #3672, opened 2021-01-08
Offer enrollment if student uses course link but is not enrolled

jonatanschroeder:
This PR replaces the 403 error message when uses a course instance link they don't have access to with an offer to enroll in it, if such an enrollment is possible. This should make it possible for instructors to post the course link directly in their external pages, and it would offer enrollment if needed.

I was a bit limited in my testing options, so I'd welcome additional testing.


mwest1066:
I appreciate the fact that redirecting students to the `enroll` page gives them an opportunity to enroll in the class. But I'm a bit worried about the fact that this doesn't provide "one-click" access into the course.

As an alternative, could we just directly insert an enrollment if it's available and allow the student to continue directly to the linked assessment (or whichever page they are trying to access)? I'm thinking that, if allowed, we could just force an enrollment here: https://github.com/PrairieLearn/PrairieLearn/blob/3676c69135d341c7e92e63c52b215df48cefb98c/sprocs/authz_course_instance.sql#L15

Actually, maybe this would be better done as part of #3020? In that branch, it looks like it would be very easy to replace this PERFORM with an UPSERT to insert an enrollment if we don't have one yet: https://github.com/PrairieLearn/PrairieLearn/blob/9ae4ea78af28005302f5c03ab5bc747b68547ce7/sprocs/authz_course_instance.sql#L58

I suggest that we could just do this unconditionally at the moment, but in the future it would also be easy to add an access rule like `allowSelfEnrollment` to control whether students can self (and auto) enroll.

Thoughts?

jonatanschroeder:
By "one-click" you mean this new enrollment page should basically (1) enroll them without asking, or (2) provide a single confirm button that, when pressed, enrolls them and directs them straight to the next URL? I for sure like the second option, but I think the first one can be dangerous. Students should be somehow warned that they are enrolling in the course on PrairieLearn.

mwest1066:
> By "one-click" you mean this new enrollment page should basically (1) enroll them without asking, or (2) provide a single confirm button that, when pressed, enrolls them and directs them straight to the next URL? I for sure like the second option, but I think the first one can be dangerous. Students should be somehow warned that they are enrolling in the course on PrairieLearn.

I was thinking about the first one. The student clicks on a link in Canvas (or wherever) to "HW1" and they just go straight to the assessment in PL. They will still be prompted to login if they haven't used PL before, so they aren't unknowingly giving away login information or anything like that.

Enrollments in PL are very lightweight (it just adds the course to the list of courses they see on the PL homepage) and they can remove it again anytime they want.

But if we do have a clickthrough agreement, then I think we should continue the redirect through to their original URL afterwards. This redirect is much easier to implement with the silent auto-add of the enrollment, however 😄

jonatanschroeder:
> I was thinking about the first one. The student clicks on a link in Canvas (or wherever) to "HW1" and they just go straight to the assessment in PL. They will still be prompted to login if they haven't used PL before, so they aren't unknowingly giving away login information or anything like that.
> 
> Enrollments in PL are very lightweight (it just adds the course to the list of courses they see on the PL homepage) and they can remove it again anytime they want.
> 
> But if we do have a clickthrough agreement, then I think we should continue the redirect through to their original URL afterwards. This redirect is much easier to implement with the silent auto-add of the enrollment, however 😄

That sounds good. I'll try to work on that.

mwest1066:
> That sounds good. I'll try to work on that.

Now that I've looked at it, this might be something for which it's worth waiting until after #3020 is merged. I talked to @tbretl today and I'm hopeful that #3020 will be merged soon (early next week).

I guess that might not be early enough for UBC?

jonatanschroeder:
Sure, that's fine, I guess I can wait.

jonatanschroeder:
From #3074 

- [ ] Add a row to the instructorAssessmentSettings page for "Student link", with a value of https://HOSTNAME/pl/course_instance/:course_instance_id/assessment/:assessment_id/. If instructors paste this URL into their course homepages, students can click on the URL to go directly to their personal assessment.
- [ ] Add docs to docs/assessment.md in a new section called "Linking to assessments" and describe that you can get a link for students to a specific assessment from the Settings page, and that these links will auto-enroll students who follow them.

tests/index.js
```diff
@@ -29,7 +29,7 @@ require('./testRealTimeGradingDisabled');
 require('./testShowClosedAssessmentScore');
 require('./testGradeRate');
 require('./testAccess');
-require('./testAccessAsStudent');
+// require('./testAccessAsStudent');
```
mwest1066, line 32: Are all of these tests no longer needed? Or do they just need to be fixed?

jonatanschroeder: The premise of the test is no longer valid. It tests whether access is denied for a student that is not enrolled. How would you like to proceed? We can remove the test, or modify it to (1) check that enrollment was added and/or (2) check whether access is denied to a user that shouldn't have access (e.g., out of instance access date range, or invalid institution).

mwest1066: > The premise of the test is no longer valid. It tests whether access is denied for a student that is not enrolled. How would you like to proceed? We can remove the test, or modify it to (1) check that enrollment was added and/or (2) check whether access is denied to a user that shouldn't have access (e.g., out of instance access date range, or invalid institution).

Can we modify it to check both (1) and (2)? 😄

jonatanschroeder: Done. (2) was done by creating a new course instance that has no access rule, than trying to acccess the assessments page for that course.

tests/testAccessAsStudent.js
```diff
@@ -1,14 +1,23 @@
...
 const baseUrl = siteUrl + '/pl';
 const courseInstanceUrl = baseUrl + '/course_instance/1';
 const assessmentsUrl = courseInstanceUrl + '/assessments';
 
 const storedConfig = {};
 
-describe('Test access with authn as student', function() {
+var newAssessmentsUrl;
+
+describe('Test student auto-enrolment', function() {
```
mwest1066, line 20:
```suggestion
describe('Test student auto-enrollment', function() {
```

mwest1066 (approved):

PR #3673, opened 2021-01-09
Add option for pl-integer-input to be provided in a non-10 base

jonatanschroeder:
This allows hex, binary, octal and other non-trivial bases for integers.

### DIFF
--- a/elements/pl-integer-input/pl-integer-input.py
+++ b/elements/pl-integer-input/pl-integer-input.py
@@ -209,6 +209,7 @@ def parse(element_html, data):
 def grade(element_html, data):
     element = lxml.html.fragment_fromstring(element_html)
     name = pl.get_string_attrib(element, 'answers-name')
+    base = pl.get_integer_attrib(element, 'base', BASE_DEFAULT)
 
     # Get weight
     weight = pl.get_integer_attrib(element, 'weight', WEIGHT_DEFAULT)
--- a/elements/pl-integer-input/pl-integer-input.py
+++ b/elements/pl-integer-input/pl-integer-input.py
@@ -26,13 +26,16 @@ def prepare(element_html, data):
     name = pl.get_string_attrib(element, 'answers-name')
     base = pl.get_integer_attrib(element, 'base', BASE_DEFAULT)
 
-    if base < 2 or base > 36:
-        raise Exception('Base must be between 2 and 36')
+    if base != 0 and (base < 2 or base > 36):
+        raise Exception('Base must be either 0, or between 2 and 36')
 
-    correct_answer = pl.string_to_integer(pl.get_string_attrib(element, 'correct-answer', CORRECT_ANSWER_DEFAULT), base)
+    correct_answer = pl.get_string_attrib(element, 'correct-answer', CORRECT_ANSWER_DEFAULT)
     if correct_answer is not None:
         if name in data['correct_answers']:
             raise Exception('duplicate correct_answers variable name: %s' % name)
+        # Test conversion, but leave as string so proper value is shown on answer panel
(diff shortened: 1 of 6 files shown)


question-servers/freeformPythonLib/prairielearn.py
```diff
@@ -578,22 +578,32 @@ def string_partition_outer_interval(s, left='[', right=']'):
...
+
     # Replace unicode minus with hyphen minus wherever it occurs
     s = s.replace(u'\u2212', '-').strip()
+
+    # If base is 16, strip "0x" prefix
+    if base == 16 and s.startswith('0x'):
+        s = s[2:].strip()
+    elif base == 16 and s.startswith('-0x'):
+        s = '-' + s[3:].strip()
+
```
nwalters512, line 599: Can you explain why this section is necessary? AFAICT, `int()` handles this for us:

```python
>>> int('FFFFFF', 16)
16777215
>>> int('0xFFFFFF', 16)
16777215
>>> int('-FFFFFF', 16)
-16777215
>>> int('-0xFFFFFF', 16)
-16777215
```

elements/pl-integer-input/pl-integer-input.py
```diff
@@ -112,7 +126,7 @@ def render(element_html, data):
             a_sub = pl.from_json(a_sub)
 
             html_params['suffix'] = suffix
-            html_params['a_sub'] = '{:d}'.format(a_sub)
+            html_params['a_sub'] = numpy.base_repr(a_sub, base=base)
```
nwalters512, line 129: Nit: it doesn't really matter and I'm not requesting a change, but as a relatively inexperienced Python dev I'm curious about when one would prefer to use keyword arguments as you've done here vs. positional arguments like you use elsewhere with the base argument to `get_integer_attrib`.

nwalters512 (approved):
Looks great! Thanks for the nice addition.

question-servers/freeformPythonLib/prairielearn.py
jonatanschroeder: Good point, I didn't realize that. In light of that I removed this piece of code, but I also added an option for special base value 0.

elements/pl-integer-input/pl-integer-input.py
jonatanschroeder: I personally like to add keyword arguments when the function receives multiple arguments and I want to make it clear which one I'm assigning a value to. In any case I removed it for consistency with the other cases.

nwalters512: What does it mean for a number to be base-0? I've never encountered that idea before - how would that work?

question-servers/freeformPythonLib/prairielearn.py
nwalters512: What does it mean for a number to be base 0? I've never encountered that idea before - how would that work?

nwalters512: Ah, I should read the docs you wrote, shouldn't I...

Would it make sense for that to be the default behavior no matter what base is specified? i.e., even if `base="7"`, I could still write "0x123abc" as an answer?

jonatanschroeder: I would prefer to use base=0 for that, otherwise you may have problems with inputs like 0777. Most students would expect this to mean base 10, except when you explicitly allow for it to be base 8.

jonatanschroeder: and if you need the base to be something other than 10, you probably want to force students to use that base as part of the question...

mwest1066: Makes sense to me.

mwest1066 (approved):

PR #3684, opened 2021-01-14
Implement Java autograder

jonatanschroeder:
Basic implementation of Java autograder. Initial version uses C grader as a basis.

- [x] ~~Basic implementation of single Java file with main function (simple executable class file).~~
    - Removed, focus on JUnit only (most common case and better testing options). An example question will be added that shows how to use JUnit to check output.
- [x] Proper implementation and testing for applications with multiple Java files.
- [x] Combine student-provided Java files with instructor-provided Java files.
- [x] Support for Java libraries and external JAR files.
- [x] Support for JUnit tests.
-  ~~Support for Maven/Gradle/Ant-based compilation rules (??).~~
    - will leave for future implementation.
- [x] Documentation.
- [x] Sample question(s) in example course.


graders/java/JUnitAutograder.java
```diff
@@ -0,0 +1,242 @@
...
+
+    private void saveResults() {
+
+        JSONArray resultsTests = new JSONArray();
+        //Collections.sort(this.tests);
+        for (AutograderTest test : this.tests)
+            resultsTests.add(test.toJson());
+
+        JSONObject results = new JSONObject();
+        results.put("score", this.maxPoints > 0 ? this.points / this.maxPoints : 0);
```
falkirks, line 194: I can't figure out why this object doesn't include `succeeded` which the spec for external grading says is required (https://prairielearn.readthedocs.io/en/stable/externalGrading/)? However when I try to debug locally it seems that something in between is adding the `succeeded` field. But if I am reading the docs right the `succeeded` field should be added here.

falkirks (commented):
This looks great! I added a few small comments. I'm really excited about this 😄 

This is probably out of scope for this PR, but it seems a bit weird that the student code is compiled also with junit in the class path, allowing students to just use junit in their code. I can't think of any really terrible things that could be done with this, but it just seems like a weird thing to have. I assume fixing this would look like changing `compileFiles` to allow for a custom class path to be passed in. This would probably be better suited for a future PR, but just wanted to flag it here.

graders/java/JUnitAutograder.java
```diff
@@ -0,0 +1,242 @@
...
+
+        this.tests.add(autograderTest);
+        this.points += autograderTest.points;
+        this.maxPoints += autograderTest.maxPoints;
+    }
+
+    private void saveResults() {
+
+        JSONArray resultsTests = new JSONArray();
+        //Collections.sort(this.tests);
```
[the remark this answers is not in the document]
jonatanschroeder: Originally I thought of having the tests sorted by test name, before I discovered that JUnit 5 allows you to set the order of the tests. So I ended up adding the tests in the order reported by JUnit, which means I don't need to sort it there.

graders/java/JUnitAutograder.java
jonatanschroeder: Seems like that documentation is incorrect. If I understand the code correctly (https://github.com/PrairieLearn/PrairieLearn/blob/ba42f43cc91adcef1f883a1ebad5ab927923aeeb/grader_host/index.js#L514), `succeeded` is ignored and set to true as long as the results.json file is properly parsed. @mwest1066 can you confirm this is indeed the case? In any case, I just `gradable` for what is described in the docs as `succeeded`, and that works for the C grader as well.

jonatanschroeder:
> This is probably out of scope for this PR, but it seems a bit weird that the student code is compiled also with junit in the class path, allowing students to just use junit in their code. I can't think of any really terrible things that could be done with this, but it just seems like a weird thing to have. I assume fixing this would look like changing `compileFiles` to allow for a custom class path to be passed in. This would probably be better suited for a future PR, but just wanted to flag it here.

That's a good point. I didn't want to have a different classpath for their code versus the test code, to avoid conflicts, and I don't think having the junit library (or even json-simple) would be a problem.

There is a problem, though, which I'm planning to address in a future PR, which is code sandboxing. The current implementation would make it viable for a student to change the `results.json` file then throw an Error to keep it from being overwritten by the Autograder class. I changed the entrypoint to allow the future PR more flexibility in how to handle this sandboxing.

graders/java/JUnitAutograder.java
mwest1066: I think this is actually a bug in `grader_host`. We are trying to distinguish between three different types of error:
1. Error in the grader code itself. This is indicated by `succeeded = false`. Ideally this should never happen. In this case we will ignore any other data returned by the grader. This grading attempt won't count against the student in any way (no points lost, no attempts used).
2. Error in the student code for which we want to abandon grading. This is indicated by setting `succeeded = true` and `gradable = false`. This will be reported back to the student as an "Invalid" submission (hopefully with detailed information from the grader about what was wrong). This grading attempt won't count against the student in any way (no points lost, no attempts used).
3. Test case errors in the student code. These are "normal" errors which cause test cases to fail. We return `succeeded = true` and `gradable = true` and points are calculated based on the number of test cases that pass.

I think that `grader_host` should not be setting `succeeded = true`. Instead, I think that it should be doing something like:
```
results.succeeded = (results.succeeded == null ? true : results.succeeded);
```
That is, if `succeeded` is missing then we'll set it to `true` to handle old graders that aren't setting it correctly.

Ideally we would do `results.succeeded = !!results.succeeded`, which would force graders to return `succeeded = true` if they want results to be processed. But I worry that this would break backwards compatibility.

@jonatanschroeder does this sound correct to you? Should we change `grader_host` and or the docs?

jonatanschroeder: It seems the docs @falkirks was referring to are not the latest version. The latest version (https://prairielearn.readthedocs.io/en/latest/externalGrading/) doesn't make any reference at all to `succeeded`. @mwest1066  how do you want to proceed?

mwest1066: Good catch! Let's leave all this as is. No changes needed.

mwest1066:
Does the resolution of the `succeeded` flag question mean that this is ready to go? All of @falkirks's items have been resolved?

falkirks (approved):

falkirks:
It all looks good to me

jonatanschroeder:
> Does the resolution of the `succeeded` flag question mean that this is ready to go? All of @falkirks's items have been resolved?

All that's needed is for you to review the news item...

mwest1066 (approved):

PR #3699, opened 2021-01-19
Fix pl-dropdown badge in case of save without grade

### LINKED ISSUE #3362: Pl-dropdown grades 0% on submission
Pl-dropdown is labelling submissions as 0% on the UI whether the submissions are incorrect or correct. It would be ideal if the submissions could not show a 0% grade beside submissions until they are graded. Other elements just show submissions as such: `highlighted text`.

0% showing up under submissions on pl-dropdown:
image

jonatanschroeder:
Fixes #3362

### DIFF
--- a/elements/pl-dropdown/pl-dropdown.mustache
+++ b/elements/pl-dropdown/pl-dropdown.mustache
@@ -32,12 +32,14 @@
         {{^parse-error}}
             <code class="user-output">{{{submitted-answer}}}</code>
 
+            {{#display-score-badge}}
             {{#correct}}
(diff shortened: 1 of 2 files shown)


elements/pl-dropdown/pl-dropdown.mustache
```diff
@@ -32,12 +32,14 @@
         {{^parse-error}}
             <code class="user-output">{{{submitted-answer}}}</code>
 
+            {{#display-score-badge}}
             {{#correct}}
```
andrewstec, line 36: This subsequent block should be indented.

jonatanschroeder: Your wish has been granted...

PR #3704, opened 2021-01-22
Clean up gitignore

astrosticks:
Cleaning up gitignore, moving related sections near each other. Only have comments when pattern doesn't explain itself.

Other changes:
 - Renamed `start.sh` to `start-custom.sh`, just to give it more of a clear intent, especially outside of the gitignore file
 - Added entry for `docker-compose.override.yml`, an officially supported, implicit extension of `docker-compose.yml` ([comment about multiple docker-compose files](https://github.com/PrairieLearn/PrairieLearn/pull/3214#issuecomment-765175831))


mebird:
If we're officially using the docker compose override, it may be worth updating the readme as well to reflect this!

mwest1066:
> If we're officially using the docker compose override, it may be worth updating the readme as well to reflect this!

Yes please!

mwest1066:
> > If we're officially using the docker compose override, it may be worth updating the readme as well to reflect this!
> 
> Yes please!

@astrosticks could you updated the docs to clarify the override filenames? Or is this in #3715?

mwest1066 (approved):

PR #3707, opened 2021-01-22
Manual Grading MVP

mebird:
Creates a manual grading MVP, that creates a grading menu overlay where an instructor can force upgrade the grade of a student.


pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.js
```diff
@@ -0,0 +1,27 @@
...
+    sqlDb.queryZeroOrOneRow(sql.get_next_unmarked_instance_question, params, (err, result) => {
+        if (ERR(err, next)) return;
+
+        // If we have no more submissions, then redirect back to manual grading page
+        if (!result.rows[0]) {
+            res.redirect(`${res.locals.urlPrefix}/assessment/${res.locals.assessment.id}/manual_grading?done`);
+            return;
+        }
+        const instance_question_id = result.rows[0].id;
+        res.redirect(res.locals.urlPrefix + '/instance_question/' + instance_question_id + '/manual_grading');
```
jonatanschroeder, line 21: A bit of a pet peeve here, but you use template literals in line 17 and string addition in line 21. Also, this is probably a case where an else would make it clearer.

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.sql
```diff
@@ -0,0 +1,14 @@
...
+    FROM instance_questions AS iq
+    JOIN variants AS v ON (v.instance_question_id = iq.id)
+    JOIN (
+        -- We only want the last submission...
+        SELECT * FROM submissions AS s
+        ORDER BY s.date DESC, s.id DESC
+        LIMIT 1
+    ) s ON (s.variant_id = v.id)
+    -- ... but ONLY IF the last submission 'graded_at' is null (has not been graded)
+WHERE iq.assessment_question_id = $assessment_question_id AND s.graded_at IS NULL
```
jonatanschroeder, line 12: probably wise to also include a filter for assessment_id, since that's what's being authorized here (otherwise I could use a course instance and assessment ID I have access to to bypass authz, and use the assessment question I want).

jonatanschroeder: Also (again, code style), have each condition on a separate line.

pages/partials/instructorGradingPanel.ejs
```diff
@@ -0,0 +1,37 @@
...
+    <div class="card mb-4 border-success">
+      <div class="card-header bg-success">Grading Panel</div>
+      <div class="card-body">
+        <% if (typeof submission_updated !== 'undefined' && submission_updated === true) { %>
+            <blockquote style="color: red">Submission updated</blockquote>
+        <% } %>
+        <form name="instance_question-manual-grade-update-form" method="POST">
+            <input type="hidden" name="__action" value="add_manual_grade">
+            <input type="hidden" name="__csrf_token" value="<%= __csrf_token %>">
+            <input type="hidden" name="assessment_instance" value="<%= assessment_instance.assessment_id %>">
```
jonatanschroeder, line 11: Is this the assessment_instance_id (in which case the value is wrong) or the assessment_id (in which case the name is wrong)?

pages/partials/navTabs/instructorAssessment.ejs
```diff
@@ -28,6 +28,13 @@
     urlSuffix:      '/assessment/'+assessment.id+'/question_statistics',
     iconClasses:    'fas fa-table',
     tabLabel:       'Question stats'}) %>
+<% if (course.manual_grading_visible) { %>
+    <%- include('navTab', {
+        activeSubPage:  'manual_grading',
+        urlSuffix:      '/assessment/'+assessment.id+'/manual_grading',
+        iconClasses:    'fas fa-exclamation',
```
jonatanschroeder, line 35: The exclamation looks weird. Maybe `fa-pen-alt` or `fa-marker`?

server.js
```diff
@@ -706,6 +710,16 @@ module.exports.initExpress = function() {
         require('./pages/studentAssessmentInstanceHomework/studentAssessmentInstanceHomework'),
         require('./pages/studentAssessmentInstanceExam/studentAssessmentInstanceExam'),
     ]);
+    app.use('/pl/course_instance/:course_instance_id/instructor/assessment/:assessment_id/assessment_question/:assessment_question_id/manual_grading', [
```
jonatanschroeder, line 713: Again pet peeve: this uses "manual_grading" in the same way as the manual_grading page above, which from the URL sounds like this will list all manual grading questions. Suggestion: "next_ungraded".

sprocs/instance_question_select_manual_grading_objects.sql
```diff
@@ -0,0 +1,45 @@
...
+    -- If variant is found but no submission, student did not submit anything
+    SELECT to_jsonb(s.*)
+    INTO submission
+    FROM
+        instance_questions AS iq
+        JOIN variants AS v ON (v.instance_question_id = iq.id)
+        JOIN submissions AS s ON (s.variant_id = v.id)
+    WHERE iq.id = iq_id
+    ORDER BY s.date DESC, s.id DESC
+    LIMIT 1;
```
jonatanschroeder, line 42: I believe this query could return a submission that doesn't match a variant for questions with multiple variants (e.g., if the last variant doesn't have submissions, but the first one has). You may want to review this to make a single select statement instead.

andrewstec:
Doing some QA work now. Just noticed that the input for the feedback expects a number that increments :)  

image

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.ejs
```diff
@@ -0,0 +1,167 @@
...
+      <!-- ###################################################################### -->
+
+      <div class="card mb-4">
+        <div class="card-header bg-primary text-white">
+          <%= assessment_set.name %> <%= assessment.number %>: Manual Grading Queue
+        </div>
+
+        <div class="card-body">
+          <p class="small">
+            A list of ungraded submissions for questions. Note that submissions that students have saved, but not submitted, may be listed here.
```
jonatanschroeder, line 29: I think this note would be better in the card-footer (see the Access page). Also,
```suggestion
            This list may include answers that students saved but did not request to be graded.
```

pages/partials/instructorGradingPanel.ejs
```diff
@@ -0,0 +1,37 @@
...
+                <input type="text" class="form-control" id="submissionScore" name="submission_score" value="<%= score_perc %>" aria-describedby="scoreHelpBlock"/>  
+                <span class="input-group-addon">%</span>
+              </div>
+              <small id="scoreHelpBlock" class="form-text text-muted">
+                This percentage will override the most recent submission score.
+              </small>
+            </div>
+            <div class="form-group">
+              <label for="submission-feedback">Feedback:</label>
+              <input type="number" name="submission_note" class="form-control" id="submission-feedback" value="<% if (submission.feedback) {%><%= submission.feedback.manual %><% } %>">
```
jonatanschroeder, line 25:
```suggestion
              <input type="text" name="submission_note" class="form-control" id="submission-feedback" value="<% if (submission.feedback) {%><%= submission.feedback.manual %><% } %>">
```

Alternatively, a small `textarea` may also be an idea...

pages/partials/instructorGradingPanel.ejs
```diff
@@ -0,0 +1,37 @@
...
+        <% } %>
+        <form name="instance_question-manual-grade-update-form" method="POST">
+            <input type="hidden" name="__action" value="add_manual_grade">
+            <input type="hidden" name="__csrf_token" value="<%= __csrf_token %>">
+            <input type="hidden" name="assessment_instance" value="<%= assessment_instance.assessment_id %>">
+            <input type="hidden" name="assessment_question_id" value="<%= instance_question.assessment_question_id %>">
+            <div class="form-group">
+              <label for="submissionScore">Score:</label>
+              <div class="input-group">
+                <input type="text" class="form-control" id="submissionScore" name="submission_score" value="<%= score_perc %>" aria-describedby="scoreHelpBlock"/>
```
jonatanschroeder, line 16:
```suggestion
                <input type="number" class="form-control" id="submissionScore" name="submission_score" value="<%= score_perc %>" min="0" max="100" step="5" aria-describedby="scoreHelpBlock"/>  
```

I should point out that max is just where the slider stops if you keep pushing up. You can still allow values over 100.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.js
```diff
@@ -0,0 +1,104 @@
...
+                submission.true_answer,
+            ];
+
+            sqlDb.callOneRow('grading_jobs_insert', params, (err, result) => {
+                if (ERR(err, next)) return;
+
+                /* If the submission was marked invalid during grading the grading job will
+                   be marked ungradable and we should bail here to prevent LTI updates. */
+                res.locals['grading_job'] = result.rows[0];
+                if (!res.locals['grading_job'].gradable) return next(error.make('Invalid submission error'));
```
jonatanschroeder, line 89:
```suggestion
                if (!res.locals['grading_job'].gradable) return next(error.make(400, 'Invalid submission error'));
```

jonatanschroeder: Actually I'm getting this error when I'm not supposed to...

jonatanschroeder (commented):
This is what I found so far, I can check more later.

sprocs/instance_question_select_manual_grading_objects.sql
andrewstec: Just want to unpack this a bit further @jonatanschroeder. I may be missing something here. Is there a scenario where we would want to manually grade the last variant without a submission if a variant prior to it has a submission? 

If there is no submission, the student did not submit an answer (ie. a student re-opens the instance question, where a new variant is produced, but a student decides not to answer again and sticks with the last submission as their final answer).

jonatanschroeder: I'm not saying that you should grade a non-submission, I'm saying that, in the situation where a new variant is created without a submission, this code would return a variant that doesn't match the submission. So for example, if the student receives variants 1, 2 and 3, submits an answer to variants 1 and 2, but not to 3, it would return the submission of 2, but the variant of 3.

Again, I'm not arguing about what needs to be graded, I'm saying that this code isn't returning elements that match each other. This is why I suggested a single select statement instead, which would both make it faster and returning matching values. For example, you could do (I haven't tested this, and the syntax may be a bit off, but it's just to illustrate):

```suggestion
    SELECT to_jsonb(q.*), to_jsonb(v.*), to_jsonb(s.*)
    INTO question, variant, submission
    FROM
        instance_questions AS iq
        JOIN assessment_questions AS aq ON (aq.id = iq.assessment_question_id)
        JOIN questions AS q ON (q.id = aq.question_id)
        LEFT JOIN variants AS v ON (v.instance_question_id = iq.id)
        LEFT JOIN submissions AS s ON (s.variant_id = v.id)
    WHERE iq.id = iq_id
    ORDER BY s.date DESC, s.id DESC
    LIMIT 1;
```

andrewstec: Just explored this issue a little bit further. 

You probably already know this, but there is a difference between when variants are produced based on 'save' and 'save & grade' actions. The former action, which we are now concentrating on to support the manual grading feature, will not create a new variant after a submission. The latter will. 

I suspect after something is manually graded, a new variant will be created. Just confirming that now.

andrewstec: I'm reading what you just posted now.

andrewstec: Ah I see. I assumed that was invalid SQL syntax, but Postgresql certainly allows it. I'm much happier with that approach.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.js
andrewstec: I've changed a query that might make a minor difference here. I am just doing some more QA work to see if I can break stuff. I should hopefully find the reason why you are getting an error when you are not supposed to.

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.js
andrewstec: Agreed. Done.

server.js
andrewstec: Agreed. Done.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.js
andrewstec: I ran into some other minor issues, but I did not find this issue that you were having. It is safe to say that a 400 error code is permissible here regardless.

jonatanschroeder: The 400 is required here anyway, that's the first parameter of the make function. As for why I'm getting the error, I'll test it later, maybe it's related to other issues you're fixing?

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.sql
andrewstec: This is quite smart and I like this idea. Done.

pages/partials/instructorGradingPanel.ejs
mebird: Dropped the `_id` suffix. Thanks for the catch.

pages/partials/navTabs/instructorAssessment.ejs
mebird: Sounds good! Wasn't completely sure which icon to go with since we have so many different "pen on paper" icons yet, so was treating the exclamation as a placeholder. I've updated to `fa-marker`.

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.ejs
mebird: My concern with this living in the footer is that there may be a large number of student questions, so the notice may be buried.

However, I'll move it into the footer since it seems like all of the other instructor pages like to put things in the footer.

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.sql
```diff
@@ -0,0 +1,25 @@
...
+FROM
+    instance_questions AS iq
+    JOIN assessment_instances AS ai ON (ai.id = iq.assessment_instance_id)
+    JOIN assessment_questions AS aq ON (aq.id = iq.assessment_question_id)
+    JOIN questions AS q ON (q.id = aq.question_id)
+    JOIN question_order(ai.id) AS qo ON (qo.instance_question_id = iq.id)
+    JOIN variants AS v ON (v.instance_question_id = iq.id)
+    JOIN (
+        -- We only want the last submission...
+        SELECT * FROM submissions AS s
```
mebird, line 18: When we have a question that's been graded, and then there's a new variant generated and saved, this seems to have some issues. The score displayed belongs to the graded question, but the grading page brings you to the saved submission.

pages/partials/instructorGradingPanel.ejs
mebird: Textarea looks good, and is responsive in the box.

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.sql
andrewstec: Ah, if I understand correctly, do you want the grading page to bring you to the graded submission instead? 

Just looking at this briefly, do we perhaps want the last graded_at submission? AKA. The last submission that graded_at != null?

andrewstec:
Things to note: @mwest1066 

Hopefully this helps you understand some of the directions taken here and where to go next.

- The manually grading logic looks for the last submission by a student that has not been graded by an internal or external grader.
- Whether the question is configured for external, internal, or manual, we use the `grading_jobs_insert_internal` sproc to force internal score calculations. 
- However, we have to configure the question for Manual type to get the correct view to render for the student, as the "Save and Grade" button cannot exist on the student view. We need the student to save a submission instead of saving and grading it for us to identify ungraded submissions, as we are looking for `graded_at` = null. 
- One must disable maxPoints as the score accumulations do not make sense in this context.

jonatanschroeder (approved):
I submitted a new commit with a few fixes, and also listed the question title. To be considered for a next PR: I'm not sure if using the student's question order is the best option, since from the instructor's point of view what matters is the instructor's order, but I won't block submission based on that. Otherwise, great job!

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.js
```diff
@@ -0,0 +1,30 @@
...
+const express = require('express');
+const router = express.Router();
+const path = require('path');
+const debug = require('debug')('prairielearn:' + path.basename(__filename, '.js'));
+const {sqlDb, sqlLoader} = require('@prairielearn/prairielib');
+const sql = sqlLoader.loadSqlEquiv(__filename);
+
+router.get('/', (req, res, next) => {
+    const params = {
+        assessment_question_id: res.locals.assessment_question_id,
```
mwest1066, line 11: Add a comment here that `assessment_question_id` is untrusted (has not been verified by an authz middleware) and so will need to be checked in the SQL. The `assessment_id` _has_ been verified by and authz middleware, so it can be trusted.

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.sql
```diff
@@ -0,0 +1,18 @@
...
+        JOIN (
+            -- We only want the LAST submissions that are created by each user to join on all
+            -- instance questions under assessment question
+            SELECT DISTINCT ON (s.auth_user_id) * FROM submissions AS s
+            WHERE s.graded_at IS NULL
+            ORDER BY s.auth_user_id, s.date DESC, s.id DESC
+        ) s ON (s.variant_id = v.id)
+WHERE 
+    iq.assessment_question_id = $assessment_question_id
+    AND a.id = $assessment_id
```
mwest1066, line 17: Add a comment that the `assessment_question_id` is untrusted, but `assessment_id` is trusted.

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.sql
```diff
@@ -0,0 +1,18 @@
+-- BLOCK get_next_unmarked_instance_question
+SELECT iq.*
+FROM 
+    instance_questions AS iq
+        JOIN assessment_questions AS aq ON (aq.id = iq.assessment_question_id)
```
mwest1066, line 5: Nit: code style is to only indent these JOINs to a total of 4 spaces.

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.sql
```diff
@@ -0,0 +1,18 @@
...
+        JOIN assessments AS a ON (a.id = aq.assessment_id)
+        JOIN variants AS v ON (v.instance_question_id = iq.id)
+        JOIN (
+            -- We only want the LAST submissions that are created by each user to join on all
+            -- instance questions under assessment question
+            SELECT DISTINCT ON (s.auth_user_id) * FROM submissions AS s
```
mwest1066, line 11: Nit: code style: FROM on new line.

mwest1066, line 11: This query won't work for group work, where multiple users might be submitting. In general, we should use `auth_user_id` for anything other than auditing. In this case, we want the most recent submission per instance question, not per `auth_user_id`.

I would actually prefer to use the instance question status to select them and ignore submissions entirely.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.js
```diff
@@ -0,0 +1,104 @@
...
+    });
+
+    debug('GET /');
+});
+
+router.post('/', function(req, res, next) {
+    if (req.body.__action == 'add_manual_grade') {
+        const note = req.body.submission_note;
+        const score = req.body.submission_score;
+        const params = [res.locals.instance_question.id];
```
mwest1066, line 58: I think we should also record the `submission_id` that is being graded (the most recent one) and use this here. This will avoid race conditions where an instructor starts grading a question with one submission, but then a student submits a new answer.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.js
```diff
@@ -0,0 +1,104 @@
...
+                submission.partial_scores,
+                score / 100, // overwrite submission score
+                submission.v2_score,
+                {manual:note}, // overwrite feedback
+                submission.submitted_answer,
+                submission.params,
+                submission.true_answer,
+            ];
+
+            sqlDb.callOneRow('grading_jobs_insert_internal', params, (err, result) => {
```
mwest1066, line 83: I think we should make a new `grading_jobs_insert_manual` for use here.

This should be similar to the existing `grading_jobs_insert_external_manual` and `grading_jobs_insert_internal`.

We should rename `grading_jobs_insert_external_manual` to just `grading_jobs_insert_external`.

sprocs/grading_jobs_insert_internal.sql
```diff
@@ -44,10 +44,6 @@ BEGIN
 
     IF NOT FOUND THEN RAISE EXCEPTION 'no such submission_id: %', submission_id; END IF;
 
-    IF grading_method != 'Internal' THEN
-        RAISE EXCEPTION 'grading_method is not Internal for submission_id: %', submission_id;
-    END IF;
-
```
mwest1066, line 50: We should undo this deletion once we've added the new `sprocs/grading_jobs_insert_manual`

sprocs/instance_question_select_manual_grading_objects.sql
```diff
@@ -0,0 +1,29 @@
+-- BLOCK instance_question_select_last_variant_with_submission
```
mwest1066, line 1: No need for BLOCK in sprocs.
```suggestion
```

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.js
```diff
@@ -0,0 +1,30 @@
...
+        // If we have no more submissions, then redirect back to manual grading page
+        if (!result.rows[0]) {
+            res.redirect(`${res.locals.urlPrefix}/assessment/${res.locals.assessment.id}/manual_grading?done`);
+            return;
+        }
+        const instance_question_id = result.rows[0].id;
+        res.redirect(`${res.locals.urlPrefix}/instance_question/${instance_question_id}/manual_grading`);
+    });
+
+    debug('GET /');
```
mwest1066, line 27: If we are going to have this `debug` line at all, it should be the first line in the function.

pages/partials/instructorGradingPanel.ejs
```diff
@@ -0,0 +1,37 @@
...
+        <% if (typeof submission_updated !== 'undefined' && submission_updated === true) { %>
+            <blockquote style="color: red">Submission updated</blockquote>
+        <% } %>
+        <form name="instance_question-manual-grade-update-form" method="POST">
+            <input type="hidden" name="__action" value="add_manual_grade">
+            <input type="hidden" name="__csrf_token" value="<%= __csrf_token %>">
+            <input type="hidden" name="assessment_id" value="<%= assessment_instance.assessment_id %>">
+            <input type="hidden" name="assessment_question_id" value="<%= instance_question.assessment_question_id %>">
+            <div class="form-group">
+              <label for="submissionScore">Score:</label>
```
mwest1066, line 14: Prefer kebab-case for HTML/CSS variables.
```suggestion
              <label for="submission-score">Score:</label>
```

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.js
```diff
@@ -0,0 +1,30 @@
...
+const {sqlDb, sqlLoader} = require('@prairielearn/prairielib');
+const sql = sqlLoader.loadSqlEquiv(__filename);
+
+router.get('/', (req, res, next) => {
+    const params = {
+        assessment_question_id: res.locals.assessment_question_id,
+        assessment_id: res.locals.assessment.id,
+    };
+    // Unmarked instance question df. Is last created submission of instance question AND has null graded_at value.
+    sqlDb.queryZeroOrOneRow(sql.get_next_unmarked_instance_question, params, (err, result) => {
```
mwest1066, line 15: Terminology: let's try and standardize on `graded` rather than `marked` (this is partially a US-vs-UK-English thing).
```suggestion
    sqlDb.queryZeroOrOneRow(sql.get_next_ungraded_instance_question, params, (err, result) => {
```

mwest1066:
I'm going to merge/deploy this for CA now as is, so that @jonatanschroeder can start using it immediately, but it would be great to do a second PR to address the comments from @jonatanschroeder and me. This is looking awesome! I can't wait to see it in use.

PR #3710, opened 2021-01-25
Fix auto-enrollment while opening unrelated queries

### LINKED ISSUE #3709: Auto-enrollment into all available instances of a course when viewing the course
After merging #3672 we have a bug as follows:

1. Make a course with two course instances, both of which are currently open for enrollments by students (e.g., with empty access rules).
2. As a student, use the "Add/remove courses" button to enroll in one of the course instances but not the other.
3. Return to the PL homepage. It will show only one course instance enrollment.
4. Click on that course instance to access it. Accessing the course instance URL causes enrollments to be generated for the other course instance from the same course.
5. Returning to the PL homepage will show that we are now enrolled in both course instances of the course.

@jonatanschroeder any ideas on what's causing this?

jonatanschroeder:
Fixes #3709

### DIFF
--- a/middlewares/authzCourseInstance.sql
+++ b/middlewares/authzCourseInstance.sql
@@ -11,7 +11,7 @@ SELECT
 FROM
     course_instances AS ci
     JOIN pl_courses AS c ON (c.id = ci.course_id)
-    JOIN LATERAL authz_course_instance($authn_user_id, ci.id, $is_administrator, $req_date) AS permissions_course_instance ON TRUE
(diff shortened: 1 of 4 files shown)


jonatanschroeder:
Pending: the API access also uses the same middleware. Should the auto-enrollment apply in this case? I imagine not, in that case I'll need to review that. Accessing the API URL will enroll them, but will not allow them to proceed because they also need instructor access.

mwest1066 (approved):

PR #3715, opened 2021-01-25
docker-compose: Separate commands for build / interactive / init

astrosticks:
Instead of using one docker-compose service, why not multiple?

All commands rely on the image `prairielearn/prairielearn:local`
- `docker-compose build` to build `prairielearn/prairielearn:local`
- `docker-compose run init` to start the server 
- `docker-compose run bash` to start an interactive `bash` session

Notes:
- Building was moved out since you likely won't want to build the image every single time you start the dev server
- `bash` will start you in `/PrairieLearn`, so you no longer have to manually cd into it every time!

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,17 +1,32 @@
 version: '3.8'
+
+x-options:
+  &dev-options
+  image: prairielearn/prairielearn:local
+  ports:
+    - 3000:3000
+  volumes:
+    - ./testCourse:/course
+    - /var/run/docker.sock:/var/run/docker.sock
+    - ${HOME}/pl_ag_jobs:/jobs
+    - .:/PrairieLearn
+  container_name: pl
+  environment:
+    - HOST_JOBS_DIR=${HOME}/pl_ag_jobs
+    - NODEMON=true
+
 services:
-  pl:
+
+  build-dev:
     build:
       context: .
     image: prairielearn/prairielearn:local
-    ports:
-      - 3000:3000
-    volumes:
-      - ./testCourse:/course
-      - /var/run/docker.sock:/var/run/docker.sock
-      - ${HOME}/pl_ag_jobs:/jobs
-      - .:/PrairieLearn
-    container_name: pl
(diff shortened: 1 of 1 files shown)


astrosticks:
If anyone can think of more understandable names for these commands, please suggest them!

coatless:
> `docker-compose run it`: to start an interactive terminal

How about `sh` instead of `it`? 

```
docker-compose run sh
```

astrosticks:
@coatless `sh` might be misleading... the command puts you into a `bash` session

coatless:
@astrosticks true. In that case, why not? 

```
docker-compose run bash
```

astrosticks:
yeah... lol. good move

mwest1066:
Does this PR also need some corresponding updates to the docs?

mebird:
> * `docker-compose build` to build `prairielearn/prairielearn:local`

> Notes:
> 
> * Building was moved out since you likely won't want to build the image every single time you start the dev server
> * `bash` will start you in `/PrairieLearn`, so you no longer have to manually cd into it every time!

By default, `docker-compose` will only build the image the first time, and then won't rebuild until you run `docker-compose build` or `docker-compose --build ...`. It's a bit misleading to "move it out into its own service".



> * `docker-compose build` to build `prairielearn/prairielearn:local`
> * `docker-compose run init` to start the server
> * `docker-compose run bash` to start an interactive `bash` session
Using services called "init" and "bash" may be misleading, especially since they aren't different "services". Running `docker-compose up` will bring up both init and bash... Which will fail since `x-options` causes both containers to use the same name. Further, you'll have two PL instances up. 

If PL was ever refactored to be multiple services (i.e. moving the DB to its own docker container, simulating an AWS service using the container, etc.) using docker-compose to run scripts doesn't work as well.

astrosticks:
Closing this for now because I realized that docker-compose isn't really built to do the super fine-tuned things I wanted to do with it. Feel free to take any bits from my changes that you want to.

PR #3720, opened 2021-01-27
Revisions to manual grading submission views

andrewstec:
Changes requested:

- [ ] The instance questions displayed on the `instructorAssessmentManualGrading.sql` view are not grouped, which makes the information hard to digest for the user. List needs to be grouped by instance question.
- [ ] Timer lock that stops grading by multiple actors if they incidentally select the same item for grading.

### DIFF
--- a/pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.sql
+++ b/pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.sql
@@ -1,28 +1,4 @@
 -- BLOCK select_questions_manual_grading
-WITH issue_count AS (
-    SELECT
-        q.id AS question_id,
-        count(*) AS open_issue_count
-    FROM
-        issues AS i
-        JOIN questions AS q ON (q.id = i.question_id)
-    WHERE
-        i.assessment_id = $assessment_id
-        AND i.course_caused
-        AND i.open
-    GROUP BY q.id
-),
-question_scores AS (
-    SELECT
-        aq.question_id,
-        avg(aq.mean_question_score) AS question_score
-    FROM
-        assessment_questions AS aq
-    WHERE
-        aq.assessment_id = $assessment_id
-    GROUP BY
-        aq.question_id
-)
 SELECT
     aq.*,
     q.qid,
@@ -30,19 +6,7 @@ SELECT
     q.id AS question_id,
     admin_assessment_question_number(aq.id) as number,
     ag.number AS alternative_group_number,
-    ag.number_choose AS alternative_group_number_choose,
     (count(*) OVER (PARTITION BY ag.number)) AS alternative_group_size,
-    z.title AS zone_title,
-    z.number AS zone_number,
-    z.number_choose as zone_number_choose,
-    (lag(z.id) OVER (PARTITION BY z.id ORDER BY aq.number) IS NULL) AS start_new_zone,
-    (lag(ag.id) OVER (PARTITION BY ag.id ORDER BY aq.number) IS NULL) AS start_new_alternative_group,
-    coalesce(ic.open_issue_count, 0) AS open_issue_count,
-    question_scores.question_score AS avg_question_score_perc,
-    z.max_points AS zone_max_points,
-    (z.max_points IS NOT NULL) AS zone_has_max_points,
-    z.best_questions AS zone_best_questions,
-    (z.best_questions IS NOT NULL) AS zone_has_best_questions,
     (SELECT COUNT(DISTINCT iq.id)
      FROM
          instance_questions AS iq
@@ -62,12 +26,8 @@ FROM
     assessment_questions AS aq
     JOIN questions AS q ON (q.id = aq.question_id)
     JOIN alternative_groups AS ag ON (ag.id = aq.alternative_group_id)
-    JOIN zones AS z ON (z.id = ag.zone_id)
-    JOIN assessments AS a ON (a.id = aq.assessment_id)
-    LEFT JOIN issue_count AS ic ON (ic.question_id = q.id)
-    LEFT JOIN question_scores ON (question_scores.question_id = q.id)
 WHERE
-    a.id = $assessment_id
+    aq.assessment_id = $assessment_id
     AND aq.deleted_at IS NULL
     AND q.deleted_at IS NULL
-ORDER BY z.number, z.id, aq.number;
+ORDER BY aq.number;
--- a/pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.sql
+++ b/pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.sql
@@ -1,22 +1,73 @@
--- BLOCK select_instance_questions_manual_grading
-SELECT DISTINCT ON (iq.id)
-    iq.*,
-    s.graded_at,
+-- BLOCK select_questions_manual_grading
+WITH issue_count AS (
+    SELECT
+        q.id AS question_id,
+        count(*) AS open_issue_count
+    FROM
+        issues AS i
+        JOIN questions AS q ON (q.id = i.question_id)
+    WHERE
+        i.assessment_id = $assessment_id
+        AND i.course_caused
+        AND i.open
+    GROUP BY q.id
+),
+question_scores AS (
+    SELECT
+        aq.question_id,
+        avg(aq.mean_question_score) AS question_score
+    FROM
+        assessment_questions AS aq
+    WHERE
+        aq.assessment_id = $assessment_id
+    GROUP BY
+        aq.question_id
+)
+SELECT
+    aq.*,
+    q.qid,
+    q.title,
     q.id AS question_id,
-    q.title AS question_title,
-    aq.max_points,
-    ai.id AS assessment_instance_id,
-    qo.question_number
+    admin_assessment_question_number(aq.id) as number,
+    ag.number AS alternative_group_number,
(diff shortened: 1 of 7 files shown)


cheeren:
i'd like to have MG as an option on the issues page. maybe it's just a different interface on the grade-change, but it seems more natural, somehow.

jonatanschroeder:
> * The instance questions displayed on the `instructorAssessmentManualGrading.sql` view are not grouped, which makes the information hard to digest for the user. List needs to be grouped by instance question.

To clarify, it needs to be grouped by assessment question, not instance question. To extend on this: I think it makes sense for this page to only list questions themselves, not individual instances. At this point we have 4500 submissions, which makes the page hard to read, slow to load, and is not very useful anyway. The list would then have, for each assessment question:
* The question's QID, title and number (instructor number, not student number, e.g., I-A1.1);
* Some basic stats (e.g., number of graded submissions, number of ungraded submissions, maybe min/max/avg score or points);
* A link that opens the list of submissions for that question only;
* A link to start grading, which would point to the redirect URL.

This page doesn't need to have a way to list or change individual grades. It doesn't matter to me that one particular student got 57% in the question if (1) I don't know who the student is and (2) I don't know what their answer is.

jonatanschroeder:
There needs to be a way to address race conditions when multiple TAs are handling the same question. I'm not sure what the best approach would be here, but it would probably involve tagging an instance question as "being graded" with an UPDATE that fails if the previous value is true (it would be ironic if this is not properly handled while marking CPSC 213, where we talk about spinlocks). There would also need to have a way to "expire" a taken tag or at least give a warning if the taken tag is on for a different marker.

andrewstec:
> There needs to be a way to address race conditions when multiple TAs are handling the same question. I'm not sure what the best approach would be here, but it would probably involve tagging an instance question as "being graded" with an UPDATE that fails if the previous value is true (it would be ironic if this is not properly handled while marking CPSC 213, where we talk about spinlocks). There would also need to have a way to "expire" a taken tag or at least give a warning if the taken tag is on for a different marker.

There are some CRON jobs running, so we can explore adding a CRON job to a taken item for grading that lasts longer than 30 minutes, etc.

andrewstec:
> > * The instance questions displayed on the `instructorAssessmentManualGrading.sql` view are not grouped, which makes the information hard to digest for the user. List needs to be grouped by instance question.
> 
> To clarify, it needs to be grouped by assessment question, not instance question. To extend on this: I think it makes sense for this page to only list questions themselves, not individual instances. At this point we have 4500 submissions, which makes the page hard to read, slow to load, and is not very useful anyway. The list would then have, for each assessment question:
> 
> * The question's QID, title and number (instructor number, not student number, e.g., I-A1.1);
> * Some basic stats (e.g., number of graded submissions, number of ungraded submissions, maybe min/max/avg score or points);
> * A link that opens the list of submissions for that question only;
> * A link to start grading, which would point to the redirect URL.
> 
> The page above would probably look similar to the Questions tab, only without Topics/tags/other assessments, and with links to list submissions and grade. It can probably borrow some elements from there, like the base query itself, the stats, and the grouping by zone.
> 
> This page doesn't need to have a way to list or change individual grades. It doesn't matter to me that one particular student got 57% in the question if (1) I don't know who the student is and (2) I don't know what their answer is.

I think this is smart, as it seems that we will go beyond supporting just Manual type questions. As we will likely display all types of questions, we will need to organize this information better to accommodate large numbers of instance questions.

jonatanschroeder:
> There are some CRON jobs running, so we can explore adding a CRON job to a taken item for grading that lasts longer than 30 minutes, etc.

I'd still be inclined to go with the warning that someone else may be grading it too.

Also, the current submit button automatically moves to the next graded submission, if a TA grades a few and needs to leave, there would need to be a way to mark that question as no longer "taken". Maybe a button that returns to the list of ungraded submissions? Or just a "Cancel"-like button? Also, the same trigger for the warning we give when students try to close a tab that has unsaved work could probably be applied here.

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.sql
```diff
@@ -1,20 +1,73 @@
...
+    z.number AS zone_number,
+    z.number_choose as zone_number_choose,
+    (lag(z.id) OVER (PARTITION BY z.id ORDER BY aq.number) IS NULL) AS start_new_zone,
+    (lag(ag.id) OVER (PARTITION BY ag.id ORDER BY aq.number) IS NULL) AS start_new_alternative_group,
+    coalesce(ic.open_issue_count, 0) AS open_issue_count,
+    question_scores.question_score AS avg_question_score_perc,
+    z.max_points AS zone_max_points,
+    (z.max_points IS NOT NULL) AS zone_has_max_points,
+    z.best_questions AS zone_best_questions,
+    (z.best_questions IS NOT NULL) AS zone_has_best_questions,
```
mwest1066, line 45: The code above looks like it was copied from `instructorAssessmentQuestions`? It has a lot of stuff in there to render zones, max points, best questions, etc. Do we want all this for the manual grading overview page? I tend to think that we should just show a simple list of questions with information about what still needs grading, with no extra extraneous info.

pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.sql
```diff
@@ -0,0 +1,29 @@
...
+    aq.max_points
+FROM
+    assessment_questions AS aq
+    JOIN questions AS q ON (q.id = aq.question_id)
+WHERE
+    aq.assessment_id = $assessment_id
+    AND aq.id = $assessment_question_id;
+
+
+-- BLOCK select_instance_questions_manual_grading
```
mwest1066, line 14: It looks to me like this is selecting all submissions for the given `assessment_question`, not all `instance_questions`. Is that what we want? If so, can we change the BLOCK name to reflect this?

But I'm a bit confused, because the JS and EJS files seem to be treating the returned data as a list of instance questions. What should it be?

pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.js
```diff
@@ -0,0 +1,98 @@
...
+        assessment_question_id: res.locals.assessment_question_id,
+    };
+    sqlDb.queryOneRow(sql.select_question, params, (err, result) => {
+        if (ERR(err, next)) return;
+        
+        // Instance question doesn't exist
+        if (result.rowCount == 0) {
+            return error.make(404, 'Assessment question not found.');
+        }
+
```
mwest1066, line 29: This check won't do anything because we are already requiring exactly one row to be returned by calling `queryOneRow`.
```suggestion
```

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.sql
jonatanschroeder: Removed zones and assessment groups, though I kept the assessment group numbering.

pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.sql
jonatanschroeder: It is the last submission of a each specific instance question (see the DISTINCT ON at the top). So in essence it's selecting all instance questions, together with the last submission of that instance question.

mebird:
> > * The instance questions displayed on the `instructorAssessmentManualGrading.sql` view are not grouped, which makes the information hard to digest for the user. List needs to be grouped by instance question.
> 
> To clarify, it needs to be grouped by assessment question, not instance question. To extend on this: I think it makes sense for this page to only list questions themselves, not individual instances. At this point we have 4500 submissions, which makes the page hard to read, slow to load, and is not very useful anyway. The list would then have, for each assessment question:
> 
> * The question's QID, title and number (instructor number, not student number, e.g., I-A1.1);
> * Some basic stats (e.g., number of graded submissions, number of ungraded submissions, maybe min/max/avg score or points);
> * A link that opens the list of submissions for that question only;
> * A link to start grading, which would point to the redirect URL.
> 
> The page above would probably look similar to the Questions tab, only without Topics/tags/other assessments, and with links to list submissions and grade. It can probably borrow some elements from there, like the base query itself, the stats, and the grouping by zone.
> 
> This page doesn't need to have a way to list or change individual grades. It doesn't matter to me that one particular student got 57% in the question if (1) I don't know who the student is and (2) I don't know what their answer is.

If we're bubbling this up to the question level rather than the instance, does it make sense to move this grading interface as an extra button in each row on the questions page? Or am I misled to believe that that would still serve the same purpose.

jonatanschroeder:
> If we're bubbling this up to the question level rather than the instance, does it make sense to move this grading interface as an extra button in each row on the questions page? Or am I misled to believe that that would still serve the same purpose.

I wouldn't be opposed to that, though I'd prefer a separate page until this is more widely available to avoid messing up with a page that is already in widespread use.

pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.sql
mwest1066: Oh right! Somehow I overlooked this. Sorry!

mwest1066:
> I wouldn't be opposed to that, though I'd prefer a separate page until this is more widely available to avoid messing up with a page that is already in widespread use.

+1 to this!

PR #3722, opened 2021-01-28
changed all @prairielearn requirements to direct requirements

### LINKED ISSUE #3624: Change PrairieLib requires to direct requires
Now that we've moved `PrairieLib` back into the monorepo, we should use direct requires. That is, everywhere that we do this: https://github.com/PrairieLearn/PrairieLearn/blob/89d265461da214d773b722b5e0465e00396f52b1/lib/question.js#L20 we should change it to:
```
const sqldb = require('../prairielib/sql-db');
```
and similarly for any require with `@prairielib`.

nitahieb:
Fixes #3624


mwest1066:
Thanks for the PR! It looks like there might have been some typos that are causing CI test failures.

PR #3728, opened 2021-01-30
Add EFS support to workspaces

nicknytko:
- [x] add config var
- [x] add db migrations
- [x] update prairielearn server
  - [x] update `initialize()` 
    - [x] check backing storage, upload to EFS/S3
  - [x] update `getGradedFiles()` to check backing storage type, pull from EFS/S3
- [x] update workspace host
  - [x] workspaces should mount their home directory directly on EFS
  - [x] code that periodically pushes to S3 should check workspace storage, EFS does not need to be pushed
  - [x] file watching code does not need to watch EFS files
    - we should get this for free if EFS files are in a different directory than S3 jobs?
  - [x] do not need to download files from S3, if they are on EFS
    - [x] `_getInitialZipAsync()`
    - [x] `_getInitialFiles()` 
- [x] test on staging


database/enums/enum_workspace_homedir_location.pg
```diff
@@ -0,0 +1 @@
+s3, efs
```
mwest1066, line 1: Rather than defining a new enum, perhaps we could use the existing `enum_file_storage_type`? It was created to solve a similar problem (but in the reverse FileSystem -> S3) direction. In any case I'd prefer to call the non-S3 version something that isn't specific to EFS (so maybe 'filesystem' or 'disk' or similar), because from PL's point of view these are just local files (which will happen to be on a mounted EFS share).

nicknytko: Ah good idea, should've known there was an existing enum for it.

lib/workspace.js
```diff
@@ -270,12 +284,55 @@ module.exports = {
...
+            } else if (workspace.homedir_location == 'FileSystem') {
+                zipPath = await module.exports.getGradedFilesFromFileSystem(workspace_id);
+            }
         }
 
         return zipPath;
     },
 
+    async getGradedFilesFromFileSystem(workspace_id) {
+        console.log('getting graded files from filesystem');
```
mwest1066, line 298: Debugging code.

lib/config.js
```diff
@@ -169,6 +169,9 @@ config.workspaceHostUnhealthyTimeoutSec = 12 * 60 * 60;
 config.workspaceHostLaunchTimeoutSec = 10 * 60;
 config.workspaceUrlRewriteCacheMaxAgeSec = 60 * 60;
 config.workspacePercentMessageRateLimitSec = 1;
+config.workspaceHomeDirRoot = '/jobs/workspaces'; /* Where the main server will store workspace files.  Note on production this should be different from the jobs directory, this is just to make local development easier. */
+config.workspaceHostHomeDirRoot = '/jobs/workspaces'; /* Where the workspace host server will store workspace files. */
+config.workspaceHostWatchJobFiles = false; /* Should be enabled if we're watching files to sync with S3 */
```
mwest1066, line 174: For backwards compatibility should we set this to `true` by default?

If I'm understanding correctly, when we first deploy this we'll be running with some workspaces backed by S3 and some by FileSystem (on the same workspace host). So we'll need to run the watcher, but if it detects a file change then we'll only take action if the file is for a workspace with S3 backing. After we are sure we no longer need workspaces with S3 backing (after the semester is over?) we'll set `workspaceHostWatchJobFiles = false`.

workspace_host/interface.js
```diff
@@ -776,19 +788,33 @@ function _recursiveDownloadJobManager(curDirPath, S3curDirPath, callback) {
...
+        localPath = `${workspacePrefix}/${localName}`;
+
+        const isDirectory = false;
+        update_queue[[zipPath, isDirectory]] = { action: 'skip' };
+        const options = {
+            owner: config.workspaceJobsDirectoryOwnerUid,
+            group: config.workspaceJobsDirectoryOwnerGid,
+        };
+        await awsHelper.downloadFromS3Async(config.workspaceS3Bucket, remotePath, zipPath, options);
+    } else if (workspace.homedir_location == 'FileSystem') {
```
mwest1066, line 807: I'm a bit confused here. If we have a workspace with FileSystem storage, why do we need to do any copying at all? Won't the files automatically be in the right place when we start up?

workspace_host/interface.js
```diff
@@ -684,7 +696,7 @@ async function _autoUpdateJobManager() {
         debug(`watch: workspace_id=${workspace_id}, localPath=${local_path}`);
         const workspace = await _getWorkspaceAsync(workspace_id);
         const workspaceSettings = await _getWorkspaceSettingsAsync(workspace_id);
-        const s3_name = workspace.s3_name;
+        const remote_name = workspace.remote_name;
```
mwest1066, line 699: In `_autoUpdateManager()`, don't we only want to take action for files for workspaces with S3 storage? I couldn't see where we were skipping files associated with FileSystem workspaces.

lib/config.js
nicknytko: Yes agreed!  We should set this to be `true` when we're running on production servers.  I have this set to `false` because I figured it's unnecessary when running locally on docker.

workspace_host/interface.js
nicknytko: I was thinking since the EFS/FileSystem jobs and the old S3 jobs will be running in different directories, the file watcher won't be tracking the FileSystem jobs.  So these files shouldn't ever show up in the update queue/update manager.

workspace_host/interface.js
nicknytko: I'm also confused 🙂 

I'm not exactly sure when the `initial.zip` file is used (if at all), so I left this in there.  If it turns out that we don't actually need it I'm happy to remove this bit.

nicknytko:
notes:

```
      if (useInitialZip) {
          debug(`init: bootstrapping workspace with initial.zip`);
          await _getInitialZipAsync(workspace);
      } else {
          debug(`init: syncing workspace from S3`);
          await _getInitialFilesAsync(workspace);
      }
```

wrap this in an if()

echuber2:
Is this the issue that affected Phys 214? Do we have a related issue for that one? I was hoping to understand better what the underlying issue was here, since it seemed like the issue was sporadic.

nicknytko:
> Is this the issue that affected Phys 214? Do we have a related issue for that one? I was hoping to understand better what the underlying issue was here, since it seemed like the issue was sporadic.

Maybe?  We're planning to move to a different storage backend for the workspaces that should remove the need for file syncing and hopefully fix those issues as well.

echuber2:
If it will use a mount from now on, is there a need to pre-populate the home directory in a different way than before?

nicknytko:
> If it will use a mount from now on, is there a need to pre-populate the home directory in a different way than before?

Yep, instead of zipping up the files and shipping them off to S3 they will now just be copied directly into the mount: https://github.com/PrairieLearn/PrairieLearn/blob/49f1ce193668d7c0cddd2bfd93e5bd783623cac0/lib/workspace.js#L246.  And the workspace servers won't need to copy and extract anything now either.

PR #3729, opened 2021-01-30
Fix error in pl-matrix-component-input for non-numpy answer

jonatanschroeder:
The `pl-matrix-component-input` element allows a matrix of ints to be set as the correct answer, even if it is not a numpy array. However, if this is done, and there is a parse error in the submission (e.g., one of the elements is left blank), this will cause a set of issues to be raised, due to a render error in this element that assumes that correct answer is a numpy array. This PR fixes that issue by ensuring the array is converted to a numpy array if necessary.


nicknytko (commented):
looked good 🙂

mwest1066:
> looked good 🙂

😂

PR #1626, opened 2019-08-04
Containerize course code execution

nwalters512:
This PR introduces a new way to execute course/element code: said code can now be executed inside isolated Docker containers. This isolates the code from PrairieLearn and other courses. More details to be written up here later!

**TODO**
- [x] Documentation
- [x] Move image from https://github.com/nwalters512/PLQS to main PL repo

PR #3731, opened 2021-02-01
Use os._exit() to exit Python worker process

nwalters512:
While working on https://github.com/PrairieLearn/PrairieLearn/pull/1626, I noticed that Python processes took on the order of hundreds of milliseconds to exit and return control to the parent. When a Python process implicitly exits (or exits with `sys.exit()`), it doesn't actually exit immediately - it allows things to gracefully shut down and clean themselves up. While this is generally a good thing, in this case, we just want to get out of the process as fast as possible so we can re-use the worker. Switching to `os._exit()` allows us to do this.

For the metric of "delta between start of question generation to the instant that all the python callers used for that question are available again", this change reduces that metric from ~`1200ms` to ~`120ms`. This should allow us to dramatically increase response times under periods of high load - or, at least, allow us to use machines with fewer CPUs.

### DIFF
--- a/lib/python-caller-trampoline.py
+++ b/lib/python-caller-trampoline.py
@@ -68,7 +68,15 @@ def worker_loop():
                 outf.write(json_outp)
                 outf.write("\n");
                 outf.flush()
-                break
+                # `sys.exit()` allows the process to gracefully shut down. however, that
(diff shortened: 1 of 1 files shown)


nicknytko:
🤯

Issue #3708, opened 2021-01-24
pl-integer-input bound by JS integer value limits


jonatanschroeder:
It seems an input that is longer than what JS supports (15 decimal digits) is silently truncated by JS, even though both PostgreSQL and Python support longer numbers. This leads to a weird situation where, if the student provides an input larger than this limit in a `pl-integer-input`, the value shown in the submission box doesn't match what was submitted.

One possible approach (suggested by @mwest1066) would be to reject numbers (format error) that are too large for JavaScript to support. In essence this would mean reject any number larger than 2^53-1 or smaller than -(2^53-1) (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER). If the question actually needs a number that large, a custom `parse`/`grade` function would be needed, working with `data["raw_submitted_answers"]`.

Other alternatives would require either the use of `BigInt` in JS, or a way to provide different arguments to the element to handle very large numbers in a different way.

jonatanschroeder:
To consider: https://www.npmjs.com/package/json-bigint

eliotwrobson:
I think the only issue here is Json serialization of the integer, correct? If so, the integer can just be stored as a string and converted back to a Python integer by grader code.

PR #3781, opened 2021-02-01
Fixed pl-integer-input issue by dealing with possible scientific nota…

### LINKED ISSUE #3080: Very large inputs will break rendering of pl-integer-input
Reproduce by putting the value `112173586448650064316584391727166410732855297644839296413224534507665844338688` into an integer input.  (How students got this value, I have no idea...)

This happens because the entry in `submitted_values` gets truncated as `1.1217358644865006e+77` when converted to JSON, and then this is assumed to be an integer when rendered.

Stack trace:

```
  File "/PrairieLearn/lib/python-caller-trampoline.py", line 167, in <module>
    worker_loop()
  File "/PrairieLearn/lib/python-caller-trampoline.py", line 113, in worker_loop
    val = method(*args)
  File "/PrairieLearn/elements/pl-integer-input/pl-integer-input.py", line 115, in render
    html_params['a_sub'] = '{:d}'.format(a_sub)
ValueError: Unknown format code 'd' for object of type 'float'
  File "/PrairieLearn/lib/python-caller-trampoline.py", line 179, in <module>
```

nitahieb:
…tion
fixed #3080 by casting the input to a float then back into an int

### DIFF
--- a/workspaces/jupyterlab/Dockerfile
+++ b/workspaces/jupyterlab/Dockerfile
@@ -1,4 +1,4 @@
-FROM jupyter/datascience-notebook
+FROM jupyter/datascience-notebook:aec555e49be6
 COPY jupyter_notebook_config.py /etc/jupyter/
 
 ENV XDG_DATA_HOME=/tmp/local/share


jonatanschroeder:
This fix indeed solves the breaking part, but the value is still not reliable due to #3708. I'd personally prefer to solve both together, as there's no point in fixing the display if the display still shows the wrong submitted value.

mwest1066:
To also fix #3708 I think it would be a good idea to detect large integers in `parse()` and reject them with a `format_error` (thereby marking the submission as invalid). That is, we should get the submitted integer and if it's outside the range of `-(2^53 - 1)` and `2^53 - 1` then we reject it. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

If we reject out-of-range integers then we won't need the `float()` fix in this PR.

@nitahieb do you want to make this change?

jonatanschroeder:
Should this be done with the correct answer as well in `prepare`, to ensure an instructor gets an error right away if the correct answer expects this kind of value?

nitahieb:
Ya, I'll get on doing this!

mwest1066:
> Should this be done with the correct answer as well in `prepare`, to ensure an instructor gets an error right away if the correct answer expects this kind of value?

Excellent idea: throw an exception if the correct answer is out of range.

nitahieb:
should I create a new format error string or do I just throw a value error and let the default format error message pop up? Or possibly change the default to include the int size constraints?

mwest1066:
> should I create a new format error string or do I just throw a value error and let the default format error message pop up? Or possibly change the default to include the int size constraints?

Like this: https://github.com/PrairieLearn/PrairieLearn/blob/master/elements/pl-integer-input/pl-integer-input.py#L33

jonatanschroeder:
I think @nitahieb is referring to the `parse` function, in that case it would be something like this:

https://github.com/PrairieLearn/PrairieLearn/blob/9892ac51710a3ce17c144fbd5069e24e8bb3071b/elements/pl-integer-input/pl-integer-input.py#L202

nitahieb:
Thank you! I think I found a problem in the code the checks are only happening if the person defines a correct-answer in the HTML because it will default to CORRECT_ANSWER_DEFAULT (which is none), if you put an incompatible answer in the python then it won't be detected https://github.com/PrairieLearn/PrairieLearn/blob/9892ac51710a3ce17c144fbd5069e24e8bb3071b/elements/pl-integer-input/pl-integer-input.py#L35-L42

mwest1066:
I agree that we don't have a great way to detect out-of-range correct answers that are directly created in the python code. I'm happy to leave this as unchecked.

jonatanschroeder:
`prepare` has access to `data["correct_answers"]`, you can check that separately...

mwest1066:
> `prepare` has access to `data["correct_answers"]`, you can check that separately...

But by that point we've already round-tripped through JSON, so we might have lost information already, I think?

jonatanschroeder:
> But by that point we've already round-tripped through JSON, so we might have lost information already, I think?

We lost information about the exact value, but not about its magnitude. The value will still be represented as a very large number, only not the exact number provided by the server.py code. So it will still be outside the acceptable range.

mwest1066:
> > But by that point we've already round-tripped through JSON, so we might have lost information already, I think?
> 
> We lost information about the exact value, but not about its magnitude. The value will still be represented as a very large number, only not the exact number provided by the server.py code. So it will still be outside the acceptable range.

Excellent point! I don't know what happens if someone enters a number outside of the float64 range (> 2e308), but I'm not too worried about that case 😄

nitahieb:
I added the html checks to prepare, added the checks for the python file, and #3080. I'm going to send a pull request. 
Thank you for the help!

mwest1066:
> I added the html checks to prepare, added the checks for the python file, and #3080. I'm going to send a pull request.
> Thank you for the help!

Thanks!

PR #3799, opened 2021-02-04
Add points for current and future attempts in question footer

jonatanschroeder:
As discussed on Slack.

image

### DIFF
--- a/pages/partials/questionFooter.ejs
+++ b/pages/partials/questionFooter.ejs
@@ -45,6 +45,20 @@
               <i class="fa fa-question-circle" aria-hidden="true"></i>
           </a>
           <% } %>
+          <% if (question_context == 'student_exam' && instance_question.points_list) { %>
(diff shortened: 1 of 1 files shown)


mwest1066 (approved):

cheeren:
This feels a little verbose, to me, but I think it's just because there seems to be redundant information -- do we need to say "k attempts left" in addition to the more informative sentence on the right, or can we just simplify the button back to save & grade?

mwest1066:
> This feels a little verbose, to me, but I think it's just because there seems to be redundant information -- do we need to say "k attempts left" in addition to the more informative sentence on the right, or can we just simplify the button back to save & grade?

I think @jonatanschroeder's earlier problem is that (some) students don't read the extra info, and miss the fact that they have limited attempts (and especially when they are on their last attempt). The hope was that putting it right on the button would help with this 😀

PR #3803, opened 2021-02-05
added support for incorrect integer input

nitahieb:
solved #3708 by restricting student input to be in bounds and limiting the creation of questions with answers outside of the bounds. Also added checks for incorrect inputs in the python file


elements/pl-integer-input/pl-integer-input.py
```diff
@@ -37,9 +37,22 @@ def prepare(element_html, data):
         if name in data['correct_answers']:
             raise Exception('duplicate correct_answers variable name: %s' % name)
         # Test conversion, but leave as string so proper value is shown on answer panel
-        if pl.string_to_integer(correct_answer, base) is None:
+        a_tru_parsed = pl.string_to_integer(correct_answer, base)
+        if a_tru_parsed is None:
             raise Exception('correct answer is not a valid input: %s' % name)
+        if a_tru_parsed > 2 ** 53 - 1 or a_tru_parsed < -((2 ** 53) - 1):
```
mwest1066, line 43: Generally prefer no spaces around exponentiation operator:
```suggestion
        if a_tru_parsed > 2**53 - 1 or a_tru_parsed < -((2**53) - 1):
```

elements/pl-integer-input/pl-integer-input.py
```diff
@@ -37,9 +37,22 @@ def prepare(element_html, data):
...
+            raise Exception('answer must be between (2**53)-1 and -((2**53)-1)')
```
mwest1066, line 44: when saying "between A and B" I think it's clearer if we order them with A < B? And indicate which answer.
```suggestion
            raise Exception('correct answer must be between -(2**53)-1 and ((2**53)-1)')
```

mwest1066: For clarity I think we should use `^` for exponentiation and also write the actual values:
```suggestion
            raise Exception('correct answer must be between -9007199254740991 and +9007199254740991 (that is, between -(2^53 - 1) and +(2^53 - 1)).')
```

elements/pl-integer-input/pl-integer-input.py
```diff
@@ -37,9 +37,22 @@ def prepare(element_html, data):
...
         data['correct_answers'][name] = correct_answer
+        return
+
+    a_tru = pl.from_json(data['correct_answers'].get(name, None))
+    if a_tru is not None:
+        if isinstance(a_tru, str):
+            a_tru = pl.string_to_integer(a_tru, base)
+        if a_tru is None:
+            raise Exception('correct answer is not a valid input: %s' % name)
```
mwest1066, line 53: Should we do `a_tru = int(a_tru)` here to force a conversion? Maybe inside a `try` block? Later we will do this: https://github.com/PrairieLearn/PrairieLearn/blob/f55ea11694ce820c2290be2d6c6a80611c2a0468/elements/pl-integer-input/pl-integer-input.py#L255

elements/pl-integer-input/pl-integer-input.py
```diff
@@ -37,9 +37,22 @@ def prepare(element_html, data):
...
+        if a_tru > 2 ** 53 - 1 or a_tru < -((2 ** 53) - 1):
+            raise Exception('answer must be between (2**53)-1 and -((2**53)-1)')
```
jonatanschroeder, line 55: Same comments as the ones @mwest1066 listed above for the other check. Also, I believe the other test above could be replaced with this one only, since you set `data["correct_answers"][name]` in line 45 anyway, so this would be checked for both cases anyway.

elements/pl-integer-input/pl-integer-input.py
```diff
@@ -215,6 +228,8 @@ def parse(element_html, data):
         a_sub_parsed = pl.string_to_integer(str(a_sub), base)
         if a_sub_parsed is None:
             raise ValueError('invalid submitted answer (wrong type)')
+        if a_sub_parsed > 2 ** 53 - 1 or a_sub_parsed < -((2 ** 53) - 1):
+            data['format_errors'][name] = 'Answer must be between (2**53)-1 and -((2**53)-1)'
```
jonatanschroeder, line 232: This error will be shown to students, some who may not be familiar with Python's representation of exponentiation. I'd suggest showing the actual numbers here as well, in a similar fashion as Matt's suggestion above.

jonatanschroeder (commented):
Looks good, check the notes. @mwest1066 do you think this warrants an update in:
* documentation for `pl-integer-input` (`docs/elements.md`), and/or
* the helper text shown when the student presses the help button?

mwest1066:
@nitahieb did you mean to close this PR?

mwest1066:
> @mwest1066 do you think this warrants an update in:
> 
> * documentation for `pl-integer-input` (`docs/elements.md`), and/or
> * the helper text shown when the student presses the help button?

Yep, both of these would be great.

elements/pl-integer-input/pl-integer-input.py
jonatanschroeder: Marking unresolved because the first comment was handled, but not the second.

nitahieb:
> @nitahieb did you mean to close this PR?

No, I'm new to git I didn't know I could add commits to this pull request

mwest1066:
> > @nitahieb did you mean to close this PR?
> 
> No, I'm new to git I didn't know I could add commits to this pull request

Haha, no worries, github pull requests are pretty magic 😄

elements/pl-integer-input/pl-integer-input.py
nitahieb: line 45 wouldn't run if there's no html input wouldn't this need to be checked in case of the python input for example being a string or something?

nitahieb: oh but if add one of the previous suggestions (casting directly to an int with a try block) I don't need to check for none because if it's not an int it'll throw an error

elements/pl-integer-input/pl-integer-input.py
nitahieb: If i add a try block do I add the except block just to throw another exception but with a more specific error message?

elements/pl-integer-input/pl-integer-input.py
jonatanschroeder: Note that none is still a valid value, in that case we assume the question has a custom grader. What's not valid is a correct value that is not an integer in a valid range, or can't be converted to one.

nitahieb: oh so I shouldn't raise an exception if it's None? only when out of bounds or can't be converted

jonatanschroeder: Yes, though note you already do a check for None in line 49.

elements/pl-integer-input/pl-integer-input.py
mwest1066: That would be great.

nitahieb: https://github.com/nitahieb/PrairieLearn/blob/2aa9eaea9927ea2afbeda189ea70c4a9aedb64a9/elements/pl-integer-input/pl-integer-input.py#L48-L56

jonatanschroeder:
> Looks good, check the notes. @mwest1066 do you think this warrants an update in:
> 
> * documentation for `pl-integer-input` (`docs/elements.md`), and/or
> * the helper text shown when the student presses the help button?

@nitahieb just checking if you are game for changing the items above?

nitahieb:
> > Looks good, check the notes. @mwest1066 do you think this warrants an update in:
> > 
> > * documentation for `pl-integer-input` (`docs/elements.md`), and/or
> > * the helper text shown when the student presses the help button?
> 
> @nitahieb just checking if you are game for changing the items above?

Ya sounds good. When a student inputs  a number out of range shouldn't the why button explain the problem(out of range)? Or is there a different help button?

I'm not quite sure what to do in the .md file. I don't see any other input talking about it's bounds? Do I add a sections similar to "Specifying a non-trivial base"? Thank you!

jonatanschroeder:
> Ya sounds good. When a student inputs a number out of range shouldn't the why button explain the problem(out of range)? Or is there a different help button?

This is the help button that shows on the input, not on the submission: 
image
 
> I'm not quite sure what to do in the .md file. I don't see any other input talking about it's bounds? Do I add a sections similar to "Specifying a non-trivial base"? Thank you!

I think a new section should be fine. You can add a suggestion that larger values can be handled with a `pl-string-input` with custom grading.

elements/pl-integer-input/pl-integer-input.py
```diff
@@ -37,9 +37,23 @@ def prepare(element_html, data):
         if name in data['correct_answers']:
             raise Exception('duplicate correct_answers variable name: %s' % name)
         # Test conversion, but leave as string so proper value is shown on answer panel
-        if pl.string_to_integer(correct_answer, base) is None:
+        a_tru_parsed = pl.string_to_integer(correct_answer, base)
+        if a_tru_parsed is None:
             raise Exception('correct answer is not a valid input: %s' % name)
+        if a_tru_parsed > 2**53 - 1 or a_tru_parsed < -((2**53) - 1):
+            raise Exception('correct answer must be between -9007199254740991 and +9007199254740991 (that is, between -(2^53 - 1) and +(2^53 - 1)).')
```
jonatanschroeder, line 44: This is probably not needed since the same test is also done in lines 51-56.

elements/pl-integer-input/pl-integer-input.py
jonatanschroeder: The current (new) version only converts to int if the value was originally a string. I suggest removing the `if instanceof` line and convert to int without checking the original type. The exception would also catch a type that cannot be converted to an int (e.g., a list).

jonatanschroeder (approved):

elements/pl-integer-input/pl-integer-input.py
nitahieb: But the 51-56 tests don't run if you get the answer from the html

nitahieb:
https://github.com/PrairieLearn/PrairieLearn/pull/3803#discussion_r577012815
for this, if I don't add the check I'd need to cast into an int by using pl.string_to_integer() because the int() function will crash if you cast an int with a base, and I set it up this way due to Matt's comment > Should we do `a_tru = int(a_tru)` here to force a conversion? Maybe inside a `try` block? Later we will do this:
> 
> https://github.com/PrairieLearn/PrairieLearn/blob/f55ea11694ce820c2290be2d6c6a80611c2a0468/elements/pl-integer-input/pl-integer-input.py#L255

elements/pl-integer-input/pl-integer-input.py
jonatanschroeder: You update `correct_answers` in line 45, why wouldn't the tests run if it comes from HTML?

jonatanschroeder:
> for this, if I don't add the check I'd need to cast into an int by using pl.string_to_integer() because the int() function will crash if you cast an int with a base

This is already inside a try block, which will catch it and handle it if it's not a type that can be converted into an int. Also note that Matt's suggestion implies the use of int if the value is *not* an int (the code he quotes converts to int if it's not a string). In any case, note that pl.string_to_integer() basically calls int, but with exception handling.

elements/pl-integer-input/pl-integer-input.py
nitahieb: wouldn't the return on line 46 make it so it never reaches the other section?

nitahieb: oh, i could just remove the return and have a check if the answer has been loaded and if not load it?,

nitahieb:
> > for this, if I don't add the check I'd need to cast into an int by using pl.string_to_integer() because the int() function will crash if you cast an int with a base
> 
> This is already inside a try block, which will catch it and handle it if it's not a type that can be converted into an int. Also note that Matt's suggestion implies the use of int if the value is _not_ an int (the code he quotes converts to int if it's not a string). In any case, note that pl.string_to_integer() basically calls int, but with exception handling.

I dealt with this by trying to convert everything except for integers. and moved some of the checks around to reduce repitition

elements/pl-integer-input/pl-integer-input.py
```diff
@@ -37,23 +37,17 @@ def prepare(element_html, data):
...
-            raise Exception('correct answer must be between -9007199254740991 and +9007199254740991 (that is, between -(2^53 - 1) and +(2^53 - 1)).')
+    if correct_answer is None:
+        correct_answer = pl.from_json(data['correct_answers'].get(name, None))
+    try:
+        if not isinstance(correct_answer, int):
+            correct_answer = int(correct_answer, base)
+    except Exception:
+        raise Exception('correct answer is not a valid input: %s' % name)
+    if correct_answer > 2**53 - 1 or correct_answer < -((2**53) - 1):
+        raise Exception('correct answer must be between -9007199254740991 and +9007199254740991 (that is, between -(2^53 - 1) and +(2^53 - 1)).')
```
jonatanschroeder, line 50: This should only be checked if `correct_answer is not None`.

jonatanschroeder (approved):

Issue #3787, opened 2021-02-03
Redis memory usage


mwest1066:
Redis memory usage went unusually high. This correlated with a large spike of users doing externally graded questions, both in-browser editing and via workspaces. I issued a FLUSHALL to redis at 17:45 to clear the memory.

<img width="355" alt="image" src="https://user-images.githubusercontent.com/875833/106679723-e2ca6180-6582-11eb-90e6-1cc1dc9826ab.png">

<img width="355" alt="image" src="https://user-images.githubusercontent.com/875833/106679773-f83f8b80-6582-11eb-9a4c-eeb42091e89d.png">

<img width="355" alt="image" src="https://user-images.githubusercontent.com/875833/106679796-ffff3000-6582-11eb-8ba1-9af1171d8d44.png">

<img width="355" alt="image" src="https://user-images.githubusercontent.com/875833/106679898-26bd6680-6583-11eb-883d-e1c7f5647a51.png">

<img width="355" alt="image" src="https://user-images.githubusercontent.com/875833/106679930-3b99fa00-6583-11eb-91bc-a1a5d5be4d3a.png">

<img width="355" alt="image" src="https://user-images.githubusercontent.com/875833/106679970-4f456080-6583-11eb-944d-69934d6745ff.png">

mwest1066:
It looks like we were actually running out of memory, because we had been setting `maxItems = 100k`: https://github.com/PrairieLearn/PrairieLearn/blob/51a67cf6e34c6b707f7f5787656e4af8db55d3d9/lib/cache.js#L29

But we ran out of memory before we hit the item cap:
<img width="355" alt="image" src="https://user-images.githubusercontent.com/875833/107127800-32a87180-687e-11eb-9b0b-8a8efb6b8203.png">
<img width="355" alt="image" src="https://user-images.githubusercontent.com/875833/107127833-6aafb480-687e-11eb-9e37-454fdde948d0.png">

mwest1066:
This will be fixed by #3806 and redis cluster reconfiguration.

PR #3806, opened 2021-02-06
Add maxItems and maxAge defaults for redis

mwest1066:


nwalters512 (approved):
Looks sensible! Is anything in particular driving this change?

mwest1066:
> Looks sensible! Is anything in particular driving this change?

See #3787. This is just one small change to get a better handle on redis memory usage.

PR #3807, opened 2021-02-06
Fix LTI nonce re-use detection

mwest1066:
The LTI authentication code stores used nonces in the cache to detect whether a nonce has been re-used (not allowed). However, the cache code that sets values with a maxAge parameter was incorrect. It used the underlying redis syntax with a `PX` argument, whereas the `redis-lru.set()` function does not need (or support) this argument:
https://github.com/facundoolano/redis-lru/blob/40fcf9f7d4690ad126528547fafecd671fbfda26/index.js#L112

Note that the underlying code _does_ use `PX` because it's talking directly to redis:
https://github.com/facundoolano/redis-lru/blob/40fcf9f7d4690ad126528547fafecd671fbfda26/index.js#L119

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -58,7 +58,7 @@ module.exports.set = function(key, value, maxAgeMS=null) {
             // care if it errors
             if (maxAgeMS) {
                 // we don't log the error because it contains the cached value, which can be huge and which fills up the logs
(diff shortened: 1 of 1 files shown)


nwalters512 (approved):
Reviewed from my phone and did not test, but I looked at the docs and this looks good!

PR #3817, opened 2021-02-10
Fix workspace file loading when using S3 backend

mwest1066:


nicknytko (approved):
looks good, this is basically undoing the changes I made previously?

mwest1066:
> looks good, this is basically undoing the changes I made previously?

I hope so! I should check the diff more carefully.

mwest1066:
> I hope so! I should check the diff more carefully.

Yep, it undoes it, except for the change `workspace.s3_name` -> `workspace.remote_name`.

PR #3826, opened 2021-02-14
Add support for normalizing non-ASCII input

jonatanschroeder:
* [x] pl-string-input
* [x] pl-file-editor
* [x] documentation


elements/pl-string-input/pl-string-input.py
```diff
@@ -184,6 +187,10 @@ def parse(element_html, data):
         data['submitted_answers'][name] = None
         return
 
+    if normalize_to_ascii:
+        a_sub = unidecode(a_sub)
```
mwest1066, line 191: Should we also normalize `data["correct_answers"][xxx]` in `grade()`? Or just a correct answer specified in the HTML?

mwest1066 (approved):

zillesc:
Needs to add "unidecode" to package.json ?

mwest1066:
@zillesc to test this branch you can do:
```
cd ~/git/PrairieLearn
git fetch
git checkout force-ascii
docker pull prairielearn/prairielearn:force-ascii
docker run -td --name=pl_container -p 3000:3000 -v ${PWD}:/PrairieLearn2 prairielearn/prairielearn:force-ascii /bin/bash
docker exec -t pl_container pip3 install -r /PrairieLearn2/environments/plbase/python-requirements.txt
docker exec -it pl_container /PrairieLearn/docker/init.sh
# wait for the message "info: Go to http://localhost:3000"
# use PL at http://localhost:3000
# type Ctrl-C to stop PL
docker rm -f pl_container
```

zillesc:
Okay.  I've checked and it (in general) seems to fix issues related to smart quotes generated by Microsoft Word.  (I have no idea what browsers introduce smart quotes...)

I do still have one set of questions with an issue.  I do a slightly gross thing where I use the external auto-grader with pl-string-input elements.  I have a script that reads from the "data.json" file that gets pushed to the container and it appears that that container is still receiving a data.json with non-ASCII characters.  I'm confused why this would be, because (from looking at the code) you seem to be re-writing the contents of data['submitted_answers'][name] which is where I'm grabbing it from.  Any ideas?

Otherwise, I'm happy with the patch.

jonatanschroeder:
> I do still have one set of questions with an issue. I do a slightly gross thing where I use the external auto-grader with pl-string-input elements. I have a script that reads from the "data.json" file that gets pushed to the container and it appears that that container is still receiving a data.json with non-ASCII characters. I'm confused why this would be, because (from looking at the code) you seem to be re-writing the contents of data['submitted_answers'][name] which is where I'm grabbing it from. Any ideas?

I have a similar question, where I get an expression from a pl-string-input, and it worked for me (the external grader got the converted string)... Did you create a new variant after setting `normalize-to-ascii` in your question.html?

zillesc:
I'm still not sure where my problem is, but it appears to be my problem, and I don't want to hold up merging this.

elements/pl-string-input/pl-string-input.py
zillesc: I don't think normalizing correct answers is necessary, personally, because those are supplied by the instructor.  If you as a developer are SPAMing smart quotes all over your questions, it seems like you might have bigger problems...

zillesc (approved):
I don't see anything that I have concerns about.

PR #3827, opened 2021-02-14
Fix access to serverFilesCourse on chunk servers

mwest1066:
This PR makes three (independent) fixes to chunk loading for question code execution:

1. **Changes all references to `course.path` to `context.course_dir`.** This is needed on chunk consumers because `course.path` is the on-disk path to the course only on chunk generators. On chunk consumers the unpacked chunks for a course will be located at a different path. We set `context.course_dir = chunks.getRuntimeDirectoryForCourse()` so that it will be correct on both chunk consumers and generators.

2. **Removes a redundant call to `chunks.ensureChunksForCourseAsync()` in `loadElementsForCourseAsync()`.** This is redundant because we already called it in the parent function `getContextAsync()`.

3. **Adds `clientFilesCourse` to the list of chunks that are pulled before running any question code.** When we run `server.py` functions we give them the on-disk location of `clientFilesCourse` as `data["options"]["client_files_course_path"]` (see https://prairielearn.readthedocs.io/en/latest/question/#accessing-files-on-disk) so we are implicitly saying that it's ok for `server.py` functions to access this on-disk directory, so we need to make sure it's actually on disk.


nicknytko:
I see a few other places we are using course.path instead of the chunked path:

https://github.com/PrairieLearn/PrairieLearn/blob/91299efa5d58f01f7171a4ff9b8dc5a68ce06eab/question-servers/freeform.js#L227
https://github.com/PrairieLearn/PrairieLearn/blob/91299efa5d58f01f7171a4ff9b8dc5a68ce06eab/question-servers/freeform.js#L873

mwest1066:
> I see a few other places we are using course.path instead of the chunked path:

Good catch! Do we also need to pull chunks explicitly for those paths?

nicknytko:
> > I see a few other places we are using course.path instead of the chunked path:
> 
> Good catch! Do we also need to pull chunks explicitly for those paths?

I think the chunks should be already loaded at this point, we just need to make sure we're getting the correct path.

mwest1066:
> I think the chunks should be already loaded at this point, we just need to make sure we're getting the correct path.

Yep, right, because they are loaded as the first thing by `getContextAsync()`, right?

mwest1066:
Given that we are loading element extension chunks here: https://github.com/PrairieLearn/PrairieLearn/blob/91299efa5d58f01f7171a4ff9b8dc5a68ce06eab/question-servers/freeform.js#L1469 can we eliminate this load: https://github.com/PrairieLearn/PrairieLearn/blob/91299efa5d58f01f7171a4ff9b8dc5a68ce06eab/question-servers/freeform.js#L146
which (as far as I can see) is only called from here: https://github.com/PrairieLearn/PrairieLearn/blob/91299efa5d58f01f7171a4ff9b8dc5a68ce06eab/question-servers/freeform.js#L1480

mwest1066:
Do we also have a bug where we aren't pulling chunks for the code path following this call? https://github.com/PrairieLearn/PrairieLearn/blob/91299efa5d58f01f7171a4ff9b8dc5a68ce06eab/sync/syncFromDisk.js#L60

mwest1066:
> Do we also have a bug where we aren't pulling chunks for the code path following this call?
> 
> https://github.com/PrairieLearn/PrairieLearn/blob/91299efa5d58f01f7171a4ff9b8dc5a68ce06eab/sync/syncFromDisk.js#L60

See #3845 for resolution of this.

mwest1066:
> can we eliminate this load:
> https://github.com/PrairieLearn/PrairieLearn/blob/91299efa5d58f01f7171a4ff9b8dc5a68ce06eab/question-servers/freeform.js#L146

Eliminated by fa356d68cecd3edf46940f6d6dc6a899336e9f98

nwalters512 (approved):
This all makes sense to me!

nicknytko (approved):

PR #3828, opened 2021-02-14
C Grader: Add support for valgrind and multiple C files

mukund0299:


jonatanschroeder (approved):
I made a few changes as discussed via Slack.

mwest1066 (approved):

PR #3095, opened 2020-09-09
Create Manual Grading RFC

mebird:
## Description
- As discussed in the past few weeks, we'd like to bring a manual grading feature to PL
- This RFC outlines some basic requirements for said feature, as well as some "stretch" or additional goals/features we find relevant to manual grading.

PR #3835, opened 2021-02-16
grading_method Column Split

bevansm:
This PR establishes the split of `grading_method` into three boolean columns (`*_external`, `*_internal`, `*_manual`) in preparation for parallel grading workflows.

This alters the `questions`, `submissions`, and `grading_jobs` columns. The migration proc is not atomic, and it may throw an error if it's run on an already migrated deploy. If that's not okay, I need to take another look on how to only trigger a table update if we've actually added columns--this may require some sort of inline function.


sync/fromDisk/questions.js
```diff
@@ -28,7 +28,9 @@ function getParamsForQuestion(q) {
         options: q.options,
         client_files: q.clientFiles || [],
         topic: q.topic,
-        grading_method: q.gradingMethod || 'Internal',
+        grading_method_internal: !q.gradingMethod || q.gradingMethod === 'Internal',
```
mebird, line 31: Here's where we'll change config parsing to allow multiple "grading methods" per assignment.

sprocs/grading_jobs_insert.sql
```diff
@@ -40,15 +42,15 @@ BEGIN
     -- ######################################################################
     -- delegate the call
 
-    IF grading_method = 'Internal' THEN
+    IF grading_method_internal = True THEN
```
mebird, line 45: Think this is going to be the choke point for adding another grading job (i.e. external/manual in parallel with internal). We may have to deal with ordering though... What if the internal job resolves, but the scores conflict with the external output? Do we have to do a split on the scores field as well?

migrations/227_grading_method_split.sql
```diff
@@ -0,0 +1,33 @@
...
+    ADD COLUMN IF NOT EXISTS grading_method_internal boolean DEFAULT True,
+    ADD COLUMN IF NOT EXISTS grading_method_external boolean DEFAULT False,
+    ADD COLUMN IF NOT EXISTS grading_method_manual boolean DEFAULT False;
+ALTER TABLE grading_jobs
+    ADD COLUMN IF NOT EXISTS grading_method_internal boolean DEFAULT True,
+    ADD COLUMN IF NOT EXISTS grading_method_external boolean DEFAULT False,
+    ADD COLUMN IF NOT EXISTS grading_method_manual boolean DEFAULT False;
+
+-- Update based on old values
+UPDATE submissions
```
mebird, line 16: Here's the update that may break on an already migrated deploy. I can add a column check on the table if needbe (i.e. only run these `UPDATE` statements if the `grading_method` column is still hanging around.

sprocs/grading_jobs_insert.sql
andrewstec: I agree that this is a choke point, but I am not sure how the internal job would resolve but then have a conflicting score with the external output if a question can only exclusively be external or internal.

mebird: For now, it's not an issue since they are exclusive. In the future, it may be.

It's not an issue with adding a parallel manual workflow though, providing we tease that out.

andrewstec: Yeah, agreed. I think the "parallel manual workflow" could be a call to a different sproc. It doesn't seem to make much sense that we have internal and external question types that are graded and then change their type to manual to be graded for manual grading, which then leads down a third possible delegated call here... 

Manual grading seems more like an action than a grading method type now.

andrewstec: One additional note, we could use the weighted score in the question configuration to segment the final score and visualize what portion accounts for manual grading on the front-end. 

Don't know if this is too far down the road to think about, but it looks very close to achieving here.

pages/partials/submissionStatus.ejs
```diff
@@ -6,13 +6,13 @@
...
     <% } else { %>
-        <% if (submission.grading_method == 'External') { %>
+        <% if (submission.grading_method_external) { %>
             <% if (submission.gradable) { %>
                 <span id="grading-status-<%= submission.id %>" class="badge badge-secondary"></span>
             <% } else { %>
                 <span class="badge badge-danger">invalid, not gradable</span>
             <% } %>
-        <% } else if (submission.grading_method == 'Manual') { %>
+        <% } else if (submission.grading_method_manual) { %>
```
andrewstec, line 15: This may require some inspection to ensure that it works with the new logic... not just the boolean changes, but the new manual grading approach in general.

migrations/227_grading_method_split.sql
```diff
@@ -0,0 +1,33 @@
...
+UPDATE questions 
+    SET grading_method_internal = grading_method = 'Internal',
+        grading_method_external = grading_method = 'External',
+        grading_method_manual   = grading_method = 'Manual';
+UPDATE grading_jobs 
+    SET grading_method_internal = grading_method = 'Internal',
+        grading_method_external = grading_method = 'External',
+        grading_method_manual   = grading_method = 'Manual';
+
+-- Drop old grading method column
```
andrewstec, line 29: We would like to eventually drop this. I wonder if we should wait until a successful test-run and after the migration though. This drop table could be a small PR on its own.

pages/instructorGradingJob/instructorGradingJob.sql
```diff
@@ -24,5 +24,5 @@ FROM
     LEFT JOIN users AS u ON (u.user_id = s.auth_user_id)
 WHERE
     gj.id = $job_id
-    AND gj.grading_method = 'External'
+    AND gj.grading_method_external = True
```
andrewstec, line 27: So this is one area that it is nice to have the grading_method_external on the grading_job. 

If a graded submission has a grading job, I believe we could just look for these boolean flags in the `grading_job` table. 

Do you believe that it is necessary to put the grading method booleans in the submission as well?

sprocs/assessment_instances_select_log.sql
```diff
@@ -167,7 +167,7 @@ BEGIN
                     LEFT JOIN users AS u ON (u.user_id = gj.auth_user_id)
                 WHERE
                     iq.assessment_instance_id = ai_id
-                    AND gj.grading_method = 'Manual'
+                    AND gj.grading_method_manual = True
```
andrewstec, line 170: Another example of where it looks like these boolean fields could live in the grading job column. We may have to unpack this logic a little to understand why these grading types are important here. 

I am not too sure what happens if an item has been internally and externally graded now, and how it relates to logging compared to if it was only an externally graded question, etc.

sprocs/grading_jobs_insert_internal.sql
```diff
@@ -25,15 +25,17 @@ DECLARE
...
+    grading_method_manual boolean;
     new_correct boolean;
 BEGIN
     -- ######################################################################
     -- get the related objects
 
     -- we must have a variant, but we might not have an assessment_instance
-    SELECT s.credit,       v.id, q.grading_method,              iq.id,                  ai.id
-    INTO     credit, variant_id,   grading_method, instance_question_id, assessment_instance_id
+    SELECT s.credit,       v.id, q.grading_method_internal, q.grading_method_external, q.grading_method_manual,                iq.id,                  ai.id
```
andrewstec, line 37: It does not necessarily make sense to insert the grading method booleans from the question into the grading_job table. 

Perhaps we should define an internal, external, and manual grading operation and once they complete, update the grading job. Or we can still delegate the calls and include a flag to update the boolean based on which delegated call was hit.

This is my first pass at this logic, so I still have to digest and think about this. Some input from @mwest1066  would be quite valuable here too.

sync/fromDisk/questions.js
andrewstec: So it seems that the idea is to still configure the question, and then grade the submission on the basis of the flags in the configuration file. This looks good!

We then create a grading job and insert some of the internal, external, and manual boolean fields that are updated. I am starting to see how you are getting pulled into the situation where this information lives on the submission and grading_job tables. I think we may want to join this information though. It will require modifying some queries. @mwest1066

andrewstec:
I think this PR works @mebird ! Overall, it looks like the question configuration is straight-forward, but we could limit the boolean fields to either the submission or grading_job tables. I do not think that they need to be in both places at this time. 

As the current code checks the grading_job table for the grading type, perhaps we can stick with that approach and see how we can integrate our logic with the new grading_job method boolean logic. 

Do we have to also include these fields in the submission?

pages/partials/submissionStatus.ejs
mebird: I know this works with the workflow from before--I am going to do a sanity check on most of these pages to ensure backwards compatibility though.

migrations/227_grading_method_split.sql
mebird: The drop itself or the data migration? Again, I'm still somewhat paranoid if this proc is ever run twice on the DB... I've tried to sniff around with a better way to do the column update and then drop, but kinda unsure.

Unless what you're saying is just to do the column add/migration in this PR, and then the drop of the old columns/enums in another?

pages/instructorGradingJob/instructorGradingJob.sql
mebird: It's just tied to the submission in so many other places in the code--I think the frontend queries base it off of the submission columns, rather than the grading job columns, in many places. I think it was tied to the submission earlier because of that--the data is easily at hand without doing table joins.

sprocs/assessment_instances_select_log.sql
mebird: Yeah, I'm a bit tentative on how to deal with joint logic now, mainly since before internally and externally graded questions have been completely disjoint. There's no support to internally and then externally grade--the external grading workflow, if it's run, will override the scores of the internally graded version of the question at this point in time. And if there's an internally graded portion, and then also an externally graded portion, we'll still have the two workflows stomping on each other, instead of creating a composite score.

After stepping back for a bit on this, what are your opinions on having an internal/external grading method column that forces the two to be disjoint, and then a second manual grading method column?

mebird:
> I think this PR works @mebird ! Overall, it looks like the question configuration is straight-forward, but we could limit the boolean fields to either the submission or grading_job tables. I do not think that they need to be in both places at this time.
> 
> As the current code checks the grading_job table for the grading type, perhaps we can stick with that approach and see how we can integrate our logic with the new grading_job method boolean logic.
> 
> Do we have to also include these fields in the submission?

Again, for whatever reason, these fields on the submission tend to get dragged through to frontend logic/weird places in the code--it's equally as prevalent as the column on the grading jobs. I can take a look into removing the submission column and seeing where I can replace it with a join on the questions table/the grading_jobs table, but I'm not 100% on it. If @mwest1066  can provide the original motivation for having it lie on all three tables, then that may be useful.

bevansm:
Okay. Thinking on this more over the past week, and how we're moving towards parallels workflows.

- We can't limit the fields to only the submission or the grading_job, since we'll need grading jobs in order to tell what steps we have left (manual v external), and we need them on the submission to persist high level information about what types of steps we might need to run.
- I think we're going to have to move the queries for "what questions we have left to manually grade" from the submissions to the grading jobs. With the current flow, if we have a question that has grades filled in by an internal grader, then the scores won't be null. Hence, the manual grading queries won't pick it up.
- When we make the config file, we should have an option for external/internal, then a secondary option for manual. Internal and External can't be parallelized right now--and I don't think they ever _will_ be (or will they?).
   - With that in mind, should we just be adding a manual grading flag to questions and adding logic that adds a manual grading job (in the grading jobs table?) so we can base the queue off those jobs.
   - We also have parallelism issues with external grading here--maybe we block manual grading if there's another job present with a non-manual type?
   - In this world, do we keep the question type (manual, internal, external, etc.) and add another optional boolean that can add manual grading as another layer?
   - Does this completely stomp over the current work we have with the TA grading locks? 
   - Does this change our perception of how manual grading jobs work?
- I can see where this could work, but does this require massive changes to manual grading? Is this too much of a paradigm shift?

bevansm:
Based upon our discussion on Tuesday, we need this col split, and it should be robust to everything currently in master. I'm still a bit shaky on the migration proc because it completely changes our enum (and deletes it), and may break if run twice. If you have any suggestions on this, matt, it'd be much appreicated.

bevansm:
Working on a separate branch for the grading workflow split itself (insomnia man)--looks like we might want to keep the enum around for the grading_jobs table so we can record different job types. Taking a look into that now.

bevansm:
Based upon our discussion today, I'm going to roll back my changes to the grading_jobs table, drag over some code from my other working branch, and re-introduce the grading_method enum, since it looks like we'll always need it on grading_jobs.

lib/question.js
```diff
@@ -579,10 +579,11 @@ module.exports = {
 
                     /* If the submission was marked invalid during grading the grading job will
                        be marked ungradable and we should bail here to prevent LTI updates. */
-                    grading_job = result.rows[0];
-                    if (!grading_job.gradable) return callback(new NoSubmissionError());
-
-                    debug('_gradeVariantWithClient()', 'inserted', 'grading_job.id:', grading_job.id);
+                    grading_jobs = result.rows;
+                    for (const grading_job of grading_jobs) {
+                        if (!grading_job.gradable) return callback(new NoSubmissionError());
```
andrewstec, line 584: Hmmmmmm... Just a question about this @bevansm . Do we know if all of the grading_jobs would be gradable or not at this point? 
 
Let's suppose external grading jobs are running at the same time. Do we have to worry about the race condition here?

sync/fromDisk/questions.js
bevansm: Yup, configure the question, then create discrete jobs for each type of grading method that we'll deal with. We'd have to look at the `gj.grading_method_manual` field to insert an initial job, and then we should be able to just look for any incomplete grading job in the DB. I know this gets hairy with how we handle conflicts in the spin lock prs (since we also have manual jobs for pending grading info... stuff), but it aligns with how we discover and handle external grading jobs.

lib/question.js
bevansm: At this point, we can have only a manual grading job or an external grading job per variant. Hence, it falls under the same logic as before, as we can't actually have multiple types of jobs, and hence multiple jobs, per question in production.

+ gradable status currently only relates to external jobs--all manual jobs should be marked as gradable. If this needs to change for grading pipelines (i.e. only mark a manual job as gradable once external jobs have finished), then this might have a different meaning.

sprocs/grading_jobs_insert_external_manual.sql
```diff
@@ -1,7 +1,10 @@
+DROP FUNCTION IF EXISTS grading_jobs_insert_external_manual(bigint, bigint);
+
 CREATE OR REPLACE FUNCTION
```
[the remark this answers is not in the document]
bevansm: Splitting the job is a +1 for me; I just wanted to establish the column split without trodding over _too_ much. If you split the sprocs in your branch, then it may be worth reverting this file, and just calling that sproc instead of passing the grading method type to this sproc (as one currently does).

database/tables/submissions.pg
```diff
@@ -9,7 +9,9 @@ columns
     format_errors: jsonb
     gradable: boolean default true
     graded_at: timestamp with time zone
-    grading_method: enum_grading_method
+    grading_method_external: boolean default false
```
andrewstec, line 12: TO DO: Come back to this for review, as I don't believe this is used by anything. This information is available at the Question level and the grading method appears in the grading_jobs.

.gitignore
```diff
@@ -77,3 +77,9 @@ __pycache__/
 .RData
 .Ruserdata
 
+# VSCode files
```
andrewstec, line 80: This is optional and can be taken if wanted.

migrations/227_grading_method_split.sql
```diff
@@ -0,0 +1,23 @@
+-- Add new grading columns
+ALTER TABLE submissions 
+    ADD COLUMN IF NOT EXISTS grading_method_internal boolean DEFAULT True,
+    ADD COLUMN IF NOT EXISTS grading_method_external boolean DEFAULT False,
```
andrewstec, line 4: TO DO: Review if submission level checks are used, as grading_jobs and question level information already exists.

testCourse/questions/externalGrade/codeUpload/question.html
```diff
@@ -0,0 +1,26 @@
+<pl-question-panel>
+  <p>
```
andrewstec, line 2: We may want to consider creating an image for the externalGrader that cannot be touched by anyone. If someone updates the image and changes something, the CI/CD tests will break and we will not know why. 

For optimization, it could be Alpine Linux, which will be much quicker in an already long pipeline. This should cut off 20-30 seconds at least in the test time.

database/tables/submissions.pg
andrewstec: Nope. Still needed for the question.js to determine the nature of the question on a submission. I suppose this could be altered to come directly from the Question file.

sync/fromDisk/questions.js
```diff
@@ -4,6 +4,36 @@ const sqldb = require('../../prairielib/lib/sql-db');
...
+/**
+ * Returns true iff the given question is an internal grading question
+ * @param {import('../course-db').Question} q
+ * @returns {boolean}
+ */
+function hasInternalGrading(q) {
+  const { gradingMethod, gradingMethods } = q;
+  if (gradingMethods) {
+    return (
+      (Array.isArray(gradingMethods) && !gradingMethods.length) || hasGradingMethod(q, 'Internal')
```
nwalters512, line 31: I don't 100% understand what's going on here - does this mean that an empty array of `gradingMethods` implies internal grading is enabled?

testCourse/questions/internalExternalManualGrade/addingNumbersCodeUpload/question.html
```diff
@@ -0,0 +1,54 @@
...
+
+<section>
+  <p>This question combines "Add two numbers", an "Internal" grading question, and "Fibonacci function, file upload", an "External" grading question.</p>
+
+  <p>This question has been configured with "Internal", "External", and "Manual" grading methods on the info.json 'gradingMethods' property. The "Internal" and "External" grading methods run configured as usual. 
+  The `server.py` files from the two questions have been merged into one `server.py` to accept both the "Internal" and "External" parameters submitted by the student when a question is answered.</p>
+
+  <p>The "Manual" grading method leverages the pre-existing "Internal" grading submission and results, and the "External" grading submission, displayed on the page. No HTML or Python boilerplate has been added to allow for the additional "Manual" grading method. 
+    
+  NOTE: It is, at the moment, not possible to display the "External" grading results on the student question submission view if the "Internal" grading method is also enabled. Therefore, as the view is re-used for the manual grading instructor view, it would also not be able to view the "External" grading results.</p>
```
nwalters512, line 12:
```suggestion
  <p>The "Manual" grading method leverages the pre-existing "Internal" grading submission and results, and the "External" grading submission, displayed on the page. No HTML or Python boilerplate has been added to allow for the additional "Manual" grading method.</p>
    
  <p>NOTE: It is, at the moment, not possible to display the "External" grading results on the student question submission view if the "Internal" grading method is also enabled. Therefore, as the view is re-used for the manual grading instructor view, it would also not be able to view the "External" grading results.</p>
```

migrations/245_grading_method_split.sql
```diff
@@ -0,0 +1,23 @@
...
+        grading_method_external = grading_method = 'External',
+        grading_method_manual   = grading_method = 'Manual';
+UPDATE questions 
+    SET grading_method_internal = grading_method = 'Internal',
+        grading_method_external = grading_method = 'External',
+        grading_method_manual   = grading_method = 'Manual';
+
+-- Drop old grading method column
+ALTER TABLE submissions DROP COLUMN IF EXISTS grading_method;
+ALTER TABLE questions DROP COLUMN IF EXISTS grading_method;
```
nwalters512, line 23: This is almost certainly going to cause problems when it comes time to deploy this change, as for a while we'll have some servers running old and new versions of this code simultaneously. Unfortunately, I think we'll need to keep the old `grading_method` column around for a while and remove it once no more servers are trying to interact with it. Alternatively, we could decide to accept some downtime to deploy this change and take PL completely offline during the migration. However, that would complicate deployments for folks who are running PL themselves.

nwalters512 (changes requested):
I'm tentatively blocking this on figuring out the deployment strategy. If we decide to just accept downtime during this deployment, I'll remove the block.

@bevansm @mebird we'll need you to sign our CLA before we can approve and merge this PR. Please sign it at your earliest convenience! Instructions can be found in this comment: https://github.com/PrairieLearn/PrairieLearn/pull/3835#issuecomment-879225020

bevansm:

bevansm:
> I have reviewed and hereby sign the CLA

You'll have to override the CLA for mebird; I've changed my account name since contributing.

bevansm:
It's been about eight months since I've contributed @nwalters512 and @andrewstec. What's the most productive way I can lend a hand to get this off the ground? I don't have much context as to the current state of the overall project.

nwalters512:
@bevansm unfortunately, Andrew will no longer be involved in development. I'm trying to pick up where y'all left off, but I also don't have a good overview of this project as a whole. Did y'all write any planning/design docs that I can take a look at to give me some context?

sprocs/grading_jobs_insert.sql
```diff
@@ -34,17 +36,27 @@ BEGIN
...
-    ELSIF grading_method = 'External' OR grading_method = 'Manual' THEN
-        grading_job := grading_jobs_insert_external_manual(submission_id, authn_user_id);
-    ELSE
-        RAISE EXCEPTION 'unknown grading_method: %', grading_method;
     END IF;
+    
+    -- delegate external/manual grading job
+    IF grading_method_external = True OR grading_method_manual = True THEN
+        RETURN NEXT grading_jobs_insert_external_manual(submission_id, authn_user_id, 'External');
+    END IF;
```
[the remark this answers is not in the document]
bevansm: We had a discussion about running manual grading as a "grading job" that should be completed by a person rather than a machine. This code is incorrect since the type of grading job inserted should be 'external'/'manual' rather than always external.

However, I don't know if this was the final representation we used for manual grading internally (i.e. if there's a separate table for manual jobs).

sync/fromDisk/questions.js
bevansm: Line 29: Null check for `gradingMethods`. In this case, fallback to `gradingMethod`.
Line 31: Yes, no `gradingMethods` implies Internal grading (just like no `gradingMethod` implies Internal (see line 34).

schemas/schemas/infoQuestion.json
```diff
@@ -77,13 +77,25 @@
             "type": "string"
         },
         "gradingMethod": {
-            "description": "The grading method used for this question.",
+            "description": "The grading method used for this question. (DEPRECATED - Use `gradingMethods` below)",
```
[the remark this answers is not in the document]
bevansm: `gradingMethod `is still supported in `questions.js`. Agreed that this probably shouldn't be marked as deprecated, but there should be some sort of validation check that at least one is set (in the case neither is set, defaults to Internal).

migrations/245_grading_method_split.sql
bevansm: Deployment wise, makes sense. Can we feature flag this to help reduce drift?

bevansm:
> @bevansm unfortunately, Andrew will no longer be involved in development. I'm trying to pick up where y'all left off, but I also don't have a good overview of this project as a whole. Did y'all write any planning/design docs that I can take a look at to give me some context?

High level as to why we needed the split:

- A question can be both manually & externally graded or just externally graded (same for manual + internal).
- Currently, PL only allows for one grading type per method.
- Adding a "grading method" table with rows of [FK to a question/submission, grading type] means we have to make calls out to the grading method table to get all the info on a question/submission rather than just calling the question/submission tables. Didn't seem like we'd have even more grading types in the future, so three binary columns that could be arbitrarily combined made sense.

Docs/resources (think you already have some of these):
- [The "Instructor Controls" Minutes](https://docs.google.com/document/d/150z0vIHpyoB0RygXTknDIleKNZC4VxLjCIm-jcWG8qY/edit). Lots of discussion of the project as we started and some technical bits and bobs. Think you're already in the loop, but the mentions of "Maja's PR" on 7/26 and before are regarding this PR.
- [Outdated RFC.](https://github.com/PrairieLearn/PrairieLearn/pull/3095/files). High level of the project & goals.
- [Project Board](https://github.com/PrairieLearn/PrairieLearn/projects/9#card-45180056). Unsure of what's on this board/what's out of date.
- [Some Notes in This PR](https://github.com/PrairieLearn/PrairieLearn/pull/3835#issuecomment-789160568)

Hopefully the brief braindump helps! Again, it's been nearly a year--I've tried not to misrepresent anything.

nwalters512:
We'll be taking a different approach for manual grading that does not require this column to be split.

PR #3836, opened 2021-02-16
Manual grading anti ta overlap

andrewstec:
Scope of changes:

- Adds segmentation of data points to instructor assessment view.
- Adds additional logic to ensure that instance questions are tagged with the user who is marking
- Fixes issues with next instance question loading
- Adds logic to lock database tables for update


pages/instructorQuestionManualGrading/instructorQuestionManualGrading.ejs
```diff
@@ -22,6 +22,32 @@
     </div>
     <div id="content" class="container">
       <div class="row">
+
+        <% if (typeof submission != null) { %>
+          <% if (submission.score == null && submission.manual_grading_user != null) { %>
+          <div class="col-lg-12 col-sm-12">
+            <div class="card mb-4">
+              <div class="card-header bg-warning text-blue">
+                <% if (authn_user.uid === user.uid) { %>
```
mwest1066, line 31: Use actual `id` for testing equality.
```suggestion
                <% if (authn_user.id === user.id) { %>
```

mwest1066, line 31: What is `user` here? Is this supposed to be `grading_user`?

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.ejs
```diff
@@ -22,6 +22,32 @@
...
+                  You are currently grading this question
```
mwest1066, line 32: I think we should skip this statement and only display something if someone else is currently grading the question.

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.sql
```diff
@@ -7,21 +7,92 @@ SELECT
...
+             GROUP BY s.auth_user_id
+         )
+         AND manual_grading_user IS NOT NULL
+         AND graded_at IS NULL) AS num_locked_submissions,
+    (SELECT COUNT(DISTINCT IQ.id)
+     FROM 
+         submissions AS S
+         JOIN variants AS V ON (s.variant_id = v.id)
+         JOIN instance_questions AS IQ ON (v.instance_question_id = iq.id)
+     WHERE (s.auth_user_id, s.date)
```
mwest1066, line 75: I'm not loving this method of finding `max(date)` and then re-finding the submission with the maximum date. Elsewhere we sort by `s.date DESC, s.id DESC` and then LIMIT the results: https://github.com/PrairieLearn/PrairieLearn/blob/5483eb8c5698e3b3e5b7e888c1d9b4e51434e270/sprocs/variants_select_submission_for_grading.sql#L18

Can we do the same here with a `DISTINCT ON (iq.id)` to get the list of instance questions that we are interested in?

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.ejs
```diff
@@ -22,6 +22,32 @@
     </div>
     <div id="content" class="container">
       <div class="row">
+
+        <% if (typeof submission != null) { %>
+          <% if (submission.score == null && submission.manual_grading_user != null) { %>
+          <div class="col-lg-12 col-sm-12">
```
mwest1066, line 28: Is this `<div>` needed? What's it doing?

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.ejs
```diff
@@ -51,10 +56,29 @@
...
+                    <a class="btn btn-primary" href="<%= urlPrefix %>/assessment/<%= assessment.id %>/assessment_question/<%= question.id %>/unlock_ungraded">
+                      Unlock <%= question.num_locked_submissions %> submission<% if (question.num_locked_submissions > 1) { %>s<% } %>
+                    </a>
+                    <% } %>
+                </td>
                 <td>
                   <% if (question.num_ungraded_submissions > 0) { %>
                   <a class="btn btn-primary" href="<%= urlPrefix %>/assessment/<%= assessment.id %>/assessment_question/<%= question.id %>/next_ungraded">
-                    Grade <%= question.num_ungraded_submissions %> submission<% if (question.num_ungraded_submissions > 1) { %>s<% } %>
+                    Grade <%= question.num_ungraded_submissions %> submission<% if (question.num_ungraded_submissions) { %>s<% } %>
```
mwest1066, line 81: I feel like a simple `Grade` button would be clearer?
```suggestion
                    Grade
```

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.ejs
```diff
@@ -26,7 +26,12 @@
                 <th></th>
                 <th>QID</th>
                 <th>Points</th>
-                <th>Number of Submissions</th>
+                <th>Total Submissions</th>
```
mwest1066, line 29: Do we care about this?

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.ejs
```diff
@@ -26,7 +26,12 @@
...
+                <th>Last Submissions for Manual Grading</th>
```
mwest1066, line 30: Is this `Need grading`?

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.ejs
```diff
@@ -26,7 +26,12 @@
...
+                <th>Manually Graded</th>
```
mwest1066, line 31: `Already graded` or `Grading complete`?

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.ejs
```diff
@@ -26,7 +26,12 @@
...
+                <th>Locked Submissions</th>
```
mwest1066, line 32: Do we care about this?

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.js
```diff
@@ -3,24 +3,24 @@ const express = require('express');
...
-        assessment_id: res.locals.assessment.id,
-    };
+    const params = [
+        res.locals.assessment.id,
+        res.locals.assessment_question_id,
+        res.locals.authn_user.user_id,
+    ];
     // Unmarked instance question df. Is last created submission of instance question AND has null graded_at value.
-    sqlDb.queryZeroOrOneRow(sql.get_next_unmarked_instance_question, params, (err, result) => {
+    sqlDb.callZeroOrOneRow('instance_questions_update_submission_for_manual_grading', params, (err, result) => {
```
mwest1066, line 15: Below we assume that `rows[0]` exists. We should use `callOneRow()` to ensure this.

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.js
```diff
@@ -3,24 +3,24 @@ const express = require('express');
...
         if (ERR(err, next)) return;
 
-        // If we have no more submissions, then redirect back to manual grading page
-        if (!result.rows[0]) {
+        // If we have no more instance questions with gradable submissions, then redirect back to manual grading page
+        if (!result.rows[0].instance_question) {
             res.redirect(`${res.locals.urlPrefix}/assessment/${res.locals.assessment.id}/manual_grading?done`);
```
mwest1066, line 20: What's the `?done` query for on the end of this redirect?

server.js
```diff
@@ -710,10 +710,24 @@ module.exports.initExpress = function() {
         require('./pages/studentAssessmentInstanceHomework/studentAssessmentInstanceHomework'),
         require('./pages/studentAssessmentInstanceExam/studentAssessmentInstanceExam'),
     ]);
+    app.use('/pl/course_instance/:course_instance_id/instructor/assessment/:assessment_id/assessment_question/:assessment_question_id/unlock_ungraded', [
```
mwest1066, line 713: This is an instructor route but it's in the section for student routes: https://github.com/PrairieLearn/PrairieLearn/blob/5483eb8c5698e3b3e5b7e888c1d9b4e51434e270/server.js#L672

See also this earlier comment: https://github.com/PrairieLearn/PrairieLearn/pull/3707/files#r565474285

server.js
```diff
@@ -710,10 +710,24 @@ module.exports.initExpress = function() {
...
+        function(req, res, next) {res.locals.assessment_question_id = req.params.assessment_question_id; next();},
+        require('./pages/instructorQuestionManualGrading/instructorQuestionManualGradingUnlockInstanceQuestions'),
```
mwest1066, line 715: We are using a convention of one route handler per directory in `pages/`. That is, `instructorQuestionManualGrading.js` should be the only handler in `pages/instructorQuestionManualGrading/`. We should either handle this query within `instructorQuestionManualGrading.js` or add another directory in `pages`.

pages/instructorQuestionManualGrading/instructorQuestionManualGradingUnlockInstanceQuestions.js
```diff
@@ -0,0 +1,23 @@
+const ERR = require('async-stacktrace');
+const express = require('express');
+const router = express.Router();
+const path = require('path');
+const debug = require('debug')('prairielearn:' + path.basename(__filename, '.js'));
+const {sqlDb, sqlLoader} = require('@prairielearn/prairielib');
+const sql = sqlLoader.loadSqlEquiv(__filename);
+
+router.get('/', (req, res, next) => {
```
mwest1066, line 9: This seems to be an explicitly state-modifying query and so should be a POST?

database/tables/submissions.pg
```diff
@@ -12,6 +12,7 @@ columns
     grading_method: enum_grading_method
     grading_requested_at: timestamp with time zone
     id: bigint not null default nextval('submissions_id_seq'::regclass)
+    manual_grading_user: bigint
```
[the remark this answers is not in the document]
mwest1066: We are introducing a lot of complexity by allowing a user to have multiple submissions being graded simultaneously. Would it be cleaner to only allow a single question/submission to be graded by any given user at a time?

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.ejs
```diff
@@ -51,10 +56,29 @@
...
+                <td class="text-center">
+                  <%= question.num_locked_submissions %>
+                </td>
+                <td class="text-center">
+                  <%= question.manual_grading_users %>
+                </td>
+                <td class="text-center">
+                  <% if (question.num_locked_submissions > 0) { %>
+                    <a class="btn btn-primary" href="<%= urlPrefix %>/assessment/<%= assessment.id %>/assessment_question/<%= question.id %>/unlock_ungraded">
+                      Unlock <%= question.num_locked_submissions %> submission<% if (question.num_locked_submissions > 1) { %>s<% } %>
```
mwest1066, line 74: If we restrict each user to only having one question at a time locked for grading then we could avoid all this unlocking code.

pages/instructorQuestionManualGrading/instructorQuestionManualGradingAbortInstanceQuestion.js
```diff
@@ -0,0 +1,25 @@
...
+const router = express.Router();
+const path = require('path');
+const debug = require('debug')('prairielearn:' + path.basename(__filename, '.js'));
+const error = require('@prairielearn/prairielib/error');
+const sqldb = require('@prairielearn/prairielib/sql-db');
+const sqlLoader = require('@prairielearn/prairielib/sql-loader');
+
+const sql = sqlLoader.loadSqlEquiv(__filename);
+
+router.get('/', function(req, res, next) {
```
mwest1066, line 12: How do we ever get to this code? I can't see that it's possible for a user to arrive here.

sprocs/instance_questions_select_manual_grading_objects.sql
```diff
@@ -0,0 +1,55 @@
+-- BLOCK instance_question_select_last_variant_with_submission
+DROP FUNCTION IF EXISTS instance_questions_select_manual_grading_objects(bigint, bigint);
+
+-- Retrieves the last variant for an instance question and last submission for the variant.
+
+CREATE OR REPLACE FUNCTION
+    instance_questions_select_manual_grading_objects(
+        IN iq_id bigint,
```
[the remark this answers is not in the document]
andrewstec: Agreed. Done. #beb437811

pages/partials/instructorGradingPanel.ejs
```diff
@@ -26,6 +26,12 @@
             </div>
             <div class="text-right">
               <button type="submit" class="btn btn-primary">Submit and Next</button>
+              <button type="submit" class="btn btn-danger" onclick="((e) => {
+                e.preventDefault();
+                console.log('<%= urlPrefix %>');
+                const thing = '<%= urlPrefix %>';
+                // $.post('<%= urlPrefix %>/abort_grading');
+              })">Skip and Leave Grading</button>
```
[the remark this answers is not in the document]
andrewstec: Fixed. Integrated this and moved the abort_grading endpoint logic into the instructorQuestionManualGrading file.

server.js
andrewstec: Done!

pages/instructorQuestionManualGrading/instructorQuestionManualGradingAbortInstanceQuestion.js
andrewstec: This code has been eliminated. It's now part of the 'InstructorQuestionManualGrading.js' logic under the action 'abort_manual_grading'. It is now integrated property so this logic will be accessible.

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.ejs
andrewstec: We could eliminate this code for sure, but it might come back in another way. For example, suppose we were to lock one item:

1) What happens if a user closes the browser? The item would still be locked.
  a) User is redirected back to question to finish grading on starting manual grading again (if and only if they start manual grading again. Hence, an item could be locked and forgotten, which requires further logic)
    i) Other users take over locked items after x period of time set in config
    ii) Other users take over locked items if there are no more items to be graded --> Could possibly run into a grading conflict here.

It would be ideal if we could lock an item at the database level and let it manage the lock-- eventually releasing it. The for update lock will end after the transaction ends, however. 

Please let me know if you are not concerned with edge cases here, as we can try it out and see if it works without this code.

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.js
andrewstec: Done.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.ejs
andrewstec: It defines a grid as full across for the badge.

image

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.ejs
andrewstec: Sounds good.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.ejs
andrewstec: It might seem redundant, but the view is similar to the question page. This message helps a new user differentiate between a grading page. It can also help a new user understand what is happening behind the scenes while they are grading. Perhaps "You are currently grading this submission. Nobody else will be able to access this submission while you are grading."

It's a nice-to-have. If you want to minimize lines of code and logic, this can go.

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.js
andrewstec: This is deprecated from what I have seen and no longer has functionality. I took it out.

database/tables/submissions.pg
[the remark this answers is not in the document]
andrewstec: The instance_question that is used in the JOIN will be locked on a FOR UPDATE.

However, on the topic of general locking logic so that items cannot be marked in parallel by multiple users, we can move that to the instance_question level. The lock was added to the submission level because the idea of marking more than one submission per instance question was discussed. The impression that I get is this is no longer the case, so it will be moved to the instance question level.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.ejs
cheeren: The right panel and perhaps heading info can be used to characterize the page as a grading page. We should save text for critical warning info indicating that the grading state of the question_instance has changed.

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.sql
andrewstec: I think the only problem here is that we are searching for ungraded submissions, which we _then_  sort and limit by. 

Hence, if I am a student who makes multiple submissions, and then someone grades my last submission, the proposed query logic above will begin to return the second to last submission that is ungraded, even though my instance question has already been graded.

I really don't like the subquery either, but retrieving the last submission and seeing if it is graded seems to be the right logic. 

I am a bit unsure about my decision, as Jonatan implemented the same logic that you are recommending.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.ejs
cheeren: or yes, as matt says, to indicate that someone else is grading the question.

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.sql
andrewstec: Just to confirm, I went a very hacky route here, so I would like to find an alternative to this too.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.ejs
andrewstec: S

> The right panel and perhaps heading info can be used to characterize the page as a grading page. We should save text for critical warning info indicating that the grading state of the question_instance has changed.

Sounds like a good plan. That will happen then.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.ejs
cheeren: So if we're not doing the "You are currently grading..." banner, then this can go.

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.ejs
cheeren: I don't think we do... The idea of a "locked submission" is a little risky imo. I think the sync issues should be handled via communication/warnings, rather than locks, anyway.

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.ejs
cheeren: DRAFT: I'm going to try to get this right, and we'll see if it's helpful. **Note**, each TA can have only one question instance open for grading, and that is the one most recently opened. Every other question instance a TA has opened should be considered stale. Any question instance which is not _currently_ open for grading, including those which are stale, _can_ be open for grading, and thus selected in the "to be graded" list. 

Suppose there are 2 TAs, and TA1 has a qinst open for grading. If TA2 opens the same qinst (somehow) they do so under a warning, but nothing changes in their access ... they can grade and even submit the grade, under the warning. When any TA submits a score, if there has been a state change in the score since the qinst was open for grading, then the TA is offered a choice between the two? scores, and a new submission is created if they insist on their own score. Possibilities for race conditions 1) two TAs insist on their own score (and thus make a new submission) -- this is a human problem and bears conversation. Do we want to facilitate that in any way? 2) both TAs defer and thus no score submission is made, at all. IMO this should just put the qinst back into the list of things to be graded. Is this always a pairwise problem? that is, can we always assume that the choice is between the current grade and the most recently submitted one?

andrewstec: > DRAFT: I'm going to try to get this right, and we'll see if it's helpful. **Note**, each TA can have only one question instance open for grading, and that is the one most recently opened. Every other question instance a TA has opened should be considered stale. Any question instance which is not _currently_ open for grading, including those which are stale, _can_ be open for grading, and thus selected in the "to be graded" list.
> 
> Suppose there are 2 TAs, and TA1 has a qinst open for grading. If TA2 opens the same qinst (somehow) they do so under a warning, but nothing changes in their access ... they can grade and even submit the grade, under the warning. When any TA submits a score, if there has been a state change in the score since they opened the qinst for grading, then the TA is offered a choice between the two? scores, and a new submission is created if they insist on their own score. Possibilities for race conditions 1) two TAs insist on their own score (and thus make a new submission) -- this is a human problem and bears conversation. Do we want to facilitate that in any way? 2) both TAs defer and thus no score submission is made, at all. IMO this should just put the qinst back into the list of things to be graded. Is this always a pairwise problem? that is, can we always assume that the choice is between the current grade and the most recently submitted one?

Regarding your first paragraph, we need to then:

1) Any time a user opens an instance question for manual grading, we can: 

 - remove grading user from any instance questions that have been opened for manual grading by this user and have been ungraded/abandoned
 
But we cannot tell when an instance question is currently open and being graded without adding more logic to send requests to the server on an interval or abandoning a question after 'x' amount of time. Hence, my last comment to Matt on this thread.

2) I would propose that we send a request to the back-end and if the submission has been graded, we can create a modal with the diff grading info. This would require more code though, so @mwest1066 would have to approve, as he is currently looking for less code for this feature.

I think the general idea would be to hope that TAs communicate if they are both grading the same question so they do not run into a conflict in grading. If they both submit a grade, the latter submission overwrites the former submission without any warnings. Perhaps that would be a good warning banner for another user grading the same question?

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.sql
```diff
@@ -7,21 +7,92 @@ SELECT
...
-    (SELECT COUNT(DISTINCT iq.id)
+    (SELECT COUNT(DISTINCT s.id)
      FROM
          instance_questions AS iq
          JOIN variants AS v ON (v.instance_question_id = iq.id)
          JOIN submissions AS s ON (s.variant_id = v.id)
      WHERE
          iq.assessment_question_id = aq.id) AS num_submissions,
-    (SELECT COUNT(DISTINCT iq.id)
+    (SELECT COUNT(DISTINCT s.id)
```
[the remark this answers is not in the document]
andrewstec: Very true. Removed the distinct.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.ejs
andrewstec: Removed.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.ejs
andrewstec: Only left in to indicate when someone else is grading the question.

sprocs/instance_questions_update_for_manual_grading.sql
```diff
@@ -0,0 +1,51 @@
...
+
+    -- Get LAST submissions, filter for ungraded entries to find next ungraded instance question
+    SELECT iq.id
+        FROM instance_questions AS iq
+            JOIN variants AS v ON (v.instance_question_id = iq.id)
+            JOIN submissions AS s ON (s.variant_id = v.id)
+    INTO iq_id
+    WHERE (s.auth_user_id, s.date) 
+        IN (
+            SELECT s.auth_user_id, MAX(s.date)
```
andrewstec, line 25: I still want to change this logic here, but aside from that, everything seems to be working and a lot more simplified. Most of your comments have been overwritten @mwest1066 .

server.js
andrewstec: This endpoint was eliminated, but I implemented convention this for the next_ungraded endpoint.

pages/instructorQuestionManualGrading/instructorQuestionManualGradingUnlockInstanceQuestions.js
andrewstec: This endpoint was eliminated, as the locking feature logic was not necessary.

sprocs/instance_questions_select_manual_grading_objects.sql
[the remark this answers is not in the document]
andrewstec: Arguments now use 'arg_' prefix.

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.sql
```diff
@@ -7,21 +7,92 @@ SELECT
...
+             FROM
+                 submissions AS s
+                 JOIN variants AS v ON (s.variant_id = v.id)
+                 JOIN instance_questions AS iq ON (v.instance_question_id = iq.id)
+             WHERE
+                 iq.assessment_question_id = aq.id
+             GROUP BY s.auth_user_id
+         )
+    ) AS num_submissions_eligible,
+    (SELECT COUNT(DISTINCT IQ.id)
```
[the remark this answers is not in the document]
andrewstec: This logic was simplified. No longer an issue.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.ejs
andrewstec: This logic was updated. If authenticated user is not the manual grading user declared on the particular instance_question row, then the warning message appears.

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.sql
andrewstec: Went a CTE route here, which works much better.

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.ejs
andrewstec: I don't think so.

This is how the view will look now:
image

I pulled out anything unnecessary.

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.ejs
andrewstec: Graded / Ungraded in the end.

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.js
andrewstec: Done.

andrewstec:
Closing and replacing with #3964

PR #3845, opened 2021-02-18
Remove unnecessary reloadElementsForCourse()

mwest1066:
We currently call `reloadElementsForCourse()` towards the end of `syncDiskToSqlWithLock()`, but this is unnecessary because we always call `loadElementsForCourseAsync()` before running any course code and:

1. In production we always check the `commit_hash` for the course and reload elements if it has changed.
2. In dev mode (where there might not be a commit) we call `flushElementCache()` from `syncDiskToSqlWithLock()` which ensures we will always reload all elements.
3. During tests we might have `commit_hash = null` but we now check for this explicitly and always reload elements in this case.

All of the above also applies to course element extensions.

### DIFF
--- a/question-servers/freeform.js
+++ b/question-servers/freeform.js
@@ -137,8 +137,9 @@ module.exports = {
     },
 
     async loadElementsForCourseAsync(course) {
-        if (courseElementsCache[course.id] !== undefined &&
-            courseElementsCache[course.id].commit_hash === course.commit_hash) {
+        if (courseElementsCache[course.id] !== undefined
+            && courseElementsCache[course.id].commit_hash
+            && courseElementsCache[course.id].commit_hash === course.commit_hash) {
             return courseElementsCache[course.id].data;
         }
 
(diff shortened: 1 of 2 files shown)


nicknytko (approved):
This looks safe to me. I'm kind of worried about courses that don't have a commit hash on prod (i.e example and test), but if we're explicitly checking for null we should be good to go?

mwest1066:
> This looks safe to me. I'm kind of worried about courses that don't have a commit hash on prod (i.e example and test), but if we're explicitly checking for null we should be good to go?

Actually all courses on prod have commit hashes (including example course), so I think we are fine. For the example course we actually have a checked-out copy of the full PL source on disk as `/disk1/courses/PrairieLearn` but we set `courses.path` to `/disk1/courses/PrairieLearn/exampleCourse`. This means that doing a sync for XC 101 runs `git pull` in the `courses.path` directory, but this is fine because `git pull` can run in subdirs of git repos and will still pull/access the full repo.

nwalters512 (approved):

Issue #3530, opened 2020-12-13
Configure question params from assessments and courses


mwest1066:
In response to a question from David Eaton, @nwalters512 proposes:

### Default params for questions

1. Add a `questionParams` object to `infoCourse.json`, `infoCourseInstance.json`, and `infoAssessment.json`. For `infoAssessment.json` this could be at the top level, but could also be at the zone or question level.
2. For each question, construct a `unifiedQuestionParams` object by taking the union of all the params, with more specific levels overriding more general levels. That is:
```
unifiedQuestionParams = {...courseQuestionParams, ...courseInstanceQuestionParams, ...assessmentQuestionParams, ...zoneQuestionParams, ...alternativesQuestionParams, ...questionQuestionParams}
```
3. When running `generate()` for a question, seed `data["params"]` with `unifiedQuestionParams` so the code can use it.

### Default params for elements

Can we extend the above scheme to also set parameters for elements? Then you could set `pl-number-input.rtol = 1e-3` in `infoCourse.json` and all `pl-number-input` questions would then default to using that relative tolerance ("rtol") for grading. Maybe we would allow all `info.json` files to contain something like:
```
"elementParams": {
    "pl-number-input": {
        "rtol": 1e-3
    }
}
```

mwest1066:
Suggestions for implementation:

1. Add these new `questionParams` properties to the relevant `info.json` files and sync them to new columns in the database. The steps for this are outlined [here](https://prairielearn.readthedocs.io/en/latest/dev-guide/#json-syncing)
2. During question generation, we need to read the `questionParams` property from all the relevant objects (course, course instance, assessment, etc) and construct `unifiedQuestionParams` as described in the issue description. This can be done in [`questions_select.sql`](https://github.com/PrairieLearn/PrairieLearn/blob/master/sprocs/questions_select.sql) which is called (indirectly) during the call [here](https://github.com/PrairieLearn/PrairieLearn/blob/master/lib/question.js#L1288) to `ensureVariant()` within `getAndRenderVariant()`.
3. Set the default value of `data.params` to be `unifiedQuestionParams` (at the moment the default is `{}`). This is done [here](https://github.com/PrairieLearn/PrairieLearn/blob/master/question-servers/freeform.js#L756)

[...]

PR #3848, opened 2021-02-18
Specifying parameters to questions at course, instance, and assessment level

nealterrell:
Addresses #3530.

This adds support for `questionParams` fields in the JSON files for courses, course instances, and assessments (and its zones, alternative groups, and assessment questions). By adding these fields to one of these objects, individual questions can have their `data["params"]` dictionary pre-filled with values specified by the assessment containing the question, or the course instance that the assessment is offered, or the course itself that owns the question.

Example motivation: in my CS1 course, we teach 12 weeks of Python and 3 weeks of Java. I have a question for identifying the Python primitive type to use for a given scenario description. I'd like to do the same type of skills practice when we move to Java, using the Java primitive type names instead. Currently I have to write two separate questions with the exact same logic; now, I can use `questionParams` to configure the question with a `language` field that is used in `server.py` to display different type names depending on the language.

For example, I can add `"questionParams": {"language": "Python"}` to `infoCourse.json` to set Python as the default language parameter; the server.py sees that in `data["params"]["language"]`, then uses Python type names for the question:

image

When we later move to Java, I add  `"questionParams": {"language": "Java"}` to my `infoAssessment.json`. (I can add it to the question object individually, or to a zone (to pass to all zone question), or to an alternatives group (likewise)). This value overrides the course parameter, so now when a student takes this assessment, the question changes to Java types:

image

Parameters are overridden in the order you would hope: Course < Course instance < Assessment < Zone < Alternatives group < Assessment qusetion.

### Changes

* Added `questionParams` fields to the info files for course, course instance, and assessment
* Synced questionParams to database
* When reading a question from database, merge question params from the available sources. (If only given a question ID, we can merge the course params; if given a question _instance_ ID, merge with course, course instance, and assessment params).
* Changed questions_select to return JSONB instead of questions.

### TODO
* ~~write tests~~
* ~~add to exampleCourse?~~


nealterrell:
I believe this is ready for review.

nealterrell:
Hi @mwest1066, I will bring this PR up to date with main, and then I'd love to get some input on getting it merged!

database/tables/alternative_groups.pg
```diff
@@ -3,6 +3,7 @@ columns
     id: bigint not null default nextval('alternative_groups_id_seq'::regclass)
     number: integer
     number_choose: integer
+    question_params: jsonb not null default '{}'::jsonb
```
jonatanschroeder, line 6: Setting this as `NOT NULL` makes the migration process really slow, and does not provide any meaningful benefit. This should probably be allowed to be NULL, and then set to an empty object as necessary when processing. Same with the equivalent fields in the other tables.

database/tables/assessment_questions.pg
```diff
@@ -32,6 +32,7 @@ columns
     number_submissions_variance: double precision
     points_list: double precision[]
     question_id: bigint not null
+    question_params: jsonb not null default '{}'::jsonb
```
jonatanschroeder, line 35: Should be allowed to be NULL. See comment above.

database/tables/assessments.pg
```diff
@@ -16,6 +16,7 @@ columns
     number: text not null default '0'::text
     obj: jsonb
     order_by: integer
+    question_params: jsonb not null default '{}'::jsonb
```
jonatanschroeder, line 19: Should be allowed to be NULL. See comment above.

database/tables/course_instances.pg
```diff
@@ -6,6 +6,7 @@ columns
     id: bigint not null default nextval('course_instances_id_seq'::regclass)
     long_name: text
     ps_linked: boolean not null default true
+    question_params: jsonb not null default '{}'::jsonb
```
jonatanschroeder, line 9: Should be allowed to be NULL. See comment above.

database/tables/pl_courses.pg
```diff
@@ -10,6 +10,7 @@ columns
     manual_grading_visible: boolean default false
     options: jsonb not null default '{}'::jsonb
     path: text
+    question_params: jsonb not null default '{}'::jsonb
```
jonatanschroeder, line 13: Should be allowed to be NULL. See comment above.

database/tables/zones.pg
```diff
@@ -5,6 +5,7 @@ columns
     max_points: double precision
     number: integer
     number_choose: integer
+    question_params: jsonb not null default '{}'::jsonb
```
jonatanschroeder, line 8: Should be allowed to be NULL. See comment above.

exampleCourse/courseInstances/Sp15/assessments/hm11-questionParameters/infoAssessment.json
```diff
@@ -0,0 +1,31 @@
...
+                {"id": "demo/questionParamsCalculation", "points": 1}
+            ]
+        },
+        {
+            "title": "Questions parameterized by the assessment",
+            "questions": [
+                {"id": "demo/questionParamsRandomSymbolic", "points": 1,
+                 "questionParams": {
+                     "coeff_lower_bound": 3
+                 }}
```
jonatanschroeder, line 26: Nit:
```suggestion
                {
                    "id": "demo/questionParamsRandomSymbolic",
                    "points": 1,
                    "questionParams": {
                        "coeff_lower_bound": 3
                    }
                }
```

exampleCourse/courseInstances/Sp15/assessments/hm11-questionParameters/infoAssessment.json
```diff
@@ -0,0 +1,31 @@
...
+            ]
+        }
+    ],
+    "text": "As an instructor, you will view questions parameterized only by the course they are in. If you view as a student, the questions will inherit parameters from the course instance and assessment."
```
jonatanschroeder, line 30: Not sure if this is the best place to add this comment...

exampleCourse/questions/demo/questionParamsCalculation/question.html
```diff
@@ -0,0 +1,7 @@
+<pl-question-panel>
+  <p> Consider two numbers in the range $\left[ {{params.lower_bound}}, {{params.upper_bound}} \right]$,
```
jonatanschroeder, line 2: I see you are trying to demonstrate adding the question parameters in the question itself, but given these parameters are not relevant to answering the question, this looks a bit confusing. Since this is the example course intended as a demo to instructors, I think it is ok to list these as a meta-comment (something like: "based on questions parameters, $a$ and $b$ are in the range $\left[ {{params.lower_bound}}, {{params.upper_bound}} \right]$.").

migrations/233_pl_course__question_params__add.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE pl_courses ADD COLUMN question_params jsonb NOT NULL DEFAULT '{}'::jsonb;
```
jonatanschroeder, line 1: Does the course actually need question params? I'm not saying it's wrong, I just want a justification since the question is part of the course already anyway. Also, it should be NULLable.

sprocs/questions_select.sql
```diff
@@ -1,15 +1,44 @@
-CREATE FUNCTION
+DROP FUNCTION IF EXISTS questions_select(bigint);
+DROP FUNCTION IF EXISTS questions_select(bigint, bigint);
+
+CREATE OR REPLACE FUNCTION
```
jonatanschroeder, line 4: `DROP FUNCTION` is no longer necessary based on recent changes.

```suggestion
CREATE FUNCTION
```

sprocs/questions_select.sql
```diff
@@ -1,15 +1,44 @@
...
+
+        IF NOT FOUND THEN RAISE EXCEPTION 'no such question_id: %', question_id; END IF;
+    ELSE 
+        SELECT to_jsonb(q.*) 
+            || jsonb_build_object('question_params', pc.question_params 
+                || ci.question_params
+                || aset.question_params 
+                || z.question_params 
+                || ag.question_params 
+                || aq.question_params)
```
jonatanschroeder, line 28: If `question_params` are NULL you can use `COALESCE(??.question_params, '{}'::jsonb)` instead.

sync/fromDisk/assessments.js
```diff
@@ -71,6 +71,7 @@ function getParamsForAssessment(assessmentInfoFile, questionIds) {
         student_group_create: !!assessment.studentGroupCreate,
         student_group_join: !!assessment.studentGroupJoin,
         student_group_leave: !!assessment.studentGroupLeave,
+        question_params: assessment.questionParams || {},
```
jonatanschroeder, line 74: Can be null in this case.

tests/sync/util.js
```diff
@@ -95,7 +97,7 @@ const syncFromDisk = require('../../sync/syncFromDisk');
   * @property {numer | number[]} maxPoints
   * @property {string} id
   * @property {boolean} forceMaxPoints
-  * @property {number} triesPerVariant
```
jonatanschroeder, line 98: I believe this was removed by mistake?

tests/sync/util.js
```diff
@@ -115,7 +118,7 @@ const syncFromDisk = require('../../sync/syncFromDisk');
  * @property {number} maxPoints
  * @property {number} maxChoose
  * @property {number} bestQuestions
- * @property {ZoneQuestion[]} questions
```
jonatanschroeder, line 118: Was this removed by mistake?

schemas/schemas/infoQuestion.json
```diff
@@ -96,6 +96,10 @@
             "description": "Options that define how the question will work, specific to the individual question type.",
             "type": "object"
         },
+        "parameters": {
+            "description": "Information on parameters that the question is prepared to accept from questionParams sources.",
+            "type": "object"
+        },
```
jonatanschroeder, line 102: I may be mistaken, but I don't see this being used anywhere, are you planning to filter the parameters for a question?

jonatanschroeder (commented):
This is a preliminary review of the code only, I haven't had a chance to test it yet, but I believe these are things you can address and take a look. Also, I find it is a good idea to draft the documentation at this stage, to make sure the use of it makes sense from a user's point of view (sometimes the way you write the documentation can affect how you implement the code).

One minor issue that's bugging me is that you use a pretty generic parameter name like "`lower_bound`" in your examples, I think a good practice would be to make it clearer regarding what it represent, at least to encourage good practice from instructor (like `sum_operand_lower_bound`, or something like that), since the same name may be used by several unrelated questions.

This may be outside of the scope of this PR, but this kind of parameter-setting may be a good starting point to discuss the possibility of removing restrictions so that the same question can be added to an assessment more than once. @mwest1066 thoughts?

This may be a discussion for when this is close to being implemented, but this sounds like a good feature to add a news item about.

database/tables/alternative_groups.pg
nealterrell: Got it!

exampleCourse/courseInstances/Sp15/assessments/hm11-questionParameters/infoAssessment.json
nealterrell: Agreed, will work this into formal documentation.

exampleCourse/questions/demo/questionParamsCalculation/question.html
nealterrell: I'm not sure if you had something specific in mind with "meta-comment", but I did add this as a <p> after the question text.

migrations/233_pl_course__question_params__add.sql
nealterrell: I added this based on the plan that was laid out in #3530. I imagine the justification is of convenience, allowing an instructor to change the default value for some common parameter that they deliberately use in many questions. For example, many of my questions choose random scenarios from JSON files of data. Those questions could be programmed to read a JSON filename from params and load scenarios from that file; then, when moving to a new semester, I could add new files for that semester and change the course parameter in order to cycle out old scenarios.

A stretch, perhaps (if I'm already adding a "fall2011.json" file to each of those questions, why not just change the server.py default value while I'm at it?), but I'm sure there are other uses.

sprocs/questions_select.sql
nealterrell: Sorry, should this be `CREATE FUNCTION` or `CREATE OR REPLACE FUNCTION`? My DROP statements were based on an older version of the file and I wasn't sure how to incorporate the changes.

sprocs/questions_select.sql
nealterrell: I *really* don't like that `{some jsonb} || NULL` gives `NULL`. Alas, this is fixed.

migrations/233_pl_course__question_params__add.sql
jonatanschroeder: That makes sense, though your example makes a good use-case for course instance, not course. Anyway, using your examples, I'd suggest adding in the documentation an example where a set of questions may be copied between courses, and keeping the source may make these questions easier to migrate.

schemas/schemas/infoQuestion.json
nealterrell: I think I was experimenting with having questions specify what parameters they recognized, but that ended up being very tedious.

sprocs/questions_select.sql
jonatanschroeder: No need for replace in new versions, see other sprocs in master.

sync/fromDisk/assessments.js
nealterrell: Unsure; do you mean `assessment.questionParams || null`?

jonatanschroeder: I mean it can be just `assessment.questionParams`, since if it's null it could be used as is. But I'm ok with the original too.

docs/question.md
```diff
@@ -447,3 +447,84 @@ For most [elements] there are four different ways of grading the student answer.
...
+
+1. `infoCourse.json`: parameters defined here will be passed to all questions in the course, including when they are previewed by an instructor without being attached to an assessment.
+2. `infoCourseInstance.json`: these parameters will be passed to questions when they are used in one of this course instance's assessments.
+3. `infoAssessment.json`: parameters can be added to the assessment itself, to one of its zones, to one of its lists of alternatives, or to an individual question object. In each case, the parameters will be passed to all questions nested under that element. For example, parameters specified at the top-level assessment apply to every question; paramters for a zone only apply to the questions in that zone; and parameters on an individual question only apply to that question.
+
+Parameters with matching names are overridden by the locations that are farther down the preceding list. (A parameter configured by `infoCourse.json` will be supplanted by one configured in `infoCourseInstance.json`, parameters on an individual question in an assessment will override all others.)
```
nwalters512, line 461: It might be worth calling out explicitly that these are not deep merged - that is, if I have `{"a": { "b": 1 } }` at the course level and `{"a": { "c": 2 } }` at the question level, you end up with `{"a": { "c": 2 } }`, not `{"a": { "b": 1, "c": 2 } }`.

docs/question.md
```diff
@@ -447,3 +447,84 @@ For most [elements] there are four different ways of grading the student answer.
...
+
+Example of a question's `server.py` that can be customized with `questionParams`:
+
+```python
+import random
+def generate(data):
+  param_names = data['params'].get('names', [{'name': 'Matt', 'photo': '001'}, {'name': 'Craig', 'photo': '002'}])
```
nwalters512, line 468: I think we should document that folks should almost always use `.get()` with a default when accessing question params so that the question will work in all contexts.

exampleCourse/questions/demo/questionParamsRandomSymbolic/server.py
```diff
@@ -0,0 +1,33 @@
...
+
+    # Modify data and return
+    data["params"]["x"] = sympy.latex(x)
+    data["params"]["f"] = sympy.latex(f)
+    data["params"]["coeff_lower_bound"] = coeff_lower_bound
+    data["params"]["coeff_upper_bound"] = coeff_upper_bound
+    data["correct_answers"]["df"] = str(df)
+
+if __name__ == "__main__":
+    generate({"params":{},"correct_answers":{}})
```
nwalters512, line 33: Let's leave this out - we don't want this to be featured in the example course. It looks like you may have copied this from another question, in which case we should remove it from there too.

```suggestion
```

migrations/245_pl_course__question_params__add.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE pl_courses ADD COLUMN question_params jsonb DEFAULT '{}'::jsonb;
```
nwalters512, line 1: IMO we should batch all these up into a single file, unless there's a particular reason they're separate now.

schemas/schemas/infoAssessment.json
```diff
@@ -351,6 +366,11 @@
                     "type": "number",
                     "default": 1
                 },
+                "questionParams": {
+                    "description": "Parameters to pass down to this question.",
```
nwalters512, line 370: Nit:
```suggestion
                    "description": "Parameters to pass to this question.",
```

sprocs/sync_course_instances.sql
```diff
@@ -20,7 +20,7 @@ BEGIN
     -- Move all our data into a temporary table so it's easier to work with
 
     CREATE TEMPORARY TABLE disk_course_instances (
-        short_name TEXT NOT NULL,
+        short_name TEXT NOT NULL,
```
nwalters512, line 23:
```suggestion
        short_name TEXT NOT NULL,
```

tests/testQuestionParams.js
```diff
@@ -0,0 +1,336 @@
...
+    it('should contain questions', function (callback) {
+      sqldb.query(sql.select_questions, [], function (err, result) {
+        if (ERR(err, callback)) return;
+        if (result.rowCount === 0) {
+          return callback(new Error('no questions in DB'));
+        }
+        locals.questions = result.rows;
+        callback(null);
+      });
+    });
```
nwalters512, line 55: You should use `async`/`await` here instead of callbacks. Ditto for the rest of the tests.

tests/testQuestionParams.js
```diff
@@ -0,0 +1,336 @@
...
+      it('should be initialized', function () {
+        locals.siteUrl = 'http://localhost:' + config.serverPort;
+        locals.baseUrl = locals.siteUrl + '/pl';
+        locals.courseInstanceBaseUrl = locals.baseUrl + '/course_instance/1';
+        locals.questionBaseUrl = locals.courseInstanceBaseUrl + '/instance_question';
+        locals.assessmentsUrl = locals.courseInstanceBaseUrl + '/assessments';
+        locals.isStudentPage = true;
+        locals.totalPoints = 0;
+      });
+    });
```
nwalters512, line 101: Nit: I suspect you copied something like this from existing tests, but there are no assertions here, so these really shouldn't be in `describe`/`it` blocks.

tests/testQuestionParams.js
```diff
@@ -0,0 +1,336 @@
...
+        assert.equal(
+          locals.assessmentUrl,
+          locals.courseInstanceBaseUrl + '/assessment/' + locals.assessment_id + '/'
+        );
+      });
+    });
+
+    describe('GET to assessment URL', function () {
+      it('should load successfully', function (callback) {
+        locals.preStartTime = Date.now();
```
nwalters512, line 158: I don't see `preStartTime` or `postStartTime` (below) used anywhere.

tests/testQuestionParams.js
```diff
@@ -0,0 +1,336 @@
...
+    });
+
+    it('should parse', function () {
+      locals.$ = cheerio.load(page);
+    });
+
+    it('should have a range set by the assessment question questionParams', function () {
+      elemList = locals.$('span:contains("[20, 30]")');
+      assert.lengthOf(elemList, 1);
+    });
```
nwalters512, line 334: After getting to the end of this file it's looking like this is just a copy/paste of existing assessment tests, so take my comments here with a grain of salt. That said, while it's unfortunate that old tests aren't written how we'd like, I'd like new tests like this to follow best practices (use async/await, `it` blocks should contain assertions and not just logic, etc.)

We should really refactor our tests so we can share boilerplate like `startAssessment` and such, but that's outside what I'd expect you to do for this PR.

docs/question.md
```diff
@@ -447,3 +447,90 @@ For most [elements] there are four different ways of grading the student answer.
...
+2. `infoCourseInstance.json`: these parameters will be passed to questions when they are used in one of this course instance's assessments.
+3. `infoAssessment.json`: parameters can be added to the assessment itself, to one of its zones, to one of its lists of alternatives, or to an individual question object. In each case, the parameters will be passed to all questions nested under that element. For example, parameters specified at the top-level assessment apply to every question; paramters for a zone only apply to the questions in that zone; and parameters on an individual question only apply to that question.
+
+Parameters with matching names are overridden by the locations that are farther down the preceding list. (A parameter configured by `infoCourse.json` will be supplanted by one configured in `infoCourseInstance.json`, parameters on an individual question in an assessment will override all others.) Parameters are **not** deep merged -- a course with parameter `{"a": { "b": 1 } }` and an assessment with parameter `{"a": { "c": 2 } }` will pass `{"a": { "c": 2 } }` to the question, not `{"a": { "b": 1, "c": 2 } }`.
```
nwalters512, line 461: Nits:
```suggestion
Parameters with matching names are overridden by the locations that are farther down the preceding list. That is, a parameter configured by `infoCourse.json` will be supplanted by one configured in `infoCourseInstance.json`, and parameters on an individual question in an assessment will override all others. Parameters are **not** deep merged: a course with parameter `{"a": { "b": 1 } }` and an assessment with parameter `{"a": { "c": 2 } }` will pass `{"a": { "c": 2 } }` to the question, not `{"a": { "b": 1, "c": 2 } }`.
```

docs/question.md
```diff
@@ -447,3 +447,90 @@ For most [elements] there are four different ways of grading the student answer.
...
+
+A `questionParams` is a JSON object that can be placed in three different PrairieLearn file types. When a question's `generate()` function is called, the `data["params"]` dictionary will be pre-filled with the key-value pairs defined by `questionParams` in these locations:
+
+1. `infoCourse.json`: parameters defined here will be passed to all questions in the course, including when they are previewed by an instructor without being attached to an assessment.
+2. `infoCourseInstance.json`: these parameters will be passed to questions when they are used in one of this course instance's assessments.
+3. `infoAssessment.json`: parameters can be added to the assessment itself, to one of its zones, to one of its lists of alternatives, or to an individual question object. In each case, the parameters will be passed to all questions nested under that element. For example, parameters specified at the top-level assessment apply to every question; paramters for a zone only apply to the questions in that zone; and parameters on an individual question only apply to that question.
```
mwest1066, line 459:
```suggestion
3. `infoAssessment.json`: parameters can be added to the assessment itself, to one of its zones, to one of its lists of alternatives, or to an individual question object. In each case, the parameters will be passed to all questions nested under that element. For example, parameters specified at the top-level assessment apply to every question; parameters for a zone only apply to the questions in that zone; and parameters on an individual question only apply to that question.
```

docs/question.md
```diff
@@ -447,3 +447,90 @@ For most [elements] there are four different ways of grading the student answer.
...
+}
+```
+
+Doing this does not change the question in the original assessment, which would not be the case if the `server.py` was updated with the new list of names.
+
+Question params can reduce the need to duplicate questions in order to change the data source of its variation. This makes maintenance of the questions easier. Examples:
+
+- A "match each nation with its capital city" question can use a different set of nations depending on what region has been studied most recently. (Instead of different questions for each region.)
+- A question on programming language syntax errors can involve more examples of errors as new features with different syntax rules are learned. (Instead of different questions for later in the course.)
+- A question that generates an indefinite integrals can use simple trig functions on an early calculus exam, and later use inverse trig functions. (Instead of two different questions that only vary in the chosen function.)
```
mwest1066, line 535:
```suggestion
- A question that generates an indefinite integral can use simple trig functions on an early calculus exam, and later use inverse trig functions. (Instead of two different questions that only vary in the chosen function.)
```

docs/question.md
```diff
@@ -447,3 +447,90 @@ For most [elements] there are four different ways of grading the student answer.
...
+
+## Initializing parameters from outside the question
+
+The `data["params"]` dictionary normally used to pass parameters to `question.html` can be "seeded" from outside the question's files, allowing some portion of its randomization to be set by an assessment, course instance, or course configuration file. Such "`questionParams`" can change the behavior of a question without needing to alter its source code, so a question can be reused on different assessments but with a different configuration of parameters to alter its actual variants as needed.
+
+A `questionParams` is a JSON object that can be placed in three different PrairieLearn file types. When a question's `generate()` function is called, the `data["params"]` dictionary will be pre-filled with the key-value pairs defined by `questionParams` in these locations:
+
+1. `infoCourse.json`: parameters defined here will be passed to all questions in the course, including when they are previewed by an instructor without being attached to an assessment.
```
mwest1066, line 457: I'm concerned that `questionParams` set in `infoCourse.json` will "pollute" the `params` of all questions. That is, every variant generated in the course will include the course `questionParams` in its stored `params`. For example, in the example below where course instructor names are set in the `infoCourseInstance.json` question params, I think that every variant for every question in that course instance would store the instructor names?

I'm not really worried about the storage, it's more that it will be messy when we display the params for variants and for most questions there are these spurious entries.

A couple of ideas off the top of my head to address this:

1. Use `questionOptions` rather than `questionParams` and merge into `data["options"]`. This is not stored on a per-variant basis. If a question wants to store the data, they can copy it from `options` into `params`.

2. Require questions to opt-in to particular question params by listing them in their `info.json` file. For example, a question would need to declare `"useQuestionParams": ["names"]` in its `info.json` file to have the `names` question param copied into its local `params`.

nealterrell: Yes, I understand your concern. It's ultimately your team's call, both solutions sound easy enough for me to implement. Here are my thoughts:

1. This is nice for more clearly differentiating "params" (which currently only come from server.py, right?) from "options" (which are passed in from external sources). I think most use of "options" will be to set bounds/configurations on the variants that are generated (at least this is true in my head), and those don't really need to be visible.  As you say, we can copy options to params if we want them visible/stored.
2. This is also nice for documenting the ways a question can be configured.

mwest1066: @nwalters512 thoughts about this?

nwalters512: @nealterrell Matt and just chatted about this for an hour or so. We've agreed that something more like option 2 is the way we want to go, i.e. use params that are persisted as part of a variant (not options) and require questions to declare somehow that they will or can use a certain parameter. We'd like to spend a little bit longer talking about this, as the design will have implications for other things that are part of our longer-term vision for PL, and we want to make sure that we're at least foreseeing potential problems now. Sorry the complexity is snowballing here, but we want to be really deliberate with how we're designing this.

Matt is heading off to Learning@Scale for a few days so we probably won't be able to talk about this more til next week. In the meantime, here are some of the things we're thinking about:

- Someday, we'll allow folks to share questions between courses or institutions. How can we make sure we're guiding people towards writing sharable questions in the context of params, which may not be present in a course that a question is shared with? We're thinking about some way to declare defaults in `info.json`, and possibly making that mandatory.
  - The alternative is of course `.get('...', DEFAULT)` in Python, but if your course or course instance already has that value in place, it would be easy to forget that you should specify a default value as you'd never hit the case of a missing param during question development.
- In the same vein as defaults, having declarative information about params in `info.json` would allow us to do things like make an editor or preview UI to show which params will take effect, or to layer on some kind of schema checking, and so on. We probably want something more than just a list of param names, but how much more? Maybe `{"foo": {"default: 5}}`? Or `[{"name": "foo", "default": 5}]`? Maybe we just use [JSON Schema](https://json-schema.org/) directly?
- What do we actually want to call these things? Calling them just "params" inside a question might be a bit confusing, as they're different from the params that a question would write into `params` during `generate`. Maybe we call them "inherited params" or "external params"? @mwest1066 might be able to expand on this thought a bit.

Feel free to chime in with your thoughts about any/all of the above. I'll keep you posted with what Matt and I chat about next week.

eliotwrobson: To chime in briefly about this, having the "params" field in the data dict copied into could cause confusion in designing the backend for a question. The easiest thing (IMO) would be to have a separate entry in the data dictionary (`questionParams`, for example) that only holds options taken from the `info.json` (similar to option 2). That way, we can add `questionParams` to the type signatures for the data dict, and it makes it easier to declare a custom dictionary for just those parameters. This seems like a really cool feature, and something that is a good stepping stone to question sharing across courses.

mwest1066: I'm certainly not opposed to having these "external" params copied into a separate entry in `data`. What to call it is not so clear to me, however 😄. `questionParams` seems like a good name to use in `infoAssessment.json`, but I feel like it would be weird to have both `data['params']` and `data['questionParams']`.

And @nwalters512 and I really need to figure this out, we'd love to get this feature merged.

echuber2: `explicitParams`? `specParams`? `configuredParams`?

eliotwrobson: Maybe `assignmentParams`? The whole point is that these parameters are fed in externally from the assignment configuration.

echuber2: The word "external" is coming up a lot, so I'd also suggest `externalParams`.

eliotwrobson: I think I agree with @echuber2, the parameters can come from the course, instance, or assignment level, so `externalParams` seems more natural. @nwalters512 what do you think?

nwalters512: I'm a big fan of `externalParams`!

eliotwrobson: With respect to declaring that information in the `info.json`, it seems like having a dictionary of the form `{"foo": {"default": 5}}` as Nathan mentioned would be best, since dict keys have to be distinct (as opposed to the second list suggestion). So the entry `externalParams` in the `info.json` for a question would hold that dict, and then the course + assignment + whatever could override it as needed (with an error being thrown if a parameter is injected that isn't on the question's approved list). Handling this with the json as opposed to just in the Python backend, the error message will be thrown as soon as the assignment configuration including the question is created, as opposed to when the question is first opened.

Also I know this has been sitting around for a while, but @nealterrell let us know if you'd like any help implementing this, since our team has grown enough that we can offer some assistance if needed, and this would be a useful feature for us.

nealterrell: @eliotwrobson I am able to work on the trivial stuff like renaming the JSON fields, but the plans for parameter schemas and validation probably exceeds my bandwidth at the moment.

eliotwrobson: Understood, I'll get someone on that piece of this as soon as I can. The parameter validation shouldn't be _too_ difficult, since you already have the parameter injection set up. Will follow up later if we run into issues.

nwalters512:
I had a great conversation with some Berkeley folks that's really relevant to the idea of external question params; sharing more details here:

The specific problem they're trying to solve is that they want to be able to essentially force a question into a particular state for QAing. There's no good way to do that now short of clicking "New variant" a bunch of times and praying that you get the variant you're looking for.

I proposed external question params as a possible solution to this. Since we're leaning in the direction of having questions declare their parameters in `question.json`, we could automatically generate a nice little UI for the instructor question preview where folks could pick specific params for their question. With a small adjustment to question code to be aware of these params, they could then pick specific values to use for this particular variant.

While I don't think such a UI should be a part of this PR, we should keep that in mind as a use case; it should then be pretty trivial to build this additional functionality on top of it.

mfbutner:
@mwest1066 Pointed me to this PR because I had an issue that was adjacent to the one that you discussed. My problem is that I have some questions where there is some "global constant" information that needs to be shared across multiple locations (serve.py, question.html, and the external grader) in the question. The most common piece of information is what files the students should be submitting. While I know I could copy this information around, it would be ideal to have it only in a single location so as to make it easier to change. 

The only change that would be needed to support this based on what you are already doing is to extend it down to the question level as well.

mwest1066:
As @eliotwrobson pointed out, with this feature it would also be very helpful to implement #9781.

eliotwrobson:
Closing in favor of #11181, as this would need to be rebased. Discussion will continue there.

PR #3861, opened 2021-02-22
 Add answer feedback for checkbox element

ashnaarya:
Add pl-feedback child


nwalters512:
Feedback for `<pl-checkbox>` answers was implemented in https://github.com/PrairieLearn/PrairieLearn/pull/4917.

PR #3865, opened 2021-02-22
Add element that displays manual feedback to students

jonatanschroeder:
This is an element to show manual feedback to students. Although there is currently a method to show it, this element makes it cleaner and more straightforward, particularly for instructors with limited technical skills.


bevansm:
In order to bring the instructor view to parity with the student view, should we be add a preview tab to the manual grading method on the question (i.e. a little modal popup/inline preview that pre-renders the given feedback as html/markdown)?

jonatanschroeder:
> In order to bring the instructor view to parity with the student view, should we be add a preview tab to the manual grading method on the question (i.e. a little modal popup/inline preview that pre-renders the given feedback as html/markdown)?

I imagine you mean the page for graders to submit their grades, which would certainly be interesting. Probably something similar to the previewer in pl-file-editor?

bevansm:
> > In order to bring the instructor view to parity with the student view, should we be add a preview tab to the manual grading method on the question (i.e. a little modal popup/inline preview that pre-renders the given feedback as html/markdown)?
> 
> I imagine you mean the page for graders to submit their grades, which would certainly be interesting. Probably something similar to the previewer in pl-file-editor?

Yes! Probably a future PR though.

docs/elements.md
```diff
@@ -1747,6 +1748,37 @@ It expects results to follow [the reference schema for external grading results]
 
 - [External Grading Reference Schema](externalGrading.md#grading-result)
 
+
+-----
+
+
+### `pl-feedback` element
```
mwest1066, line 1755: What's our thinking about manual grading feedback from in-browser manual grading? I feel like it would be best if instructors don't have to put a `<pl-feedback>` element into their manually-graded questions, but instead we just display the feedback automatically.

Are we thinking that it's still helpful to have this element for grade upload feedback?

jonatanschroeder: You're probably right, but for grade upload you still need something. I'm ok with closing this altogether and discussing this in the context of manual grading, since that may be more long-term plan. I have this element in my courses anyway, in the meantime.

mwest1066: Yeah, I'm thinking that we probably want to handle manual feedback (at least from in-browser grading) outside of the regular element flow. Let's close this for now and think harder about manual grading.

PR #3868, opened 2021-02-23
Fix chunk symlinks to ensure overwriting old links

mwest1066:
The `ensureSymlink()` function silently does nothing if the destination already exists, even if it isn't the correct symlink (see jprichardson/node-fs-extra#786, jprichardson/node-fs-extra#869, jprichardson/node-fs-extra#826). This means that with the new symlink chunk code (see #3354) we will never update a question chunk, even when there is a new version of the question available. To work around this we create the link with a temporary name and move it to the correct link name.

### DIFF
--- a/lib/chunks.js
+++ b/lib/chunks.js
@@ -701,7 +701,15 @@ const ensureChunk = async (courseId, chunk) => {
     });
 
     // Finally, link targetPath -> relativeUnpackPath
-    await fs.ensureSymlink(relativeUnpackPath, targetPath);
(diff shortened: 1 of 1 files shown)


nicknytko (approved):
Just as a quick sanity check, `rename()` will overwrite any existing file if `targetPath` already exists, right?

mwest1066:
> Just as a quick sanity check, `rename()` will overwrite any existing file if `targetPath` already exists, right?

Yep, I tested it because I no longer trust anything 😄

mwest1066:
Successfully tested on staging.

PR #3877, opened 2021-02-26
Fix templated v2 questions on chunk servers

mwest1066:
v2 questions can have a `template_directory` specified which should be the QID of another question. This means that any file missing from the question should instead be taken from the template question directory. This can operate recursively.

This means that v2 questions need to pull not only the chunks for the question itself, but also for any template question, and the template of the template, etc.


pages/legacyQuestionFile/legacyQuestionFile.js
```diff
@@ -25,16 +26,24 @@ router.get('/:filename', function(req, res, next) {
         if (!result.rows[0].access_allowed) return next(error.make(403, 'Access denied', {locals: res.locals, filename: filename}));
 
         const coursePath = chunks.getRuntimeDirectoryForCourse(course);
-        const chunk = {
-            'type': 'question',
-            'questionId': question.id,
-        };
-        chunks.ensureChunksForCourse(course.id, chunk, (err) => {
+
+        sqldb.call('questions_select_templates', [question.id], (err, result) => {
```
nwalters512, line 30: Nit: we could probably avoid this call if `!question.template_directory`, right? Probably not worth the effort though, especially since this is implemented with callbacks.

sprocs/questions_select_templates.sql
```diff
@@ -0,0 +1,27 @@
...
+    FROM
+        questions AS q
+        JOIN questions AS tq ON (tq.qid = q.template_directory AND tq.course_id = q.course_id)
+    WHERE q.id = question_id
+    -- required UNION for a recursive WITH statement
+    UNION
+    -- recursive term that references template_questions again
+    SELECT tq.*
+    FROM
+        template_questions AS q
```
nwalters512, line 23: Whoa - am I reading correctly that a recursive function is implemented as a sort of pseudo-table? Never seen a recursive PG function before!

nwalters512 (approved):
Did not personally test, but the code looks reasonable to me!

pages/legacyQuestionFile/legacyQuestionFile.js
mwest1066: Good point, I did this.

sprocs/questions_select_templates.sql
mwest1066: Yep! They are pretty neat. They are perfect for self-referential data structures like graphs (or the `question->template_question->template_question->...` pointer chasing). Traditionally graphs are a real pain to store in SQL, but with these recursive queries they become straightforward.

PR #3947, opened 2021-03-04
Manual Grading Parallel Workflows Split

bevansm:
By taking advantage of how we handle grading jobs, we can effectively use them to triage grading requests.

Note that this is very much a draft--I want to pull in Andrew's work from #3836 so I can look at what procs I might need to alter so we're basing the "things we need to grade" off of the grading_jobs rather than the submissions/variants column. 

### Backend
- [ ] Update manual grading page procs to look for grading jobs from the jobs table rather than looking at variants
- [ ] Update TA overlap procs to handle timestamp updates in the grading jobs and interact with the jobs table nicely
- [ ] Add proper `grading_job_*_at` timestamps based upon TA interaction with different endpoints (ties into #3836)
- [ ] Add warnings/resolve issues when we have a manual grading job queued, but external job hasn't processed yet 
   - NOTE: Personally prefer disabling anything where we have an external grading job, or show a warning, and display it in a "preparing for manual grading" bucket on the main manual grading page
   - No need to try and staple manual grading as the "step after external grading" or deal with complicated pipelining issues
- [ ] Make sure that we're feeding `beginGradingJob` external grading jobs, since we can no longer implicitly assume an "ungraded" grading job is an external grading job.
### UI
- [ ] Display submit button for manually graded questions, but alert student that the question may be manually graded
- [ ] Show final grade as pending until all grading jobs are complete for a certain question
- [ ] Add external grading display to any question with external grading 
#### Integration With Current Features
- [ ] Do we ever allow instructors to automatically open scores once everything is graded? Do we need to add extra checks that all grading jobs for a given assessment are completed, or are the current checks agnostic to us adding more grading types?
- [ ] Will this play nicely with regrades? Do we have to worry about anything messing with the current workflow, or can we assume that as long as we cancel the jobs and queue new ones, everything will be fine?
- [ ] What's the best way to test this (we have to deal with external jobs, and none of the current questions from the test course have external jobs)?
#### Design/Conventions
- [ ] Are we happy with an array option on the config, or do we want to go with explicit booleans for each sort of grading type?
- [ ] Are we releasing parallel grading workflows in any sort of docs at this time, or should we only talk about (Internal OR External) + Manual?

## Future Work
- [ ] Look at feeding in current scores/partials into external graders in `externalGrader.beginGradingJob` (assume that external graders don't have access rn)
- [ ] Clean up Student UI for questions w/ multiple grading types
   - may be implicit, but UI may become confusing b/c students may look for their autograder at the bottom of the page, but overwritten score/compound score may be available elsewhere
   - Consider displaying side by side or more dropdown menus when there's multiple types of grading?
- [ ] Kick people off the GET endpoint for the manual grading page(s) if there's no jobs for that variant/submission (currently, errors on post, but don't want people accidently spending hours grading smth just for it to fail)
   - Currently no direct link to this page in this case, so it's probably fine, but someone is going to use their own hardcoded links and be disappointed


sprocs/instance_question_select_manual_grading_objects.sql
```diff
@@ -2,7 +2,16 @@
...
-
+-- TODO: Move this query to grab the most recent manual grading job for a given submission?
+--       Should we add the variant_id onto the grading_jobs table, and then query the variant ID
+--          off of that?   
+--       Should we be only presenting a question if it doesn't have an outstanding external grading
+--          job, or should we bubble up completion messages for the job?
+--       No, we should be basing this off of the submission. We should find the most recent submission
+--          for the given question (as we do), but instead of doing fun queries, we can do it based upon
+--          the most recently submitted manual grading job.
+--       We should let someone submit a manual grade, but we should warn them if an external job is hanging.
```
andrewstec, line 13: - This query can quite quickly be modified to include grading jobs without any issues.
- Warning message if the external grader is hanging. We may have to learn about the external grader and how it fails to manage this. Don't know if we should include it in this or find a simpler way to work around this issue for now to avoid the added complexity.
- In the new logic, we always look for the last submission for an instance question and gather the necessary dependency objects based off of that, so it looks like some of this work has been done for you. We might need to merge our branches soon.

PR #3962, opened 2021-03-11
Remove uploaded files from instructor assessment instance log

### LINKED ISSUE #2168: Snipping File Contents in Log
We should trim the "_files" "contents" output in the log for file upload questions.  (The one where you go to see student submissions, when they generated a variant, etc.)

Some of the files students upload are rather large, and looking through the log entails scrolling through pages of meaningless letters.  We could instead just display some or none of it and then put a [snipped].  We could have a "show remainder" link people could click if they wanted to see the rest.

mwest1066:
Until #3035 or equivalent is merged we are storing files as JSONB in the DB. This produces problems when viewing the assessment instance log because the files are too big. Recent math exams have been having students upload large PDFs which break the rendering of the assessment instance page because of this problem.

To fix the problem, this PR strips the files from the submitted answer JSON when getting the log. This is the same technique used when downloading submissions: https://github.com/PrairieLearn/PrairieLearn/blob/a4380b697baa6db8a288d35995424c833f363567/pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql#L111 This PR only conditionally strips the files. It strips them for the HTML page and the corresponding CSV download, but it keeps them for the API access to the log.

Fixes #2168

### DIFF
--- a/api/v1/endpoints/courseInstanceAssessmentInstances/index.js
+++ b/api/v1/endpoints/courseInstanceAssessmentInstances/index.js
@@ -54,7 +54,7 @@ router.get('/:assessment_instance_id/submissions', (req, res, next) => {
 });
 
 router.get('/:assessment_instance_id/log', (req, res, next) => {
-    const params = [req.params.assessment_instance_id];
+    const params = [req.params.assessment_instance_id, true];
     sqldb.call('assessment_instances_select_log', params, (err, result) => {
         if (ERR(err, next)) return;
         res.status(200).send(result.rows);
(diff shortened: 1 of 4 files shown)


jonatanschroeder:
This handles file uploads, but not file editors, which include a `_file_editor_{uuid}` field in the `submitted_answer`, with a value corresponding to the base-64 encoding of the set file. Should that be removed as well? Potentially from the CSV too?

mwest1066:
> This handles file uploads, but not file editors, which include a `_file_editor_{uuid}` field in the `submitted_answer`, with a value corresponding to the base-64 encoding of the set file. Should that be removed as well? Potentially from the CSV too?

Good point! I'll defer that to after #3035 is implemented because I don't have a really simple way of stripping the file editor properties at the moment. Also, the urgent problem that this PR fixes is the case where very large files break page/CSV rendering, which I think isn't (typically) caused by file editors?

PR #3966, opened 2021-03-14
Remove superfluous "* 2" on color palette

echuber2:
The palette shown already specifies all 16 colors explicitly, so there's no need to duplicate to get the second 8 (the "bright" set). Coincidentally, the library automatically generates a "bright" variant of the first eight colors for the second eight, if they aren't already specified.

By the way, the built-in "xterm" palette is very similar to the one defined here.


echuber2:
I'm guessing the `*2` is vestigial based on the example mentioned in this other PR: https://github.com/pycontribs/ansi2html/pull/126#issuecomment-798897672

PR #3972, opened 2021-03-16
Add support for group assessments in score upload

### LINKED ISSUE #3480: Support for group submissions in assessment grade upload
The Upload option for questions supports the Groupname column, the Upload option for total scores doesn't (at least not with this column name, and no other documented name is listed).

jonatanschroeder:
Fixes #3480 .

Quick note: I used "groupname" instead of "group_name" because that is what is in use in the file downloads. I'm happy to switch to "group_name" if that looks better.


lib/score-upload.js
```diff
@@ -334,16 +334,27 @@ module.exports = {
     },
 
     _updateAssessmentInstanceFromJson(json, assessment_id, authn_user_id, callback) {
-        if (!_.has(json, 'uid')) return callback(new Error('"uid" not found'));
+        let query, id;
+        if (_.has(json, 'uid')) {
+            query = sql.select_assessment_instance_uid;
+            id = json.uid;
+        }
+        else if (_.has(json, 'groupname')) {
```
mwest1066, line 342: Style nit: `else` on same line as `}`

lib/score-upload.js
```diff
@@ -334,16 +334,27 @@ module.exports = {
...
+            query = sql.select_assessment_instance_group;
+            id = json.groupname;
+        }
+        else
+            return callback(new Error('"uid" or "groupname" not found'));
```
mwest1066, line 347: Style nit: use braces unless it's a single-line `if` statement

pages/instructorAssessmentUploads/csvHelpAssessmentInstanceScores.ejs
```diff
@@ -6,15 +6,21 @@ mwest@illinois.edu,1,63.5
...
 <p>
   Alternatively, the total assessment points can be changed with a CSV
-  containing a <tt>points</tt> column, like:
+  containing a <code>points</code> column, like:
 </p>
 <pre class="ml-4">uid,instance,points
 mwest@illinois.edu,1,120
 zilles@illinois.edu,1,130.27</pre>
+<p>
+  For assessments using groupwork, use the <code>groupname</code> column instead:
```
mwest1066, line 22:
```suggestion
  For assessments using group work, use the <code>group_name</code> column instead:
```

mwest1066:
> Quick note: I used "groupname" instead of "group_name" because that is what is in use in the file downloads. I'm happy to switch to "group_name" if that looks better.

Thanks for pointing this out! Let's switch to `group_name`. Changing this in the file downloads would also be great. I tried to catch all of these in the various group work PRs but I clearly missed some of them 😄

jonatanschroeder:
> Thanks for pointing this out! Let's switch to `group_name`. Changing this in the file downloads would also be great. I tried to catch all of these in the various group work PRs but I clearly missed some of them 😄

The download had "Groupname" (capital G), and other columns with spaces and user-friendly (rather than upload-friendly) column names. I used "Group name" instead, to follow the same convention, but it should be easy enough for an instructor to use group_name instead. Everything else is handled.

mwest1066:
> The download had "Groupname" (capital G), and other columns with spaces and user-friendly (rather than upload-friendly) column names. I used "Group name" instead, to follow the same convention, but it should be easy enough for an instructor to use group_name instead. Everything else is handled.

Perfect, thanks!

PR #3977, opened 2021-03-18
New Script for Anonymizing Course Database Dumps Downloaded from the API

SethPoulsen:
IRB often requires a 3rd party to download a course data dump and anonymize it before sending it to the research team. 

This should make the process easier. 

I also removed the old script, which @mwest1066 said doesn't work anymore.


trombonekenny:
When I've done this kind of anonymization work before, researchers had cared about consistency in the anonymity mapping across executions-- either new data sets like if we did this six months later, or if found a buglet somewhere and needed to immediately run it again. It doesn't look like this script does that (but could with some sort of local data source: sqlite or some language specific hash-dump). Is that a requirement for you, @SethPoulsen?

SethPoulsen:
This is a great question. That's not a requirement for me right now, because the course is over and so I'll just be getting all the data at once. However, I can see how that would be a useful feature for if you were, e.g. getting anonymized data incrementally over the course of a semester.

trombonekenny:
At that point, why not code this functionality into the API? Then you don't need a middle-man?

SethPoulsen:
This is a great idea which @mwest1066 and I talked about, and actually we do still want to do it. 

The reason I didn't do it now is because I think it would require touching some rather complicated systems (e.g. adding a new type of user role where a user is authenticated to access course data only anonymously), and I need this data within the next week.

PR #3992, opened 2021-03-23
added an option to auto-refresh table when looking at the students tab

nitahieb:
@jonatanschroeder 
Added a feature to auto refresh the students tab when looking at an assessment

### DIFF
--- a/pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
+++ b/pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
@@ -189,7 +189,7 @@
                 
               </div>
             </div>
-            <input type="checkbox" id="refreshCheck" checked>
+            <input type="checkbox" id="refreshCheck">
             <label for="refreshCheck"> Auto-Refresh </label><br>
           </div>
         </div>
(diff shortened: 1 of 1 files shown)


nitahieb:
For just the autorefresh feature I was wondering if there's any standardization on what the website should look like?
RIght now I put the autorefresh under the action for all instances.
image

jonatanschroeder:
> For just the autorefresh feature I was wondering if there's any standardization on what the website should look like?
> RIght now I put the autorefresh under the action for all instances.

I think it may be best to show it at the bottom. Also, in terms of style, take a look at boostrap: https://getbootstrap.com/docs/4.0/components/forms/#checkboxes-and-radios-1

nitahieb:
Do you mean at the bottom of the page outside of the table? Or at the bottom of the table?

jonatanschroeder:
> Do you mean at the bottom of the page outside of the table? Or at the bottom of the table?

Not sure, can you try both and see what looks better?

jonatanschroeder:
I'm going to close this, since (based on #4543 and a possible expectation of migrating this page to `bootstrap-table` as well) it may be better to work with a different paradigm to refresh data using JSON instead of full page refresh.

Issue #3616, opened 2020-12-27
Add option to allow students to see assessment but not create/submit


jonatanschroeder:
Add an option in access rules for an assessment that allows students to see an assessment (either not yet started or already completed), but not start a new instance or submit any questions for open instances. Completed assessment overview/answers are shown based on `showClosedAssessment`, and grades are shown based on `showClosedAssessmentScore`. This could be useful for:

* Students to take a look at future assessments, for planning purposes;
* Avoid questions like "I don't see the exam on PL yet!" a few minutes before the assessment is supposed to start;
* After an assessment is closed, students can see their grades, but students that didn't start it cannot submit anything.

Assessment can be seen, as well as the dates/times in the future where assessment will be submittable.

This could also encompass the issue raised in #3553.

jonatanschroeder:
Additional ideas discussed today with @mwest1066 and @cheeren that could be implemented with this:

There should be a way to set an assessment instance grades as "released". Then it should be possible to set "showClosedAssessmentScore" to "false" only if grades are released. Then, instructors would have a button to release grades for an individual assessment instance, or to all instances of an assessment.

PR #3993, opened 2021-03-23
Add "active" access restriction

echau01:
This is a first-draft implementation of the access rule restriction described in #3616.

@jonatanschroeder @cheeren

TODO:
- [x] Unit tests
- [x] ReadTheDocs documentation
- [x] If the active access rule has `submittable = false`, and there is a future time slot where students can submit answers for credit, then display the start of the next time slot in the "Available Credit" string.


lib/question.js
```diff
@@ -1522,7 +1532,7 @@ module.exports = {
             const locals = {};
             config.setLocals(locals);
             _.assign(locals, this._buildQuestionUrls(urlPrefix, variant, question, instance_question, assessment));
-            _.assign(locals, this._buildLocals(variant, question, instance_question, assessment, assessment_instance, assessment_question));
+            _.assign(locals, this._buildLocals(variant, question, instance_question, assessment, assessment_instance, assessment_question, null));
```
jonatanschroeder, line 1535: I'm not sure if `null` is appropriate here... I may be a bit paranoid, but I imagine if a grading job that takes a really long time could cause a page that was supposed to be non-submittable to end up showing a submit button.

sprocs/check_assessment_access.sql
```diff
@@ -20,10 +20,13 @@ CREATE OR REPLACE FUNCTION
         OUT seb_config JSONB,         -- SEBKeys (if any) for this assessment.
         OUT show_closed_assessment boolean, -- If students can view the assessment after it is closed.
         OUT show_closed_assessment_score boolean, -- If students can view their grade after the assessment is closed
+        OUT submittable boolean,     -- If the assessment is visible but not submittable
         OUT access_rules JSONB       -- For display to the user. The currently active rule is marked by 'active' = TRUE.
     ) AS $$
 DECLARE
     active_access_rule_id bigint;
+    assessment_start_date TIMESTAMP WITH TIME ZONE;
+    assessment_end_date TIMESTAMP WITH TIME ZONE;
```
jonatanschroeder, line 29: Not sure I understand why these are needed...

jonatanschroeder (commented):
I left some minor things to consider. The code looks fine, though I didn't have a chance to test it yet. Great job so far!

sprocs/check_assessment_access.sql
echau01: Oops, this was left over from my previous work. It shouldn't be there.

echau01:

echau01:
recheck

jonatanschroeder (commented):
Overall it looks great. Some general comments:

* It may be a good idea to add a column to the Access page (pages/instructorAssessmentAccess), to avoid confusion.
* I don't think the submittable rules should have precedence over non-submittable, rather the order should be considered.
* Check pages/studentAssessments, lines 38 and 50. These tests make the assessment name not show a link to students if the assessment is an assessment with multiple instances. If you add `aa.submittable` to the corresponding query, you can add it as an additional condition that doesn't show a link in those cases.

sprocs/check_assessment_access.sql
```diff
@@ -100,6 +145,7 @@ BEGIN
         seb_config = NULL;
         show_closed_assessment = TRUE;
         show_closed_assessment_score = TRUE;
+        submittable = TRUE;
     END IF;
 
     -- List of all access rules that will grant access to this user/mode/role at some date (past or future),
```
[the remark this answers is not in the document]
jonatanschroeder: (To clarify, I'm talking about the query in lines 153 onwards).

lib/question.js
jonatanschroeder: I can't think of a better way to handle this. I guess `null` is an appropriate heuristic in this case, as the corner cases where this happens to affect the result can then be caught in submission errors instead.

sprocs/check_assessment_access.sql
```diff
@@ -71,9 +77,35 @@ BEGIN
         aar.assessment_id = check_assessment_access.assessment_id
         AND caar.authorized
     ORDER BY
+        aar.submittable DESC,
```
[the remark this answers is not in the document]
echau01: When the submittable rule is used, the available credit should be 0 (because if one cannot submit answers, it doesn't make sense to have credit be available). However, it is possible for someone to mistakenly set the available credit to a nonzero value. For example:

```
"allowAccess": [
    {
        "startDate": "2021-01-01T00:00:01",
        "endDate": "2021-01-01T23:59:59",
        "credit": 75
    },
    {
        "submittable": false,
        "credit": 100
    }
]
```

If I did not give submittable assessments priority, then the access rule with `submittable` being `false` would always be the active access rule. The assessment would never be submittable.

I considered just putting a warning in the documentation saying that if `submittable` is set to `false`, then the available `credit` must be 0. However, I felt that I should do more to handle the potential mistake. Ideally, I'd throw some sort of error if there is an access rule with `submittable` being `false` and `credit` being nonzero, but I don't see an easy way to do that. Should I just put a warning in the documentation and leave it at that?

echau01:
Thank you very much for the feedback. Some questions:

> It may be a good idea to add a column to the Access page (pages/instructorAssessmentAccess), to avoid confusion.

What should the column say?

> Check pages/studentAssessments, lines 38 and 50. These tests make the assessment name not show a link to students if the assessment is an assessment with multiple instances. If you add aa.submittable to the corresponding query, you can add it as an additional condition that doesn't show a link in those cases.

To confirm: there should be no link if the assessment is not submittable?

sprocs/check_assessment_access.sql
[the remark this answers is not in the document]
jonatanschroeder: That is an interesting point, but I think the proper course of action is not to give priority to submittable, it's to ensure that non-submittable rules don't have credit. This could be done with something on sync that rejects a submittable rule with non-zero credit (or that gives a warning and sets the credit to zero).

jonatanschroeder:
> What should the column say?

The title can be Submittable, and the text "True" or "False".

> To confirm: there should be no link if the assessment is not submittable?

I don't think students need to see a rule that has no submission. This list is intended for students to see when they can work on the assessment, so filtering out non-submittable rules probably makes sense here.

jonatanschroeder:
> > To confirm: there should be no link if the assessment is not submittable?
> 
> I don't think students need to see a rule that has no submission. This list is intended for students to see when they can work on the assessment, so filtering out non-submittable rules probably makes sense here.

Just realized I answered the wrong question. No, there should be no link if the assessment is not submittable.

docs/accessControl.md
```diff
@@ -198,6 +199,32 @@ To block students from viewing closed assessment scores, set `"showClosedAssessm
 
 The `showClosedAssessment` access rule restriction is only useful in conjunction with [disabling real-time grading](assessment.md#disabling-real-time-grading) and setting `"showClosedAssessment": false`.
 
+## Submittable assessments
+
+The `submittable` access rule restriction is useful for allowing students to see what assessments they have coming up. If `submittable` is set to `false`, students can see the assessment on the Assessments page, but they cannot create a new assessment instance or submit answers to questions. If an assessment is currently not submittable but will be in the future, students can see when the assessment will become submittable by looking at the `Available credit` column on the Assessments page. Assessments are submittable by default.
+
+**Note**: if `submittable` is set to `false` in an access rule, the available `credit` for that access rule must be 0.
```
jonatanschroeder, line 206:
```suggestion
**Note**: if `submittable` is set to `false` in an access rule, the available `credit` cannot be set to any value other than 0 (the default value).
```

Just to clarify that they don't need to set the credit, they only need to ensure that it is not set to a non-zero value.

docs/accessControl.md
```diff
@@ -198,6 +199,32 @@ To block students from viewing closed assessment scores, set `"showClosedAssessm
...
+    },
+    {
+        "startDate": "2015-01-01T00:00:01",
+        "endDate": "2015-02-15T23:59:59",
+        "submittable": false
+    }
+]
+```
+
+In the example above, from January 1st to February 15th, students can see the assessment on the Assessments page but cannot begin the assessment. If students try to click into the assessment during this time, they will see a message saying that the assessment will be available on February 16th. The first access rule applies on February 16th, and since `submittable` is `true` (by default), students can start the assessment and submit answers to questions on that day.
```
jonatanschroeder, line 226:
```suggestion
In the example above, from January 1st to February 15th, students can see the assessment on the Assessments page but cannot begin the assessment. They will see a message saying that the assessment will be available on February 16th. The first access rule applies on February 16th, and since `submittable` is `true` (by default), students can start the assessment and submit answers to questions on that day.
```

jonatanschroeder (approved):
Great job!

mwest1066:
Looks great! Could we modify one of the existing `exampleCourse` assessments to demonstrate the usage of this?

And maybe it’s worth adding a news item? This is a pretty cool feature.

docs/accessControl.md
```diff
@@ -198,6 +199,32 @@ To block students from viewing closed assessment scores, set `"showClosedAssessm
 
 The `showClosedAssessment` access rule restriction is only useful in conjunction with [disabling real-time grading](assessment.md#disabling-real-time-grading) and setting `"showClosedAssessment": false`.
 
+## Submittable assessments
+
+The `submittable` access rule restriction is useful for allowing students to see what assessments they have coming up. If `submittable` is set to `false`, students can see the assessment on the Assessments page, but they cannot create a new assessment instance or submit answers to questions. If an assessment is currently not submittable but will be in the future, students can see when the assessment will become submittable by looking at the `Available credit` column on the Assessments page. Assessments are submittable by default.
```
mwest1066, line 204: I think this should also be used when releasing exams to students for post-exam review? Using this new restriction will ensure that a student who had never started the exam will be unable to start it during the post-exam review period.

mwest1066 (approved):

docs/accessControl.md
echau01: It depends on what the "post-exam review period" looks like. If `submittable` is set to false, students who have not started the exam will not be able to see the questions at all. That's fine if the instructor only wants the students who did the exam to be able to see the questions. On the other hand, if the instructor wants to allow everyone to see the exam questions, they should *not* set `submittable` to false. Instead, they should set `credit: 0` (or omit the `credit` field so that it defaults to 0) in the access rules. In the end, it depends on whether or not the instructor wants to let everyone see the questions.

mwest1066: You describe the situation exactly correctly, including the current practice. I think we should recommend that people change this current practice to instead use `"submittable": false` when returning exams. It would be great to specify this in the docs, both in the section about the `submittable` restriction and also here: https://prairielearn.readthedocs.io/en/latest/faq/#how-do-i-give-students-access-to-view-their-exams-after-they-are-over

exampleCourse/courseInstances/Sp15/assessments/exam05-submittableAccessRestriction/infoAssessment.json
```diff
@@ -0,0 +1,44 @@
...
+            "credit": 100,
+            "timeLimitMin": 50,
+            "startDate": "2100-01-02T00:00:01",
+            "endDate": "2100-01-02T23:59:59",
+            "showClosedAssessment": false
+        },
+        {
+            "mode": "Public",
+            "submittable": false,
+            "showClosedAssessment": false
```
mwest1066, line 31: I think we don't want `"showClosedAssessment": false` here, because it will block students from viewing their closed assessments?

Also, maybe we can add a `"comment": "This access rule is for viewing the exam after it is finished"`?

And we should probably have a `startDate` set so that exam viewing only starts a couple of days after the exam ended?

exampleCourse/courseInstances/Sp15/assessments/exam05-submittableAccessRestriction/infoAssessment.json
```diff
@@ -0,0 +1,44 @@
...
+        },
+        {
+            "mode": "Public",
+            "credit": 100,
+            "timeLimitMin": 50,
+            "startDate": "2100-01-02T00:00:01",
+            "endDate": "2100-01-02T23:59:59",
+            "showClosedAssessment": false
+        },
```
mwest1066, line 27: We seem to have two identical access rules above?

news_items/029_non_submittable/index.html
```diff
@@ -0,0 +1,79 @@
...
+     margin-bottom: -1em;
+     line-height: 125%;
+ }
+</style>
+
+<p>
+  Thanks to great work by 
+  <a href="https://github.com/echau01">Eugene Chau</a>, a student in
+  the University of British Columbia, it is now possible to allow
+  students to see assessments that are not available for submission,
```
mwest1066, line 32: I feel like it's not really clear here that "not available for submission" also means "can't be started / can't be viewed". I wonder whether it's worth separating out the "before release data" and "after deadline" cases more, and explaining them each in more detail. For the "before release" case I also find that phrases like "allow students ... to see it" (from a later paragraph) is confusing. Could we be clearer about the two meanings of "see it": (1) the assessment shows up on the list of assessments, with future dates; and (2) being able to view the assessment, including all the questions. I'm thinking about something like:

> 1. Before an assessment has been released, using `"submittable": false` means the students will see the assessment on the course homepage, including the release date, but they won't be able to start the assessment or see any of the questions.
> 2. After students have completed an assessment, using `"submittable": false` means the students will be able to view their work, including seeing all questions, but they won't be able to submit any further answers. If they had not yet started the assessment by this time, then they won't be able to view it at all.

Trying to explain this makes me wonder whether there is a better word than `submittable`. Maybe `active`? Another possibility would be to separate out the `startAsessment` and `submitAnswers` restrictions. But I think that it's better to keep them together, just like in the current PR. But I do worry that `submittable` isn't very clear for the "before release" case.

echau01: I really like the name "active". Thanks for the suggestion.

exampleCourse/courseInstances/Sp15/assessments/exam05-submittableAccessRestriction/infoAssessment.json
echau01: Fixed

exampleCourse/courseInstances/Sp15/assessments/exam05-submittableAccessRestriction/infoAssessment.json
echau01: Fixed

docs/accessControl.md
echau01: Documentation has been added

docs/faq.md
```diff
@@ -23,12 +23,13 @@ To allow students to see their entire exam after the semester is over you can ad
...
+Students who wrote the exam will then have public access to their exams after the `startDate` until the end of the course instance, while students who did not write the exam will not be able to view it. Students with access will not be able to answer questions for further credit (there is no `"credit": 100` line), but they will be able to see the entire exam in exactly the same state as when they were doing the exam originally. Because students have public access to the exam, it should be assumed that all the questions will be posted to websites such as Chegg and Course Hero. To let students see their exams with some additional security, consider only allowing (limited access post-exam under controlled conditions)[#should-students-be-able-to-review-their-exams-after-they-are-over] (although this requires in-person access by students and doesn't work online).
```
mwest1066, line 32: I'm afraid that there is significant regional variation in the verb used for students answering questions on a test, with "take an exam" (US, UK, Australia), "sit an exam" (UK, Australia), "write an exam" (Canada), and "give an exam" (India) all being used. It's especially confusing because the US uses both "write an exam" and "give an exam" to describe the actions taken by the _professor_ in preparing and delivering the exam.

In these situations PL defaults to US conventions, just to pick a standard.

```suggestion
Students who took the exam will then have public access to their exams after the `startDate` until the end of the course instance, while students who did not write the exam will not be able to view it. Students will not be able to answer questions (due to `"active": false`), but they will be able to see the entire exam in exactly the same state as when they were doing the exam originally. Because students have public access to the exam, it should be assumed that all the questions will be posted to websites such as Chegg and Course Hero. To let students see their exams with some additional security, consider only allowing (limited access post-exam under controlled conditions)[#should-students-be-able-to-review-their-exams-after-they-are-over] (although this requires in-person access by students and doesn't work online).
```

docs/accessControl.md
```diff
@@ -198,6 +199,32 @@ To block students from viewing closed assessment scores, set `"showClosedAssessm
 
 The `showClosedAssessment` access rule restriction is only useful in conjunction with [disabling real-time grading](assessment.md#disabling-real-time-grading) and setting `"showClosedAssessment": false`.
 
+## Active assessments
+
+The `active` access rule restriction is useful for allowing students to see what assessments they have coming up. It should also be used when [returning exams to students](faq.md#how-do-i-give-students-access-to-view-their-exams-after-they-are-over). If `active` is set to `false`, students can see the assessment on the Assessments page, but they cannot create a new assessment instance or submit answers to questions. If an assessment is currently not active but will be in the future, students can see when the assessment will become active by looking at the `Available credit` column on the Assessments page. The `active` property in an access rule is `true` by default.
```
mwest1066, line 204:
```suggestion
The `active` access rule restriction is useful for allowing students to see what assessments they have coming up. It should also be used when [returning exams to students](faq.md#how-do-i-give-students-access-to-view-their-exams-after-they-are-over). If `active` is set to `false`, students can see the assessment on the Assessments page, but they cannot start the assessment, create a new assessment instance, or submit answers to questions. If an assessment is currently not active but will be in the future, students can see when the assessment will become active by looking at the `Available credit` column on the Assessments page. The `active` property in an access rule is `true` by default.
```

news_items/029_non_active/index.html
```diff
@@ -0,0 +1,79 @@
...
+  This feature allows students to visualize and plan for future
+  assessments, potentially from the start of the term. It also allows
+  instructors to release completed assessments for review while
+  blocking any further submission.
+</p>
+
+<img class="border rounded" src="non-active-exam.png" width="900px">
+
+<p>
+  You can set up your assessments to be visible, but not active
```
mwest1066, line 46:
```suggestion
  You can set up your assessments to be visible, but not active,
```

news_items/029_non_active/index.html
```diff
@@ -0,0 +1,79 @@
...
+  by creating a new access rule with the <code>active</code>
+  setting set to <code>false</code>. This will allow students that
+  have not started the assessment to see it while prevent them from
```
mwest1066, line 49:
```suggestion
  have not started the assessment to see it while preventing them from
```

news_items/029_non_active/index.html
```diff
@@ -0,0 +1,79 @@
...
+  starting a new instance of the assessment. It will also allow
+  students that completing the assessment to see their answers, but
```
mwest1066, line 51:
```suggestion
  students that complete the assessment to see their answers, but
```

news_items/029_non_active/index.html
echau01: I decided to reword some parts of this paragraph. Can you please check it?

mwest1066: Looks good! I'll add one more comma 😄

Issue #3622, opened 2020-12-30
Remove the concept of "closing" assessments


mwest1066:
After extensive discussion with @jonatanschroeder in the #3600 thread, we propose to eliminate the concept of assessments being "closed/open". This would supersede #3344.

The basic idea is that "open/closed" is one more piece of state that instructors need to keep track of. If we can achieve all the functionality that we need without this, then it would simplify everyone's lives. Instructors wouldn't need to remember to re-open student assessment instances when they want to give them access again, etc. Removing "open/closed" would also be helpful for bringing Homework and Exam assessments more inline with each other as part of #3347.

To eliminate "open/closed" we would:
* [ ] Remove the `assessment_instances.open` column.
* [ ] For timed exams, replace checks for an "open" exam with checking that the timer has not expired. When a timer expires we would continue 
* [ ] Rename `showClosedAssessment` to `showCompletedAssessment` (keeping the old one for back-compat). Similarly rename `showClosedAssessmentScore`. Both of these would now trigger on an expired timer rather than "open/closed".
* [ ] Eliminate the `autoClose` option for Exams. This was originally introduced as a hack to get around exam security issues in the CBTF, but then have been long since solved in a much better way. Now `autoClose` only causes headaches and confusion for people. We should leave it as an undocumented option in `infoAssessment.json` files, for back-compat, and throw an error if we find `"autoClose": true` anywhere.
* [ ] For all assessments, automatically grade any outstanding submissions after the deadline has expired. To do this, we add a flag `instance_questions.has_ungraded_submissions BOOLEAN`. Every time we save a submission for this instance question, we set `has_ungraded_submissions := true`. We repurpose `cron/autoFinishExams` to search for any instance questons that have `has_ungraded_submissions = true` and which are not accessible to the student, grade them, and set `has_ungraded_submissions := false`. We would need an index on this flag to make the search query fast.
* [ ] On the `instructorAssessmentInstances` page we add a column to show `has_ungraded_submissions`. We also add a button to "Grade all ungraded submissions".
* [ ] Getting rid of "closing" would require implementing #3616 so that instructors can give students access to view (but not edit) their exams after they are over, even without the concept of "closing".

This is related to #3600 and #2750.

jonatanschroeder:
> For all assessments, automatically grade any outstanding submissions after the deadline has expired. To do this, we add a flag `instance_questions.has_ungraded_submissions BOOLEAN`.

[...]

PR #3996, opened 2021-03-25
Add Grade All button

### LINKED ISSUE #3644: Add buttons for Grade all/Close all in assessment instance list

jonatanschroeder:
Fixes #3644

### DIFF
--- a/pages/instructorAssessmentInstances/instructorAssessmentInstances.js
+++ b/pages/instructorAssessmentInstances/instructorAssessmentInstances.js
@@ -105,9 +105,9 @@ router.post('/', function(req, res, next) {
                 res.redirect(req.originalUrl);
             });
         });
-    } else if (req.body.__action == 'grade_all') {
+    } else if (req.body.__action == 'grade_all' || req.body.__action == 'close_all') {
         const assessment_id = res.locals.assessment.id;
-        const close = false;
+        const close = req.body.__action == 'close_all';
         const overrideGradeRate = true;
         assessment.gradeAllAssessmentInstances(assessment_id, res.locals.authn_user.user_id, close, overrideGradeRate, function(err) {
             if (ERR(err, next)) return;
--- a/pages/instructorAssessmentInstances/instructorAssessmentInstances.js
+++ b/pages/instructorAssessmentInstances/instructorAssessmentInstances.js
@@ -105,6 +105,14 @@ router.post('/', function(req, res, next) {
                 res.redirect(req.originalUrl);
             });
         });
+    } else if (req.body.__action == 'grade_all') {
+        const assessment_id = res.locals.assessment.id;
+        const close = false;
+        const overrideGradeRate = true;
+        assessment.gradeAllAssessmentInstances(assessment_id, res.locals.authn_user.user_id, close, overrideGradeRate, function(err) {
+            if (ERR(err, next)) return;
+            res.redirect(req.originalUrl);
+        });
     } else if (req.body.__action == 'delete_all') {
         const params = [
             res.locals.assessment.id,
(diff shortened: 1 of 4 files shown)


cheeren:
omg thank you.

pages/instructorAssessmentInstances/instructorAssessmentInstances.js
```diff
@@ -105,6 +105,14 @@ router.post('/', function(req, res, next) {
                 res.redirect(req.originalUrl);
             });
         });
+    } else if (req.body.__action == 'grade_all') {
+        const assessment_id = res.locals.assessment.id;
+        const close = false;
```
mwest1066, line 110: Why is this not also closing the assessment instances? For per-instance actions we only have the "Grade & Close" action, so I would have thought that the "all" button would mimic this. Or are you thinking ahead to #3622?

pages/instructorAssessmentInstances/instructorAssessmentInstances.js
```diff
@@ -105,6 +105,14 @@ router.post('/', function(req, res, next) {
...
+        const overrideGradeRate = true;
+        assessment.gradeAllAssessmentInstances(assessment_id, res.locals.authn_user.user_id, close, overrideGradeRate, function(err) {
```
mwest1066, line 112: This is a potentially long-running function so I think that it should be run under a server jobSequence? My thinking is that this would get the list of assessment instances to grade, but wouldn't lock them. It would then loop over these instances and call the existing `assessment.gradeAssessmentInstance()` function on each of them, which will take care of locking/unlocking each one individually, so that we only hold the lock briefly. Ideally we'd also print output so the instructor would see:
```
Grading assessment instance #1 for mwest@illinois.edu
Grading assessment instance #1 for zilles@illinois.edu
...
```

mwest1066:
> omg thank you.

Yes! I think we can all agree with this sentiment. This is a highly-desired feature!

pages/instructorAssessmentInstances/instructorAssessmentInstances.js
jonatanschroeder: Yes, good idea, I'll take a look at this. With that in mind, about external grading jobs, currently the jobs are started by this process, but the result page is loaded before they complete. Should these jobs be executed as part of the grade all job sequence or in the background? They could still run in parallel, but should the page return a success once they are triggered or once they are finished?

pages/instructorAssessmentInstances/instructorAssessmentInstances.js
jonatanschroeder: I was thinking that the "Close All" could be achieved by expiring the time limit for all assessments, which would work for now and for after #3622 is implemented, but I don't have a problem with changing this to also close all assessments. The idea was to give instructors the ability to grade them without closing, and then closing them separately. I can also have both options ("Grade All" and "Close All"), and add the Grade only button in per-instance actions.

jonatanschroeder: The problem with expiring the time limit for all assessments is that it doesn't work for Homework-type assessments, since they don't check the time limit.

jonatanschroeder: I added both options ("Grade All" and "Close All").

pages/instructorAssessmentInstances/instructorAssessmentInstances.js
jonatanschroeder: Just pushed what seems to be a working version of using a job sequence. My tests seem to work, and I think this is done or close to done, but I never worked with job sequences before, I'd appreciate a review of the implementation. I used `async.eachSeries` inside the job, is that necessary or should I use something simpler? Also, should I list the completed/pending "grade all" jobs somewhere?

mwest1066 (approved):

pages/instructorAssessmentInstances/instructorAssessmentInstances.js
mwest1066: The use of `eachSeries()` is perfect!

PR #4000, opened 2021-03-25
Add safeguard to ensure submission uploads match the UID

### LINKED ISSUE #3997: Grade upload: require student ID and submission ID to match
If a CSV file has both student ID and submission ID, currently the file ignores the student ID. This can lead to confusion if the file, due to an error, shows one user but uploads to another. I believe that if multiple identifying fields are used, then all must match in order to select a submission to grade. In other words, if the file contains a submission ID and a student ID, and the student associated to the specific submission doesn't match the student in the file, then the upload should fail. Same goes for instance ID.

jonatanschroeder:
Fixes #3997

### DIFF
--- a/sprocs/instance_questions_update_score.sql
+++ b/sprocs/instance_questions_update_score.sql
@@ -9,7 +9,7 @@ CREATE OR REPLACE FUNCTION
         -- identify the instance_question/submission
         IN arg_submission_id bigint,        -- must provide submission_id
         IN arg_instance_question_id bigint, -- OR instance_question_id
-        IN arg_uid text,                    -- OR (uid, assessment_instance_number, qid)
+        IN arg_uid_or_group text,           -- OR (uid/group, assessment_instance_number, qid)
         IN arg_assessment_instance_number integer,
         IN arg_qid text,
 
@@ -24,7 +24,7 @@ DECLARE
     submission_id bigint;
     instance_question_id bigint;
     assessment_instance_id bigint;
-    found_uid text;
+    found_uid_or_group text;
     max_points double precision;
     new_score_perc double precision;
     new_points double precision;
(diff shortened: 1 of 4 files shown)


sprocs/instance_questions_update_score.sql
```diff
@@ -33,8 +34,8 @@ BEGIN
     -- ##################################################################
     -- get the assessment_instance, max_points, and (possibly) submission_id
 
-    SELECT        s.id,                iq.id,                  ai.id, aq.max_points
-    INTO submission_id, instance_question_id, assessment_instance_id,    max_points
+    SELECT        s.id,                iq.id,                  ai.id, aq.max_points, u.uid
+    INTO submission_id, instance_question_id, assessment_instance_id,    max_points, found_uid
```
mwest1066, line 38: Is this safe for group assessments? I assume so, but I haven't traced it carefully to check.

jonatanschroeder: Good point. I just tested, and the update happens as usual in this case. The fact that `found_uid` will be NULL for these assessments means that `found_uid != arg_uid` (in the IF that checks for this value in line 67) will be NULL (equivalent to false), so regardless of if there is a `uid` column in the file the exception will never be raised. The download file for manual grading doesn't include a UID column for group assessments, so this would only take place if the instructor explicitly includes a `uid` column, which would lead me to believe that the instructor is expecting the safeguard. So options are:

1. if there is a UID column in a group assessment, trigger an error always (since no assessment is linked to a single student);
2. for group assessments check the `group_name` column instead;
3. for group assessments, check if at least one user in the group matches the uid.

Option 1 is the simplest. Option 2 is probably the best, but would require either refactoring the sproc and its calls to add a group name, or repurposing the `arg_uid` to mean `arg_uid_or_group_name` instead.

jonatanschroeder: Option 2 implemented.

mwest1066: Excellent, that's definitely the best-but-painful option, so thanks for doing it.

Issue #3843, opened 2021-02-18
POGIL roles for group assessments


mfsilva22:
Add a management tool at the beginning of group assessments, where students can define group roles. We will start with Recorded, Reflector and Manager. 

Roles will be selected for each student in the group. The roles will control how students have access to different questions. Think also about a mechanism for modifying the group size (for when students are missing the group activity)

reteps:
@nwalters512 is this closed?

PR #4004, opened 2021-03-26
Add POGIL roles to database

renzol2:
To integrate POGIL roles directly into group assessments (per #3843), we would like to create a new `user_roles` table that connect users to a POGIL role `enum`. This addition lets us implement a number of things related to POGIL role functionality in group assessments in the future, such as

- restricting certain functionalities based on student roles (ex. only Recorders can Save & Grade, only Reflectors can fill out reflections, etc.)
- allowing students to change roles more easily during an assessment
- accessing role information outside the context of a question

We are also implementing a PostgreSQL function `update_group_pogil_roles` that is called when the Manager of a group submits new or updated roles for their group. This function ensures that certain constraints regarding POGIL roles are kept in place:

- every group must have exactly one Manager, one Recorder, and one Reflector
- every person is assigned to at least one role
- no one is assigned to Contributor unless the group size is bigger than 3

If there are any other constraints that are of importance, please let me or @caleqw2 know!


alawinia:
Reviewed all updates and everything looks good to me.

migrations/232_enum_pogil_role__create.sql
```diff
@@ -0,0 +1,7 @@
+DO $$
+BEGIN
+    IF NOT EXISTS (SELECT * FROM pg_type WHERE typname = 'enum_pogile_role') THEN
```
nicknytko, line 3:
```suggestion
    IF NOT EXISTS (SELECT * FROM pg_type WHERE typename = 'enum_pogil_role') THEN
```

Is this confirmed to work locally? If you were to run this locally without the migration (by say, starting the docker container and opening postgres with `psql`), and then apply the migration, does it work okay?

migrations/232_enum_pogil_role__create.sql
```diff
@@ -0,0 +1,7 @@
...
+        CREATE TYPE enum_pogil_role AS ENUM ('None', 'Manager', 'Recorder', 'Reflector', 'Contributor');
+    END IF;
+END;
+$$;
```
nicknytko, line 7: Tiny nit, but please put a newline at the end of files 😀

sprocs/update_group_pogil_role.sql
```diff
@@ -0,0 +1,19 @@
...
+    -- Update the user's pogil role
+    UPDATE
+        group_users as gu
+    SET
+        pogil_role = arg_group_role
+    WHERE
+        gu.user_id = arg_user_id
+        AND gu.group_id = arg_group_id;
+END;
+$$ LANGUAGE plpgsql VOLATILE;
```
nicknytko, line 19: Nice!

migrations/232_enum_pogil_role__create.sql
renzol2: For this migration, I had copied from another migration [008_enum_instance_question_status__create.sql](https://github.com/PrairieLearn/PrairieLearn/blob/master/migrations/008_enum_instance_question_status__create.sql) that also added a new enum type. Just to check that it worked, I ran the docker container without migrations and using `psql`, created the enum (and ran again just to make sure it exists):

```
postgres=# CREATE TYPE enum_pogil_role AS ENUM ('None', 'Manager', 'Recorder', 'Reflector', 'Contributor');
CREATE TYPE

postgres=# CREATE TYPE enum_pogil_role AS ENUM ('None', 'Manager', 'Recorder', 'Reflector', 'Contributor');
ERROR:  type "enum_pogil_role" already exists
```

We also ran the `233` migration to manually add the enum as a column to the `group_users` table:
```
postgres=# ALTER TABLE group_users ADD COLUMN pogil_role enum_pogil_role DEFAULT 'None'::enum_pogil_role;
ALTER TABLE
postgres=# \d group_users
                          Table "public.group_users"
   Column   |      Type       | Collation | Nullable |         Default
------------+-----------------+-----------+----------+-------------------------
 group_id   | bigint          |           | not null |
 user_id    | bigint          |           | not null |
 pogil_role | enum_pogil_role |           |          | 'None'::enum_pogil_role
Indexes:
    "group_users_pkey" PRIMARY KEY, btree (group_id, user_id)
    "group_users_group_id_key" btree (group_id)
    "group_users_user_id_key" btree (user_id)
Foreign-key constraints:
    "group_users_group_id_fkey" FOREIGN KEY (group_id) REFERENCES groups(id) ON UPDATE CASCADE ON DELETE CASCADE
    "group_users_user_id_fkey" FOREIGN KEY (user_id) REFERENCES users(user_id) ON UPDATE CASCADE ON DELETE CASCADE
```

sprocs/update_group_pogil_role.sql
```diff
@@ -0,0 +1,19 @@
+DROP FUNCTION IF EXISTS update_group_pogil_role(enum_pogil_role,bigint,bigint);
+CREATE OR REPLACE FUNCTION
+    update_group_pogil_role (
```
mwest1066, line 3: Please name sprocs starting with the table name (in this case `group_users`). I think this sproc should be `group_users_update_pogil_role`.

sprocs/update_group_pogil_role.sql
```diff
@@ -0,0 +1,19 @@
...
+        arg_group_role enum_pogil_role, -- This is the new POGIL role!
+        arg_user_id bigint,
+        arg_group_id bigint
```
mwest1066, line 6: Put the user_id and group_id as the first args, because they are the primary key of the row. Also put the group_id first because that's the primary key order. Also don't use `arg_group_role` when the column is called `pogil_role` -- instead use `arg_pogil_role`. So the args will be `arg_group_id, arg_user_id, arg_pogil_role`.

sprocs/update_group_pogil_role.sql
```diff
@@ -0,0 +1,19 @@
...
+    ) RETURNS void
+AS $$
+BEGIN
+    -- Update the user's pogil role
```
mwest1066, line 10: Comment is not needed.

renzol2:
@nicknytko @mwest1066 

After discussing with the rest of the POGIL team today, we've realized that we need to allow students to take on multiple roles in the case of a group having less students than required roles since we are implementing restricted and unique functionality for each role. For example, if a group had only 2 students, one student may have to be both a Recorder (able to check and submit answers) and a Reflector (fills out reflection on behalf of the group). 

This poses a problem for our current database changes, since we've assumed that each user only takes on one role. Database-wise, what would be the best way to go about allowing multiple roles per student? Prof. Alawini suggested we create a new table to allow a user to carry multiple different roles in the same group, but we wanted to check and make sure that this approach lines up with the overall database design.

mwest1066:
> This poses a problem for our current database changes, since we've assumed that each user only takes on one role. Database-wise, what would be the best way to go about allowing multiple roles per student? Prof. Alawini suggested we create a new table to allow a user to carry multiple different roles in the same group, but we wanted to check and make sure that this approach lines up with the overall database design.

Another table sounds fine. Make sure you think about whether you want to add constraints such as "you can't have two people in the same role in the same group".

renzol2:

caleqw2:

nicknytko (approved):
I don't see anything horrendously wrong here, though @mwest1066 is generally more adept at SQL and can probably give better feedback.  

I did notice that there is lots of logic in the sproc for making sure there are the correct number of roles, everyone is assigned to a role, etc.  I am wondering if this would be duplicated on the frontend/js side if we are to provide detailed feedback as to why a group update fails?

nwalters512:
@renzol2 @caleqw2 is it fair to say that #5200 supersedes this work?

caleqw2:
>

PR #4023, opened 2021-03-31
Fix handling of group users in instance downloads

jonatanschroeder:
Some instance downloads (e.g., final submissions, best submissions) don't return all submissions in case of group work. This seems to be related to the use of `unique_group` in the downloads file, since the code aims to only return one element per group, but that one element could not match the final or best submission.
```
        (row_number() OVER (PARTITION BY v.id ORDER BY s.date DESC, s.id DESC)) = 1 AS final_submission_per_variant,
        (row_number() OVER (PARTITION BY v.id ORDER BY s.score DESC NULLS LAST, s.date DESC, s.id DESC)) = 1 AS best_submission_per_variant,
        (row_number() OVER (PARTITION BY g.id, v.id, s.id)) = 1 AS unique_group,
...
WHERE
    ($include_all
        OR ($include_final AND final_submission_per_variant)
        OR ($include_best AND best_submission_per_variant))
        AND (($group_work AND unique_group) OR ($group_work = false))
```

Since the group user isn't used anywhere in the download anyway (the uids are obtained via a separate sproc), I removed the joins with group users to ensure that only one entry is returned for each submission/instance.

* [x] Test the changes

### DIFF
--- a/pages/instructorAssessmentDownloads/instructorAssessmentDownloads.js
+++ b/pages/instructorAssessmentDownloads/instructorAssessmentDownloads.js
@@ -1,5 +1,4 @@
 const ERR = require('async-stacktrace');
-const _ = require('lodash');
 const express = require('express');
 const router = express.Router();
 const path = require('path');
@@ -51,14 +50,15 @@ router.get('/', function(req, res, _next) {
 });
 
 var sendInstancesCsv = function(res, req, columns, options, callback) {
-    var params = {assessment_id: res.locals.assessment.id};
+    var params = {
+        assessment_id: res.locals.assessment.id,
+        highest_score: options.only_highest,
+        group_work: options.group_work,
+    };
     sqldb.query(sql.select_assessment_instances, params, function(err, result) {
         if (ERR(err, callback)) return;
 
         var rows = result.rows;
-        if (options.only_highest) {
-            rows = _.filter(rows, 'highest_score');
-        }
         csvMaker.rowsToCsv(rows, columns, function(err, csv) {
             if (ERR(err, callback)) return;
             res.attachment(req.params.filename);
--- a/pages/instructorAssessmentDownloads/instructorAssessmentDownloads.js
+++ b/pages/instructorAssessmentDownloads/instructorAssessmentDownloads.js
@@ -59,9 +59,6 @@ var sendInstancesCsv = function(res, req, columns, options, callback) {
         if (options.only_highest) {
             rows = _.filter(rows, 'highest_score');
         }
-        if (options.group_work) {
-            rows = _.filter(rows, 'unique_group');
-        }
         csvMaker.rowsToCsv(rows, columns, function(err, csv) {
             if (ERR(err, callback)) return;
             res.attachment(req.params.filename);
--- a/pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql
+++ b/pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql
@@ -302,7 +302,7 @@ WITH all_submissions_with_files AS (
         JOIN assessment_instances AS ai ON (ai.assessment_id = a.id)
         LEFT JOIN group_configs AS gc ON (gc.assessment_id = a.id)
         LEFT JOIN groups AS g ON (g.id = ai.group_id AND g.group_config_id = gc.id)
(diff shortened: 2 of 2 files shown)


pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql
```diff
@@ -301,15 +296,13 @@ WITH all_submissions_with_files AS (
...
-        (row_number() OVER (PARTITION BY g.id, v.id, s.id)) = 1 AS unique_group,
         g.name AS group_name
     FROM
         assessments AS a
         JOIN assessment_instances AS ai ON (ai.assessment_id = a.id)
         LEFT JOIN group_configs AS gc ON (gc.assessment_id = a.id)
         LEFT JOIN groups AS g ON (g.id = ai.group_id AND g.group_config_id = gc.id)
-        LEFT JOIN group_users AS gu ON (gu.group_id = g.id)
-        JOIN users AS u ON (u.user_id = ai.user_id OR u.user_id = gu.user_id)
+        JOIN users AS u ON (u.user_id = ai.user_id)
```
mwest1066, line 305: Should this be a LEFT JOIN? If it's just a plain JOIN it will require that `ai.user_id`) is NOT NULL, which will exclude group work assessment instances.

pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql
```diff
@@ -168,8 +165,7 @@ WITH all_submissions AS (
         JOIN assessment_instances AS ai ON (ai.assessment_id = a.id)
         LEFT JOIN group_configs AS gc ON (gc.assessment_id = a.id)
         LEFT JOIN groups AS g ON (g.id = ai.group_id AND g.group_config_id = gc.id)
-        LEFT JOIN group_users AS gu ON (gu.group_id = g.id)
-        JOIN users AS u ON (u.user_id = ai.user_id OR u.user_id = gu.user_id)
+        LEFT JOIN users AS u ON (u.user_id = ai.user_id)
```
mwest1066, line 168: Should we also include a `LEFT JOIN users AS authn_u ON (u.user_id = s.auth_user_id)` and select `authn_u.uid AS submitter_uid`? If we include `submitter_uid` as a column in the output CSV then it will show the actual submitting user for both group work and also non group work (e.g., maybe a TA makes a submission for a student).

This is not a required change for this PR -- it's just a thought I had while reading the PR.

pages/instructorAssessmentDownloads/instructorAssessmentDownloads.js
```diff
@@ -51,17 +50,15 @@ router.get('/', function(req, res, _next) {
 });
 
 var sendInstancesCsv = function(res, req, columns, options, callback) {
-    var params = {assessment_id: res.locals.assessment.id};
+    var params = {
+        assessment_id: res.locals.assessment.id,
+        highest_score: options.only_highest,
+        group_work: options.group_work,
+    };
```
mwest1066, line 57: Nice! It's great to move this into the SQL.

pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql
jonatanschroeder: Isn't this what `submission_user` is about? Not my change, was there already...

pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql
jonatanschroeder: My bad, this one was missing. Fixed now.

pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql
mwest1066: D'oh! 😄

mwest1066 (approved):

mwest1066:
LGTM, but I see you have an unchecked "Test the changes" item in the PR description. I'll wait to hear from you about when you've finished testing.

jonatanschroeder:
> LGTM, but I see you have an unchecked "Test the changes" item in the PR description. I'll wait to hear from you about when you've finished testing.

Forgot to check it, should be good to go.

PR #4029, opened 2021-04-04
Allow specifying assessment instructions in an assessment.html file

### LINKED ISSUE #4028: Allow specifying assessment instructions in an assessment.html file
PrairieLearn supports a `text` property in `infoAssessment.json` that can be used to show student-facing information for assessments. It supports EJS templates for building links to assessment resources (cheat sheets, etc.). It's rendered in a light gray box ("For this assessment..."):

Screen Shot 2021-04-03 at 4 35 11 PM

To make it easier to build more complex sets of instructions, we should support an `assessment.html` file that's adjacent to `infoAssessment.json`. If this file is present, it will be templated with Mustache and rendered in place of `assessment.text`. If it's not present, we'll fall back to `assessment.text` for backwards compatibility.

nwalters512:
This PR implements the feature described in #4028, allowing the creation of an `assessment.html` file in an assessment directory that will be shown to students in place of the `text` property in `infoAssessment.json`.

I opted to store the file's contents in the database at sync time so that we don't have to worry about creating additional chunks for assessment contents.

We could opt to deprecate the `text` property in `infoAssessment.json` and add a warning for that property at sync time, but I don't think that's necessary at the moment. Happy to add that if folks think differently!

Closes #4028


exampleCourse/courseInstances/Sp15/assessments/exam02-realTimeGradingDisabled/assessment.html
```diff
@@ -0,0 +1 @@
+For this quiz you can use the <a target="_blank" href="{{ client_files_course }}/formulas.pdf">formula sheet</a>.
```
nwalters512, line 1: IMO, we could remove this from every assessment except `exam01-bestPractices`, I don't think it serves much value to repeat it so many times.

lib/assessment.js
```diff
@@ -44,27 +45,30 @@ module.exports = {
...
+     * Render the `html` or `text` property of an assessment.
      *
-     * @param {Object} assessment - The assessment to render the text for.
+     * @param {import('../sync/course-db').Assessment} assessment - The assessment to render the text for.
      * @param {string} urlPrefix - The current server urlPrefix.
      * @param {function} callback - A callback(err, html) function.
      */
-    renderText(assessment, urlPrefix, callback) {
-        if (!assessment.text) return callback(null, null);
+    renderHtmlOrText(assessment, urlPrefix) {
```
nwalters512, line 54: There was no reason for this function to be async, so I dropped the callback in favor of just returning a value.

pages/studentAssessmentInstanceExam/studentAssessmentInstanceExam.js
```diff
@@ -59,34 +60,27 @@ router.post('/', function(req, res, next) {
     }
 });
 
-router.get('/', function(req, res, next) {
+router.get('/', asyncHandler(async (req, res, next) => {
```
nwalters512, line 63: An intermediate implementation used an async `renderHtmlOrText`, so I converted this to an async handler. Though it's not necessary any more, I'm leaving it as this is paying down a bit of tech debt.

jonatanschroeder:
About using sync instead of loading from the directory, isn't the directory stored in the disk anyway because of clientFilesAssessment? So wouldn't it make sense to just use the same structure as the question, and keep it consistent? This would mean that the html doesn't need to be in the database or handled on sync, and for dev mode changes in this setting could be done locally without having to Load from Disk. This would make it consistent with other sync processes (for instructors), where only changes in JSON require a "Load from disk", and changes in PY/HTML files would be loaded automatically.

docs/assessment.md
```diff
@@ -296,3 +296,11 @@ The text of the honor code was based on the University of Maryland's [Honor Pled
...
 
 The appropriate link to provide to students can be found by opening the "Settings" tab of the Assessment. This page includes, among other useful information, a Student Link that can be provided to students. This link points students directly to the specific assessment, enrolling them automatically in the course if they are not yet enrolled.
+
+## Providing instructions for an assessment
+
+To provide instructions or other information to students, create an `assessment.html` adjacent to the `infoAssessment.json` for your assessment. This file will be rendered on a student's assessment page. `client_files_course`, `client_files_course_instance`, and `client_files_assessment` variables can be accessed via Mustache templates to construct links to resources in your course. For example, to provide access to a formula sheet located in `clientFilesCourse/formulas.pdf`, you can add the following to your `assessment.html`:
```
jonatanschroeder, line 302: Most of the instructor documentation avoids the use of library names, like Mustache.
```suggestion
To provide instructions or other information to students, create an `assessment.html` adjacent to the `infoAssessment.json` for your assessment. This file will be rendered on a student's assessment page. `client_files_course`, `client_files_course_instance`, and `client_files_assessment` variables can be accessed using double-curly-braces to construct links to resources in your course. For example, to provide access to a formula sheet located in `clientFilesCourse/formulas.pdf`, you can add the following to your `assessment.html`:
```

nwalters512:
> About using sync instead of loading from the directory, isn't the directory stored in the disk anyway because of clientFilesAssessment? So wouldn't it make sense to just use the same structure as the question, and keep it consistent? This would mean that the html doesn't need to be in the database or handled on sync, and for dev mode changes in this setting could be done locally without having to Load from Disk. This would make it consistent with other sync processes (for instructors), where only changes in JSON require a "Load from disk", and changes in PY/HTML files would be loaded automatically.

Good point about consistency with other `*.html` files in dev mode. Currently only `clientFilesAssessment` is stored as a chunk, so `assessment.html` wouldn't be avilable to machines that render student assessments. I was trying to avoid the need to create additional chunks, though I suppose it'd be worth it to achieve consistency! I can make that change.

docs/assessment.md
nwalters512: We do mention it in the question and client/server files docs. I don't think we do instructors any favors by not naming the specific syntax we're using in case they want to look up more information on it on their own.

jonatanschroeder: Good point, I hadn't read the entire sentence...

echuber2:
What happens if an assessment has both the text property and the new html file?

echuber2:
Looks like if both `assessment.html` is present and the `text` attribute is defined, it favors the html, but there's no warning or error. I'd suggest making this an error to help staff avoid making a mistake.

nwalters512:
This PR is extremely out of date and would probably be best off is started from scratch when we decide to prioritize this work. I'm going to close it for now.

PR #4031, opened 2021-04-05
Introduce assessment chunks

nwalters512:
https://github.com/PrairieLearn/PrairieLearn/pull/4029 requires assessment files not in `clientFilesAssessment` to be stored as chunks so that they can be loaded to disk on chunk servers. Rather that introduce a bespoke `assessmentHtml` chunk type, we want to introduce an `assessment` chunk type instead and cut over implementations to use that instead of `clientFilesAssessment`. This PR is the first of two that will be necessary to make that transition.

This PR introduces code for creating `assessment` chunks during sync. Once this has been deployed to production and the appropriate chunks have been created, we'll be able to merge and deploy a second PR that cuts over to using the new chunk type instead of `clientFilesAssessment`. We can't do this in a single PR as that wouldn't let us guarantee that the new `assessment` chunks would have been created before they're needed to serve some request.

This PR also introduces a button to create chunks for all assessments. Once this PR is deployed, we'll need to click that button and make sure the job completes successfully before we deploy any code that relies on `assessment` chunks.


lib/chunks.js
```diff
@@ -296,24 +319,46 @@ module.exports.identifyChunksToGenerate = async ({ coursePath, courseId, courseD
...
-                path.join(coursePath, 'courseInstances', ciid, 'assessments', tid, 'clientFilesAssessment'),
-            );
+            const assessmentPath = path.join(coursePath, 'courseInstances', ciid, 'assessments', tid);
+            const hasAssessmentDirectory = await fs.pathExists(assessmentPath);
+            const hasClientFilesAssessmentDirectory = await fs.pathExists(path.join(assessmentPath, 'clientFilesAssessment'));
+            if (hasAssessmentDirectory && (
+                (
+                    !existingCourseChunks.courseInstances[ciid] ||
+                    !existingCourseChunks.courseInstances[ciid].assessments ||
+                    !existingCourseChunks.courseInstances[ciid].assessments.has(tid)
```
nwalters512, line 329: Are we running on Node 14 yet in prod? It'd be so nice to be able to use optional chaining here:
```suggestion
                    !existingCourseChunks.courseInstances[ciid]?.assessments?.has(tid)
```

nicknytko: Yes!

```
[ec2-user@ip-10-0-0-162 ~]$ node -v
v14.15.4
```

nwalters512: Sweet - just refactored in 25e32a4!

nwalters512:
This PR is extremely out of date and would probably be best off is started from scratch when we decide to prioritize this work. I'm going to close it for now.

PR #4067, opened 2021-04-17
Preserve order of blocks after submission (bugfix)

jonatanschroeder:
The order of blocks in `pl-order-blocks` was not maintained once there is a submission. This was particularly relevant in case of `source-blocks-order="ordered"`, since the blocks would no longer be ordered once a submission was sent.


Frozemint (approved):
This makes a lot of sense :sweat_smile: :sweat_smile:

PR #4164, opened 2021-05-06
Proper handling of exception in sys.exit in Python caller

jonatanschroeder:
The Python caller trampoline could occasionally cause an exception on termination, where `sys.exit(0)` would fail. This caused the `exit` event not to be called, which could cause the PythonCaller waitForFinish event never to exit (since the process is presumed to still be closing). This can cause tests to hang forever in an infinite loop.

This fix causes `os._exit(0)` to be called if `sys.exit(0)` fails on signal-based termination, causing the process to exit as expected. Hopefully this would resolve the issue where occasional CI tests fail due to timeout (or at least one of the reasons for that).

### DIFF
--- a/lib/python-caller-trampoline.py
+++ b/lib/python-caller-trampoline.py
@@ -167,9 +167,9 @@ def terminate_worker(signum, stack):
     try:
         sys.exit(0)
     except Exception as e:
-        # print(e, file=sys.stderr)
-        # traceback.print_exc()
-        # print('terminate_worker: sys.exit(0) failed, calling os._exit(0)', file=sys.stderr)
+        print(e, file=sys.stderr)
+        traceback.print_exc()
+        print('terminate_worker: sys.exit(0) failed, calling os._exit(0)', file=sys.stderr)
         os._exit(0)
 
 signal.signal(signal.SIGTERM, terminate_worker)
--- a/lib/python-caller-trampoline.py
+++ b/lib/python-caller-trampoline.py
@@ -164,7 +164,12 @@ worker_pid = 0
 def terminate_worker(signum, stack):
     if worker_pid > 0:
         os.kill(worker_pid, signal.SIGKILL)
-    sys.exit(0)
+    try:
+        print('terminate_worker: calling sys.exit(0)')
+        sys.exit(0)
+    except:
+        print('terminate_worker: calling os._exit(0)')
+        os._exit(0)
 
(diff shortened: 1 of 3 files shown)


mwest1066:
@nwalters512 does this look ok to you in the context of #1626?

jonatanschroeder:
Problem was that the exception that was happening was not a subclass of `Exception` so I changed to `BaseException` and it now works.

lib/python-caller-trampoline.py
```diff
@@ -164,7 +164,13 @@ def worker_loop():
 def terminate_worker(signum, stack):
     if worker_pid > 0:
         os.kill(worker_pid, signal.SIGKILL)
-    sys.exit(0)
+    try:
+        sys.exit(0)
+    except BaseException as e:
```
nwalters512, line 169: Given that `sys.exit(0)` raises a `SystemExit` exception, and that [`SystemExit` extends from `BaseException`](https://docs.python.org/3/library/exceptions.html#SystemExit), won't we hit the `except` case 100% of the time?

echuber2: Then could `os._exit(0)` be used right away, or instead, add an initial handler to re-raise the `SystemExit` case? Like this:

``` python
try:
    sys.exit(0)
except SystemExit as e:
    raise e
except BaseException as e:
    # ...
```

I don't know if re-raising is proper for this, though.

echuber2: On second thought, my second suggestion doesn't help if the exit call does raise a failure exception, since it will not be handled by the other case when re-raised. So, what about resorting to `os._exit(0)` initially?

jonatanschroeder: I considered using `os._exit(0)` originally, but I thought that `sys.exit(0)` would probably clean up some relevant data. If that is not necessary (considering that #3731 replaced it in other points of the code), I'm happy to change it to use that instead. @nwalters512 can you think of any reason why keeping `sys.exit(0)` would be needed in this case? Maybe a parent process that needs to handle the child process?

nwalters512: Yeah, I was thinking about that. We should only hit this case if either a) the parent is specifically trying to kill the zygote, or b) if the user sends SIGINT, and in both cases we shouldn't have to worry about cleanup, since zygotes are meant to be stateless and free of side effects. I think that should be safe.

mwest1066 (approved):

PR #4194, opened 2021-05-08
Replace `correct_answer` with `correct_answers`

firasm:
I had a look at all of the example problems and all of them use `data["correct_answers"]` rather than `data["correct_answer"]`. In some cases in the docs, `correct_answer` is used and generates an error (specifically in the symbolic_input question.

It would be nice if someone could confirm this is the case before updating the documentation.

### DIFF
--- a/docs/elements.md
+++ b/docs/elements.md
@@ -262,7 +262,7 @@ Attribute | Type | Default | Description
 --- | --- | --- | ---
 `answers-name` | string | — | Variable name to store data in.
 `weight` | integer | 1 | Weight to use when computing a weighted average score over elements.
-`correct-answers` | float | special | Correct answer for grading. Defaults to `data["correct_answers"][answers-name]`.
+`correct-answer` | float | special | Correct answer for grading. Defaults to `data["correct_answers"][answers-name]`.
 `label` | text | — | A prefix to display before the input box (e.g., `label="$F =$"`).
 `suffix` | text | — | A suffix to display after the input box (e.g., `suffix="$\rm m/s^2$"`).
 `display` | "block" or "inline" | "inline" | How to display the input field.
@@ -525,7 +525,7 @@ Attribute | Type | Default | Description
 --- | --- | --- | ---
 `answers-name` | string | — | Variable name to store data in. If the correct answer `ans` is a `sympy` object, you should use `import prairielearn as pl` and `data['correct_answers'][answers-name] = pl.to_json(ans)`.
 `weight` | integer | 1 | Weight to use when computing a weighted average score over elements.
-`correct-answers` | float | special | Correct answer for grading. Defaults to `data["correct_answers"][answers-name]`.
+`correct-answer` | float | special | Correct answer for grading. Defaults to `data["correct_answers"][answers-name]`.
 `label` | text | — | A prefix to display before the input box (e.g., `label="$F =$"`).
 `display` | "block" or "inline" | "inline" | How to display the input field.
 `variables` | string | — | A comma-delimited list of symbols that can be used in the symbolic expression.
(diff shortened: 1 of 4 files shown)


firasm:

docs/elements.md
```diff
@@ -262,15 +262,15 @@ Attribute | Type | Default | Description
 --- | --- | --- | ---
 `answers-name` | string | — | Variable name to store data in.
 `weight` | integer | 1 | Weight to use when computing a weighted average score over elements.
-`correct-answer` | float | special | Correct answer for grading. Defaults to `data["correct_answers"][answers-name]`.
+`correct-answers` | float | special | Correct answer for grading. Defaults to `data["correct_answers"][answers-name]`.
```
jonatanschroeder, line 265: This is the argument expected by `pl-number-input`, it was correct originally. The problem was in the Python version, not the HTML argument.

docs/elements.md
```diff
@@ -516,16 +516,16 @@ def generate(data):
...
 ```
 
 #### Customizations
 
 --- | --- | --- | ---
-`answers-name` | string | — | Variable name to store data in. If the correct answer `ans` is a `sympy` object, you should use `import prairielearn as pl` and `data['correct_answer'][answers-name] = pl.to_json(ans)`.
+`answers-name` | string | — | Variable name to store data in. If the correct answer `ans` is a `sympy` object, you should use `import prairielearn as pl` and `data['correct_answers'][answers-name] = pl.to_json(ans)`.
-`correct-answer` | float | special | Correct answer for grading. Defaults to `data["correct_answers"][answers-name]`.
```
jonatanschroeder, line 528: Same as above.

docs/elements.md
```diff
@@ -583,7 +583,7 @@ Attribute | Type | Default | Description
 --- | --- | --- | ---
 `answers-name` | string | — | Variable name to store data in.
 `weight` | integer | 1 | Weight to use when computing a weighted average score over elements.
-`correct-answer` | string | special | Correct answer for grading. Defaults to `data["correct_answers"][answers-name]`.
+`correct-answers` | string | special | Correct answer for grading. Defaults to `data["correct_answers"][answers-name]`.
```
jonatanschroeder, line 586: Ditto

jonatanschroeder:
That's a good catch. I just did a quick search and noticed a similar problem in https://github.com/PrairieLearn/PrairieLearn/blob/6512ff99b48aac753f90cafec88f31369091ed1e/elements/pl-symbolic-input/pl-symbolic-input.py#L42 (and following lines) where it's using `correct-answers` instead of `correct_answers`. Do you want to fix that too?

Same for https://github.com/PrairieLearn/PrairieLearn/blame/master/docs/pl-drawing/index.md#L33

docs/elements.md
firasm: Hmm.

I get an error if I use `correct-answer` for symbolic input.

Screen Shot 2021-05-08 at 2 13 23 PM

jonatanschroeder: There are two things you might be confusing here. The Python way to indicate a correct answer is `data["correct_answers"][field] = value`, but you can also set it in the HTML element itself:
`<pl-symbolic-input answers-name="fieldname" correct-answer="x+y^2" ...></pl-symbolic-input>`

docs/elements.md
firasm: Same for this one, the string input questions expect `correct-answers`

Screen Shot 2021-05-08 at 2 16 48 PM

docs/elements.md
firasm: I think I am not understanding something crucial... 

For numeric input, `correct-answer` produces an error in the problem.

Screen Shot 2021-05-08 at 2 19 31 PM

docs/elements.md
firasm: Ohhh I see!! Thank you.

I think that is very confusing. Am I wrong about this? The parameter names should be the same shouldn't they? I suppose that would be a much bigger fix...

jonatanschroeder: The `correct-answer` parameter to the HTML element is the answer to that particular element. The `correct_answers` dict on Python is a dictionary of all the correct answers for all the elements.

firasm: Ah and you can have multiple elements, but in a python dict it's a list of all of them.

Got it. Thanks, I'll revert the changes on these

firasm:
I think I've got them all now @jonatanschroeder

jonatanschroeder (approved):

mwest1066:
Thanks for the fixes @firasm!

PR #4216, opened 2021-05-10
Fix a very small typo in docs.

CrashAndSideburns:
The docs' example for symbolic inputs incorrectly refers to data["correct_answer"] instead of data["correct_answers"]. Quite a minor error, but I thought it was worth correcting.


CrashAndSideburns:
Ah, beat me to it. Thanks!

PR #4230, opened 2021-05-18
Add API endpoint for course instance info

mepstein-gh:
Added an api call for basic info about a course instance.

Sample call URL: [http://localhost:3000/pl/api/v1/course_instances/2/course_instance_info](http://localhost:3000/pl/api/v1/course_instances/2/course_instance_info)

Sample output:
```
[
  {
    "deleted_at": null,
    "course_title": "Test Course",
    "display_timezone": "America/Chicago",
    "course_short_name": "QA 101",
    "course_instance_id": 2,
    "hide_in_enroll_page": false,
    "course_instance_course_id": 2,
    "course_instance_long_name": "Spring 2015",
    "course_instance_short_name": "Sp15"
  }
]
```


nwalters512:
Per REST API conventions, I would expect this to be served from a route like `pl/api/v1/course_instances/2`. Other than that, this looks reasonable!

mepstein-gh:
I actually was planning to do it that way initially, but then I wasn't quite sure how to do it, and thought that might lead to a conflict/ambiguity in the processing of the different URLs.  But I just made a slight change to `/api/v1/index.js`, which seems to do the trick.  Is this a reasonable approach?  (Note that even if it is, I still need to update a few things, in `api.md`, `testApi.js`, and `api_download.py` -- which means the tests won't pass.  I'll update those after I hear back on this question.)

nwalters512:
Yep, that approach is correct - I _believe_ the reason that this works is that router paths match more "strictly" than middleware paths (`/foo/bar/` would match a `/foo` middleware, but not a `/foo` route). Would have to test that myself to be sure though, the Express docs didn't clarify for me.

mepstein-gh:
OK, I updated those other pieces, the tests now pass.  This is good to go on my end, unless you have other concerns.  Thanks.

tools/api_download.py
```diff
@@ -25,6 +25,7 @@ def download_course_instance(args, logfile):
     log(logfile, f'starting download at {local_iso_time()} ...')
     start_time = time.time()
     course_instance_path = f'/course_instances/{args.course_instance_id}'
+    course_instance_info = get_and_save_json(f'{course_instance_path}', 'course_instance_info', args, logfile)
```
nwalters512, line 28: nit: no need for the f-string, you can use the raw `course_instance_path`.

nwalters512 (approved):

tools/api_download.py
mepstein-gh: Yep.  The need for that went away after I removed the "course_instance_info" part of the URL per your earlier comment.

tests/testApi.js
```diff
@@ -511,4 +511,31 @@ describe('API', function () {
...
+        page = body;
+        callback(null);
+      });
+    });
+    it('should parse as JSON', function () {
+      locals.json = JSON.parse(page);
+    });
+    it('should have one item of info', function () {
+      assert.lengthOf(locals.json, 1);
+    });
```
nwalters512, line 539: Ahhh, looking at this tests, I'm realizing this is returning an array. There should only ever be one course instance for a specific ID, so we can safely have this return just an object.

nwalters512: Resolved in 330ef44.

PR #4244, opened 2021-05-23
Bugfix: change invalid function test to no longer use atan

jonatanschroeder:
#4217 added `atan` as a valid function in `pl-symbolic-input`, but `atan` was being used in tests as an example of an invalid function. This PR fixes that by replacing the invalid function with something that is unlikely to be valid any time in the future.


jonatanschroeder:
Note that this bug was causing undeterministic CI failures, since if the test for any question with a symbolic input element happened to use the invalid function test by coincidence, it would succeed when a failure was expected.

mwest1066 (approved):

Issue #3606, opened 2020-12-24
Gradebook ideas


mwest1066:
This is a collection of ideas for future full gradebook functionality.

1. Allow the creation of extra assessments that are purely for score upload. Assessment instances will get made during instructor score upload, not by students.
2. Add curving for assessments and overall course score. See https://edatscale.org/2020/12/24/simple-exam-score-curving-with-mtm/
3. Add an "exempt" flag for assessments, so that an instructor can mark a student as exempt from a particular assessment. This could be a separate table, or we could store it in the assessment instances.
4. Add a "weight" parameter to assessments and compute assessment-set average scores (e.g., average quiz score) with these weights.
5. Allow assessments to be marked as retries of other assessments, with a combining function (e.g., alpha*max + (1-alpha)*min).
6. Add a "weight" parameter to assessment sets, and use this to compute an overall course score.
7. Add a map from overall course scores to letter grades.
8. Allow manual upload/download of overall course score and letter grades.
9. Allow a course-instance `server.py` with a `grade()` function that is given all assessment score information and can update overall scores and letter grades.

mwest1066:
See #3605 for a similar idea about `grade()` functions on a per-assessment level.

jonatanschroeder:
For computing assessment-set averages, also include a setting for "drop lowest/highest X assessments".

There also needs to be some discussion on what is shown on assessment-set averages while there are unstarted or ungraded assessments, or assessments with unpublished grades (e.g., showClosedAssessmentScore set to false).

daviddalpiaz:
One minor gradebook thought I'll put here while I'm thinking about it: Allow for sorting of columns?

Currently, if you click a column heading, it takes you to that assessment. I know this, because I do it a lot expecting it to sort the columns. The use case here is something like sorting on an exam grade to get a quick look at something like homework grades for students that didn't do well.

alvaromonge:
Here's a suggestion for the list... Provide a **student view of the stats for assessments and even questions**. It's good for students to understand where they are in the assessment relative to the class. Otherwise knowing that you have a 75% isn't helpful unless you know that the highest score was 82% and other information like the standard deviation and even the score distribution. Perhaps the instructor can also control when to enable this view and for which assessments.

firasm:
ooooh **9** sounds really useful:

> Allow a course-instance server.py with a grade() function that is given all assessment score information and can update overall scores and letter grades.

jpolitz:
[...]

PR #4257, opened 2021-05-25
Add score only assessments

Kieran-Weaver:
This pull request adds the `scoreOnly` flag to `assessmentInfo.json`, which specifies that an assessment is only for score upload. As described in issue #3606, score only assessments cannot be attempted by students. Instead, when an instructor uploads scores, assessment instances are created automatically if they do not already exist.

@jonatanschroeder 

TODO:

- [ ] Unit tests
- [ ] ReadTheDocs Documentation


jonatanschroeder (commented):
The feature you created, instead of making an assessment "score only", makes it so a new instance is created on upload when one doesn't exist. I'd imagine a score-only assessment would need a bit more thought (e.g., possibly the ability to have no questions, and some further thoughts on how to handle cases where students open this kind of assessment, before or after a grade is pushed). Maybe having this feature in combination with the submittable feature that Eugene has submitted may make more sense, but I believe in this case it may be better to focus on what this feature actually does. This means replacing "scoreOnly" with "autoInstanceCreation" or something similar. Thoughts?

For the sprocs: 

* for the arguments in the sprocs, instead of user_uid I believe it would be more generic to make it user_id, and have the caller obtain the user_id instead;
* dates in general come from SQL, not from JS. Instead of date as an argument, use SQL date expressions like `current_timestamp`;
* instead of passing `score_only` as an argument, note that you already fetch `group_work` and `mode` from the assessment, might as well fetch the `score_only` flag from there as well.
* Uploads of group assessments are typically based on the group name, not on one of the students' UIDs. Based on the code above it would make sense for the user ID to also have a group ID in case of group work.

Kieran-Weaver:
Renaming "scoreOnly" makes sense, would "createInstancesOnUpload" be a better name? I don't know if something like "autoInstanceCreation" would be intuitive to use.

jonatanschroeder:
Maybe createInstanceOnGrading? Upload may be too restrictive if we want to allow other types of manual grading later.

Kieran-Weaver:
createInstanceOnGrading sounds like a better name. For the sprocs, I used the user id for group assessments because `assessment_instances_insert` takes a user id regardless of `group_work`. Is there a better way to create group assessment instances?

jonatanschroeder:
> For the sprocs, I used the user id for group assessments because `assessment_instances_insert` takes a user id regardless of `group_work`. Is there a better way to create group assessment instances?

Good point. Let's keep it like this for now and discuss how to handle group assessments later. I imagine the best option would probably require more thought, potentially creating a group for the user if one does not exist.

nwalters512:
@Kieran-Weaver would you still like to see this PR through?

Kieran-Weaver:
No, I'm not working on this anymore. Sorry!

PR #4336, opened 2021-06-02
Facelift to authenticated homepage

astrosticks:
This is a redesign for the authenticated home page that is more accessibility-minded with the use of headers and lists instead of a single-column table. Visually, it's inspired by sites such as Gradescope and ClassTranscribe. 

# Screenshots

## Managing courses
| Before | After |
| --- | --- |
|  |  |

## No managed courses
| Before | After |
| --- | --- | 
| home-nomanage-before | home-nomanage-after |

You'll notice that the "Your courses" h1 tag replaces the h2 tag for "Course instances" if you're not staff on any courses. This ensures that there aren't too many headers on-screen if they aren't needed, and also makes sure that there's always an h1 tag.

# Loose ends
 - [ ] Fix tests that use this page after the design gets approved.
- @tdy Back when the landing page dropped, some people were asking why the landing and courses overview page were part of the same template. Now that time has passed, are you still committed to keeping these pages part of the same file? Now that only a couple of elements are shared between these pages, I think it would be nice for maintainability to move them into separate EJS files completely, and use partials to manage the shared elements.

### DIFF
--- a/pages/home/courseInstanceListEntry.ejs
+++ b/pages/home/courseInstanceListEntry.ejs
@@ -1,0 +1,15 @@
+<li class="d-flex mb-1" style="max-width: 100%;">
+  <a href="<%= href %>" class="py-2 px-3 mr-1 border rounded" style="max-width: 350px; overflow: hidden;">
+    <div style="font-size: 125%;">
+      <div class="font-weight-bold">
+        <%= course_instance.short_name %>
+      </div>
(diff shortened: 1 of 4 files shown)


nwalters512:
I like the direction we're heading in here! I think ultimately we'll implement something similar as part of a more holistic visual refactor. I'm going to close this for now.

PR #4343, opened 2021-06-03
Add custom format in pl-number-input correct/submitted answers

### LINKED ISSUE #4253: pl-number-input: Add user-controlled threshold for switching to scientific notation
Currently in the `pl-number-input` element, large numbers are only formatted as scientific notation if they are above 1E12. Personally, I think that should be reduced to 5 or 6, but ideally this would be a user-controlled parameter - maybe `scinot-threhold` or something similar?

In [`pl-number-input.py`](https://github.com/PrairieLearn/PrairieLearn/blob/master/elements/pl-number-input/pl-number-input.py#L47) the relevant code is:

```
def format_true_ans(element, data, name):
    a_tru = pl.from_json(data['correct_answers'].get(name, None))
    if a_tru is not None:
        # Get comparison parameters
        comparison = pl.get_string_attrib(element, 'comparison', COMPARISON_DEFAULT)
        if comparison == 'relabs':
            # FIXME: render correctly with respect to rtol and atol
            a_tru = '{:.12g}'.format(a_tru)
```

I was about to start working on a PR to implement this but there's a `# FIXME` note from @nicknytko (if I read git blame correctly) that I don't fully understand.

P.S. Because of this hard-coded setting we (@jake-bobowski and I)  have to go through work-arounds with `parse` to correctly format the numbers:

119284306-404af280-bbf4-11eb-90b8-dcfce6cb8811

jonatanschroeder:
Resolves #4253


firasm:
(quick note of thanks for starting this PR @jonatanschroeder ! This does exactly what I want! )

mwest1066 (approved):

Issue #4338, opened 2021-06-02
More display customization for assessments in student views


astrosticks:
# Overview

This issue summarizes some recent discussions between @mfsilva22 and myself.

## Assessment units

We'd like to add some options to `infoCourse.json` to give instructors more flexibility for displaying the assessment list to students. Instead of grouping assessments together by `assessmentSet`, instructors could display them by "assessmentUnit" (@mwest1066, let us know if you like this name- we were also considering "assessmentTopic"). Think of a unit as, well, a unit in a course: Probably a couple of homework assessments, then maybe a practice exam or a lab or two, followed by the real exam. 

@mfsilva22 has already hacked together a workaround on the live PL site to implement "units" using assessmentSets: 

<img width="228" alt="image" src="https://user-images.githubusercontent.com/12501485/120567549-b2cb8780-c3d7-11eb-979f-04e55c4b8fa4.png">

Here's a rough idea of what both kinds of course structure could look like:

### Outline: Assessments grouped by `assessmentSet` (old method)

- Homeworks
  - HW1
  - HW2
  - HW3
  - HW4
- Labs
  - LAB1
  - LAB2
- Exams
  - E1
  - E2

### Outline: Assessments grouped by `assessmentUnit` (proposed method)

- Unit 1: Vectors
  - HW1
  - HW2
  - LAB1
  - E1
- Unit 2: Matrices
  - HW3
  - HW4
  - LAB2
  - E2

## Reverse-chronological ordering

Additionally, student responses collected by @mfsilva22 suggest that students would prefer a reverse-chronological ordering in the assessments page- i.e. most recent/relevant assessments appear at the top of the page. We could introduce an `ascendingOrder` key in `infoCourse.json` to accomplish this, but there are still some outstanding design considerations: 
- ~~Should `ascendingOrder` re-order the list of assessments inside of a group of assessments, just the groups themselves, or both?~~ Answer: These flags are flexible enough not to need special parameters for reverse-ordering.
- ~~Could instructors just move around the entries in `assessmentUnits` or `assessmentSets` to accomplish the same thing?~~

# Example usage

## `infoCourse.json`
```js
"assessmentUnits": [
  {"name":"add", "heading": "Adding numbers"},
  {"name":"sub", "heading": "Subtracting numbers"},
],
"groupBy": "unit", // default: "set"
"ascendingOrder": false, // default: true
```

## `infoAssessment.json`
```js
[...]

PR #4360, opened 2021-06-07
feat: Add option to group assessments by Module

astrosticks:
Inspired by discussion in #4338, this PR implements assessment "modules" which provide a more ergonomic grouping of the assessments in a course instance. This is an alternative to the traditional (and still default) way of grouping them by set.

This replaces #4332.

# Screenshots

## `"groupBy": "set"` (default, current functionality)

image 

## `"groupBy": "module"` (new option)

image


database/enums/enum_assessment_grouping.pg
```diff
@@ -0,0 +1 @@
+set, unit
```
jonatanschroeder, line 1: Not saying it's incorrect, just raising the issue: most non-techinical enums use CamelCase for items, would it be more consistent to have: 
```suggestion
Set, Unit
```

docs/assessment.md
```diff
@@ -34,6 +34,10 @@ Assessments are organized into `sets` (e.g., `Homework`, `Quiz`, `Exam`) and wit
 
 You can select a set from the list of [standardized assessment sets](course.md#standardized-assessment-sets) or create your [own](course.md#adding-your-own-assessment-sets).
 
+## Assessment units
+
+An assessment will belong to the unit named by its `unit` property. By default, an assessment defined without an explicit unit will belong to the "Default" unit. Units are optional fields that are used to control the order that assessments are displayed to students. For more information, see [Course configuration](course.md#assessment-units).
```
jonatanschroeder, line 39: I think it might be a good idea to be a bit more verbose with examples of what a unit may represent.
```suggestion
Instructors may optionally group their assessments by course units, corresponding to course topics or individual parts of the course. This allows students to see all assessments corresponding to a particular unit, which can help them plan their study and practice habits.

By default, every assessment is assigned to the "Default" unit. Instructors may assign an assessment to a specific unit by setting its `unit` property. 

For more information about assessment units, see [Course configuration](course.md#assessment-units).
```

You may also want to add a `unit` line in the example in line 15.

docs/course.md
```diff
@@ -124,6 +124,31 @@ The assessment set order in `infoCourse.json` is the order in which the assessme
     ],
 ```
 
+## Units
+
+Each assessment in the course belongs to a *unit* defined in `infoCourse.json`. Units should be thought of as a unit of instruction, typically with a consistent number of assessments per unit, such as 3 homeworks, 2 labs, and 1 exam. Properties for assessment units are as follows.
```
jonatanschroeder, line 129: "typically with a consistent number of assessments per unit, such as 3 homeworks, 2 labs, and 1 exam"

I think this sounds a bit restrictive. I'm ok with making this kind of suggestion, but in my courses I have some units that take one class (and thus less homeworks), while other units take a couple of weeks.

docs/course.md
```diff
@@ -124,6 +124,31 @@ The assessment set order in `infoCourse.json` is the order in which the assessme
...
+
+Property | Description
+--- | ---
+`name` | One word to identify the unit, lowercase.
+`heading` | Longer title for the unit, displayed to users.
+
```
jonatanschroeder, line 135: I think this should move to after the example.

docs/course.md
```diff
@@ -124,6 +124,31 @@ The assessment set order in `infoCourse.json` is the order in which the assessme
...
+Units are optional and do not affect any behavior by default. Setting the `infoCourse.json` property `"groupBy"` to `"unit"` (over the default `"set"`) will cause assessments to be grouped together by unit on the student assessments overview page. Because units are defined in an array, the order that they are defined in will be the order that they display in, top-down, to students. This can let students view their list of assessments in a chronological order, rather than simply grouped by set.
```
jonatanschroeder, line 136:
```suggestion
Units are optional and do not affect any behavior by default. Setting the `infoCourse.json` property `"groupBy"` to `"unit"` (instead of the default `"set"`) will cause assessments to be grouped together by unit on the student assessments overview page. The order that they are defined in will be the order that they display in, top-down, to students. This can let students view their list of assessments in a chronological order, rather than simply grouped by set.
```

pages/instructorAssessments/instructorAssessments.ejs
```diff
@@ -65,6 +66,7 @@
                   <% } %></a>
                   <%- include('../partials/issueBadge', {count: row.open_issue_count}); %>
                 </td>
+                <td><%= row.assessment_unit_name %></td>
```
jonatanschroeder, line 69: Would it make sense to group it by unit for instructors as well, so instructors see them in the same order as students?

sprocs/sync_assessments.sql
```diff
@@ -414,6 +416,14 @@ BEGIN
...
         AND a.course_instance_id = syncing_course_instance_id
         AND (da.errors IS NOT NULL AND da.errors != '');
+    
+    -- Ensure all assessments have an assessment unit, default number=0.
+    UPDATE assessments AS a
+    SET
+        assessment_unit_id = COALESCE(a.assessment_unit_id,
+            (SELECT id FROM assessment_units WHERE number = 0 AND course_id = syncing_course_id))
+    WHERE a.deleted_at IS NULL
+    AND a.course_instance_id = syncing_course_instance_id;
```
jonatanschroeder, line 426: I believe this is cleaner and faster, but I haven't tested it:

```suggestion
    UPDATE assessments AS a
    SET
        assessment_unit_id = 
            (SELECT id FROM assessment_units WHERE number = 0 AND course_id = syncing_course_id)
    WHERE a.deleted_at IS NULL
    AND a.course_instance_id = syncing_course_instance_id
    AND a.assessment_unit_id IS NULL;
```

jonatanschroeder (commented):
I will test more extensively later, but here are some preliminary comments.

pages/instructorAssessments/instructorAssessments.ejs
astrosticks: Makes sense. Added. Do you think I should remove this column from this page?

database/enums/enum_assessment_grouping.pg
astrosticks: You're right, added.

sprocs/sync_assessments.sql
astrosticks: I'm not sure why, but a couple of assessments cause an error on sync when I use your suggestion.

jonatanschroeder: Did you find any particular feature in common between these assessments (e.g., they are assessments with a unit, without a unit, with a unit listed in the course.json, or not, etc.)?

pages/instructorAssessments/instructorAssessments.ejs
jonatanschroeder: Maybe show the unit if sorting by set, or the set if sorting by unit?

astrosticks: The set is already indicated by the badge next to the assessment name, and if groupBy is set to "set", then knowing the unit is pretty much pointless knowledge

jonatanschroeder: Good point.

sprocs/sync_assessments.sql
astrosticks: I see now that a couple of new assessments have been added to exampleCourse, obv. without units. The error is being thrown from line 472, presumably because they're making it to the end of the sync procedure with a null assessment_unit_id. Not sure why your code wouldn't work though.

On that note, I think I should come up with new units for the exampleCourse, so that when people do add new courses to it, it's more obvious what unit they should belong to.

pages/instructorAssessments/instructorAssessments.ejs
astrosticks: I think I'll go ahead and remove it. If people want it, it's not that hard to add back in

migrations/999_assessment_units__add.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS assessment_units (
```
jonatanschroeder, line 1: This is close to being merged, can you update the migration number?

pages/instructorAssessments/instructorAssessments.js
```diff
@@ -38,6 +38,7 @@ router.get('/', function(req, res, next) {
         course_instance_id: res.locals.course_instance.id,
         authz_data: res.locals.authz_data,
         req_date: res.locals.req_date,
+        assessments_group_by: res.locals.course.assessments_group_by,
```
jonatanschroeder, line 41: I personally would prefer to have this coming from `pl_courses` itself in the query, though I wouldn't mind being convinced otherwise.

pages/instructorAssessments/instructorAssessments.sql
```diff
@@ -53,11 +60,13 @@ FROM
     LEFT JOIN LATERAL assessments_duration_stats(a.id) AS dstats ON TRUE
     LEFT JOIN LATERAL authz_assessment(a.id, $authz_data, $req_date, ci.display_timezone) AS aa ON TRUE
     LEFT JOIN issue_count AS ic ON (ic.assessment_id = a.id)
+    LEFT JOIN assessment_units AS au ON (au.id = a.assessment_unit_id)
 WHERE
     ci.id = $course_instance_id
     AND a.deleted_at IS NULL
     AND aa.authorized
 ORDER BY
+    (CASE WHEN $assessments_group_by = 'Unit' THEN au.number END),
```
jonatanschroeder, line 69:
```suggestion
    (CASE WHEN $assessments_group_by = 'Unit' THEN au.number, au.id END), 
```
The number is not unique in the database, so I'd prefer to have a unique ID in the sort. I believe the syntax above is fine, otherwise you may need to adjust it in some way.

pages/studentAssessments/studentAssessments.sql
```diff
@@ -124,10 +140,19 @@ SELECT
...
+    ) AS start_new_assessment_group,
+    (CASE WHEN $assessments_group_by = 'Set' THEN assessment_set_heading ELSE assessment_unit_heading END) AS assessment_group_heading
 FROM
     all_rows
 WHERE
     authorized
-ORDER BY
-    assessment_set_number, assessment_order_by, assessment_id, assessment_instance_number NULLS FIRST;
+ORDER BY 
+    (CASE WHEN $assessments_group_by = 'Unit' THEN assessment_unit_number END),
```
jonatanschroeder, line 156: Same as instructor page comment.

jonatanschroeder (commented):
Looking good.

@mwest1066 news item here?

jonatanschroeder:
Can you also add the Unit to the Settings tab of the assessment?

pages/instructorAssessments/instructorAssessments.js
astrosticks: This value is already being authorized by the middleware, and it saves a few lines of SQL joins, I think they're both just as valid

pages/instructorAssessments/instructorAssessments.sql
astrosticks: The number is unique in a course/courseInstance, cuz it isn't explicitly declared in an infoFile, it's taken from the index of the assessmentUnits array in infoCourse.json

jonatanschroeder: I understand that, but I'd prefer to have it either reflected in the DB with a unique constraint or have this id in the sort. It probably won't have a practical effect, but it's foolproofing it for future changes.

astrosticks: Do we know that the unit ids will always be increasing alongside the numbers?

astrosticks: I will go ahead and add it

jonatanschroeder: > Do we know that the unit ids will always be increasing alongside the numbers?

They don't need to, that's why you have both the number and the id

pages/instructorAssessmentSettings/instructorAssessmentSettings.ejs
```diff
@@ -43,11 +43,24 @@
...
+                    <tr>
+                        <th scope="row">Set</th>
+                        <td>
+                            <%= assessment_set.name %> 
+                            <span class="text-muted">(<%= assessment_set.abbreviation %>)</span>
+                        </td>
+                    </tr>
+                    <tr><th scope="row">Number</th><td><%= assessment.number %></td></tr>
+                    <tr>
+                        <th scope="row">Set</th>
```
jonatanschroeder, line 57:
```suggestion
                        <th scope="row">Unit</th>
```

jonatanschroeder (approved):
Missing a news item and the change below. @mwest1066 will review it tomorrow too.

pages/instructorAssessmentSettings/instructorAssessmentSettings.ejs
astrosticks: thanks for catching that haha

pages/studentAssessments/studentAssessments.ejs
```diff
@@ -28,20 +28,20 @@
           </thead>
           <tbody>
             <% rows.forEach(function(row) { %>
-            <% if (row.start_new_set) { %>
+            <% if (row.start_new_assessment_group) { %>
             <tr>
-              <th colspan="4"><%= row.assessment_set_heading %></th>
+              <th class="pl-assessment-group-heading" colspan="4"><%= row.assessment_group_heading %></th>
```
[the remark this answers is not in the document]
nwalters512: Changed in b4897d1.

tests/testAssessmentOrdering.js
```diff
@@ -0,0 +1,123 @@
...
+
+const config = require('../lib/config');
+const sqldb = require('../prairielib/sql-db');
+const sqlLoader = require('../prairielib/sql-loader');
+const sqlEquiv = sqlLoader.loadSqlEquiv(__filename);
+
+const helperServer = require('./helperServer');
+const helperClient = require('./helperClient');
+
+describe('Course with assessments grouped by unit vs set', function () {
```
nwalters512, line 13: I would really like to rewrite this so that it isn't tied to the current state of the `testCourse`. We can hopefully use the sync testing utils to construct arbitrary courses here.

nwalters512:
We should also introduce tests for assessment unit syncing, similar to `tests/sync/assessmentSetsSync.js`.

docs/course.md
```diff
@@ -129,32 +144,127 @@ The following list of standardized assessments sets is automatically included in
...
+    }
+  ]
+}
+```
+
+## Assessment units
+
+Each assessment in the course belongs to a _unit_ defined in `infoCourse.json`. Units should be thought of as a unit of instruction, or in PrairieLearn terms, a collection of assessments related to one another, but not necessarily of the same type. This means you can have a unit with two homeworks, one lab, and one exam, for example.
+
+Units are optional and do not affect any behavior by default. Setting the `infoCourse.json` property `"groupBy"` to `"Unit"` (instead of the default `"Set"`) will cause assessments to be grouped together by unit on the student assessments overview page. The order that they are defined in will be the order that they display in, top-down, to students. This can let students view their list of assessments in a chronological order, rather than simply grouped by set.
```
mfsilva22, line 231: groupBy should not be a property of infoCourse.json, at least not only be defined there. I can see newly created courses that want to use "Units" at the start for all courses instances. However, I believe the use case for this property is the most useful at the course instance level.

mfsilva22 (changes requested):

mfsilva22:
I moved the property `groupBy` out of the course level into the course instance level. The example course remains using the default `groupBy: Set`. I created another course instance for testing, and it all looks good. I have also tested this PR using my course CS 357 (which is organized by units already using sets in infoCourse.json, and the result looks great! 

I was not able to address any of the comments left by @nwalters512  :-(

mfsilva22:
I did not make changes to the "test" files, and they seem to be passing right now. However, I don't think they are actually correct, since they currently `assume assessments_group_by` is still a column in `pl_courses` table. Maybe the file `testAssessmentOrdering.sql` should have instead the following code:

-- BLOCK get_test_course
SELECT assessments_group_by, id
FROM course_instances
WHERE short_name = 'Sp15';

-- BLOCK test_course_assessments_group_by_unit
UPDATE course_instances
SET assessments_group_by = 'Unit'
WHERE short_name = 'Sp15'
RETURNING id;

mfsilva22:
As I was updating my course repo last night, I just realized two aspects of this PR. I am not sure they are concerns, but just want to point them out here.

 Here is an image of my course using this PR, and groupBy: Unit:

image

- Since `assessmentUnits` does not have the property color (that in this case would override the ones defined by the Set property), each Unit has colorful labels, depending on the sets used in the course (I use many of them!). I actually like the use of different colors, since students can identify the type of assessment by its color. I am not sure if my preference will be shared by others. Should we have the color property in `assessmentUnits` as an option?

- when an assessment is selected, the menu bar gives a pull down menu for easy switch between assessments.

image

as of now, the assessments that appear in the pull down menu are still the ones that belong to the same "set", even when "unit" is used for grouping option. Is this the desired behavior? I am not sure I have a preference here, but I don't know if this was an accidental implementation, or part of a design plan.

mwest1066 (changes requested):
Sorry to bikeshed this, but can we think about the name "unit"? At least to me this doesn't immediately suggest what it does. I'm wondering about something like "section" or "topic"?

mfsilva22:
> 

I agree with you. I also prefer "Topic"

mfsilva22:
Made the changes to the PR to replace "Units" with "Modules".

image

nwalters512:
I'm working on improving the tests locally. One thing that came to mind as I was doing so: if someone has opted to group by module, should we still force them to set a set for their assessments? Currently, `set` is marked as a required property for all assessments: https://github.com/PrairieLearn/PrairieLearn/blob/8589f0c44cfb7a54f829b3d81c7d2e550d64e118/schemas/schemas/infoAssessment.json#L11.

schemas/schemas/infoCourseInstance.json
```diff
@@ -93,6 +93,15 @@
                     }
                 }
             }
+        },
+        "groupBy": {
```
nwalters512, line 97: Should we call this `groupAssessmentsBy` instead?

mwest1066:
> I'm working on improving the tests locally. One thing that came to mind as I was doing so: if someone has opted to group by module, should we still force them to set a set for their assessments?

I think we should still force this, at least initially. We can always review this later on to relax it. On a practical level, we assume in many places in the code that we have a valid assessment set (e.g., to render the colored label like "HW1" for an assessment).

schemas/schemas/infoCourseInstance.json
mwest1066: Agreed.

nwalters512: I made this change in [c7bf646](https://github.com/PrairieLearn/PrairieLearn/pull/4360/commits/c7bf646e4c21e794539d4e6b7c7487eac32029fb)

mfsilva22:
@nwalters512 I checked my course branch using this PR, and it looks great! I will be super happy to see this merged :-)

mfsilva22 (approved):

pages/instructorAssessments/instructorAssessments.sql
```diff
@@ -53,11 +59,14 @@ FROM
     LEFT JOIN LATERAL assessments_duration_stats(a.id) AS dstats ON TRUE
     LEFT JOIN LATERAL authz_assessment(a.id, $authz_data, $req_date, ci.display_timezone) AS aa ON TRUE
     LEFT JOIN issue_count AS ic ON (ic.assessment_id = a.id)
+    LEFT JOIN assessment_modules AS au ON (au.id = a.assessment_module_id)
```
mwest1066, line 62: Could we rename `au` here to `am`, and similarly in other SQL files?

pages/studentAssessments/studentAssessments.sql
```diff
@@ -103,6 +118,7 @@ WITH
             JOIN assessment_sets AS aset ON (aset.id = a.assessment_set_id)
             LEFT JOIN assessment_instances AS ai ON (ai.assessment_id = a.id AND (ai.user_id = $user_id OR ai.group_id = gu.group_id))
             LEFT JOIN LATERAL authz_assessment(a.id, $authz_data, $req_date, ci.display_timezone) AS aa ON TRUE
+            LEFT JOIN assessment_modules AS au ON (au.id = a.assessment_module_id)
```
mwest1066, line 121: Another place to rename `au` to `am`.

middlewares/selectAndAuthzAssessment.sql
```diff
@@ -2,11 +2,13 @@
 SELECT
     to_jsonb(a) AS assessment,
     to_jsonb(aset) AS assessment_set,
+    to_jsonb(au) AS assessment_module,
     to_jsonb(aa) AS authz_result,
     assessment_label(a, aset) AS assessment_label
 FROM
     assessments AS a
     JOIN assessment_sets AS aset ON (aset.id = a.assessment_set_id)
+    JOIN assessment_modules AS au ON (au.id = a.assessment_module_id)
```
mwest1066, line 11: Another `au` to `am` rename.

mwest1066 (approved):
Approved conditional on the `au` -> `am` rename.

news_items/038_organize_by_module/index.html
```diff
@@ -0,0 +1,20 @@
...
+    You can now declare a list of modules for your course inside
+    of your course configuration. These are collections of
+    assessments that can belong to different sets, which get 
+    displayed together in the order that you declare them. You 
+    can think of them as chapters, sections, or topic lists in your course. 
+    If you want to group together several types of assessments 
+    (for example, two homeworks and an exam), now you can!
+</p>
+<img 
+    src="module-example.png"
```
mwest1066, line 11: In this example, could we change the names of the modules to be something that's more obviously freely chosen (not just numbered). Maybe:
* "Introductory module"
* "Another module"
* "The final module"

mfsilva22: @mwest1066 this example no longer exists (this was a separate course instance that I created at some point, using the Module for sorting, but we ended up deleting it). Should I just create another example?

mwest1066: If it's not too hard to do then I think it would be nice. The existing example makes it look like the modules are called "Module 1", "Module 2", etc (like HWs, Exams, etc)

PR #4362, opened 2021-06-09
Regrading a single question for a single student

echau01:
This is a continuation of the work done in pull request #3452.

A summary of changes made since the end of #3452:
- regrading multiple submissions on an exam or homework question
- disallow regrading of questions with external grading (for now)
- extensive unit tests for regrading exam and homework questions
- in `lib/question.js`, merged `_regradeVariantWithClient()` function into `_gradeVariantWithClient()` function, thus reducing code duplication
- implemented regrading for submissions incorrectly marked as invalid (as long as the submission is "regradable" in the sense of #4406)
- resolved the "error: undefined {"data":{"rollback":"success"}}" logger message described in one of the TODO items of #3452.
- rewording of messages
- rewrote documentation for regrading


echau01:
## Issues requiring clarification

### What happens to the database if we keep the original score instead of the regraded score?

When a question is regraded for a student, all regrading occurs inside a single transaction. During regrading, several tables are modified, such as `grading_jobs`, `submissions`, `variants`, `instance_questions`, and `assessment_instances`. If we want to keep the original score after regrading (because the regraded score is lower), should we "rollback" all of the changes made during the transaction? The current behaviour is to rollback **some** changes, but not all changes -- for example, the score for the instance question is reset, but the scores for each submission are not reset. I think we should rollback all changes made to the database for two reasons:

- The main reason is to avoid inconsistencies in the database. For example, suppose a student makes one submission for a question instance. When graded, that submission earns 100%, so the instance question's score is 100%. Later, the `grade` function is modified. Suppose that when the question is regraded, the submission earns 50%, but the instructor chooses to discard the regraded score if it is lower than the original score, so the question's score remains at 100%. Now, we have an inconsistency: the question score is 100%, but the submission score is 50%. This inconsistency shows up in question statistics.
- The secondary reason is that if we only rollback some changes, then we need to write more code to explicitly rollback the desired changes, instead of just letting `sqldb.endTransaction` rollback everything.

### Regrading invalid submissions

Should submissions that are marked "invalid" be regraded? My thoughts on this issue:

- Suppose a question's `parse` function is coded incorrectly so that submissions are being marked as invalid when they are supposed to be valid. When the `parse` function is corrected and the question is regraded, I think we should regrade the invalid submissions because they might now be valid (i.e. gradable).
- I think we should only regrade the invalid submissions for which grading was requested. However, we currently cannot distinguish invalid submissions made using the "Save & Grade" button vs. submissions made using the "Save Only" button. Going forward, we can add a new boolean `grading_attempted` column to the `submissions` table in order to identify invalid submissions that we have tried to grade previously. However, I don't see a way to update already-existing submissions in the database -- currently, we cannot tell whether or not grading was attempted for any particular invalid submission.

### Question-specific `regradeActive` flags?

In the original PR, one of the TODO items is "to modify the `regradeActive` flag so it's question specific." If my understanding is correct, the suggestion is to set a `regradeActive` property for each question (probably in the `info.json` file) to indicate whether a specific question should be regraded or not. However, I don't see the need for this -- the instructor can already choose which questions to regrade because each question has its own "Regrade" button. If someone could please clarify this TODO item, I would appreciate it.

echau01:
### An unsolved problem: the difficulty of regrading a question

Currently, to regrade a question, instructors must write a new grading function (or modify the existing one) in the question's `server.py` or `server.js` file. Many people would find it easier if they could just modify the values of the correct answers in the question configuration. However, this will not work because the correct answers for a question variant are only computed when the variant is generated; (re)grading a submission does not compute the correct answers anew.

Regrading is difficult because in order to write a grading function, one needs some knowledge of the internal workings of the specific PL elements used in the question. As an example, consider a multiple choice question with the following `question.html` file:

```
<pl-question-panel>
    <p>The correct answer should be "Option 2".</p>
</pl-question-panel>

<pl-multiple-choice answers-name="ans" weight="1">
    <pl-answer correct="true">Option 1</pl-answer>
    <pl-answer>Option 2</pl-answer>
</pl-multiple-choice>
```

We have mistakenly set the "Option 1" choice to be the correct answer when in fact "Option 2" should be correct. To regrade this question, we cannot just set the "Option 2" choice to be the correct answer in the `question.html` file. Instead, we must write a `grade` function in `server.py` like so:

```
def grade(data):
    ans_key = data['submitted_answers']['ans']
    selected_option = None
    
    for choice in data['params']['ans']:
        if choice['key'] == ans_key:
            selected_option = choice
            break
    
    if selected_option['html'] == 'Option 2':
        data['partial_scores']['ans'] = {'score': 1, 'weight': 1}
        data['score'] = 1
    else:
        data['partial_scores']['ans'] = {'score': 0, 'weight': 1}
        data['score'] = 0
```

To be able to write the `grade` function above, we need to know the following characteristics specific to the multiple choice element:
- `data['submitted_answers']['ans']` contains the alphabetic key (`'a'`, `'b'`, `'c'`, etc.) of the multiple choice option that was submitted
- `data['params']['ans']` is an array containing the choices in the multiple choice question. Additionally, each element in the array is a `dict` of the form `{"key": ..., "html": ...}`.

However, if the question used a different PL element (e.g. `pl-order-blocks`), the structure of `data['submitted_answers']['ans']` and the elements of `data['params']['ans']` will be different. The grading function would then have to be tailored toward the internal workings of that PL element. Ultimately, the current regrading system demands too much knowledge of internal implementations of elements.

Other concerns:
- The grading function can easily become very complex if the question has many parts. Normally, PL elements grade themselves using the "correct answers" computed at question variant generation. However, when regrading, the elements no longer grade themselves correctly, so instructors have to essentially re-implement the `grade` functions for the elements used in the question. Bugs are likely to occur as a result.
- If an element's internal implementation changes, then the grading function might break.

mwest1066:
@echau01 to make things easier for reviewers, would you mind switching this PR to a branch in a directly-checked-out copy of the PL repo (rather than a branch in a fork, as this is currently using)?

PR #4366, opened 2021-06-10
Move "assessmentSets", "topics", "tags" JSON retrieval to a single method

astrosticks:
This is a helpful refactor, but it isn't within scope of the other PR I'm working on (#4360), so I put it onto its own branch. This will let me add another section to `infoCourse.json` in the style of "assessmentSets", "topics", etc. without having to copy a bunch of redundant code. 

As I'm looking more and more into the syncFromDisk code, I see a lot of redundant code that's been copy-pasted whenever another one of these fields needed to be implemented. If there's any interest in more PRs like this that collapse all that code into generic implementations, let me know.

### DIFF
--- a/sync/course-db.js
+++ b/sync/course-db.js
@@ -585,6 +585,47 @@ module.exports.loadCourseInfo = async function(coursePath) {
 
     const info = loadedData.data;
 
+    /**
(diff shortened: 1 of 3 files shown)


astrosticks:
@jonatanschroeder So #4360 includes these commits, but they're also here if you want to take a look at them.

PR #4368, opened 2021-06-12
Add Julia autograder and example

ctessum:
Let me know what you think!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,6 +17,8 @@ jobs:
       run: sh ./tools/build-image-if-needed.sh graders/c prairielearn/grader-c
     - name: Build the grader-java docker image if needed
       run: sh ./tools/build-image-if-needed.sh graders/java prairielearn/grader-java
+    - name: Build the grader-julia docker image if needed
+      run: sh ./tools/build-image-if-needed.sh graders/julia prairielearn/grader-julia
     - name: Build the grader-python docker image if needed
       run: sh ./tools/build-image-if-needed.sh graders/python prairielearn/grader-python
     - name: Build the grader-r docker image if needed
--- a/exampleCourse/questions/demo/autograder/julia/fibonacci/info.json
+++ b/exampleCourse/questions/demo/autograder/julia/fibonacci/info.json
@@ -1,0 +1,15 @@
+{
+    "uuid": "29fff7a7-9f67-4dc2-93d1-6d05486bfd3a",
+    "title": "Demo: Julia code, fibonacci function, in-browser editor, external grading",
+    "topic": "Autograder",
+    "tags": ["code", "v3", "julia"],
+    "type": "v3",
+    "singleVariant": true,
+    "gradingMethod": "External",
+    "externalGradingOptions": {
+        "enabled": true,
+        "image": "prairielearn/grader-julia",
+        "entrypoint": "/run.sh",
+        "timeout": 20
+    }
+}
--- a/exampleCourse/questions/demo/autograder/julia/fibonacci/initial_code.jl
+++ b/exampleCourse/questions/demo/autograder/julia/fibonacci/initial_code.jl
@@ -1,0 +1,3 @@
+function fib(n::Int)::Int
+    nothing
+end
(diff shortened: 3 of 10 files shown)


ctessum:

ctessum:
recheck

jonatanschroeder:
I'm trying to check if I can find someone that knows enough Julia to be able to review this more confidently, since I know virtually nothing of Julia, but I have a small concern:

https://github.com/ctessum/PrairieLearn/blob/jl/graders/julia/install.jl#L5

The line seems to be downloading code from your own repo, instead of including it in the grader itself. I believe this could be dangerous, since although this is open source it would require any future contributors to push changes to your repo instead of to PrairieLearn, leading to some confusion over PR practices. Also, there would need to be some form of synchronization to ensure the PL grader image is properly updated when your code changes. @mwest1066 thoughts?

mwest1066:
> The line seems to be downloading code from your own repo, instead of including it in the grader itself.

Good point!

@ctessum can you include the grader itself with this PR? For example, the python autograder is in [graders/python/python_autograder](https://github.com/PrairieLearn/PrairieLearn/tree/master/graders/python/python_autograder) and the C autograder is in [graders/c/cgrader](https://github.com/PrairieLearn/PrairieLearn/tree/master/graders/c/cgrader).

ctessum:
> @ctessum can you include the grader itself with this PR? For example, the python autograder is in [graders/python/python_autograder](https://github.com/PrairieLearn/PrairieLearn/tree/master/graders/python/python_autograder) and the C autograder is in [graders/c/cgrader](https://github.com/PrairieLearn/PrairieLearn/tree/master/graders/c/cgrader).

Sure! My original thought was that by making it platform-agnostic it would be more likely to attract contributions from the Julia community, but this makes sense too.

ctessum:
> @ctessum can you include the grader itself with this PR?

So one thing is that Grader.jl currently has its own CI ([here](https://github.com/ctessum/Grader.jl/tree/main/.github/workflows)), which runs a test suite which by convention is assumed to be in the `/test` directory, which is close to but not the same as the existing `/tests` directory in PL. This directory location appears to be hard-coded into the `julia-runtests` github action. Is there a recommendation for the best way to handle this? The options I can think of are 1) adding a new `/test` directory for use by Julia, 2) making a custom github action that uses `/tests` instead of `/test`, or 3) dropping the test suite. None of these options seem really elegant.

I'm willing to do whatever you all think is best, but If, alternatively, Grader.jl stays in its own repository, it would be straightforward to dictate a specific version to install, e.g. `Pkg.add(url="https://github.com/ctessum/Grader.jl:v0.1.0")`. Changing the version in `install.jl` would trigger a new Docker build, and if I stop being responsive for whatever reason someone can make a fork and change the dependency, or the repository could be moved to the PrairieLearn organization to start with, to make it less tied to me.

Either way, it's your decision, just let me know what you prefer.

nwalters512:
Hey @ctessum, I was just running through some old PRs. I think the direction we want to move in the future is to have graders like these live in separate repositories so that they can be fully owned by you, as it doesn't scale well to have the main repository contain code that the core maintainers aren't familiar with. We'd of course be happy to link to third-party graders in our docs.

We haven't settled on this approach though, so I wanted to see how that sounds to you!

ctessum:
I think that makes sense. I think the grader code itself needs to be in a separate repository for version control purposes, otherwise a breaking change to the grader would break every course everywhere that uses it.

The other part though is the example question(s) that show people how to create a question that uses Julia (or some other language). It probably doesn't scale to have those live in the main repository either, but would it make sense to set up some kind of system with like a library of 3rd-party repos providing example questions in different languages / courses etc?

nwalters512:
While that would be neat to build in the long term, it's not something that's likely to be a priority for the time being. I'd be much more willing to add a few questions to the example course than an entire grader implementation!

ctessum:
Sounds good. If you want to link to the grader, it is here: https://docs.juliahub.com/Grader/

Perhaps if someone else wants to use Julia and PL together just let me know and we can figure out the best way to make that happen.

PR #4406, opened 2021-06-23
Add new eligible_for_regrading column to submissions table

echau01:
This pull request will help with #4362. 

A submission is `eligible_for_regrading` if it is meant to be automatically graded. More specifically, any submission that we try to grade using `_gradeVariantWithClient()` in `lib/question.js` is eligible for regrading. As a result, the following submissions have `eligible_for_regrading = True`:
- submissions made with the "Save & Grade" button
- submissions made with the "Save Only" button that are automatically graded when an assessment closes

Note that manual grading does not change the `eligible_for_regrading` status of a submission.


database/tables/submissions.pg
```diff
@@ -5,6 +5,7 @@ columns
     credit: integer
     date: timestamp with time zone default CURRENT_TIMESTAMP
     duration: interval default '00:00:00'::interval
+    eligible_for_regrading: boolean not null default false
```
mwest1066, line 8: Maybe rename to `regradable`?

echau01: I like that name -- it's a lot shorter.

PR #4411, opened 2021-06-23
Directly require() PrairieLib files instead of using it as a package

mwest1066:
Now that PrairieLib is directly in the monorepo, this PR changes all files to directly `require()` files from `./prairielib/` rather than using them via the `@prairielearn/prairielib` package.

The conversion here is done by creating the script `tools/replace_prairielib_paths.py` and then running it. The script itself does all the conversions to other files.

The advantage of using PrairieLib files directly is that:
1. The full test-suite checks any changes to PrairieLib.
2. We can make a simultaneous change to PrairieLib and PrairieLearn and test both changes together. This is used in PR #4412, which changes how SQL stored procedures are run.


nwalters512 (approved):
Time is a flat circle 😛 it's only a matter of time before this all ends up back in `/lib`!

Took a quick skim through, this all looks correct, and the passing CI gives me confidence to approve.

mwest1066:
> Time is a flat circle 😛 it's only a matter of time before this all ends up back in `/lib`!

It's the external monolith/micro-services tradeoff!

> Took a quick skim through, this all looks correct, and the passing CI gives me confidence to approve.

Thanks!

Issue #3824, opened 2021-02-13
Problems with deployment of #2626


mwest1066:
Although #2626 was merged, problems were encountered during deployment and so it was reverted. The issues are:

1. After deployment we started to accumulate large numbers of "stuck" running queries from here: https://github.com/PrairieLearn/PrairieLearn/blob/7543da03935e54f0462d5b2aca0f52f7577cd55e/middlewares/selectAndAuthzInstanceQuestion.sql#L1 I don't know what's causing this. I can't see anything that looks risky in the PR, but something is clearly wrong. I'll try and do some more testing on a live staging server sometime next week.

2. While attempting to test the above issue, I noticed that the migration creates `effective_advance_score_perc` as `NULL`, which would be fine except that this comparison: https://github.com/PrairieLearn/PrairieLearn/blob/fb99f8f1f779354876288f9b74224438d3845308/sprocs/question_order.sql#L23 will then always result in `locking = NULL` for questions other than the first (because `FALSE OR NULL = NULL`, see https://www.postgresql.org/docs/current/functions-logical.html) and hence `sequence_locked = NULL` lower in the query.

    The `sequence_locked` value is eventually used here: https://github.com/PrairieLearn/PrairieLearn/blob/fb99f8f1f779354876288f9b74224438d3845308/middlewares/selectAndAuthzInstanceQuestion.sql#L89 and `AND NOT sequence_locked` becomes `AND NOT NULL` which is `AND NULL`, so access to every question is locked. This will be fixed the next time someone syncs that course, but until then no student can access any question in the course (except the first question in each assessment).

    I haven't thought carefully about whether it would be better to make `effective_advance_score` be `DEFAULT 0 NOT NULL` and leave all the code as is, or to change the code to deal safely with `effective_advance_score = NULL`.  I guess the former is probably simpler.

astrosticks:
@mwest1066 Attempting to merge the latest changes from master into forceSequential, I'm getting an error when trying to start the server: `{...,"routine":"errorMissingColumn","message":"column aq.effective_advance_score_perc does not
 exist"},...`. Could this have something to do with the fact that `migrations/227_force_sequential_questions__add.sql` was reverted, but its index hasn't been replaced by another migration since then, leaving the docker image thinking that migration 227 already happened?

mwest1066:
We are trying again with #4497 now that #4412 is deployed.

nwalters512:
@mwest1066 i believe we can close this now?

[...]

PR #4412, opened 2021-06-23
Use separate PostgreSQL schemas for sprocs

mwest1066:
This PR changes `sql-db` so that it can use a local [DB schema](https://www.postgresql.org/docs/current/ddl-schemas.html) for all queries. We use this to store a separate set of sprocs (stored procedures) for each invocation of the DB. The startup sequence is:

1. Start with only the default `public` schema.
2. Run migrations, which will modify the global public tables.
3. Set the default schema to a random string (with fallback to `public`).
4. Run sproc init, which will create all the sprocs from scratch in the random schema.
5. For all future SQL calls, use the random schema so that we get our local copy of the sprocs.

This is intended to avoid problems where one server startup modifies sprocs that are currently in use by another server. Although Postgres has transactional DDL, this can still cause problems with deadlocking via explicit locks. I think this is responsible for #3824

Note that this PR requires #4411 to work, because the changes are in PrairieLib and so we need to be using it directly to pick up the new code.


server.js
```diff
@@ -1136,6 +1136,12 @@ if (config.startServer) {
                 callback(null);
             });
         },
+        function(callback) {
```
nwalters512, line 1139: I'd appreciate a few lines of comments here talking about why we do this and the importance of the sequence between migrations, random schema, and sproc creation.

nwalters512 (approved):
This makes sense to me - left a few comments, but I don't think they're blocking. I can't see any way this could have a negative impact (famous last words, right)?

I don't see anything that specifically instructs Postgres to create sprops in the random schema - will Postgres automatically create them in the first element of `search_path`?

mwest1066:
> I don't see anything that specifically instructs Postgres to create sprops in the random schema - will Postgres automatically create them in the first element of `search_path`?

Yep. This is also why it's important to run migrations (which might create tables) before setting the random schema.

See https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PATH where it says:

> The first schema in the search path that exists is the default location for creating new objects.

I also checked that we really are creating the sprocs in the random schema, as expected.

mwest1066:
> > I don't see anything that specifically instructs Postgres to create sprops in the random schema - will Postgres automatically create them in the first element of `search_path`?
> 
> Yep. This is also why it's important to run migrations (which might create tables) before setting the random schema.

An alternative which I considered is to explicitly set the schema when you create objects. For example, we could do
```
CREATE FUNCTION ${schemaName}.variants_insert
```
for all our sprocs to make sure they are created in exactly the schema we want. However, this is pretty invasive and I believe that using the implicit "create in first search_path entry" behavior is safe.

nwalters512:
Yep, as long as that behavior is documented, I'm good with relying on it! Thanks for the link to the docs.

mwest1066:
@nwalters512 Another question is whether we should do any automatic cleanup of either:
1. The old sprocs in the public schema.
2. The new schemas that we are creating every time we restart the server.

My current thoughts are to defer (1) until after we've deployed this PR and are happy with things. We could then add a migration to clean up the public sprocs.

For (2) we can either ignore it for now, relying on the fact that having extra schemas lying around doesn't really hurt anything, or we could add code to automatically delete any old schemas for our current server name. That is, the schemas are named like `i_329c8aef93_2021-06-21-SDLFK` where the first bit is the EC2 instance ID, so on startup we can delete any schemas that start with our own instance ID. We always run all servers with unique IDs, although it would be possible to override this which would lead to problems if we auto-delete.

Thoughts?

nwalters512:
Yeah, I'm happy to ignore the ones in the public schema almost indefinitely (though we can of course clean them up in the future).

I believe (2) gets trickier once we throw auto scaling and general transient servers into the mix. We're not ever guaranteed to even restart a server on an instance once the process is killed or dies. I'm inclined to instead use the date somehow, e.g. deleting all schemas with creation dates over 30 days ago. That would of course cause issues if there happens to be a server that's alive for 30 days. I'll have to think more about this!

nwalters512:
Yeah, I'm happy to ignore the ones in the public schema almost indefinitely (though we can of course clean them up in the future).

I believe (2) gets trickier once we throw auto scaling and general transient servers into the mix. We're not ever guaranteed to even restart a server on an instance once the process is killed or dies. I'm inclined to instead use the date somehow, e.g. deleting all schemas with creation dates over 30 days ago. That would of course cause issues if there happens to be a server that's alive for 30 days. I'll have to think more about this!

mwest1066:
> Yeah, I'm happy to ignore the ones in the public schema almost indefinitely (though we can of course clean them up in the future).

Sounds good.

> I believe (2) gets trickier once we throw auto scaling and general transient servers into the mix. We're not ever guaranteed to even restart a server on an instance once the process is killed or dies. I'm inclined to instead use the date somehow, e.g. deleting all schemas with creation dates over 30 days ago. That would of course cause issues if there happens to be a server that's alive for 30 days. I'll have to think more about this!

That's a good point that autoscaling would mean instances never come back. However, for autoscaling we will soon need to add a table like [server_loads](https://github.com/PrairieLearn/PrairieLearn/blob/master/database/tables/server_loads.pg) (which is for the grader hosts) but for `webserver_loads`, so we can track how many webservers there are and load levels. Each server updates its row every 10 seconds while it's alive, so it's easy to tell when servers are dead. We use this already for grader hosts to kill them off. The same logic would let us clean up old schemas.

prairielib/lib/sql-db.js
```diff
@@ -419,14 +479,35 @@ module.exports.query = function(sql, params, callback) {
             return true;
         };
         if (handleError(err)) return;
-        paramsToArray(sql, params, function(err, newSql, newParams) {
-            if (err) err = error.addData(err, {sql: sql, sqlParams: params});
+
+        const setSearchPath = function(callback) {
+            if (searchSchema != null) {
+                const setSearchPathSql = `SET search_path TO ${searchSchema},public`;
```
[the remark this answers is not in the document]
mwest1066: For legacy reasons the "explicit client" functions (`getClient()`, `queryWithClient()`, `queryWithClientOneRow()`, etc.) are completely separate from the "implicit client" functions (`query()`, `queryOneRow()`, etc.). This means that the `SET search_path` needs to be done in two separate places in the code.

Any given caller will either be using the "explicit client" functions, so they'll get a client with `search_path` set, or they just be doing a one-off `queryXXX()` call with an implicit client, which will set the `search_path` for just that single call.

Maybe it would be cleaner to rewrite `query()` to use `getClient()` underneath, so that we only have to do `SET search_path` in one place?

prairielib/lib/sql-db.js
```diff
@@ -180,7 +182,65 @@ module.exports.getClient = function(callback) {
             }
             return ERR(err, callback); // unconditionally return
         }
-        callback(null, client, done);
+        if (searchSchema != null) {
+            const setSearchPathSql = `SET search_path TO ${searchSchema},public`;
```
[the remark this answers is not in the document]
mwest1066: I discovered that `node-postgres` has an [undocumented](https://github.com/brianc/node-postgres/issues/1978) function called [`escapeIdentifier()`](https://github.com/brianc/node-postgres/pull/396) that does what we want, so I used that as part of the interpolation.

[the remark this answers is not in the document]
mwest1066: An alternative would be to do something like:
```
EXECUTE 'SET search_path TO ' || quote_ident(schema) || ',public;';
```
or
```
EXECUTE format('SET search_path TO %I,public;', schema);
```
See https://www.postgresql.org/docs/12/plpgsql-statements.html#PLPGSQL-STATEMENTS-EXECUTING-DYN

The downside of the above approach is that it requires PL/pgSQL for EXECUTE, not plain SQL, but this isn't to bad if we use a [`DO`](https://www.postgresql.org/docs/12/sql-do.html) statement.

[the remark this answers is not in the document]
mwest1066: Actually, it turns out that you can't use parameters with `DO` statements in postgres. See [this](https://github.com/brianc/node-postgres/issues/812). So I'll stick with using `escapeIdentifier()`.

prairielib/lib/sql-db.js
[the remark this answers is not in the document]
mwest1066: I rewrote `query()` to just call `getClient()` and `queryWithClient()`. This is a pretty fundamental function to rewrite, but the new version is really short and seems safe. This is a big cleanup in that the old `query()` was pretty messy and we now only have `SET search_path` in one place in the code.

server.js
mwest1066: Excellent point. I added both a comment here in the code and also docs in `docs/dev-guide.md`.

mwest1066:
@nwalters512 I updated a bunch of things and I believe it's ready to go. Mind taking another quick look?

docs/dev-guide.md
```diff
@@ -275,6 +275,31 @@ FROM
...
+
+```
+set search_path to server_2021_06_29t19_28_32_247z_e3khma,public;
+```
+
+* During startup we initially have no non-public schema in use. We first run the migrations to update all tables in the `public` schema, then we call `sqldb.setRandomSearchSchema()` to activate a random per-execution schema, and we run the sproc creation code to generate all the stored procedures in this schema. This means that every invocation of PrairieLearn will have its own local copy of the stored procedures which are the correct versions for its code. This lets us upgrade PrairieLearn servers one at a time, while old servers are still running with their own copies of their sprocs. When PrairieLearn first starts up it has `search_path = public`, but later it will have `search_path = server_2021_06_29t19_28_32_247z_e3khma,public` so that it will first search the random schema and then fall back to `public`. The naming convention for the random schema uses the local instance name, the date, and a random string.
```
nwalters512, line 300: Very small nit: "local" implied "on the instance running PL" the first time I read it, I think this is slightly clearer?
```suggestion
* During startup we initially have no non-public schema in use. We first run the migrations to update all tables in the `public` schema, then we call `sqldb.setRandomSearchSchema()` to activate a random per-execution schema, and we run the sproc creation code to generate all the stored procedures in this schema. This means that every invocation of PrairieLearn will have its own locally-scoped copy of the stored procedures which are the correct versions for its code. This lets us upgrade PrairieLearn servers one at a time, while old servers are still running with their own copies of their sprocs. When PrairieLearn first starts up it has `search_path = public`, but later it will have `search_path = server_2021_06_29t19_28_32_247z_e3khma,public` so that it will first search the random schema and then fall back to `public`. The naming convention for the random schema uses the local instance name, the date, and a random string.
```

prairielib/lib/sql-db.js
```diff
@@ -403,31 +420,12 @@ module.exports.endTransactionAsync = promisify(module.exports.endTransaction);
...
-            const sqlError = JSON.parse(JSON.stringify(err));
-            sqlError.message = err.message;
-            err = error.addData(err, {sqlError: sqlError, sql: sql, sqlParams: params});
-            ERR(err, callback);
-            return true;
-        };
-        if (handleError(err)) return;
-        paramsToArray(sql, params, function(err, newSql, newParams) {
-            if (err) err = error.addData(err, {sql: sql, sqlParams: params});
+    module.exports.getClient((err, client, done) => {
```
nwalters512, line 423: Love this cleanup! 🚀

sprocs/array_and_number.sql
```diff
@@ -1,7 +1,6 @@
...
-    IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'array_and_number') THEN
-        CREATE TYPE array_and_number AS (arr DOUBLE PRECISION[], number INTEGER);
-    END IF;
-END$$;
+-- We do not test whether the type already exists.
+-- This is because we will be creating it within the current default schema, which will be empty.
+-- We could test whether the type exists in any schema by querying pg_type like in migrations/000_initial_state.sql.
+-- But that would test for the type in any schema, and we only care about schema on the search path.
+-- It's hard to restrict the "does this type exist" query to just the current search path.
+CREATE TYPE array_and_number AS (arr DOUBLE PRECISION[], number INTEGER);
```
nwalters512, line 6: Ahh, this is interesting. Was checking for the existence of the type actually causing problems? It makes sense that we don't check for existence, but it's not clear to me if that's because we don't have to or because it actually causes problems if we do.

nwalters512 (approved):
Awesome! Fingers crossed that this makes deploys even more robust 🤞

sprocs/array_and_number.sql
mwest1066: It was actually causing problems. It took me an embarrassingly long time to figure out what was causing the following bug:
1. Start with an empty DB.
2. Start PL. It successfully creates the sprocs in `random_schema_1`, including the type `array_and_number`.
3. Stop and start PL again. It creates `random_schema_2` and starts creating the sprocs.
4. It runs `sprocs/array_and_number.sql` which first checks whether the type exsists.
5. The `array_and_number` exists in the `pg_type` table (because it's in the DB in `random_schema_1`, even though this is not in the current search path). We thus don't create the `array_and_number` type again in `random_schema_2`.
6. It runs `sprocs/array_avg.sql` which tries to use the `array_and_number` type and fails, because this type only exists in `random_schema_1` which is not in the current search path.

The fundamental issue here is that postgres doesn't have a `CREATE TYPE IF NOT EXISTS`, which is why we were using the more complex solution of looking for the type in the system `pg_type` table. Postgres has `CREATE TABLE IF NOT EXISTS`, `CREATE FUNCTION IF NOT EXISTS`, etc, but doesn't have the equivalent for types (so annoying!).

There are various ways to fix this:
1. Figure out how to check whether the type exists just in the current search path. This is possible using the [`current_schema()`](https://www.postgresql.org/docs/12/functions-info.html) function, but the checking query is a bit complex.
2. Shift all types to the `public` schema which we maintain with `migrations/`. This means that would only ever be one global copy of the type, so our existing checking system would continue to work fine. The downside of this is that this type is only used for internal purposes inside the `sprocs/array_avg` function, so it's conceptually cleaner to keep it together with the function.
3. The solution I used: just skip the "exists" check and assume we will always be creating into an empty schema. This is fine with our new schema setup, but it will break if we revert back to the old way of putting everything into the public schema.

Do you have any suggestions for how to explain this better in the comment in the file?

nwalters512: Gotcha, thanks for the detailed explanation. Let me take a shot an an improved comment first...

```suggestion
-- Ideally, we'd use something like `CREATE TYPE IF NOT EXISTS` here, but Postgres doesn't offer that.
-- We used to approximate this by checking `IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = ...)`.
-- However, now that we're creating sprocs and types in a local per-process schema, this would no longer work,
-- as the type would exist in `pg_type` but not necessarily in the schemas on the search path.
-- So, instead, we blindly create the type without checking if it first exists, as it should never exist since we're
-- creating it in a fresh, empty schema.
CREATE TYPE array_and_number AS (arr DOUBLE PRECISION[], number INTEGER);
```

Feel free to workshop that a bit more 😛 

Regarding "fixing" this, I think what you have is fine - if we ever decide to go back to just using the public schema for everything, we can re-adjust this type creation to handle that.

I _did_ find something interesting [on Stack Overflow](https://stackoverflow.com/a/48382296/2444312):

```sql
DO $$ BEGIN
    CREATE TYPE my_type AS (/* fields go here */);
EXCEPTION
    WHEN duplicate_object THEN null;
END $$;)
```

It looks like this might be mroe forgiving if we ever switch back to using only the public schema.

mwest1066: Excellent find on the "catch exception" method! That exactly handles the "is this type in the current search path" query.

mwest1066:
Commit 4a966d13ed6cc8afae5646b7dee53584a9c43d01 above changes all sprocs from the old pattern:
```
DROP FUNCTION IF EXISTS my_func()
CREATE OR REPLACE FUNCTION my_func()
```
to the new pattern:
```
CREATE FUNCTION my_func()
```

The reason for this is that the old pattern causes a bug when deploying in an existing multi-server environment. The problem is:
1. Consider running with an existing `prod` server and a `chunk` server, both of which are using the pre-existing code with all sprocs in the `public` schema.
2. Deploy this new random-schema code to `prod`.
3. The `prod` server runs code like `DROP FUNCTION IF EXISTS my_func()`, which will delete the copy of `my_func()` in the `public` schema.
4. The `prod` server creates a new function `my_func()` in its random schema.
5. The `chunk` server is still only using the `public` schema so when it tries to use `my_func()` it dies with a SQL error of `cache lookup failed for function`.

The same problem happens for sprocs without a `DROP FUNCTION` because they all use `CREATE OR REPLACE FUNCTION` which is equivalent to a `DROP FUNCTION` followed by a `CREATE FUNCTION`. In the case that the existing function is in the public schema and we have `search_path = <random_schema>,public`, running `CREATE OR REPLACE FUNCTION` removes the `public` function and creates a new one in the random schema.

To fix this, we could either:
1. On server startup, first use just the `public` schema to run migrations, then set `search_path` to only the random schema (with no fallback to `public`) to run all the `DROP/CREATE` code for sprocs. Finally, set `search_path` to `<random_schema>,public` for actual running.
2. Change all the sprocs to just do plain `CREATE` with no `DROP` or `OR REPLACE`.

I went with (2) even though it's a much more invasive code change because it makes the resulting code much simpler: we have a simpler startup procedure and also individual sproc functions are simpler and cleaner.

mwest1066:
In the process of doing 4a966d13ed6cc8afae5646b7dee53584a9c43d01 I also discovered that we were using an explicit `public` schema here: https://github.com/PrairieLearn/PrairieLearn/blob/27e279f761c940dc44038ffc00cfc9ff2b51251f/sprocs/first.sql#L2 and here: https://github.com/PrairieLearn/PrairieLearn/blob/27e279f761c940dc44038ffc00cfc9ff2b51251f/sprocs/last.sql#L2

I removed both of these so that we will default to the current random schema.

I personally don't like these sprocs because they are using names which are much too generic (`first` and `last`) and could easily cause confusion or conflicts with built-in functions. I'm leaving them as-is for now, however.

PR #4417, opened 2021-06-24
Change sanitize function to properly handle null values

jonatanschroeder:
The `sanitizeObject` function in prairielib uses `typeof value === 'object'` as a way to check if `value` is an object, but this expression is also true for `null`. This has implications in grader hosts, whenever autograders include, in the results, a key that points to `null`. Instead of treating the object as null, this would trigger a parsing error on the results object. This PR fixes it by handling null values explicitly.


nwalters512 (approved):
LGTM, thanks for the fix!

PR #4429, opened 2021-07-01
fixes for collecting files from nested paths inside workspaces

echuber2:
This fixes a few issues the CS 128 staff encountered with collecting nested files from workspaces:

- Make sure the path to the file is created inside the autograder first, or else unpacking the file will fail.
- Guard against `..` anywhere in the resolved relative path, not just at the beginning.
- Clarify documentation about how to specify paths to files. (Currently it *does not work* to just specify a directory and expect the whole directory to be collected, as the docs previously implied; only individual files are collected.)


lib/externalGraderCommon.js
```diff
@@ -117,11 +117,14 @@ module.exports.buildDirectory = function(dir, submission, variant, question, cou
...
-                    if (relPath.split(path.sep)[0] == '..' || path.isAbsolute(relPath)) {
+                    if (relPath.split(path.sep).includes('..') || path.isAbsolute(relPath)) {
                         callback(new Error('Invalid filename'));
                         return;
                     }
-                    fs.writeFile(path.join(dir, 'student', file.name), decodedContents, callback);
+                    // We need to make sure that the directories are created first.
+                    // Here fs is actually fs-extra, so we can use ensureDirSync.
+                    fs.ensureDirSync(path.dirname(fullPath));
+                    fs.writeFile(fullPath, decodedContents, callback);
```
nicknytko, line 127: Can we convert the `ensureDirSync` call to callback style?  I recall that sync code is avoided in the server as it can lead to performance issues.

echuber2: @nicknytko How about now? I put an async/await wrapper.

echuber2: Hold on, this isn't valid. I'll fix it.

echuber2: I think it's correct now. I removed the wrapper (I hadn't properly included the file write within the async function to begin with). Now it's just using the Promise's `catch` and `then`.

lib/externalGraderCommon.js
```diff
@@ -117,11 +117,16 @@ module.exports.buildDirectory = function(dir, submission, variant, question, cou
...
                         callback(new Error('Invalid filename'));
                         return;
                     }
-                    fs.writeFile(path.join(dir, 'student', file.name), decodedContents, callback);
+                    // The version of ensureDir provided by fs-extra
+                    // automatically creates intermediate paths recursively.
+                    const targetDir = path.dirname(fullPath);
+                    fs.ensureDir(targetDir).then(() => {
+                        fs.writeFile(fullPath, decodedContents, callback);
+                    });
```
echuber2, line 129: The part that was in question. Is this version all right?

echuber2 (commented):
@nicknytko Revised

lib/externalGraderCommon.js
mwest1066: Because we are in a callback-style function, we should be using old-style callback code to make sure no exceptions are thrown. This should be something like:
```
fs.ensureDir(targetDir, (err) => {
    fs.writeFile(fullPath, decodedContents, (err) => {
        callback(null);
    });
});
```

mwest1066: See https://prairielearn.readthedocs.io/en/latest/dev-guide/#stack-traces-with-callback-style-functions for more details and examples of this.

lib/externalGraderCommon.js
```diff
@@ -117,11 +117,16 @@ module.exports.buildDirectory = function(dir, submission, variant, question, cou
                     // the directory hierarchy
                     const fullPath = path.join(dir, 'student', file.name);
                     const relPath = path.relative(path.join(dir, 'student'), fullPath);
-                    if (relPath.split(path.sep)[0] == '..' || path.isAbsolute(relPath)) {
+                    if (relPath.split(path.sep).includes('..') || path.isAbsolute(relPath)) {
```
mwest1066, line 120: Is this change to catch potential security issues with `/../` in the middle of paths?

echuber2: Yes, keeping in the spirit of the original version that checked the first part only. Otherwise `foo/../../bar` for example would achieve the same effect as `../bar`. However since the name of the collected file is specified by the question, maybe this isn't so much a security issue as a sanity check on what the staff are doing?

lib/externalGraderCommon.js
echuber2: @mwest1066 I replaced the code with your suggestion. (I wonder if the innermost callback could have just been `callback` as before, though?)

lib/externalGraderCommon.js
mwest1066: Because [`path.join()`](https://nodejs.org/api/path.html#path_path_join_paths) normalizes its output, both arguments to `path.relative()` will be normalized on the previous line. As I understand it, this means that `relPath` can only contain `..` segments at the start of the path, so I think this change has no effect.

I'm ok to make the change anyway, because it's harmless and maybe slightly clearer.

mwest1066 (approved):

lib/externalGraderCommon.js
echuber2: Ah, that makes sense.

Issue #4430, opened 2021-07-01
Bug: pl-checkbox can display more correct options than specified by max-correct attribute


echau01:
I made a new question with the following `question.html` file:

```html
<pl-question-panel>
    <p>In the question.html file, <code>min-correct="2"</code> and <code>max-correct="4"</code>, but all 5 correct options show up as seen below.</p>
</pl-question-panel>


<pl-checkbox answers-name="ans" min-correct="2" max-correct="4" detailed-help-text="true" partial-credit="true">
    <pl-answer>Incorrect 1</pl-answer>
    <pl-answer>Incorrect 2</pl-answer>
    <pl-answer>Incorrect 3</pl-answer>
    <pl-answer>Incorrect 4</pl-answer>
    <pl-answer correct="true">Correct 1</pl-answer>
    <pl-answer correct="true">Correct 2</pl-answer>
    <pl-answer correct="true">Correct 3</pl-answer>
    <pl-answer correct="true">Correct 4</pl-answer>
    <pl-answer correct="true">Correct 5</pl-answer>
</pl-checkbox>
```

The question looks like this:

![Bugged_Checkbox](https://user-images.githubusercontent.com/25561432/124066729-11435000-d9ee-11eb-8482-03dea68d8323.PNG)

All 5 correct options appear, but `max-correct` is set to 4. As a result, a maximum of 4 options can be selected; trying to select all 5 correct options results in an invalid submission. Therefore, there is no way to get 100% on this question.

I think this bug is caused by the following line in `pl-checkbox.py`:

https://github.com/PrairieLearn/PrairieLearn/blob/27e279f761c940dc44038ffc00cfc9ff2b51251f/elements/pl-checkbox/pl-checkbox.py#L67

When the RHS is evaluated, we have:
- `len_correct == 5`
- `number_answers == 9`
- `len_incorrect == 4`
- `min_correct == 2` (this comes from the `min-correct="2"` in the `question.html`)

Thus the RHS is equal to `min(5, min(9, max(0, max(9 - 4, 2))))`, which is 5, so `min_correct` is assigned the value 5. As a result, when the answers are sampled later on (in the lines shown below), there will always be at least 5 correct answers chosen.

https://github.com/PrairieLearn/PrairieLearn/blob/27e279f761c940dc44038ffc00cfc9ff2b51251f/elements/pl-checkbox/pl-checkbox.py#L76-L79

echau01:
[...]

PR #4454, opened 2021-07-04
Checkbox coverage grading scheme

### LINKED ISSUE #4424: "Coverage" checkbox grading
For Physics exams at Illinois we would like a grading scheme for `pl-checkbox` that mimics the way pen-and-paper multiple-choice exams were graded. The idea is that there is a single correct option, but students can select one, two, or three options. The grading scheme is then:
* If the student selects just one option, they get full points if it's correct and zero points otherwise.
* If the student selects two options, they get 1/2 points if one of their selected options is the correct one, and zero otherwise.
* If they select three options, they get 1/3 points if one of the selected options is correct, otherwise zero.
* If the student selects four or more options, they get zero points. In PL we could handle this case as an invalid answer.

To implement this, we should add a new grading scheme for `pl-checkbox` (not the [existing schemes](https://prairielearn.readthedocs.io/en/latest/elements/#details)). It could be called `COV` for "coverage", because the aim is to cover the true options with the student-selected options. We'll define the grading scheme assuming there are `n_total` options and `n_correct` of them are correct. The student selects `n_selected` options, of which `k` are correct. The score is then:
* `base_score = k / n_correct` (this is the fraction of correct answers that were selected)
* `guessing_factor = k / n_selected` (this is the fraction of selected answers that are correct)
* `score = base_score * guessing_penalty`

For example, if there are 3 correct options and the student selects 5 options, two of which are correct, then `base score = 2/3` because they got 2/3 of the correct options, but `guessing_factor = 2/5` because only 2/5 of their selected answers were correct, so their final value is `score = 2/3 * 2/5 = 4/15 = 0.267`.

We should also add `pl-checkbox` options called `min_select` and `max_select`. If the student selects fewer than `min_select` or more than `max_select` options, then their answer is considered invalid and an error is reported. By default `min_select = 0` and `max_select = number_answers`. If either value is not the default then a help message should be displayed next to the checkbox that tells the student the non-default value.

We could also add an option `show-number-correct` that adds help text like `There are exactly 2 correct options in the list above.` This would be especially useful when `n_correct = 1` and we want students to know this.

Given a certain set of selected options, this grading scheme has the properties that:
* It is always helpful to add another correct option (this will increase both `k` and `n_selected` by one, so both the `base_score` and `guessing_factor` will increase).
* It is always harmful to add another incorrect option (this will increase `n_selected` without changing `k`).
* Switching an incorrect answer to a correct option is always helpful (this will increase `k` while keeping `n_selected` the same).
* Switching a correct option to an incorrect one is always harmful (the opposite of the above case).

Additionally:
* The maximum score is one, which is achieved if and only if `k = n_selected = n_correct`.
* If any correct option is selected then the score will be more than zero.
* If all options are selected then the score is `n_correct / n_total`.
* If there is only one correct answer and `max_select = 3` then COV grading is exactly the Illinois Physics multiple-choice grading scheme.

echau01:
Closes #4424.

Summary of changes:
- New coverage (COV) grading scheme implemented as described in original issue
- `show-number-correct` option added to `pl-checkbox` as described in original issue
- `min-select` and `max-select` options added to `pl-checkbox`:
    - The default values of `min-select` and `max-select` are 1 and `number_answers`, respectively.
    - Any submission that selects less than `min-select` options or more than `max-select` options is marked invalid.
    - If `min-select` and/or `max-select` are specified, then the "help text" below the checkboxes displays the range of the number of options that must be selected (i.e. "You must select between X and Y options.").
    - Since the attributes `min-correct` and `max-correct` also restrict the number of selected options in a valid submission, I had to think of a way to handle/avoid potential conflicts with `min-select` and `max-select`. My solution: if `min-select` and/or `max-select` is explicitly specified in the HTML, then the number of selected options a valid submission must contain is determined entirely by the values of `min-select` and `max-select`. As a result, if `min-correct` and/or `max-correct` are specified along with `min-select` and/or `max-select`, then `min-correct` and `max-correct` only restricts the number of correct answers that appear, while `min-select` and `max-select` restrict the number of selected options in valid submissions.


exampleCourse/questions/element/checkbox/question.html
```diff
@@ -183,14 +183,68 @@
         Part 7
     </div>
 
+    <div class="card-body">
```
mwest1066, line 186: I like the new examples in the `element/checkbox` question. Could you also add another question, called something like `demo/multiAnswerMC` which demonstrates how to use `pl-checkbox` to grade in the way described in the first paragraph of #4424. That is, a checkbox question where we have 5 numerical answers, only one of which is correct, and we have `max-select = 3` and `COV` grading?

docs/elements.md
```diff
@@ -171,12 +171,15 @@ Attribute | Type | Default | Description
...
-`hide-help-text` | boolean | false | Help text with hint regarding the selection of answers. Popover button describes the selected grading algorithm ('all-or-nothing', 'EDC' or 'PC')
+`partial-credit-method` | string | 'PC' | Three grading methods for partial credit: 'COV' (Coverage), 'EDC' (Every Decision Counts), and 'PC' (Percent Correct). See explanation below.
+`hide-help-text` | boolean | false | Help text with hint regarding the selection of answers. Popover button describes the selected grading algorithm ('all-or-nothing', 'COV', 'EDC' or 'PC')
 `detailed-help-text` | boolean | false | Display detailed information in help text about the number of options to choose.
 `hide-answer-panel` | boolean | false | Option to not display the correct answer in the correct panel.
 `hide-letter-keys` | boolean | false | Hide the letter keys in the answer list, i.e., (a), (b), (c), etc.
 `hide-score-badge` | boolean | false | Hide badges next to selected answers.
+`min-select` | integer | special | The minimum number of answers that must be selected in any valid submission.
+`max-select` | integer | special | The maximum number of answers that can be selected in any valid submission.
```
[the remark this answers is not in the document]
mwest1066: Can we add some more documentation about exactly how `min/max-correct` interact with `min/max-select`? For example, which of them take priority if we set both of them?

docs/elements.md
```diff
@@ -171,12 +171,15 @@ Attribute | Type | Default | Description
...
+`show-number-correct` | boolean | false | Display the number of correct choices in the help text.
```
mwest1066, line 182: What happens if we select both `show-number-correct` and `detailed-help-text`? Could we add some docs to describe this? Or maybe we should add some checks in the code to only allow one of them to be selected at a time?

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -73,6 +74,16 @@ def prepare(element_html, data):
...
 
+    min_select = pl.get_integer_attrib(element, 'min-select', 1)
+    max_select = pl.get_integer_attrib(element, 'max-select', number_answers)
+
+    if min_select < 1:
+        raise ValueError('The attribute min-select is {:d} but must be at least 1'.format(min_select))
+    if min_select > max_select:
+        raise ValueError('min-select (%d) is greater than max-select (%d)' % (min_select, max_select))
+    if min_select > number_answers:
+        raise ValueError('min-select (%d) is greater than the total number of answers (%d)' % (min_select, number_answers))
```
mwest1066, line 85: Should we add any checks for the relationship between `min/max-correct` and `min/max-select`? I think it doesn't make sense for `min-select > max-correct` or `max-select < min-correct`? I might be mistaken here, so please think about this! 😄

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -73,6 +74,16 @@ def prepare(element_html, data):
     if not (0 <= min_incorrect <= max_incorrect <= len_incorrect):
         raise ValueError('INTERNAL ERROR: incorrect number: (%d, %d, %d, %d)' % (min_incorrect, max_incorrect, len_incorrect, len_correct))
 
+    min_select = pl.get_integer_attrib(element, 'min-select', 1)
+    max_select = pl.get_integer_attrib(element, 'max-select', number_answers)
+
+    if min_select < 1:
```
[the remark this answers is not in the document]
echau01: The way that checkboxes are currently implemented, any answer that doesn't select any checkboxes are considered invalid. That's why I made `min_select` be at least 1. If we want to make `min_select` be 0, submissions that check 0 checkboxes must be considered valid.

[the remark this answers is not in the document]
mwest1066: Right, it's probably just too confusing to allow zero selected answers. Let's forget about this 😄

mwest1066:
@echau01 would you mind replying to the review comments and saying whether you've addressed each of the issues, or whether you decided to do something different? Thanks!

elements/pl-checkbox/pl-checkbox.py
echau01: Fixed. The code throws an error if `min-select > min-correct` or `max-select < max-correct` (I changed the conditions slightly).

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -73,6 +74,16 @@ def prepare(element_html, data):
     if not (0 <= min_incorrect <= max_incorrect <= len_incorrect):
         raise ValueError('INTERNAL ERROR: incorrect number: (%d, %d, %d, %d)' % (min_incorrect, max_incorrect, len_incorrect, len_correct))
 
+    min_select = pl.get_integer_attrib(element, 'min-select', 1)
```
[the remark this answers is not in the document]
echau01: Done -- I also found a place to introduce a constant `MIN_CORRECT_DEFAULT`.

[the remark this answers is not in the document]
mwest1066: Nice!

echau01:
Everything that is now marked as "resolved" has been addressed -- if I modified your suggestions a bit, I left a comment indicating what I did.

mwest1066:
> Everything that is now marked as "resolved" has been addressed -- if I modified your suggestions a bit, I left a comment indicating what I did.

Awesome, many thanks!

docs/elements.md
echau01: They are independent of each other. Setting `show-number-correct` tells you how many correct options there are -- it adds the text "There are exactly X correct options in the list above" to the help text. On the other hand, `detailed-help-text` tells you how many options you have to select -- it adds the text "Select between Y and Z options" to the help text. Therefore, if both are specified, then the help text looks like this: "Select between Y and Z options. There are exactly X correct options."

echau01: I think the fix here is to clarify the documentation for `detailed-help-text`.

echau01: I should also add that in some situations, setting `detailed-help-text="true"` has no effect. This happens if `min/max-select` are set to some non-default value -- in that situation, we always tell students to "Select between `min-select` and `max-select` options" no matter whether `detailed-help-text` is specified or not.

echau01: Documentation for `detailed-help-text` has been added in my latest commit.

docs/elements.md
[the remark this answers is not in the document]
echau01: I wrote up this documentation in my latest commit.

echau01:
To help with review and testing, I created a table below which specifies the full relationship between `min/max-select`, `min/max-correct`, and `detailed-help-text`. These five attributes determine the minimum and maximum number of options that can be selected, as well as what help text is displayed.

Notation:
- `num_ans` denotes the number of displayed checkbox options
- "default" indicates the default value that an attribute takes. The default value for `min-select` and `min-correct` is 1, while the default value for `max-select` and `max-correct` is `num_ans`.
- X, Y, Z, W are non-default values.
- "true" and "false" are the boolean values for `detailed-help-text`

|min-select|max-select|min-correct|max-correct|detailed-help-text|Minimum options to select|Maximum options to select|Help text
:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:
X | Y | Z or default | W or default | true or false | X | Y | Select between X and Y options
X | default | Z or default | W | true | X | W | Select between X and W options
X | default | Z or default | W | false | X | `num_ans` | Select at least X options
default | Y | Z | W or default | true | Z | Y | Select between Z and Y options
default | Y | Z | W or default | false | 1 | Y | Select at most Y options
X | default | Z or default | default | true | X | `num_ans` | Select between X and `num_ans` options
X | default | Z or default | default | false | X | `num_ans` | Select at least X options
default | Y | default | W or default | true | 1 | Y | Select between 1 and Y options
default | Y | default | W or default | false | 1 | Y | Select at most Y options
default | default | Z | W | true | Z | W | Select between Z and W options
default | default | Z | W | false | 1 | `num_ans` | Select at least 1 option
default | default | default | default | true | 1 | `num_ans` | Select between 1 and `num_ans` options
default | default | default | default | false | 1 | `num_ans` | Select at least 1 option

The min/max number of options to select are determined by the following rules, which are explained in the documentation:
1. If `min/max-select` is explicitly specified, we always use the specified value.
2. If `min/max-select` are not set, but `min/max-correct` are set **and** `detailed-help-text="true"`, then `min/max-correct` determines the min/max number of options that can be selected.
3. If rules 1 and 2 do not apply, then the min/max number of options that can be selected takes on the default values of 1 and `num_ans`.

If `detailed-help-text="true"`, the help text always displays the values of the min/max number of options to select. Otherwise, we only display the min/max values if they are non-default values.

exampleCourse/questions/demo/multiAnswerMC/question.html
```diff
@@ -0,0 +1,31 @@
...
+        This example question shows how we can use the Coverage (COV) grading scheme in a
+        multiple-choice question where only one option is correct, but multiple options can be
+        selected. In this question, COV grading gives a nonzero amount of credit to any valid
+        submission that selects the correct option. Notice also that at most 3 options can be
+        selected; we accomplish this by setting <code>max-select="3"</code> in the <code>pl-checkbox</code>
+        attributes.
+    </p>
+</pl-question-panel>
+
+<div class="card my-2">
```
mwest1066, line 12: We are trying to avoid bootstrap markup in example questions, because instructors tend to copy it unnecessarily. In this case I think it would be fine to just delete all the `card` markup.

exampleCourse/questions/demo/multiAnswerMC/info.json
```diff
@@ -0,0 +1,7 @@
+{
+    "uuid": "3f030556-e75b-11eb-ba80-0242ac130004",
+    "title": "Demo: Multi-Answer MC with Coverage Grading Scheme",
```
mwest1066, line 3:
```suggestion
    "title": "Demo: Multi-answer MC with coverage grading scheme",
```

exampleCourse/questions/demo/multiAnswerMC/question.html
```diff
@@ -0,0 +1,20 @@
...
+    <p>
+        This example question shows how we can use the Coverage (COV) grading scheme in a
+        multiple-choice question where only one option is correct, but multiple options can be
+        selected. In this question, COV grading gives a nonzero amount of credit to any valid
+        submission that selects the correct option. Notice also that at most 3 options can be
+        selected; we accomplish this by setting <code>max-select="3"</code> in the <code>pl-checkbox</code>
+        attributes.
+    </p>
+    <p>
+        Integer Review: Which of these mathematical statements are true?
```
mwest1066, line 11:
```suggestion
        Integer review: Which of these mathematical statements are true?
```

exampleCourse/questions/demo/multiAnswerMC/question.html
```diff
@@ -0,0 +1,20 @@
+<pl-question-panel>
+    <p>
+        This example question shows how we can use the Coverage (COV) grading scheme in a
+        multiple-choice question where only one option is correct, but multiple options can be
+        selected. In this question, COV grading gives a nonzero amount of credit to any valid
+        submission that selects the correct option. Notice also that at most 3 options can be
+        selected; we accomplish this by setting <code>max-select="3"</code> in the <code>pl-checkbox</code>
+        attributes.
+    </p>
```
mwest1066, line 9:
```suggestion
```

exampleCourse/questions/demo/multiAnswerMC/question.html
```diff
@@ -0,0 +1,20 @@
...
+        Integer Review: Which of these mathematical statements are true?
+    </p>
+</pl-question-panel>
+
+<pl-checkbox answers-name="ans" partial-credit="true" partial-credit-method="COV" max-select="3">
+    <pl-answer correct="true">If an integer is divisible by 4, it must be divisible by 2.</pl-answer>
+    <pl-answer>If an integer is divisible by 2, it must be divisible by 4.</pl-answer>
+    <pl-answer>An integer divided by an integer is always an integer.</pl-answer>
+    <pl-answer>The number 3.5 is an integer.</pl-answer>
+</pl-checkbox>
```
mwest1066, line 20:
```suggestion
</pl-checkbox>

    <p><i>
        This example question shows how we can use the Coverage (COV) grading scheme in a
        multiple-choice question where only one option is correct, but multiple options can be
        selected. In this question, COV grading gives a nonzero amount of credit to any valid
        submission that includes the correct option. Notice also that at most 3 options can be
        selected; we accomplish this by setting <code>max-select="3"</code> in the <code>pl-checkbox</code>
        attributes.
    </i></p>
```

mwest1066 (approved):

exampleCourse/questions/demo/multiAnswerMC/question.html
```diff
@@ -0,0 +1,21 @@
...
+
+    <p><i>
```
mwest1066, line 14:
```suggestion
<pl-question-panel>
    <p><i>
```

exampleCourse/questions/demo/multiAnswerMC/question.html
```diff
@@ -0,0 +1,21 @@
...
+        This example question shows how we can use the Coverage (COV) grading scheme in a
+        multiple-choice question where only one option is correct, but multiple options can be
+        selected. In this question, COV grading gives a nonzero amount of credit to any valid
+        submission that includes the correct option. Notice also that at most 3 options can be
+        selected; we accomplish this by setting <code>max-select="3"</code> in the <code>pl-checkbox</code>
+        attributes.
+    </i></p>
```
mwest1066, line 21:
```suggestion
    </i></p>
</pl-question-panel>
```

mwest1066:
> To help with review and testing, I created a table below which specifies the full relationship between `min/max-select`, `min/max-correct`, and `detailed-help-text`. These five attributes determine the minimum and maximum number of options that can be selected, as well as what help text is displayed.

Thanks for the super-detailed analysis of this! Amazing job. This all seems correct to me. Do you also think it's as good as possible?

echau01:
The table looks quite complicated at first. I had a few principles in mind when designing the table.

First, I wanted to preserve the already-existing interactions between `min/max-correct` and `detailed-help-text`. The `detailed-help-text` attribute has two functions: one is, of course, to show help text, but its other function is to restrict the number of options selected. As a result, if the help text says "Select between X and Y options", then PrairieLearn makes sure that the student actually selects between X and Y options. Before the `min/max-select` attributes were introduced, the numbers X and Y are determined by `min/max-correct`. I wanted to preserve that behaviour so as not to break any existing questions.

The other principle I followed was to make sure that `min/max-select`, if specified, completely determine the min/max number of options to select. These attributes override any restrictions imposed by `min/max-correct` in this area.

The above table is a result of these two principles. I think it's the right way to go.

mwest1066:
> The above table is a result of these two principles. I think it's the right way to go.

I agree, and I think your principles outlined above are sensible and clear.

Just to make sure I'm understanding correctly, the code and the table are consistent with each other now, right? That is, we don't need any further code modifications?

Or do we still need to do something to address #4430?

echau01:
#4430 is not addressed by this PR. I'll be working on it when I have the time.

mwest1066:
> #4430 is not addressed by this PR. I'll be working on it when I have the time.

Got it, thank for confirming!

PR #4455, opened 2021-07-04
Change pl-input to integer

sentz2:
Changes the default question that appears when selecting "Add Question".  Only changes pl-input from "pl-number-input"  to "pl-integer-input".


sentz2:
recheck

PR #4542, opened 2021-07-21
Manual grading col split schema fix

andrewstec:
After implementing an array with enumerated string types (ie. gradingMethod: ["External", "Manual"]) in a Question info.json configuration file, I found a bug during the JSON schema validation import. This PR fixes the bug with one possible approach.

The current implementation on PrairieLearn/master is that a gradingMethod can support one enumerated string type. (ie. One of "Manual", "Internal", or "External"). The implementation on the `manual-grading-col-split` branch intended to change `gradingMethod` from an enumerated string type to a UNION enumerated string type OR array of enumerated strings type. 

Note that we cannot do union types in strict mode of our schema validation. Therefore, implementing the string and array union possibility on the `gradingMethod` property is not possible in the `manual-grading-col-split` branch. Furthermore, it is not recommended from a modeling standpoint, as it is a singular word and a plural word is more suitable for arrays.

Keeping the `gradingMethod` property in the schema would be necessary as long as question imports continue to use the `gradingMethod` property. It would be possible to update all `gradingMethod` properties to the new `gradingMethods` convention by writing a script to update the repositories, but I am not too sure if this is the right approach. Alternatively, we can reject this PR and disable strict mode to allow union types.

Bug current example on `manual-grading-col-split`:

image

Resolved with this PR and changing implementation to gradingMethods: 

image

Actionable outcome: 

- Need agreement on approach with deprecation. 

Decided on July 26, 2021 meeting that `gradingMethod` will be kept and honoured for backwards compatibility. We will move forward with new `gradingMethods` implementation as well.

### DIFF
--- a/schemas/schemas/infoQuestion.json
+++ b/schemas/schemas/infoQuestion.json
@@ -77,12 +77,24 @@
             "type": "string"
         },
         "gradingMethod": {
-            "description": "The grading method used for this question.",
+            "description": "The grading method used for this question. (DEPRECATED - Use `gradingMethods` below)",
             "enum": [
                 "Internal",
                 "External",
                 "Manual"
             ]
+        },
+        "gradingMethods": {
+            "description": "The grading methods to enable during question grading. (Overrides `gradingMethod` if implemented`)",
+            "type": "array",
+            "items": {
+                "type": "string",
+                "enum": [
+                    "Internal",
+                    "External",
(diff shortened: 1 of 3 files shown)


jonatanschroeder:
I don't see a problem with this code, but I'd prefer to defer to Matt to review this, sorry.

Issue #965, opened 2017-12-26
add student assessment page


egunter:
This s just a suggestion for a potential future feature.

The tools for examining an assessment and being able to drill down to questions and then students and on various views of what a student did has been extremely helpful in giving students help with individual assignments.  If this is the column major view of things, it would also be very helpful to have the row major view:  to have a page of students, comparable to the page of assessments, and to be able to click on a student and see the student performance on every assessment, possibly be able to sort the assessment by score or submit time, maybe to see their score for each assessment dependent on how it is sorted, and to be able to click on an assessment and be taken to the same student assessment page as you get to through the assessment page.  This would facilitate seeing how individual students are performing and to see when they have trouble in one place how it might be linked back to other troubles they have had.  
---Elsa

trombonekenny:
Like the Gradebook tab at the top, i.e. https://prairielearn.engr.illinois.edu/pl/course_instance/3634/instructor/gradebook -- or something else?

egunter:
The grade book has more of the functionality I was requesting than I realized, but if that is the portal for this view, it is quite difficult to impossible to use the way I am desiring for a course of 250-300 students and dozens of assessments.  If I want to review what the students did on the 20th, 23rd and 27th assessments and the netid starts with r, it is almost impossible.  I have no idea how hard this would be, but making it so that, as you scroll down, the header row stays visible, and as you scroll right, the UID and/or Name column(s) stayed visible would largely accomplish the aims I was seeking, given the current functionality.  It would be a little easier if I could click of the UID or Name of the student and go to a page with all and just that student's assessments, but this is not required to achieve the review ability for which I was suggesting the extension.

nwalters512:
@mwest1066 and I were discussing changes similar to that, namely switching away from the classic "massive table" gradebook towards a per-student view. The current form is potentially not very useful for large numbers of students and assessments. Is there any part of your workflow that would be broken if the all-encompasing table was eliminated?

egunter:
@nwalters512 If your comment is directed to me, I would not loose much if the table went away.  The table is a little useful for seeing how the whole class is doing, but dumping to excel would accomplish much the same functionality.  I would suspect I'm no the only user or potential user of the grade book, however.

nwalters512:
Course instances now have a "Students" page where you can drill into individual students and do the following:

- View their enrollment history
- View their score on all assessments
- View the gradebook as the student
- View the course instance as the student

[...]

PR #4543, opened 2021-07-21
Sorting, search and UI improvements to Gradebook instructor page

jonatanschroeder:
Replaces #3177


mwest1066:
Thanks for the new branch! Makes testing it much easier for me 😄

mwest1066:
Minor display issue: when scrolling horizontally, the first column loses its right-hand border (on Safari):

no_border

pages/instructorGradebook/instructorGradebook.ejs
```diff
@@ -73,7 +106,7 @@
                 <td><%= row.role %></td>
                 <% row.scores.forEach(function(score, iScore) { %>
                 <td>
-                  <% if (score.score_perc == null) { %>
+                  <% if (score.score_perc === null) { %>
```
mwest1066, line 109: The `== null` will catch both `null` and `undefined`, which is why I try and always use it by default. In this case I think we are probably safe with `===` but as a matter of safety I'd tend to prefer `==` unless there is a specific reason we want the triple. (there's another case of this below)

pages/instructorGradebook/instructorGradebook.ejs
```diff
@@ -2,42 +2,57 @@
...
-          form.append('<button type="submit" class="btn btn-primary">Change</button>')
-          return form
-        },
-        title: 'Change total percentage score',
-        trigger: 'click',
-      })
-    })
+        $('#gradebook-table').bootstrapTable({
+            onResetView: function () {
+                $('button.edit-score').popover({
```
mwest1066, line 14: I'm completely fine to leave this as is, but this is one place where using ajax would be very nice to avoid the page reload. At the moment it's a bit annoying when you have carefully scrolled to a specific location the page and then the edit page reload scrolls back to the top 😄

mwest1066:
I'm not sure whether the "card toggle" is so useful? I think a better solution for this would be to add a per-student gradebook page, that shows list of all scores for a student and lets you edit them. Basically the page that Elsa is describing here: https://github.com/PrairieLearn/PrairieLearn/issues/965 (the "student performance on every assessment" page).

pages/instructorGradebook/instructorGradebook.ejs
jonatanschroeder: Ok, I'll change back in the next commit. For some reason I was thinking that `null` would be equal to 0, but indeed that's not the case.

jonatanschroeder:
> I'm not sure whether the "card toggle" is so useful? I think a better solution for this would be to add a per-student gradebook page, that shows list of all scores for a student and lets you edit them. Basically the page that Elsa is describing here: #965 (the "student performance on every assessment" page).

I'm looking into the option to enable "Detailed view", where you can expand the list of grades for a specific student. I'll remove the card view regardless, but the detailed view may be an interesting alternative.

jonatanschroeder:
Upon further review, using `detail-view` looks like an interesting way to show the grades in a more visually appealing way, but I'll leave that for a future PR.

For your other comments, all addressed, except for ajax update. I can try to work on it later, unless you want to separate that into a different PR.

pages/instructorGradebook/instructorGradebook.ejs
jonatanschroeder: Ajax grade update implemented. The interesting thing about this particular update is that, for group assessments, the update needs to change more than one row, and the other rows may not be visible.

In any case, in order to get this to work properly with `bootstrap-table` I had to move the formatting of the grade fields to client side (since an update also requires reformatting), but I believe the current implementation makes this a bit more streamlined than before. This means that tests had to be changed.

Assuming the updated tests now pass, this should be ready for review.

jonatanschroeder:
> no_border

From the screenshot above I noticed that the user name is now wrapping, I can change it to `text-nowrap`, what do you think? I hadn't noticed this locally because my local database doesn't have populated student names...

mwest1066:
> From the screenshot above I noticed that the user name is now wrapping, I can change it to `text-nowrap`, what do you think? I hadn't noticed this locally because my local database doesn't have populated student names...

I think it will basically always wrap, right? That's going to be annoying because it makes the rows extra tall. Should we set `no-wrap` and `width=20em` and `text-overflow=ellipses`? Or some other way to handle the overflow?

jonatanschroeder:
> I think it will basically always wrap, right? That's going to be annoying because it makes the rows extra tall. Should we set `no-wrap` and `width=20em` and `text-overflow=ellipses`? Or some other way to handle the overflow?

Bootstrap has `text-nowrap` as a class, probably the easiest way to handle this. Let me know if you have any concerns with that, it's now pushed.

mwest1066:
> Bootstrap has `text-nowrap` as a class, probably the easiest way to handle this. Let me know if you have any concerns with that, it's now pushed.

Excellent, that looks like a perfect solution. If a student has a really long name then it's going to look weird, but I think that's an edge case we can ignore for now.

pages/instructorGradebook/instructorGradebook.js
```diff
@@ -71,7 +71,15 @@ router.post('/', function(req, res, next) {
         ];
         sqldb.call('assessment_instances_update_score_perc', params, function(err, _result) {
             if (ERR(err, next)) return;
-            res.redirect(req.originalUrl);
+
+            params = {
+                assessment_instance_id: req.body.assessment_instance_id,
```
mwest1066, line 76: Data in `req.body` is untrusted, so it's possible that this `assessment_instance_id` is one that the instructor does not have permission to modify. We should actually be checking this above in `assessment_instances_update_score_perc()`. Our standard approach to this would be to pass `res.locals.assessment.id` (trusted, because it's in `res.locals`) to `assessment_instances_update_score_perc()` and inside `assessment_instances_update_score_perc()` we first check whether the provided `assessment_instance_id` belongs to the `assessment_id` (throwing a access exception if not). After that we can trust that the `assessment_instance_id` is valid.

pages/instructorGradebook/instructorGradebook.sql
```diff
@@ -73,3 +73,16 @@ SELECT user_id,uid,uin,user_name,role,
 FROM scores
 GROUP BY user_id,uid,uin,user_name,role
 ORDER BY role DESC, uid;
+
+-- BLOCK assessment_instance_score
+SELECT
+    COALESCE(ai.user_id, gu.user_id) AS user_id,
```
mwest1066, line 79: Very nice handling of group assessment updates!

pages/instructorGradebook/instructorGradebook.js
jonatanschroeder: The grade update was not modified, only the retrieval of the updated grade, so the vulnerability was there already. That said, I'm happy to update the sproc, though this would also require updating it on the other places where this sproc is called: the assessment instance log, and the upload by assessment. These other two places do not have the vulnerability (the log uses the instance from `res.locals`, score upload searches for the assessment instance ID from the assessment ID).

mwest1066: Sorry for not being clearer --- this was absolutely a pre-existing vulnerability that I only noticed now because of your nearby changes 😄

If the other call sites have easy access to the `assessment_id` then I'd suggest just passing it in and doing the check in all cases. If that's a pain, then an alternative is to make the `assessment_id` argument optional for `assessment_instances_update_score_perc()` and only do the check if we get the argument passed in.

mwest1066:
Also, given how cool an update this is, do you want to add a news item for it? I think that many people currently don't use the gradebook page because it is clunky, but with this PR then might start using it.

pages/instructorGradebook/instructorGradebook.js
jonatanschroeder: assessment_id is not in `res.locals`, since this is not an assessment-related page... Should I use the course instance instead?

jonatanschroeder:
> Also, given how cool an update this is, do you want to add a news item for it? I think that many people currently don't use the gradebook page because it is clunky, but with this PR then might start using it.

To be honest I was planning to use the same restructuring in the Students tab for an assessment as a follow up PR (one of our students already did some preliminary work on formatters for that), if you're ok with that. In that case I'd prefer to have a single news item for both.

mwest1066:
> To be honest I was planning to use the same restructuring in the Students tab for an assessment as a follow up PR (one of our students already did some preliminary work on formatters for that), if you're ok with that. In that case I'd prefer to have a single news item for both.

Yeah, I saw you closed the issue on that and I assumed you were thinking along those lines 😄

I'm fine with whatever you want!

pages/instructorGradebook/instructorGradebook.js
```diff
@@ -28,6 +28,20 @@ router.get('/', function(req, res, next) {
...
+                var scores = {
+                    user_id: row.user_id,
+                    uid: row.uid,
+                    uin: row.uin ?? '',
+                    user_name: row.user_name ?? '',
+                    role: row.role,
+                };
+                row.scores.forEach(function(score) {
+                    scores[`score_${score.assessment_id}`] = score.score_perc;
+                    scores[`score_${score.assessment_id}_ai`] = score.assessment_instance_id;
```
mwest1066, line 41: A few other places in the code have used `ai_id` as an abbreviation for `assessment_instance_id`. It's not an essential change, however.

pages/instructorGradebook/instructorGradebook.js
mwest1066: Yep, good call. Another option here is to do something like this: https://github.com/PrairieLearn/PrairieLearn/blob/f37c625f9c67f2037d57277326d2e68927b9d9d4/pages/instructorAssessmentInstances/instructorAssessmentInstances.js#L83

This is somewhat less efficient because it adds another DB round trip, but we already do multiple queries for auth so one more isn't too bad.

jonatanschroeder: Yeah, that would probably be simpler. Should I do this "checkBelongs" version for course instance vs assessment instance in the gradebook page itself or should I create a new library for it? Or are you aware of a library function that already exists for this test?

mwest1066: I don't think we have a `checkBelongs()` for an `assessment_instance_id` in a `course_instance_id`. Making a library function for this would be great. I don't have a great suggestion for an existing file to put it in, so maybe a new `lib/course.js`? (mimicking `lib/assessment.js` and `lib/question.js`)

jonatanschroeder: Would `lib/courseInstance.js` make more sense?

mwest1066: Yes, except that I was thinking about the analogy with `lib/assessment.js` and `lib/question.js`. We don't have `lib/assessmentInstance.js`, for example. But I don't have a strong opinion here, so go with whatever you think is most natural.

mwest1066:
The example table at https://examples.bootstrap-table.com uses a fixed-height table which has vertical scrolling within the table. It also has a button to expand to full page and pagination. What do you think about any of those setups?

The PR in its current form is already fantastic, so these are just minor questions. I'm fine with whatever layout you think is the most useable.

jonatanschroeder:
I added the fullscreen button. As for fixed-height, in early versions of this PR (back when it was just fixed-height) I tried with it and it was weird. I think a fixed height makes sense if the table is one element among several in the page, but this is not the case for this page.

mwest1066:
> I added the fullscreen button.

When playing around with this, it seems that clicking the fullscreen button has no advantage over just scrolling down the page? But fullscreen is confusing, because scrolling up won't bring back the navbar and you need to click the fullscreen button again. I think the fullscreen button is only of use when using a table is one element among several in the page, as you describe below. Sorry for mentioning it!

> As for fixed-height, in early versions of this PR (back when it was just fixed-height) I tried with it and it was weird. I think a fixed height makes sense if the table is one element among several in the page, but this is not the case for this page.

Makes sense, thanks!

jonatanschroeder:
I believe the only thing potentially missing here is the warning about changing the grade for other group members.

pages/instructorGradebook/instructorGradebook.ejs
```diff
@@ -2,42 +2,99 @@
 <html lang="en">
   <head>
     <%- include('../partials/head'); %>
+    <script src="<%= node_modules_asset_path('bootstrap-table/dist/bootstrap-table.min.js') %>"></script>
+    <script src="<%= node_modules_asset_path('bootstrap-table/dist/extensions/sticky-header/bootstrap-table-sticky-header.min.js') %>"></script>
+    <link href="<%= node_modules_asset_path('bootstrap-table/dist/bootstrap-table.min.css') %>" rel="stylesheet" />
+    <link href="<%= node_modules_asset_path('bootstrap-table/dist/extensions/sticky-header/bootstrap-table-sticky-header.min.css') %>" rel="stylesheet" />
   </head>
+  <script id="grade-data">
+    let gradeData = <%- JSON.stringify(user_scores_data) %>;
```
mwest1066, line 11: I'm worried about https://sophiebits.com/2012/08/03/preventing-xss-json.html

In the file editor page we do this on the server: https://github.com/PrairieLearn/PrairieLearn/blob/f37c625f9c67f2037d57277326d2e68927b9d9d4/pages/instructorFileEditor/instructorFileEditor.js#L124

and embed it into the HTML with: https://github.com/PrairieLearn/PrairieLearn/blob/f37c625f9c67f2037d57277326d2e68927b9d9d4/pages/instructorFileEditor/instructorFileEditor.ejs#L43

and then in client-JS we unpack it: https://github.com/PrairieLearn/PrairieLearn/blob/f37c625f9c67f2037d57277326d2e68927b9d9d4/pages/instructorFileEditor/instructorFileEditorClient.js#L82

jonatanschroeder: Resolved by escaping the input strings.

mwest1066:
> I believe the only thing potentially missing here is the warning about changing the grade for other group members.

Yep, that would be nice, because it could definitely be a bit unexpected.

pages/instructorGradebook/instructorGradebook.ejs
mwest1066: Perfect!

mwest1066:
Looks like we need some extra space before the edit button?

image

mwest1066:
Is it possible to get the row highlight to extend to the first column? No big deal if this is too messy or hard.

image

mwest1066:
No sort buttons for role?

image

mwest1066:
While you are cleaning up this page, how about adding a "?" icon next to the "Role" header with a popover that contains the role description text from the footer? This is more discoverable and we could then delete the entire footer (including the download link).

mwest1066:
I feel a bit dumb for asking this, but how does horizontal scrolling work if you don't have a trackpad? That is, I just use two fingers to drag the gradebook sheet sideways, which feels extremely fluid and natural. But if you are on an old-school Windows desktop, for example, then how to you scroll sideways? Is there a horizontal scrollbar somewhere? I hope you don't need to scroll all the way to the bottom of the page to find a horizontal scrollbar?

pages/instructorGradebook/instructorGradebook.ejs
```diff
@@ -2,42 +2,110 @@
...
+            return "—";
+        } else {
+            var container = $('<div>');
+            var link = $('<a/>').attr('href', `<%= urlPrefix %>/assessment_instance/${assessment_instance_id}`)
+                .text(Math.floor(score) + '%').appendTo(container);
+            if (<%= authz_data.has_instructor_edit %>) {
+                var button = $('<button/>').attr('type', 'button').addClass('btn btn-xs btn-secondary edit-score')
+                    .attr('tabindex', 0).attr('data-ai-id', assessment_instance_id).attr('data-score', score)
+                    .attr('data-other-users', JSON.stringify(other_users))
+                    .append($('<i class="fa fa-edit" aria-hidden="true"></i>'))
```
mwest1066, line 94: I wonder whether we should switch to https://icons.getbootstrap.com/icons/pencil-square/ at some point? Font Awesome is becoming increasingly annoying with their commercial push.

jonatanschroeder:
> Looks like we need some extra space before the edit button?

Done.

> No sort buttons for role?

Done.

> While you are cleaning up this page, how about adding a "?" icon next to the "Role" header with a popover that contains the role description text from the footer? This is more discoverable and we could then delete the entire footer (including the download link).

Will do.

> I feel a bit dumb for asking this, but how does horizontal scrolling work if you don't have a trackpad? That is, I just use two fingers to drag the gradebook sheet sideways, which feels extremely fluid and natural. But if you are on an old-school Windows desktop, for example, then how to you scroll sideways? Is there a horizontal scrollbar somewhere? I hope you don't need to scroll all the way to the bottom of the page to find a horizontal scrollbar?

I don't have an easy answer for this. This could be resolved with a fixed height table, though settling on a height could be a bit tricky. I can play around with this and try to figure out an option. That said, instructors have the option to remove the columns they don't need, in that case horizontal scrolling is not necessary.

jonatanschroeder:
> Is it possible to get the row highlight to extend to the first column? No big deal if this is too messy or hard.

I'll try to play around a bit, I tried before but didn't insist on it that much.

pages/instructorGradebook/instructorGradebook.ejs
jonatanschroeder: Done.

jonatanschroeder:
@mwest1066 All requests are now handled. I don't think horizontal scrolling is that easy to solve at this point in a reliable way, so I'd say the click and keyboard directionals should make do for now.

PR #4559, opened 2021-07-25
Remove NOT NULL constraint on submissions.regradable

mwest1066:
Adding a column with a `NOT NULL` constraint requires rewriting the entire table and blocks writes while it's happening. This is too expensive for a large table like `submissions`. While it's nice to have `NOT NULL` on this column, it shouldn't actually be required.

Weirdly, everything I [can](https://brandur.org/postgres-default) [find](https://www.depesz.com/2018/04/04/waiting-for-postgresql-11-fast-alter-table-add-column-with-a-non-null-default/) about PG 11 seems to suggest that `ADD COLUMN ... NOT NULL DEFAULT FALSE` should be just as fast as `ADD COLUMN ... DEFAULT FALSE` (without the NOT NULL). But in my experiments the latter version is super fast (maybe 1 second for 45-million rows) while the former version is super slow (at least 10 minutes, at which point it was auto-killed).

### DIFF
--- a/migrations/234_submissions__regradable__add.sql
+++ b/migrations/234_submissions__regradable__add.sql
@@ -1,4 +1,4 @@
-ALTER TABLE submissions ADD COLUMN IF NOT EXISTS regradable boolean NOT NULL DEFAULT FALSE;
+ALTER TABLE submissions ADD COLUMN IF NOT EXISTS regradable boolean DEFAULT FALSE;
 
 UPDATE submissions
 SET regradable = (graded_at IS NOT NULL) OR (grading_requested_at IS NOT NULL);
--- a/database/tables/submissions.pg
+++ b/database/tables/submissions.pg
@@ -17,7 +17,7 @@ columns
     params: jsonb
     partial_scores: jsonb
     raw_submitted_answer: jsonb
-    regradable: boolean not null default false
+    regradable: boolean default false
     score: double precision
     sid: text
     submitted_answer: jsonb


migrations/234_submissions__regradable__add.sql
```diff
@@ -1,4 +1,4 @@
-ALTER TABLE submissions ADD COLUMN IF NOT EXISTS regradable boolean NOT NULL DEFAULT FALSE;
+ALTER TABLE submissions ADD COLUMN IF NOT EXISTS regradable boolean DEFAULT FALSE;
 
 UPDATE submissions
 SET regradable = (graded_at IS NOT NULL) OR (grading_requested_at IS NOT NULL);
```
jonatanschroeder, line 4: @mwest1066 If the concern is updating the whole table, isn't this a problem too?

jonatanschroeder, line 4: @mwest1066 if the concern is updating the whole table, isn't it better to remove this update too?

jonatanschroeder (approved):

migrations/234_submissions__regradable__add.sql
mwest1066: The real problem is not the full table rewrite, but the fact that the table is locked as `ACCESS EXCLUSIVE` while `ALTER TABLE` is running, which blocks all other accesses (including reads and writes). This basically freezes PL while the table is being rewritten. In contrast, `UPDATE` only needs the `ROW EXCLUSIVE` [table lock](https://www.postgresql.org/docs/current/explicit-locking.html), which permits reads and writes to continue. At least this is my understanding 😄

echau01 (approved):
I always thought it's better to not have null entries in a database. If your experimentation shows otherwise, though, then let's go with your way.

mwest1066:
> I always thought it's better to not have null entries in a database. If your experimentation shows otherwise, though, then let's go with your way.

You are absolutely correct that it's better not to allow NULLs in the DB when they have no meaning. The problem is that we are migrating large existing DBs on live production applications, and it's very expensive (and blocking) to add these constraints. I'll leave off the `NOT NULL` constraint for now.

Issue #3344, opened 2020-10-15
Change assessment instance "open/closed" to "open/locked"?


mwest1066:
At the moment the "open/closed" status of assessment instances seems to be a bit confusing. Maybe it would be clearer if we renamed "closed" to "locked" and showed it as a lock icon on assessment instances? Instructors could click this lock icon to get a popover that says "This assessment instance is locked. The student can't submit answers to it anymore. Click [here] to unlock it. Note that the student will need [access] to submit answers. They currently [do/do not] have access."

Issue #3340, opened 2020-10-14
Live mode for assessment instances list


jonatanschroeder:
It would be great to have a "live" view of the assessments currently open (or recently closed), with the remaining time updating without the need for a refresh.

jonatanschroeder:
Such a screen could include some information like:

- which questions each student is working on;
- options to add time/close/etc;
- groups (interactions);
- ???

jonatanschroeder:
One possibility is to have this page interact with a web socket that uses the `LISTEN` directive in PostgreSQL, and relevant events would trigger `NOTIFY`: https://www.postgresql.org/docs/current/sql-notify.html

jonatanschroeder:
The assessment instances page already provides some of this functionality, as of #4560 it should provide additional provisions. Still pending is a break-down of status per question, and (possibly) an automatic refresh based on web sockets triggered by actual events.

PR #4560, opened 2021-07-25
UI improvements to Assessment instances page

### LINKED ISSUE #1130: instructor assessment instance list CSS popup problem
If you have an assessment that only has 1-2 assessment instances, the bottom of the Action popup (to close, re-open, etc.) gets hidden under the bottom of the panel footer. Scrolling doesn't work properly and not all the buttons are visible.

screen shot 2018-04-25 at 7 31 28 am

### LINKED ISSUE #1202: User with TA permissions can see "Regrade all" button on instructorAssessment
For users with TA-level permissions, the "Regrade All" button appears above assessment instances, but clicking results in a (somewhat opaque) "Not Found" error. Not sure if the right solution is to hide the button or to create a more meaningful "You don't have this permission" modal/error page.

Compare the individual assessment instance "Action" drop-down, which is empty for users with TA permissions.

I think this might be a particular case of #758

### LINKED ISSUE #1316: instructorAssessmentInstances: show message if not has_instructor_edit
https://github.com/PrairieLearn/PrairieLearn/blob/f4192f2e0dddf2efc9e4fbc3b808b6e6ac6a81c5/pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs#L221

Update the UI so that if a TA opens the drop down to edit student assessment instances, they get a message about needing Instructor access to edit instead of a blank (broken-looking) dropdown.

### LINKED ISSUE #2149: Remaining and Duration columns semi-broken in Students tab
If you view a given `pl/course_instance/nnnnn/instructor/assessment/nnnnnnn/instances` page (the "Students" tab for a given assessment instance), the Remaining and Duration columns don't always seem to display correct info:

The Duration column appears to show "0s" as the duration for a session that closed due to time limit running out. Example available here (may not be persistent if we need to correct the grade manually later and that changes something): https://prairielearn.engr.illinois.edu/pl/course_instance/64437/instructor/assessment/1413729/instances Possibly connected to the recent PR about LTI https://github.com/PrairieLearn/PrairieLearn/pull/2144 This was a situation where the student interacted with the first variant for about 15 seconds, and then may have had some issue that led to logging in again through the direct PrairieLearn URL that should not be available, so there may be more than one thing going on. In this case, it's also strange that the session timed out after 100 minutes when it should have timed out after 90 minutes. (Maybe this warrants a separate issue. I'm testing how long it takes for another session to time out, now.)

The Remaining column seems to show "0 min" while a live exam is in session. (Tested this several times myself over the past few months, but can't link an example that will persist.)

### LINKED ISSUE #3339: Implement Pagination for Assessment Instances
I have a practice test for students that students have generated over 1300 instances for (though this is not terribly surprising--with 450 students that's an average of 3 attempts per student), and it lags my browser out when trying to sort them. Would be great to be able to limit to, say, 50 or 100 at at time.

### LINKED ISSUE #3343: Show student access status on "Students" tab
At the moment the "Students" tab shows Open/Closed status for assessment instances, but not whether each student currently has access. It would be helpful to show this directly. Should we show different icons for different levels of access? Like "can see the whole exam", "can see total score only", "can't see anything"?

### LINKED ISSUE #3658: Create another column in the Students Page (tab)
Current page for an individual assessment:

image

I propose we create another column (maybe the third column) that has the header "Details" and the following text for each row entry:

<short_assessment_name>  + (<assessment_instance_number>) + for + <user_uid>

For the example above, the entry would be `HW6(1) for mfsilva@illinois.edu`.

This text would have a link that points to the same page when the button "Details" is clicked. The columns "Details" could then be removed.

For a group assessment, the row entry would be 

<short_assessment_name>  + (<assessment_instance_number>) + for + <name_entry>

image

For the example above,  `L7(1) for Mari`.

____

In the future, we would also add a link to the column `UID`, which would direct to a page that contains informations about that student (assessments, assessment summary and statistics, etc). This page does not exist yet (another PR!), but once we have this page,  we should add a link to it in all occurrences of UID (the above example, the gradebook, ...)

jonatanschroeder:
Changes similar to those done in #4543 in assessment instances page. This replaces #3992 as well, since a refresh button is provided too, and an auto-refresh button can also be provided (https://bootstrap-table.com/docs/extensions/auto-refresh/).

Resolves #1130.

Still TODO:

* [x] More extensive tests;
* [x] Ajax actions like change time limits, close, grade and delete assessments, followed by refresh;
* [x] Auto-refresh;
* [x] Update CI tests that rely on the data in this page (e.g., `tests/testInstructorAssessment.js`);
* [x] Change details link to a first column with AI identification (resolves #3658);
* [x] Add column for total time limit (initially hidden);
* [x] Pagination (resolves #3339);
* [x] Add helper box (similar to Role) to explain meaning of Duration column (resolves #2149);
* [x] Review and clarify action button absence when user does not have edit permission, e.g., TA (resolves #1202, resolves #1316);
* [x] News item.


mwest1066:
Open issues related to this page are:
* #3344
* #3343
* #3340
* #3339
* #2149
* #1316
* #1202
* #1130
* #2263

I'm definitely not saying that all of the above _should_ be implemented or fixed. Some of them might be too hard or just bad ideas 😄

jonatanschroeder:
> * [Change assessment instance "open/closed" to "open/locked"? #3344](https://github.com/PrairieLearn/PrairieLearn/issues/3344)

This probably requires a more extensive discussion instead of just a UI update.

> * [Live mode for assessment instances list #3340](https://github.com/PrairieLearn/PrairieLearn/issues/3340)

Although this PR helps in this direction, I'm still hoping to have a page with more extensive info, e.g., broken down by question and with more options for automatic refresh than just a re-request every 60 seconds.

> * [Remaining and Duration columns semi-broken in Students tab #2149](https://github.com/PrairieLearn/PrairieLearn/issues/2149)

I believe Remaining is correct now, and Duration only needs an explanation of what it means. Added to TODO.

> * [instructor assessment instance list CSS popup problem #1130](https://github.com/PrairieLearn/PrairieLearn/issues/1130)

This is fixed already.

> * [Instructor assessment instance view is missing tab-bar #2263](https://github.com/PrairieLearn/PrairieLearn/issues/2263)

This is for the single instance page, not for the list of instances. I'll focus this PR on the list.

All other items you listed have been added to the TODO list.

mwest1066:
> All other items you listed have been added to the TODO list.

Awesome! I agree with all the other points you made. Do you want to close #1130?

mwest1066:
Oooops! Sorry! I was thinking about closing #1130 and accidentally closed your PR!!

jonatanschroeder:
> Do you want to close #1130?

I mean this PR fixes the issue with #1130, so I linked it, but it's already done in this PR.

mwest1066:
> I mean this PR fixes the issue with #1130, so I linked it, but it's already done in this PR.

Got it! That's great.

jonatanschroeder:
@mwest1066 I made a stub of a news item, if you have time I'd prefer if you fill the fluff in it, is that ok? I can provide more info if necessary.

mwest1066:
> @mwest1066 I made a stub of a news item, if you have time I'd prefer if you fill the fluff in it, is that ok? I can provide more info if necessary.

Will do!

mwest1066:
The "Action" button on the Students tab looks weird (should it be a button?):

image

And the dropdown menu under Action seems to be missing spacing after the icons:

image

mwest1066:
With the "showing 50 rows of 200" in the footer, I always worry that this will be off the bottom of the screen and so will be missed. Should we shift this to the top left of the table (where the "Download..." link is now)? Maybe move the download link to use the bootstrap-table download icon in the icon bar?

I'm not convinced that these are good ideas, but rather I'm throwing them out for consideration.

mwest1066:
I was a bit too hasty in my description of "Duration", sorry! The 1-hour inactivity exception actually only applies to Homeworks, and Exams measure the full time from start to final submission. Also see #4609

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -243,168 +151,219 @@
...
+                <h4 class="modal-title" id="role-help-title">Time Remaining</h4>
+                <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                  <span aria-hidden="true">&times;</span>
+                </button>
+              </div>
+              <div class="modal-body">
+                <p>For open assessments with a time limit, this column
+                  will indicate the number of minutes (rounded down)
+                  the student has left to complete the assessment. If
+                  the value is <strong>0 min</strong>, the student has
```
mwest1066, line 351: Could we display `< 1 min` in this case? Only if it isn't too messy to do so.

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -243,168 +151,219 @@
...
+        <table id="usersTable"
+               data-unique-id="assessment_instance_id"
+               data-classes="table table-sm table-hover table-bordered"
+               data-url="<%= urlPrefix %>/assessment/<%= assessment.id %>/instances/raw_data.json"
+               data-search="true"
+               data-show-columns="true"
+               data-show-refresh="true"
+               data-auto-refresh="true"
+               data-auto-refresh-status="false"
+               data-auto-refresh-interval="30"
```
mwest1066, line 174: Is there any way to indicate how long until the next auto-refresh? Is this what the `data-auto-refresh-silent` option does?

mwest1066:
Should we also make the first column sticky? I'm thinking both to help folks on small screens right now, and also future-proofing as more columns are added.

admin_queries/generate_assessment_instances.json
```diff
@@ -0,0 +1,14 @@
+{
+    "description": "Simulates all students enrolled in a course starting an assessment (groupwork not supported)",
```
mwest1066, line 2: These are very nice admin queries!

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
jonatanschroeder: Yes, it's possible, will do.

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
jonatanschroeder: I don't know if it's possible to show that, I'll check. As for the silent option, it's to refresh it without redrawing the entire table, i.e., only update the rows/cells that changed.

jonatanschroeder: I can't find a easy and reliable way to show the time for the next auto-refresh through the BT extension, other than to bypass the extension and do the whole auto-refresh process separately, or to extend the BootstrapTable class functionality (which may not be reliable if the npm package is later updated and the changes become inconsistent). I don't think it's worth the effort.

jonatanschroeder:
> With the "showing 50 rows of 200" in the footer, I always worry that this will be off the bottom of the screen and so will be missed. Should we shift this to the top left of the table (where the Gradebook "Download..." link is now)? Maybe move the Gradebook download link to use the bootstrap-table download icon in the icon bar? The same thing goes for the Students page, but there we have no download link so the upper-left corner is free.
> 
> I'm not convinced that these are good ideas, but rather I'm throwing them out for consideration.

image

Thoughts?

mwest1066:
> Thoughts?

Love it!

jonatanschroeder:
All your comments are handled, other than the time left for auto-refresh.

Makefile
```diff
@@ -1,7 +1,11 @@
 PATH := node_modules/.bin/:$(PATH)
 
+ifdef NODEMON
+start: start-nodemon
+else
 start:
-	node server.js
+	@node server.js
+endif
```
mwest1066, line 8: Maybe this leaked in by accident from #4614?

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
mwest1066: Agreed. It's fine as is.

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -243,168 +169,232 @@
...
+                  <li>12:00 - student submits answer.</li>
+                </ul>
+                <p>In the above example, the "duration" would be 2
+                  hours: one hour from 08:00 to 09:00, and another
+                  hour from 11:00 to 12:00. The two-hour gap between
+                  09:00 to 11:00 is not counted as part of the
+                  duration.</p>
+                <p><strong>For Exam assessments</strong>, a student is
+                  considered to be actively working between the start
+                  of the assessment and the less submission,
```
mwest1066, line 354:
```suggestion
                  of the assessment and the last submission,
```

mwest1066 (approved):

Makefile
jonatanschroeder: The perils of `git -a`...

jonatanschroeder:
@mwest1066 want to review the news item yet?

mwest1066:
> @mwest1066 want to review the news item yet?

Haha, oops, I completely forgot! Thanks for the reminder!!

Makefile
mwest1066: I know 😄

mwest1066:
> > @mwest1066 want to review the news item yet?
> 
> Haha, oops, I completely forgot! Thanks for the reminder!!

Done now. Let me know if my news item changes look ok to you.

PR #4580, opened 2021-07-28
Add enableNetworking option for workspaces

### LINKED ISSUE #3193: Restrict networking in workspaces
(breaking this out of #2504)

Add an option like `allowNetworkAccess` (default `false`) for workspaces. Possible implementation strategies:
* https://stackoverflow.com/a/47008421
* https://www.eisfunke.com/article/docker-wireguard-systemd.html

mwest1066:
Resolves #3193


docs/workspaces/index.md
```diff
@@ -64,6 +64,7 @@ The question's `info.json` should set the `singleVariant` and `workspaceOptions`
     * `args` (optional, default none): command line arguments to pass to the Docker image
     * `syncIgnore` (optional, default none): list of files or directories that will be excluded from sync
     * `rewriteUrl` (optional, default true): if true, the URL will be rewritten such that the workspace container will see all requests as originating from /
+    * `enableNetworking` (optional, default false): whether the workspace should be allowed to connect to the public internet. This is disabled by default to make secure, isolated execution the default behavior. This restriction is not enforced when running PrairieLearn in local development mode.
```
nwalters512, line 67: What's the likelihood that this will break existing workspaces? I think this'll be worth it regardless, but might warrant advanced warning to folks before deploying.

workspace_host/interface.js
```diff
@@ -967,10 +968,22 @@ function _createContainer(workspace, callback) {
     } else {
         args = args.split(' ');
     }
+
+    let networkMode = 'bridge';
+    if (!workspace.settings.workspace_enable_networking) {
+        if (config.workspaceSupportNoInternet) {
+            networkMode = 'no-internet';
+        } else {
+            logger.verbose('Workspace requested unsupported feature enableNetworking:false');
```
nwalters512, line 977: Thinking out loud - should we log a warning at sync time if someone tries to sync a no-network workspace to a PL instance that doesn't support it? We can suppress it in dev mode to avoid giving false warnings.

workspace_host/interface.js
```diff
@@ -967,10 +968,22 @@ function _createContainer(workspace, callback) {
     } else {
         args = args.split(' ');
     }
+
+    let networkMode = 'bridge';
+    if (!workspace.settings.workspace_enable_networking) {
+        if (config.workspaceSupportNoInternet) {
+            networkMode = 'no-internet';
```
nwalters512, line 975: Nit: it may be advantageous to make the name of the network configurable? Not necessary based on the current setup though.

nwalters512 (approved):

docs/workspaces/index.md
mwest1066: Good point, I should have documented the deployment plan. My intention is to deploy in a non-enforcing mode, so that people have a chance to add `"enableNetworking":true` to their `info.json` files. At this point I'll send email to anyone using workspaces and warn them about the change, with a deadline (2 weeks?) for when we'll start enforcing it.

An alternative would be to wait until after final exams are over at UIUC and then just start enforcing it. A news item about the change might be a good idea in this case?

workspace_host/interface.js
mwest1066: Unfortunately it's a bit tricky because it's the workspace hosts that do the enforcing, but the PL webservers that do the syncing. Let's discuss this in person to brainstorm ideas.

nwalters512: Ahh - as in, the config option applies to workspace hosts, not the PL server?

mwest1066: Exactly

workspace_host/interface.js
mwest1066: Skipping for now.

PR #4581, opened 2021-07-28
Add initial role select table

renzol2:
Adding table before start of group assessment for users to select POGIL roles


nwalters512:
Confirmed offline with @caleqw2 that this PR is stale and can be closed.

PR #3818, opened 2021-02-11
Replace npm scripts with make targets

jonatanschroeder:
Given some changes in behaviour in NPM 7, this PR replaces all `npm run` scripts with `make` targets.

PR #4590, opened 2021-07-30
Remove all use of npm scripts and npx

### LINKED ISSUE #4586: matplotlib error while starting up tests
I'm running tests from the docker instance on my machine (KDE Neon, Ubuntu 20.04 base)

```
# start container 
docker run -it --rm -p 3000:3000 -v ~/git/PrairieLearn/:/PrairieLearn --name pl prairielearn/prairielearn

# enter the container (from a different terminal)
docker exec -it pl bash

# run the tests
cd PrairieLearn/
npm test
```

I will get the following error: 
```
bash-4.2# npm test
> PrairieLearn@3.2.0 test
> make test
nyc --reporter=lcov mocha tests/index.js
  database
    ✔ should match the database described in /database (713ms)
    ✔ should not contain "ON DELETE CASCADE" foreign keys from soft-delete to hard-delete tables (621ms)
  Administrator pages
    1. view administrator overview page
      ✔ should load successfully (225ms)
      ✔ should parse (61ms)
      ✔ should contain QA 101
error: Unexpected STDERR data: Matplotlib created a temporary config/cache directory at /tmp/matplotlib-inyn61_c because the default path (/root/.config/matplotlib) is not a writable directory; it is highly recommended to set the MPLCONFIGDIR environment variable to a writable directory, in particular to speed up the import of Matplotlib and to better support multiprocessing.
 {"childIsNull":false,"callbackIsNull":true,"timeoutIDIsNull":true,"outputStdout":"","outputStderr":"","outputBoth":"","outputData":"","stack":"Error\n    at PythonCaller._errorData (/PrairieLearn/lib/code-caller.js:87:4587)\n    at PythonCaller._logError (/PrairieLearn/lib/code-caller.js:87:5113)\n    at PythonCaller._handleStderrData (/PrairieLearn/lib/code-caller.js:86:1534)\n    at Socket.emit (events.js:375:28)\n    at Socket.emit (domain.js:470:12)\n    at addChunk (internal/streams/readable.js:290:12)\n    at readableAddChunk (internal/streams/readable.js:261:11)\n    at Socket.Readable.push (internal/streams/readable.js:204:10)\n    at Pipe.onStreamRead (internal/stream_base_commons.js:188:23)","lastCallData":null}
  AdministratorQuery page
    ✔ visit queries page (81ms)
Deprecated as of 10.7.0. highlight(lang, code, ...args) has been deprecated.
Deprecated as of 10.7.0. Please use highlight(code, options) instead.
```

@echuber2 suggested this hack to make it happy: 
```
# hack to fix the matplotlib error
mkdir -p /root/.config/matplotlib
chmod a+rx /root /root/.config
chmod -R a+rwx /root/.config/matplotlib
```

it prevents the errors from happening during start up, but any tests which use matplotlib will still fail. 

Other things tried: running docker as root, and running the docker container with `-v /var/run/docker.sock:/var/run/docker.sock` don't fix the problem.

mwest1066:
npm v7 has a horrible misfeature:

> When npm is run as root, scripts are always run with the effective uid and gid of the working directory owner.

(see https://blog.npmjs.org/post/626173315965468672/npm-v7-series-beta-release-and-semver-major)

This causes lots of problems when mounting a local copy of the PrairieLearn source into the container at `/PrairieLearn` because the working directory will be owned by a different user to the files inside the container (e.g., `/root/.gitconfig`), causing hard-to-debug errors due to permissions being wrong. For example, #4586 is caused by this.

@jonatanschroeder fixed most of this by switching from `npm` scripts to `make` in #3818, but we missed a few cases. This PR eradicates all remaining references to `npm` scripts and `npx`.

Fixes #4586

### DIFF
--- a/docs/installingLocal.md
+++ b/docs/installingLocal.md
@@ -11,31 +11,64 @@ This page describes the procedure to run PrairieLearn within Docker, but using a
 git clone https://github.com/PrairieLearn/PrairieLearn.git
 ```
 
-* Install the Node.js packages.  This will use the version of `npm` that is pre-installed in the Docker image, so you don't need your own copy installed.
+* Run PrairieLearn with:
 
 ```sh
-docker run --rm -w /PrairieLearn -v /path/to/PrairieLearn:/PrairieLearn prairielearn/prairielearn /usr/local/bin/npm ci
+docker run -it --rm -p 3000:3000 -v /path/to/PrairieLearn:/PrairieLearn prairielearn/prairielearn /bin/bash
+
+# following commands are inside the container:
+cd /PrairieLearn
+npm ci                   # install packages, repeat this after switching branches or pulling new code
+docker/start_support.sh  # start the DB, cache, etc.
+node server              # run PrairieLearn itself
+
+# now you can Ctrl-C and run "node server" again to restart PrairieLearn (after code edits, for example)
+# or Ctrl-C to stop PL and Ctrl-D to exit the container
 ```
 
-The path `/path/to/PrairieLearn` should be replaced with the *absolute* path to the PrairieLearn source on your computer.  If you're in the root of the source directory already, you can substitute `%cd%` (on Windows cmd), `${PWD}` (on Windows PowerShell), or `$PWD` (Linux, MacOS, and WSL) for `/path/to/PrairieLearn`.
+The path `/path/to/PrairieLearn` above should be replaced with the *absolute* path to the PrairieLearn source on your computer.  If you're in the root of the source directory already, you can substitute `%cd%` (on Windows cmd), `${PWD}` (on Windows PowerShell), or `$PWD` (Linux, MacOS, and WSL) for `/path/to/PrairieLearn`.
 
-By default, PrairieLearn will load `exampleCourse`, `testCourse`, and any courses mounted at `/course` and `/course[2-9]` in the Docker container.  To override this behavior, you can create a custom [`config.json` file](configJson.md).
 
-* Run it with:
+## Running the test suite
+
+The linters and tests for the Python and JavaScript code can be run with the following commands in a shell instance:
 
 ```sh
-docker run -it --rm -p 3000:3000 -v /path/to/PrairieLearn:/PrairieLearn prairielearn/prairielearn
+docker run -it --rm -p 3000:3000 -v /path/to/PrairieLearn:/PrairieLearn prairielearn/prairielearn /bin/bash
+
+# following commands are inside the container:
+cd /PrairieLearn
+./docker/lint_js.sh
+./docker/lint_python.sh
+./docker/test_js.sh
+./docker/test_python.sh
 ```
 
-By default, PrairieLearn does not monitor for server-side changes, so it will **not** automatically restart the node server when you change the node source. To enable automatic live-reloading of server-side changes, use the `-e` flag to set the `NODEMON=true` environment variable:
+
+## Updating or building the Docker image
+
+The commands above all run PrairieLearn using local source inside the `prairielearn/prairielearn` image. This image has Python packages and other supporting files already installed. This should be periodically updated with:
 
 ```sh
-docker run -it --rm -p 3000:3000 -e NODEMON=true -v /path/to/PrairieLearn:/PrairieLearn prairielearn/prairielearn
+docker pull prairielearn/prairielearn
 ```
 
-### Running a specific branch
+You can also build a local copy of this image and use it to make sure you have a version that corresponds exactly to your local source:
 
-By default, the above command will run PrairieLearn from the branch that is currently checked out in the directory `/path/to/PrairieLearn`. So, to run a different branch, just use commands like `git checkout BRANCH_NAME` or equivalent.
+```
+cd /path/to/PrairieLearn
+docker build -t local_prairielearn .
+
+# now use it with:
+docker run -it --rm -p 3000:3000 -v /path/to/PrairieLearn:/PrairieLearn local_prairielearn /bin/bash
+```
+
+Above we are using the `local_prairielearn` image in place of the pre-built `prairielearn/prairielearn` image.
+
+
+## Running a specific branch in a pre-built container
+
+By default, the commands above will run PrairieLearn from the branch that is currently checked out in the directory `/path/to/PrairieLearn`. So, to run a different branch, just use commands like `git checkout BRANCH_NAME` or equivalent.
 
 It is also possible to run a branch other than `master` without cloning or checking out the code for that branch, as long as the branch has been pushed to GitHub.  If you would like to run a different branch (to test it, for example), the branch name can be appended to the end of the image name as such:
 
@@ -50,7 +83,8 @@ docker pull prairielearn/prairielearn:BRANCH_NAME
 docker run -it --rm -p 3000:3000 prairielearn/prairielearn:BRANCH_NAME
 ```
 
-### Running commands in Docker
+
+## Running commands in Docker
 
 If needed, you can run the container with a different command:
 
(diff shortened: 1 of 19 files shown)


jonatanschroeder (approved):
A few more places that could be considered too:
https://github.com/PrairieLearn/PrairieLearn/blob/50c8c4017cc9ccbfe10c4febec5e1fb8cd777a2a/docker/test_grader_host.sh#L5

https://github.com/PrairieLearn/PrairieLearn/blob/50c8c4017cc9ccbfe10c4febec5e1fb8cd777a2a/docker/test_prairielib.sh#L5

https://github.com/PrairieLearn/PrairieLearn/blob/50c8c4017cc9ccbfe10c4febec5e1fb8cd777a2a/docs/installingNative.md#L80

mwest1066:
> A few more places that could be considered too:

Thanks! I changed the docs (and did some rearranging of the local-install docs). I did not change `grader_host` and `prairielib` because they don't currently have Makefiles and I didn't want to mess with them. This might be a mistake and I'm sure I'll regret it when it causes trouble in six months time 😄. @nwalters512 has been wanting to switch from `npm` to `yarn`, and I'm starting to agree!

mwest1066:
> I did not change `grader_host` and `prairielib` because they don't currently have Makefiles, they work with npm v7, and I didn't want to mess with them.

I actually went ahead and changed them as well, for future safety.

mwest1066:
I also deleted all the `"scripts"` inside `package.json` files so that we can't accidentally run scripts like `npm test` anymore. The only valid use of `npm` should now be `npm ci` to actually do package installation.

docs/installingLocal.md
```diff
@@ -11,31 +11,64 @@ This page describes the procedure to run PrairieLearn within Docker, but using a
...
-docker run --rm -w /PrairieLearn -v /path/to/PrairieLearn:/PrairieLearn prairielearn/prairielearn /usr/local/bin/npm ci
+docker run -it --rm -p 3000:3000 -v /path/to/PrairieLearn:/PrairieLearn prairielearn/prairielearn /bin/bash
+
+# following commands are inside the container:
+cd /PrairieLearn
+npm ci                   # install packages, repeat this after switching branches or pulling new code
+docker/start_support.sh  # start the DB, cache, etc.
+node server              # run PrairieLearn itself
+
+# now you can Ctrl-C and run "node server" again to restart PrairieLearn (after code edits, for example)
```
jonatanschroeder, line 25:
```suggestion
make start               # run PrairieLearn itself

# now you can Ctrl-C and run "make start" again to restart PrairieLearn (after code edits, for example)
```

docs/installingLocal.md
```diff
@@ -11,31 +11,64 @@ This page describes the procedure to run PrairieLearn within Docker, but using a
...
 ```sh
-docker run -it --rm -p 3000:3000 -v /path/to/PrairieLearn:/PrairieLearn prairielearn/prairielearn
+docker run -it --rm -p 3000:3000 -v /path/to/PrairieLearn:/PrairieLearn prairielearn/prairielearn /bin/bash
+
+# following commands are inside the container:
+cd /PrairieLearn
+./docker/lint_js.sh
+./docker/lint_python.sh
+./docker/test_js.sh
+./docker/test_python.sh
```
jonatanschroeder, line 44: Linters could also be invoked with:
```
make lint-js
make lint-python
```
I believe the `make` target is easier to memorize. It may be a good idea to create `make` targets that include the test plus initialization as well, though I don't have a strong opinion on it.

docs/installingLocal.md
```diff
@@ -64,34 +98,27 @@ This can be used to, e.g., run scripts distributed with PrairieLearn by opening
...
 
-and when any modifications are made, you can close the server with `<ctrl-C>` and re-run the init script.
 
-#### Tests from shell
+## Auto-restarting the node server
 
-The linters and tests for the Python and JavaScript code can be run with the following commands in a shell instance:
+The description at the start of this page suggests manually stopping and restarting PrairieLearn after you have edited any JavaScript files. You can alternatively use the `nodemon` package to watch for changes to code and auto-restart PrairieLearn. Two different ways to do this are:
+
+* Run PrairieLearn as described at the start of this page, but replace `node server` with `make start-nodemon`.
```
jonatanschroeder, line 112:
```suggestion
* Run PrairieLearn as described at the start of this page, but replace `make start` with `make start-nodemon`.
```

jonatanschroeder (approved):
I added suggestions regarding replacing scripts with make targets where they exist, if you'd prefer to keep the current versions I'm happy with those too.

docs/installingLocal.md
mwest1066: I changed to use `make` targets and I added targets for `test-js` and `test-python`. I didn't add targets that run the initialization because that causes problems with re-running those targets. I want to keep it so that you can run `make test` or `make lint` repeatedly, as part of a debug-edit-check cycle.

nwalters512:
Drive-by comment: I would vastly prefer to just revert to npm v6 - v7 has done nothing but caused problems without giving us any benefits. I don't like the idiosyncracies from the rest of the Node ecosystem that this introduces, it feels mostly like debt that'll need to be undone in the future.

mwest1066:
> Drive-by comment: I would vastly prefer to just revert to npm v6 - v7 has done nothing but caused problems without giving us any benefits. I don't like the idiosyncracies from the rest of the Node ecosystem that this introduces, it feels mostly like debt that'll need to be undone in the future.

My current feeling is that it was a weird decision by npm to include a command-runner into a build tool, and somehow other node build tools (yarn, pnpm) have followed along. No other package manager ecosystem does this, as far as I know, and I don't understand the point. We already have perfectly good ways of running commands (shell scripts or make).

What am I missing here?

I'm not necessarily against moving to yarn or pnpm or whatever as an alternative to npm, except that I feel like we are caught in node package manager churn. At least if we run our scripts using stable tools (like make) then it's easy to switch package managers without having to also change our command runner.

I'm not super excited about downgrading to npm 6, because that just feels like a stop-gap and we are going to have to deal with this properly sooner or later?

I guess all of this is a "convince me why it isn't just easier to run commands via make, which I know and understand and has been stable for 1000 years" 😄

nwalters512:
I certainly don't have anything against make, and I don't have a particularly compelling argument besides "it matches what the rest of the ecosystem does". If I run `yarn lint` or `npm run lint` in *any* Node project, I know what'll happen. Violating conventions introduces additional friction that makes it harder for me to work (I can no longer rely on muscle memory) and harder for new contributors to find their way in the project.

All that said, this is a minor issue in the grand scheme of things and one that we can certainly document our way around and won't fundamentally impart our ability to build a good product 🙂

mwest1066:
> I certainly don't have anything against make, and I don't have a particularly compelling argument besides "it matches what the rest of the ecosystem does". If I run `yarn lint` or `npm run lint` in _any_ Node project, I know what'll happen. Violating conventions introduces additional friction that makes it harder for me to work (I can no longer rely on muscle memory) and harder for new contributors to find their way in the project.
> 
> All that said, this is a minor issue in the grand scheme of things and one that we can certainly document our way around and won't fundamentally impart our ability to build a good product 🙂

Totally agreed -- I'm just feeling annoyed after wasting a bunch more time tracking down weird behavior due to npm 7 and this is sending me into unix graybeard mode where I claim that everything was better when we edited files with `ed` 😄

It's a very fair point you make that we shouldn't deviate from conventions unless there's a good reason (looking at you, npm v7). I'm on board for a yarn switch.

At this point I'm inclined to not switch to yarn and to merge this PR to at least deal with npm breakage. I acknowledge that it means typing `make lint` rather than `npm lint`, but I feel like this is livable until we get around to switching to yarn.

nwalters512:
Yep - happy to forge ahead with this to make sure things work for the time being! Did not mean to block this from landing.

mwest1066:
> Yep - happy to forge ahead with this to make sure things work for the time being! Did not mean to block this from landing.

No worries, it's very good to have you pull me back when I'm charging off in a weird direction 😄

PR #4621, opened 2021-08-03
Add automated test for Proof Blocks

SethPoulsen:


elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -24,6 +24,15 @@
 INDENT_OFFSET = 0
 TAB_SIZE_PX = 50
 
+DAG_FIRST_WRONG_FEEDBACK = [
```
mwest1066, line 27: Can we make this a dict with descriptive keys? Maybe `feedback-none` and `feedback-first-wrong`?

SethPoulsen: sure!

mwest1066 (approved):

PR #4622, opened 2021-08-03
Docker: Add multi-arch build to support ARM64

ctessum:
(a.k.a. Apple Silicon). See here for more info:
https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

### DIFF
--- a/docs/installingLocal.md
+++ b/docs/installingLocal.md
@@ -33,6 +33,8 @@ By default, PrairieLearn does not monitor for server-side changes, so it will **
 docker run -it --rm -p 3000:3000 -e NODEMON=true -v /path/to/PrairieLearn:/PrairieLearn prairielearn/prairielearn
 ```
 
+**NOTE**: On MacOS with "Apple Silicon" (ARM64) hardware, the use of R is not currently supported.
+
 ### Running a specific branch
 
 By default, the above command will run PrairieLearn from the branch that is currently checked out in the directory `/path/to/PrairieLearn`. So, to run a different branch, just use commands like `git checkout BRANCH_NAME` or equivalent.
--- a/docs/installingLocal.md
+++ b/docs/installingLocal.md
@@ -33,8 +33,6 @@ By default, PrairieLearn does not monitor for server-side changes, so it will **
 docker run -it --rm -p 3000:3000 -e NODEMON=true -v /path/to/PrairieLearn:/PrairieLearn prairielearn/prairielearn
 ```
 
-**NOTE**: On MacOS with "Apple Silicon" (ARM64) hardware, the use of R is not currently supported.
-
 ### Running a specific branch
 
 By default, the above command will run PrairieLearn from the branch that is currently checked out in the directory `/path/to/PrairieLearn`. So, to run a different branch, just use commands like `git checkout BRANCH_NAME` or equivalent.
--- a/tools/build-image-if-needed.sh
+++ b/tools/build-image-if-needed.sh
@@ -13,5 +13,5 @@ if git diff --exit-code remotes/origin/master...HEAD -- ${BUILD_DIRECTORY}; then
   echo "${BUILD_DIRECTORY} files not modified; no rebuild required"
 else
   echo "${BUILD_DIRECTORY} files modified; ${TAG_NAME} requires a rebuild"
-  docker buildx build --push --platform linux/arm64/v8,linux/amd64 ${BUILD_DIRECTORY} -t ${TAG_NAME}
+  docker build ${BUILD_DIRECTORY} -t ${TAG_NAME}
 fi
--- a/tools/build-image-if-needed.sh
+++ b/tools/build-image-if-needed.sh
@@ -13,5 +13,5 @@ if git diff --exit-code remotes/origin/master...HEAD -- ${BUILD_DIRECTORY}; then
   echo "${BUILD_DIRECTORY} files not modified; no rebuild required"
 else
   echo "${BUILD_DIRECTORY} files modified; ${TAG_NAME} requires a rebuild"
-  docker build ${BUILD_DIRECTORY} -t ${TAG_NAME}
+  docker buildx build --push --platform linux/arm64/v8,linux/amd64 ${BUILD_DIRECTORY} -t ${TAG_NAME}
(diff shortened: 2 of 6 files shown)


ctessum:
The reason for this change is that running PL locally on Docker under qemu emulation mode is almost completely unusable, it constantly freezes or crashes. So this is PR attempts to add native ARM64 builds.

This is a work in progress; I think the main remaining problem is that R is not supported on ARM64 unless you build the whole thing from source yourself.

I've temporarily removed R from the arm64 build, and currently the ARM64 build works fine, but the amd64 build of plbase fails with this error:

```
/plbase-install.sh: line 69:  7394 Killed                  conda install r-essentials
```

There's not much information here to tell what went wrong, and notably this error doesn't actually stop the build process, it keeps going on until it's not able to install `rpy2`, and then it errors out.

With the changes to the Dockerfiles included here I've been able to get it to build locally, so I don't have any huge drive to troubleshoot this the rest of the way, but I thought I would put it here in case someone else wants to pick up where I'm leaving off.

tools/release-image.sh
```diff
@@ -42,12 +42,4 @@ if [ -z "${t}" ]; then
...
-docker build -t prairielearn/$1 .
-
-if [ $? -ne 0 ]; then
-  echo "ERR: building image failed. skipping upload." >& 2
-  exit 4
-fi
-
-docker tag prairielearn/$1 prairielearn/$1:${t}
-docker push prairielearn/$1:${t}
+docker buildx build --push --platform linux/arm64/v8,linux/amd64 -t prairielearn/$1:${t} .
```
ctessum, line 45: `buildx` is not able to load multi-arch images into the local docker instance (see [here](https://docs.docker.com/engine/reference/commandline/buildx_build/#output)), so this builds and pushes in one step. I don't know if that will break anything.

tools/build-image-if-needed.sh
```diff
@@ -13,5 +13,5 @@ if git diff --exit-code remotes/origin/master...HEAD -- ${BUILD_DIRECTORY}; then
   echo "${BUILD_DIRECTORY} files not modified; no rebuild required"
 else
   echo "${BUILD_DIRECTORY} files modified; ${TAG_NAME} requires a rebuild"
-  docker build ${BUILD_DIRECTORY} -t ${TAG_NAME}
+  docker buildx build --push --platform linux/arm64/v8,linux/amd64 ${BUILD_DIRECTORY} -t ${TAG_NAME}
```
ctessum, line 16: `buildx` is not able to load multi-arch images into the local docker instance (see [here](https://docs.docker.com/engine/reference/commandline/buildx_build/#output)), so this builds and pushes in one step. I suspect that this might not be the desired behavior here, but I'm not sure what is. If this is just for a local development build, then maybe this file doesn't need to change at all.

mwest1066:
> I've temporarily removed R from the arm64 build, and currently the ARM64 build works fine, but the amd64 build of plbase fails with this error:
> 
> ```
> /plbase-install.sh: line 69:  7394 Killed                  conda install r-essentials
> ```

If this is a `Killed` error from `conda` then I’ve seen it before when `conda` ran out of memory during the “solving” phase of package installation. On my laptop I had to increase the amount of memory in the Docker Desktop preferences. Maybe you could try bumping the container memory and running a manual `conda install r-essentials` in the `arm64` version of `plbase`?

tools/build-image-if-needed.sh
mwest1066: We use this script in CI to rebuild various images if need: https://github.com/PrairieLearn/PrairieLearn/blob/c0c0e638dda4e3f0cc8a921bdc9fd6ef6628279b/.github/workflows/main.yml#L15

I think we can leave this file unchanged, because we only really need to build for local usage here.

tools/release-image.sh
mwest1066: This should be fine.

docs/installingLocal.md
```diff
@@ -33,6 +33,8 @@ By default, PrairieLearn does not monitor for server-side changes, so it will **
 docker run -it --rm -p 3000:3000 -e NODEMON=true -v /path/to/PrairieLearn:/PrairieLearn prairielearn/prairielearn
 ```
 
+**NOTE**: On MacOS with "Apple Silicon" (ARM64) hardware, the use of R is not currently supported.
```
mwest1066, line 36: Maybe shift this to just after: https://github.com/PrairieLearn/PrairieLearn/blob/c0c0e638dda4e3f0cc8a921bdc9fd6ef6628279b/docs/installing.md#L42 ?

ctessum:
I think I've responded to the comments above, let me know what you think.

Building both architectures now works for me locally after increasing my docker memory limit.

If I try to install R on ARM64 I get this error:

``` bash
ctessum@christophersmbp plbase % docker run -it --rm ctessum/plbase
bash-4.2# conda install r-essentials
Collecting package metadata (current_repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Solving environment: failed with repodata from current_repodata.json, will retry with next repodata source.
Collecting package metadata (repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Solving environment: \ 
Found conflicts! Looking for incompatible packages.
This can take several minutes.  Press CTRL-C to abort.
failed                                                                                                                                     

UnsatisfiableError: The following specifications were found to be incompatible with each other:

Output in format: Requested package -> Available versions
```

I think that's because conda-forge doesn't have an ARM64 build for R, as shown here: https://anaconda.org/conda-forge/r-essentials. It does say there is a "no-arch" build, but I don't see how that would be possible because my understanding is that a lot of R is written in C, which needs platform-specific compilation.

mwest1066 (approved):

mwest1066:
> I think that's because conda-forge doesn't have an ARM64 build for R, as shown here: https://anaconda.org/conda-forge/r-essentials. It does say there is a "no-arch" build, but I don't see how that would be possible because my understanding is that a lot of R is written in C, which needs platform-specific compilation.

Makes sense, I agree that dropping R on ARM64 is the right way to go. Thanks!

PR #4625, opened 2021-08-03
Bug fix: changing call type for return of multiple rows

andrewstec:
Bug (two parts): 

1) When an 'External' enum is enabled on the `gradingMethods` property, internal grading and external grading occurs because internal grading is enabled on all questions. This returns two rows instead of one row, which breaks the SQL call expecting one row returned. 
2) After the SQL call is changed to allow for two rows returned, we ONLY see 'Internal' grading actions on the front-end. 

After 1 is fixed on the back-end, 2 becomes apparent.

**We only see this Internal action:**
image
As we see, it is configured for external grading too though.
image

**We should also see this External action (taken from a working External grading question):** 

image

### DIFF
--- a/lib/question.js
+++ b/lib/question.js
@@ -574,7 +574,7 @@ module.exports = {
                     data.params,
                     data.true_answer,
                 ];
-                sqldb.callWithClientOneRow(client, 'grading_jobs_insert', params, (err, result) => {
+                sqldb.callWithClient(client, 'grading_jobs_insert', params, (err, result) => {
                     if (ERR(err, callback)) return;
 
                     /* If the submission was marked invalid during grading the grading job will


andrewstec:
I merged another fix into this PR so I could continue to work through getting the external grader submission to appear on the Question page with the internal grader submission when the `gradingMethods` property contains ['External', 'Internal'] (or ['External'] because internal is configured to run as true by default).

After the merge, we can now see the external grader running on the back-end, but we still only see the internal grading submission on the front-end on the question view. 

Back-end external grader running:
image

Only Internal grading appearing on front-end:
image

Hence, we know bug part 1 is resolved but part 2 still needs to be resolved.

PR #4636, opened 2021-08-04
Add news item for workspace default internet block

mwest1066:
image


jonatanschroeder (commented):
I think this item (or the related documentation) should:

1. Warn about the implications of giving access to the Internet when not necessary;
2. Describe how this option would impact CBTF-held exams, and in which situations (e.g., Homeworks) it would be more suitable;
3. Include a practical example of a workspace that could benefit from networking (e.g., a programming environment that involves working with sockets).

mwest1066:
> I think this item (or the related documentation) should:
> 
> 1. Warn about the implications of giving access to the Internet when not necessary;
> 2. Describe how this option would impact CBTF-held exams, and in which situations (e.g., Homeworks) it would be more suitable;
> 3. Include a practical example of a workspace that could benefit from networking (e.g., a programming environment that involves working with sockets).

Done.

jonatanschroeder (approved):

PR #4637, opened 2021-08-05
Add filter in gradebook and students page to list only Students

jonatanschroeder:
I'm open to suggestions about tweaks in UI (e.g., different button order, different title).

### DIFF
--- a/pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
+++ b/pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
@@ -89,6 +89,11 @@ $(function() {
             modal.modal('hide');
         });
     });
+
+    $('[data-toggle="modal"]').click(function (e) {
+        e.stopPropagation(); // Keep click from changing sort
+        $($(e.currentTarget).data('target')).modal('show');
+    })
 });
 
 function refreshTable() {
--- a/pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
+++ b/pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
@@ -7,6 +7,7 @@ $(function() {
             studentsOnly: {
                 text: 'Students Only',
                 icon: 'fa-user-graduate',
+                attributes: { title: 'List only enrolled students' },
                 event: () => {
                     const table = $('#usersTable');
                     const filterOn = !table.data('filter-student-only');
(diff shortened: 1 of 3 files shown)


pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
```diff
@@ -3,6 +3,20 @@ var bsTable;
 $(function() {
 
     bsTable = $('#usersTable').bootstrapTable({
+        buttons: {
+            studentsOnly: {
```
mwest1066, line 7: Can we add hover text to say what the button does, like the current Refresh and Column buttons have? (the Download button is also missing hover text)

pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
```diff
@@ -3,6 +3,20 @@ var bsTable;
...
+                text: 'Students Only',
+                icon: 'fa-user-graduate',
+                event: () => {
+                    const table = $('#usersTable');
+                    const filterOn = !table.data('filter-student-only');
+                    table.data('filter-student-only', filterOn);
+
+                    $('.columns button[name=studentsOnly]').toggleClass('active', filterOn);
+                    table.bootstrapTable('filterBy', filterOn ? { role: 'Student' } : {});
```
mwest1066, line 16: It looks like this button defaults to "off" (that is, by default we show all users). Is that correct? I like this default.

pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
jonatanschroeder: Done.

pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
jonatanschroeder: Yes, it defaults to show all users.

jonatanschroeder: To clarify, do you mean "I like the default to stay the way it is" or "I'd like this filter to become on by default"?

mwest1066: I like the default to stay the way it is 😄

mwest1066 (approved):

PR #4644, opened 2021-08-07
Fix checkbox unit testing bug

### LINKED ISSUE #4642: pl-checkbox unit tests sometimes fail
The `element/checkbox` question tests sometimes fail. This can be seen both in CI and also if you click "test 100 times" for the `element/checkbox` question.

The failures occur for test type `incorrect` and they report:
```
Error: "format_errors keys" mismatch: expected "[]" but got "["group_ans_v9"]"
```

The problem seems to be that we are trying to generate an incorrect answer, but we actually generate an invalid one. It's invalid due to `group_ans_v9` which uses `min-select` and `max-select`: https://github.com/PrairieLearn/PrairieLearn/blob/a2eb14e678899b5708d31acb42a5b477db3d3f11/exampleCourse/questions/element/checkbox/question.html#L247

I think the problem is here: https://github.com/PrairieLearn/PrairieLearn/blob/a2eb14e678899b5708d31acb42a5b477db3d3f11/elements/pl-checkbox/pl-checkbox.py#L461

In the code above we are only checking min/max-select if we _don't_ have `detailed-help-text`. But in `group_ans_v9` we still want min/max-select because we set them explicitly.

@echau01 I'm not sure what the best way is to fix this, so could you take a look at it? Thanks!

### Excerpted failing test data

The expected submission is:

```
expected_submission:
{
    ...
    "format_errors": {},
    ...
    "raw_submitted_answer": {
        ...
        "group_ans_v9": [
            "g"
        ],
    },
}
```

The test submission is:

```
test_submission:
{
    ...
    "format_errors": {
        "group_ans_v9": "You must select between <b>3</b> and <b>8</b> options."
    },
    ...
    "raw_submitted_answer": {
        ...
        "group_ans_v9": [
            "g"
        ],
    },
}
```

echau01:
Fixes #4642

### DIFF
--- a/elements/pl-checkbox/pl-checkbox.py
+++ b/elements/pl-checkbox/pl-checkbox.py
@@ -441,6 +441,9 @@ def test(element_html, data):
     number_answers = len(data['params'][name])
     all_keys = [pl.index2key(i) for i in range(number_answers)]
 
+    min_options_to_select = _get_min_options_to_select(element, MIN_SELECT_DEFAULT)
+    max_options_to_select = _get_max_options_to_select(element, number_answers)
+
     result = data['test_type']
 
     if result == 'correct':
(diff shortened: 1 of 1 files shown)


mwest1066:
Awesome! Thanks for the fast and clean fix!

mwest1066 (approved):

PR #4645, opened 2021-08-07
Popover details for sync errors and warnings in assessments list

### LINKED ISSUE #2925: Sync error indicator does not indicate what it is
If we have a sync error in an `infoAssessment.json` file (for example) then we render a red cross next to the assessment on the instructor assessments page. But there is no indication about what that red cross means, and clicking on it doesn't reveal any further information or take you anywhere.

jonatanschroeder:
Fixes #2925 (as needed for #3020).


mwest1066:
You are too fast! 😄

Do we have the same missing popover here? https://github.com/PrairieLearn/PrairieLearn/blob/9a7fba37c3ff2c7cf88e87de9465d878b19ea6d9/pages/instructorQuestions/instructorQuestions.ejs#L59

And should we add the sync error/warning icons (and popovers) to `pages/instructorAssessmentQuestions`?

mwest1066:
I feel a bit dumb, but I can't get the popovers to display. I see them in the source, and I see the popover init script at the top of the page, but nothing is happening when I click or hover over the error icons.

jonatanschroeder:
> I feel a bit dumb, but I can't get the popovers to display. I see them in the source, and I see the popover init script at the top of the page, but nothing is happening when I click or hover over the error icons.

Does it work with a different browser? And does something show in the browser's console window?

jonatanschroeder:
> You are too fast! 😄

I'm in the zone 😉 
 
> Do we have the same missing popover here?
> 
> https://github.com/PrairieLearn/PrairieLearn/blob/9a7fba37c3ff2c7cf88e87de9465d878b19ea6d9/pages/instructorQuestions/instructorQuestions.ejs#L59

Done.

> And should we add the sync error/warning icons (and popovers) to `pages/instructorAssessmentQuestions`?

Do you mean for question sync errors? Instance question sync errors are reported at the assessment level.

jonatanschroeder:
> > I feel a bit dumb, but I can't get the popovers to display. I see them in the source, and I see the popover init script at the top of the page, but nothing is happening when I click or hover over the error icons.
> 
> Does it work with a different browser? And does something show in the browser's console window?

I changed the trigger to hover instead of focus, is this better?

> > And should we add the sync error/warning icons (and popovers) to `pages/instructorAssessmentQuestions`?
> 
> Do you mean for question sync errors? Instance question sync errors are reported at the assessment level.

I added the question sync errors to the buttons. I added it before the QID to avoid breaking the numbering system (and to make it clear that the issue is in the question, not the assessment), but I'm happy to discuss alternative locations.

mwest1066:
> Does it work with a different browser? And does something show in the browser's console window?

Good call! It was working in Chrome but not in Safari. No errors in the JS console.

> I changed the trigger to hover instead of focus, is this better?

It now works in both Chrome and Safari, and I think it's better anyway! Nice!

mwest1066:
> Do you mean for question sync errors? Instance question sync errors are reported at the assessment level.

Yes, sorry, I should have been clearer that I was talking about question sync errors. But you are so deep in the zone that you understood anyway 😉

> I added the question sync errors to the buttons. I added it before the QID to avoid breaking the numbering system (and to make it clear that the issue is in the question, not the assessment), but I'm happy to discuss alternative locations.

That's awesome. I think it's very clear that something is wrong with the question.

mwest1066:
This is a huge improvement already, but now I'm getting greedy and asking for more 😄

Two UI issues that I'm seeing:

1. When the popover displays,  it's hard for the user to figure out exactly which file they need to navigate to in order to fix the issue. In contrast, the page-level sync error displays have a nice "Edit info.json" button:

image

Could we add an "Edit info.json" button to the popover? But then "hover" wouldn't work? And can we change the text to be "Edit info.json to fix this error" or something like that? Maybe in primary blue rather than the current white?

2. The page-level sync errors (like in the image above) are good for making it really clear that something is wrong, and giving the button to take the user to where it needs to be fixed. However, these page-level errors don't show up on the Files tab, and (even worse) they don't show up when you are actually editing the broken `info.json`. So the user sees that there is an error, gets to the place to fix it, but then they've forgotten what the error was. Can you see any way to fix this?

jonatanschroeder:
> Could we add an "Edit info.json" button to the popover? (I know this will break "hover") 

In all cases the button is right next to a link that leads them to the a page that has the same error popping up as a main item. I believe this should be a good enough flow at this point, though I understand that this will be one extra click compared to what you suggested.

> And can we change the text to be "Edit info.json to fix this error" or something like that? Maybe in primary blue rather than the current white?

Sure, that can be done, in the main question/assessment item. I can do that for course instance and course sync errors as well.

> 2. The page-level sync errors (like in the image above) are good for making it really clear that something is wrong, and giving the button to take the user to where it needs to be fixed. However, these page-level errors don't show up on the Files tab, and (even worse) they don't show up when the user is actually editing the broken `info.json`. So the user sees that there is an error, gets to the place to fix it, but then they've forgotten what the error was. Can you see any way to fix this?

I'll check if something can be done easily.

jonatanschroeder:
Like this?

image

jonatanschroeder:
> > 1. The page-level sync errors (like in the image above) are good for making it really clear that something is wrong, and giving the button to take the user to where it needs to be fixed. However, these page-level errors don't show up on the Files tab, and (even worse) they don't show up when the user is actually editing the broken `info.json`. So the user sees that there is an error, gets to the place to fix it, but then they've forgotten what the error was. Can you see any way to fix this?
> 
> I'll check if something can be done easily.

Actually the file edit page already shows these, but this has a bug whereas if the QID has a `/` (i.e., questions use directory structures), this doesn't work. I'll fix it for QID, but just to check, can AIDs or CIIDs also have directory structures?

jonatanschroeder:
Answering my own question: AID can have directories. Can CIIDs have directories?

jonatanschroeder:
My implementation is not particularly fancy, it just assumes there won't be a CIID where one of the components is called `assessments` (which I believe is a reasonable assumption for 99.999% of courses).

mwest1066:
> Like this?

Perfect!

mwest1066:
> Answering my own question: AID can have directories. Can CIIDs have directories?

Yep, just to make sure we are on the same page, everything (QIDs, AIDs, CIIDs) can have hierarchical directory structures. This is because they all call `loadInfoForDirectory()` with `recursive = true`: https://github.com/PrairieLearn/PrairieLearn/blob/296e3555619127f7325230480734f6eb8c540a9e/sync/course-db.js#L1048

jonatanschroeder:
Don't merge yet, I still have one commit to push, and I'm AFK

mwest1066:
> Don't merge yet, I still have one commit to push, and I'm AFK

Sounds good, I won't merge until you give the "all clear". I'm still testing it anyway 😄

jonatanschroeder:
Pushed.

mwest1066:
> Pushed.

Awesome, I'll take another look in the morning before merge, and then I'll get this into tomorrow's deploy.

mwest1066:
This is looking really great! I agree with your comment above that we don't need a "Click to edit" button on the popover because it's really easy to click on the offending object to get to such a button.

Two more places where we are currently missing sync error/warning display:

1. On the "file view" page for the relevant `infoXXX.json` file. After your fix above we now display the errors on the "file edit" page, but they are missing from the equivalent "file view".

2. On the "Files" page for the relevant assessment, course instance, etc. That is, the view that shows a directory listing containing the broken `infoXXX.json` file. I think we would ideally show both the banner error (with the "Click here to fix..." button) as well as the red X icon next to the broken JSON file (with a popover).

jonatanschroeder:
> Two more places where we are currently missing sync error/warning display:
> 
> 1. On the "file view" page for the relevant `infoXXX.json` file. After your fix above we now display the errors on the "file edit" page, but they are missing from the equivalent "file view".

Done.

> 2. On the "Files" page for the relevant assessment, course instance, etc. That is, the view that shows a directory listing containing the broken `infoXXX.json` file. I think we would ideally show both the banner error (with the "Click here to fix..." button) as well as the red X icon next to the broken JSON file (with a popover).

I've added the popover, but not the general banner. Do you think it's really necessary?

mwest1066:
> > 1. On the "file view" page for the relevant `infoXXX.json` file...
>
> Done.

Awesome!

> > 1. On the "Files" page ...
>
> I've added the popover, but not the general banner. Do you think it's really necessary?

I tried it out and I agree that it's not essential.

However, it does feel a bit weird to me that the user can click back and forwards between the different tabs for an assessment and the only tab where the banner _isn't_ shown is the Files tab. When clicking on different tabs it's a bit visually jarring to have the banner "disappear" on the Files tab, and it also feels a bit logically weird to not have the banner there.

Are you thinking that it's just more visual noise for the user? Or is it hard to implement?

jonatanschroeder:
> > I've added the popover, but not the general banner. Do you think it's really necessary?
> 
> I tried it out and I agree that it's not essential.
> 
> However, it does feel a bit weird to me that the user can click back and forwards between the different tabs for an assessment and the only tab where the banner _isn't_ shown is the Files tab. When clicking on different tabs it's a bit visually jarring to have the banner "disappear" on the Files tab, and it also feels a bit logically weird to not have the banner there.
> 
> Are you thinking that it's just more visual noise for the user? Or is it hard to implement?

I'm fine with adding the banner, but I think it's weird to have both,  especially since the popover will typically show the same info as the banner. I can replace the popover with a banner instead, do you think that makes more sense? The code in that case may even be relatively simpler depending on how I implement it.

jonatanschroeder:
And should I show the banner for all files of a question/CI/assessment/course or just the info*.json? Also, for subdirectories (e.g., clientFilesCourse) should I add it too?

mwest1066:
> I'm fine with adding the banner, but I think it's weird to have both, especially since the popover will typically show the same info as the banner. I can replace the popover with a banner instead, do you think that makes more sense? The code in that case may even be relatively simpler depending on how I implement it.

Given a choice, I'd rather have the banner than the popover, because the banner has the clear "Click here to fix it" button. But I personally think that it's better to have both. My general feeling here is that this is a really serious problem and we need to provide all possible indicators about what's wrong, in the hope that the user follows at least one of them 😄

(but I'd be ok with just the banner, if you think it's just too ugly to have the popover as well)

> And should I show the banner for all files of a question/CI/assessment/course or just the info*.json? Also, for subdirectories (e.g., clientFilesCourse) should I add it too?

My feeling is that we should add it in as many places as possible 😄 (all files and all subdirectories sounds fine to me)

jonatanschroeder:
Here's what's implemented now:
image

I think it looks weird to have both:
image

mwest1066 (approved):

PR #4646, opened 2021-08-08
 Regrading a single question for a single student

echau01:
(I accidentally closed #4362, oops.)

The regrading branch is in this repository now. Copying over the PR description from #4362:

This is a continuation of the work done in pull request #3452.

A summary of changes made since the end of #3452:
- regrading multiple submissions on an exam or homework question
- disallow regrading of questions with external grading (for now)
- extensive unit tests for regrading exam and homework questions
- in `lib/question.js`, merged `_regradeVariantWithClient()` function into `_gradeVariantWithClient()` function, thus reducing code duplication
- implemented regrading for submissions incorrectly marked as invalid (as long as the submission is "regradable" in the sense of #4406)
- fixed the "error: undefined {"data":{"rollback":"success"}}" logger message described in one of the TODO items of #3452. It shouldn't appear anymore.
- rewording of messages
- rewrote documentation for regrading
- added logs for regrading in instructor assessment instance page


echau01:
Copying over unresolved issues raised in #4362:

### What happens to the database if we keep the original score instead of the regraded score?

When a question is regraded for a student, all regrading occurs inside a single transaction. During regrading, several tables are modified, such as `grading_jobs`, `submissions`, `variants`, `instance_questions`, and `assessment_instances`. If we want to keep the original score after regrading (because the regraded score is lower), should we "rollback" all of the changes made during the transaction? The current behaviour is to rollback **some** changes, but not all changes -- for example, the score for the instance question is reset, but the scores for each submission are not reset. 

One reason to rollback all changes made to the database is to avoid inconsistencies in the database. For example, suppose a student makes one submission for a question instance. When graded, that submission earns 100%, so the instance question's score is 100%. Later, the `grade` function is modified. Suppose that when the question is regraded, the submission earns 50%, but the instructor chooses to discard the regraded score if it is lower than the original score, so the question's score remains at 100%. Now, we have an inconsistency: the question score is 100%, but the submission score is 50%. This inconsistency shows up in question statistics.

However, there is a disadvantage of rolling back everything: a submission score may be different from the actual mark that the submission should earn, which may confuse/mislead students. If a submission earned 100% at first, but should really earn 50% after regrading, we don't want the student to think that their submission is completely correct. 

I am personally in favour of **not** rolling back everything because of the disadvantage explained above. As for the possible inconsistencies: manually editing a student's grade makes the database inconsistent as well, and that's already an established feature of PL, so it seems that the inconsistencies don't actually matter.

### An unsolved problem: the difficulty of regrading a question

Currently, to regrade a question, instructors must write a new grading function (or modify the existing one) in the question's `server.py` or `server.js` file. Many people would find it easier if they could just modify the values of the correct answers in the question configuration. However, this will not work because the correct answers for a question variant are only computed when the variant is generated; (re)grading a submission does not compute the correct answers anew.

Regrading is difficult because in order to write a grading function, one needs some knowledge of the internal workings of the specific PL elements used in the question. As an example, consider a multiple choice question with the following `question.html` file:

```html
<pl-question-panel>
    <p>The correct answer should be "Option 2".</p>
</pl-question-panel>

<pl-multiple-choice answers-name="ans" weight="1">
    <pl-answer correct="true">Option 1</pl-answer>
    <pl-answer>Option 2</pl-answer>
</pl-multiple-choice>
```

We have mistakenly set the "Option 1" choice to be the correct answer when in fact "Option 2" should be correct. To regrade this question, we cannot just set the "Option 2" choice to be the correct answer in the `question.html` file. Instead, we must write a `grade` function in `server.py` like so:

```python
def grade(data):
    ans_key = data['submitted_answers']['ans']
    selected_option = None
    
    for choice in data['params']['ans']:
        if choice['key'] == ans_key:
            selected_option = choice
            break
    
    if selected_option['html'] == 'Option 2':
        data['partial_scores']['ans'] = {'score': 1, 'weight': 1}
        data['score'] = 1
    else:
        data['partial_scores']['ans'] = {'score': 0, 'weight': 1}
        data['score'] = 0
```

To be able to write the `grade` function above, we need to know the following characteristics specific to the multiple choice element:
- `data['submitted_answers']['ans']` contains the alphabetic key (`'a'`, `'b'`, `'c'`, etc.) of the multiple choice option that was submitted
- `data['params']['ans']` is an array containing the choices in the multiple choice question. Additionally, each element in the array is a `dict` of the form `{"key": ..., "html": ...}`.

If the question used a different PL element (e.g. `pl-order-blocks`), the structure of `data['submitted_answers']['ans']` and the elements of `data['params']['ans']` will be different. The grading function would then have to be tailored toward the internal workings of that PL element. Ultimately, the current regrading system demands too much knowledge of internal implementations of elements.

Other concerns:
- The grading function can easily become very complex if the question has many parts. Normally, PL elements grade themselves using the "correct answers" computed at question variant generation. However, when regrading, the elements no longer grade themselves correctly, so instructors have to essentially re-implement grading for the elements used in the question. The more complex the grade function, the more likely that bugs will be introduced.
- If an element's internal implementation changes, then the grading function might break.

echau01:
Another potential issue I thought of: there is no way to undo a regrade once it's done. Suppose an instructor wants to keep the highest score when regrading, but accidentally chooses to allow the score to decrease. They can't undo the regrade (except via manual score override).

mwest1066:
> Another potential issue I thought of: there is no way to undo a regrade once it's done. Suppose an instructor wants to keep the highest score when regrading, but accidentally chooses to allow the score to decrease. If the score decreases, the instructor can't undo the regrade (except via manual score override). Regrading a second time with "keep highest score" will have no effect.

Yep, true. I think we should just make sure to put a warning on the "regrade" button, like we do for other not-undoable actions: https://github.com/PrairieLearn/PrairieLearn/blob/68ca439f9be6ac340257cf16ec4b4c52074e7c2d/pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs#L82

firasm:
Thanks for getting this going @echau01 ! I was just going through the discussion in this PR and wanted to make a quick comment about this.

> Instead, we must write a grade function in server.py like so....

In general, I understand and appreciate the need to regrade a question by updating the `grade()` function in `server.py`, but I worry about the implications of this `grade()` function left in the question code, especially when re-running the course in the future.

In the example you've got, it makes complete sense, and even if that code was left in there for a future assessment, it would be harmless and innocuous (just redundant) because only `Option 2` would be marked correct. 

However, I'm imagining a more complex regrading scenario where say, in the initial version of the question, a default `rtol = 0.01` was not sufficient. So to regrade the question this year, I'd write a new `grade()` function that accepts any answer +/- 7%. If I use this question in the future, and this time more intelligently craft the question so such a high tolerance isn't needed, and then set `rtol = 0.03`, then my grade function overwrites the effect of `rtol = 0.07`, and most importantly this happens silently in the background.

In general, I hate to overcomplicate things, but if we wanted to "protect users from themselves", would a better option be to introduce a new function called `regrade()` instead to decouple the `grade()` action and the `regrade()` action? The advantage would be that the `regrade()` function would only be run when the Regrade button is clicked.

To be clear, I waffled back and forth whether to even post this because there is something so elegant and simple by just having 4 functions: `generate`, `parse`, `prepare`, and `grade`... so feel free to ignore this comment if purity should beat practicality in this case.

nwalters512:
@echau01 do you intend to revisit this? If not, I'm inclined to close it, as a large number of merge conflicts have built up in the past few years.

echau01:
Hi @nwalters512 , I don't intend to revisit this. More elaboration would be needed if we were to pick up this project again.

Also consider reverting #4406 if the "regradable" column is never used elsewhere.

PR #4647, opened 2021-08-09
Add GitHub action to publish prairielearn/* docker images

mwest1066:
Includes multi-arch builds to support ARM64/M1, following on from #4622

### DIFF
--- a/.github/workflows/images.yml
+++ b/.github/workflows/images.yml
@@ -11,24 +11,132 @@ jobs:
     steps:
       -
         name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v2 # https://github.com/marketplace/actions/checkout
       -
         name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
+        uses: docker/setup-qemu-action@v1 # https://github.com/marketplace/actions/docker-setup-qemu
       -
         name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@v1 # https://github.com/marketplace/actions/docker-setup-buildx
       -
         name: Login to DockerHub
-        uses: docker/login-action@v1
+        uses: docker/login-action@v1 # https://github.com/marketplace/actions/docker-login
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      ######################################################################################################
+      # plbase
       -
-        name: Build and push
+        name: Check whether images/plbase was modified
+        run: tools/check_path_modified.sh images/plbase images_plbase_modified
+      -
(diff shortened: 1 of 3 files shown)


.github/workflows/images.yml
```diff
@@ -0,0 +1,142 @@
+name: Images
+
+on:
+  push:
+    branches:
+      - 'master'
```
nwalters512, line 6: I believe Docker Hub gave us builds for branches too, but I think this is reasonable to avoid chewing up all our build minutes on image building.

tools/check_path_modified.sh
```diff
@@ -0,0 +1,21 @@
...
+if [ "$#" -ne 2 ]; then
+    echo "USAGE: $0 check_path env_var" >& 2
+    echo "Example: $0 images/plbase plbase_modified" >& 2
+    exit 1
+fi
+
+CHECK_PATH=$1
+ENV_VAR=$2
+
+if git diff --exit-code HEAD~1..HEAD -- ${CHECK_PATH}; then
```
nwalters512, line 16: The gaps in my shell script knowledge are showing 😂 I thought this was backwards at first, but it turns out shells treat `0` as true. The script is fine, this was just a TIL for me!

.github/workflows/images.yml
```diff
@@ -0,0 +1,142 @@
...
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2 # https://github.com/marketplace/actions/checkout
```
nwalters512, line 14: Nit, I tend to prefer the slightly more compact version with the `-` on the same line as the key. Not blocking on this at all.
```suggestion
      - name: Checkout
        uses: actions/checkout@v2 # https://github.com/marketplace/actions/checkout
```

nwalters512 (approved):

.github/workflows/images.yml
mwest1066: Yep, my thoughts exactly. We never really made much use of the branch builds. But you reminded me to delete the docs about this.

tools/check_path_modified.sh
mwest1066: Yeah, it's the world's most confusing thing!

.github/workflows/images.yml
mwest1066: This was a copy from [the example](https://github.com/marketplace/actions/build-and-push-docker-images#git-context) and I inherited their weird YAML syntax 😄. I agree with you and I changed the dashes to be on the same line.

tools/check_path_modified.sh
echuber2: Feel free to ping me on bash nonsense. (I've had to work with way too much bash since last year due to some poor choices.)

PR #4648, opened 2021-08-09
Improved tmux script

echuber2:
It looked like the old `start_workspace.sh` tmux launcher script was removed, so I renovated it to support the new makefile and added some features.

I renamed it to `start_workspace_tmux.sh` to clarify that it's mainly for getting a tmux session.


docker/start_workspace_tmux.sh
```diff
@@ -0,0 +1,123 @@
...
+# kill any containers with a name like workspace-*
+CONTAINERS=$(docker ps -aq --filter "name=workspace-")
+if [[ ! -z "$CONTAINERS" ]] ; then
+  echo Killing existing workspace containers: $CONTAINERS
+  docker kill $CONTAINERS
+  docker rm $CONTAINERS
+fi
+
+# ---
+# Not sure if this is still needed.
```
mwest1066, line 84: I don't think it's needed anymore.

docker/start_workspace_tmux.sh
```diff
@@ -0,0 +1,123 @@
+#!/bin/bash
```
mwest1066, line 1: Let's shift this to `tools/` --- I'm trying to keep `docker/` for the critical scripts needed for CI.

PR #4653, opened 2021-08-10
Enable bootstrap-table in questions page

### LINKED ISSUE #4558: Use bootstrap-table for Questions page
#4543 massively improved the Gradebook page using [bootstrap-table](https://bootstrap-table.com). Using this would also allow us to improve the Questions page and finally add support for comparing the usage of questions across course instances (see #988, #1018, #1503). We could have the current Questions page, but with one column per course instance, showing when questions were used. By default we could only show the current course instance (if there is one), which would mimic the current experience.

The use-case for this is to allow instructors to easily see which questions have been used on different assessment types in different terms. For example, you could see that a question hadn't been used in the last few terms and decide to put it on an exam in the current term. Alternatively, you might see that a question has been used too much recently, and decide to take it off the exam.

jonatanschroeder:
Resolves #4558

### DIFF
--- a/pages/instructorQuestions/instructorQuestions.ejs
+++ b/pages/instructorQuestions/instructorQuestions.ejs
@@ -3,6 +3,7 @@
   
   <head>
     <%- include('../partials/head'); %>
+    <script src="<%= node_modules_asset_path('underscore/underscore-min.js') %>"></script>
     <script src="<%= node_modules_asset_path('bootstrap-table/dist/bootstrap-table.min.js') %>"></script>
     <script src="<%= node_modules_asset_path('bootstrap-table/dist/extensions/sticky-header/bootstrap-table-sticky-header.min.js') %>"></script>
     <script src="<%= node_modules_asset_path('bootstrap-table/dist/extensions/filter-control/bootstrap-table-filter-control.min.js') %>"></script>
@@ -16,6 +17,7 @@
       $(function() {
           $('#questionsTable').bootstrapTable({
               //url: '<%= urlPrefix %>/<%= navPage %>/gradebook/raw_data.json',
+              data: <%- JSON.stringify(questions) %>,
               onPreBody: function() {
               },
               onResetView: function () {
@@ -31,9 +33,53 @@
           });
       });
 
+      function titleFormatter(title, question) {
+          var text = '';
+          if (question.sync_errors) {
+              text += `<button class="btn btn-xs mr-1" data-toggle="popover" data-title="Sync Errors"
+                               data-content="<pre style=&quot;background-color: black&quot; class=&quot;text-white rounded p-3&quot;>${_.escape(question.sync_errors_ansified)}</pre>">
+                    <i class="fa fa-times text-danger" aria-hidden="true"></i>
+                  </button>`;
+          } else if (question.sync_warnings) {
+              text += `<button class="btn btn-xs mr-1" data-toggle="popover" data-title="Sync Warnings"
+                          data-content="<pre style=&quot;background-color: black&quot; class=&quot;text-white rounded p-3&quot;>${_.escape(question.sync_warnings_ansified)}</pre>">
+                    <i class="fa fa-exclamation-triangle text-warning" aria-hidden="true"></i>
+                  </button>`;
+          }
+          text += `<a class="formatter-data" href="<%= urlPrefix %>/question/${question.id}/">${_.escape(question.title)}</a>`;
+          if (question.open_issue_count > 0) {
+              text += `<a class="badge badge-pill badge-danger ml-1" href="<%= urlPrefix %>/course_admin/issues?q=qid%3A${_.escape(question.qid)}">${question.open_issue_count}</a>`;
+          }
+          return text;
+      }
+
       function topicFormatter(topic, question) {
-          topic = JSON.parse(topic);
-          return `<span class="badge color-${topic.color}">${topic.name}</span>`;
+          return `<span class="badge color-${topic.color}">${_.escape(topic.name)}</span>`;
+      }
+
+      function tagsFormatter(tags, question) {
+          return _.map(tags ?? [], tag => `<span class="badge color-${tag.color}">${tag.name}</span>`).join(' ');
+      }
+
+      function versionFormatter(version, question) {
+          return `<span class="badge color-${version == 'v3' ? 'green1' : 'red1'}">${_.escape(version)}</span>`;
+      }
+
+      function assessmentsFormatter(assessments, question) {
+          return _.map(assessments ?? [], assessment => `<a href="<%= urlPrefix %>/assessment/${assessment.assessment_id}" class="badge color-${assessment.color} color-hover" onclick="event.stopPropagation();"><span>${_.escape(assessment.label)}</span></a>`).join(' ');
+      }
+
+      function genericFilterSearch(search, value, field, data) {
+          console.log($('<div>').html(value).find('.formatter-data').text(), search);
+          return $('<div>').html(value).find('.formatter-data').text().includes(search);
+      }
+
+      function elementListFilterSearch(search, value, field, data) {
+          var values = $('<div>').html(value).find('span')
+              .filter((i, elem) => $(elem).text().toUpperCase() == search.toUpperCase())
+              .length;
+          console.log(value, values, !!values);
+          return !!values;
       }
     </script>
     <style>
@@ -94,7 +140,9 @@
               <th data-field="title"
                   data-sortable="true"
                   data-class="align-middle"
+                  data-formatter="titleFormatter"
                   data-filter-control="input"
+                  data-filter-custom-search="genericFilterSearch"
                   data-switchable="true">Title</th>
               <th data-field="qid"
                   data-sortable="true"
@@ -105,92 +153,40 @@
                   data-sortable="true"
                   data-class="align-middle"
                   data-formatter="topicFormatter"
+                  data-filter-control="select"
+                  data-filter-data="json:<%= JSON.stringify(all_topics) %>"
+                  data-filter-custom-search="elementListFilterSearch"
                   data-switchable="true">Topic</th>
               <th data-field="tags"
                   data-sortable="false"
                   data-class="align-middle"
+                  data-formatter="tagsFormatter"
+                  data-filter-control="select"
+                  data-filter-data="json:<%= JSON.stringify(all_tags) %>"
+                  data-filter-custom-search="elementListFilterSearch"
                   data-switchable="true">Tags</th>
               <th data-field="display_type"
                   data-sortable="true"
                   data-class="align-middle"
+                  data-formatter="versionFormatter"
+                  data-filter-control="select"
+                  data-filter-data="json:<%= JSON.stringify(all_versions) %>"
+                  data-filter-custom-search="elementListFilterSearch"
                   data-switchable="true">Version</th>
               <% if (locals.course_instance) { %>
               <th data-field="assessments"
                   data-sortable="true"
                   data-class="align-middle"
+                  data-formatter="assessmentsFormatter"
+                  data-filter-control="select"
+                  data-filter-data="json:<%= JSON.stringify(all_assessments) %>"
+                  data-filter-custom-search="elementListFilterSearch"
                   data-switchable="true">Assessments</th>
               <% } %>
             </tr>
           </thead>
-            <tbody>
-              <% questions.forEach(function(question) { %>
-              <tr>
-                <td class="align-middle">
-                  <% if (question.sync_errors) { %>
-                  <button class="btn btn-xs mr-1" data-toggle="popover" data-title="Sync Errors"
-                          data-content="<pre style=&quot;background-color: black&quot; class=&quot;text-white rounded p-3&quot;><%= question.sync_errors_ansified %></pre>">
-                    <i class="fa fa-times text-danger" aria-hidden="true"></i>
-                  </button>
-                  <% } else if (question.sync_warnings) { %>
-                  <button class="btn btn-xs mr-1" data-toggle="popover" data-title="Sync Warnings"
-                          data-content="<pre style=&quot;background-color: black&quot; class=&quot;text-white rounded p-3&quot;><%= question.sync_warnings_ansified %></pre>">
-                    <i class="fa fa-exclamation-triangle text-warning" aria-hidden="true"></i>
-                  </button>
-                  <% } %>
-                  <a href="<%= urlPrefix %>/question/<%= question.id %>/"><%= question.title %></a>
-                  <%- include('../partials/issueBadge', {count: question.open_issue_count, issueQid: question.qid}); %>
-                </td>
-                <td class="align-middle">
-                  <%= question.qid %>
-                </td>
-                <td class="align-middle">
-                  <%= JSON.stringify(question.topic) %>
-                </td>
-                <td class="align-middle">
-                  <%- include('../partials/tags', {tags: question.tags}); %>
-                </td>
-                <td class="align-middle">
-                  <%- include('../partials/types', {type: question.display_type}); %>
-                </td>
-                <% if (locals.course_instance) { %>
-                <td class="align-middle">
-                  <%- include('../partials/assessments', {assessments: question.assessments}); %>
-                </td>
-                <% } %>
-              </tr>
-              <% }); %>
-            </tbody>
           </table>
         </div>
-        <script>
-                // $(function() {
-                //     $("#questionsTable").tablesorter({
-                //         widgetOptions: {
-                //             zebra: ["even", "odd"],
-                //             filter_reset: "#resetFilter",
-                //             filter_cssFilter: "form-control",
-                //             filter_functions: {
-                //                 3: {
-                //                     <% all_tags.forEach(function(tag) { %>
-                //                     "<%= escapeRegExp(tag.name) %>": function(e, n, f, i, $r, c, data) {
-                //                         return /\b<%= escapeRegExp(tag.name) %>\b/.test(e);
-                //                     },
-                //                     <% }); %>
-                //                 },
-                //                 <% if (locals.course_instance) { %>
-                //                 5: {
-                //                     <% all_assessments.forEach(function(assessment) { %>
-                //                     "<%= escapeRegExp(assessment.label) %>": function(e, n, f, i, $r, c, data) {
-                //                         return /\b<%= escapeRegExp(assessment.label) %>\b/.test(e);
-                //                     },
-                //                     <% }); %>
-                //                 },
-                //                 <% } %>
-                //             },
-                //         },
-                //     });
-                // });
-        </script>
       </div>
     </div>
 </body>
--- a/pages/instructorQuestions/instructorQuestions.ejs
+++ b/pages/instructorQuestions/instructorQuestions.ejs
@@ -1,136 +1,197 @@
 <!DOCTYPE html>
 <html lang="en">
-
-<head>
+  
+  <head>
     <%- include('../partials/head'); %>
-    <link href="<%= node_modules_asset_path('tablesorter/dist/css/theme.bootstrap.min.css') %>" rel="stylesheet" />
-    <script src="<%= node_modules_asset_path('tablesorter/dist/js/jquery.tablesorter.min.js') %>"></script>
-    <script src="<%= node_modules_asset_path('tablesorter/dist/js/jquery.tablesorter.widgets.min.js') %>"></script>
-</head>
+    <script src="<%= node_modules_asset_path('bootstrap-table/dist/bootstrap-table.min.js') %>"></script>
+    <script src="<%= node_modules_asset_path('bootstrap-table/dist/extensions/sticky-header/bootstrap-table-sticky-header.min.js') %>"></script>
+    <script src="<%= node_modules_asset_path('bootstrap-table/dist/extensions/filter-control/bootstrap-table-filter-control.min.js') %>"></script>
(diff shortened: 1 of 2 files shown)


mwest1066:
This is exciting!

jonatanschroeder:
> This is exciting!

@mwest1066 how about this?

image

mwest1066:
> @mwest1066 how about this?

Looks awesome! What do you think about using just "2021S1" for the column headers, to try and keep them narrow? Many too unclear?

mwest1066:
Would it be too awful of me to suggest considering #2343 as part of this work? Not merging the PRs, because that would create a horrible mess, but thinking about whether we could apply #2343 (or a variant of it) after this current PR?

mwest1066:
I'm wondering whether we should swap the order of the first two columns, and treat the QID as the unique identifier of the question (which it is, of course). That is, we'd make QID the first column and use it as the link that goes to the question. The title would become just another informational column like topics or tags.

To push this a bit further, is there anything useful we could do to take advantage of the hierarchical (directory) structure of QIDs? I'm thinking about tree-view sort of things, although I can't see any useful options in bootstrap-table. Anyway, I'm just bringing this up in case you have any good ideas 😄

mwest1066:
Any thoughts about [showing button text](https://examples.bootstrap-table.com/#options/show-button-text.html)? Would this encourage button usage or is it just too verbose? I'm thinking that it would be helpful for writing docs, because it's much easier to say "Click the 'XXX' button" when it's actually labeled XXX, but maybe we don't really want to write docs that describe buttons anyway.

mwest1066:
More things that I'm thinking about (but which I know shouldn't go in this PR):
* Inline editing of data (titles, QID, tags, etc)
* Add "X" buttons on the Assessment badges that remove the question from that assessment.
* Add a "+" button to the Assessment column cells to add the question to an assessment.
* Add a delete-question button (or a Action button) to each row.
* Add a [checkbox column](https://examples.bootstrap-table.com/#column-options/checkbox.html) to allow the selection of multiple questions, with an Action button that lets us operate on all checked questions (e.g., add all of them to an assessment).

One use case of the checkboxes would be to enable directory renaming within the questions directory. At the moment, if we have questions `foo/q1`, `foo/q2`, `foo/q3` and we want to rename `foo -> bar`, then there is no good way to do this from the web UI. I'm thinking we could select all three questions and have an Action -> Rename Folder command that lets us mass-rename their parent directory.

jonatanschroeder:
> More things that I'm thinking about (but which I know shouldn't go in this PR):
> 
> * Inline editing of data (titles, QID, tags, etc)
> * Add "X" buttons on the Assessment badges that remove the question from that assessment.
> * Add a "+" button to the Assessment column cells to add the question to an assessment.
> * Add a delete-question button (or a Action button) to each row.
> * Add a [checkbox column](https://examples.bootstrap-table.com/#column-options/checkbox.html) to allow the selection of multiple questions, with an Action button that lets us operate on all checked questions (e.g., add all of them to an assessment).
> 
> One use case of the checkboxes would be to enable directory renaming within the questions directory. At the moment, if we have questions `foo/q1`, `foo/q2`, `foo/q3` and we want to rename `foo -> bar`, then there is no good way to do this from the web UI. I'm thinking we could select all three questions and have an Action -> Rename Folder command that lets us mass-rename their parent directory.

These sound like interesting ideas. Do you want to create an issue? 😏 😄 

> I'm wondering whether we should swap the order of the first two columns, and treat the QID as the unique identifier of the question (which it is, of course). That is, we'd make QID the first column and use it as the link that goes to the question. The title would become just another informational column like topics or tags.

That's actually a great idea, since the title can technically be empty which would make the experience a bit weird. I also made it sticky in case the number of columns becomes large with instances. I also disabled word wrapping.

> To push this a bit further, is there anything useful we could do to take advantage of the hierarchical (directory) structure of QIDs? I'm thinking about tree-view sort of things, although I can't see any useful options in bootstrap-table. Anyway, I'm just bringing this up in case you have any good ideas 😄

The bootstrap-table [treegrid extension](https://bootstrap-table.com/docs/extensions/treegrid/) may be an option, but we may need to discuss the UI of directories.

> Would it be too awful of me to suggest considering #2343 as part of this work? Not merging the PRs, because that would create a horrible mess, but thinking about whether we could apply #2343 (or a variant of it) after this current PR?

That certainly looks like an interesting idea for future PR.

> Looks awesome! What do you think about using just "2021S1" for the column headers, to try and keep them narrow? Many too unclear?

It looks a bit unclear.

image

mwest1066:
> These sound like interesting ideas. Do you want to create an issue? 😏 😄

Hahahaha, done. #4655

> > I'm wondering whether we should swap the order of the first two columns, and treat the QID as the unique identifier of the question (which it is, of course). That is, we'd make QID the first column and use it as the link that goes to the question. The title would become just another informational column like topics or tags.
> 
> That's actually a great idea, since the title can technically be empty which would make the experience a bit weird. I also made it sticky in case the number of columns becomes large with instances. I also disabled word wrapping.

Sticky is an excellent point! I really like this system where we have a unique "primary key" as the first column in every table, and it's sticky.

> The bootstrap-table [treegrid extension](https://bootstrap-table.com/docs/extensions/treegrid/) may be an option, but we may need to discuss the UI of directories.

Oh wow, I love this! Let's use this! I think this would really help people to keep their questions organized.

I've been thinking for some time that directories in questions is really better than "topics". I feel like topics are now a bit redundant. What do you think? Basically the only advantage of "topics" is that they are color-coded.

> > Would it be too awful of me to suggest considering #2343 as part of this work? Not merging the PRs, because that would create a horrible mess, but thinking about whether we could apply #2343 (or a variant of it) after this current PR?
> 
> That certainly looks like an interesting idea for future PR.

Very diplomatic of you 😄. I agree we can forget about it for now.

> > Looks awesome! What do you think about using just "2021S1" for the column headers, to try and keep them narrow? Many too unclear?
> 
> It looks a bit unclear.

Yeah, agreed, but I really like the space saving 😄. What about using [column groups](https://examples.bootstrap-table.com/#welcomes/group-columns.html) to put a common "Assessments" label over all the course instance columns? I guess it will make the other columns look horrible, especially in the common case when we only have one course instance displayed.

jonatanschroeder:
> Any thoughts about [showing button text](https://examples.bootstrap-table.com/#options/show-button-text.html)? Would this encourage button usage or is it just too verbose? I'm thinking that it would be helpful for writing docs, because it's much easier to say "Click the 'XXX' button" when it's actually labeled XXX, but maybe we don't really want to write docs that describe buttons anyway.

Done:

image

mwest1066:
> > Any thoughts about [showing button text](https://examples.bootstrap-table.com/#options/show-button-text.html)? Would this encourage button usage or is it just too verbose? I'm thinking that it would be helpful for writing docs, because it's much easier to say "Click the 'XXX' button" when it's actually labeled XXX, but maybe we don't really want to write docs that describe buttons anyway.
> 
> Done:

Nice!

I'm not sure if this is relevant to your screenshots above, but for some reason, if I have Mac screenshots set to capture directly to the clipboard (either using Grab or Cmd-Ctrl-4) and then I paste straight into a github comment, the image comes out the correct size. If I first save to file or open in Preview and then copy, the image shoes up enlarged.

jonatanschroeder:
@mwest1066 I think this is in a merge-able state, two things to confirm:

* do you want to show button texts in gradebook and students page?
* the "Add Question" is now by itself as an action in the card header, should it stay there? Any change in style?

pages/instructorQuestions/instructorQuestions.ejs
```diff
@@ -1,136 +1,228 @@
...
+                  data-switchable="true">Topic</th>
+              <th data-field="tags"
+                  data-sortable="false"
+                  data-class="align-middle text-nowrap"
+                  data-formatter="tagsFormatter"
+                  data-filter-control="select"
+                  data-filter-data="func:tagsList"
+                  data-filter-custom-search="badgeFilterSearch"
+                  data-switchable="true">Tags</th>
+              <th data-field="display_type"
```
mwest1066, line 202: Maybe hide this column by default? Only old courses will have legacy v2 questions in them. Or maybe show it by default only if the course actually contains at least one v2 question?

pages/instructorQuestions/instructorQuestions.ejs
```diff
@@ -1,136 +1,228 @@
...
+              <th data-field="qid"
+                  data-sortable="true"
+                  data-class="align-middle sticky-column"
+                  data-formatter="qidFormatter"
+                  data-filter-control="input"
+                  data-switchable="true">QID</th>
+              <th data-field="title"
+                  data-sortable="true"
+                  data-class="align-middle text-nowrap"
+                  data-filter-control="input"
```
mwest1066, line 183: The title filter seems to be broken for me (the QID filter works fine). For example:

image

image

pages/instructorQuestions/instructorQuestions.ejs
```diff
@@ -1,136 +1,228 @@
...
-            </script>
+          </div>
+        </div>
+
+        <table id="questionsTable"
+               data-data="<%= JSON.stringify(questions) %>"
+               data-classes="table table-sm table-hover table-bordered"
+               data-thead-classes="thead-light"
+               data-filter-control="true"
+               data-show-filter-control-switch="true"
```
mwest1066, line 159: Do we really want this button? Do you think people want to be able to hide the filter controls?

image

pages/instructorQuestions/instructorQuestions.ejs
```diff
@@ -1,136 +1,228 @@
...
+               data-show-search-clear-button="true"
```
mwest1066, line 160: Assuming this is the button below, I'm a bit worried that users won't know what "filters" are, and the trash icon looks scary. Could we maybe label it something like "Show all questions" or "Show everything"? With a friendlier icon of some kind? 😄

image

pages/instructorQuestions/instructorQuestions.ejs
```diff
@@ -1,136 +1,228 @@
...
+                  data-filter-control="input"
+                  data-switchable="true">QID</th>
+              <th data-field="title"
+                  data-sortable="true"
+                  data-class="align-middle text-nowrap"
+                  data-filter-control="input"
+                  data-filter-custom-search="genericFilterSearch"
+                  data-switchable="true">Title</th>
+              <th data-field="topic"
+                  data-sortable="true"
```
mwest1066, line 187: Sorting by topic seems incorrect.

image

pages/instructorQuestions/instructorQuestions.ejs
```diff
@@ -1,136 +1,228 @@
...
+                  data-class="align-middle text-nowrap"
+                  data-formatter="versionFormatter"
+                  data-filter-control="select"
+                  data-filter-data="func:versionList"
+                  data-filter-custom-search="badgeFilterSearch"
+                  data-switchable="true">Version</th>
+              <% locals.course_instances.forEach(course_instance => { %>
+              <th data-field="assessments_<%= course_instance.id %>"
+                  data-sortable="true"
```
mwest1066, line 212: This column has sort buttons but they don't seem to do anything when I click them.

pages/instructorQuestions/instructorQuestions.ejs
mwest1066: Also, I need to click the topic sort button 3 times to cycle between the different states. It seems to do (1) sort ascending, (2) nothing, (3) sort descending.

pages/instructorQuestions/instructorQuestions.ejs
```diff
@@ -1,136 +1,228 @@
...
+               data-pagination-v-align="top"
+               data-page-list="[50,100,200,500,unlimited]"
+               data-page-size="50"
+               data-smart-display="false"
+               data-show-extended-pagination="true"
+               data-toolbar=".fixed-table-pagination"
+               data-sticky-header="true">
+          <thead>
+            <tr>
+              <th data-field="qid"
```
mwest1066, line 174: Any chance of a tree view? 😄 (waiting for a future PR is completely reasonable)

mwest1066:
> * do you want to show button texts in gradebook and students page?

I think so, if it's possible to still fit everything on the page. What do you think?

> * the "Add Question" is now by itself as an action in the card header, should it stay there? Any change in style?

I'm not in love with the current placement, but I'm not sure I've got a better idea. Could it be integrated with the other table buttons? (I don't know whether bootstrap-table lets you add your own buttons)

pages/instructorQuestions/instructorQuestions.ejs
jonatanschroeder: I remembered this a few seconds after submitting my comment, I'm working on it right now. Will let you know if it looks good.

mwest1066: ❤️

jonatanschroeder: treeview seems quite unstable from my limited tests. I'd prefer to move it to a separate PR where this can be revised more extensively.

pages/instructorQuestions/instructorQuestions.ejs
jonatanschroeder: Sort by assessment removed.

pages/instructorQuestions/instructorQuestions.ejs
jonatanschroeder: This was a leftover from the custom search used when the title had a potential popover/issues count, which now moved to QID. Fixed.

jonatanschroeder:
> > * the "Add Question" is now by itself as an action in the card header, should it stay there? Any change in style?
> 
> I'm not in love with the current placement, but I'm not sure I've got a better idea. Could it be integrated with the other table buttons? (I don't know whether bootstrap-table lets you add your own buttons)

Done, can you check if it looks better?

jonatanschroeder:
> > * do you want to show button texts in gradebook and students page?
> 
> I think so, if it's possible to still fit everything on the page. What do you think?

Also done, can you check?

jonatanschroeder:
> Yeah, agreed, but I really like the space saving 😄. What about using [column groups](https://examples.bootstrap-table.com/#welcomes/group-columns.html) to put a common "Assessments" label over all the course instance columns? I guess it will make the other columns look horrible, especially in the common case when we only have one course instance displayed.

I'm really not a fan, it looks weird. The table header is already pretty tall due to the filters, I think adding a new group row would probably make it a bit too much.

pages/instructorQuestions/instructorQuestions.ejs
mwest1066: Very reasonable.

pages/instructorQuestions/instructorQuestions.ejs
mwest1066: It turns out that I dislike "Show all questions" even more, sorry! I actually clicked it thinking that it was going to remove pagination 😄

I guess I think we should go back to "Clear filters". Perhaps with a generic X icon (or no icon) rather than the trash?

mwest1066:
> > > * do you want to show button texts in gradebook and students page?
> 
> Also done, can you check?

Looks great! For both of those pages:

* Also put the pagination left/right controls to the left of the pagination text
* Add 10 and 20 as pagination options
* Change tooltip text for buttons to be more than just their labels.
    - "Refresh" tooltip: "Refresh all data on page"
    - "Auto Refresh" tooltip: "Auto refresh all data every 30 seconds" (or whatever the time period is)
    - "Columns" tooltip: "Select columns to display"
    - "Download" tooltip: "Download gradebook data in CSV format"
    - "Student Only" -- already has a good tooltip

mwest1066:
> > > * the "Add Question" is now by itself as an action in the card header, should it stay there? Any change in style?
> > 
> > 
> > I'm not in love with the current placement, but I'm not sure I've got a better idea. Could it be integrated with the other table buttons? (I don't know whether bootstrap-table lets you add your own buttons)
> 
> Done, can you check if it looks better?

I like it! Much better!

The buttons have different combinations of rounded and square edges on them:

image

In contrast, the Gradebook icons are using what looks like a bootstrap [button group](https://getbootstrap.com/docs/4.6/components/button-group/):

image

I don't have a strong preference for which style is best.

mwest1066:
> > Yeah, agreed, but I really like the space saving 😄. What about using [column groups](https://examples.bootstrap-table.com/#welcomes/group-columns.html) to put a common "Assessments" label over all the course instance columns? I guess it will make the other columns look horrible, especially in the common case when we only have one course instance displayed.
> 
> I'm really not a fan, it looks weird. The table header is already pretty tall due to the filters, I think adding a new group row would probably make it a bit too much.

I reluctantly agree 😄

mwest1066:
The right-hand button group is not centered on the Gradebook and Students pages. Alternatively, we could get rid of the 10-point vertical margin on the pagination controls?

image

image

mwest1066:
Also, the Gradebook page card should be titled "Gradebook" rather than "Users".

mwest1066:
> Unfortunately at this point it doesn't seem like `bootstrap-table` has support to change the title to something different than the text

No worries, this is absolutely not critical. We can just leave it as is. I'd rather keep our code clean than hack things to try and add better tooltips.

> All your other comments are addressed.

Awesome! Thanks!

mwest1066:
> Is it safe to assume that `locals.course_instances` corresponds to the instances the current user has access to?

Yes.

> I'm thinking of filtering them by the course instances that are in `locals.course_instances`, does that make sense?

Absolutely, good catch!

pages/instructorQuestions/instructorQuestions.ejs
```diff
@@ -1,136 +1,253 @@
...
+                  data-filter-control-placeholder="(All Versions)"
+                  data-filter-data="func:versionList"
+                  data-filter-custom-search="badgeFilterSearch"
+                  data-visible="<%= has_legacy_questions %>"
+                  data-switchable="true">Version</th>
+              <% locals.course_instances.forEach(course_instance => { %>
+              <th data-field="assessments_<%= course_instance.id %>"
+                  data-class="align-middle text-nowrap"
+                  data-formatter="assessments<%= course_instance.id %>Formatter"
+                  data-filter-control="select"
```
mwest1066, line 239: Any chance of a natural sort here?

image

pages/instructorQuestions/instructorQuestions.ejs
```diff
@@ -1,136 +1,253 @@
...
+  
+  <body>
+    <script>
+      $(function() {
+          $('#questionsTable').bootstrapTable({
+              icons: {
+                  columns: 'fa-th-list',
+              },
+              buttons: {
+                  clearFilters: {
```
mwest1066, line 23: This button doesn't seem to do anything when I click it. The current filters all stay there with no changes. No errors on the browser JS console.

mwest1066: I'd be fine with using the built-in clear-filters button (tooltip is not that important).

jonatanschroeder:
> > Is it safe to assume that `locals.course_instances` corresponds to the instances the current user has access to?
> 
> Yes.
> 
> > I'm thinking of filtering them by the course instances that are in `locals.course_instances`, does that make sense?
> 
> Absolutely, good catch!

How to I test this? In other words, what pattern of user permission (before #3020) would have access to the Questions page but would not have access to a particular course instance?

mwest1066:
> How to I test this? In other words, what pattern of user permission (before #3020) would have access to the Questions page but would not have access to a particular course instance?

I believe this happens if someone has no course role (that is, they are not listed on the course-level Access page) but they are an `Instructor` in individual `infoCourseInstance.json` files.

jonatanschroeder:
> > How to I test this? In other words, what pattern of user permission (before #3020) would have access to the Questions page but would not have access to a particular course instance?
> 
> I believe this happens if someone has no course role (that is, they are not listed on the course-level Access page) but they are an `Instructor` in individual `infoCourseInstance.json` files.

I tried that, but that user doesn't have access to the page at all.

mwest1066:
> I tried that, but that user doesn't have access to the page at all.

I will admit that I'm so deep into #3020 that I can't remember what the current system does 😄

I suggest filtering to only use CIs in `locals.course_instances` and not worrying too much about testing. Alternatively, you could make a temporary branch to merge together this PR with #3020 and test there?

pages/instructorQuestions/instructorQuestions.ejs
jonatanschroeder: I tried playing around with this, but I believe there's a bug in the extension that is ignoring the [`data-filter-order-by`](https://bootstrap-table.com/docs/extensions/filter-control/#filterorderby) option and always sorting in ascending order of text representation. Sorry, if there is a solution, it would be hacky and hard.

pages/instructorQuestions/instructorQuestions.ejs
jonatanschroeder: Fixed.

pages/instructorQuestions/instructorQuestions.ejs
mwest1066: No worries, we can just forget about this. Thanks for trying!

jonatanschroeder:
> I suggest filtering to only use CIs in `locals.course_instances` and not worrying too much about testing. Alternatively, you could make a temporary branch to merge together this PR with #3020 and test there?

I tested it by brute-forcing the list of CIs, which now works as expected. I believe that's the last pending issue.

mwest1066:
> I tested it by brute-forcing the list of CIs, which now works as expected. I believe that's the last pending issue.

Excellent solution 😄

mwest1066 (approved):

PR #4656, opened 2021-08-10
Add "feedback" attribute to pl-multiple-choice

Cris-Moran:
Allows for developers to pass in a hint attribute to the `pl-answer` element in `pl-multiple-choice`. A visual demo of a hint is shown below:
Screen Shot 2021-08-10 at 3 51 00 PM
Potential issues:
- No way to display a hint next to answer choices created by `all-of-the-above` and `none-of-the-above` attributes using this  method, since they don’t appear as a pl-answer in question.html
- On some mobile devices, the hint and score badge can stick outside of their parent container.


Cris-Moran:
**_I have reviewed and hereby sign the CLA_**

mwest1066:
recheck

mwest1066:
@Cris-Moran I'm sorry for the poor wording from the CLA bot, but it actually needs you to paste that text into a comment _without formatting_. Just the plain text "I have reviewed and hereby sign the CLA" (no quotes) in a comment. Thanks!

Cris-Moran:
Ah okay, thank you for the clarification!

Cris-Moran:

elements/pl-multiple-choice/pl-multiple-choice.css
```diff
@@ -7,3 +7,7 @@
     min-width: 1.25em;
     overflow-wrap: normal;
 }
+
+.pl-multiple-choice-hint {
+    border-left: 4px solid #999;
```
jonatanschroeder, line 12: I wonder if it may be better to move all styling for this hint to the CSS (i.e., replace classes `mr-1 ml-1 pl-2` in the hint div with `margin-right`, `margin-left` and `padding-left` properties here). This could help with for consistency and make it easier to change in the future.

Also, I think it may be a good idea to introduce a light background or very thin border-top/bottom/right to distinguish it from the question itself.

elements/pl-multiple-choice/pl-multiple-choice.mustache
```diff
@@ -14,7 +13,13 @@
             {{^hide_letter_keys}}<div class="pl-multiple-choice-key-label">({{key}})</div>{{/hide_letter_keys}}
             <div class="ml-1 mr-1">{{{html}}}</div>
         </label>
-            
+
+        {{#display_hint}}
+        <div id="hint" class="pl-multiple-choice-hint ml-4 mr-1 pl-2">
```
jonatanschroeder, line 18:
```suggestion
        <div class="pl-multiple-choice-hint">
```

See comment in CSS file above. Also, this element doesn't need an `id`.

elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -15,6 +15,7 @@
 HIDE_LETTER_KEYS_DEFAULT = False
 EXTERNAL_JSON_CORRECT_KEY_DEFAULT = 'correct'
 EXTERNAL_JSON_INCORRECT_KEY_DEFAULT = 'incorrect'
+HINT_DEFAULT = ''
```
jonatanschroeder, line 18:
```suggestion
HINT_DEFAULT = None
```

elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -222,7 +224,9 @@ def render(element_html, data):
                 'key': answer['key'],
                 'checked': (submitted_key == answer['key']),
                 'html': answer['html'],
-                'display_score_badge': display_score and submitted_key == answer['key']
+                'display_score_badge': display_score and submitted_key == answer['key'],
+                'display_hint': submitted_key == answer['key'] and answer['hint'] != HINT_DEFAULT,
```
jonatanschroeder, line 228:
```suggestion
                'display_hint': submitted_key == answer['key'] and answer['hint'] is not None,
```

jonatanschroeder (commented):
This looks great! Can you please also delete the .DS_Store file from the commit?

Before this is approved, please:
* Describe the new attribute in the documentation (`docs/elements.md`);
* Add the attribute to some example course question (probably `exampleCourse/questions/element/multipleChoice/question.html`).

If you're feeling adventurous, would you be ok with adding a similar option in `pl-checkbox`? Maybe `pl-dropdown` too, let me know what you think.

elements/pl-multiple-choice/pl-multiple-choice.css
Cris-Moran: Will do. I added a light gray background to the hint, and a border radius so the corners don't look sharp. I'll commit the changes shortly.
Screen Shot 2021-08-12 at 1 58 21 PM

Cris-Moran:
I'll play around with adding hints for `pl-checkbox` and `pl-dropdown` as well. Once I figure out how to implement hints for them I'll submit more PRs.

Cris-Moran:
Since the hint is now more distinguishable from the answer, I reduced the spacing between them by a little bit. Here's how it looks now:
Screen Shot 2021-08-13 at 11 27 12 AM

docs/elements.md
```diff
@@ -129,6 +129,7 @@ a `pl-answer` that has attributes:
 Attribute | Type | Default | Description
 --- | --- | --- | ---
 `correct` | boolean | false | Is this a correct answer to the question?
+`hint` | string | special | Text to be displayed next to submitted answer after question is graded.
```
jonatanschroeder, line 132:
```suggestion
`hint` | string | — | Helper text to be displayed to the student next to the option after question is graded if this option has been selected by the student.
```

exampleCourse/questions/element/multipleChoice/question.html
```diff
@@ -169,3 +169,29 @@
...
+
+<div class="card my-2">
+    <div class="card-header">
+        Part 7
+    </div>
+
+    <div class="card-body">
+
+        <pl-question-panel>
+            <p>Here, you can observe the behavior of the <code>hint</code> attribute in <code>pl-answer</code>.</p>
```
jonatanschroeder, line 181: You probably want to expand on what the hint is expected to convey, as a suggestion, and comment on what is expected to happen when an option with a hint is selected.

elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -185,8 +187,8 @@ def prepare(element_html, data):
     # overwriting previous choice(s)
     display_answers = []
     correct_answer = None
-    for (i, (index, correct, html)) in enumerate(sampled_answers):
-        keyed_answer = {'key': pl.index2key(i), 'html': html}
+    for (i, (index, correct, html, hint)) in enumerate(sampled_answers):
```
jonatanschroeder, line 190: Since you're changing this tuple to have 4 elements, you also need to update other places that handle this tuple, like lines 51 and 54, which get correct/incorrect answers from JSON files.

jonatanschroeder: And lines 174 and 182, for all-of-these/none-of-these.

elements/pl-multiple-choice/pl-multiple-choice.css
```diff
@@ -7,3 +7,12 @@
     min-width: 1.25em;
     overflow-wrap: normal;
 }
+
+.pl-multiple-choice-hint {
+    border-left: .25rem solid #999;
+    background-color: #F9F9F9;
```
jonatanschroeder, line 13:
```suggestion
    background-color: rgba(0,0,0,0.05);
```
This is a suggestion, feel free to adjust if needed. The rationale for using opacity instead of color is that, if the background is set to another color (e.g., the instructor changed it to something else with CSS), this will still make the hint slightly darker than the background. A similar change may be a good consideration for the border as well.

jonatanschroeder:
> * No way to display a hint next to answer choices created by `all-of-the-above` and `none-of-the-above` attributes using this  method, since they don’t appear as a pl-answer in question.html

This could be addressed with an element-level attribute, like `none-of-the-above-hint`, if you think this is important.

> * On some mobile devices, the hint and score badge can stick outside of their parent container.

I don't have an easy way to test this, can you include a screenshot? Though I don't think this would be a big problem.

elements/pl-multiple-choice/pl-multiple-choice.py
Cris-Moran: Alright, I'll put `None` into the JSON file tuples in lines 51 and 54 for now since I'm not sure how hints will work with them at the moment, and I'll add something similar to the `none-of-the-above-hint` attribute you talked about which will be inserted into the tuple in lines 174 and 182.

Cris-Moran:
> I don't have an easy way to test this, can you include a screenshot? Though I don't think this would be a big problem.

This issue only came up when I emulated a Galaxy Fold display in Chrome Dev Tools. As you can see the red box with the 'x' extends beyond its border:
Screen Shot 2021-08-13 at 5 43 12 PM

I don't have a Galaxy Fold to test this on. However, I tested the hints using Chrome on my own mobile device (which isn't a Galaxy Fold) and it looked exactly the same as when I selected it in the Chrome Dev Tools, so I believe it's accurate.

elements/pl-multiple-choice/pl-multiple-choice.css
Cris-Moran: Good suggestion, I changed both the border and background to use this instead.

jonatanschroeder:
We were talking about this feature today, and here are a few things that I'd like to ask you to change:

* Instead of "hint", call this "feedback". The idea of hint usually comes before submission, so to avoid confusion we'd like to clarify that this comes after submission.
* The `grade(data)` function in `pl-multiple-choice.py` should include this feedback as part of the `data['partial_scores']` object (together with score and weight). This would allow the feedback to be:
    * updated by the `grade` function of the question, which would allow instructors more flexibility in manipulating it;
    * better listed in logs;
    * included in downloaded files and updated via file upload.
* To ensure the update above can be used properly, the `render(data)` function in `pl-multiple-choice.py` should take the feedback from `partial_scores` (similar to the how the score for the badge is obtained) instead of from the answers list.
* This will probably come as consequence from the items below, but the feedback should only show after "Save & Grade", not after "Save Only".

Cris-Moran:
Alright, just got done making the requested changes.

exampleCourse/questions/element/multipleChoice/question.html
```diff
@@ -169,3 +169,30 @@
...
+<div class="card my-2">
+    <div class="card-header">
+        Part 7
+    </div>
+
+    <div class="card-body">
+
+        <pl-question-panel>
+            <p>Here, you can observe the behavior of the <code>feedback</code> attribute in <code>pl-answer</code>. Feedback is expected to convey a small piece of information that explains why an 
+                answer is correct or incorrect. When an answer with a hint is selected and graded, a gray stripe followed by a light gray box containing the hint will be displayed next to the answer.</p>
```
jonatanschroeder, line 182:
```suggestion
            <p>Here, you can observe the behavior of the <code>feedback</code> attribute in <code>pl-answer</code>. Feedback is expected to convey a small piece of information that explains why an 
                answer is correct or incorrect. When an answer with a feedback setting is selected and graded, a small box containing the feedback is displayed next to the answer.</p>
```

Cris-Moran (commented):
Sorry, this was a mistake. Please dismiss this review.

elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -333,9 +340,13 @@ def grade(element_html, data):
 
     score = 0
     if (submitted_key is not None and submitted_key == correct_key):
-        score = 1
+        score = 1           
 
-    data['partial_scores'][name] = {'score': score, 'weight': weight}
+    for option in data['params'][name]:
+        if option['key'] == submitted_key:
+            feedback = option['feedback']
```
Cris-Moran, line 347: One thing I'd like to bring up: in the `grade()` function, the `data['submitted_answers']` object doesn't contain the feedback I inserted into the answer tuples, so at the moment I'm not sure if there's another way to grab the feedback from the submitted answer without using a for loop.

jonatanschroeder: I believe this is a reasonable solution.

jonatanschroeder:
It seems tests are failing because you haven't updated the `test` function. You should update `data['partial_scores']` with the value that is expected to be set by `grade`, which includes the feedback.

Cris-Moran:
Ah I see. I'll fix that shortly.

Cris-Moran:
I fixed the issue by setting feedback to `None` in the `test()` function

Cris-Moran:
Hmm it seems to still be failing. Looks like I need to put something other than `None` into the `data['partial_scores']` object

elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -364,11 +364,18 @@ def test(element_html, data):
     result = data['test_type']
     if result == 'correct':
         data['raw_submitted_answers'][name] = data['correct_answers'][name]['key']
-        data['partial_scores'][name] = {'score': 1, 'weight': weight}
+        for option in data['params'][name]:
+            if option['key'] == correct_key:
+                feedback = option['feedback']
+        data['partial_scores'][name] = {'score': 1, 'weight': weight, 'feedback': feedback}
```
jonatanschroeder, line 370:
```suggestion
        feedback = data['correct_answers'][name].get('feedback', None)
        data['partial_scores'][name] = {'score': 1, 'weight': weight, 'feedback': feedback}
```

I'm not 100% sure this would work, but you should test it.

Cris-Moran: I tested this locally and it appears to work. I'll go ahead and commit it.

jonatanschroeder (approved):

Cris-Moran:
Before this is merged, I'd like to propose changing the spacing of the feedback again. I tested the feature with a different question and to me it looks cramped. For this reason I'd like to change the `margin-left` of the feedback from `0.5rem` (first image) to `0.75rem` (second image). It's a subtle difference, but I think it looks better.
Screen Shot 2021-08-18 at 6 03 15 PM
Screen Shot 2021-08-18 at 6 01 38 PM

elements/pl-multiple-choice/pl-multiple-choice.mustache
```diff
@@ -14,7 +13,13 @@
             {{^hide_letter_keys}}<div class="pl-multiple-choice-key-label">({{key}})</div>{{/hide_letter_keys}}
             <div class="ml-1 mr-1">{{{html}}}</div>
         </label>
-            
+
+        {{#display_feedback}}
+        <div class="pl-multiple-choice-feedback">
+            {{{feedback}}}
+        </div>
+        {{/display_feedback}}
```
mwest1066, line 21: I love the display of feedback in the question panel next to the selected item. But in the case where the student makes multiple submissions, the feedback will only be visible for the most recent submission (in the question panel). Can we additionally display the feedback in the submission panel, next to the submitted answer? That way the student will be able to see the history of their feedback together with their history of submitted answers.

docs/elements.md
```diff
@@ -129,6 +131,7 @@ a `pl-answer` that has attributes:
 Attribute | Type | Default | Description
 --- | --- | --- | ---
 `correct` | boolean | false | Is this a correct answer to the question?
+`feedback` | string | — | Helper text to be displayed to the student next to the option after question is graded if this option has been selected by the student.
```
mwest1066, line 134: The feedback is actually HTML, I think? That is, we template it in with triple-braces. Maybe clarify that here?
```suggestion
`feedback` | string | — | Helper text (HTML) to be displayed to the student next to the option after question is graded if this option has been selected by the student.
```

mwest1066: What happens if the feedback is very long? Does it look horrible? I wonder whether we should add an option like `show-feedback-at-end` to put the feedback after the multiple choice element? Maybe I'm overthinking this, however.

Cris-Moran: Here's what a long hint looks like:
Screen Shot 2021-08-30 at 8 43 49 PM
For an example of how this looks on a smaller screen, see https://github.com/PrairieLearn/PrairieLearn/pull/4656#issuecomment-898783046

elements/pl-multiple-choice/pl-multiple-choice.mustache
Cris-Moran: Just implemented this, here's what it looks like:
Screen Shot 2021-08-30 at 8 49 01 PM

Cris-Moran:
Just pushed all the requested changes.

docs/elements.md
mwest1066: Ok, that's kind of horrible but at least you can still read it 😄 --- I'm fine with this.

elements/pl-multiple-choice/pl-multiple-choice.mustache
mwest1066: Perfect!

mwest1066 (approved):

mwest1066:
@jonatanschroeder please merge when you are ready.

exampleCourse/questions/element/checkbox/question.html
```diff
@@ -257,3 +257,28 @@
...
+            <pl-answer>               The ability to work on a project by yourself.</pl-answer>
+            <pl-answer>               Having only one person make all of the decisions.</pl-answer>
+            <pl-answer correct="true">Constructive dialog and increased internal group motivation.</pl-answer>
+            <pl-answer correct="true">Personal accountability to the team to work on a project.</pl-answer>
+            <pl-answer correct="true">Friendship.</pl-answer>
+            <pl-answer>               The resources to work on a small project instead of a large project. </pl-answer>
+            <pl-answer correct="true">Prepares you for a team-environment in the workplace or a research group.</pl-answer>
+        </pl-checkbox>
+    </div>
+</div>
```
jonatanschroeder, line 284: @Cris-Moran did you add this to the wrong PR?

Cris-Moran: I don't remember writing this. I believe this appeared when I pulled from upstream.

jonatanschroeder: Doesn't matter, I removed it, it was causing tests to fail and is not relevant anyway.

PR #4658, opened 2021-08-11
Replace references to core scripts with node modules

### LINKED ISSUE #4654: Replace usage of underscore.js with lodash
There are currently two libraries in use by the system that define a `_` variable: lodash and underscore.js. lodash is used extensively more, but a few pages use underscore.js. There are differences in behaviour (e.g., `_.keyBy` in lodash roughly corresponds to `_.object` in underscore.js) which may cause confusion by developers. I suggest we replace all internal uses of underscore.js with lodash for consistency. I don't think we should remove underscore.js, since some elements have been using it and it's plausible that some instructors may have copied those elements for customization (or used underscore.js in general), but we should avoid using it where possible.

jonatanschroeder:
Replaced references to:

* `underscore.js` and `lodash.min.js` -> `node_modules/lodash/lodash.min.sh` (resolves #4654 );
* `sha1.js` -> not in use in modern elements/codes/questions;
* `dropzone.js` -> `node_modules/dropzone/dist/dropzone.js`;

Old copies of core scripts are kept for backwards compatibility. References to `PrairieUtil.js` (mainly code to create a copy to clipboard button) are kept, as well as any code for deprecated elements and v2 questions.

Since `sylvester.js` doesn't have an easy browser-friendly NPM package (the existing packages are mostly for node), it's kept as is for now as well.

Still missing: documentation to encourage new development using node modules instead of core scripts were appropriate.

### DIFF
--- a/docs/devElements.md
+++ b/docs/devElements.md
@@ -125,7 +125,7 @@ configure them for your element as follows:
     "controller": "pl-my-element.py",
     "dependencies": {
         "coreScripts": [
-            "d3.min.js"
+            "PrairieUtil.js"
         ],
         "nodeModulesScripts": [
             "three/build/three.min.js"
@@ -149,7 +149,7 @@ The different types of dependency properties currently available are summarized 
 Property | Description
 --- | ---
 `coreStyles` | The styles required by this element, relative to `[PrairieLearn directory]/public/stylesheets`.
-`coreScripts` | The scripts required by this element, relative to `[PrairieLearn directory]/public/javascripts`.
+`coreScripts` | The scripts required by this element, relative to `[PrairieLearn directory]/public/javascripts`. Scripts in this directory are mainly used for compatibility with legacy elements and questions, and should be avoided if an alternative is available.
 `nodeModulesStyles` | The styles required by this element, relative to `[PrairieLearn directory]/node_modules`.
 `nodeModulesScripts` | The scripts required by this element, relative to `[PrairieLearn directory]/node_modules`.
 `elementStyles` | The styles required by this element relative to the element's directory, which is either `[PrairieLearn directory]/elements/this-element-name` or `[course directory]/elements/this-element-name`.
--- a/docs/devElements.md
+++ b/docs/devElements.md
@@ -110,23 +110,14 @@ The above function descriptions describe the typical variables that will be read
 
 ### Element dependencies
 
-It's likely that your element will depend on certain client-side assets,
-such as scripts or stylesheets. To keep clean separation of HTML, CSS, and JS,
-you can place those dependencies in other files. If you depend on libraries like
-`lodash` or `d3`, you can also express that need. PrairieLearn will compile a list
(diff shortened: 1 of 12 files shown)


elements/pl-drawing/info.json
```diff
@@ -2,12 +2,11 @@
...
         "nodeModulesScripts": [
-           "fabric/dist/fabric.js"
+            "fabric/dist/fabric.js",
+            "lodash/lodash.js"
         ],
         "coreScripts": [
-            "sylvester.js",
-            "sha1.js",
-            "underscore.js"
+            "sylvester.js"
```
mwest1066, line 9: Is `sylvester.js` actually used in `pl-drawing`? @nicknytko?

jonatanschroeder: It is, removing it breaks the question.

jonatanschroeder: https://github.com/PrairieLearn/PrairieLearn/blob/master/elements/pl-drawing/mechanicsObjects.js#L4

nicknytko: Yes, its used for vector math

mwest1066: Cool, thanks for confirming!

docs/devElements.md
```diff
@@ -125,7 +125,7 @@ configure them for your element as follows:
     "controller": "pl-my-element.py",
     "dependencies": {
         "coreScripts": [
-            "d3.min.js"
+            "PrairieUtil.js"
```
mwest1066, line 128: How about we just remove `coreScripts` from the example completely, so that people are less likely to copy it?

docs/devElements.md
```diff
@@ -149,7 +149,7 @@ The different types of dependency properties currently available are summarized
 Property | Description
 --- | ---
 `coreStyles` | The styles required by this element, relative to `[PrairieLearn directory]/public/stylesheets`.
-`coreScripts` | The scripts required by this element, relative to `[PrairieLearn directory]/public/javascripts`.
+`coreScripts` | The scripts required by this element, relative to `[PrairieLearn directory]/public/javascripts`. Scripts in this directory are mainly used for compatibility with legacy elements and questions, and should be avoided if an alternative is available.
```
mwest1066, line 152: I'd be happy to completely remove `coreScripts` from the docs, so that it only exists as an undocumented legacy feature. Or maybe we should be even stronger in our language and say `This directory is deprecated and should not be used.`

jonatanschroeder: I took the properties out of the table and added stronger warning. I also added the [`deprecated`](http://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.3) flag on the schemas.

docs/devElements.md
jonatanschroeder: Done.

docs/devElements.md
mwest1066: Nice!

mwest1066 (approved):

PR #4659, opened 2021-08-13
Drop enum role to complete permissions migration

mwest1066:
This PR has been split out of #3020 to perform the removal of the old `role` permission system. This should only be merged and deployed after a successful deploy of #3020. In principle it's fine to deploy both together, but this way is safer and permits rollback if anything goes wrong.

* [x] Also revert #4688


migrations/238_enum_role__drop.sql
```diff
@@ -0,0 +1,10 @@
+DELETE FROM assessment_access_rules WHERE role > 'Student';
+DELETE FROM course_instance_access_rules WHERE role > 'Student';
```
mwest1066, line 2: We could also delete `enrollments WHERE role > 'Student'` as requested by @jonatanschroeder to ensure old course statistics are still correct.

nwalters512:
@mwest1066 is it still worth it applying this migration?

mwest1066:
I think we should apply this to keep our DB tidy. But we should update to include the suggestion from @jonatanschroeder ?

jonatanschroeder:
> I think we should apply this to keep our DB tidy. But we should update to include the suggestion from @jonatanschroeder ?

Are you referring to #4798?

nwalters512:
Ha, it definitely does! A lot has changed in the past 3 years...

nwalters512:
Split out into #9615.

nwalters512 (approved):
@mwest1066 I'll let you make the final call about merging this.

mwest1066:
> @mwest1066 I'll let you make the final call about merging this.

I can't see any reason to retain these unused columns at this point so I'll go ahead and merge this.

mwest1066:
Actually, what is going to happen with the deletion of historical `assessment_access_rules` and `course_instance_access_rules` that have `role > Student`? Aren't we just going to recreate them on the next sync but with no `role`? To be clear, I don't think that this is a security issue because we already ignore `role`, but it makes it seem a bit pointless to delete these rules if we are just going to recreate them.

nwalters512:
@mwest1066 we haven't been syncing any access rules with a non-empty `role` that isn't `Student` since #3020 landed. You can see here where we do that filtering:

https://github.com/PrairieLearn/PrairieLearn/blob/55d022d190a3914f2d96c7150aec50c81f64cf24/apps/prairielearn/src/sync/fromDisk/assessments.ts#L58-L80

The only reason we'd have any rows with a non-Student role in the DB is a very old or deleted course that hasn't been synced since August 2021. This migration will handle those.

mwest1066:
> @mwest1066 we haven't been syncing any access rules with a non-empty `role` that isn't `Student` since #3020 landed.

Excellent, thanks for reminding me about all of this.

PR #4663, opened 2021-08-15
Java autograder basic sandbox

### LINKED ISSUE #3831: Java autograder sandboxing
The Java autograder needs to implement a more extensive sandboxing mechanism. Currently a student's code may be able to generate a result file with a successful grade without major problems.

jonatanschroeder:
Fixes #3831 

* [x] Move code compilation to autograder script;
* [x] Run autograder as sandbox user;
* [x] Ensure Java autograder cannot write to results.json file directly;
* [ ] Review what files sandbox user has access to;
* [ ] More extensive tests.

### DIFF
--- a/graders/java/Dockerfile
+++ b/graders/java/Dockerfile
@@ -4,8 +4,7 @@ LABEL maintainer="jonatan@cs.ubc.ca"
 
 ARG DEBIAN_FRONTEND=noninteractive
 
-RUN apt update
-RUN apt install -y python3.8 python3-pip make openjdk-11-jdk-headless && apt clean
+RUN apt update && apt install -y python3.8 python3-pip make openjdk-11-jdk-headless jq && apt clean
 
 RUN ln -sf /usr/bin/python3.8 /usr/bin/python3
 
--- a/graders/java/JUnitAutograder.java
+++ b/graders/java/JUnitAutograder.java
@@ -25,10 +25,6 @@ import static org.junit.platform.engine.discovery.DiscoverySelectors.selectClass
 
 public class JUnitAutograder implements TestExecutionListener {
 
-    private static final File baseDirectoryForStudentCode = new File("/grade/student");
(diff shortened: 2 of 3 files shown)


jonatanschroeder:
@mwest1066 About the news items we were talking about this morning: in essence what is blocked now and wasn't before is access to any data within `/grade` (except for a new directory that is created by the autograder, called `/grade/classpath`, where all compiled classes are copied/compiled to) and the `/javagrader` directory (containing the autograder classes). I made sure the sandbox user (`sbuser`) has a home directory (`/home/sbuser`), and the sandbox environment has access to this directory with no restrictions, and the working directory is that home directory. This means any code (JUnit test or student code) would be able to create and access files in the home directory.

I believe under these circumstances the only conceivable/realistic location that worked before and is not working now would be code that accesses `/grade/data/data.json`, or code that requires student-submitted files in `/grade/student` other than compiled Java classes. I don't think enough people would have done any Java-based code that depends on these, so a news item may not be necessary. Thoughts?

nwalters512:
> Do you mean a code that would try to exploit the docker code, that would be successful to change the code before this change and unsuccessful after it? I can come up with an example or two, how do you want me to document it? Is a java file here enough?

Yeah, that'd be great - you can just put it here in a comment. I'm mostly looking to independently verify that these changes do what they mean to.

jonatanschroeder:
Here's a code that would give them 100% before, but would fail now:

```java
import java.io.*;
import java.util.*;

public class Example {
    
    public static int square(int x) {
        File f = new File("/grade/results/results.json");
        PrintWriter w = null;
        try {
            f.getParentFile().mkdirs();
            w =  new PrintWriter(new FileWriter(f));
            w.println("{\"score\": 1}");
        } catch (IOException ex) {
            ex.printStackTrace();
        } finally {
            w.close();
            f.setWritable(false);
            System.exit(0);
        }
        return 0;
    }
}
````

PR #4664, opened 2021-08-16
UI tweaks to questions, gradebook and students tables

jonatanschroeder:
* [x] Ctrl-F (or Cmd-F) sends focus to search/filter function in table
* [x] Add pagination to bottom of screen as well.


mwest1066 (approved):
Works great in my local testing!

PR #4667, opened 2021-08-17
Fixes for two pl-order-blocks issues

### LINKED ISSUE #4665: pl-order-blocks not rendering special HTML characters correctly
Blocks containing HTML special characters (&lt;, &gt;, &quot;, &amp;, etc.) are being rendered with HTML encoding, escaping those characters as literal `&lt;` etc. This was not an issue in the past; a block like this:

```
<pl-answer>if number < 12 or number > 24:</pl-answer>
```

used to look like this (running on the Berkeley PL instance):

image

The same question running on a local up-to-date PL:

image

The `user_code.py` that is generated for external grading also sees the characters written with their HTML encoding, which is obviously bad.

The `pl-answer` above is actually malformed HTML, but the issue persists in all ways even when rewritten with an escaped inner HTML on the answer:

```
<pl-answer>if number &lt; 12 or number &gt; 24:</pl-answer>
```

@jonatanschroeder identified a quick fix for the rendering in the browser:

image

This does fix the rendering, but the user_code.py is not fixed.

### LINKED ISSUE #4666: pl-order-blocks crashes when indentation="false"
I get this error on any pl-order-blocks element with indentation="false", either by omitting the attribute or explicitly setting to false.

```
  File "/PrairieLearn/lib/python-caller-trampoline.py", line 182, in <module>
    worker_loop()
  File "/PrairieLearn/lib/python-caller-trampoline.py", line 128, in worker_loop
    val = method(*args)
  File "/PrairieLearn/elements/pl-order-blocks/pl-order-blocks.py", line 334, in parse
    indent = int(answer['indent'])
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
  File "/PrairieLearn/lib/python-caller-trampoline.py", line 194, in <module>
```

nealterrell:
Fix #4666 and Fix #4665.


elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -331,8 +332,8 @@ def parse(element_html, data):
 
         answer_code = ''
         for index, answer in enumerate(student_answer):
-            indent = int(answer['indent'])
-            answer_code += ('    ' * indent) + answer['inner_html'] + '\n'
+            indent = int(answer['indent'] or 0)
+            answer_code += ('    ' * indent) + html.unescape(answer['inner_html']) + '\n'
```
jonatanschroeder, line 336: This may be a problem if the inner HTML has any formatting (e.g., `x = <b>2 *</b>y`). The user will see a Python line that happens to have a bold part. Should those be stripped? This can be done with (note that `lxml.html` is already imported):
```suggestion
            answer_code += ('    ' * indent) + lxml.html.fromstring(answer['inner_html']).text_content() + '\n'
```

SethPoulsen (approved):
Looks good to me!

elements/pl-order-blocks/pl-order-blocks.py
nealterrell: What's the official stance on whether an element like this is legal/well-formed?

```
<pl-answer>x < 10</pl-answer>
```

I realize now that other element types (radio, checkbox) expect this to be escaped properly in the question.html. With the `{{{inner_html}}}` change, we get proper rendering of  a block like

```
<pl-answer>x = &quot;&lt;b&gt;10&gt;/b&lt;</pl-answer>
```

as `x = "<b>10</b>"`, and with `html.unescape` the correct text is written to user_code. I think that's correct for all cases?

jonatanschroeder: > What's the official stance on whether an element like this is legal/well-formed?
> 
> ```
> <pl-answer>x < 10</pl-answer>
> ```

This is not well-formed HTML, so it shouldn't be accepted. I believe having `<pl-answer>x = <b>5</b></pl-answer>` should add formatting to the block text and show `x = 5` in the generated file, while `<pl-answer>x = &lt;b&gt;5&lt;/b&gt;</pl-answer>` should show `<b>5</b>` verbatim in the block and in the file, as that makes more sense in terms of what I believe an instructor would expect to see. Is this what happens when you use this code?

nealterrell: The second works as desired; the first leaves `x = <b>5</b>` in the file. I'll try the lxml change!

nealterrell: I can confirm that the block `x = &quot;&lt;b&gt;True&lt;/b&gt;&quot;` renders to the screen and to user_code.py as  "x = &quot;&lt;b&gt;True&lt;/b&gt;&quot;", and the block `x = <b>5</b>` renders to screen as "x = <b>5</b>" [the 5 is bold] and to user_code as "x = 5".

nealterrell:
I have improved the solution with Jonatan's suggestion!

elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: hmm...it seems like this is consistent with what the `pl-code` element is doing here: 
https://github.com/PrairieLearn/PrairieLearn/blob/master/elements/pl-code/pl-code.mustache#L10

Am I missing something? It seems like `pl-code` would also allow the malformed html. (I'm not saying it _should_, just saying that it currently does. Maybe we should make it a future issue to do proper validation both here and in `pl-code`, and in `pl-file-editor`?)

`pl-code` does do `html.escape` when the code is brought in from a file: 
https://github.com/PrairieLearn/PrairieLearn/blob/master/elements/pl-code/pl-code.py#L156

jonatanschroeder: Well, I think `pl-code` is expected to show raw formatting, while `pl-order-blocks` doesn't have it as a main goal. In other words, `pl-code` will show the file exactly as is, while `pl-order-blocks` can be used to show stuff that is not necessarily code. Who knows, maybe you can add `<pl-answer><pl-code>...</pl-code></pl-answer>`...

nealterrell: I can confirm that these changes do allow `<pl-answer><pl-code>` inside a block. The rendered layout isn't great, but it does work, and the generated file only contains the code text, not the `pl-code` block tags.

nealterrell: image

pl-code inside a block.

jonatanschroeder (approved):

PR #4672, opened 2021-08-18
Fix questions page case where no instance is available

### LINKED ISSUE #4673: Course "Viewer" with no course instance access cannot view Questions
Reported by Eric Larson.

This occurred in XC 101 where users default to have `course_role = Viewer` and no course instance access.

pl_error_xc101_demo_access

jonatanschroeder:
Fixes issue listed recently on Slack.

Fixes #4673

### DIFF
--- a/pages/instructorQuestions/instructorQuestions.ejs
+++ b/pages/instructorQuestions/instructorQuestions.ejs
@@ -138,7 +138,8 @@
           return !!values;
       }
       
-      //<% locals.course_instances.forEach(course_instance => { %>
+      //<% if (locals.course_instances) {
+               locals.course_instances.forEach(course_instance => { %>
       function assessments<%= course_instance.id %>List() {
           return assessmentsByCourseInstanceList(<%= course_instance.id %>);
       }
@@ -146,7 +147,8 @@
       function assessments<%= course_instance.id %>Formatter(_, question) {
           return assessmentsByCourseInstanceFormatter(<%= course_instance.id %>, question);
       }
-      //<% }); %>
+      //<%     });
+           } %>
 
     </script>
     <style>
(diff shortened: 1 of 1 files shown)


mwest1066:
Hahaha, #4674

jonatanschroeder:
Different approaches. Which do you prefer? Should it remain NULL?

mwest1066:
> Different approaches. Which do you prefer? Should it remain NULL?

I tend to think it's cleaner to always have `locals.course_instances` be an array, so we don't have to check whether it's NULL before every use?

mwest1066:
Can you reproduce the bug? I'm failing to repro locally in XC 101.

mwest1066:
> Can you reproduce the bug? I'm failing to repro locally in XC 101.

Oh sorry, I can repro just fine. And I confirmed that #4674 fixes the problem.

mwest1066:
> > Different approaches. Which do you prefer? Should it remain NULL?
> 
> I tend to think it's cleaner to always have `locals.course_instances` be an array, so we don't have to check whether it's NULL before every use?

Which fix do you think is cleaner?

jonatanschroeder:
> I tend to think it's cleaner to always have `locals.course_instances` be an array, so we don't have to check whether it's NULL before every use?
> 
> Which fix do you think is cleaner?

We already check before use, at least where I could find, e.g.:

https://github.com/PrairieLearn/PrairieLearn/blob/ec5d41c856feb0c47328b88b63938feadb1199f1/pages/partials/navbarInstructor.ejs#L71

The line above should work as expected anyway, so I don't have a problem with either approach. We could even use both 😄  I'll review yours, it's actually a bit cleaner. Closing this one.

mwest1066:
> We could even use both 😄 I'll review yours, it's actually a bit cleaner. Closing this one.

Sounds good 😄

PR #4685, opened 2021-08-20
Add collaborative editing to Jupyter workspace

nicknytko:
Adds real-time editing by default (https://jupyterlab.readthedocs.io/en/stable/user/rtc.html), and unfreezes the base docker image version.  For some reason, there's some bug in newer versions of Jupyter where it doesn't recognize the `port` from the config python file, so I've instead added it as a launch argument.  

A demo of this running: https://www.prairielearn.org/pl/course_instance/128596/instructor/question/9064781/preview


nicknytko:
After some discussion w/ Matt, this is going to be temporarily pushed to a new image: `prairielearn/workspace-jupyterlab-collab` that we'll test on the CS 357 kids.  If nothing horribly breaks, then this should be okay to enable globally in the default jupyter workspace.

mwest1066:
Do you want to make this PR a draft until we've tested the image in 357? That would involve making/pushing the `-collab` image to docker hub by hand.

nicknytko:
I think we've been testing this for a few group activities now without any real breakage -- @mfsilva22 do you have any comments or thoughts on things you've noticed?

mfsilva22:
I did not notice anything, not got negative feedback from students. I think this is a fantastic feature.

mwest1066:
I know the actual change is very small, but the impact of this seems very big (and awesome!). Should we add a news item?

mwest1066:
> I know the actual change is very small, but the impact of this seems very big (and awesome!). Should we add a news item?

And is there any way to record a gif of two mirrored notebook windows next to each other, with one being edited and the other one showing the synchronized edits?

coatless:
Was there any reason this PR stalled out? It seems like the change here would yield good benefits by being enabled on the default workspace-jupyterlab container: 

https://prairielearn.slack.com/archives/C266KEH9A/p1666818420266999

The only issue here is the underlying `jupyterlab` docker image version might need to be bumped.

coatless:
Was there any reason this PR stalled out? It seems like the change here would yield good benefits by being enabled on the default workspace-jupyterlab container: 

https://prairielearn.slack.com/archives/C266KEH9A/p1666818420266999

The only issue here is the underlying `jupyterlab` docker image version might need to be bumped.

news_items/036_collaborative_jupyter/index.html
```diff
@@ -26,7 +26,7 @@
 </style>
 
 <p>
-    Thanks to an update by <a href="https://github.com/nicknytko">Nicolas Nytko</a> (CS MS '22), Jupyter Lab workspaces now support collaborative editing by multiple users in group assessments.  Students can now edit the same notebook and run code cells, seeing these changes occur <i>in real time</i>.
+    Thanks to an update by <a href="https://github.com/nicknytko">Nicolas Nytko</a> (CS PhD), Jupyter Lab workspaces now support collaborative editing by multiple users in group assessments.  Students can now edit the same notebook and run code cells, seeing these changes occur <i>in real time</i>.
```
coatless: @nicknytko Congrats on leveling up!

nwalters512:
https://github.com/jupyterlab/jupyterlab/issues/12596 mentions "content duplication", which is almost certainly the same thing that Mariana has been seeing in CS 357 group notebooks. It looks like there's no fix for that yet, unfortunately.

news_items/041_collaborative_jupyter/info.json
```diff
@@ -0,0 +1,6 @@
+{
+    "uuid": "161df1e0-a543-4147-948b-97bd30ee3c1a",
+    "title": "Collaborative Editing in Jupyter Workspaces",
```
mwest1066, line 3:
```suggestion
    "title": "Collaborative editing in Jupyter Workspaces",
```

news_items/041_collaborative_jupyter/index.html
```diff
@@ -0,0 +1,43 @@
...
+ .code-block div {
+     padding: 0.5rem;
+     display: block;
+     margin-bottom: -1em;
+     line-height: 125%;
+ }
+</style>
+
+<p>
+    Thanks to an update by <a href="https://github.com/nicknytko">Nicolas Nytko</a> (CS PhD), Jupyter Lab workspaces now support collaborative editing by multiple users in group assessments.  Students can now edit the same notebook and run code cells, seeing these changes occur <i>in real time</i>.
```
mwest1066, line 29:
```suggestion
    Thanks to an update by <a href="https://github.com/nicknytko">Nicolas Nytko</a> (CS Ph.D.), JupyterLab workspaces now support collaborative editing by multiple users in group assessments.  Students can now edit the same notebook and run code cells, seeing these changes occur <i>in real time</i>.
```

workspaces/jupyterlab/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM jupyter/datascience-notebook:aec555e49be6
+FROM jupyter/datascience-notebook:lab-3.5.0
```
mwest1066, line 1: Update this to 3.6.1, which I think is the most recent?

mwest1066 (approved):

nicknytko:
Update: I've bumped the base image version to 3.6.1 like Matt has pointed out and pushed the new image to `prairielearn/workspace-jupyter-collab` to trial in CS357.  We can see going forward if this fixes the duplication issues.

nwalters512:
Mariana has been experiencing some issues since using this latest image. Here's one report I filed: https://github.com/jupyterlab/jupyterlab/issues/13930#issuecomment-1448716566

nicknytko:
> 

How frequently has this been popping up?  Do we need to roll back the image to the previous version?

mfsilva22:
This seems to be impacting many students today during the group activity. Students mostly report:
-"Can the development team check if the autosave function is fixing the window to the very top every time it executes? It is very annoying that it forces us to jump to the very top every once a while. The only difference I see with current notebook is the autosave function. If we cannot resolve this issue, can we temporarily revert back to the manual save version? That would be really helpful."
- "Jupyter notebook keeps jumping to the top. This has been happening to me for the past two days and is currently happening during GA time which is annoying. Any way to fix this? This wasn't happening before."

I guess we should revert to the previous image? Sadly, we fixed one issue to just have another one.

nwalters512:
I reported this issue upstream: https://github.com/jupyterlab/jupyterlab/issues/14106. I feel a little bad because we don't have a ton of information yet, but it seems widespread enough to not be a fluke.

nwalters512:
@nicknytko what do you think about upgrading to a newer Jupyter image but *not* enabling RTC until we're more confident that the scroll-to-top issue has been resolved?

nicknytko:
> @nicknytko what do you think about upgrading to a newer Jupyter image but _not_ enabling RTC until we're more confident that the scroll-to-top issue has been resolved?

Yeah, I like this idea. I can spin off the changes into a separate PR (and update image to latest) and keep this one as the PR that flips the switch for collaborative editing.

nwalters512:
@nicknytko is there anything in this PR you want to save? Otherwise I'm inclined to close it. We're all set up to enable it in the new `prairielearn/workspace-jupyterlab-python` image once the upstream issue with `jupyter-collaboration` is resolved:

https://github.com/PrairieLearn/PrairieLearn/blob/c95fa0dbc3b04b3a6b6dc0d5a834e8053f446821/workspaces/jupyterlab-python/jupyter_server_config.py#L16-L20

PR #4691, opened 2021-08-22
Add workaround for Firefox not showing borders in sticky columns

jonatanschroeder:
Firefox has a bug whereas if a background is set to a `td` where the table has `border-collapse`, the border disappears. The workaround is to use `background-clip: padding-box`, which ensures the background doesn't apply to the border itself. Should not make a difference for other browsers, since this setting should only make a difference from the default if the border uses opaque/transparent elements.

See https://bugzilla.mozilla.org/show_bug.cgi?id=688556, https://developer.mozilla.org/en-US/docs/Web/CSS/background-clip

### DIFF
--- a/pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
+++ b/pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
@@ -16,6 +16,7 @@
       position: sticky;
       left: 0;
       background: white;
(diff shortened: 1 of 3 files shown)


mwest1066:
10 year old bug 😱 (although some of ours are also getting old!)

mwest1066 (approved):

PR #4693, opened 2021-08-23
Update automata-lib version

jasonxia17:
We would like to use new features in version 5.0.0 to develop questions for CS 374. It would be great if this could be merged and deployed relatively soon, because we're aiming to release these questions in a week or two. Thanks!

### DIFF
--- a/images/plbase/python-requirements.txt
+++ b/images/plbase/python-requirements.txt
@@ -26,3 +26,4 @@ PuLP==2.4
 text-unidecode==1.3
 statsmodels==0.12.2
 openpyxl==3.0.7
+automata-lib==5.0.0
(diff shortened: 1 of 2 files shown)


jasonxia17:

images/plbase/python-requirements.txt
```diff
@@ -26,3 +26,4 @@ PuLP==2.4
 text-unidecode==1.3
 statsmodels==0.12.2
 openpyxl==3.0.7
+automata-lib==5.0.0
```
mwest1066, line 29: We are trying to avoid adding more specialized libraries to the base image. Instead, you can install this yourself into the course `serveFilesCourse` directory and import it directly from question code. Would that work for you as an alternative?
```suggestion
```

jasonxia17: Could you elaborate on how to install it into `serverFilesCourse`? I'm not super familiar with how pip works.

PR #4699, opened 2021-08-24
Add new R autograder based on STAT 430/477 DSPM grader

eddelbuettel:
The PR adds the autograder used by the 'Data Science Programming Methods' course (initially STAT 430 (2018-2020),  now STAT 447).  It is based on the 'current Ubuntu LTS stack' from Rocker which combines backports of the current R package (now: R 4.1.1, built my Michael Rutter of Penn U at Erie based on the Debian package I maintain, and distributed via CRAN) along with over 5k CRAN binaries for Ubuntu LTS that Michael maintains via a Launchpad PPA.   In combination, this makes for very fast Docker builds (== almost all binaries) based on widely used and tested binaries.  It also adds the components we have found useful in writing R questions for homework and quizzen 'just like CI tests' wrapped in a security-minded evaluation helper that can also display colorized diffs in case of incorrect answers.

As per earlier discussion, it renames the earlier R autograder directory to r-previous.  The two could coexist; the earlier one might need an update as it picked a version-stable Rocker build and is therefore ... still at the now two years old versions of its components.  

Both approaches have their merits; I happen to prefer the approach this PR brings and had good experiences over the years.

The plan is to complement this PR with another adding two or three sample questions showing the autograder in use, and highlighting some of its capabilties.


graders/r/README.md
```diff
@@ -1,10 +1,31 @@
-## PrairieLearn Autograding R Docker Image
+## rocker-pl
 
+This is a Docker image working as an external grader to be paired with PrairieLearn.  It has been
+used in the [STAT 430](https://stat430.com) (2018-2020) and [STAT 447](https://stat430.com) (since
+2021) courses on _Data Science Programming Methods_ at the University of Illinois at
+Urbana-Champiagn. In a previous course iteration, a CentOS7-based container was used via Python. We
```
ctessum, line 6: *Champaign

eddelbuettel: Ouch. Will fix. Thanks for spotting that.

Done in 3191a19 (after rebasing to origin/master)

ctessum:
Thanks for adding this. Speaking on behalf of the instructor team for CEE 202, would it be possible to add an example or two of best practices for using this to autograde problems? Thanks for considering!

eddelbuettel:
You're ears must have been rigning.  @daviddalpiaz was just talking about you.  Come on over to the PL slack into and into channel #r .  We are getting a little closer to critical mass and I plan to start a repo with exactly these types of  'demo' or 'best practices' questions.

ctessum:
Will do, Thanks!

graders/r/Dockerfile
```diff
@@ -1,19 +1,64 @@
...
+	r-cran-xts \
+	sqlite3 \
+        sudo \
+        && echo "options(diffobj.brightness=\"dark\")" >> /etc/R/Rprofile.site
+
+# Install additional R packages from CRAN (on top of the ones pre-built as r-cran-*)
+RUN install.r bench flexdashboard microbenchmark lintr ttdo unix
+
+# Install plr -- for now (?) from GH; also install visualTest
+RUN installGithub.r stat447/plr MangoTheCat/visualTest
```
mwest1066, line 56: Can we include `stat447/plr` directly into the `graders/r` directory? For example, like the way `graders/python/python_autograder` contains the python autograder library directly within `graders/python`. We prefer this because it keeps all the code for the grader together in the same repo.

graders/r/README.md
```diff
@@ -1,10 +1,31 @@
-## PrairieLearn Autograding R Docker Image
+## rocker-pl
 
+This is a Docker image working as an external grader to be paired with PrairieLearn.  It has been
+used in the [STAT 430](https://stat430.com) (2018-2020) and [STAT 447](https://stat430.com) (since
+2021) courses on _Data Science Programming Methods_ at the University of Illinois at
+Urbana-Champaign. In a previous course iteration, a CentOS7-based container was used via Python. We
+now drive R directly from PrairieLearn evaluating code in this container. However, the container is
+perfectly generic and can be used for general R work as well.
```
mwest1066, line 8: I suggest leaving out historical information unless it's important for understanding the current code?
```suggestion
Urbana-Champaign. However, the container is
perfectly generic and can be used for general R work as well.
```

graders/r/README.md
```diff
@@ -1,10 +1,31 @@
...
+
+and more. We also include a few command-line applications such as `git` and `sqlite3`.
+
+### Authors
+
+Alton Barbehenn and Dirk Eddelbuettel
+
+### License
+
+AGPL-3
```
mwest1066, line 31: No need for licenses on individual source components.
```suggestion
```

graders/r/README.md
```diff
@@ -1,10 +1,31 @@
...
+releases---both current versions of R itself, as well as current CRAN packages.
 
+### R Packages Installed
 
+See the Dockerfile for full details but we include
+
+- tidyverse and data.table for data manipulation;
+- RUnit, testthat and tinytest for testing;
+- microbenchmark, rbenchmark and bench for timing;
+- shiny, flexdashboard, dygraphs for shiny;
```
mwest1066, line 21: Can we include some example questions in `exampleCourse/questions/demo/autograder/r`? These would parallel the example questions in the neighboring `python`, `c`, and `java` directories.

It would also be great to include a question called `exampleCourse/questions/demo/workspace/rstudio` that uses both the new `prairielearn/workspace-rstudio` image and grades the resulting code with this autograder.

mwest1066 (commented):
Looks great!

graders/r/README.md
eddelbuettel: That is planned.  See https://github.com/eddelbuettel/pl-r-demos   -- and I think you I even briefly riffed on Slack that we want to bring this in first, then adjust the question and bring the questions in.

graders/r/README.md
eddelbuettel: Done in bc30a9661 (after a rebase too).

graders/r/Dockerfile
mwest1066: I guess I was hoping to have the dev version be the version directly in `graders/r/plr/...`.

But I'm also fine to keep it in a separate repo in the PL org. I can make `PrairieLearn/plr` if you'd rather do this? Let me know.

graders/r/README.md
eddelbuettel: With that, any other outstanding issues you would like to see addressed, and that are not covered in my push earlier?

mwest1066: Just a decision about https://github.com/PrairieLearn/PrairieLearn/pull/4699#discussion_r701345955 --- can we either put the (unpacked) dev version of `plr` into `graders/r/plr/...` or shift it to `PrairieLearn/plr`? I'm fine either way, just let me know which one you prefer.

graders/r/Dockerfile
eddelbuettel: Sorry -- I plainly misunderstood.

R packages have a very standard layout, and it (to my preferences, and to that of many others) works best when they sit top-level in their own repo.  `plr` is first and foremost an R package, which is why I am hesitant to have it a few levels down in the pl repo.  I would prefer not to.  (There are counter examples for other 'large scope multi-language projects' as _e.g._ Apache Arrow, or the MLPACK repo.  This is a little different.)

So yes `PrairieLearn/plr` (or another name) would be my preference.  It is currently an incomprehensibly short name ... because we had no prior constraints.  So the repo could be `PrairieLearn/pl-r-helpers` or `-misc` or ...

mwest1066: Great! I created https://github.com/PrairieLearn/pl-r-helpers and you are Admin for it. I left the repo uninitialized in case you want to push the full history of `stat447/plr` into the new repo.

Let me know after you've switched the current PR over to use `PrairieLearn/pl-r-helpers` and I'll merge it.

eddelbuettel: Awesome.  Was on a call for a bit.   We could also
- nuke pl-r-helpers (to create room)
- I transfer my existing plr
- we double check I can admin, else you do, and we rename it to pl-r-helpers
and that way keep the (trivial, but it is the right thing (tm)) history.  Worth it?

mwest1066: If you have a checked-out local copy of `stat447/plr` then you should be able to do:
```
git remote rename origin old-origin
git remote add origin git@github.com:PrairieLearn/pl-r-helpers.git
git branch -M main
git push -u origin main
```

This will switch your checked-out repo to use `PrairieLearn/pl-r-helpers` as it's origin and then push a full copy of the history into this new repo.

eddelbuettel: Doh. I was looking for something like that and I overcomplicated.  Decentralised revision control systems for the win, and a very nice wave of the clue bat!  I tried to do it all inside `magit` which fell apart when I realised I had not accepted your invitation so I couldn't actually write there.  Doh.  

All good now.

With that, back to installing the R package it represent  from git remote as before (now hitting a sibbling directory) ?

mwest1066: Yep, jut switch to install from `PrairieLearn/pl-r-helpers`.

eddelbuettel: Done in c371ea9de

mwest1066: Thanks!

mwest1066:
@coatless just a heads up that this PR is proposing to move the existing R autograder in PL to `graders/r-previous` and make `graders/r` be the grader maintained by Dirk. Let me know if you have any objection to this. Thanks!

coatless:
@mwest1066 no objection on my end re: naming

mwest1066:
> @mwest1066 no objection on my end re: naming

Great, thanks! If you'd prefer a different name than `r-previous` then let me know.

PR #4706, opened 2021-08-25
Manual grading col split tests

andrewstec:
## These tests ensure that:

- the `gradingMethod` deprecated property on the infoQuestion.json does not break with future changes.
- the new `gradingMethods` property logic works when a question is configured with it
- adds an implemented `gradingMethods` question with Internal, External, and Manual grading methods to see that it functions.

## Problems to highlight:

-  One note of concern is that the external grading jobs are overwriting internal grading jobs, as they finish after the Internal grading job. We can see this by running the new `gradingMethods` configured question.
- If a question is configured with both Internal and External grading methods, only the Internal past submission panel appears on the Question.js student/instructor views. This will have to be fixed to fully leverage the view for Manual grading. Adding the Manual grading method does not have any such effect.

We should probably merge this PR, after which I will merge the `manual-grading-col-split` branch into the `manual-grading-anti-ta-overlap-spin-lock-views`. After some minor integration work, and some test updates, the manual grading feature should be functional and ready for a code review-- and also covered with code. We can then fix the two issues under the "Problems to highlight".


SethPoulsen:
> **CLA Assistant Lite bot:** Thanks for your contribution! Like many open source projects, we ask that you sign our [Contributor License Agreement](https://github.com/PrairieLearn/cla/blob/main/CLA.md) before we can accept your contribution. You can sign the CLA by posting a comment with the below format:
> 
> _**I have reviewed and hereby sign the CLA**_
> 
> **7** out of **8** committers have signed the CLA.
>  @andrewstec
>  @jonatanschroeder
>  @echuber2
>  @SethPoulsen
>  @SanjitBasker
>  @nealterrell
>  @jasonxia17
>  @tbretl
> You can retrigger this bot by commenting **recheck** in this Pull Request

Seems like there are some people listed here that shouldn't be (like myself). Is there some sort of merging or rebasing problem going on here?

PR #4717, opened 2021-08-28
Remove unused enrollment counts in authorization

mwest1066:
Since #3020 was deployed we've been seeing much higher DB CPU loads. The main culprit is `sprocs/users_is_instructor_in_course_instance`, which we are currently calling about 20,000 times per second at the moment (averaged over the day, much more during peak load). I think the vast majority of these calls are coming from here: https://github.com/PrairieLearn/PrairieLearn/blob/1dfd2d9be1f713d96f75d625dbc2f449e4844c69/sprocs/course_instances_with_staff_access.sql#L68

The current flow is:
* On every authz in `middlewares/middlewares/authzCourseOrInstance.sql` we call `sprocs/course_instances_with_staff_access` to get the list of course instances to put in the navbar dropdown menu that allows rapid switching between course instances.
* In addition to finding the list of accessible course instances, `sprocs/course_instances_with_staff_access` also currently computes the enrollment in each one. I guess this was intended for display to the user in the dropdown switcher, but we don't currently display it (or use it in any other way).
* This enrollment calculation is counting the number of `enrollments` in each course instance. It is excluding instructors from this count by calling `sprocs/users_is_instructor_in_course_instance` on every enrollment.
* The function `sprocs/users_is_instructor_in_course_instance` is quite cheap (average time to execute is 33 microseconds) but, given the number of calls, it ends up being very expensive in total.

This PR removes the unused calculation of the enrollment count for the course instances. After this, I think the cost of authorization should be roughly the same as it was before #3020.

### DIFF
--- a/pages/instructorCourseAdminInstances/instructorCourseAdminInstances.js
+++ b/pages/instructorCourseAdminInstances/instructorCourseAdminInstances.js
@@ -15,6 +15,7 @@ const { CourseInstanceAddEditor } = require('../../lib/editors');
 
 const fs = require('fs-extra');
 const async = require('async');
+const _ = require('lodash');
 
 router.get('/', function(req, res, next) {
     async.series([
@@ -25,6 +26,21 @@ router.get('/', function(req, res, next) {
                         res.locals.needToSync = true;
                     } else return ERR(err, callback);
                 }
+                callback(null);
+            });
+        },
+        (callback) => {
+            if (!res.locals.authz_data || !res.locals.authz_data.course_instances)
+                return callback(null);
+            const params = {
+                course_id: res.locals.course.id,
+            };
+            sqldb.query(sql.select_enrollment_counts, params, (err, result) => {
+                if (ERR(err, callback)) return;
+                res.locals.authz_data.course_instances.forEach(ci => {
+                    var row = _.find(result.rows, row => row.course_instance_id == ci.id);
+                    ci.number = row?.number || 0;
+                });
                 callback(null);
             });
         },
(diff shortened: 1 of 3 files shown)


jonatanschroeder:
The field is used here:
https://github.com/PrairieLearn/PrairieLearn/blob/1dfd2d9be1f713d96f75d625dbc2f449e4844c69/pages/instructorCourseAdminInstances/instructorCourseAdminInstances.ejs#L69

I think the most appropriate action here is to calculate the enrollment count in `pages/instructorCourseAdminInstances/instructorCourseAdminInstances.js`, based on a page query rather than a middleware/sproc query.

jonatanschroeder (approved):

jonatanschroeder:
> The field is used here:
> https://github.com/PrairieLearn/PrairieLearn/blob/1dfd2d9be1f713d96f75d625dbc2f449e4844c69/pages/instructorCourseAdminInstances/instructorCourseAdminInstances.ejs#L69
> 
> I think the most appropriate action here is to calculate the enrollment count in `pages/instructorCourseAdminInstances/instructorCourseAdminInstances.js`, based on a page query rather than a middleware/sproc query.

Changes sent, please check if you're ok with them.

mwest1066:
> The field is used here:

Excellent catch! I thought I'd searched for usage, but clearly not carefully enough.

mwest1066:
> Changes sent, please check if you're ok with them.

Looks great, thanks! I added a comment about why we use `authz_data.course_instances` rather than running a simple DB query to find all CIs together with their enrollments.

PR #4719, opened 2021-08-28
Add authz cache for workspace proxy

mwest1066:
This PR adds a cookie-based authz cache for workspaces. After successful authz for a workspace, we give the client a cookie that lets it bypass authz for this specific workspace for the next 60 seconds.

An alternative would be to cache authz server-side, by keying a redis entry off of the `pl_authn` cookie and the `workspace_id`. I don't have a strong opinion about whether its better to do this with a cookie or a server-side cache, but I went with the cookie because it's simple and doesn't add any load to our cache.

### DIFF
--- a/middlewares/authzWorkspaceCookieCheck.js
+++ b/middlewares/authzWorkspaceCookieCheck.js
@@ -1,0 +1,24 @@
+const asyncHandler = require('express-async-handler');
+
+const csrf = require('../lib/csrf');
+const config = require('../lib/config');
+
+module.exports = asyncHandler(async (req, res, next) => {
+    // This middleware looks for a workspace_id-specific cookie and,
+    // if found, skips the rest of authn/authz. The special cookie is
+    // set by middlewares/authzWorkspaceCookieSet.js
+
+    const workspace_id = res.locals.workspace_id;
+    const cookieName = `pl_authz_workspace_${workspace_id}`;
+    if (cookieName in req.cookies) {
+        // if we have a workspace authz cookie then we try and unpack it
+        const cookieData = csrf.getCheckedData(req.cookies[cookieName], config.secretKey, {maxAge: config.workspaceAuthzCookieMaxAgeMilliseconds});
+        // if the cookie unpacking failed then cookieData will be null
+        if (!cookieData) return next();
+        // if we have a valid cookie with matching workspace_id then skip the rest of authn/authz
+        if (cookieData?.workspace_id == workspace_id) return next('router');
+    }
+
+    // otherwise we fall through and proceed to the full authn/authz stack
+    next();
+});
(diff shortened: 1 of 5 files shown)


jonatanschroeder:
I don't understand enough of the mechanics of router to give a proper review, sorry. But the code that I've seen makes sense to me.

middlewares/authzWorkspaceCookieSet.js
```diff
@@ -0,0 +1,21 @@
...
+
+module.exports = asyncHandler(async (req, res, next) => {
+    // We should only have arrived here if we passed authn/authz and
+    // are authorized to access res.locals.workspace_id. We will set a
+    // short-lived cookie specific to this workspace_id that will let
+    // us bypass authn/authz in the future. This checking is done by
+    // middlewares/authzWorkspaceCookieCheck.js
+    
+    const workspace_id = res.locals.workspace_id;
+    const cookieName = `pl_authz_workspace_${workspace_id}`;
```
nwalters512, line 14: Nice - as I started reading this PR I was thinking "what about multiple workspaces open at the same time" but you already thought of that!

middlewares/authzWorkspaceCookieSet.js
```diff
@@ -0,0 +1,21 @@
+const asyncHandler = require('express-async-handler');
+
+const csrf = require('../lib/csrf');
+const config = require('../lib/config');
+
+module.exports = asyncHandler(async (req, res, next) => {
```
nwalters512, line 6: Nit: nothing async here, so this isn't strictly necessary.

middlewares/authzWorkspaceCookieCheck.js
```diff
@@ -0,0 +1,25 @@
+const asyncHandler = require('express-async-handler');
+
+const csrf = require('../lib/csrf');
+const config = require('../lib/config');
+
+module.exports = asyncHandler(async (req, res, next) => {
```
nwalters512, line 6: Nit: nothing async here, so this isn't strictly necessary.

nwalters512 (approved):
This looks great!

middlewares/authzWorkspaceCookieSet.js
mwest1066: 😄

middlewares/authzWorkspaceCookieSet.js
mwest1066: Removed.

middlewares/authzWorkspaceCookieCheck.js
mwest1066: Removed.

nicknytko (approved):
This all looks good to me!

mwest1066:
> This all looks good to me!

Thanks!

Issue #4701, opened 2021-08-24
Implement new <pl-template> element


mwest1066:
Instructors often have blocks of "boilerplate" text that they want to include in questions. This can be done at the moment by hand-written templating in `server.py`. To make this easier, we could have an element like:
```
<pl-template file-name="my_template.html" directory="serverFilesCourse">
    <pl-variable name="zip-file" value="starter_files.zip"></pl-variable-name>
    <pl-variable name="upload-file" value="fibonacci.py"></pl-variable-name>
</pl-template>
```
This would load the file `my_template.html` from `serverFilesCourse`, which might look like:
```
<pl-question-panel>
    You should download the {{zip-file}}, unpack it, write your code, and upload the {{upload-file}}.
</pl-question-panel>
```
This file should be templated (using `chevron` in Python, which is `mustache` templates) with the given variables and inserted in place of the original `<pl-template>` element.

A few thoughts:
* We should be able to use templates recursively, so a template can include other templates. I think this will just work automatically.
* The options for the template should be similar to those for [`pl-file-download`](https://prairielearn.readthedocs.io/en/latest/elements/#pl-file-download-element) or other elements that include files.
* Should we also pass all of `data["params"]` into the template? We can always "relay" variables from `params` like this:
```
<pl-template file-name="my_template.html" directory="serverFilesCourse">
    <pl-variable name="zip-file" value="{{params.zip_file}}"></pl-variable-name>
    <pl-variable name="upload-file" value="{{params.upload_file}}"></pl-variable-name>
</pl-template>
```

mwest1066:
Until we implement this, a workaround from @jonatanschroeder is as follows:

* Put a file called `make_template.py` into `serverFilesCourse` containing:
    ```
    import chevron
    
    def render(template_name, **params):
         with open(template_name) as f:
             return chevron.render(f, params)
    ```
* In a question's `server.py` do something like this:
    ```
    import make_template
    
    def generate(data):
        data["params"]["filled_template"] = make_template.render(
            template_name="my_template.html",
            zip_file="starter.tar.gz",
            upload_file="fibonacci.py
        )
    ```
* In a question's `question.html` put:
    ```
    {{{params.filled_template}}}
    ```
* In this example, we would have a file called `my_template.html` in `serverFilesCourse` that contains:
    ```
    <pl-question-panel>
        You should download the {{zip_file}}, unpack it, write your code, and upload the {{upload_file}}.
    </pl-question-panel>
    ```

[...]

PR #4720, opened 2021-08-29
Add FAQ entry describing local templates

mwest1066:
This is a temporary workaround until #4701 is implemented.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -373,11 +373,11 @@ If you have a block of text that you want to re-use in many questions, possibly 
         with open(os.path.join(data["options"]["server_files_course_path"], template_filename)) as f:
             return chevron.render(f, params)
     ```
-2. Put a template (this example is called `code_instructions.html`) into `serverFilesCourse`:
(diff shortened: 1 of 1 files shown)


echuber2:
@mwest1066 It may be confusing that the example doesn't create a clickable link to either of the named files (they are just text labels). Would it better to expand the example to include clickable links to stuff in `clientFilesQuestion`, or to simplify the example (for example, just parameterize some integers in a math problem)?

mwest1066:
> @mwest1066 It may be confusing that the example doesn't create a clickable link to either of the named files (they are just text labels). Would it better to expand the example to include clickable links to stuff in `clientFilesQuestion`, or to simplify the example (for example, just parameterize some integers in a math problem)?

Done.

echuber2:
OK, looks good

echuber2:
Maybe we should add a note about where two vs. three `{` are needed? This seems like a footgun if the instructor is new to Mustache.

mwest1066:
> Maybe we should add a note about where two vs. three `{` are needed? This seems like a footgun if the instructor is new to Mustache.

Done.

PR #4727, opened 2021-08-31
RFC: decouple PrairieSchedule and PrairieLearn

nwalters512:
WIP!


trombonekenny:
PLpeek is another component that needs to get tracked or removed

PR #4730, opened 2021-08-31
Add GUI desktop workspace image

### LINKED ISSUE #3290: Run full graphical desktop inside workspace
Pulled from #2504:

> Make a demo question to run a full graphical desktop inside a workspace. We'd need a browser-based VNC/RDP client like https://guacamole.apache.org or https://novnc.com in a workspace container that runs a linux graphical desktop and a VNC server. This would be useful for running things like Matlab, CAD programs, etc.

nicknytko:
A workspace image based on Ubuntu 20.04 running an XFCE 4 desktop environment.  Also adds a demo question as `demo/workspace/novnc` in the example course.

image

This is set up as:
 - x11vnc running on port 5900
   - creates a virtual xorg server with Xvfb
 - Websockify acting as middleware between TCP 5900 and websocket on 5901
    - technically x11vnc can accept ws requests directly, but I've found the implementation is very unstable
 - Small express server that serves the novnc client and proxies ws requests to 5901

This image is uploaded as `prairielearn/workspace-desktop` to dockerhub.

This will resolve #3290


echuber2:
Do you have a screenshot of the "tiny screen" issue? If there's no elegant way to detect it, perhaps desktop shortcuts could be made for students to click on? 1024x768, 1920x1080, etc.? Or incremental ones like "wider", "taller", etc.? (I'm wondering what screen dimensions people might actually have in remote learning.)

nicknytko:
> Do you have a screenshot of the "tiny screen" issue? If there's no elegant way to detect it, perhaps desktop shortcuts could be made for students to click on? 1024x768, 1920x1080, etc.? Or incremental ones like "wider", "taller", etc.? (I'm wondering what screen dimensions people might actually have in remote learning.)

So by default the resolution is something like 1280x1024.  I think at the very least what we could do is detect the workspace resolution on the client, which is then passed to the server and used to boot up X.  There’s not really a great way that I can tell to change the resolution once X has been started, though.  I guess you could restart the server whenever the window is resized?

mwest1066:
Amazing! I just had a quick play with it and it's incredible that this works 😄. I'll take a more complete look later.

nicknytko:
I was talking to Geoffrey about this, I think something that we'd want to add before merging is clipboard support.  We can probably do something similar to what the full NoVNC client does which is to have a hideable textbox that you can copy+paste into and controls the vnc clipboard.  Though, it would also be cool if there is some JS api that can transparently handle clipboard reads+writes for us and would alleviate the need for a clunky textbox.

mwest1066:
Maybe we should just call this `workspaces/desktop`?

nicknytko:
Todo when talking to @mwest1066 today:

- [ ] Potentially changing the name of the image?
- [ ] Have the server resize only on boot, refreshing should not kill x session.  Students can reboot to resize the virtual screen.
- [ ] Do we want to have a news item for this?

mwest1066:
> Potentially changing the name of the image?

`workspaces/desktop`?

> Do we want to have a news item for this?

Yes!

nicknytko:
@mwest1066 created news item 035 that you're free to edit

workspaces/desktop/config/xfce4/xfconf/xfce-perchannel-xml/xfce4-desktop.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<channel name="xfce4-desktop" version="1.0">
+  <property name="backdrop" type="empty">
+    <property name="screen0" type="empty">
+      <property name="monitorscreen" type="empty">
+        <property name="workspace0" type="empty">
+          <property name="color-style" type="int" value="0"/>
+          <property name="image-style" type="int" value="5"/>
+          <property name="last-image" type="string" value="/home/prairielearner/.local/wallpaper.jpg"/>
```
mwest1066, line 10: Is this wallpaper image freely licensed?

news_items/035_desktop_workspace/index.html
```diff
@@ -0,0 +1,76 @@
...
+ .code-block div {
+     padding: 0.5rem;
+     display: block;
+     margin-bottom: -1em;
+     line-height: 125%;
+ }
+</style>
+
+<p>
+    Thanks to <a href="https://github.com/nicknytko">Nicolas Nytko</a> (CS MS '22), PrairieLearn has a new <i>desktop workspace</i> image.  This is a full Linux desktop that will allow students to run whatever graphical or terminal applications that they are provided with.  Like the existing workspaces, this desktop environment is fully persistent and is coupled with the external autograder system.
```
mwest1066, line 29: This is an excellent sentence!

> Like the existing workspaces, this desktop environment is fully persistent and is coupled with the external autograder system.

workspaces/desktop/config/xfce4/xfconf/xfce-perchannel-xml/xfce4-desktop.xml
nicknytko: It’s stolen from https://publicaffairs.illinois.edu/resources/zoom-backgrounds/, so I have no idea?  I can find a CC or similar replacement for it.

mwest1066: It's so annoying when people say "here's some content" but fail to license it. I think we should probably find a different image, sadly.

I'm actually also wondering about whether we'd be better off using a plain color background to save on bandwidth? I think VNC has to push chunks of the background to the client every time we drag a window and reveal new bits of the background, and a highly-textured image like this one won't compress well. I could easily be overthinking this, however, and maybe a nice background isn't much more expensive? I love the look of a nice background.

news_items/035_desktop_workspace/index.html
```diff
@@ -0,0 +1,76 @@
...
+</p>
+
+<p>
+    This new image is provided as <code class="user-output">prairielearn/workspace-desktop</code> and comes with a small set of pre-installed applications.  Installing extra software can be easily done by creating your own Docker image that uses this as a base.
+</p>
+
+<img class="border rounded shadow modal-image" src="workspace.png" width="550px">
```
mwest1066, line 36: If you need to regenerate this screenshot with a new background, I'm wondering whether it would be more impactful to include the full browser window, including the PL workspace header and also the browser controls themselves. Just to drive home the point that this is _inside the browser_.

mwest1066:
Oops, I mistakenly referenced this issue from #4812 -- please ignore.

nicknytko:
Updated the wallpaper and news item screenshot and pushed support for moving data to the container clipboard since I already had this working.

mwest1066:
> pushed support for moving data to the container clipboard since I already had this working.

Sweet!

mwest1066 (approved):

PR #4731, opened 2021-08-31
RStudio workspace Dockerfiles (official PR)

echuber2:
@mwest1066 @eddelbuettel


workspaces/rstudio/Dockerfile
```diff
@@ -0,0 +1,120 @@
...
+    # Cleanup:
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* && \
+    find /tmp -not -path /tmp -delete
+
+# Make sure we'll be able to run as user 1001. This is hacky.
+USER 0
+RUN mkdir -p /var/run/s6 && \
+    chown -R 1001:1001 /var/run/s6 && \
+    chown -R 1001:1001 /usr/local/lib/R && \
```
eddelbuettel, line 37: There is an old precedent here of making the (effectively single-user) user (of the multi-user) be part of (either one of) `staff` or `adm` (and I never remember which).  

If it is just one user, setting uid=gid is fine too, of course.  The main point, and it is important, is to allow user to install packages as non-root, and this ensures it.

eddelbuettel (approved):
I'll give this a 'LGTM' as it contains some stuff I am simply not familiar with (hello, nginx) and other stuff I never bothered to look into (deeper run-time customization of RStudio panes from non-standard starting directories).  I take your word for it.

workspaces/rstudio/Dockerfile
echuber2: Ah, I figured if anything more needed to be installed in the workspace, then the admins had better roll a new image (otherwise, it'll be hundreds of students reinstalling ephemeral packages every time they reload the workspace).

trombonekenny:
Can we get an example course question that uses this build included in the PR?

eddelbuettel:
Yes indeed -- that was a TODO I had for myself.  Please see here for one:   https://github.com/eddelbuettel/pl-r-demos/tree/master/questions/R_Demos/test_R_object -- and the commit I just made to enable that isolates what needs to be added.

There is one other possibly gotcha that the 'launcher' for PL (at least on Linux) needs an additional option `--add-host=host.docker.internal:172.17.0.1 ` but I would not know what macOS or other OSs require here.

eddelbuettel:
It would also be good (cough, cough, hi @mwest1066 ) to get this merged and/or to get the container pushed because local use still adds noise:

```
info: Pulling docker image: prairielearn/workspace-rstudio
error: Error pulling "prairielearn/workspace-rstudio" image; attempting to fall back to cached version. (HTTP code 404) unexpected - pull access denied for prairielearn/workspace-rstudio, repository does not exist or may require 'docker login': denied: requested access to the resource is denied  {"statusCode":404,"json":null,"stack":"Error: (HTTP code 404) unexpected - pull access denied for prairielearn/workspace-rstudio, repository does not exist or may require 'docker login': denied: requested access to the resource is denied \n    at /PrairieLearn/node_modules/docker-modem/lib/modem.js:301:17\n    at IncomingMessage.<anonymous> (/PrairieLearn/node_modules/docker-modem/lib/modem.js:328:9)\n    at IncomingMessage.emit (events.js:412:35)\n    at IncomingMessage.emit (domain.js:470:12)\n    at endReadableNT (internal/streams/readable.js:1317:12)\n    at processTicksAndRejections (internal/process/task_queues.js:82:21)"}
error: Error proxying workspace request: Error: socket hang up {"err":{"code":"ECONNRESET"},"url":"/pl/workspace/1/container/events/get_events"}

```

eddelbuettel:
Maybe this will do (untested):

```diff
modified   workspaces/rstudio/Dockerfile
@@ -68,7 +68,8 @@ EXPOSE 3939
 
 USER rstudio
 ENV EDITOR_FOCUS_DIR "/home/rstudio/workspace"
-RUN mkdir -p "$EDITOR_FOCUS_DIR"
+RUN mkdir -p "$EDITOR_FOCUS_DIR" && \
+    chown rstudio:rstudio "$EDITOR_FOCUS_DIR"
 
 # The .Rprofile hack will set the working directory of the left code panes but
 # not the right file browser pane. The rsession.conf hack below will set both.
```

echuber2:
I'll look into it.

eddelbuettel:
It's PL. 

If I run the container 'as is' the directory is owned the right way.  So ... something happens when we mount or copy from PL?

echuber2:
This is an issue caused by the fact that by default, local docker PL runs workspaces as root instead of 1001. @mwest1066 @nicknytko we were discussing this in relation to the OCaml workspace too. I can probably use gosu to implement the same fix here. In the long run, it might be worth defaulting the local PL to run as 1001:1001.

I can confirm that adjusting the local config to run as 1001:1001 fixes the issue too.

echuber2:
I mean in config.json. I bake this into my version of PrairieLearn locally.

eddelbuettel:
Hm so how could I make this modification without having to rebuild pl/pl ?

echuber2:
> Hm so how could I make this modification without having to rebuild pl/pl ?

@eddelbuettel Um, I'm not sure there's a convenient way. I suggest trying the latest files I just added to this PR. It should work nicely now.

eddelbuettel:
Nicely done -- works as prescribed:

image

echuber2:
OK, so I guess what's missing yet is an example question and maybe some documentation.

eddelbuettel:
Example questions (I can use plural, there are two, one using this) are here:  https://github.com/eddelbuettel/pl-r-demos

mwest1066 (approved):

PR #4801, opened 2021-09-07
Add units input element

Cadenze:
- Added input element that allows input and grading of a number and/or unit
- Added units.py, a module that parses units as two numbers, representing multiplier (how much of a quantity) and dimension (what quantity)
- Allows partial marks for correct unit (i.e. correct dimension), but incorrect answer


Cadenze:

Cadenze:
recheck

eliotwrobson:
@Cadenze @nwalters512 is this still being actively worked on? I think this has some cool features that could be useful in our course (as well as in general). If not, I'd be happy to try and push this through when I have some spare time (as this is very feature complete and the code looks really good).

Cadenze:
I don't think there's any work done on this since 2021, but from what I remember there shouldn't be too many loose ends. If there is something you'd like to know/fix feel free to just message me.

eliotwrobson:
Sounds good. I think after one of my other PRs gets merged, I can take a pass over this and simplify some of the code and add type hints. My main concern is whether people will view this as too similar to the existing number-input element, but overall it seems different enough that it's worth using a separate element (instead of folding all of the features into that one).

nwalters512:
#6790 is close to being merged; I'm going to close this in favor of that PR. Thanks for the initial work here @Cadenze!

PR #4803, opened 2021-09-09
Add access control via pt_reservations

mwest1066:


migrations/240_pt_tables__create.sql
```diff
@@ -0,0 +1,17 @@
...
+    user_id BIGINT NOT NULL REFERENCES users
+);
+
+CREATE TABLE IF NOT EXISTS pt_reservations (
+    id bigserial PRIMARY KEY,
+    exam_id BIGINT NOT NULL REFERENCES pt_exams,
+    enrollment_id BIGINT NOT NULL REFERENCES pt_enrollments,
+    access_start TIMESTAMPTZ,
+    access_end TIMESTAMPTZ
+);
```
nwalters512, line 17: I believe that this migration and the "same" migration in PT will interact poorly if we deploy PL to a database and later try to deploy PT on top of it. I understand this is not a concern at the moment, but something to keep in mind!

See https://github.com/PrairieLearn/PrairieSchedule/blob/7c5ccd51d0a1c8fcabe38745fba564443bd0fefb/migrations/000_initial_state.sql#L81 for an example of how PS deals with this.

I have vague thoughts about changing the way we do migrations to avoid this problem entirely, inspired by [this GitHub blog post](https://github.blog/2020-02-14-automating-mysql-schema-migrations-with-github-actions-and-more/), but they are very much just thoughts at this point 😅

sprocs/check_assessment_access_rule.sql
```diff
@@ -74,12 +74,30 @@ BEGIN
...
+            -- is there a checked-in pt_reservation?
+            PERFORM 1
+            FROM
+                pt_reservations AS r
+                JOIN pt_enrollments AS e ON (e.id = r.enrollment_id)
+                JOIN pt_exams AS x ON (x.id = r.exam_id)
+            WHERE
+                (date BETWEEN r.access_start AND r.access_end)
+                AND e.user_id = check_assessment_access_rule.user_id
+                AND x.uuid = assessment_access_rule.exam_uuid;
```
nwalters512, line 91: Do we want the same "pick at most one, and pick the longest-lasting one if there are multiple" logic that we use for PS?

sprocs/ip_to_mode.sql
```diff
@@ -33,5 +33,19 @@ BEGIN
...
     END IF;
+
+    -- Do we have an active PT reservation? If so, force mode to 'Exam'
+    PERFORM 1
+    FROM
+        pt_reservations AS r
+        JOIN pt_enrollments AS e ON (e.id = r.enrollment_id)
+    WHERE
+        e.user_id = ip_to_mode.authn_user_id
+        AND (date BETWEEN r.access_start AND r.access_end);
```
nwalters512, line 44: Asking out of complete ignorance of how PT functions: do we need to check for deletion/"checked in" status, null dates, etc.? Basically, any of the things that we do for PS.

nwalters512 (approved):

migrations/240_pt_tables__create.sql
mwest1066: Ha! I was also thinking about this exact same problem! I have a much hackier solution in mind than what GitHub did 😄. But as you say, we can figure this out later.

sprocs/check_assessment_access_rule.sql
mwest1066: From what I can see, that logic isn't really doing anything. It selects into the block-local variable `reservation`, but we never refer to this variable again. So I can't see that it matters which `reservation` we pick, or even if we pick one at all. This is why I just used `PERFORM 1`. Does this also make sense to you?

sprocs/ip_to_mode.sql
mwest1066: Nope, it's strictly checking date between start/end times. Both of these times will be NULL if we are not checked in, so the BETWEEN evaluates to NULL which is not TRUE. When we check out, we set `access_end = now()` so any future checks will fail the BETWEEN check. We don't have a separate `checked_in` flag.

Note that BETWEEN cares about order, so "6 between 10 and 2" is FALSE. Doing "a BETWEEN b AND c" is just checking "b < a" and "a < c". See https://www.postgresql.org/docs/current/functions-comparison.html

sprocs/check_assessment_access_rule.sql
nwalters512: Ah! That makes perfect sense. As long as _some_ reservation is active at this point, that's all we care about.

sprocs/ip_to_mode.sql
nwalters512: Gotcha! The way `BETWEEN` works with `NULL` seems very clever but does make intuitive sense. This all looks good to me!

sprocs/check_assessment_access_rule.sql
mwest1066: Exactly!

sprocs/ip_to_mode.sql
mwest1066: Yeah, it's maybe a bit too magic that BETWEEN handles all the NULL cases, but it does make the check very short and easy to write 😄 

Thanks for all the checking!

mwest1066:
Just a note that this PR should not be merged to `master` just yet. I need to handle deploy of this a bit carefully. I converted the PR to Draft to prevent accidental merges.

jonatanschroeder (commented):
I won't officially approve it, since there are parts of the code that use a logic I'm not familiar with re: PrairieSchedule, but everything looks ok. I have a similar concern as Nathan about the migration with tables not matching exactly, but I'll defer to your judgement on this.

mwest1066:
> I won't officially approve it, since there are parts of the code that use a logic I'm not familiar with re: PrairieSchedule, but everything looks ok. I have a similar concern as Nathan about the migration with tables not matching exactly, but I'll defer to your judgement on this.

Great, thanks for the checking!

The migration issue is definitely a problem, but we can solve that next week 😄

PR #4805, opened 2021-09-09
Add alt-text attribute for <pl-figure>

glherman:
Added an alt text field for the pl-figure element for accessiblity reasons. Updated the element *.py files and documentations.

Couldn't use alt or alt-text as a variable name in Python because alt is a reserved word, so apologies on the inconsistent naming of the variables.

Chose "PrairieLearn Figure" as the default text but would happily change if there is a better choice here. Could also go with an empty string.

### DIFF
--- a/docs/elements.md
+++ b/docs/elements.md
@@ -1194,6 +1194,7 @@ Attribute | Type | Default | Description
 `directory` | text | "clientFilesQuestion" | The directory that contains the file, either 'clientFilesQuestion' or 'clientFilesCourse' (see [client and server files](clientServerFiles.md)). A directory cannot be specified if `type='dynamic'`.
 `width` | number | `None` | Width of image (e.g., '250px').
 `inline` | boolean | false | Display figure inline with text (true) or on a separate line (false).
+`alt-text` | text | "PrairieLearn Figure" | Provide alt (alternative) text to improve accessibility of figures by describing the image or the purpose of the image.
 
 #### Dynamically generated figures
 
--- a/docs/elements.md
+++ b/docs/elements.md
@@ -1194,7 +1194,7 @@ Attribute | Type | Default | Description
 `directory` | text | "clientFilesQuestion" | The directory that contains the file, either 'clientFilesQuestion' or 'clientFilesCourse' (see [client and server files](clientServerFiles.md)). A directory cannot be specified if `type='dynamic'`.
 `width` | number | `None` | Width of image (e.g., '250px').
 `inline` | boolean | false | Display figure inline with text (true) or on a separate line (false).
-`alt-text` | text | "PrairieLearn Figure" | Provide alt (alternative) text to improve accessibility of figures by describing the image or the purpose of the image.
+`alt` | text | "" | Provide alt (alternative) text to improve accessibility of figures by describing the image or the purpose of the image. Default is an empty string.
 
 #### Dynamically generated figures
 
--- a/elements/pl-figure/pl-figure.py
+++ b/elements/pl-figure/pl-figure.py
@@ -8,7 +8,7 @@ WIDTH_DEFAULT = None
 TYPE_DEFAULT = 'static'
 DIRECTORY_DEFAULT = 'clientFilesQuestion'
 INLINE_DEFAULT = False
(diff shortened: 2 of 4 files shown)


glherman:

nicknytko (approved):
This all looks reasonable to me!

elements/pl-figure/pl-figure.py
```diff
@@ -8,11 +8,12 @@
 TYPE_DEFAULT = 'static'
 DIRECTORY_DEFAULT = 'clientFilesQuestion'
 INLINE_DEFAULT = False
+ALT_TEXT_DEFAULT = 'PrairieLearn Figure'
```
tdy, line 11:
```suggestion
ALT_TEXT_DEFAULT = 'PrairieLearn logo'
```
I suggest being as specific as possible. It's generally recommended to avoid generic terms like "image" and "figure" in alt text.

docs/elements.md
```diff
@@ -1194,6 +1194,7 @@ Attribute | Type | Default | Description
 `directory` | text | "clientFilesQuestion" | The directory that contains the file, either 'clientFilesQuestion' or 'clientFilesCourse' (see [client and server files](clientServerFiles.md)). A directory cannot be specified if `type='dynamic'`.
 `width` | number | `None` | Width of image (e.g., '250px').
 `inline` | boolean | false | Display figure inline with text (true) or on a separate line (false).
+`alt-text` | text | "PrairieLearn Figure" | Provide alt (alternative) text to improve accessibility of figures by describing the image or the purpose of the image.
```
tdy, line 1197:
```suggestion
`alt-text` | text | "PrairieLearn Figure" | Provide alt (alternative) text to improve accessibility of figures by describing the image or the purpose of the image. For basic guidelines on creating good alt text, see [WebAIM's article on alternative text](https://webaim.org/techniques/alttext/).
```
Maybe it's worth linking to some basic alt text guidelines? Just a thought.

elements/pl-figure/pl-figure.py
mwest1066: I thinking that we should default to empty alt text, assuming that the image is described by the surrounding text as is common in many questions. My understanding is that when the image is described by the surrounding text, we should not specify alt text?

docs/elements.md
mwest1066: Can you update the example course questions with appropriate alt texts? At least `exampleCourse/questions/element/figure`.

elements/pl-figure/pl-figure.py
glherman: Would you prefer for the alt text field in the img tog to be `alt=""` or would you prefer for there just to be no alt field whatsoever?

mwest1066: According to the link that @tdy posted, all images should have an `alt` attribute and we are supposed so specify `alt=""` if it should be empty. So let's always add it and just set `alt=""`.

glherman: done

docs/elements.md
glherman: add alt text to all figures in `exampleCourse/questions/element/figure`

nwalters512: > Couldn't use alt or alt-text as a variable name in Python because alt is a reserved word, so apologies on the inconsistent naming of the variables.

I understand this, but I think `alt` would be a better name for the attribute so that it matches the underlying web platform. I don't think that would cause an issue with Python reserved words.

nwalters512: Also, I can't find any documentation about `alt` being a reserved word - could you point me to that?

mwest1066: Good point. I agree that we should use `alt` for the attribute name, but `alt_text` for the Python variable.

mwest1066: Also, I think the default value needs to be changed to `""` in the docs above.

glherman: hmmm...I'm not getting the error with alt in Python that I was getting earlier. It refused to work for me to until I changed alt to alt_text. It's behaving now with alt or alt_text. Leaving html/mustache as alt and leaving Python as alt_text for now. Updated the examples in `exampleCourse/questions/element/figure`

mwest1066 (approved):

PR #4810, opened 2021-09-11
Extra Docker install notes for MacOS

maxieds:
Moving some helpful documentation for MacOS from PR #4334:

### Alternate solution to long development build times

To replace the core `python` files one modifies in development there is another work around. Most likely, the developer is posting commits onto a fork of the `PrairieLearn` repository on GitHub and expects to eventually submit a pull request for the active branch they use locally. Suppose that your local changes to the development code are located in the directory `~/PrairieLearn`.

To build a `docker` image from the fork with the source for a particular branch, run the following commands first (this initial step will take a while to complete): 
```bash
$ docker build --no-cache https://github.com/TheGHUserName/PrairieLearn.git#exact-branch-name
$ export PLDIR_MAPS="-v $(greadlink -f ~/PrairieLearn/docs):/PrairieLearn/docs \
      -v $(greadlink -f ~/PrairieLearn/elements):/PrairieLearn/elements \
      -v $(greadlink -f ~/PrairieLearn/exampleCourse):/PrairieLearn/exampleCourse \
      -v $(greadlink -f ~/PrairieLearn/lib):/PrairieLearn/lib" 
$ docker images
$ docker run -it --rm -p 3000:3000 $PLDIR_MAPS MOST-RECENT-IMAGE-NAME-FROM-ABOVE
```
Then you can connect to the running instance of `PrairieLearn` through the web browser method listed above. Note that `greadlink` is available on MacOS by running `brew install coreutils`. On Linux and other Unix machines, a standard `readlink -f path` will work in the `export` commands provided in the last terminal listings.

### DIFF
--- a/docs/installingLocal.md
+++ b/docs/installingLocal.md
@@ -50,6 +50,22 @@ docker pull prairielearn/prairielearn:BRANCH_NAME
 docker run -it --rm -p 3000:3000 prairielearn/prairielearn:BRANCH_NAME
 ```
 
+### Alternate solution to long development build times (build from scratch from a specific branch)
+
+To replace the core `python` files one modifies in development there is another work around. Most likely, the developer is posting commits onto a fork of the `PrairieLearn` repository on GitHub and expects to eventually submit a pull request for the active branch they use locally. Suppose that your local changes to the development code are located in the directory `~/PrairieLearn`.
+
+To build a `docker` image from the fork with the source for a particular branch, run the following commands first (this initial step will take a while to complete): 
+```bash
+$ docker build --no-cache https://github.com/TheGHUserName/PrairieLearn.git#exact-branch-name
+$ export PLDIR_MAPS="-v $(greadlink -f ~/PrairieLearn/docs):/PrairieLearn/docs \
+      -v $(greadlink -f ~/PrairieLearn/elements):/PrairieLearn/elements \
+      -v $(greadlink -f ~/PrairieLearn/exampleCourse):/PrairieLearn/exampleCourse \
+      -v $(greadlink -f ~/PrairieLearn/lib):/PrairieLearn/lib" 
+$ docker images
+$ docker run -it --rm -p 3000:3000 $PLDIR_MAPS MOST-RECENT-IMAGE-NAME-FROM-ABOVE
+```
+Then you can connect to the running instance of `PrairieLearn` through the web browser method listed above. Note that `greadlink` is available on MacOS by running `brew install coreutils`. On Linux and other Unix machines, a standard `readlink -f path` will work in the `export` commands provided in the last terminal listings. 
+
 ### Running commands in Docker
 
 If needed, you can run the container with a different command:


echuber2:
Is the idea here that you just want to bind-mount the local working directories _instead of_ rebuilding every time, to see the changes? I think there's already a note about this in the "installingLocal" page under the dev guide section:

https://prairielearn.readthedocs.io/en/latest/installingLocal/

maxieds:
@echuber2 @mwest1066 
The installing local docs are good. Instructions on howto build the initial image (into which we bind-mount the local changes) from a **fork** of the PL repo are still missing. This is the way that pull requests are developed and contributed, e.g., the following line is still relevant and missing from current docs:
```bash
$ docker build --no-cache https://github.com/TheGHUserName/PrairieLearn.git#exact-branch-name
```

echuber2:
I think it would be better to propose a revision for the `installingLocal` page (part of the dev guide) in this case, rather than the `installing` page that is more general and consumer-oriented.

echuber2:
Can you clarify a bit more about a few things: why are you bind-mounting several PL subdirectories separately instead of the entire PL directory at once? (That's what the original installingLocal docs suggest.) Why use no-cache instead of allowing cache, which might slightly increase the build? Why is it essential to teach the `git` detail about identifying a branch directly in the build command, here? (Instead, one could checkout the specific branch, and do `git build -t whatever -f whatever .` within the directory, etc.

Since this is the official documentation, it comes with an implicit recommendation to new contributors about their workflow. So, not everyone's personalized tricks or shortcut scripts are added. (I've had suggestions rejected or heavily edited before, myself.) Some of those things go into the `tools` subdirectory as contributor scripts instead.

maxieds:
@echuber2 
I am still using the instructions in the branch associated with this PR on my Mac.  I guess the instructions I added here were a result of inexperience with ``docker`` when I first started using it a few months back. The results of scouring the ``docker`` documentation online produced these commands. For some reason, the initial install local instructions to build the PL ``docker`` image did not work on my MacOS desktop (10.14, partially upgraded and patched to 10.15). Have we checked this recently on Mac, or even better with the new BigSur releases (I am curious). I only bind mounted a subset of the PL contents because that happens to be the folders I was actively editing while doing testing. It should be clear enough from context that more (less) can be mounted that way.

The use case I had after forking the main PL repository is what the appended notes cover. I just want to reiterate the question, which is does adding these extra notes really hurt? It would be nice to stash them somewhere that I can find them periodically when I float back to the PL development projects. If you have concerns that it will confuse users, maybe we should just add a note to the effect of "The following commands are NOT recommended ... " (or however you would want that caveat to read.

nwalters512:
>  For some reason, the initial install local instructions to build the PL docker image did not work on my MacOS desktop (10.14, partially upgraded and patched to 10.15). Have we checked this recently on Mac, or even better with the new BigSur releases (I am curious)?

Yep - they've continued to work for at least myself all the way through Monterey, the current latest version. Do you by chance have a M1 Mac? That's the only scenario where things might be a little wonky now, and I don't think your instructions would change that. If you're having issues, we'd appreciate an issue detailing exactly what problems you're encountering!

> does adding these extra notes really hurt?

"Hurt" is maybe too strong of a word, but it does create additional maintenance burden (we need to maintain these docs) and additional text for users to try to comb through as they navigate our already-complex development docs.

Before I'd consider merging this, I'd want to see the docs updated to explain what problem(s) these new docs solve, i.e. why I as a new PrairieLearn developer would choose them over other docs on the page.

PR #4815, opened 2021-09-14
Fix pt_reservations linking

mwest1066:
The key fix is to also check `exam_uuids` against `pt_exams` (not just `ps_exams`) in `sync/fromDisk/assessments.sql`. Without this, we would record a "UUID not found" error when we linked to an exam in `pt_exams`, which would then cause PL to not load the JSON file at all and so the `exam_uuid` would not make it into the DB.

Also add `pt_exams` links on Access page.


lib/config.js
```diff
@@ -131,6 +131,7 @@ config.shibLinkColors = {
 config.hasAzure = false;
 config.hasOauth = false;
 config.syncExamIdAccessRules = false;
+config.ptHost = 'http://localhost';
```
jonatanschroeder, line 134:
```suggestion
config.ptHost = 'http://localhost:4000';
```
If this is supposed to be the default for local environment tests, this is more likely to be the correct option.

jonatanschroeder (approved):

PR #4819, opened 2021-09-16
Add fixed-options-order to pl-matching

### LINKED ISSUE #4817: pl-matching: fixed order of options
It would be good to have a way to fix the order of the options while still randomizing the order of the statements. I have a question where the matching options are network layers, and having the layers show in the traditional order may be helpful in that case.
@nealterrell

### LINKED ISSUE #4822: pl-matching docs
Documentation for pl-matching includes an image that doesn't match the example:
Docs

Reported by @mfsilva22 
@nealterrell if you have time? You mentioned a patch in pl-matching, can you add this?

### LINKED ISSUE #4831: pl-matching counters don't match the options in answer panel
When submitting a pl-matching answer using counter-type "upper-alpha", the answer panel shows incorrect option counters with correct option text. In essence, the text of the answer shows the correct option, but the counter doesn't match the test (in my case the answer always showed A/B/C/D/E regardless of the correct answer). The same thing happens with decimal or lower-alpha, but full-text works fine.
@nealterrell

nealterrell:
Fixes #4817. I find this especially helpful when using full-text counters. 

Also fixes #4822 and fixes #4831.


docs/elements.md
```diff
@@ -675,7 +675,8 @@ Attribute | Type | Default | Description
 --- | --- | --- | ---
 `answers-name` | string | — | Variable name to store data in.
 `weight` | integer | 1 | Weight to use when computing a weighted average score over elements.
-`fixed-order` | boolean | False | Whether or not to display the statements in a fixed order; otherwise they are shuffled. Options are always shuffled.
+`fixed-order` | boolean | False | Whether or not to display the statements in a fixed order; otherwise they are shuffled.
+`fixed-options-order` | boolean | False | Whether or not to display the options in a fixed order; otherwise they are shuffled.
```
jonatanschroeder, line 679: It may be a good idea to include (either here or around line 697) in which order the options are printed if there are implicit options created based on statement matches (maybe encourage instructors to use explicit `pl-option` entries for all options to ensure the order is as expected).

jonatanschroeder (commented):
The code looks fine, but I haven't had a chance to test it yet. It may be a good idea to also add an example in the exampleCourse sample question.

nealterrell:
Added fixes for the two reported pl-matching errors.

elements/pl-matching/pl-matching.py
```diff
@@ -157,17 +158,25 @@ def prepare(element_html, data):
             # Add a sample of the distractors.
             needed_options.extend(random.sample(distractors, more_needed))
```
jonatanschroeder, line 159: There is a big problem here for fixed order, since the distractors are always added at the end (same in line 153). This not only causes the order to deviate from documentation, but also gives a huge hint to students that know the behaviour, since they can focus on the top N options as correct when there are more options than statements, and assume the distractors are at the end.

jonatanschroeder:
```suggestion
            distractor_sample = random.sample(distractors, more_needed)
            needed_options_keys = needed_options_keys.union(set(o['name'] for o in distractor_sample))
            needed_options = [o for o in options if o['name'] in needed_options_keys]
```

for line 153, `needed_options = options` probably resolves the issue.

jonatanschroeder: @nealterrell I have committed these changes, if you have a chance to test that would be great (I'd prefer a second set of eyes in these changes). Otherwise I tagged @mwest1066 and @nwalters512 to review it.

echuber2: I thought this might have been marked resolved by mistake so I unresolved it for visibility, but sorry if I was wrong to do that. Feel free to mark resolved again.

elements/pl-matching/pl-matching.py
```diff
@@ -149,25 +150,35 @@ def prepare(element_html, data):
         more_needed = number_options - len(needed_options)
         if more_needed >= len(distractors):
             # Add all distractors.
```
nwalters512, line 152: The original comment made more sense to me with the original implementation; hopefully this updated comment is more clear?
```suggestion
            # Use all options, both correct and distractors.
```

nwalters512 (approved):
I once again find myself wishing we had an established way of unit testing elements 😅

Issue #4652, opened 2021-08-09
pl-order-blocks/ranking grading algorithm and feedback


jeffgerickson:
The `ranking` grading algorithm for `pl-order-blocks` questions removes submitted items that should be excluded, and then counts the number of indices i such that either

- i=1 and ranking(i)==1, or
- i>1 and (ranking(i)==ranking(i-1) or ranking(i)==ranking(i-1)+1)

and divides by the correct number of items.  This grading algorithm is rather opaque, especially because there is currently no support for feedback along with the grade.

I'd like to suggest three alternatives; I can split these into separate issues if necessary.

### Non-consecutive ranks

This requires the least change from the current algorithm. The grading algorithm removes submitted items that should be excluded, and then counts the number of indices i where either 

- i=1 and ranking(i) is the smallest ranking in every correct solution
- i>1 and ranking(i) ≥ ranking(i-1)

and divides by the correct number of items.  The only difference is that rankings in the question do not have to be consecutive. 

There should be a corresponding `feedback` option that shows which items are out of order.

### Longest correct prefix

The grade is the length of the longest correctly ordered prefix of the correct items;.  This should be used with `feedback="first-wrong"`, which is currently not supported.  (I'm starting from the assumption that all grades should include feedback, that is consistent with the grading algorithm.)

### Longest non-decreasing subsequence

The grade is the length of the longest non-decreasing subsequence of the submitted rankings divided by the number of items.  As a graded solution, the student sees their submitted list, with items in the longest non-decreasing subsequence shaded green, and other items shaded red.  ("_Items in red should be ranked either above some earlier item, or below some later item._")  Rankings supplied in the question do not need to be consecutive.  I'd recommend requiring **all** items to be marked `correct="true"`; otherwise, there should be an additional penalty for including incorrect items.

I would use this grading method in this question:
https://www.prairielearn.org/pl/course_instance/128710/instructor/question/9063329/preview

SethPoulsen:
These all seem like sensible grading methods to me. Some other questions we should probably ask as well like--how transparent should we be to students about which grading method we are using? should the help-text we show for students tell them which grading method is being used to assign them partial credit? I would think probably yes.

I agree having a default feedback that corresponds to each grading mode seems a good idea.

SethPoulsen:
I'm going to tag some other people here who I know are using the element so we can make sure we implement something that works well for everyone: @Frozemint @jonatanschroeder @cheeren @ygertner @jasonxia17 

[...]

PR #4820, opened 2021-09-16
Fix exception in ordered grading of order blocks.

### LINKED ISSUE #4792: pl-order-blocks exception with ordered grading and indentation
I have this very simple ordered blocks question:

```html
<pl-order-blocks answers-name="answer" grading-method="ordered" indentation="true">
    <pl-answer correct="true" indent="0">Place this block first.</pl-answer>
    <pl-answer correct="true" indent="0">Place this block second.</pl-answer>
    <pl-answer correct="true" indent="1">Place this block third. Indent it once.</pl-answer>
    <pl-answer correct="false" indent="0">Unused block</pl-answer>
    <pl-answer correct="false" indent="0">Incorrect block</pl-answer>
</pl-order-blocks>
```

It grades correctly if the student only uses up to three blocks in their answer. However, if they use more than 3 blocks, we get the following error:

```
  File "/PL/CODE-pldev/lib/python-caller-trampoline.py", line 175, in <module>
    worker_loop()
  File "/PL/CODE-pldev/lib/python-caller-trampoline.py", line 121, in worker_loop
    val = method(*args)
  File "/PL/DATA-pldev/courses/csulb-cecs174/elements/pl-order-blocks/pl-order-blocks.py", line 569, in grade
    if true_answer_indent[i] == "-1" or int(indent) == true_answer_indent[i]:
IndexError: list index out of range
  File "/PL/CODE-pldev/lib/python-caller-trampoline.py", line 187, in <module>
```

(The path shows my local repository instead of the PL source, because I copied the element's code from PL to my course due to the Berkeley server being out of date.) 

I believe the issue is that `true_answer_indent` is built only from the `correct="true"` blocks, so it has 3 entries; however, the statement throwing the error is in a loop over the list `student_answer`, which in the error case has 4 entries.

nealterrell:
Fixes #4792.

I believe this fix is sufficient, but a second pair of eyes would be great.

The justification: the error occurs when "ordered" grading is used (student must put specific blocks in one specific order; every other arrangement is scored a 0), and the student uses a larger number of blocks than required in the official solution. The loop on line 417 enumerates over the student's selected blocks, but uses these indices to compare against the solution blocks; therefore giving out-of-bounds exceptions if there are more student blocks than solution. There seems to be an assumption on this line that the two lists have equivalent blocks at equivalent indices, i.e., that the student got the question correct and now indentation needs to be checked. This assumption is violated if the student didn't get the question right, for example by using too many blocks.

Such a submission is graded a 0 by the existing code. We can avoid checking indentation at all by requiring a non-zero final_score before doing so. Since the indentation score is ultimately multiplied by the final_score, it is pointless to check indentation on a submission with a final_score of 0.

This should suffice until a meaningful way of giving partial credit in this grading method is achieved, as discussed in #4652 .

### DIFF
--- a/elements/pl-order-blocks/pl-order-blocks.py
+++ b/elements/pl-order-blocks/pl-order-blocks.py
@@ -409,7 +409,7 @@ def grade(element_html, data):
                 else:
                     feedback = DAG_FIRST_WRONG_FEEDBACK['wrong-at-block'].format(str(first_wrong + 1))
 
-    if check_indentation:
+    if check_indentation and final_score > 0:
         student_answer_indent = filter_multiple_from_array(data['submitted_answers'][answer_name], ['indent'])
         student_answer_indent = list(map(lambda x: x['indent'], student_answer_indent))
         true_answer_indent = filter_multiple_from_array(data['correct_answers'][answer_name], ['indent'])


Frozemint (approved):
Good fix, it makes no sense to check indentation if the answer the student gave is deemed to be completely wrong. Removing the unneeded computation here is a really neat and elegant fix for all the unwanted side effect that comes with performing unnecessary computation.

Issue #3540, opened 2020-12-13
MathJax equations sometimes have parts missing on Safari


mwest1066:
The `demo/calculation` example question does this on Safari (works on Chrome and Firefox):

<img width="832" alt="safari_mathjax" src="https://user-images.githubusercontent.com/875833/102022881-4d171080-3d4f-11eb-8f4f-e66d5efb8507.png">

mwest1066:
Maybe this is related to mathjax/MathJax#2435

mwest1066:
As a workaround we could use SVG output for Safari. Unfortunately we can't switch to SVG for all browsers because of #3520

PR #4824, opened 2021-09-17
Use SVG mode for MathJax when Linux detected (all browsers)

echuber2:
We have another issue with horizontal lines disappearing on Linux browsers (Firefox, Chrome-alikes both). Switching to SVG MathJax seems to resolve it.


echuber2:
For reference:

SVG mode is also enabled for Safari for similar reasons. https://github.com/PrairieLearn/PrairieLearn/issues/3540

SVG mode is only used selectively on Windows+Chrome because of issues it has. https://github.com/PrairieLearn/PrairieLearn/pull/3520

More testing would be required, but I have a theory that SVG mode should work everywhere _except_ Windows+Chrome as noted above. So, we may need to add even more such opt-ins in the future for SVG mode.

nwalters512 (approved):

PR #4830, opened 2021-09-19
Error handling in pl-matching if params is not set

jonatanschroeder:
I had problems with this for a student that opened a question that had been released by mistake before it was properly set, but this can also be a problem if a question is changed after a submission had been made (bad practice, but should have a cleaner solution). Current version creates issues every time the question is opened. This should remove that problem

### DIFF
--- a/elements/pl-matching/pl-matching.py
+++ b/elements/pl-matching/pl-matching.py
@@ -226,7 +226,7 @@ def parse(element_html, data):
 def render(element_html, data):
     element = lxml.html.fragment_fromstring(element_html)
     name = pl.get_string_attrib(element, 'answers-name')
-    display_statements, display_options = data['params'].get(name)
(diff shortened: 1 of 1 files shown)


jonatanschroeder:
@nealterrell I can't add you as reviewer because you don't have the permissions, but can you check if this makes sense?

nealterrell:
So an empty pl-matching element? I can see how this would cause an error. Your fix looks correct to me; it should cause nothing to be rendered, instead of throwing an error.

jonatanschroeder:
This error should not happen in normal cases, it's mostly in cases where either the question changes or the version of pl-matching changes from when the students first opened the question to when they see it later. IOW, it's a sanity check. For my case in particular, the student opened the assessment before pl-matching had been deployed, so there were no data["params"] object linked to the pl-matching element.

PR #4834, opened 2021-09-21
adding SQLite3 to RStudio workspace

barbehenna:
`sqlite3` is a command line tool that UIUC STAT 447 uses, others may find it useful as well.


eddelbuettel:
Context is that we have a set of homework, then quiz a week later, questions with SQL content coming up... (and `sqlite3` is in the RStudio Cloud instance students can access if they purchase the token for the term etc).

mwest1066 (approved):

PR #4835, opened 2021-09-21
Fix handling of non-ASCII characters in pl-rich-text-editor

jonatanschroeder:
Non-ASCII Unicode characters (accents, non-English alphabets, emojis, etc.) are properly encoded when submitted via `pl-rich-text-editor`, but the resulting text is not decoded properly when the result is shown again after submission. This should fix the issue (https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem).


elements/pl-rich-text-editor/pl-rich-text-editor.js
```diff
@@ -27,6 +27,9 @@ window.PLRTE = function(uuid, options) {
     quill.setContents(quill.clipboard.convert(contents));
     
     quill.on('text-change', function(_delta, _oldDelta, _source) {
-        inputElement.val(btoa(unescape(encodeURIComponent(quill.root.innerHTML))));
+        inputElement.val(btoa(he.encode(quill.root.innerHTML, {
+            allowUnsafeSymbols: true, // HTML tags should be kept
```
mwest1066, line 31: This sounds bad, but I assume it's actually ok?

jonatanschroeder: This means characters like `<`, `>`, `"`, etc. are not converted, which is what I actually want in this case.

jonatanschroeder: https://github.com/mathiasbynens/he#allowunsafesymbols

> The default value for the allowUnsafeSymbols option is false. This means that characters that are unsafe for use in HTML content (&, <, >, ", ', and `) will be encoded. When set to true, only non-ASCII characters will be encoded.

mwest1066: Ok, cool, thanks for explanation!

mwest1066 (approved):

PR #4840, opened 2021-09-23
desktop workspace: improvements, low-latency tweaks

echuber2:
I made a smaller copy of the (big) wallpaper file, although it doesn't help much.

Turned off vblank and compositing for xfce, although these may already have been happening by default with quiet failures. It seemed to slightly improve things though.

Not in this PR: Replacing the xfce window manager with icewm seems to improve latency of key inputs somewhat. Also icewm sessions start up instantly compared to xfce. Maybe we could try other things than xfce; icewm is rough around the edges, but maybe Enlightenment or Mate would work well too.

I wasn't able to replicate a Visual Studio Code crash locally or on the server, so no idea what helps there yet.


nicknytko (approved):
I'd be very happy to just not have the HD copy of the wallpaper here, otherwise this looks fine to me

echuber2:
I added the low-latency tweaks and made a few more tweaks to the apt-get sequence while I had the opportunity. (Please see commit message)

PR #4845, opened 2021-09-26
Add config for DB pool initialization

mwest1066:


jonatanschroeder (approved):
I haven't tested because I don't have an environment where these would make a difference, but it makes sense to me...

mwest1066:
> I haven't tested because I don't have an environment where these would make a difference, but it makes sense to me...

Thanks, getting another set of eyes on this (even though it's trivial) is most appreciated.

PR #4847, opened 2021-09-27
Add note about useNewQuestionRenderer in course config doc

echuber2:
I think this explanation belongs here. Edited the example shown also.


jonatanschroeder:
Given #3548 I'm not sure if this is desirable...

mwest1066:
We are a bit stuck with #3548 because of backwards-compatibility concerns. I think this PR is a good idea until we can figure out a strategy for #3548.

mwest1066 (approved):

Issue #3813, opened 2021-02-09
Groupwork data missing


Yucheng-Jiang:
I managed to download all data files of cs233(fall2020) but found all the students data related to ''group work'' are missing.  The submission, log,  and instance questions files only record TA and instructor's data. Also students group work scores are null in gradebook.

mwest1066:
@frankyhao can you take a look at this?

mwest1066:
I think we just need to change the JOINs on `users` to be LEFT JOINs and add a LEFT JOIN on groups: https://github.com/PrairieLearn/PrairieLearn/blob/c0beb25780ae5eeab62341a95b849dd6eecf13b3/api/v1/endpoints/queries.sql#L67

frankyhao:
Yeah, I only changed JOIN for queries in pages and may forget to change this one for gruops

Yucheng-Jiang:
any update?

PR #4850, opened 2021-09-28
change the JOINs on users to be LEFT JOINs in api/v1/endpoints

Yucheng-Jiang:
change the JOINs on users to be LEFT JOINs and add a LEFT JOIN on groups in api/v1/endpoints, fix bug mentioned in [here](https://github.com/PrairieLearn/PrairieLearn/issues/3813)


Yucheng-Jiang:

nwalters512:
Could you attempt to add a test for this to `tests/testApi.js`? There should be some prior art for group-related tests in `tests/testGroupStudent.js` and `tests/testGroupScoreAndSync.js`. I'd like to see the test fail without this change and then succeed with it.

Yucheng-Jiang:
emm this change is not about group student and group score. Previously, we cannot get any group activity assessment_instance created by student by calling api (reported issue [here](https://github.com/PrairieLearn/PrairieLearn/issues/3813). I followed Matthew and Frankyhao's suggest and looked into what Frankyhao changes in sql queries in `pages/instructorAssessmentInstance/instructorAssessmentInstance.sql`.

nwalters512:
I understand; I'm asking specifically for new tests for the API. We don't currently have any group-work-specific API tests yet, so I was providing links to other tests that do include group work so they can serve as a reference point to you.

api/v1/endpoints/queries.sql
```diff
@@ -65,7 +65,9 @@ WITH object_data AS (
         JOIN course_instances AS ci ON (ci.id = a.course_instance_id)
         JOIN assessment_sets AS aset ON (aset.id = a.assessment_set_id)
         JOIN assessment_instances AS ai ON (ai.assessment_id = a.id)
-        JOIN users AS u ON (u.user_id = ai.user_id)
+        LEFT JOIN group_info($assessment_id) AS gi ON (gi.id = ai.group_id)
+        LEFT JOIN group_users AS gu ON (gu.group_id = gi.id)
+        JOIN users AS u ON (u.user_id = ai.user_id OR u.user_id = gu.user_id)
```
mwest1066, line 70: This JOIN will result in multiple rows for the same assessment instance (one row per group user), which I don't think is what we want here (this is a query that is supposed to return one row per assessment instance). Instead I think you should do the same thing as this: https://github.com/PrairieLearn/PrairieLearn/blob/5b997eabb69a0b5707f6b0e2ab2d615069455756/pages/instructorAssessmentInstances/instructorAssessmentInstances.sql#L49

Also add new output columns for the group info, like this: https://github.com/PrairieLearn/PrairieLearn/blob/5b997eabb69a0b5707f6b0e2ab2d615069455756/pages/instructorAssessmentInstances/instructorAssessmentInstances.sql#L6

api/v1/endpoints/queries.sql
```diff
@@ -65,7 +65,9 @@ WITH object_data AS (
         JOIN course_instances AS ci ON (ci.id = a.course_instance_id)
         JOIN assessment_sets AS aset ON (aset.id = a.assessment_set_id)
         JOIN assessment_instances AS ai ON (ai.assessment_id = a.id)
-        JOIN users AS u ON (u.user_id = ai.user_id)
+        LEFT JOIN group_info($assessment_id) AS gi ON (gi.id = ai.group_id)
```
mwest1066, line 68: We can't assume that `$assessment_id` is not-NULL here, as you can see from the WHERE statement below: https://github.com/PrairieLearn/PrairieLearn/blob/5b997eabb69a0b5707f6b0e2ab2d615069455756/api/v1/endpoints/queries.sql#L71

Instead you should use `a.id` which is guaranteed to be not-NULL.

nwalters512:
per @mfsilva22, @Yucheng-Jiang is no longer interested in contributing this change. I'm going to move this commit to a new PR so that I can push changes to it.

nwalters512:
Superseded by #5271.

Issue #4733, opened 2021-09-01
Passing environment variables through ?


eddelbuettel:
We have an open PR in our nascent 'R question demos' repo where I support 'debug mode' by setting a toggle in `run.sh`, see https://github.com/eddelbuettel/pl-r-demos/pull/2

Now, Docker itself lets us set environment variables at the command-line and that is use by numerous containers to great effect.  But here we are in a Docker-in-Docker situation that I am less familiar with.  Is there a best practices standard for passing (some, all, any, none, ...) environment variable from the (outer) call to the (inner) container running inside the container?  It would be sweet to just turn 'verbose' or 'debug' mode on but I haven't really thought through if we'd unleash any undesirable monsters if we permitted that as a more general measure.  Does anybody here know?

(I figure I can also proxy it by just `touch`-ing a flag file inside the work directory and tickling a change a behavior is such a file is seen.  Possibly safer and likely simpler.)

echuber2:
You can set them on the entrypoint in info.json, here: https://github.com/eddelbuettel/pl-r-demos/blob/feature/update_run_sh/questions/R_Demos/test_R_function/info.json

But I'm not sure I fully understand the use case you had in mind.

echuber2:
e.g. if `/tmp/foo.sh` contains:

```
#!/bin/bash
echo "$ZOUNDS"
```

Then you can set the entrypoint like:

```
ZOUNDS="Spaceman Spiff" /tmp/foo.sh
```

eddelbuettel:
I call `prairielearn/prairielearn` via `docker` and can give `-e foo=bar` all day long.  

Now `prairielearn/prairielearn` calls, in my course, `stat447/pl` but  does _not_ as far as I understand it pass environments variable on when it does the black magic of docker-in-docker.  I could be wrong, I didn't test.   Will `stat447/pl` see `${foo}` ?

echuber2:
No, the autograder or workspace docker will not automatically inherit environment variables from the PrairieLearn server. However you can arrange for options to be passed through other means.

[...]

PR #4852, opened 2021-09-29
Add `environment` option for workspaces

tdy:
Env vars for workspaces (half of #4733)

### DIFF
--- a/sync/course-db.js
+++ b/sync/course-db.js
@@ -255,6 +255,7 @@ const FILE_UUID_REGEX = /"uuid":\s*"([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4
  * @property {string[]} syncIgnore
  * @property {string} rewriteUrl
  * @property {boolean} enableNetworking
+ * @property {string[]} environment
  */
 
  /**
--- a/sync/fromDisk/questions.js
+++ b/sync/fromDisk/questions.js
@@ -45,6 +45,7 @@ function getParamsForQuestion(q) {
         workspace_sync_ignore: (q.workspaceOptions && q.workspaceOptions.syncIgnore),
         workspace_url_rewrite: (q.workspaceOptions && q.workspaceOptions.rewriteUrl),
         workspace_enable_networking: (q.workspaceOptions && q.workspaceOptions.enableNetworking),
+        workspace_environment: (q.workspaceOptions && q.workspaceOptions.environment),
     };
 }
 
--- a/tests/sync/util.js
+++ b/tests/sync/util.js
@@ -159,6 +159,7 @@ const syncFromDisk = require('../../sync/syncFromDisk');
  * @property {string[]} syncIgnore
  * @property {string} rewriteUrl
  * @property {boolean=} enableNetworking
+ * @property {string[]} environment
  */
 
  /**
--- a/database/tables/questions.pg
+++ b/database/tables/questions.pg
@@ -27,6 +27,7 @@ columns
     uuid: uuid
     workspace_args: text
     workspace_enable_networking: boolean
+    workspace_environment: text[] default ARRAY[]::text[]
     workspace_graded_files: text[] default ARRAY[]::text[]
     workspace_home: text
     workspace_image: text
--- a/docs/workspaces/index.md
+++ b/docs/workspaces/index.md
@@ -65,6 +65,7 @@ The question's `info.json` should set the `singleVariant` and `workspaceOptions`
     * `syncIgnore` (optional, default none): list of files or directories that will be excluded from sync
     * `rewriteUrl` (optional, default true): if true, the URL will be rewritten such that the workspace container will see all requests as originating from /
     * `enableNetworking` (optional, default false): whether the workspace should be allowed to connect to the public internet. This is disabled by default to make secure, isolated execution the default behavior. This restriction is not enforced when running PrairieLearn in local development mode. It is strongly recommended to use the default (no networking) for exam questions, because network access can be used to enable cheating. Only enable networking for homework questions, and only if it is strictly required, for example for downloading data from the internet.
+    * `environment` (optional, default none): list of environment variables to pass into the workspace container, e.g. `["DEBUG=true", "FOO=bar"]`
 
 #### `info.json` for ungraded workspace
 
--- a/migrations/243_questions__workspace_environment__add.sql
+++ b/migrations/243_questions__workspace_environment__add.sql
@@ -1,0 +1,1 @@
+ALTER TABLE questions ADD COLUMN workspace_environment text[] DEFAULT ARRAY[]::text[];
--- a/schemas/schemas/infoQuestion.json
+++ b/schemas/schemas/infoQuestion.json
@@ -269,6 +269,10 @@
                 "enableNetworking": {
                     "description": "Whether the workspace should have network access. Access is disabled by default.",
                     "type": "boolean"
+                },
+                "environment": {
+                    "description": "Environment variables to pass into the workspace container.",
+                    "type": "array"
                 }
             }
         }
--- a/sprocs/sync_questions.sql
+++ b/sprocs/sync_questions.sql
@@ -137,6 +137,7 @@ BEGIN
         workspace_sync_ignore = jsonb_array_to_text_array(src.data->'workspace_sync_ignore'),
         workspace_url_rewrite = (src.data->>'workspace_url_rewrite')::boolean,
         workspace_enable_networking = (src.data->>'workspace_enable_networking')::boolean,
+        workspace_environment = jsonb_array_to_text_array(src.data->'workspace_environment'),
         sync_errors = NULL,
         sync_warnings = src.warnings
     FROM
--- a/workspace_host/interface.js
+++ b/workspace_host/interface.js
@@ -631,7 +631,10 @@ async function _getWorkspaceSettingsAsync(workspace_id) {
         workspace_args: result.rows[0].workspace_args || '',
         workspace_sync_ignore: result.rows[0].workspace_sync_ignore || [],
         workspace_enable_networking: !!result.rows[0].workspace_enable_networking,
+        workspace_environment: result.rows[0].workspace_environment || [],
     };
+    const env_workspace_base_url = `WORKSPACE_BASE_URL=/pl/workspace/${workspace_id}/container/`;
+    settings.workspace_environment.push(env_workspace_base_url);
 
     if (config.cacheImageRegistry) {
         const repository = new dockerUtil.DockerName(settings.workspace_image);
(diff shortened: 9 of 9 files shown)


tdy:

echuber2:
We should be sure to test that this doesn't mess up the base URL passthrough that some workspaces had previously relied upon. (Looks like you have it taken care of.)

eddelbuettel (commented):
I cannot speak to the implementation guts that I am mostly unfamiliar with but this seems to quite nicely address the (entirely optional) use of environment variables in the inner grading container and could be useful.

One minor comment. As I think of these as shell environment variable the space in the JSON example strikes me as ... "borderline".  Maybe best to protect these two words by single quotes?  Ie maybe make `"FOO=bar baz",` be `"FOO='bar baz'",` ?

tdy:
> One minor comment. As I thing of these as shell environment variable the space in the JSON example strikes me as ... "borderline". Maybe best to protect these two words by single quotes? Ie maybe make `"FOO=bar baz",` be `"FOO='bar baz'",` ?

I actually included that example on purpose because the result felt a bit unexpected. Docker escapes the input, so `"FOO='bar baz'"` will actually include literal quotes (basically `\'bar\ baz\'`).

echuber2:
I agree @tdy, I just tried it out... we want to avoid the quotes in this case.

eddelbuettel:
Maybe just avoid complexity and stick with `KEY=value`.  The `ILL=ini` is extra cute.  Any second one will do.

echuber2:
I just tried this with the JupyterLab workspace and it still seems to work okay (including passthrough of the base URL).

docs/externalGrading.md
```diff
@@ -70,6 +70,8 @@ External grading configuration is done on a per-question basis. All configuratio
 
 * `enableNetworking`: Allows the container to access the public internet. This is disabled by default to make secure, isolated execution the default behavior.
 
+* `environment`: List of environment variables to set inside the grading container in the form accepted by the [Docker Engine API](https://docs.docker.com/engine/api/v1.41/#operation/ContainerCreate): `["VAR=value", ...]`. A variable without `=` is removed from the environment, rather than to have an empty value.
```
nwalters512, line 73: IMO there's no reason to follow the pattern set by the Docker API - we're building something higher-level on top of that, so I think implementing as an object (key/value pair) makes more sense for a user-facing API. I spot-checked a few other projects and AWS Lambda, GCP Functions, Azure Functions, and Kubernetes all do (roughly) that instead of using `FOO=BAR`.

nwalters512 (commented):
This all looks pretty good! I want to chat about `{"FOO": "BAR"}` vs. `["FOO=BAR"]` a bit before landing this change.

docs/externalGrading.md
echuber2: Then I think it's a good choice to use the object, as it removes this ambiguity about the parsing without quotes.

echuber2: How should unsetting variables be handled?

tdy: Sure, I can reformat. Thoughts on how (or whether) to support `unset`ting of variables? Docker will `unset FOO` given `["FOO"]`.

Maybe we can `unset` given `{"FOO": null}`, e.g.:

```js
Object.keys(env).map(key => env[key] === null ? key : `${key}=${env[key]}`)
```

echuber2: We'll need to make sure the key doesn't have spaces or other forbidden characters, perhaps:

```
> const env = {
... 'name': 'bob',
... 'age': 30,
... 'address': null,
... 'favorite food': 'green eggs and takoyaki'
... }
undefined
> Object.keys(env).map(key => env[key] === null ? key : `${key}=${env[key]}`)
[
  'name=bob',
  'age=30',
  'address',
  'favorite food=green eggs and takoyaki'
]
```

echuber2:
So, I've found something strange. `docker run` will let you set env vars with spaces in the names, although afaik this is against POSIX standards. Interestingly, Bash shows it with the space in the name as part of the environment (`printenv`), but there seems to be no actual way to dereference the variable once declared.

nwalters512:
> So, I've found something strange. `docker run` will let you set env vars with spaces in the names, although afaik this is against POSIX standards. Interestingly, Bash shows it with the space in the name as part of the environment (`printenv`), but there seems to be no actual way to dereference the variable once declared.

This isn't unique to Docker - plain old `env` will let you do that, and you can definitely dereference it, at least in Node:

```sh
env "FOO BAR=baz" node -e "console.log(process.env['FOO BAR'])"
```

sync/course-db.js
```diff
@@ -243,6 +243,7 @@ const FILE_UUID_REGEX = /"uuid":\s*"([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4
  * @property {string[]} serverFilesCourse
  * @property {number} timeout
  * @property {boolean} enableNetworking
+ * @property {Object} environment
```
nwalters512, line 246: Nit, this is more precise:
```suggestion
 * @property {Record<string, string | null>} environment
```

sync/course-db.js
```diff
@@ -255,6 +256,7 @@ const FILE_UUID_REGEX = /"uuid":\s*"([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4
  * @property {string[]} syncIgnore
  * @property {string} rewriteUrl
  * @property {boolean} enableNetworking
+ * @property {Object} environment
```
nwalters512, line 259: Nit, this is more precise:
```suggestion
 * @property {Record<string, string | null>} environment
```

eddelbuettel:
Good point.  Ditto from/to the R prompt:

```r
> Sys.setenv("FOO BAR"=42L)
> Sys.getenv("FOO BAR")
[1] "42"
> as.integer(Sys.getenv("FOO BAR")) == 42L
[1] TRUE
> 
```

sync/fromDisk/questions.js
```diff
@@ -36,6 +36,7 @@ function getParamsForQuestion(q) {
         external_grading_entrypoint: (q.externalGradingOptions && q.externalGradingOptions.entrypoint),
         external_grading_timeout: (q.externalGradingOptions && q.externalGradingOptions.timeout),
         external_grading_enable_networking: (q.externalGradingOptions && q.externalGradingOptions.enableNetworking),
+        external_grading_environment: (q.externalGradingOptions && q.externalGradingOptions.environment) || {},
```
nwalters512, line 39: Nit: we're running Node 14, so we can use optional chaining and nullish coalescing to make this a bit more concise:
```suggestion
        external_grading_environment: q.externalGradingOptions?.environment ?? {},
```

sync/fromDisk/questions.js
```diff
@@ -45,6 +46,7 @@ function getParamsForQuestion(q) {
         workspace_sync_ignore: (q.workspaceOptions && q.workspaceOptions.syncIgnore),
         workspace_url_rewrite: (q.workspaceOptions && q.workspaceOptions.rewriteUrl),
         workspace_enable_networking: (q.workspaceOptions && q.workspaceOptions.enableNetworking),
+        workspace_environment: (q.workspaceOptions && q.workspaceOptions.environment) || {},
```
nwalters512, line 49: Nit: we're running Node 14, so we can use optional chaining and nullish coalescing to make this a bit more concise:
```suggestion
        workspace_environment: q.workspaceOptions?.environment ?? {},
```

tests/sync/util.js
```diff
@@ -147,6 +147,7 @@ const syncFromDisk = require('../../sync/syncFromDisk');
  * @property {string[]=} serverFilesCourse
  * @property {number=} timeout
  * @property {boolean=} enableNetworking
+ * @property {Object=} environment
```
nwalters512, line 150: Nit, this is more precise:
```suggestion
 * @property {Record<string, string | null>=} environment
```

tests/sync/util.js
```diff
@@ -159,6 +160,7 @@ const syncFromDisk = require('../../sync/syncFromDisk');
  * @property {string[]} syncIgnore
  * @property {string} rewriteUrl
  * @property {boolean=} enableNetworking
+ * @property {Object=} environment
```
nwalters512, line 163: Nit, this is more precise:
```suggestion
 * @property {Record<string, string | null>=} environment
```

lib/externalGraderLocal.js
```diff
@@ -89,6 +89,8 @@ class Grader {
             (callback) => {
                 docker.createContainer({
                     Image: question.external_grading_image,
+                    // Convert {key: 'value'} to ['key=value'] and {key: null} to ['key'] for Docker API
+                    Env: Object.entries(question.external_grading_environment).map(([k, v]) => v === null ? k : `${k}=${v}`),
```
nwalters512, line 93: @mwest1066 remind me what the latest with PG is - are we still running off of https://github.com/PrairieLearn/PrairieGrader in production, or are we using the local `grader_host` directory?

@tdy pending the above, you may have to add code in PrairieGrader to handle this, but at a minimum, you'll need to update https://github.com/PrairieLearn/PrairieLearn/blob/master/lib/externalGraderSqs.js and https://github.com/PrairieLearn/PrairieLearn/blob/master/grader_host/index.js to handle this as well.

nwalters512 (commented):
Will approve once the `externalGraderSqs` and `grader_host` (and possibly PrairieGrader) code has been updated to support this new setting!

echuber2:
> This isn't unique to Docker - plain old `env` will let you do that, and you can definitely dereference it, at least in Node:

Well, this is news to me that `env` allows it. Interestingly, `bash` cannot dereference such variable names although Node and R can:

https://stackoverflow.com/questions/18468397/how-can-you-access-an-environment-variable-that-has-a-space-in-its-name-in-bash/18469375#18469375

docs/externalGrading.md
```diff
@@ -70,6 +70,8 @@ External grading configuration is done on a per-question basis. All configuratio
 
 * `enableNetworking`: Allows the container to access the public internet. This is disabled by default to make secure, isolated execution the default behavior.
 
+* `environment`: Environment variables to set inside the grading container. Set variables using `{"VAR": "value", ...}`, and unset variables using `{"VAR": null}` (no quotes around `null`).
```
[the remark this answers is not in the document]
echuber2: Scripts may respond differently to the cases of a variable being not declared at all vs. assigned a blank value. For example, Node differentiates these as `undefined` and `''`. It may also be desired to hide suggestive variable names leftover from deployment or something.

At any rate, since the environment (e.g. bash `unset`) and Docker allow it, it makes sense to me to support it.

lib/externalGraderLocal.js
tdy: > at a minimum, you'll need to update `lib/externalGraderSqs.js` and `grader_host/index.js`

done in 2a2061c28db7be6194e913ce34950f3747bf4389 (I think)

PR #4911, opened 2021-10-03
Reorganize tools directory

firasm:
**This is still a work in progress**

Here are the things this PR will eventually do:

- [x] Move every script/tool into its own folder
- [x] Create an index of all the scripts so it's easy to find them
- [x] Add a template README file to each script so instructions can be added
- [ ] Categorize the scripts
- [ ] Find the original contributors and attribute scripts to them 
- [ ] Archive obsolete scripts


firasm:
Yikes! Didn't realize a GH action relied on a script in this directory.

firasm:
Note to self: it's a really bad idea to open PRs from a forked repo as a draft !

nwalters512:
@firasm would you still like to get this merged? Otherwise, I'm going to close this out to trim down our list of PRs.

firasm:
let's close it and if I get some time in the summer, I can resurrect it.

eliotwrobson:
Note if this gets revived: Since we're now trying to format + lint most of the Python code in the repo, it would be great if we could do that in the tools directory also. One of the obstacles to this has been the fact that there's some Python 2 files in this directory, so it would be great to have those ported over to Python 3.

PR #4916, opened 2021-10-04
Pass student identity to questions

akshitdewan:
Use Case: Two main reasons
1) so we can quickly sync the score with our primary grading server 
2) so we can try some controlled experiments, such as what type of feedback to give on an incorrect answer (hint videos, step-by-step walkthroughs, links to related exercises, etc).

In order to get access to student identity, one will have to add a flag to the question's info.json:
```
"requiresStudentIdentity": true
```
Then, server.py will receive the student identity in both the `generate` and the `grade` functions in the `data` object. Specifically, data['student_identity'] will be a dictionary following the same structure as the `res.locals.user`, which I believe follows the schema in users.pg.

I tested this ad-hoc by displaying the student identity in the question HTML, with and without the flag. Will add an explicit question or two that tests this in the example course. Is that the right place?

My view of the codebase is fairly concentrated on how this one feature works, so definitely happy to clean up this PR in whatever way you all see fit!

### DIFF
--- a/database/tables/course_instances.pg
+++ b/database/tables/course_instances.pg
@@ -11,6 +11,7 @@ columns
     sync_job_sequence_id: bigint
     sync_warnings: text
     uuid: uuid
+    pass_student_identity_to_questions: boolean default false
 
 indexes
     course_instances_pkey: PRIMARY KEY (id) USING btree (id)
--- a/database/tables/course_instances.pg
+++ b/database/tables/course_instances.pg
@@ -11,7 +11,6 @@ columns
     sync_job_sequence_id: bigint
     sync_warnings: text
     uuid: uuid
-    pass_student_identity_to_questions: boolean default false
 
 indexes
     course_instances_pkey: PRIMARY KEY (id) USING btree (id)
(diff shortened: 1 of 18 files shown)


akshitdewan:

jonatanschroeder:
Not providing a formal review here, but just adding that your code uses the user from authentication/effective user. Although this will often match, note that this might give you unexpected behaviour in group assessments, since the question will have different parameters depending on which partner opens the question first, and which partner submits the question. You may want to review the code to consider using the user/group from the assessment instead.

akshitdewan:
> Not providing a formal review here, but just adding that your code uses the user from authentication/effective user. Although this will often match, note that this might give you unexpected behaviour in group assessments, since the question will have different parameters depending on which partner opens the question first, and which partner submits the question. You may want to review the code to consider using the user/group from the assessment instead.

Thanks @jonatanschroeder I will take a look at how to get data from the assessment instead, and would appreciate any relevant code pointers.

nwalters512:
Hey @akshitdewan 👋 are you still interested in contributing this? Otherwise I'll go ahead and close out this PR.

nwalters512:
This PR has accumulated a lot of conflicts and hasn't seen any activity in almost a year. I'm going to close this out; @akshitdewan if you're still interested in contributing, please open a new rebased PR!

PR #4917, opened 2021-10-05
Add "feedback" attribute to pl-checkbox

Cris-Moran:
This feature is similar to the `feedback` feature for `pl-multiple-choice` (https://github.com/PrairieLearn/PrairieLearn/pull/4656). When a user submits an answer through pl-checkbox, all of the selected answers that contain feedback will have its feedback displayed.

Demonstration:
Screen Shot 2021-10-04 at 10 11 59 PM


elements/pl-checkbox/pl-checkbox.css
```diff
@@ -7,3 +7,23 @@
...
+.pl-checkbox-submission-feedback {
+    border-left: 0.25rem solid rgba(0, 0, 0, 0.4);
+    background-color: rgba(0, 0, 0, 0.03);
+    padding: 0.25rem 0.5rem;
+    margin-left: 0.75rem;
+    margin-right: 0.5rem;
+    margin-top: 0.25rem;
+    margin-bottom: 0.25rem;
+    border-radius: 0.2rem;
+}
```
nwalters512, line 29: AFAICT, these are virtually the same (expect the latter has additional top/bottom margins. Consider collapsing the shared bits into a single `.pl-checkbox-feedback` class.

Separately, we generally prefer to use Bootstrap classes to style things - this means applying classes like `ml-2`, `rounded`, etc. instead of using bespoke styles like these in a CSS file.

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -481,8 +494,9 @@ def test(element_html, data):
                 raise ValueError(f'Unknown value for partial_credit_method: {partial_credit_method}')
         else:
             score = 0
+        submitted_keys_to_feedback = {option['key']: option['feedback'] for option in data['params'][name] if option['key'] in ans}
         data['raw_submitted_answers'][name] = ans
-        data['partial_scores'][name] = {'score': score, 'weight': weight}
+        data['partial_scores'][name] = {'score': score, 'weight': weight, 'submitted_keys_to_feedback': submitted_keys_to_feedback}
```
nwalters512, line 499: I'd rather we use just `feedback` here . That would be consistent with how elements work, including those with multiple pieces of feedback, for instance `pl-matrix-component-input`:

https://github.com/PrairieLearn/PrairieLearn/blob/5b997eabb69a0b5707f6b0e2ab2d615069455756/elements/pl-matrix-component-input/pl-matrix-component-input.py#L381-L413

nwalters512 (commented):
Thanks for the contribution! Before we can land this, you'll need to update the docs for this element in `docs/elements.md` and add an example to `exampleCourse/questions/element/checkbox/question.html`. I've left a few comments for discussion too.

Cris-Moran:
I made the suggested changes a while ago, please review them when possible :)

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -152,14 +154,17 @@ def render(element_html, data):
...
         answerset = []
         for answer in display_answers:
             answer_html = {
                 'key': answer['key'],
                 'checked': (answer['key'] in submitted_keys),
                 'html': answer['html'].strip(),
-                'display_score_badge': score is not None and show_answer_feedback and answer['key'] in submitted_keys
+                'display_score_badge': score is not None and show_answer_feedback and answer['key'] in submitted_keys,
+                'display_feedback': feedback and answer['feedback'] in feedback and feedback[0] is not None,
+                'feedback': feedback.pop(0) if feedback and answer['feedback'] in feedback else None
```
nwalters512, line 167: I'm a little confused with what's going on here:

- I don't understand what `answer['feedback'] in feedback` and `feedback[0] is not None` are checking for, and specifically why they're referencing the `feedback` variable - it seems like it would be sufficient to just write `answer.get('feedback', None) is not None`.
- Similarly for `feedback.pop(0) if feedback and answer['feedback'] in feedback else None`, I suspect that could just be written as `answer.get('feedback', None)`.

Regardless of all that, we need to avoid unchecked access to `answer['feedback']` for backwards compatibility with existing question instances (e.g., trying to render an existing question fails with `KeyError`).

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -295,6 +300,8 @@ def render(element_html, data):
                 if answer_item['display_score_badge']:
                     answer_item['correct'] = (submitted_key in correct_keys)
                     answer_item['incorrect'] = (submitted_key not in correct_keys)
+                answer_item['display_feedback'] = submitted_answer['feedback']
+                answer_item['feedback'] = submitted_answer['feedback'] if submitted_answer['feedback'] else None
```
nwalters512, line 304: You'll need to use `submitted_answer.get('feedback', None)` instead of `submitted_answer['feedback']` for backwards compatibility with existing question instances - this will fail with a `KeyError` otherwise.

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -426,7 +434,7 @@ def grade(element_html, data):
         else:
             raise ValueError(f'Unknown value for partial_credit_method: {partial_credit_method}')
 
-    data['partial_scores'][name] = {'score': score, 'weight': weight}
+    data['partial_scores'][name] = {'score': score, 'weight': weight, 'feedback': feedback}
```
nwalters512, line 437: Can you explain why this is being stored on `partial_scores`? I think all the feedback is already stored in `data['params'][name][*]['feedback']`, and I'd rather keep that as the "canonical" source for feedback if possible.

nwalters512 (changes requested):
Blocking on unchecked access to `['feedback']` key - backwards compatibility with existing questions is very important.

elements/pl-checkbox/pl-checkbox.py
Cris-Moran: Yeah I could have made this more clear, my bad. Allow me to explain:

- `feedback` is an attribute of the `partial scores` object. In `grade()` it gets populated with the feedback for each answer choice that was selected by the student, in order. For example, if  A, C, and E are chosen, `feedback` would look like `[A's feedback, C's feedback, E's feedback]`. `answer['feedback']` is the feedback attached to the answer itself.

To answer your questions:

- `answer['feedback'] in feedback` is meant to check if the answer choice was chosen by the student. Looking at this again I suppose it would be better to replace it with `answer['key'] in submitted_keys`. In the example in the previous section, if A didn't have feedback, `A's feedback` would be `None`. `feedback[0] is not None` checks that if a student submitted one of these answers that don't contain feedback, it doesn't get displayed.  Otherwise, an empty gray box would appear.
- `feedback.pop(0) if feedback and answer['feedback'] in feedback else None` gets the actual feedback to display from the `feedback` list. Using the same reasoning as before, I suppose I can replace `answer['feedback'] in feedback` with `answer['key'] in submitted_keys`. 

With these edits, there are no more accesses to `answer['feedback']`. I'm hesitant to use `answer.get('feedback', None)` so I can be consistent with `pl-multiple-choice`, where I was told to put `feedback` into `partial_scores` and use it in `render()` (https://github.com/PrairieLearn/PrairieLearn/pull/4656#issuecomment-899773221). I'll have the changes pushed shortly, let me know what you think :)

elements/pl-checkbox/pl-checkbox.py
Cris-Moran: See https://github.com/PrairieLearn/PrairieLearn/pull/4656#issuecomment-899773221

elements/pl-checkbox/pl-checkbox.py
Cris-Moran: That should work, to be consistent with `pl-multiple-choice` though I'll modify this to use the `partial_scores` dictionary.

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -453,7 +462,8 @@ def test(element_html, data):
             data['raw_submitted_answers'][name] = correct_keys
         else:
             pass  # no raw_submitted_answer if no correct keys
-        data['partial_scores'][name] = {'score': 1, 'weight': weight}
+        feedback = [option['feedback'] for option in data['params'][name] if option['key'] in correct_keys]
```
nwalters512, line 465: This shouldn't break in practice, but let's use `.get()` to be extra careful.

```suggestion
        feedback = [option.get('feedback', None) for option in data['params'][name] if option['key'] in correct_keys]
```

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -400,6 +408,7 @@ def grade(element_html, data):
     submitted_keys = data['submitted_answers'].get(name, [])
     correct_answer_list = data['correct_answers'].get(name, [])
     correct_keys = [answer['key'] for answer in correct_answer_list]
+    feedback = [option['feedback'] for option in data['params'][name] if option['key'] in submitted_keys]
```
nwalters512, line 411: We need to handle the situation where the `feedback` key doesn't exist - this will be the case for any variant generated before this change gets deployed.

```suggestion
    feedback = [option.get('feedback', None) for option in data['params'][name] if option['key'] in submitted_keys]
```

elements/pl-checkbox/pl-checkbox.py
nwalters512: Super helpful - thanks! Jonatan's comment on the other PR for `pl-multiple-choice` makes a lot of sense.

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -481,8 +491,9 @@ def test(element_html, data):
                 raise ValueError(f'Unknown value for partial_credit_method: {partial_credit_method}')
         else:
             score = 0
+        feedback = [option['feedback'] for option in data['params'][name] if option['key'] in ans]
```
nwalters512, line 494: Let's use `.get()` here too for safety.

```suggestion
        feedback = [option.get('feedback', None) for option in data['params'][name] if option['key'] in ans]
```

nwalters512 (changes requested):
We'll need to update the remaining unchecked accesses to the `feedback` property in a few more places before this can be merged. For reference, I tested this by checking out `master`, generating variants of the MC examples there, then checking out this branch and trying to grade submissions - I got errors when I did that.

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -152,14 +154,17 @@ def render(element_html, data):
...
         answerset = []
         for answer in display_answers:
             answer_html = {
                 'key': answer['key'],
                 'checked': (answer['key'] in submitted_keys),
                 'html': answer['html'].strip(),
-                'display_score_badge': score is not None and show_answer_feedback and answer['key'] in submitted_keys
+                'display_score_badge': score is not None and show_answer_feedback and answer['key'] in submitted_keys,
+                'display_feedback': feedback and answer['key'] in submitted_keys and feedback[0] is not None,
+                'feedback': feedback.pop(0) if feedback and answer['key'] in submitted_keys else None
```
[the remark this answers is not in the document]
Cris-Moran: I was able to do this using a counter variable. I'll have the changes pushed shortly.

Cris-Moran:
So it looks like a made a mistake a while ago where I committed to the master branch in my fork :/ I'll make a new fork so that this doesn't cause any problems.

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -148,19 +150,26 @@ def render(element_html, data):
 
     correct_answer_list = data['correct_answers'].get(name, [])
     correct_keys = [answer['key'] for answer in correct_answer_list]
+    # Pointer used to index feedback
+    ptr = 0
```
nwalters512, line 154: I believe this and the below logic is more complex than it needs to be. I don't think we need to conditionally increment this - something like this is simpler to understand and more Pythonic:

```py
for i, answer in display_answers:
	answer_html = {
    # ...
    'feedback': feedback[i] if feedback and answer['key'] in submitted_keys else None
  }
```

Apologies if there's a specific reason that we're conditionally incrementing the index and I missed it. If there is, let's document it.

Cris-Moran: The conditional is used to make sure we don't go over a piece of feedback as we're iterating through the answer choices.

For example, let's say I have five questions in `display_answers` and I select two. `feedback` is now two elements long.

As I'm iterating through `display_answers`, `i` might not index to the correct item in `feedback`.

After some thought, I can get rid of the pointer syntax by making `feedback` a dictionary, or a list of dictionaries. Let me know your thoughts.

nwalters512: Ahhh, I understand - the elements of `feedback` correspond to what's been selected, not necessarily the elements of `display_answers`?

I think there are two potential ways to improve this:

1. refactor this so that `feedback` is a dict that maps answer keys to their feedback:
```py
{
  'a': 'a feedback',
  'b': 'b feedback',
  'c': None
}
```
2. refactor this so that feedback is an array whose elements correspond 1:1 with the items in `display_answers`:
```py
[
  'a feedback',
  'b feedback',
  None
]
```

In both cases, `feedback` would contain feedback for *all* answer choices, not just the ones that the user chose for a particular submission. IMO, that makes this easier to reason about. Whether or not we actually end up rendering the feedback would still be determined by if the user chose that coice, etc.

Cris-Moran: Got it. Will push changes shortly.

nwalters512:
After your latest changes, attempting to grade the new example question you added results in the following runtime error:

```
  File "/Users/nathan/git/PrairieLearn/lib/python-caller-trampoline.py", line 182, in <module>
    worker_loop()
  File "/Users/nathan/git/PrairieLearn/lib/python-caller-trampoline.py", line 128, in worker_loop
    val = method(*args)
  File "/Users/nathan/git/PrairieLearn/elements/pl-checkbox/pl-checkbox.py", line 166, in render
    'display_feedback': answer['key'] in submitted_keys and feedback.get(answer['key'], None) is not None,
AttributeError: 'NoneType' object has no attribute 'get'
  File "/Users/nathan/git/PrairieLearn/lib/python-caller-trampoline.py", line 194, in <module>
```

Cris-Moran:
My bad, I forgot to check if `feedback` exists. It should be fixed now.

nwalters512 (approved):

PR #4919, opened 2021-10-05
Fix handling of blank scores in score upload

jonatanschroeder:
Making a distinction between a blank `score_perc` (no change) and a zero (explicitly set to zero). This also fixes an issue whereby a grade of 0 wouldn't update the instance question status from `grading` to `complete`.


nwalters512 (approved):
Looks great!

This is a good reminder to me that https://eslint.org/docs/rules/eqeqeq is not enabled in this repo - I'll open a followup PR turning that on.

mwest1066 (approved):

PR #4923, opened 2021-10-05
chore: Enable eqeqeq ESLint rule

nwalters512:
This PR is a followup to https://github.com/PrairieLearn/PrairieLearn/pull/4919. It enables the [`eqeqeq`](https://eslint.org/docs/rules/eqeqeq) ESLint rule that forbids the use of `==` except in very specific cases (comparing two literals, evaluating the value of `typeof`, or comparing against `null`).

This was a huge change, as you can see from the diff. I did a lot of the work with good old find-replace, (e.g. I assumed everything that compared with a string literal could trivially safely use `===`, so I replaced all instances of ` == '` with ` === '` and ` != '` with ` !== '), but some required much more manual changes.

Specifically, we actually _rely_ on the type coercion provided by `==` when comparing IDs in many places, since even Postgres sometimes returns them as strings and other times as numbers. The approach I took to fix this was to introduce a new `idsEqual` function that should be used whenever we're comparing IDs. Currently, it's just a wrapper around `==`, but it could be enhanced to do other things in the future. At any rate, I felt that was better than having a bunch of `//eslint-disable` comments scattered everywhere. See [this commit](https://github.com/PrairieLearn/PrairieLearn/commit/954af298a21774cc6f2d72cc88eddcc3dfc5c7ab) to see where I added this function and changed implementations.


docker/start_s3rver.sh
```diff
@@ -13,8 +13,8 @@ if lsof -i :5000 > /dev/null ; then
     exit
 fi
 
-mkdir -p /s3rver
-node_modules/.bin/s3rver --directory /s3rver --port 5000 --configure-bucket workspaces --configure-bucket chunks --configure-bucket file-store > /dev/null &
+mkdir -p ./s3rver
+node_modules/.bin/s3rver --directory ./s3rver --port 5000 --configure-bucket workspaces --configure-bucket chunks --configure-bucket file-store > /dev/null &
```
nwalters512, line 17: I tried to run this with `make start` on my machine outside of docker, but the root of my machine (`/`) is read-only, so this failed. I opted to change this to create the directory in PrairieLearn instead. See the corresponding change to `.gitignore`.

lib/id.js
```diff
@@ -0,0 +1,29 @@
...
+ * In the future, the existence of this function could make it easier to start
+ * parsing IDs as `BigInt` objects as they come out of the database.
+ * 
+ * @param {string | number} id1 
+ * @param {string | number} id2 
+ */
+module.exports.idsEqual = function idsEqual(id1, id2) {
+  // eslint-disable-next-line eqeqeq
+  return id1 == id2;
+};
```
nwalters512, line 29: Here's the `idsEqual` function I mentioned in the PR description.

nwalters512:
The tests pass, which gives me a reasonable amount of confidence that this is good to go!

docker/start_s3rver.sh
mwest1066: This seems fine. We were keeping in `/` assuming everyone was running inside a container, but I'm on board with supporting non-containerized dev.

lib/id.js
mwest1066: Should we add a note about this to https://github.com/PrairieLearn/PrairieLearn/blob/master/docs/dev-guide.md ?

mwest1066:
Everything here makes sense to me.

I'm nervous about this PR because:
1. Are there other cases where we are relying on `==` behavior? The ones that occur to me are `== null` (taken care of), the fact that postgres 64-bit BIGINT maps to string due to Number limits (taken care of), and possible issues with inputs (e.g., form inputs, data from JSON) that we are silently coercing at the moment (I don't know whether there are any).
2. This might cause a lot of merge noise with other outstanding PRs. I don't know how much of an issue this really is.
3. I'm generally a bit nervous of a patch that's so invasive. The risk of weird issues seems high.

But I'm happy to proceed anyway 😄

mwest1066:
> Unrelated, but I think I got through about 20 or so files before realizing how many changes were actually made in this PR 😀

I fortunately looked at the "changed files" count before I started, although I would also have been altered by the fact that my browser slowed down when trying to view the diff 😄

lib/editors.js
```diff
@@ -1172,12 +1173,12 @@ class QuestionRenameEditor extends Editor {
...
                                         question.alternatives.forEach((alternative) => {
-                                            if (alternative.id == this.question.qid) {
+                                            if (idsEqual(alternative.id, this.question.qid)) {
                                                 alternative.id = this.qid_new;
                                                 found = true;
                                             }
                                         });
-                                    } else if (question.id == this.question.qid) {
+                                    } else if (idsEqual(question.id, this.question.qid)) {
                                         question.id = this.qid_new;
```
jonatanschroeder, line 1182: I believe these are for QID, so I imagine `===` would be fine...

ataraxie:
I just started watching the PL repo, but I'll feel so free to say my opinion (please just ignore if you think this is no good, haha).

I think this is an important change because `==` is a very common source for bugs. However, I'd also be a bit shaky changing it everywhere all at once in the code base. It might make sense to roll this out a bit more carefully. For example, you could first only replace all instances that are simple string comparisons (not comparing against empty string) or number comparisons (except comparing against 0 and 1). This would be a very safe change. Then, once it's obvious that this didn't break anything (like in a future sprint), you could continue to change another group of checks, etc. But yes, this would require some temporary `//eslint-disable` annotations. But this can be actually kind of nice because it makes it very obvious what still needs to be changed.

P.S. I'm generally on the careful side of things when it comes to these issues and people have often convinced me that something should just be done all-at-once 😅

pages/instructorFileTransfer/instructorFileTransfer.js
```diff
@@ -44,11 +46,12 @@ function getFileTransfer(file_transfer_id, user_id, callback) {
 
 router.get('/:file_transfer_id', function(req, res, next) {
     if (config.filesRoot == null) return next(new Error('config.filesRoot is null'));
+    console.log('res.locals.user', res.locals.user);
```
nwalters512, line 49:
```suggestion
```

nwalters512 (commented):
Comments about this being a bit risky and invasive are completely valid - I agree. To de-risk a bit, I did a self-review of my own changes and found a few things that I'd like a second opinion on. @mwest1066 @jonatanschroeder @nicknytko sorry to ping again, but there's a relatively small set of things that I'd appreciate input on!

jonatanschroeder (approved):
You pushed a change as I was reviewing, so you may have fixed something already...

pages/studentAssessmentInstanceHomework/studentAssessmentInstanceHomework.js
```diff
@@ -61,7 +61,7 @@ router.get('/', function(req, res, next) {
                     sqldb.query(sql.get_group_info, params, function(err, result) {
                         if (ERR(err, next)) return;
                         res.locals.group_info = result.rows;
-                        if (res.locals.group_info[0] == undefined) return next(error.make(403, 'Not a group member', res.locals));
+                        if (res.locals.group_info[0] === undefined) return next(error.make(403, 'Not a group member', res.locals));
```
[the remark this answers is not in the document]
jonatanschroeder: I believe this query allows multiple rows, so `if (result.rowCount > 0)`, or even `if (result.rowCount)`, should be better. That said, this query is weird...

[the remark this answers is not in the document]
nwalters512: Ahhhh, very good catch. This query is indeed weird, I had to go diving all the way down into `pages/shared/groupWorkInfoContainer.ejs` to see that we do actually expect `group_info` to have multiple rows. I'll update my suggestion.

elements/pl-order-blocks/pl-order-blocks.js
```diff
@@ -68,7 +68,7 @@ function update_indent(leftDiff, id, ui) {
 
     // leftDiff is the direction to move the MCQ answer tile, in px
     // we limit leftDiff to be increments of TABWIDTH, whether positive or negative
-    if (currentIndent != ''){
+    if (currentIndent !== ''){
```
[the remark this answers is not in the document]
Frozemint: I think using `!==` would work, the intention is to check whether a block is indented by checking the `marginLeft` CSS style is set, as in `var currentIndent = ui.item[0].style.marginLeft;` 10 lines earlier. 

I can't immediately say whether checking `if (currentIndent)` would work, because I don't know too much about Javascript truthy falsy to say if changing the check to `if (currentIndent)` would add any unintentional behaviour (because javascript). Would the javascript wizards of this repo mind chiming in? There might be another much better way to check if the CSS property is set tho.

pages/instructorFileTransfer/instructorFileTransfer.js
```diff
@@ -19,11 +20,12 @@ function getFileTransfer(file_transfer_id, user_id, callback) {
...
                 if (ERR(err, callback)) return;
                 file_transfer = result.rows[0];
-                if (file_transfer.transfer_type != 'CopyQuestion') {
+                if (file_transfer.transfer_type !== 'CopyQuestion') {
                     return callback(new Error(`bad transfer_type: ${file_transfer.transfer_type}`));
                 }
-                if (file_transfer.user_id != user_id) {
-                    return callback(new Error(`must have same user_id: ${file_transfer.user_id} and ${user_id}`));
+                if (!idsEqual(file_transfer.user_id, user_id)) {
+                    console.log('file_transfer', file_transfer);
```
nwalters512, line 27:
```suggestion
```

pages/instructorFileTransfer/instructorFileTransfer.js
```diff
@@ -19,11 +20,12 @@ function getFileTransfer(file_transfer_id, user_id, callback) {
...
+                    return callback(new Error(`must have same user_id: ${file_transfer.user_id} and ${user_id} (types: ${typeof file_transfer.user_id}, ${typeof user_id})`));
```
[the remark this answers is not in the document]
nwalters512: Resolved in e7231f9

mwest1066:
I think that once we move to a staged deployment model (next week) we should go ahead and land this PR. I'll feel confident deploying it if we can try it out with some live data.

nwalters512:
> I think that once we move to a staged deployment model (next week) we should go ahead and land this PR. I'll feel confident deploying it if we can try it out with some live data.

Sounds good - I'll hold off on merging until we get that working!

nwalters512:
Merging master (and thus #5008) into this branch was somewhat of a nightmare. Note to self: don't queue up multiple huge changes like this in the future 🙂 

I'm going to go back and review each and every changed file (again, ugh) to make sure merging occurred without any unintended changes.

PR #4934, opened 2021-10-08
Execute v2 question JavaScript code in subprocess

### LINKED ISSUE #3846: v2 questions are not properly reloaded on multi-servers
v2 questions load `server.js` into the main running PL process using RequireJS. To make sure we've got the latest on-disk copy, we call `lib/require-frontend.undefQuestionServers()` from a variety of locations in the code (basically anywhere that might change on-disk v2 question code). This function removes all JS files in the course from the RequireJS cache.

However, this all fails when we are running multiple servers because server2 doesn't know to call `undefQuestionServers()` after server1 changed an on-disk v2 question. Some ideas to fix this:

1. **Unconditionally call `undefQuestionServers()` before every attempt to load a `server.js` files.** We already do this in dev mode via `middlewares/undefCourseCode.js`. I worry a bit about doing this in production because we will have multiple simultaneous in-flight requests all accessing the same question and I don't know exactly what happens when you `undef()` a module that's currently in use. On the other hand, [the docs](https://requirejs.org/docs/api.html#undef) do seems to suggest that it might be ok. I'm also not sure what the performance impact of this strategy would be. The advantage of this strategy is that it's super simple and (if it works) robust, and it would allow us to rip out the calls to `undefQuestionServers()` that are currently sprinkled throughout the code.
2. **Somehow detect that a `server.js` has changed on disk.** I don't have a great strategy for this, but maybe we could check the course `commit_hash` like we do when deciding whether to reload the element cache for freeform questions?
3. **Broadcast a message to all running servers.** If server1 calls `undefQuestionServers()` then it could also do a pub/sub message to all other servers via redis to tell them to also call `undefQuestionServers()`. This feels horribly complicated and fragile, however.
4. **Shift execution of v2 question code into subprocess.** For safety/robustness reasons we really want to shift the execution of course code into a subprocess anyway (preferably containerized: #1626). If the subprocess does not retain state between question executions then it would additionally fix this issue. One idea might be to have the `python-caller-trampoline.py` be able to run JS files by spawning a NodeJS process that loads and runs `server.js` for the question. This would be a bit inefficient, but that's probably fine. Node has very fast startup so it would probably work fine to spawn a separate process every time we want to call a `server.js` function.

nwalters512:
This PR moves the execution of server-side JavaScript code for v2 questions into a subprocess. Unfortunately, Node doesn't allow POSIX-style forks, so we can't use the same zygote model that we do for Python. This means that we incur the cost of a Node process start-up for every function call. This should be ok, since at most we'll do one or two function calls per page render (we don't have elements that necessitate potentially dozens of IPC calls per render).

This PR _does_ keep the old way of executing code around, and in fact that's still the code that'll ultimately be used to handle requests. The old code and new code are executed at the same time with a library like [GitHub's scientist](https://github.com/github/scientist), and the results are compared to ensure that they match. In general, this should work well, but it's likely to fail if people use `Math.random` instead of `PrairieRandom` (the former is non-deterministic, but the latter is generally seeded with a random value that's unique to a variant).

This parallel-run code is instrumented with Honeycomb, so we can easily track results:

Screen Shot 2021-10-12 at 9 58 09 AM

One thing that may be less-than-desirable with the current implementation is that there's no limit to the max number of subprocesses that can be active at once. I think this should be fine, but it's something I wanted to call out - I'm happy to try to add that if folks desire.

Will also fix #3846.


question-servers/calculation.js
```diff
@@ -1,221 +1,195 @@
...
+        const dataMismatched =
+          controlHasData && candidateHasData && !_.isEqual(controlData, candidateData);
+
+        if (errorsMismatched || dataMismatched) {
+          span.setAttributes({
+            'experiment.result': 'mismatched',
+            'experiment.control': observationPayload(controlError, controlResult),
+            'experiment.candidate': observationPayload(candidateError, candidateResult),
+          });
+          span.setStatus({
```
nwalters512, line 121: @mwest1066 now that Sentry is a part of our stack, I'm wondering if perhaps we should use that instead to report any mismatches instead of Honeycomb? Its lossless retention of errors seems to be what we want, without any need for messing around with sample rates.

python/zygote.py
```diff
@@ -125,6 +126,45 @@ def worker_loop():
...
+                # Node generally boots up very quickly, so this should be fine.
+                result = subprocess.run(
+                    [
+                        "node",
+                        "./question-servers/calculation-worker.js",
+                    ],
+                    cwd=cwd,
+                    capture_output=True,
+                    # TODO: consider manual check of `result.returncode != 0` instead
+                    # check=True,
```
nwalters512, line 145:
```suggestion
```

lib/file-paths.js
```diff
@@ -1,125 +1,138 @@
-var ERR = require('async-stacktrace');
-var fs = require('fs');
-var path = require('path');
+// @ts-check
```
mwest1066, line 1: Nice!

question-servers/calculation.js
mwest1066: Agreed! I think that Sentry is just a better fit for error reporting than HC.

question-servers/calculation.js
```diff
@@ -1,186 +1,164 @@
...
+const { experimentAsync } = require('tzientist');
+const _ = require('lodash');
+const { trace, context, SpanStatusCode } = require('@opentelemetry/api');
 
-var error = require('../prairielib/lib/error');
-var chunks = require('../lib/chunks');
-var filePaths = require('../lib/file-paths');
-var requireFrontend = require('../lib/require-frontend');
+const calculationInprocess = require('./calculation-inprocess');
+const calculationSubprocess = require('./calculation-subprocess');
```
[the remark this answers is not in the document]
mwest1066: I'm very happy if we can close out this thread by using Sentry instead 😅

question-servers/calculation-worker.js
```diff
@@ -0,0 +1,202 @@
+// @ts-check
+// This is meant to be invoked from `./question-servers/calculation.js`. It
+// serves to isolate code from the main process that's handling requests.
```
mwest1066, line 3: Does this comment need updating? We are actually invoking this from the python zygote now?

question-servers/calculation-worker.js
```diff
@@ -0,0 +1,202 @@
...
+      didResolve = true;
+      resolve(null);
+    });
+  });
+}
+
+if (require.main === module) {
+  // Redirect `stdout` to `stderr` so that we can ensure that no
+  // user code can write to `stdout`; we need to use `stdout` to send results
+  // instead.
```
mwest1066, line 141: I think that lots of legacy v2 questions write junk to stderr/stdout, which will now trigger an issue on the question, I think? I'm 100% happy that we will get this junk out of our syslog, but I'm a bit nervous about "breaking" v2 questions in courses that have been "working" for years. I'm almost tempted to silently drop both stderr and stdout here in the interest of just keeping things working?

mwest1066 (approved):

question-servers/calculation-worker.js
nwalters512: Yep! Good call; I'll change this.

question-servers/calculation-worker.js
nwalters512: That's fair, yeah. This would mean that there's no easy way for instructors to debug their questions, though I suppose that's ok? The v2 question experience doesn't seem to be worth optimizing at this point.

echuber2: CS 421 has a fair amount of v2 content with sporadic bugs that surfaced in just the past year. We already ran in to dead ends trying to sort that out, so I was thinking it made the most sense for them to migrate content to v3.

nwalters512: Upon further reviewing the code, the "error if there's output" piece is actually part of the v3 question processing code. We actually already just ignore the output property on the code caller result (see `question-servers/calculation-subprocess.js#callFunction`), so there's no code change to make here. I'll add a comment explaining that though.

This still means that output will no longer be user-facing; I still think that's ok though.

question-servers/calculation.js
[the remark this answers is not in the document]
nwalters512: Yep - I switched to Sentry in 00aaba2!

question-servers/calculation-subprocess.js
```diff
@@ -0,0 +1,224 @@
...
+      stdio: ['pipe', 'pipe', 'pipe', 'pipe'], // stdin, stdout, stderr, and an extra one for data
+      timeout: config.questionTimeoutMilliseconds,
+    });
+
+    child.stdout.setEncoding('utf-8');
+    child.stderr.setEncoding('utf-8');
+    // @ts-expect-error
+    child.stdio[3].setEncoding('utf-8');
+
+    child.stdin.write(JSON.stringify(callData));
```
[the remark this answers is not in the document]
nwalters512: I ultimately switched to invoking the worker from the Python zygote so that we can seamlessly use containerization.

question-servers/calculation.js
nwalters512: I made the switch in 00aaba2.

question-servers/calculation-worker.js
nwalters512: Resolved in d1a147c.

question-servers/calculation-worker.js
nwalters512: I documented this in 00aaba2.

Issue #4938, opened 2021-10-08
pl-order-blocks did not allow tiles to be dragged on iPad...


cheeren:
Two students reported (and I saw) that they could not drag the blocks when using their ipad. Strangely, one of the students DID drag ONE block, but then could not drag more.  I assume they were using safari, but didn't ask. (I could do so.)

SethPoulsen:
Thanks for reporting this Cinda! I'm going to close this now as a duplicate: https://github.com/PrairieLearn/PrairieLearn/issues/4912

Good news is, I already have a fix here: https://github.com/PrairieLearn/PrairieLearn/pull/4954

Issue #4912, opened 2021-10-03
pl-order-blocks without a mouse (keyboard accessibility)


jonatanschroeder:
When answering on a touch only device (e.g., phone) the pl-order-blocks element doesn't provide a way to provide an answer. I imagine a similar issue could also affect accessibility. I'm wondering if something like small buttons to add entries to the answer and to reorder them may address both of these issues.

SethPoulsen:
I have a fix for touch screens here: https://github.com/PrairieLearn/PrairieLearn/pull/4954

This issue should remain open because Accessibility for people with low vision who navigate only with a keyboard is still a problem and it is something that I would like to support when I get around to it.

nwalters512:
A possible starting point: https://www.smashingmagazine.com/2018/01/dragon-drop-accessible-list-reordering/

There's some good discussion happening on [this Slack thread](https://prairielearn.slack.com/archives/C0XBTPZMZ/p1634661735378500) around this; lifting a few thoughts into this issue for when free Slack inevitably wipes those messages from history 😢:

* We could probably use left/right and MOD+left/MOD+right to control indentation and moving items between the choices and answer zones, where MOD is shift or command or some other modifier. It's not clear which keys would be better suited for indentation vs. moving between dropzones.
* We probably want focus to stay with items as they're moved around.

SethPoulsen:
It seems some other people have hacked accessibility features onto the jqueryui sortable, which is what we're using: 

https://codepen.io/bartz/pen/rrgyjw

https://stackoverflow.com/questions/16844616/jqueryui-sortable-with-keyboard

echuber2:
If touch has been resolved in https://github.com/PrairieLearn/PrairieLearn/pull/4954 then maybe we should rename this issue to focus on accessibility. (I couldn't find it until Seth gave me the link.)

echuber2:
Discord has an MIT-licensed library for accessible drag-and-drop (if I'm reading this correctly): https://github.com/discord/react-dnd-accessible-backend

SethPoulsen:
@echuber2 Looks cool thanks for the reference! I obviously can't use that since it got a million dependencies, but good to know it exits. I also was unaware of the new HTML 'draggable' attribute & friends. I'll be sure to look into that when I get to making sure the elements are properly annotated for screen readers.

PR #4954, opened 2021-10-09
Added touch support for pl-order-blocks

SethPoulsen:
Resolves https://github.com/PrairieLearn/PrairieLearn/issues/4938 and part of https://github.com/PrairieLearn/PrairieLearn/issues/4912. 

It would be nice if someone else could test this for me. Its not too bad, just: 

1. Clone my branch
2. Run `npm ci` to pull the new dependency
3. Run prairielearn on your machine, overriding the PrairieLearn folder in the docker container with the new version
4. Make sure you have a phone/tablet on the same network as the computer you are running PL on, then open `http://<your computer's IP address>:3000/pl/course_instance/3/instructor/question/95/preview` to open the example question and try it out.

edit: I only tested on my Android phone, if someone could test with iOS, that would be nice though since I'm just using a standard node package I don't think it should be an issue.

### DIFF
--- a/elements/pl-order-blocks/pl-order-blocks.py
+++ b/elements/pl-order-blocks/pl-order-blocks.py
@@ -230,7 +230,7 @@ def render(element_html, data):
             return ''  # external grader is responsible for displaying results screen
 
         student_submission = ''
-        score = 0
+        score = None
         feedback = None
 
         if answer_name in data['submitted_answers']:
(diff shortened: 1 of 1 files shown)


SethPoulsen:
@nwalters512 I added the new dependency by `npm i --save jquery-ui-touch-punch` and then I committed the changes that the command made to `package.lock.json`, was that the right thing to do? I've never had to deal with this thing before.

SethPoulsen:
Thanks @jonatanschroeder 

Ya, that happened on my phone too. We should make a new issue for that. I think its probably fine on tablets.

SethPoulsen:
I made a new issue for that here: https://github.com/PrairieLearn/PrairieLearn/issues/4955

To be clear, I did test with my phone _before_ making the change to support touch, and this layout issue is definitely unrelated/not cause by this change.

jonatanschroeder:
> I made a new issue for that here: #4955 (I can upload my own screenshot if you object to me using yours)
> 
> To be clear, I did test with my phone _before_ making the change to support touch, and this layout issue is definitely unrelated/not caused by this change.

Go for it, it doesn't have anything revealing (part of one of my questions, but nothing that gives away any information).

PR #4963, opened 2021-10-12
Prevent negative margin of dropped blocks (#4926)

### LINKED ISSUE #4926: pl-order-blocks UI sometimes allows negative indentation
Seems to be an intermittent problem. Need to test more to see when exactly it occurs.

ataraxie:
This should fix #4926 (the problem with the negative margin of some blocks). The actual bugfix, as I described it in the issue, is in lines 8-12.

With this PR, I also suggest a few minor improvements alongside the fix, but it's also ok to just fix the bug in line 8-12. I changed `var`s to `let` or `const` because `const` had already been used in this script, so I assume that the browsers that PL supports are ES6 compatible (as far as I can read from the docs non-edge IE is not supported). A few very minor things have also just been fixed by my IDE. I'll leave some more comments along the code.

### DIFF
--- a/elements/pl-order-blocks/pl-order-blocks.js
+++ b/elements/pl-order-blocks/pl-order-blocks.js
@@ -2,15 +2,10 @@
 
 const TABWIDTH = 50;    // defines how many px the answer block is indented by, when the student
                         // drags and indents a block
-let MAX_INDENT = 4;     // defines the maximum number of times an answer block can be indented
+var MAX_INDENT = 4;     // defines the maximum number of times an answer block can be indented
 
 function set_max_indent(event) {
-    // We read the indent from the attribute. This will be a string, so we need to parse it.
-    const potentialNewIndent = parseInt(event.target.getAttribute('indent'));
-    // It should only be used if it is a positive intege
-    if (Number.isInteger(potentialNewIndent) && potentialNewIndent >= 0) {
-        MAX_INDENT = potentialNewIndent;
-    }
+    MAX_INDENT = event.target.getAttribute('indent');
 }
 
 function check_block(event, ui) {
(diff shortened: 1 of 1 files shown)


ataraxie:
Sorry, my commits appear with author `fgrund` (my bitbucket username). I'll fix my local Git config before I commit next time.

nwalters512:
> I'm not sure whether this file is actually linted. There are a lot of syntax violations in there...

This file should indeed be linted, so I'd be surprised if there were syntax errors here. Could you let me know what they are?

> I was already wondering about multiple conventions that seem "violated" in this file.

We've unfortunately relied on manual enforcement of conventions by humans for many things, and depending on who reviews a particular PR, they may or may not pick up on different things. This is obviously less than ideal, and as I'm getting back into PL development after a hiatus of several years, it's very much on my list to automate as much enforcement of code standards as possible. There's just a lot to do, and I haven't gotten around to it yet - I appreciate your patience in the meantime!

> I'm just not really sure what I should do with this PR.

I think the best thing to do here is to make the minimal set of changes to fix the bug described in https://github.com/PrairieLearn/PrairieLearn/issues/4926. I'm generally a big fan of drive-by refactoring + cleanup, but I realize I commented on things that you didn't change and it's not reasonable to ask you to modify them when you're just trying to fix a bug 🙂 I'd ask that you back out the jQuery changes, change only what's needed to fix the bug (and possibly rename non-constants to not use SCREAMING_SNAKE), and we'll get this merged. If you'd like to contribute additional cleanup in another PR, you're more than welcome too, but I'd also be happy to do so myself.

Thank you very much for contributing a fix - I'm hopeful that we'll be able to improve the experience for contributors soon!

PR #4978, opened 2021-10-13
chore: Convert to Yarn

nwalters512:
This PR switches us from npm to Yarn. It specifically uses Yarn v1, which has installation behavior that matches that of npm.

We do not currently use any of the features that Yarn provides, but this change will set us up to be able to use Yarn workspaces to better manage our monorepo development workflow.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,16 +33,8 @@ jobs:
       run: docker exec test_container /PrairieLearn/docker/lint_js.sh
     - name: Run the Python linter
       run: docker exec test_container /PrairieLearn/docker/lint_python.sh
-    - name: Run the PrairieLib linter
-      run: docker exec test_container /PrairieLearn/docker/lint_prairielib.sh
-    - name: Run the grader_host linter
-      run: docker exec test_container /PrairieLearn/docker/lint_grader_host.sh
     - name: Run typechecker
       run: docker exec test_container /PrairieLearn/docker/typecheck.sh
-    - name: Run the PrairieLib tests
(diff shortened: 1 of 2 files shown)


mwest1066:
I built a new `plbase` image from this branch, but inside it I can't start postgres (which causes the build of the main `prairielearn` image to fail). Steps to reproduce:
```
$ cd images/plbase
$ docker build -t plbase .
$ docker run -it --rm plbase /bin/bash
# su postgres -c "pg_ctl --silent --pgdata=/var/postgres --log=/var/postgres/postgresql.log start"
pg_ctl: could not start server
Examine the log output.
# cat /var/postgres/postgresql.log
2021-10-14 22:25:31.317 UTC [12] LOG:  listening on IPv4 address "127.0.0.1", port 5432
2021-10-14 22:25:31.317 UTC [12] LOG:  could not bind IPv6 address "::1": Cannot assign requested address
2021-10-14 22:25:31.317 UTC [12] HINT:  Is another postmaster already running on port 5432? If not, wait a few seconds and retry.
2021-10-14 22:25:31.321 UTC [12] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
2021-10-14 22:25:31.324 UTC [12] LOG:  listening on Unix socket "/tmp/.s.PGSQL.5432"
2021-10-14 22:25:31.335 UTC [12] FATAL:  could not open log file "log/postgresql-Thu.log": No such file or directory
2021-10-14 22:25:31.338 UTC [12] LOG:  database system is shut down
```

Has anyone else seen anything like this?

nwalters512:
@mwest1066 looks like the tests pass, so we were able to successfully build a Docker container and use the database in CI.

PR #4989, opened 2021-10-15
refactor: Use async/await for database library code

nwalters512:
This PR refactors `prairielib/sql-db` to use `async`/`await` for all core implementations, and uses `util.callbackify` to provide callback-based versions where needed. This should result in the same behavior as before, but it makes it so that any code using only the async variant doesn't use `callback`/`ERR` at all, thus retaining the nice stack traces that Node now provides for async functions.


nwalters512:
Welp, the tests are hanging forever without doing anything - some callback/promise is not getting triggered/resolved! Looking into this now.

PR #5031, opened 2021-10-25
enhancements for tmux script in light of yarn, local mounting

echuber2:
I've made several additions to the optional tmux entrypoint script in `tools/`. Now it can optionally invoke yarn upon load and re-chown generated files in a dev mount (e.g. `node_modules`...) back to the local user's uid:gid. Also, kill lingering workspaces upon exit.


nwalters512 (approved):
This is mostly a rubber stamp; I don't use this tool so I can't comment on whether this is working in a sensible way or not. I assume you've tested it appropriately.

This feels like a lot of complexity for a tool that is probably not very widely used.

Issue #4960, opened 2021-10-11
Partial Credit Grading for Proof Blocks


SethPoulsen:
I've had a lot of conversations with people on how to do partial credit for the `dag` grading mode of `pl-order-blocks`, especially with @jeffgerickson, @jasonxia17, and @mwest1066. 

Now that I feel like we've reached some sort of consensus on what the best thing to do is moving forward, I'm going to give it an outline here so its in writing--both so people can give feedback on it, and so that I have a record of what to do once I have the time to do it.

First things first: **How should students be assigned partial credit?**
The original prototype code for proof blocks would give partial credit based on where the first wrong line of the proof was. For example, if the first 3 lines of your proof were correct, and the correct proof had 10 lines, you would get 3/10. Its not very hard to see this doesn't always work well. For example, what if a student has the first line of the Proof wrong, but every other line was correct? Shouldn't that be almost full credit?  

Well, yes, it should. Instead, we've concluded that the best way to assign partial credit is based on the **mimimum edit distance from the solution they give to _some_ correct Proof** (i.e. to some topological sort of the blocks).

We are also hoping that once we can have a concrete edit sequence to some correct solution, this can also be used to give students better feedback, but feedback is not the main focus (yet), so we'll just leave that out for now.

The next question to answer is **How do we actually compute this?**
There is a relatively straightforward exponential algorithm for finding all topological sorts, as implemented in the 
"[all_topological sorts](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.dag.all_topological_sorts.html)" function from the `networkx` python library. 

We can then, calculate the edit distance (we'll probably use Longest Common Subsequence) to each topological sort from the student input we can probably just use something built in in Python, ~~e.g. the `SequenceMatcher` from [difflib](https://docs.python.org/3/library/difflib.html).~~

It seems to be an open problem whether or not there is a better algorithm for finding the edit distance from a given sequence to some topological sort, but after trying to find one, and discussing a bit, we decided that it would be best to just try out the naive algorithm and see what happens. 

Many Proof Blocks questions use as many as 10 blocks, which could be a problem with an exponential time algorithm, _but_
the exponential algorithm for finding all topo sorts of a DAG hits its worst case for very wide DAGs, but most of the dozens of Proof Blocks problems we have seen used in courses so far seem to be very narrow but tall DAGs, so we think it is very unlikely that we will hit the worst case.

[...]

PR #5066, opened 2021-11-01
pl-order-blocks: Refactor DAG grading code

SethPoulsen:
Code is now much shorter and simpler. This refactor also helps prepare
for implementing #4960.

Tested on over 9000 submissions from prior courses to ensure the
refactor didn't change any behavior.


elements/pl-order-blocks/dag_checker.py
```diff
@@ -1,72 +1,50 @@
...
+        elif group_id is None and cur_group is not None:
+            return i
+        elif group_id is not None and cur_group is not None:
+            if group_id == cur_group:
+                cur_group_size += 1
+                if cur_group_size == group_sizes[cur_group]:
+                    cur_group = None
+                    cur_group_size = None
+            else:
+                return i
```
SethPoulsen, line 32: an alternative way to write this for loop would be: 

```
    for i, node in enumerate(order):
        group_id = group_belonging.get(node)
        if group_id == cur_group:
            if group_id is None:
                continue
            cur_group_size += 1
            if cur_group_size == group_sizes[cur_group]:
                cur_group = None
                cur_group_size = None
        else:
            if group_id is not None and cur_group is None:
                cur_group = group_id
                cur_group_size = 1
            else:
                return i
```

I don't really love either way, I'm definitely open to suggestions on how to make this prettier.

SethPoulsen:
@nwalters512 What should I make of this build failing during `yarn install`? https://github.com/PrairieLearn/PrairieLearn/runs/4070792812?check_suite_focus=true 

I didn't touch anything having to do with dependencies, or even javascript at all for that matter.

elements/pl-order-blocks/dag_checker.py
```diff
@@ -1,72 +1,50 @@
...
-    def get_current_group_blocks(self):
-        return self.current_group_len
+from collections import Counter
+import networkx as nx
+
+
+def check_topological_sorting(order, graph):
+    seen = set()
+    for i, node in enumerate(order):
+        if node is None or not all([u in seen for (u, _) in graph.in_edges(node)]):
```
nwalters512, line 8: Nit: it's more efficient to use a generator expression instead of a list comprehension here:
```suggestion
        if node is None or not all(u in seen for (u, _) in graph.in_edges(node)):
```

nwalters512 (approved):
I'd really appreciate some docstrings/comments to help explain what's going on here - at the very least, docstrings explaining the arguments and return values for each function here.

> Tested on over 9000 submissions from prior courses to ensure the
refactor didn't change any behavior.

Ack, though this kind of self-contained module is a great candidate for unit tests to give us increased confidence in the code now and in changes in the future! We have a single piece of prior art for testing Python code in PL - see [the test file](https://github.com/PrairieLearn/PrairieLearn/blob/184a711c76b260eea422d0bb9eb00b83403e21cb/question-servers/freeformPythonLib/prairielearn_test.py) and [the part of the Makefile that invokes it](https://github.com/PrairieLearn/PrairieLearn/blob/184a711c76b260eea422d0bb9eb00b83403e21cb/Makefile#L32). I won't block on adding tests, but I would greatly appreciate the addition of a few.

SethPoulsen:
Yes I would love to add tests! I actually didn't know you had a way to do freestanding python unit tests, I thought it all had to be done through the element testing framework (which I also have a TODO item to improve for this element...). Should I just add a new line to the makefile for the test file I make? And should I put the new test file also in `freeformPythonLib`?

On Docstrings, I'm not sure what kinds of details/level of details you would like, can you point me toward some examples in the codebase?

nwalters512:
> Should I just add a new line to the makefile for the test file I make?

I think that's the best solution for now. We should really have better infra for Python tests - ideally we'd just use pytest and have it run every `*_test.py` file as a test, but that's a problem for another PR 🙂 

> And should I put the new test file also in `freeformPythonLib`?

Nope, the only reason that test is there is because it's testing `prairielearn.py`, which is directly adjacent to it. You should do the same thing with this test file, i.e. locate it at `elements/pl-order-blocks/dag_checker_test.py`.

> On Docstrings, I'm not sure what kinds of details/level of details you would like, can you point me toward some examples in the codebase?

To be perfectly honest: I don't think I can point to any great examples in the current codebase 😅 documenting Python code is something we haven't historically been particularly diligent about, and that's something I've trying to change now. I'm partial to the Sphinx format (https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html). You can also consider using type hints (https://docs.python.org/3/library/typing.html) to annotate your code - while we don't have any tooling to actually typecheck Python at the moment, they're still a valuable form of documentation!

SethPoulsen:
Ah yes, I've actually been meaning to figure out python type annotations, just haven't gotten around to it yet. Having tooling for that would be awesome (typescript too of course, but on thing at a time, right?). 

Sometime later this week I should have some more time to work on this.

nwalters512:
> Having tooling for that would be awesome (typescript too of course, but on thing at a time, right?).

Absolutely - this is very much on my list, though not quite currently at the top of it 🙂

SethPoulsen:
@nwalters512 I just tried to add type checking for the file, but I broke the build. Any idea what I did wrong? I haven't had to touch `plbase` before so I'm not really sure.

SethPoulsen:
I get the same error when running it locally--for some reason rebuilding `plbase` makes it so it can't find `flake8` anymore, very odd.

SethPoulsen:
Ok wow, that was dumb, I used `=` instead of `==` when specifying the `mypy` version so the whole pip install failed. Lets try again...

SethPoulsen:
I opened a new PR to make the `plbase` build fail if the installation script doesn't work. This way the error actually shows up in the correct place in the future: https://github.com/PrairieLearn/PrairieLearn/pull/5075

SethPoulsen:
@nwalters512 I'm happy for this to be merged now if you are. 

Further improvements (discussed here and elsewhere) will come later.

elements/pl-order-blocks/dag_checker.py
```diff
@@ -1,72 +1,60 @@
...
+
+def grade_dag(order: list[str], depends_graph: Mapping[str, list[str]], group_belonging: Mapping[str, Optional[int]]) -> tuple[int, int]:
+    """In order for a student submission to a DAG graded question to be deemed correct, the student
+    submission must be a topological sort of the DAG and blocks which are in the same pl-block-group
+    as one another must all appear contiguously.
+    :param order: the block ordering given by the student
+    :param depends_graph: The dependency graph between blocks specified in the question
+    :param group_belonging: which pl-block-group each block belongs to, specified in the question
+    :return: length of list that meets both correctness conditions, starting from the beginning,
+    index of first wrong (-1 if none wrong)
```
nwalters512, line 45:
```suggestion
    :param group_belonging: which pl-block-group each block belongs to, specified in the question
    :return: a tuple containing the length of list that meets both correctness conditions,
        starting from the beginning, and the index of first wrong (-1 if none wrong)
```

This is just a question (not necessarily asking for a code change), but is returning both values necessary? Seems like the caller could use a single value - if `ret == len(order)`, full credit, else use `ret` as the index of the first wrong.

elements/pl-order-blocks/dag_checker.py
```diff
@@ -1,72 +1,60 @@
...
-            first_wrong = i
-            break
-
-    return len(answer_state.valid), first_wrong
+from collections import Counter
+from typing import Mapping, Optional
+import networkx as nx  # type: ignore
+
+
+def check_topological_sorting(order: list[str], graph: nx.DiGraph) -> int:
```
nwalters512, line 6: Could you add a quick docstring here as well? It looks like this just returns the index of the first element in `order` that is not topologically sorted, but I'd like that to be explicitly documented.

elements/pl-order-blocks/dag_checker.py
```diff
@@ -1,72 +1,60 @@
...
+    seen = set()
+    for i, node in enumerate(order):
+        if node is None or not all(u in seen for (u, _) in graph.in_edges(node)):
+            return i
+        seen.add(node)
+    return len(order)
+
+
+def check_grouping(order: list[str], group_belonging: Mapping[str, Optional[int]]) -> int:
```
nwalters512, line 15: Would appreciate a docstring here too.

nwalters512 (approved):
Approval stands - if you'd rather handle my comments elsewhere, I'll go ahead and merge! Thanks for taking the time to add tests and docs, future maintainers will thank you 🙏

elements/pl-order-blocks/dag_checker.py
SethPoulsen: No, its definitely not necessary. The main reason that I kept it this way for now is that I wanted to keep this refactor contained to just this file. 

The return value here is definitely going to change when I implement partial credit (coming soon!) anyway.

elements/pl-order-blocks/dag_checker.py
SethPoulsen: done.

elements/pl-order-blocks/dag_checker.py
SethPoulsen: done.

elements/pl-order-blocks/dag_checker.py
SethPoulsen: No one has proposed a simpler way to write this code, so I'm going to resolve this. 

Maybe I'll think something up later.

elements/pl-order-blocks/dag_checker.py
```diff
@@ -1,72 +1,71 @@
...
-    def set_valid(self, stmt_id):
-        self.valid.add(stmt_id)
-        if self.current_group is not None:
-            self.current_group_len += 1
+def check_grouping(order: list[str], group_belonging: Mapping[str, Optional[int]]) -> int:
+    """
+    :param order: candidate solution
+    :param group_belonging: group that each block belongs to
+    :return: index of first element breaking condition that members of the same group must be
+    adjacent, or length of list if they all mee
```
nwalters512, line 25: typo?

SethPoulsen: ya, fixed

SethPoulsen:
@nwalters512 I've merged this to the latest master, still all good. Feel free to merge whenever.

PR #5070, opened 2021-11-02
Encapsulate and format JavaScript for pl-order-blocks element

### LINKED ISSUE #4980: Clean up pl-order-blocks.js to comply to best practices
the file pl-order-blocks.js doesn't conform well to coding best practices and PrairieLearn conventions and could use some cleanup. 

Specifically: 

- pollutes the global namespace with `TABWIDTH`, `MAX_INDENT`, and with all the functions it uses. It would be better to only add one object to the global namespace and attach all methods and variables defined in the file to that
- using all caps for `MAX_INDENT` even thought it isn't a constant anymore (it used to be)
- uses snake_case instead of camelCase for function names

Feel free to comment anything else we should clean up there

SethPoulsen:
fix #4980

### DIFF
--- a/elements/pl-order-blocks/pl-order-blocks.js
+++ b/elements/pl-order-blocks/pl-order-blocks.js
@@ -31,11 +31,8 @@ window.PLOrderBlocks = function (uuid, options) {
       }
     }
 
-    var textfieldName = event.target.getAttribute('name');
-    if (textfieldName === null) {
-      return;
-    }
-    textfieldName = '#' + textfieldName + '-input';
+    var textfieldName = '#' + event.target.getAttribute('name');
+    +'-input';
     $(textfieldName).val(JSON.stringify(studentAnswers));
   }
 
--- a/elements/pl-order-blocks/pl-order-blocks.js
+++ b/elements/pl-order-blocks/pl-order-blocks.js
@@ -22,12 +22,12 @@ window.PLOrderBlocks = function (uuid, options) {
           answerIndent = Math.round(answerIndent / TABWIDTH); // get how many times the answer is indented
         }
 
-        var answerJson = {
+        var answer = {
           inner_html: answerText,
           indent: answerIndent,
           uuid: uuid,
         };
-        studentAnswers.push(answerJson);
+        studentAnswers.push(answer);
       }
     }
 
--- a/elements/pl-order-blocks/pl-order-blocks.js
+++ b/elements/pl-order-blocks/pl-order-blocks.js
@@ -9,8 +9,7 @@ window.PLOrderBlocks = function (uuid, options) {
   let dropzoneElementId = '#order-blocks-dropzone-' + uuid;
 
   function setAnswer(event) {
-    // We only care about when this function is fired
-    // from an answer dropzone, aka dropzones with yellow backgrounds
+    // We only care about when this function is fired from an answer dropzone
     var domObjs = $(dropzoneElementId).children();
     var studentAnswers = [];
     for (var i = 0; i < domObjs.length; i++) {
--- a/elements/pl-order-blocks/pl-order-blocks.js
+++ b/elements/pl-order-blocks/pl-order-blocks.js
@@ -4,20 +4,21 @@ window.PLOrderBlocks = function (uuid, options) {
   const TABWIDTH = 50; // defines how many px the answer block is indented by, when the student
   // drags and indents a block
   let maxIndent = options.maxIndent; // defines the maximum number of times an answer block can be indented
+  let enableIndentation = options.enableIndentation;
 
   let optionsElementId = '#order-blocks-options-' + uuid;
   let dropzoneElementId = '#order-blocks-dropzone-' + uuid;
 
   function setAnswer(event) {
-    var domObjs = $(dropzoneElementId).children();
+    var answerObjs = $(dropzoneElementId).children();
     var studentAnswers = [];
-    for (var i = 0; i < domObjs.length; i++) {
-      if (!$(domObjs[i]).hasClass('info-fixed')) {
-        var answerText = domObjs[i].getAttribute('string');
-        var uuid = domObjs[i].getAttribute('uuid');
+    for (var i = 0; i < answerObjs.length; i++) {
+      if (!$(answerObjs[i]).hasClass('info-fixed')) {
+        var answerText = answerObjs[i].getAttribute('string');
+        var uuid = answerObjs[i].getAttribute('uuid');
         var answerIndent = null;
-        if (domObjs[i].parentElement.classList.contains('enableIndentation')) {
-          answerIndent = parseInt($(domObjs[i]).css('marginLeft').replace('px', ''));
+        if (enableIndentation) {
+          answerIndent = parseInt($(answerObjs[i]).css('marginLeft').replace('px', ''));
           answerIndent = Math.round(answerIndent / TABWIDTH); // get how many times the answer is indented
         }
 
@@ -35,10 +36,7 @@ window.PLOrderBlocks = function (uuid, options) {
   }
 
   function calculateIndent(ui, parent) {
-    if (
-      !parent[0].classList.contains('dropzone') ||
-      !parent[0].classList.contains('enableIndentation')
-    ) {
+    if (!parent[0].classList.contains('dropzone') || !enableIndentation) {
       // don't indent on option panel or solution panel with indents explicitly disabled
       return 0;
     }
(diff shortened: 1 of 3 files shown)


SethPoulsen:
@nwalters512 I left the helper functions here as local functions inside the constructor, so that they are not exposed as public methods on the object when they don't need to be. However, I noticed that some other elements (e.g. PLFileUpload) attach function to the object prototype so that they are public, even if they are only used internally. 

Is there an official convention we are supposed to be following?

elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -1,4 +1,14 @@
 {{#question}}
+<script>
+$(function() {
+  var UUID = "{{uuid}}";
+
+  new window.PLOrderBlocks(UUID, {
+      maxIndent: {{max_indent}}
+  });
+});
+</script>
```
nwalters512, line 10: Nit: indentation is all over in this file. Let's stick to either 2 or 4 spaces.

elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -8,27 +18,27 @@
                     </a>
                 </li>
             </ul>
-            <ul id="{{answer_name}}-options" name="{{answer_name}}" class="pl-order-blocks-connected-sortable {{inline}}" indent={{max_indent}}>
+            <ul id="order-blocks-options-{{uuid}}" name="{{answer_name}}" class="pl-order-blocks-connected-sortable {{inline}}" indent={{max_indent}}>
```
nwalters512, line 21: I believe the `indent` attribute is no longer in use and can be removed.
```suggestion
            <ul id="order-blocks-options-{{uuid}}" name="{{answer_name}}" class="pl-order-blocks-connected-sortable {{inline}}">
```

elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -8,27 +18,27 @@
...
-            </div>
+        <div class="column {{dropzone_layout}}">
+            <ul class="list-group-item list-group-item-primary header">
+                <li><strong>{{solution-header}}</strong>
+                    <a role="button" class="btn btn-light btn-sm" data-toggle="popover" data-html="true" title="Order Blocks" data-content="{{help_text}}" data-trigger="focus" tabindex="0">
+                    <i class="fa fa-question-circle" aria-hidden="true"></i>
+                </a>
+                </li>
+            </ul>
+            <ul id="order-blocks-dropzone-{{uuid}}" name="{{answer_name}}" class="list-group pl-order-blocks-connected-sortable dropzone {{check_indentation}}" indent={{max_indent}}>
```
nwalters512, line 35: I believe the `indent` attribute is no longer in use and can be removed.
```suggestion
            <ul id="order-blocks-dropzone-{{uuid}}" name="{{answer_name}}" class="list-group pl-order-blocks-connected-sortable dropzone {{check_indentation}}">
```

elements/pl-order-blocks/pl-order-blocks.js
```diff
@@ -1,112 +1,90 @@
...
-function check_block(event, ui) {
-  var block_parent_name = $(ui.item.parent()[0]).attr('name');
-  var block_destination_name = event.target.getAttribute('name');
-  if (block_parent_name !== block_destination_name) {
-    return false;
-  }
-  return true;
-}
+  function setAnswer(event) {
+    var domObjs = $(dropzoneElementId).children();
```
nwalters512, line 12: Can we give this a more meaningful name than `domObjs`?

nwalters512 (approved):
🙌 This is a great start! I'm happy to accept this as-is, though there are some other changes that we could make as well to ensure this element is adhering to all our conventions and best practices (I'm happy to push them off to another PR though):

* For attributes like `string` and `uuid`, we should use prefix them with `data-`, e.g. `data-string` or `data-uuid` - this ensures that they're actually valid HTML.
* We currently check if indentation is enabled with an `enableIndentation` class - that's exactly the kind of thing we should instead be providing via the `options` argument to `new PLOrderBlocks`.
* A lot of the CSS includes global selectors like `div.row`, `div.column`, etc. This is _really_ bad, as this means that this element can negatively impact the styling of any other HTML on the page. As much as possible, we should limit any CSS to only apply to markup that's rendered by this element. A few approaches to this are to include the element name in classes (e.g. `.pl-order-blocks-row` instead of just `.row`) or to wrap the whole element in a `<div>` with an `id` and target that specific ID in all selectors (see https://github.com/PrairieLearn/PrairieLearn/blob/184a711c76b260eea422d0bb9eb00b83403e21cb/elements/pl-file-upload/pl-file-upload.css for an example).

> edit: I'm of the opinion that we should hide all the functions we can and only leave things publicly accessible if necessary, but I also don't necessarily want to fight convention.

I'm very much on board like this - I don't think there's a good reason that I wrote the other elements this way to begin with.

elements/pl-order-blocks/pl-order-blocks.js
```diff
@@ -1,112 +1,90 @@
...
+        // when the user stops interacting with the list
+        let leftDiff = calculateIndent(ui, ui.item.parent());
+        ui.item[0].style.marginLeft = leftDiff + 'px';
+        setAnswer(event);
+      },
+    });
+  }
+
+  initialize(optionsElementId, dropzoneElementId);
+  initialize(dropzoneElementId, optionsElementId);
```
nwalters512, line 86: Is it necessary to `initialize(...)` twice? In the old code, it looks like a single call to `$(...).sortable(...)` was sufficient.

elements/pl-order-blocks/pl-order-blocks.mustache
SethPoulsen: Nice catch.

elements/pl-order-blocks/pl-order-blocks.mustache
SethPoulsen: nice catch

elements/pl-order-blocks/pl-order-blocks.js
SethPoulsen: Well, it was initializing all items on the page in one call based on the class, and connecting all of them to each other! so I'd say this is a huge improvement. Let me check though, I might be able to pass it a list of the 2 IDs instead of calling it twice.

elements/pl-order-blocks/pl-order-blocks.js
SethPoulsen: sure

Frozemint:
This is kind of neat, changing the "visibility" of the functions. I haven't had a chance to test this yet since I'm at school, but so far things looks good. I could definitely look at being smarter when using CSS rules and not use rules that are too strong.

elements/pl-order-blocks/pl-order-blocks.mustache
SethPoulsen: LOL I think its actually 4 throughout the file except for right here where I pasted from somewhere else and didn't fix it. 

Side note, are we going to have auto-formatting on mustache files too in the future?

SethPoulsen:
Lets push the CSS fixes, and the renaming of HTML attributes to another PR. I think I've addressed everything else.

elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -1,4 +1,15 @@
 {{#question}}
+<script>
+$(function() {
+    var UUID = "{{uuid}}";
+
+    new window.PLOrderBlocks(UUID, {
+        maxIndent: {{max_indent}},
+        enableIndentation: {{#check_indentation}}true{{/check_indentation}}{{^check_indentation}}false{{/check_indentation}}
```
SethPoulsen, line 8: This isn't that pretty, is there a better way to do this with moustache? I tried just directly including the variable but it dropped it in capitlized Python-style which obviously didn't work for JavaScript.

SethPoulsen: If there isn't a nice mustache construct for it, I could just do something like
```
"true" if check_indentation else "false"
```
in the python, which is still not that cool but is probably prettier than this.

nwalters512: `"true" if condition else "false"` is the pattern we use in other elements - I think that's totally fine to use here as well.

elements/pl-order-blocks/pl-order-blocks.mustache
nwalters512: Possibly, though that gets trickier because things like `maxIndent: {{max_indent}}` isn't something that a JS formatter should touch. If we were to pass options as `JSON.parse("{{options}}")` instead, that _could_ make auto-formatting via Prettier possible. Food for thought!

elements/pl-order-blocks/pl-order-blocks.js
nwalters512: Ah, I had completely missed the fact that the old selector actually matched multiple things on the page! To be clear, those questions were not necessarily asking for changes, just clarification - I would have been happy with the code as-is and the explanation you provided.

nwalters512 (approved):
Approval stands, though I'll hold off on merging until you update the `enableIndentation` templating.

SethPoulsen:
All done. Go ahead and merge whenever.

PR #5075, opened 2021-11-04
If plbase-install.sh fails, the plbase build should also fail

SethPoulsen:
Otherwise, you could have cases where the plbase image succeeds to
build, but doesn't function properly, as in this build where python
packages weren't installed properly:
https://github.com/PrairieLearn/PrairieLearn/runs/4101152775?check_suite_focus=true


nwalters512 (approved):
😅 If I had a dollar for every time this had bitten us... thanks much!

PR #5077, opened 2021-11-05
Add Python type checking with Pyright

SethPoulsen:
There are 4 main competitors for python type checking: 

1. pyright (Microsoft)
2. pytype (Google)
3. pyre (Facebook)
4. mypy (some Python people)

From my brief googling and trying things out with VS Code, it seems like pyright has best editor support for VS Code and emacs, which are the most used editors among us I believe. Supposedly there is sublime support too, but I don't know the details.

We should probably get more peoples opinions, and do a little more research before deciding to merge this, but it could be good!


SethPoulsen:
there seem to be lots of configuration options: https://github.com/microsoft/pyright/blob/main/docs/configuration.md

eliotwrobson:
Figured I'd chime in, since this is something we've set up in the repo for our course (we're using mypy, I don't think the specific tool you use really matters that much). Adding typechecking has saved us a ton of headache, since it catches a whole class of errors without having to deploy live, and prevents issues with refactoring (we have it configured to run over all question code). Since the backend for PL is getting substantial, I would **strongly** recommend you get this set up as soon as possible.

jasonxia17:
I agree that this is a good idea. Thanks for taking the lead on this Seth! :)

SethPoulsen:
@nwalters512 What are your thoughts on merging this? 

I honestly don't have time at the moment to make the type checking work for all python files, but I would like to have it on the part of the codebase that I'm maintaining, and we could do it incrementally.

I mostly chose Pyright because it seemed to have the best IDE support for what people are using. No one else seems to be weighing in opinions about which checker to use.

tdy:
> it seems like pyright has best editor support for VS Code and emacs

pyright is also what I use in neovim -- [coc-pyright](https://github.com/fannheyward/coc-pyright) has gotten the most traction in the coc.nvim ecosystem.

eliotwrobson:
> I honestly don't have time at the moment to make the type checking work for all python files, but I would like to have it on the part of the codebase that I'm maintaining, and we could do it incrementally.

You can set it up to typecheck the whole codebase now, then incrementally add the type annotations. By default, it shouldn't typecheck any functions that don't have annotations on them.

SethPoulsen:
> By default, it shouldn't typecheck any functions that don't have annotations on them.

I think this is true of mypy, but not pyright. 

Either way, I think it makes more sense to explicitly state what we are checking and what we are not rather than have it done implicitly by lack of annotations.

package.json
```diff
@@ -148,6 +148,7 @@
         "node-fetch": "^2.6.1",
         "nyc": "^15.1.0",
         "prettier": "2.4.1",
+        "pyright": "^1.1.184",
```
nwalters512, line 151: Whoa, it's slightly wild to me that Python tooling is distributed via npm 😂

nwalters512 (approved):
> I honestly don't have time at the moment to make the type checking work for all python files, but I would like to have it on the part of the codebase that I'm maintaining, and we could do it incrementally.

IMO this is absolutely the right way to go about things - I've been doing pretty much exactly that with JSDoc type annotations and `tsc` and that's worked reasonably well so far.

package.json
SethPoulsen: Ya it is funny. It looks like they built it in Typescript for fast & easy VSCode integration, since that was their first priority.

PR #5078, opened 2021-11-07
Remove grading/finish instructions if assessment instance is closed

jonatanschroeder:
As discussed on Slack. This removes from the assessment overview page for closed exams the footer information associated to how to save, grade, create new attempts of finish the assessment, since these options are not available for closed assessments. The "Finish assessment" button would trigger an error if pressed in this case.

Pending: the error that happens when closing an assessment that is already closed can still happen if the user has an open exam, but double-clicks the "Finish assessment" button in the modal. This should probably be changed to a clean ignore operation that returns success without changing anything (maybe logging?), so it returns to the appropriate version of the page without an error.


nwalters512 (approved):

nwalters512:
> Pending: the error that happens when closing an assessment that is already closed can still happen if the user has an open exam, but double-clicks the "Finish assessment" button in the modal. This should probably be changed to a clean ignore operation that returns success without changing anything (maybe logging?), so it returns to the appropriate version of the page without an error.

@jonatanschroeder is there a separate issue to track this yet?

jonatanschroeder:
> > Pending: the error that happens when closing an assessment that is already closed can still happen if the user has an open exam, but double-clicks the "Finish assessment" button in the modal. This should probably be changed to a clean ignore operation that returns success without changing anything (maybe logging?), so it returns to the appropriate version of the page without an error.
> 
> @jonatanschroeder is there a separate issue to track this yet?

Just created, #5090.

PR #5081, opened 2021-11-08
Fix course element extensions caching

nicknytko:
The `courseHash` and `urlPrefix` args were switched in `assets.courseElementExtensionAssetPath`, which was causing the question `demo/drawing/extensions ` to break.  This flips them to match `assets.courseElementAssetPath`.


nwalters512 (approved):
Whew, great catch!

PR #5089, opened 2021-11-15
Fix manual grading pages

SethPoulsen:
I've been testing these features so that I can get them working for courses I'm involved with. I've run into a number of issues both when testing locally, and when running in production. 

Errors when testing locally: 
1.  error finding socket.io.slim.js on question manual grading page
2.  "next question" button would not always work on question manual grading page, often just keeping you on the same page
3.  "start grading" button on instructor assessment manual grading page would sometimes not work, just leaving you on the same page

(1) Was a pretty easy fix, I just needed to fix the import. However, I don't see a need for this import on this page. We may want to remove it all together, unless I am missing something.

(2) and (3) were a bit more complicated. They both came down to bugs in SQL queries which made it so the different pages thought that there were different numbers of submissions left to grade. I have fixed the SQL queries, and I have also collapsed 2 of the queries into 1 where it made sense to. Less SQL queries in total means less SQL queries that can have bugs in them.

Errors when testing in Production: 
- Any time I hit the `next_question` endpoint in production (on the course `pl-pb101`), I will get a 504 gateway timeout, and the server will never respond. @mwest1066 and I spent a while debugging looking through the code as well as the logs, and we were unable to figure out what was going on. I'm hoping that fixing the issues that I saw when testing locally will also fix this issue, but I'm not positive that it will. As such, I've added some extra log statements to help with debugging this in production. These log statements will be removed before this feature is opened up more broadly so as to not overwhelm the production logs.


SethPoulsen:
I'm also adding user name and user id of students on the page that show the full list of things to be graded, because this is very easy to add and critical for usability.

SethPoulsen:
After talking to @mwest1066 we decided to go back to separate SQL queries for each page to minimize data sent over the network.

SethPoulsen:
I've just opened a new issue for one of the things that I wasn't able to do in this PR, but I hope to soon: https://github.com/PrairieLearn/PrairieLearn/issues/5091

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.js
```diff
@@ -44,6 +44,7 @@ router.get('/', (req, res, next) => {
               );
             }
 
+            logger.info('QuestionManualGrading: Found Question To Grade in DB.');
```
mwest1066, line 47: For all of these logging calls, can we add some simple context to help track things? For example:
```suggestion
            logger.info('QuestionManualGrading: Found Question To Grade in DB.', {instance_question_id: res.locals.instance_question.id, result_row: result.rows[0]});
```

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.sql
```diff
@@ -1,18 +1,16 @@
...
-WHERE 
-    iq.assessment_question_id = $assessment_question_id
-    AND a.id = $assessment_id
+    JOIN assessment_instances AS ai ON (ai.id = iq.assessment_instance_id)
+    JOIN variants AS v ON (v.instance_question_id = iq.id)
+    JOIN submissions s ON (s.variant_id = v.id)
+WHERE
+    ai.assessment_id = $assessment_id
+    AND iq.assessment_question_id = $assessment_question_id
+ORDER BY iq.id DESC, s.date DESC, s.id DESC) as submission_info
```
mwest1066, line 14: Can you reformat to indent the subquery? Something like:
```
SELECT * FROM (
    SELECT DISTINCT ON ...
    FROM ...
    WHERE ...
    ORDER BY ...
) AS submission_info
WHERE ...
LIMIT 1;
```

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.sql
```diff
@@ -1,18 +1,16 @@
...
+WHERE submission_info.graded_at is NULL
```
mwest1066, line 15:
```suggestion
WHERE submission_info.graded_at IS NULL
```

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.sql
mwest1066:
```suggestion
ORDER BY iq.id DESC, s.date DESC, s.id DESC) AS submission_info
```

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.sql
```diff
@@ -1,18 +1,16 @@
...
-            SELECT DISTINCT ON (s.auth_user_id) * FROM submissions AS s
-            WHERE s.graded_at IS NULL
-            ORDER BY s.auth_user_id, s.date DESC, s.id DESC
-        ) s ON (s.variant_id = v.id)
-WHERE 
-    iq.assessment_question_id = $assessment_question_id
-    AND a.id = $assessment_id
+    JOIN assessment_instances AS ai ON (ai.id = iq.assessment_instance_id)
+    JOIN variants AS v ON (v.instance_question_id = iq.id)
+    JOIN submissions s ON (s.variant_id = v.id)
```
mwest1066, line 10:
```suggestion
    JOIN submissions AS s ON (s.variant_id = v.id)
```

pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.js
```diff
@@ -32,7 +32,11 @@ router.get('/', function (req, res, next) {
       assessment_id: res.locals.assessment.id,
       assessment_question_id: res.locals.assessment_question_id,
     };
-    sqlDb.query(sql.select_instance_questions_manual_grading, params, function (err, result) {
+
+    let query = res.locals.assessment.group_work
+      ? sql.select_instance_questions_manual_grading_groups
+      : sql.select_instance_questions_manual_grading;
```
mwest1066, line 38: We'd like to be robust to the situation where `group_work` is changed while an assessment is in progress, so some assessment instances are for individual students and some are for groups. To do this, one simple strategy is to LEFT JOIN on both groups and users like this: https://github.com/PrairieLearn/PrairieLearn/blob/b3c07403d2f65bc21eba563b43544fe308b9a00e/pages/instructorAssessmentInstances/instructorAssessmentInstances.sql#L49

Ideally the EJS would also render group or user for each row, rather than using `assessment.group_work` to decide, although I see that `instructorAssessmentInstances.ejs` doesn't do this well: https://github.com/PrairieLearn/PrairieLearn/blob/b3c07403d2f65bc21eba563b43544fe308b9a00e/pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs#L220

It's not necessary to fix this before merging this PR, but it would be good to come back to it quickly in a follow-up PR so we don't forget.

mwest1066:
@SethPoulsen I haven't actually run this but it all looks fine and it's contained to the manual grading code so I'm happy to try it out. I'd prefer to have the cosmetic changes done before merging and to add some more info to all the `logger()` output.

pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.js
SethPoulsen: Ok, sounds good.

SethPoulsen: Ya,  I totally didn't realize that having both types at once was a thing, because I just assumed that what `instructorAssessmentInstances.ejs` was doing was correct.

mwest1066: I think it would be fine to choose the headers based on the value of `assessment.group_work` (which will almost always be correct) but we should just display something sensible in the case of rows of the "wrong" type. For example, a group work assessment could have a column called "Group Name" but then for rows without a group (but with a user_id) could just put the user name in that column.

SethPoulsen: I've made an issue so we don't forget: https://github.com/PrairieLearn/PrairieLearn/issues/5092

PR #5093, opened 2021-11-16
Fix undefined variable in manual grading info logging

SethPoulsen:


pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.js
```diff
@@ -29,11 +29,14 @@ router.get('/', (req, res, next) => {
...
     }
 
+    const instance_question_id = result.rows[0].id;
     logger.info(
       'ManualGradingNextInstanceQuestion: Found next submission to grading, redirecting.',
-      { instance_question_id: instance_question_id, result_row: result.rows[0] }
+      {
+        instance_question_id: instance_question_id,
+        result_row: result.rows[0],
+      }
```
SethPoulsen, line 38: I think its hilarious that prettier decided to format this differently, based on whether or not it was before or after the variable declaration. I can't see how this makes any sense, or how it is making these decisions.

nwalters512: This is based on the trailing semicolon after `result.rows[0]`, I think - if you remove it, it's happy to place everything on one line.

nwalters512 (approved):
We should really turn on https://eslint.org/docs/rules/no-use-before-define, which would have caught this. I filed https://github.com/PrairieLearn/PrairieLearn/issues/5095 to track that.

PR #5094, opened 2021-11-16
fix: Ensure DB client is released if transaction errors

nwalters512:
This was introduced by me in #4989 - I didn't properly translate the old code into async/await. This meant that we were leaking DB clients if transactions ever errored.


jonatanschroeder:
I'm not familiar enough with DB pools to test this and properly opine on this PR, sorry.

mwest1066 (approved):

PR #5096, opened 2021-11-16
Add optional comment attribute to pl-order-blocks DAG grader to help with research documentation

SethPoulsen:
We are doing a research study to figure out what kinds of distractor items are most compelling for students when writing math proofs with pl-order-blocks. It would be nice if we could directly annotate our HTML with what kinds of distractors they are.


elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -80,7 +80,8 @@ def prepare_tag(html_tags, index, group=None):
         elif grading_method in ['ranking', 'ordered']:
             pl.check_attribs(html_tags, required_attribs=[], optional_attribs=['correct', 'ranking', 'indent'])
         elif grading_method == 'dag':
-            pl.check_attribs(html_tags, required_attribs=[], optional_attribs=['correct', 'tag', 'depends'])
+            # 'distractor-for' and 'distractor-type' are to help keep track of types of distractors for research purposes not publicly documented because we might change them later
+            pl.check_attribs(html_tags, required_attribs=[], optional_attribs=['correct', 'tag', 'depends', 'distractor-for', 'distractor-type'])
```
SethPoulsen, line 84: thoughts on the attribute names @mwest1066 ? `distractor-type` and `distractor-for` are a bit long, it would be nice to have something shorter, but I couldn't think of anything.

mwest1066: What about having an extra attribute called `comment` or something like that, and you can put `comment="for:abc,type:xyz"` or whatever other custom format you want? That would mean we don't need add more attributes in the future.

SethPoulsen: Ya I think you're right, this makes more sense.

mwest1066 (approved):

PR #5097, opened 2021-11-17
Install newer versions of Python deps for Apple Silicon compat

nwalters512:
Apple Silicon support for Python is... iffy, right now. Many packages, especially all the ones that we use that rely heavily on native code, don't have prebuilt binaries available. So, I had to build them from source, which required installing a lot of extra dependencies. Even then, some things wouldn't build without a version bump, which is what this PR does!

As far as I can tell, `statsmodels==0.13.0` didn't actually contain any breaking changes, per the [release notes](https://www.statsmodels.org/devel/release/version0.13.0.html). So I feel pretty good about this not breaking anything.

### DIFF
--- a/images/plbase/python-requirements.txt
+++ b/images/plbase/python-requirements.txt
@@ -9,7 +9,7 @@ pandas==1.3.2
 pycryptodome==3.10.1
 pyquaternion==0.9.9
 regex==2021.8.28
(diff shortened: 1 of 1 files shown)


mwest1066 (approved):

nwalters512:
Closing in favor of #5119.

PR #5098, opened 2021-11-17
Add Points Input Box to Manual Grading Page

SethPoulsen:
Instructors should be able to specify points OR percentage (previously they could only do it by percentage).

Submission points updating

### DIFF
--- a/pages/instructorQuestionManualGrading/instructorQuestionManualGrading.js
+++ b/pages/instructorQuestionManualGrading/instructorQuestionManualGrading.js
@@ -51,6 +51,7 @@ router.get('/', (req, res, next) => {
             res.locals.question = result.rows[0].question;
             res.locals.variant = result.rows[0].variant;
             res.locals.submission = result.rows[0].submission;
+            res.locals.max_points = result.rows[0].max_points;
             res.locals.score_perc = res.locals.submission.score * 100;
             callback(null);
           }
@@ -72,6 +73,7 @@ router.get('/', (req, res, next) => {
             variant: res.locals.variant,
             submission: res.locals.submission,
           });
+          // console.log(res.locals);
           callback(null);
         });
       },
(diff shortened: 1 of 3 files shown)


SethPoulsen:
The input boxes don't really line up that well with the text. 

Are there some standard CSS classes we have or something that I should use to make it a little prettier?

nwalters512:
> Are there some standard CSS classes we have or something that I should use to make it a little prettier?

Given that we use Bootstrap, the answer to this question is generally "yes"! Check out the second example under https://getbootstrap.com/docs/5.1/forms/overview/#form-text - that should serve as a good starting point.

BenjaminCosman:
This looks great, thanks! This is definitely required for usability - I frequently grade things out of say 6 or 7, and typing in percentages for those would be obnoxious.

tdy:
> Check out the second example under https://getbootstrap.com/docs/5.1/forms/overview/#form-text - that should serve as a good starting point.

Note that those are v5 docs. If Seth isn't well-versed in v4 vs v5, I'd suggest using the v4 docs: https://getbootstrap.com/docs/4.6/components/forms/#supported-elements

I'd use the prepend/append pattern here: `input-group-prepend` for "Points" and "Percent" and `input-group-append` for "/5" and "%".

nwalters512:
> > Check out the second example under [getbootstrap.com/docs/5.1/forms/overview/#form-text](https://getbootstrap.com/docs/5.1/forms/overview/#form-text) - that should serve as a good starting point.
> 
> Note that those are v5 docs. If @SethPoulsen isn't well-versed in v4 vs v5, I'd suggest using the v4 docs: [getbootstrap.com/docs/4.6/components/forms/#supported-elements](https://getbootstrap.com/docs/4.6/components/forms/#supported-elements)
> 
> I'd use the prepend/append pattern here: `input-group-prepend` for "Points" and "Percent" and `input-group-append` for "/5" and "%".

Fantastic point, I need to keep in mind that we're lagging behind.

> However, note that prepend/append is one of v5's breaking changes (I'm not sure what @nwalters512 has in mind for migrating v4->v5 in PL).

😬 this was vaguely a disaster when we made the leap from v3 -> v4, namely because questions themselves come to rely on specific Bootstrap features from a version that break or disappear when jumping major versions. I unfortunately don't have any great ideas for this at the moment. This is one of the problems with mixing platform and user-provided code on the same page, any styles present on the page become part of your API contract with users and changing it is hard 😕

mwest1066 (approved):

SethPoulsen:
Note for future reference: 

Matt opted to merge this right away so that @BenjaminCosman can start using it now. 

I'm happy to clean up the styling later.

PR #5099, opened 2021-11-17
fix: Use DATE_PART instead of EXTRACT in Postgres queries

nwalters512:
Postgres 14 changed `EXTRACT()` to return a `numeric` type ([changelog](https://www.postgresql.org/docs/release/14.0/)):

> Change EXTRACT() to return type numeric instead of float8 (Peter Eisentraut)
>
> This avoids loss-of-precision issues in some usages. The old behavior can still be obtained by using the old underlying function date_part().
>
> Also, EXTRACT(date) now throws an error for units that are not part of the date data type.

Our Postgres library treats `numeric` values coming from Postgres as strings and doesn't try to coerce them into numbers because of the possible loss of precision. We generally don't care *that* much about precision when working with dates/times, so we could configure `node-pg` to convert `numeric` types to `number`s automatically, but it seemed less risky to just use `DATE_PART` instead of `EXTRACT`, which will give us a double-precision floating point number that can natively be represented in JS.

I made this change with a naive find/replace, replacing `extract\s*\(epoch from ` with `DATE_PART('epoch', `). I _think_ this should be safe and sufficient.

Yes, I know I should just be developing in the container and not running bleeding-edge versions of Postgres 😂 but we'll be running on Postgres 14 _someday_, and we may as well get ready for that now in a backwards-compatible way that (famous last words) shouldn't hurt anything.

For those with access to the repo, this mirrors the changes made in https://github.com/PrairieLearnInc/PrairieTest/pull/216.

---

Here are two queries showing the problem:

```
postgres=# SELECT extract(epoch from interval '2 hours') as secs;
    secs
-------------
 7200.000000
```

Changing to `date_part` avoids this problem:

```
postgres=# select date_part('epoch', interval '2 hours') as secs;
 secs
------
 7200
```


jonatanschroeder (approved):
I haven't tested, but looks good.

mwest1066 (approved):

PR #5103, opened 2021-11-18
Use instance_questions_update_score to save scores in manual grading

SethPoulsen:
Previously, the manual grading page was using `grading_jobs_insert_internal` which was a total hack, because that's supposed to be for autograded questions. This caused a couple of issues including: 
- questions could only be graded once from the manual grading page, it wouldn't save properly if you tried to go back and change the grade
- inconsistent behavior based on 'credit' depending on which page you manually enter the grade on (still not totally clear what we should do here)

This solution isn't probably the end of this, we need to think more deeply about how scores from manual grading should be handled.

However, it will get use something that 'mostly works' for now, which is better than what we currently have.

### DIFF
--- a/pages/instructorQuestionManualGrading/instructorQuestionManualGrading.js
+++ b/pages/instructorQuestionManualGrading/instructorQuestionManualGrading.js
@@ -7,6 +7,7 @@ const question = require('../../lib/question');
 const debug = require('debug')('prairielearn:' + path.basename(__filename, '.js'));
 const error = require('../../prairielib/lib/error');
 const sqlDb = require('../../prairielib/lib/sql-db');
+const ltiOutcomes = require('../../lib/ltiOutcomes');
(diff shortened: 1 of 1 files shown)


pages/instructorQuestionManualGrading/instructorQuestionManualGrading.js
```diff
@@ -89,7 +90,7 @@ router.get('/', (req, res, next) => {
 router.post('/', function (req, res, next) {
   if (req.body.__action === 'add_manual_grade') {
     const note = req.body.submission_note;
-    const score = req.body.submission_score_percent;
+    const score_perc = req.body.submission_score_percent;
```
mwest1066, line 93: Can we change the `req.body` variable to also be `req.body.score_perc` for consistency? (not needed immediately for this PR)

mwest1066 (approved):

Issue #5091, opened 2021-11-15
Manual Grading Sould go to next submission, not Beginning when you click "Submit and next Question"


SethPoulsen:
Currently, when you are on the manual grading page and you click "Submit and Next", no matter where you are in the list of submissions to grade, it will take you back to the front of the list of ungraded submissions. 

Instead, it should go to the next one in the list after where you currently are.

e.g. If you choose on the overview page to grade the 10th student, when you click "Submit and Next", you should go to the 11th submission, not the 1st. 

This will also make is MUCH easier for multiple people to work on grading the same question at the same time, because one person can start at the top and one person can start halfway down. 

The best way to implement this will probably be adding an extra parameter to the `next_question` endpoint which says what the current question is.

SethPoulsen:
This feature together with the confilct warning & resolution system (being worked on over here: https://github.com/PrairieLearn/PrairieLearn/pull/3964#issuecomment-969385928) will make a pretty good experience for multiple people grading the same question at once.

PR #5104, opened 2021-11-19
Manual grading: return questions in  order when selecting 'next'

SethPoulsen:
Fixes https://github.com/PrairieLearn/PrairieLearn/issues/5091

### DIFF
--- a/pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.sql
+++ b/pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.sql
@@ -13,22 +13,22 @@ WHERE
 
 -- BLOCK select_instance_questions_manual_grading
 SELECT * FROM (
-SELECT DISTINCT ON (iq.id)
-    iq.*,
-    s.graded_at,
-    ai.id AS assessment_instance_id,
-    u.uid,
-    COALESCE(g.name, u.name) AS user_or_group_name
-FROM
-    instance_questions AS iq
-    JOIN assessment_instances AS ai ON (ai.id = iq.assessment_instance_id)
-    JOIN variants AS v ON (v.instance_question_id = iq.id)
-    JOIN submissions s ON (s.variant_id = v.id)
-    LEFT JOIN users u ON (u.user_id = ai.user_id)
-    LEFT JOIN groups g ON (g.id = ai.group_id)
-WHERE
-    ai.assessment_id = $assessment_id
-    AND iq.assessment_question_id = $assessment_question_id
-ORDER BY iq.id DESC, s.date DESC, s.id DESC
+    SELECT DISTINCT ON (iq.id)
+        iq.*,
+        s.graded_at,
+        ai.id AS assessment_instance_id,
+        u.uid,
+        COALESCE(g.name, u.name) AS user_or_group_name
+    FROM
+        instance_questions AS iq
+        JOIN assessment_instances AS ai ON (ai.id = iq.assessment_instance_id)
+        JOIN variants AS v ON (v.instance_question_id = iq.id)
+        JOIN submissions s ON (s.variant_id = v.id)
+        LEFT JOIN users u ON (u.user_id = ai.user_id)
+        LEFT JOIN groups g ON (g.id = ai.group_id)
+    WHERE
+        ai.assessment_id = $assessment_id
+        AND iq.assessment_question_id = $assessment_question_id
+    ORDER BY iq.id DESC, s.date DESC, s.id DESC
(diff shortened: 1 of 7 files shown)


SethPoulsen:
I've also changed the manual grading pages to show things to be graded in alphabetical order, instead of ordered by the instance question ID, as that makes more sense.

pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.sql
```diff
@@ -12,38 +12,23 @@ WHERE
 
 
 -- BLOCK select_instance_questions_manual_grading
+SELECT * FROM (
```
mwest1066, line 15: Can we indent the following subquery?

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.sql
```diff
@@ -1,16 +1,27 @@
...
+            LEFT JOIN groups g ON (g.id = ai.group_id)
+        WHERE
+            ai.assessment_id = $assessment_id
+            AND iq.assessment_question_id = $assessment_question_id
+        ORDER BY iq.id DESC, s.date DESC, s.id DESC) AS submission_info
+    WHERE submission_info.graded_at IS NULL OR submission_info.id = $prior_instance_question_id
+)
+SELECT * FROM to_grade
+WHERE graded_at IS NULL
+ORDER BY is_next DESC
```
mwest1066, line 25:
```suggestion
ORDER BY is_next DESC -- choose next question if there is one, otherwise choose any question
```

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.sql
```diff
@@ -1,16 +1,27 @@
...
+            LEFT JOIN users u ON (u.user_id = ai.user_id)
+            LEFT JOIN groups g ON (g.id = ai.group_id)
+        WHERE
+            ai.assessment_id = $assessment_id
+            AND iq.assessment_question_id = $assessment_question_id
+        ORDER BY iq.id DESC, s.date DESC, s.id DESC) AS submission_info
+    WHERE submission_info.graded_at IS NULL OR submission_info.id = $prior_instance_question_id
+)
+SELECT * FROM to_grade
+WHERE graded_at IS NULL
```
mwest1066, line 24: After the change in #5103, will manually grading a question actually set `graded_at` on the submission? I worry that it won't, so here we'll just keep selecting these questions.

SethPoulsen: It is, right here: 
https://github.com/PrairieLearn/PrairieLearn/blob/0473cb9a848c5b93f3803f371da83fc12e7eac81/sprocs/instance_questions_update_score.sql#L126-L151

I just noticed by looking there, however, that they are also being marked as `grading_method='External'`, which isn't really correct.

pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.sql
SethPoulsen: sure

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.sql
mwest1066: Ok, great. I see we _are_ correctly marking the `grading_job` as `grading_method='Manual'`. Want to fix the `External`?

SethPoulsen: Wait, this same function call has been for instructors manually editing grades for ages. Has this been wrong for that case all along too?

mwest1066: I think so? I don't think we use `grading_method` for anything important, so it hasn't been causing a problem?

SethPoulsen: I guess. Let's just open a new issue for it.

SethPoulsen: https://github.com/PrairieLearn/PrairieLearn/issues/5110

mwest1066 (approved):

Issue #4671, opened 2021-08-18
Docker: prairielearn/prairielearn still does not have linux/arm64 architecture, but plbase does


ctessum:
You can see that the prairielearn/plbase docker image [does have an entry for linux/arm64](https://hub.docker.com/r/prairielearn/plbase/tags?page=1&ordering=last_updated), but prairielearn/prairielearn [does not](https://hub.docker.com/r/prairielearn/prairielearn/tags?page=1&ordering=last_updated).

Is there a way to build prairielearn/prairielearn using buildx, like what was added in #4622 ?

It works fine if I build it locally on my laptop, but I don't know how to add it to whatever workflow PL uses.

mwest1066:
Sorry, I should have documented this better. I actually did enable arm64 builds for `prairielearn/prairielearn` (see eff0e8a009bde406f516f0106e9e1fc0a2b262f0) but the arm64 build would often hang at the DB initialization step, holding up the entire image build. I suspect that the problem is the same as this one: postgis/docker-postgis#216. The build infrastructure uses qemu for the arm build, and it sounds like there is a qemu bug that affects postgresql.

I tried a few different options and currently (see 2dce3bc772841612c5dd235a3d2ec9523f4d7ef6) we are pushing an arm build at a separate tag (`prairielearn/prairielearn:arm`), as the last image build step so it doesn't matter so much if it fails.

One possible solution might be to disable DB initialization on arm, and instead do it when the container starts up? At the moment we start postgres, run an initial "migrate-and-exit" pass of PL to initialize the DB, and then stop postgres again. All of this is just an optimization, so we could skip it in the arm build and everything should just work. These are the lines to skip: https://github.com/PrairieLearn/PrairieLearn/blob/5247b532062768431abbd4edb2c8a0c9592fdb62/Dockerfile#L28

Do you want to make a PR to skip the DB steps on build on arm (and test it)?

For reference, here is are the build steps at the moment: https://github.com/PrairieLearn/PrairieLearn/blob/5247b532062768431abbd4edb2c8a0c9592fdb62/.github/workflows/images.yml#L42 and on arm: https://github.com/PrairieLearn/PrairieLearn/blob/5247b532062768431abbd4edb2c8a0c9592fdb62/.github/workflows/images.yml#L163

nwalters512:
Reopening - I reverted my changes in https://github.com/PrairieLearn/PrairieLearn/pull/5109.

mwest1066:
[...]

PR #5105, opened 2021-11-19
fix: Build arm64 version of PrairieLearn image

nwalters512:
As far as we can tell, the arm64 variant of the `prairielearn/prairielearn` image has been building successfully on master for some time now. This should allow us to build it normally with the amd64 image, instead of separately at the end of the workflow with `continue-on-error: true` set to ignore errors.

There's no great way to test this from a branch - I'll monitor the master build once this lands, and if it turns out that this broke something, I'll revert it again.

Closes https://github.com/PrairieLearn/PrairieLearn/issues/4671.

### DIFF
--- a/.github/workflows/images.yml
+++ b/.github/workflows/images.yml
@@ -43,7 +43,7 @@ jobs:
         uses: docker/build-push-action@v2 # https://github.com/marketplace/actions/build-and-push-docker-images
         with:
           context: .
-          platforms: linux/amd64
+          platforms: linux/amd64,linux/arm64
           push: true
           no-cache: true
           tags: prairielearn/prairielearn:latest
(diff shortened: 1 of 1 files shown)


mwest1066 (approved):

nwalters512:
After merging the master build failed, and also failed on a subsequent retry 😢 

```
#14 7.783 [1/4] Resolving packages...
#14 20.77 [2/4] Fetching packages...
#14 173.1 info There appears to be trouble with your network connection. Retrying...
#14 206.3 info There appears to be trouble with your network connection. Retrying...
#14 239.5 info There appears to be trouble with your network connection. Retrying...
#14 272.7 info There appears to be trouble with your network connection. Retrying...
#14 309.6 error An unexpected error occurred: "registry.yarnpkg.com/date-fns/-/date-fns-2.23.0.tgz: ESOCKETTIMEDOUT".
```

Both times, it failed on the same `date-fns` package. This appears to be the issue described in https://github.com/yarnpkg/yarn/issues/8242, which unfortunately was closed as "resolved in Yarn v2". I'm going to try increasing the network timeout to 60s in another PR.

PR #5106, opened 2021-11-19
chore: Increase Yarn network timeout

nwalters512:
This is a followup to https://github.com/PrairieLearn/PrairieLearn/pull/5105 that tries to fix the new build failures on master by increasing the Yarn timeout. If this doesn't work, we may have to switch to Yarn 2.

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,0 +1,1 @@
+network-timeout 60000


mwest1066 (approved):

mwest1066:
From a spot-check of the `images` workflow before #5105 it looks like yarn might be timing out more for the `arm` builds than for `amd`? Am I imagining this, or is there actually a difference in package hosting?

nwalters512:
@mwest1066 the hypothesis from https://github.com/nodejs/docker-node/issues/1335 is that "emulation is slow" which increases our likelihood of running into timeouts. Which seems plausible, but the default timeout is 30s, which already seems quite high, so 🤷

PR #5111, opened 2021-11-21
Add support for Check unit testing in C grader

jonatanschroeder:
The current C grader focuses on tests based on stdout, which can be limiting depending on the application. This change implements the ability to use the [Check framework](https://libcheck.github.io/check/) for more fine-tuned testing.

Some minor changes included in the same PR:
* Clean up of Dockerfile and use of `requirements.txt` file for pip install;
* Argument to pass environment variables to `run_command`;
* New function to compile a file without creating a test result;
* Allow use of `pkg-config` to set up required flags;

Implementation steps:
* [x] Implementation of features in grader code;
* [x] Testing;
* [x] Question(s) in example course using Check-based grading;
* [x] Docs.


graders/c/cgrader/cgrader.py
```diff
@@ -238,7 +263,48 @@ def add_test_result(self, name, description='', points=True,
...
+            tree = ET.parse(log_file)
+            for suite in tree.getroot().findall('{*}suite'):
+                suite_title = suite.findtext('{*}title') if use_suite_title else ''
+                for test in suite.findall('{*}test'):
+                    result = test.get('result')
+                    test_id = test.findtext('{*}id') if use_unit_test_id else ''
+                    iteration = f" (run {test.findtext('{*}iteration')})" if use_iteration else ''
+                    case_name = test.findtext('{*}description') if use_case_name else ''
+                    self.add_test_result(f'{suite_title}{separator_1}{case_name}{separator_2}{test_id}{iteration}',
+                                         points=result == 'success',
```
nwalters512, line 300: Food for thought: in the Catch-based autograder we developed for CS 225 at UIUC, we could use tags to specify additional metadata or execution instructions for individual tests, such the number of points a given test case is worth, or whether or not a given test case should be executed with Valgrind. Looks like recent version of Check include `tcase_set_tags`, which could also be used for this! Unfortunately, per the versions chart at https://github.com/libcheck/check#packaging, the version of Check published for Ubuntu 18.04 isn't recent enough to include that 😢

graders/c/requirements.txt
```diff
@@ -0,0 +1,19 @@
...
+numpy==1.20.1
+Pillow==8.1.0
+pycrypto==2.6.1
+pygobject==3.26.1
+pyparsing==2.4.7
+python-dateutil==2.8.1
+pyxdg==0.25
+SecretStorage==2.3.1
+six==1.11.0
+sympy==1.7.1
```
nwalters512, line 19: Are all of these actually required? This seems excessive at first glance.

graders/c/cgrader/cgrader.py
```diff
@@ -76,14 +76,18 @@ def test_compile_file(self, c_file, exec_file, main_file=None,
...
             flags.append('-Wl,--allow-multiple-definition')
 
         if not compiler:
             compiler = self.compiler
 
+        if pkg_config_flags:
+            if isinstance(pkg_config_flags, str):
+                pkg_config_flags = shlex.split(pkg_config_flags)
+            out_flags = self.run_command(['pkg-config', '--cflags', '--libs'] + pkg_config_flags)
+            if out_flags: flags.extend(shlex.split(out_flags))
```
nwalters512, line 89: Nit for consistency:

```suggestion
            if out_flags:
            		flags.extend(shlex.split(out_flags))
```

exampleCourse/questions/demo/autograder/c/square/info.json
```diff
@@ -1,6 +1,6 @@
 {
     "uuid": "1e1a7e46-1511-49b8-8869-744276a309a0",
-    "title": "Autograder demo: Square function in C",
+    "title": "Autograder demo: Square function in C (Standard Output Override)",
```
nwalters512, line 3: What does "override" imply in the name?

nwalters512 (commented):
Thanks for this! Holding off on a ✅ until we can talk through some of these comments, particularly the ones about `main_file` backwards compatibility, some mistakes in the docs, and the number of Python deps being installed.

graders/c/cgrader/cgrader.py
```diff
@@ -60,12 +63,9 @@ def run_command(self, command, input=None, sandboxed=True, timeout=None):
                     out += out2.decode('utf-8', 'backslashreplace')
                 return out + tostr
 
-    def test_compile_file(self, c_file, exec_file, main_file=None,
```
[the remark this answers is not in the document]
jonatanschroeder: It still works in `test_compile_file` for backwards compatibility, but not in `compile_file` (the new function).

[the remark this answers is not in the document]
nwalters512: Ah, perfect!

graders/c/cgrader/cgrader.py
```diff
@@ -24,19 +25,21 @@ def __init__(self, compiler='gcc'):
             self.data = json.load(file)
         self.compiler = compiler
 
-    def run_command(self, command, input=None, sandboxed=True, timeout=None):
+    def run_command(self, command, input=None, sandboxed=True, timeout=None, env=None):
         if isinstance(command, str):
             command = shlex.split(command)
         if sandboxed:
+            envl = [f'{shlex.quote(k)}={shlex.quote(v)}' for k, v in env.items()] if env else []
```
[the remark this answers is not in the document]
jonatanschroeder: That's actually a good question. I think there was some leftover issues from previous version of the grader that used other methods of keeping the path, but indeed this is not necessary. I'll remove it and make some tests to ensure this doesn't affect functionality.

graders/c/cgrader/cgrader.py
```diff
@@ -238,7 +263,48 @@ def add_test_result(self, name, description='', points=True,
...
+        if not isinstance(args, list):
+            args = [args]
+
+        log_file_dir = tempfile.mkdtemp()
+        log_file = os.path.join(log_file_dir, 'tests.xml')
+        self.change_mode(log_file_dir, '777', change_parent=False)
+    
+        out = self.run_command([exec_file] + args,
+                               env={'CK_XML_LOG_FILE_NAME': log_file, 'TEMP': '/tmp'})
+        print(out)
```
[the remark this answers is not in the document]
jonatanschroeder: May make debugging graders easier for instructors, since they can check the grading job log...

graders/c/cgrader/cgrader.py
```diff
@@ -238,7 +263,48 @@ def add_test_result(self, name, description='', points=True,
...
+
+        separator_1 = ': ' if use_suite_title and use_case_name else ''
+        separator_2 = ' - ' if use_unit_test_id and (use_suite_title or use_case_name) else ''
+        try:
+            with open(log_file) as f:
+                print('DEBUG: LOG FILE')
+                print(f.read())
```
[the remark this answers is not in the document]
jonatanschroeder: Good point, removing.

graders/c/cgrader/cgrader.py
jonatanschroeder: I was considering using this opportunity to upgrade to 20.04, but that is not far enough either. And I'd prefer to avoid non-LTS versions... I'll figure some things out.

graders/c/requirements.txt
jonatanschroeder: Good point. I just did a dump on the status of what was in the previous version. I'll trim it down a bit.

graders/c/cgrader/cgrader.py
nwalters512: I was thinking the same things. I'm be no means asking for a change here - just throwing out an idea that may someday be useful!

exampleCourse/questions/demo/autograder/c/square/info.json
jonatanschroeder: The main function in this code replaces the standard output to ensure the provided code does not print some result that could be acceptable in the autograder. In other words, it ensures that students return the value, not print it.

graders/c/cgrader/cgrader.py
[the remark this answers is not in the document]
jonatanschroeder: Turns out this is needed for `PATH`, but not for other envs... Removed.

graders/c/cgrader/cgrader.py
jonatanschroeder: Added as issue #5126 .

exampleCourse/questions/demo/autograder/c/square/info.json
nwalters512: 👍 thanks for explaining!

graders/c/requirements.txt
```diff
@@ -0,0 +1,4 @@
+matplotlib==3.3.4
+numpy==1.20.1
+Pillow==8.1.0
+sympy==1.7.1
```
nwalters512, line 4: Out of curiosity, why are these required at all? I don't see them used in `cgrader.py`.

nwalters512 (approved):

graders/c/requirements.txt
jonatanschroeder: They are not used in cgrader itself, but:
* matplotlib is used in https://github.com/PrairieLearn/PrairieLearn/blob/7a7647fe1b6624f7388e64f6b4969ba016d93dfc/exampleCourse/questions/demo/autograder/c/drawTriangle/tests/test.py#L5
* sympy and numpy are used in some of the questions I originally used this grader for, so they're kept for compatibility with those (plus any other production questions that may still depend on them);
* Pillow is not directly used, but sympy depends on it, and I've had problems with the new version of Pillow on pip3 (it has some undisclosed dependencies). Version 8.1.0 does not cause these issues.

nwalters512: Ah, ok.

I think we need a better house view on things like this - my expectation would be that if specific courses need additional libraries, they should make their own image that installs anything extra they may need. Otherwise these core images keep getting fatter and fatter.

All that said, I acknowledge that we _don't_ have a well defined house view like that and that maintaining backwards-compat is important, so this looks fine!

PR #5113, opened 2021-11-23
Proof blocks partial credit

SethPoulsen:
Implements https://github.com/PrairieLearn/PrairieLearn/issues/4960


docs/elements.md
```diff
@@ -413,20 +413,21 @@ Element to arrange given blocks of code or text that are displayed initially in
...
+| `solution-placement`  | "right" or "bottom"     | "right"                        | `right` shows the source and solution areas aligned side-by-side. `bottom` shows the solution area below the source area.                                                                                                                                                                                                                                                                 |
+| `partial-credit`      | string                  | Depends on `grading-method`    | For the `dag` grading method, you may specify `none` for no partial credit (defualt) or `lcs` for partial credit based on the LCS edit-distance from the student solution to some correct solution. For the other grading methods, using this property is not yet supported. Grading methods `unordered` and `ranking` will assign partial credit, and grading method `ordered` will not. |
```
nwalters512, line 429: Edit for possibly increased clarity:
```suggestion
| `partial-credit`      | string                  | Depends on `grading-method`    | For the `dag` grading method, you may specify `none` for no partial credit (defualt) or `lcs` for partial credit based on the LCS edit-distance from the student solution to some correct solution. For the other grading methods, using this property is not yet supported. Grading methods `unordered` and `ranking` will always assign partial credit, and grading method `ordered` will never do so. |
```
Also, instead of `string` for the type, let's enumerate the possible values like the other properties here do.

elements/pl-order-blocks/dag_checker.py
```diff
@@ -45,27 +46,92 @@ def check_grouping(order: list[str], group_belonging: Mapping[str, Optional[int]
     return len(order)
 
 
-def grade_dag(order: list[str], depends_graph: Mapping[str, list[str]], group_belonging: Mapping[str, Optional[int]]) -> tuple[int, int]:
+def dag_to_nx(depends_graph: Mapping[str, list[str]]) -> nx.DiGraph:
+    """Convert input graph format from HTML into NetworkX object to utilize their algorithms."""
+    graph = nx.DiGraph()
+    for node in depends_graph:
+        for node2 in depends_graph[node]:
+            graph.add_edge(node2, node)
```
nwalters512, line 54: To clarify understanding: `depends_graph` is a dict of reverse edges? So if the graph is `A -> B`, `depends_graph` would be `{'B': ['A']}`?

elements/pl-order-blocks/dag_checker.py
```diff
@@ -45,27 +46,92 @@ def check_grouping(order: list[str], group_belonging: Mapping[str, Optional[int]
     return len(order)
 
 
-def grade_dag(order: list[str], depends_graph: Mapping[str, list[str]], group_belonging: Mapping[str, Optional[int]]) -> tuple[int, int]:
+def dag_to_nx(depends_graph: Mapping[str, list[str]]) -> nx.DiGraph:
+    """Convert input graph format from HTML into NetworkX object to utilize their algorithms."""
```
nwalters512, line 50: Nit: we're really far removed from HTML at this point, I don't think we should mention it here.

elements/pl-order-blocks/dag_checker.py
```diff
@@ -45,27 +46,92 @@ def check_grouping(order: list[str], group_belonging: Mapping[str, Optional[int]
...
+            distractors += 1
+            continue
+
+        for node2 in seen:
+            if trans_clos.has_edge(node, node2):
+                subgraph.add_edge(node, node2)
+
+        seen.add(node)
+
+    if len(subgraph) == 0:
```
nwalters512, line 122: Nit: I'd rather use this instead for extra clarity - I had to go and look up how `len(nx.DiGraph())` was defined, but this should be unambiguous.
```suggestion
    if subgraph.number_of_nodes() == 0:
```

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -115,6 +126,9 @@ def prepare_tag(html_tags, index, group=None):
         elif html_tags.tag == 'pl-block-group':
             if grading_method != 'dag':
                 raise Exception('Block groups only supported in the "dag" grading mode.')
+            if partial_credit_type != 'none':
+                raise Exception('Partial credit not yet supported in questions using block groups')
```
nwalters512, line 130: Nit for consistency:
```suggestion
                raise Exception('Partial credit not yet supported in questions using block groups.')
```

elements/pl-order-blocks/dag_checker.py
```diff
@@ -45,27 +46,92 @@ def check_grouping(order: list[str], group_belonging: Mapping[str, Optional[int]
...
+    graph = dag_to_nx(depends_graph)
+    trans_clos = nx.algorithms.dag.transitive_closure(graph)
+
+    seen = set()
+    subgraph = nx.DiGraph()
+    distractors = 0
+    for node in order:
+        if node is None:
+            distractors += 1
+            continue
```
nwalters512, line 114: I had to trace the code back to find this - can you confirm that the implicit assumption here is that distractors will not have a defined `tag`?

elements/pl-order-blocks/dag_checker.py
```diff
@@ -45,27 +46,92 @@ def check_grouping(order: list[str], group_belonging: Mapping[str, Optional[int]
...
+        solution until it is the correct solution, so you can directly calculate the edit distance.
+    :param order: the block ordering given by the student
+    :param depends_graph: The dependency graph between blocks specified in the question
+    :return: edit distance from the student submission to some correct solution
+    """
+    graph = dag_to_nx(depends_graph)
+    trans_clos = nx.algorithms.dag.transitive_closure(graph)
+
+    seen = set()
+    subgraph = nx.DiGraph()
```
nwalters512, line 109: Nit: let's name this `problematic_subgraph` for consistency with the docs above.

nwalters512 (commented):
I confess to not having a great understanding of the partial credit algorithm used here; I trust that that's correct! Most of my comments are stylistic or intended to improve comprehension.

elements/pl-order-blocks/dag_checker.py
```diff
@@ -45,27 +46,92 @@ def check_grouping(order: list[str], group_belonging: Mapping[str, Optional[int]
...
-    correctness = top_sort_correctness if top_sort_correctness < grouping_correctness else grouping_correctness
-    first_wrong = -1 if correctness == len(order) else correctness
-    return correctness, first_wrong
+    return top_sort_correctness if top_sort_correctness < grouping_correctness else grouping_correctness
+
+
+def is_vertex_cover(G: nx.DiGraph, vertex_cover: Iterable):
+    """ this function from
+    https://docs.ocean.dwavesys.com/projects/dwave-networkx/en/latest/_modules/dwave_networkx/algorithms/cover.html#is_vertex_cover
+    """
```
tdy, line 81:
```suggestion
    """Determines whether the given set of vertices is a vertex cover of graph G.
    A vertex cover is a set of vertices such that each edge of the graph is incident with at least one vertex in the set.
    This function is taken from:
    https://docs.ocean.dwavesys.com/projects/dwave-networkx/en/latest/_modules/dwave_networkx/algorithms/cover.html#is_vertex_cover
    :param G: the graph on which to check the vertex cover
    :param vertex_cover: the iterable of nodes
    :return: True if the given iterable forms a vertex cover
    """
```
It might be good to mirror part of the docstring for clarity, in addition to the attribution link.

elements/pl-order-blocks/dag_checker.py
SethPoulsen: Ya... when writing the questions as an instructor it makes more sense to think about the edges one way, but when I thought up the grading algorithms it was easier to think about them the other way. I guess it would only take changing a couple lines of code to make the grading algorithms work with the edges the other direction. Do you think it would significantly increase code readability if I did that?

elements/pl-order-blocks/dag_checker.py
SethPoulsen: sure.

elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: sure

elements/pl-order-blocks/dag_checker.py
SethPoulsen: sounds good.

elements/pl-order-blocks/dag_checker.py
SethPoulsen: Yes, but I don't like it very much actually. 

Right now, during the `parse` function, we are assigning it to `None`. I actually don't like this because it is a little hard to tell what's going on (as you noticed), and also, the information about which distractor the student picked is not saved in the database. On the other hand, I don't want to require the person writing the question to put a `tag` on distractor lines. Maybe we could put uniquely generated tags on distractors internally or something? I don't know, I'll spend a little more time thinking about it. Let me know if you have an idea.

elements/pl-order-blocks/dag_checker.py
SethPoulsen: sure

elements/pl-order-blocks/dag_checker.py
SethPoulsen: I generally don't like having documentation that is longer than the function itself, because I think it just gets in the way of reading and understanding the code.

But if other people really want it I can put it in.

elements/pl-order-blocks/dag_checker.py
SethPoulsen: Since there's so much going on here, I'm not going to make any big changes on this right now (even though I would like to eventually). I will leave a comment though.

docs/elements.md
SethPoulsen: ok

SethPoulsen:
> I confess to not having a great understanding of the partial credit algorithm used here

Ya, its a bit complicated I'll admit. 

I'm in the process of writing a paper about the algorithm (submitting to AIED in February most likely), which I hope will serve as good, detailed documentation. I will certainly link it in the docs once its actually published.

elements/pl-order-blocks/dag_checker.py
SethPoulsen: After a lot of thinking, and discussing with @mwest1066, I'm going to go ahead and leave this the way that it is. The reason why is because if I flip the edges the other way for the algorithm, then: 
- the diagrams of graphs I show in my paper would have to have the edges pointing upward instead of downward through the proof, which is pretty unconventional
- correct solutions to a proof would then become _reverse_ topological sorts instead of topological sorts.

I think both of these things together would cause more confusion than the current way of doing things does.

Instead, I will add a comment here clarifying that the depends graph lists the _incoming_ edges of a given node.

nwalters512 (approved):
Took another spin through this, it looks good! Since it sounds like @mwest1066 collaborated with you on this a lot, might be good to seek final approval from him as well.

docs/elements.md
```diff
@@ -414,20 +414,21 @@ Element to arrange given blocks of code or text that are displayed initially in
...
+| `solution-placement`  | "right" or "bottom"     | "right"                        | `right` shows the source and solution areas aligned side-by-side. `bottom` shows the solution area below the source area.                                                                                                                                                                                                                                                                                |
+| `partial-credit`      | "none" or "lcs"         | Depends on `grading-method`    | For the `dag` grading method, you may specify `none` for no partial credit or `lcs` (defualt) for partial credit based on the LCS edit-distance from the student solution to some correct solution. For the other grading methods, using this property is not yet supported. Grading methods `unordered` and `ranking` will always assign partial credit, and grading method `ordered` will never do so. |
```
mwest1066:
```suggestion
| `partial-credit`      | "none" or "lcs"         | Depends on `grading-method`    | For the `dag` grading method, you may specify `none` for no partial credit or `lcs` (default) for partial credit based on the LCS edit-distance from the student solution to some correct solution. For the other grading methods, using this property is not yet supported. Grading methods `unordered` and `ranking` will always assign partial credit, and grading method `ordered` will never do so. |
```

mwest1066 (approved):

PR #5114, opened 2021-11-23
Typecheck pl-order-blocks.py

SethPoulsen:
The next step in expanding type checking of the Python code. 
- Defined Python type for the `data` object that gets passed around all the element functions
- Turn on type checking for `pl-order-blocks.py`
- Let the typechecker know where to find the `prairielearn` module


SethPoulsen:
Ah, I just realized the build is failing because I didn't tell it to copy the pyright config into the docker container. Will fix.

question-servers/freeformPythonLib/prairielearn.py
```diff
@@ -17,6 +18,31 @@
...
+    raw_submitted_answers: Dict[str, Any]
+    editable: bool
+    panel: Literal['question', 'submission', 'answer']
+    extensions: Dict[str, Any]
+
+
+class ElementTestData(QuestionData):
+    test_type: Literal['correct', 'incorrect', 'invalid']
+
+
```
SethPoulsen, line 45: I'm not sure if this file is the best

I also don't like how many `Any`s there are here, there are a few different things we could do about it, but I'm not convinced which one is best yet. 

- leave them as `Any`: don't have to write as much code, but less type safe
- change them to `unknown`: individual elements will need to do explicit run time checking or casts in order to access them, but then it will be type safe
- Have individual elements make a new type inheriting from this one that declares their actual type for their use case: more code, may or may not play together well with inheriting from this type to make certain things `Final` we could end up with a complicated inheritance structure. That could be good though if it gives us really nice type safety everywhere

SethPoulsen: Oh wait, it seems like types of properties can't be redefined in subclasses, even if the new type is a subtype of the type in the parent class. That is really annoying. I guess that means that if we want to mark certain things as `Final` depending on which function we are calling (e.g. `options` should be `Final` in the `prepare` function because it can't be edited there), or if we want to more specific types for the `Any`s, we might have to make entirely new classes instead of just inheriting. That's very annoying.

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -318,6 +322,9 @@ def parse(element_html, data):
     if answer_raw_name in data['raw_submitted_answers']:
         student_answer = data['raw_submitted_answers'][answer_raw_name]
 
+    if student_answer is None:
+        student_answer = '[]'
```
nwalters512, line 326: Nit: I would simplify this and the preceding lines to something like `student_answer = data['raw_submitted_answers'].get(answer_raw_name, '[]')`.

question-servers/freeformPythonLib/prairielearn.py
nwalters512: > we might have to make entirely new classes instead of just inheriting. That's very annoying.

IMO this would be completely acceptable. These types virtually never change, so I'm not concerned about duplicating a bit of code.

question-servers/freeformPythonLib/prairielearn.py
```diff
@@ -17,6 +18,31 @@
 import collections
 
 
+# TODO in the future it would be nice to have a way to use the `Final` type
+# around some of these values for the different functions so that the type
+# checker knows which functions are allowed to modify which values.
+# This may require subclassing QuestionData, and constructing the `data`
+# dictionary through class constructors instead of the usual was
```
nwalters512, line 25: Is this comment out of date given "it seems like types of properties can't be redefined in subclasses, even if the new type is a subtype of the type in the parent class"?

nwalters512 (approved):

question-servers/freeformPythonLib/prairielearn.py
SethPoulsen: Ya, fixed.

elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: oh, duh

nwalters512 (approved):

PR #5121, opened 2021-11-29
Add back button to question manual grading page

SethPoulsen:
Without a back button on this page, you have to go through 4+ clicks just to switch which student you are grading: 

Assessments -> Assessment Name -> Manual Grading -> Question Name

This is a bit much, I think we should have the back button here. 

Currently looks like this: 
image

Which is a good start, but I'm happy for input on the UI, I'm not totally stuck on it or anything.

### DIFF
--- a/pages/partials/instructorGradingPanel.ejs
+++ b/pages/partials/instructorGradingPanel.ejs
@@ -5,6 +5,11 @@
         <% if (typeof submission_updated !== 'undefined' && submission_updated === true) { %>
             <blockquote style="color: red">Submission updated</blockquote>
         <% } %>
+        <a class="btn btn-primary mb-2" href="<%= urlPrefix %>/assessment/<%= assessment_instance.assessment_id %>/assessment_question/<%= instance_question.assessment_question_id %>/manual_grading">
+          <i class="fas fa-arrow-left"></i>
+          Back to Question
+        </a>
+
         <form name="instance_question-manual-grade-update-form" method="POST">
             <input type="hidden" name="__action" value="add_manual_grade">
             <input type="hidden" name="__csrf_token" value="<%= __csrf_token %>">


nwalters512 (approved):
Looks good from the screenshot! I'd appreciate documented instructions on how to get into + test the manual grading flow locally just so I can see what this looks/feels like for myself in the context of the entire page.

SethPoulsen:
> I'd appreciate documented instructions on how to get into + test the manual grading flow locally just so I can see what this looks/feels like for myself in the context of the entire page.

Sounds good, whereabouts should such documentation go? 

Also, it would be really nice to have some proper testing for this manual grading stuff, even just to make sure that the multiple different SQL queries involved are behaving in a sensible way. The trouble is, you have to get the database into a rather complicated state (multiple students will multiple submissions to multiple problems) before you can even do meaningful tests. Do we have infrastructure for doing this sort of testing? 

These last few weeks have been my first time touching the database at all, so I'm just getting familiar with that part of the codebase.

nwalters512:
> Sounds good, whereabouts should such documentation go?

I really hate this answer, but... until we can get our documentation into a better state, just dump it in a new file under `docs` and make it available under the ["Developer Guide"](https://github.com/PrairieLearn/PrairieLearn/blob/4c13318aaba9fb20440e1ccbd1885cbe48593e47/mkdocs.yml#L43) 😅 

> Do we have infrastructure for doing this sort of testing?

Roughly speaking, yes. We do some of the various pieces of this (multiple submissions, multiple problems, multiple users) across various test files, e.g. https://github.com/PrairieLearn/PrairieLearn/blob/master/tests/testGroupScoreAndSync.js and https://github.com/PrairieLearn/PrairieLearn/blob/master/tests/testHomework.js. I will be the first to say that it's not exactly _good_ infrastructure, but it should certainly be possible to piece some meaningful tests together. I'd rather have suboptimal tests that we have to refactor someday than no tests at all!

PR #5122, opened 2021-11-29
Use pytest to execute Python tests

### LINKED ISSUE #5073: Use pytest to discover and run Python tests
As discussed in https://github.com/PrairieLearn/PrairieLearn/pull/5066, we should have a better setup for writing Python tests. Currently, we'd need to add a new command to the Makefile for every new test we add, which is cumbersome and requires a new Python process per test, which is slower than necessary.

We should use [`pytest`](https://docs.pytest.org/) to automatically discover and run any test (a file with the pattern `*_test.py`) in the PrairieLearn repo. This reduces the barrier to adding tests, so hopefully we can gain additional test coverage!

neneal:
I've implemented the use of pytest as outlined in #5073. I also updated the Makefile to use pytest to search the entire repository for Python tests. As for the existing Python tests specified in the Makefile with the test-python directive, they have been converted to use the pytest format instead of unittest.

Closes #5073.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,8 +29,8 @@ test-grader-host:
 test-nocoverage: start-support
 	@mocha tests/index.js
 test-python:
-	@python3 -m pytest
-
+	@python3 -m pytest --ignore graders/python/python_autograder/pl_unit_test.py
+	
 lint: lint-js lint-python
 lint-js:
 	@eslint --ext js "**/*.js"
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,9 @@ test-grader-host:
 test-nocoverage: start-support
 	@mocha tests/index.js
 test-python:
+# `pl_unit_test.py` has an unfortunate file name - it matches the pattern that
+# pytest uses to discover tests, but it isn't actually a test file itself. We
+# explicitly exclude it here.
 	@python3 -m pytest --ignore graders/python/python_autograder/pl_unit_test.py
 	
 lint: lint-js lint-python
--- a/Makefile
+++ b/Makefile
@@ -29,8 +29,7 @@ test-grader-host:
 test-nocoverage: start-support
 	@mocha tests/index.js
 test-python:
-	@python3 question-servers/freeformPythonLib/prairielearn_test.py
-	@python3 elements/pl-order-blocks/dag_checker_test.py
+	@python3 -m pytest
 
 lint: lint-js lint-python
 lint-js:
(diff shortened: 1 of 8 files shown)


neneal:

Makefile
```diff
@@ -29,8 +29,7 @@ test-grader-host:
 test-nocoverage: start-support
 	@mocha tests/index.js
 test-python:
-	@python3 question-servers/freeformPythonLib/prairielearn_test.py
-	@python3 elements/pl-order-blocks/dag_checker_test.py
+	@python3 -m pytest
```
nwalters512, line 32: This doesn't matter in CI because the `graders` directory isn't included in the Docker image where tests are run, but it does impact me running `make test-python` locally.

```suggestion
# `pl_unit_test.py` has an unfortunate file name - it matches the pattern that
# pytest uses to discover tests, but it isn't actually a test file itself. We
# explicitly exclude it here.
	@python3 -m pytest --ignore graders/python/python_autograder/pl_unit_test.py
```

elements/pl-order-blocks/dag_checker_test.py
```diff
@@ -35,15 +34,11 @@
 ]
 
 
-class TestDAGChecker(unittest.TestCase):
+class TestDAGChecker:
```
nwalters512, line 37: Wrapping tests in a class is unnecessary with `pytest` - let's make `test_grade_dag` a top-level function.

question-servers/freeformPythonLib/prairielearn_test.py
```diff
@@ -4,19 +4,14 @@
 sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '../../lib')))
 
 import prairielearn as pl  # noqa: E402
-import unittest            # noqa: E402
 import lxml.html           # noqa: E402
 
 
-class TestPrairielearnLib(unittest.TestCase):
+class TestPrairielearnLib:
```
nwalters512, line 10: Wrapping tests in a class is unnecessary with `pytest` - let's make `test_inner_html` a top-level function.

nwalters512 (commented):
Thanks for doing this! Just a few small tweaks and this should be good to go.

neneal:
Good catch with the `pl_unit_test.py` file. I made your suggested changes on the branch. Thanks for the feedback!

Makefile
nwalters512: @neneal can you please update the makefile to include the comment I suggested here? Documenting the rationale behind edge cases like this is really important.

neneal: Added.

PR #5179, opened 2021-12-01
fix: Build changed images on master

### LINKED ISSUE #5178: Changes to plbase image are not reflected in master build
I noticed this after we landed https://github.com/PrairieLearn/PrairieLearn/pull/5122. That PR added the `pytest` dependency to `plbase`, but when the master CI build ran, it was still using the _old_ version of `plbase` that did not include that dependency, so it failed. We should ensure that either a) the master build _also_ rebuilds `plbase` for itself, or that b) the `Images` workflow that'll publish a new version of `plbase` completes before the main workflow executes.

nwalters512:
Fixes #5178.

### DIFF
--- a/tools/build-image-if-needed.sh
+++ b/tools/build-image-if-needed.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -e
+set -ex
 
 if [ "$#" -ne 2 ]; then
   echo "USAGE: $0 build_directory tag_name" >& 2
--- a/tools/build-image-if-needed.sh
+++ b/tools/build-image-if-needed.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -ex
+set -e
 
 if [ "$#" -ne 2 ]; then
   echo "USAGE: $0 build_directory tag_name" >& 2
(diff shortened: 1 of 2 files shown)


tools/build-image-if-needed.sh
```diff
@@ -11,7 +11,17 @@ fi
...
+
+# If this script is being run *on* the master branch, then we want to diff
+# with the previous commit on master. Otherwise, we diff with master itself.
+if [[ "$BRANCH" == "master" ]]; then
+  DIFF_SOURCE="HEAD^1"
+else
+  DIFF_SOURCE="remotes/origin/master"
+fi
+
+if git diff --exit-code $DIFF_SOURCE...HEAD -- ${BUILD_DIRECTORY}; then
```
nwalters512, line 24: I still can't tell from the Git docs if `...` or `..` is actually desired here 🙃

mwest1066: I believe `...` is what we want. Consider:
```
branch_origin ---- master_changes --- master
        |
        +------ branch_changes ---- HEAD
```

Assume that both `branch_changes` AND `master_changes` introduce changes to `BUILD_DIRECTORY`. If we use `..` then we will get a diff that _adds_ the changes from `branch_changes` and _removes_ the changes from `master_changes` (these are the changes needed to turn `master` into `HEAD`). On the other hand, if we use `...` then we _only_ get the changes in `branch_changes` (these are the changes needed to turn `branch_origin` into `HEAD`).

We only want to build the image if we've introduced changes in the current branch (we don't care about what `master` has done in the meantime) so we want `...`.

Does that sound correct to you?

mwest1066 (approved):

tools/build-image-if-needed.sh
nwalters512: That sounds correct! I'll roll with that.

PR #5182, opened 2021-12-07
fix: Don't cache rejected chunk promises

### LINKED ISSUE #3939: Chunk download caching also caches rejections
We add the promise for the chunk fetch to the caching map: https://github.com/PrairieLearn/PrairieLearn/blob/d65d09050494d13f2d6937341c2eaddf7e7a17f4/lib/chunks.js#L753 but we never remove it.

This means that we will cache promise rejections and not properly retry (and we are accumulating memory in the Map). To fix this I think we need to add a `finally()` handler to clean up. We could use a wrapper around `ensureChunk()` which does something like:
```
const ensureChunkOneAtATime = async (courseId, chunk, pendingChunkKey) => {
    const pendingChunkPromise = pendingChunksMap.get(pendingChunkKey);
    if (pendingChunkPromise) {
        // If this chunk is already being loaded, reuse the existing promise
        return pendingChunkPromise;
    }
    const chunkPromise = ensureChunk(courseId, chunk).finally(() => pendingChunksMap.delete(pendingChunkKey));
    // is there are race condition here because we are installing the delete handler before we set the Map entry?
    pendingChunksMap.set(pendingChunkKey, chunkPromise);
    return chunkPromise;
};
```

Alternatively, perhaps we could use [async.memoize()](https://caolan.github.io/async/v3/docs.html#memoize) which does not cache errors. Memoization is a bit of a memory leak, of course, but that might be ok.

nwalters512:
As described in #3939, we were previously caching chunk promises indefinitely. This is a memory leak (although we haven't yet had it impact anything in production) and also means we were caching _rejected_ promises. So, if a chunk failed to load, it would fail to load forever since we had cached the failure.

This PR ensures that a chunk loading promise is removed from our cache once it settles.

Closes #3939.

### DIFF
--- a/lib/chunks.js
+++ b/lib/chunks.js
@@ -938,6 +938,14 @@ module.exports.ensureChunksForCourseAsync = async (courseId, chunks) => {
     }
     const chunkPromise = ensureChunk(courseId, chunk);
     pendingChunksMap.set(pendingChunkKey, chunkPromise);
+    chunkPromise.finally(() => {
+      // Once the promise has settled, remove it from our collection of
+      // pending promises. This helps prevent memory leaks and, more
+      // importantly, ensures we don't cache rejected promises - if loading
+      // a chunk fails for some reason, this will ensure we try to load it
+      // again the next time it's requested.
+      pendingChunksMap.delete(pendingChunkKey);
+    });
     return chunkPromise;
   });
 


mwest1066 (approved):
Very elegant solution!

PR #5192, opened 2021-12-14
Submissions-artifact-pdf-upload: Part 2 - Decode PDF Upload, Upload Files to API, and Fetch/Display on Question View

andrewstec:


pages/home/home.ejs
```diff
@@ -77,6 +77,12 @@
                         <a href="https://prairielearn.readthedocs.io/en/latest" class="btn btn-<%= isAuthenticated ? 'xs my-0 align-text-bottom' : 'sm mb-1' %> btn-outline-primary justify-content-center">
                           View docs
                         </a>
+                        <a href="<%= plainUrlPrefix %>/scrap_paper" class="btn btn-<%= isAuthenticated ? 'xs my-0 align-text-bottom' : 'sm mb-1' %> btn-outline-primary justify-content-center">
```
andrewstec, line 80: I would recommend disabling this until we can get feedback from a small group of testing users.

lib/question.js
```diff
@@ -1894,6 +1896,59 @@ module.exports = {
             callback(null);
           });
         },
+        (callback) => {
+          // if barcodes found, get available pdf files
+          const barcodedSubmissions = locals.submissions
+            .slice(Math.max(locals.submissions.length - 5, 0)) // only last 5 displayed
```
andrewstec, line 1902: In my latest test, more than 5 submissions can be displayed on a question element. I am not too sure if we need this math max bit of code any longer. Has the max 5 displayed submissions rule changed?

tests/index.js
```diff
@@ -31,6 +31,7 @@ require('./testShowClosedAssessmentScore');
 require('./testActiveAccessRestriction');
 require('./testGradeRate');
 require('./testBonusPoints');
+require('./testScrapPaper'); // must go before testCron.js or breakage only in CI/CD pipeline; view commits before and after #541987f
 require('./testAccess');
```
andrewstec, line 35: For some reason, if the testCron.js job runs first, the after() function in the tests cannot complete in `testScrapPaper.js`.

andrewstec: This only occurs in CI/CD. Inside the docker container on my laptop, this is not an issue.

lib/scrapPaperReader.js
```diff
@@ -1,5 +1,80 @@
+const config = require('./config');
+const path = require('path');
+const { v4: uuidv4 } = require('uuid');
+const fileStore = require('../lib/file-store');
+const fs = require('fs').promises;
+
+const quagga = require('quagga').default;
+const imagemagick = require('imagemagick');
```
nwalters512, line 8: From https://github.com/rsms/node-imagemagick:

> Note: This code has been unmaintained for a long time. Please consider using the gm module instead.

I'd very much like to heed their advice.

lib/scrapPaperReader.js
```diff
@@ -10,4 +85,410 @@ const readPdf = async (pdfBuffer) => {
...
+const processScrapPaperPdf = async (pdfBuffer, originalName, userId, logger) => {
+  const job = async () => _processScrapPaperPdf(pdfBuffer, originalName, userId, logger);
+  Queue.enqueue(job, logger);
+};
+
+const _decodeJpegs = async (jpegs, logger) => {
+  for (let i = 0; i < jpegs.length; i++) {
+    const start = new Date();
+    logger.info(`_decodeJpegs() - start decoding of ${jpegs[i].jpegFilename} / page ${i}`);
+    jpegs[i]['barcode'] = await _decodeJpeg(jpegs[i], logger);
```
nwalters512, line 238: This isn't so much decoding an image as it is extracting/reading a barcode. I'd recommend naming this something like `extractBarcode`. Ditto for the enclosing `_decodeJpegs` function.

lib/scrapPaperReader.js
```diff
@@ -10,4 +85,410 @@ const readPdf = async (pdfBuffer) => {
...
+  return jpegs;
+};
+
+/**
+ * Detects and extracts a barcode on a jpeg file
+ * We probably want to start working on some image analysis that helps read
+ * barcodes based on uploaded scans. Many likely will not read. This will require
+ * some optimization here.
+ * @param {string} jpeg jpeg file object from arr produced by _convertPdf()
+ * @returns {string} a code-128 formatted barcode or undefined if not found
```
nwalters512, line 253:
```suggestion
 * @returns {Promise<string>} a code-128 formatted barcode or undefined if not found
```

lib/scrapPaperReader.js
```diff
@@ -10,4 +85,410 @@ const readPdf = async (pdfBuffer) => {
...
+  }
+  return jpegs;
+};
+
+/**
+ * Detects and extracts a barcode on a jpeg file
+ * We probably want to start working on some image analysis that helps read
+ * barcodes based on uploaded scans. Many likely will not read. This will require
+ * some optimization here.
+ * @param {string} jpeg jpeg file object from arr produced by _convertPdf()
```
nwalters512, line 252: I won't leave comments on every function, but: if you're going to include type annotations, please ensure that they a) include all parameters and b) correctly reflect the type of the parameters. For instance, this function is missing a `logger` param and `jpeg` is most definitely not a string.

lib/scrapPaperReader.js
```diff
@@ -10,4 +85,410 @@ const readPdf = async (pdfBuffer) => {
...
+          numOfWorkers: 0,
+          inputStream: {
+            size: 1920, // restrict input-size resolution to be 1920 in width (long-side)
+          },
+          decoder: {
+            readers: ['code_128_reader'],
+          },
+        },
+        (result) => {
+          if (result && result.codeResult && result.codeResult.code) {
```
nwalters512, line 318: Nit, optional chaining can make this a little more legible:
```suggestion
          if (result?.codeResult?.code) {
```

lib/scrapPaperReader.js
```diff
@@ -10,4 +85,410 @@ const readPdf = async (pdfBuffer) => {
...
+};
+
+/**
+ * Splits jpeg image into vertical slices to work with Quagga barcode reader.
+ * @param {number} pageNum the page number folder that the segmented files should be placed in
+ * @param {string} jpegFilepath filepath referencing jpeg full page image
+ * @param {string} workingDir working directory where image processing occurs
+ * @returns {string} contains filepaths referencing jpeg segments
+ */
+const _segmentJpeg = async (pageNum, jpegFilepath, workingDir, logger, optionNum) => {
```
nwalters512, line 338: What is `optionNum`?

lib/scrapPaperReader.js
```diff
@@ -10,4 +85,410 @@ const readPdf = async (pdfBuffer) => {
...
+        '-quality',
+        '100',
+        '-adaptive-sharpen',
+        '0x3',
+        filepath,
+      ],
+      (err) => {
+        // stdout 2nd arg, but always empty
+        if (err) {
+          reject(err);
```
nwalters512, line 379: You should `return` after this or else the promise will always resolve even after it rejects.

pages/instructorScanPaper/instructorScanPaper.ejs
```diff
@@ -0,0 +1,37 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <%- include('../partials/head'); %>
+    <link href="<%= asset_path('stylesheets/issues.css') %>" rel="stylesheet" />
+	<title></title>
```
nwalters512, line 6:
```suggestion
```

pages/instructorScanPaper/instructorScanPaper.ejs
```diff
@@ -0,0 +1,37 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <%- include('../partials/head'); %>
+    <link href="<%= asset_path('stylesheets/issues.css') %>" rel="stylesheet" />
```
nwalters512, line 5: Blind copy/paste?
```suggestion
```

pages/instructorScanPaper/instructorScanPaper.ejs
```diff
@@ -0,0 +1,37 @@
...
+	<title></title>
+</head>
+<body>
+    <%- include('../partials/navbar', {navPage: 'news_item'}); %>
```
nwalters512, line 9: This is not a `news_item` page.

pages/instructorScanPaper/instructorScanPaper.ejs
```diff
@@ -0,0 +1,37 @@
...
+                        <div class="form-group">
+                            <label for="pdf-collection">Choose PDF file</label>
+                            <input class="form-control-file" id="pdf-collection" name="file" type="file" accept="application/pdf">
+                        </div>
+                        <button type="submit" class="btn btn-primary">Submit</button>
+                    </div>
+                </form>
+            </div>
+            </div>
+        </div>
```
nwalters512, line 34: Indentation and nesting is all over the place here, and I think you have a few extraneous closing tags. Please patch this up.

pages/instructorScanPaper/instructorScanPaper.ejs
```diff
@@ -0,0 +1,37 @@
...
+<body>
+    <%- include('../partials/navbar', {navPage: 'news_item'}); %>
+
+    <div id="content" class="container">
+        <div class="card mb-4">
+            <div class="card-header bg-primary text-white">Scan a PDF Collection of Barcoded Scrap Paper Documents</div>
+
+            <div class="card-body scan-paper-description">
+                <p>Each page of the PDF collection that you are uploading should contain a barcode. The PDF should not be greater than 25MB. High-resolution, but compressed images, are recommended for each page of the PDF upload.</p>
+                <p>970 out of 1000 barcodes are currently decoded from PDF upload in testsuite and further optimization is needed. Feedback for optimization is requested. Please share scanning resolution and other related success/failures in #pl-dev to help in the ongoing development of this feature.</p>
```
nwalters512, line 17:
```suggestion
                <p>970 out of 1000 barcodes are currently decoded from PDF upload in testsuite and further optimization is needed. Feedback for optimization is requested. Please share scanning resolution and other related success/failures in the #pl-dev Slack channel to help in the ongoing development of this feature.</p>
```

pages/instructorScanPaper/instructorScanPaper.js
```diff
@@ -0,0 +1,80 @@
...
+const debug = require('debug')('prairielearn:' + path.basename(__filename, '.js'));
+
+const { processScrapPaperPdf } = require('../../lib/scrapPaperReader');
+const serverJobs = require('../../lib/server-jobs');
+
+const error = require('../../prairielib/lib/error');
+const ERR = require('async-stacktrace');
+
+router.get('/', (req, res) => {
+  res.render(__filename.replace(/\.js$/, '.ejs'), res.locals);
```
nwalters512, line 13: As with the other PR, you'll need an authz check here too.

pages/instructorScanPaper/instructorScanPaper.js
```diff
@@ -0,0 +1,80 @@
...
+});
+
+router.post('/', function (req, res, next) {
+  if (!res.locals.authz_data.has_course_instance_permission_edit) return next();
+  if (req.body.__action === 'scan_scrap_paper') {
+    if (!req.file) {
+      return next(error.make(400, 'Missing barcoded pdf collection file data'));
+    }
+    if (!res.locals || !res.locals.authn_user || !res.locals.authn_user.user_id) {
+      return next(error.make(400, 'Authn_user required on file-store API'));
```
nwalters512, line 23: This error message is pretty opaque. Can you improve it?

pages/instructorScanPaper/instructorScanPaper.js
```diff
@@ -0,0 +1,80 @@
...
+
+router.get('/', (req, res) => {
+  res.render(__filename.replace(/\.js$/, '.ejs'), res.locals);
+});
+
+router.post('/', function (req, res, next) {
+  if (!res.locals.authz_data.has_course_instance_permission_edit) return next();
+  if (req.body.__action === 'scan_scrap_paper') {
+    if (!req.file) {
+      return next(error.make(400, 'Missing barcoded pdf collection file data'));
```
nwalters512, line 20:
```suggestion
      return next(error.make(400, 'Missing PDF file'));
```

pages/instructorScanPaper/instructorScanPaper.js
```diff
@@ -0,0 +1,80 @@
...
+          type: 'decode_pdf_collection',
+          description:
+            'Decodes each barcode on each page in the pdf collection used as scrap paper by students.',
+          job_sequence_id: jobSequenceId,
+          last_in_sequence: false,
+        };
+
+        return serverJobs.createJobAsync(jobOptions);
+      })
+      .then((job) => {
```
nwalters512, line 49: Let's rewrite this with async/await to avoid the need for explicit promise chaining.

pages/partials/submission.ejs
```diff
@@ -24,6 +24,11 @@
   <div class="collapse <% if (submissionIdx == submissionCount) { %>show<% } %>" id="submission-<%= submission.id %>-body">
     <div class="card-body submission-body" >
       <%- submissionHtml %>
+      <% if (typeof submission != 'undefined' && typeof submission.pdf_barcode_scan != 'undefined') { %>
+        <div class="submission-body-pdf-barcode-scan-container">
+          <iframe class="submission-body-pdf-barcode-scan" width="100%" type="text/html" src="data:application/pdf;base64,<%= submission.pdf_barcode_scan %>"></iframe>
```
nwalters512, line 29: Are there missing custom styles for these? I see classes but no corresponding CSS styles.

pages/instructorScanPaper/instructorScanPaper.js
```diff
@@ -0,0 +1,80 @@
...
+      .catch((err) => {
+        if (ERR(err, next)) return;
+      });
+
+    // TO DO:
+
+    // discuss how/if we want to handle multiple barcodes on same submission ie.
+    // 1. we could use javascript to automatically add additional barcode submission element if an option is included in ,
+    // 2. store multiple submission referneces in barcodes table (probably need a barcode_submissions table)
+    // 3. decide if we need to do this now or can do it later with a migration to keep backwards compatibility.
```
nwalters512, line 69: Does this TODO belong here? This seems orthogonal to the scanning+uploading process.

tests/testScrapPaper.js
```diff
@@ -171,26 +172,26 @@ describe('Barcode generation, student submission, and scanning process', functio
...
-      //     assert.equal(recomputedChecksum, checksum);
-      //     validBarcodes.push(jpeg.barcode);
-      //   });
-      // });
+      it('barcodes should ALL be scannable by barcode reader', async () => {
+        decodedJpegs = await decodeBarcodes(
+          pdfBuffer,
+          `${config.imageProcessingDir}/test`,
+          'Any original filename.pdf',
+          console
```
nwalters512, line 180: Ideally this should be a no-op logger instead of dumping everything to the console.

nwalters512:
I also noticed a ton of logs like this in the output after I uploaded a PDF - we'll need to remove them before this gets merged.

```
[ 583, 360, 4 ]
FrameGrabber {"videoSize":[583,360],"canvasSize":[1920,1185],"stepSize":[0.30364583333333334,0.3037974683544304],"size":[1920,1152],"topRight":{"x":0,"y":0}}
[ 583, 32, 4 ]
FrameGrabber {"videoSize":[583,32],"canvasSize":[1920,105],"stepSize":[0.30364583333333334,0.3047619047619048],"size":[1920,96],"topRight":{"x":0,"y":0}}
[ 583, 360, 4 ]
FrameGrabber {"videoSize":[583,360],"canvasSize":[1920,1185],"stepSize":[0.30364583333333334,0.3037974683544304],"size":[1920,1152],"topRight":{"x":0,"y":0}}
```

lib/scrapPaperReader.js
```diff
@@ -1,5 +1,80 @@
...
 const pdfjsLib = require('pdfjs-dist/legacy/build/pdf.js'); // legacy means node support
 
+const sqldb = require('../prairielib/lib/sql-db');
+const sqlLoader = require('../prairielib/lib/sql-loader');
+const sql = sqlLoader.loadSqlEquiv(__filename);
+
+const processingDir = config.imageProcessingDir;
+const segmentOptions = [
+  // NOTE for Developer: Add/remove segmentation possibilities once you have real PDF uploads to analyze
+  // We need to collect data. Overall this shotgun approach works, but it is slow.
```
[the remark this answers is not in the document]
andrewstec: We have the ability to determine what segmentation and Quagga configuration frame-size combination used for decoding a PDF successfully or unsuccessfully. 

Some combinations have better success rates. About 977/1000 are decoded with a manual test of 1000 pages submitted directly after the PDF creation.

We need to prioritize the highest success rates based on our decoding configuration and segmentation option, but we also want to set up a new dimension to analyze by based on the quality of the upload (ie. how many pixels in the PDF and other metadata ie. color versus b&w). 

The metadata for a PDF upload could go in a new table that could be removed after analysis has been deemed complete. But you could consider permanently adding the metadata for the what successful segmentation and quagga frame-size led the barcode to be decoded.

Those are the key points one can consider, but there are obviously a lot of data experts in the area who can give additional advice. If anyone is somehow familiar with graphic processing and barcode reading, that's a massive bonus.

lib/scrapPaperReader.js
```diff
@@ -1,5 +1,80 @@
...
+  ['-sharpen', '0x7', '-crop', 'x400', '+repage'],
+  ['-crop', 'x400', '+repage'],
+  ['-resize', '400%', '-adaptive-sharpen', '0x3', '-crop', 'x400', '+repage'],
+];
+class Queue {
+  static queue = [];
+  static pendingPromise = false;
+
+  static notifyPositions(queued) {
+    queued.forEach((item, i) => item.logger.info(`Queued in position ${i}\n.`));
```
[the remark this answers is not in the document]
andrewstec: Sounds good!

pages/instructorScanPaper/instructorScanPaper.js
andrewstec: Does async await work on the route functions? I thought promise chains were needed.

pages/instructorScanPaper/instructorScanPaper.js
andrewstec: Done.

pages/instructorScanPaper/instructorScanPaper.js
andrewstec: Done.

lib/scrapPaperReader.js
```diff
@@ -1,5 +1,80 @@
...
+  }
+
+  static enqueue(promise, logger) {
+    return new Promise((resolve, reject) => {
+      logger.info(`PDF Barcode Processing Queued in Position ${this.queue.length}\n.`);
```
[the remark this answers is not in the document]
andrewstec: If people want to know where their PDF is in the queue and if it is working, we should keep this. Some user feedback is needed.

lib/scrapPaperReader.js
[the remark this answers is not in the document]
andrewstec: On retrospect, it's a bad idea to take out the user feedback. They will have no idea what is happening on the back-end or if their job is ever coming up.

lib/scrapPaperReader.js
```diff
@@ -10,4 +85,410 @@ const readPdf = async (pdfBuffer) => {
   return pdfjsLib.getDocument(pdfBuffer).promise;
 };
 
-module.exports = { readPdf };
+/**
+ * Helper method to upload a pdf page to S3 via file-store API
+ * @param {Array<object>} decodedJpegs array jpeg file meta data object list
+ * @param {Array<object>} submissions contains assessment_instance_id, instance_question_id, submission id meta data
```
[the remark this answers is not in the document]
andrewstec: Done.

lib/scrapPaperReader.js
```diff
@@ -10,4 +85,410 @@ const readPdf = async (pdfBuffer) => {
...
+        updateValues = updateValues.substring(0, updateValues.length - 1);
+      }
+    }
+
+    const query = sql.update_barcodes.replace('$updateValues', updateValues);
+    const updated = (await sqldb.queryAsync(query, {}))[3];
+    return updated;
+  }
+
+  // TO DO FIX: Does not work when I rely on queryAsync to substitute values.
```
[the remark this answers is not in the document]
andrewstec: This is just so you are aware why I took another approach. The TO DO can be removed now. I think the substitution is adding in a character that isn't playing friendly in the query.

I manually substituted it with ```    const query = sql.update_barcodes.replace('$updateValues', updateValues);```

I didn't have time to debug the issue further. Hence the dirty quick note.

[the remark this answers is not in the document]
andrewstec: Removed the TO DO:

```


  // TO DO FIX: Does not work when I rely on queryAsync to substitute values.
  // -- BLOCK update_barcodes_with_submission

  // DROP TABLE IF EXISTS barcode_submission_ids_temp;

  // CREATE TEMP TABLE barcode_submission_ids_temp
  //     (submission_id BIGINT NOT NULL PRIMARY KEY, barcode TEXT);

  // INSERT INTO
  //     barcode_submission_ids_temp(submission_id, barcode)
  // VALUES
  //     $updateValues;
  //     ^
  //     |
  //     + ERROR POSITION SHOWN ABOVE

  // UPDATE
  //     barcodes b
  // SET
  //     submission_id = bs_temp.submission_id
  // FROM
  //     barcode_submission_ids_temp bs_temp
  // WHERE
  //     b.barcode = bs_temp.barcode
  // RETURNING *;
  // SQL params:

  // {
  //     "updateValues": "(1, '2D581')"
  // }

```

lib/scrapPaperReader.js
```diff
@@ -10,4 +85,410 @@ const readPdf = async (pdfBuffer) => {
...
+const _processScrapPaperPdf = async (pdfBuffer, originalName, userId, logger) => {
+  const workingDir = path.join(processingDir, uuidv4());
+
+  try {
+    const allResults = await decodeBarcodes(pdfBuffer, workingDir, originalName, logger);
+    const readBarcodedPages = allResults.filter((decodedPdfPage) => decodedPdfPage.barcode);
+
+    logger.info(`JSON decoder uploaded pages: ${allResults.length}`);
+    logger.info(`JSON decoder successfully read barcodes on pages: ${readBarcodedPages.length}`);
+    logger.info(`JSON decoder all results data dump: ${JSON.stringify(allResults, null, 2)}`);
```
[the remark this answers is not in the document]
andrewstec: This is what the user would see on their view, as they are waiting for the decoding to occur. If you want the user to see this information, I believe you want to keep it as logger.info. 

It is probably safer to give the user more information to see vaguely what is happening during the debugging process.

lib/scrapPaperReader.js
andrewstec: That really goes without saying. It is almost as if there is more than meets the eye here. 

Did you want the comments out then? Just let me know what you prefer.

tests/testScrapPaper.js
andrewstec: Can we get a no op logger without implementing more code?

pages/instructorScanPaper/instructorScanPaper.js
andrewstec: Done.

tests/testScrapPaper.js
andrewstec: Done.

pages/instructorScanPaper/instructorScanPaper.ejs
andrewstec: Done.

lib/scrapPaperReader.js
```diff
@@ -10,4 +85,410 @@ const readPdf = async (pdfBuffer) => {
...
 };
 
-module.exports = { readPdf };
+/**
+ * Helper method to upload a pdf page to S3 via file-store API
+ * @param {Array<object>} decodedJpegs array jpeg file meta data object list
+ * @param {Array<object>} submissions contains assessment_instance_id, instance_question_id, submission id meta data
+ * @return {Array<object>} metadata
+ */
+const _uploadPages = async (decodedPdfs, fileMetadata, userId) => {
```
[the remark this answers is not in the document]
andrewstec: It's generally used in the JavaScript ecosystem where private classes cannot be declared.

lib/scrapPaperReader.js
```diff
@@ -10,4 +85,410 @@ const readPdf = async (pdfBuffer) => {
...
+  // RETURNING *;
+  // SQL params:
+
+  // {
+  //     "updateValues": "(1, '2D581')"
+  // }
+};
+
+const _processScrapPaperPdf = async (pdfBuffer, originalName, userId, logger) => {
+  const workingDir = path.join(processingDir, uuidv4());
```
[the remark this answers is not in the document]
andrewstec: Outdated. It no longer does.

pages/instructorScanPaper/instructorScanPaper.js
nwalters512: Express doesn't support that natively, but we wrap routes with [`express-async-handler`](https://www.npmjs.com/package/express-async-handler) - see [this example](https://github.com/PrairieLearn/PrairieLearn/blob/2f3b359987bebf00d45a8c32d58e3b302d117c4a/pages/administratorQueries/administratorQueries.js#L14).

lib/scrapPaperReader.js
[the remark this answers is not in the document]
nwalters512: This is a module though - everything is implicitly private unless it's explicitly exported.

lib/scrapPaperReader.js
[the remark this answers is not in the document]
nwalters512: Ah, I didn't catch that this `logger` referred to a *job's* logger and not our global logger that goes to stdout. You can probably ignore most of these comments about logs then!

lib/scrapPaperReader.js
andrewstec: Done.

lib/scrapPaperReader.js
nwalters512: I am a huge proponent of type annotations in comments, I just want them to actually reflect what the types are. Otherwise that creates unnecessary cognitive overhead to understanding the code *and* makes our work more difficult when this code is ultimately type-checked in the future.

lib/scrapPaperReader.js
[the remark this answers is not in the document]
andrewstec: Yeah, it just helps people understand where to run methods from in a large file without understanding all the details, as if it were a class. It is a convention for communication, but has no technical effect.

lib/scrapPaperReader.js
andrewstec: Quagga chose to use the word decode instead of extract, so I just wanted to keep it consistent.

pages/partials/submission.ejs
andrewstec: Done. Taken out.

andrewstec: Added one back in for a test that uses it.

lib/scrapPaperReader.js
andrewstec: Patch size array. Updated.

lib/scrapPaperReader.js
```diff
@@ -10,4 +85,410 @@ const readPdf = async (pdfBuffer) => {
...
+ * @return {Array<object>} metadata
+ */
+const _uploadPages = async (decodedPdfs, fileMetadata, userId) => {
+  const uploaded = [];
+  const confirmation = [];
+  for (let i = 0; i < decodedPdfs.length; i++) {
+    for (let j = 0; j < fileMetadata.rows.length; j++) {
+      if (
+        fileMetadata.rows[j].barcode === decodedPdfs[i].barcode &&
+        confirmation.indexOf(decodedPdfs[i].barcode) === -1
```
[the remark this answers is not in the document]
andrewstec: Fixed. Using one array to be clearer.

lib/scrapPaperReader.js
```diff
@@ -10,4 +85,410 @@ const readPdf = async (pdfBuffer) => {
...
+ * A Pdf Scan is a collection of scanned barcoded single pieces of paper with
+ * student written work. It is expected that one barcode exists on each page in the PDF.
+ * If a barcode is readable by the decoder, the page will be uploaded to S3 and stored with
+ * metadata information in `files` and `barcodes` table to later allow viewing of doc by student.
+ * If a student submits a barcode after file upload, it can still be displayed on front-end.
+ * @param {Buffer} pdfBuffer buffered application/pdf scanned document
+ * @param {string} originalName uploaded filename
+ * @param {string} userId authn user id to be stored in file api
+ * @param {object} logger
+ * @returns updates barcode table rows
```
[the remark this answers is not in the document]
andrewstec: Moved this comment and fixed.

nwalters512:
The latest version of this branch is not working. I generated a PDF with part 1 and tried to scan it with part 2, and it failed to find a barcode on every page. It appears as though the tests pass. This will require more work and investigation.

server.js
```diff
@@ -123,13 +123,14 @@ module.exports.initExpress = function () {
     round: 0,
   });
   app.post(
-    '/pl/scan_paper',
+    '/pl/course_instance/:course_instance_id/instructor/scan_paper',
     // Exception for PDF uploads, which may be quite a bit bigger than other file uploads
+    // Raw pdf does not persists. Segmented into jpeg pages downstream
```
nwalters512, line 128:
```suggestion
    // This particular route accepts files up to ~50MB, as they may be
    // significantly larger than other files. Note that these PDFs are
    // processed in memory and split up into separate images; the raw
    // PDF is currently not persisted anywhere.
```

lib/question.js
nwalters512: All submissions are rendered into the page; some (currently all but the 3 most recent) are just hidden inside a collapsible panel. We'll need to fetch PDFs for all submissions here, not just the first 5.

nwalters512:
Hmmmm, even the last commit that passed the tests on CI (63e3785) is failing for me when I check it out locally.

I've updated the tests in 2e9d884 to fix the assertions so that they actually reflect the root of the problem, which is that `file_id` in the barcodes table is actually `null`. The test was passing because `isDefined` reports true even for `null`:

```sh
node -e 'require("chai").assert.isDefined(null)'
```

PR #5195, opened 2021-12-17
pl-number-input: handle fraction with denominator set to zero

### LINKED ISSUE #5194: pl-number-input creates issue if fraction divides by 0
If a pl-number-input that allows fraction receives an answer with a zero as denominator, it marks the submission as invalid (as it should), but also creates an issue:

image

I don't think an issue should be created, rather a simple format error (or similar message) should be provided to students.

jonatanschroeder:
Fixes #5194


jonatanschroeder:
Just realized that this should have been handled by https://github.com/PrairieLearn/PrairieLearn/blob/f74ea7b90ab12af0662975dde01eec44630d93f5/question-servers/freeformPythonLib/prairielearn.py#L711 but wasn't. I'll review with that in mind.

jonatanschroeder:
> Just realized that this should have been handled by
> 
> https://github.com/PrairieLearn/PrairieLearn/blob/f74ea7b90ab12af0662975dde01eec44630d93f5/question-servers/freeformPythonLib/prairielearn.py#L711
> 
> but wasn't. I'll review with that in mind.

Turns out that the problem is that numpy throws a different exception, and only if the error state is to raise an error. Updated to use numpy for it.

nwalters512 (approved):

PR #5199, opened 2021-12-18
feat: Add length limit for group name

### LINKED ISSUE #5124: Add length limit to student created groups name
Currently, there are no character limits on name length of student created groups; it is therefore possible students to deliberately create a long group name (screenshot below). Long group names also mess up the instructor UI. 

image

Please see the linked slack thread https://prairielearn.slack.com/archives/C266KEH9A/p1634856860011300?thread_ts=1634856718.011200&cid=C266KEH9A for detailed discussions.

Friendseeker:
Fixes #5124


Friendseeker:

pages/shared/groupWorkInitial.ejs
```diff
@@ -4,7 +4,7 @@
       <div class="col-sm bg-light py-4 px-4 border">
         <form id="create-form" name="create-form" method="POST">
           <h6>Group name</h6>
-          <input type="text" class="form-control" id="groupNameInput" name="groupName" placeholder="e.g. teamOne" aria-describedby="groupNameHelp">
+          <input type="text" class="form-control" id="groupNameInput" name="groupName" maxlength="10" placeholder="e.g. teamOne" aria-describedby="groupNameHelp">
```
tdy, line 7: We should probably combine this with server-side validation.

Friendseeker: Thanks for your feedback! The new commit added server-side validation.

nwalters512 (commented):
10 characters seems like a pretty small limit. Can we bump this up to something larger that'll still avoid issues with path lengths when unzipping?

Friendseeker:
> 10 characters seems like a pretty small limit. Can we bump this up to something larger that'll still avoid issues with path lengths when unzipping?

Thank you for your feedback!
Indeed 10 can be needlessly strict. Therefore, I adjusted the length limit from 10 to 30.

pages/shared/groupWorkInitial.ejs
```diff
@@ -4,7 +4,7 @@
       <div class="col-sm bg-light py-4 px-4 border">
         <form id="create-form" name="create-form" method="POST">
           <h6>Group name</h6>
-          <input type="text" class="form-control" id="groupNameInput" name="groupName" placeholder="e.g. teamOne" aria-describedby="groupNameHelp">
+          <input type="text" class="form-control" id="groupNameInput" name="groupName" maxlength="30" placeholder="e.g. teamOne" aria-describedby="groupNameHelp">
           <small id="groupNameHelp" class="form-text text-muted">Group names can only contain letters and numbers</small>
```
nwalters512, line 8: Can you update this helper text and the error message that's shown when the group name is invalid? We need to make sure that users actually know that their group name is invalid.

Friendseeker: My thought was that the HTML maxlength modifier is sufficient UX-wise to indirectly tell user that the length is capped (as when group name reach 30 characters, the user can no longer type in more characters)

But indeed, it can be better to explicitly inform user that the length is capped, hence I made a new commit to update the helper text & error message.

nwalters512 (approved):

PR #5200, opened 2021-12-20
Support custom group roles in database

renzol2:
## Feature

This PR lays the database foundation for supporting **custom group roles** in group assessments, which implements role-based collaboration directly into PrairieLearn.

Instructors can opt-in to define custom group roles for their group assessments. Roles can be assigned different permissions to facilitate role-based teamwork, such as:
- Submitting specific questions
- Viewing questions
- Assigning group roles at the start of an assessment
- Assigning group roles during an assessment 
- Submitting the assessment (in context of an exam)

Instructors would define their roles and associated permissions in the JSON schema for their group assessments. These JSON changes are automatically synced in database for retrieval later, where instructors could evaluate the performance of their students by role.

## Changes

- [x] Update JSON schema in `course-db.js`
- [x] Create migration
- [x] Edit sync parser to parse new JSON fields
- [x] Edit sync query to sync JSON changes with database
- [x] Create restrictions on JSON schema
- [x] Add tests (if needed)


sprocs/sync_assessments.sql
```diff
@@ -189,6 +195,33 @@ BEGIN
...
+                    can_assign_roles_at_start,
+                    can_assign_roles_during_assessment
+                ) VALUES (
+                    (group_role->>'role_name'),
+                    new_assessment_id,
+                    -- Insert default values where necessary
+                    CASE WHEN group_role ? 'minimum' THEN (group_role->>'minimum')::integer ELSE 0 END,
+                    CASE WHEN group_role ? 'maximum' THEN (group_role->>'maximum')::integer ELSE (valid_assessment.data->>'group_min_size')::integer END,
+                    CASE WHEN group_role ? 'can_assign_roles_at_start' THEN (group_role->>'can_assign_roles_at_start')::boolean ELSE FALSE END,
+                    CASE WHEN group_role ? 'can_assign_roles_during_assessment' THEN (group_role->>'can_assign_roles_during_assessment')::boolean ELSE FALSE END
```
renzol2, line 215: We'd like to have it so instructors don't necessarily have to specify explicitly the minimum and maximum number of group users allowed to be a specific role. In these cases, we want to have the minimum and maximum number default to some reasonable values, e.g. `minimum = 0` and `maximum = the group's minimum size`.

We're not sure if this is the cleanest way to implement this kind of default behavior - what are your thoughts @nwalters512?

migrations/245_group_role_tables__create.sql
```diff
@@ -0,0 +1,38 @@
...
+    role_name TEXT NOT NULL,
+    assessment_id BIGINT NOT NULL,
+    minimum INT DEFAULT 0,
+    maximum INT DEFAULT 4,
+    can_assign_roles_at_start BOOLEAN DEFAULT FALSE,
+    can_assign_roles_during_assessment BOOLEAN DEFAULT FALSE,
+    can_submit_assessment BOOLEAN DEFAULT TRUE,
+    CONSTRAINT group_roles_assessment_id_fkey
+        FOREIGN KEY (assessment_id)
+            REFERENCES assessments(id) ON UPDATE CASCADE ON DELETE CASCADE
```
nwalters512, line 13: Instead of separately declaring a foreign key constraint, you should do it when you create the column:

```suggestion
    assessment_id BIGINT REFERENCES assessments(id) ON UPDATE CASCADE ON DELETE CASCADE,
    minimum INT DEFAULT 0,
    maximum INT DEFAULT 4,
    can_assign_roles_at_start BOOLEAN DEFAULT FALSE,
    can_assign_roles_during_assessment BOOLEAN DEFAULT FALSE,
    can_submit_assessment BOOLEAN DEFAULT TRUE
```

migrations/245_group_role_tables__create.sql
```diff
@@ -0,0 +1,38 @@
+-- Create table and index for group roles
+CREATE TABLE IF NOT EXISTS group_roles (
+    id BIGSERIAL PRIMARY KEY,
+    role_name TEXT NOT NULL,
+    assessment_id BIGINT NOT NULL,
+    minimum INT DEFAULT 0,
+    maximum INT DEFAULT 4,
```
nwalters512, line 7: I would be inclined to default `maximum` to NULL. A max of 4 seems completely arbitrary, unless there's a specific reason this is like this?

schemas/schemas/infoAssessment.json
```diff
@@ -136,6 +143,41 @@
...
+                "minimum": {
+                    "description": "The minimum number of users that should be in this role in a group.",
+                    "type": "number",
+                    "default": 1
+                },
+                "maximum": {
+                    "description": "The maximum number of users that should be in this role in a group.",
+                    "type": "number",
+                    "default": 1
+                },
```
nwalters512, line 163: AFAICT the defaults of `1` here don't match the implementation.

sprocs/sync_assessments.sql
nwalters512: Ah, I mentioned this above too, but I think we can default to a min of 0 and no max. Is there a certain reason why you were thinking of defaulting the max to the minimum group size?

sprocs/sync_assessments.sql
```diff
@@ -347,6 +380,85 @@ BEGIN
                         question_id = EXCLUDED.question_id
                     RETURNING aq.id INTO new_assessment_question_id;
                     new_assessment_question_ids := array_append(new_assessment_question_ids, new_assessment_question_id);
+
+                    IF (valid_assessment.data->>'group_work')::boolean THEN
+                        -- Create temporary tables for roles that can view the question
```
nwalters512, line 385: Can you clarify why this works with temporary tables instead of just inserting directly into `assessment_question_role_permissions`?

sprocs/sync_assessments.sql
```diff
@@ -189,6 +195,33 @@ BEGIN
                 student_authz_join = EXCLUDED.student_authz_join,
                 student_authz_leave = EXCLUDED.student_authz_leave,
                 deleted_at = NULL;
+
+            -- Insert all group roles
+            FOR group_role IN SELECT * FROM JSONB_ARRAY_ELEMENTS(valid_assessment.data->'groupRoles') LOOP
```
nwalters512, line 200: This won't handle the case where a group role is deleted.

sprocs/sync_assessments.sql
```diff
@@ -347,6 +380,85 @@ BEGIN
...
+                            group_role_id BIGINT,
+                            can_view BOOLEAN,
+                            can_submit BOOLEAN
+                        ) ON COMMIT DROP;
+
+                        INSERT INTO temp_role_permissions
+                        SELECT * FROM roles_can_view NATURAL LEFT JOIN roles_can_submit;
+
+                        FOR permission in SELECT * FROM temp_role_permissions LOOP
+                            INSERT INTO assessment_question_role_permissions (
```
nwalters512, line 443: You'll need to handle the case where a role permission is removed, e.g.

```diff
{
-  "canSubmit": ["Manager", "Recorder"],
+  "canSubmit": ["Manager"],
}
```

sync/fromDisk/assessments.js
```diff
@@ -151,6 +161,8 @@ function getParamsForAssessment(assessmentInfoFile, questionIds) {
             forceMaxPoints: question.forceMaxPoints || false,
             triesPerVariant: question.triesPerVariant || 1,
             gradeRateMinutes: questionGradeRateMinutes,
+            canView: question.canView || null,
+            canSubmit: question.canSubmit || null,
```
nwalters512, line 165: So, the pattern we've used elsewhere is to define all defaults in JS instead of setting `null` here and checking for null in the syncing sproc. I'd be inclined to do the same here.

sync/fromDisk/assessments.js
```diff
@@ -210,6 +224,17 @@ function getParamsForAssessment(assessmentInfoFile, questionIds) {
...
+  const groupRoles = assessment.groupRoles || [];
+  assessmentParams.groupRoles = groupRoles.map((role) => {
+    return {
+      role_name: role.name,
+      minimum: role.minimum,
+      maximum: role.maximum,
+      can_assign_roles_at_start: role.canAssignRolesAtStart,
+      can_assign_roles_during_assessment: role.canAssignRolesDuringAssessment,
+    };
+  });
```
nwalters512, line 236:
```suggestion
  assessmentParams.groupRoles = (assessment.groupRoles ?? []).map((role) => {
    return {
      role_name: role.name,
      minimum: role.minimum,
      maximum: role.maximum,
      can_assign_roles_at_start: role.canAssignRolesAtStart,
      can_assign_roles_during_assessment: role.canAssignRolesDuringAssessment,
    };
  });
```

nwalters512 (commented):
I really appreciate you doing the database work ahead of any UI stuff - that makes this really easy to review! I left a number of inline comments, but here are a few more high-level notes:

- We should do more validation at sync time. The fact that everything is validated against the JSON schema gets us halfway there, but we should also do things like validating that the entries in `canSubmit`, etc. correspond to actual group roles names. This would fail when we actually try to sync to the database, but we can provide nicer error messages if we handle it in JS. See https://github.com/PrairieLearn/PrairieLearn/blob/b21d66f54ebf412f3bc9b3d8beb7cb851b69989b/sync/course-db.js#L1086 for the existing implementation of assessment validation. We should report errors if any role permissions don't correspond to actual roles.
- Let's add some tests! See https://github.com/PrairieLearn/PrairieLearn/blob/b21d66f54ebf412f3bc9b3d8beb7cb851b69989b/tests/sync/assessmentsSync.js for the existing assessment syncing tests. I'd like the tests to cover at least the following:
  - Ensuring all information from the JSON file actually ends up in the DB correctly
  - Ensuring that roles/permissions are deleted from the DB when they're deleted from the JSON file
  - Ensuring that sync errors are produced when a permission corresponds to a role that does not exist.

migrations/245_group_role_tables__create.sql
renzol2: We had the idea of having 4 as a default max since the four-role POGIL system has been widespread across a few different courses at UIUC, and we thought it would be a reasonable max for this case. However, we realize that we're trying to generalize this for any course and group role system, so having a max of NULL does make more sense. Thank you for pointing that out!

renzol2: For `sync_assessments`, we thought that having the maximum number for a role be set by default to the minimum group size would be more flexible than hard-coding the max to 4, but it seems best to be as general as possible for our implementation first before worrying. We'll make changes accordingly, thanks again!

sprocs/sync_assessments.sql
renzol2: We realized that we don't need to use temporary tables, so we've updated the implementation to just insert directly into `assessment_question_role_permissions`.

sync/course-db.js
```diff
@@ -1215,6 +1227,30 @@ async function validateAssessment(assessment, questions) {
...
+    validRoleNames.add(role.name);
+  });
+
+  (assessment.zones || []).forEach((zone) => {
+    (zone.questions || []).forEach((zoneQuestion) => {
+      (zoneQuestion.canView || []).forEach((roleName) => {
+        if (!validRoleNames.has(roleName)) {
+          errors.push(
+            `A zone question's "canView" permission contains the non-existent group role name "${roleName}".`
+          );
```
renzol2, line 1241: @nwalters512 Should we specify in the sync errors which question ID failed the validation? If so, should we have separate code to handle questions with alternatives vs. without alternatives?

tests/sync/assessmentsSync.test.js
```diff
@@ -316,6 +316,339 @@ describe('Assessment syncing', () => {
...
+        .length,
+      2
+    );
+    assert.equal(
+      syncedPermissions.filter((p) => parseInt(p.group_role_id) === parseInt(foundContributor.id))
+        .length,
+      2
+    );
+
+    // Remove group roles and re-sync
```
nwalters512, line 494:
```suggestion
    // Remove the "Contributor" group role and re-sync
```

tests/sync/assessmentsSync.test.js
```diff
@@ -316,6 +316,339 @@ describe('Assessment syncing', () => {
     assert.isEmpty(uids, 'uids should be empty');
   });
 
+  it('syncs group roles correctly', async () => {
```
nwalters512, line 319: Thanks for all the thorough tests! 🙌

yarn.lock
```diff
@@ -11003,4 +11003,4 @@ zip-stream@^4.1.0:
 zwitch@^1.0.0:
   version "1.0.5"
   resolved "https://registry.yarnpkg.com/zwitch/-/zwitch-1.0.5.tgz#d11d7381ffed16b742f6af7b3f223d5cd9fe9920"
-  integrity sha512-V50KMwwzqJV0NpZIZFwfOD5/lyny3WlSzRiXgA0G7VUnRlqttta1L6UQIHzd6EuBY/cHGfwTIck7w1yH6Q5zUw==
+  integrity sha512-V50KMwwzqJV0NpZIZFwfOD5/lyny3WlSzRiXgA0G7VUnRlqttta1L6UQIHzd6EuBY/cHGfwTIck7w1yH6Q5zUw==
```
nwalters512, line 11006: Nit: can you revert this change?

nwalters512 (approved):
This is looking great! Thanks again for the thorough tests.

I've approved this, though I left a few final non-blocking comments. Feel free to address them if you'd like, and let me know when you're ready for this to be merged.

sprocs/sync_assessments.sql
```diff
@@ -361,6 +401,47 @@ BEGIN
...
+                            ) VALUES (
+                                new_assessment_question_id,
+                                valid_group_role.id,
+                                (valid_group_role.role_name IN (SELECT * FROM JSONB_ARRAY_ELEMENTS_TEXT(assessment_question->'can_view')))
+                            ) ON CONFLICT (assessment_question_id, group_role_id) 
+                            DO UPDATE
+                            SET
+                                assessment_question_id = EXCLUDED.assessment_question_id,
+                                group_role_id = EXCLUDED.group_role_id,
+                                can_view = EXCLUDED.can_view;
```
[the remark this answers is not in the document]
mwest1066: That's correct.

nwalters512:
@caleqw2 @renzol2 finals are over, which means we can merge this! To triple-check, are you all done making changes here?

renzol2:
> @caleqw2 @renzol2 finals are over, which means we can merge this! To triple-check, are you all done making changes here?

@nwalters512 Hi Nathan! Yes, we're done making changes. These changes are ready to be merged.

PR #5201, opened 2021-12-20
Add manual grading override logic

### LINKED ISSUE #5005: Confusing submission grade if question grade was manually updated
For instance questions where the grade was manually updated in the assessment log, the grading status in the question box and the submission panel both show the original submission grade which doesn't match the instance question grade. This can be confusing for students, especially in manually graded questions, since they see a 0% in the page.

image

### LINKED ISSUE #5193: Manually regrading a question updates the score, but not the UI glyph
When I regrade a question by uploading the modified `..._for_manual_grading.csv` file, the scores are updated correctly, but the UI glyphs stay the same.

## Before manual regrading

Screen Shot 2021-12-14 at 7 49 18 PM

## After manual regrading

Screen Shot 2021-12-14 at 7 49 03 PM

This is the case everywhere these scores appear, including on the assessment overview and the individual question page.

Let me know if there's anything else I can contribute to make this issue clearer!

rayhua12:
Fixes #5193 and Fixes #5005

When "highest_submission_score" is not equal to instance_questions / max_points, then it is assumed that manual grading has occurred, and so the ui should reflect the latter score, not the former.

Note: Depending on if "allow real time grading" is turned on, manually graded assignments might still show "Best Submission" rather than the more accurate "Submission Status".


rayhua12:

rayhua12:
recheck

jonatanschroeder:
You found the source of the issue, but the solution is not correct. Note that the code above won't work for exams that have reduced points in multiple attempts (e.g., 5 points for first attempt, 3 points for second attempt, etc.). In this case, the `highest_submission_score` may indeed not match the calculation you used, but in that case it should remain as is.

I believe the correct approach is not to compute a different value when printing, but instead to change the manual grading sprocs (probably `instance_questions_update_score`) to update `highest_submission_score`.

rayhua12:
> You found the source of the issue, but the solution is not correct.

Now you sound like my therapist :) 

I've moved the logic back into sprocs. However, after thinking about it some more, I want to double check that if real time grading is **True**, should it still override the visual percentage mark? I ask because usually, RTG's percentage does not match the score, for reasons as you outlined above. The current commit has some added logic to only change the displayed percent if and only if RTG is **False**.

jonatanschroeder:
Note that this sproc is only used for manual grading, so real time grading does not have an effect on this update. Real time grading only affects grades that are computed on submission, so no manual grading.

rayhua12:
What I'm more getting at is the text "Best Submission" (which appears when RTG is on) vs "Submission Status" (which appears when RTG is of). Should the % shown after "Best Submission" change after manual grading?

jonatanschroeder:
That affects the view of the data, not its update. Note that RTG only affects open assessments, and if it is off then there is no marking involved while the assessment is open, only after closing is the assessment marked. So updating the field should happen regardless of if real-time-grading is enabled, since by the time marking is done RTG no longer takes effect.

sprocs/instance_questions_update_score.sql
```diff
@@ -31,6 +31,7 @@ DECLARE
     new_score double precision;
     new_correct boolean;
     current_partial_score jsonb;
+
```
jonatanschroeder, line 34:
```suggestion
```

Nit

jonatanschroeder (approved):

PR #5203, opened 2021-12-22
Set search-formatter = false where needed

### LINKED ISSUE #5108: the name "max" matches all rows on student search page
As @jonatanschroeder  points out, the default bootstrap function matches html, and since our page uses max-width, every row is selected. 

https://bootstrap-table.com/docs/api/column-options/#searchformatter default should be set to false.

rayhua12:
Fixes #5108 . It turns out, most search tables were not affected as the "formatting" that was the issue was correlated with buttons and badges. As such, all columns with buttons and badges now have the appropriate flag. This is why the questions page is not affected, as the columns that would have been affected don't allow for free-input search.


nwalters512 (approved):
Thanks for the contribution! I tested this by searching for "pencil" on the gradebook page and "max" on the assessment instances page. That used to match every row, but now it correctly filters down to zero rows.

PR #5205, opened 2021-12-30
chore: Add .git-blame-ignore-revs file

nwalters512:
This PR adds a `.git-blame-ignore-revs` file, which can be used to configure Git to exclude certain commits when doing a `git blame`. Notably, this excludes 2f3b359987bebf00d45a8c32d58e3b302d117c4a, where I bulk-reformated all JS files with Prettier.

Note that GitHub does not yet understand this file; see https://github.community/t/support-ignore-revs-file-in-githubs-blame-view/3256 for the feature request.


nwalters512:
Going to go ahead and land this - let me know if you have any comments after the fact!

mwest1066:
Excellent idea! This has been a minor irritant several times when I've done `blame`, so it's awesome to fix it.

PR #5260, opened 2022-01-03
chore(deps): Install all minor/patch version bumps

nwalters512:
This PR tries to consolidate many of dependabot's version bumps into a single PR to avoid the need to manually merge + update, merge + update, etc. dozens of individual PRs.


sync/course-db.js
```diff
@@ -656,11 +656,12 @@ module.exports.loadInfoFile = async function ({
     }
 
     // Validate file against schema
+    /** @type {import('ajv').ValidateFunction<T>} */
     const validate = ajv.compile(schema);
     try {
       const valid = validate(json);
       if (!valid) {
-        const result = { uuid: json.uuid };
+        const result = { uuid: /** @type {any} */ (json).uuid };
```
nwalters512, line 664: TypeScript now [correctly propagates type guards that are extracted into constants](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#cfa-aliased-conditions) like we have here - all the other changes in this file stemmed from that.

PR #5262, opened 2022-01-06
Partial Credit for pl-order-blocks problems that use block groups

SethPoulsen:
needs to be merged after #5113


SethPoulsen:
I've found an edge case where this algorithm doesn't work: 

Suppose you have: 

```html
<pl-order-blocks grading-mode="dag">
  <pl-block-group>
     <pl-answer tag="1" depends="">1</pl-answer>
     <pl-answer tag="3" depends="1">3</pl-answer>
  </pl-block-group>
     <pl-answer tag="2" depends="1">1</pl-answer>
</pl-order-blocks
```
The only solution here is 1, 3, 2. But if the student submission is 2, 3. They haven't violated the block group contraints or the DAG constraints. But there is no way for the solution to be completed. 

The most likely fix to this is going to be making it so that answers inside a pl-block group can only reference things within the same pl-block group, but I need to do more thinking to make sure that actually solves the problem.

mwest1066 (approved):

SethPoulsen:
After discussing with @mwest1066 , we decided the best path forward is to merge this now, and then quickly follow it up with another PR which will change the syntax used with `pl-block-group`, making it so that the problem above is invalid, and that the algorithm in this PR is provably correct in all cases.

Issue #3856, opened 2021-02-20
Unpin code-server upstream image?


mwest1066:
We pinned the code-server workspace image at 3.4.1: https://github.com/PrairieLearn/PrairieLearn/blob/5483eb8c5698e3b3e5b7e888c1d9b4e51434e270/exampleCourse/questions/demo/workspace/vscode/info.json#L12

This was done because 3.5.0 had a "reconnect" bug: cdr/code-server#2144

See also https://github.com/PrairieLearn/PrairieLearn/blob/5483eb8c5698e3b3e5b7e888c1d9b4e51434e270/exampleCourse/questions/demo/workspace/vscode/question.html#L4

It looks like cdr/code-server#2144 has been closed so maybe we should un-pin this version again?

trombonekenny:
Bump?

jonatanschroeder:
@trombonekenny I believe this may be superseded by #9302.

reteps:
Closing -- `workspaces/vscode-base` now gets updates via dependabot and is at `FROM codercom/code-server:4.99.3-noble`

Issue #3170, opened 2020-09-22
Make better template for code-server workspaces


echuber2:
As discussed in previous issues, there are some problems with getting permissions right in images, and this can make it hard to install extensions and other things. It seems like we're going to standardize on 1001:1001 for the main user. With that in mind, here's an example code-server dockerfile that preloads an extension and fixes the permissions ahead of time. It does this in the image specifically so that fixuid doesn't have to do work at launch time to change a lot of permissions back. (I think jupyter handles this differently by just making sure the effective user is part of the group that owns the workspace files, even if the user ID is not exactly the same as expected. That may be a more elegant solution.)

```dockerfile

FROM codercom/code-server:latest

# We want to standardize on using 1001:1001 for the user.
USER root
RUN export OLD_UID=$(id -u coder) && \
	export OLD_GID=$(id -g coder) && \
	export NEW_UID=1001 && \
	export NEW_GID=1001 && \
	groupmod -g 1001 coder && \
	usermod -u 1001 -g 1001 coder && \
	find /home -user $OLD_UID -execdir chown -h $NEW_UID {} + && \
	find /home -group $OLD_GID -execdir chgrp -h $NEW_GID {} + && \
	unset OLD_UID OLD_GID NEW_UID OLD_GID

# Cleaning up the cache in the same step when we install is important
# if you want to minimize the image size.
USER coder
RUN ( code-server --disable-telemetry --force \
		# Install only one extension per line.
		--install-extension hediet.vscode-drawio \
		# --install-extension something.else \
	) && \
	rm -rf /home/coder/.local/share/code-server/CachedExtensionVSIXs

USER root
RUN	rm -f /run/fixuid.ran && \
	echo "" > /usr/bin/entrypoint.sh && chmod 0755 /usr/bin/entrypoint.sh && \
	/usr/bin/env echo -e '#!/usr/bin/env sh \n' \
		'set -eu \n' \
		'if [ "${DOCKER_USER-}" ]; then \n' \
		'  echo "$DOCKER_USER ALL=(ALL) NOPASSWD:ALL" | sudo tee -a /etc/sudoers.d/nopasswd > /dev/null \n' \
		'  sudo usermod --login "$DOCKER_USER" \ \n' \
		'    --move-home --home "/home/$DOCKER_USER" \ \n' \
		'    coder \n' \
		'  sudo groupmod -n "$DOCKER_USER" coder \n' \
[...]

PR #5263, opened 2022-01-06
Create VS Code workspace image

nicknytko:
Creates a base VS Code image that we can add stuff to, instead of using the coder image directly.  Currently just adds some python libraries.  Also resolves https://github.com/PrairieLearn/PrairieLearn/issues/3856.


workspaces/vscode/install.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Apt dependencies
+apt-get update
+apt-get upgrade --yes
+apt-get install --yes python3-pip gosu
+
+# Install Python dependencies
+pip install scipy numpy matplotlib pandas
```
nwalters512, line 9: I'd like to get thoughts from other folks, but IMO we shouldn't be opinionated about what software we install in this base image. If someone wants a workspace with a specific set of dependencies, they should create their own image with `prairielearn/workspace-vscode` as a base and install any deps they need. Otherwise, this image will just grow without bounds forever as people shoehorn more and more languages and libraries into it (or maybe that's all FUD 🤷).

workspaces/vscode/pl-gosu-helper.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+# The purpose of this script is to prevent subtle issues that might occur if
+# you are testing in the local PrairieLearn Docker container and it attempts
+# to launch workspaces as root.
+
+# Put a line like "ENV PL_USER rstudio" or "ENV PL_USER coder"
+# (whatever is appropriate) in your Dockerfile. This name should
+# match user 1001.
```
nwalters512, line 9: I'd like a little more documentation than this, specifically with regards to "whatever is appropriate" - what _is_ appropriate?

nwalters512 (commented):
Before we merge this, you'll need to update `.github/workflows/images.yml` and `.github/workflows/main.yml` to build this image.

workspaces/vscode/install.sh
mfsilva22: I think we should have one image with the basic python libraries, such as the ones above. I can imagine most people using python will want to use them. And if they want to get more libraries, yes, they can create their own image. 

I guess this is my selfish opinion. This request came because I am setting up a course for Brian (MechSE) for Finite Element Analysis, and I needed a workspace with these basic functions.

nwalters512: I guess my issue is that this isn't a Python workspace, it's a generic VSCode workspace. If we want an opinionated Python+VSCode workspace, maybe we can do that ourselves by introducing a separate `vscode-python` image that builds on top of this generic `vscode` one?

nicknytko: Sure, totally on-board with having a separate Python-based vscode image.

workspaces/vscode/pl-gosu-helper.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+# The purpose of this script is to prevent subtle issues that might occur if
+# you are testing in the local PrairieLearn Docker container and it attempts
+# to launch workspaces as root.
```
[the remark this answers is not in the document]
nicknytko: This is actually a launch script we've been using in some of the other images (see `workspaces/rstudio` and `workspaces/desktop`), courtesy of @echuber2.  Currently, workspace containers launch with root on local dev instead of the user they are supposed to be, which can lead to issues with permissions and things working correctly on local and not on prod, or vice-versa.  
I suppose it would be more appropriate to fix this in PL, rather than having these workarounds in all of the containers (but trying to work through the workspace logic is kind of scary 😢 )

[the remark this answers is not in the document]
nwalters512: > I suppose it would be more appropriate to fix this in PL, rather than having these workarounds in all of the containers (but trying to work through the workspace logic is kind of scary 😢 )

Yeah, this absolutely sounds like the right thing to do. By "the workspace logic", are you referring to everything in `workspace_host`? Agree that that's kinda hairy, but if you have a good sense of the abstract things that need to be done, it'd be awesome if you could open an issue capturing that.

workspaces/vscode/install.sh
nicknytko: Continuing the conversation, is there anything we should consider essential to a generic vscode workspace?

nwalters512: IMO, the base image should contain just the stuff we need to get VSCode in the browser. Anything language-specific should be delegated to sub images, whether maintained by us or by individual courses.

That said, that's just _my_ opinion - let me know if that's too strict or impractical 😅

workspaces/vscode/pl-gosu-helper.sh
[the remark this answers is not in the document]
echuber2: This was a pretty ad-hoc solution that worked for Dirk's rstudio situation and Elsa's OCaml (VS Code) situation. I agree it would be good if the local PL docker always tried to run workspaces as 1001 to emulate what the server does, but maybe some staff users' local OS won't allow that for whatever reason.

I think it would work to have PL try to run workspaces as 1001 by default, and if the OS won't allow it, then to issue a verbose error message to the local session terminal along with offering an override flag.

workspaces/vscode/pl-gosu-helper.sh
echuber2: It's used in `chown -R 1001:1001 "/home/$PL_USER"`, and the intent is that PL_USER should be the name of the expected user 1001 based on however the workspace image is designed.

echuber2:
I'd encourage everyone to take a look at https://github.com/PrairieLearn/PrairieLearn/issues/3170 -- I think I have a more simplified, newer draft than this that I could merge, but the important things are that 1001:1001 is effectively set on the 1000:1000 stuff in the base image, and extensions can be preinstalled. These things combined make the fixuid part at startup run _much_ faster (instantly), which greatly reduces startup latency and prevents timeouts. Without this, CS 225 had timeouts based on large extensions they were trying to install. My version also changes the working directory for the VS Code workspace to prevent system dotfiles and such being mixed into the student's view.

workspaces/vscode/pl-gosu-helper.sh
echuber2: For the workspace images I helped develop, I tried to make sure everything would be 1001:1001 beforehand to prevent slow chowning. But in the case where the local user runs as root, the specific staff user's setup might be creating or mounting stuff into the home tree with different perms, and some things can get messed up if the files in the home tree aren't owned by 1001:1001, so this was a stopgap measure. In particular, I think Dirk ran into this while testing RStudio.

echuber2: ...And I'm thinking this may not be necessary with VS Code, since its startup sequence already does the fixuid routine to accomplish something similar. (However, as noted in https://github.com/PrairieLearn/PrairieLearn/issues/3170, relying on fixuid to take care of everything can cause very slow launches, so it's better to make sure as many things as possible are baked into the image as 1001:1001.)

echuber2:
@nicknytko Is it okay for me to make commits to your PR branch? (Otherwise, is it possible for me to make a PR to your branch?) I need to do some fixups for CS 233's workspace and then it may be most efficient to help get the same fixes into the official base image.

nicknytko:
> @nicknytko Is it okay for me to make commits to your PR branch? (Otherwise, is it possible for me to make a PR to your branch?) I need to do some fixups for CS 233's workspace and then it may be most efficient to help get the same fixes into the official base image.

Yeah, that’s fine.  Commit to your heart’s content.

nwalters512:
@nicknytko are you still interested in getting this merged?

Issue #4928, opened 2021-10-06
pl-rich-text-editor: implement a "math mode"


jonatanschroeder:
Allow students to type formulas into the pl-rich-text-editor. This may help: https://lucidar.me/en/rich-content-editor/mathjax-and-quill-editor/

nwalters512:
Done in #5268!

PR #5268, opened 2022-01-10
Add math to pl-rich-text-editor

rayhua12:
Implements #4928

This is a tremendously awkward implementation, but I'm not sure how better to implement this. Quill, long story short, was not designed to work with MathJax.
Some good things with this implementation:
1. It overrides Quill's default formula implementation, meaning that the UI needs not be touched. 

Some bugs/weaknesses exist with this implementation:

1. A space character is added at the end of the mathjax text. This is required as when rendered after a page refresh, the cursor disappears if the mathjax text is the last text in the box.  
2. MathJax rendering *must* occur after the page is fully loaded. While I tried to get various async functions to work, the reality remained that none of them would work if they were initially called prior to page load completion. Hence, the timeouts. 
3. When using the "bubble" theme, it may be expected that highlighting text would change that text into LaTeX. It does not. This is intended behavior in Quill, but it is counterintuitive and un-user-friendly.

### DIFF
--- a/elements/pl-rich-text-editor/pl-rich-text-editor.js
+++ b/elements/pl-rich-text-editor/pl-rich-text-editor.js
@@ -1,5 +1,5 @@
 /* eslint-env browser,jquery */
-/* global Quill,he */
+/* global Quill,he, MathJax */
 
 window.PLRTE = function (uuid, options) {
   if (!options.modules) options.modules = {};
@@ -8,7 +8,7 @@ window.PLRTE = function (uuid, options) {
   } else {
     options.modules.toolbar = [
       ['bold', 'italic', 'underline', 'strike'],
-      ['blockquote', 'code-block', { script: 'sub' }, { script: 'super' }],
+      ['blockquote', 'code-block', { script: 'sub' }, { script: 'super' }, 'formula'],
       [{ list: 'ordered' }, { list: 'bullet' }, { indent: '-1' }, { indent: '+1' }],
       [{ size: ['small', false, 'large'] }],
       [{ header: [1, 2, 3, false] }],
@@ -34,3 +34,36 @@ window.PLRTE = function (uuid, options) {
     );
   });
 };
+
+// Override default implementation of 'formula'
+
+var Embed = Quill.imports.parchment.Embed;
+
+class MathFormula extends Embed  {
+  static create(value) {
+    const node = super.create(value);    
+    if (typeof value === 'string') {
+      let html = MathJax.tex2chtml(value);
+      let formatted = html.innerHTML
+      node.innerHTML = "&#65279;" + formatted +"&#65279;";
+      MathJax.typeset();
+      node.setAttribute('data-value', value);
+      node.contentEditable = 'false';
+    }
+    return node;
+  }
+
+  static value(domNode) {
+    return domNode.getAttribute('data-value');
+  }
+
+  html() {
+    const { formula } = this.value();
+    return `<span>${formula}</span>`;
+  }
+}
+MathFormula.blotName = 'formula';
+MathFormula.className = 'ql-formula';
+MathFormula.tagName = 'SPAN';
+
+Quill.register('formats/formula', MathFormula, true);
--- a/elements/pl-rich-text-editor/pl-rich-text-editor.js
+++ b/elements/pl-rich-text-editor/pl-rich-text-editor.js
@@ -43,14 +43,25 @@ class MathFormula extends Embed  {
   static create(value) {
     const node = super.create(value);    
     if (typeof value === 'string') {
-      let html = MathJax.tex2chtml(value);
-      let formatted = html.innerHTML
-      node.innerHTML = "&#65279;" + formatted +"&#65279;";
-      MathJax.typeset();
-      node.setAttribute('data-value', value);
-      node.contentEditable = 'false';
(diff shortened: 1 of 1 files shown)


elements/pl-rich-text-editor/pl-rich-text-editor.js
```diff
@@ -34,3 +34,45 @@ window.PLRTE = function (uuid, options) {
...
+class MathFormula extends Embed {
+  static create(value) {
+    const node = super.create(value);
+    if (typeof value === 'string') {
+      this.waitUntilLoaded(node, value);
+    }
+    return node;
+  }
+
+  static waitUntilLoaded(node, value) {
```
nwalters512, line 51: I'd recommend naming this function something other than `waitUntilLoaded` - maybe `updateNode`? That's the main this function does, the waiting is just am implementation detail IMO.

Related: please add an inline comment explaining _why_ waiting is necessary. I saw that in the PR description, but we really want that to be located in the source code so it's easy for future maintainers to discover.

elements/pl-rich-text-editor/pl-rich-text-editor.js
```diff
@@ -34,3 +34,45 @@ window.PLRTE = function (uuid, options) {
...
+    if (document.readyState !== 'complete') {
+      window.setTimeout(this.waitUntilLoaded, 200, node, value);
+    } else {
+      let html = MathJax.tex2chtml(value);
+      let formatted = html.innerHTML;
+      node.innerHTML = '&#65279;' + formatted + '&#65279;' + ' ';
```
nwalters512, line 57: Can you add a comment about the trailing space too? Also, what is `&#65279`?

elements/pl-rich-text-editor/pl-rich-text-editor.js
```diff
@@ -34,3 +34,45 @@ window.PLRTE = function (uuid, options) {
...
+      MathJax.typeset();
```
nwalters512, line 58: Why is this call to `typeset` necessary?

nwalters512 (commented):
This is awesome! All in all, the implementation looks pretty straightforward, kudos for figuring out the issues related to MathJax. This is making me yet again consider whether we should switch in part or in whole to KaTeX... but that's a question for another day 🙂 

Will approve once this has inline comments explaining the quirks.

elements/pl-rich-text-editor/pl-rich-text-editor.js
rayhua12: Typeset is used to "render" the LaTeX, since Mathjax does some weird things with the CSS to render the LaTeX. WIthout it, the LaTeX just doesn't render

rayhua12:
I've added things suggested by comments and refactored to code to work with slow connections. I discovered that the previous implementation, using timeout, would fail silently when on slow connections, for reasons I do not know. Now, using event listeners instead.

elements/pl-rich-text-editor/pl-rich-text-editor.js
```diff
@@ -34,3 +34,53 @@ window.PLRTE = function (uuid, options) {
...
+  }
+
+  static value(domNode) {
+    return domNode.getAttribute('data-value');
+  }
+
+  html() {
+    const { formula } = this.value();
+    return `<span>${formula}</span>`;
+  }
```
nwalters512, line 80: Taking another look at this, I'm not sure how this is supposed to work - `this.value()` expects a DOM node as an argument and doesn't return anything that we can destructure `formula` out of. I can't find a reference to this function in the Quill docs. Do you know how we can test this?

elements/pl-rich-text-editor/pl-rich-text-editor.js
```diff
@@ -34,3 +34,53 @@ window.PLRTE = function (uuid, options) {
...
+    if (document.readyState === 'complete') {
+      let html = MathJax.tex2chtml(value);
+      let formatted = html.innerHTML;
+      // Without trailing whitespace, cursor will not appear at end of text if LaTeX is at end
+      node.innerHTML = formatted + '&#8201;';
+      MathJax.typeset();
+      node.contentEditable = 'false';
+      node.setAttribute('data-value', value);
+    } else {
+      var n = node;
```
nwalters512, line 62: Is this necessary, or can we just pass `node` directly to `updateNode`?

elements/pl-rich-text-editor/pl-rich-text-editor.js
rayhua12: [I copied it directly out of formula.js just in case it was used](https://github.com/quilljs/quill/blob/9cf0285caa6514356a7bba9db132ec7229eb254c/formats/formula.js#L23), though you do make a good point that it doesn't seem to be used, or usable?

elements/pl-rich-text-editor/pl-rich-text-editor.js
rayhua12: Ah, that's a leftover.

elements/pl-rich-text-editor/pl-rich-text-editor.js
nwalters512: Ah, it looks like that original one is broken too. I'd rather we just drop this - AFAICT, if it _were_ used, it would break.

nwalters512 (approved):

PR #5279, opened 2022-01-18
chore: Add HTMLHint to lint HTML files

nwalters512:
I noticed in #4497 that some of our questions were improperly formatted - incorrect syntax, missing/mismatched tags, etc. I fixed that in [`a90e41d` (#4497)](https://github.com/PrairieLearn/PrairieLearn/pull/4497/commits/a90e41dfeb1ed349093f3958e229949f6771b3d2), but I'd like to automate the process of discovering invalid HTML.

This PR introduces [`HTMLHint`](https://github.com/htmlhint/HTMLHint) to our tooling - it'll check any `question.html` in `exampleCourse` or `testCourse` for invalid HTML.

~This PR will not pass CI yet - I still have to actually fix all the existing broken questions!~


exampleCourse/questions/element/richTextEditor/question.html
```diff
@@ -15,7 +15,7 @@
 
 <pl-hide-in-panel answer="true">
 
-  <hr style="margin-top: 3rem;"></hr>
+  <hr style="margin-top: 3rem;">
```
nicknytko, line 18: I know you didn't write this, but while we're focusing on code style do we want to change this over to `class="mt-*"` instead of an inline style?

nicknytko (approved):
lgtm; thanks for cleaning all this up!

exampleCourse/questions/element/richTextEditor/question.html
nwalters512: Done in e92f8f0 - I ended up removing the custom spacing entirely.

PR #5080, opened 2021-11-08
Combine example and demo courses

mfsilva22:
This PR  merges the demo course "XC 100" with the example course "XC 101". 

The idea is to create different assessments with appropriate rules that can be copied to other courses (for example, a template for HW, exam, group work, etc). These assessments appear in the assessment page when someone sign-in as a student (not an instructors with a course space). 

Other assessments collect questions by type, similar to how XC 101 used to be. They do not have access rule, and hence do not show in the assessment page in student mode. 

Assessment page in Staff View:
image

Assessment page in Student View:
image

To be done (after receiving some feedback):

- [ ] check links from documentation, to make sure they are not broken (I don't think they are, since I did not change question paths. List of broken links:

- https://prairielearn.readthedocs.io/en/latest/assessment/#adding-text-and-links-to-assessments

- [ ] update prairielearn.com with the appropriate links, to stop directing to xc 100, and start directing to xc 101.
- [ ] create more documentation explaining the use of each assessment.

Issue #5025, opened 2021-10-24
exampleCourse has incorrect GitHub links on questions settings pages


echuber2:
It looks like whatever formula is used to generate the GitHub links on question settings pages is not working correctly for the exampleCourse that sits in the main PL repo.

Example:
https://www.prairielearn.org/pl/course/108/question/4942648/settings
The "view on GitHub" link leads to here:
https://github.com/PrairieLearn/PrairieLearn/tree/master/questions/demo/drawing/graphs
but it should lead to here:
https://github.com/PrairieLearn/PrairieLearn/tree/master/exampleCourse/questions/demo/drawing/graphs

nwalters512:
Good catch - it looks like the bit of logic that generates that code doesn't take into account the fact that the example course is a subdirectory of a repo, whereas we assume every other course is at the repo root. We special case the example course in many other places throughout PL; we'll have to do the same here too.

echuber2:
Perhaps this is a difficult suggestion if it's already being special-cased in many places, but would it be better if each course had a setting for root directory? That would at least generalize the problem.

trombonekenny:
Is this resolved? Perhaps by #5288?

nwalters512:
Yep, I think we can close this!

PR #5288, opened 2022-01-21
Use correct URLs for example course questions

nwalters512:
#5080 did not correctly update the URLS that were embedded in various questions and assessments to point to `PrairieLearn/exampleCourse` instead of `pl-demo-course`. This PR fixes that.


echuber2:
That reminds me of this issue: #5025

PR #5291, opened 2022-01-25
Reorganize example course assessments

nwalters512:
This PR reorganizes the assessments for the example course by removing the numbered prefixes and trying to group the assessments inside of directories. This is mostly successful, except for the questions assessments:

Screen Shot 2022-01-27 at 2 41 30 PM

@mfsilva22 I'd like your input here - IMO it'd be nice to standardize on either "demo" or "gallery", and possibly remove the "templates" assessment entirely, as the template assessment actually mixes `template`, `demo`, and `elements` directories.

As part of this, I wrote a small script (`tools/validate-links.mjs`) to validate that links to our GitHub repository actually refer to files/directories that exist. This caught a lot of unrelated broken links, which I decided to fix in this PR anyways.


tools/validate-links.mjs
```diff
@@ -0,0 +1,45 @@
+import fs from 'fs-extra';
```
nwalters512, line 1: Here's the script that validates links. We're now running on a new-enough version of Node that we can use `.mjs` files with native ES modules!

mfsilva22:
I think the changes look good. If we are not worried about broken links, I think we should be okay with this change. 

To be consistent with the idea of grouping assessments inside folders, I think AID `workshop` should be replaced with `workshop/summer2022`, and `lecture` with `lecture/self-guided` (I am planning to create another lecture example with videos, but first need to find time to create a couple of videos).

To give a little of background with the "Collection of example questions zone": this zone is not available when someone is checking the course example from pl.org, but only from local development, so we assume the person is already hooked with PL (hence, this zone is not used for marketing purposes). The idea was to use this zone to support the documentation, and be used as starting point for instructors when creating questions. The proposed organization had the following logic:

- "C4" has one question for each element as it appears in the documentation. It gives a more detailed explanation of the element, and its different features. They are not ideal to be used as template.
- "C6" has the questions that should be used as templates. We tried to organized it with zones for better clarity.

I think the "template" assessment is actually the most useful to new instructors. How you think we should make these questions available to instructors, if we remove the assessment? I am not attached at all to this organization, it was just an idea at the time.

nwalters512:
> To be consistent with the idea of grouping assessments inside folders, I think AID `workshop` should be replaced with `workshop/summer2022`, and `lecture` with `lecture/self-guided` (I am planning to create another lecture example with videos, but first need to find time to create a couple of videos).

Good call - I'll make that change.

Re: organization of the assessments that show questions, thanks for explaining the current rationale. I agree with the existence of those two assessments; I guess my issue lies with the demo vs gallery vs template distinction. This is not a huge issue though, and thanks for pointing out that this won't be visible to folks coming to PL (although maybe it should be, to show off all we have to offer!). I'll avoid any future changes to the question assessments for now.

mfsilva22 (approved):
Looks all good! We can later talk more about making the "Collection" section available, but I think it will be distracting. I think something polished and with purpose (smaller) is better for marketing, and if they are curious for more, they will get to see everything once they create their course space.

PR #5294, opened 2022-01-26
fix: Include page view events to group assessment instance log

### LINKED ISSUE #5286: "View Variant" event miss from groupwork assessment log
Issue: "View Variant" event tag is missed from groupwork assessment log. However, It functions normal in all other types of assessments. 

- Reproduce the error:
1. Open a groupwork assessment in a course
2. Click on Question 1
3. (optional) Click on assessment overview and then come back to Question 1 again
4. Check log file (method to check log file is described below), there is no log activity tagged with "View Variant"
5. Open any other type of assessment (hw, exams, ...), repeat same behavior, there should be at least one log activity tagged with "view variant"

Method to view logs:
  In instructor / TA view "Go to your course > your course instance > assessment name > Students tab. The links in the first column on that sheet take you to their results and log unified in one page. Scroll down to the bottom to see the detailed log" (quote from Eric Huber's instruction).

  It is also missed from log files in API, which can be downloaded using the end point `/pl/api/v1/course_instances/:course_instance_id/assessment_instances/:assessment_instance_id/log`

jonatanschroeder:
Fixes #5286.

The log used to limit page view log entries to those associated to the assessment's user themselves, presumably to avoid page view events triggered by the instructor. I considered using a similar filter to limit to group members if there was a group, but this would stop events from showing if a user were to be removed from a group. I'd prefer to keep the side-effect of including instructor views in the log, but if that is not desirable I guess the appropriate way to handle it would be either switching group users to soft-delete or having some differentiation in the log entry itself, I'd appreciate some suggestions in that case.


jonatanschroeder:
Found the source of where this was originally filtered: https://github.com/PrairieLearn/PrairieLearn/pull/1165/files#r194761337

So this all depends if we want this to apply to the instance log: https://github.com/PrairieLearn/PrairieLearn/blob/90049a27c302c011358b78a1636c3fd11d470124/middlewares/logPageView.js#L22-L26

jonatanschroeder:
One possible workaround here is to use `group_logs` to restrict views to users that were part of the assessment group at the time of the event...

jonatanschroeder:
The filter based on `group_logs` is implemented. The issue with it, though, is that `group_logs` does not have an index that can be used for this query, which could make this query slower. That said, I believe this is an issue with the table, not with the implementation, since other uses of `group_logs` also have this issue. My suggestion here is to consider an index in `group_logs` by `group_id` at least.

sprocs/assessment_instances_select_log.sql
```diff
@@ -340,7 +340,13 @@ BEGIN
                 WHERE
                     pvl.assessment_instance_id = ai_id
                     AND pvl.page_type = 'studentInstanceQuestion'
-                    AND pvl.authn_user_id = ai.user_id
+                    AND (pvl.authn_user_id = ai.user_id
+                         OR EXISTS (SELECT 1
+                                    FROM group_logs gl
```
nwalters512, line 345: Nit for consistency:
```suggestion
                                    FROM group_logs AS gl
```

sprocs/assessment_instances_select_log.sql
```diff
@@ -366,7 +372,13 @@ BEGIN
                 WHERE
                     pvl.assessment_instance_id = ai_id
                     AND pvl.page_type = 'studentAssessmentInstance'
-                    AND pvl.authn_user_id = ai.user_id
+                    AND (pvl.authn_user_id = ai.user_id
+                         OR EXISTS (SELECT 1
+                                    FROM group_logs gl
```
nwalters512, line 377: Nit for consistency:
```suggestion
                                    FROM group_logs AS gl
```

sprocs/assessment_instances_select_log.sql
```diff
@@ -340,7 +340,13 @@ BEGIN
                 WHERE
                     pvl.assessment_instance_id = ai_id
                     AND pvl.page_type = 'studentInstanceQuestion'
-                    AND pvl.authn_user_id = ai.user_id
+                    AND (pvl.authn_user_id = ai.user_id
+                         OR EXISTS (SELECT 1
```
nwalters512, line 344: It's entirely possible that the query planner/optimizer will do the right thing here by default, but I wonder if this should instead be `OR (ai.group_id IS NOT NULL AND EXISTS (SELECT 1 ... )`, at least until we can get an index on the `group_id` column? An `EXPLAIN ...` could probably help us decide one way or the other with more confidence.

nwalters512 (commented):
> That said, I believe this is an issue with the table, not with the implementation, since other uses of `group_logs` also have this issue. My suggestion here is to consider an index in `group_logs` by `group_id` at least.

I think this is a good idea. While checking it out for myself, I realized that the `group_logs` table is completely missing any sort of referential integrity (no foreign keys 😢).

jonatanschroeder:
> I think this is a good idea. While checking it out for myself, I realized that the `group_logs` table is completely missing any sort of referential integrity (no foreign keys 😢).

I think that's on purpose for log tables, you don't want them to be affected if a row is deleted.

sprocs/assessment_instances_select_log.sql
jonatanschroeder: That's a good point. I'm not sure the query planner would do the right thing for non-group assessments with views by instructors in this case. I'll add the qualifier.

nwalters512:
> I think that's on purpose for log tables, you don't want them to be affected if a row is deleted.

Yeah, Matt pointed this out to me. It's interesting because _some_ of our log tables (like `page_view_logs`) do actually maintain foreign key relationships. I guess we're erring on the side of "retain all data" though.

sprocs/assessment_instances_select_log.sql
```diff
@@ -340,7 +340,14 @@ BEGIN
...
                     AND pvl.page_type = 'studentInstanceQuestion'
-                    AND pvl.authn_user_id = ai.user_id
+                    AND (pvl.authn_user_id = ai.user_id
+                         OR (ai.group_id IS NOT NULL
+                             AND EXISTS (SELECT 1
+                                         FROM group_logs AS gl
+                                         WHERE gl.action = 'join'
+                                               AND gl.group_id = ai.group_id
+                                               AND gl.user_id = pvl.authn_user_id
+                                               AND gl.date <= pvl.date)))
```
nwalters512, line 350: Can you explain why this is necessary?

nwalters512 (approved):
@jonatanschroeder would you mind filing an issue to track the desire for an index on `group_logs.group_id`?

sprocs/assessment_instances_select_log.sql
jonatanschroeder: If you mean the date, it is not strictly necessary, it would probably only make a difference for someone that has access to the assessment but is not part of the group (e.g., a staff member), and in that case only if that user then becomes part of the group, which presumably means this group is for staff testing only (which probably would be a bit fuzzy to determine if their view should be included). The main reason I kept it was in the spirit of "was a member of the group at the time of the view action", so "time" came into account. I'd be perfectly fine if you feel strongly about removing it. Is that what you were asking about?

jonatanschroeder:
> @jonatanschroeder would you mind filing an issue to track the desire for an index on `group_logs.group_id`?

#5306

sprocs/assessment_instances_select_log.sql
nwalters512: Yep, I was referring to the date. I figured it was something like that, but I had the same "hmmm..." thinking about this only really making sense if someone viewed it without being part of the group and then later joined the group. I don't feel particularly strongly about this (my PR approval stands), but I'm inclined to remove it because a) it seems to be here to support an unlikely edge case and b) I don't think it handles the case where someone (realistically, a staff member) joins a group, then leaves it, then views again. Rather than going down the edge case rabbit hole, I'm pretty comfortable saying "if user X was part of a group at any point, include their views".

All that said, good call on using `group_logs` instead of just `group_users` to ensure the case of someone leaving a group is handled correctly!

jonatanschroeder: Done. I also added a comment to clarify the filter.

nwalters512 (approved):

PR #5295, opened 2022-01-26
fix: Exclude deleted questions when fetching question templates

### LINKED ISSUE #5293: v2 questions report "path" error
(reported by @glherman)

Since we deployed 90049a27c, some v2 questions in production have been reporting `The "path" argument must be of type string. Received null`.

To reproduce, go to CS 233 Spring 2022, change to student view, select assessment W2.03PRE, and click on the question "Full Adder Specification".

The stack trace is:
```
Async Stacktrace:
    at /PrairieLearn/pages/studentInstanceQuestionHomework/studentInstanceQuestionHomework.js:175:9
    at /PrairieLearn/lib/question.js:1970:13
    at /PrairieLearn/lib/question.js:1829:21
    at /PrairieLearn/lib/question.js:426:17
    at /PrairieLearn/lib/question.js:417:21
    at /PrairieLearn/lib/question.js:340:17
    at /PrairieLearn/lib/question.js:252:13
    at /PrairieLearn/lib/question.js:117:13
    at /PrairieLearn/question-servers/calculation.js:102:11
    at Object.<anonymous> (/PrairieLearn/question-servers/calculation.js:34:13)

TypeError [ERR_INVALID_ARG_TYPE]: The \"path\" argument must be of type string. Received null
    at new NodeError (internal/errors.js:322:7)
    at validateString (internal/validators.js:124:11)
    at Object.join (path.js:1148:7)
    at Object.module.exports.coursePathForChunk (/PrairieLearn/lib/chunks.js:210:19)
    at /PrairieLearn/lib/chunks.js:975:38
    at Array.map (<anonymous>)
    at Object.module.exports.ensureChunksForCourseAsync (/PrairieLearn/lib/chunks.js:971:19)
```

It looks like the error is being thrown at: https://github.com/PrairieLearn/PrairieLearn/blob/90049a27c302c011358b78a1636c3fd11d470124/lib/chunks.js#L210 which is called from: https://github.com/PrairieLearn/PrairieLearn/blob/90049a27c302c011358b78a1636c3fd11d470124/lib/chunks.js#L975

The [NodeJS docs for `path.join()`](https://nodejs.org/api/path.html#pathjoinpaths) indicate that exactly this error message will be reported if any of the arguments are not strings, so this is consistent. This means either `coursePath` or `chunkMetadata.questionName` is `null` for v2 questions, but I can't see why this would be the case.

I was unable to reproduce this error in staging -- the question referenced above worked fine in student view in staging.

nwalters512:
Fixes #5293 - see that ticket for more details.

### DIFF
--- a/lib/chunks.sql
+++ b/lib/chunks.sql
@@ -60,6 +60,7 @@ WITH RECURSIVE template_questions AS (
     FROM
         template_questions AS q
         JOIN questions AS tq ON (tq.qid = q.template_directory AND tq.course_id = q.course_id)
+    WHERE tq.deleted_at IS NULL
 )
 SELECT id FROM template_questions LIMIT 100; -- LIMIT prevents infinite recursion on circular templates
 


mwest1066:
Ah!! So that’s why it only affected v2 questions. Good sleuthing!

mwest1066 (approved):

PR #5297, opened 2022-01-27
fix: Exclude deleted questions when fetching question templates

nwalters512:
Followup to #5295 - I missed half of the recursive query!


lib/chunks.sql
```diff
@@ -52,7 +52,9 @@ WITH RECURSIVE template_questions AS (
     FROM
         questions AS q
         JOIN questions AS tq ON (tq.qid = q.template_directory AND tq.course_id = q.course_id)
-    WHERE q.id = $question_id
+    WHERE
+        q.id = $question_id
+        AND q.deleted_at IS NULL
```
nwalters512, line 57: Wait, damn, I think *this* is it 😭 
```suggestion
        AND tq.deleted_at IS NULL
```

mwest1066 (approved):

Issue #4650, opened 2021-08-09
pl-order-blocks question failing automatic tests


jeffgerickson:
I have created a pl-order-blocks question at https://www.prairielearn.org/pl/course_instance/128710/instructor/question/9063329

When I select "Test once with full details” on the question settings page, I get two "Question test failure" errors, complaining about a score mismatch (_expected "1" but got "0.4"_).  Here are the messages in detail:
https://www.prairielearn.org/pl/course_instance/128710/instructor/question/9063329/preview?variant_id=38920997

Also for some previous variants, even _opening_ the issue page generates three new issues:
https://www.prairielearn.org/pl/course_instance/128710/instructor/question/9063329/preview?variant_id=38870280

SethPoulsen:
Ok, so there are a couple of things causing this problem. Both are simplifications we made in trying to ship `pl-order-blocks`, and have been on my list of things to fix, but I haven't gotten to them yet.

1. The testing code assumes that the ordering of the blocks in the HTML is a correct solution, without verifying that they are
2. The testing code assumes that removing the first block in the correct answer will result in a completely incorrect answer. This is obviously a bad assumption for some question/grading mode combinations

SethPoulsen:
> 1. The testing code assumes that the ordering of the blocks in the HTML is a correct solution, without verifying that they are

Will be solved by #5871

> 2. The testing code assumes that removing the first block in the correct answer will result in a completely incorrect (0 credit) answer.

This was already fixed between #5261 and #5113.

So I'll go ahead and close this once #5871 is merged!

Feel free to re-open after that if there still ends up being issues.

PR #5302, opened 2022-01-28
pl-order-blocks: fix incorrect internal conversion from ranking to dag grading

### LINKED ISSUE #5300: pl-order-blocks incorrectly grading ranking problem
```
<pl-order-blocks answers-name="order-frames" grading-method="ranking">
  <pl-answer correct="true" ranking="2">ARP Request</pl-answer>
  <pl-answer correct="true" ranking="3">ARP Reply</pl-answer>
  <pl-answer correct="true" ranking="1">IP Frame 1</pl-answer>
  <pl-answer correct="true" ranking="4">IP Frame 2</pl-answer>
</pl-order-blocks>

```
image

Reported by Fraida Fund on PL Slack

SethPoulsen:
Fix #5300 

The new code to convert the ranking to the DAG grading wrongly assumed that the answer blocks would be in some correct order in the HTML.

NOTE that users should still write their problems with the answer in the correct order in the HTML, because the order given in the HTML is the order that will be shown to students as the 'correct' solution. This is a documented known problem that I just haven't gotten around to fixing yet (see e.g. #4650).

### DIFF
--- a/elements/pl-order-blocks/pl-order-blocks.py
+++ b/elements/pl-order-blocks/pl-order-blocks.py
@@ -417,15 +417,17 @@ def grade(element_html, data):
 
         if grading_mode == 'ranking':
             true_answer_list = sorted(true_answer_list, key=lambda x: int(x['ranking']))
-            true_answer = [answer['ranking'] for answer in true_answer_list]
-            lines_of_rank = {rank: [str(i) for i, x in enumerate(true_answer) if x == rank] for rank in set(true_answer)}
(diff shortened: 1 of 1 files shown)


nwalters512 (approved):
I don't fully understand what's going on here, but going to rubber-stamp on the assumption that you've tested this 🙂 

Aside: it would be fantastic to get some unit testing set up for elements. Since they're just composed of a number of functions, it should be pretty straightforward to set up an HTML blob, call some element functions, and assert things about the state of `data` or the resulting HTML.

PR #5307, opened 2022-01-29
Refactor API tests

nwalters512:
This PR refactors the API tests to use `node-fetch` instead of the deprecated `request` library. It also tries to make less use of `locals` in favor of more explicit variables, and makes test ordering more explicit via `mocha-steps`.

I've seen that contributors (unsurprisingly) tend to look towards existing tests for examples of how to write their own, and I want to make sure our tests exemplify our own best practices. This is a first step in that direction.


lib/workspace.js
```diff
@@ -185,7 +185,7 @@ module.exports = {
     if (res.ok) return;
 
     // if there was an error, we should have an error message from the host
-    const json = res.json();
+    const json = await res.json();
```
nwalters512, line 188: Thanks TypeScript!

PR #5352, opened 2022-02-01
Allow tests to run in parallel

nwalters512:
This PR allows us to execute our Mocha tests in parallel. We see only moderate speedup in CI (where we're only running with 2 cores), but it really shines locally on machines with more cores - the whole test suite now runs in just 80 seconds when running natively on my laptop!


nwalters512:
I can reproduce the CI timeout failure locally with the following:

```
DEBUG=prairielearn:cron ./node_modules/.bin/mocha --full-trace --require tests/mocha-hooks.mjs tests/cron.test.js tests/exampleCourseQuestions.test.js
```

I then see logs like the following:

```
  prairielearn:cron stop() +12s
  prairielearn:cron stop(): requesting stop for 3 +0ms
  prairielearn:cron stop(): clearing timeout for daily +1ms
  prairielearn:cron stop(): waiting for 1 jobs to stop +0ms
  prairielearn:cron { '3': -1 } +0ms
  prairielearn:cron stop(): waiting for 1 jobs to stop +103ms
  prairielearn:cron { '3': -1 } +0ms
  prairielearn:cron stop(): waiting for 1 jobs to stop +101ms
  prairielearn:cron { '3': -1 } +0ms
  prairielearn:cron stop(): waiting for 1 jobs to stop +101ms
  prairielearn:cron { '3': -1 } +0ms
  prairielearn:cron stop(): waiting for 1 jobs to stop +101ms
  prairielearn:cron { '3': -1 } +0ms
```

And so on, forever and ever.

This doesn't appear when I just run `make test-prairielearn` locally - presumably this has something to do with test ordering, or which tests get run together in specific subprocesses.

nwalters512:
Woof, this was a bug in how we handled marking abandoned jobs as errored - see f32af7c.

lib/server-jobs.js
```diff
@@ -393,6 +393,8 @@ module.exports.errorAbandonedJobs = function (callback) {
             }
             callback(null);
           });
+        } else {
+          callback(null);
```
nwalters512, line 397: This was causing test failures, and looks to just be a very longstanding bug!

server.js
```diff
@@ -1598,6 +1598,25 @@ module.exports.startServer = async () => {
     throw new Error('unknown serverType: ' + config.serverType);
   }
 
+  // Wait for the server to either start successfully or error out.
+  await new Promise((resolve, reject) => {
```
nwalters512, line 1602: It's up to us whether we want to actually leave this in. This was useful for identifying the root cause of a test failure where multiple servers were trying to listed on the same port.

tests/chunkAssessment.test.js
```diff
@@ -1,6 +1,7 @@
 const util = require('util');
 const assert = require('chai').assert;
 const { step } = require('mocha-steps');
+const tmp = require('tmp-promise');
```
nwalters512, line 4: I'm not 100% sure why this was necessary, but without this, the tests failed because they couldn't create the path `/chunks`. This may have been unique to me because I develop outside of Docker. At any rate, the new setup mirrors what happens in `tests/chunks`.

tests/courseEditor.test.js
```diff
@@ -5,21 +5,25 @@ const fs = require('fs-extra');
...
 
 const locals = {};
 let page, elemList;
 
-const baseDir = path.join(__dirname, 'testFileEditor');
-const courseTemplateDir = path.join(baseDir, 'courseTemplate');
+const courseTemplateDir = path.join(__dirname, 'testFileEditor', 'courseTemplate');
+
+// Set up temporary writeable directories for course content
+const baseDir = tmp.dirSync().name;
```
nwalters512, line 26: Previously, this and `tests/fileEditor` expected to have exclusive use of the directories in `testFileEditor`, which is an invalid assumption when running tests in parallel. We now avoid this problem by giving each test its own temporary directory.

I think there are some `.gitignore` rules that we can now remove.

tests/groupScoreAndSync.test.js
```diff
@@ -27,6 +27,16 @@ const questions = _.keyBy(question, 'qid');
...
 
+  const storedConfig = {};
+  before('store authenticated user', () => {
+    storedConfig.authUid = config.authUid;
+    storedConfig.authName = config.authName;
+    storedConfig.authUin = config.authUin;
+  });
+  after('unset authenticated user', () => {
+    Object.assign(config, storedConfig);
+  });
```
nwalters512, line 38: This whole test had a side effect of mutating `storedCofig.auth*`. We now properly restore the original values like the other tests.

tests/honorCode.test.js
```diff
@@ -78,6 +78,7 @@ describe('Exam assessment response to `requireHonorCode`', function () {
     assert.equal(response.$('#start-assessment').text(), 'Start assessment');
 
     // We should not see the honor code div anymore
-    assert.lengthOf(response.$('div.test-suite-honor-code'), 0);
+    // TODO: fix this test. Exam 2 in `testCourse` no longer disables the honor code.
+    // assert.lengthOf(response.$('div.test-class-honor-code'), 0);
```
nwalters512, line 82: Calling this out.

tests/honorCode.test.js
```diff
@@ -41,7 +41,7 @@ describe('Exam assessment response to `requireHonorCode`', function () {
   context.baseUrl = `${context.siteUrl}/pl`;
   context.courseInstanceBaseUrl = `${context.baseUrl}/course_instance/1`;
 
-  before('set up testing server', helperServer.before);
+  before('set up testing server', helperServer.before());
```
nwalters512, line 44: This whole test suite wasn't being executed. Now that it's included, I had to fix tests that had been broken since the suite was added.

tests/sync/courseDb.test.js
```diff
@@ -187,27 +187,6 @@ describe('course database', () => {
         assert.isOk(result.data);
       });
     });
-
-    it('warns about a default assessment set being present', async () => {
```
nwalters512, line 191: This test suite wasn't being executed before; I had to fix some tests that were broken without us knowing.

tests/helperDb.js
```diff
@@ -114,46 +115,58 @@ async function databaseExists(dbName) {
...
-    await createFullDatabase(postgresqlDatabaseTemplate, true, mochaThis);
-    await createFromTemplate(postgresqlDatabase, postgresqlDatabaseTemplate, true, mochaThis);
+    await module.exports.createTemplate(mochaThis);
+    await createFromTemplate(dbName, POSTGRES_DATABASE_TEMPLATE, true, mochaThis);
   }
-  await establishSql(postgresqlDatabase);
+  await establishSql(dbName);
 }
 
-module.exports = {
```
SethPoulsen, line 127: Why this change? a style guide thing or something?

.github/workflows/main.yml
```diff
@@ -57,14 +57,3 @@ jobs:
...
-      run: docker cp test_container:/PrairieLearn/coverage coverage
-    - name: Directory listing for debug
-      run: ls -lR
-    - name: Head of lcov.info for debug
-      run: head -1000 coverage/lcov.info
-    - name: Send code-coverage data to Coveralls
-      uses: coverallsapp/github-action@master
-      with:
-        github-token: ${{ secrets.github_token }}
-      continue-on-error: true
```
SethPoulsen, line 70: Why are we getting rid of this?

SethPoulsen (commented):
Awesome, I'm always excited to have faster builds.

tests/helperDb.js
nwalters512: personal preference slash better symmetry with ESM `import`/`export` syntax that we'll hopefully move to soonish!

.github/workflows/main.yml
nwalters512: `nyc` mangled stack traces, making it really difficult to debug tests. I've personally almost never looked at code coverage when looking at PRs, so I opted to drop it rather than trying to figure out how to coerce `nyc` into presenting correct stack traces. If you feel strongly about keeping it, I can look into that.

nwalters512:
@mwest1066 The build is finally green here - this should finally be ready for review!

.github/workflows/main.yml
SethPoulsen: nope, just curious

tests/honorCode.test.js
nwalters512: Fixed in 31a70b3.

.github/workflows/main.yml
SethPoulsen: One thought actually, it might be interesting just to glance at coverage once just to make a list of all the things that need more tests, e.g. a list of which elements, pages, API endpoints, etc. don't have any tests, or don't have nearly enough.

The we can throw that list in with all the other lists of things to do :joy: .

nwalters512: You can check out the current coverage on https://coveralls.io/github/PrairieLearn/PrairieLearn!

lib/server-jobs.js
mwest1066: Oh wow, good find!

server.js
mwest1066: Looks fine to me to keep.

tests/chunkAssessment.test.js
mwest1066: The new setup certainly seems cleaner in any case.

tests/courseEditor.test.js
mwest1066: This is the best change ever! A horrible side-effect of the way things used to be was that killing a test half-way through would leave the temporary directories in place, which would then stop the tests from running next time. I'm so happy to see this fixed!

tests/groupScoreAndSync.test.js
mwest1066: Nice cleanup!

tests/honorCode.test.js
mwest1066: Huh!!!

tsconfig.json
```diff
@@ -4,7 +4,8 @@
     "esModuleInterop": true,
     "noEmit": true,
     // Allow spreading Maps/Sets into arrays
-    "downlevelIteration": true
+    "downlevelIteration": true,
+    "allowSyntheticDefaultImports": true
```
mwest1066, line 8: Not that I really mind either way, but if `tsconfig.json` allows trailing commas then should we be using them?

Makefile
```diff
@@ -27,13 +27,11 @@ start-s3rver:
 test: test-js test-python
 test-js: test-prairielearn test-prairielib test-grader-host test-packages
 test-prairielearn: start-support
-	@nyc --reporter=lcov mocha --full-trace tests/index.js
+	@mocha --full-trace --parallel --require tests/mocha-hooks.mjs "tests/**/*.test.{js,mjs}"
```
mwest1066, line 30: Maybe we should add a target to run the tests serially, and document this in the dev docs? I can imagine that there will be some cases where it will be much easier to debug a failing test if we can run it serially.

Also, do we need any updates to the docs about how to run individual tests? See: https://github.com/PrairieLearn/PrairieLearn/blob/2add41190efddfe4eae142634cef6ef57de862a3/docs/installingLocal.md?plain=1#L49

mwest1066 (approved):
I tried it out and it seems to work for me, although I was only giving it 2 cores so there wasn't much obvious speedup.

nwalters512:
There's a regression in the testing experience that was introduced by the combination of this PR and https://github.com/PrairieLearn/PrairieLearn/pull/5396. See https://github.com/open-telemetry/opentelemetry-js-contrib/issues/970 for a description. I'll probably hold off on merging this until that's fixed 🙏

tsconfig.json
nwalters512: Done in c3b624b.

Makefile
mwest1066: Nice!

nwalters512:
I've submitted a fix for the above. Hopefully it gets merged and released soon! https://github.com/open-telemetry/opentelemetry-js-contrib/pull/972

In the meantime, it looks like tests are failing for some reason 👀 I'll investigate.

nwalters512:
I ended up removing the code that disables instrumentations in #5640, so this PR is safe to proceed!

PR #5353, opened 2022-02-01
chore: Cache Docker layers in GitHub Actions

nwalters512:


nwalters512:
Unfortunately this seems to be just about as bad as before - any time gained by caching layers is lost again by the need to actually restore the cache and then upload the new cache at the end.

Issue #3451, opened 2020-11-02
Consider a percent-based credit


jonatanschroeder:
The current credit system is based on a maximum value of a question, rather than a percentage. In some of our courses the policies are based on a percentage instead (e.g., credit of 80% means that whatever grade is obtained during that time is multiplied by 0.8 instead of being capped at 80). It would be useful then to allow some mechanism for this kind of credit to apply.

mwest1066:
See #818 and #1921. I completely agree with you that `credit` should apply to work done during that time. It would be awesome to fix this!

PR #5355, opened 2022-02-02
Introduce percent based calculation

DawtDawt:
This PR adds the optional `percentageCreditGrading` boolean flag to switch credit calculations (<100) from the default `min(credit, points / max_points * 100)` to percent based calculation mentioned in #3451.

The implementation defaults to the previous method of calculation.

### DIFF
--- a/api/v1/endpoints/queries.sql
+++ b/api/v1/endpoints/queries.sql
@@ -100,7 +100,8 @@ WITH object_data AS (
         aar.show_closed_assessment_score,
         format_date_iso8601(aar.start_date, ci.display_timezone) AS start_date,
         aar.time_limit_min,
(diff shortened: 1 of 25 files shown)


DawtDawt:

jonatanschroeder:
I personally think that this would probably be clearer if, instead of reusing the `credit` option, used a separate value for the percentage (something like `percentageGrade: 90` or `gradePenaltyPercentage: 10`) that is totally independent of credit. Trying to reuse the same value could lead to confusion.

DawtDawt:
Thanks for the feedback! I will work on implementing the seperate value soon.

(Nice to see you too Jonatan! Hope you're doing well in York!)

PR #5362, opened 2022-02-05
fix: Update formatters in question page to account for change in bootstrap-table behavior

jonatanschroeder:
The new version of bootstrap-table (1.19.1) seems to have an issue with filter control related to custom searches as reported by David on Slack. The easiest solution here is to restore it to version 1.18.3.

A preliminary check of the code seems to be that filter control in 1.19.1 assumes fields to be strings, which can affect how it formats the data for fields that are not strings, but this is just a very quick assessment, I may be wrong. I'm just submitting this PR to ensure that, if a deployment is done this weekend, it doesn't go with this bug, and it gives us time to review how to properly address this issue (1.19.1 still has a couple of desirable bug fixes in other areas, so I'm hoping to check this further at a future time).

### DIFF
--- a/lib/question.js
+++ b/lib/question.js
@@ -1796,7 +1796,8 @@ module.exports = {
       [
         (callback) => {
           if (variant_id != null) {
-            sqldb.callOneRow('variants_select', [variant_id], (err, result) => {
+            const params = [variant_id, locals.question.id, locals.instance_question?.id];
+            sqldb.callOneRow('variants_select', params, (err, result) => {
               if (ERR(err, callback)) return;
               debug('variants_select', result.rows[0].variant);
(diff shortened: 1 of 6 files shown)


jonatanschroeder:
Found a solution that seems to work with version 1.19.1. The formatter has issues with fields that correspond to objects/arrays, so changing the formatting to use the fields themselves from the row instead of the field parameter value seems to fix it.

nwalters512:
Issue reported to `bootstrap-table`: https://github.com/wenzhixin/bootstrap-table/issues/6044

`bootstrap-table` PR that may fix this: https://github.com/wenzhixin/bootstrap-table/pull/6045

PR #5367, opened 2022-02-08
feat: Password-protect assessment instance creation

### LINKED ISSUE #5365: Change password entry to protect assessments (not just assessment instances)
At the moment an exam with a password only asks for the password once the assessment instance has been created, and thus after the exam timer has started. This is a problem if the proctor takes a long time to enter the password.

To fix this, we should also trigger the password check if we have an assessment with no assessment instance. This will capture the start-exam scenario, while leaving post-exam behavior (when we have an assessment instance) unchanged. I think we just need to add a condition like `|| (res.locals.authz_result?.password && res.locals.assessment && ! res.locals.assessment_instance)` here: https://github.com/PrairieLearn/PrairieLearn/blob/8715c094ffac16eec1c75d28f1812f443a45908e/middlewares/studentAssessmentAccess.js#L55

nwalters512:
This PR ensures, for exams that require a password, that the student has already entered it before they are allowed to create a new assessment instance. This avoids a problem with prompting for the password *after* the assessment instance has been created: for time-limited exams, the timer would start counting down as soon as the instance is created, but before the user had entered the password.

Regarding the implementation, I don't love the fact that we now have two pieces of code that need to check for the password, but I don't see a great way around that - the logic of when to actually prompt for a password when hitting a `studentAssessment{Exam,Homework}` page can't be really easily expressed up at the middleware level. I tried to compromise by reusing code and leaving plenty of comments 🤷 

Closes #5365.


pages/studentAssessmentExam/studentAssessmentExam.js
```diff
@@ -2,6 +2,7 @@ var ERR = require('async-stacktrace');
 var express = require('express');
 var router = express.Router();
 
+const { checkPasswordOrRedirect } = require('../../middlewares/studentAssessmentAccess');
```
mwest1066, line 5: I’m not in love with requiring middleware from page routes, but I don’t strongly object either. We could shift this function to `lib/`, but I’m not convinced it would be neater. I wanted to at least raise the question to see whether you had any other ideas!

mwest1066 (approved):

mwest1066:
> the logic of when to actually prompt for a password when hitting a `studentAssessment{Exam,Homework}` page can't be really easily expressed up at the middleware level. I tried to compromise by reusing code and leaving plenty of comments 🤷

Just for my own curiosity, can you say something a bit more about why it wasn’t enough to do the type of simple check that I was thinking of in #5365?

nwalters512:
> Just for my own curiosity, can you say something a bit more about why it wasn’t enough to do the type of simple check that I was thinking of in #5365?

Yes! The main issue is that the `studentAssessmentAccess` middleware would run on the `studentAssessment{Exam,Homework}` pages before we've even had a chance to check if an assessment instance exists, so `res.locals.assessment_instance` will *always* be null for those pages. This means that landing on `studentAssessment{Exam,Homework}` for an assessment that already has a *closed* instance, we'll incorrectly prompt the user for the password. At least, I think this is how it would all work - does that sound correct?

The alternative would be to move the "does assessment instance exist for this assessment" logic into middleware that runs before `studentAssessmentAcccess`.

mwest1066:
> Yes! The main issue is that the `studentAssessmentAccess` middleware would run on the `studentAssessment{Exam,Homework}` pages before we've even had a chance to check if an assessment instance exists, so `res.locals.assessment_instance` will _always_ be null for those pages. This means that landing on `studentAssessment{Exam,Homework}` for an assessment that already has a _closed_ instance, we'll incorrectly prompt the user for the password. At least, I think this is how it would all work - does that sound correct?

Excellent point! We don't load the AI until we are in the page handler, which is after the middleware has run.

> The alternative would be to move the "does assessment instance exist for this assessment" logic into middleware that runs before `studentAssessmentAcccess`.

Yeah, I'm not sure this is cleaner in any way.

pages/studentAssessmentExam/studentAssessmentExam.js
nwalters512: Yeah, I don't love this either, but if we move this to `lib/`, I think we'd have to move much of `studentAssessmentAccess` too (along with the `.ejs` file), which basically feels like we're still importing middleware, just without `/middleware/` in the path 😅 I'm inclined to leave this as-is.

mwest1066: Agreed. Let's stick with things as you have them.

PR #5370, opened 2022-02-08
pl-order-blocks: Allow partial credit when using ranking grading

### LINKED ISSUE #5368: pl-order-blocks ranking problems don't accept partial credit
I believe pl-order-blocks should support partial credit in `ranking` mode now (#5261), and the docs say it should be:

> For the dag and ranking grading methods, you may specify none for no partial credit or lcs (default) for partial credit based on the LCS edit-distance from the student solution to some correct solution.

but it doesn't: 

https://github.com/PrairieLearn/PrairieLearn/blob/5d9562c2cd3fbf77dd0901206aa8faeaca77801f/elements/pl-order-blocks/pl-order-blocks.py#L63

image

SethPoulsen:
Fix #5368

I should probably have test cases with every possible combination of configuration options


nwalters512 (approved):
+1 for element unit tests!

PR #5123, opened 2021-11-29
feat: Add ee directory and corresponding license

nwalters512:
This PR lays the groundwork for developing "enterprise" features (e.g. billing, multi-institution SSO, advanced autoscaling) without having to split the code between private and public repositories. Specifically, it adds an `ee` directory with a separate license that's based on [Gitlab's enterprise license](https://gitlab.com/gitlab-org/gitlab/-/blob/master/ee/LICENSE).

It also moves the `workspaceHostLoads` code into the `ee` directory as an initial proof of concept for splitting code between the EE and CE versions.

Closes #5117.

PR #5371, opened 2022-02-08
chore: Add infrastructure to support multiple packages

nwalters512:
This PR introduces infrastructure to support publishing multiple packages from the monorepo, and also adds a basic `@prairielearn/html` package that we can use to test the workflow. It utilizes the following tools:

- [Turborepo](https://turborepo.org/) is used as the task runner (will be more useful once we have more packages)
- [changesets/changesets](https://github.com/changesets/changesets) is used to automate the versioning and publishing of packages

The primary motivation of this change is to allow easier code sharing with PrairieTest (closed-source). Currently, PrairieTest has PrairieLearn as a submodule, which works ok, but Yarn isn't aware that PrairieLearn contains a `package.json` and thus isn't able to actually ensure that the correct versions of PL dependencies are installed. By breaking things out into modules that are actually published to npm, we're able to treat them as standard packages that can have their dependencies resolved.

However, breaking things out into subpackages also allows us to start incrementally adopting TypeScript in target ways instead of having to go all-in with the whole repo. And with [Turborepo](https://turborepo.org/), breaking things up into separate packages also means we can cache results and avoid duplicate work when a particular package's code hasn't changed!

## TODO

- [x] Publishing automation


Makefile
```diff
@@ -1,5 +1,11 @@
 export PATH := node_modules/.bin/:$(PATH)
 
+build:
+	@turbo run build
```
mwest1066, line 4: Works for me!

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Setup Node.js 14.x
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14.x
+
+      - name: Install Dependencies
+        run: yarn --frozen-lockfile
+
+      - name: Create Release Pull Request or Publish to npm
```
mwest1066, line 27: I didn't test this, but it looks good and is awesome

mwest1066 (approved):

packages/html/package.json
```diff
@@ -0,0 +1,19 @@
+{
```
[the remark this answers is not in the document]
nwalters512: I like the idea! Let's land https://github.com/PrairieLearn/PrairieLearn/pull/5123 first, since that lays a small amount of groundwork for using multiple licenses throughout different parts of the codebase. We can come back to this in a small followup PR.

.github/workflows/release.yml
nwalters512: I have also not tested this 😂 I'll see what happens when we merge to master and open followup PRs to fix things if necessary.

packages/html/package.json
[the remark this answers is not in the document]
mwest1066: Sounds good!

.github/workflows/release.yml
mwest1066: 😂

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
...
+#
+# See the `if` condition on the `release` job for the second half of this logic.
+on:
+  workflow_run:
+    workflows:
+      - CI
+    branches:
+      - master
+    types:
+      - completed
```
nwalters512, line 15: FYI, I changed this. I think this new version is what we want!

mwest1066: Makes sense to me!

PR #5375, opened 2022-02-11
fix: Java autograder repeat test if multiple test classes

jonatanschroeder:
As reported by @NormanCHutchinson, the Java autograder was repeating tests any time more than one JUnit test class was found in the junit folder. So if there were classes A, B and C, the execution of A would run the tests of A, the execution of B would run the tests of A and B, and the execution of C would run the tests of A, B and C. Caused by the reuse of a request builder in the JUnit launcher API.

### DIFF
--- a/graders/java/JUnitAutograder.java
+++ b/graders/java/JUnitAutograder.java
@@ -60,7 +60,7 @@ public class JUnitAutograder implements TestExecutionListener {
     public void runTests() throws UngradableException {
 
         for (String classSrcName : testClasses) {
-	    LauncherDiscoveryRequestBuilder requestBuilder = LauncherDiscoveryRequestBuilder.request();
+            LauncherDiscoveryRequestBuilder requestBuilder = LauncherDiscoveryRequestBuilder.request();
             String className = classSrcName
                 .replaceFirst("^/grade/tests/junit/", "")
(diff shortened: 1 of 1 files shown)


graders/java/JUnitAutograder.java
```diff
@@ -59,9 +59,8 @@ public static void main(String args[]) {
 
     public void runTests() throws UngradableException {
 
-        LauncherDiscoveryRequestBuilder requestBuilder = LauncherDiscoveryRequestBuilder.request();
-
         for (String classSrcName : testClasses) {
+	    LauncherDiscoveryRequestBuilder requestBuilder = LauncherDiscoveryRequestBuilder.request();
```
nwalters512, line 63: Let's avoid mixing tabs and spaces.
```suggestion
        for (String classSrcName : testClasses) {
            LauncherDiscoveryRequestBuilder requestBuilder = LauncherDiscoveryRequestBuilder.request();
```

nwalters512 (approved):

graders/java/JUnitAutograder.java
jonatanschroeder: Ah, the caveats of using multiple editors... fixed.

PR #5406, opened 2022-02-14
Use Google auth name if it exists

jonatanschroeder:
From https://developers.google.com/identity/protocols/oauth2/openid-connect:

> The user's full name, in a displayable form. Might be provided when:
> * The request scope included the string "profile"
> * The ID token is returned from a token refresh
>
> When name claims are present, you can use them to update your app's user records. Note that this claim is never guaranteed to be present.

The Google auth login currently uses `profile` as a scope, so the name might be provided. This PR uses the `name` if it exists, and keeps using `email` if it doesn't. This should allow the name to be used if it was provided.


mwest1066 (approved):

pages/authCallbackOAuth2/authCallbackOAuth2.js
```diff
@@ -49,8 +49,8 @@ router.get('/', function (req, res, next) {
     }
     const params = [
       identity.email, // uid
-      identity.email, // name
-      null, // uin
+      identity.name || identity.email, // name (use email if name is not present)
+      identity.sub, // uin
```
mwest1066, line 53: Do you have any idea what `sub` looks like? For example, is it a 256-character monstrosity that we shouldn't try and render on pages?

jonatanschroeder: The docs say at most 255 characters, but the example has ~30 digits. I imagine it's closer to the latter, but I don't have a simple way to test.

jonatanschroeder: May not be a bad idea to render the UIN with a max-width and ellipsis overlay anyway, though (in a different PR).

mwest1066: Agreed.

jonatanschroeder:
I don't want to merge this without merging the PT one (assuming they're deployed together), otherwise one will override the other...

nwalters512:
I tested this in our staging environment and this looks good! As you point out, we'll want to make sure this is deployed together with PT, so I'll hold off merging this until we're ready to deploy both.

PR #2747, opened 2020-07-18
Automatically add exampleCourse viewer permissions on attempted access

tdy:
* [x] Remove restrictions from XC 101 -- all users will be added as `Viewer`
* [x] Move the sql from the home page to preview page -- users might be linked directly to the question (e.g., workshop docs)

---

* [x] Move to `authzCourse` (a0c0c15)
* [x] Split `home.sql` into separate queries (e288094)

---

* [x] Lift `pl_courses.options->isExampleCourse` to `pl_courses.is_example_course` (a2369d0, 740f63b)

cc @mfsilva22

PR #5407, opened 2022-02-14
feat: Show prompt for student data permissions on staff pages

### LINKED ISSUE #5284: Clarification on lack of student data view for users with course permissions
A very common question on the PL slack relates to course staff that fail to give themselves student data permission to the course instances. This is an understandably confusing case, since these users often have full control of the course from term to term, and often don't need access to student data for some time.

My proposal: keep the gradebook page (and other pages that require student data access) accessible to users that have course Viewer permission or above, but if the staff member doesn't have student data access, directs them to request this permission from the course owner (maybe list the owners?). If the staff member is an owner, provide a link to the staff page itself.

nwalters512:
It's confusing a lot of course staff that they're unable to see the gradebook page. This is because they don't have student data view permissions, but this is in no way obvious from the UI. This PR ensures that anyone with access to a course can see and visit the gradebook tab, but if they don't have student data permissions, it prompts them to request permissions from a course owner (or, if they are the owner, gives them a link to the "Staff" page).

Fixes #5284.


sprocs/authz_course.sql
```diff
@@ -34,10 +34,10 @@ BEGIN
 
     IF NOT FOUND THEN
         course_role := 'None';
-    END IF;
 
-    IF is_example_course AND allow_example_course_override THEN
-        course_role := 'Viewer';
+        IF is_example_course AND allow_example_course_override THEN
+            course_role := 'Viewer';
+        END IF;
```
nwalters512, line 40: The previous arrangement made it so that it was difficult to test these changes with the `dev@illinois.edu` user - even if I made that user a course owner, this condition would always trigger and revert me to a viewer. I don't think this change should have any negative impact, but I wanted to call it out for review since it's auth code and that's really important to get right.

mwest1066: I checked on prod and we have a lot of users who are `Editor` for XC 101. I'm not sure why this happened, but I'm worried that this change will allow all of them to actually act as editors and not be force-reverted to viewers.

We could delete all course-level permissions for XC 101 in prod, but we should probably check why we have so many Staff listed there. Are we automatically adding users to course staff in XC 101? Or did we used to do this in the past but now we just automatically give people staff access without adding them? I should remember the state of this, but I don't 😭

pages/instructorGradebook/instructorGradebook.js
```diff
@@ -21,9 +21,11 @@ var csvFilename = function (locals) {
...
 
 router.get('/', function (req, res, next) {
-  if (!res.locals.authz_data.has_course_instance_permission_view) {
-    return next(error.make(403, 'Access denied (must be a student data viewer)'));
-  }
+  // We don't actually forbid access to this page if the user is not a student
+  // data viewer, because we want to allow users to click the gradebook tab and
+  // see instructions for how to get student data viewer permissions. Otherwise,
+  // users just wouldn't see the tab at all, and this caused a lot of questions
+  // about why staff couldn't see the gradebook tab.
```
mwest1066, line 28: This is excellent! But we've seen the same problem happen for pages other than the Gradebook. I remember that people have been confused about why they don't see Downloads and the Students page, in particular. Should we take this opportunity to change all pages to be always visible in navigation (but with the message that this page needs more perms)?

Another case of this that has caused lots of support traffic is the Staff page, where non-Owners can't see it and get confused by the docs that say to go to the Staff page. The same solution would solve this issue.

pages/instructorGradebook/instructorGradebook.ejs
```diff
@@ -165,62 +165,78 @@
...
+          <div class="card-body">
+            <h2>Missing student data access</h2>
+            <p>You must have permission to view student data in order to access the gradebook.</p>
+            <% if (authz_data.has_course_permission_own) { %>
+              <p>
+                You can grant yourself access to student data on the course's
+                <a href="<%= urlPrefix %>/course_admin/staff">Staff tab</a>.
+              </p>
+            <% } else { %>
+              <p>You can request access to the student data by contacting a course owner.</p>
```
mwest1066, line 178: Would it be reasonable to explicitly list the course owners? We have the problem every semester where someone has inherited a course and comes to us to ask for more permissions because they don't know who the owner is that they need to ask for perms.

mwest1066 (approved):

mwest1066 (changes requested):
Sorry, I meant to block this pending resolution of the existing Editors in XC 101 on prod.

sprocs/authz_course.sql
nwalters512: Ah, thank you so much for actually checking!

As far as I can tell, we are not automatically adding users as editors (or anything else) for the example course, though I'm hesitant to say that with certainty. I'd appreciate a second set of eyes on this.

I'll leave this change around for now as I continue to iterate on this PR, but I'm happy to back it out if we can't gain confidence that this would be safe to do.

pages/instructorGradebook/instructorGradebook.ejs
nwalters512: I think that's totally reasonable, yeah. I'll make that change.

pages/instructorGradebook/instructorGradebook.js
nwalters512: Ahhhh, great call. Yes to all of that! I'll just do that as part of this PR too.

pages/instructorGradebook/instructorGradebook.ejs
mwest1066: If it's painful to obtain the list of owners on every page, we could just unconditionally direct people to the (now globally visible) Staff page, and just show the list of owners there.

sprocs/authz_course.sql
tdy: > we are not automatically adding users as editors (or anything else) for the example course

I'm not sure about `Editor`, but we auto-add XC 101 `Viewer` upon attempted access (#2747). I believe the idea was to make it easier to demo questions to non-PL users.

> we have a lot of users who are `Editor` for XC 101. I'm not sure why this happened

Was this due to the PL Workshop a while back?

mwest1066: To be clear, I agree with this change. I just want to make sure that it's safe to deploy 😄

I just deleted all XC 101 course staff on prod. Let's wait and see whether any of them re-appear before we merge this PR. If they stay deleted then I think we are safe to deploy this (we'll also need to delete XC 101 staff in other regions).

pages/instructorGradebook/instructorGradebook.ejs
```diff
@@ -165,62 +165,78 @@
...
+            <% } %>
+            <p></p>
```
jonatanschroeder, line 180: I'm not really a fan of empty `p` paragraphs. If the intention is to increase the bottom spacing, maybe use `pb-2` in the enclosing `div`, or `mb-2` in the previous paragraph?

pages/instructorGradebook/instructorGradebook.js
jonatanschroeder: Would it make sense here to skip the SQL query in this case, since the data is not needed?

pages/instructorGradebook/instructorGradebook.ejs
nwalters512: Neither am I, this was just a typo! I'll remove it, thanks for pointing it out.

pages/instructorGradebook/instructorGradebook.ejs
```diff
@@ -154,32 +154,38 @@
...
             <% } else { %>
-              <p>You can request access to the student data by contacting a course owner.</p>
+              <p>Contact one of the below course owners to request access.</p>
+              <ul>
+              <% course_owners.forEach(function(owner) { %>
+                <li>
+                  <%= owner.uid %> <% if (owner.name) { %> (<%= owner.name %>) <% } %>
+                </li>
+              <% }); %>
+              </ul>
```
jonatanschroeder, line 179: This could be a partials to avoid duplication...

mwest1066 (approved):

pages/instructorGradebook/instructorGradebook.ejs
nwalters512: Removed in e43c7d1.

PR #5410, opened 2022-02-16
chore: Add logs to help diagnose freezing chunk servers

nwalters512:
In an effort to gain more information about the intermittent "stuck in transaction" problem that affects chunk servers, this PR adds logs that will dump out some information if transactions don't complete within 60 seconds. Hopefully this will allow us to identify a common cause during failures (a specific question, or course, a particularly large submission, etc.).

### DIFF
--- a/lib/question.js
+++ b/lib/question.js
@@ -605,7 +605,7 @@ module.exports = {
       logger.error('saveSubmission transaction was not closed within 60 seconds', {
         variant_id: variant.id,
         question_id: question.id,
-        course_id: course,
+        course_id: course.id,
         submission,
       });
     }, 60 * 1000);
@@ -863,7 +863,7 @@ module.exports = {
       logger.error('gradeVariant transaction was not closed within 60 seconds', {
         variant_id: variant.id,
         question_id: question.id,
-        course_id: course,
+        course_id: course.id,
       });
     }, 60 * 1000);
 
@@ -934,7 +934,7 @@ module.exports = {
       logger.error('saveAndGradeSubmission transaction was not closed within 60 seconds', {
         variant_id: variant.id,
         question_id: question.id,
-        course_id: course,
+        course_id: course.id,
       });
     }, 60 * 1000);
 
@@ -1345,7 +1345,7 @@ module.exports = {
       logger.error('_testQuestion transaction was not closed within 60 seconds', {
         variant_id: variant.id,
         question_id: question.id,
-        course_id: course,
+        course_id: course.id,
       });
     }, 60 * 1000);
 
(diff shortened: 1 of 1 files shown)


lib/question.js
```diff
@@ -583,6 +598,18 @@ module.exports = {
...
    */
   saveSubmission(submission, variant, question, course, callback) {
+    // This is (hopefully) temporary code to debug an issue where chunk servers
+    // get stuck inside of a transaction. We can possibly remove it once we
+    // identify and fix the issue.
+    const timeoutId = setTimeout(() => {
+      logger.error('saveSubmission transaction was not closed within 60 seconds', {
+        variant_id: variant.id,
+        question_id: question.id,
+        course_id: course,
```
mwest1066, line 608: Not that it really matters, but I think `course` is an object here? (and similarly in later changes)
```suggestion
        course_id: course.id,
```

lib/question.js
```diff
@@ -583,6 +598,18 @@ module.exports = {
    * @param {function} callback - A callback(err, submission_id) function.
    */
   saveSubmission(submission, variant, question, course, callback) {
+    // This is (hopefully) temporary code to debug an issue where chunk servers
+    // get stuck inside of a transaction. We can possibly remove it once we
+    // identify and fix the issue.
+    const timeoutId = setTimeout(() => {
+      logger.error('saveSubmission transaction was not closed within 60 seconds', {
```
mwest1066, line 605: Might be worth including some information about the submission, in case it's bad submissions causing this? Could even just be something like `JSON.stringify(submission).substring(0, 1000)` so that we truncate it if it's too big?

lib/question.js
```diff
@@ -886,6 +926,18 @@ module.exports = {
    */
   saveAndGradeSubmission(submission, variant, question, course, overrideGradeRateCheck, callback) {
     debug('saveAndGradeSubmission()');
+
+    // This is (hopefully) temporary code to debug an issue where chunk servers
+    // get stuck inside of a transaction. We can possibly remove it once we
+    // identify and fix the issue.
+    const timeoutId = setTimeout(() => {
+      logger.error('saveAndGradeSubmission transaction was not closed within 60 seconds', {
```
mwest1066, line 934: Also add something about the submission here?

mwest1066 (approved):

lib/question.js
nwalters512: Good catch! Done in 3d0021d.

lib/question.js
nwalters512: I like your thinking, I made this change in 4de3e5e.

mwest1066 (approved):

PR #5412, opened 2022-02-17
fix: Only consider gradable submissions in question stats

jonatanschroeder:
As reported on Slack, the stats were giving weird results if the first submission was non-gradable (score set to null). In that case the first submission has a grade of null, but max and average did not count that submission. For consistency this PR updates it so only gradable submissions are included in stats.

### DIFF
--- a/sprocs/instance_questions_calculate_stats.sql
+++ b/sprocs/instance_questions_calculate_stats.sql
@@ -15,6 +15,7 @@ WITH first_calculation AS (
         variants AS v
         JOIN submissions AS s ON (s.variant_id = v.id)
     WHERE v.instance_question_id = instance_question_id_param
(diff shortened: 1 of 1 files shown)


sprocs/instance_questions_calculate_stats.sql
```diff
@@ -15,6 +15,7 @@ WITH first_calculation AS (
         variants AS v
         JOIN submissions AS s ON (s.variant_id = v.id)
     WHERE v.instance_question_id = instance_question_id_param
+          AND s.gradable IS TRUE
```
nwalters512, line 18: Nit for consistent formatting:
```suggestion
         AND s.gradable IS TRUE
```

nwalters512 (approved):
This seems reasonable!

PR #5413, opened 2022-02-18
Add functionality to tag questions for manual grading

jonatanschroeder:
This PR in essence replaces the functionality of selecting questions for manual grading from "questions where the last submission  hasn't been graded" to "questions tagged for manual grading". While the original functionality remains the default in this PR, this allows questions to be tagged for grading even if (1) they were graded already or (2) they were graded using internal/external grading.

Also in this PR are the database changes to assign a question to individual graders. The functionality for this may come in a future PR, but some related queries are already in place.

* [x] database changes;
* [x] implementation of basic functionality;
* [x] major revamping of assessment question page for manual grading, including functionality to:
    * [x] cleaner view of assessment questions;
    * [x] ability to filter;
    * [x] ability to select and tag questions for manual grading.
* [ ] CI.


pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.js
```diff
@@ -24,30 +20,52 @@ router.get('/', function (req, res, next) {
...
+    const action_data = JSON.parse(req.body.batch_action_data) || {};
+    const instance_question_ids = Array.isArray(req.body.instance_question_id)
+      ? req.body.instance_question_id
+      : [req.body.instance_question_id];
+    const params = {
+      assessment_id: res.locals.assessment.id,
+      assessment_question_id: res.locals.assessment_question_id,
+      instance_question_ids,
+      requires_manual_grading: !!action_data?.requires_manual_grading,
+      assigned_grader: action_data?.assigned_grader,
```
jonatanschroeder, line 60: @nwalters512 @mwest1066 do we need any sort of validation here to ensure the assigned grader is a staff member in the course instance? I'm thinking that a malicious user with TA access in one course could potentially use this to get information (name, etc.) from a user_id, but maybe I'm overthinking.

nwalters512: We should definitely validate. I'm not too concerned about malicious users (I could already use the "Staff" page to add someone to my course and see their name), but just as a matter of principle, we don't want to allow someone to assign to a user outside of their course instance.

jonatanschroeder:
@mwest1066 @nwalters512 I think this PR is at a stage where it could benefit from feedback from staff in courses with manual grading enabled. It is by no means final, but I believe it's stable enough to ship.

migrations/248_instance_questions__requires_manual_grading__add.sql
```diff
@@ -0,0 +1,28 @@
...
+FROM
+    assessment_questions AS aq
+    JOIN questions AS q ON (q.id = aq.question_id)
+    JOIN ranked_submissions AS rs ON (TRUE)
+WHERE
+    aq.id = iq.assessment_question_id
+    AND q.grading_method = 'Manual'
+    AND rs.instance_question_id = iq.id
+    AND rs.submission_row = 1
+    AND rs.graded_at IS NULL;
```
jonatanschroeder, line 28: If this PR were to be deployed in production with migrations being done separately from code changes, this update should probably be done just before code changes are implemented to avoid discrepancies. Not sure what the best approach should be in that case.

pages/instructorAssessmentManualGrading/instructorAssessmentManualGrading.ejs
```diff
@@ -23,13 +33,13 @@
...
                 <th>Points</th>
-                <th>Number of Submissions</th>
+                <th>Submissions to grade</th>
                 <th></th>
               </tr>
             </thead>
             <tbody>
               <% questions.forEach(function(question, i) { %>
-              <tr>
+              <tr <% if (question.num_instance_questions_assigned > 0) { %>class="bg-light"<% } %> >
```
jonatanschroeder, line 42: This is a very subtle highlighting of questions assigned to the current user. I'm happy to discuss something stronger or removing it altogether.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.js
```diff
@@ -89,58 +72,39 @@ router.get('/', (req, res, next) => {
...
-            if (ERR(err, next)) return;
-            res.redirect(
-              `${res.locals.urlPrefix}/assessment/${req.body.assessment_id}/assessment_question/${req.body.assessment_question_id}/next_ungraded?instance_question=${res.locals.instance_question.id}`
-            );
-          });
-        });
-      }
-    );
+        res.redirect(
+          `${res.locals.urlPrefix}/assessment/${req.body.assessment_id}/assessment_question/${req.body.assessment_question_id}/next_ungraded?prior_instance_question=${res.locals.instance_question.id}`
```
nwalters512, line 104: Seems like we could possibly redirect them straight to the next ungraded one instead of requiring a second redirect from the `next_ungraded` route? Not necessary to change now, but may be worth it to add a `TODO:` if you think that's possible.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.ejs
```diff
@@ -21,6 +21,11 @@
       <%- include('../partials/questionSyncErrorsAndWarnings'); %>
     </div>
     <div id="content" class="container">
+      <% if (assessment_instance.open) { %>
```
nwalters512, line 24: Unrelated to this change, but I might consider renaming this page's files to `instructorInstanceQuestionManualGrading` to differentiate between this and `instructorAssessmentQuestionManualGrading`.

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.js
```diff
@@ -3,45 +3,38 @@ const express = require('express');
...
-    // If we have no more submissions, then redirect back to manual grading page
-    if (!result.rows[0]) {
-      logger.info(
-        'ManualGradingNextInstanceQuestion: No more submissions, back to manual grading page.',
-        params
-      );
+      // If we have no more submissions, then redirect back to manual grading page
+      if (!result.rows[0]) {
+        res.redirect(
+          `${res.locals.urlPrefix}/assessment/${res.locals.assessment.id}/assessment_question/${res.locals.assessment_question_id}/manual_grading?done`
```
nwalters512, line 27: Is the `done` query param used?

sprocs/submissions_insert.sql
```diff
@@ -38,12 +39,16 @@ BEGIN
     IF NOT FOUND THEN RAISE EXCEPTION 'invalid variant_id = %', variant_id; END IF;
     
     -- we must have a variant, but we might not have an assessment_instance
-    SELECT              iq.id,                  ai.id
-    INTO instance_question_id, assessment_instance_id
+    SELECT iq.id,                ai.id,
+           iq.requires_manual_grading OR q.grading_method = 'Manual'
+    INTO   instance_question_id, assessment_instance_id,
+           new_requires_manual_grading
```
nwalters512, line 45: Nit, as long as we're modifying this code:
```suggestion
    SELECT
        iq.id,
        ai.id,
        iq.requires_manual_grading OR q.grading_method = 'Manual'
    INTO
        instance_question_id,
        assessment_instance_id,
        new_requires_manual_grading
```

nwalters512 (commented):
On the whole this is looking great! I took a first pass and left some comments.

pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.ejs
```diff
@@ -2,18 +2,272 @@
...
+          $.post($(e.target).attr('action'), $(e.target).serialize(), function(data) {
+              $('#grading-table').bootstrapTable('refresh');
+          }, 'json').fail(function(data) {
+              // TODO Better user notification of update failure
+          });
+      }
+
+      function gradingTagDropdown() {
+          const course_staff = JSON.parse('<%- JSON.stringify(course_staff) %>') || [];
+          const dropdown = $('<div class="dropdown ml-1">');
```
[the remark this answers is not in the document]
jonatanschroeder: I wanted to ensure the button and the dropdown are in the same div separate from other buttons to avoid issues later on if another button with a dropdown is included (e.g., to select columns). I figured out that it's possible to add a btn-group inside another btn-group to take care of this, should be ok now.

pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.ejs
```diff
@@ -2,18 +2,272 @@
...
+      function updateGradingTagButton() {
+          $('.grading-tag-button').prop('disabled', !$('#grading-table').bootstrapTable('getSelections').length);
+      }
+
+      //<% include('../partials/pointsFormatter'); %>
+      window.getStringFromFloat = <%- getStringFromFloat.toString() %>;
+
+      function pointsFormatter(points, row) {
+          let text = `${getStringFromFloat(row.points)} <small>/<span class="text-muted"><%= max_points %></span></small> `, button = '';
+          //<% if (authz_data.has_course_instance_permission_edit) { %>
```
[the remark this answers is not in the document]
jonatanschroeder: the comment there is mostly to avoid having the text editor trying to interpret it as weird javascript. With the comment, emacs just treats it as a comment and doesn't try to mess up the rest of formatting.

pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.ejs
```diff
@@ -2,18 +2,272 @@
...
+          }));
+          item_no_grader.text(` No grader assigned`).prepend('<i class="fas fa-tag"></i>');
+
+          menu.append('<div class="dropdown-divider">');
+
+          const item_graded = $('<button class="dropdown-item" type="submit">').appendTo(menu);
+          item_graded.attr('data-batch-action', JSON.stringify({
+              requires_manual_grading: false,
+          }));
+          item_graded.text(` Tag as graded`).prepend('<i class="fas fa-check-square"></i>');
```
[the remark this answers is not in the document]
jonatanschroeder: There is: tag as requiring grading with no grader assigned. Do you mean a tag that keeps the current assigned grader?

pages/instructorQuestionManualGrading/instructorQuestionManualGradingNextInstanceQuestion.js
jonatanschroeder: Leftover from the previous version. I don't think it did anything before, I'm not opposed to removing it.

pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.ejs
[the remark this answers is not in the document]
nwalters512: Ahh, that makes sense now that you point it out. Maybe my problem is that there weren't any staff listed adjacent to "No grader assigned", so I interpreted that as a current status instead of an action?

Thinking out loud here, but maybe we could separate "update grading needed" and "assign graders" to make this clearer? I'm thinking something like the following (copy/icons all subject to change, of course):

Screen Shot 2022-03-16 at 09 17 25

pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.ejs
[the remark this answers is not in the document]
jonatanschroeder: Maybe something like this?

https://github.com/PrairieLearn/PrairieLearn/blob/73803d56ebbf5f8f8816863d836fd3653aa0bca7/pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.ejs#L171

pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.ejs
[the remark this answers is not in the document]
jonatanschroeder: Typically if you are assigning grading to someone you are also tagging it for grading (doesn't make sense to assign a grader without also tagging it for grader, from an instructor's perspective). While we can still maintain the actions above and just reshuffle the buttons, that could be confusing for instructors.

One option is to add a header to the second section, similar to the first section, to help clarify that they are different options?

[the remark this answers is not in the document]
nwalters512: Ah, yes - we could still have the "Assign grader" actions _also_ tag as ungraded?

I think the confusing thing for me is that there are two pieces of state ("is graded" and "who is assigned") that are kind of independent and kind of not, and we have one dropdown where different actions control different subsets of that state 😅 maybe I'm overcomplicating this...

pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.ejs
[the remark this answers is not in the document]
nwalters512: That's definitely better! In the long term, I still want zero EJS interpolation in JavaScript, but that can certainly wait.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.js
jonatanschroeder: I moved the computation of next ungraded to a lib called by both the next_ungraded route and this redirect. I also async'ed some stuff in the process.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.ejs
jonatanschroeder: I'm wondering if it may be a better idea to use ManualGrading as a prefix to help organize these pages? So:
* `instructorManualGradingAssessment`
* `instructorManualGradingAssessmentQuestion`
* `instructorManualGradingInstanceQuestion`
* `instructorManualGradingNextUngradedQuestion`

Thoughts?

pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.ejs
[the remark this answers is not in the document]
jonatanschroeder: Moved it to a script-level variable for client-side check.

pages/instructorAssessmentQuestionManualGrading/instructorAssessmentQuestionManualGrading.ejs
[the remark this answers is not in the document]
jonatanschroeder: Updated to match behaviour above.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.ejs
SethPoulsen: Either of these sound better than the current naming scheme. I was so confused all the time when I was working on this.

Do we want the file names to correspond somehow with the URLs too? I know @mwest1066 mentioned not being particularly happy with the URLs at some point, so I don't know if that's something we want to revisit.

SethPoulsen:
My main concern about the design of this portion of the codebase (not specific to this PR, but the manual grading stuff from the start) is that there are 3-4 different SQL queries which return _almost_ they same thing. 

This code duplication is a real pain for maintenance, and makes it really easy to introduce bugs (inconsistencies between the queries is why manual grading was completely broken in production on launch initially!). 

It seems like it would be really cool if these could be refactored in to a single sproc to avoid the duplication, and make testing easier. 

Either way, it would be really nice to have tests on the queries to make sure they are actually returning the expecting values under various sets of circumstances with different numbers of submissions at different times to different question instances, etc.

I haven't really had a chance to read the rest of this PR. I may in a few weeks if it hasn't already been merged by then.

pages/instructorQuestionManualGrading/instructorQuestionManualGrading.ejs
mwest1066: @jonatanschroeder I think that it makes sense to group all the manual grading pages together as you suggest. We could also consider using a deeper hierarchy and having `pages/instructorManualGrading/assessment/assessment.js` and similar. I don't have a strong opinion about this.

jonatanschroeder: > @jonatanschroeder I think that it makes sense to group all the manual grading pages together as you suggest. We could also consider using a deeper hierarchy and having `pages/instructorManualGrading/assessment/assessment.js` and similar. I don't have a strong opinion about this.

@mwest1066 Sounds good. Thoughts about changing the URLs as @SethPoulsen mentioned?

mwest1066: Do you have a plan for how to change them? So long as we maintain the necessary structure for authz, I'm pretty flexible.

migrations/248_instance_questions__requires_manual_grading__add.sql
jonatanschroeder: @mwest1066 @nwalters512 any thoughts on this?

sprocs/instance_questions_update_score.sql
```diff
@@ -163,7 +163,9 @@ BEGIN
             score_perc_in_grading = 0,
             status = 'complete',
             modified_at = now(),
-            highest_submission_score = new_score
+            highest_submission_score = new_score,
+            requires_manual_grading = FALSE,
+            last_grader = arg_authn_user_id
```
nwalters512, line 168: Since this sproc is also used for score updates, does this mean we're treating score uploads as a form of manual grading? Intuitively this makes sense, but I do worry that setting `requires_manual_grading` and `last_grader` could be a little confusing for instructors... I'm not sure what I'd expect the behavior to be tbh.

nwalters512 (approved):
This is looking really good! I left a few comments, but nothing is blocking.

sprocs/instance_questions_update_score.sql
jonatanschroeder: > Since this sproc is also used for score updates, does this mean we're treating score uploads as a form of manual grading? Intuitively this makes sense, but I do worry that setting `requires_manual_grading` and `last_grader` could be a little confusing for instructors... I'm not sure what I'd expect the behavior to be tbh.

Yes, that is my expectation, that (at least for now) any manual update of the score would reset the manual grading flag and update the last grader. I'm happy to add that in the docs once we start updating docs... I added a note in our scope document to ensure this is documented. While I can certainly entertain other options once this is more widely distributed, I believe this accounts for the common cases people use to update grades for manual grading questions without the manual grading interface.

nwalters512: Makes sense! Thanks for noting that we'll have to document this. Once this feature is generally available, we might consider documenting this inline on the "Uploads" page. No need to change anything right now though.

migrations/252_instance_questions__requires_manual_grading__update.sql
```diff
@@ -0,0 +1,22 @@
+WITH ranked_submissions AS (
```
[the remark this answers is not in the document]
jonatanschroeder: > This is the query that'll need to be run again after we've deployed new code to all servers, correct? If so, could you add a comment here to that effect?

Yes, that is correct. I have updated it to also change the flag back to false if the question is graded in the period between canary and production. Comments also added.

middlewares/selectAndAuthzAssessmentQuestion.js
```diff
@@ -0,0 +1,24 @@
...
+    authz_data: res.locals.authz_data,
+    req_date: res.locals.req_date,
+  };
+  sqldb.query(sql.select_and_auth, params, function (err, result) {
+    if (ERR(err, next)) return;
+    if (result.rowCount === 0) return next(error.make(403, 'Access denied'));
+    _.assign(res.locals, result.rows[0]);
+    next();
+  });
+};
```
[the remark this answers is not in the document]
jonatanschroeder: Done.

PR #5419, opened 2022-02-24
fix: Avoid acquiring additional database clients inside transactions

nwalters512:
When we generate/save/grade questions, we do so with a transaction held open to avoid duplicate work + data. However, on chunk servers, we actually run additional database queries while the transaction is held open, but we try to run the queries with a new client. During periods of high load (or if we just happen to get 10+ requests in the same small window of time), this means that we'd end up with 10 transactions held open. The only way to get out of the transaction is through a query that requires a new DB client, but all the DB clients were in use via the transactions. This is a classic deadlock scenario that ended up with entire servers being unable to respond to requests.

This PR fixes that by using [`AsyncLocalStorage`](https://nodejs.org/api/async_context.html#class-asynclocalstorage) to ensure that any queries executed during a transaction use the transaction's client instead of trying to acquire a new one.

@mwest1066 and I realized that there's really no good reason to hold transactions open while executing question code. In a future PR, we'll come back and refactor to avoid the need for transactions at all. I opened #5420 to track this.

This resolves https://github.com/PrairieLearnInc/sysconf/issues/28.

### DIFF
--- a/prairielib/lib/sql-db.js
+++ b/prairielib/lib/sql-db.js
@@ -4,6 +4,17 @@ const pg = require('pg');
 const path = require('path');
 const debug = require('debug')('prairielib:' + path.basename(__filename, '.js'));
 const { callbackify } = require('util');
+const { AsyncLocalStorage } = require('async_hooks');
+
+/**
+ * @type {AsyncLocalStorage<import('pg').PoolClient>}
+ *
+ * We use this to propagate the client associated with the current transaction
+ * to any nested queries. In the past, we had some nasty bugs associated with
+ * the fact that we tried to acquire new clients inside of transactions, which
+ * ultimately lead to a deadlock.
+ */
+const alsTransaction = new AsyncLocalStorage();
 
 const error = require('./error');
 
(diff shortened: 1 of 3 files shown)


prairielib/lib/sql-db.js
```diff
@@ -203,7 +215,16 @@ module.exports.getClientAsync = async function () {
...
-  const client = await pool.connect();
+  /** @type {import('pg').PoolClient} */
+  let client;
+
+  // If we're inside a transaction, we'll reuse the same client to avoid a
+  // potential deadlock.
+  if (alsClient.getStore() === undefined) {
+    client = await pool.connect();
+  } else {
+    client = alsClient.getStore();
```
mwest1066, line 226: Not that it really matters, but could we just return this `client` immediately here and avoid re-setting the search schema?

mwest1066 (approved):

prairielib/lib/sql-db.js
mwest1066: Ignore this --- I actually read the code below and realized that we'll short-circuit because `client[SEARCH_SCHEMA]` will already equal `searchSchema`.

PR #5421, opened 2022-02-25
Fix slow load of student assessment page

mwest1066:
Fixes an ambiguous variable reference introduced by #4497. Without this fix, everything works fine locally and in staging, but the `question_order()` function takes about 60 seconds to run in production.

### DIFF
--- a/sprocs/question_order.sql
+++ b/sprocs/question_order.sql
@@ -1,6 +1,6 @@
 CREATE FUNCTION
     question_order (
-        assessment_instance_id BIGINT
+        arg_assessment_instance_id BIGINT
     ) RETURNS TABLE (
         instance_question_id BIGINT,
         row_order INTEGER,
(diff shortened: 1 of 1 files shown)


sprocs/question_order.sql
```diff
@@ -26,7 +26,7 @@ WITH locks_next AS (
         JOIN instance_questions AS iq ON (iq.assessment_instance_id = ai.id)
         JOIN assessment_questions AS aq ON (aq.id = iq.assessment_question_id)
     WHERE
-        ai.id = assessment_instance_id
+        ai.id = arg_assessment_instance_id
```
mwest1066, line 29: Without the `arg_` this was being interpreted by postgres as `ai.id = iq.assessment_instance_id`, so it was searching over all assessment instances in the DB, not just the one we wanted.

nwalters512 (approved):
Great catch! Let this emphasize the importance of testing in production (or at least canaries with access to production data), we presumably never catch something like this in stage.

PR #5425, opened 2022-02-27
Fix forced-question-order for unsynced courses

mwest1066:
PR #4497 only worked if a course was synced after the new code was active. It added the column `assessment_questions.effective_advance_score_perc` which had no default, so all existing questions have NULL in this column. The sync code was modified by #4497 to set a default value of `0` for the column, but that only happens after a course is synced.

The `effective_advance_score_perc` is used by #4497 in the authz code to compute the `is_sequence_locked` variable, assuming that `effective_advance_score_perc` is not NULL: https://github.com/PrairieLearn/PrairieLearn/blob/822235c5f67018a478195acf167425f0c2253f7d/sprocs/question_order.sql#L22

For unsynced courses, the NULL value in `effective_advance_score_perc` means that `is_sequence_locked` will be NULL for all questions. This is used to check authorization here: https://github.com/PrairieLearn/PrairieLearn/blob/822235c5f67018a478195acf167425f0c2253f7d/middlewares/selectAndAuthzInstanceQuestion.sql#L88

Because `is_sequence_locked = NULL`, the expression `NOT is_sequence_locked` is also NULL, so we deny access to the question.

This PR fixes the bug by explicitly using `coalesce(effective_advance_score_perc, 0)` everywhere.

### DIFF
--- a/middlewares/selectAndAuthzInstanceQuestion.sql
+++ b/middlewares/selectAndAuthzInstanceQuestion.sql
@@ -56,7 +56,7 @@ SELECT
             ELSE aq.max_points
         END,
         'remaining_points', iq.points_list[(iq.number_attempts + 2):array_length(iq.points_list, 1)],
-        'advance_score_perc', aq.effective_advance_score_perc,
+        'advance_score_perc', coalesce(aq.effective_advance_score_perc, 0),
         'sequence_locked', iqi.sequence_locked
     ) AS instance_question_info,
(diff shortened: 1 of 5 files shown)


nwalters512 (approved):
The explanation makes sense, and the code looks good! All the usual caveats from a review from my phone apply 😛

mwest1066:
Closing in favor of #5426

PR #5426, opened 2022-02-27
Fix forced-question-order for unsynced courses

mwest1066:
This is a cleaner fix for #5425 that simply adds a default value for `assessment_questions.effective_advance_score_perc`. As of PG v11 adding a default value does not cause any DB writes until rows are rewritten, so it's effectively free.

### DIFF
--- a/database/tables/assessment_questions.pg
+++ b/database/tables/assessment_questions.pg
@@ -11,7 +11,7 @@ columns
     average_submission_score_variance: double precision
     deleted_at: timestamp with time zone
     discrimination: double precision
-    effective_advance_score_perc: double precision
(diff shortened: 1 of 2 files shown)


nwalters512 (approved):
> As of PG v11 adding a default value does not cause any DB writes until rows are rewritten, so it's effectively free.

Ah, this is really cool. So the default is applied on-the-fly as the row is read, but is only persisted when it's rewritten?

mwest1066:
> Ah, this is really cool. So the default is applied on-the-fly as the row is read, but is only persisted when it's rewritten?

Exactly! https://www.2ndquadrant.com/en/blog/add-new-table-column-default-value-postgresql-11/

PR #5484, opened 2022-03-03
Allow UIN, Name and Role to be hidden in gradebook

### LINKED ISSUE #5408: Limit rendered size of UIN if too long
There may be cases where the UIN is too long and not necessary for student identification. In these cases it should be possible to render it with a max-width (or other limiting CSS property) and ellipsis overlay. This may be particularly relevant in the Gradebook page. This would be done in CSS only, so the information is still there if needed, and would not affect CSV files or downloads.

_Originally posted by @jonatanschroeder in https://github.com/PrairieLearn/PrairieLearn/pull/5406#discussion_r806356835_

### LINKED ISSUE #5483: Ability to hide UIN column in Gradebook
The UIN column takes up a lot of screen real estate in the gradebook

image

and it's not meaningful/useful for me. (for my students, it's 21 digits long)

It would be good to have the ability to hide this column, like we can hide any assessment column.

jonatanschroeder:
Resolves #5483 
Also limits the rendered size of UIN if it's too long (Resolves #5408).

### DIFF
--- a/pages/instructorGradebook/instructorGradebook.ejs
+++ b/pages/instructorGradebook/instructorGradebook.ejs
@@ -149,6 +149,11 @@
     .fixed-table-toolbar div.pagination-detail {
         margin: 0 1em 0 0 !important;
     }
+    .gradebook-uin {
+        max-width: 10em;
+        overflow: hidden;
+        text-overflow: ellipsis;
+    }
   </style>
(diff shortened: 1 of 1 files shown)


nwalters512 (approved):
Tested locally, LGTM!

Issue #4662, opened 2021-08-14
Allow passing of dynamically generated files to workspaces


glherman:
Example use case: In our class, we generate a random, obtuse piece of code that can't be easily read or interpreted. The goal is that students need to create a series of test cases to determine what the code does. Currently, we generate the code in servey.py, have the students download the code from the question.html page, and then copy-paste it into the workspace.

trombonekenny:
Expanding this issue to also cover static files, copied in from the question or ClientFiles into the workspace. For example, copying in datasets. From Slack chat on 9/2/2021

echuber2:
Would it be appropriate to do this by base64-stuffing the files into `data`? Or maybe not? If there were an additional mount (similar to the "home" mount) where things like this could be auto-generated, then the entrypoint script for the workspace could copy them into place in the user's home directory or something.

echuber2:
Oh, I don't think `data` is available in the workspace, only in the grader...

eddelbuettel:
Yes, having a set of assets (i.e. directory with content, falling back to just a file) would be very useful for us too.

echuber2:
A directory with content (if not dynamic) is already possible by putting a `workspace` subdirectory in your question. That is:

- info.json
- question.html
- etc.
- workspace
  - file1.txt ...

These things are copied into an empty, persisted mount at whatever is specified as "home" for the workspace in `info.json`. For the RStudio workspace, I suggested setting "home" as `/home/rstudio/workspace` (note, not `/home/rstudio`).

If you want _dynamic_ files, I'm not sure there's a way to customize these per-student and also make the grader aware of them before passing them through the workspace's chain of custody (thus, could be tampered with by the student). Dirk suggested he might have a hacky workaround for that already. To elaborate on the chain of custody issue, you could set up the entrypoint for the workspace to handle starter file generation; the problem is that the resulting configuration can't be passed back to the grader except by collecting the student files, and so students could always mess with the problem specification by which they would be graded.

eddelbuettel:
Ah.  We care (at least right now!) mostly about static files to let me test this.  Seems like it does what we need.

echuber2:
[...]

PR #5485, opened 2022-03-03
Mustache processing autograder

steven-wolfman:
# Mustache-Processing Version of Python Autograder for PL

This extends the Python autograder for PL to add in automatic mustache processing in a way that feels somewhat similar to how `question.html` is mustache-processed.

## Basic Idea

While staging the `tests` subfolder, the autograder's run script performs mustache processing on any file in that folder with a name that contains `.mustache`. (Hidden directories are presently skipped. Files with nothing before `.mustache` are skipped. `.mustache` can appear at the end or separated by `.` from other extensions, e.g., either `file.mustache.py` or `file.py.mustache`.) The result is stored in a file by the same name but with `.mustache` removed.

For question writers: If you have a testing file you want to have mustache processed, you add `.mustache` into its extensions and use mustache syntax as you like.

Key code:
+ `exampleCourse/questions/demo/autograder/python/mustacheProcess`: a sample question that uses mustache processing in tests
+ `graders/python`: changes to `README.md` (documentation), `install.sh` (add `mustache` program), `python_autograder/mustache-process.sh` (`.mustache` file discovery and processing script), and `python_autograder/run.sh` (inserted a mustache processing stage)

## Local Testing

(Update: The steps below work, but you should probably refer to https://prairielearn.readthedocs.io/en/latest/installingLocal/ instead!)

Unless and until the official PrairieLearn docker images are changed to add the changes to the Python autograder (the Python grading image) and the new question (the main image), testing locally is a bit painful.

To test this new code out locally, clone the pull request. Then in
the folder `graders/python/`, run:

```
docker build -t test_python_autograder .  # Builds the image named test_python_autograder
docker run -d -t test_python_autograder   # Runs and leaves running (the -t option) so you can play with it in the CLI
```

Then you need to change the sample question to use your `test_python_autograder` image. In `exampleCourse/questions/demo/autograder/python/mustacheProcess/info.json`, change the line

```
        "image": "prairielearn/grader-python",
```

to

```
        "image": "test_python_autograder",
```

Next, build a local image of PrairieLearn as a whole to test with and run it:

```
docker build -t my_pl .
docker run -it --rm -p 3000:3000  \
       -v "$PWD":/exampleCourse \
       -v $HOME/pl_ag_jobs:/jobs \
       -e HOST_JOBS_DIR=$HOME/pl_ag_jobs \
       -v /var/run/docker.sock:/var/run/docker.sock \
       --add-host=host.docker.internal:172.17.0.1 \
       my_pl
```

(Or use a command closer to what you normally use to run PrairieLearn, but note the use of `my_pl` and not `prairielearn/prairielearn`.)

Then open `XC 101` and look for the mustache test under Questions.

## Notes/Issues/Questions:

+ There's little specific to the Python autograder here. Should this be more general?
+ The mustache processing happens when the autograder prepares to run tests. It doesn't happen before that. So, e.g., you cannot currently have a file like `tests/initial_code.mustache.py` processed into `tests/initial_code.py` when rendering the question. Maybe that would be good?
+ The mustache processor does not delete the original pre-processing `*.mustache.*` files. That's easy to change by passing the `-d` option, however.
+ Might not want the `.mustache` name convention. (So in that case just process all files in a particularly subtree?)


steven-wolfman:
***I have reviewed and hereby sign the CLA***

nwalters512:
@steven-wolfman you'll have to post that without the bold/italic formatting. Sorry it's a bit confusing! It also looks like you made some commits from an email that GitHub doesn't recognize. Could you either add that email to GitHub or rebase your commits with your GitHub email?

I'll approve your CI run so we can get this moving!

steven-wolfman:

steven-wolfman:
> @steven-wolfman you'll have to post that without the bold/italic formatting. Sorry it's a bit confusing! It also looks like you made some commits from an email that GitHub doesn't recognize. Could you either add that email to GitHub or rebase your commits with your GitHub email?
> 
> I'll approve your CI run so we can get this moving!

@nwalters512 : Took me a while, but I figured out which bold/italic you meant 😝 . FWIW, I copied the bold/italic from the bot's provided text.

The checks also passed; so, maybe I'm all good otherwise?

nwalters512:
Yep, next step is code review. @nicknytko is the expert when it comes to the Python autograder, so hopefully he'll have time to review soon, but I'll take a look next week too.

nicknytko:
> Yep, next step is code review. @nicknytko is the expert when it comes to the Python autograder, so hopefully he'll have time to review soon, but I'll take a look next week too.

I'll take a look at this PR early next week, but please do poke me if I forget!

nwalters512 (commented):
On the whole, this looks great. I love what this enables instructors to do, and I think it really plays to PL's strengths of making truly individualized questions. I left a few comments that we can discuss before I approve.

I do wonder if this should be on by default or be opt-in. Hypothetically, someone could be having students author mustache files, or they could already include mustache files in their test cases. This PR could break that. This could be made opt in via an environment variable, e.g.

```json
{
  "externalGradingOptions": {
    "image": "prairielearn/grader-python",
    "environment": {
      "PROCESS_MUSTACHE_TEMPLATES": "true"
    }
  }
}
```

That said, I'd like to hear what others think before we change that.

exampleCourse/questions/demo/autograder/python/mustacheProcess/clientFilesQuestion/style.css
```diff
@@ -0,0 +1,3 @@
+table, th, td {
+    border: 1px solid;
+}
```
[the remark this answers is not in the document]
steven-wolfman: Took me a long time to get back to this. Sorry. For this one, I think the best solution is to eliminate the formatting completely. It was only there because of an example I was adapting and needlessly complicates the example. I'll also remove the folder and the reference to this from info.json.

exampleCourse/questions/demo/autograder/python/mustacheProcess/question.html
```diff
@@ -0,0 +1,44 @@
+<pl-question-panel>
+    <p>
+        <i>
+            This is a small example question that shows how we can use mustache processing on test files
+            much like <samp>question.html</samp>.  <samp>server.py</samp> randomizes a function name and
```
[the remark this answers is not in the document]
steven-wolfman: Sure. Changed and committed.

graders/python/install.sh
```diff
@@ -8,7 +8,10 @@ yum install -y \
     make \
     dos2unix \
     graphviz \
-    graphviz-devel
+    graphviz-devel \
+    rubygems
+
+gem install mustache
```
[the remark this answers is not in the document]
steven-wolfman: Ah, much better! OK. I've updated this and the `python_autograder/mustache_process.sh` file to use chevron. I tested this both on my computer and in a docker container, and it seems to be working correctly.

I personally may find some incompatibilities in my already-created problems btw chevron and mustache, but it should be no big deal for others adopting this fresh!

Note that I used the current-since-Jan-2021 0.14.0 version of chevron in `requirements.txt`. I'm guessing that's old enough to be a reasonable choice.

steven-wolfman:
I've updated the three smaller comments and prototyped using an environment variable to control whether mustache substitution occurs. I used the environment variable `PRAIRIELEARN_PROCESS_MUSTACHE_TEMPLATES`. That's pretty long and not what you proposed @nwalters512 , but I wanted to make sure that just `PROCESS_MUSTACHE_TEMPLATES` would be OK. Is there a standard in PL for the names of environment variables? Do they all start with something?

If this all looks good to you, should I try to merge the main branch back into this PR to see if I've created any new problems?

echuber2:
> Is there a standard in PL for the names of environment variables? Do they all start with something?

I don't think so. In a few cases `PL_` has been used, but several environment variables have no decoration.

It seems like this is a useful feature but I wish there was a clean way to extend the functionality to other graders too, as a middleware or something. Would it make sense to add the feature to PL itself instead of building it into each grading Docker? Maybe in the future if not now?

We've wanted something similar for passing customized starter files to workspaces. (See https://github.com/PrairieLearn/PrairieLearn/issues/4662) Since individualized starter files would go hand in hand with a customized grader (or test cases), maybe a general solution would allow customizing the question in all of these ways.

exampleCourse/questions/demo/autograder/python/mustacheProcess/server.py
```diff
@@ -0,0 +1,32 @@
...
+        "output": output
+    } for (input, output) in zip(input_words, output_words)]
+
+    # Pick the function name
+    function_name = random.choice(FUN_CHOICES)
+
+    data["params"]["names_for_user"] = []
+    data["params"]["names_from_user"] = [
+        {"name": function_name, "description": "the requested function", "type": "function (str -> str)"}
+    ]
```
nwalters512, line 28: Nit: are these params used? I don't see them referenced anywhere.

exampleCourse/questions/demo/autograder/python/mustacheProcess/tests/setup_code.py
```diff
@@ -0,0 +1,2 @@
+# Anything in this file will be run before the student's code
+# Use it to generate anything needed for the solution
```
nwalters512, line 2: We can remove this file.

exampleCourse/questions/demo/autograder/python/mustacheProcess/tests/ans.mustache.py
```diff
@@ -0,0 +1,8 @@
+{{#params}}
+def {{function_name}}(input):
+    {{#pairs}}
+    if input == "{{input}}":
+        return "{{output}}"
+    {{/pairs}}
+    return {{default_output}}
+{{/params}}
```
nwalters512, line 8: Seeing GitHub try and fail to deal with the syntax here, I'm wondering if we should mandate that `.mustache` is at the end of a filename, rather than in the middle of it? Is there a specific use case for `*.mustache.*`?

exampleCourse/questions/demo/autograder/python/mustacheProcess/tests/test.mustache.py
```diff
@@ -0,0 +1,27 @@
+from pl_helpers import name, points, not_repeated
+from pl_unit_test import PLTestCaseWithPlot, PLTestCase
+from code_feedback import Feedback
+from functools import wraps
+
+{{#params}}
+class Test(PLTestCaseWithPlot):
```
nwalters512, line 7: Nit: AFAICT, this example question doesn't deal with plots. Is this intentional? If not, let's use `PLTestCase` instead.

nwalters512 (commented):
My last outstanding question is re: `*.mustache` vs `*.mustache.*`.

> It seems like this is a useful feature but I wish there was a clean way to extend the functionality to other graders too, as a middleware or something. Would it make sense to add the feature to PL itself instead of building it into each grading Docker? Maybe in the future if not now?

Yeah, I think this is broadly useful. I think as soon as someone wants to introduce this functionality into another grader image, we should implement this in PL instead. It'd be pretty straightforward to add this functionality to [`buildDirectory`](https://github.com/PrairieLearn/PrairieLearn/blob/cccb4941e888eb9dc558c86c220be6dbaad6a171/lib/externalGraderCommon.js#L27). We could then add config to the `externalGradingOptions` object in `info.json`, e.g.:

```json
{
  "externalGradingOptions": {
    "mustache": {
      "enabled": true,
      "files": ["/grade/tests"]
    }
  }
}
```

I'm happy to merge this PR and save this for a future task though.

nwalters512:
One more thing: could you document this in `docs/python-grader/index.md`?

exampleCourse/questions/demo/autograder/python/mustacheProcess/server.py
steven-wolfman: Thanks! These are unused and now removed.

exampleCourse/questions/demo/autograder/python/mustacheProcess/tests/setup_code.py
steven-wolfman: Done. Thanks!

exampleCourse/questions/demo/autograder/python/mustacheProcess/tests/ans.mustache.py
steven-wolfman: I don't know what the right choice is here, but I can tell you why I allowed the `*.mustache.*` version: I don't have any mustache syntax highlighting/checking in my editor. So, it's the language dialect that does a better job for me. If I put `.mustache.` in the middle, that means I get as much Python (or whatever language) highlighting as the editor can give me.

To make this work a little better, I try to avoid mustache tags outside of strings and comments where that's clean to do. I've updated my files to do this (but there are still 'visible' mustache tags because of things like identifier names or unquoted values).

Thoughts?

I guess the perfect world solution would be to get some sort of hybrid of mustache and language-specific formatting, but I don't think I'm going to get there.

exampleCourse/questions/demo/autograder/python/mustacheProcess/tests/test.mustache.py
steven-wolfman: Entirely unintentional. I switched this to `PLTestCase` and also deleted unused imports (including `not_repeated`). I haven't tested this yet!

steven-wolfman:
@nwalters512 : I've pushed resolutions for all the small things.

This leaves two larger remaining issues:

1. Can we do this as "middleware" of some sort so any external grader could benefit, and it could apply to more than just test files? I'd love to do this, but I haven't mucked about in the JS sections of PL yet. I'll put some notes/thoughts below on what I might do and where I'm unsure how to proceed. Can you advise? Should I just finish off this PR and worry about generalization later?
2. Documentation, as in `docs/python-grader/index.md`. After resolving 1, I'll work on this.

So, how would I think about proceeding on middleware?

You suggested looking at [`buildDirectory`](https://github.com/PrairieLearn/PrairieLearn/blob/cccb4941e888eb9dc558c86c220be6dbaad6a171/lib/externalGraderCommon.js#L27). I think to implement that, I need to decide and learn a few things:

1. Question: If I'm mucking about in the JS, any tips on where I should look to get started with that (e.g., to make sure I can test/debug what I'm doing effectively)?
2. Decision: Should I reimplement the functionality of [`mustache-process.sh`](https://github.com/steven-wolfman/prairielearn-autograder-prolog/blob/main/serverFilesCourse/prolog-grader/mustache-process.sh) in JavaScript?

    If so, can I use the `glob` library? If that's not already available in PL, how do I add it to the set of libraries required?

    If not, how should I appropriately call on `mustache-process.sh` properly from the JS (and where do I put `mustache-process.sh` to make that happen)? (I can look up how to make a shell call in JS.. but seems worth asking to try to do it right!) 
4. Double-check: I assume however I implement this that I should insert it as a new action in the `async.series` array just before the one that refers to `submission.submitted_answer._files`.
5. Question: How do I access the external grader options from inside my action?

I suspect I'll run into more problems as I dive into the JS. Still, this seems like a better choice than doing a Python version and then immediately saying "oh, and I hope I can use this for Prolog and Haskell in my CPSC 312 course as well!"

eliotwrobson:
I know this is old and I don't want to rain on the parade too much here (since as @nwalters512 said above, this PR allows for lots of randomization that is one of the big strengths of PL), but I have _strong_ reservations about the way that this implements randomization of autograded questions. In particular, the use of mustache templating adds an extra layer of complexity to the autograder and makes it difficult to both debug and use popular tools during development (I don't think that `black` will even run on the files in here).

Unlike C (where something like this might be more appropriate), Python is already a super flexible language. In particular, the use case in the example question can be handled more cleanly just by having the autograder take in parameter strings from the `data` dictionary and using the `getattr` function. Having some functionality like this would be great to add to the existing autograder, but I think we should have a more detailed discussion about the potential use cases and whether or not they can be implemented more cleanly by other means.

eliotwrobson:
Now that https://github.com/PrairieLearn/PrairieLearn/pull/7903 has been merged with a version of the example question from this PR, I'm going to go ahead and close this one. As I said in my previous comment, I think allowing greater randomization is a good idea, but we need some use cases that aren't already covered by the existing autograder before considering adding a feature like this.

PR #5492, opened 2022-03-07
fix: Reuse DB client for queries executed with lock held

nwalters512:


nwalters512:
After conversing with @mwest1066, we've decided to take a different approach.

PR #5493, opened 2022-03-08
fix: Invoke callback even if server job is running

nwalters512:
If the `errorAbandonedJobs` cronjob happens to run while a job is in progress, it will fail to invoke `callback()`, which in turn won't ever release the lock. Since locks are implemented via DB clients, it'll hold the client until our idle-in-transaction timeout is hit 10 minutes later and kill the server.

This PR ensures that we always call `callback()`.

### DIFF
--- a/lib/server-jobs.js
+++ b/lib/server-jobs.js
@@ -393,6 +393,8 @@ module.exports.errorAbandonedJobs = function (callback) {
             }
             callback(null);
           });
+        } else {
+          callback(null);
         }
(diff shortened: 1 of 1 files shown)


mwest1066 (approved):
I can't wait to have these annoying `idle-in-transaction timeout` errors gone forever! 🎉

mwest1066:
@nicknytko you will enjoy this PR 😄

PR #5489, opened 2022-03-04
Execute question code outside of transactions

nwalters512:
Closes #5420.

Most of the work here involved ensuring that we can safely grade variants and assessments multiple times, even concurrently, which is required for correct behavior without locks. See inline comments for more discussion of what changed in this PR!

PR #5494, opened 2022-03-08
refactor: Wrap database pool in class for reusability

nwalters512:
This PR introduces a `PostgressPool` class that encapsulates our opinionated layer on top of `node-postgres`. The `sqldb` module retains the same interface as before - that is, we still expose a default pool. However, this change allows us to potentially have multiple database pools, which we'll take advantage of in the the future to improve how our named locks work.


prairielib/lib/sql-db.js
```diff
@@ -124,762 +112,809 @@ function escapeIdentifier(identifier) {
...
-  // Note that this schema name will need quoting because of characters like ':', '-', etc
-  const schema = `${truncPrefix}_${timestamp}_${suffix}`;
-  await module.exports.setSearchSchema(schema);
-  return schema;
+const defaultPool = new PostgresPool();
+
+module.exports = {
+  ...defaultPool,
+  // Class methods are non-enumerable, so we can't spread them. We'll manually
+  // attach them to `module.exports`.
```
mwest1066, line 891: Ah, I didn't know this. Explicit pass-throughs still aren't too bad here.

prairielib/lib/sql-db.test.js
```diff
@@ -0,0 +1,32 @@
+/* eslint-env jest */
+const sqldb = require('./sql-db');
+
+describe('sqldb', () => {
+  it('exports the full PostgresPool interface', () => {
```
mwest1066, line 5: Very nice to check this here!

prairielib/lib/sql-db.js
```diff
@@ -124,762 +112,809 @@ function escapeIdentifier(identifier) {
...
+    /**
+     * @type {AsyncLocalStorage<import('pg').PoolClient>}
+     * @private
+     *
+     * We use this to propagate the client associated with the current transaction
+     * to any nested queries. In the past, we had some nasty bugs associated with
+     * the fact that we tried to acquire new clients inside of transactions, which
+     * ultimately lead to a deadlock.
+     */
+    this.alsClient = new AsyncLocalStorage();
```
mwest1066, line 134: Do we in fact still want the AsyncLocalStorage code here? I'm worried that we might inadvertently rely on it, only to have it break in some complex and impossible-to-debug way. An alternative might be to use AsyncLocalStorage to throw an error if anyone tries to nest a DB call inside a transaction?

mwest1066 (approved):

prairielib/lib/sql-db.js
nwalters512: Until we merge https://github.com/PrairieLearn/PrairieLearn/pull/5489, I want to keep this here. Even then, I think it's still useful, and it mirrors how transactions work in things like Ruby on Rails (no need to manually pipe around a client, which is a footgun). That said, after the other PR, we really shouldn't have any more transactions, so this code should remain unused.

PR #5497, opened 2022-03-08
Use separate database connection pool for named locks

nwalters512:
This PR changes our "named locks" code to use a separate pool of database connections to avoid deadlocks. This should help with resolving some of the idle-in-transaction timeout and 504 errors that we've seen recently.


lib/named-locks.js
```diff
@@ -45,7 +45,27 @@ const sql = sqlLoader.loadSqlEquiv(__filename);
  * form `{client, done}`, where `client` and `done` are sqldb
  * transaction objects. These are simply the objects we need to end
  * the transaction, which will release the lock.
+ *
+ * Importantly, we use a separate pool of database connections for acquiring
+ * and holding locks. This ensures that we don't end up with a deadlock.
+ * For instance, if we have a pool of 10 clients and there are 10 locks held at
+ * once, if any code inside of a lock tries to acquire another database client,
+ * we'd deadlock if we weren't separating the two pools of connections.
+ */
```
mwest1066, line 54: Should we add a warning somewhere that we must not nest locks? Or do we think it's ok to nest them, given the AsyncLocalStorage code?

prairielib/lib/sql-db.js
```diff
@@ -916,5 +916,6 @@ module.exports = {
   setSearchSchema: defaultPool.setSearchSchema,
   getSearchSchema: defaultPool.getSearchSchema,
   setRandomSearchSchemaAsync: defaultPool.setRandomSearchSchemaAsync,
-  PostgresPool,
 };
+
+module.exports.PostgresPool = PostgresPool;
```
mwest1066, line 921: Just out of curiosity, was this change necessary?

server.js
```diff
@@ -1720,28 +1721,30 @@ if (config.startServer) {
...
-        sqldb.init(pgConfig, idleErrorHandler, function (err) {
-          if (ERR(err, callback)) return;
-          logger.verbose('Successfully connected to database');
-          callback(null);
-        });
+        }
+
+        logger.verbose(`Connecting to ${pgConfig.user}@${pgConfig.host}:${pgConfig.database}`);
+
+        await sqldb.initAsync(pgConfig, idleErrorHandler);
```
mwest1066, line 1741: Does this init also need to be added to the test initialization code? https://github.com/PrairieLearn/PrairieLearn/blob/073606f8c65addc40684c12fa7d74cd98cc7c769/tests/helperServer.js#L47

mwest1066 (approved):

server.js
nwalters512: Yes! Great catch.

prairielib/lib/sql-db.js
nwalters512: I'm not 100% sure *why*, but TypeScript didn't like it when I tried to do something like `/** @type import('./sql-db').PostgresPool */` with the old way 🤷

server.js
nwalters512: Done in 099d7f9.

lib/named-locks.js
```diff
@@ -1,13 +1,13 @@
...
 
-const sqldb = require('../prairielib/lib/sql-db');
+const { PostgresPool } = require('../prairielib/lib/sql-db');
 const sqlLoader = require('../prairielib/lib/sql-loader');
 
 const sql = sqlLoader.loadSqlEquiv(__filename);
 
 /** @typedef {import('pg').PoolClient} PoolClient */
-/** @typedef {{ client: PoolClient, done: (release?: any) => void }} Lock */
+/** @typedef {{ client: PoolClient }} Lock */
```
nwalters512, line 10: Calling out that I got rid of an unused `done` on the Lock object.

lib/named-locks.js
```diff
@@ -120,42 +144,43 @@ module.exports.doWithLock = async function (name, options, func) {
...
     if (options.wait && options.timeout > 0) {
-      await sqldb.callWithClientAsync(client, 'lock_timeout_set', [options.timeout]);
+      // SQL doesn't like us trying to use a parameterized query with
+      // `SET LOCAL ...`. So, in this very specific case, we do the
+      // parameterization ourselves using `escapeLiteral`.
+      await pool.queryWithClientAsync(
+        client,
+        `SET LOCAL lock_timeout = ${client.escapeLiteral(options.timeout.toString())}`,
+        {}
+      );
```
nwalters512, line 162: Noting that we no longer have access to sprocs here (since we never set the search path to the random one where we created the sprocs. I suppose I could pass that through to here, but it was simpler to just remove the usage of the sproc entirely.

lib/named-locks.sql
```diff
@@ -11,6 +11,9 @@ VALUES
     ($name)
 ON CONFLICT (name) DO NOTHING;
 
+-- BLOCK set_lock_timeout
+SET LOCAL lock_timeout TO $timeout;
+
```
nwalters512, line 16: Left over from experimentation.

```suggestion
```

tests/helperDb.js
```diff
@@ -28,13 +29,18 @@ async function runMigrationsAndSprocs(dbName, mochaThis, runMigrations) {
...
+  await namedLocks.init(pgConfig, idleErrorHandler);
+
   if (runMigrations) {
     // @ts-expect-error
     await util.promisify(migrations.init)(path.join(__dirname, '..', 'migrations'), 'prairielearn');
   }
 
   await sqldb.setRandomSearchSchemaAsync('test');
   await util.promisify(sprocs.init)();
+  await namedLocks.close();
```
nwalters512, line 43:
```suggestion

  await namedLocks.close();
```

lib/named-locks.js
nwalters512: Until we can remove everything in favor of `doWithLock()`, we can't rely on ALS to work correctly at all here, unfortunately. I'll add a comment.

nwalters512: Added a note in d7bb094.

mwest1066: Looks good.

mwest1066 (approved):

lib/named-locks.js
nwalters512: I think if anything, we should go the *other* direction and make it `{ __lockInternal }` or something - I think we really don't want people using this client to make queries of their own.

mwest1066: I'm happy to just yell at anyone who tries 😄. To really hide it I think we could do:
```
const my_secret_symbol = Symbol();
const lock = { my_secret_symbol: client };
return lock;
```

PR #5499, opened 2022-03-09
Fix broken clientFilesQuestion URLs in manual grading page

jonatanschroeder:
Questions with references to images or other URL-based files, either static or dynamic, were not being properly displayed in manual grading pages. This was due to the way URLs for client files and generated files are built. This PR introduces entrypoints for client files and generated files that will allow these files to be properly retrieved.

The alternatives would be to change the URL for these types of files to use an existing URL entrypoint, but that would either require unnecessary student-based authentication, or swapping the URL of the files altogether to the question preview page. I believe this is the simplest approach.


server.js
```diff
@@ -1229,6 +1229,23 @@ module.exports.initExpress = function () {
...
+      require('./middlewares/selectAndAuthzInstanceQuestion'),
+      require('./pages/clientFilesQuestion/clientFilesQuestion'),
+    ]
+  );
+
+  app.use(
+    '/pl/course_instance/:course_instance_id/instructor/instance_question/:instance_question_id/generatedFilesQuestion',
+    [
+      require('./middlewares/selectAndAuthzInstanceQuestion'),
+      require('./pages/studentGeneratedFilesQuestion/studentGeneratedFilesQuestion'),
```
nwalters512, line 1245: I did a double-take seeing a `student*` page used on an instructor route, but this makes sense because we're looking at a student's instance question. I looked at `studentGeneratedFilesQuestion` and `instructorGeneratedFilesQuestion` and they're virtually identical; I think we could merge them into a single `generatedFilesQuestion` page in a separate PR. We could unconditionally look at `res.locals.question.id`, just like `clientFilesQuestion`.

nwalters512 (approved):
I agree with this approach!

server.js
nwalters512: Actually, that might need a little more consideration so that students can't access generated files from variants that they don't own...

nwalters512: I opened #5500 with this refactor.

mwest1066 (approved):

PR #5500, opened 2022-03-10
refactor: Consolidate generatedFilesQuestion code

nwalters512:
I noticed in https://github.com/PrairieLearn/PrairieLearn/pull/5499#discussion_r823218099 that `studentGeneratedFilesQuestion` and `instructorGeneratedFilesQuestion` were identical except for the fact that the former looks at `instance_question.id` and the latter looks at `question.id`. This PR consolidates them into a single handler that always checks `question.id` and checks `instance_question.id` if it's present.


pages/generatedFilesQuestion/generatedFilesQuestion.sql
```diff
@@ -0,0 +1,13 @@
...
+FROM
+    variants AS v
+WHERE
+    v.id = $variant_id
+    -- We check both `question_id` and `instance_question_id` to make sure that
+    -- this variant is actually accessible by the user. Both will have been
+    -- validated by middleware. For the instructor preview page, we don't have
+    -- an instance question, so we ignore it in that case.
+    AND v.question_id = $question_id
+    AND (($has_instance_question IS FALSE) OR (v.instance_question_id = $instance_question_id));
```
nwalters512, line 13: I couldn't find a better way to write this, since unfortunately it seems like we can't use query parameters multiple times in the same query. I didn't want to make this a full-blown sproc, so this is where I ended up. I'm tired and kinda sick though, so I'm probably missing some obviously-better way to write this 😅

pages/generatedFilesQuestion/generatedFilesQuestion.js
```diff
@@ -12,6 +12,10 @@ router.get('/variant/:variant_id/*', function (req, res, next) {
   var variant_id = req.params.variant_id;
   var filename = req.params[0];
   var params = {
+    // The instance question generally won't be present if this is used on
+    // an instuctor route.
```
jonatanschroeder, line 16:
```suggestion
    // an instructor route.
```

jonatanschroeder (approved):

pages/generatedFilesQuestion/generatedFilesQuestion.sql
mwest1066: I might be misunderstanding what you mean here, but you _can_ use `$instance_question_id` multiple times in the same query, for example. The value of `xxx` just gets templated in for every occurrence of `$xxx`.

PR #5503, opened 2022-03-10
fix: Set timeout for GitHub Actions workflow

nwalters512:
This avoids the scenario where CI runs for 6 hours after the JS tests get stuck 😬


.github/workflows/main.yml
```diff
@@ -4,8 +4,8 @@ on: [push, pull_request]
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
+    timeout-minutes: 30
```
nwalters512, line 8: This is a little over 2x the average time our CI runs take. If this exceeds that in the future, we can tweak this.

Although... this doesn't account for a PR where we need to rebuild a bunch of images. Maybe setting the timeout specifically on the JS tests step would be better? I don't think we've seen other things frequently time out.

nwalters512: I ended up making this change in d6f233b.

mwest1066 (approved):

PR #5520, opened 2022-03-14
feat: Improve performance of assessment_instances_select_log

nwalters512:
This PR introduces two performance optimizations to the `assessment_instances_select_log` sproc:

- Restructures how we fetch from `page_view_logs`. Previously, we had a nested `SELECT 1 ...` query that would check for group `join` events in order to determine which users we had to show logs for. However, an `EXPLAIN ANALYZE ...` revealed that this resulted in a sequential scan of `group_logs` *per page view log*, which was very slow. Now, we first do a single filter of `group_logs` based on the `assessment_instance` ID to find all the user IDs we should fetch `page_view_logs` for, and then we filter `page_view_logs` on `assessment_instance_id` and `authn_user_id`. We can then reuse the results of that when we check for assessment/variant view logs.
- Adds indexes to the `group_logs` table so that we don't have to do a full sequential scan of it every time we want to query for logs from a particular group. This index may or may not end up being used in the `assessment_instances_select_log` sproc, as the group logs table is still small enough that the query planner may choose to do a sequential scan instead of an index scan.

### DIFF
--- a/migrations/248_group_logs__indexes_add.sql
+++ b/migrations/248_group_logs__indexes_add.sql
@@ -1,0 +1,6 @@
+-- Optimizes for "find all logs for a certain user in a certain group"
+-- queries that's used for things like generating the log of all assessment events.
+CREATE INDEX group_logs_group_id_user_id_idx ON group_logs (group_id, user_id);
+
+-- Optimizes for "find all logs for a certain user" queries.
+CREATE INDEX group_logs_user_id_idx ON group_logs (user_id);
--- a/migrations/248_group_logs__indexes_add.sql
+++ b/migrations/248_group_logs__indexes_add.sql
@@ -1,6 +1,6 @@
--- Optimizes for "find all logs for a certain user in a certain group"
--- queries that's used for things like generating the log of all assessment events.
-CREATE INDEX group_logs_group_id_user_id_idx ON group_logs (group_id, user_id);
+-- Optimizes for "find all logs for a certain group" queries that are used for
+-- things like the instructor assessment instance logs.
+CREATE INDEX group_logs_group_id_idx ON group_logs (group_id);
 
 -- Optimizes for "find all logs for a certain user" queries.
 CREATE INDEX group_logs_user_id_idx ON group_logs (user_id);
--- a/sprocs/assessment_instances_select_log.sql
+++ b/sprocs/assessment_instances_select_log.sql
@@ -24,6 +24,29 @@ AS $$
 BEGIN
     RETURN query
         WITH
+        group_users AS (
+            SELECT gl.user_id
+            FROM
+                assessment_instances AS ai
+                JOIN group_logs AS gl ON (gl.group_id = ai.group_id)
+            WHERE ai.id = ai_id
+            AND gl.action = 'join'
+        ),
+        user_page_view_logs AS (
+            SELECT pvl.*
+            FROM
+                assessment_instances AS ai
+                JOIN page_view_logs AS pvl ON (pvl.assessment_instance_id = ai.id)
+            WHERE
+                ai.id = ai_id
+                -- Only include events for the assessment's user or, in case of
+                -- group assessments, for events triggered by any user that at
+                -- some point was part of the group.
+                AND (
+                    (ai.user_id IS NOT NULL AND pvl.authn_user_id = ai.user_id)
+                    OR (ai.group_id IS NOT NULL AND pvl.authn_user_id IN (SELECT * FROM group_users))
+                )
+        ),
         event_log AS (
             (
                 SELECT
@@ -337,19 +360,7 @@ BEGIN
                     JOIN questions AS q ON (q.id = pvl.question_id)
                     JOIN users AS u ON (u.user_id = pvl.authn_user_id)
                     JOIN assessment_instances AS ai ON (ai.id = pvl.assessment_instance_id)
-                WHERE
-                    pvl.assessment_instance_id = ai_id
-                    AND pvl.page_type = 'studentInstanceQuestion'
-                    -- Only include events for the assessment's user or, in case of
-                    -- group assessments, for events triggered by any user that at
-                    -- some point was part of the group.
-                    AND (pvl.authn_user_id = ai.user_id
-                         OR (ai.group_id IS NOT NULL
-                             AND EXISTS (SELECT 1
-                                         FROM group_logs AS gl
-                                         WHERE gl.action = 'join'
-                                               AND gl.group_id = ai.group_id
-                                               AND gl.user_id = pvl.authn_user_id)))
+                WHERE pvl.page_type = 'studentInstanceQuestion'
             )
             UNION
             (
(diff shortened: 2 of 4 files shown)


migrations/248_group_logs__indexes_add.sql
```diff
@@ -0,0 +1,6 @@
+-- Optimizes for "find all logs for a certain user in a certain group"
+-- queries that's used for things like generating the log of all assessment events.
+CREATE INDEX group_logs_group_id_user_id_idx ON group_logs (group_id, user_id);
+
+-- Optimizes for "find all logs for a certain user" queries.
+CREATE INDEX group_logs_user_id_idx ON group_logs (user_id);
```
nwalters512, line 6: After rewriting the queries, I'm wondering if we actually just want to change to a single non-composite index on `group_id` - we no longer have any queries where we filter on `user_id` in a `JOIN` or `WHERE` clause.

If we don't change, I assume Postgres is smart enough to use the composite index even if we don't filter by `user_id`?

sprocs/assessment_instances_select_log.sql
```diff
@@ -24,6 +24,30 @@ AS $$
...
+            FROM
+                page_view_logs AS pvl
+                JOIN assessment_instances AS ai ON (ai.id = pvl.assessment_instance_id)
+            WHERE
+                pvl.assessment_instance_id = ai_id
+                -- Only include events for the assessment's user or, in case of
+                -- group assessments, for events triggered by any user that at
+                -- some point was part of the group.
+                AND (
+                    (ai.user_id IS NOT NULL AND pvl.authn_user_id = ai.user_id)
```
nwalters512, line 47: The `ai.user_id IS NOT NULL` is new here. I don't think it's strictly necessary, but it matches what we do on the next line for groups? Alternatively, these two lines could instead become the following, since `ai_group_users` should be an empty set for non-group assessments:
```sql
AND (pvl.authn_user_id = ai.user_id OR pvl.authn_user_id IN (SELECT * FROM ai_group_users)
```

tools/pg_describe
```diff
@@ -11,6 +11,7 @@ const databaseDescribe = require('../lib/databaseDescribe');
 
 const yargs = require('yargs')
     .usage('Usage: $0 <database name> [options]')
+    .demandCommand(1)
```
nwalters512, line 14: This tool was failing with `Unknown argument: postgres` when I tried to run `./tools/pg_describe postgres -o database`. I'm not sure when this started failing (new major version of `yargs` at some point?) but this change fixes it.

sprocs/assessment_instances_select_log.sql
mwest1066: And why are we using `pvl.authn_user_id` rather than `pvl.user_id`? We should be using the effective user ID in almost all cases, I think?

tools/pg_describe
mwest1066: I run it as `pg_describe -o database postgres` because when I was at kindergarten they taught us to always put options before positional arguments 😄

sprocs/assessment_instances_select_log.sql
```diff
@@ -24,6 +24,30 @@ AS $$
 BEGIN
     RETURN query
         WITH
+        ai_group_users AS (
```
mwest1066, line 27: Add a comment that this is not just current group users (which would be cleaner) but all users who were ever in the group?

migrations/248_group_logs__indexes_add.sql
mwest1066: Postgres is indeed smart enough to use the composite index for single-column lookups, but I'm also not opposed to just having a `(group_id)` index. I'm slightly inclined to keep the `(user_id)` index because I imagine that we'll want to look up logs by user at some point, but then again, YAGNI.

mwest1066 (approved):

sprocs/assessment_instances_select_log.sql
jonatanschroeder: I agree with Matt, IS NOT NULL looks redundant. As for authn, I believe we don't want to include in the logs page views by the student if the instructor is viewing as effective student?

jonatanschroeder (approved):

sprocs/assessment_instances_select_log.sql
nwalters512: We've used `authn_user_id` for as long as this query has been around - see https://github.com/PrairieLearn/PrairieLearn/blame/ef36c97de71cde0a6b05f3dc6eb1a12646bee03e/pages/instructorAssessmentInstance/instructorAssessmentInstance.sql (this is before it made the leap to a sproc). I imagine there's value in differentiating between "instructor impersonating student and viewing variant" and "student viewing variant"?

tools/pg_describe
nwalters512: Lol, FWIW, that's what I tried originally:

```
./tools/pg_describe -o database postgres
Usage: pg_describe <database name> [options]

Options:
      --version         Show version number                            [boolean]
  -o, --output          Specify a directory to output files to
      --ignore-tables   a list of tables to ignore                       [array]
      --ignore-enums    a list of enums to ignore                        [array]
      --ignore-columns  a list of columns to ignore, formatted like
                        [table].[column]                                 [array]
  -h, --help            Show help                                      [boolean]

Examples:
  pg_describe postgres
  pg_describe userdb -o db_description --ignore-tables a b --ignore-columns
  a.col1 a.col2

Unknown argument: postgres
```

The example I posted here was just the most recent thing from my shell history 😂 

It's curious that that worked for you though - when was the last time you ran it successfully?

sprocs/assessment_instances_select_log.sql
nwalters512: Re "cleaner", I just saw a real-world example in Mariana's course where a student created a group with themselves in it, did the assessment, and then left the group. If we didn't show all past members too, we'd end up with an empty assessment log in that case, which seems undesirable from an instructor's perspective.

nwalters512: Done in 4d37fde. I also inlined my above comment in case someone ever tries to be clever and change that code 😂

sprocs/assessment_instances_select_log.sql
nwalters512: I dropped the `IS NOT NULL` bits in 4d37fde. I'm going to stick with `authn_user_id` unless we have a really compelling reason to change it.

migrations/248_group_logs__indexes_add.sql
nwalters512: 👍 I'm going to change the multicolumn index to just be an index of `group_id`. I suspect that the cardinality of `group_logs` for a given group will always be small enough that we won't see much gain from including `user_id` on a multicolumn index.

I'll keep the `user_id` index around too.

nwalters512: Done in ba3a9ac.

tools/pg_describe
mwest1066: I ran it successfully this morning, but I was running it in PT so it was using the PT version of `node_modules` 😬

sprocs/assessment_instances_select_log.sql
mwest1066: Fair enough on `authn_user_id`. At some point we should make some firm rules about when an instructor emulating a student should show up in logs.

migrations/248_group_logs__indexes_add.sql
mwest1066: 👍

sprocs/assessment_instances_select_log.sql
mwest1066: Haha, nice!

mwest1066 (approved):

PR #5522, opened 2022-03-15
fix: Speed up enrollments admin query

mwest1066:
This switches from using `submissions.date` to `instance_questions.modified_at` to count active questions. This results in about a 20x speedup.


mwest1066:
I tested this on UIUC Staging.

nwalters512 (approved):

PR #5525, opened 2022-03-20
feat: More stats for active course instance admin query

mwest1066:
This PR adds some more useful statistics to the `course_instance_active_enrollments` admin query. These should help to identify the term in which course instances were active.

I've tested this on the UIUC Staging server.


nwalters512 (approved):
This all looks reasonable. Is performance still acceptable with all the extra data?

mwest1066:
> This all looks reasonable. Is performance still acceptable with all the extra data?

Yep, performance seems fine (a few seconds on the CA server).

PR #5527, opened 2022-03-21
fix: Add missing argument when grading all assessment instances

### LINKED ISSUE #5526: Missing argument in gradeAllAssessmentInstances()
It looks like we missed one update in #5489. The `gradeAssessmentInstance()` call here: https://github.com/PrairieLearn/PrairieLearn/blob/21c26abd8efedc4e0c1d4558c01b901280c7b0c8/lib/assessment.js#L401 is missing the `requireOpen()` argument.

Looking at this code again, at the moment we will throw an error (and terminate the grading of all instances) if any of these is unexpectedly closed (e.g., due to a race). Should we change `requireOpen` to cause a simple return-with-no-error if the assessment is closed? I don't think we rely on the actually-throw-an-error behavior anywhere if an instance is unexpectedly closed.

mwest1066:
Fixes #5526.

As discussed in #5526, I think we should make `gradeAssessmentInstance()` return without error if the instance is closed and `requireOpen == true`. That could wait for a later PR, however.

### DIFF
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -398,9 +398,11 @@ module.exports = {
                 job.verbose(
                   `Grading assessment instance #${row.instance_number} for ${row.username}`
                 );
+                const requireOpen = true;
                 module.exports.gradeAssessmentInstance(
                   row.assessment_instance_id,
                   authn_user_id,
+                  requireOpen,
                   close,
                   overrideGradeRate,
                   (err) => {


jonatanschroeder (approved):
I just encountered this problem in my own independent local testing of something else, and this PR fixed it.

nwalters512 (approved):

PR #5534, opened 2022-03-23
Updated package set for R autograder (thanks, @daviddalpiaz)

eddelbuettel:
This PR updates the set of packages in the R autograder, and was prepared by @daviddalpiaz and myself.  I ran the Docker build locally, so it should not create any sort of surprises or hickups when you merge.  

Any questions, please just hit us up here or on Slack.

### DIFF
--- a/graders/r/Dockerfile
+++ b/graders/r/Dockerfile
@@ -33,8 +33,13 @@ RUN apt-get update && apt-get install -y \
 	r-cran-gh \
 	r-cran-git2r \
 	r-cran-igraph \
+	r-cran-lahman \
 	r-cran-memoise \
+	r-cran-microbenchmark \
+	r-cran-nycflights13 \
+	r-cran-palmerpenguins \
 	r-cran-png \
+	r-cran-profmem \
 	r-cran-rcpparmadillo \
 	r-cran-rex \
 	r-cran-rsqlite \
@@ -50,7 +55,7 @@ RUN apt-get update && apt-get install -y \
         && echo "options(diffobj.brightness=\"dark\")" >> /etc/R/Rprofile.site
 
 # Install additional R packages from CRAN (on top of the ones pre-built as r-cran-*)
-RUN install.r bench flexdashboard microbenchmark lintr ttdo unix
+RUN install.r bench flexdashboard gapminder lintr ttdo unix
 
 # Install plr (from sibbling PL repo), and visualTest from Mango
 RUN installGithub.r PrairieLearn/pl-r-helpers MangoTheCat/visualTest


graders/r/Dockerfile
```diff
@@ -50,7 +55,7 @@ RUN apt-get update && apt-get install -y \
         && echo "options(diffobj.brightness=\"dark\")" >> /etc/R/Rprofile.site
 
 # Install additional R packages from CRAN (on top of the ones pre-built as r-cran-*)
-RUN install.r bench flexdashboard microbenchmark lintr ttdo unix
+RUN install.r bench flexdashboard gapminder lintr ttdo unix
```
mwest1066, line 58: Are we confident that no one else is using `microbenchmark`?

eddelbuettel: See line 38. I flipped it from using 'from CRAN source' and switched to r-cran-microbenchmark.

Excellent catch, and no, I generally try _not_ to reduce the surface of what is there for these very reasons.

daviddalpiaz: No, but we're still including `microbenchmark`, just the prebuilt binary version.

eddelbuettel: (Also @daviddalpiaz and I were bouncing the diff back and forth via slack, but of course you can't see that here...)

daviddalpiaz: `r-cran-microbenchmark`

mwest1066 (approved):

graders/r/Dockerfile
eddelbuettel: <HindsightAlertMode> I could-would-shoulda have been clearer there. My bad. </HindsightAlertMode>

Now that we crossed that bridge, I could actually propose a real change to go from container

      rocker/r-ubuntu:20.04

to container 

      rocker/r-bspm:20.04

which uses `bspm` by my pal (and for [this arXiv paper, coauthor](https://arxiv.org/abs/2103.08069) @enchufa2 which "injects" a small layer of Python logic and a modified version of R's own `install.packages()` to transform all calls like

       install.packages(c("dplyr", "ggplot2"))

and alike into

       sudo apt install r-cran-dplyr r-cran-ggplot2
       
i.e. makes automagic what I did here (clumsily).  I have been using that for a year or two for all my CI at GitHub Actions.  It. Just. Works.  But it looks a little different so I didn't push it.  And the time gain is immaterial here for this container.  

(And bspm also works with Fedora giving almost all of CRAN (all 18+ thousand packages...) as rpms.  I pretend that doesn't exist to not rock my .deb-based world :sunglasses:   If I find some time or funding or a unicorn I'll (re-)build the same service for Ubuntu.  I once was very close....)

mwest1066: > ```
>    sudo apt install r-cran-dplyr r-cran-ggplot2
> ```
> 
> [...]
> 
> (And bspm also works with Fedora giving almost all of CRAN (all 18+ thousand packages...) as rpms. I pretend that doesn't exist to not rock my .deb-based world 😎 If I find some time or funding or a unicorn I'll (re-)build the same service for Ubuntu. I once was very close....)

Oh! This looks cool! Could we use this to install R packages into the `plbase` image? That is, replace this: https://github.com/PrairieLearn/PrairieLearn/blob/ea7cb1dbc1002ffd18494dc32e4484f963b21fc4/images/plbase/plbase-install.sh#L69
with a list of `yum install ...`? Or maybe RSPM is the right choice, given that `plbase` is built on Amazon Linux 2 (RedHat-like) and not Fedora?

daviddalpiaz: No to go off into too much of a tangent, but is anyone actually using the R install within `plbase`? 

I could be wrong, but it's possible that my STAT 432 was the only course to do so, and I'm not currently teaching the course, and looks like I won't be in the near future. As much as I'd like to be able to spin that course up again, I'm wondering how big that R install is, and if it would be better to just get rid of it. (And if I ever teach 432 again, add it back then, or better yet, I just do the work to transition to the actual R autograder.)

nwalters512: That's a great question. R has been somewhat of a consistent problem for us. One that we're happy to support if folks are using it, but still, the packages account for a significant part of the size of `plbase` and also takes a long time to install when managing production machines. If we could gain confidence that no one is using it, it'd make me very happy to rip it all out 🙂

daviddalpiaz: Gosh, if you can audit this and I'm the only one, then please, by all means, rip it out. Hate to burden everyone just so maybe someday I can return to a course.

eddelbuettel: Happy to discuss (but between calls now).  RSPM does _not_ have hard system-depends, bspm does.  So we have some strong (and, hey. "fact-based") feelings here.  I don't personally use rpm-based distros but I would be more than happy to help.  Maybe we can scope out a test setup.

eddelbuettel: (Previous comment come out of sequence as was a mor direct reply to @mwest1066's comment.)

The reason I set up the autograder in the first place was for added modularity so I too would say 'go ahead' and pull R from `plbase`.  Maybe a quick poll (at Slack ?) of the userbase may be prudent though...)

Issue #5540, opened 2022-03-25
Some elements are not properly disabled when submissions are not permissible


jonatanschroeder:
Element development allows the elements to use the `data["editable"]` property to disable or otherwise avoid new user input in situations like closed exams or inactive assessments (or, related to #5413, the manual grading page). While most of the simplest elements (integer, number, string, matrix input, matrix component, symbolic, multiple choice, checkbox) become disabled in this scenario, not all elements use this information. Here are the official PL elements that, at this moment, as far as I can identify, would need to be updated to disallow answer editing:

* [x] `pl-file-editor`;
* [x] `pl-file-upload`;
* [x] `pl-dropdown`;
* [x] `pl-order-blocks` (@SethPoulsen tagging you in case you'd like to do something about it);
* [x] `pl-matching` (#10038);
* [x] `pl-rich-text-editor` (tagging myself just to be fair to the people above @jonatanschroeder 😄);
* [x] `pl-drawing` (#10041).

~~`pl-threejs` (not sure if and how precisely this is still being maintained).~~ DEPRECATED

SethPoulsen:
Completed for `pl-orded-blocks` by @Frozemint in #7901

jonatanschroeder:
@nwalters512 @nicknytko with #5540 underway the only element left to address is `pl-drawing` (since `pl-threejs` is deprecated, and I assume won't be maintained anymore for issues like this).

nwalters512:
@jonatanschroeder that should be a trivial fix, I'll do that now.

nicknytko:
> @nwalters512 @nicknytko with #5540 underway the only element left to address is `pl-drawing` (since `pl-threejs` is deprecated, and I assume won't be maintained anymore for issues like this).

This might be as simple as changing this line: https://github.com/PrairieLearn/PrairieLearn/blob/28476ef6cfcf65562c9ec445c3ed5d1377b5f410/apps/prairielearn/elements/pl-drawing/pl-drawing.py#L297

nwalters512:
@nicknytko that's more or less exactly what I did 🙂

PR #5541, opened 2022-03-25
Disable elements when new answers are no longer allowed

jonatanschroeder:
Addresses part of #5540, in particular for pl-dropdown, pl-rich-text-editor, pl-file-editor and pl-file-upload.

### DIFF
--- a/elements/pl-file-editor/pl-file-editor.js
+++ b/elements/pl-file-editor/pl-file-editor.js
@@ -19,6 +19,7 @@ window.PLFileEditor = function (uuid, options) {
   this.editor.setTheme('ace/theme/chrome');
   this.editor.getSession().setUseWrapMode(true);
   this.editor.setShowPrintMargin(false);
+  this.editor.setReadOnly(!!options.readOnly);
   this.editor.getSession().on('change', this.syncFileToHiddenInput.bind(this));
 
   if (options.aceMode) {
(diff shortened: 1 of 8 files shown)


elements/pl-file-editor/pl-file-editor.js
```diff
@@ -19,6 +19,7 @@ window.PLFileEditor = function (uuid, options) {
   this.editor.setTheme('ace/theme/chrome');
   this.editor.getSession().setUseWrapMode(true);
   this.editor.setShowPrintMargin(false);
+  this.editor.setReadOnly(!!options.readOnly);
```
nwalters512, line 22: Nit: this should always be defined, so the `!!` shouldn't be necessary.
```suggestion
  this.editor.setReadOnly(options.readOnly);
```

elements/pl-rich-text-editor/pl-rich-text-editor.py
```diff
@@ -68,7 +68,7 @@ def render(element_html, data):
             'editor_uuid': uuid,
             'question': data['panel'] == 'question',
             'submission': data['panel'] == 'submission',
-            'read_only': 'true' if data['panel'] == 'submission' else 'false'
+            'read_only': 'true' if (data['panel'] == 'submission' or not data['editable']) else 'false'
```
nwalters512, line 71: Is checking the panel still necessary with the addition of the other condition?

nwalters512 (approved):
Have not tested, but the code looks good!

elements/pl-rich-text-editor/pl-rich-text-editor.py
jonatanschroeder: Yes, it is needed. `data['editable']` can still be true in the submission panel. In essence, `data['editable']` indicates that the question still allows more submissions, even if the panel being rendered is not the question panel itself.

PR #5546, opened 2022-03-28
chore(deps): Install latest patch/minor versions of all JS deps

nwalters512:
This PR consolidates a bunch of upgrades that would have otherwise been done one-by-one with Dependabot.


jonatanschroeder:
This is just for JS dependencies, right? Not python/plbase?

mwest1066 (approved):

Issue #5416, opened 2022-02-22
Implement server-side sessions


nwalters512:
Right now, users authenticate solely via a cookie. If the cookie says it's valid and the signature matches, we trust it. However, this leaves us unable to invalidate sessions, which would be really useful during a potential security incident or if we find a bug with authentication code.

We should switch to using server-side sessions that are persisted in the database. We can build this on top of http://expressjs.com/en/resources/middleware/session.html with a custom session store to interact with a new `sessions` table, which we're envisioning would have the following schema:

- `session_id`: A UUID that uniquely identifies a session; used as the session ID (the value of the cookie)
- `user_id`: Maps to a user in the `users` table
- `authn_provider_name`: Just like the field that we currently store in the signed cookie. This is useful for displaying messages that are specific to LTI sessions.
- `created_at`: Date the session was created
- `expires_at`: Stores when the session will expire. If the cookie for this session is used after this date, it's considered invalid.
- `invalidated_at`: Date the session was invalidated. We'll do soft-deletes instead of removing the row entirely so that we can use this as forensic data during a hypothetical investigation.
- `data`: A JSON column storing other arbitrary data about this session (flash messages, etc.)

We'll need some logic to handle packing and unpacking `user_id` and `created_at` from the session object. That is, `req.session.user_id` should correspond to the user ID, but we don't actually want to store that in the `data` JSON blob.

Currently PL extends sessions by N hours every time there's a request. However, we'd place a lot of pressure on Postgres if we actually tried to update the row on every single request. So, instead of that, we'll only update `expires_at` to be N hours from now if `(NOW() + N hours) - expires_at >= 24 hours`.

We should ensure that sessions and their corresponding cookies are invalidated when a user explicitly logs out.

To avoid breaking existing sessions when we deploy this change, we should write some code to migrate existing `pl_authn` cookies to the new system.

trombonekenny:
We should consider the case where a user may have different authentication scopes: like, for example, Shibboleth auth for most courses but LTI authentication for a single course instance. Right now you can only have one authenticated session so they invalidate each other when they context switch. If there was a way to have multiple sessions and have the right one scoped to the course instance it's tied to, that would make it easier for the user?

nwalters512:
Infrastructure for this was added in #7143, but we still don't use sessions for storing auth data. Currently, they're just used for flash messages.

nwalters512:
[...]

PR #5547, opened 2022-03-28
Manual grading: identify conflicting scores

jonatanschroeder:
This PR introduces logic to identify cases where two different graders provide conflicting scores to the same instance question.

### DIFF
--- a/pages/instructorAssessmentInstance/instructorAssessmentInstance.sql
+++ b/pages/instructorAssessmentInstance/instructorAssessmentInstance.sql
@@ -53,6 +53,7 @@ WHERE
 -- BLOCK select_instance_questions
 SELECT
     iq.*,
+    to_jsonb(iq.modified_at) AS modified_at,
     ((lag(z.id) OVER w) IS DISTINCT FROM z.id) AS start_new_zone,
     z.id AS zone_id,
     z.title AS zone_title,
--- a/pages/instructorAssessmentInstance/instructorAssessmentInstance.sql
+++ b/pages/instructorAssessmentInstance/instructorAssessmentInstance.sql
@@ -53,7 +53,10 @@ WHERE
 -- BLOCK select_instance_questions
 SELECT
     iq.*,
-    to_jsonb(iq.modified_at) AS modified_at,
+    -- Convert modified_at to a text representation suitable for
+    -- PostgreSQL so it can be properly interpreted when a grade
+    -- update POST is received back.
+    CAST(iq.modified_at AS TEXT) AS modified_at,
     ((lag(z.id) OVER w) IS DISTINCT FROM z.id) AS start_new_zone,
     z.id AS zone_id,
     z.title AS zone_title,
--- a/pages/instructorAssessmentManualGrading/instanceQuestion/instanceQuestion.js
+++ b/pages/instructorAssessmentManualGrading/instanceQuestion/instanceQuestion.js
@@ -18,6 +18,10 @@ router.get(
       return next(error.make(403, 'Access denied (must be a student data viewer)'));
     }
 
+    res.locals.attempted_score = req.query.attempted_score
+      ? JSON.parse(req.query.attempted_score)
+      : null;
+
     res.locals.manualGradingInterface = true;
     await util.promisify(question.getAndRenderVariant)(null, null, res.locals);
 
@@ -50,6 +54,7 @@ router.post(
         null, // uid
         null, // assessment_instance_number
         null, // qid
+        req.body.modified_at,
         req.body.submission_score_percent, // score_perc
         null, // points
         { manual: req.body.submission_note }, // feedback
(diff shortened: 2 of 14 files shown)


pages/instructorAssessmentManualGrading/instanceQuestion/instanceQuestion.js
```diff
@@ -66,7 +71,16 @@ router.post(
...
-      await sqlDb.callAsync('instance_questions_update_score', params);
+      const result = await sqlDb.callAsync('instance_questions_update_score', params);
+      if (result.rows[0]?.modified_since_query) {
+        const attempted_score = JSON.stringify({
+          score_perc: req.body.submission_score_percent,
+          feedback: req.body.submission_note,
+        });
+        return res.redirect(
+          req.baseUrl + `?attempted_score=${encodeURIComponent(attempted_score)}`
+        );
```
jonatanschroeder, line 82: @nwalters512 I'd appreciate your thoughts on this strategy. I'm using a GET query to send the attempted score/feedback information to the page, but I'm not sure if that's the best approach. Having the POST itself call the EJS renderer is an option too, but I'm not a fan of that approach. Any other thoughts?

sprocs/instance_questions_update_score.sql
```diff
@@ -74,6 +95,13 @@ BEGIN
         RAISE EXCEPTION 'found submission with id=%, but question does not match %', arg_submission_id, arg_qid;
     END IF;
 
+    IF arg_modified_at IS NOT NULL AND current_modified_at != arg_modified_at THEN
+        modified_since_query = TRUE;
```
jonatanschroeder, line 99: @nwalters512 do you think a grading log should be added in this case, even if unsuccessful?

pages/instructorAssessmentManualGrading/instanceQuestion/instanceQuestion.js
jonatanschroeder: Alternative option: instead of the sproc just returning a boolean false, there could be the creation of a `grading_jobs` record, and the URL could have the attempted grading job ID instead. This would then cause the page to retrieve the score and feedback from the grading job.

sprocs/instance_questions_update_score.sql
nwalters512: I don't think so - I think that'd be a bit confusing/misleading?

pages/instructorAssessmentManualGrading/instanceQuestion/instanceQuestion.js
nwalters512: This is fine for now. Once we implement https://github.com/PrairieLearn/PrairieLearn/issues/5416, we can implement something like [Rails flash messages](https://api.rubyonrails.org/classes/ActionDispatch/Flash.html) to persist and fetch this information on the next request. Until then, query params will be our best option, I think. We do use this pattern elsewhere, FWIW.

sprocs/instance_questions_update_score.sql
jonatanschroeder: what about a grading job?

pages/instructorAssessmentManualGrading/instanceQuestion/instanceQuestion.js
nwalters512: > I'm a bit concerned with using session-based info for this. If a grader chooses to discard their changes, I'm not sure if we should open it again if the grader decides to open the same submission, the persistent information may get weird. Even if this is expected behaviour, it should happen for the user, not for the session (i.e., if the grader switches devices or browsers, the persistent information should either go with them to the other device, or not show at all).

This information would persist only for a fraction of the second between the `POST` and the subsequent `GET`. If the subsequent `GET` fails or if the user manages to reload the page on a different device in the fraction of the second between `GET` and `POST`, I'm frankly perfectly comfortable with weird behavior here.

That said, I'm fine with attaching this information to a grading job instead 👍

pages/instructorAssessmentInstance/instructorAssessmentInstance.sql
```diff
@@ -53,6 +53,7 @@ WHERE
 -- BLOCK select_instance_questions
 SELECT
     iq.*,
+    to_jsonb(iq.modified_at) AS modified_at,
```
nwalters512, line 56: I don't actually understand what this is doing - why `to_jsonb`?

nwalters512 (approved):

pages/instructorAssessmentInstance/instructorAssessmentInstance.sql
jonatanschroeder: I forgot to make a comment on this. If `modified_at` is used as is, it will be interpreted by JS as a `Date` object, which will print the value (using regular methods like `toISOString()`) with millisecond precision. The problem is that PostgreSQL is checking the date with microsecond precision. This conversion ensures the date is converted to a string format that will be interpreted and converted back to timestamptz with the appropriate values when the data is passed back through the POST request.

I used `to_jsonb` because I noticed that the precision problem didn't happen for the middleware query, which returns a JSONB object for the instance question. If you know of a better alternative I'm happy to consider it.

nwalters512: Ahh, that makes perfect sense. I wonder if `CAST(iq.modified_at AS TEXT)` would make the intent a little clearer? I don't feel strongly about that though. But with either this or `to_jsonb`, an inline comment would be helpful 👍

jonatanschroeder: Ah, right, that looks clearer. I'll make the changes and add a comment.

PR #5553, opened 2022-03-31
fix: Speed up migration 252 for manual grading

mwest1066:
I tested the `WITH ranked_submissions` CTE from [migrations/252] on the production DB and it took about an hour and used over 200 GB of disk space to page out the CTE results. This PR speeds this up about 50x by a combination of:
1. Early restriction to questions with `grading_method = 'Manual'` (this is only a small fraction of questions).
2. Restriction to instance questions modified this semester (a small fraction of instance questions).
3. Avoiding the use of `s.*` in a CTE, which selects the `submitted_answer` and pages it to disk, including a lot of large base64-encoded python notebooks.

I also used DISTINCT with ORDER BY, rather than the window function, but I'm not sure whether this is any faster.

I believe that this PR doesn't change the result of the migration (except the date restriction (2) above) but I haven't tested it in any really useful way.

### DIFF
--- a/migrations/252_instance_questions__requires_manual_grading__update.sql
+++ b/migrations/252_instance_questions__requires_manual_grading__update.sql
@@ -12,13 +12,19 @@
 -- grading when it shouldn't.
 
 WITH ranked_submissions AS (
-    SELECT
-        v.instance_question_id,
-        s.*,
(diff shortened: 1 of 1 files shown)


jonatanschroeder (approved):
Just tested locally, and it does what it is intended for.

mwest1066:
> Actually it was by coincidence that I had just `git pull`ed and that I noticed the new branch, so I was already taking a look before you created the PR :-D

Haha, nice 😄

PR #5602, opened 2022-04-01
fix: Don't error on 'null' in HTML templates

mwest1066:
Because `typeof null === 'object'`, we would previously throw an error when trying to render `null` values: https://github.com/PrairieLearn/PrairieLearn/blob/69b1ad97bd8f2bb88fb205b487ea23929bcfdfe5/packages/html/src/index.ts#L12

This PR also prints a more informative error message when an object can't be rendered.

@nwalters512 can you remind me how to release a new version of this library after this PR is merged?

### DIFF
--- a/.changeset/fresh-months-tap.md
+++ b/.changeset/fresh-months-tap.md
@@ -1,0 +1,5 @@
+---
+'@prairielearn/html': patch
+---
+
+Fix rendering of null values in templates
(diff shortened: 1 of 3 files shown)


nwalters512 (approved):
Would appreciate it if you could add a test for this new behavior!

mwest1066:
> Would appreciate it if you could add a test for this new behavior!

Thanks for the reminder! Added in 48be0d2e2712f111fffc25dced3b8ac744d21bdb

PR #5603, opened 2022-04-01
fix: Don't install anything R-related if SKIP_R_PACKAGES is set

nwalters512:
R packages take up a lot of time and space, so I'll often set `SKIP_R_PACKAGES=yes` when I'm trying to work quickly in a non-production environment, e.g. build a new EC2 AMI to test. However, even with that flag set, we would still install `r-essentials` and `rpy2`. This PR changes that such that `SKIP_R_PACKAGES=yes` won't install *anything* R-related.

### DIFF
--- a/images/plbase/plbase-install.sh
+++ b/images/plbase/plbase-install.sh
@@ -29,8 +29,7 @@ yum -y install \
     git \
     graphviz \
     graphviz-devel \
-    # Needed by the Pillow package
-    libjpeg
+    libjpeg # Needed by the Pillow package
 
 yum clean all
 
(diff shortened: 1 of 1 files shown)


images/plbase/plbase-install.sh
```diff
@@ -69,6 +72,7 @@ if [[ "${arch}" != "aarch64" ]]; then # R is not yet supported on ARM64.
         Rscript /r-requirements.R
     fi
```
mwest1066, line 73: Can we remove the `if` statement here and unconditionally install `r-requirements.R`?

mwest1066 (approved):

images/plbase/plbase-install.sh
nwalters512: Done in 06a2da4.

PR #5606, opened 2022-04-01
fix: Improve security of cookies

nwalters512:
This change ensures that our authentication cookie won't be accessible to client-side JavaScript, which is a security best practice (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#restrict_access_to_cookies).

### DIFF
--- a/docs/dev-guide.md
+++ b/docs/dev-guide.md
@@ -811,7 +811,7 @@ function foo(p, callback) {
 
   1. We first redirect to a remote authentication service (either Shibboleth or Google OAuth2 servers). For Shibboleth this happens by the “Login to PL” button linking to `/pl/shibcallback` for which Apache handles the Shibboleth redirections. For Google the “Login to PL” button links to `/pl/auth2login` which sets up the authentication data and redirects to Google.
 
-  2. The remote authentication service redirects back to `/pl/shibcallback` (for Shibboleth) or `/pl/auth2callback` (for Google). These endpoints confirm authentication, create the user in the `users` table if necessary, set a signed `pl_authn` cookie in the browser with the authenticated `user_id`, and then redirect to the main PL homepage.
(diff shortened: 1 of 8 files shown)


mwest1066:
1. Maybe we should also set `Secure: true` on the cookies? This has almost no impact, but I think it can't hurt and could potentially mitigate a MITM intercept where the user is tricked into visiting the `http` version of the site.

2. Any reason that we don't also set `HttpOnly` in the two places below? I realize that they are much less important than the `pl_authn` cookie, but it also can't hurt? Here: https://github.com/PrairieLearn/PrairieLearn/blob/2f3b359987bebf00d45a8c32d58e3b302d117c4a/middlewares/authzWorkspaceCookieSet.js#L17 and here: https://github.com/PrairieLearn/PrairieLearn/blob/435580fee3490bc2fd1814f0d2ef77270f3fd9cb/pages/authPassword/authPassword.js#L22

mwest1066 (approved):

nwalters512:
@mwest1066 changes made in e3c596e - LMK if they look good to you!

mwest1066 (approved):

trombonekenny:
We should confirm things still work okay for local development, or other places where the cookies and https aren't necessarily a given. That's a part of my #1724, changing devMode from "the default" to one option of many, which would mean the cookies have to work. Sessions would also incorporate where we're currently setting a few different cookies, I think.

nwalters512:
@trombonekenny cookies should indeed still work locally; see https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#restrict_access_to_cookies:

> A cookie with the `Secure` attribute is only sent to the server with an encrypted request over the HTTPS protocol. It's never sent with unsecured HTTP **(except on localhost)**

trombonekenny:
@nwalters512 It's not out of character for me to point Windows and MacOS at my Linux local dev of PL. It's okay that this is broken, I guess, because we'll revisit all of it with `express-session` and be able to set all of those cookie parameters in one place.

trombonekenny:
Or I'll point my local PL behind ngrok so the other clients get https. I'm just warning that cookies are tricky. :)

PR #5607, opened 2022-04-01
feat: Wrap database migrations in transactions

### LINKED ISSUE #5530: Wrap each migration in a transaction
At the moment a single migration file can partially execute. For example, consider a migration file that contains:
```
ALTER TABLE foo ADD COLUMN bar;
UPDATE foo SET bar = complex_calculation();
```
We might interrupt this migration while it's running the second line, which would leave the first line already run but the migration won't be listed as complete in the `migrations` table, so we'll try and run it again the next time we do migrations. This will then fail, because we are trying to re-ADD a column.

Our current workaround for this is to write transactions so they can be re-applied multiple times, but this makes them more complicated to write.

To fix this properly, we should wrap each migration file in a separate transaction when running it.

nwalters512:
This PR ensures each database transaction is executed and recorded in an atopic fashion. This ensures that the database is left in a consistent state if a migration errors or is aborted while executing.

Closes #5530.

### DIFF
--- a/prairielib/lib/migrations.js
+++ b/prairielib/lib/migrations.js
@@ -127,29 +127,31 @@ module.exports._initWithLock = function (callback) {
             async.waterfall(
               [
                 (callback) => {
-                  fs.readFile(path.join(migrationDir, file.filename), 'utf8', (err, sql) => {
-                    if (ERR(err, callback)) return;
-                    callback(null, sql);
-                  });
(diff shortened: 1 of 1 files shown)


nwalters512:
Shoot:

```
ALTER TYPE ... ADD cannot run inside a transaction block
```

Looks like this won't work 😢

PR #5608, opened 2022-04-01
fix: Duplicate submission rows for group assessments

### LINKED ISSUE #4848: All submissions download file duplicate submission in case of group work
The file download for assessments for all submissions includes all submissions multiplied by the number of students in the group, even though the group info is found as a JSON array in one of the columns. I haven't tested other files.

### LINKED ISSUE #5511: Duplicate files in assessment downloads all_files.zip
In assessment downloads, the all_files.zip file seems to be duplicating some of the files, i.e., the zip has the same file name repeated multiple times. I believe the issue is related to groupwork, where if a group has multiple people files are duplicate, while a group with one person doesn't have duplicates. The code seems to lean on that conclusion, but it would require additional checking.

jonatanschroeder:
Fixes #5511
Fixes #4848 

Files associated to all submissions or to instance questions were being duplicated based on a join of group_users that wasn't even used.

### DIFF
--- a/pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql
+++ b/pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql
@@ -72,8 +72,7 @@ FROM
     JOIN course_instances AS ci ON (ci.id = a.course_instance_id)
     LEFT JOIN group_configs AS gc ON (gc.assessment_id = a.id)
     LEFT JOIN groups AS g ON (g.id = ai.group_id AND g.group_config_id = gc.id)
-    LEFT JOIN group_users AS gu ON (gu.group_id = g.id)
-    JOIN users AS u ON (u.user_id = ai.user_id OR u.user_id = gu.user_id)
+    LEFT JOIN users AS u ON (u.user_id = ai.user_id)
 WHERE
     a.id = $assessment_id
 ORDER BY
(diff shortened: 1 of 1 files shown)


pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql
```diff
@@ -72,8 +72,7 @@ FROM
     JOIN course_instances AS ci ON (ci.id = a.course_instance_id)
     LEFT JOIN group_configs AS gc ON (gc.assessment_id = a.id)
     LEFT JOIN groups AS g ON (g.id = ai.group_id AND g.group_config_id = gc.id)
-    LEFT JOIN group_users AS gu ON (gu.group_id = g.id)
-    JOIN users AS u ON (u.user_id = ai.user_id OR u.user_id = gu.user_id)
+    LEFT JOIN users AS u ON (u.user_id = ai.user_id)
```
nwalters512, line 75: Checking my understanding: I believe this means that `uid`, `uin`, `name`, and `role` will be null for group assessment instances, but this is fine because of this line?

https://github.com/PrairieLearn/PrairieLearn/blob/f207b375c361055840a12d65c5efdca1a6cbb9c9/pages/instructorAssessmentDownloads/instructorAssessmentDownloads.js#L126-L128

nwalters512 (approved):

pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql
jonatanschroeder: Yup, that's basically it. For all files that use these 3 queries, the UID/UIN/Name/Role are not used at all as columns or in any processing. The files where rows are per user (instead of per group) use a different query that remains joined with group_users:
https://github.com/PrairieLearn/PrairieLearn/blob/d149ab41ec6fbedc0b50c2c1dd91d38a22a29644/pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql#L29-L30

PR #5610, opened 2022-04-04
feat: Log an error on unused SQL params

nwalters512:


trombonekenny:
I presume this PR also fixes the places where warnings popped up with unused parameters? That's why it's so many files?

nwalters512:
> I presume this PR also fixes the places where warnings popped up with unused parameters? That's why it's so many files?

Yep - at least, for all logged issues I saw while running the tests. Just need to get the tests passing now...

lib/score-upload.js
```diff
@@ -387,24 +387,29 @@ module.exports = {
...
 
   _updateAssessmentInstanceFromJson(json, assessment_id, authn_user_id, callback) {
-    let query, id;
+    let query, params, id;
     if (_.has(json, 'uid')) {
       query = sql.select_assessment_instance_uid;
+      params = {
+        assessment_id,
+        uid: json.uid,
+        instance_number: json.instance_number,
```
nwalters512, line 396:
```suggestion
        instance_number: json.instance,
```

PR #5625, opened 2022-04-06
fix: Use data attributes in pl-xss-safe

dwang:
Instead of templating the student-provided content to a variable in the `pl-xss-safe` element, it is passed as a data attribute (`data-code`). This removes an XSS vector where a closing `</script>` tag gets interpreted, so the content is never sanitized.

For consistency, the language is also passed as a data attribute (`data-language`).

@jonatanschroeder


dwang:

nwalters512 (approved):
Tested locally - this is looking good! @jonatanschroeder I'll wait for your review before merging - security things like this could use as many eyes as possible.

jonatanschroeder (approved):

PR #5626, opened 2022-04-06
Serve user content from separate subdomains

nwalters512:
## TODO

- [ ] Test and fix v2 questions
- [ ] Figure out how to replace usages of `localStorage`.
  - Persist on the backend? 
  - Switch to cookies?


nwalters512:
An interesting case: navigate to https://q1216.nathan-prairielearn.ngrok.io/pl/course_instance/13/instructor/question/1216/preview, observe that https://nathan-prairielearn.ngrok.io/pl/course_instance/13/instructor/question/1315/clientFilesQuestion/logic-diagram.png is served successfully from the main domain. This is because there's no `Origin` header on this request, as it's made "automatically" by the browser and not by code like `fetch()`.

Things do break as expected when user code tries to request things from a different question, e.g. from that same page, fetching https://nathan-prairielearn.ngrok.io/pl/course_instance/13/instructor/question/1216/clientFilesQuestion/true.png fails with a 403.

apps/prairielearn/src/middlewares/subdomain.ts
```diff
@@ -0,0 +1,223 @@
+import express = require('express');
+import error = require('@prairielearn/error');
+import { config } from '../lib/config';
+
+/**
+ * Specifies all routes that should always be accessible from any subdomain.
+ */
+const ALLOWED_FROM_ANY_SUBDOMAIN = [/^\/assets/, /^\/cacheable_node_modules/, /^\/favicon.ico/];
```
nwalters512, line 8: There may be more routes we have to handle here - check `server.js`.

apps/prairielearn/src/middlewares/subdomain.ts
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * Specifies a list of subdomain patterns and the routes that pages served from
+ * that subdomain should be able to access.
+ */
+const SUBDOMAINS = [
```
nwalters512, line 14: I believe we'll need to add a route here for manual grading.

nwalters512:
Another problem: `<a download>` tags only work for same-origin requests: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#download. This breaks `<pl-file-download>` when loading files from `clientFilesCourse`/etc., which in the current implementation are served from different subdomains from questions themselves.

One possible solution: always serve `clientFiles*` from something nested under the question path, e.g. `/pl/course/2/question/127/clientFilesCourse`, which in turn means we could reliably serve them from the same subdomain. These files are already not cached, which means we don't lose anything by serving the same files from a bunch of different paths.

PR #5632, opened 2022-04-08
chore: Add s3rver file store to ignore list for nodemon

jonatanschroeder:
I noticed that every time I saved a file in the Web interface `nodemon` would restart the server. This seems to be related to the fact that saving a file in the file editing interfaces causes a JSON file to be added to `s3rver/file-store`, which triggers `nodemon` to restart the server. This change causes this directory to be ignored by `nodemon`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -177,7 +177,8 @@
         "ignore": [
             "public/*",
             "exampleCourse/*",
-            "testCourse/*"
+            "testCourse/*",
+            "s3rver/file-store/*"
         ],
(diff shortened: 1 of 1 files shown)


jonatanschroeder:
It may not be a bad idea to add the full `/s3rver` directory instead of just `file-store`, any problems with doing that? I'm not that familiar with how `s3rver` works.

mwest1066 (approved):

PR #5637, opened 2022-04-12
docs: Fix indentation in `workspaceOptions` list

mwest1066:
Readthedocs markdown needs 4 space indent for nested lists.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,1 +1,2 @@
 mkdocs>=1.0
+mdx_truly_sane_lists>=1.2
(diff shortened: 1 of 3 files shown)


nwalters512:
> Readthedocs markdown needs 4 space indent for nested lists.

Prettier, on the other hand... 😅 

@mfsilva22 had the same problem recently, and our "fix" was to rework things so that we didn't need an indented list. Is there some mkdocs config we can use to make it take 2-space-indented lists?

nwalters512:
Building over at https://readthedocs.org/projects/prairielearn/builds/16633212/ to make sure it works 🤞

nwalters512:
This looks good now!

Screen Shot 2022-04-12 at 08 51 26

PR #5639, opened 2022-04-13
feat: Add assessment submissions API endpoint

chrononyan:
We're currently scraping all submissions across all assessment instances to calculate potential student grades at arbitrary deadlines. It would be nice to have this endpoint, as the 2 current alternatives are:
- `all_submissions.csv`, which is frontend-only and not easily automatable
- `GET /api/v1/course_instances/:course_instance_id/assessments/:assessment_id/assessment_instances` + multiple `GET /api/v1/course_instances/:id/assessment_instances/:assessment_instance_id/submissions` calls, which gets fairly expensive as we reach several hundred/thousand assessment instances

### DIFF
--- a/api/v1/endpoints/courseInstanceAssessmentInstances/index.js
+++ b/api/v1/endpoints/courseInstanceAssessmentInstances/index.js
@@ -43,6 +43,7 @@ router.get('/:unsafe_assessment_instance_id/instance_questions', (req, res, next
 router.get('/:unsafe_assessment_instance_id/submissions', (req, res, next) => {
   const params = {
     course_instance_id: res.locals.course_instance.id,
+    unsafe_assessment_id: null,
     unsafe_assessment_instance_id: req.params.unsafe_assessment_instance_id,
     unsafe_submission_id: null,
   };
--- a/api/v1/endpoints/courseInstanceAssessments/index.js
+++ b/api/v1/endpoints/courseInstanceAssessments/index.js
@@ -62,4 +62,17 @@ router.get('/:unsafe_assessment_id/assessment_access_rules', (req, res, next) =>
   });
 });
 
+router.get('/:unsafe_assessment_id/submissions', (req, res, next) => {
+  const params = {
+    course_instance_id: res.locals.course_instance.id,
+    unsafe_assessment_id: req.params.unsafe_assessment_id,
+    unsafe_assessment_instance_id: null,
+    unsafe_submission_id: null,
+  };
+  sqldb.queryOneRow(sql.select_submissions, params, (err, result) => {
+    if (ERR(err, next)) return;
+    res.status(200).send(result.rows[0].item);
+  });
+});
+
 module.exports = router;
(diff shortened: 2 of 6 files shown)


chrononyan:

nwalters512:
Thanks for the contribution! As outlined in our [contributor guidelines](https://github.com/PrairieLearn/PrairieLearn/blob/master/CONTRIBUTING.md), please include the rationale for making this change in the PR body.

chrononyan:
Sorry, updated

nwalters512:
> which gets fairly expensive as we reach several hundred/thousand assessment instances

This is actually why I'm concerned about this endpoint being introduced. This PR shifts the "expensiveness" from the client to the server. More specifically, this would require us to buffer potentially massive amounts of data in memory, both on our database host and on the web server itself. That's already with a problem with the CSV downloads, and I'm really hesitant to introduce it in another place.

I think this kind of endpoint *could* work if we introduced pagination to our API, but we don't have any precedent for that and I haven't given much thought at all for what that would look like.

When you say "gets fairly expensive" with regards to using the current API endpoints for this, are you referring to expense in terms of time to download, or something else?

mwest1066:
> I think this kind of endpoint _could_ work if we introduced pagination to our API, but we don't have any precedent for that and I haven't given much thought at all for what that would look like.

I feel like PL effectively paginates at the moment via the "iterate over assessment instances" model, which helps to keep the per-request data size under control. I'm concerned that "all submissions for an assessment" could be a huge amount of data.

chrononyan:
Yeah that's fair, I wasn't sure if the CSV downloads were problematic or not. I think we're an edge case -- large Berkeley course, and thus heavily automated where possible.

> When you say "gets fairly expensive" with regards to using the current API endpoints for this, are you referring to expense in terms of time to download, or something else?
Pretty much, it takes several minutes to iterate over N assessments * M instances if I throttle to low RPS. Haven't extensively tested what per-request overhead looks like so I'm scared to go higher.

nwalters512:
> Pretty much, it takes several minutes to iterate over N assessments * M instances if I throttle to low RPS. Haven't extensively tested what per-request overhead looks like so I'm scared to go higher.

I can't speak to the Berkeley installation, but yeah, low RPS is a good idea - too many parallel requests (especially for the large, complex queries for submissions) does bad things to the DB.

Keep in mind that these requests are indeed pretty DB-intensive compared to most other queries, so even if we were to merge this PR, you're likely going to just end up with one request that takes a few minutes instead of N*M comparably fast requests that still take a few minutes collectively. And at that point, you run the risk of load balancer timeouts, so you might have to fall back to doing individual requests anyways!

chrononyan:
Makes sense, a potential multi-minute endpoint does sound undesirable.

`all_submissions.csv` normally takes <5s to generate a 17466-row CSV for a homework assignment in our course -- do you have any insight regarding what the difference may be? Instance over-provisioning, or does the request differ in some way?

nwalters512:
What do your submissions look like? Are they numerical answers, or code/file uploads?

I'm approaching this with the worst case scenario in mind, which is "thousands of instances each with dozens of submissions each with multi-megabyte file uploads".

chrononyan:
Oh, yes, no multi-megabyte files. Got it.

PR #5643, opened 2022-04-14
Fix desktop workspace build on arm64

nicknytko:
Was previously broken because we were always downloading x64 vscode.  Now we download the correct version based on the platform.

### DIFF
--- a/workspaces/desktop/Dockerfile
+++ b/workspaces/desktop/Dockerfile
@@ -1,7 +1,7 @@
 FROM ubuntu:20.04
 
 COPY pre-install.sh /pre-install.sh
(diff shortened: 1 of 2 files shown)


nwalters512:
Build is failing :( I wonder if `uname -r` reports something different in GitHub Actions?

nicknytko:
I think I've broken the x64 build by accident 😅 -- going to keep playing with this and ping you once I remember how bash scripting works

nwalters512:
Looks like the image built successfully! 🎉 

(Just saw that this was in draft, sorry for the premature comment)

nwalters512 (approved):

PR #5645, opened 2022-04-14
feat(opentelemetry): Don't handle SIGTERM by default

nwalters512:
Before this change, a process receiving `SIGTERM` wouldn't actually terminate. Now, it does!

Instead of having `@prairielearn/opentelemetry` handle `SIGTERM`, I opted to just expose a `shutdown()` function that a consumer could call. This makes it possible to coordinate with other handlers that need to do something before the process exits.

I tested this locally by setting `"openTelemetryEnabled": true` in my `config.json`, starting PL, finding its PID with `ps aux | grep "node server.js"`, and then running `kill -TERM $PID`

### DIFF
--- a/packages/opentelemetry/src/index.ts
+++ b/packages/opentelemetry/src/index.ts
@@ -111,6 +111,8 @@ instrumentations.forEach((i) => {
   i.enable();
 });
 
+let tracerProvider: NodeTracerProvider;
+
 export interface OpenTelemetryConfig {
   openTelemetryEnabled: boolean;
   openTelemetryExporter?: 'console' | 'honeycomb';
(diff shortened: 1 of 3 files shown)


packages/opentelemetry/src/index.ts
```diff
@@ -209,11 +211,16 @@ export async function init(config: OpenTelemetryConfig) {
   tracerProvider.register();
 
   instrumentations.forEach((i) => i.setTracerProvider(tracerProvider));
+}
 
-  // When the process starts shutting down, terminate OTEL stuff.
-  process.on('SIGTERM', () => {
-    tracerProvider.shutdown().catch((error) => console.error('Error terminating tracing', error));
-  });
```
mwest1066, line 216: Nice! I agree that we don't want to install signal handlers from inside library code. Shifting this responsibility to the caller is the right thing to do.

mwest1066 (approved):

PR #5654, opened 2022-04-16
feat: Introduce new @prairielearn/html-ejs package

nwalters512:
This PR splits EJS-related functionality out of `@prairielearn/html` and into `@prairielearn/html-ejs` so that the main `@prairielearn/html` package can safely be bundled into and used from client-side JavaScript.

We were actually relying on the `escapeXML` function from EJS to escape HTML, so I duplicated that function inside `@prairielearn/html` to remove that dependency.


packages/html/package.json
```diff
@@ -7,11 +7,8 @@
     "dev": "tsc --watch --preserveWatchOutput",
     "test": "mocha --no-config --require ts-node/register src/index.test.ts"
   },
-  "dependencies": {
-    "ejs": "^3.1.6"
-  },
   "devDependencies": {
-    "@types/ejs": "^3.1.0",
+    "@prairielearn/tsconfig": "*",
```
mwest1066, line 11: Is this `tsconfig` package also released to `npm`?

mwest1066 (approved):

packages/html/package.json
nwalters512: Nope, it's marked as private: https://github.com/PrairieLearn/PrairieLearn/pull/5654/files#diff-d75b7a6d7a2f99f72c90ad29d2d4953ec3ad2f4640af5285b1886c1a85bdf674R4

mwest1066: Ah! Very good.

PR #5656, opened 2022-04-17
feat(html): Add type constraints for interpolated values

nwalters512:
Now, we'll get compile-time warnings for trying to interpolate objects in an `html` tagged template literal. This matches the runtime behavior, which is to throw an error.

Closes https://github.com/PrairieLearnInc/PrairieTest/issues/464.

### DIFF
--- a/packages/html/src/index.test.ts
+++ b/packages/html/src/index.test.ts
@@ -31,6 +31,7 @@ describe('html', () => {
 
   it('errors when interpolating object', () => {
     assert.throws(
+      // @ts-expect-error
       () => html`<p>${{ foo: 'bar' }}</p>`.toString(),
       'Cannot interpolate object in template'
     );
--- a/packages/html/src/index.ts
+++ b/packages/html/src/index.ts
@@ -54,7 +54,9 @@ export class HtmlSafeString {
   }
 }
 
-export function html(strings: TemplateStringsArray, ...values: any[]): HtmlSafeString {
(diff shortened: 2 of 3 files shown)


packages/html/src/index.test.ts
```diff
@@ -31,6 +31,7 @@ describe('html', () => {
 
   it('errors when interpolating object', () => {
     assert.throws(
+      // @ts-expect-error
```
mwest1066, line 34: Oh nice! I didn't realize we could do that, but it makes sense that we want to.

packages/html/src/index.ts
```diff
@@ -54,7 +54,9 @@ export class HtmlSafeString {
   }
 }
 
-export function html(strings: TemplateStringsArray, ...values: any[]): HtmlSafeString {
+export type HtmlValue = string | number | boolean | HtmlSafeString | undefined | null | HtmlValue[];
```
mwest1066, line 57: _Recursive_ type definitions?? 🤯 Awesome!

mwest1066 (approved):

packages/html/src/index.test.ts
nwalters512: Yeah, `// @ts-expect-error` is awesome. Unlike `// @ts-ignore`, it actually ensures that there *is* an error there, and if there stops being an error at some point in the future, the lack of error itself becomes an error and you're forced (in a good way) to deal with it!

packages/html/src/index.ts
nwalters512: Isn't TS great?? 🤩

PR #5659, opened 2022-04-19
Update documentation of pl-block-group to match new behavior

SethPoulsen:
Better documentation of changes introduced in #5356


SethPoulsen:
@jasonxia17 Feel free to review this as well. For some reason it wouldn't let me request a review from you (maybe because you aren't part of the organization? I don't know).

nwalters512 (approved):
LGTM, though I'd like input from folks who've had experience with this element to make sure this description is accurate and useful (it is to me, at least!)

docs/elements.md
```diff
@@ -448,7 +448,7 @@ Different grading options are defined via the attribute `grading-method`:
...
+- `dag`: in this method, the `depends` attibute of the `pl-answer` options are used to declare the directed acyclic graph relation between the blocks, and a correct answer is any topological sort of that directed acyclic graph. If `pl-block-group` elements are used to divide some blocks into groups, then a correct answer is a topological sort of the lines of the proof with the added condition that the lines of each group must be listed contiguously (as an example, this is useful for expressing a proof by cases when the `dag` grader is used for mathematical proofs). Blocks inside a `pl-block-group` element may only depend or be depended on by blocks within the same `pl-block-group`, and `pl-blocks-group`s can be given their own tag so that an entire group may depend on a block, or a block may depend on an entire group.
```
jasonxia17, line 451:
```suggestion
- `dag`: in this method, the `depends` attibute of the `pl-answer` options are used to declare the directed acyclic graph relation between the blocks, and a correct answer is any topological sort of that directed acyclic graph. If `pl-block-group` elements are used to divide some blocks into groups, then a correct answer is a topological sort of the lines of the proof with the added condition that the lines of each group must be listed contiguously (as an example, this is useful for expressing a proof by cases when the `dag` grader is used for mathematical proofs). Blocks inside a `pl-block-group` element may only depend or be depended on by blocks within the same `pl-block-group`, and `pl-blocks-group`s can be given their own tag so that an entire group may depend on a block, or a block may depend on an entire group, or a group may depend on another group.
```

jasonxia17 (approved):
Looks good to me! I just have one minor change to suggest.

docs/elements.md
SethPoulsen: I'm making a slightly different edit to get the same point across in less words. Thanks!

docs/elements.md
```diff
@@ -448,7 +448,7 @@ Different grading options are defined via the attribute `grading-method`:
...
+- `dag`: in this method, the `depends` attibute of the `pl-answer` options are used to declare the directed acyclic graph relation between the blocks, and a correct answer is any topological sort of that directed acyclic graph. If `pl-block-group` elements are used to divide some blocks into groups, then a correct answer is a topological sort of the lines of the proof with the added condition that the lines of each group must be listed contiguously (as an example, this is useful for expressing a [proof by cases](demo/proofBlocks) when the `dag` grader is used for mathematical proofs). Blocks inside a `pl-block-group` element may only depend or be depended on by blocks within the same `pl-block-group`, and `pl-blocks-group`s can be given their own `tag` and `depends` properties so that a block may depend on an entire group, or a group may depend on a block or group.
```
SethPoulsen, line 451: `[proof by cases](demo/proofBlocks)`
hmm...wanted to add a link to the example problem here, @nwalters512 why is this not working like the link in the bullet points down below this does?

nwalters512: @SethPoulsen you'll need to add that to the bottom of this file: https://github.com/PrairieLearn/PrairieLearn/blob/df80443bbe31e40e84791d84686799ae1108c763/docs/elements.md?plain=1#L1969

PR #5660, opened 2022-04-19
chore: Add more `pl-order-blocks` tests

SethPoulsen:
pl-order-blocks has become increasingly complex, with many different possible combinations of options. It has become increasingly easy to accidental break certain configurations without noticing (e.g. #5368, #4666), so we need more automated tests.


eliotwrobson:
Do you think it would be a good idea to add type hints to the backend for this? Doing that has made the code for some of our custom elements cleaner.

SethPoulsen:
Ya, we're definitely all about type hints. We actually _had_ them in this codebase, but we had to remove them due to one of the PrairieLearn servers running an older version of Python (see https://github.com/PrairieLearn/PrairieLearn/pull/5292)

However, this PR is actually trying to solve a different problem--the fact that during question writing, there are many different combinations of options you can express in the HTML.

mwest1066 (approved):

PR #5667, opened 2022-04-25
Adding distractor feedback for pl-order-blocks

CoffeePoweredComputers:
This pull request add the following tags to the pl-order-blocks element:
* The `display-distractor-feedback` attribute has been added to `<pl-order-blocks>` which toggles whether or not distractors, and any feedback associated with a distractor, will be displayed in the correct answer panel. When `true`, this information is displayed, and, when `false`, the panel simply displays the correct answer.
* The `distractor-feedback` attribute has been added to `<pl-answer>`. This allows a message with some feedback on why the distractor is incorrect to be associated with the distractor block. If `display-distractor-feedback` is true, these messages will be displayed along side the distractors in the correct answer panel.

<table>

<tr>
<td>With feedback</td>
<td> Without feedback</td>
</tr>

<tr>
<td>

 image

</td>
<td> 

image

</td>
</tr>

<tr>
<td>

```html
<pl-order-blocks ... display-distractor-feedback=true>
    <pl-answer correct="true" indent=0 ranking=1>correct_block</pl-answer>
    <pl-answer correct="false"  distractor-feedback="incorrect_block">distractor</pl-answer>
</pl-order-blocks>
```

</td>

<td>

```html
<pl-order-blocks ... display-distractor-feedback=true>
    <pl-answer correct="true" indent=0 ranking=1>correct_block</pl-answer>
    <pl-answer correct="false" >distractor</pl-answer>
</pl-order-blocks>
```

</td>

</tr>

</table>

Additionally, if the `feedback` parameter is set for `"first-wrong"` it will display an indication of the block that was incorrect and, if feedback was associated with that distractor via the `<pl-answer>` tag it will include that feedback in the submission.

image
```html
<pl-order-blocks answers-name="python-ordered" grading-method="ranking" feedback='first-wrong' indentation="true" display-distractor-feedback=true>

    <pl-answer correct="true" ranking="1" indent="0">def my_sum(first, second):</pl-answer>
    <pl-answer correct="true" ranking="2" indent="1">sum = 0</pl-answer>
    <pl-answer correct="true" ranking="3" indent="1">sum += first</pl-answer>
    <pl-answer correct="true" ranking="3" indent="1">sum += second</pl-answer>
    <pl-answer correct="true" ranking="4" indent="1">return sum</pl-answer>
    <pl-answer correct="false"  distractor-feedback="This computes a sum but fails to store the result." >sum + first</pl-answer>
    <pl-answer correct="false" distractor-feedback="This computes a sum but fails to store the result." >sum + second</pl-answer>
</pl-order-blocks>
```

### DIFF
--- a/docs/elements.md
+++ b/docs/elements.md
@@ -65,6 +65,7 @@ images, files, and code display. The following **decorative** elements are avail
 - [`pl-overlay`](#pl-overlay-element): Allows layering existing elements on top of one another in specified positions.
 - [`pl-external-grader-variables`](#pl-external-grader-variables-element): Displays expected and given variables for externally graded questions.
 - [`pl-xss-safe`](#pl-xss-safe-element): Removes potentially unsafe code from HTML code.
+- [`pl-file-preview`](#pl-file-preview-element): Displays a preview of submitted files.
 
 **Conditional** elements are meant to improve the feedback and question structure.
 These elements conditionally render their content depending on the question state.
@@ -81,8 +82,7 @@ The following **Conditional** elements are available:
 - [`pl-manual-grading-only`](#pl-manual-grading-only-element): Shows content only in manual grading.
 - [`pl-external-grader-results`](#pl-external-grader-results-element):
   Displays results from questions that are externally graded.
-- [`pl-hidden-hints`](#pl-hidden-hints): Displays hints as a student submits more on the current variant.
-- [`pl-file-preview`](#pl-file-preview): Displays a preview of submitted files.
+- [`pl-hidden-hints`](#pl-hidden-hints-element): Displays hints as a student submits more on the current variant.
 
 Note: PrairieLearn Elements listed next have been
 **deprecated**. These elements are still supported for backwards
@@ -1019,37 +1019,10 @@ The `focus` attribute defaults to `"false"`. Setting this to true will cause the
 #### See also
 
 - [`pl-file-upload` to receive files as a submission](#pl-file-upload-element)
-- [`pl-file-preview` to display previously submitted files](#pl-file-preview)
+- [`pl-file-preview` to display previously submitted files](#pl-file-preview-element)
 - [`pl-external-grader-results` to include output from autograded code](#pl-external-grader-results)
 - [`pl-code` to display blocks of code with syntax highlighting](#pl-code-element)
 - [`pl-string-input` for receiving a single string value](#pl-string-input-element)
-
----
-
-### `pl-file-preview` element
-
-Provides an in-browser preview of pure-text or image files submitted by a student as part of an external grading system.
-Does not support other file types (e.g., PDF). Shows the submitted file in the corresponding submission panel.
-Used in conjunction with submission elements like `pl-file-editor`, `pl-file-upload`, and `pl-rich-text-editor`.
-Commonly appears in the submission panel with companion `pl-external-grader-results` element.
-
-#### Sample element
-
-```html
-<pl-file-preview></pl-file-preview>
-```
-
-#### Example implementations
-
-- [element/fileEditor]
-- [demo/autograder/codeEditor]
-
-#### See also
-
-- [`pl-file-editor` to provide an in-browser code environment](#pl-file-editor-element)
-- [`pl-file-upload` to receive files as a submission](#pl-file-upload-element)
-- [`pl-external-grader-results` to include output from autograded code](#pl-external-grader-results)
-- [`pl-code` to display blocks of code with syntax highlighting](#pl-code-element)
 
 ---
 
@@ -1894,6 +1867,33 @@ Note that only one of the attributes `source-file-name`, `submitted-file-name` o
 
 ---
 
+### `pl-file-preview` element
+
+Provides an in-browser preview of pure-text or image files submitted by a student as part of an external grading system.
+Does not support other file types (e.g., PDF). Shows the submitted file in the corresponding submission panel.
+Used in conjunction with submission elements like `pl-file-editor`, `pl-file-upload`, and `pl-rich-text-editor`.
+Commonly appears in the submission panel with companion `pl-external-grader-results` element.
+
+#### Sample element
+
+```html
+<pl-file-preview></pl-file-preview>
+```
+
+#### Example implementations
+
+- [element/fileEditor]
+- [demo/autograder/codeEditor]
+
+#### See also
+
+- [`pl-file-editor` to provide an in-browser code environment](#pl-file-editor-element)
+- [`pl-file-upload` to receive files as a submission](#pl-file-upload-element)
+- [`pl-external-grader-results` to include output from autograded code](#pl-external-grader-results)
+- [`pl-code` to display blocks of code with syntax highlighting](#pl-code-element)
+
+---
+
 ## Conditional Elements
 
 ### `pl-question-panel` element
@@ -2030,59 +2030,6 @@ element contents only in a specific panel.
 
 ---
 
-### `pl-hidden-hints` element
-
-Display progressive hints that become accessible as the number of student submissions increases for the current variant.
-Hints are only open on page load when they are first revealed (when first reaching the desired submission count).
-Otherwise hints start closed and must be opened by the user. The submission counter is reset when new variants are
-generated. Note that **this element does not reveal new hints across variants.**
-
-Best used in situations where there is a penalty for more submissions to a given variant. This prevents students from
-spamming incorrect submissions to reveal all hints right away.
-
-#### Sample element
-
-```html
-<pl-hidden-hints>
-  <pl-hint> This is a hint that will be accessible immediately. </pl-hint>
-
-  <pl-hint show-after-submission="3">
-    This is a hint that will be accessible after three incorrect submissions for the current
-    variant.
-  </pl-hint>
-
-  <pl-hint show-after-submission="5">
-    This is a hint that will be accessible after five incorrect submissions for the current variant.
-  </pl-hint>
-</pl-hidden-hints>
-```
-
-#### Customizations
-
-For the inner `pl-hint` element:
-
-| Attribute               | Type | Default | Description                                                                                                                                                                                                 |
-| ----------------------- | ---- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `show-after-submission` | int  | -       | Number of submissions on the current variant needed before the hint is accessible. If not set, hint is always accessible. **Should only be set for questions that allow multiple submissions per variant.** |
-
-#### Details
-
-Add hints to a variant that are revealed with more submissions using the `show-after-submission` attribute. By default, hints without
-`show-after-submission` set are always shown. Hints with the same `show-after-submission` appear in the order they're written in the
-question HTML.
-
-#### Example implementations
-
-- [element/hiddenHints]
-
-#### See also
-
-- [`pl-question-panel` for displaying the question prompt.](#pl-question-panel-element)
-- [`pl-submission-panel` for changing how a submitted answer is displayed.](#pl-submission-panel-element)
-- [`pl-hide-in-panel` to hide contents in one or more display panels.](#pl-hide-in-panel-element)
-
----
-
 ### `pl-hide-in-manual-grading` element
 
 Hide the contents so that it is **not** displayed to graders in the manual grading page.
@@ -2165,6 +2112,59 @@ It expects results to follow [the reference schema for external grading results]
 #### See also
 
 - [External Grading Reference Schema](externalGrading.md#grading-result)
+
+---
+
+### `pl-hidden-hints` element
+
+Display progressive hints that become accessible as the number of student submissions increases for the current variant.
+Hints are only open on page load when they are first revealed (when first reaching the desired submission count).
+Otherwise hints start closed and must be opened by the user. The submission counter is reset when new variants are
+generated. Note that **this element does not reveal new hints across variants.**
+
+Best used in situations where there is a penalty for more submissions to a given variant. This prevents students from
+spamming incorrect submissions to reveal all hints right away.
+
+#### Sample element
+
+```html
+<pl-hidden-hints>
+  <pl-hint> This is a hint that will be accessible immediately. </pl-hint>
+
+  <pl-hint show-after-submission="3">
+    This is a hint that will be accessible after three incorrect submissions for the current
+    variant.
+  </pl-hint>
+
+  <pl-hint show-after-submission="5">
+    This is a hint that will be accessible after five incorrect submissions for the current variant.
+  </pl-hint>
+</pl-hidden-hints>
+```
+
+#### Customizations
+
+For the inner `pl-hint` element:
+
+| Attribute               | Type | Default | Description                                                                                                                                                                                                 |
+| ----------------------- | ---- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `show-after-submission` | int  | -       | Number of submissions on the current variant needed before the hint is accessible. If not set, hint is always accessible. **Should only be set for questions that allow multiple submissions per variant.** |
+
+#### Details
+
+Add hints to a variant that are revealed with more submissions using the `show-after-submission` attribute. By default, hints without
+`show-after-submission` set are always shown. Hints with the same `show-after-submission` appear in the order they're written in the
+question HTML.
+
+#### Example implementations
+
+- [element/hiddenHints]
+
+#### See also
+
+- [`pl-question-panel` for displaying the question prompt.](#pl-question-panel-element)
+- [`pl-submission-panel` for changing how a submitted answer is displayed.](#pl-submission-panel-element)
+- [`pl-hide-in-panel` to hide contents in one or more display panels.](#pl-hide-in-panel-element)
 
 ---
 


CoffeePoweredComputers:
@SethPoulsen Mind taking a look at this when you get a chance?

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -456,6 +481,9 @@ def grade(element_html, data):
             elif feedback_type == 'first-wrong':
                 if first_wrong == -1:
                     feedback = FIRST_WRONG_FEEDBACK['incomplete']
+                elif display_distractor_feedback and student_answer[first_wrong]['tag'] is None:
+                    text = student_answer[first_wrong]['inner_html']
+                    feedback += FIRST_WRONG_FEEDBACK['distractor-feedback'].format(str(first_wrong + 1), text, distractor_feedback[text] or '')
```
SethPoulsen, line 486: Currently, the way you have your `display-distractor-feedback` option set up, it will affect what feedback is shown during intermediate attempts at a problem, AND the feedback that is shown to students after their last attempt at a problem (when they are shown the correct answer). 

I think that these things represent different instructional design decisions and so they should be controlled separately from one another.

Also, up to this point, the feedback that students are given on intermediate attempts has been purposely vague, not telling them all the information about why their solution is wrong. If we are going to give them more specific feedback in the case then they chose a distractor, do we also want to give them more specific feedback for the other cases? (out of order, or unfinished block-group)? In my mind, the option `feedback=first_wrong` should mean that the only thing the student is told is which block is the first one that is wrong, and if we want to give them more information than that, it should be a different option, `feedback=???` I don't know what to call it though.

SethPoulsen (commented):
Looks awesome overall!

I think there are just a few places we need to think more carefully about what behaviors we are introducing and how they are going to affect all the others using the element before we can merge.

elements/pl-order-blocks/pl-order-blocks.py
CoffeePoweredComputers: In terms of thing to call this feedback here's a few things that come to mind:
* `first_wrong_verbose`
* `first_wrong_w_distractor_feedback`
* `first_wrong_w_feedback`
* `first_wrong_show_feedback`
Any of these, or perhaps a fork of one of these, look okay?

CoffeePoweredComputers: I ended up going with first-wrong-verbose for the time being.

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -374,15 +396,19 @@ def parse(element_html, data):
...
     student_answer = data['submitted_answers'][answer_name]
+
     grading_mode = pl.get_string_attrib(element, 'grading-method', GRADING_METHOD_DEFAULT)
     check_indentation = pl.get_boolean_attrib(element, 'indentation', INDENTION_DEFAULT)
     feedback_type = pl.get_string_attrib(element, 'feedback', FEEDBACK_DEFAULT)
     answer_weight = pl.get_integer_attrib(element, 'weight', WEIGHT_DEFAULT)
     partial_credit_type = pl.get_string_attrib(element, 'partial-credit', 'lcs')
-
     true_answer_list = data['correct_answers'][answer_name]
+    display_distractor_feedback = pl.get_boolean_attrib(element, 'display-distractor-feedback', DISPLAY_DISTRACTORS_DEFAULT)
```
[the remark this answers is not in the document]
CoffeePoweredComputers: This line ended up being unneeded due to the new `'first-wrong-verbose'` tag.

CoffeePoweredComputers:
Updates:

* The `distractor-info` attribute has been added to `<pl-order-blocks>` which has the following options:
    * `show-distractors` which only displays the distractors in the correct answer panel.
    * `show-feedback` which displays the distractors and the feedback associated with each distractor.
    *  `none` (default) whic displays no information on distractors

<table>

<tr>
<td>With feedback</td>
<td> Without feedback</td>
</tr>

<tr>
<td>

 image

</td>
<td> 

image

</td>
</tr>

<tr>
<td>


```html
<pl-order-blocks ... distractor-info="show-feedback">
    <pl-answer correct="true" indent=0 ranking=1>correct_block</pl-answer>
    <pl-answer correct="false"  distractor-feedback="incorrect_block">distractor</pl-answer>
</pl-order-blocks>
```

</td>

<td>


```html
<pl-order-blocks ... display-distractor-feedback="show-distractors">
    <pl-answer correct="true" indent=0 ranking=1>correct_block</pl-answer>
    <pl-answer correct="false" >distractor</pl-answer>
</pl-order-blocks>
```

</td>

</tr>

</table>

Additionally, if the `feedback` parameter is set for `"first-wrong"` it will display an indication of the block that was incorrect and, if feedback was associated with that distractor via the `<pl-answer>` tag it will include that feedback in the submission.

image
```html
<pl-order-blocks .... feedback='first-wrong-verbose' >

    <pl-answer correct="true" ranking="1" indent="0">def my_sum(first, second):</pl-answer>
    <pl-answer correct="true" ranking="2" indent="1">sum = 0</pl-answer>
    <pl-answer correct="true" ranking="3" indent="1">sum += first</pl-answer>
    <pl-answer correct="true" ranking="3" indent="1">sum += second</pl-answer>
    <pl-answer correct="true" ranking="4" indent="1">return sum</pl-answer>
    <pl-answer correct="false"  distractor-feedback="This computes a sum but fails to store the result." >sum + first</pl-answer>
    <pl-answer correct="false" distractor-feedback="This computes a sum but fails to store the result." >sum + second</pl-answer>
</pl-order-blocks>
```

CoffeePoweredComputers:
@mwest1066 Seth mentioned that you have some changes you'd like made to the attributes and the design you were planning to request on this PR. I just thought I'd bump this thread to see if I could get those requests so I can try and update things this week.

docs/elements.md
```diff
@@ -417,22 +417,25 @@ Element to arrange given blocks of code or text that are displayed initially in
...
+| `format`              | "code" or "default"     | "default"                      | If this property is set to "code", then the contents of each of the blocks will be wrapped with a `pl-code` element.                                                                                                                                                                                                                                                                                     |
+| `code-language`       | string                  | -                              | The programming language syntax highlighting to use. Only available when using `format="code"`.                                                                                                                                                                                                                                                                                                          |
+
```
SethPoulsen, line 438: This shouldn't be showing up here, looks like some sort of merge issue.

SethPoulsen: oh sorry i was reading the diff wrong

elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -62,7 +62,12 @@ $(function() {
         </ul>
         <ul class="list-group submission_table">
             {{#student_submission}}
-            	<li class="{{block_formatting}}" style="margin-left: {{indent}}px">{{{inner_html}}}</li>
+            	<li class="{{block_formatting}}" style="margin-left: {{indent}}px">
+                <div class="badge {{badge_type}}" style='display: table-cell; vertical-align: middle;'>
```
SethPoulsen, line 66: The badge should probably be in a `span`, not a `div`, because its supposed to be inline. Once you do that, you should have to have the `display: table-cell` hack.

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -233,17 +247,18 @@ def render(element_html, data):
         solution_header = pl.get_string_attrib(element, 'solution-header', SOLUTION_HEADER_DEFAULT)
 
         mcq_options = data['params'][answer_name]
-        mcq_options = filter_multiple_from_array(mcq_options, ['inner_html', 'uuid'])
+        mcq_options = filter_multiple_from_array(mcq_options, ['inner_html', 'uuid', 'distractor-feedback'])
 
         if answer_name in data['submitted_answers']:
             student_previous_submission = filter_multiple_from_array(data['submitted_answers'][answer_name], ['inner_html', 'uuid', 'indent'])
-            mcq_options = [opt for opt in mcq_options if opt not in filter_multiple_from_array(student_previous_submission, ['inner_html', 'uuid'])]
+            mcq_options = [opt for opt in mcq_options if opt['uuid'] not in {block['uuid'] for block in student_previous_submission}]
```
SethPoulsen, line 255: @CoffeePoweredComputers the change on this line should fix the bug you had with blocks being duplicated. The blocks were failing the set membership check here because of the new 'distractor-info' property you added. Obviously, this was a poorly designed check, so I changed it to just check UUIDs instead (as we should). 

I can't test the fix myself though, because I'm getting errors other places in the code, you must have fixed them locally but not pushed yet. Anyway, let me know how it goes!

SethPoulsen:
Something very odd going on here, the green strip reaches the top of the block but the red strip _doesn't quite_, there's a little space:

image

elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: I figured out what was causing the crash (the problem that I was using it on has feedback=none, so the 'badge_type' was never set), so I was able to test. 

It works!

elements/pl-order-blocks/pl-order-blocks.mustache
SethPoulsen: oh wait, you might still have a problem with it showing inline in the case of `format=code`, because of the fact that `pl-code` always get translated into a `div`. If there _is_ a `pl-code` specific hack that is needed, it should probably go over here: 
https://github.com/PrairieLearn/PrairieLearn/blob/00fe7665ac9c16ca94d650c5b48f2805af289447/elements/pl-order-blocks/pl-order-blocks.css#L19-L22
because that's where we already have one such hack.

CoffeePoweredComputers:
@SethPoulsen 
image
image
After some tinkering, here's what the feedback badges look like now for the default and code (with python) options, respectively.

elements/pl-order-blocks/pl-order-blocks.css
```diff
@@ -14,8 +14,18 @@ div.pl-order-blocks-bottom {
...
+  padding: 0px !important;
 }
 
+li.pl-order-blocks-code .badge{
+    margin: 5px !important;
+}
+
+li.pl-order-blocks-code .mb-2{
+  margin-bottom: 0px !important;
+}
```
SethPoulsen, line 26: instead of adding this, why not just edit the rule right below this one so that its `:last_child` instead of `:only_child`?

elements/pl-order-blocks/pl-order-blocks.css
```diff
@@ -14,8 +14,18 @@ div.pl-order-blocks-bottom {
 
 li.pl-order-blocks-code {
   list-style-type: none;
+  padding: 0px !important;
```
SethPoulsen, line 17: why is this needed?

elements/pl-order-blocks/pl-order-blocks.css
```diff
@@ -14,8 +14,18 @@ div.pl-order-blocks-bottom {
...
 }
 
+li.pl-order-blocks-code .badge{
+    margin: 5px !important;
+}
```
SethPoulsen, line 22: Seems odd to need to do this. How are these badges being styled by the other elements? Is there some default badge style somewhere we can just refer to?

elements/pl-order-blocks/pl-order-blocks.css
```diff
@@ -52,13 +62,22 @@ ul.ui-sortable {
...
+    width: 20px;
+    text-align: center;
+}
+
 ul.ui-sortable li, ul.answer_table li, ul.submission_table li {
   margin-bottom: 5px;
   border: 1px solid grey !important;
   border-radius: 5px !important;
   line-height: 1.2em;
+  padding: 0.4em;
```
SethPoulsen, line 77: love the change, but think its a little bit small for me, maybe `0.6em` is good?

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -118,16 +122,24 @@ def prepare_tag(html_tags, index, group_info={'tag': None, 'depends': None}):
...
         inner_html = pl.inner_html(html_tags)
         ranking = pl.get_integer_attrib(html_tags, 'ranking', -1)
+        distractor_info = pl.get_string_attrib(element, 'distractor-info', DISTRACTOR_INFO_DEFAULT)
+        distractor_feedback = pl.get_string_attrib(html_tags, 'distractor-feedback', None)
+
+        if (distractor_info == 'none') and (distractor_feedback is not None):
+            raise Exception("distractor-info must be either 'show-distractors' or 'show-feedback' for the distractor_feedback tag to be used in <pl-answer>")
+
+        if (distractor_info == 'show-feedback') and (distractor_feedback is None) and (not is_correct):
+            raise Exception("All distractors must have 'distractor-feedback' tag populated if 'show-feedback' option is used.")
```
SethPoulsen, line 142: I can't see any reason why to have this restriction. Why can't an instructor put explanations on some of the distractors but not others?

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -118,16 +122,24 @@ def prepare_tag(html_tags, index, group_info={'tag': None, 'depends': None}):
...
 
         is_correct = pl.get_boolean_attrib(html_tags, 'correct', PL_ANSWER_CORRECT_DEFAULT)
         answer_indent = pl.get_integer_attrib(html_tags, 'indent', None)
         inner_html = pl.inner_html(html_tags)
         ranking = pl.get_integer_attrib(html_tags, 'ranking', -1)
+        distractor_info = pl.get_string_attrib(element, 'distractor-info', DISTRACTOR_INFO_DEFAULT)
+        distractor_feedback = pl.get_string_attrib(html_tags, 'distractor-feedback', None)
+
+        if (distractor_info == 'none') and (distractor_feedback is not None):
+            raise Exception("distractor-info must be either 'show-distractors' or 'show-feedback' for the distractor_feedback tag to be used in <pl-answer>")
```
SethPoulsen, line 139: `distractor_feedback` -> `distractor-feedback`, also change the message to talk about the new `distractor-info` values.

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -345,22 +366,45 @@ def render(element_html, data):
...
+                    distractor['margin'] = '0px'
+                elif distractor_info == 'show-feedback':
+                    distractor['margin'] = '10px'
+                else:
+                    raise Exception('The distractor info tag must be show-distractors, show-feedback or none')
 
+        else:
+            all_distractors = []
+
+        if answer_name in data['correct_answers']:
```
SethPoulsen, line 386: do we need this if check here? Even if there is a possibility that the key is not in the dictionary, doing a call to `get` with a default value is a nicer solution than having this whole `if` statement.

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -345,22 +366,45 @@ def render(element_html, data):
...
+
+            html_params = {
+                'true_answer': True,
+                'question_solution': question_solution,
+                'grading_mode': grading_method,
+                'indentation_message': indentation_message,
+                'distractor-feedback': all_distractors,
+                'show-distractors': distractor_info in ['show-feedback', 'show-distractors'],
+                'show-feedback': distractor_info == 'show-feedback',
+                'block_formatting': block_formatting
```
SethPoulsen, line 400: Variable names should be snake_case. I know its confusing because HTML property names and values are kebab-case.

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -294,7 +312,10 @@ def render(element_html, data):
         if answer_name in data['submitted_answers']:
             student_submission = [{
                 'inner_html': attempt['inner_html'],
-                'indent': ((attempt['indent'] or 0) * TAB_SIZE_PX) + INDENT_OFFSET
+                'indent': ((attempt['indent'] or 0) * TAB_SIZE_PX) + INDENT_OFFSET,
+                'badge_type': attempt['badge_type'] if (feedback_type == 'first-wrong') and 'badge_type' in attempt else '',
+                'icon': attempt['icon'] if (feedback_type == 'first-wrong') and ('icon' in attempt) else '',
+                'first-wrong': feedback_type == 'first-wrong'
```
SethPoulsen, line 318: Can we somehow have just one variable here instead of three new ones? Maybe actually construct the HTML for the badge right here, and then if there shouldn't be a badge, just use the empty string? Or some other way of making it cleaner?

elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -62,7 +62,17 @@ $(function() {
         </ul>
         <ul class="list-group submission_table">
             {{#student_submission}}
-            	<li class="{{block_formatting}}" style="margin-left: {{indent}}px">{{{inner_html}}}</li>
+            	<li class="{{block_formatting}}" style="margin-left: {{indent}}px;">
+                    {{#first-wrong}}
+                    <div class="feedback_badge_container">
+                        <span class="badge {{badge_type}} "><i class="fa {{icon}}" aria-hidden="true"></i></span>
+                        <span>{{{inner_html}}}</span>
+                    </div>
```
SethPoulsen, line 70: Is the whole badge container thing necessary? why can't you just drop in the badge in front of the other html? Also, I don't think the name "badge container" makes sense when the content is also going in the container.

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -345,22 +366,45 @@ def render(element_html, data):
...
+            for distractor in all_distractors:
+                if (distractor['distractor-feedback'] is None) and (distractor_info == 'show-feedback'):
+                    raise Exception('Feedback must be associated with each distractor to use the show-feedback option')
+
+                if distractor_info == 'show-distractors' or distractor_info == 'none':
+                    distractor['margin'] = '0px'
+                elif distractor_info == 'show-feedback':
+                    distractor['margin'] = '10px'
+                else:
+                    raise Exception('The distractor info tag must be show-distractors, show-feedback or none')
```
SethPoulsen, line 381: Aren't we doing this differently now? 

Also, this check should only need to be happening in the `prepare` phase, and I think you already have it there, so you don't need to also check it here.

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -345,22 +366,45 @@ def render(element_html, data):
...
-        }
-        with open('pl-order-blocks.mustache', 'r', encoding='utf-8') as f:
-            html = chevron.render(f, html_params)
-        return html
+        distractor_info = pl.get_string_attrib(element, 'distractor-info', DISTRACTOR_INFO_DEFAULT)
+        if distractor_info != 'none':
+            all_distractors = [item for item in data['params'][answer_name] if not item['is_correct']]
+            for distractor in all_distractors:
+                if (distractor['distractor-feedback'] is None) and (distractor_info == 'show-feedback'):
+                    raise Exception('Feedback must be associated with each distractor to use the show-feedback option')
```
SethPoulsen, line 374: this check not needed

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -478,6 +525,20 @@ def grade(element_html, data):
         num_initial_correct, true_answer_length = grade_dag(submission, depends_graph, group_belonging)
         first_wrong = -1 if num_initial_correct == len(submission) else num_initial_correct
 
+        if feedback_type == 'first-wrong':
+            for i, block in enumerate(student_answer):
+                if i >= num_initial_correct: break
```
SethPoulsen, line 530: minor: use a list slice here to control the for loop instead of a break, just like you did a few lines below this.

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -490,9 +551,12 @@ def grade(element_html, data):
         if final_score < 1:
             if feedback_type == 'none':
                 feedback = ''
-            elif feedback_type == 'first-wrong':
+            elif feedback_type.startswith('first-wrong'):
                 if first_wrong == -1:
                     feedback = FIRST_WRONG_FEEDBACK['incomplete']
+                elif feedback_type.endswith('-verbose') and student_answer[first_wrong]['inner_html'] in distractor_feedback:
+                    text = student_answer[first_wrong]['inner_html']
+                    feedback += FIRST_WRONG_FEEDBACK['distractor-feedback'].format(str(first_wrong + 1), text, distractor_feedback[text] or '')
```
SethPoulsen, line 559: we're not doing it like this anymore, right? we are just going to have `feedback='first-wrong'` and then `distractor-info` be separate things?

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -478,6 +525,20 @@ def grade(element_html, data):
...
+                block["badge_type"] = "badge-success"
+                block["icon"] = "fa-check"
+
+            if first_wrong != -1:
+                student_answer[first_wrong]["badge_type"] = "badge-danger"
+                student_answer[first_wrong]["icon"] = "fa-xmark"
+
+                for block in student_answer[first_wrong + 1: ]:
+                    block["badge_type"] = ""
+                    block["icon"] = ""
```
SethPoulsen, line 540: shouldn't this for loop be indented 1 less time than it is?

SethPoulsen (commented):
Love the progress. I think most of the functionality is all good, I'm mostly being picky about code style at this point.

elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen:
```
for block in student_answer[:first_wrong + 1]:
```

SethPoulsen, line 530:
```suggestion
            for block in student_answer[:first_wrong + 1]:
```

SethPoulsen: oooh I forgot I can just directly suggest an edit. let me do that.

elements/pl-order-blocks/pl-order-blocks.css
SethPoulsen: or just get rid of it, I guess

elements/pl-order-blocks/pl-order-blocks.css
```diff
@@ -14,8 +14,18 @@ div.pl-order-blocks-bottom {
...
+li.pl-order-blocks-code .badge{
+    margin: 5px !important;
+}
+
+li.pl-order-blocks-code .mb-2{
+  margin-bottom: 0px !important;
+}
+
+
 .pl-order-blocks-code > .pl-code:only-child div {
```
SethPoulsen, line 29:
```suggestion
.pl-order-blocks-code > .pl-code div {
```

elements/pl-order-blocks/pl-order-blocks.py
CoffeePoweredComputers: It could be, but if `first_wrong` is -1then doesn't that mean that the solution so far is correct? The body of this if statement adds an X badge to the incorrect block and then populates any remaining blocks with blank badges.

elements/pl-order-blocks/pl-order-blocks.css
CoffeePoweredComputers: Yeah, that seems good. The reasoning for 0.4 was it kept the default and code styled blocks about the same width.

elements/pl-order-blocks/pl-order-blocks.css
CoffeePoweredComputers: When the format is changed to `code` for some reason the formatting of the badge gets strange. That line has since been changed to the following, just to cut down on whitespace:
```css
margin-left: 5px !important;
margin-right: 0px !important;
```
If those are excluded we get badges that look like this when the formatting is included as `code`:
image

CoffeePoweredComputers: When they're added back it looks like this:
image

elements/pl-order-blocks/pl-order-blocks.css
CoffeePoweredComputers: Yeah, I was trying things along those lines to get the rule below it to impact the submission panel but I couldn't get it working. Removing the one below has the same impact so I might go ahead and remove it in favor of this rule if that's okay.

elements/pl-order-blocks/pl-order-blocks.css
SethPoulsen: so odd. Any ideas why other elements dont have this problem?

elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: oh wait, this also has problems when first_wrong is -1. I stil think there should be a more elegant solution than the break though...maybe not

elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: ah, my bad I thought that this for loop was under the other for loop. Angry face at python's indentation-based scoping.

elements/pl-order-blocks/pl-order-blocks.py
CoffeePoweredComputers: Fixed in latest push.

elements/pl-order-blocks/pl-order-blocks.mustache
CoffeePoweredComputers: Using the `d-inline-block` rule, which is how the badge is placed elsewhere works for the default styling but not for code styling. The name was just a placeholder so I'm happy to replace it with something else but the css associated with that rule was the only way I was able to get the badge placement right for both default and code format: 
image
image

CoffeePoweredComputers: The above is what happens both when using the existing d-inline-block and when removing the div and just dropping the badge in front.

elements/pl-order-blocks/pl-order-blocks.py
CoffeePoweredComputers: Removed.

elements/pl-order-blocks/pl-order-blocks.py
CoffeePoweredComputers: Yeah, that's the issue was was running into so I opted for this. I'll mull it over and see if I can think of a simpler solution.

elements/pl-order-blocks/pl-order-blocks.py
CoffeePoweredComputers: Yeah, I still need to separate them. Should have that updated in the next round of changes.

elements/pl-order-blocks/pl-order-blocks.css
CoffeePoweredComputers: Not sure. :( I'm afraid I don't have enough experience with the other elements to say. Is there an example you had in mind of where a badge is placed inline with a block of pl-code I could take a look at?

elements/pl-order-blocks/pl-order-blocks.css
CoffeePoweredComputers: The padding changes that made the default format option need to overridden for blocks containing pl-code so they don't get bigger.

elements/pl-order-blocks/pl-order-blocks.css
SethPoulsen: most of them do it, just start with looking at a simple one like pl-integer-input and see what you learn

elements/pl-order-blocks/pl-order-blocks.py
CoffeePoweredComputers: Fixed.

elements/pl-order-blocks/pl-order-blocks.py
CoffeePoweredComputers: I had some redundant checks here so the updated code should be a bit cleaner. I could construct the HTML here but I feel like that would be more busy than keeping the HTML template in the mustache file and just setting an additional attribute here.

SethPoulsen:
This is not necessarily a blocking issue, because it could be fixed in a later PR, but the way that the distractor feedback is being formatted is breaking LaTeX/MathJax layout:
image

elements/pl-order-blocks/pl-order-blocks.py
CoffeePoweredComputers: Sorry, for this one, are you referring to the if on line 386 or the if-else blocks above it?

elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -88,5 +98,27 @@ $(function() {
...
+        <li class="list-group-item list-group-item-danger"><strong>
+        Incorrect Blocks (these blocks should not be included in your solution):</strong></li>
+        </ul>
+        <ul class="list-group answer_table" style="list-style-type:none;">
+        {{#distractor-feedback}}
+            <li> 
+                <div class="{{block_formatting}}">{{{inner_html}}}</div> 
+                {{#has-distractor-feedback}}
+                <div style="margin: 10px" > 
+                    <div style="display: flex; align-items: center;">
```
SethPoulsen, line 113: get rid of the style here to fix the issue noted in my other comment (https://github.com/PrairieLearn/PrairieLearn/pull/5667#issuecomment-1239756742)

SethPoulsen, line 113:
```suggestion
                    <div>
```

elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: Ya, line 386.

SethPoulsen:
I just realized that having the `distractor-feedback` as a property means that you can't write arbitrary HTML inside it. Say for example you wanted to have some formatting inside the feedback that you give, like: 

```html
<pl-answer correct="false" distractor-feedback="This answer is wrong for reason 1 <i>and</i> for reason 2.">Some really wrong answer</pl-answer>
```
well, you can't do that because its invalid HTML. We also probably shouldn't use `&lt` and `&gt` for the tags because those should actually show up as those actual characters on the page. 

An alternative would be to have the distractor feedback in a separate tag, but that's a bit cumbersome.
```html
<pl-answer correct="false" >Some really wrong answer
  <pl-distractor-feedback>This answer is wrong for reason 1 <i>and</i> for reason 2.</pl-distractor-feedback>
</pl-answer>
```
or else
```html
<pl-answer tag="wrong1" correct="false" >Some really wrong answer</pl-answer>
<pl-distractor-feedback feedback-for>This answer is wrong for reason 1 <i>and</i> for reason 2.</pl-distractor-feedback>
```

I don't know what the 'right' or 'best solution is here, I just realized this. The best solution may be to just leave it as it is for now and worry about it later if people start wanting to use more complex formatting.

PR #5668, opened 2022-04-25
Fix handling of failed docker pulls in grader_host

mwest1066:
This removes the [IIFE](https://flaviocopes.com/javascript-iife/) and simplifies the error handling. The `callbackCalled` guard should not be needed but is there for redundancy.

### DIFF
--- a/grader_host/lib/pullImages.js
+++ b/grader_host/lib/pullImages.js
@@ -48,41 +48,52 @@ module.exports = function (callback) {
           images,
           config.parallelInitPulls,
           (image, callback) => {
-            ((callback) => {
-              var ourAuth = {};
-              logger.info(
-                `Pulling latest version of "${image}" image from ${
-                  config.cacheImageRegistry || 'default registry'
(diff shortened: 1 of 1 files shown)


grader_host/lib/pullImages.js
```diff
@@ -48,41 +48,52 @@ module.exports = function (callback) {
...
+                (err) => {
+                  if (err) {
+                    // Just log the error, don't pass it to callback() because
+                    // we want to keep going.
+                    logger.error(`Error pulling "${image}" (followProgress)`, err);
+                  }
+                  if (!callbackCalled) {
+                    callbackCalled = true;
                     callback(null);
-                  },
```
nicknytko, line 76: do we want to print any debug info (e.g. a stacktrace?) if the callback has already been called?

mwest1066: Good idea! But I'm going to close this PR and just file an issue for now.

PR #5670, opened 2022-04-25
Add progress bar to manual grading page

jonatanschroeder:
As I'm following the progress of our TAs marking an assessment, it is easier to follow if there is some better visual indication of how many questions still have pending marking. This PR adds a progress bar to the page.


jonatanschroeder:
Minor issue, though: the label is for submissions that require grading, while the progress bar shows submissions that are graded. Textually I believe this is clear enough, but I'm wondering if this could cause some initial visual confusion. @nwalters512 thoughts?

nwalters512 (approved):
> Minor issue, though: the label is for submissions that require grading, while the progress bar shows submissions that are graded. Textually I believe this is clear enough, but I'm wondering if this could cause some initial visual confusion.

I think this is fine! I vaguely remember this "problem" existing in PT and the scheduler too, and I don't think anyone's complained much. It's cheap to change in the future if folks complain.

jonatanschroeder:
> > Minor issue, though: the label is for submissions that require grading, while the progress bar shows submissions that are graded. Textually I believe this is clear enough, but I'm wondering if this could cause some initial visual confusion.
> 
> I think this is fine! I vaguely remember this "problem" existing in PT and the scheduler too, and I don't think anyone's complained much. It's cheap to change in the future if folks complain.

This was actually fixed in another change after this comment, by keeping a light background in the progress bar and adding a remaining bar to the right with a darker background, which made this clearer.

Issue #5669, opened 2022-04-25
Incorrect error handling in pullImages for grader hosts


mwest1066:
During pre-loading of images on grader hosts, if we encounter an error then there is something wrong with the handling of it in the callback stack. An example error trace is:
```
info: Querying for recent images
info: Need to pull 7 images
/PrairieLearn/node_modules/async/dist/async.js:324
if (fn === null) throw new Error("Callback was already called.");
^
Error: Callback was already called.
at /PrairieLearn/node_modules/async/dist/async.js:324:36
at /PrairieLearn/grader_host/lib/pullImages.js:85:15
at module.exports (/PrairieLearn/node_modules/async-stacktrace/ERR.js:52:7)
at /PrairieLearn/grader_host/lib/pullImages.js:69:21
at /PrairieLearn/node_modules/dockerode/lib/docker.js:118:7
at Modem.buildPayload (/PrairieLearn/node_modules/docker-modem/lib/modem.js:320:19)
at ClientRequest.<anonymous> (/PrairieLearn/node_modules/docker-modem/lib/modem.js:302:10)
at ClientRequest.emit (events.js:400:28)
at ClientRequest.emit (domain.js:475:12)
at Socket.socketErrorListener (_http_client.js:475:9)
at Socket.emit (events.js:400:28)
at Socket.emit (domain.js:475:12)
at emitErrorNT (internal/streams/destroy.js:106:8)
at emitErrorCloseNT (internal/streams/destroy.js:74:3)
at processTicksAndRejections (internal/process/task_queues.js:82:21)
```

mwest1066:
I had a brief attempt at simplifying the callback logic in #5668 but I wasn't confident that it would actually fix anything.

mwest1066:
Here's another similar stack trace that we just had, but not in different code:
```
info: Received job!
info: received [REDACTED] from queue
info: Logging job [REDACTED] to S3: [REDACTED]
/PrairieLearn/node_modules/async/dist/async.js:324
if (fn === null) throw new Error("Callback was already called.");
^
Error: Callback was already called.
at /PrairieLearn/node_modules/async/dist/async.js:324:36
at /PrairieLearn/node_modules/async/dist/async.js:3674:17
at module.exports (/PrairieLearn/node_modules/async-stacktrace/ERR.js:52:7)
at /PrairieLearn/grader_host/index.js:274:20
at /PrairieLearn/node_modules/dockerode/lib/docker.js:118:7
[...]

PR #5676, opened 2022-04-26
varying timeouts on dockerode (docker-modem) for pullImages

echuber2:
Trying to diagnose #5669


echuber2:
@nwalters512 mentioned this didn't help. It seems the docker-modem library inside dockerode is using the callback for multiple https responses.

PR #5678, opened 2022-04-26
fix: Ensure createImage callback only executes once

nwalters512:
This is a bandaid for #5669 that should hopefully allow us to get things operational for students and instructors again while we work on identifying the root cause and a better fix.

Originally, I did this with `_.once`, but ultimately realized that it was more robust to just refactor this to use promises. Promises have nice semantics where they can only transition from pending to fulfilled exactly once, so that prevents any of these kind of errors from occurring.

### DIFF
--- a/grader_host/index.js
+++ b/grader_host/index.js
@@ -8,7 +8,6 @@ const AWS = require('aws-sdk');
 const { exec } = require('child_process');
 const path = require('path');
 const byline = require('byline');
-const _ = require('lodash');
 const sqldb = require('../prairielib/lib/sql-db');
 const sanitizeObject = require('../prairielib/lib/util').sanitizeObject;
 
@@ -255,7 +254,7 @@ function initDocker(info, callback) {
           callback(null);
         }
       },
-      (callback) => {
+      async () => {
         logger.info(`Pulling latest version of "${image}" image`);
         var repository = new dockerUtil.DockerName(image);
         if (config.cacheImageRegistry) {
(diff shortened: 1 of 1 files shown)


mwest1066:
And just to document from the ES2015 spec:
> Attempting to resolve or reject a resolved promise has no effect.

(see https://262.ecma-international.org/6.0/#sec-promise-objects)

mwest1066 (approved):

echuber2:
Given the weird behavior of docker-modem, what will happen if the Promise is rejected before it is (i.e., would have been) resolved? Will the job retry automatically?

nwalters512:
We always retry jobs that fail for any reason, up to three total executions.

echuber2:
If rejections preempt resolutions too many times, it may be helpful to manipulate the rejection case to give extra tries in that particular case.

nwalters512:
Can you explain your thinking there? It's not obvious to me why we'd want to treat different kinds of failures differently.

FWIW, what we were seeing was a successful response from the Docker daemon following by a connection reset (an error). That would be a resolution preempting an error that didn't actually matter to us. We shouldn't ever expect to see an HTTP request that errors but then later resolves with a successful response.

echuber2:
Since we don't know why docker-modem was behaving as it did, I thought the two responses might be received in the opposite order at some point (unpredictably), and this could happen repeatedly. We can cross that road when we come to it. The reason for handling it specially would be that the error is a buggy side-effect of a library that is preferentially ignored in this case rather than letting it exhaust (potentially) three retries and fail hard.

But I mean, ideally they can just fix docker-modem and no other special casing is needed.

echuber2:
(This is kind of similar to the fix here in spirit: the bug's not in our code but I suggested wrapping to make sure it happens only once. We shouldn't have to do this either.)

PR #5683, opened 2022-04-28
Ensure remove-spaces on pl-string-input removes all whitespace

ffund:
The `remove-leading-training` option removes all whitespace from the beginning and end of a string, but `remove-spaces` only removed the space character. Other whitespace (tabs, for example) were not removed.

This PR updates `remove-spaces` to remove other whitespace as well.

### DIFF
--- a/elements/pl-string-input/pl-string-input.py
+++ b/elements/pl-string-input/pl-string-input.py
@@ -243,8 +243,8 @@ def grade(element_html, data):
 
     # Remove the blank spaces between characters
     if (remove_spaces):
-        a_sub = "".join(a_sub.split())
-        a_tru = "".join(a_tru.split())
+        a_sub = ''.join(a_sub.split())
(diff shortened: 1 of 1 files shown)


ffund:

elements/pl-string-input/pl-string-input.py
```diff
@@ -243,8 +243,8 @@ def grade(element_html, data):
 
     # Remove the blank spaces between characters
     if (remove_spaces):
-        a_sub = a_sub.replace(' ', '')
-        a_tru = a_tru.replace(' ', '')
+        a_sub = ''.join(a_sub.split())
+        a_tru = ''.join(a_tru.split())
```
nwalters512, line 247: This is consistent with the usage of `strip()` above for `remove-leading-trailing`, but in the future we might consider using a regex with `\s` instead, which will handler a wider variety of whitespace characters. Not requesting a change here though!

PR #5687, opened 2022-04-29
fix regression for LTI grading

### LINKED ISSUE #5671: LTI grades only being sent by individual regrade (Coursera)
The CS 445 staff flagged this issue and I've confirmed that with LTI items linked from Coursera, the grades are currently only being sent back if we select Regrade for student instance records *one at a time*. Attempting to close+grade or expire/reopen/re-close+grade all of the assignments doesn't work. This is different from the behavior in past semesters where LTI grades went back as soon as the student finished the exam item.

Is this a bug or an intentional change in behavior? If we don't attempt to manually regrade items, will they all be transmitted in bulk automatically after some period of time? Would it be possible to add the transmission trait to the "grade all instances" action?

Tagging @trombonekenny for his expertise on LTI.

trombonekenny:
Fixes #5671 from @nwalters512 troubleshooting

### DIFF
--- a/lib/question.js
+++ b/lib/question.js
@@ -624,7 +624,7 @@ module.exports = {
             data.score,
             data.v2_score,
           ];
-          sqldb.callOneRow('grading_jobs_update_after_grading', params, (err, result) => {
+          sqldb.callOneRow('grading_jobs_insert', params, (err, result) => {
             if (ERR(err, callback)) return;
 
             // If the submission was marked invalid during grading the grading
--- a/lib/question.js
+++ b/lib/question.js
@@ -624,7 +624,7 @@ module.exports = {
             data.score,
             data.v2_score,
           ];
-          sqldb.callOneRow('grading_jobs_insert', params, (err, result) => {
+          sqldb.callOneRow('grading_jobs_update_after_grading', params, (err, result) => {
             if (ERR(err, callback)) return;
 
             // If the submission was marked invalid during grading the grading
--- a/sprocs/grading_jobs_update_after_grading.sql
+++ b/sprocs/grading_jobs_update_after_grading.sql
@@ -1,20 +1,21 @@
 CREATE FUNCTION
     grading_jobs_update_after_grading(
-        grading_job_id bigint,
-        received_time timestamptz,
-        start_time timestamptz,
-        finish_time timestamptz,
-        new_submitted_answer jsonb, -- NULL => no change
-        new_format_errors jsonb,
-        new_gradable boolean,
-        new_broken boolean,
-        new_params jsonb, -- NULL => no change
-        new_true_answer jsonb, -- NULL => no change
-        new_feedback jsonb,
-        new_partial_scores jsonb,
-        new_score double precision,
-        new_v2_score double precision
-    ) RETURNS grading_job grading_jobs
+        IN grading_job_id bigint,
+        IN received_time timestamptz,
+        IN start_time timestamptz,
+        IN finish_time timestamptz,
+        IN new_submitted_answer jsonb, -- NULL => no change
+        IN new_format_errors jsonb,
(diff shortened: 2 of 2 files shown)


nwalters512 (changes requested):
The problem is not that we're calling the wrong sproc, it's that the new sproc doesn't return the same information that the old sproc did (the grading job).

trombonekenny:
Yeah, I'm learning it's not a one liner. :)

I'm also learning I don't have the knowledge or setup to `make test` locally because things are failing there too.

So maybe this isn't a good PR for me.

trombonekenny:
Or maybe it can be.... but how do we really test it?

trombonekenny:
I'm not sure how to re-request a review but I changed up the sproc and local and GH tests are passing.

nwalters512:
> I'm not sure how to re-request a review

I'll take a look right now! For future reference, there should be a little arrows-in-a-circle icon next to my name that you can click:

Screen Shot 2022-04-29 at 13 38 39

trombonekenny:
image

nwalters512:
@trombonekenny ahhh, you already clicked that button 😛 

Screen Shot 2022-04-29 at 13 42 39

nwalters512:
I took a look at this locally, and I think the `grading_job` in the `DECLARE` block is shadowing the `OUT` arg. I tested locally by adding a `console.log()` after this line:

https://github.com/PrairieLearn/PrairieLearn/blob/529f6dffb1b5b79d1d0ebf777a99741eab8f6aa7/lib/question.js#L633

I then graded a question. In the logs, I saw a `grading_jobs` row with all null values. Deleting `grading_job` from the `DECLARE` block produced a `grading_job` with the expected values.

trombonekenny:
haha Oops. Well, this time I mean it! :-D

echuber2:
I think we need a reliable way to test this. I wish we could have that in CI too.

trombonekenny:
> I took a look at this locally, and I think the `grading_job` in the `DECLARE` block is shadowing the `OUT` arg. I tested locally by adding a `console.log()` after this line:

I replicated that locally and pushed a fix. With the console logging in question.js I saw the grading object returned and the next stanza with ltiOutcomes was run.

We could implement more testing (always), but I think this bugfix works.

nwalters512 (approved):

echuber2:
Is it possible this will be deployed before finals?

mwest1066:
> Is it possible this will be deployed before finals?

I'm normally extremely conservative about deploying new code shortly before finals. But we could get this deployed if it will make a big difference to you?

echuber2:
@mwest1066 I think for this to help with CS 445's final exam it would need to be up soon this week, although that course has a workaround for transferring grades manually if LTI just won't work out this time.

However, summer courses starting as early as May 9 (soft launch) or May 16 may want to use LTI for homework, so hopefully it will be deployed in time for May 16 or May 23 at least.

echuber2:
@mwest1066 Update: CS 445 will skip LTI for this final so there's no need to rush this, although the summer courses might use it.

nwalters512:
Sounds great - we'll get this merged and deployed once finals are done!

echuber2:
Once this is deployed, will the "grade all instances" feature work to push any outstanding grades to the LTI consumer? (I'm referring to Assessment > Students > Action for all instances > Grade all instances.)

trombonekenny:
> Once this is deployed, will the "grade all instances" feature work to push any outstanding grades to the LTI consumer? (I'm referring to Assessment > Students > Action for all instances > Grade all instances.)

I believe so, yes. Tracing the functions, it eventually runs https://github.com/PrairieLearn/PrairieLearn/blob/57b6672d2a19a603692f2dd4312f028fcc7a7185/lib/question.js#L647

PR #5751, opened 2022-05-03
convert start_postgres.sh to honor PGDATA

### LINKED ISSUE #6329: Workspace runs before migrations, may start as unhealthy
When starting a Docker container, it starts two different servers in parallel: the workspace host, and the regular PL server. When there are pending migrations, the PL server will identify those migrations and run them. However, depending on which migrations and code exist, (e.g., if there are sprocs that use tables that were created by a migration), the workspace host will start before the migration runs, so it may fail to run as expected, and will start in an unhealthy state. This is the output that I get when running PL with branch `mg-rubrics` (#6271) while the Docker image is on the current `master` branch:

```
Starting PrairieLearn...
yarn run v1.22.19
$ /PrairieLearn/node_modules/.bin/nodemon -L server.js
[nodemon] 2.0.19
[nodemon] to restart at any time, enter `rs`
[nodemon] watching path(s): *.*
[nodemon] watching extensions: js,json,sql
[nodemon] starting `node server.js`
error: Error initializing workspace host: type "rubrics" does not exist {"code":"42704","data":{"sql":"CREATE FUNCTION\n    rubric_gradings_insert(\n        IN rubric_id BIGINT,\n        IN adjust_points DOUBLE PRECISION,\n        IN applied_rubric_items JSONB,\n        INOUT computed_points DOUBLE PRECISION,\n        OUT rubric_grading_id BIGINT\n    )\nAS $$\nDECLARE\n    rubric rubrics;\nBEGIN\n    SELECT * INTO rubric FROM rubrics AS r WHERE r.id = rubric_id;\n\n    IF NOT FOUND THEN\n        RAISE EXCEPTION 'Could not find rubric %d.', rubric_id;\n    END IF;\n\n    IF computed_points IS NULL THEN\n        SELECT LEAST(GREATEST(rubric.starting_points + SUM(COALESCE(ari.points, ri.points)) + COALESCE(adjust_points, 0),\n                              rubric.min_points), rubric.max_points)\n        INTO computed_points\n        FROM JSONB_TO_RECORDSET(applied_rubric_items) AS ari(rubric_item_id BIGINT, points DOUBLE PRECISION, note TEXT)\n             JOIN rubric_items AS ri ON (ri.id = ari.rubric_item_id);\n    END IF;\n\n    INSERT INTO rubric_gradings\n        (rubric_id, computed_points, adjust_points)\n    VALUES\n        (rubric_id, computed_points, adjust_points)\n    RETURNING id INTO rubric_grading_id;\n\n    INSERT INTO rubric_grading_items\n        (rubric_grading_id, rubric_item_id, points, note)\n    SELECT\n        rubric_grading_id, ari.rubric_item_id, COALESCE(ari.points, ri.points), ari.note\n    FROM JSONB_TO_RECORDSET(applied_rubric_items) AS ari(rubric_item_id BIGINT, points DOUBLE PRECISION, note TEXT)\n         JOIN rubric_items AS ri ON (ri.id = ari.rubric_item_id);\nEND;\n$$ LANGUAGE plpgsql VOLATILE;\n","sqlError":{"code":"42704","file":"parse_type.c","length":95,"line":"275","message":"type \"rubrics\" does not exist","name":"error","position":"272","routine":"typenameType","severity":"ERROR"},"sqlFile":"rubric_gradings_insert.sql","sqlParams":[]},"file":"parse_type.c","length":95,"line":"275","name":"error","position":"272","routine":"typenameType","severity":"ERROR","stack":"Async Stacktrace:\n    at /PrairieLearn/workspace_host/interface.js:203:13\n    at /PrairieLearn/sprocs/index.js:183:11\n    at PostgresPool.<anonymous> (/PrairieLearn/sprocs/index.js:177:15)\n\nerror: type \"rubrics\" does not exist\n    at Parser.parseErrorMessage (/PrairieLearn/node_modules/pg-protocol/dist/parser.js:287:98)\n    at Parser.handlePacket (/PrairieLearn/node_modules/pg-protocol/dist/parser.js:126:29)\n    at Parser.parse (/PrairieLearn/node_modules/pg-protocol/dist/parser.js:39:38)\n    at Socket.<anonymous> (/PrairieLearn/node_modules/pg-protocol/dist/index.js:11:42)\n    at Socket.emit (node:events:513:28)\n    at Socket.emit (node:domain:489:12)\n    at addChunk (node:internal/streams/readable:315:12)\n    at readableAddChunk (node:internal/streams/readable:289:9)\n    at Socket.Readable.push (node:internal/streams/readable:228:10)\n    at TCP.onStreamRead (node:internal/stream_base_commons:190:23)"}
warn: Marked self as unhealthy: error: type "rubrics" does not exist
info: Running migration 20220905173159__rubrics__create.sql
info: PrairieLearn server ready, press Control-C to quit
info: Go to http://localhost:3000
```
This causes the local instance to have issues with workspaces, which are unable to run. It will also cause some weird issues like the terminal eventually printing messages that `_pushAllRunningContainersToS3` isn't running, so the small timer that logs this error is active, but the  timer that actually pushes the containers is not.

I realize this is a pretty niche issue, but just reporting in case there is some way to address it.

trombonekenny:
Makes `docker/start_postgres.sh` honor the PGDATA environment variable to change the location of the PG database in docker.  Defaults to `/var/postgres` if not set, so it's backwards compatible.

Adds `docker/start_postgres.sh init` as a manual call to bootstrap a postgres database at the PGDATA location, with the commands taken from various parts of the Dockerfiles. 

I use it like this (with a RAM disk and environment variable set in the docker launch):

```
docker --rm -it run -p 3000:3000 --tmpfs=/pgtmpfs -e PGDATA=/pgtmpfs prairielearn/prairielearn /bin/bash
# /PrairieLearn/docker/start_postgres.sh init
# time make test
```
On the same hardware, the full `make test` suite takes me 5 minutes 38 seconds. 
Pointing the PL database to a RAM disk makes it finish in 1 minute 43 seconds.

Also resolves #6329


docker/start_postgres.sh
```diff
@@ -13,12 +15,25 @@ if [[ "$ACTION" == "start" ]]; then
     fi
 fi
 
+# Bootstrap the PGDATA folder
+if [[ "$ACTION" == "init" ]]; then
```
jonatanschroeder, line 19: Would it make sense to, instead of testing `init`, have the script check if `PGDATA` is provided and (if the action is `start`) initialize that instead? This way no additional manual init is required, and a simple change in the invocation of `docker run` is sufficient.

trombonekenny: If PGDATA is provided, `init` isn't needed. Perhaps you mount in a persistent volume that's already initialized, so it wouldn't need that setup step. The `init` is just a helper to bootstrap an empty folder (at $PGDATA) to make getting started easier.

trombonekenny: It felt a little creepy making it somehow detect if PGDATA was ready-to-roll and if not, automatically do it. I don't think I could anticipate all of the failure cases?

My workflow with PL dev is usually to run a shell in the container anyway and not the launch scripts. Since then I need to run `/PrairieLearn/docker/start_postgres.sh` first anyway, making the `init` command separate seemed reasonable.

jonatanschroeder: That actually makes sense. Long shot: what about running `pg_ctl status` and checking the status code? If the status code is 4, the directory is not valid and needs to be initialized (https://www.postgresql.org/docs/current/app-pg-ctl.html).

trombonekenny: I can noodle on that.

trombonekenny: OK, I have a version of the script that automatically initializes the database sbased on `pg_ctl status` as you described. I think I'm running into a new race condition with workspaces bootstrapping. I'll post about that in a new comment.

trombonekenny:
OK, with the latest push I have a `start_postgres.sh` that automatically makes the postgres database so (in theory) it should be drop-in-able for the normal docker entrypoint (`docker/init.sh` I think, which is `make start start-workspace-host`). Here's what I run and see:

```
docker run -it --rm -v $PWD:/PrairieLearn -e PGDATA=/fooball prairielearn/prairielearn

Making new postgres database at /fooball
error: Error initializing workspace host: type course_instance_access_rules does not exist ..... longdumpoferror

error: Could not mark self as unhealthy relation "workspace_hosts" does not exist .... longdumpoferror
```

If I duplicate that as separate steps:
```
10:42 discovery$ docker run -it --rm -v $PWD:/PrairieLearn -e PGDATA=/fooball prairielearn/prairielearn /bin/bash
bash-4.2# cd /PrairieLearn/
bash-4.2# make start start-workspace-host
Making new postgres database at /fooball
info: PrairieLearn server ready, press Control-C to quit
info: Go to http://localhost:3000

```
I don't get the error and I'm not sure why.

nwalters512:
I don't know why things work if you run scripts manually (maybe run `psql` and see if a schema exists before you run `make start start-workspace-host`?), but I do know why they break otherwise. PL itself is what sets up the database schema; if the workspace host launches before PL, there won't be a schema and it won't be able to run its initialization queries.

Perhaps running `server.js --migrate-and-exit` before the services themselves are started would fix this?

echuber2:
`init.sh` does `make -s -C /PrairieLearn -j 2 start start-workspace-host`. Does removing `-j 2` fix it?

jonatanschroeder:
> `make start start-workspace-host`

The problem is that this waits until PL ends before starting the workspace host, if I recall. The `-j 2` is needed to run both in parallel. Probably the `migrate-and-exit` option that @nwalters512 mentioned is the most promising option.

jonatanschroeder:
Adding this to the top of `init.sh` resolved the issue:

```
make -s -C /PrairieLearn start-support
cd /PrairieLearn && node server.js --migrate-and-exit
```

trombonekenny:
> `init.sh` does `make -s -C /PrairieLearn -j 2 start start-workspace-host`. Does removing `-j 2` fix it?

That's what's different when I run it from `init.sh` versus from the command line (the `-j 2`), good catch. Is the parallelism really needed?

jonatanschroeder:
This may resolve #6329, by the way.

nwalters512:
Parallelism is essential; that's what allows both PL and the workspace host to run at the same time. I believe running the migration before that line is the correct thing to do here.

jonatanschroeder:
This is what I have:

```sh
#!/bin/bash

echo 'Starting PrairieLearn...'

make -s -C /PrairieLearn start-support
cd /PrairieLearn && node server.js --migrate-and-exit

if [[ $NODEMON == "true" ]]; then
    # start-nodemon is listed first so it can use standard input
    make -s -C /PrairieLearn -j 2 start-nodemon start-workspace-host-nodemon
else
    make -s -C /PrairieLearn -j 2 start start-workspace-host
fi
```

trombonekenny:
OK, your `init.sh` changes are in the branch now.

jonatanschroeder:
Sorry for merging with master, I pressed the wrong button 😬

docker/start_postgres.sh
```diff
@@ -13,12 +15,26 @@ if [[ "$ACTION" == "start" ]]; then
...
 fi
 
 if [[ "$ACTION" == "start" ]]; then
     # wait for postgres to start (local or PG_HOST)
     until pg_isready -q ; do sleep 1 ; done
 fi
+
+if [[ $INIT_RESOLVE ]]; then
+    su postgres -c "createuser -s root"
+fi
```
nwalters512, line 40: Nit, trailing newline:
```suggestion
fi

```

nwalters512 (approved):
Looks reasonable! I assume you've tested this with and without `$PGDATA` and/or a volume mount?

echuber2:
In retrospect, it's funny that the tmux script accidentally avoided the issue entirely, since starting the server in one pane and the workspace host in another implicitly runs them in parallel. (I modified that code from some scripts I think @nicknytko and @tdy had written before, I think.)

trombonekenny:
Yeah, all the testing combos I've done seem to work without complaint.

If you wanted to see if CI would benefit from it, I think we'd just need to add something like `-e PGDATA /pgtmpfs --tmpfs=/pgtmpfs` to https://github.com/PrairieLearn/PrairieLearn/blob/367c12df16833f420ae210be48bb0fbf33747b26/.github/workflows/main.yml#L67

Should we do that in this PR or a different one?

trombonekenny:
> In retrospect, it's funny that the tmux script accidentally avoided the issue entirely, since starting the server in one pane and the workspace host in another implicitly runs them in parallel. (I modified that code from some scripts I think @nicknytko and @tdy had written before, I think.)

Did you want to throw that change into this PR? I don't want to lose it in the comments if we don't do anything with it and don't incorporate it.

echuber2:
> Did you want to throw that change into this PR? I don't want to lose it in the comments if we don't do anything with it and don't incorporate it.

Maybe I should make a separate PR for it so I can test it thoroughly. The existing parallel startup method with tmux was working, but the ability to honor PGDATA is good to adapt. Admittedly I don't know if anyone still uses the tmux script but me. Perhaps I should refactor that script so that it calls the maintained versions of functions from the other scripts directly instead of having ad-hoc copies of code (for example, I have the kill workspaces lines verbatim).

docker/init.sh
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash
 
 echo 'Starting PrairieLearn...'
+
+make -s -C /PrairieLearn start-support
+cd /PrairieLearn
```
jonatanschroeder, line 6: the `-C` option changes the directory where `make` operates. If we're already changing the working directory anyway, it's not necessary.
```suggestion
cd /PrairieLearn
make -s start-support
```

Same for the `make` calls below.

jonatanschroeder (approved):

docker/init.sh
trombonekenny: OK, I like that. It's cleaner.

Issue #4924, opened 2021-10-06
pl-order-blocks: easier code formatting


SethPoulsen:
It is possible to get code formatting by wrapping the text in each of the blocks with `<code></code`, but this is a little clumsy. 

Maybe we should put a new option on pl-order-blocks so that it will just wrap each of the blocks with `<code></code>` for you?

lslavice:
Aha! Your comments made me think I was missing something obvious, and I was. My local copy of PL I think must be out of date because when I was testing things locally, none of the answers that were in code font would appear. But I just pushed it to the actual course on PL, and I get at least what you are talking about: lines are in code font, but with extra margins. And I can have a block with multiple lines in it, which also wasn't rendering locally. See below (this is just using `<code>` not `<pl-code>`:

![image](https://user-images.githubusercontent.com/70032904/137359267-4b11e92d-0ccd-46ae-9243-fa78dc4c3929.png)

(And that's what I meant by "multiple lines", although I'm totally down for the dragging multiples of the same thing over like in issue #4055  so I don't have all these random curly braces and weird ranking logic or whatever.)

I am still new to PL, so thank you for your help and sorry if I caused confusion! Mostly, I just wanted to throw support at the `pl-order-blocks` having a nice way to make them all be `pl-code` styling, because that would be very nice. :)

SethPoulsen:
Ya, that double margin thing really isn't very pretty. 

I think we should probably have a way to have the code styling apply directly on the blocks instead of having the extra box inside of them, especially since this element is so commonly used for code.

SethPoulsen:
Currently no one is working on this anyone can feel free to have a go.

echuber2:
Trying to clarify: Is it desired that nested blocks *should* auto-indent on the solution assembly column? Or that they _should not_?

SethPoulsen:
The element has no way at all of being aware of the content of the blocks, and we have no way of doing auto-indentation or any plans for adding that.

This issue is about making the formatting of code inside of pl-order-blocks more pleasant, by making it look better and by making it easier to write the HTML.

My comments above are with regards too making it so that there is only a single margin in the blocks. 

Currently, is you do `pl-code` inside of `pl-order-blocks`, there is a margin between the edge of the block and the code formatting, then there is _another_ margin between the edge of the code formatting and the actual code. 

We need the code formatting to applied directly to the order block, instead of having another div inside it. this will look much better.

echuber2:
[...]

PR #5753, opened 2022-05-04
Add code option to pl-order-blocks

SethPoulsen:
implement #4924


SethPoulsen:
Before: 
image


After: 
image

elements/pl-code/pl-code.py
```diff
@@ -184,7 +184,7 @@ def render(element_html, data):
 
     formatter_opts = {
         'style': 'friendly',
-        'cssclass': 'mb-2 rounded',
```
SethPoulsen, line 187: This extra bottom margin was causing the code to not be vertically centered in the blocks. 

I couldn't think of any good reason for it to ever be on blocks of code so I just removed it always. Let me know if this is wrong.

elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -35,7 +40,12 @@ $(function() {
             </ul>
             <ul id="order-blocks-dropzone-{{uuid}}" name="{{answer_name}}" class="list-group pl-order-blocks-connected-sortable dropzone">
                 {{#submission_dict}}
-                    <li uuid="{{uuid}}" string="{{inner_html}}" style='margin-left: {{indent}}px;' class="list-group-item shadow">{{{inner_html}}}</li>
+                    {{#code}}
+                    <li class="shadow" uuid="{{uuid}}" style="list-style-type:none;" string="{{inner_html}}" style='margin-left: {{indent}}px;><pl-code language="{{code_language}}">{{{inner_html}}}</pl-code></li>
+                    {{/code}}
+                    {{^code}}
+                    <li class="list-group-item shadow" uuid="{{uuid}}" string="{{inner_html}}" style='margin-left: {{indent}}px;>{{{inner_html}}}</li>
+                    {{/code}}
```
SethPoulsen, line 48: I'm not really a fan of all the extra complexity added to this file, but I'm not sure if there is a better way.

SethPoulsen: I've fixed it by just adding the `pl-code` tags in the prepare function. This also fixed my other bug!

SethPoulsen:
Fixed & ready to go. 

The only thing we may want to discuss: 

Do we really want to have the option be called `code` and be a boolean? 

suppose we wanted to add other formatting options in the future, then we would have to make _another_ new property to support that. Alternatively, we could instead have the options be something like `formatting=code` and `formatting=default` or something like that. Then it would be more easily extensible in the future. I'm interested to hear what @mwest1066 and @nwalters512 think.

We also need to consider how the choice of the name might affect other elements that might want to add a similar feature in the future.

SethPoulsen:
@mwest1066 @nwalters512 

This is ready for another round of review.

elements/pl-code/pl-code.py
nicknytko: It looks like without it we get some weird behavior when we have two `<pl-code>` tags next to each other vertically:
image
(Hard to see, but those *are* supposed to be two separate blocks!)

We could add some hacky attribute to pl-code that turns that off, or it might be better to have an `inline` attribute that turns off margins/padding and renders inline instead of as a block?

SethPoulsen: This is quite annoying. 

@mwest1066, @nicknytko  how about this: we can add some CSS with a fancy selector to override the margin of the pl-code element only in the case when it is the only child of one of the blocks?

echuber2: I think we can use selectors to fix this. @SethPoulsen Would this be for when it's the only pl-code item, or when it's the only child of its parent at all?

echuber2: Or we could just add contextual selectors for the specific case when it's pl-code inside of order blocks.

echuber2: @SethPoulsen I took the liberty of making a commit just now. Please check it out and revert if it's no good.

SethPoulsen: Ya! That's _almost_ exactly what I was going to do, but I was going to make the selector a little more selective--if someone puts a `pl-code` inside a `pl-order-block` along with other content (and I have seen people do this), it should still have the margin. The margin should only be removed if it is the [only child](https://developer.mozilla.org/en-US/docs/Web/CSS/:only-child).

echuber2: Can you give an example of the heterogeneous mixture? I thought the `format="code"` applied to the entire order blocks element.

echuber2: OK, I think I see what you meant. I added some more rules for the mixed case as well as an ugly example to illustrate. (The example commit probably should be reverted, since it's ugly and redundant for end-users.)

SethPoulsen: Sorry I wasn't more specific, I just meant to make the existing selector more specific. 

I guess you are right it doesn't matter much because we are only using that CSS class in the case when `format="code"`, but I still think its nice to have the selector be as specific as possible.

SethPoulsen:
@mwest1066 This is ready to merge from my end.

mwest1066 (approved):

Issue #5280, opened 2022-01-19
Support transparency in filled pl-drawing shapes (triangle, rectangle, circle, polygon)


ffund:
Currently pl-drawing shapes can't be made transparent - specifying e.g. `color="#0000ffff"` doesn't work.

I would like to write questions where students annotate an image (i.e. "place a rectangle around the part of the image that shows X"). I can get almost all the way there with pl-overlay + pl-drawing, but need transparent rectangles (just the outline).

According to @echuber2 via Slack:

> It appears that the color attribute itself indeed only takes a 6-digit hex, and since the parsing function is used in a lot of places on PL, we can't easily extend it to an 8-digit without more thorough testing. But it might be relatively easy to add some fill-opacity and stroke-opacity attributes for pl-drawing shapes (edited)

echuber2:
@ffund You mentioned in the merged PR #5754 that it _partially_ addresses this issue. Is there more left to do, or is this resolved?

ffund:
@echuber2 that PR adds an overall opacity attribute. I wanted a separate fill-opacity and stroke-opacity too, but I couldn't figure out how to make it work.

PR #5754, opened 2022-05-04
Add opacity attribute to some drawing elements

ffund:
This PR adds an opacity attribute to: rectangle, triangle, polygon, circle, line, arc, and point elements.

The `opacity` attribute partially addresses #5280 . It allows us to use `pl-drawing` and `pl-overlay` in combination, to ask questions where students "annotate" an image or graph by placing semi-transparent drawing elements on it.


nwalters512 (approved):
LGTM - thanks for the contribution!

PR #5758, opened 2022-05-08
Add config vars for PG pool size and timeout

mwest1066:
The main server process already has `config.postgresqlPoolSize` and `config.postgresqlIdleTimeoutMillis`. This PR adds them for workspace hosts and grader hosts. No defaults are changed by this PR so there will be no change in behavior unless the new config options are set.

### DIFF
--- a/grader_host/config/base.yaml
+++ b/grader_host/config/base.yaml
@@ -91,3 +91,9 @@ properties:
   postgresqlPassword:
     default: null
     envVar: PG_PASSWORD
+  postgresqlPoolSize:
+    default: 2
+    envVar: PG_POOLSIZE
+  postgresqlIdleTimeoutMillis:
+    default: 30000
+    envVar: PG_IDLETIMEOUTMILLIS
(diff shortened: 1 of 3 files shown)


grader_host/config/base.yaml
```diff
@@ -91,3 +91,9 @@ properties:
   postgresqlPassword:
     default: null
     envVar: PG_PASSWORD
+  postgresqlPoolSize:
+    default: 2
+    envVar: PG_POOLSIZE
+  postgresqlIdleTimeoutMillis:
+    default: 30000
+    envVar: PG_IDLETIMEOUTMILLIS
```
nwalters512, line 99: Heh, we've gotta delete this bespoke config stuff I wrote 😅

nwalters512 (approved):

grader_host/config/base.yaml
mwest1066: We'll get to it at some point 😄

PR #5787, opened 2022-05-23
fix: Grading methods test not submitting manual answer

jonatanschroeder:
The current grading methods tests was not properly submitting an answer to test if the submission process worked. As a result, the testing was done based on outdated variable values from internal grading instead of from manual grading itself. This PR fixes the test to actually submit a value.

A result of this distinction is that it detected a case where questions with manual grading were allowing a grade action which had been disabled in the UI. This PR also fixes the error that the test originally failed to identify.

### DIFF
--- a/lib/question.js
+++ b/lib/question.js
@@ -510,6 +510,10 @@ module.exports = {
     async.series(
       [
         (callback) => {
+          if (question.grading_method == 'Manual') return callback(new NoSubmissionError());
+          callback(null);
+        },
+        (callback) => {
           var params = [variant.id, check_submission_id];
           sqldb.callZeroOrOneRow(
             'variants_select_submission_for_grading',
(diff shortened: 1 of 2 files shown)


lib/question.js
```diff
@@ -509,6 +509,10 @@ module.exports = {
     let questionModule, courseIssues, data, submission, grading_job;
     async.series(
       [
+        (callback) => {
+          if (question.grading_method === 'Manual') return callback(new NoSubmissionError());
+          callback(null);
```
nwalters512, line 514:
```suggestion
          if (question.grading_method === 'Manual') {
            // The button to make a submission won't be available to students
            // on manualy-graded questions; this is just here to handle the
            // case where someone tries to be clever and manually post a response
            // with a `grade` action.
            return callback(new NoSubmissionError());
          }
          callback(null);
```

nwalters512 (approved):

PR #5788, opened 2022-05-23
Only run CI once for PRs

nwalters512:
This PR changes the triggers for our CI workflow so that we only run one CI job per commit on a pull request. Previously, we ran two, which more quickly exhausted our maximum parallel jobs and made folks wait longer to get a green result to merge.

Per suggestion from @SethPoulsen, I'm lifting this change out of #1626 so it can be reviewed and merged independently.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,10 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
 
 jobs:
   build:


SethPoulsen:
This also makes it so that CI will not be run on pushes to non-master branches _unless_ they are a pull request. 

This seems like a sensible change though--usually if I'm working on a branch, I don't want the CI to run on it until I'm ready to turn it into a PR anyway.

nwalters512:
> This also makes it so that CI will not be run on pushes to non-master branches _unless_ they are a pull request.
> 
> This seems like a sensible change though--usually if I'm working on a branch, I don't want the CI to run on it until I'm ready to turn it into a PR anyway.

Great callout, yeah. And if someone still wants feedback from CI without opening up code for review, they can either open a draft PR or run `make lint && make build && make test` locally!

nicknytko:
Why were we running two CI jobs before?  One for the pull request and one for the branch?

nwalters512:
> Why were we running two CI jobs before?

The "why" is just an oversight from when we first set up GitHub Actions back in #1700.

nicknytko (approved):

coatless:
@nwalters512 could you add a manual trigger: 

```
on:
  push:
    branches:
      - master
  pull_request:
  workflow_dispatch:
```

PR #5789, opened 2022-05-23
Manual grading: test cases for typical user flow in latest version

jonatanschroeder:
Test cases checks if:

* [x] Submitting an answer to manual grading question works;
* [x] MG flag is set;
* [x] Manual grading assessment page lists its count;
* [x] Manual grading AQ page lists the instance question;
* [x] Manual grading IQ page opens properly;
* [x] All three pages warn about assessment being open;
* [x] After closing, all three pages no longer show warning;
* [x] Tagging question to a grader updates all three page listings;
* [x] "Grade next submission" shows for grader with questions tagged;
* [x] "Grade next submission" does not show for grader with no questions tagged;
* [x] URL for next submission lists next untagged question;
* [x] URL for next submission lists next question tagged to current grader;
* [x] URL for next submission does not list next question tagged to another grader;
* [x] URL for next submission does not list graded questions (without flag);
* [x] Submitting a grade updates the grade in DB;
* [x] Submitting a grade removes the MG flag;
* [x] After submission, manual grading assessment page updates count of pending questions;
* [x] After submission, manual grading AQ page updates the flag value;
* [x] After submission, student view sees new grade value and status.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -185,6 +185,13 @@ config.workspaceHomeDirRoot = '/jobs/workspaces'; // Where the main server will 
 config.workspaceHostHomeDirRoot = '/jobs/workspaces'; // Where the workspace host server will store workspace files.
 config.workspaceHostWatchJobFiles = false; // Should be enabled if we're watching files to sync with S3
 config.workspaceHomeDirLocation = 'FileSystem';
+config.workspaceDockerMemory = 1 << 30; // 1 GiB
+config.workspaceDockerMemorySwap = 1 << 30; // same as Memory, so no access to swap
+config.workspaceDockerKernelMemory = 1 << 29; // 512 MiB
+config.workspaceDockerDiskQuota = 1 << 30; // 1 GiB
+config.workspaceDockerCpuPeriod = 100000; // microseconds
+config.workspaceDockerCpuQuota = 90000; // portion of the CpuPeriod for this container
+config.workspaceDockerPidsLimit = 1024;
 
 config.chunksS3Bucket = 'chunks';
 config.chunksGenerator = false; // Should we generate chunks?
--- a/workspace_host/interface.js
+++ b/workspace_host/interface.js
@@ -1182,16 +1182,15 @@ function _createContainer(workspace, callback) {
                 ],
               },
               Binds: [`${workspacePath}:${containerPath}`],
-              // Copied directly from externalGraderLocal.js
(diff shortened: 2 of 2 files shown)


tests/manualGrading.test.js
```diff
@@ -0,0 +1,408 @@
...
+        );
+        assert.equal(ciStaffResult.rowCount, 1);
+      })
+    );
+  });
+
+  after('reset default user', () => setUser(defaultUser));
+
+  describe('Submit and grade a manually graded question', () => {
+    it('load page as student', async () => {
```
nwalters512, line 130: I'd recommend using [`mocha-steps`](https://www.npmjs.com/package/mocha-steps) here for these tests that request specific things to be run sequentially in order. There's prior art in PL, e.g. [here in the active access restriction tests](https://github.com/PrairieLearn/PrairieLearn/blob/0cfb2c71da637f7f4b5b82aebe4e8eaf0ae3cd3d/tests/activeAccessRestriction.test.js#L52).

Alternatively, this could all go into one big `test()`/`it()` function.

tests/manualGrading.test.js
```diff
@@ -0,0 +1,408 @@
...
+  const hm1Body = await loadHomeworkPage(user);
+  const $hm1Body = cheerio.load(hm1Body);
+  return siteUrl + $hm1Body(`a:contains("${manualGradingQuestionTitle}")`).attr('href');
+};
+
+/**
+ * Gets the score text for the first submission panel on the page.
+ *
+ * @param {import('cheerio')} $
+ * @returns {string}
```
nwalters512, line 78: I appreciate the effort to add type annotations! Could you add `// @ts-check` as the first line of this file? That will ensure that this file is actually type-checked when `make typecheck` is run.

nwalters512: > but it will take time as that would require the entire code to be reviewed and type annotated

I'm not asking for everything to be type-annotated, just that we opt the file in to typechecking. I added the `// @ts-check` annotation locally and only two lines had to change for typechecking to succeed:

```diff
- const fetch = require('node-fetch');
+ const fetch = require('node-fetch').default;
...
-           submission_score_percent: score_percent,
+           submission_score_percent: score_percent.toString(),
```

Much of the code will still use implicit `any` for types, but a lot of it (especially third-party code like `node-fetch`, `fs-extra`, `path`, `chai`, `cheerio`, etc...) does have useful type coverage!

Issue #3160, opened 2020-09-21
Cron errorAbandonedJobs is not multi-server safe


mwest1066:
`cron/errorAbandonedJobs.js` calls https://github.com/PrairieLearn/PrairieLearn/blob/622417be74eb5f4f03dcf345e1dc631fd1690154/lib/server-jobs.js#L345

This checks which running jobs are not present in the _local_ server's job list: https://github.com/PrairieLearn/PrairieLearn/blob/622417be74eb5f4f03dcf345e1dc631fd1690154/lib/server-jobs.js#L349

Any missing jobs are killed.

This will kill any jobs started by a different server, because they won't be in the local server's job list.

To fix this, we should store the server `instance_id` in the `jobs` (or `job_sequences`?) table so we can only kill our own jobs. Maybe we should add a new table called `hosts` that has one row per VM instance so we can refer to them. This table could have a `heartbeat_at` column that we update every 10 seconds or so.

It would also be good to add a cron job that looks for jobs that belong to hosts that are no longer alive (last heartbeat too long ago) and kills them.

nicknytko:
I think this should be okay, since we're only running cron on the main server anyway?  Or do we want to move away from this in the future?

mwest1066:
> I think this should be okay, since we're only running cron on the main server anyway? Or do we want to move away from this in the future?

I think we should fix it so that we can run multiple main servers if needed (and at some point we might figure out a scheme to have chunk servers handle all the routes of main servers, in which case it would be nice to be truly multi-server safe).

I don't think this is very high priority, however, for the reason you mentioned (we can avoid multiple main servers). This is why I didn't label this as "high impact" 😄

nwalters512:
This caused a few impacts to staff lately - a job was killed before it could release its lock, and that caused a few courses to be temporarily un-syncable.

nwalters512:
This is appearing more regularly. This is the root cause of #6454.

PR #5791, opened 2022-05-23
Fix race condition in server jobs

nwalters512:
After enabling extra logging in #5690, it became clear that the intermittent test failures in CI were related to our ["error abandoned jobs" cronjob](https://github.com/PrairieLearn/PrairieLearn/blob/master/cron/errorAbandonedJobs.js), as all the failing server jobs were labeled as "Job abandoned by server".

I was able to reliably reproduce the failures locally by setting the interval for that cron job to `0.0001` (10 thousand times a second, I believe). This uncovers a race condition in the `errorAbandonedJobs` function. If a job is running when the `select_running_jobs` query is executed but is then marked as succeeded in between the query result being received and checked against `liveJobs`, it will no longer be present in `liveJobs` and will thus be marked as abandoned.

This is a similar problem to what's in `#3160` - we're trying to maintain the same state both in the database and in memory, and doing that incorrectly can lead to problems. I believe that making server jobs multi-server safe would also fix this bug. But until then, the changes in this PR should at least fix the flaky tests.

### DIFF
--- a/lib/server-jobs.js
+++ b/lib/server-jobs.js
@@ -106,12 +106,6 @@ Job.prototype.debug = function (_msg) {
 Job.prototype.finish = function (code, signal) {
   var that = this;
 
-  if (!module.exports.liveJobs[that.id]) {
-    // Guard against handling job exit more than once
-    return;
-  }
-  delete module.exports.liveJobs[that.id];
-
   var params = {
     job_id: that.id,
     stderr: that.stderr,
@@ -120,7 +114,15 @@ Job.prototype.finish = function (code, signal) {
     exit_code: code,
     exit_signal: signal,
   };
-  sqldb.queryOneRow(sql.update_job_on_close, params, function (err, result) {
+  sqldb.queryZeroOrOneRow(sql.update_job_on_close, params, function (err, result) {
+    if (result?.rowCount === 0) {
+      // No jobs were actually updated; a job probably finished twice.
+      // Do nothing.
+      return;
+    }
+
+    delete module.exports.liveJobs[that.id];
+
     socketServer.io.to('job-' + that.id).emit('update');
     if (that.options.job_sequence_id != null) {
       socketServer.io.to('jobSequence-' + that.options.job_sequence_id).emit('update');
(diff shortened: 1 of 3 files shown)


nwalters512:
CI is green, but the JS tests are now taking many times longer to complete for some reason. And in fact, I think they're actually failing - but somehow failing with a zero exit code? I see this right around where I'd expect the end of the logs to be, but then it continues on to other test targets in the Makefile.

```
/PrairieLearn/lib/load.js:142
    if (!_.has(estimators, jobType)) throw new Error(`endJob(): no such estimator: ${jobType}`);

                                     ^

Error: endJob(): no such estimator: request
    at Object.endJob (/PrairieLearn/lib/load.js:142:44)
    at /PrairieLearn/server.js:404:12
    at AsyncResource.runInAsyncScope (async_hooks.js:197:9)
    at listener (/PrairieLearn/node_modules/on-finished/index.js:170:15)
    at onFinish (/PrairieLearn/node_modules/on-finished/index.js:101:5)
  GET /
    at callback (/PrairieLearn/node_modules/ee-first/index.js:55:10)
    at Socket.onevent (/PrairieLearn/node_modules/ee-first/index.js:93:5)
    at Socket.emit (events.js:412:35)
    at Socket.emit (domain.js:475:12)
    at TCP.<anonymous> (net.js:686:12)
    at TCP.callbackTrampoline (internal/async_hooks.js:130:17)
    1) "before all" hook: set up testing server for "should load successfully"
    2) "after all" hook: shut down testing server for "should contain QA 101"
```

lib/server-jobs.sql
```diff
@@ -66,6 +66,8 @@ WITH updated_jobs AS (
         exit_signal = $exit_signal
     WHERE
         j.id = $job_id
+        -- Ensure we can't finish the same job multiple times.
+        AND status = 'Running'::enum_job_status
```
mwest1066, line 70: Not that it matters, but the typecast should not be needed here (Postgres can infer types in this situation).

lib/server-jobs.js
```diff
@@ -369,33 +362,39 @@ module.exports.killJob = function (job_id, callback) {
...
+        // Filter out rows that correspond to running jobs that we know about.
+        if (_.has(module.exports.liveJobs, row.id)) {
+          return callback(null);
         }
+
+        // This job *might* be abandoned by the server, but it's also possible
+        // that in between `select_running_jobs` and now, it actually finished
+        // normally. `update_job_on_error` is written such that it will only
+        // modify rows that still have `status = 'Running'`, so we can safely
+        // execute it no matter what.
```
mwest1066, line 379: Makes sense, thanks for the comment here!

mwest1066 (approved):

lib/server-jobs.sql
nwalters512: Changed in 4cafc0c!

PR #5793, opened 2022-05-25
Manual grading: add feedback panel to question page

jonatanschroeder:
image

Feedback is assumed to be in markdown, as discussed.


nwalters512:
Unfortunately the grading method tests are a little brittle since they work with classnames. I believe this function is the culprit:

https://github.com/PrairieLearn/PrairieLearn/blob/00222118ff2f00eca539d21a75b58829eed75dcc/tests/manualGrading.test.js#L77-L79

I believe updating the selector to `.card[id^="submission"] .pastsubmission-block .card-header .badge` should fix most/all of the failing tests.

jonatanschroeder:
> I believe updating the selector to `.card[id^="submission"] .pastsubmission-block .card-header .badge` should fix most/all of the failing tests.

That is still vulnerable to changes in the future. I added a custom class to the object that needs to be checked by the tests (which is what the `pastsubmission-block` is used for as well).

pages/partials/submission.ejs
```diff
@@ -1,15 +1,38 @@
...
       <div>
         <span class="mr-2">
           Submitted answer
           <% if (submissionCount > 1) { %>
-          <%= submissionIdx %>
+          <%= submission.submission_number %>
           <% } %>
         </span>
-        <span><%- include('submissionStatus', {submission: submission}); %></span>
+        <span class="submission-status"><%- include('submissionStatus', {submission: submission}); %></span>
```
nwalters512, line 35: If something is strictly to be used for testing, I'd rather use a `data-*` attribute that makes it really clear exactly what it is and what it's being used for:

```suggestion
        <span data-testid="submission-status"><%- include('submissionStatus', {submission: submission}); %></span>
```

You can then query this in the test with `$('[data-testid="submission-status"]')`. Check out this section of this blog post that talks about why this is better than classnames: https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change#whats-with-the-data-testid-query

pages/partials/submission.ejs
```diff
@@ -1,15 +1,38 @@
...
+      <% if (submissionCount > 1) { %>
+      (for submitted answer <%= submission.submission_number %>)
+      <% } %>
+    </div>
+    <button type="button" class="expand-icon-container btn btn-outline-light btn-sm">
+      <i class="fa fa-angle-up fa-fw ml-1 expand-icon"></i>
+    </button>
+  </div>
+  <div class="collapse <% if (submission.submission_number == submissionCount) { %>show<% } %>" id="submission-feedback-<%= submission.id %>-body">
+    <div class="card-body feedback-body">
```
nwalters512, line 18: Ditto:

```suggestion
    <div class="card-body" data-testid="feedback-body">
```

pages/partials/submission.ejs
jonatanschroeder: I have changed only the objects I created/modified in this PR, this is probably something to be added in a documentation or something like that for future items (and to eventually update in existing tests).

nwalters512 (approved):

PR #5795, opened 2022-05-26
Ability to change font size of Ace editor

ffund:
The default font size in the Ace editor is small and not very readable. This PR lets us set the font size.


mwest1066:
@jonatanschroeder I know you were looking into editors recently -- any thoughts about this?

ffund:
> Should we change the default if it isn't appropriate?

Increasing the default would make some existing initial code wrap, that was previously not wrapping.

jonatanschroeder:
I saw this, and one concern is that having the font size should probably be changed the student, not the instructor, based on their own comfort level.

Also, if we change the font (or styling in general) for pl-file-editor, should we also allow other elements to change their sizes (or fonts, etc.)? I'm worried this could open a precedent that could require a lot of work in other elements. Most other instructors that mess up with the formatting do it with question-level CSS (which could be done in this case, as well), though I don't know how advisable it is at this point considering it can affect multiple panels (unless the panels move towards iframes, which I understand is a plan in the future).

ffund:
> I saw this, and one concern is that having the font size should probably be changed the student, not the instructor, based on their own comfort level.

I'm not sure why font size is different from something like ace theme in that respect. In general, the instructor is responsible for designing the appearance of the question.

> Also, if we change the font (or styling in general) for pl-file-editor, should we also allow other elements to change their sizes (or fonts, etc.)? I'm worried this could open a precedent that could require a lot of work in other elements. Most other instructors that mess up with the formatting do it with question-level CSS (which could be done in this case, as well), though I don't know how advisable it is at this point considering it can affect multiple panels (unless the panels move towards iframes, which I understand is a plan in the future).

I think this is different, because the file editor does not inherit the appearance of the rest of the page.
but, it is true that I can change font size in the file editor with question-level CSS. It's just a little bit hack-y 🤷

docs/elements.md
```diff
@@ -923,6 +923,7 @@ def fib(n):
...
 | `ace-theme`          | string  | `ace/theme/chrome` | Specifies an Ace editor theme; see the full list of themes [here](https://github.com/ajaxorg/ace/tree/master/lib/ace/theme).                                                                                                                                                                                                                                                                                                                          |
+| `font-size`          | float   | None               | Specifies the font size (in px) for the Ace editor.                                                                                                                                                                                                                                                                                                                                                                                                   |
```
jonatanschroeder, line 926: It seems font size actually supports either a number represented as pixels or a string with a CSS-style size (so `1rem` or something similar would also be acceptable). Not sure if it's worth accepting this option or not, just bringing it up.

As for the default, it would probably bring more questions than information to keep the None here. Documentation is not very clear about the default, but the source code for ace seems to indicate its default as 12px.

ffund: Sure, that makes sense.

jonatanschroeder (approved):

PR #5799, opened 2022-05-27
chore(deps): Replace colors with chalk

nwalters512:
This PR replaces [`colors`](https://www.npmjs.com/package/colors), which was [recently vandalized by its maintainer](https://blog.sonatype.com/npm-libraries-colors-and-faker-sabotaged-in-protest-by-their-maintainer-what-to-do-now), with [`chalk`](https://www.npmjs.com/package/chalk), which is what we use everywhere else we need to color text for the terminal.

This will unblock https://github.com/PrairieLearnInc/PrairieTest/pull/453.


lib/databaseDescribe.js
```diff
@@ -1,7 +1,8 @@
+// @ts-check
```
mwest1066, line 1: I'm looking forward to this being real TS!

tools/pg_describe.js
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env node
```
mwest1066, line 1: Any idea why these files don't show up as rename in GitHub? I assume the changes in the files were minor?

mwest1066 (approved):

tools/pg_describe.js
nwalters512: "minor" in the sense that Prettier was run on them for the first time ever 😛 if it helps make the history cleaner (or at least easier to `git blame`), I can save the rename for another PR.

nwalters512: I did this in [c802148](https://github.com/PrairieLearn/PrairieLearn/pull/5799/commits/c8021484ae941fdd9aac538f5f753a49af2f6305).

nwalters512: Actually, changed my mind, I'm splitting this up into a few more PRs to keep history clean:
#5800 and #5801.

mwest1066: Nice, thanks, I was struggling to see what was going on but now it's crystal clear.

nwalters512: The overall diff for this PR should be much clearer now!

PR #5806, opened 2022-05-29
fix: Use Yarn to execute Node tools

mwest1066:
I believe this is needed to use the copy of `turbo` that's installed by `yarn`. Without this I get a "command not found" for `turbo` because I don't have it globally installed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,10 @@
 export PATH := node_modules/.bin/:$(PATH)
 
 build:
-	@turbo run build
+	@yarn turbo run build
 
 dev:
-	@turbo run dev
+	@yarn turbo run dev
 
 start: start-support
 	@node server.js
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,3 @@
-export PATH := node_modules/.bin/:$(PATH)
-
 build:
 	@yarn turbo run build
 
@@ -9,7 +7,7 @@ dev:
 start: start-support
 	@node server.js
 start-nodemon: start-support
-	@nodemon -L server.js
+	@yarn nodemon -L server.js
 start-workspace-host: start-support kill-running-workspaces
 	@node workspace_host/interface.js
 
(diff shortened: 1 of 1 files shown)


nwalters512:
This should "just work" as-is just like Mocha, etc. since we add the `node_modules` bin directory to `PATH` in the Makefile. Can you confirm that a `turbo` executable exists in `node_modules/.bin`? If not, you may need to run `yarn`.

mwest1066:
> This should "just work" as-is just like Mocha, etc. since we add the `node_modules` bin directory to `PATH` in the Makefile. Can you confirm that a `turbo` executable exists in `node_modules/.bin`? If not, you may need to run `yarn`.

After further investigation it turns out the problem is the old version of `make` that ships with MacOS (v3.81). This version does something funky with command execution. Consider this Makefile:
```
export PATH := node_modules/.bin:$(PATH)
build:
    which turbo
    @turbo run build
test-prairielearn:
    which mocha
    @mocha "tests/**/*.test.{js,mjs}"
```
Using `make` v3.81 this will correctly find and run `node_modules/.bin/mocha` but it will complain that it can't find `turbo`. Weirdly, the `which turbo` line _does_ correctly find `turbo`.

Reading the `make` v3.79 manual [it says](https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_5.html#SEC46):
> When it is time to execute commands to update a target, they are executed by making a new subshell for each line. (In practice, make may take shortcuts that do not affect the results.)

I suspect that `make` thinks it can execute `turbo` directly because none of its arguments need shell expansion, so it bypasses its usual subshell execution model. For `mocha` the arguments need to be processed by a shell, so `make` uses a subshell and everything works.

As semi-confirmation of this, the following rule fails (that is, `which mocha` gives the correct answer, but `make` v3.81 fails to actually run `mocha`):
```
test-prairielearn-no-expansion:
    which mocha
    @mocha tests/testDatabase.js
```

While we could require everyone to upgrade the system `make` on MacOS, I'm inclined to add `yarn` to the front of all the `make` commands so that it will work out of the box. This has the side benefit of making it clearer how to run these commands yourself if someone is reading the Makefile to see how things work.

What do you think?

nwalters512:
Oh man, this again 😂 the "macOS ships with a version of `make` from 2006" has bitten us many times before. Yep, I think using `yarn` to execute everything is the right thing to do here.

mwest1066:
> Yep, I think using `yarn` to execute everything is the right thing to do here.

Done!

nwalters512 (approved):

Issue #5665, opened 2022-04-22
pl-order-blocks displaying incorrect "correct answer" in answer panel


bowmnath:
The "correct answer" displayed by `pl-order-blocks` is sometimes not a valid correct answer.  I have not yet noticed a particular pattern in how it fails, but here is a very simple MWE `question.html`:

```html
<pl-order-blocks answers-name="order" grading-method="ranking" partial-credit="none">
  <pl-answer ranking=5>1</pl-answer>
  <pl-answer ranking=4>2</pl-answer>
  <pl-answer ranking=3>3</pl-answer>
  <pl-answer ranking=2>4</pl-answer>
  <pl-answer ranking=1>5</pl-answer>
</pl-order-blocks>
```

When I tested that question on the live server, it displayed the correct answer as `1, 2, 3, 4, 5`, when the only valid answer is actually `5, 4, 3, 2, 1`.  

Note that the element is still **grading** correctly.  In this case, it (correctly) accepted `5, 4, 3, 2, 1` and (correctly) did not accept `1, 2, 3, 4, 5`.

SethPoulsen:
Ya, this is a known issue that has been on my list of things to fix for a long time, I just haven't gotten to it. Actually, for a long time it wasn't possible to do because we couldn't guarantee a solution to problems in the `dag` grading mode, but now we can so it is totally fixable at this point.

`pl-order-blocks` just assumes that the order you write the items in the question HTML is a correct solution when it goes to display the correct solution.

I don't think there is a dedicated issue for this so we'll keep it open, but I do mention it over in #4650

SethPoulsen:
@bowmnath I've opened a PR to fix this at #5871. Feel free to leave it a code review if you want to help speed the process up!

PR #5871, opened 2022-06-03
pl-order-blocks: don't assume the order given by the HTML is correct

### LINKED ISSUE #4650: pl-order-blocks question failing automatic tests

SethPoulsen:
fix #4650, #5665


elements/pl-order-blocks/dag_checker.py
```diff
@@ -44,10 +68,10 @@ def check_grouping(submission, group_belonging):
         group_id = group_belonging.get(node)
         if group_id is not None and cur_group is None:
             cur_group = group_id
-            cur_group_size = 1
         elif group_id is None and cur_group is not None:
             return i
-        elif group_id is not None and cur_group is not None:
+
+        if group_id is not None and cur_group is not None:
```
SethPoulsen, line 74: This was actually a hilarious bug--we were not grading correctly when there was a `pl-block-group` with only one answer in it. This is of course, a useless edge case, but I still find it funny that the bug was here for a year without anyone noticing.

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -294,33 +322,30 @@ def render(element_html, data):
...
+            ordering_message = 'in any order'
+        elif grading_method == 'dag' or grading_method == 'ranking':
+            ordering_message = 'one possible correct order'
         else:
-            grading_mode = 'in the specified order'
+            ordering_message = 'in the specified order'
         check_indentation = pl.get_boolean_attrib(element, 'indentation', INDENTION_DEFAULT)
         indentation_message = ', with correct indentation' if check_indentation is True else None
 
-        if answer_name in data['correct_answers']:
```
SethPoulsen, line 307: I couldn't figure out a good reason for this check to be here (it isn't in any of the other elements).

testCourse/questions/orderBlocks/question.html
```diff
@@ -48,3 +48,35 @@
 </pl-order-blocks>
 
 <!-- TODO add more to cover the rest of the possible option configurations. -->
+
+
```
nwalters512, line 52:
```suggestion
```

Or move this back to the bottom of this file, and ideally list which configurations aren't already covered.

nwalters512 (approved):

elements/pl-order-blocks/dag_checker_test.py
```diff
@@ -77,13 +78,23 @@ def test_grade_dag():
...
+    with pytest.raises(Exception):
+        dag_to_nx(problem_3_invalid_dag_3, problem_3_groups)
+    dag_to_nx(problem_3_dag, problem_3_groups)
+
+
+def test_solve_dag():
+    problems = [(problem_1_dag, problem_1_groups), (problem_2_dag, problem_2_groups), (problem_3_dag, problem_3_groups)]
+    for depends_graph, group_belonging in problems:
+        solution = solve_dag(depends_graph, group_belonging)
+        assert len(solution) == grade_dag(solution, depends_graph, group_belonging)[0]
```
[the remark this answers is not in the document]
SethPoulsen: I prefer to leave it as is, there's actually a good reason for it.

If we were to hard code a particular solution here, then our tests passing would depend on the implementation details of `nx.topological_sort`, which in theory could return _any_ topological sort of the graph. It would seem weird if we had to update our tests because an implementation detail in a library changed. It would also seem weird to make it look like a particular solution is _the_ correct solution when it is actually just _one of many_ correct solutions.

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -169,17 +189,26 @@ def prepare_tag(html_tags, index, group_info={'tag': None, 'depends': None}):
...
 
     data['params'][answer_name] = mcq_options
     data['correct_answers'][answer_name] = correct_answers
 
+    # if the order of the blocks in the HTML is a correct solution, leave it unchanged, but if it
+    # isn't we need to change it into a solution before displaying it as such
+    data_copy = deepcopy(data)
+    data_copy['submitted_answers'] = {answer_name: correct_answers}
+    data_copy['partial_scores'] = {}
+    grade(element_html, data_copy)
```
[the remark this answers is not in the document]
SethPoulsen: Thanks! I was pretty pleased with this trick, but I was wondering, do you think there are any potential gotchas in doing this?

testCourse/questions/orderBlocks/question.html
SethPoulsen: Listing which ones aren't already covered is half the work of actually doing it :sweat_smile:. I can just remove it for now.

elements/pl-order-blocks/dag_checker_test.py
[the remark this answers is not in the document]
nwalters512: Ah! Thanks for that explanation, makes total sense.

elements/pl-order-blocks/pl-order-blocks.py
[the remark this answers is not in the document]
nwalters512: I don't think so... it'd be *nice* to refactor this such that we don't need to rely on any "tricks", but certainly not required.

SethPoulsen:
@mwest1066 Do you care to take a look at this before it gets merged or no?

elements/pl-order-blocks/dag_checker.py
```diff
@@ -16,6 +16,30 @@ def validate_grouping(graph, group_belonging):
...
+    :param group_belonging: which pl-block-group each block belongs to, specified in the question
+    :return: a list that is a topological sort of the input DAG with blocks in the same group occuring
+    contiguously, making it a solution to the given problem
+    """
+    graph = dag_to_nx(depends_graph, group_belonging)
+    sort = list(nx.topological_sort(graph))
+
+    # We need to ensure that blocks from the same block group occur contiguously. Because we enforce the syntactic
+    # constraint that dependence relationships (edges in the DAG) can't cross group boundaries, we can move
+    # blocks in each group next to one another while maintaining a topological sort.
```
mwest1066, line 31:
```suggestion
    # blocks in each group back earlier to be next to one another while maintaining a topological sort.
```

mwest1066 (approved):

PR #5873, opened 2022-06-03
Remove unnecessary call to getContext during rendering

nwalters512:
This is a continuation of https://github.com/PrairieLearn/PrairieLearn/pull/5762 - I missed a call to `getContext`!

I recommend reviewing this with "Hide whitespace changes" enabled.


mwest1066 (approved):
LGTM. This was a diff that really needed whitespace-hiding to be readable!

PR #5874, opened 2022-06-03
Add instrumentation to freeform question rendering

nwalters512:
This PR adds some instrumentation to the `renderPanel` function used for freeform questions. I'm doing this primarily in preparation for #1626, where we'll want to have detailed visibility into how that's performing, both in terms of errors and timing.


question-servers/freeform.js
```diff
@@ -1173,13 +1190,47 @@ module.exports = {
...
+    panel,
+    pc,
+    submission,
+    variant,
+    question,
+    course,
+    locals,
+    context
+  ) {
+    return instrumented(`freeform.renderPanel:${panel}`, async (span) => {
```
nwalters512, line 1212: Currently, I only have instrumentation for renders, not any of the other operations we have (`generate`, `parse`, `grade`, etc.). Adding instrumentation has proven to be easiest with Promise-based functions, and I didn't want this PR to explode into a full-blown refactor of this code. I intend to do that in a separate PR and then come back to add instrumentation to `processQuestion`, which should give us much better visibility than just instrumenting renders.

packages/opentelemetry/src/index.ts
```diff
@@ -220,8 +244,33 @@ export async function init(config: OpenTelemetryConfig) {
 export async function shutdown(): Promise<void> {
   if (tracerProvider) {
     await tracerProvider.shutdown();
+    tracerProvider = null;
   }
 }
 
+export async function instrumented<T>(
```
nwalters512, line 251: This is the interesting bit, the other changes to this file were mostly to make it possible to test functionality here. I'm honestly surprised that this functionality isn't part of the core OTel API!

packages/opentelemetry/src/index.ts
```diff
@@ -202,11 +227,10 @@ export async function init(config: OpenTelemetryConfig) {
     );
   }
 
-  const tracerProvider = new NodeTracerProvider({
+  tracerProvider = new NodeTracerProvider({
```
nwalters512, line 230: I think this was a bug 🙃 without this, `tracerProvider` would always be falsy in `shutdown()`.

packages/opentelemetry/README.md
```diff
@@ -23,13 +23,28 @@ await init({
...
 
-To manually instrument code, you can use the `trace` export:
+To easily instrument individual pieces of functionality, you can use the `instrumented()` helper function:
 
 ```ts
-import { trace } from '@prairielearn/opentelemetry';
+import { instrumented } from '@prairielearn/opentelemetry';
+
+async function doThing() {
+  return instrumented('span.name', async (span) => {
```
mwest1066, line 32: Nice!

packages/opentelemetry/src/index.ts
mwest1066: Yes! Makes sense.

packages/opentelemetry/src/index.ts
mwest1066: I really like this kind of wrapper. I've been enjoying `runInTransaction()` recently which is similar.

question-servers/freeform.js
```diff
@@ -1173,13 +1190,47 @@ module.exports = {
...
+      const result = await promisify(module.exports.renderPanel)(
+        panel,
+        pc,
+        variant,
+        submission,
+        course,
+        locals,
+        context
+      );
+      span.setAttribute('cache.status', result.cacheHit ? 'hit' : 'miss');
```
mwest1066, line 1229: Cool!

mwest1066 (approved):

packages/opentelemetry/src/index.ts
nwalters512: Yes! Me too. I with there was language-level support for this like Python's `with ...` syntax, but the function form isn't so bad either.

question-servers/freeform.js
nwalters512: Yes! Once we prove this out in prod, I'd like to rip out the cache tracking that we currently have in the `page_view_logs` table.

packages/opentelemetry/src/index.ts
mwest1066: That would be so nice!

question-servers/freeform.js
mwest1066: Totally agreed.

Issue #1844, opened 2019-11-07
sympy to/from json loses Symbol assumptions


trombonekenny:
`server.py`
```
import prairielearn as pl
import sympy
import python_helper_sympy as phs

def generate(data):
    c = sympy.var('c', positive=True, real=True)
    ans = 1 / sympy.sqrt(c)
    data['correct_answers']['foo'] = pl.to_json(ans)
```
When the `grade` function unpacks the JSON'd answer back into Sympy (`python_helper_sympy.json_to_sympy`) it loses the assumptions made with the c Symbol so the answers don't match.

(Because this is [one of those edge cases](https://docs.sympy.org/latest/tutorial/simplification.html#powers) where the `positive=True, real=True` assumptions need to be set with the variable. If we do that on the `generate` it doesn't persist through to the `grade`

@tbretl How should we fix this?

mwest1066:
I'm assigning this to @tbretl in case he finds some extra time on the flight back from China 😄

trombonekenny:
@coatless Suggested https://github.com/PrairieLearn/pl-stat400/blob/master/questions/symbolic_differentiation_sqrt/server.py would be a good test question too.

coatless:
@tbretl did you ever get a chance to look into preserving assumptions set within `sympy` when converted with `pl.to_json()`?

echuber2:
Here's a relevant chunk of the code:

https://github.com/PrairieLearn/PrairieLearn/blob/905d6d62afb11f6b187653aea7a178b0aeb58844/lib/python_helper_sympy.py#L251

Could have `sympy_to_json` include another key in the object, `_assumptions`, that is a dict mapping each variable in `_variables` to yet another dict of that variable's assumptions. I think you can already get that with the `assumptions0` property:

```
>>> c = sympy.var('c', positive=True, real=True)
>>> c.assumptions0
{'negative': False, 'infinite': False, 'extended_nonpositive': False, 'nonzero': True, 'extended_negative': False, 'complex': True, 'zero': False, 'extended_real': True, 'positive': True, 'finite': True, 'commutative': True, 'hermitian': True, 'extended_nonzero': True, 'extended_positive': True, 'imaginary': False, 'nonnegative': True, 'nonpositive': False, 'extended_nonnegative': True, 'real': True}
```

These would have to be reapplied in the constructor when the variables are recreated in `convert_string_to_sympy`:

[...]

PR #5880, opened 2022-06-04
draft: pl-symbolic-input: attempts at implicit multiplication

cbm755:
Before doing the AST stuff, we attempt to muck around in the string
using SymPy's transformations.  Specifically, we use
`implicit_multiplication_application` which will change `14x` into
`14*x`.  In theory, this also gives `13xyz` into `13*x*y*z` and
`cos(3x)` into `cos(3*x)`, although in this commit I think it will give
`c*o*s*(3*x)` due to empty `global_dict` which I believe is an allow
list, usually set to (roughly) `from sympy import *`.

Completely untested as I have no idea how to run PL except via Docker!

### DIFF
--- a/lib/python_helper_sympy.py
+++ b/lib/python_helper_sympy.py
@@ -185,6 +185,17 @@ def evaluate(expr, locals_for_eval={}):
     if '#' in expr:
         raise HasCommentError(expr.find('#'))
 
+    from sympy.parsing.sympy_parser import stringify_expr
+    from sympy.parsing.sympy_parser import standard_transformations
+    from sympy.parsing.sympy_parser import implicit_multiplication_application
+    # TODO: may want `function_exponentiation` as well
+    T = standard_transformations + (implicit_multiplication_application, )
+    # TODO: we probably need this to avoid sin -> s*i*n.  How does this relate
+    # to the allow list below?
+    global_dict = {}
+
+    expr = stringify_expr(expr, locals_for_eval, global_dict, T)
+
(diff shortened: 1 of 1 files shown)


cbm755:

cbm755:
Help very much welcome!  So far I think I have learned how to run my local install by reading this: https://prairielearn.readthedocs.io/en/latest/installingLocal/

Rather predictably my code does not work ;-)    I've been testing with http://localhost:3000/pl/course_instance/1/instructor/question/24/preview

cbm755:
So one complication is that `pl-symbolic-input` does not accept `srepr()` strings.  That is, you cannot write `2*Symbol("x")` and have it work.  That is unfortunate!  As the rewrite stuff I'm trying here gives back such things...

This might also be why assumptions are broken #1844.  One cannot do `Symbol('x', real=True)` and have that survive the various to-from json thingies...

eliotwrobson:
@cbm755 is there any roadmap to getting this working? The backend for the symbolic input got rewritten substantially, so this will have to be integrated with that before merging. If it doesn't seem feasible then better to close this PR.

eliotwrobson:
@cbm755 going to go ahead and close since this is stale, and the same feature is being implemented as part of https://github.com/PrairieLearn/PrairieLearn/pull/7020. Would be great to continue the discussion of this feature there!

cbm755:
Very good, I will follow up there at #7020.   I didn't pursue it as everyone seemed swamped at the time I started trying this.

PR #5888, opened 2022-06-07
Configure ESLint to report unused disable directives

nwalters512:
This PR enables the [`--report-unused-disable-directives`](https://eslint.org/docs/user-guide/command-line-interface#--report-unused-disable-directives) feature of ESLint and removes all unused disable directives.


server.js
```diff
@@ -1729,7 +1729,7 @@ if (config.startServer) {
             (time) => {
               const msg = `BLOCKED: Blocked for ${time}ms (set config.blockedAtWarnEnable for stack trace)`;
               logger.verbose(msg, { time });
-              console.log(msg); // eslint-disable-line no-console
+              console.log(msg);
```
mwest1066, line 1732: Why was this one unused? I thought we _were_ warning on the use of `console.log()`?

mwest1066 (approved):

server.js
nwalters512: I can't find any references to that rule ever being enabled in our config 😬 maybe we do want to leave it enabled?

mwest1066: Huh! Maybe I'm thinking about PT where do have `'no-console': 'error'` set. I having this set to help catch leftover debugging logs.

PR #5890, opened 2022-06-07
Docs: update manual grading display

jonatanschroeder:
Update docs to reflect the change in #5793.


jonatanschroeder:
@mwest1066 @nwalters512 I'm wondering if it's worth adding this change as a news item, encouraging instructors to remove the feedback from their questions based on the docs above. Although manual grading isn't enabled for most courses, the manual feedback is still used by instructors that use the upload feature.

mwest1066:
I think a news item is a good idea if there are user-visible changes (for non-beta features, like feedback via upload).

docs/manualGrading.md
```diff
@@ -70,6 +57,10 @@ Each question will have its score and/or feedback updated and the total assessme
 
 If you prefer to use points rather than a percentage score, rename the `score_perc` column in the CSV file to `points`.
 
+If a feedback column is provided, it will be shown by default to the student above the submission panel, as seen in the image below. This feedback accept Markdown formatting.
```
nwalters512, line 60: Nit:
```suggestion
If a `feedback` column is provided, it will be shown by default to the student above the submission panel, as seen in the image below. The feedback can utilize Markdown formatting.
```

nwalters512 (approved):

news_items/037_feedback_view/index.html
```diff
@@ -0,0 +1,73 @@
...
+  encouraged to remove this snippet from your questions.
+</p>
+<pre class="border rounded bg-light p-2">
+&lt;pl-submission-panel&gt;
+  {{#feedback.manual}}
+  &lt;p&gt;Feedback from course staff:&lt;/p&gt;
+  &lt;markdown&gt;{{{feedback.manual}}}&lt;/markdown&gt;
+  {{/feedback.manual}}
+&lt;/pl-submission-panel&gt;
+</pre>
```
mwest1066, line 69: This is all fine, but it takes until the very end of the news item to tell instructors what to do. Could we get this up in the first paragraph? Say something roughly like "There will be changes coming to manual grading in the coming months. The first change is that manual grading feedback will be shown automatically to students. If you have been using manual grading feedback then you can remove <code>{{{feedback.manual}}}</code> from your questions to avoid showing it to students twice. More details are below."

jonatanschroeder: I have reviewed the first few paragraphs with this in mind. Let me know if it looks clearer.

mwest1066: Looks great -- thanks!

PR #5893, opened 2022-06-11
Duplicate answer names

### LINKED ISSUE #5881: Better documentation and/or error message on duplicate answers-name
Documentation is currently vague on the fact that answers-name should be unique within a question. While this is something easier to expect for someone with a programming background or when the correct answer is set in `server.py`, for instructors in non-technical areas this should probably be made more evident. What doesn't help is also the fact that the elements don't properly check this on `prepare`, leading to weird errors on submission instead of a clearer error on question preview.

So one of these should probably be considered (preferably both):
* clarify in docs, for every element that has an `answers-name` field (or equivalent), that this field must be unique within the question;
* include checks in `prepare` functions for gradable elements that verify if the answer name is used by any other element.

Related: #1278

tanyongzhi:
- Clarify in the docs for each element in a question to have unique `answers-name` values
- Add a new field, `answers_names` in `prepare` to store the answer-names attributes of elements in a question
- Check and throw error message if a duplicate answer-names is encountered

Fixes #5881


elements/pl-checkbox/pl-checkbox.py
```diff
@@ -30,6 +30,11 @@ def prepare(element_html, data):
     pl.check_attribs(element, required_attribs, optional_attribs)
     name = pl.get_string_attrib(element, 'answers-name')
 
+    if name in data['answers_names']:
+        raise Exception('duplicate answer-name attribute: %s' % name)
+    else:
+        data['answers_names'][name] = True
```
jonatanschroeder, line 36: Instead of a dictionary, I wonder if it may be better to store this as a list instead, since the `True` value brings no meaning at all. Alternatively, if there is any useful information here (e.g., something that can add information to the exception message to help the instructor find the duplicate), the `True` value could be replaced with that, but I don't think this would make a big difference.

elements/pl-matrix-component-input/pl-matrix-component-input.py
```diff
@@ -25,6 +25,12 @@ def prepare(element_html, data):
     optional_attribs = ['weight', 'label', 'comparison', 'rtol', 'atol', 'digits', 'allow-partial-credit', 'allow-feedback', 'allow-fractions', 'rows', 'columns', 'allow-blank', 'blank-value']
     pl.check_attribs(element, required_attribs, optional_attribs)
     name = pl.get_string_attrib(element, 'answers-name')
+
+    if name in data['answers_names']:
+        raise Exception('duplicate answer-name attribute: %s' % name)
```
eliotwrobson, line 30: Best to avoid raising the base `Exception` when possible. Also, should use f-strings.

```suggestion
        raise ValueError(f'duplicate answer-name attribute: {name}')
```

elements/pl-checkbox/pl-checkbox.py
eliotwrobson: I think storing as a list would be a mistake, since the data structure would allow for insertion of duplicates. Making this a dict, although a little weird, makes it clear that keys should be unique.

apps/prairielearn/python/prairielearn.py
```diff
@@ -67,12 +67,21 @@ class QuestionData(TypedDict):
...
 class ElementTestData(QuestionData):
     test_type: Literal["correct", "incorrect", "invalid"]
 
 
+def check_answers_names(data: QuestionData, name: str) -> None:
+    """Checks that answers names are distinct using property in data dict."""
+    if name in data["answers_names"]:
+        raise KeyError('Duplicate answer-name attribute: "{name}"')
+    else:
+        data["answers_names"][name] = True
```
SethPoulsen, line 82: @nwalters512 is it ok to throw an error here, or are we worried it is going to break questions in production?

It seems like it would be nice to have a method of providing warnings during question rendering that would show up on the instructor preview page but not for students.

eliotwrobson: I think it's fine to throw an error, any question that has a duplicate answers name should be considered broken (i.e., things won't grade right, and rendering will be weird).

For your second item, I've already created an issue: https://github.com/PrairieLearn/PrairieLearn/issues/7337

There's a few instances where it would be nice to display warnings for the question that are just reading from standard error (or similar).

apps/prairielearn/elements/pl-big-o-input/pl-big-o-input.py
```diff
@@ -59,6 +59,8 @@ def prepare(element_html: str, data: pl.QuestionData) -> None:
     pl.check_attribs(element, required_attribs, optional_attribs)
 
     name = pl.get_string_attrib(element, "answers-name")
+    pl.check_answers_names(data, name)
+
```
SethPoulsen, line 63: I don't like that we have to call this function manually from every single element. Isn't there some way we can have it just executed automatically before the `prepare` function of each function? or even implement it in the Javascript?

eliotwrobson: Something like this might be possible in the experimental renderer. But right now, there's no way to collect this single attribute value across all elements with the current separation of logic. This goes back to a structural issue about elements in general, the Python backends are really just a collection of functions and it's hard to share functionality between them (which is why all the graded elements have a bunch of similar-looking boilerplate code). Even the use of `answers-name` as an attribute is by convention.

eliotwrobson (approved):
LGTM!

apps/prairielearn/elements/pl-big-o-input/pl-big-o-input.py
nwalters512: I have the same reservation about this change. As it stands, I had to manually check every element, and this PR indeed neglects to make the appropriate change for some of them. At the same time, as Eliot points out, this name is just a convention. An element could in theory use `answers-name` to identify a specific key to read from (not write to), which would be completely valid *and* fail any kind of automatic checks we build into the renderer.

This concern shouldn't _necessarily_ stand in the way of merging this change, but I'm also not going to rubber-stamp it without a bit more thought/conversation.

eliotwrobson: I think that's perfectly reasonable. As you said Nate, the reason I'm in support of a change like this is the lack of a unified system for dealing with element attributes. The ideal thing to do would be to have a dataclass-like interface for the developer, where they could specify the attributes on each element (with appropriate defaults) and define some type of validation logic on top of all of this, instead of doing everything manually the way it is now. Then this change could be integrated in the background as part of that validation logic.

However, until something like this gets written, we don't really have a choice and any addition of validation / consistency logic has to be done manually. Even bolting this on top of something like `check_attribs` could cause unexpected issues with people's client code.

elements/pl-checkbox/pl-checkbox.py
eliotwrobson: Actually, I'm kinda flip-flopping on this. Since this logic is now handled by a helper function, using a list might not make a difference (since users won't really see this). Do you have any thoughts @nwalters512 ?

apps/prairielearn/elements/pl-threejs/pl-threejs.py
```diff
@@ -160,6 +160,7 @@ def get_objects(element, data):
 def render(element_html, data):
     element = lxml.html.fragment_fromstring(element_html)
     answer_name = pl.get_string_attrib(element, "answer-name")
```
nwalters512, line 162: 😬 `answer-name` vs `answers-name`, which means that the raised exception will have an incorrect error message.

(FWIW, I actually rather dislike `answers-name` as an attribute name, though we're in _much_ too deep to do anything about that now lmao)

nwalters512 (approved):

elements/pl-checkbox/pl-checkbox.py
nwalters512: I think a dictionary still makes sense here. What we *really* want is a set, but a dictionary with `True` values is as close as we can get with JSON.

apps/prairielearn/elements/pl-threejs/pl-threejs.py
eliotwrobson: Ahhh, I didn't really think about this until you just mentioned it. Yeah, it's used in way too many places to really change. I'm also not sure what the status of this threejs element is, from some other issues it seems like there are a bunch of feature requests, but people on those have said this thing isn't really being maintained or used in any substantial way.

PR #5894, opened 2022-06-11
docs typo fix: "the answer" not "or answer"

cbm755:
Actually my understanding of panels is so incredibly low * that someone
should double-check whether this is the correct fix.

* bad attitude due to 1.5 hours of `*-panel` debugging.  If I actually learn enough to understand I will try to improve these docs but prognosis is dim


cbm755:
sorry wrong branch, closed

Issue #5885, opened 2022-06-06
Assessment instances page: minor issues with group assessments


jonatanschroeder:
The assessment instances page (Students tab) has a button for "Students only", that filters assessments listing only those assigned to students (i.e., not including staff instances). This button doesn't work for group assessments. Using this filter lists no entry at all.
<img width="610" alt="image" src="https://user-images.githubusercontent.com/1004907/172234176-399eb801-95c1-4f4a-a6a3-f145d70db8d4.png">

Another issue in the same page: when deleting an assessment, the modal for deletion does not list the group being deleted. 
<img width="374" alt="image" src="https://user-images.githubusercontent.com/1004907/172234109-47c7dafb-d39c-4e96-b55a-0b841d6bf8cf.png">

tanyongzhi:
Hi! I would like to tackle this issue, but I'm having trouble replicating it. I am struggling to access the homework using the "Student" role. Changing the view to "student view" via the dropdown on the upper right corner and attempting the homework still shows a Staff role in the Students tab. 

Any help would be greatly appreciated!

jonatanschroeder:
Hi, @tanyongzhi, welcome to the team! I assume you are trying to access the homework as student to be able to enrol as a student in a local environment and attempt the assessment, is that correct? If so, there are a few possible ways to do it, but one better option is to:

1. Go to the PrairieLearn home page (https://localhost:3000/ or whatever suits your environment);
2. At the top menu, click on "Queries";
3. Select "generate_and_enroll_users.sql";
4. Type the ID of the course instance you want to add students to (this is the number that follows "`course_instance/`" in the URL) and run the query;
5. For group assessments: back to the assessment, go to the Groups tab, and select "Auto" to create random groups;
6. Back to the queries page, select "generate_assessment_instances.sql" (for individual assessments) or "generate_assessment_instances_group.sql" (for group assessments), and run the query with the appropriate assessment ID.

The sequence above should enroll a set of students (30 by default), assign them to groups, and create an assessment instance for each group. This should be enough to test your implementation of the code above.

Good luck, and feel free to ask more questions if you have any concerns.

tanyongzhi:
> I assume you are trying to access the homework as student to be able to enrol as a student in a local environment and attempt the assessment, is that correct?

That's right! I'll give your suggestion a go, thanks for the help!

tanyongzhi:
[...]

PR #5899, opened 2022-06-12
Fix "Students Only" button for group assessments

tanyongzhi:
- Add sql function `group_get_displayed_role` that returns the a list of corresponding roles for the input user id array
- Add a new column that displays the list of roles for for each group member in the group assessment instructor table 
- Modify the "Students Only" button for group assessments to filter out all groups with no group member with the "Student" role

Example of filter off:
image

Example of filter on:
image

Part of #5885


sprocs/group_get_displayed_role.sql
```diff
@@ -0,0 +1,18 @@
...
+    user_id integer;
+BEGIN
+    IF user_ids IS NOT NULL THEN
+        FOREACH user_id IN ARRAY user_ids
+        LOOP
+            roles = array_append(roles, users_get_displayed_role(user_id, course_instance_id));
+        END LOOP;
+    END IF;
+END;
+$$ LANGUAGE plpgsql STABLE;
```
jonatanschroeder, line 18: I'm not convinced that a separate sproc is warranted in this case, instead of adding a column to the `group_info` sproc instead. This may require some consideration as to the performance impact if the sproc is used elsewhere that doesn't require roles, but there could be options to remediate that. Also, it would probably make sense to have this computation be done through a query (`array_agg`) instead of a loop, and to return it with unique roles.

pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
```diff
@@ -14,7 +14,19 @@ $(function() {
                     table.data('filter-student-only', filterOn);
 
                     $('.columns button[name=studentsOnly]').toggleClass('active', filterOn);
-                    table.bootstrapTable('filterBy', filterOn ? { role: 'Student' } : {});
+                    <% if (assessment.group_work) { %>
+                        table.bootstrapTable('filterBy', {id: 1}, {
```
jonatanschroeder, line 18: why is id there?

pages/instructorAssessmentInstances/instructorAssessmentInstances.ejs
```diff
@@ -234,6 +234,15 @@
                   data-class="text-center align-middle text-wrap"
                   data-formatter="listFormatter"
                   data-switchable="true">Group Member Names</th>
+              <th data-field="group_roles"
+                  data-sortable="true"
+                  data-sorter="roleSorter"
```
jonatanschroeder, line 239: I believe the `roleSorter` function is leftover from a previous version of this code and is no longer in place. Probably should be removed. Also, since this field will be a list, you probably should consider using a listFormatter, particularly to consider cases of groups where all students left the group (empty), or groups where all students dropped the course (None role).

jonatanschroeder (commented):
Thanks for your initial work on this PR. I have added a few comments to consider.

sprocs/group_get_displayed_role.sql
tanyongzhi: I agree with removing the need for a loop, we probably can rewrite the query from scratch and not rely on `users_get_displayed_roles`.

We can probably place the logic in the existing `group_info` sproc, although I'm not too sure how performance for other queries will be affected. Do you have any suggestions on how the performance impact can be minimized for use cases where roles aren't required?

tanyongzhi:
Thank you for the review. Unfortunately, I will only be able to work on code on Fridays and the weekends. I'll make the relevant changes then. Thank you for your patience!

sprocs/group_get_displayed_role.sql
jonatanschroeder: I just did a quick scan of all places that use this sproc, and most of them could be simplified by removing the sproc call altogether. In essence, these three references could easily replace `gi.id` with `ai.group_id` which would remove all need for this PR in general.
* `pages/instructorAssessmentStatistics/instructorAssessmentStatistics.sql`
* `sprocs/assessment_stats.sql`
* `sprocs/assessment_duration_stats.sql`

This leaves two places: this page and the API endpoint for assessment instances. I don't believe the API call would have an impact that I'd consider worrisome, so I think you can easily add the student role as a column without problems, and remove the reference to this sproc in the files above.

As for `users_get_displayed_role`, I believe it is better to keep its use for consistent computation of the role, but use it inside the query instead of in a loop.

And this issue is by no means urgent, if you can't work on it for a few days it's understandable.

pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
tanyongzhi: For the `filterBy` method, it seems that a `filter` has to be specified in the parameters (link [here](https://bootstrap-table.com/docs/api/methods/#filterby)). If the `filter` parameter is not specified, the filter does not do anything, and no roles will be filtered out. I added a comment that will hopefully clarify things a little, and also renamed some variable names for clarity.

jonatanschroeder: can you set the `filter` to a blank object (`{}`)? I haven't tested, just wondering.

tanyongzhi: Setting the filter to a blank object will not filter out any rows.

jonatanschroeder: What about `group_roles: "Student"`? Since it has no effect, at least keep it with a meaningful value...

tanyongzhi: Alright, let's do that. I'll keep the comments to clarify too.

nwalters512: It looks as though this is fixed in the latest version of `bootstrap-table` via https://github.com/wenzhixin/bootstrap-table/pull/6104. @jonatanschroeder do you know of any reasons we shouldn't install the latest version?

sprocs/group_info.sql
```diff
@@ -1,11 +1,14 @@
...
+        IN assessment_id bigint,
+        IN course_instance_id bigint DEFAULT null
     ) RETURNS TABLE (
         id bigint,
         name text,
         uid_list text[],
-        user_name_list text[]
+        user_name_list text[],
+        user_id_list bigint[],
+        roles text[]
```
nwalters512, line 11: Nit: consider calling this `user_roles_list` instead for consistency with the other output values? Tbh, I never liked the `_list` suffix, so I'd also be happy if you refactored to remove that 🙂

nwalters512 (commented):
> if the way that role retrieval as done here may affect performance in a significant way.

I don't have great intuition here. I get a little nervous when I see that a sproc is being called many times during the execution of a particular query. Let's keep an eye on this once it's deployed, and we can revisit if needed.

> if the way the new column is presented should be changed (I was thinking of having it show unique roles, or just show student if there is at least one student, or something like that).

Yeah, showing `Student, Student, Student, ...` isn't the most useful thing IMO. I like the idea of unique roles!

pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
jonatanschroeder: I don't have any specific reason, I'd just like to make sure we test it properly before upgrading bootstrap-table. The last time we upgraded it (to 1.19.1) it introduced some weird errors (see #5362), and bootstrap-table functionality isn't covered by our CI tests, so I'd prefer to do it in a separate PR (dependabot has #5794) with proper human testing.

sprocs/group_info.sql
tanyongzhi: I think I'll refactor it as `user_roles_list` for now. Perhaps we can remove the `_list` suffixes in another PR.

tanyongzhi:
> Yeah, showing `Student, Student, Student, ...` isn't the most useful thing IMO. I like the idea of unique roles!

Yes, this sounds like a good idea!

jonatanschroeder (approved):
Tested locally, works as expected. I just made a suggestion related to the column title (no need to indicate it's unique there), other than that all good.

PR #5900, opened 2022-06-12
Fix bug in delete group assessment instance

tanyongzhi:
- Fix an issue with delete group assessments not showing the group name and member UIDs

Sample:
image

#5885

### DIFF
--- a/pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
+++ b/pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
@@ -66,6 +66,9 @@ $(function() {
         var uid = button.data('uid'); // Extract info from data-* attributes
         var name = button.data('name');
         var number = button.data('number');
+        var group_name = button.data('group-name');
+        var uid_list = button.data('uid-list');
+        uid_list = uid_list.replace(/,/g, ', '); // adds a space after each comma to prevent text overflow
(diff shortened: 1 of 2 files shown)


pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
```diff
@@ -301,6 +306,8 @@ function actionButtonFormatter(_value, row, index) {
         .attr('data-name', row.name)
         .attr('data-number', row.number)
         .attr('data-date-formatted', row.date_formatted)
+        .attr('data-group-name', row.group_name)
+        .attr('data-uid-list', row.uid_list)
```
jonatanschroeder, line 310: This `uid_list` is actually an array that is being forced into a string here. For consistency, I suggest:

```suggestion
        .attr('data-uid-list', row.uid_list?.join(', ') || 'empty')
```

This way you also don't need to add spaces to commas in the other function, and it will more properly handle groups with no students.

jonatanschroeder (commented):
I tested locally and it looks good, so just a few minor tweaks.

tanyongzhi:
Thank you for the review. Unfortunately, I will only be able to work on code on Fridays and the weekends. I'll make the relevant changes then. Thank you for your patience!

jonatanschroeder (approved):

PR #5901, opened 2022-06-13
Move correctness badge outside input-group-text for pl-integer and pl-number-input

firasm:
I'd like to improve the `pl-number-input` and `pl-integer-input` elements by adjusting the feedback they see when they enter correct, partially correct, or incorrect answers.

## Current Behaviour

A student has just put in an incorrect answer on a number-input question and they get shown a red badge with the text 0% indicating they got no credit for this question.

However, the red badge is immediately adjacent to the info question mark `?` and clicking on the red badge shows them a popover modal telling them what a valid submission should comprise, for that question.
Clearly, this is no longer relevant information given that they've made a valid (but incorrect) submission.

Screen Shot 2022-06-12 at 6 42 38 PM

Screen Shot 2022-06-12 at 6 42 15 PM

## Suggested behaviour

I thinking taking away the info question mark `?` is a bit heavy-handed, so instead, I'm proposing to move the badge of correct, partially correct, and incorrect **outside** the `input-group` like so:
Screen Shot 2022-06-12 at 6 38 49 PM

Screen Shot 2022-06-12 at 6 39 04 PM

I will make the same changes as above to `pl-number-input` as well (in the same PR).


jonatanschroeder:
If this is to be submitted I'd guess other elements with similar look-and-feel should probably be considered as well (not saying it should, I'll leave that decision to higher-ups, but just raising it):
* pl-string-input;
* pl-symbolic-input;
* pl-matrix-component-input;
* pl-matrix-input.

firasm:
Agreed - it should be consistent with the other elements as well. 

Happy to make the change there too if the higher-ups are supportive.

eliotwrobson:
Any updates on this? Seems like a pretty straightforward change, and would be nice to have this in if it's something people feel like would improve things.

eliotwrobson:
Going to go ahead and close this, since the changes here are going to be spread across PRs associated with https://github.com/PrairieLearn/PrairieLearn/discussions/7353.

PR #5905, opened 2022-06-13
fix: Gracefully handle missing LTI callback URL

nwalters512:
[As reported on Slack](https://prairielearn.slack.com/archives/C266KEH9A/p1655148217555639), it's possible that an LTI assessment might be missing a callback URL to report the score, resulting in an error like the following:

```
options.uri is a required argument
```

This PR updates our LTI code to gracefully handle that case.

### DIFF
--- a/lib/ltiOutcomes.js
+++ b/lib/ltiOutcomes.js
@@ -64,6 +64,15 @@ function updateScore(ai_id, callback) {
 
     var info = result.rows[0];
 
+    // It's possible for a user to enter via LTI but for the assessment to not
+    // have a callback URL set, for instance, if the instructor forgot to set
+    // one or if the assessment is a practice exam that doesn't need to have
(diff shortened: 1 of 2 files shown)


mwest1066 (approved):

trombonekenny:
I feel like we'd be better off fixing the issue in ltiOutcomes.sql than checking after the fact? Do you want me to go into the branch and make those changes?

nwalters512:
@trombonekenny could you elaborate on what you have in mind? What's the underlying issue you're talking about?

trombonekenny:
> @trombonekenny could you elaborate on what you have in mind? What's the underlying issue you're talking about?

It's a one liner so I made the change in a different way and pushed the change. We already check for a null result in the SQL query, so no need to change the JS just the SQL.

PR #5908, opened 2022-06-15
Add flag to question for hiding correct answer panel

### LINKED ISSUE #623: Add flag to hide "Correct answer" panel on questions
Externally-graded (read: coding) questions don't have a correct answer in the sense that a multiple choice or true/false question does. Therefore, we should have the ability to hide the "Correct answer" panel on questions that don't technically have one.

eliotwrobson:
Adds a flag that allows instructors to manually hide correct answer panel for any question. Resolves #623.


eliotwrobson:

lib/question.js
```diff
@@ -1534,6 +1534,11 @@ module.exports = {
       locals.showTrueAnswer = true;
     }
 
+    // Manually disable correct answer panel
+    if (question && question.hide_correct_answer) {
```
nwalters512, line 1538: Nit (though I think `question` should always be defined here, so checking if it's falsy if probably unnecessary?)
```suggestion
    if (question?.hide_correct_answer) {
```

database/tables/questions.pg
```diff
@@ -12,6 +12,7 @@ columns
     external_grading_image: text
     external_grading_timeout: integer
     grading_method: enum_grading_method not null default 'Internal'::enum_grading_method
+    hide_correct_answer: boolean default false
```
nwalters512, line 15: For consideration: we use `show...` elsewhere, e.g. `showClosedAssessment` and `showClosedAssessment` in `infoAssessment.json`. There might be value in being consistent and doing the same here?

eliotwrobson: I think it's used in enough places that it's worth being consistent. Will switch over.

nwalters512:
@eliotwrobson looks like this is failing our lint checks. If you run `make format`, that should fix that right up.

PR #5941, opened 2022-06-29
Update faq about hiding correct answer

echuber2:
Prompted by: https://github.com/PrairieLearn/PrairieLearn/pull/5908

CC @eliotwrobson @nwalters512


echuber2:
I don't know if I should write "other strategies" instead of "_older_ strategies," but anyway, it doesn't seem like we should just remove the old tip. It may be useful in some cases.

jonatanschroeder:
I think there is a bit of wiggle room here. Some instructors may wish to combine the item above with other concerns, e.g., hide the correct answer for some elements but not others, or hide the correct answer until a specific date/time. The new flag is just all or nothing, other strategies are still valid. So it's not a replacement, but a new feature.

Personally I don't like to add "As of Summer 2022" or equivalent, particularly since this isn't done for other docs and can be confusing if used extensively.

jonatanschroeder:
> The new flag is just all or nothing, other strategies are still valid. So it's not a replacement, but a new feature.

Also, there is a difference between the functionality in the new flag, which hides the correct answer box altogether, and the hide-in-panel functionality, which displays the box but hides the content.

jonatanschroeder (approved):

PR #5943, opened 2022-06-30
Override runMigrations config with --migrate-and-exit

nwalters512:
This PR allows the `--migrate-and-exit` flag to override the value of `config.runMigrations`.


nwalters512:
I talked to Matt about this PR, we decided it makes sense both in principle and for our specific use case.

PR #5946, opened 2022-07-01
Link to documentation from grader image READMEs

nwalters512:
This should help to improve the discoverability of the documentation for each grader image.


jonatanschroeder:
Should this instead point to readthedocs? Since I imagine this README file is also listed in Docker Hub, where the relative link wouldn't make sense, but I may be wrong.

nwalters512:
@jonatanschroeder good call - it looks like it is shown there: https://hub.docker.com/r/prairielearn/grader-python. I'll update.

PR #5985, opened 2022-07-01
Partial credit for multiple choice questions

### LINKED ISSUE #5796: pl-multiple-choice partial credit
As far as I know, it doesn't seem like there's a built-in way to give partial credit in pl-multiple-choice elements. It would be really nice to have this as a feature, and then have a different icon display the partial credit given.

eliotwrobson:
Adds a partial credit attribute for multiple choice questions. Resolves #5796. Surprisingly not that many code changes, since it repurposes dead code that supports this feature in the existing element. Open to changes on the backend behavior of this element.

Screenshot_20220701_232312

Result of backward compatibility test (submission 2 is before change, submission 3 is after):
Screenshot_20220701_231356


elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -339,16 +356,13 @@ def grade(element_html, data):
...
-    score = 0
-    if (submitted_key is not None and submitted_key == correct_key):
-        score = 1
 
     feedback = None
+    score = 0
     for option in data['params'][name]:
         if option['key'] == submitted_key:
             feedback = option.get('feedback', None)
+            score = option.get('score', SCORE_INCORRECT_DEFAULT)
```
nwalters512, line 365: Questions that were generated prior to this change won't have `score` in their params, so this will give students a score of 0 even if they answered correctly. We should ensure this change is made in a backwards-compatible way. Something like this may work:

```suggestion
            score = option.get('score', SCORE_CORRECT_DEFAULT if submitted_key == correct_key else SCORE_INCORRECT_DEFAULT)
```

elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -25,11 +27,21 @@ def categorize_options(element, data):
...
+            pl.check_attribs(child, required_attribs=[], optional_attribs=['score', 'correct', 'feedback'])
             correct = pl.get_boolean_attrib(child, 'correct', False)
             child_html = pl.inner_html(child)
             child_feedback = pl.get_string_attrib(child, 'feedback', FEEDBACK_DEFAULT)
-            answer_tuple = (index, correct, child_html, child_feedback)
+
+            default_score = SCORE_CORRECT_DEFAULT if correct else SCORE_INCORRECT_DEFAULT
+            score = pl.get_float_attrib(child, 'score', default_score)
+
+            if not (SCORE_INCORRECT_DEFAULT <= score <= SCORE_CORRECT_DEFAULT):
```
nwalters512, line 38: TIL that this works in Python! Pretty neat.

exampleCourse/questions/element/multipleChoice/question.html
```diff
@@ -196,3 +196,31 @@
...
+
+        <pl-multiple-choice answers-name="sky-color8">
+            <pl-answer correct="true" feedback="Correct! The sky is indeed blue.">Blue</pl-answer>
+            <pl-answer>Pink</pl-answer>
+            <pl-answer score=0.5>Purple</pl-answer>
+            <pl-answer feedback="Not quite" score=0.1>Orange</pl-answer>
+            <pl-answer>Yellow</pl-answer>
+            <pl-answer score=0.3>Brown</pl-answer>
+            <pl-answer score=0.2>Red</pl-answer>
+          </pl-multiple-choice>
```
nwalters512, line 223: Nit, consistent indentation:
```suggestion
        </pl-multiple-choice>
```

nwalters512 (commented):
My main concern here is backwards compatibility. I think it shouldn't be too hard to address, but I'd encourage you to test this by generating a multiple choice question on `master`, switching to this branch, and then trying to grade it.

It looks like some of the automated tests are failing, you'll need to fix those before we can merge this.

elements/pl-multiple-choice/pl-multiple-choice.py
eliotwrobson: Did something similar to this. Tested briefly (see screenshot in description). Should probably double check for backward compatibility one more time but should be working ok.

eliotwrobson:
@nwalters512 Any updates on when this might get merged? There's not really a rush on our end, but it would be nice to have it in sooner rather than later and I don't want things to get stale.

docs/elements.md
```diff
@@ -133,10 +133,12 @@ incorrect answers and displays them in a random order as radio buttons.
...
-| `feedback` | string  | —       | Helper text (HTML) to be displayed to the student next to the option after question is graded if this option has been selected by the student. |
+| Attribute             | Type    | Default | Description                                                                                                                                    |
+| --------------------- | ------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------- |
+| `correct`             | boolean | false   | Is this a correct answer to the question?                                                                                                      |
+| `feedback`            | string  | —       | Helper text (HTML) to be displayed to the student next to the option after question is graded if this option has been selected by the student. |
+| `score`               | float   | -       | Score given to answer choice if selected by student. Defaults to 1.0 for correct answers and 0.0                                               |
+| to incorrect answers. |
```
nwalters512, line 141:
```suggestion
| `score`               | float   | -       | Score given to answer choice if selected by student. Defaults to 1.0 for correct answers and 0.0 for incorrect answers. |
```

nwalters512 (commented):
@eliotwrobson sorry for the delay! This looks great, just one final formatting fix for the docs and this will be good to go.

docs/elements.md
eliotwrobson: Added the suggestion and then re-ran `make format`. Should be good to go!

nwalters512:
@eliotwrobson unfortunately it looks like a new step added to CI doesn't work well for PRs from forks 😢 I could force-merge this, but if it's alright with you, I'd like to try to see if I can get CI working as expected here!

eliotwrobson:
@nwalters512 sounds like a plan!

nwalters512:
CI is green!

PR #5989, opened 2022-07-07
Emit Node metrics to Cloudwatch

nwalters512:
This PR introduces a few new custom metrics that will be emitted to CloudWatch:

- `NodeEventLoopDelayMean`
- `NodeMemoryRss`
- `NodeMemoryHeapTotal`
- `NodeMemoryHeapUsed`
- `NodeMemoryExternal`
- `NodeCpuUser`
- `NodeCpuSystem`

This should give us a bit more visibility into how PL behaves under load.

Closes https://github.com/PrairieLearnInc/sysconf/issues/162. Closes https://github.com/PrairieLearnInc/sysconf/issues/164.


lib/node-metrics.js
```diff
@@ -0,0 +1,83 @@
...
+    {
+      MetricName: 'NodeCpuUser',
+      Unit: 'Microseconds',
+      Value: cpuStats.user,
+    },
+    {
+      MetricName: 'NodeCpuSystem',
+      Unit: 'Microseconds',
+      Value: cpuStats.system,
+    },
```
nwalters512, line 50: I'm not sure how useful these are to us. They give us microsecond values, which I don't really have a good intuition for. Perhaps this will just be valuable to show us when CPU usage spikes? i.e., the absolute values aren't too useful, but if we see CPU usage double, we might have a problem?

lib/config.js
```diff
@@ -68,6 +68,7 @@ config.cronIntervalWorkspaceTimeoutWarnSec = 60;
 config.cronIntervalWorkspaceHostLoadsSec = 10;
 config.cronIntervalWorkspaceHostTransitionsSec = 10;
 config.cronDailySec = 8 * 60 * 60;
+config.nodeMetricsIntervalSec = 5;
```
mwest1066, line 71: Should we make this default to a value (maybe 0) that causes this to be inactive? Even if we default to active, I think it would be good to have a way to deactivate it via a config option (either treat 0 as inactive, or add an explicit boolean flag).

lib/node-metrics.js
mwest1066: If we pass the previous value then I think we can get the increment: https://nodejs.org/api/process.html#processcpuusagepreviousvalue

We could then divide by the 5 second interval to get a percentage, which might be more useful?

mwest1066: It might also be helpful to separately report the total CPU% (user + system)?

lib/config.js
nwalters512: How about using `null` to disable?

mwest1066 (approved):

PR #5990, opened 2022-07-08
Add typing extensions library

eliotwrobson:
Adds `typing-extensions` library to base pl image and python autograder. Library is for extensions to built in static typechecking. There are some instances for us where it would be nice to use features only available in the typing extensions module. This module back-ports typechecking features available in later versions of Python to earlier versions.


eliotwrobson:
@nwalters512 any opinions on merging this? This library is pretty standard (in fact, it only contains back ports of things from future versions of Python) and should help with development and forward compatibility.

nwalters512:
@eliotwrobson seems reasonable! I'll merge #5985 as soon as I get CI working, and I'll get this branch up to date with master and CI should be able to pass 🎉

nwalters512 (approved):

PR #5994, opened 2022-07-09
Upgrade to NodeJS v16 (from v14)

mwest1066:
NodeJS v16 is now ["Active"](https://nodejs.org/en/about/releases/) so we should upgrade to it.

### DIFF
--- a/images/plbase/plbase-install.sh
+++ b/images/plbase/plbase-install.sh
@@ -39,7 +39,7 @@ cd /nvm
 git checkout `git describe --abbrev=0 --tags --match "v[0-9]*" $(git rev-list --tags --max-count=1)`
 source /nvm/nvm.sh
 export NVM_SYMLINK_CURRENT=true
-nvm install 14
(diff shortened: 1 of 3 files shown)


nwalters512 (approved):
Woohoo! Let's get this rolled out into production, and then we can start using Node-16-specific goodies and drop our little bits of backwards-compat code. A few more thoughts:

- Let's consider building a Node version check into startup that validates that the correct version of Node is being used and aborts the process if it's outdated.
- We should announce this somewhere so folks doing self-hosted PL are aware of this. Since we're not numerically versioned and aren't maintaining a changelog, I'm not totally sure where to put this note...

nwalters512:
> How about I just put this in the #announce channel along with the release announcement? It's not a great place to put it, but it's better than nothing. The new engines spec above will at least make this clear when someone tries to upgrade.

This seems reasonable for now. I don't love that we lose our Slack history after just a few months, but... yeah 😓

PR #5995, opened 2022-07-09
Upgrade C grader to Ubuntu 22.04

jonatanschroeder:
The C grader was still based on Ubuntu 18.04. This PR updates the C grader to Ubuntu 22.04 (latest LTS). It also updates Python to 3.10 and relevant pip packages to their latest versions.

The PR is here to ensure this is done eventually, it may be a good idea to merge this update after the Summer terms in Illinois/UBC/York/etc. but before the Fall term.

Supersedes #5918 and #5988.


mwest1066:
> The PR is here to ensure this is done eventually, it may be a good idea to merge this update after the Summer terms in Illinois/UBC/York/etc. but before the Fall term.

Do you want to convert this to a Draft to make sure we don't accidentally merge it early? I'll rely on you to set a reminder somewhere to actually merge this after Summer.

mwest1066 (approved):

jonatanschroeder:
> Do you want to convert this to a Draft to make sure we don't accidentally merge it early? I'll rely on you to set a reminder somewhere to actually merge this after Summer.
> 

Another option is to merge it now and rely on instructors not syncing their course's view of the image until they have tested it locally or until after the term. Is the cached image on production per course or shared?

mwest1066:
> Another option is to merge it now and rely on instructors not syncing their course's view of the image until they have tested it locally or until after the term. Is the cached image on production per course or shared?

Shared, which is good and bad. Mainly bad in this case.

jonatanschroeder:
Ok, so after term it is. I have a fork of the C grader in my own Docker Hub account that I'm using for beta-testing it with my course, though.

mwest1066:
> Ok, so after term it is. I have a fork of the C grader in my own Docker Hub account that I'm using for beta-testing it with my course, though.

Sounds good!

jonatanschroeder:
Is there a way to check if there is any active course using the C autograder? I don't see a strong reason to wait before merging this, since York (me) uses the new version already, UBC (from what I could tell) doesn't have any course that uses C in Summer term 2, and UIUC has its last day of Summer classes today.

nwalters512:
Given the existence of various self-hosted instances, we're well past the point of Matt and I having perfect visibility into how any piece of PrairieLearn is used. That's part of the reason I was advocating for us to just release brand-new images instead of modifying the existing ones in place: we will never be sure who is using them, or what their semester schedule looks like.

I can spot check the instances that we own, but I still think we need to shift our approach to releasing changes.

mwest1066:
Are we anticipating breakage from this? If not, then I think we should just go ahead and merge it now. I agree that longer-term we should think about versioning, but I don’t want to hold up this PR.

jonatanschroeder:
> Are we anticipating breakage from this? If not, then I think we should just go ahead and merge it now. I agree that longer-term we should think about versioning, but I don’t want to hold up this PR.

I don't expect it to break anything, though there are a few packages being upgraded, so it's hard to be 100% sure. I'll merge it then.

nwalters512:
Sounds good! @jonatanschroeder can you post something about this in the #announce channel on Slack once this lands?

jonatanschroeder:
> Sounds good! @jonatanschroeder can you post something about this in the #announce channel on Slack once this lands?

For sure. Should I also mention the recent Java upgrade?

nwalters512:
Yeah, that'd be great 👍

jonatanschroeder:
Another option is to merge it now and rely on instructors not syncing their course's view of the image until they have tested it locally or until after the term. Is the cached image on production per course or shared?

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

Issue #1287, opened 2018-09-25
Add integration tests for externally graded questions


nwalters512:
We've been bitten a number of times now by deploys that have broken subtly because of not fully testing live results rendering for externally-graded questions under all possible conditions. We should add full automated integration tests for this.

coatless:
From @mwest1066's https://github.com/PrairieLearn/PrairieLearn/pull/1999#issuecomment-577304869 on unit tests for the external grading feature.

---

Questions have a `startTestQuestion()` method, which calls this code:  https://github.com/PrairieLearn/PrairieLearn/blob/master/lib/question.js#L863

These tests currently get called from two places:

1. The "Test once" button (and "Test 100 times") on the instructor question page: https://github.com/PrairieLearn/PrairieLearn/blob/master/pages/instructorQuestionSettings/instructorQuestionSettings.ejs#L83

2. The test suite, especially from https://github.com/PrairieLearn/PrairieLearn/blob/master/tests/testExampleCourseQuestions.js -- this actually just simulates a button click on "Test once" for each question

We currently comment-out externally-graded questions in `tests/testExampleCourseQuestions.js` and we disable the "Test once" buttons for them.

If we could fix the underlying problem, then it would be easy to reactivate testing for externally-graded questions.

I believe the underlying problem is that in the test code we only call `this._gradeVariantWithClient()` to grade the question: https://github.com/PrairieLearn/PrairieLearn/blob/925fd037e9239b0d15a7cea0aa49c98f5988b6dc/lib/question.js#L822

This works fine for _internally_ graded questions, but _externally_ graded questions need an extra step. If we look at the full version of `gradeVariant()`, we see that it calls `_gradeVariantWithClient()` but then it also does a second step to call `externalGrader.beginGradingJob()`: https://github.com/PrairieLearn/PrairieLearn/blob/925fd037e9239b0d15a7cea0aa49c98f5988b6dc/lib/question.js#L552

This second step is missing from the test-question code.

I think that "all" we need to do is make sure we call `externalGrader.beginGradingJob()` for test questions, and then we need to somehow wait for the external grading job to finish. One slightly hacky but reasonable option would be to just check every second whether the `grading_job` in the DB has been set as `graded`.

jonatanschroeder:
Adding here based on #5997: this CI should also test cases like when external grader is disabled.

PR #5997, opened 2022-07-10
Fix handling of disabled external grader

### LINKED ISSUE #5996: Rendering error for questions with disabled external grading
When a question has `externalGradingOptions.enabled` set to false, upon submission, a rendering error is triggered:

```
Error: Invalid state before render: data.format_errors is not an object: null
```

jonatanschroeder:
Resolves #5996. In essence this resolves two issues:

* The question freeform processing assumed `format_errors` being present meant it was not null, which is incorrect.
* Students were not receiving the intended error message when grading was disabled. Changed to mark the question as not gradable and fixed the format of the object that needed to be sent in this case.

Pending: I'll add a test case to ensure this is handled appropriately in the future.


jonatanschroeder:
I'll push CI tests to #1287 (doesn't make sense to test disabled external grading without testing enabled external grading).

mwest1066 (approved):

PR #5998, opened 2022-07-10
Upgrade Java grader to Java 18 / JUnit Platform 1.8.2

jonatanschroeder:
Similar to the C grader update: Ubuntu 22.04 (latest LTS), Java 18 and JUnit 5 platform 1.8.2.


mwest1066:
Similarly to #5995, should we make this Draft?

mwest1066 (approved):

nwalters512:
Just got a request specifically for Java 18 from someone who's interested in using PrairieLearn. It'll be great to see this merged soon!

nwalters512:
That said, I wonder if we should instead release this as a separate `grader-java-18` image? That might be a better long-term strategy for potentially-breaking changes in graders and workspaces. cc @jonatanschroeder @mwest1066

jonatanschroeder:
I usually just checkout this version and build/push to my own docker hub account when I need the latest version.

nwalters512:
Understood, I'm talking about other folks may not want to build and maintain their own copies of images.

jonatanschroeder:
That said, I'm less concerned with this PR breaking things than I am with the C grader. I think it should be relatively safe to merge now.

jonatanschroeder:
It's also possible for instructors to compile with older versions of Java: https://prairielearn.readthedocs.io/en/latest/java-grader/#changing-compilation-options

nwalters512:
IMO we should think about these images conceptually the way that AWS thinks about [Lambda runtimes](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html), where a given runtime is a language/version combo. We're no longer at a small enough scale where we can risk a breakage, which seems likely during a jump 7 major versions forward. That said, this might just be my unfamiliarity with Java (and getting burned by breaking changes in major-version jumps of Node/Python). if you're 100% confident that this change won't cause breakage, let's get this merged.

PR #5999, opened 2022-07-11
Remove useWorkers config option and related code

nwalters512:
We don't ever want to got back to *not* using persistent Python workers; the performance improvements are too valuable to go without in production.


mwest1066 (commented):
Oops, sorry, I thought I'd approved this but somehow I failed to do so.

PR #6000, opened 2022-07-11
Use generic-pool library to manage Python workers

nwalters512:
This PR replaces our custom pool/queue implementation with [`generic-pool`](https://github.com/coopernurse/node-pool). This allows us to simplify our code and will make error handling easier in some upcoming changes from https://github.com/PrairieLearn/PrairieLearn/pull/1626.


mwest1066:
Please ignore my force-push above. I accidentally committed to this branch and needed to undo the changes.

lib/workers.js
```diff
@@ -1,120 +1,97 @@
...
-        return callback(null);
-      }
-    };
-    testFinished();
+    util.callbackify(module.exports.finishAsync)(callback);
   },
 
   getPythonCaller(callback) {
+    const jobUuid = uuidv4();
+    load.startJob('python_callback_waiting', jobUuid);
```
nwalters512, line 63: There's a subtle change in behavior here. Before, if there was an available worker, we'd avoid recording load for `python_callback_waiting` entirely. Now, however, we'll momentarily report that load and then immediately end it. Based on the following line, that might end up reporting more load than desired:

https://github.com/PrairieLearn/PrairieLearn/blob/94bf9b5d8aaf8be782fbf92f7848efe97abe2e4f/lib/load.js#L73

One possible solution would be to switch to using [`hrtime`](https://nodejs.org/api/process.html#processhrtimebigint) to compute load deltas, which would free us from needing to round up from zero milliseconds. Or, we can just not worry about this?

lib/workers.js
```diff
@@ -1,120 +1,93 @@
...
-    for (let i = 0; i < numWorkers; i++) {
-      const pc = new codeCaller.PythonCaller();
-      pc.number = i;
-      pythonCallers.push(pc);
-      load.startJob('python_worker_idle', pc.uuid);
-      availablePythonCallers.push(pc);
-    }
-    module.exports._warmUpWorkers();
+    const numWorkers = config.workersCount ?? Math.ceil(config.workersPerCpu * os.cpus().length);
+    pool = genericPool.createPool(
```
mwest1066, line 31: If I'm understanding correctly, this [defaults to `idleTimeoutMillis: 30000`](https://github.com/coopernurse/node-pool#createpool). It wasn't quite clear to me whether this means it will auto-destroy workers after this time? It also defaults to `evictionRunIntervalMillis: 0` which means it doesn't run eviction checks, so that might mean that it never auto-destroys.

Do you have any logging to check whether workers are being auto-destroyed after a 30 second timeout?

lib/workers.js
mwest1066: I'm think it should be fine to just not worry about it. Presumably 1 ms is significantly less time than it will take to process the request, so this should be just a small amount of noise on the signal.

mwest1066:
I meant to comment on my review that I love this change! Such a nice cleanup (and improvement) of the existing code.

lib/workers.js
```diff
@@ -1,120 +1,93 @@
...
-        load.startJob('python_callback_waiting', callback.__load_uuid);
-        waitingCallbacks.push(callback);
-      } else {
-        debug(`getPythonCaller(): no workers available, waiting to error`);
-        setTimeout(() => {
-          debug(`getPythonCaller(): no workers available, reporting error`);
-          callback(new Error('Server is overloaded. Please try again.'));
-        }, config.workerOverloadDelayMS);
+
+    pool.acquire().then(
```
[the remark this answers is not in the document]
nwalters512: It is by default FIFO, yes. Check out the `fifo` option in the docs - it defaults to true, but can be changed to false to make it behave like a stack.

lib/workers.js
nwalters512: The docs under https://github.com/coopernurse/node-pool#idle-object-eviction make it clear that idle eviction is disabled by default. To triple check, I added a log to the `destroy` factory function and left a locally-running server idle for several minutes and didn't see any logs from it.

lib/workers.js
```diff
@@ -1,120 +1,93 @@
...
-      load.startJob('python_worker_idle', pc.uuid);
-      availablePythonCallers.push(pc);
-    }
-    module.exports._warmUpWorkers();
+    const numWorkers = config.workersCount ?? Math.ceil(config.workersPerCpu * os.cpus().length);
+    pool = genericPool.createPool(
+      {
+        create: async () => {
+          const pc = new codeCaller.PythonCaller();
+          pc.ensureChild();
```
nwalters512, line 35: @mwest1066 one more thing: I didn't see any easy way to maintain the existing staggered warmup behavior, so I opted to just immediately invoke `ensureChild` here. Do you foresee any problems with that?

mwest1066: This seems fine. I think the staggered behavior was just being very cautious and I don't think it's actually needed.

lib/workers.js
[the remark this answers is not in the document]
mwest1066: I think the `fifo` option is referring to the FIFO behavior of the worker pool, not the waiting acquire queue? That is, `fifo: true` means that the oldest released worker will be the next one used (in the case where we have more workers than acquirers), whereas I'm thinking about the ordering of which waiting acquire call is given the next available worker (in the case where we have more acquirers than workers).

lib/workers.js
mwest1066: Excellent, thanks for checking!

lib/workers.js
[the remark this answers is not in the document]
nwalters512: Ahhhh, ok, you're reading this more carefully than I am! I'm going to assume that the queueing of requested resources is a queue because who in their right mind would build that as a stack???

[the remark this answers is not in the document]
mwest1066: Yeah, I agree that it would be insane to use anything except a FIFO queue for the acquire queue. I'm happy to assume they are being sensible!

lib/workers.js
```diff
@@ -1,213 +1,126 @@
...
+          load.startJob('python_worker_idle', pc.uuid);
+          return pc;
+        },
+        destroy: async (pc) => {
+          pc.done();
```
mwest1066, line 40: Do we need a `load.endJob('python_worker_idle', pc.uuid);` here?

mwest1066 (approved):

lib/workers.js
nwalters512: Good call! Yes, I'll add that.

nwalters512: Done in aa6ad64.

PR #6001, opened 2022-07-12
Add gcc g++ compiler to plbase image

nwalters512:
Doing this independently of #5993 so that we can build and publish a new base image. It looks as though the `pltest` image used in CI is built from the *latest* version of `prairielearn/plbase` from Dockerhub, not the version of `plbase` that we actually build during CI. That's a problem we should solve, but not one I can right now.

### DIFF
--- a/images/plbase/plbase-install.sh
+++ b/images/plbase/plbase-install.sh
@@ -11,6 +11,9 @@ amazon-linux-extras install -y \
     postgresql11 \
     redis4.0
 
+# Notes:
+# - `libjpeg-devel` is needed by the Pillow package
+# - `gcc-c++` is needed to build the native bindings in `packages/bind-mount`
 yum -y install \
     postgresql-server \
     postgresql-contrib \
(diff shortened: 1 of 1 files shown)


images/plbase/plbase-install.sh
```diff
@@ -29,7 +32,8 @@ yum -y install \
     git \
     graphviz \
     graphviz-devel \
-    libjpeg-devel # Needed by the Pillow package
+    libjpeg-devel \
+    gcc-c++
```
mwest1066, line 36: One [hack](https://stackoverflow.com/a/12797512) for inline comments is like this:
```suggestion
    libjpeg-devel `# for the python Pillow package` \
    gcc-c++ `# for building the native bindings in packages/bind-mount`
```

mwest1066 (approved):

images/plbase/plbase-install.sh
nwalters512: I'm going to land this as-is so I can make forward progress without waiting for another base image build in CI. We can tweak later if desired!

mwest1066: Yeah, I'm fine either way -- inline is easier to keep up to date, but I worry that this is maybe a bit too devious 😄

PR #6003, opened 2022-07-12
Remove hardcoded references to Node 14

nwalters512:
This is a small followup to #5994. I just grepped for `node.*14` and replaced what I found!


mwest1066 (approved):
Thanks for thinking to do this! Especially the CI change is a good find.

PR #6016, opened 2022-07-15
Show correct answer panel after scoring 100 percent on single variant HW question

### LINKED ISSUE #5914: Single variant answer panel visibility
@ffund was in office hours today asking when students see the answer panel in single variant questions.

She would like the functionality where students can see the correct answer panel after they get the question right (or in case of a homework, after having earned all of the available points).

Something like being able to see a closed type:exam assessment, but also for homeworks.

From the Slack discussion, she uncovered issues #895 #2083 and #3191 but we weren't sure what the expected behavior was

ffund:
Fixes #5914

The correct answer panel is the standard place to share extra detail and explanation with students.

Currently, on Homework-type assessments, students only see the correct answer panel after they've exhausted all attempts at a variant. Since students get unlimited attempts on single variant questions in Homework-type assessments, they will never see the correct answer panel and the accompanying explanation.

With this PR, students will see the correct answer panel on a single variant Homework question after scoring 100%. 

(note: they are not restricted from further submissions, so they can still try out other possible answers, e.g. in the case where there is more than one correct answer.)

### DIFF
--- a/lib/question.js
+++ b/lib/question.js
@@ -1474,6 +1474,9 @@ module.exports = {
           locals.variantAttemptsLeft = assessment_question.tries_per_variant - variant.num_tries;
           locals.variantAttemptsTotal = assessment_question.tries_per_variant;
         }
+        if (question.single_variant && instance_question.score_perc >= 100.0) {
+          locals.showTrueAnswer = true;
+        }
       }
       if (assessment.type === 'Exam') {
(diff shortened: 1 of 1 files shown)


ffund:
@mwest1066 per our discussion in office hours.

nwalters512 (approved):

PR #6022, opened 2022-07-19
Reorganize code callers

nwalters512:
This is a followup to #1626 that does a bit of cleanup and big fixing:

- Fixes issue where errors logged from code callers would fail with the following ([wtf winston](https://github.com/winstonjs/winston/blob/3998df01449da35a60cf4b90730b91f4b2466bc5/lib/winston/create-logger.js#L73)):
  ```
  TypeError: self.log is not a function
      at Object.DerivedLogger.<computed> [as errorLogger] (/Users/nathan/git/PrairieLearn/node_modules/winston/lib/winston/create-logger.js:96:19)
      at PythonCaller._logError (/Users/nathan/git/PrairieLearn/lib/code-caller/code-caller-python.js:571:18)
      at PythonCaller._handleStderrData (/Users/nathan/git/PrairieLearn/lib/code-caller/code-caller-python.js:353:12)
  ```
- Removes all callback-based functions from the code callers in favor of async ones.
- Moves all code caller code into `lib/code-caller`
- Renames `code-caller-python` to `code-caller-native` and `code-caller-docker` to `code-caller-container`
- Renames `withPythonCaller` to `withCodeCaller`.


lib/code-caller/code-caller-native.js
```diff
@@ -44,61 +43,7 @@ const EXITED = Symbol('EXITED');
 
 /** @typedef {Error & { data?: ErrorData }} CodeCallerError */
 
-/*
-  Usage
```
nwalters512, line 48: This information was all out of date, and (IMO) is no longer super useful. Happy to revert this and edit it to be up to date if that's preferred.

lib/code-caller/index.js
```diff
@@ -28,27 +28,31 @@ const { PythonCaller } = require('./code-caller-python');
...
     debug('init()');
 
     const { workersExecutionMode } = config;
     if (workersExecutionMode !== 'container' && workersExecutionMode !== 'native') {
       throw new Error(`unknown config.workersExecutionMode: ${workersExecutionMode}`);
     }
 
+    if (config.workersExecutionMode === 'container') {
+      await initCodeCallerDocker();
+    }
```
nwalters512, line 41: I moved this from `server.js` to here so that the server startup routine doesn't need to be directly aware of execution modes, etc.

lib/code-caller/index.js
```diff
@@ -28,27 +28,31 @@ const { PythonCaller } = require('./code-caller-python');
...
+
     const numWorkers = config.workersCount ?? Math.ceil(config.workersPerCpu * os.cpus().length);
     pool = genericPool.createPool(
       {
         create: async () => {
           let codeCallerOptions = {
             questionTimeoutMilliseconds: config.questionTimeoutMilliseconds,
-            errorLogger: logger.error,
+            errorLogger: logger.error.bind(logger),
```
nwalters512, line 49: Grrrr: https://github.com/winstonjs/winston/blob/3998df01449da35a60cf4b90730b91f4b2466bc5/lib/winston/create-logger.js#L73

mwest1066: Reason number 1000 to hate Winston 🤯

lib/deferred.js
```diff
@@ -0,0 +1,27 @@
...
+ * @typedef {Object} DeferredPromise
+ * @template T
+ * @property {Promise<T>} promise
+ * @property {(resolve: (value: T) => void) => void} resolve
+ * @property {(reject: (reason: any) => void) => void} reject
+ */
+
+/**
+ * Returns an object that can be used to resolve or reject a promise from
+ * the outside.
```
mwest1066, line 11: Cute!

server.js
```diff
@@ -1883,14 +1882,8 @@ if (config.startServer) {
...
       },
-      async () => {
-        if (config.workersExecutionMode === 'container') {
-          await codeCallerDocker.init();
-        }
-      },
-      function (callback) {
-        codeCallers.init();
-        callback(null);
+      async function () {
```
mwest1066, line 1885: Nit for consistency:
```suggestion
      async () => {
```

mwest1066 (approved):

lib/code-caller/index.js
mwest1066: Much nicer!

lib/code-caller/code-caller-native.js
mwest1066: I agree it doesn't seem necessary anymore.

Issue #6025, opened 2022-07-20
Add dockerized production documentation


Prostigus:

firasm:
this can probably be closed now right @Prostigus

PR #6027, opened 2022-07-20
Add docs for running in production with Docker Compose

### LINKED ISSUE #6026: Add docker-compose.yml file for running in production

Prostigus:
I've added in a new docker-compose file for users that would like to run the system in production. Documentation on how to use this will be added soon.
Closes #6026


Prostigus:

Prostigus:
recheck

nwalters512:
👋 I'd prefer to land the Dockerfile and any related documentation in on go. Could you add that to this PR?

Prostigus:
Understood, it's still a work in progress as I'm not sure about how to implement all the different authentications. I'll get right to it!

Prostigus:
Add docs pertaining to production docker file closes #6025

Prostigus:
@nwalters512 I've added docs to this PR and it's ready for review.

docker-compose-production.yml
```diff
@@ -0,0 +1,17 @@
+version: '3.8'
+services:
+  pl:
+    build:
+      context: .
+    image: prairielearn/prairielearn:local
+    ports:
+      - 3000:3000
+    volumes:
```
nwalters512, line 9: Should we put the Postgres state on a volume so it isn't lost if the container is destroyed?

docs/authentication.md
```diff
@@ -0,0 +1,47 @@
+# Authentication
```
nwalters512, line 1: This and the other page should be included in our MkDocs config: https://github.com/PrairieLearn/PrairieLearn/blob/master/mkdocs.yml. Now that we have several pages, I'd recommend introducing a new section, something like "Running in production", and placing these and the existing "Running in production" page under it.

docs/authentication.md
```diff
@@ -0,0 +1,47 @@
+# Authentication
+PrairieLearn currently has a few ways to do user authentication. 
+- [Google OAuth 2](#google-oauth-2)
+- [Azure](#azure)
+- [Shibboleth](#shibboleth)
+- [LTI](./courseInstance.md#lti-overview)
+
+
+## Google OAuth 2
+> Take a look in `pages/authLoginOAuth2/authLoginOAuth2.js` and `pages/authCallbackOAuth2/authCallbackOAuth2.js` to get an idea of what variables you need to modify.
```
nwalters512, line 10: This isn't semantically a block quote; I'd leave this as plain text. Ditto for elsewhere.

docs/authentication.md
```diff
@@ -0,0 +1,47 @@
...
+  - Proceed to `Credentials` and create a new `OAuth client ID`.
+  - Select `Web application`.
+  - Under Authorized JavaScript origins, click `ADD URI` and add in your domain address.
+  - Under Authorized redirect URIs, click `ADD URI` and add `https://yourdomain/pl/oauth2callback` which is the route to the Google OAuth callback.
+  - Click `Create` which will give you a `Client ID` and a `Client Secret`. **WARNING: DO NOT UPLOAD THESE KEYS ANYWHERE**
+
+  
+Now add the keys to `config.json`:
+
+```sh
```
nwalters512, line 26:
```suggestion
```json

```

docs/authentication.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Azure
+> Take a look in `pages/authLoginAzure/authLoginAzure.js` and `pages/authCallbackAzure/authCallbackAzure.js` to get an idea of what variables you need to modify.
+
+## Shibboleth
+> Take a look in `pages/authCallbackShib/authCallbackShib.js` to get a sense of what headers you will need your reverse proxy to pass.
+
+## LTI
+
+Look here to read up more on [LTI](./courseInstance.md#lti-overview).
```
nwalters512, line 46:
```suggestion
Check out the [course instance LTI docs](./courseInstance.md#lti-overview) to learn more about LTI.

```

docs/running-in-production-dockerized.md
```diff
@@ -0,0 +1,100 @@
+# Running in Production Dockerized
```
nwalters512, line 1: There's a lot of duplication between this and the existing "Running in production" page. I'd recommend leaving most of this information in the existing page and moving this to a separate page that's focused on specifically what's necessary for Docker Compose.

docker-compose-production.yml
Prostigus: I agree, this would be the best way to implement the database. How do you recommend we approach this as we need to populate the DB before adding it to the docker volume.

nwalters512: Hmm, good question. We may just have to have folks manually create the DB and run the migrations. Alternatively, we could avoid using the version of Postgres built into the Docker container and use a separate [`postgres`](https://hub.docker.com/_/postgres) container, set up the DB and credentials via environment variables, and then have folks set `"runMigrations": true` in their config file.

Do you know how other Docker Compose setups handle this problem?

Prostigus: I am currently doing research into this and will have to get back to you. From what I've seen so far, Docker Compose setups have the database in another container. 

I will try both implantations very soon and then we can discuss about the pros and cons of each method.

docs/authentication.md
Prostigus: Add new section and moved docs

docs/authentication.md
Prostigus: Done

docs/authentication.md
Prostigus: Done

docs/authentication.md
Prostigus: Done

docs/running-in-production-dockerized.md
```diff
@@ -0,0 +1,88 @@
...
+{
+"serverCanonicalHost": "https://yourdomain"
+}
+```
+
+## Authentication
+
+PrairieLearn currently has 4 ways to do user authentication. Read more at [authentication](./authentication.md).
+
+## Admin User
```
nwalters512, line 42: This core information here (adding an admin user) is broadly useful to folks whether or not they're running PL inside of Docker. I'd like to see that info moves somewhere else that we can link to from both the Docker and non-Docker version of this documentation.

docs/running-in-production-dockerized.md
```diff
@@ -0,0 +1,88 @@
...
+
+A reverse proxy can be implemented using something like [Apache](https://httpd.apache.org/docs/2.4/howto/reverse_proxy.html) or [NGINX](https://docs.nginx.com/nginx/admin-guide/web-server/reverse-proxy/).
+
+In `config.json` to configure your domain add:
+
+```json
+{
+"serverCanonicalHost": "https://yourdomain"
+}
+```
```
nwalters512, line 36: This seems relevant to non-Docker installations as well. Let's move this to the primary "Running in production" docs.

mkdocs.yml
```diff
@@ -56,6 +55,10 @@ nav:
   - 'Code execution': 'codeExecution.md'
   - 'Docker packaging': 'docker.md'
   - 'Contributing to PL development': 'contributing.md'
+- Running in Production:
+  - 'Running in production': 'running-in-production.md'
```
nwalters512, line 59: Let's call this page "Setup". Could you also move these and the other pages to a `docs/running-in-production` directory to keep them separated from the rest of the docs?

mkdocs.yml
```diff
@@ -56,6 +55,10 @@ nav:
...
+  - 'Running in production dockerized': 'running-in-production-dockerized.md'
```
nwalters512, line 60: Let's call this "Using Docker Compose".

docs/running-in-production-dockerized.md
```diff
@@ -0,0 +1,88 @@
+# Running in Production Dockerized
+
+As the PrairieLearn source code is publicly-available, it's possible to run PrairieLearn on your own infrastructure. Running a single instance of the PrairieLearn server may be appropriate for tens or hundreds of total users, and a number of universities have done this successfully.
```
nwalters512, line 3: Rather than repeating this here, I'd encourage you to leave this in the main setup page and just cross-link these two pages.

nwalters512 (commented):
I'm still trying to figure out the right information architecture for this; thanks for bearing with me through multiple rounds of reviews!

You'll need to run `make format` locally to reformat your Markdown files with Prettier.

docker-compose-production.yml
Prostigus: @nwalters512 For some reason on windows, when adding a volume that points to the database on the docker container it gets permission errors. I think this has to do with the way windows mounts folders.

It took me a bit but I think I've got a solution that would work for all systems. We can use docker volumes and this would give us a persistent database. What do you think of this solution?

nwalters512: Candidly, I don't think many folks are going to be running PrairieLearn in production on Windows machines, so I'm not particularly concerned with that use case, especially if it adds any complexity. If there's a very straightforward way to make it work though, let's check it out!

@firasm that works because we pre-populate the database when we build the `prairielearn/prairielearn` image. As soon as we try to mount something over the Postgres data directory, we'd lose that.

Prostigus: The volume approach is the most straight forward and bulletproof in my opinion, no room for user error. 

You can test it out by changing your docker-file to look like this:
```sh
version: '3.8'
services:
  pl:
    build:
      context: .
    image: prairielearn/prairielearn:local
    ports:
      - 3000:3000
    volumes:
      - psql:/var/postgres
      - /var/run/docker.sock:/var/run/docker.sock
      - ${HOME}/pl_ag_jobs:/jobs
      
    container_name: pl
    environment:
      - HOST_JOBS_DIR=${HOME}/pl_ag_jobs
      - NODE_ENV=production
      
volumes:
   psql:
```

This will create a docker volume `psql` and use that in place of all the files in the database. A user can always backup a volume when they like.

docs/running-in-production-dockerized.md
Prostigus: Moved to file `admin-user.md`

docs/running-in-production-dockerized.md
Prostigus: moved to running in production docs

mkdocs.yml
Prostigus: Moved to directory and renamed to setup

mkdocs.yml
Prostigus: Renamed

docs/running-in-production-dockerized.md
Prostigus: Remove repeated section

Prostigus:
@nwalters512 I've updated the documentation and also included a new way to launch PrairieLearn in vanilla mode. This would be good for people just wanting to get the project up and running in production mode.

docker-compose-production.yml
Prostigus: After doing further research I think it is best to let the user decide how they would like to have the database state as a volume on docker. I've added to the docs the way you could do this for both Linux/Mac and for Windows.

docs/running-in-production/running-in-production.md
```diff
@@ -2,9 +2,13 @@
 
 As the PrairieLearn source code is publicly-available, it's possible to run PrairieLearn on your own infrastructure. Running a single instance of the PrairieLearn server may be appropriate for tens or hundreds of total users, and a number of universities have done this successfully.
 
+## Running in Production Dockerized
+
+PrairieLearn can also be run in production mode in a docker container [using docker compose](./running-in-production-dockerized.md).
```
nwalters512, line 7:
```suggestion
PrairieLearn can also be run in a Docker container [using Docker Compose](./running-in-production-dockerized.md).
```

Dockerfile-alternate
```diff
@@ -0,0 +1,3 @@
+FROM prairielearn/prairielearn
```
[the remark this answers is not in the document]
Prostigus: This removes all the installation stuff and uses the latest PrairieLearn image directly, speeds up starting the production server by quite a bit. If we don't want this I can remove it.

docs/running-in-production/running-in-production-dockerized.md
```diff
@@ -0,0 +1,123 @@
...
+
+```
+CONTAINER ID  IMAGE                      COMMAND              CREATED      STATUS      PORTS                   NAMES
+e0f522f41ea4  prairielearn/prairielearn  "/bin/sh -c /Prai…"  2 hours ago  Up 2 hours  0.0.0.0:3000->3000/tcp  upbeat_roentgen
+```
+
+- Then copy out postgres contents to where you would like the database to exist
+
+```sh
+docker cp <containerId>:/var/postgres /host/path/target
```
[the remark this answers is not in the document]
Prostigus: Ok, we can do it this way, should I include how to do this in the documentation or just leave these changes in `docker-compose-production.yml`?

mkdocs.yml
```diff
@@ -56,6 +55,11 @@ nav:
   - 'Code execution': 'codeExecution.md'
   - 'Docker packaging': 'docker.md'
   - 'Contributing to PL development': 'contributing.md'
+- Running in Production:
+  - 'Setup': 'running-in-production/running-in-production.md'
+  - 'Using Docker Compose': 'running-in-production/running-in-production-dockerized.md'
```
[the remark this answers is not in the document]
Prostigus: renamed

docs/running-in-production/admin-user.md
```diff
@@ -0,0 +1,48 @@
...
+```
+CONTAINER ID  IMAGE                      COMMAND              CREATED      STATUS      PORTS                   NAMES
+e0f522f41ea4  prairielearn/prairielearn  "/bin/sh -c /Prai…"  2 hours ago  Up 2 hours  0.0.0.0:3000->3000/tcp  upbeat_roentgen
+```
+
+- Open a shell in your PrairieLearn container by running
+
+```sh
+docker exec -it CONTAINER_NAME /bin/bash
+```
```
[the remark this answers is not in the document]
Prostigus: moved

docs/running-in-production/admin-user.md
```diff
@@ -0,0 +1,48 @@
...
+
+- Connect to the postgres database by running
+
+```sh
+psql postgres
+```
+
+- Check the users table by running
+
+```sh
```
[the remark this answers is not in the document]
Prostigus: Changed

docs/running-in-production/admin-user.md
```diff
@@ -0,0 +1,48 @@
...
+SELECT * FROM users;
+```
+
+Which will display a table of users in the database:
+
+```sh
```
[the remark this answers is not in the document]
Prostigus: Changed

docs/running-in-production/admin-user.md
```diff
@@ -0,0 +1,48 @@
...
+ user_id |       uid        |    uin    |   name   | lti_course_instance_id | lti_user_id | lti_context_id | institution_id | deleted_at
+---------+------------------+-----------+----------+------------------------+-------------+----------------+----------------+------------
+       1 | dev@illinois.edu | 000000000 | Dev User |                        |             |                |              1 |
+```
+
+Add the desired user to the administrators table by running
+
+```sh
```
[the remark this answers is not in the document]
Prostigus: Changed

docs/running-in-production/running-in-production-dockerized.md
```diff
@@ -0,0 +1,123 @@
+# Using Docker Compose
+
+## Getting started
+
+Follow the steps to [install PrairieLearn with local source code](../installingLocal.md). Then run this command in the root folder:
+
+```sh
+docker-compose -f docker-compose-production.yml up
```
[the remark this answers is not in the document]
Prostigus: Using compose v2

docs/running-in-production/running-in-production.md
Prostigus: Fixed

docs/running-in-production/running-in-production-dockerized.md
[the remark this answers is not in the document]
Prostigus: With further reconsideration, it would be best to leave the changes in the docker-file for less user error.

Dockerfile-alternate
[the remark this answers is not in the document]
nwalters512: Could we just remove the `build` section from the `pl` service in the Docker Compose config and instead rely on the `prairielearn/prairielearn` image that's published to Docker Hub? I think that should be the default case.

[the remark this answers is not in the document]
echuber2: Independent users of PL may want to pin their server version (suppose PL makes a major breaking update or something, and they're not ready for it). If there's a rebuild involved here they can specify a different upstream image based on its hash or version tag. Would that be a reasonable use case?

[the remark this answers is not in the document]
echuber2: On the other hand, this may encourage someone to carelessly downgrade the image, and I don't know if PL would handle migrations or DB breakage going in the other direction.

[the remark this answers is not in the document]
Prostigus: @echuber2 I agree with this, also if they would like to install some new dependencies on top of the PrairieLearn image, they could do so easily through the `Docker-alternate` file.

i.e. adding the line to the file:
`RUN pip install [new dependency]`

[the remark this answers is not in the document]
nwalters512: @echuber2 if pinning is something that folks want support for, we can start publishing images tagged by the Git hash of the commit it's built from. And if we do that, we don't need a custom Dockerfile, they could just specify: `image: prairielearn/prairielearn:GIT_HASH_HERE` in their Docker Compose config.

@Prostigus candidly, I don't think we're interested in officially supporting the use case of including additional dependencies. That would make it more difficult for anyone trying to do local course development, i.e. they'd have to also build that custom image locally instead of just doing `docker run prairielearn/prairielearn`. Of course, we can't stop you from doing that if you want to, but I don't want that potential complexity to leak into our official docs.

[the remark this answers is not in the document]
echuber2: I'd agree with tagging the images with the commit. (I've had to pin versions of other upstream projects before when they did something breaking.)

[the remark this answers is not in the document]
echuber2: What would happen if somebody ran the latest build of PL and it did migrations, then they tried to roll back to a pinned previous version? Would PL be able to handle it gracefully?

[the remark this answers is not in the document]
nwalters512: We don't test rollbacks of code after applying database migrations, nor do we make any guarantees about what would happen if you tried that.

You're welcome to open an issue or PR for tagging images, but you can even pin to specific versions without making any changes to our tagging strategy, e.g. `prairielearn/prairielearn@sha256:0238c426cfdc347d972b34eafef584207cb7a1b1ffceb72885e1ad33c5807f62`. You can find the image SHA with `docker inspect prairielearn/prairielearn`.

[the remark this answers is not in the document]
Prostigus: Agreed, I've removed `build` and changed `prairielearn/prairielearn:local` to `prairielearn/prairielearn:latest`

Prostigus:
@nwalters512 Updated the docker production docs with some new stuff that I thought was necessary in getting the server up and running. Also removed the alternate file and docs pertaining to that. 

Thanks for being so patient with me, I feel like I'm learning something new everyday with dockerizing this project that the docs keep getting bigger.

nwalters512 (approved):
@Prostigus and thanks to *you* for working with us to figure out how to get this all documented! Once CI passes, I'll merge this in :shipit:

PR #6028, opened 2022-07-20
Install latest version of cla-assistant/github-action

nwalters512:
This pulls in the latest changes, including those from https://github.com/contributor-assistant/github-action/pull/104 that should improve the experience for folks signing the CLA for the first time!


mwest1066:
So nice to see upstream fixes flowing back!

PR #6034, opened 2022-07-22
Remove centos7-ocaml and centos7-cs225 images

nwalters512:
While these images are still in use by the courses, they haven't changed in 4 years. In addition, the OCaml image contains grading code that is incompatible with the way that external grading now works.

Our best practice is now that courses should maintain their own images. If these courses ever want new code in external graders, the Dockerfiles will be available in the Git history for them to reference when creating a new image.


mwest1066 (approved):
Just to double-confirm, this won't remove the actual images themselves from docker hub, right? So there won't be any impact to courses using these images?

nwalters512:
Correct, this will not delete anything!

mwest1066:
> Correct, this will not delete anything from Docker Hub!

Excellent, thanks!

Issue #6031, opened 2022-07-21
"History" of awarded points should show all attempts


ffund:
Currently, the "history" of awarded points in the Homework view (on the assessment page and the question page) only includes variants where 100% credit was awarded, plus the last variant where partial credit was awarded (unless they earned 100% on the last attempt...)

The "history" should show all attempted variants, including zero-credit and partial-credit attempts, because:

* the current behavior is unintuitive and undocumented (as far as I can tell)
* greater transparency for current and future non-trivial question grading schemes
* will allow the badges in the "history" section to be used for #5763 (then students will be able to e.g. click on an attempt where they earned zero credit to review it with a TA)
* provides some slight motivation to make a serious attempt on the first try, instead of submitting random values and clicking through to see a solved example (students will want to have that "all green" assessment page)
* students will see if they have many failed attempts at a question. This could encourage them to seek help or go review their notes before re-attempting, instead of trying to brute-force their way to a correct answer.
* when students complain to their instructor that "I tried this question a million times and I still can't get it" they will be able to give a more precise estimate of the number of failed attempts (it was actually only 80 attempts, etc.)

Disadvantages: 

* it would be discouraging to see lots of failed attempts on the assessment page

ffund:
My alternative proposal if ^ seems too discouraging:

- rename "History" to "Points History" on assessment and question page
- on question page only, in the points box on the right side, add an "Attempt History" list under "Points History"

ffund:
Following some discussion with Matt, we decided that on both the assessment page and the question page we should show all of the attempts (*instead* of the current "awarded points" list, not in addition to.)

To avoid being too discouraging, we will use the non-judgmental "secondary" badge color instead of the success/danger/warning colors:

![image](https://user-images.githubusercontent.com/2316553/182611115-bb68a177-eafe-4fa8-93e4-1161d81b3838.png)

Also, if there is an unreasonable number of attempts, we should show a smaller number of badges with a `...` or equivalent to reveal more.

mwest1066:
> To avoid being too discouraging, we will use the non-judgmental "secondary" badge color instead of the success/danger/warning colors.

I love this idea!

Issue #5763, opened 2022-05-13
Let students see their own previous variants by clicking on score badge


ffund:
Currently, if a student submits a (not single variant) HW question, and has no more tries for that variant, once they click away they cannot see that submission. When they open the question again, they see a new variant.

However, students like to see their own previous submission, either to help with exam review, or to revisit incorrect attempts.

It would be cool if these score badges were click-able, and let students review the variant + submission associated with the score:

![image](https://user-images.githubusercontent.com/2316553/168375234-a5e797f1-5181-4932-9276-923399c68c9c.png)

(on both assessment overview and question pages.)

jonatanschroeder:
Related: #3158, #29

glherman:
My students have been asking for this functionality this semester and recommend this implementation

Issue #3158, opened 2020-09-21
Add a way for students to see their previous submissions in Homework-type assessments


jonatanschroeder:
Please add a way for students to see their own history of submissions in Homework-type assessments. This is particularly relevant in file-editor and file-upload style questions (like code).

mwest1066:
For questions like the ones you reference, you might want to set `"singleVariant": true` in the question `info.json`. This will show all previous submissions.

On the question page, PL shows all the submissions for the current question variant (the particular random instance of the question). On exams there is only one variant per question, which is why exams show all previous submissions. Homeworks normally generate new variants for every attempt, which is why they don't show previous submissions. A homework question that has `"singleVariant": true` will behave like an example question in so far as submission-display is concerned.

jonatanschroeder:
Thanks, that solves most common issues, but we still have coding questions with multiple variants that we want to keep, and give students a chance to practice with those variants.

mwest1066:
Yep, understood. This is just a work-around until we can address all the cases properly.

I think this issue is actually a duplicate of #29 (ancient issue!) 😄 -- I'm going to leave this one open as well because it has valuable additional context.

Issue #29, opened 2014-05-22
Add historical "activity" view to client


mwest1066:
Add a client page that shows all submissions for the current user.

mwest1066:
This would be helpful for students doing Homework assessments. Exam assessments already show all submissions to a question, but for Homeworks the history information vanishes.

It would be great to add a new page called `pages/studentInstanceQuestionHistory` that shows a list of all submissions to a particular instance question. This could be linked from a "History" button in the sidebar of Homework questions.

The list of submissions should not show the submissions themselves, but rather should just be a big table listing the time of submission, the score, and a link to the question page with the `?variant_id=` query parameter set so that it will show that exact variant.

PR #6105, opened 2022-08-03
Show list of variants with links instead of awarded points history

ffund:
Currently, if a student submits a (not single variant) HW question, and has no more tries for that variant, once they click away they cannot see that submission. When they open the question again, they see a new variant. However, students like to see their own previous submission, either to help with exam review, or to revisit incorrect attempts.

This PR will show students a list of clickable previous variants in place of the current "awarded points" history list.

(Will resolve https://github.com/PrairieLearn/PrairieLearn/issues/6031, resolve https://github.com/PrairieLearn/PrairieLearn/issues/5763, resolve https://github.com/PrairieLearn/PrairieLearn/issues/3158, and resolve https://github.com/PrairieLearn/PrairieLearn/issues/29)


ffund:
To do:

[ ] Show only variants that have been graded, instead of all variants
[ ] Show point value instead of attempt number
[ ] If there have been an unreasonable number of attempts, only show some + click to reveal the rest to avoid UI madness

jonatanschroeder:
@ffund just FYI, in previous discussion I mentioned about doing this in #5909 , but this PR is now pretty huge with other aspects, so it probably makes sense to make it separately. So if you still want to proceed with this PR go for it. Sorry for the confusion.

ffund:
> @ffund just FYI, in previous discussion I mentioned about doing this in #5909 , but this PR is now pretty huge with other aspects, so it probably makes sense to make it separately. So if you still want to proceed with this PR go for it. Sorry for the confusion.

Great, thanks for the update!

If you have any suggestion on the discussion post https://github.com/PrairieLearn/PrairieLearn/discussions/6106  I'd love to hear them :)

ffund:
Show badges (with non-judgmental gray color) that link to previous variants on homework assessments.

In the assessment overview:

prev-variants-hw-page

On the question page:

prev-variants-question-page

nwalters512 (approved):
I tested this out, it's looking good to me! @mwest1066 did you want to take a final look before we land this?

pages/partials/questionAwardedPoints.ejs
```diff
@@ -1,11 +1,9 @@
...
-        <% if (awarded_points[i] > 0) { %>
-            <span class="badge badge-warning"><%= getStringFromFloat(awarded_points[i]) %></span>
-        <% } %>
-    <% } else { %>
-        <span class="badge badge-success"><%= getStringFromFloat(awarded_points[i]) %></span>
+<% if (previous_variants) { %>
+    <% for (var i = 0; i < previous_variants.length; i++) { %>
+        <span class="badge badge-secondary">
+            <a class="text-white" href="<%= urlPrefix %>/instance_question/<%= instance_question_id %>/?variant_id=<%= previous_variants[i].variant_id %>"><%= Math.round(previous_variants[i].max_submission_score * 100) + "%"  %></a> 
+        </span>
```
mwest1066, line 7: As we discussed, I'm a bit worried about what happens when a student builds up a lot of previous variants. Here's an attempt to fix this by only showing the most recent 10 variants and providing a "..." clickable link that will reveal the earlier ones. Thoughts?

image

New `questionAwardedPoints.ejs`:
```
<% include('pointsFormatter'); %>

<% const MAX_DISPLAYED_VARIANTS = 10; %>

<% if (previous_variants) { %>
    <%
        const collapseId = `variants-points-collapse-${instance_question_id}`;
        const collapseButtonId = `variants-points-collapse-button-${instance_question_id}`;
        let startHiddenIndex = null;
        let endHiddenIndex = null;
    %>
    <% if (previous_variants.length > MAX_DISPLAYED_VARIANTS) { %>
        <a id="<%= collapseButtonId %>" role="button" onclick="$(`#<%= collapseId %>`).css('display','inline'); $(`#<%= collapseButtonId %>`).css('display','none');">&hellip;</a>
        <% startHiddenIndex = 0; %>
        <% endHiddenIndex = previous_variants.length - MAX_DISPLAYED_VARIANTS; %>
    <% } %>
    <% for (var i = 0; i < previous_variants.length; i++) { %>
        <% if (i == startHiddenIndex) { %>
            <span id="<%= collapseId %>" style="display:none">
        <% } %>
        <% if (i == endHiddenIndex) { %>
            </span>
        <% } %>
        <span class="badge badge-secondary">
            <a class="text-white" href="<%= urlPrefix %>/instance_question/<%= instance_question_id %>/?variant_id=<%= previous_variants[i].variant_id %>"><%= Math.round(previous_variants[i].max_submission_score * 100) + "%"  %></a> 
        </span>
    <% } %>
<% } %>
```

pages/partials/questionScorePanel.ejs
```diff
@@ -10,8 +10,8 @@
       <% } %>
       <% if (assessment.type == 'Homework') { %>
       <tr>
-        <td>History:</td>
-        <td><%- include('../partials/questionAwardedPoints', {awarded_points: instance_question.variants_points_list, init_points: assessment_question.init_points});  %></td>
+        <td>Previous variants:</td>
+        <td><%- include('../partials/questionAwardedPoints', {awarded_points: instance_question.variants_points_list, init_points: assessment_question.init_points, instance_question_id: instance_question.id, previous_variants: instance_question_info.previous_variants});  %></td>
```
mwest1066, line 14: This ends up being a really big vertical list. I'm wondering whether we should render this as a single cell using `colspan="2"` to let it use the full table width for wrapping?

Current style:
image

Proposed change:
```
      <% if (assessment.type == 'Homework') { %>
      <tr>
        <td colspan="2" style="white-space: normal;">
          Previous variants:
          <%- include('../partials/questionAwardedPoints', {awarded_points: instance_question.variants_points_list, init_points: assessment_question.init_points, instance_question_id: instance_question.id, previous_variants: instance_question_info.previous_variants});  %>
        </td>
      </tr>
      <% } %>
```

With the proposed change:
image

pages/studentAssessmentInstanceHomework/studentAssessmentInstanceHomework.ejs
```diff
@@ -66,22 +66,22 @@
...
-              <th rowspan="2">Question</th>
-              <th class="text-center" colspan="3">Auto-grading</th>
-              <th class="text-center" rowspan="2">Manual grading points</th>
-              <th class="text-center" rowspan="2">Total points</th>
+              <th>Question</th>
+              <th class="text-center">Value</th>
+              <th class="text-center">Previous variants</th>
+              <th class="text-center">Auto-grading points</th>
+              <th class="text-center">Manual grading points</th>
+              <th class="text-center">Total points</th>
```
mwest1066, line 74: Why don't we need the `colspan` and `rowspan` attributes anymore? Or said another way, why did we used to need them?

SethPoulsen:
What are we waiting on for this to be merged? I just had a student ask for it so I came to find this PR.

SethPoulsen:
Do we just need to implement what @mwest1066 suggested in his latest comment then it will be good to go?

mwest1066:
Yep, if someone could just implement my proposed changes then we can merge this. I meant to do it but it keeps failing to get to the top of my ToDo list 😭

jonatanschroeder:
> Yep, if someone could just implement my proposed changes then we can merge this. 

Not sure which is more work: implementing the changes or handling the merge conflicts that accumulated over time...

pages/partials/questionAwardedPoints.ejs
mwest1066: Done in 94d9da247170bc5fee7b100a2a0cd39b3d11bcee

pages/partials/questionScorePanel.ejs
mwest1066: Done in 94d9da247170bc5fee7b100a2a0cd39b3d11bcee

pages/studentAssessmentInstanceHomework/studentAssessmentInstanceHomework.ejs
mwest1066: This no longer seems to be relevant in the current diff.

mwest1066 (approved):

apps/prairielearn/src/pages/partials/questionAwardedPoints.ejs
```diff
@@ -1,11 +1,28 @@
...
+            <span id="<%= collapseId %>" style="display:none">
+        <% } %>
+        <% if (i == endHiddenIndex) { %>
+            </span>
         <% } %>
-    <% } else { %>
-        <span class="badge badge-success"><%= getStringFromFloat(awarded_points[i]) %></span>
+        <span class="badge badge-secondary">
+            <a class="text-white" href="<%= urlPrefix %>/instance_question/<%= instance_question_id %>/?variant_id=<%= previous_variants[i].variant_id %>"><%= Math.floor(previous_variants[i].max_submission_score * 100) + "%"  %></a> 
+        </span>
```
jonatanschroeder, line 26: I believe the `a` should be outside the `span`, otherwise only the text of the badge is a link, not the badge itself.

apps/prairielearn/src/pages/partials/questionScorePanel.ejs
```diff
@@ -14,8 +14,10 @@
       <% } %>
       <% if (assessment.type == 'Homework') { %>
       <tr>
-        <td>History:</td>
-        <td><%- include('../partials/questionAwardedPoints', {awarded_points: instance_question.variants_points_list, init_points: assessment_question.init_points});  %></td>
+        <td colspan="2" style="white-space: normal;">
```
jonatanschroeder, line 17: Is the `white-space: normal` necessary here?

apps/prairielearn/src/pages/partials/questionAwardedPoints.ejs
```diff
@@ -1,11 +1,28 @@
...
+
+<% if (previous_variants) { %>
+    <%
+        const collapseId = `variants-points-collapse-${instance_question_id}`;
+        const collapseButtonId = `variants-points-collapse-button-${instance_question_id}`;
+        let startHiddenIndex = null;
+        let endHiddenIndex = null;
+    %>
+    <% if (previous_variants.length > MAX_DISPLAYED_VARIANTS) { %>
+        <a id="<%= collapseButtonId %>" role="button" onclick="$(`#<%= collapseId %>`).css('display','inline'); $(`#<%= collapseButtonId %>`).css('display','none');">&hellip;</a>
```
jonatanschroeder, line 13: Nit: could this logic be moved to a separate script tag for cleanness? And in that case, either switch to vanilla JS, or use JQuery show/hide.

apps/prairielearn/src/pages/studentAssessmentInstance/studentAssessmentInstance.sql
```diff
@@ -7,6 +7,49 @@ WHERE
...
+      JOIN variants AS v ON (v.instance_question_id = iq.id)
+      LEFT JOIN variant_max_submission_scores AS vmss ON (vmss.variant_id = v.id)
+    WHERE
+      iq.assessment_instance_id = $assessment_instance_id
+      AND aq.deleted_at IS NULL
+      AND NOT v.open
+      AND NOT v.broken
+    GROUP BY
+      iq.id
+  )
```
jonatanschroeder, line 52: Similar comments as before about aq.deleted_at.

```suggestion
WITH
  variant_max_submission_scores AS (
    SELECT
      v.id AS variant_id,
      max(s.score) AS max_submission_score
    FROM
      instance_questions AS iq
      JOIN variants AS v ON (v.instance_question_id = iq.id)
      JOIN submissions AS s ON (s.variant_id = v.id)
    WHERE
      iq.assessment_instance_id = $assessment_instance_id
      AND s.score IS NOT NULL
    GROUP BY
      v.id
  ),
  instance_question_variants AS (
    SELECT
      iq.id AS instance_question_id,
      jsonb_agg(
        jsonb_build_object(
          'variant_id',
          v.id,
          'max_submission_score',
          COALESCE(vmss.max_submission_score, 0)
        )
        ORDER BY
          v.date
      ) AS variants
    FROM
      instance_questions AS iq
      JOIN variants AS v ON (v.instance_question_id = iq.id)
      LEFT JOIN variant_max_submission_scores AS vmss ON (vmss.variant_id = v.id)
    WHERE
      iq.assessment_instance_id = $assessment_instance_id
      AND NOT v.open
      AND NOT v.broken
    GROUP BY
      iq.id
  )
```

apps/prairielearn/src/pages/partials/questionAwardedPoints.ejs
jonatanschroeder: Also, use `button class="btn"` instead of `a role="button"`.

apps/prairielearn/src/pages/partials/questionAwardedPoints.ejs
mwest1066: Good point, it should just be a single `<a class="badge">`. Fixed in f4793b734a97d9fe1002466a3c3e770101f697d3

apps/prairielearn/src/middlewares/selectAndAuthzInstanceQuestion.sql
```diff
@@ -43,6 +43,50 @@ WITH
...
+      JOIN assessment_instances AS ai ON (ai.id = iq.assessment_instance_id)
+      JOIN variants AS v ON (v.instance_question_id = iq.id)
+      LEFT JOIN variant_max_submission_scores AS vmss ON (vmss.variant_id = v.id)
+    WHERE
+      iq.assessment_instance_id = ai.id
+      AND aq.deleted_at IS NULL
+      AND NOT v.open
+      AND NOT v.broken
+    GROUP BY
+      iq.id
```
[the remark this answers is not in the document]
mwest1066: Great point, apparently I never read these queries in my review 🤦. Applied in 029783f37

apps/prairielearn/src/middlewares/selectAndAuthzInstanceQuestion.sql
```diff
@@ -123,6 +169,7 @@ FROM
   LEFT JOIN users AS u ON (u.user_id = ai.user_id)
   LEFT JOIN users AS uag ON (uag.user_id = iq.assigned_grader)
   LEFT JOIN users AS ulg ON (ulg.user_id = iq.last_grader)
+  LEFT JOIN instance_question_variants AS iqv ON (iqv.instance_question_id = iq.id)
```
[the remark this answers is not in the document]
mwest1066: Applied in 108c3b4a1

apps/prairielearn/src/pages/studentAssessmentInstance/studentAssessmentInstance.sql
mwest1066: Applied in ca475996c

apps/prairielearn/src/pages/partials/questionScorePanel.ejs
mwest1066: Yep, without it you end up with this:

image

apps/prairielearn/src/pages/partials/questionAwardedPoints.ejs
mwest1066: I replaced the jQuery with plain JS in 964127603c9aec23044d8e583b6105d91383b545. I left the code in the `onclick` handler because it's quite short and self-contained. In the future we can move this to an asset when we convert this all to `html` templates.

Regarding the use of a button, I actually deliberately chose to style this as a link instead of a button because a button takes up too much vertical space, expanding the row height (and a `btn-xs` uses a too-small font).

apps/prairielearn/src/pages/partials/questionScorePanel.ejs
jonatanschroeder: What about adding [the text-wrap class](https://getbootstrap.com/docs/4.5/utilities/text/#text-wrapping-and-overflow)?

apps/prairielearn/src/pages/partials/questionAwardedPoints.ejs
jonatanschroeder: What about btn-link? https://getbootstrap.com/docs/4.6/components/buttons/#examples

mwest1066: That still includes all the usual button padding and margin. I think this is a case where using an actual `<a>` link but indicating that it should function as a button (`role="button"`) is cleanest?

apps/prairielearn/src/pages/partials/questionScorePanel.ejs
mwest1066: Good call, that does exactly the same thing. Changed to `.text-wrap` in 4009693c923a4ca707d4aff75f3f5a9dedd392ef

apps/prairielearn/src/pages/partials/questionAwardedPoints.ejs
mwest1066: Even using `btn btn-link p-0 m-0 border-0 rounded-0` I still couldn't get the button to preserve the line height. I think the button element itself is slightly too high (and using `btn-xs` shrinks the font). I thus left it as a `<a role="button">`.

jonatanschroeder: The ellipsis don't have any visual indication that they can be clicked with the a styling, that looks weird to me. I tried with `badge-light`, it looks a little bit better, but I'll leave that decision to you. And I also wasn't able to get the button to style in a decent way.

One thing I tried that made it look a bit better was to replace `&ellip;` with `ctdot;`. Up to you.

apps/prairielearn/src/pages/partials/questionAwardedPoints.ejs
```diff
@@ -4,25 +4,32 @@
...
     <% if (previous_variants.length > MAX_DISPLAYED_VARIANTS) { %>
-        <a id="<%= collapseButtonId %>" role="button" onclick="$(`#<%= collapseId %>`).css('display','inline'); $(`#<%= collapseButtonId %>`).css('display','none');">&hellip;</a>
-        <% startHiddenIndex = 0; %>
-        <% endHiddenIndex = previous_variants.length - MAX_DISPLAYED_VARIANTS; %>
+        <a
+            id="<%= collapseButtonId %>"
+            role="button"
+            onclick="
+                // show all the hidden variant score buttons
+                document.querySelectorAll('.<%= collapseClass %>').forEach(e => e.style.display = 'inline-block');
```
jonatanschroeder, line 16: Nit: the usual suggestion I see in this case to set display to `''` (blank), so it takes the default display of the context.

jonatanschroeder (approved):
As I was testing this, I noticed an interesting thing that I'm not sure if it should be addressed or not. When I click on a previous variant, the button for "Try a new variant" would show. This button would create a new variant if no open variant exists, but if an open/incomplete variant exists, it would switch to the open variant (i.e., a variant that has no submission, or one that has been saved but not graded, or one where the submission is ungradable, or one with multiple tries per variant). I believe this behaviour is expected and should remain, but the text of the button ("Try a new variant") is misleading. Do you think it's worth the effort to change it? In other words, if there is an open variant in the current instance question that is not the current one, change the text to something that says "Go to latest variant" or equivalent?

Also, that latest variant is never shown in the history. Should it? At least if it has a submission? In that case it could show "Open", or if it has a valid submission (e.g., >1 tries per submission) the percentage of the submissions so far plus "/Open"? Possibly with a different colour?

I'm approving the PR in the case where you believe the comments above should not be addressed in the current PR, but if you decide to address them here feel free to request a new review.

apps/prairielearn/src/pages/partials/questionAwardedPoints.ejs
mwest1066: I switched to a `<button>` by avoiding Bootstrap's `.btn` styling and explicitly styling a built-in button. This gives the correct spacing.

I agree with you that I'd prefer a better visual indication that this element is clickable. It does have the mouse pointer change on hover, which helps for platforms with pointers but not for touch platforms. I tried using a `.badge` but that doesn't play well with `<button>` and I'd rather not fall back to `<a>` now that I've got `<button>` working (for better accessibility support).

I like the idea of `&ctdot;` -- I switched to that.

apps/prairielearn/src/pages/partials/questionAwardedPoints.ejs
mwest1066: Oh nice! TIL! Changed in 78fe96881285365ac201672163f65482e41d1ddc

mwest1066:
> As I was testing this, I noticed an interesting thing that I'm not sure if it should be addressed or not. When I click on a previous variant, the button for "Try a new variant" would show. This button would create a new variant if no open variant exists, but if an open/incomplete variant exists, it would switch to the open variant (i.e., a variant that has no submission, or one that has been saved but not graded, or one where the submission is ungradable, or one with multiple tries per variant). I believe this behaviour is expected and should remain, but the text of the button ("Try a new variant") is misleading. Do you think it's worth the effort to change it? In other words, if there is an open variant in the current instance question that is not the current one, change the text to something that says "Go to latest variant" or equivalent?
> 
> Also, that latest variant is never shown in the history. Should it? At least if it has a submission? In that case it could show "Open", or if it has a valid submission (e.g., >1 tries per submission) the percentage of the submissions so far plus "/Open"? Possibly with a different colour?
> 
> I'm approving the PR in the case where you believe the comments above should not be addressed in the current PR, but if you decide to address them here feel free to request a new review.

Thanks. These are all reasonable questions, but I'm going to merge this PR as is because I think it brings useful functionality even in its current state and I don't have the bandwidth to devote to PR anymore (I was only trying to get @ffund's original PR merged). I'd be very happy to discuss the questions above or review a followup PR.

PR #6107, opened 2022-08-03
Split paid hosting link on to its own line

SethPoulsen:
I think it looks a little better formatted this way, feel free to discard the PR if you prefer the way it is.


README.md
```diff
@@ -4,5 +4,5 @@ PrairieLearn is an online problem-driven learning system for creating homeworks
 
 [![GitHub Actions build status](https://github.com/PrairieLearn/PrairieLearn/actions/workflows/main.yml/badge.svg)](https://github.com/PrairieLearn/PrairieLearn/actions/workflows/main.yml) [![License](https://img.shields.io/github/license/PrairieLearn/PrairieLearn.svg)](https://github.com/PrairieLearn/PrairieLearn/blob/master/LICENSE)
 
-Documentation website: [https://prairielearn.readthedocs.io/](https://prairielearn.readthedocs.io/)
+Documentation website: [https://prairielearn.readthedocs.io/](https://prairielearn.readthedocs.io/)\
```
nwalters512, line 7: Nit, let's just use linebreaks, the trailing slash is kinda non-obvious:
```suggestion
Documentation website: [https://prairielearn.readthedocs.io/](https://prairielearn.readthedocs.io/)

```

nwalters512 (approved):
Thanks, the original formatting was totally unintentional!

README.md
echuber2: Yes, blank lines in between cause a paragraph separation, subsequent lines with no trailing slash run into the same formatted line, and the trailing lash inserts a newline in the same paragraph. However, if you end a line with _two_ spaces (instead of a slash), that will also induce a newline.

echuber2: I avoid the two trailing spaces method just because I sometimes use a "remove trailing spaces" macro in my editor.

SethPoulsen: I have a strong dislike of the trailing double space because (1) you can't see it at all and (2) I typically like having my editor auto-trim trailing whitespace whenever possible.

nwalters512: Separate paragraphs are totally fine. Alternatively, a bulleted list would also be good 👍🏻

PR #6109, opened 2022-08-04
Implement group role selection

renzol2:
### Feature

As a follow-up to #5200, this PR implements functionality for students to select their group roles before starting an assessment.

The group roles they select from are defined by instructors in the assessment’s `infoAssessment.json` file (see #5200). All changes are backwards-compatible if the instructor opts not to use group roles for a given assessment.

We will implement the functionality of group roles during an assessment (i.e. enabling or restricting viewing/submitting questions) in a future PR.

image

### Changes

Added a table component in the UI to allow students to select their group roles before an assessment

Added backend logic to update students’ roles in the database

Added logic to enforce that student-selected group role assignments are valid before an assessment starts, including:

- Preventing too many/few students in a given role
- Ensuring there’s always someone with the role that can assign other student roles
- Ensuring optional group roles aren’t selectable until all required roles are filled
- Ensuring each student is only assigned one role unless the group size is smaller than the number of required roles (in which case having multiple roles per student is allowed)

Added logic to enforce the above rules when students create, join, and leave a group by:

- Automatically assigning the group’s creator a role that can assign other roles
- Giving new group members appropriate required/optional roles based on group size
- Reassigning roles of leaving group members to other members, if necessary

Added logic when syncing course content that enforces that, in every assignment with group roles, there exists a required role that has the ability to assign other roles

Implemented tests for all new logic


lib/groups.sql
```diff
@@ -6,6 +6,19 @@ FROM
 WHERE
     gc.assessment_id = $assessment_id AND gc.deleted_at IS NULL;
 
+-- BLOCK get_creator_role
```
nwalters512, line 9: To check my understanding: this gets a role that should be assigned to the creator, not the role or whoever created the group? I know that's a subtle distinction, but the name makes me think it's the latter.

migrations/20220830011729_group_configs__using_group_roles__add.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE group_configs ADD COLUMN using_group_roles boolean DEFAULT false;
+
+ALTER TABLE group_users DROP CONSTRAINT group_users_pkey;
+ALTER TABLE group_users ADD PRIMARY KEY (group_id, user_id, group_role_id);
```
nwalters512, line 4: To maintain integrity, you should do these in a single statement:

```suggestion
ALTER TABLE group_users
DROP CONSTRAINT group_users_pkey,
ADD PRIMARY KEY (group_id, user_id, group_role_id);
```

@mwest1066 could I get your eyes on this too?

lib/groups.sql
```diff
@@ -29,48 +42,63 @@ create_log AS (
...
+    gr.*
+FROM
+    group_roles as gr
+WHERE
+    gr.assessment_id = $assessment_id;
+
+-- BLOCK get_assessment_level_permissions
+SELECT 
+    bool_or(gr.can_assign_roles_at_start) AS can_assign_roles_at_start, 
+    bool_or(gr.can_assign_roles_during_assessment) AS can_assign_roles_during_assessment
```
nwalters512, line 100: Why is `bool_or` being used with a single argument?

pages/partials/studentGroupControls.ejs
```diff
@@ -28,10 +28,33 @@
...
+    <% if (typeof validationErrors !== 'undefined' && validationErrors.length > 0) { %>
+      <% validationErrors.forEach(function ({ role_name, minimum, maximum, num_assigned }) { %>
+        <div class="alert alert-danger my-2" role="alert">
+          <%= role_name %>
+        <% if (num_assigned < minimum) { %>
+          has too few assignments (found  
+          <% } else { %>
+          has too many assignments (found
+          <% } %>
+          <%= num_assigned %> but expected <%= minimum %> minimum and <%= maximum %> maximum.)
```
nwalters512, line 51:
```suggestion
          <%= num_assigned %> but expected between <%= minimum %> and <%= maximum %>.)
```

Consider special-casing the case where `minimum === maximum` - regardless of what phrasing is used, it'd be weird to read "expected 1 minimum and 1 maximum" or "expected between 1 and 1".

pages/shared/groupRoleSelectTable.ejs
```diff
@@ -0,0 +1,42 @@
+<form id="role-select-form" name="role-select-form" method="POST">
+  <table class="table table-bordered table-striped mt-3 pl-3 pr-3">
+    <thead>
+      <tr>
+        <th scope="col">Select Role</th>
```
nwalters512, line 5: I believe this is more correct?

```suggestion
        <th scope="col">User</th>
```

pages/shared/groupRoleSelectTable.ejs
```diff
@@ -0,0 +1,42 @@
...
+              name="<%= role.id %>-<%= user.uid %>"
+              value="checked"
+              <% if (disabledRoles.includes(role.role_name)) {%>
+                disabled
+              <% } else if (user.role_names.includes(role.role_name)) { %>
+                checked
+              <% }%>
+              form="role-select-form"
+            >
+            <label for="<%= role.id %>-<%= user.uid %>" class=<%= disabledRoles.includes(role.role_name) && "text-muted"%>>
```
nwalters512, line 28:
```suggestion
            <label for="<%= role.id %>-<%= user.uid %>" <% if (disabledRoles.includes(role.role_name)) { %> class="text-muted" <% } %>>
```

database/tables/group_users.pg
```diff
@@ -1,10 +1,10 @@
 columns
     group_id: bigint not null
-    group_role_id: bigint
+    group_role_id: bigint not null
     user_id: bigint not null
 
 indexes
-    group_users_pkey: PRIMARY KEY (group_id, user_id) USING btree (group_id, user_id)
+    group_users_pkey: PRIMARY KEY (group_id, user_id, group_role_id) USING btree (group_id, user_id, group_role_id)
```
nwalters512, line 7: Does including the role in the key imply that a given user may have multiple roles for a given group?

pages/partials/studentGroupControls.ejs
```diff
@@ -28,10 +28,33 @@
...
   </div>
 <% } %>
+
 <%- include('../shared/groupWorkInitial.ejs', {permissions: permissions}); %>
 <% } %>
 <% } else {%>
   <%- include('../shared/groupWorkInfoContainer.ejs', {permissions: permissions, group_info: group_info}); %>
+  <% if (usingGroupRoles && !rolesAreBalanced) { %>
+    <div class="alert alert-danger my-2" role="alert">
+      A user has too many roles. Every student should be assigned to exactly <strong>one</strong> role with group size <strong><%= groupsize %></strong>
```
nwalters512, line 38: My question here relates to my other question: can a given student have multiple roles in the same group? If so, this message is misleading. It's also not clear to me why the group size is relevant to an error message about a user having too many roles.

Backing up: how is it possible to get into a state where a user has "too many roles" - wouldn't we verify that that's not the case before inserting assignments into the DB?

lib/groups.js
```diff
@@ -139,12 +179,54 @@ module.exports.createGroup = function (groupName, assessmentId, userId, authnUse
...
+module.exports.getAssessmentLevelPermissions = function (assessmentId, userId, callback) {
   const params = {
     assessment_id: assessmentId,
     user_id: userId,
-    authn_user_id: authnUserId,
   };
-  sqldb.query(sql.leave_group, params, function (err) {
+  sqldb.query(sql.get_assessment_level_permissions, params, function (err, result) {
+    if (ERR(err, callback)) return;
+    callback(result.rows[0]);
```
nwalters512, line 196: When using callbacks, the first argument should always be the error:

```suggestion
    callback(null, result.rows[0]);
```

lib/groups.js
```diff
@@ -139,12 +179,54 @@ module.exports.createGroup = function (groupName, assessmentId, userId, authnUse
...
 
 module.exports.leaveGroup = function (assessmentId, userId, authnUserId, callback) {
+  const params = [assessmentId, userId, authnUserId];
+  sqldb.call('group_leave', params, function (err) {
+    if (ERR(err, callback)) return;
+    callback();
+  });
+};
+
+module.exports.getAssessmentLevelPermissions = function (assessmentId, userId, callback) {
```
nwalters512, line 189: This and the below function (`updateGroupRoles`) should be written as `async` functions instead of using callbacks.

lib/groups.js
```diff
@@ -104,28 +139,33 @@ module.exports.joinGroup = function (joinCode, assessmentId, userId, authnUserId
...
-          if (ERR(err, callback)) return;
-          const permissions = result.rows[0];
-          const uniqueGroupName = true;
-          callback(null, false, uniqueGroupName, null, permissions);
-        });
-      }
+    sqldb.query(sql.get_creator_role, params, function (err, result) {
+      if (ERR(err, callback)) return;
+      params.group_role_id = result.rows[0].id;
+      sqldb.query(sql.create_group, params, function (err, _result) {
```
nwalters512, line 157: Rather than having two separate queries, I'd recommend using a CTE inside `create_group` to assign a role to the group creator.

lib/groups.js
```diff
@@ -104,28 +139,33 @@ module.exports.joinGroup = function (joinCode, assessmentId, userId, authnUserId
...
+        if (!err) {
```
nwalters512, line 158: I know this is existing code that you didn't touch, but I don't understand what's happening here: is `create_group` expected to error? Why do we `get_config_info` if there was an error?

lib/groups.js
```diff
@@ -139,12 +179,54 @@ module.exports.createGroup = function (groupName, assessmentId, userId, authnUse
...
+
+module.exports.updateGroupRoles = function (
+  requestBody,
+  assessmentId,
+  userId,
+  authnUserId,
+  callback
+) {
+  // Convert form data to valid input format for a SQL function
+  const roleKeys = Object.keys(requestBody).filter((key) => !key.startsWith('__'));
```
nwalters512, line 208: Rather than looking specifically for things that *don't* start with `__`, I'd recommend adding a known prefix to the input names, e.g. `user_role_`.

lib/groups.js
```diff
@@ -139,12 +179,54 @@ module.exports.createGroup = function (groupName, assessmentId, userId, authnUse
...
+  const uidToRoleIdMap = {};
+  for (const roleKey of roleKeys) {
+    const [roleId, uid] = roleKey.split('-');
```
nwalters512, line 211: Is the thing after the first `-` really a UID? If so, this isn't safe - what if a UID itself contains a `-`? I'd recommend using the `user_id` instead.

lib/groups.js
```diff
@@ -139,12 +179,54 @@ module.exports.createGroup = function (groupName, assessmentId, userId, authnUse
...
+  }
+
+  const roleAssignments = [];
+  Object.entries(uidToRoleIdMap).forEach((entry) => {
+    const roleAssignment = {
+      uid: entry[0],
+      group_role_ids: entry[1].map((id) => parseInt(id)),
+    };
+    roleAssignments.push(roleAssignment);
+  });
```
nwalters512, line 226: Nit, this is a bit more concise:

```suggestion
  const roleAssignments = Object.entries(uidToRoleIdMap).map((entry) => ({
    uid: entry[0],
    group_role_ids: entry[1].map((id) => parseInt(id)),
  }));
```

sprocs/assessment_groups_update.sql
```diff
@@ -13,7 +13,25 @@ DECLARE
     arg_user_id bigint;
     arg_group_id bigint;
     group_user text[];
+    default_role_id bigint;
 BEGIN
+    IF NOT EXISTS (
```
nwalters512, line 18: Same comment as before: this smells funky to me. Let's see if we can do without this special value.

sprocs/group_leave.sql
```diff
@@ -0,0 +1,104 @@
...
+            -- When group_size (pre-leave) > num_required roles:
+
+            -- 1. Get whether the leaving user has a required role
+            SELECT gr.minimum > 0
+            INTO arg_has_required_role
+            FROM group_users gu LEFT JOIN group_roles gr ON gu.group_role_id = gr.id
+            WHERE gu.group_id = arg_group_id AND gu.user_id = arg_user_id;
+
+            IF arg_has_required_role THEN
+                -- 2. If true, find a group member with a non-required role
```
nwalters512, line 75: What if there's no group member with a non-required role? Does this code still work correctly?

sprocs/group_roles_validate.sql
```diff
@@ -0,0 +1,77 @@
+CREATE FUNCTION
+    group_roles_validate (
```
nwalters512, line 2: Is it necessary for this code to be a sproc? I suspect we could do this instead with a single `SELECT` and some JavaScript, which would be my preference.

sprocs/group_users_insert.sql
```diff
@@ -56,16 +59,50 @@ BEGIN
...
     INSERT INTO group_users
-        (user_id, group_id)
+        (user_id, group_id, group_role_id)
     VALUES
-        (arg_user_id, arg_group_id);
-
-    INSERT INTO group_logs
-        (authn_user_id, user_id, group_id, action)
-    VALUES
-        (arg_authn_user_id, arg_user_id, arg_group_id, 'join');
```
nwalters512, line 68: Did you mean to delete this logging?

sync/fromDisk/assessments.js
```diff
@@ -77,6 +77,7 @@ function getParamsForAssessment(assessmentInfoFile, questionIds) {
     student_group_join: !!assessment.studentGroupJoin,
     student_group_leave: !!assessment.studentGroupLeave,
     advance_score_perc: assessment.advanceScorePerc,
+    using_group_roles: assessment.groupRoles ? true : false,
```
nwalters512, line 80: Nit:

```suggestion
    using_group_roles: !!assessment.groupRoles,
```

database/tables/group_users.pg
renzol2: Yes, a given user can have multiple roles in any given group. This allows groups to continue completing assessments with the necessary roles even if users leave the group while completing the assessment.

This feature is partially why we had the code-smell value that you pointed out [here](https://github.com/PrairieLearn/PrairieLearn/pull/6109#discussion_r1037628409). We weren't sure exactly how to implement multiple roles per user in a group while also ensuring that `group_users` could still represent users in groups that don't use group roles. We'll try and think of a workaround, but any advice would be greatly appreciated!

pages/partials/studentGroupControls.ejs
renzol2: Regarding the unclear error message: we asked the SIIP team about their thoughts on the potential for group role configurations like the following, where one user has all the "required" roles (a role with a minimum > 0) and the rest of the users have non-required roles. Based on their feedback, we decided to prevent role configurations like these by enforcing that every user should have one role if `# of users >= # of required roles`.

image

Admittedly, we didn't specify this requirement for users at all in the frontend (other than the error itself), so if this requirement still makes sense to implement, we can detail it in `studentGroupControls`.

Regarding the possibility of having "too many roles" in the first place: we originally wanted to verify that role assignments were valid before inserting them into the database. We first implemented this in the `POST` endpoint for the `update_group_roles` action, where we would call the `group_roles_validate` sproc to check if the role assignments were valid before inserting them into the database - if they weren't valid, they would never be inserted.

However, we couldn't find a way to cleanly display errors from the sproc to users from that `POST` endpoint - if the new role assignments were valid, we could just redirect to the same page, but if we wanted to add error messages to `res.locals`, we had to duplicate code from the `GET` endpoint to populate `res.locals` with the necessary variables as well as the validation error messages. 

Instead, we decided to allow *any* role assignment configuration to be inserted into the database, but validate whenever users load the page (in the `GET` endpoint) and subsequently disable users from starting the assessment if their group's role assignments were invalid. By validating within the `GET` endpoint, we could avoid duplicating code while also sending error messages if roles were invalid. This does mean that there could be invalid role assignments in the database, which is why we validate them when the page is loaded so we can disable the assessment from starting.

Do you think we should change our approach back to preventing invalid role assignments in the first place?

lib/groups.sql
caleqw2: If the user has multiple roles, `gr.can_assign_roles_at_start` etc. would have multiple rows, so `bool_or` was intended to reduce these rows in order to tell if a student has at least one role with the relevant permission.

sprocs/group_users_insert.sql
caleqw2: It's still there on line 107

database/tables/group_configs.pg
```diff
@@ -10,6 +10,7 @@ columns
     student_authz_create: boolean default false
     student_authz_join: boolean default false
     student_authz_leave: boolean default false
+    using_group_roles: boolean default false
```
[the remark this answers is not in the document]
caleqw2: Renamed column and corresponding JS/EJS

lib/groups.sql
```diff
@@ -29,48 +42,63 @@ create_log AS (
...
-                        WHERE a.id = $assessment_id
-                        AND g.deleted_at IS NULL
-                        AND gc.deleted_at IS NULL)
-    RETURNING group_id
-)
-INSERT INTO group_logs
-    (authn_user_id, user_id, group_id, action)
-SELECT $authn_user_id, $user_id, group_id, 'leave'
-FROM log;
+-- BLOCK get_group_info_with_roles
```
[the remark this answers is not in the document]
caleqw2: Condensed queries into one

sprocs/group_roles_validate.sql
caleqw2: Moved to a single select statement in `groups.sql`.

sprocs/group_leave.sql
caleqw2: Updated `group_leave` logic to meet the case when there's no one with an optional role, and also the case when the leaving user has multiple roles in a group where `group_size` > `num_required_roles`.

lib/groups.js
caleqw2: Modified this function and `group_roles_update.sql` to use `user_id` instead of `uid`

lib/groups.js
caleqw2: `create_group` will error if the group name is already taken. If the group name is taken or invalid, the page needs to re-render with the appropriate errors, so the callback function ends up calling `res.render(...)`. `get_config_info` is called to get `permissions` which is needed in `res.locals` for the page to load.

This is a mini version of the problem that Renzo mentioned [here](https://github.com/PrairieLearn/PrairieLearn/pull/6109#discussion_r1037784714), where we can't use a POST/redirect/GET pattern because the GET call can't remember the errors to display.

lib/groups.sql
caleqw2: Renamed and moved to CTE

pages/studentAssessmentInstanceHomework/studentAssessmentInstanceHomework.js
```diff
@@ -67,44 +67,31 @@ router.get('/', function (req, res, next) {
       assessment.renderText(
         res.locals.assessment,
         res.locals.urlPrefix,
-        function (err, assessment_text_templated) {
+        async function (err, assessment_text_templated) {
```
caleqw2, line 70: Might be an issue to pass an async function as a callback, like Nathan mentioned earlier.

sprocs/group_leave.sql
```diff
@@ -0,0 +1,118 @@
...
+                END IF;
+            END LOOP;
+        END IF;
+    END IF;
+
+    -- Delete the user from the group
+    DELETE FROM group_users
+    WHERE user_id = arg_user_id AND group_id = arg_group_id;
+
+    DELETE FROM group_user_roles
```
caleqw2, line 111: I wonder if there's a way to cascade this delete so that we only have to delete from `group_users`

migrations/20220830011729_group_configs__using_group_roles__add.sql
```diff
@@ -0,0 +1,10 @@
...
+
+-- Insert role for group assessments that don't use group roles
+INSERT INTO group_roles 
+    (role_name, maximum)
+VALUES
+    ('No group roles', 99);
```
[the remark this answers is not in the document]
renzol2: Since our original solution required `group_users` have all three columns `group_id`, `user_id`, and `group_role_id` to be non-`NULL` as they were part of the primary key, we used this value to get around that.

I committed a change to create a new relational table `group_user_roles` and to revert `group_users` back to having only a `group_id` and `user_id`. See: https://github.com/PrairieLearn/PrairieLearn/pull/6109/commits/74baeda59120fef0d283ccc4653af27bb16907ba

We originally thought to implement roles this way, but we didn't want to duplicate information from `group_users` into another table. Is there a better way to go about implementing this?

sprocs/assessment_groups_update.sql
renzol2: See: https://github.com/PrairieLearn/PrairieLearn/pull/6109#discussion_r1042685844

pages/studentAssessmentExam/studentAssessmentExam.js
```diff
@@ -37,22 +37,50 @@ router.get('/', function (req, res, next) {
           maxsize,
```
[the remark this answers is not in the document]
renzol2: Done with this commit: https://github.com/PrairieLearn/PrairieLearn/pull/6109/commits/5bb0f1977c1dc794132ff1abc6a0cd4dfc929508

We'll still have to change `studentAssessmentInstanceExam.js`. I'm a little unsure of how to do it correctly since the callback code there doesn't have async versions

pages/studentAssessmentInstanceHomework/studentAssessmentInstanceHomework.js
renzol2: @nwalters512 Would it be possible to rewrite the enclosing `router.get()` callback function in this file to be entirely async? I'm not sure how to do so because most of these methods in `assessment.js` do not have async versions.

If it's not possible, would we use the `async` library with `async.series()`, or should we leave this as is?

migrations/20220830011729_group_configs__using_group_roles__add.sql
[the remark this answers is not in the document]
nwalters512: Probably the best way would have been to not use `(group_id, user_id)` as a primary key for `group_users`, but instead use an auto-incrementing ID. You could then have a `group_user_roles` table that references that ID (`group_user_id`).

That said, separating group membership and group roles into separate tables is already a much better way to go! And in fact, if you treat `(group_id, user_id)` as a `FOREIGN KEY` in the `group_user_roles` table, you won't *really* be duplicating any information, and you can take advantage of cascading deletes too. Let me know if you want me to expand on that.

lib/groups.sql
nwalters512: Ah! Gotcha. I didn't realize `bool_or` was an aggregate function - this makes sense.

database/tables/group_users.pg
nwalters512: It sounds like this has been solved by moving role assignments to another table! Feel free to resolve if that's the case.

sprocs/group_leave.sql
nwalters512: I believe you can! See https://github.com/PrairieLearn/PrairieLearn/pull/6109#discussion_r1044683215.

pages/partials/studentGroupControls.ejs
nwalters512: Yeah, I apologize for the lack of the ability to show useful errors etc. in response to a `POST`. We'll be moving to a stateful session model soon, which should make this possible. In the meantime, I think the current validate-on-GET approach makes sense.

A related question: can roles be re-assigned once the assessment has been started? If so, what happens if the new configuration is no longer valid?

caleqw2: Roles currently can't be reassigned once an assessment has started and no one can view the table, but the configuration should stay valid upon students leaving and joining as long as the assessment was started with a valid configuration. The SIIP team advised Renzo and I to include mid-assessment role reassignment in our next PR in order to keep the size of this current PR manageable and to start the review process sooner.

Is mid-assessment role reassignment something you'd like to see in this PR?

tests/groupRole.test.js
```diff
@@ -0,0 +1,1349 @@
...
+const storedConfig = {};
+
+describe('Test group based assessments with custom group roles from student side', function () {
+  this.timeout(20000);
+  before('set authenticated user', function (callback) {
+    storedConfig.authUid = config.authUid;
+    storedConfig.authName = config.authName;
+    storedConfig.authUin = config.authUin;
+    callback(null);
+  });
```
nwalters512, line 32: Using a callback isn't necessary here since this is a synchronous function.

tests/groupRole.test.js
```diff
@@ -0,0 +1,1349 @@
...
+  before('set up testing server', helperServer.before(locals.courseDir));
+  after('shut down testing server', helperServer.after);
+  after('unset authenticated user', function (callback) {
+    Object.assign(config, storedConfig);
+    callback(null);
+  });
+
+  describe('1. the database', function () {
+    it('should contain a group-based homework assessment with roles', function (callback) {
```
nwalters512, line 41: The function passed here should be `async` instead of using callbacks; ditto everywhere else in this file.

tests/groupRole.test.js
```diff
@@ -0,0 +1,1349 @@
...
+      sqldb.query(sql.select_group_work_assessment_with_roles, [], function (err, result) {
```
nwalters512, line 42: We should use `queryAsync` and friends in all new code instead of using these callback functions; ditto everywhere else in this file.

tests/groupRole.test.js
```diff
@@ -0,0 +1,1349 @@
...
+        callback(null);
+      });
+    });
+    it('should contain the 4 roles for the assessment', function (callback) {
+      const params = {
+        assessment_id: locals.assessment_id,
+      };
+      sqldb.query(sql.select_assessment_group_roles, params, function (err, result) {
+        if (ERR(err, callback)) return;
+        assert.lengthOf(result.rows, 4);
```
nwalters512, line 69: If you're going to use a giant shared object, you should use `mocha-steps` to ensure that if one step fails, the entire test aborts. See https://github.com/PrairieLearn/PrairieLearn/blob/master/tests/bonusPoints.test.js for an example. When you do that, you'll need to move all steps that depend on previous steps into the same `describe` block so that they can be aborted as a whole.

tests/groupRole.test.js
```diff
@@ -0,0 +1,1349 @@
...
+      config.authUid = locals.studentUsers[0].uid;
+      config.authName = locals.studentUsers[0].name;
+      config.authUin = '00000001';
+      config.userId = locals.studentUsers[0].user_id;
+    });
+  });
+
+  describe('3. POST to assessment page to create group', function () {
+    it('should load assessment page successfully', function (callback) {
+      request(locals.assessmentUrl, function (error, response, body) {
```
nwalters512, line 115: This should use `node-fetch` and promises instead of `request` + callbacks. See https://github.com/PrairieLearn/PrairieLearn/pull/7206/files for an example of a PR where I made similar changes; note that that was with `request-promise-native` instead of just `request`, but the general idea should be the same.

tests/groupRole.test.js
```diff
@@ -0,0 +1,1038 @@
...
+    storedConfig.authUin = config.authUin;
+  });
+  before('set up testing server', helperServer.before(locals.courseDir));
+  after('shut down testing server', helperServer.after);
+  after('unset authenticated user', function () {
+    Object.assign(config, storedConfig);
+  });
+
+  describe('1. the database', function () {
+    step('should contain a group-based homework assessment with roles', async function () {
```
nwalters512, line 39: Nesting `step` inside of `describe` is less than ideal: if a step inside a single `describe` block fails, only _that_ `describe` will be aborted, but the rest will keep executing. There are two things you could do here:

- Move all `step`s up to the root `describe` and get rid of the `describe('1. ...)` wrappers
- Instead of having multiple steps within each `describe`, merge the nested steps into a single step and use `step` instead of `describe to group. For example:

```ts
// Change from this:
describe('root', () => {
  describe('a', () => {
    step('1', () => console.log(1));
    step('2', () => console.log(2));
  });
});

// To this:
describe('root', () => {
  step('a', () => {
    console.log(1);
    console.log(2);
  })
});
```

lib/groups.js
```diff
@@ -3,66 +3,108 @@ var ERR = require('async-stacktrace');
...
-        group_info,
-        join_code,
-        start,
-        used_join_code
+module.exports.getGroupInfo = async function (assessmentId, userId) {
+  const groupInfo = {};
+
+  // Get config info
+  let params = { assessment_id: assessmentId, user_id: userId };
+  let result = await sqldb.queryAsync(sql.get_config_info, params);
```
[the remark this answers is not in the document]
renzol2: That looks great! To clarify, should we update all new database queries we write to use these data validating functions?

lib/groups.js
```diff
@@ -3,66 +3,108 @@ var ERR = require('async-stacktrace');
...
+  groupInfo.permissions = result.rows[0];
+  groupInfo.minSize = result.rows[0].minimum || 0;
+  groupInfo.maxSize = result.rows[0].maximum || 999;
+  groupInfo.hasRoles = result.rows[0].has_roles;
+
+  // Get group member info
+  result = await sqldb.queryAsync(sql.get_group_members, params);
+  groupInfo.groupMembers = result.rows;
+  groupInfo.groupSize = result.rowCount;
+  groupInfo.isGroupMember = result.rowCount > 0;
```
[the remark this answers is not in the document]
caleqw2: It represents whether the user is a member of any active group for the current assessment. I took this out of `getGroupInfo()` and made a new variable `notInGroup` in `studentAssessmentHomework.js` and `studentAssessmentExam.js`.

migrations/20220830011729_group_configs__using_group_roles__add.sql
```diff
@@ -0,0 +1,20 @@
+ALTER TABLE group_configs
+ADD COLUMN has_roles boolean DEFAULT false;
+
+ALTER TABLE group_users
+DROP COLUMN IF EXISTS group_role_id;
```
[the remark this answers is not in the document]
renzol2: No; we added this column in our first migration when we thought that each user in a group would have only up to 1 role. However, I don't see any application code that used this column, so I believe it's safe to remove it.

database/tables/group_user_roles.pg
```diff
@@ -0,0 +1,15 @@
+columns
+    group_id: bigint not null
+    group_role_id: bigint not null
+    user_id: bigint not null
+
+indexes
+    group_user_roles_pkey: PRIMARY KEY (group_id, user_id, group_role_id) USING btree (group_id, user_id, group_role_id)
```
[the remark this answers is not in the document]
caleqw2: I see he mentioned that `assessment_question_role_permissions` and `group_users` are also issues. Do you want us to fix these as well, or leave them since they've already been merged?

pages/studentAssessmentInstanceHomework/studentAssessmentInstanceHomework.js
```diff
@@ -40,76 +40,67 @@ router.get('/', function (req, res, next) {
   if (res.locals.assessment.type !== 'Homework') return next();
   debug('is Homework');
 
-  ensureUpToDate(res.locals, (err) => {
+  ensureUpToDate(res.locals, async (err) => {
```
renzol2, line 43: @nwalters512 I'd like to have this entire endpoint wrapped in `asyncHandler` from `express-async-handler`. However, `ensureUpToDate` calls [`update` from `assessments.js`](https://github.com/PrairieLearn/PrairieLearn/blob/ed27bcfa732e8a8e1ff6833a06499a34fa85d7bf/lib/assessment.js#L126), which employs the use of callbacks. Would it be sensible to rewrite `assessment.update` to remove the need for a callback, or is wrapping this GET endpoint in `express-async-handler` even necessary?

For context, I was able to use all async code in the `studentAssessmentInstanceExam.js` equivalent of this endpoint: https://github.com/PrairieLearn/PrairieLearn/blob/2399e8accdf4dfda3e7a2d5d8bb5792414c80602/pages/studentAssessmentInstanceExam/studentAssessmentInstanceExam.js#L104-L106

database/tables/group_user_roles.pg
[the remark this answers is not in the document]
nwalters512: Unfortunately I don't see a clear path to doing so in a way that can tolerate zero-downtime rolling deploys 😢 I think it's fine to leave this alone.

pages/studentAssessmentInstanceHomework/studentAssessmentInstanceHomework.js
nwalters512: You should be able to wrap `assessment.update` in `utils.promisify`. Here's an example:

https://github.com/PrairieLearn/PrairieLearn/blob/ed27bcfa732e8a8e1ff6833a06499a34fa85d7bf/lib/server-jobs.js#L344

Then you can make this local `ensureUpToDate` function async.

pages/shared/groupWorkInfoContainer.ejs
```diff
@@ -52,8 +52,12 @@
...
       <% } %>
       <span id='group-member'><b>Group members: </b></span>
-      <% group_info.forEach(function(row) {%>
-      <li><%= row.user_name %> (<%= row.uid %>)</li>
+      <% groupMembers.forEach(function(user) {%>
+      <% if (hasRoles) { %>
+        <li><%= user.uid %> - <%= rolesInfo.roleAssignments[user.uid] ? rolesInfo.roleAssignments[user.uid].join(", ") : "None"%></li>
+      <% } else { %>
+        <li><%= user.uid %></li>
+      <% } %>
```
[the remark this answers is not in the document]
renzol2: I think having a `roleAssignments` property for each item in `groupMembers` would make more sense in this frontend code, but we're unsure on how to cleanly combine the `get_role_assignments` and `get_group_members` SQL queries. We thought it would be cleaner to keep `roleAssignments` as a separate object within `rolesInfo` in the case that a group assessment doesn't have roles.

If you would prefer we combine these before this gets merged though, let us know!

lib/groups.js
```diff
@@ -191,38 +332,148 @@ module.exports.createGroup = function (groupName, assessmentId, userId, authnUse
...
+              user_id: assigneeUser.user_id,
+              group_role_id: groupRoleId,
+              group_id: groupId,
+            });
+          }
+
+          await sqldb.queryAsync(sql.clear_group_roles, { group_id: groupId });
+          await sqldb.queryAsync(sql.reassign_group_roles_after_leave, {
+            role_assignments: JSON.stringify(groupRoleAssignmentUpdates),
+          });
```
nwalters512, line 394: You should be able to combine these into a single query using CTEs; you can probably even combine it with `delete_not_required_roles` below. While not strictly necessary, we generally like to limit the number of database round-trips we have to do.

lib/groups.js
```diff
@@ -191,38 +332,148 @@ module.exports.createGroup = function (groupName, assessmentId, userId, authnUse
...
+ */
+module.exports.leaveGroup = async function (assessmentId, userId, authnUserId) {
+  await sqldb.runInTransactionAsync(async () => {
+    const groupId = await module.exports.getGroupId(assessmentId, userId);
+    const groupConfig = await module.exports.getGroupConfig(assessmentId);
+
+    if (groupConfig.has_roles) {
+      const groupInfo = await module.exports.getGroupInfo(groupId, groupConfig);
+
+      // Reassign roles if there is more than 1 user
```
nwalters512, line 348: Now that we're doing just a little bit of DB I/O at the start and end of this function, all this role reassignment feels like it could be pretty easily unit-tested if it were split into a separate function!

lib/groups.js
```diff
@@ -191,38 +332,148 @@ module.exports.createGroup = function (groupName, assessmentId, userId, authnUse
...
-  let params = [assessmentId, roleAssignments, userId, authnUserId];
-  await sqldb.callAsync('group_roles_update', params);
+    await sqldb.queryAsync(sql.clear_group_roles, { group_id: groupId });
+    await sqldb.queryAsync(sql.update_group_roles, {
+      group_id: groupId,
+      role_assignments: JSON.stringify(roleAssignments),
+      user_id: userId,
+      authn_user_id: authnUserId,
+    });
+  });
```
nwalters512: Ditto for merging into one query.

lib/groups.js
```diff
@@ -191,38 +332,148 @@ module.exports.createGroup = function (groupName, assessmentId, userId, authnUse
...
-  // Convert form data to valid input format for a SQL function
-  const roleKeys = Object.keys(requestBody).filter((key) => key.startsWith('user_role_'));
-  const userIdToRoleIdMap = {};
-  for (const roleKey of roleKeys) {
-    const [roleId, userId] = roleKey.replace('user_role_', '').split('-');
-    if (userIdToRoleIdMap[userId] === undefined) {
-      userIdToRoleIdMap[userId] = [roleId];
-    } else {
-      userIdToRoleIdMap[userId].push(roleId);
+  await sqldb.runInTransactionAsync(async () => {
```
nwalters512, line 439: @mwest1066 what do you think - should we be explicitly locking the `group_configs` row when doing all this?

lib/groups.sql
```diff
@@ -172,3 +179,78 @@ FROM
...
+          id
+        FROM
+          group_roles
+        WHERE
+          assessment_id = $assessment_id
+          AND minimum = 0
+      )
+  )
+SELECT
+  1;
```
nwalters512: Why the CTE? You should be able to just directly `DELETE FROM ...`.

migrations/20220830011729_group_configs__using_group_roles__add.sql
```diff
@@ -7,12 +7,15 @@ DROP COLUMN IF EXISTS group_role_id;
...
   group_user_roles (
-    group_id BIGINT REFERENCES groups (id),
-    user_id BIGINT REFERENCES users (user_id),
-    group_role_id BIGINT REFERENCES group_roles (id),
-    PRIMARY KEY (group_id, user_id, group_role_id)
+    id BIGSERIAL,
+    group_id BIGINT NOT NULL REFERENCES groups (id),
+    user_id BIGINT NOT NULL REFERENCES users (user_id),
+    group_role_id BIGINT NOT NULL REFERENCES group_roles (id),
+    PRIMARY KEY (id)
```
nwalters512, line 14:
```suggestion
    id BIGSERIAL PRIMARY KEY,
    group_id BIGINT NOT NULL REFERENCES groups (id),
    user_id BIGINT NOT NULL REFERENCES users (user_id),
    group_role_id BIGINT NOT NULL REFERENCES group_roles (id)
```

pages/studentAssessmentInstanceExam/studentAssessmentInstanceExam.js
```diff
@@ -13,44 +13,83 @@ var groupAssessmentHelper = require('../../lib/groups');
...
+          if (ERR(err, next)) return;
+          if (req.body.__action === 'timeLimitFinish') {
+            res.redirect(req.originalUrl + '?timeLimitExpired=true');
+          } else {
+            res.redirect(req.originalUrl);
+          }
+        }
+      );
+    } else if (req.body.__action === 'leave_group') {
+      if (!res.locals.authz_result.active) return next(error.make(400, 'Unauthorized request.'));
```
nwalters512, line 81: Since this is an async function, you can just directly throw the error.

```suggestion
      if (!res.locals.authz_result.active) throw error.make(400, 'Unauthorized request.');
```

pages/studentAssessmentInstanceHomework/studentAssessmentInstanceHomework.js
```diff
@@ -101,54 +107,54 @@ router.get('/', function (req, res, next) {
...
-      function (err) {
+  })
+);
+
+router.post(
+  '/',
+  asyncHandler(async function (req, res, next) {
+    if (res.locals.assessment.type !== 'Homework') return next();
+    if (!res.locals.authz_result.authorized_edit) {
+      return next(error.make(403, 'Not authorized', res.locals));
```
nwalters512, line 118:
```suggestion
      throw error.make(403, 'Not authorized', res.locals);
```

Ditto elsewhere in the function.

lib/groups.js
```diff
@@ -118,33 +276,182 @@ module.exports.createGroup = function (groupName, assessmentId, userId, authnUse
...
+                };
+              } else {
+                return { user_id, group_role_id, group_id: groupId };
+              }
+            }
+          );
+
+          // Assign remaining required roles to another group user
+          const assigneeUser = groupInfo.groupMembers.find((m) => m.user_id !== userId.toString());
+          if (assigneeUser === undefined) {
```
caleqw2, line 353: `assigneeUser` should never be `undefined` here, because we're checking above that `currentSize > 1. Is throwing an error like this the cleanest way to avoid type checking errors? Ditto for other errors thrown in this file for similar `null`/`undefined` checks.

lib/groups.js
caleqw2: Would we need to unit-test this if we've already testing the functionality of this logic in `groupRole.test.js` (which is more of an integration test I think)? That said, I can see how this is enough code to warrant being in its own function.

lib/groups.js
nwalters512: Yep - throwing an error is the way to go!

migrations/20220830011729_group_configs__using_group_roles__add.sql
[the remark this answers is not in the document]
nwalters512: Thanks for confirming!

pages/studentAssessmentExam/studentAssessmentExam.js
```diff
@@ -1,201 +1,214 @@
...
+            );
+            res.locals.canViewRoleTable = result.can_assign_roles_at_start;
+          }
         }
-      );
+
+        res.render(__filename.replace(/\.js$/, '.ejs'), res.locals);
+      } else {
+        res.render(__filename.replace(/\.js$/, '.ejs'), res.locals);
+      }
```
nwalters512, line 64: Nit, minor simplification:

```suggestion
      }
      res.render(__filename.replace(/\.js$/, '.ejs'), res.locals);
```

nwalters512, line 64: Nit, minor simplification:

```suggestion
      }
      res.render(__filename.replace(/\.js$/, '.ejs'), res.locals);
```

pages/studentAssessmentHomework/studentAssessmentHomework.js
```diff
@@ -47,35 +49,38 @@ router.get('/', function (req, res, next) {
...
+          const groupInfo = await groupAssessmentHelper.getGroupInfo(groupId, groupConfig);
+          res.locals.groupSize = groupInfo.groupSize;
+          res.locals.groupMembers = groupInfo.groupMembers;
+          res.locals.joinCode = groupInfo.joinCode;
+          res.locals.groupName = groupInfo.groupName;
+          res.locals.start = groupInfo.start;
+          res.locals.rolesInfo = groupInfo.rolesInfo;
+
+          if (groupConfig.has_roles) {
+            const result = await groupAssessmentHelper.getAssessmentLevelPermissions(
```
nwalters512, line 75: Nit: idk if we need the word "level" here? `getAssessmentPermissions` has the same meaning to me.

lib/assessment.js
```diff
@@ -81,9 +80,10 @@ module.exports = {
     try {
       assessment_text_templated = ejs.render(assessment.text, context);
     } catch (e) {
-      return ERR(e, callback);
+      debug('renderText()', 'error when rendering text of assessment ', assessment.id, assessment);
+      assessment_text_templated = null;
```
nwalters512, line 84: Why are we catching this error? I think this should continue to fail fast like the original implementation.

sprocs/group_users_insert.sql
```diff
@@ -11,6 +11,9 @@ DECLARE
     arg_group_id bigint;
     arg_cur_size bigint;
     arg_max_size bigint;
+    arg_has_roles boolean;
+    arg_default_group_role_id bigint;
+    arg_required_roles_count bigint;
```
nwalters512, line 16: Nit: I know you're just pattern-matching against existing code, but we shouldn't use the `arg_` prefix for things that are not actually arguments. I'd appreciate it if you could update both the new and old variables here.

sync/course-db.js
```diff
@@ -1208,6 +1208,32 @@ async function validateAssessment(assessment, questions) {
     });
   });
 
+  // Ensure at least one role can assign roles before and during an assessment
+  if (assessment.groupRoles) {
```
nwalters512, line 1212: Is it worth doing some additional validation here? For instance, asserting that `role.minimum < groupMaxSize` and `role.maximum < groupMaxSize`?

database/tables/group_configs.pg
```diff
@@ -3,6 +3,7 @@ columns
     course_instance_id: bigint not null
     date: timestamp with time zone default CURRENT_TIMESTAMP
     deleted_at: timestamp with time zone
+    has_roles: boolean default false
```
nwalters512, line 6: `NOT NULL` constraint?

database/tables/group_user_roles.pg
```diff
@@ -0,0 +1,17 @@
...
+    group_role_id: bigint not null
+    id: bigint not null default nextval('group_user_roles_id_seq'::regclass)
+    user_id: bigint not null
+
+indexes
+    group_user_roles_pkey: PRIMARY KEY (id) USING btree (id)
+    group_user_roles_group_id_user_id_group_role_id_key: UNIQUE USING btree (group_id, user_id, group_role_id)
+    group_user_roles_group_id_key: USING btree (group_id)
+    group_user_roles_group_role_id_key: USING btree (group_role_id)
+    group_user_roles_user_id_key: USING btree (user_id)
```
nwalters512, line 12: Are all these indexes actually necessary? AFAICT, every query of this table always specifies `group_id`, so I suspect that the `UNIQUE` index alone will get us the performance we want.

tests/groupRole.test.js
nwalters512, line 1: I very much appreciate the comprehensive tests, but a ~3400-line file is a bit of a smell to me. Consider breaking this into multiple files in a `tests/groupRoles` directory? Or maybe you can decompose the four (?) massive top-level `describe(...)` flows into more smaller, granular ones? `tests/sync/assessmentsSync.js` is a reasonably good example of tests: each `it(...)` is self-contained and generally a few dozen lines at most.

Feel free to discuss this with me either here or on Slack before you make any changes. You're more familiar with these tests than I am, so maybe there's a good reason that they're implemented they way they currently are.

renzol2: I first refactored some re-used code in the test files (joining/leaving a group, performing group role updates) into functions and it cut down by ~1000 lines. Let us know how this looks! I think we could cut down further if we need to be more concise.

apps/prairielearn/src/pages/studentAssessmentExam/studentAssessmentExam.js
```diff
@@ -1,201 +1,211 @@
...
+              res.locals.uniqueGroupName = uniqueGroupName;
+            }
+            res.locals.permissions = permissions;
+            res.locals.groupsize = 0;
+            res.render(__filename.replace(/\.js$/, '.ejs'), res.locals);
+          }
+        }
+      );
+    } else if (req.body.__action === 'update_group_roles') {
+      await groupAssessmentHelper.updateGroupRoles(
```
nwalters512, line 186: Does this check for `can_assign_roles_at_start` permissions?

renzol2: No, not explicitly in this endpoint. When rendering the page, we only display the form to update group roles to the user(s) with the `can_assign_roles_at_start` permission, but we didn't write anything in the endpoint to explicitly check if the user who sent the request has that permission.

nwalters512: We should _never_ rely on "I didn't show the form to the user" as a form of security. It's always important to check on the backend that the user is actually allowed to perform the action they're taking!

renzol2: We'll write a fix for that here and for homework assessments as well! Thanks for pointing it out!

apps/prairielearn/src/lib/groups.js
```diff
@@ -458,6 +458,14 @@ module.exports.updateGroupRoles = async function (requestBody, assessmentId, use
         "Couldn't access the user's group ID with the provided assessment and user IDs"
       );
     }
+
+    const permissions = await module.exports.getAssessmentPermissions(assessmentId, userId);
+    if (!permissions.can_assign_roles_at_start) {
+      throw new Error(
+        'User does not have permission to assign roles at the start of this assessment'
+      );
```
nwalters512, line 466: This should throw an error with a 403 status code. This is both more accurate *and* keeps this kind of error from showing up in our Sentry reporting of unexpected errors. Search for `error.make(403` in the codebase to see other examples.

apps/prairielearn/src/tests/groupRole.test.js
```diff
@@ -409,6 +409,30 @@ describe('Test group based assessments with custom group roles from student side
...
+      __csrf_token: locals.__csrf_token,
+      ...checkedElementIds,
+    };
+    const res = await fetch(locals.assessmentUrl, {
+      method: 'POST',
+      body: new URLSearchParams(form),
+    });
+
+    // Second user cannot update group roles
+    assert.isNotOk(res.ok);
```
nwalters512, line 433: This should assert that `res.status` is 403 once you make the other change.

PR #6111, opened 2022-08-04
Allow PrairieTest authentication via PrairieLearn

nwalters512:
This PR will allow PrairieTest users to authenticate via PrairieLearn. This simplifies our auth setup, as we don't have to maintain separate Shibboleth configurations for PT anymore.

This uses a flow that's reminiscent of OAuth or SAML:

- A user on PrairieTest clicks a button that takes them to `prairielearn.com/pl/prairietest/auth`
- A user ID is encoded in a JWT and signed with a secret shared key
- The user is sent to a page that will `POST` a form to `prairietest.com/pt/auth/prairielearn/callback`
- PrairieTest will decode and validate the JWT and create a session for the corresponding `user_id`.

This works because PrairieTest and PrairieLearn share a database with the same `users` table, so a user ID is all that we need to pass back and forth.

For folks with access to PrairieTest, see https://github.com/PrairieLearnInc/PrairieTest/pull/738 for the other half of this implementation.


server.js
```diff
@@ -422,6 +423,11 @@ module.exports.initExpress = function () {
   // app.use('/pl/downloadSEBConfig', require('./pages/studentSEBConfig/studentSEBConfig'));
   app.use(require('./middlewares/authn')); // authentication, set res.locals.authn_user
   app.use('/pl/api', require('./middlewares/authnToken')); // authn for the API, set res.locals.authn_user
+
+  if (isEnterprise()) {
+    app.use('/pl/prairietest/auth', require('./ee/auth/prairietest'));
```
nwalters512, line 428: I'm playing around with different code organization in the `ee/` directory.

lib/config.js
```diff
@@ -228,6 +228,13 @@ config.chunksMaxParallelUpload = 20;
  */
 config.isEnterprise = false;
 
+/**
+ * Used to sign JWTs that PrairieLearn provides to PrairieTest for authentication.
+ * PrairieTest should be configured with the same value for
+ * `prairieLearnAuthSecret`.
+ */
+config.prairieTestAuthSecret = 'THIS_IS_A_SECRET_KEY';
```
nwalters512, line 236: It may be more secure to make this `null` by default and only enable this feature if it's non null, that way we can't accidentally deploy both PL and PT with the same default secret.

ee/auth/prairietest.html.js
```diff
@@ -0,0 +1,66 @@
...
+          continueButton.disabled = true;
+
+          (() => {
+            function submitForm() {
+              // Artificially slow down the submission so that the user can see
+              // a little more of what's happening.
+              setTimeout(() => {
+                const form = document.getElementById('form');
+                form.submit();
+              }, 1500);
```
mwest1066, line 49: Do we really care to slow this down? I'm thinking about the login dance that AWS+Google do, where there appear to be about 20 redirects but at the end I just end up logged in, which is what I want, and I don't really care about what all the redirects are doing.

ee/auth/prairietest.js
```diff
@@ -0,0 +1,40 @@
...
+    console.log(res.locals);
+    const key = crypto.createSecretKey(config.prairieTestAuthSecret, 'utf-8');
+
+    // Generate a signed JWT containing just the user ID. PrairieTest shares a
+    // database with PrairieLearn, so it can use the same user ID to look up any
+    // relevant information about PrairieTest.
+    const jwt = await new jose.SignJWT({ user_id: res.locals.authn_user.user_id })
+      .setProtectedHeader({ alg: 'HS512' })
+      .setIssuedAt()
+      .setExpirationTime('1m')
```
mwest1066, line 24: Nice!

lib/config.js
mwest1066: Or give them different default names on the PL/PT sides so that it is broken by default? Maybe `THIS_SHOULD_MATCH_THE_PT_KEY` (and correspondingly `..._PL_...`).

mwest1066 (approved):

ee/auth/prairietest.html.js
nwalters512: Fair! I'll remove the delay.

lib/config.js
nwalters512: I like this a lot! I'll change that.

ee/auth/prairietest.html.js
nwalters512: Done in 1b98b76.

lib/config.js
nwalters512: Done in 190c788.

mwest1066 (approved):

PR #6112, opened 2022-08-04
Implement multitenant SAML authentication

nwalters512:
This PR implements per-institution SAML authentication for Enterprise Edition. Currently it can only be enabled and configured by global administrators, but once we build out institution-level administrator functionality, we can expose it to them too.

This is designed to replace the Shibboleth authentication provider. It had the unfortunate limitation that there could only be a single Shibboleth IdP per PrairieLearn instance. This worked fine when PrairieLearn was predominantly used at UIUC, but now that we're increasing the number of tenants on a single PrairieLearn instance, we need the ability to configure a per-institution authentication provider.


ee/auth/saml/index.js
```diff
@@ -0,0 +1,44 @@
...
+const strategy = new MultiSamlStrategy(
+  {
+    passReqToCallback: true,
+    getSamlOptions(req, done) {
+      getInstitutionSamlProvider(req.params.institution_id).then(
+        (samlProvider) => {
+          if (!samlProvider) {
+            return done(new Error('No SAML provider found for given institution'));
+          }
+          const host = req.headers.host;
```
nwalters512, line 15: I suppose we could use `config.serverCanonicalHost` here?

ee/auth/saml/router.html.js
```diff
@@ -0,0 +1,110 @@
...
+  const hasUidAttribute = !!uidAttribute;
+  const hasUinAttribute = !!uinAttribute;
+  const hasNameAttribute = !!nameAttribute;
+
+  const hasUid = !!uid;
+  const hasUin = !!uin;
+  const hasName = !!name;
+
+  console.log(uidAttribute, uinAttribute, nameAttribute);
+
```
nwalters512, line 23:
```suggestion
```

ee/auth/saml/router.js
```diff
@@ -0,0 +1,126 @@
...
+      })
+    )(req, res);
+
+    // Fetch this institution's attribute mappings.
+    const institutionId = req.params.institution_id;
+    const institutionSamlProvider = await getInstitutionSamlProvider(institutionId);
+    const uidAttribute = institutionSamlProvider.uid_attribute;
+    const uinAttribute = institutionSamlProvider.uin_attribute;
+    const nameAttribute = institutionSamlProvider.name_attribute;
+    console.log(institutionSamlProvider);
```
nwalters512, line 52:
```suggestion
```

ee/institution/admin/sso.html.js
```diff
@@ -0,0 +1,132 @@
...
+
+const InstitutionAdminSso = ({
+  institution,
+  allAuthenticationProviders,
+  institutionSamlProvider,
+  institutionAuthenticationProviders,
+  resLocals,
+}) => {
+  const hasSamlProvider = !!institutionSamlProvider;
+  // TODO: only show authentication providers that were enabled in `config.json`.
```
nwalters512, line 12: Note to self:

package.json
```diff
@@ -133,7 +135,8 @@
         "xml2js": "^0.4.23",
         "xss": "^1.0.13",
         "yargs": "^17.5.1",
-        "yargs-parser": "^21.0.1"
+        "yargs-parser": "^21.0.1",
+        "zod": "^3.17.10"
```
nwalters512, line 139: I'm trying out this instead of `io-ts` like we use in PT. Features seem to be the same, but there's no functional programming hoops to jump though 😂

pages/authLogin/authLogin.js
```diff
@@ -1,14 +1,63 @@
...
+          };
+        })
+        .filter(Boolean);
+    }
+
+    // We could set res.locals.config.hasOauth = false (or
+    // hasAzure) to not display those options inside the CBTF, but
+    // this will also need to depend on which institution we have
+    // detected (e.g., UIUC doesn't want Azure during exams, but
+    // ZJUI does want it).
```
nwalters512, line 58: Is this comment still useful?

server.js
```diff
@@ -417,6 +417,11 @@ module.exports.initExpress = function () {
   app.use('/pl/shibcallback', require('./pages/authCallbackShib/authCallbackShib'));
   app.use('/pl/azure_login', require('./pages/authLoginAzure/authLoginAzure'));
   app.use('/pl/azure_callback', require('./pages/authCallbackAzure/authCallbackAzure'));
+
+  if (isEnterprise()) {
+    app.use('/pl/auth/institution/:institution_id/saml', require('./ee/auth/saml/router'));
+  }
```
nwalters512, line 423: The general approach I'm taking with enterprise features is to mount a single router here in `server.js` and delegate all functionality to the code in the `ee` directory.

ee/institution/admin/saml.sql
```diff
@@ -0,0 +1,44 @@
...
+
+-- BLOCK delete_institution_saml_provider
+WITH deleted_authn_provider AS (
+  DELETE FROM institution_authn_providers
+  USING authn_providers AS ap
+  WHERE
+    ap.name = 'SAML'
+    and authn_provider_id = ap.id
+)
+DELETE FROM saml_providers WHERE institution_id = $institution_id;
```
mwest1066, line 44: Should we log this to `audit_logs`? (And similarly for the INSERT above?)

ee/institution/admin/sso.sql
```diff
@@ -0,0 +1,17 @@
...
+    institution_id,
+    authn_provider_id
+  )
+  SELECT
+    $institution_id AS institution_id,
+    unnest($enabled_authn_provider_ids::bigint[]) as authn_provider_id
+  ON CONFLICT DO NOTHING
+)
+UPDATE institutions SET default_authn_provider_id = $default_authn_provider_id
+WHERE id = $institution_id;
```
mwest1066, line 17: Logging?

package.json
mwest1066: Oh nice! I'm not a fan of all that functional stuff, although I will admit that `io-ts` has been very robust in practice. I have no preference between `zod` and `io-ts` but I would really rather only use one of them. I'm a bit nervous that we might end up using both of them in different bits of our codebase.

pages/authLogin/authLogin.js
mwest1066: No. I think this is superseded by having default providers per institution.

pages/authLogin/authLogin.ejs
```diff
@@ -62,6 +62,29 @@
     border-color: <%- config.shibLinkColors.active.border %>;
     color: <%- config.shibLinkColors.active.text %>;
   }
```
mwest1066, line 64: All of the things above were used to brand the institution login buttons. I assume we are doing away with this? If so, I think we can eliminate all the `config.shibLinkColors` options.

pages/authLogin/authLogin.ejs
```diff
@@ -104,6 +127,18 @@
...
+            Institution sign-on
+          </div>
+          <div class="login-methods">
+          <% institutionAuthnProviders.forEach((provider) => { %>
+            <a href="<%= provider.url %>" class="btn btn-outline-dark btn-block">
+              <%= provider.name %>
+            </a>
+          <% }) %>
+          </div>
+        <% } %>
```
mwest1066, line 141: Could we add a second copy of the most-recently-used institution (stored in a cookie) in a separate area at the top? That is, I'd like to get an obvious "Sign in with Illinois" button when I hit this page (as well as having Illinois being listed in it's usual order in case I read down the list).

pages/partials/navTabs/institutionAdmin.ejs
```diff
@@ -0,0 +1,10 @@
+<%- include('navTab', {
```
mwest1066, line 1: Not that it really matters, but if all the institution admin stuff is going under `ee/` (which I think it should) then should this file go in `ee/partials` or similar? It's also fine if it's more convenient to keep it where it is.

pages/partials/navbarInstitution.ejs
```diff
@@ -0,0 +1,11 @@
+<li class="nav-item">
```
mwest1066, line 1: This could also conceivably go in `ee/partials`.

pages/partials/logout.ejs
```diff
@@ -1,10 +1,5 @@
 <% if (typeof authn_user != 'undefined') { %>
-   <% if (authn_provider_name == 'Shibboleth' || authn_provider_name == 'Google' || authn_provider_name == 'Azure'
-          || authn_provider_name == 'LTI') { %>
-       <a class="dropdown-item" href="<%= plainUrlPrefix %>/logout">Log out</a>
-   <% } else { %>
-       <a class="dropdown-item disabled">Log out</a>
-   <% } %>
+   <a class="dropdown-item" href="<%= plainUrlPrefix %>/logout">Log out</a>
```
mwest1066, line 2: Excellent.

mwest1066 (approved):

database/tables/saml_providers.pg
```diff
@@ -0,0 +1,18 @@
+columns
+    certificate: text not null
+    id: bigint not null default nextval('saml_providers_id_seq'::regclass)
+    institution_id: bigint
```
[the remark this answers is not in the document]
nwalters512: Yes!

package.json
nwalters512: I decided to see how bad it would be to convert PT to zod and the answer was "not very" 😂 
 https://github.com/PrairieLearnInc/PrairieTest/pull/748

nwalters512: We intend to land this change to PrairieTest, so I feel good about using it here in PL.

mwest1066: 💯

database/tables/saml_providers.pg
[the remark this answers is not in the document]
nwalters512: Done in a57b069

migrations/20220808163441_262_authn_providers__saml__add.sql
```diff
@@ -0,0 +1 @@
+INSERT INTO authn_providers (id, name) VALUES (5, 'SAML');
```
nwalters512, line 1: TODO: make this idempotent?

nwalters512: @mwest1066 should this work?

```suggestion
INSERT INTO authn_providers (id, name) VALUES (5, 'SAML') ON CONFLICT DO NOTHING;
```

mwest1066: Yep, should work.

pages/authLogin/authLogin.js
nwalters512: I removed this comment in 92590a8.

pages/authLogin/authLogin.ejs
nwalters512: We found that some folks are still using these in production, so we'll leave them alone for now.

pages/authLogin/authLogin.ejs
nwalters512: Would you be willing to pull this out into a separate ticket + PR?

mwest1066: Yep! https://github.com/PrairieLearn/PrairieLearn/issues/6137

pages/partials/navTabs/institutionAdmin.ejs
nwalters512: I'll leave these where they are for now, I don't think there's a ton of value in moving them. We can easily move in the future if we develop different conventions for EE code.

pages/partials/navbarInstitution.ejs
nwalters512: As mentioned on the other comment, I'll leave this where it is for now.

pages/authLogin/authLogin.ejs
nwalters512: Thanks! Going to mark this resolved; we can pick that one up later.

ee/institution/admin/saml.sql
nwalters512: Yep, that seems reasonable. I won't expose them anywhere yet, but it'll at least be good to have that raw data available if we need it.

nwalters512: I added audit logs in some recent commits.

ee/institution/admin/sso.sql
nwalters512: I added audit logs in some recent commits.

pages/partials/navTabs/institutionAdmin.ejs
mwest1066: Makes sense!

pages/partials/navbarInstitution.ejs
mwest1066: Agreed.

ee/institution/admin/saml.sql
```diff
@@ -2,47 +2,84 @@
...
+    row_id,
+    old_state,
+    new_state
+  ) SELECT
+    $authn_user_id as authn_user_id,
+    'saml_providers' as table_name,
+    (CASE WHEN esp.id IS NULL THEN 'insert' ELSE 'update' END) as action,
+    $institution_id as institution_id,
+    usp.id AS row_id,
+    (to_jsonb(esp) || jsonb_build_object('private_key', 'REDACTED')) AS old_state,
```
mwest1066, line 54: With my security hat on, I like the general policy of stripping sensitive info from logs. But with my customer support hat on, I'm less sure. Am I correct in thinking that the `private_key` is never exposed to the user (or outside the server)? If so, I'm not sure I see what the attack is that would make use of old private keys.

However, I also don't think that in practice we'll need to recover old private keys, so I'm not very worried one way or the other.

ee/institution/admin/saml.sql
```diff
@@ -2,47 +2,84 @@
...
+      'id', usp.id,
+      'sso_login_url', usp.sso_login_url,
+      'issuer', usp.issuer,
+      'certificate', usp.certificate,
+      'uid_attribute', usp.uid_attribute,
+      'uin_attribute', usp.uin_attribute,
+      'name_attribute', usp.name_attribute,
+      'public_key', usp.public_key,
+      'private_key', 'REDACTED' 
+    ) AS new_state
```
mwest1066, line 65: Could this be written using a similar expression to the `to_json(esp)` line above?

ee/institution/admin/saml.sql
```diff
@@ -51,14 +88,28 @@ WITH deleted_authn_provider AS (
...
     old_state
-  ) SELECT ...
-), audit_logs_deleted_saml_provider (
+  ) SELECT
+    $authn_user_id as authn_user_id,
+    'institution_authn_providers' as table_name,
+    'delete' as action,
+    $institution_id as institution_id,
+    dap.id as row_id,
+    to_jsonb(dap) AS old_state
```
mwest1066, line 97: Note that the matching between insert columns and values is by order, not name, so the `AS ...` entries above are not necessary. I'm happy to have them as documentation if that was your intention.

ee/institution/admin/saml.sql
```diff
@@ -51,14 +88,28 @@ WITH deleted_authn_provider AS (
...
     row_id,
     old_state
-  ) SELECT ...
+  ) SELECT
+    $authn_user_id as authn_user_id,
+    'saml_providers' as table_name,
+    'delete' as action,
+    $institution_id as institution_id,
+    dsp.id as row_id,
+    (to_jsonb(dsp) || jsonb_build_object('private_key', 'REDACTED')) AS old_state
```
mwest1066, line 113: Similar comment here about the `AS ...` names being unnecessary (but fine to leave if you prefer).

mwest1066 (approved):

ee/institution/admin/saml.sql
nwalters512: > Am I correct in thinking that the private_key is never exposed to the user (or outside the server)?

Correct, yeah. I'm thinking forward to a future where, like in PT, we have user-facing audit logs pages, and I think even in that case we'd never want the possibility of exposing the private key. That said, yeah, I don't know what vulnerabilities we'd actually get if folks had our private key; I think at worst, if they were somehow able to intercept the redirect from the IDP, they'd be able to decrypt the IDP identity assertion?

ee/institution/admin/saml.sql
nwalters512: Yes! Let me change that.

ee/institution/admin/saml.sql
nwalters512: Ah fair - I'll remove them!

ee/institution/admin/saml.sql
nwalters512: Removed in eaa756d.

ee/institution/admin/saml.sql
nwalters512: Removed in eaa756d.

ee/institution/admin/saml.sql
nwalters512: Changed in eaa756d.

PR #6117, opened 2022-08-07
Fix disappearing feedback in gradeAsync

### LINKED ISSUE #6116: ansiOutput question is no longer working
This question is no longer working for some reason: https://github.com/PrairieLearn/PrairieLearn/tree/master/exampleCourse/questions/demo/autograder/ansiOutput

echuber2:
Fixes #6116 

This was a subtle typo in the recent async refactoring in `freeform.js`: `data.feedback` should be `resultData.feedback`.


nwalters512 (approved):
Great find! It's so easy for little bugs to slip into those huge refactors 😅

Issue #2590, opened 2020-06-19
Text underneath pl-checkbox multiple choice is inaccurate


astrosticks:
From the description of the requirements, "Select all possible options" seems like it should allow you to submit zero options, but it actually won't. 
![image](https://user-images.githubusercontent.com/32315760/85088333-7f332580-b1a5-11ea-8acc-6256bcd3394a.png)
![image](https://user-images.githubusercontent.com/32315760/85088362-89552400-b1a5-11ea-8cbf-af2927c679c0.png)

reteps:
It looks like allow-blank was never implemented for pl-checkbox. We should do that, and make sure the text is dynamic based on that case.

PR #6119, opened 2022-08-07
Add allow-blank attribute to pl-checkbox and pl-multiple-choice

natmarbal:
Motivation for adding `allow-blank`: in longer, multiple-part homework questions, it is difficult to use the `pl-checkbox` element due to the fact that it cannot be be left blank without the invalid warnings.

Changes:
- added an `allow-blank` attribute in `pl-checkbox.py`
- added subsequent edits to `prepare()`, `render()`, `parse()` and `_get_min_options_to_select()`
- updated the docs to include an `allow-blank` description

Behaviour:
- `allow-blank` is set to `false` by default. If `allow-blank = false`, there is a normal invalid message when checkbox is left blank.
- If `allow-blank = true`, the checkbox input is marked as incorrect, not invalid.
- If `allow-blank = true` and `min-select` >= 1, `allow-blank` takes precedence.
- If `allow-blank = false` and `min-select` >=1, there is normal `min-select` behaviour with respect to the invalid messages.

cc: @cbm755


natmarbal:

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -365,23 +378,26 @@ def parse(element_html, data):
...
-    # Check that the selected options are a subset of the valid options
-    # FIXME: raise ValueError instead of treating as parse error?
-    submitted_key_set = set(submitted_key)
-    all_keys_set = set(all_keys)
-    if not submitted_key_set.issubset(all_keys_set):
-        one_bad_key = submitted_key_set.difference(all_keys_set).pop()
-        # FIXME: escape one_bad_key
-        data['format_errors'][name] = 'You selected an invalid option: {:s}'.format(str(one_bad_key))
-        return
+    if allow_blank == False:
```
cbm755, line 386: options here are `if allow_blank is False` (if its important to distinguish between False and None). 

If not important (usual case), the I think `if not allow_blank:`

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -514,10 +530,14 @@ def _get_min_options_to_select(element, default_val):
...
     detailed_help_text = pl.get_boolean_attrib(element, 'detailed-help-text', DETAILED_HELP_TEXT_DEFAULT)
 
+    allow_blank = pl.get_boolean_attrib(element, 'allow-blank', ALLOW_BLANK_DEFAULT)
+
     if pl.has_attrib(element, 'min-select'):
         min_options_to_select = pl.get_integer_attrib(element, 'min-select')
     elif pl.has_attrib(element, 'min-correct') and detailed_help_text:
         min_options_to_select = pl.get_integer_attrib(element, 'min-correct')
+    elif allow_blank == True:
+        min_options_to_select = MIN_SELECT_BLANK
```
cbm755, line 540: I think this should be `elif allow_blank:`

cbm755:
> If allow-blank = true, the checkbox input is marked as incorrect, not invalid.

Is it possible to setup a question where the correct answer is to leave them all blank?

natmarbal:
No, it's not currently possible to have the correct answer be a blank checkbox. The following error gets thrown:
``` 
  File "/PrairieLearn/lib/python-caller-trampoline.py", line 182, in <module>
    worker_loop()
  File "/PrairieLearn/lib/python-caller-trampoline.py", line 128, in worker_loop
    val = method(*args)
  File "/course/elements/pl-checkbox-custom/pl-checkbox-custom.py", line 61, in prepare
    raise ValueError('At least one option must be true.')
ValueError: At least one option must be true.
  File "/PrairieLearn/lib/python-caller-trampoline.py", line 194, in <module>
```

For multiple-part questions, it would be wrong to have a blank checkbox be the correct answer, because students who haven't yet reached the checkbox section would have it marked correct "by accident".

natmarbal:
New Changes:
- added an allow-blank attribute in pl-multiple-choice.py
- updated the docs to include an allow-blank description for that element

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -189,6 +197,8 @@ def render(element_html, data):
             else:
                 number_correct_text = ''
```
eliotwrobson, line 198: I think that the `number_correct_text` should be altered by the `allow-blank` flag (see https://github.com/PrairieLearn/PrairieLearn/issues/2590). The error message mentioned in the linked issue is set on line 242 here (and uses the `number_correct_text`.

elements/pl-checkbox/pl-checkbox.py
```diff
@@ -78,11 +83,14 @@ def prepare(element_html, data):
     if not (0 <= min_incorrect <= max_incorrect <= len_incorrect):
         raise ValueError('INTERNAL ERROR: incorrect number: (%d, %d, %d, %d)' % (min_incorrect, max_incorrect, len_incorrect, len_correct))
 
-    min_select = pl.get_integer_attrib(element, 'min-select', MIN_SELECT_DEFAULT)
-    max_select = pl.get_integer_attrib(element, 'max-select', number_answers)
+    if allow_blank == True:
```
eliotwrobson, line 86: I think this is a little overcomplicated. I think this can be replaced by something like:
```python
min_select_default = MIN_SELECT_BLANK if allow_blank else MIN_SELECT_DEFAULT
min_select = pl.get_integer_attrib(element, 'min-select', min_select_default)
if min_select < min_select_default:
    raise ValueError("...")

File: elements/pl-checkbox/pl-checkbox.py [hunk 5]
```diff
@@ -218,7 +228,10 @@ def render(element_html, data):
                     insert_text = f' at most <b>{max_options_to_select}</b> options.'
                 else:
                     # This is the case where we reveal nothing about min_options_to_select and max_options_to_select.
-                    insert_text = ' at least 1 option.'
+                    if allow_blank == True:
```
eliotwrobson, line 231:
```suggestion
                    if allow_blank:
```

eliotwrobson (changes requested):
Looks like a good feature to add overall. A couple stylistic things (including handling that error message issue) but logic is sound.

elements/pl-checkbox/pl-checkbox.py
natmarbal: Thank you for your review!

Could the issue mentioned in #2590 please be clarified? I'm not quite sure what the bug is, since the following box seems to be quite explicit about the fact that at least one checkbox must be selected for a correct answer:
Screen Shot 2023-01-07 at 15 53 03

As I understand, the number of correct options checked, and thus graded, is separate from `allow-blank`, which is just the attribute that allows a blank submission without the usual invalid warning.

natmarbal:
@eliotwrobson Thank you for the changes in #8143.

I'll close this PR, since it's no longer needed.

PR #6128, opened 2022-08-10
Set exam mode based on pt_location_networks

mwest1066:


database/tables/pt_location_networks.pg
```diff
@@ -0,0 +1,10 @@
+columns
```
nwalters512, line 1: Can you update the `fix-pl-db.sh` in the PrairieTest repo to account for these new tables?

sprocs/ip_to_mode.sql
```diff
@@ -44,7 +49,48 @@ BEGIN
         AND (date BETWEEN r.access_start AND r.access_end);
 
     IF FOUND THEN
-        mode := 'Exam';
+        -- We have a checked-in pt_reservation. Check whether we are in a
+        -- testing center location that requires network filtering.
```
nwalters512, line 53: The use of "we are in..." throughout the following few comments was a little confusing on the first read through since it's used to mean two different things. I'd recommend disambiguating by saying either "the reservation is in a testing center location..." or "the user is physically in the location...".

nwalters512 (approved):
LGTM!

SethPoulsen:
Is "pt" supposed to stand for "physical testing"?

mwest1066:
PT = PrairieTest

database/tables/pt_location_networks.pg
mwest1066: Good point. Done in https://github.com/PrairieLearnInc/PrairieTest/pull/776

sprocs/ip_to_mode.sql
mwest1066: Excellent point. The two uses of "we are in" were somehow clear in my head while writing, but they are completely unclear while reading! Fixed in 8ba21d5649f19cf40425509c18bb29f85c5d6ff2

PR #6129, opened 2022-08-11
Rework migrations to use timestamps instead of indexes

nwalters512:
In retrospect, it was never a good idea to use unique indexes to identify database migrations. They frequently cause merge conflicts, and those merge conflicts get especially hairy when we're deploying from branches to our staging or production environments.

This PR starts of the process of changing all our migrations to use timestamps instead of indexes. It keeps the index around so we can match up migration files with existing rows in the `migrations` table, and introduces a lot of code to ensure there isn't data loss during the transition. Read the code comments for more details!


prairielib/lib/migrations.js
```diff
@@ -34,136 +35,272 @@ module.exports.init = function (dir, proj, callback) {
...
+        //
+        // First, we'll deploy a version of the code that still has the index embedded
+        // in the filename. We'll use that to sync the new timestamps to the migrations
+        // table.
+        //
+        // Only after that will we deploy a version that removes the index from the
+        // migration filenames. As part of that, we'll add some safety checks that
+        // will ensure that if we're deploying a migrations directory that doesn't
+        // have *any* indexes in the filenames, we'll error out and prompt the user
+        // to deploy an earlier version that still has timestamps.
```
nwalters512, line 224: We could have also done this matching by looking at the text portion of the migration filename, but I couldn't be 100% sure that we never changed the names of migrations after they were applied, so that felt less robust. We *could* validate that there's a 1:1 mapping from existing files (the text portion) to the each entry in the `filename` column in the database and error if that's not the case, giving the operator an opportunity to manually fix the filename in the database. Thoughts?

prairielib/lib/migrations.js
```diff
@@ -34,136 +35,272 @@ module.exports.init = function (dir, proj, callback) {
   });
 };
 
+/**
+ * Timestamp prefixes will be of the form `YYYYMMDDHHMMSS`, which will have 14 digits.
```
mwest1066, line 39: Do we care about the timezone? I think not really, but maybe we should define these as UTC just for completeness? And could include a comment about how to make these timestamp strings? The simplest expression I could find is:
```
node -e "console.log(new Date().toISOString().replace(/\D/g,'').slice(0,14))"
```
Maybe we should add a `README.md` to the `migrations/` directory with information about the filename format and instructions on how to generate a new name?

prairielib/lib/migrations.js
```diff
@@ -34,136 +35,272 @@ module.exports.init = function (dir, proj, callback) {
...
+
+    if (!match) {
+      throw new Error(`Invalid migration filename: ${mf}`);
+    }
+
+    const timestamp = match[1] ?? null;
+    const index = match[2] ?? null;
+
+    if (timestamp === null && index === null) {
+      throw new Error(`Migration ${mf} is missing a timestamp or index`);
```
mwest1066, line 70:
```suggestion
      throw new Error(`Migration ${mf} does not have either a timestamp or index`);
```

prairielib/lib/migrations.js
```diff
@@ -34,136 +35,272 @@ module.exports.init = function (dir, proj, callback) {
...
+      .map((m) => m.filename)
+      .join(', ');
+    throw new Error(`The following migration files are missing indexes: ${filesMissingIndex}`);
+  }
+
+  return migrations;
+}
+
+/**
+ * Determine the ordering of migrations. We will have already validates that either all migrations
```
mwest1066, line 136:
```suggestion
 * Determine the ordering of migrations. We will have already validated that either all migrations
```

prairielib/lib/migrations.js
```diff
@@ -34,136 +35,272 @@ module.exports.init = function (dir, proj, callback) {
...
+  // If no migrations have ever been run, run them all.
+  if (executedMigrations.length === 0) {
+    return migrationFiles;
+  }
+
+  // If our migration files have timestamps, use them to determine which
+  // have or have not executed. Otherwise, use the index.
+  if (migrationFiles.every((m) => m.timestamp)) {
+    return migrationFiles.filter(
+      (m) => !executedMigrations.some((e) => e.timestamp === m.timestamp)
```
mwest1066, line 170: Use a `Set` to avoid the O(N^2) filter? As in https://github.com/PrairieLearnInc/PrairieTest/commit/168c783e62aa9d2f36980610b3d73ff9506ab1da

prairielib/lib/migrations.js
```diff
@@ -34,136 +35,272 @@ module.exports.init = function (dir, proj, callback) {
...
+    );
+  } else {
+    return migrationFiles.filter((m) => !executedMigrations.some((e) => e.index === m.index));
```
mwest1066, line 173: Technically this is a change in behavior because we used to filter on `index > maxAppliedIndex`. I don't _think_ that we would ever have ended up in a situation where we skipped some migrations, but I wanted to at least point this out in case I'm missing some other impact of this change.

tools/rewrite-migrations.js
```diff
@@ -0,0 +1,56 @@
+const fs = require('fs-extra');
+const execa = require('execa');
+
+const prefixZero = (value) => ('0' + value).slice(-2);
+
+function formatDate(date) {
+  return `${date.getFullYear()}${prefixZero(date.getMonth() + 1)}${prefixZero(
+    date.getDate()
+  )}${prefixZero(date.getHours())}${prefixZero(date.getMinutes())}${prefixZero(date.getSeconds())}`;
```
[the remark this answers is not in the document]
mwest1066: Another option would be `date.toISOString().replace(/\D/g,'').slice(0,14)` which will also ensure these are in UTC.

prairielib/lib/migrations.test.js
```diff
@@ -0,0 +1,226 @@
...
+const tmp = require('tmp-promise');
+const fs = require('fs-extra');
+
+const {
+  readAndValidateMigrationsFromDirectory,
+  sortMigrationFiles,
+  getMigrationsToExecute,
+} = require('./migrations');
+
+async function withMigrationFiles(files, fn) {
```
mwest1066, line 14: I really like these `with` helpers that you are using recently!

mwest1066 (approved):
I like this PR a lot! It should really help when we've got multiple PRs in flight.

prairielib/lib/migrations.js
nwalters512: Glad you remembered that we've don this! The current strategy in this PR is "auto-fix based on index when we first switch to timestamps", so I'll stick with that.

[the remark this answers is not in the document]
nwalters512: Yes! Thanks.

tools/rewrite-migrations.js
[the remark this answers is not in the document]
nwalters512: Ack! I keep forgetting about `Intl.DateTimeFormat` - thanks for the reminder. I intend to delete this file before merging this branch as there's no reason for us to ever run it in the future (besides in PT).

prairielib/lib/migrations.test.js
nwalters512: <3 Still wishing JS had a `with` keyword like Python's that didn't uh, completely suck like the [*actual* JS `with` keyword](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/with) 😭

prairielib/lib/migrations.js
nwalters512: This is actually deliberate - in the case of working with multiple branches, IMO it's a feature not a bug that we'd apply any migrations that got added when we `git merge master` into a working branch.

prairielib/lib/migrations.js
nwalters512: Fixed up in 43ea2d8.

prairielib/lib/migrations.js
nwalters512: I like that idea, yeah. Maybe even a Makefile target like `make new-migration` that'll run that bit of Node code? I'll write some docs and think about a Makefile change.

nwalters512: I added a README in 53b4703.

prairielib/lib/migrations.js
```diff
@@ -34,136 +35,272 @@ module.exports.init = function (dir, proj, callback) {
...
-          },
-          (err) => {
-            if (ERR(err, callback)) return;
-            callback(null);
+        // Second pass: reconcile the timestamps with the list of migrations in the database.
+        // This should only matter a single time (the first time this code is deployed after
+        // adding timestamps to all migrations).
+        for (const { filename, timestamp, index } of migrationFiles) {
+          if (!timestamp) {
+            continue;
```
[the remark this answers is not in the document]
nwalters512: I originally didn't do that because I wanted to make sure this wouldn't block PT migrations (i.e. I wanted to be able to consume a new PL version without immediately cutting over to new migration filenames). Thinking about that some more though, we should probably just rip the bandaid off and enforce that.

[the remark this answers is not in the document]
nwalters512: I made timestamps mandatory in 3a56633.

prairielib/lib/migrations.js
mwest1066: Oh yes, I'm 100% on board for the future behavior working like this. I was only highlighting that it will also apply retroactively (I think?). But I don't think this is a concern in practice.

prairielib/lib/migrations.js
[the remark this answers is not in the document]
mwest1066: Oh right, good point. But I agree with also converting PT immediately.

prairielib/lib/migrations.js
nwalters512: Ah, yeah - if we for some reason have unexecuted migrations lying around, we probably have bigger problems 😂.

prairielib/lib/migrations.js
```diff
@@ -34,136 +34,274 @@ module.exports.init = function (dir, proj, callback) {
...
+
+    if (!match) {
+      throw new Error(`Invalid migration filename: ${mf}`);
+    }
+
+    const timestamp = match[1] ?? null;
+    const index = match[2] ?? null;
+
+    if (timestamp === null) {
+      throw new Error(`Migration filename ${mf} does not have a timestamp`);
```
nwalters512, line 69:
```suggestion
      throw new Error(`Migration ${mf} does not have a timestamp`);
```

mwest1066 (approved):

PR #6133, opened 2022-08-15
Remove Broken PrairieGrader link from docs

SethPoulsen:


docs/externalGrading.md
```diff
@@ -250,7 +250,7 @@ For a working example of this, see `PrairieLearn/elements/pl_file_upload/pl_file
 
 ## Running locally for development
 
-In production, PrairieLearn runs external grading jobs on a distributed system called [https://github.com/PrairieLearn/PrairieGrader](PrairieGrader). This system uses a variety of AWS services to efficiently run many jobs in parallel. When developing questions locally, you won't have access to this infrastructure, but PrairieLearn allows you to still run external grading jobs locally with a few workarounds.
+In production, PrairieLearn runs external grading jobs on a distributed system called PrairieGrader. This system uses a variety of AWS services to efficiently run many jobs in parallel. When developing questions locally, you won't have access to this infrastructure, but PrairieLearn allows you to still run external grading jobs locally with a few workarounds.
```
nwalters512, line 253: Thanks for catching this! Let's just drop the name "PrairieGrader" entirely; we no longer really use it to refer to that system.

```suggestion
In production, PrairieLearn runs external grading jobs on a distributed system that uses a variety of AWS services to efficiently run many jobs in parallel. When developing questions locally, you won't have access to this infrastructure, but PrairieLearn allows you to still run external grading jobs locally with a few workarounds.
```

nwalters512 (approved):

PR #6136, opened 2022-08-16
Fix database migration application

nwalters512:
To my credit, the code was so robust that I didn't notice this problem during my initial testing since the timestamp would always be synced on the second boot 😂

### DIFF
--- a/prairielib/lib/migrations.sql
+++ b/prairielib/lib/migrations.sql
@@ -12,13 +12,13 @@ CREATE TABLE IF NOT EXISTS migrations (
 
 -- BLOCK add_projects_column
 ALTER TABLE migrations ADD COLUMN IF NOT EXISTS project TEXT DEFAULT 'prairielearn';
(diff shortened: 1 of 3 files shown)


prairielib/lib/migrations.sql
```diff
@@ -5,7 +5,9 @@ CREATE TABLE IF NOT EXISTS migrations (
   index INTEGER,
   project TEXT DEFAULT 'prairielearn',
   applied_at TIMESTAMP WITH TIME ZONE,
-  UNIQUE(project, index)
+  timestamp TEXT,
+  UNIQUE(project, index),
+  UNIQUE(project, filename)
```
nwalters512, line 10: This change wasn't a bug, but it did result in an extraneous log being printed when populating a fresh database.

mwest1066 (approved):

prairielib/lib/migrations.sql
nwalters512, line 10:
```suggestion
  timestamp TEXT,
  UNIQUE(project, index),
  UNIQUE(project, timestamp)
```

PR #6138, opened 2022-08-17
Fix assessment authorization with modules

mwest1066:
PR #4360 introduced modules and does an unconditional JOIN on `assessment_modules` in `selectAndAuthzAssessment.sql`. However, modules are optional and so any assessment without one will return no rows from this query, thus failing authorization.

I will check other authz paths as well before merging this.

### DIFF
--- a/middlewares/selectAndAuthzAssessment.sql
+++ b/middlewares/selectAndAuthzAssessment.sql
@@ -8,7 +8,7 @@ SELECT
 FROM
     assessments AS a
     JOIN assessment_sets AS aset ON (aset.id = a.assessment_set_id)
-    JOIN assessment_modules AS am ON (am.id = a.assessment_module_id)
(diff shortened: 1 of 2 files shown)


mwest1066:
@nwalters512 I actually can't see why the `JOIN assessment_modules` is in `selectAndAuthzAssessment.sql` at all. There is no corresponding JOIN in `selectAndAuthzAssessmentInstances.sql` so we can't be using the module data anywhere downstream (and I can't find any usage). Can you see any reason that we shouldn't just rip out this JOIN entirely?

mwest1066:
Actually I think it's used here: https://github.com/PrairieLearn/PrairieLearn/blob/ad1c19c4c626bad593247786f77a473ca968d3fb/pages/instructorAssessmentSettings/instructorAssessmentSettings.ejs#L57-L61

But that code is also wrong because it's assuming the module exists.

nwalters512 (approved):

PR #6139, opened 2022-08-17
feat: Support sets and tuples in pl-symbolic-input

SethPoulsen:


eliotwrobson:
@SethPoulsen Should this be closed and reopened later if people want this feature? The backend for the symbolic input element has been refactored substantially, so this feature will need a significant rework before being ready to go.

SethPoulsen:
Sure, if that helps you get organized, I don't mind.

eliotwrobson:
Sounds good. I think a lot of the use cases for this are handled by the planned symbolic matrix element: https://github.com/PrairieLearn/PrairieLearn/discussions/7027

If there are requests for something outside of the above we can reexamine.

PR #6143, opened 2022-08-18
Fix grouping of assessments by module

nwalters512:
This should fix the issue that @mfsilva22 reported where assessments weren't being grouped correctly on the assessment pages.

I believe this works because the order of rows in the partition need to match that of the result set so that we put the "new section" marker in the correct place.

As for why this issue showed up in production and not locally: this is because ordering was dependent on IDs, which are not consistent between environments.

### DIFF
--- a/pages/instructorAssessments/instructorAssessments.sql
+++ b/pages/instructorAssessments/instructorAssessments.sql
@@ -46,7 +46,7 @@ SELECT
         LAG (CASE WHEN $assessments_group_by = 'Set' THEN aset.id ELSE am.id END)
         OVER (
             PARTITION BY (CASE WHEN $assessments_group_by = 'Set' THEN aset.id ELSE am.id END)
-            ORDER BY aset.id, a.order_by, a.id NULLS FIRST
+            ORDER BY aset.number, a.order_by, a.id NULLS FIRST
         ) IS NULL
     ) AS start_new_assessment_group,
     (CASE WHEN $assessments_group_by = 'Set' THEN aset.heading ELSE am.heading END) AS assessment_group_heading,
--- a/pages/instructorAssessments/instructorAssessments.sql
+++ b/pages/instructorAssessments/instructorAssessments.sql
@@ -46,7 +46,7 @@ SELECT
         LAG (CASE WHEN $assessments_group_by = 'Set' THEN aset.id ELSE am.id END)
         OVER (
             PARTITION BY (CASE WHEN $assessments_group_by = 'Set' THEN aset.id ELSE am.id END)
-            ORDER BY aset.number, a.order_by, a.id NULLS FIRST
+            ORDER BY aset.number, a.order_by, a.id
         ) IS NULL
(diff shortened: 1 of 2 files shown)


mwest1066:
Great fix!

> I believe this works because the order of rows in the partition need to match that of the result set so that we put the "new section" marker in the correct place.

That's correct. We are marking the first row in each partition with `start_new_assessment_group = TRUE`. But our understanding of "first" here has to be consistent. This is why the partition ORDER BY has to match the overall query ORDER BY, which this PR fixes.

> As for why this issue showed up in production and not locally: this is because ordering was dependent on IDs, which are not consistent between environments.

Yep, I agree with this explanation.

pages/instructorAssessments/instructorAssessments.sql
```diff
@@ -46,7 +46,7 @@ SELECT
         LAG (CASE WHEN $assessments_group_by = 'Set' THEN aset.id ELSE am.id END)
         OVER (
             PARTITION BY (CASE WHEN $assessments_group_by = 'Set' THEN aset.id ELSE am.id END)
-            ORDER BY aset.id, a.order_by, a.id NULLS FIRST
+            ORDER BY aset.number, a.order_by, a.id NULLS FIRST
```
mwest1066, line 49: Incidentally, I believe the NULLS FIRST is unnecessary here, as none of these values can be NULL. @nwalters512 correctly raised this point during review in https://github.com/PrairieLearn/PrairieLearn/pull/4360/files#r783535458 but that was never resolved.

mwest1066 (approved):

pages/instructorAssessments/instructorAssessments.sql
nwalters512: Great, I'll remove it so as not to confuse ourselves in the future 🙂

pages/studentAssessments/studentAssessments.sql
```diff
@@ -155,5 +155,4 @@ WHERE
 ORDER BY 
     CASE WHEN $assessments_group_by = 'Module' THEN assessment_module_number END, 
     CASE WHEN $assessments_group_by = 'Module' THEN assessment_module_id END,
-    assessment_set_number, assessment_order_by, assessment_id, assessment_instance_number 
-    NULLS FIRST;
+    assessment_set_number, assessment_order_by, assessment_id, assessment_instance_number;
```
nwalters512, line 158: @mwest1066 I grepped for `NULLS FIRST` and this was the only other occurrence. I believe it's safe to remove, but would appreciate a second look!

pages/instructorAssessments/instructorAssessments.sql
nwalters512: Removed in 3e2955c.

pages/studentAssessments/studentAssessments.sql
mwest1066: I agree this is also safe to remove. (I hope I don't regret saying this 😄)

PR #6146, opened 2022-08-19
Use correct branch name when performing git operations

### LINKED ISSUE #6145: Incorrect file edit resets when using `main` rather than `master`
PL allows course GitHub repos to use an arbitrary branch for syncing, which is normally set to either `'main'` (new repos) or `'master'` (legacy repos). The name of this branch is stored in: https://github.com/PrairieLearn/PrairieLearn/blob/5cd1fd81e15b0529a9e6af10be7da32462b500f0/database/tables/pl_courses.pg#L2

In some places we correctly use this branch name when resetting: https://github.com/PrairieLearn/PrairieLearn/blob/5cd1fd81e15b0529a9e6af10be7da32462b500f0/pages/shared/syncHelpers.js#L135

But there appear to be at least two places where we've kept the old hard-coded use of `'master'`: https://github.com/PrairieLearn/PrairieLearn/blob/5cd1fd81e15b0529a9e6af10be7da32462b500f0/lib/editors.js#L168
https://github.com/PrairieLearn/PrairieLearn/blob/5cd1fd81e15b0529a9e6af10be7da32462b500f0/pages/instructorFileEditor/instructorFileEditor.js#L771

These both need to be changed to use `course.branch` instead of `'master'`.

nwalters512:
This should fix #6145.

### DIFF
--- a/lib/editors.js
+++ b/lib/editors.js
@@ -165,7 +165,7 @@ class Editor {
           type: 'reset_from_git',
           description: 'Reset state to remote git repository',
           command: 'git',
-          arguments: ['reset', '--hard', 'origin/master'],
+          arguments: ['reset', '--hard', `origin/${this.course.branch}`],
           working_directory: this.course.path,
           env: gitEnv,
           on_success: on_success,
--- a/pages/instructorFileEditor/instructorFileEditor.js
+++ b/pages/instructorFileEditor/instructorFileEditor.js
@@ -768,7 +768,7 @@ function saveAndSync(fileEdit, locals, callback) {
         type: 'reset_from_git',
         description: 'Reset state to remote git repository',
         command: 'git',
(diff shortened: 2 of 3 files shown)


lib/editors.js
```diff
@@ -165,7 +165,7 @@ class Editor {
           type: 'reset_from_git',
           description: 'Reset state to remote git repository',
           command: 'git',
-          arguments: ['reset', '--hard', 'origin/master'],
+          arguments: ['reset', '--hard', `origin/${this.course.branch}`],
```
mwest1066, line 168: I think I would have blindly written `locals` here as well, so I'm glad you got it right!

mwest1066 (approved):

lib/editors.js
nwalters512: The benefits of local pattern-matching 😛

PR #6150, opened 2022-08-19
Fix error handling during SAML authentication

nwalters512:
Without this, an error during SAML authentication would crash the server, which is obviously no good!


mwest1066 (approved):
Good call. I find async/await so much easier to get right than this kind of promise code.

nwalters512:
Absolutely! If only `getSamlOptions` could be an async function 😢

PR #6151, opened 2022-08-19
Show friendly error message if user uses unsupported auth provider

nwalters512:
This will provide an easy way for users to recover from trying to authenticate via a provider that isn't supported by their institution. This should be a big improvement over the current behavior, which is to dump students onto a page that just says "ABC authentication provider is not allowed for institution XYZ".


ee/auth/saml/router.js
```diff
@@ -83,8 +83,30 @@ router.post(
...
+      // Check if this is because the provider is not allowed.
+      if (err.message?.includes('authentication provider is not allowed for institution')) {
+        // Parse institution ID from error message.
+        const institutionIdMatch = err.message.match(/\((\d+)\)$/);
+        if (institutionIdMatch) {
+          const institutionId = institutionIdMatch[1];
+          res.redirect(`/pl/auth/institution/${institutionId}/not_allowed`);
+          return;
+        }
+      }
```
nwalters512, line 104: I copy/pasted this block of code around, which is probably no worse than all the other auth code we've duplicated in 4 places?

Grepping through the error message to tell what it is and which institution is in play is kinda gross too 🙃

pages/authNotAllowed/authNotAllowed.html.js
```diff
@@ -0,0 +1,80 @@
...
+    case 'Shibboleth':
+      return '/pl/shibcallback';
+    default:
+      return null;
+  }
+}
+
+function AuthNotAllowed({ institutionAuthnProviders, resLocals }) {
+  console.log(institutionAuthnProviders);
+
```
nwalters512, line 21:
```suggestion
```

pages/authNotAllowed/authNotAllowed.html.js
```diff
@@ -0,0 +1,80 @@
...
+  // LTI providers are special in that we don't currently allow an LTI session
+  // to be initiated from PrairieLearn. We'll handle them specially.
+  const hasLtiProvider = institutionAuthnProviders.some((p) => p.name === 'LTI');
+  const nonLtiAuthnProviders = institutionAuthnProviders.filter((p) => p.name !== 'LTI');
+  const hasNonLtiProviders = nonLtiAuthnProviders.length > 0;
+
+  // Check if there's a default authentication provider.
+  const defaultAuthnProvider = nonLtiAuthnProviders.find((p) => p.is_default);
```
nwalters512, line 29:
```suggestion
  const defaultAuthnProvider = nonLtiAuthnProviders.find((p) => p.is_default);

```

pages/partials/navbar.ejs
```diff
@@ -18,9 +18,10 @@
   <button class="navbar-toggler" type="button" data-toggle="collapse" data-target=".navbar-collapse" aria-expanded="false" aria-label="Toggle navigation">
     <span class="navbar-toggler-icon"></span>
   </button>
+  <% if (typeof navbarType === 'undefined' || navbarType !== 'unauthenticated') { %>
   <div id="course-nav" class="collapse navbar-collapse">
     <ul class="nav navbar-nav mr-auto" id="main-nav">
-      <% if (typeof navbarType == 'undefined' || navbarType == 'plain') { %>
+      <% if (typeof navbarType == 'undefined' || navbarType == 'plain' || navbarType == 'unauthenticated') { %>
```
nwalters512, line 24:
```suggestion
      <% if (typeof navbarType == 'undefined' || navbarType == 'plain') { %>
```

ee/auth/saml/router.js
mwest1066: Yeah, this is indeed rather gross.

How about we return a `result` from `users_select_or_insert` which is either `"success"` or `"invalid_provider"` and also return the `institution_id` in all cases? (and still raise exceptions if something else weird goes wrong)

From what I can see we don't have all that many call sites for `users_select_or_insert` so it wouldn't be too bad to check a `result` at each of them?

That being said, I won't block on this if you prefer to parse the `institution_id` out of the error message.

pages/authNotAllowed/authNotAllowed.html.js
```diff
@@ -0,0 +1,79 @@
...
+            ${hasNonLtiProviders ? 'Please use an alternative authentication method.' : ''}
+            ${!hasNonLtiProviders && !hasLtiProvider
+              ? 'Please contact your institution for more information.'
+              : ''}
+          </p>
+          ${defaultAuthnProvider
+            ? html`
+                <div class="mb-3">
+                  <a href="${getUrlForProvider(defaultAuthnProvider)}" class="btn btn-primary">
+                    Sign in with institution single sign-on
```
mwest1066, line 53: Should we make this look identical to the button they should have clicked on originally, to train them which one is the right one? That is, if we wanted them to have originally click on "U of I" then this button should also say "U of I".

pages/authNotAllowed/authNotAllowed.html.js
```diff
@@ -0,0 +1,79 @@
...
+                  </a>
+                </div>
+              `
+            : // If there's no default, show all options (except LTI).
+              nonLtiAuthnProviders.map(
+                (provider) => html`
+                  <div class="mb-3">
+                    <a href="${getUrlForProvider(provider)}" class="btn btn-primary">
+                      Sign in with
+                      ${provider.name === 'SAML' ? 'institution single sign-on' : provider.name}
```
mwest1066, line 63: Should this also look exactly the same as the buttons on the original login screen? ("U of I" rather than "institution SSO")

mwest1066 (approved):

ee/auth/saml/router.js
nwalters512: SGTM! I'll make the change.

nwalters512: Done in 8cbdf8c. The code looks much cleaner now!

pages/authNotAllowed/authNotAllowed.html.js
nwalters512: Hmm, interesting point... taking a different approach, what if I just merged this error page into the login page via some query parameters, e.g. you'd get redirected to `prairielearn.org/pl/login?institution_id=1&unsupported_authn_provider=true` which then shows something like this (I'll obviously improve the copy/layout/etc.):

Screen Shot 2022-08-22 at 14 01 39

pages/authNotAllowed/authNotAllowed.html.js
nwalters512: I think the solution here is going to be the institution dropdown that basically filters the login buttons. I can do that in this PR if you'd like.

mwest1066: I'm actually slightly unsure of how this will look with a dropdown. What were you thinking?

Some ideas:
1. A two-part control with a dropdown and a "Login" button. The dropdown starts on "Choose institution" and the button starts out disabled. Selecting an institution from the dropdown enables the button.
2. An immediately activating dropdown, where selecting an entry immediately proceeds to login with that institution.
3. Not a dropdown at all, but instead a vertically scrollable div with buttons saying "Login with U of I", etc. Clicking a button immediately proceeds to login. The scrollable area should show about 5 buttons at a time.
4. Something with a text box to allow searching for the institution, possibly together with a dropdown and a Login button.

mwest1066: Whatever we do here, I think we should definitely try and not have buttons that say things like "institution single sign-on" -- I think people are much more like to click "sign on with U of I" or something else with their institutions name on it. A generic "Login" button next to something that says their institution name would also be fine, I think.

nwalters512: I had (1) in mind! Do you have any preference one way or another?

ee/auth/saml/router.js
mwest1066: Nice!

pages/authNotAllowed/authNotAllowed.html.js
nwalters512: Re: your other comment, I guess I'm struggling with how to handle the fact that an institution can have multiple allowable providers. Say an institution had both SAML and Google, and they set Google as the default: how do we label the SAML button? Say they don't set a default at all: which button should say "sign in with U of X"? My current design might be optimizing for cases like these that might be unlikely in practice, but I think we still need to handle them.

ee/auth/saml/router.js
mwest1066: Do we need to modify this call in any way? https://github.com/PrairieLearn/PrairieLearn/blob/7a5ab555f9ea094af519f7ec293839ca22773aaf/middlewares/authn.js#L105

I'm actually not sure we want to be calling `users_select_or_insert` here, rather than just a `users_select`. I don't think that `insert` should ever happen from this call site?

Maybe it's fine to call `users_select_or_insert ` here and ignore the invalid-institution case because it can't happen here?

pages/authNotAllowed/authNotAllowed.html.js
mwest1066: (1) sounds good to me!

mwest1066: (especially with a cookie to speed up re-logins)

mwest1066: Re: institutions setting google as the default. This all gets a bit messy, I agree. How about this:

For a plain login screen (no query param indicating an institution), we provide:

* All institutions get a "Login with U of X" institution button. The meaning of this button is:
    * If the institution has SAML setup, the institution button goes to SAML login
    * Otherwise, if the institution has a default provider, the institution button goes to the default provider
    * Otherwise, the institution button goes to the first of [Google, Azure] that's enabled
* We always display the "Login with Google/Azure" buttons.

For a login screen with a query param specifying an institution, we display the same buttons as above, but filtered to only display those allowed for the specific institution.

This system means that if you enable SAML but set google as your default, then the default is effectively ignored. I think this is ok because this seems like a weird edge case to me.

apps/prairielearn/src/pages/authLogin/authLogin.html.js
```diff
@@ -0,0 +1,265 @@
+// @ts-check
```
nwalters512, line 1: This page is probably quite outdated at this point. I'm going to convert it in another PR, and it should then be obvious what changes are *actually* made in this PR.

apps/prairielearn/src/ee/institution/admin/sso.html.ts
```diff
@@ -120,6 +125,7 @@ export const InstitutionAdminSso = ({
 
               // Sync disabled state of default provider option.
               var defaultProviderOption = $('#defaultProvider option[value="' + providerId + '"]');
+              console.log(defaultProviderOption);
```
nwalters512, line 128:
```suggestion
```

pages/authNotAllowed/authNotAllowed.html.js
nwalters512: Discussed with Matt. We're going to avoid any changes to the initial login page for now and focus only on improving the case where the user picks a provider that isn't supported by their institution.

pages/authNotAllowed/authNotAllowed.html.js
nwalters512: I ultimately made this change!

apps/prairielearn/src/sprocs/users_select_or_insert.sql
```diff
@@ -76,7 +80,8 @@ BEGIN
             AND ap.name = authn_provider_name;
 
         IF NOT FOUND THEN
-            RAISE EXCEPTION '"%" authentication provider is not allowed for institution "%"', authn_provider_name, institution.long_name;
+            result := 'invalid_authn_provider';
```
jonatanschroeder, line 83: I wonder if it may be better to use `RAISE EXCEPTION ... WITH ERRCODE` (or something similar, like `DETAIL` or `HINT`) similar to what is done with the status code, then catch the exception and get the info from `err.data.sqlError`. I think this  kind of construct is a bit weird...

apps/prairielearn/src/lib/authn.js
```diff
@@ -47,6 +47,11 @@ module.exports.loadUser = async (req, res, authnParams, optionsParams = {}) => {
     let userSelectOrInsertRes = await sqldb.callAsync('users_select_or_insert', params);
 
     user_id = userSelectOrInsertRes.rows[0].user_id;
+    const { result, user_institution_id } = userSelectOrInsertRes.rows[0];
+    if (result === 'invalid_authn_provider') {
```
jonatanschroeder, line 51: See comment in users_select_or_insert.sql about exceptions.

jonatanschroeder (approved):
Looks good, I won't be offended if you ignore my comments.

apps/prairielearn/src/sprocs/users_select_or_insert.sql
nwalters512: That is in fact how I originally implemented things, see https://github.com/PrairieLearn/PrairieLearn/pull/6151/commits/22a4c0ec6a26f30ee48eb9857891ec8e79ae9d26. I changed the implementation as discussed here: https://github.com/PrairieLearn/PrairieLearn/pull/6151#discussion_r950714886

One change that occurred since this PR was introduced is that we now only really have a single site where `users_select_or_insert` is called, which makes the "catch error and parse the institution ID out of it" approach a little more tenable. I'm inclined to not make a change here, though if @mwest1066 agrees with you, I'll reconsider.

jonatanschroeder: The difference between my suggestion and the code you originally had was that your code was based on the error message, while I'm suggesting using another field (e.g., code, detail, hint) to identify the error type. This should be more robust against changes in wording of the exception text. But it's up to you.

apps/prairielearn/src/ee/pages/institutionAdminSso/institutionAdminSso.html.ts
```diff
@@ -94,11 +95,15 @@ export const InstitutionAdminSso = ({
...
+                  const isEnabled = institutionAuthenticationProviders.some(
+                    (p) => p.id === provider.id,
+                  );
+
                   return html`
                     <option
                       value="${provider.id}"
                       ${provider.id === institution.default_authn_provider_id ? 'selected' : ''}
-                      ${provider.name === 'SAML' && !hasSamlProvider ? 'disabled' : ''}
+                      ${!isEnabled ? 'disabled' : ''}
```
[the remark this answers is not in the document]
nwalters512: Not directly related; it's just something I came across while testing this PR, which required testing different combinations of enabled/disabled/default providers. Previously, it was possible to set a provider that wasn't enabled as the default.

apps/prairielearn/src/sprocs/users_select_or_insert.sql
nwalters512: I don't see a meaningful difference between those and the exception message; we'd still have to parse an institution ID out of *something*. I suppose we could directly set `detail` to an institution ID, but that almost feels worse than an explicit `user_instititution_id` in the results.

jonatanschroeder: To be clear, I'm ok with institution ID as an output parameter, it's just the `result` column that I have a problem with.

nwalters512: Unless I'm missing something, a sproc cannot both raise an exception and return a value. We only actually need the institution ID when we're handling a failure case (that is, when you're proposing we raise an exception).

jonatanschroeder: Ah, right, didn't think of that.

PR #6155, opened 2022-08-23
Allow file URLs to be used in renderEjs and loadSqlEquiv

nwalters512:
This will help ease the transition to ESM, as we can just pass `import.meta.url` directly to these functions.

### DIFF
--- a/packages/html-ejs/src/index.ts
+++ b/packages/html-ejs/src/index.ts
@@ -24,5 +24,5 @@ export function renderEjs(filePathOrUrl: string, template: string, data: any = {
     resolvedPath = fileURLToPath(filePathOrUrl);
   }
 
-  return unsafeHtml(ejs.render(template, data, { views: [path.dirname(filePathOrUrl)] }));
+  return unsafeHtml(ejs.render(template, data, { views: [path.dirname(resolvedPath)] }));
 }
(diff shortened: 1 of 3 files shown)


packages/html-ejs/src/index.ts
```diff
@@ -9,11 +10,19 @@ import { unsafeHtml, type HtmlSafeString } from '@prairielearn/html';
...
+export function renderEjs(filePathOrUrl: string, template: string, data: any = {}): HtmlSafeString {
+  let resolvedPath = filePathOrUrl;
+
+  // This allows for us to pass `import.meta.url` to this function in ES Modules
+  // environments where `__filename` is not available.
+  if (filePathOrUrl.startsWith('file://')) {
+    resolvedPath = fileURLToPath(filePathOrUrl);
+  }
+
+  return unsafeHtml(ejs.render(template, data, { views: [path.dirname(filePathOrUrl)] }));
```
mwest1066, line 27: Should this be:
```suggestion
  return unsafeHtml(ejs.render(template, data, { views: [path.dirname(resolvedPath)] }));
```

mwest1066 (approved):

packages/html-ejs/src/index.ts
nwalters512: Yes it should 😂 I'm getting sloppy!

PR #6162, opened 2022-08-25
Limit SAML authentication to within an institution

nwalters512:
This PR adds code to improve how we handle SAML auth:

- When evaluating a response from a SAML IDP, ensure that the UID matches the `uid_regexp` for the institution associated with the SAML provider.
- Enforce UIN uniqueness only within a given institution, not globally.

I took the opportunity to refactor `tests/sprocs/users_select_or_insert.test.js` to make the tests a bit easier to read and maintain (async/await, objects instead of arrays, etc).

Closes https://github.com/PrairieLearnInc/sysconf/issues/257.


ee/auth/saml/router.js
```diff
@@ -83,7 +83,7 @@ router.post(
       throw new Error('Missing one or more SAML attributes');
     }
 
-    const params = [authUid, authName, authUin, 'SAML'];
+    const params = [authUid, authName, authUin, 'SAML', institutionId];
```
mwest1066, line 86: How secure is `req.params.institution_id`? Normally I would think it's untrusted user data. Can we verify it somehow.

I'm thinking of a case where a user is working with an evil IdP. The user gets the IdP to generate a signed auth message for `nathan@prairielearn.com` and then the user sends that to the route with `institution_id` for `prairieLearn.com`. PL blindly trusts the `institution_id` from the route and auths the user.

middlewares/authn.js
```diff
@@ -101,7 +101,7 @@ module.exports = function (req, res, next) {
       authUin = '100000000';
     }
 
-    let params = [authUid, authName, authUin, 'dev'];
+    let params = [authUid, authName, authUin, 'dev', null];
```
mwest1066, line 104: In plpgsql you can declare a function argument as [`DEFAULT NULL`](https://www.postgresql.org/docs/current/xfunc-sql.html#XFUNC-SQL-PARAMETER-DEFAULTS) so we wouldn't need to pass `null` here.

migrations/20220825210520_265_users__uin_unique_per_institution.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE users ADD CONSTRAINT users_uin_institution_id_key UNIQUE (uin, institution_id);
```
mwest1066, line 1: Could we use this syntax so we can include `IF NOT EXISTS` (for general robustness): https://github.com/PrairieLearn/PrairieLearn/blob/df2771dd8c10ff21dbe8ba1951468af4e74688f4/migrations/20200913144809_216_questions__course_id__uuid__create_index.sql#L1

sprocs/users_select_or_insert.sql
```diff
@@ -4,6 +4,7 @@ CREATE FUNCTION
         IN name text,
         IN uin text,
         IN authn_provider_name text,
+        In institution_id bigint,
```
mwest1066, line 7: Avoid needing to pass `null` everywhere. I'm also fine if you prefer to not have a default value so as to force to explicitly pass `null`.
```suggestion
        IN institution_id bigint DEFAULT NULL,
```

pages/authCallbackOAuth2/authCallbackOAuth2.js
```diff
@@ -52,6 +52,7 @@ router.get('/', function (req, res, next) {
       identity.name || identity.email, // name (use email if name is not present)
       identity.sub, // uin
```
mwest1066, line 53: I think we should not provide the UIN from Google anymore?

mwest1066 (approved):

pages/authCallbackOAuth2/authCallbackOAuth2.js
mwest1066: Or maybe we do want this, actually. But we should add a warning to the institution auth setup page to warn that enabling Google auth will overwrite UINs if you use it in conjunction with SAML? You basically should never turn on both Google and SAML for an institution, I think.

ee/auth/saml/router.js
nwalters512: I believe this is handled by signing: it would be impossible for the evil IdP to have signed a response that would correctly validate with the _real_ IdP's public certificate. Does that sound correct to you?

migrations/20220825210520_265_users__uin_unique_per_institution.sql
nwalters512: Yes!

pages/authCallbackOAuth2/authCallbackOAuth2.js
nwalters512: Yeah, I'm really not sure about this. My gut tells me that Azure and Google (OAuth) should basically behave the same, and since Azure doesn't include a UIN, neither should Google? I'd be happy to just remove this.

sprocs/users_select_or_insert.sql
nwalters512: Ah nice - I'll switch to using a default.

migrations/20220825210520_265_users__uin_unique_per_institution.sql
nwalters512: Done in d946fa0.

middlewares/authn.js
nwalters512: Changed in 0b0624a.

sprocs/users_select_or_insert.sql
nwalters512: Changed in 0b0624a.

pages/authCallbackOAuth2/authCallbackOAuth2.js
mwest1066: I guess I'm thinking that if an institution is using Google as their primary auth source then we _do_ want to use the UIN from it, to support changes in the UID. Ideally we'd also get an equivalent stable identifier from Azure. At the moment, if an institution uses Azure as it's auth source, we can't support UID renames.

nwalters512: Ah, that's fair. In the interest of shipping these changes ASAP, I won't make that change here, but I did open a new issue to track that: #6164. Does that sound good to you?

ee/auth/saml/router.js
mwest1066: But how do we know that the `institution_id`-from-route matches the `institution_id` in the signed response? When we call `passport.authenticate()` we don't pass it the `institution_id` from the route, we just process the SAML response. And then later we do `const institutionId = req.params.institution_id` but we don't check that against anything. Am I missing a checking step somewhere?

nwalters512: I think you are (at least, I hope I'm not). We use `MultiSamlStrategy`, and the options for that include `getSamlOptions`, which will retrieve the details (including certs) based on `req.params.institution_id`:

https://github.com/PrairieLearn/PrairieLearn/blob/6bb8c1d183977ad61139a1e8cb0c88adfedd6af0/ee/auth/saml/index.js#L6-L10

pages/authCallbackOAuth2/authCallbackOAuth2.js
mwest1066: Perfect

ee/auth/saml/router.js
mwest1066: Ok, great, I buy that. Thanks for explaining!

PR #6168, opened 2022-08-26
Add exampleCourse photoRosterNames question

trombonekenny:
Adds a flashcard-like question to learn student names based on roster data from Canvas. A python helper script is included to help download the roster data from Canvas.

I wasn't sure which folder in exampleCourse/questions to tuck this kind of interesting, course-staff focused question but it seemed like the fun things we'd do in a workshop.

I'm also not sure if I should add it to an assessment or make a new one?

What's the right way to share this question with others?


trombonekenny:
Stale, I'll update with a newer copy when I'm ready to merge (and have figured out where to drop it).

Issue #3800, opened 2021-02-05
Changing Homework -> Exam breaks existing instance questions


mwest1066:
As described in #3797, starting an assessment as `"type": "Homework"` and switching to `"type": "Exam"` once students are working will result in the students being unable to access their questions, because they won't have `instance_questions.points_list` set (among other problems). We should check which `instance_questions` columns are being used by the points calculation for Homeworks and Exams and make sure that things won't crash if questions are switched back and forwards. This will be needed eventually for #3347 anyway, and will improve robustness in the meantime. It's not necessary for this compatibility to do anything sensible --- it just shouldn't crash.

Issue #6169, opened 2022-08-26
Crash due to null `points_list`


mwest1066:
```
lib/question.js:1486
locals.variantAttemptsLeft = instance_question.points_list.length;
^
TypeError: Cannot read properties of null (reading 'length')
at Object._buildLocals (lib/question.js:1486:70)
at async.series.params.variant_id (lib/question.js:1623:44)
at node_modules/async/dist/async.js:3674:28
at replenish (node_modules/async/dist/async.js:446:21)
at iterateeCallback (node_modules/async/dist/async.js:430:21)
at node_modules/async/dist/async.js:327:20
at node_modules/async/dist/async.js:3679:17
at lib/question.js:1594:17
at PostgresPool.<anonymous> (lib/question.js:307:18)
at processTicksAndRejections (node:internal/process/task_queues:83:21)
```

jonatanschroeder:
Did #6171 resolve this? Should it be closed?

nwalters512:
Yes! GitHub just didn't autoclose because I misformatted my comment.

PR #6171, opened 2022-08-26
Handle case where buildLocals throws an error

nwalters512:
This puts a bandaid over the issue described in #6169. While it doesn't prevent an error, it does prevent that error from crashing the entire server.

### DIFF
--- a/lib/question.js
+++ b/lib/question.js
@@ -1609,7 +1609,12 @@ module.exports = {
           _.assign(locals, urls);
           callback(null);
         },
-        (callback) => {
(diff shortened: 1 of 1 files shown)


mwest1066:
Also see #3347

Maybe this PR should also close #3800?

mwest1066 (approved):

nwalters512:
@mwest1066 good call about #3800 - I'll make this PR close that out! This (should) fulfill the desired behavior of "just don't crash".

PR #6172, opened 2022-08-26
Scaffold automated accessibility testing; fix issues

### LINKED ISSUE #1508: Assessment questions stats page has two tables with same HTML ID
The assessment overview question stats table view page (whew!) has two different tables, "statistics" vs. "detailed statistics," that both have the same ID, `id="questionsTable"`. Unless I'm mistaken, this is a no-no. I wonder if it's related to the export issue raised here: #1507

nwalters512:
This PR adds automated accessibility testing based on [`axe-core`](https://github.com/dequelabs/axe-core). While automated testing *will not* catch all issues, particularly those surrounding interactions, it serves as a useful starting point for identifying high-volume, low-hanging-fruit problems.

To get this PR passing the new tests, I naturally had to fix our pages! There are a few broad categories of changes I made:

- Ensuring every page has a `<main>` landmark.
- Adding titles for all table columns. See inline discussions about this.
- Mixing markup to ensure correct nesting of tags after parsing.
- Adding or correcting `aria-*` attributes.

Fixes #1508.


mwest1066:
I love this concept! I'm very interested to see what kind of issues `axe` can identify.

tests/accessibility/index.test.js
```diff
@@ -0,0 +1,163 @@
...
+    courseInstanceId,
+  };
+}
+
+// These are meant to be a representative sample of pages from across the site.
+// We hardcode database IDs to 1, as such entities will always exist. We don't
+// case exactly which entities are used, just that we get a representative page
+// of each "kind".
+const pages = [
+  '/',
```
SethPoulsen, line 92: It would be cool if we could automatically generate this list from `server.js` somehow so that it wasn't an extra thing to remember to update when I'm don't know of any clean way to do that though :sweat_smile:.

Maybe easier to implement: should we also make sure we have an example of every element in here too? We could grab the list from `tests/exampleCourseQuestions.test.js` and/or `tests/testCourseQuestions.test.js`

nwalters512: I was thinking about the former a bit, but ultimately I think it may be more trouble than it's worth.

It'd be great to have some elements represented in these tests too!

SethPoulsen: I wonder how far a simple Regex could take us...something like just doing `/'pl/course_instance.*'/` on all of `server.js`. 
Of course as you say, it could be more trouble than its worth.

nwalters512: You're welcome to try to figure something out, but I think our time would be better spent remediating accessibility issues than trying to find clever ways to ensure perfect coverage.

pages/administratorAdmins/administratorAdmins.html.js
```diff
@@ -54,7 +54,7 @@ function AdministratorAdmins({ admins, resLocals }) {
                   <tr>
                     <th>UID</th>
                     <th>Name</th>
-                    <th></th>
+                    <th>Actions</th>
```
nwalters512, line 57: I could decide exactly what to do in all these cases. In some headings (like here), I just added a generic "Actions" title; in others, I still added that title but made it only visible to screen readers. We should pick one and be consistent; I lean towards always making these titles visible.

In the long-term, it'd be better on many pages to not use a table at all. That's outside the scope of this PR though.

server.js
```diff
@@ -1832,7 +1833,7 @@ module.exports.insertDevUser = function (callback) {
   });
 };
 
-if (config.startServer) {
+if (config.startServer && require.main === module) {
```
nwalters512, line 1836: I made this change so I could safely require this file without mucking around with the config.

tests/accessibility/index.test.js
```diff
@@ -0,0 +1,361 @@
...
+const config = require('../../lib/config');
+const helperServer = require('../helperServer');
+
+const SITE_URL = 'http://localhost:' + config.serverPort;
+const EXAMPLE_COURSE_DIR = path.resolve(__dirname, '..', '..', 'exampleCourse');
+
+/**
+ * Several pages have very large DOMs or attributes that AXE runs very slow on.
+ * None of these elements have accessibility components, so we'll special-case these
+ * and remove them before running AXE.
```
nwalters512, line 29: I don't love these special cases, but they were the only way I could find to avoid adding another minute to our test execution time.

tests/accessibility/index.test.js
```diff
@@ -0,0 +1,361 @@
...
+
+  // TODO: create an assessment instance and create an instance question so we can test these pages.
+  '/pl/course_instance/:course_instance_id/assessment_instance/:assessment_instance_id',
+  '/pl/course_instance/:course_instance_id/instance_question/:instance_question_id',
+  '/pl/course_instance/:course_instance_id/instructor/assessment_instance/:assessment_instance_id',
+  '/pl/course_instance/:course_instance_id/instructor/assessment/:assessment_id/manual_grading/assessment_question/:assessment_question_id',
+  '/pl/course_instance/:course_instance_id/instructor/assessment/:assessment_id/manual_grading/instance_question/:instance_question_id',
+
+  // TODO: submit an answer to a question so we can test this page.
+  '/pl/course_instance/:course_instance_id/instructor/grading_job/:job_id',
```
nwalters512, line 242: I'm currently inclined to leave these for the future, but will address them in this PR if folks feel strongly.

tests/accessibility/index.test.js
nwalters512: It turns out this wasn't actually *too* bad to get working! See 252b52c and subsequent commits.

SethPoulsen: Very cool!

pages/administratorAdmins/administratorAdmins.html.js
SethPoulsen: I would vote for "Actions" title to be visible everywhere, I can't see any harm in it.

I agree that changing from table layouts is beyond the scope of this PR.

pages/authPassword/authPassword.js
```diff
@@ -11,14 +11,12 @@ router.get('/', function (req, res) {
...
 
-    var pwCookie = csrf.generateToken({ password: req.body.password, maxAge }, config.secretKey);
-    res.cookie('pl_assessmentpw', pwCookie, { maxAge, httpOnly: true, secure: true });
-    res.clearCookie('pl_pw_origUrl');
-    return res.redirect(pl_pw_origUrl);
-  }
+  var pwCookie = csrf.generateToken({ password: req.body.password, maxAge }, config.secretKey);
+  res.cookie('pl_assessmentpw', pwCookie, { maxAge, httpOnly: true, secure: true });
+  res.clearCookie('pl_pw_origUrl');
+  return res.redirect(redirectUrl);
```
SethPoulsen, line 20: It seems like this would be a really tragic thing to accidentally break, I assume we have some automated tests covering this. (this is one of the only times I've thought 'it would be nice to have a code coverage report' since we stopped generating them a year ago or whenever it was)

pages/partials/navbar.ejs
```diff
@@ -518,8 +518,10 @@ const navPagesTabs = {
...
 if (typeof navPage != 'undefined' && navPagesTabs.hasOwnProperty(navPage)) { %>
     <!-- Instructor navigation tabs -->
-    <ul class="nav nav-tabs pl-nav-tabs-bar pt-2 px-3 bg-light">
-        <%- include(navPagesTabs[navPage]); %>
-    </ul>
+    <nav>
+        <ul class="nav nav-tabs pl-nav-tabs-bar pt-2 px-3 bg-light">
+            <%- include(navPagesTabs[navPage]); %>
+        </ul>
+    </nav>
```
SethPoulsen, line 525: this is cool, I didn't even know `<nav>` existed.

tests/accessibility/index.test.js
SethPoulsen: Ya, this is annoying. It would be nice to deep dive on why it takes so long and come up with a better solution but I think its fair to push that off until later.

tests/accessibility/index.test.js
SethPoulsen: I'm fine with leaving them until later

tests/accessibility/index.test.js
```diff
@@ -0,0 +1,357 @@
...
+];
+
+// const ONLY_ROUTES = ['/pl/course_instance/:course_instance_id/effectiveUser'];
+const ONLY_ROUTES = [];
```
SethPoulsen, line 246: Is this commented out code supposed to be here?

If so maybe leave a comment about why.

SethPoulsen (approved):

pages/authPassword/authPassword.js
nwalters512: Is there a specific concern you have with this diff?

tests/accessibility/index.test.js
nwalters512: Good call, removed!

pages/administratorAdmins/administratorAdmins.html.js
mwest1066: I'm on board with making it visible, at least until we can come up with something better than the table layout.

PR #6174, opened 2022-08-27
Update R autograder to Ubuntu 22.04, R 4.2.1 and current packages

eddelbuettel:
This also switches the CRAN package injection from using the c2d4u repo with its ~5k packages to the newer r2u with full coverage, which reduces on RUN step as we no longer need any 'additional' CRAN packages installed from source.

Otherwise standard updated which @barbehenna tested locally based on a build we did in the predecessor repo.  The PR is otherwise very narrow and touches nothing outside the r-autograder directory.


mwest1066 (approved):

mwest1066:
Looks great, thanks!

PR #6178, opened 2022-08-30
Mark workspaces as stopped when their host terminates

nwalters512:
If a workspace host unexpectedly dies or is terminated, we need to mark any workspaces that were allocated to run on it as stopped.


mwest1066 (approved):

nicknytko (approved):
lgtm

PR #6181, opened 2022-08-30
Don't proxy workspace traffic if workspace is not running

nwalters512:
This should clean up our logs a bit.

### DIFF
--- a/server.js
+++ b/server.js
@@ -262,12 +262,16 @@ module.exports.initExpress = function () {
         const match = req.url.match(/^\/pl\/workspace\/([0-9]+)\/container\//);
         if (!match) throw new Error(`Could not match URL: ${req.url}`);
         const workspace_id = match[1];
-        const result = await sqldb.queryOneRowAsync(
-          `SELECT hostname FROM workspaces WHERE id = $workspace_id;`,
+        const result = await sqldb.queryZeroOrOneRowAsync(
+          `SELECT hostname FROM workspaces WHERE id = $workspace_id AND state = 'running';`,
           { workspace_id }
         );
-        const url = `http://${result.rows[0].hostname}/`;
-        return url;
+
+        if (result.rows.length === 0) {
+          throw new Error(`Workspace ${workspace_id} is not running`);
+        }
+
+        return `http://${result.rows[0].hostname}/`;
       } catch (err) {
         logger.error(`Error in router for url=${req.url}: ${err}`);
         return 'not-matched';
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -316,10 +309,12 @@ module.exports.initExpress = function () {
   app.use('/pl/workspace/:workspace_id/container', [
     cookieParser(),
     (req, res, next) => {
+      // Needed for workspaceAuthRouter and `selectAndValidateWorkspace` middleware.
       res.locals.workspace_id = req.params.workspace_id;
       next();
-    }, // needed for workspaceAuthRouter
+    },
     workspaceAuthRouter,
+    require('./middlewares/selectAndValidateWorkspace'),
```
nwalters512, line 317: Instead of checking the workspace host/state inside the proxy, I instead moved it to separate middleware that runs before the proxy. This gives us greater control over the error code in case we encounter a request bound for a workspace that isn't running.

server.js
```diff
@@ -258,20 +258,12 @@ module.exports.initExpress = function () {
...
-        const result = await sqldb.queryOneRowAsync(
-          `SELECT hostname FROM workspaces WHERE id = $workspace_id;`,
-          { workspace_id }
-        );
-        const url = `http://${result.rows[0].hostname}/`;
-        return url;
-      } catch (err) {
-        logger.error(`Error in router for url=${req.url}: ${err}`);
-        return 'not-matched';
+      if (!req.workspace_hostname) {
```
nwalters512, line 261: I removed the `try`/`catch` around all this that returned `not-matched` if an error was encountered. AFAICT, that wasn't actually treated as a special value by any of the `http-proxy` code; I'd rather just throw an error than try to route to a nonsensical hostname.

server.js
```diff
@@ -258,20 +258,12 @@ module.exports.initExpress = function () {
...
+        throw new Error('Workspace host not found');
       }
+
+      // This value is set in the `selectAndValidateWorkspace` middleware.
+      return `http://${req.workspace_hostname}/`;
```
nwalters512, line 266: I know we don't like these magic values attached to `req` or `res`, but I felt it was appropriate in this case to avoid duplicate queries: this code will run very frequently and is pretty performance-sensitive.

server.js
nicknytko: 🙏 i agree this seems like a much more sensible thing to do

server.js
nicknytko: what’s the `originalUrl` key used for here?

nicknytko: ah, gotcha! makes sense that `req.url` might get mangled

mwest1066 (approved):

server.js
mwest1066: Yep, special cases like this seem fine to me.

PR #6182, opened 2022-08-30
Set up Sentry integration

nwalters512:
This PR sets up the Sentry SDK if `sentryDsn` is defined in the config file. This will give us greater visibility into crashes and errors that are happening in the wild.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,13 @@ on:
   pull_request:
 
 concurrency:
-  # Ensure that we only run one concurrent job for Pull Requests.
+  # Ensure that we only run one concurrent job for Pull Requests. This ensures
+  # that someone can't kill our throughput by pushing a bunch of commits to a
+  # single branch in rapid succession.
+  #
+  # However, for master builds, we allow maximum concurrency. This is achieved
+  # because `github.head_ref` isn't defined there, and `github.run_id` is
+  # globally unique in this repo.
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,6 +6,11 @@ on:
       - master
   pull_request:
 
+concurrency:
+  # Ensure that we only run one concurrent job for Pull Requests.
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/middlewares/logResponse.js
+++ b/middlewares/logResponse.js
@@ -10,7 +10,6 @@ module.exports = function (req, res, next) {
       // debug errors. No PII is added.
       if (config.sentryDsn) {
         Sentry.configureScope((scope) => {
-          console.log(scope);
           if (res.locals?.course?.id) {
             scope.setContext('course', {
               id: res.locals.course?.id,
--- a/middlewares/logResponse.js
+++ b/middlewares/logResponse.js
@@ -1,8 +1,58 @@
+const Sentry = require('@prairielearn/sentry');
+
+const config = require('../lib/config');
 const logger = require('../lib/logger');
 
 module.exports = function (req, res, next) {
   if (req.method !== 'OPTIONS') {
-    res.on('finish', function () {
+    res.once('finish', function () {
+      // Attach information to the Sentry scope so that we can more easily
+      // debug errors. No PII is added.
+      if (config.sentryDsn) {
+        Sentry.configureScope((scope) => {
+          console.log(scope);
+          if (res.locals?.course?.id) {
+            scope.setContext('course', {
+              id: res.locals.course?.id,
+              name: res.locals.course?.name,
+            });
+          }
+
+          if (res.locals?.course_instance?.id) {
+            scope.setContext('course_instance', {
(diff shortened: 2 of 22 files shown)


server.js
```diff
@@ -429,6 +436,18 @@ module.exports.initExpress = function () {
   app.use(require('./middlewares/authn')); // authentication, set res.locals.authn_user
   app.use('/pl/api', require('./middlewares/authnToken')); // authn for the API, set res.locals.authn_user
 
+  if (config.sentryDsn) {
+    // Attach user ID to Sentry events.
+    app.use((req, res, next) => {
+      if (res.locals.authn_user) {
+        Sentry.configureScope((scope) => {
+          scope.setUser({ id: res.locals.authn_user.user_id.toString() });
```
nwalters512, line 444: I configured this to test out how Sentry works with user information, but I'm not sure if we actually want to land this. Under GDPR, I don't believe internal user IDs constitute PII, but we may want to avoid sending this to be on the safe side?

server.js
```diff
@@ -1723,6 +1745,23 @@ if (config.startServer) {
           serviceName: 'prairielearn',
         });
 
+        // Same with Sentry configuration.
+        if (config.sentryDsn) {
+          // Use the current Git has to identify the version of the code.
```
nwalters512, line 1750:
```suggestion
          // Use the current Git hash to identify the release.
```

server.js
```diff
@@ -1946,8 +1985,7 @@ if (config.startServer) {
     function (err, data) {
       if (err) {
         logger.error('Error initializing PrairieLearn server:', err, data);
-        logger.error('Exiting...');
-        process.exit(1);
+        throw err;
```
nwalters512, line 1988: This allows Sentry to correctly pick up crashes during startup.

server.js
mwest1066: I think this is fine and it's very helpful debugging context.

Can we add other things here, like `res.locals.response_id` and many of the other fields in [`logRequest`](https://github.com/PrairieLearn/PrairieLearn/blob/master/middlewares/logRequest.js) and [`logResponse`](https://github.com/PrairieLearn/PrairieLearn/blob/master/middlewares/logResponse.js)?

mwest1066 (approved):

nwalters512:
#6186 should fix the failing typechecking during CI.

server.js
```diff
@@ -82,6 +83,11 @@ module.exports.initExpress = function () {
   app.set('trust proxy', config.trustProxy);
   config.devMode = app.get('env') === 'development';
 
+  // If we're set up with Sentry, use its middleware to record errors.
```
nwalters512, line 86:
```suggestion
  // If we're set up with Sentry, use its middleware to record requests.
```

middlewares/logResponse.js
```diff
@@ -1,8 +1,58 @@
...
 
 module.exports = function (req, res, next) {
   if (req.method !== 'OPTIONS') {
-    res.on('finish', function () {
+    res.once('finish', function () {
+      // Attach information to the Sentry scope so that we can more easily
+      // debug errors. No PII is added.
+      if (config.sentryDsn) {
+        Sentry.configureScope((scope) => {
+          console.log(scope);
```
nwalters512, line 13:
```suggestion
```

server.js
```diff
@@ -1607,6 +1632,9 @@ module.exports.initExpress = function () {
   app.use(require('./middlewares/notFound'));
 
   app.use(require('./middlewares/redirectEffectiveAccessDenied'));
+
+  // The Sentry error handler must come first.
+  app.use(Sentry.Handlers.errorHandler());
```
nwalters512, line 1637:
```suggestion
  if (config.sentryDsn) {
    app.use(Sentry.Handlers.errorHandler());
  }
```

server.js
nwalters512: Yes! I added some more context in 5fb320c8.

.github/workflows/main.yml
```diff
@@ -6,6 +6,11 @@ on:
       - master
   pull_request:
 
+concurrency:
+  # Ensure that we only run one concurrent job for Pull Requests.
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
```
nwalters512, line 12: I moved this into #6187.

middlewares/logResponse.js
```diff
@@ -1,8 +1,57 @@
...
+            });
+          }
+
+          if (res.locals?.instance_question?.id) {
+            scope.setContext('instance_question', {
+              id: res.locals.instance_question?.id,
+            });
+          }
+        });
+      }
```
mwest1066, line 53: Could we also add `res.locals.response_id`, `req.method`, `req.path`, and things like that?

What about `user_id`, or are we considering that to be PII?

mwest1066 (approved):

middlewares/logResponse.js
nwalters512: Ah yes, I forgot the one thing you asked about 😂 I'll add `response_id`.

The method and path are already captured automagically by Sentry:

Screen Shot 2022-08-31 at 16 22 27

nwalters512: `user_id` is captured elsewhere, although I should move it to this file for consistency.

mwest1066: > Ah yes, I forgot the one thing you asked about 😂 I'll add `response_id`.

🙏 

> The method and path are already captured automagically by Sentry:

Nice!

PR #6257, opened 2022-09-02
fix: Proper error message if external grader has invalid score

### LINKED ISSUE #6255: External grading does not properly complete if results.json fields are invalid
As reported on Slack: if there are errors while interpreting the results.json file of an external grader (e.g., score is not a number, or not between 0 and 1), then the process will not trigger the grading job to be marked as complete, which means the student's submission will remain as "Grading in progress" without a resolution. Preliminary debugging seems to point to lib/assessment.js, function `processGradingResult`, where these checks are performed. In case of an error, this function logs the error and updates the socket, but doesn't update the grading job, submission, etc.

https://github.com/PrairieLearn/PrairieLearn/blob/bb8cae5491a96d05136be245b7925b67a8464632/lib/assessment.js#L457-L463

The best approach here might need some discussion, but I believe the best approach would be to replace the error handling above with something that updates `gradable` and `succeeded` to false, `feedback` with an appropriate message, and `score` with 0; then proceeds to add the grade as usual instead of triggering an error.

jonatanschroeder:
Resolves #6255. The format of the message is somewhat analogous to what would be printed if results.json did not exist or could not be parsed as JSON.

### DIFF
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -436,18 +436,26 @@ module.exports = {
       [
         (callback) => {
           if (!_(content.grading).isObject()) {
-            return callback(error.makeWithData('invalid grading', { content: content }));
+            content.grading = {
+              feedback: {
+                succeeded: false,
+                gradable: false,
+                message: 'Error parsing external grading results: result data is not an object.',
+                original_content: content,
+              },
+            };
           }
           if (_(content.grading).has('feedback') && !_(content.grading.feedback).isObject()) {
-            return callback(
-              error.makeWithData('invalid grading.feedback', {
-                content: content,
-              })
-            );
+            content.grading.feedback = {
+              succeeded: false,
+              gradable: false,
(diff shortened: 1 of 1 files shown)


lib/assessment.js
```diff
@@ -447,26 +447,39 @@ module.exports = {
...
+            if (typeof content.grading.score === 'undefined') {
+              content.grading.feedback = {
+                results: { succeeded: false, gradable: false },
+                message: 'Error parsing external grading results: score was not provided.',
+                original_feedback: content.grading.feedback,
+              };
+              content.grading.score = 0;
+              gradable = false;
+            }
             if (!_(content.grading.score).isNumber()) {
```
mwest1066, line 466: Also exclude +/- infinity and NaN:
```suggestion
            if (!_(content.grading.score).isFinite()) {
```

mwest1066 (approved):

PR #6260, opened 2022-09-02
Fix workspace proxying

nwalters512:
After testing https://github.com/PrairieLearn/PrairieLearn/pull/6181 in our staging environment, I realized that it had broken proxying for workspaces. There were two issues in play:

- `express-async-handler` apparently can't be used for Express middleware, so requests were just hanging forever.
- Websocket requests don't flow through Express middleware, so `req.workspace_host` wasn't set.

This PR fixes both issues.


mwest1066:
> * `express-async-handler` apparently can't be used for Express middleware, so requests were just hanging forever.

Oops, sorry, I should have realized this in the original PR. You actually can do it, but it needs to be like this with an explicit call to `next()`: https://github.com/PrairieLearnInc/PrairieTest/blob/984ce2588990dbaafc66c02a15f8694683444a36/src/middlewares/authzAdmin.ts#L15

mwest1066:
> * Websocket requests don't flow through Express middleware, so `req.workspace_host` wasn't set.

I didn't realize this! I still find websockets more mysterious than I should.

mwest1066 (approved):

nwalters512:
> Oops, sorry, I should have realized this in the original PR. You actually can do it, but it needs to be like this with an explicit call to `next()`: https://github.com/PrairieLearnInc/PrairieTest/blob/984ce2588990dbaafc66c02a15f8694683444a36/src/middlewares/authzAdmin.ts#L15

Ugh, that's so gross! I'm once again looking longingly at https://koajs.com/.

echuber2:
I don't have access to the PrairieTest repo, but I always put async Express routes like this:

```javascript
router.get('*', async (req, router_res, next) => {
  try {
    // ...
  } catch(err) {
    next(err);
  }
});
```

I read that in Express 5 this is no longer necessary though. Wouldn't be against learning a new framework if it's better.

PR #6265, opened 2022-09-04
Add shortcut to re-open assessment instance

### LINKED ISSUE #4811: Better UI for reopening assessment instance
It is possible to reopen an assessment instance by editing the remaining time column, but that is not always clear. We probably need a better hint/method to make this clearer, possibly change the button's text/icon if the assessment is closed?

### LINKED ISSUE #5663: Re-opening exams is not discoverable
If a student's exam is unexpectedly terminated before the session finishes (e.g., like the fire alarm yesterday during the exam) then the CBTF staff will ask instructors to "delete or re-open the student exams" so that students can either start the exam again or continue from where they left off.

"Deleting" an exam is discoverable, but "Re-open" no longer is. "Re-open" used to be an option on the "Action" menu, but now it's no longer there and instead it's accessed as "Remove time limit" under several layers of buttons.

In practice we've found this to be undiscoverable by instructors.

One solution might be to just restore the "Re-open" menu item on the "Action" menu?

@jonatanschroeder I think you reworked some of these dialogs?

jonatanschroeder:
Resolves #4811, resolves #5663.

### DIFF
--- a/pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
+++ b/pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
@@ -236,6 +236,7 @@ function timeRemainingLimitFormatter(value, row) {
     var link = $('<a>')
         .addClass('btn btn-secondary btn-xs ml-1 time-limit-edit-button')
         .attr('role', 'button')
+        .attr('id', `row${row.assessment_instance_id}PopoverTimeLimit`)
         .attr('tabindex', 0)
         .attr('data-row', JSON.stringify(row))
         .append($('<i class="bi-pencil-square" aria-hidden="true">'))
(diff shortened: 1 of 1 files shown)


nwalters512 (approved):

pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
```diff
@@ -343,6 +344,16 @@ function actionButtonFormatter(_value, row, index) {
         .append('Grade &amp; Close')
         .appendTo(menu);
 
+    $('<a>')
```
nwalters512, line 347: I know you're just matching existing code, but we should really just use a `<button>` here instead of jumping through all the hoops of `<a role="button" tabindex=0">`.

PR #6266, opened 2022-09-05
feat: Script to convert Canvas quizzes to PrairieLearn assessments

### LINKED ISSUE #5903: Canvas to PrairieLearn converter
@jonatanschroeder  has created this canvas to PrairieLearn converted:

 https://github.com/ubc-cpsc/canvasgrading

We should incorporate documentation and add this to `tools/question_converters`.

jonatanschroeder:
Resolves #5903.

@mfsilva22 if you have anyone that is able to test this more extensively that'd be great. I basically did some improvements to the script I used when at UBC, but York doesn't use Canvas, so my tests were somewhat limited.


mwest1066 (approved):
Nice!

PR #6270, opened 2022-09-05
feat: Use ace modelist to determine mode for file editor

### LINKED ISSUE #1985: File editor shows greyed out edit button for some file types
As seen in the screenshot below, the `.R` files have a greyed out `Edit` button.  The files can be edited by first clicking the name of the file and then clicking the `Edit` button from within that window, but you cannot enter the editing screen from the overview page in the screenshot.

Screen Shot 2020-01-17 at 1 04 49 PM

jonatanschroeder:
Resolves #1985. In essence this allows any type supported by the Ace editor to be edited, instead of a custom whitelist.


pages/instructorFileBrowser/instructorFileBrowser.ejs
```diff
@@ -12,9 +14,16 @@
...
     <script>
       $(function() {
-        $('[data-toggle="popover"]').popover({
-            sanitize: false
-        })
+          $('[data-toggle="popover"]').popover({
+              sanitize: false
+          })
+
+          const modelist = ace.require('ace/ext/modelist');
```
nwalters512, line 21: I'd much prefer to do this all on the server instead of on the client. You should be able to install [`ace-code`](https://www.npmjs.com/package/ace-code) and then do the following from within server-side code:

```js
const modelist = require('ace-code/src/ext/modelist');
```

nwalters512 (commented):
I'd like to explore moving this logic to the server before approving.

pages/instructorFileBrowser/instructorFileBrowser.ejs
jonatanschroeder: I moved the Edit disable logic to the server side (thanks for pointing to ace-code, I was struggling to get ace-builds to work server-side). For the editor itself and the choice of mode, given the JS for Ace is already used in that side, I believe it makes more sense to keep the logic that chooses the mode (rather than checks if it's editable) client side, but I'm open to discussion.

pages/instructorFileBrowser/instructorFileBrowser.js
```diff
@@ -213,7 +194,7 @@ function browseDirectory(file_browser, callback) {
             fs.lstat(filepath, (err, stats) => {
               if (ERR(err, callback)) return;
               if (stats.isFile()) {
-                const editable = canEditFile(filepath);
+                const editable = modelist.getModeForPath(filename)?.extRe?.test(filename);
```
nwalters512, line 197: I was just looking at Matt's comment over on https://github.com/PrairieLearn/PrairieLearn/issues/1985 - is there any reason not to use `isBinary` here? IMO we should allow folks to edit any text files, even if we don't have syntax highlighting for them.

pages/instructorFileEditor/instructorFileEditorClient.js
```diff
@@ -7,6 +7,9 @@ window.InstructorFileEditor = function (options) {
     throw new Error(`Instructor file editor element ${options.elementId} was not found!`);
   }
 
+  const modelist = ace.require('ace/ext/modelist');
+  const aceMode = modelist.getModeForPath(options.fileName).mode;
```
nwalters512, line 11: I saw you mention this in your other PR. I think it's pretty low-effort to do this on the server, so I'd rather keep that there to avoid the network + JS overhead associated with doing this on the client.

pages/instructorFileBrowser/instructorFileBrowser.js
jonatanschroeder: What comes to mind is that `isBinary` requires this operation to check the content of every file in the directory. I don't mind doing that in line 323, where we read the contents anyway, but in here it would require additional I/O. Do you think that's worth it?

jonatanschroeder: Changed to `isText` for file browsing, let me know your thoughts on doing this for directory browsing as well.

nwalters512: Ah, great point about reading every file... I suppose one option is to do what GitHub does and only show the action buttons (Edit, etc.) on the "view individual file" page? Having the "Edit" button behave differently depending on which page it's on is kind of strange to me.

jonatanschroeder: I don't like requiring two clicks instead of one for operations that may be relatively common for instructors using the Web interface. I get your concern about consistency, not sure what the best approach is in that case.

nwalters512: @jonatanschroeder check out the change for `isBinaryFile` in https://github.com/PrairieLearn/PrairieLearn/pull/6370/files: instead of reading all the contents into memory, you can give the function a path and it will read just the first few thousand bytes. It's definitely still *more* I/O than none at all, but it should still be plenty fast for an average directory with no more than a few dozen files.

jonatanschroeder: Testing this change I found this: https://github.com/PrairieLearn/PrairieLearn/pull/6370/files#r981661573. I can fix it in this PR, just let me know your thoughts.

pages/instructorFileBrowser/instructorFileBrowser.js
```diff
@@ -207,54 +187,50 @@ function browseDirectory(file_browser, callback) {
         file_browser.dirs = [];
         async.eachOfSeries(
```
jonatanschroeder, line 188: @nwalters512 given we have some I/O in all of the cases I wonder if it makes sense to replace this `eachOfSeries` with a `Promise.all` for a `map`, so that they can run in parallel. Thoughts?

nwalters512: I'm a little wary of `Promise.all` with an unbounded number of promises. If you want to add concurrency, I'd go with something like [`async.mapLimit`](https://caolan.github.io/async/v3/docs.html#mapLimit) with a low-ish concurrency limit (3?).

pages/instructorFileBrowser/instructorFileBrowser.js
```diff
@@ -202,89 +181,74 @@ function browseDirectory(file_browser, callback) {
...
+                  contains(invalidRootPath, filepath)
+                ),
+              };
+            } else {
+              return null;
+            }
           }
         );
+        file_browser.files = all_files.filter((f) => f?.isFile).sort((f) => f.name);
+        file_browser.dirs = all_files.filter((f) => f?.isDirectory).sort((f) => f.name);
```
nwalters512, line 238: `Array.prototype.sort` takes a comparison function, not something that maps to a property to be sorted on. I think this is what you want (untested):

```suggestion
        file_browser.files = all_files.filter((f) => f?.isFile).sort((a, b) => a.name.localeCompare(b.name));
        file_browser.dirs = all_files.filter((f) => f?.isDirectory).sort((a, b) => a.name.localeCompare(b.name));
```

nwalters512 (commented):
I think this will be good to go once the sorting issue is resolved.

pages/instructorFileBrowser/instructorFileBrowser.js
jonatanschroeder: Indeed, I was thinking of the Python [sort](https://docs.python.org/3/library/stdtypes.html#list.sort) function.

pages/instructorFileBrowser/instructorFileBrowser.js
```diff
@@ -202,89 +181,74 @@ function browseDirectory(file_browser, callback) {
...
-                callback(null);
-              }
-            );
-          },
-          (err) => {
-            if (ERR(err, callback)) return;
-            callback(null);
-          }
-        );
+      async () => {
```
nwalters512, line 240: Minor nit: we could collapse these into a single `async () => { ... }` function; now that we're not using callbacks, there's no need for them to be separate.

nwalters512 (approved):

pages/instructorFileBrowser/instructorFileBrowser.js
jonatanschroeder: Done in 946fdd6.

nwalters512 (approved):

PR #6271, opened 2022-09-06
feat: Manual grading rubrics

jonatanschroeder:
Rubrics for manual grading.

* [x] Database migration and description
* [x] Tests
* [x] Update instance question grade based on rubric settings
* [x] Re-compute grades for all instance questions based on rubric change
    * [x] Limit it to rubrics that actually changed
    * [x] Re-tag instance questions for manual grading based on option
* [x] Grader interface
    * [x] Create and populate rubric
    * [x] Grade based on rubric
    * [x] Rubric changes
    * [x] Mustache processing of rubric text
    * [x] Markdown processing of rubric text
    * [x] Adjust points interface
* [x] Student interface
* [x] Assessment instance log


jonatanschroeder:
@nwalters512 @mwest1066 I still want to add some UI tests to this interface, other than that this PR should be in a pretty good shape for review.

pages/instructorAssessmentManualGrading/instanceQuestion/gradingPointsSection.ejs
```diff
@@ -0,0 +1,141 @@
...
+    </span>
+  </label>
+</div>
+<% }); %>
+<div class="js-adjust-points d-flex justify-content-end">
+  <button type="button" class="js-adjust-points-enable btn btn-sm btn-link <%= rubric.rubric_grading_data?.adjust_points ? 'd-none' : '' %>">
+    Apply adjustment
+  </button>
+  <div class="js-adjust-points-input-container w-25 <%= rubric.rubric_grading_data?.adjust_points ? '' : 'd-none' %>">
+    <div class="js-manual-grading-points">
```
jonatanschroeder, line 115: TODO Add a label.

pages/instructorAssessmentManualGrading/instanceQuestion/instanceQuestion.js
```diff
@@ -79,13 +86,78 @@ router.get(
   })
 );
 
+router.get(
+  '/grading_panel',
```
nwalters512, line 90: Very minor nit: consider renaming to reflect that this isn't just the grading panel?

pages/instructorAssessmentManualGrading/instanceQuestion/instanceQuestion.js
```diff
@@ -132,6 +197,23 @@ router.post(
           res.locals.instance_question.id
         )
       );
+    } else if (req.body.__action === 'modify_rubric_settings') {
+      const rubric_items = Object.values(qs.parse(qs.stringify(req.body)).rubric_item || {});
```
nwalters512, line 201: Why is `req.body` round-tripped through a stringify and parse?

migrations/20220905173159__rubrics__create.sql
```diff
@@ -0,0 +1,58 @@
...
+    rubric_item_id BIGINT NOT NULL REFERENCES rubric_items(id) ON DELETE CASCADE ON UPDATE CASCADE,
+    score DOUBLE PRECISION NOT NULL DEFAULT 1,
+    points DOUBLE PRECISION NOT NULL,
+    short_text TEXT,
+    note TEXT
+);
+
+-- TODO Consider other indices
+
+ALTER TABLE rubric_grading_items DROP CONSTRAINT IF EXISTS rubric_grading_items_rubric_grading_id_rubric_item_id_key;
```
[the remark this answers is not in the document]
jonatanschroeder: It's to make it idempotent. It should only exist if you run the migration twice, and in that case this statement avoids an error. It's like add if not exists, but constraints don't have that option.

pages/instructorAssessmentManualGrading/instanceQuestion/instanceQuestion.js
jonatanschroeder: `req.body` does not handle inputs that represent objects (e.g., `rubric_item[points]`), it just creates a flat object. `qs` builds an object for that. If you have a better suggestion to handle these I'm open.

migrations/20220905173159__rubrics__create.sql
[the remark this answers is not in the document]
nwalters512: Gotcha. @mwest1066 this is something I haven't really thought about in a while, but why do we care if migrations are idempotent? When would they ever be run twice (except for situations where we need to manually rerun portions, but even then, we shouldn't have to design migrations as a whole to be idempotent).

We *shouldn't* ever be running migrations twice, but if we do, giving ourselves a window (however short) where constraints aren't enforced seems deeply undesirable.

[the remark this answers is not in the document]
jonatanschroeder: I believe (and I may be wrong) that, since these two are running in the same transaction the table would be locked until the constraint is added back. That said, I don't mind removing it.

migrations/20220905173159__rubrics__create.sql
```diff
@@ -0,0 +1,58 @@
...
+    id BIGSERIAL PRIMARY KEY,
+    rubric_grading_id BIGINT NOT NULL REFERENCES rubric_gradings(id) ON DELETE CASCADE ON UPDATE CASCADE,
+    rubric_item_id BIGINT NOT NULL REFERENCES rubric_items(id) ON DELETE CASCADE ON UPDATE CASCADE,
+    score DOUBLE PRECISION NOT NULL DEFAULT 1,
+    points DOUBLE PRECISION NOT NULL,
+    short_text TEXT,
+    note TEXT
+);
+
+-- TODO Consider other indices
```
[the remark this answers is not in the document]
jonatanschroeder: I'll remove this comment. I was considering cases where gradings are selected bottom up instead of top down, but given gradings can be associated to grading jobs and old logs, that would select rows that are no longer active, so top-down is the way to go.

database/tables/rubric_grading_items.pg
```diff
@@ -0,0 +1,16 @@
+columns
+    id: bigint not null default nextval('rubric_grading_items_id_seq'::regclass)
+    note: text
```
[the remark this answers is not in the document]
jonatanschroeder: Does it make a big difference to add that restriction to the database instead of the UI/backend? Should I replace this with a varchar?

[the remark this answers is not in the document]
jonatanschroeder: According to https://www.postgresql.org/docs/current/datatype-character.html, there is no performance advantage of restricting this in the database.

migrations/20220905173159__rubrics__create.sql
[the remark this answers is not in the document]
jonatanschroeder: Removed in ad969bcb5.

migrations/20220905173159__rubrics__create.sql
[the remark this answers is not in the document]
jonatanschroeder: Removed in ad969bcb5.

lib/markdown.js
```diff
@@ -90,6 +99,17 @@ const defaultProcessor = unified()
...
 
+const inlineProcessor = unified()
+  .use(markdown)
+  .use(math)
+  .use(handleMath)
+  .use(gfm)
+  .use(remark2rehype, { allowDangerousHtml: true })
+  .use(raw)
+  .use(sanitize)
+  .use(makeHandler(visitCheckSingleParagraph))
```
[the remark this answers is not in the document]
jonatanschroeder: Done in c373968a6.

database/tables/rubric_items.pg
```diff
@@ -0,0 +1,20 @@
+columns
+    deleted_at: timestamp with time zone
+    description: text
```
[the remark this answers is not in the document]
jonatanschroeder: Done in 19db65d2c.

pages/instructorAssessmentManualGrading/instanceQuestion/instanceQuestion.js
jonatanschroeder: @nwalters512 taking your brain here: the reason I moved to a separate path here was because middlewares that contain some of the data needed in the partials ran before the rubric update, so are out-of-date. This redirection forces the re-running of the middleware. However, the issue I'm having in this particular case now is that the CSRF-token doesn't match since it's generated with a different URL, so attempting to change the settings after this redirection causes a CSRF fail. So I need a different approach altogether. Here are some ideas:

* Instead of redirecting to this new URL, return the data in the POST code itself. This is often used in other contexts, and would likely be my preferred option if possible, since it is also one less RTT for redirection. The problem is that this would mean some of the locals values used in these EJS files would be out-of-date, since the middleware that updated them would have run before the updates, so it would require one of:
    * re-running the middlewares that are affected by the result;
    * updating the libs to re-select the associated values;
    * something else?
* replace this path with a query argument in the regular get, so the originalUrl is still the main page;
* re-compute the csrf-token in here with a custom URL (I don't really like this approach).

Do you (or @mwest1066) have a better idea on how to approach this?

pages/instructorAssessmentManualGrading/instanceQuestion/instanceQuestionClient.js
```diff
@@ -0,0 +1,354 @@
...
+  });
+
+  $('#submission-feedback')
+    .on('input', function () {
+      this.style.height = '';
+      const style = window.getComputedStyle(this);
+      this.style.height =
+        this.scrollHeight + parseFloat(style.paddingTop) + parseFloat(style.paddingBottom) + 'px';
+    })
+    .trigger('input');
```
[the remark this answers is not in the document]
jonatanschroeder: Comments added here, I'll try to add some more comments throughout as I move this as a separate asset.

pages/instructorAssessmentManualGrading/instanceQuestion/instanceQuestion.js
```diff
@@ -2,6 +2,10 @@ const express = require('express');
 const router = express.Router();
 const asyncHandler = require('express-async-handler');
 const util = require('util');
+const qs = require('qs');
```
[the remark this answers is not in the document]
jonatanschroeder: Added as a dependency. I considered these options, but they don't allow for nesting (see https://github.com/PrairieLearn/PrairieLearn/pull/6271#discussion_r1053872582).

tests/manualGrading.test.js
```diff
@@ -576,5 +591,274 @@ describe('Manual Grading', function () {
         feedback_note
       );
     });
+
+    step('set rubric settings should succeed', async () => {
```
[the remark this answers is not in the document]
jonatanschroeder: I agree with this. This is also the case in pages that have dynamic elements like those using bootstrap-table (see the checks in the assessment question page in this same test file). Matt recently mentioned that you use cypress in PT, and I have recently played with puppeteer, so these may be interesting options in the future. I'll mark this as resolved as (to the best of my knowledge) this isn't supported in PL, but I certainly support the need for such tests.

pages/instructorAssessmentManualGrading/instanceQuestion/instanceQuestionClient.js
```diff
@@ -0,0 +1,354 @@
+/* eslint-env browser,jquery */
+/* global assessment_question_max_points, assessment_question_max_manual_points, assessment_question_max_auto_points, Cookies */
```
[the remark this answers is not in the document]
jonatanschroeder: Done in 3ff451ef3.

lib/question.js
```diff
@@ -1711,6 +1712,47 @@ module.exports = {
             callback(null);
           });
         },
+        async () => {
+          if (locals.instance_question) {
+            const manualResult = await sqldb.callAsync('rubric_gradings_select_data', [
```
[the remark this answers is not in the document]
jonatanschroeder: Replaced rubric_gradings_select_data to a lib. The other sprocs are used by each other, so I'll have to work on them in a more coordinated effort to ensure multiple updates use the same transaction.

jonatanschroeder:
@mwest1066 @nwalters512 As I was considering this PR, I just realized some aspects still pending (other than the one in https://github.com/PrairieLearn/PrairieLearn/pull/6271#discussion_r1054484071).

One is how this is handled in the modal presented in case of grading conflict, I'll work on it later this week.

The other issue is in terms of feedback consistency. Currently, feedback is associated to the submission. This PR, however, assigns the rubrics to the instance question. If we had a new submission after manual grading, the rubric items would show up as related to the latest submission, even though that wasn't graded yet. I could assign the rubrics to the submission instead, if you believe that's the best approach. The other option is to use grading jobs as a way to link a submission to a rubric grading, since it has references to both.

Related to the item above: if a new submission is made after manual grading, currently the manual grading page shows a blank feedback as starting point, even if a previous submission had feedback. The rubrics, though, are shown again as before. Is this inconsistent? Should this behaviour change?

I'd still appreciate reviews in the PR in the meantime, if you have the cycles, though I understand if you'd want to discuss this further and review once after this is decided.

jonatanschroeder:
Still pending:

* [ ] Restrict rubric item short text size (https://github.com/PrairieLearn/PrairieLearn/pull/6271#discussion_r1053737576). Client-side restriction is implemented, should consider server-side.
* [ ] Rubric settings modal does not work if called twice in the same page (https://github.com/PrairieLearn/PrairieLearn/pull/6271#discussion_r1054484071).
* [ ] https://github.com/PrairieLearn/PrairieLearn/pull/6271#discussion_r1053859227: is this a valid approach?
* [ ] Review level of feedback, submission vs IQ (https://github.com/PrairieLearn/PrairieLearn/pull/6271#issuecomment-1365481679)
* [ ] Tests:
    * [ ] Rubric with negative grading
    * [ ] Floor/ceiling
    * [ ] Adjust points
    * [ ] Changes in rubric settings recompute existing grades
    * [ ] Changes in rubric item points recompute existing grades that use the item, but not those that don't
    * [ ] New submission after manual grade (depending on discussion for level of feedback above)

pages/instructorAssessmentManualGrading/instanceQuestion/instanceQuestion.js
jonatanschroeder: Handled by saving the CSRF token before submit and restoring it after the panels are replaced.

jonatanschroeder:
> The other issue is in terms of feedback consistency. Currently, feedback is associated to the submission. This PR, however, assigns the rubrics to the instance question. If we had a new submission after manual grading, the rubric items would show up as related to the latest submission, even though that wasn't graded yet. I could assign the rubrics to the submission instead, if you believe that's the best approach. The other option is to use grading jobs as a way to link a submission to a rubric grading, since it has references to both.

Related: I assume that a change in settings in this case would only cause a regrade for the latest submission of an instance question. We should probably discuss this as well.

> Related to the item above: if a new submission is made after manual grading, currently the manual grading page shows a blank feedback as starting point, even if a previous submission had feedback. The rubrics, though, are shown again as before. Is this inconsistent? Should this behaviour change?

Moved to separate issue.

pages/instructorAssessmentManualGrading/instanceQuestion/gradingPanel.ejs
```diff
@@ -0,0 +1,82 @@
...
+        <small id="submission-feedback-help-<%= context %>" class="form-text text-muted">
+          Markdown formatting, such as *emphasis* or `code`, is permitted
+          and will be used to format the feedback when presented to the
+          student.
+        </small>
+    </label>
+  </div>
+  <div class="text-right">
+    <% if (!locals.disable && authz_data.has_course_instance_permission_edit) { %>
+      <button type="submit" class="btn btn-primary" name="__action" value="add_manual_grade">Submit</button>
```
mwest1066, line 61: When I click "Submit" I get the following error on the JS console:
```
An invalid form control with name='score_manual_percent' is not focusable.
```

pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
```diff
@@ -0,0 +1,226 @@
...
+                <input class="form-check-input" type="checkbox" value="true"
+                       name="tag_for_manual_grading">
+                Require all graded submissions to be manually graded/reviewed
+              </label>
+            </div>
+          </div>
+        </div>
+        <div class="modal-footer">
+          <% if (authz_data.has_course_instance_permission_edit) { %>
+          <button type="submit" class="btn btn-primary">Confirm</button>
```
mwest1066, line 219: Sometimes (I can't spot an exact pattern) when I click "Confirm" I get JS console errors:
image

pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
```diff
@@ -0,0 +1,226 @@
...
+                           <% if ((rubric?.starting_points ?? 0) == 0) { %>checked<% } %>>
+                    Positive grading
+                  </label>
+                </div>
+                <div class="form-check">
+                  <label class="form-check-label" data-toggle="tooltip"
+                         title="Start grading at 100% (<%= max_points %> points) and subtract penalties">
+                    <input class="form-check-input js-negative-grading" name="starting_points" type="radio" value="<%= max_points %>"
+                           <% if (rubric?.starting_points == max_points) { %>checked<% } %>>
+                    Negative grading
```
mwest1066, line 41: For the manual rubric (on an auto-graded question) I can't seem to save a rubric as "positive". I select "Positive grading" and click "Confirm" but then when I open the dialog again it has reset to "Negative grading".

mwest1066 (commented):
I'm still reviewing the rest of the code but I'm post an initial batch of comments now to let you start looking at them.

pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
```diff
@@ -0,0 +1,226 @@
...
+          <div class="js-rubric-settings-info">
+            <hr/>
+            <div class="form-row">
+              <div class="col-4">
+                <div class="form-check">
+                  <label class="form-check-label" data-toggle="tooltip"
+                         title="Start grading at zero and add points">
+                    <input class="form-check-input" name="starting_points" type="radio" value="0"
+                           <% if ((rubric?.starting_points ?? 0) == 0) { %>checked<% } %>>
+                    Positive grading
```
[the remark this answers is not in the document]
jonatanschroeder: > Could we add a (?) with a popover (or similar) to explain what these mean? I'm not confident that everyone will know the phrases "positive/negative grading" (also "custom starting points").

A tooltip is already in place in the label, which activates on hover and when selecting the option. Do you think that's enough?

pages/instructorAssessmentManualGrading/instanceQuestion/gradingPanel.ejs
```diff
@@ -0,0 +1,82 @@
...
+      <div class="custom-control custom-switch mx-2">
+        <input class="custom-control-input js-manual-grading-pts-perc-select"
+               name="use_score_perc" type="checkbox">
+        <span class="custom-control-label">Percentage</span>
+      </div>
+    </label>
+  </div>
+  <%- include('gradingPointsSection',
+    { type: 'manual', type_label: 'Manual',
+      points: locals.manual_points ?? instance_question.manual_points,
```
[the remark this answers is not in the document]
jonatanschroeder: > I had a question that was purely auto-graded. I added a rubric for it with Maximum Points: 4. But in the UI it still displayed the max points as being 0.

These are different values with different fields, but I can see where you get the confusion. The rubric's maximum points is the ceiling for points computed from rubric items. It is different from the number of points the question is worth. So you can have a question that is worth 3 points, and based on the rubrics you can have them go at most 3 points, or at most e.g., 4 points (if you want to allow a rubric for bonus), or at most 2 points (unlikely to be used, but would force an adjustment to reach the maximum). This is the opposite of the floor (minimum points in the settings), which keeps the question from going below 0 for penalties, though you can have that value be -1 (if you allow some penalty that affects e.g., autopoints) or 1 (again unlikely, but that ensures the question will always have at least one point).

What is shown in the grading panel is how many manual points are assigned to the question, which indeed is zero. Since the question's points are based on auto-points only, this would make sense.

Do you have a suggestion for a better way to name these to avoid this confusion?

pages/instructorAssessmentManualGrading/instanceQuestion/gradingPanel.ejs
```diff
@@ -0,0 +1,82 @@
...
+      <span class="">Points</span>
+      <div class="custom-control custom-switch mx-2">
+        <input class="custom-control-input js-manual-grading-pts-perc-select"
+               name="use_score_perc" type="checkbox">
+        <span class="custom-control-label">Percentage</span>
+      </div>
+    </label>
+  </div>
+  <%- include('gradingPointsSection',
+    { type: 'manual', type_label: 'Manual',
```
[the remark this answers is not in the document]
jonatanschroeder: > How does the manual grading section work for questions that don't have manual grading enabled? Does adding a rubric here enable manual grading? I know I should read the code but I'm being lazy and asking 😅

In this case the question would remain with the same overall number of points, and the manual grading would not affect the auto-grading, but would be an "adjustment" to the auto-grading. For example, if you have a question worth 0 manual points and 20 auto points, and the autograder gave the student 50% (i.e., 10 points), you can have 5 manual points, which would give the question 15/20 overall points, which are 10/20 auto-points and 5/0 manual points. Similarly, you can have a negative number of manual points as a penalty (e.g., for using something that was not allowed but wasn't catched by the auto-grader), so 10/20 auto-points and -5/0 manual points give you 5/20 overall points.

This is already in place since split points were deployed, it's not new in rubrics. If you believe there is a better way to show this I'm open to hear suggestions.

pages/instructorAssessmentManualGrading/instanceQuestion/gradingPointsSection.ejs
```diff
@@ -0,0 +1,145 @@
...
+    <label for="js-<%= type %>-score-value-input-points-<%= context %>" class="js-manual-grading-points">
+      <%= type_label %> Points:
+    </label>
+    <label for="js-<%= type %>-score-value-input-percentage-<%= context %>" class="js-manual-grading-percentage">
+      <%= type_label %> Score:
+    </label>
+    <span class="float-right">
+      <% if (rubric || type == 'auto') { %>
+        <span class="js-manual-grading-points js-<%= type %>-score-value-info">
+          <span class="js-value-<%= type %>-points"><%= Math.round(points * 100) / 100 %></span> / <%= max_points %>
```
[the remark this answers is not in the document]
jonatanschroeder: > Would this be better?

`points.toFixed(2)` always shows 2 decimals, even if the number of points is an integer. I could replace it with something like:

https://github.com/PrairieLearn/PrairieLearn/blob/a49d8742121dd15a34812df4535d82fb8111d497/pages/partials/pointsFormatter.ejs#L8

pages/instructorAssessmentManualGrading/instanceQuestion/gradingPointsSection.ejs
```diff
@@ -0,0 +1,145 @@
...
+    </div>
+  <% }); %>
+  <div class="js-adjust-points d-flex justify-content-end">
+    <button type="button" class="js-adjust-points-enable btn btn-sm btn-link <%= (rubric_grading?.adjust_points || locals.disable || !authz_data.has_course_instance_permission_edit) ? 'd-none' : '' %>">
+      Apply adjustment
+    </button>
+    <div class="js-adjust-points-input-container w-25 <%= rubric_grading?.adjust_points ? '' : 'd-none' %>">
+      <label>
+        <span class="small">Adjustment:</span>
+        <div class="js-manual-grading-points">
```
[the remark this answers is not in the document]
jonatanschroeder: > Somehow the max points are not being applied to the total points? (I think total points should equal manual points + auto points?)
> 
> image

Total points here is manual plus auto. What's happening is that the adjustment is being limited by the maximum from the rubric (I'm not sure if it should), so auto points is 3.

pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
jonatanschroeder: > For the manual rubric (on an auto-graded question) I can't seem to save a rubric as "positive". I select "Positive grading" and click "Confirm" but then when I open the dialog again it has reset to "Negative grading".

I can see how this may be triggered. In particular, if the question has 0 points for that type (manual or auto), then there is no real difference between positive or negative grading, since in both cases the points start at 0 (which is what this input controls).

Maybe I should hide the radios in case the maximum is already zero, to avoid confusion, and only allow a custom starting point on a checkbox? Or have a combined radio that has better information about this case?

pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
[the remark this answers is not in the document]
jonatanschroeder: Done in 55549cc4a

pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
```diff
@@ -0,0 +1,226 @@
...
+                </div>
+              </div>
+              <div class="form-group col-4">
+                <label data-toggle="tooltip" title="Minimum number of points to be assigned to a student from the rubric (out of <%= max_points %>). Rubric items with negative points will not cause the grade to become lower than this value.">
+                  Minimum points
+                  <input class="form-control" name="min_points" type="number" value="<%= rubric?.min_points ?? 0 %>">
+                </label>
+              </div>
+              <div class="form-group col-4">
+                <label data-toggle="tooltip" title="Minimum number of points to be assigned to a student from the rubric (out of <%= max_points %>). Rubric items with positive points will not cause the grade to become higher than this value.">
```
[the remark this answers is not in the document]
jonatanschroeder: Fixed in 55549cc4a.

pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
```diff
@@ -0,0 +1,226 @@
...
+                  <tr>
+                    <th style="width: 1px"><!-- Order --></th>
+                    <th>Points</th>
+                    <th>Text</th>
+                    <th>Detailed description</th>
+                    <th>Grader note</th>
+                    <th>In use</th>
+                </thead>
+                <tbody>
+                  <% (rubric?.rubric_items || []).forEach((item, index) => { %>
```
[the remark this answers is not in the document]
jonatanschroeder: Done in 55549cc4a

pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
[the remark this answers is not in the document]
jonatanschroeder: Is this clearer?

image

pages/instructorAssessmentManualGrading/instanceQuestion/gradingPanel.ejs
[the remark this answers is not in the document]
jonatanschroeder: Is this clearer?

image

pages/instructorAssessmentManualGrading/instanceQuestion/gradingPanel.ejs
```diff
@@ -0,0 +1,82 @@
...
+      points: locals.manual_points ?? instance_question.manual_points,
+      max_points: assessment_question.max_manual_points,
+      rubric: locals.rubric_data_manual,
+      rubric_grading: submission.rubric_grading_manual
+  }); %>
+  <% if (assessment_question.max_auto_points > 0 || (locals.auto_points ?? instance_question.auto_points) > 0) { %>
+    <%- include('gradingPointsSection',
+      { type: 'auto', type_label: 'Auto',
```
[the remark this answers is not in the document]
jonatanschroeder: I didn't want to overwhelm the interface, so here's an option with list groups:

image

pages/instructorAssessmentManualGrading/instanceQuestion/gradingPointsSection.ejs
```diff
@@ -0,0 +1,145 @@
...
+        <span class="input-group-text">%</span>
+      </span>
+    </div>
+  </div>
+</div>
+<% if (rubric) { %>
+  <style>
+   .js-selectable-rubric-item-label:has(input:checked) {
+       border-color: rgba(0,0,0,.125);
+       border-width: 1px;
```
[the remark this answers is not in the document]
jonatanschroeder: Resolved in 47b637971dc2524bfc99bbdf78571f50d8a58aa4.

database/tables/rubric_items.pg
```diff
@@ -0,0 +1,20 @@
...
+    id: bigint not null default nextval('rubric_items_id_seq'::regclass)
+    key_binding: text
+    number: bigint not null
+    points: double precision not null
+    rubric_id: bigint not null
+    short_text: text not null
```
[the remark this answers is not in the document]
jonatanschroeder: Renamed in eec53c63e, both in UI and in database.

database/tables/rubric_items.pg
```diff
@@ -0,0 +1,20 @@
...
+    staff_instructions: text
```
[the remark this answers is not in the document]
jonatanschroeder: Renamed in eec53c63e.

pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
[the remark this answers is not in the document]
jonatanschroeder: Done in 3cab158d3.

pages/instructorAssessmentManualGrading/instanceQuestion/gradingPointsSection.ejs
```diff
@@ -0,0 +1,145 @@
...
+    <label for="js-<%= type %>-score-value-input-percentage-<%= context %>" class="js-manual-grading-percentage">
+      <%= type_label %> Score:
+    </label>
+    <span class="float-right">
+      <% if (rubric || type == 'auto') { %>
+        <span class="js-manual-grading-points js-<%= type %>-score-value-info">
+          <span class="js-value-<%= type %>-points"><%= Math.round(points * 100) / 100 %></span> / <%= max_points %>
+        </span>
+        <span class="js-manual-grading-percentage js-<%= type %>-score-value-info">
+          <span class="js-value-<%= type %>-percentage"><%= Math.round(points * 10000 / (max_points || assessment_question.max_points)) / 100 %></span>%
```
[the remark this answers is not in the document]
jonatanschroeder: Fixed in e2212d0051ec759c4834aed12644193f6c304a28.

pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
jonatanschroeder: I fixed one particular case where something similar could happen, which would be when the grading job conflict modal is dismissed before opening the rubric settings modal. Not sure if this is the behaviour you're seeing?

pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
jonatanschroeder: This is made clearer in d909578

pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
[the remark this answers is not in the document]
mwest1066: Nice! I love it.

pages/instructorAssessmentManualGrading/instanceQuestion/gradingPanel.ejs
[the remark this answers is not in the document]
mwest1066: Definitely better, I like it!

pages/instructorAssessmentManualGrading/instanceQuestion/gradingPointsSection.ejs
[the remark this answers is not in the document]
mwest1066: Oh right, yes, `toFixed()` is not really what we want here. I'm fine to stay with your existing code. Sorry for the noise!

jonatanschroeder:
Notes from a meeting with @mwest1066:

* For the "Use rubrics" checkbox: it should be removed, and settings should be visible by default. If the question does not currently have a rubric, the "Save" button should say "Set rubric" (or "Enable", or "Add"?). If the question already has a rubric, it should say "Update rubric". In the latter case, a link/button in the bottom left should give the option to disable rubric.
* Ceiling/floor: for clarity the functionality should be changed to indicate the cap for number of points above the max (or below zero) that are allowed to be set. There should be some visual indication when this cap is reached.
* Custom starting points: use case should be rare, remove the option from the UI (though allow it in the backend). If the question is out of 0 (e.g., no manual points), Positive/Negative grading setting should be removed.

jonatanschroeder:
One more: adjustment should not be bound by caps.

jonatanschroeder:
> * For the "Use rubrics" checkbox: it should be removed, and settings should be visible by default. If the question does not currently have a rubric, the "Save" button should say "Set rubric" (or "Enable", or "Add"?). If the question already has a rubric, it should say "Update rubric". In the latter case, a link/button in the bottom left should give the option to disable rubric.

@mwest1066 thoughts?

image

pages/instructorAssessmentManualGrading/instanceQuestion/gradingPointsSection.ejs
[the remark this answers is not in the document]
jonatanschroeder: This is now resolved. The issue was that adjustment was bound to floor/ceiling.

pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
jonatanschroeder: Please let me know if you're still seeing these issues.

pages/instructorAssessmentManualGrading/instanceQuestion/gradingPanel.ejs
jonatanschroeder: This should be resolved, please let me know if this issue persists after the latest changes.

pages/instructorAssessmentManualGrading/instanceQuestion/gradingPanel.ejs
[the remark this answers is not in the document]
jonatanschroeder: Changed to "max extra points", so this confusion shouldn't come up.

jonatanschroeder:
@mwest1066 I believe all pending issues from the latest review are now addressed. The "Disable rubric" link should probably have some sort of confirmation before updating, since it's somewhat destructive, but I'll try to handle this in the near future.

jonatanschroeder:
@mwest1066 I found the problem that was causing an error to pop up when submitting a change in the rubric settings. The problem happens when changing settings twice in the same page load. In essence this happens because the settings change causes new panels to be loaded, including a new version of the settings page, and these settings have a new CSRF token that is considered invalid since the URL of the panels request does not match the URL of the POST page. I'm working on a fix, just adding it here to document it in case I'm unable to come back to it today.

jonatanschroeder:
> @mwest1066 I found the problem that was causing an error to pop up when submitting a change in the rubric settings. The problem happens when changing settings twice in the same page load. In essence this happens because the settings change causes new panels to be loaded, including a new version of the settings page, and these settings have a new CSRF token that is considered invalid since the URL of the panels request does not match the URL of the POST page. I'm working on a fix, just adding it here to document it in case I'm unable to come back to it today.

And it's fixed in 6cef3e858.

pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
jonatanschroeder: This has been resolved in 6cef3e858. Issue was CSRF tokens, which were handled in a previous version, but the switch from jQuery to vanilla JS broke this.

pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
SethPoulsen, line 1: Can the new `ejs` files in this PR be done using `@prairielearn/html` instead? 

Some examples can be seen in recent PRs of Nathan and I:
Batched migrations page: https://github.com/PrairieLearn/PrairieLearn/blob/master/pages/administratorBatchedMigrations/administratorBatchedMigrations.html.js
Question sharing settings page (still on a branch): https://github.com/PrairieLearn/PrairieLearn/blob/question-sharing-impl/pages/instructorCourseAdminSharing/instructorCourseAdminSharing.html.js

lib/assessment.sql
```diff
@@ -347,6 +389,8 @@ WITH
         JOIN assessment_questions AS aq ON (aq.id = iq.assessment_question_id)
         JOIN questions AS q ON (q.id = aq.question_id)
         LEFT JOIN users AS u ON (u.user_id = gj.auth_user_id)
+        LEFT JOIN rubric_gradings AS mrg ON (mrg.id = gj.manual_rubric_grading_id)
+        LEFT JOIN rubric_gradings AS arg ON (mrg.id = gj.auto_rubric_grading_id)
```
SethPoulsen, line 393:
```suggestion
        LEFT JOIN rubric_gradings AS mrg ON (mrg.id = gj.manual_rubric_grading_id)
        LEFT JOIN rubric_gradings AS arg ON (arg.id = gj.auto_rubric_grading_id)
```
possible copy-paste error here?

lib/manualGrading.js
```diff
@@ -34,4 +51,559 @@ module.exports.nextUngradedInstanceQuestionUrl = async (
...
+
+    return { id: rubric_grading_result.id, computed_points };
+  });
+}
+
+/** Manually updates the score of an instance question.
+ * @param {number} assessment_id - The ID of the assessment associated to the instance question. Assumed to be safe.
+ * @param {number} instance_question_id - The ID of the instance question to be updated. May or may not be safe.
+ * @param {number|null} submission_id - The ID of the submission. Optional, if not provided the last submission if the instance question is used.
+ * @param {string|null} check_modified_at - The value of modified_at when the question was retrieved, optional. If provided, and the modified_at value does not match this value, a grading job is created but the score is not updated.
```
SethPoulsen, line 364: I always love seeing types declared :)

SethPoulsen (commented):
Well done Jonaton, this is such an amazing feature, I'm super excited to use it myself in the future.

I've not read all the code yet, but I did leave some comments on what I read so far.

I am surprised to see such a big change come with no documentation changes. Is the documentation in another PR or has it not been written yet?

The current documentation for Manual Grading at https://prairielearn.readthedocs.io/en/latest/manualGrading/#manual-grading-using-the-web-user-interface
feels a bit like a wall of text at the moment. It would be really nice to have it broken up with some screenshots and sub-headings, including a new subsection to give details on how to use these new rubrics, etc.

I've checked out this branch locally and I actually can't even figure out how to get the manual grading modals open. (I did run the Admin query to enable manual grading, not sure what else I need to do.)

It looks like the rubrics are being defined completely in the modals, is that right? Is there a reason we aren't having JSON or Mustache files to define the rubrics instead? It seems like that would make it easier to copy around and edit them.

We might also want to add some stuff to `exampleCourse` to give more examples of manual & auto rubrics.

lib/assessment.sql
jonatanschroeder: Great catch, thanks. Fixed in cc5d61e.

pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
jonatanschroeder: > Can the new `ejs` files in this PR be done using `@prairielearn/html` instead?

This is just part of a page that uses this as an include. If this were a new page this could be considered, but note that this PR started in September...

jonatanschroeder:
> I am surprised to see such a big change come with no documentation changes. Is the documentation in another PR or has it not been written yet?

Good question! We don't have a lot of documentation on features that are UI-only. Given this is still behind feature flags, I didn't prioritize documentation, but I can do it in future PRs.

> The current documentation for Manual Grading at https://prairielearn.readthedocs.io/en/latest/manualGrading/#manual-grading-using-the-web-user-interface feels a bit like a wall of text at the moment. It would be really nice to have it broken up with some screenshots and sub-headings, including a new subsection to give details on how to use these new rubrics, etc.

Good point. Created #7560.

> I've checked out this branch locally and I actually can't even figure out how to get the manual grading modals open. (I did run the Admin query to enable manual grading, not sure what else I need to do.)

There should be a "Rubric" button beside the manual points in the manual grading page for an instance question.

> It looks like the rubrics are being defined completely in the modals, is that right? Is there a reason we aren't having JSON or Mustache files to define the rubrics instead? It seems like that would make it easier to copy around and edit them.

This was a discussion we've had. We want to allow instructors to change these rubrics without the need for sync to get in the way. Note that the use-case for these is that new rubric items may be added as the grading process progresses, and instructors/TAs should be able to update the rubric as new cases are found. There are plans to have the ability to import rubrics in the future, either from JSON files or from other assessments using the same question, but this is future work.

> We might also want to add some stuff to `exampleCourse` to give more examples of manual & auto rubrics.

Given rubrics are set by the instructor, not in JSON files, I'm not sure these would be beneficial at this stage.

SethPoulsen:
> "Rubric" button beside the manual points in the manual grading page for an instance question.

So this means there must have already been submissions made to the question in order to view and edit the rubric? That seems rather inconvenient as I can see a lot of instructors wanting to construct the rubric when assembling the assessment in the first place. 

I kept looking around on the question preview page trying to find some way to get to it. Although I guess the fact that the rubric is linked to the `assessment_question` makes it difficult for it to be shown on the question preview page.

Speaking of which, what are the arguments for having the rubric linked to the `assessment_question` instead of the `question`?

SethPoulsen:
Something is going wrong here with negative grading. When I select negative grading, it still adds the points instead of subtracting them:
image

SethPoulsen:
I don't really understand what is going on with the autograde rubrics. Where do I access those? What's the purpose of those, is it to enable more configurability to autograding?

SethPoulsen:
What is going to be the story for scaling this to our power users that don't like clicking things in the UI, but rather doing as much as possible through configuration files? 

I know the fact that question sharing is done through the UI and not the config files is a bother to some people--we are planning to add 'bulk editing' features. 

I'm not saying we have to have this all figured out now, I just think that we should make sure we know what the story is going to be and how its going to work before we commit ourselves to a particular DB structure or anything. For example, how can you easily copy a rubric used from one question to another? (whether it is for another instance of the same question being used, or a different one).

I love the user interface for grading questions, I really appreciate the keyboard shortcuts and everything. 

I'm not sold on having the rubric construction be in the UI instead of the JSON. I think would be nice to have the rubric declared in assessmentConfig.json right next to where the points for the question are declared. In fact, I think its quite odd right now that you select the point values for the rubric items in the UI, but if you want to change the total number of points, you have to go somewhere else entirely to edit it in the JSON. Having it in JSON would also make it so much easier to copy rubrics from one place to the next. Suppose I want to reuse a test from prior semester that has a bunch of questions with manual grading rubrics? I need to go click a button to copy it over for each one of them? The story is a lot nicer if its in the JSON, you just copy the file, just like how everything else about assessment configuration works.

SethPoulsen:
image

The message here is a little bit odd--it makes it sounds like PL itself chose 0 to 5 as the default range, but that's not case, is it? The 5 point max is because of the value set in the `assessmentConfig.json`? Right now it is totally unclear to a user in this modal how to go about editing that value, or that it is even possible to edit.

jonatanschroeder:
> So this means there must have already been submissions made to the question in order to view and edit the rubric? That seems rather inconvenient as I can see a lot of instructors wanting to construct the rubric when assembling the assessment in the first place.

This is indeed how we're handling it for now, using Gradescope and similar systems as a model. We may consider something different in the future, with a different starting point.

> I kept looking around on the question preview page trying to find some way to get to it. Although I guess the fact that the rubric is linked to the `assessment_question` makes it difficult for it to be shown on the question preview page.
> 
> Speaking of which, what are the arguments for having the rubric linked to the `assessment_question` instead of the `question`?

You can reuse the same question in multiple assessments over multiple course instances. The decision on rubric is dependent on the assessment, in the same way the number of points is linked there. Different assessments can give different weights and measure different outcomes when using rubrics.

jonatanschroeder:
> Something is going wrong here with negative grading. When I select negative grading, it still adds the points instead of subtracting them: image

For negative grading, the rubric items should have negative points. Note that adding an item when negative grading is selected will default to a negative point.

jonatanschroeder:
> The message here is a little bit odd--it makes it sounds like PL itself chose 0 to 5 as the default range, but that's not case, is it? The 5 point max is because of the value set in the `assessmentConfig.json`? Right now it is totally unclear to a user in this modal how to go about editing that value, or that it is even possible to edit.

This is based on the assessment question's number of points. I can try to make this clearer.

jonatanschroeder:
> I don't really understand what is going on with the autograde rubrics. Where do I access those? What's the purpose of those, is it to enable more configurability to autograding?

We don't expect them to be used much, mostly to allow instructors to override the autograder for students that don't get points from automatic tests.

jonatanschroeder:
> What is going to be the story for scaling this to our power users that don't like clicking things in the UI, but rather doing as much as possible through configuration files? 

Remember that this is an MVP. Those decisions will be done later. There will be options to import rubrics from JSON et al later. This is also why this feature is behind a flag now.

> I'm not saying we have to have this all figured out now, I just think that we should make sure we know what the story is going to be and how its going to work before we commit ourselves to a particular DB structure or anything. For example, how can you easily copy a rubric used from one question to another? (whether it is for another instance of the same question being used, or a different one).

The fact that rubrics are a separate table already keeps this in mind. The rubric is not linked to the question, it's the other way around.

> I'm not sold on having the rubric construction be in the UI instead of the JSON. I think would be nice to have the rubric declared in assessmentConfig.json right next to where the points for the question are declared. In fact, I think its quite odd right now that you select the point values for the rubric items in the UI, but if you want to change the total number of points, you have to go somewhere else entirely to edit it in the JSON.

Note that many decisions talked to manual grading are done after knowing what kind of answer you get, which is one of the features of manual grading vs automated grading. You don't always know what kinds of errors students will make. As for the total, that is the kind of information that you shouldn't change after submissions, since students may have made decisions as to what questions to prioritize based on this information.

> Having it in JSON would also make it so much easier to copy rubrics from one place to the next. Suppose I want to reuse a test from prior semester that has a bunch of questions with manual grading rubrics? I need to go click a button to copy it over for each one of them? The story is a lot nicer if its in the JSON, you just copy the file, just like how everything else about assessment configuration works. 

There are some ideas to simplify this process in the future. I'll try to create an issue to keep track of some of these ideas.

jonatanschroeder:
Thinking about the case for auto rubrics: since we don't expect them to be used consistently, I was thinking that maybe we should hide the "Rubric" button for auto points until the instructor clicks on the pencil button. This should avoid cases where the instructor chooses this option by mistake, by clarifying that it should only be done to manually change the auto points if necessary.

SethPoulsen:
> hide the "Rubric" button for auto points until the instructor clicks on the pencil button

I agree it makes sense to have things more-hidden if they are going to be used rarely only by advanced users. I must admit though I still don't understand the purpose of auto rubrics at all. 

Isn't there already a `weight` attribute on elements that can be used to weight autograded elements? Are we deprecating that? does the auto rubric give same functionality or different?

SethPoulsen:
Similar to the max points thing, its not clear in the modal that these values are actually set by the user in the python file (right?)
image

SethPoulsen:
Also, what is the "Apply Adjustment" thing supposed to be used for?

jonatanschroeder:
> I agree it makes sense to have things more-hidden if they are going to be used rarely only by advanced users. I must admit though I still don't understand the purpose of auto rubrics at all.
> 
> Isn't there already a `weight` attribute on elements that can be used to weight autograded elements? Are we deprecating that? does the auto rubric give same functionality or different?

The idea is for when the autograder of a question is too strict or unable to check for partial grades, and we want to allow students to get a manual review of the code. That said, this may be a case for moving this kind of functionality to manual rubrics instead and get rid of auto rubrics. @mwest1066 or @nwalters512, can we get in touch at some point to brainstorm this?

> Similar to the max points thing, its not clear in the modal that these values are actually set by the user in the python file (right?) image

Sure, I can make this clearer.

> Also, what is the "Apply Adjustment" thing supposed to be used for?

In some cases instructors may want to bypass the rubric and add or subtract credit based on something that doesn't neatly fit into individual rubric items (e.g., a 1 point deduction or credit for an unconventional quirk in the solution).

SethPoulsen:
Another thing we should make sure we think about: how do the rubrics behave when the max points is changed in the JSON? Does it behave the same whether the rubric is in point or percentage mode? I don't know if I have the answers I just want to make sure we talk about it and have reasonable behavior that is documented and tested.

SethPoulsen:
> That said, from some experience in the past I should point out that PL does not usually behave well in that case as well, rubrics aside, so I don't think this discussion should happen here. Currently, if there are submissions in an assessment and the points allocation of that assessment changes, the grades will not necessarily follow suit.

Good point.

jonatanschroeder:
Auto rubrics removed altogether. Replaced with a setting that allows manual grading rubrics to override auto points.

apps/prairielearn/src/pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
```diff
@@ -0,0 +1,276 @@
...
+          <hr/>
+          <div class="form-check">
+            <label class="form-check-label">
+              <input class="form-check-input" type="checkbox" value="true"
+                     name="tag_for_manual_grading">
+              Require all graded submissions to be manually graded/reviewed
+            </label>
+          </div>
+        </div>
+        <div class="js-settings-error-alert-placeholder"></div>
```
mwest1066, line 259: Maybe move this up to be just above the "Add item" button? I think it's better to have the warning close to the controls that are causing the problem, and preferably the button to fix the problem should be just after the warning message.

apps/prairielearn/src/pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
```diff
@@ -0,0 +1,276 @@
...
+            </div>
+            <div class="col-6">
+              <div class="form-check">
+                <label class="form-check-label">
+                  <input class="form-check-input js-replace-auto-points-input" name="replace_auto_points"
+                         type="radio" value="false" required
+                         data-max-points="<%= assessment_question.max_manual_points %>"
+                         <% if (!(locals.rubric_data?.replace_auto_points ?? !assessment_question.max_manual_points)) { %>checked<% } %>
+                         aria-describedby="rubric-settings-replace-auto-explanation">
+                  Apply rubric to manual points (keep auto points)
```
mwest1066, line 41: Because this is the default option (and the "normal" method of operation) I think it should be first?

apps/prairielearn/src/pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
```diff
@@ -0,0 +1,276 @@
...
+                </label>
+              </div>
+            </div>
+          </div>
+          <small id="rubric-settings-replace-auto-explanation" class="form-text text-muted">
+            If the rubric is applied to total points, then a student's auto points will be ignored, and the rubric items will be based on the total points of the question (<%= assessment_question.max_points %> points).
+            <br/>
+            If the rubric is applied to manual points only, then a student's auto points are kept, and the rubric items will be added to (or subtracted from) the autograder results.
```
mwest1066, line 49: I really like having lots of help text in this dialog, but it makes everything very busy (especially when you see the dialog for the first time). Could we shift the help text to popovers triggered by (?) icons at the end of different options? It would be fine if multiple popovers have the same help-text in them, such as for the positive/negative grading help text which applies to both radio options.

apps/prairielearn/src/pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
```diff
@@ -0,0 +1,276 @@
...
+        <div class="js-settings-error-alert-placeholder"></div>
+        <div class="modal-footer">
+          <% if (authz_data.has_course_instance_permission_edit) { %>
+          <% if (locals.rubric_data) { %>
+          <button type="button" class="btn btn-link btn-sm js-disable-rubric-button mr-auto">
+            Disable rubric
+          </button>
+          <% } %>
+          <button type="submit" class="btn btn-primary">
+            <%= locals.rubric_data ? 'Update rubric' : 'Set rubric' %>
```
mwest1066, line 268: I think I'd prefer just "Save" in all cases. I was initially confused about "Set rubric".

apps/prairielearn/src/pages/instructorAssessmentManualGrading/instanceQuestion/gradingPanel.ejs
```diff
@@ -0,0 +1,101 @@
...
+    </li>
+    <% if (!assessment_question.max_auto_points && !(locals.auto_points ?? instance_question.auto_points)) { %>
+    <li class="list-group-item">
+      <%- include('manualPointsSection'); %>
+      <%- include('rubricInputSection'); %>
+    </li>
+    <% } else { %>
+    <li class="list-group-item">
+      <%- include('manualPointsSection'); %>
+      <% if (!locals.rubric_data?.replace_auto_points) { %><%- include('rubricInputSection'); %><% } %>
```
mwest1066, line 34: At the moment the "Rubric" button that lets us edit the rubric can end up far away from the rubric itself (see image). We could put the auto points above the manual points (they do come first, temporally) and we could also shift the "Rubric" button to be right next to the rubric items if we have a rubric (rather than being next to the manual grading points)?

image

apps/prairielearn/src/migrations/20220905173159__rubrics__create.sql
```diff
@@ -0,0 +1,60 @@
...
+
+CREATE TABLE IF NOT EXISTS
+  rubric_grading_items (
+    id BIGSERIAL PRIMARY KEY,
+    rubric_grading_id BIGINT NOT NULL REFERENCES rubric_gradings (id) ON DELETE CASCADE ON UPDATE CASCADE,
+    rubric_item_id BIGINT NOT NULL REFERENCES rubric_items (id) ON DELETE CASCADE ON UPDATE CASCADE,
+    score DOUBLE PRECISION NOT NULL DEFAULT 1,
+    points DOUBLE PRECISION NOT NULL,
+    description VARCHAR(100) NOT NULL,
+    note VARCHAR(10000)
```
mwest1066, line 47: Match the column name from `rubric_items`?
```suggestion
    grader_note VARCHAR(10000)
```

apps/prairielearn/src/migrations/20220905173159__rubrics__create.sql
```diff
@@ -0,0 +1,60 @@
...
+  );
+
+ALTER TABLE rubric_grading_items
+ADD CONSTRAINT rubric_grading_items_rubric_grading_id_rubric_item_id_key UNIQUE (rubric_grading_id, rubric_item_id);
```
mwest1066, line 51: Not that it really matters, but you can write these UNIQUE constraints inside the CREATE TABLE expressions by writing `UNQIUE (...)` as the last line in the table definition.

testCourse/questions/manualGrade/codeUpload/server.py
mwest1066, line 1: When doing this question as a student it was not actually auto-grading my code uploads (it just immediately reported 0% auto points all the time). I assume this isn't what it's supposed to be doing?

apps/prairielearn/assets/scripts/instructorAssessmentManualGradingInstanceQuestion.js
```diff
@@ -0,0 +1,515 @@
+import ClipboardJS from 'clipboard';
+
+$(() => {
```
mwest1066, line 3: I appreciate that this file contains very little jQuery. Should we eliminate it entirely? We'd need to copy over `src/lib/document-ready.ts` from PT and make a few other minor changes below.

apps/prairielearn/assets/scripts/instructorAssessmentManualGradingInstanceQuestion.js
```diff
@@ -0,0 +1,515 @@
...
+          `.js-selectable-rubric-item[data-key-binding="${event.key}"]:not(:disabled)`
+        )
+        .forEach((item) => item.dispatchEvent(new MouseEvent('click')));
+    }
+  });
+  const modal = document.querySelector('#conflictGradingJobModal');
+  if (modal) {
+    $(modal)
+      .modal({})
+      .on('shown.bs.modal', function () {
```
mwest1066, line 27: I think this can be replaced by:
```
document.querySelector('#conflictGradingJobModal')?.addEventListener('show.bs.modal', (e) => {
    const modal = e.target as HTMLElement;
```

apps/prairielearn/assets/scripts/instructorAssessmentManualGradingInstanceQuestion.js
```diff
@@ -0,0 +1,515 @@
...
+        modal
+          .querySelectorAll('.js-submission-feedback')
+          .forEach((item) => item.dispatchEvent(new Event('input')));
+      });
+  }
+});
+
+function resetInstructorGradingPanel() {
+  $('[data-toggle="tooltip"]').tooltip();
```
mwest1066, line 36: Use something like:
```
new window.bootstrap.Tooltip(el);
```

apps/prairielearn/assets/scripts/instructorAssessmentManualGradingInstanceQuestion.js
```diff
@@ -0,0 +1,515 @@
...
+    clipboard.on('success', (e) => {
+      e.trigger.animate(
+        [
+          { backgroundColor: '', color: '', offset: 0 },
+          { backgroundColor: '#000', color: '#fff', offset: 0.5 },
+          { backgroundColor: '', color: '', offset: 1 },
+        ],
+        500
+      );
+      $(e.trigger)
```
mwest1066, line 51: Can use something like:
```
window.bootstrap.Tooltip.getInstance(el)?.dispose();
```

apps/prairielearn/src/pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
jonatanschroeder: Done in 0b32e15ac. This warning was reusing the same placeholder used for submit errors, these were separated now.

mwest1066 (approved):
Approving this subject to the comments above.

apps/prairielearn/src/pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
jonatanschroeder: Done in 61330bb5f. Also added the max points for each option for clarity.

apps/prairielearn/src/pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
jonatanschroeder: I replaced the text with tooltips in 61330bb5f, let me know if this is clearer. I'm happy to add a circle-info icon as well, if you prefer that.

apps/prairielearn/src/migrations/20220905173159__rubrics__create.sql
jonatanschroeder: Deleted in 8badea102.

apps/prairielearn/src/migrations/20220905173159__rubrics__create.sql
jonatanschroeder: Updated in 8badea102.

testCourse/questions/manualGrade/codeUpload/server.py
jonatanschroeder: This question has a grading method Manual, and the assessment does not assign auto points, so there is no autograder here.

apps/prairielearn/assets/scripts/instructorAssessmentManualGradingInstanceQuestion.js
jonatanschroeder: It still uses `tooltip`, `popover` and `modal` from bootstrap, which in v4 still rely on jQuery, so eliminating it entirely is not possible, unfortunately.

apps/prairielearn/assets/scripts/instructorAssessmentManualGradingInstanceQuestion.js
jonatanschroeder: Not in bootstrap v4.

apps/prairielearn/src/lib/manualGrading.js
```diff
@@ -34,4 +51,531 @@ module.exports.nextUngradedInstanceQuestionUrl = async (
...
+ * @param {Object} [score.feedback] - Feedback data to be provided to the student. Freeform, though usually contains a `manual` field for markdown-based comments.
+ * @param {Object} [score.partial_scores] - Partial scores associated to individual elements. Must match the format accepted by individual elements. If provided, auto_points are computed based on this value.
+ * @param {Object} [score.manual_rubric_data] - Rubric items associated to the grading of manual points. If provided, overrides manual points.
+ * @param {Object} [score.manual_rubric_data.rubric_id] - Rubric ID to use for manual grading.
+ * @param {AppliedRubricItem[]} [score.manual_rubric_data.applied_rubric_items] - Applied rubric items.
+ * @param {number | string} [score.manual_rubric_data.adjust_points=0] - number of points to add (positive) or subtract (negative) from the total computed from the items.
+ * @param {number} authn_user_id - The user_id of the logged in user.
+ * @returns {Promise<Object>}
+ */
+async function updateInstanceQuestionScore(
```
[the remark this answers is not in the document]
jonatanschroeder: This function (and previously the sproc) was mainly used for manual grading (UI, upload, or manually setting the points in the assessment instance page). I think the placement here makes sense. IMHO the fact that it is able to change the auto-points is not relevant, since it would be used in manual updates of those points.

apps/prairielearn/assets/scripts/instructorAssessmentManualGradingInstanceQuestion.js
jonatanschroeder: The modal creation requires Bootstrap v4 anyway, so adding the event listener in vanilla JS would not be a big improvement since we already have the jQuery object anyway.

apps/prairielearn/src/pages/instructorAssessmentManualGrading/instanceQuestion/gradingPanel.ejs
jonatanschroeder: Moved to total points in 50d5a9d96.

apps/prairielearn/src/pages/instructorAssessmentManualGrading/assessmentQuestion/assessmentQuestion.ejs
```diff
@@ -12,304 +12,305 @@
   </head>
   <body>
     <script>
```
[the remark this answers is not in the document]
jonatanschroeder: This PR barely touches this page, I'd prefer to do that in a future PR.

testCourse/questions/manualGrade/codeUpload/server.py
mwest1066: Ah ok, all good then. Thanks for clarifying! Should we change this so that it does auto-grade, to demonstrate the functionality?

jonatanschroeder: This question is used in tests, which rely on its manual-only configuration.

apps/prairielearn/src/pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
mwest1066: I personally prefer explicit "(?)" buttons to click on to get help, where there is space to do it. I reserve tooltips for when there really isn't space (or the help buttons would clutter the interface too much). For example, in GH's comment editor they have tooltips on all the icons above the text area, which seems like a good use for tooltips. In this case I think we have enough space for help buttons.

mwest1066: @nwalters512 thoughts about this?

apps/prairielearn/src/migrations/20220905173159__rubrics__create.sql
mwest1066: Ah ok, I misunderstood the purpose of this. Thanks for explaining!

apps/prairielearn/assets/scripts/instructorAssessmentManualGradingInstanceQuestion.js
mwest1066: Ah, ok, fair enough!

apps/prairielearn/assets/scripts/instructorAssessmentManualGradingInstanceQuestion.js
mwest1066: Sure, I'm fine with this.

apps/prairielearn/src/lib/manualGrading.js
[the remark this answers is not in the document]
mwest1066: Ok, sounds good, thanks!

apps/prairielearn/src/pages/instructorAssessmentManualGrading/assessmentQuestion/assessmentQuestion.ejs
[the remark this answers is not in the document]
mwest1066: Completely reasonable.

apps/prairielearn/src/pages/instructorAssessmentManualGrading/instanceQuestion/rubricSettingsModal.ejs
nwalters512: In case my experience was useful: I just opened up the rubric modal for the first time and was like "wow, we should have some explanatory text here", not realizing that it would appear when I hovered on inputs! I think an explicit trigger would be an improvement, but I also wonder if we should err on the side of not hiding it in the first place, a la this type of thing used throughout GitHub's settings:

Screenshot 2023-05-17 at 14 44 25
Screenshot 2023-05-17 at 14 44 05

IMO we've misused popovers pretty frequently throughout PL's UI, and I think "just show it to people" would be an improvement here.

mwest1066: Jonatan's first version actually had the text explicitly written everywhere (like the GH example above). I found it to be rather overwhelming in a dialog. I'm on board with the explicit text for things like settings pages where we are infrequently accessing them and we really want information about the options. This dialog is going to be used frequently by people who will know which options they want after the first few times using the dialog. We should certainly come up with some better guidelines for our use of help text.

@jonatanschroeder I guess you should go with whatever you think here (but probably not tooltips). Either your original design with the text just in the dialog or something with explicit popover triggers seem reasonable to me (with my vote going for "(?)" buttons).

jonatanschroeder: Would an ℹ️ with a tooltip (instead of a popover) work?

mwest1066: Yep, works for me!

jonatanschroeder: Done in 489180211. Let me know if you want to change the icon or colour.

image

PR #6298, opened 2022-09-08
Add changeset for recent version bumps

nwalters512:
I've installed https://github.com/apps/changeset-bot so that I don't forget about this in the future.


mwest1066:
This is for things like `prairielearn/html`, right?

mwest1066:
Ah yes, it is.

nwalters512:
Correct: it's for all the packages we publish so we can use them elsewhere, like PT.

mwest1066 (approved):

PR #6301, opened 2022-09-08
Add config options for compliance with Chinese regulations

nwalters512:
We were notified by ZJUI that we need to make specific changes to our site to comply with local regulations:

- We need to change the `<title>` on all pages to a specific value.
- We need to add the website registration number and a link to the [Ministry of Industry and Information Technology](https://en.wikipedia.org/wiki/Ministry_of_Industry_and_Information_Technology) domain registration website.

Adding these config values will allow us to comply with regulations without hardcoding anything specific into PrairieLearn's source code.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -263,6 +263,13 @@ config.honeycombDataset = 'prairielearn-dev';
 config.sentryDsn = null;
 config.sentryEnvironment = 'development';
 
+/**
+ * In some markets, such as China, the title of all pages needs to be a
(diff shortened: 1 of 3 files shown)


mwest1066 (approved):

trombonekenny:
I asked for clarification if those things were needed on every page, or just the homepage. I haven't heard back yet.

trombonekenny:
> They did not say, but based on the observation of other websites, I think it is only needed for the home page.

PR #6307, opened 2022-09-11
Minor change to the text of the manual feedback panel

firasm:
Submitting a minor change to the text of the manual feedback panel to make it a bit 'friendlier'

"Grader Feedback" ---> "Feedback from the Course Team"

Screen Shot 2022-09-10 at 10 07 28 PM

Let me know if there is anywhere else I need to change.

### DIFF
--- a/pages/partials/submission.ejs
+++ b/pages/partials/submission.ejs
@@ -5,7 +5,7 @@
 <div class="card mb-4 grading-block border-info">
   <div class="card-header bg-info text-white d-flex submission-header <% if (submission.submission_number != submissionCount) { %> collapsed<% } %>" data-toggle="collapse" data-target="#submission-feedback-<%= submission.id %>-body">
     <div class="mr-auto">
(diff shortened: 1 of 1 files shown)


pages/partials/submission.ejs
```diff
@@ -5,7 +5,7 @@
 <div class="card mb-4 grading-block border-info">
   <div class="card-header bg-info text-white d-flex submission-header <% if (submission.submission_number != submissionCount) { %> collapsed<% } %>" data-toggle="collapse" data-target="#submission-feedback-<%= submission.id %>-body">
     <div class="mr-auto">
-      Grader feedback
+      Feedback from the Course Team
```
jonatanschroeder, line 8:
```suggestion
      Feedback from the Course Staff
```
This matches with the naming we're using elsewhere in the system.

jonatanschroeder (approved):

PR #6309, opened 2022-09-11
Add pandoc to graders/r/Dockerfile

eddelbuettel:
This minimal PR adds `pandoc` ~and `pandoc-citeproc`~ to the R autograder container.

### DIFF
--- a/graders/r/Dockerfile
+++ b/graders/r/Dockerfile
@@ -31,6 +31,7 @@ ENV PYTHONIOENCODING=UTF-8
 RUN apt update && apt install -y \
         curl \
 	git \
+        pandoc \
         r-cran-bench \        
 	r-cran-data.table \
 	r-cran-devtools \


nwalters512:
Thanks for the PR! Can you compare the size of the Docker image before and after this change? We need to be cognizant of how heavy these "base" images are.

eddelbuettel:
Completely agree on size being a factor.  If anything, the (pre-PR) image leans heavier on the R package content than we "must" but has a few nice data packages. I would trim there.  `pandoc` has become a base layer we really need, it was my omission that it dropped.  (Our `Dockerfile` in the pre-decessor repo had it.)

```sh
edd@rob:~$ docker images | grep grader-r
prarielearn/grader-r                 withPR6309      fbc7a958fdd0   26 seconds ago   1.39GB
prairielearn/grader-r                latest          f598864eabf0   2 weeks ago      1.12GB
edd@rob:~$ 
```

The container builds in under a minute (as all the CRAN packages comes as binaries, we'll soon have a mirror for that binary repo to announce too).

nwalters512:
Gotcha! According to https://github.com/jgm/pandoc-citeproc, `pandoc-citeproc` is no longer maintained, as `pandoc` comes with `citeproc` built-in. Is it possible to remove that package?

eddelbuettel:
Great catch.  It's kinda reflexive for me as they used install as pairs on Debian / Ubuntu systems.  

Commit adjusted and force-pushed.  Now adds only `pandoc`.

nwalters512 (approved):

PR #6310, opened 2022-09-11
Don't send "Workspace is not running" errors to Sentry

nwalters512:
Throwing an error in the `router` function of `http-proxy-middleware` currently produces an unhandled exception. Until https://github.com/chimurai/http-proxy-middleware/pull/823 is merged, we need special handling to avoid blowing through our Sentry error budget.

### DIFF
--- a/server.js
+++ b/server.js
@@ -275,10 +275,6 @@ module.exports.initExpress = function () {
         { workspace_id }
       );
 
-      if (req.upgrade === true) {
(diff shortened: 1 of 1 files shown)


mwest1066 (approved):
Looks fine, although I have no idea whether `event.message` is the string that is showing up as `Workspace is not running` in Sentry.

nwalters512:
I also don't know for sure... let me try to verify that locally real quick.

PR #6311, opened 2022-09-11
Fix deletion of non-existent chunks

### LINKED ISSUE #6308: missing clientFilesCourseInstance folder throws unhelpful error message
We discovered this with an assessmentInfo.json `text` pointing to `<%= clientFilesCourseInstance =>/formulas.pdf` when the course instance `clientFilesCourseInstance` folder doesn't exist. PL errors:

The "path" argument must be of type string. Received null.

image

We should make that error more descriptive about missing folders or paths.

nwalters512:
Fixes the following Sentry error: https://sentry.io/organizations/prairielearn-inc/issues/3580869456/?project=6704263

Closes #6308.

When we're fetching chunks, we use the presence of an `id` field on the results of [`select_metadata_for_chunks`](https://github.com/PrairieLearn/PrairieLearn/blob/8e752630568f07b018096aaa7f93d92f0dd1335c/lib/chunks.sql#L18) to indicate whether or not a given chunk exists. If it does not, we remove that chunk from disk. However, this was failing for `clientFilesAssessment` chunks, as the resulting row from the database did not include the course instance name or the assessment name. This ultimately causes the following `path.join` call to fail since `chunkMetadata.courseInstanceName` and `chunkMetadata.assessmentName` are both `null`:

https://github.com/PrairieLearn/PrairieLearn/blob/8e752630568f07b018096aaa7f93d92f0dd1335c/lib/chunks.js#L219-L226

Similar issues occur for `clientFilesCourseInstance`; this PR fixes those too.

By changing the joins to use the input values directly instead of the columns on the `chunks` table, we ensure that we select all the necessary metadata.

### DIFF
--- a/lib/chunks.js
+++ b/lib/chunks.js
@@ -536,23 +536,43 @@ module.exports.diffChunks = async ({ coursePath, courseId, courseData, changedFi
   );
 
   // Check for any deleted course instances or their assessments.
-  Object.entries(existingCourseChunks.courseInstances).forEach(([ciid, courseInstanceInfo]) => {
-    if (!courseData.courseInstances[ciid]) {
-      deletedChunks.push({
-        type: 'clientFilesCourseInstance',
-        courseInstanceName: ciid,
-      });
-      existingCourseChunks.courseInstances[ciid].assessments.forEach((tid) => {
+  await Promise.all(
+    Object.entries(existingCourseChunks.courseInstances).map(async ([ciid, courseInstanceInfo]) => {
+      const courseInstanceExists = !!courseData.courseInstances[ciid];
+      const clientFilesCourseInstanceExists = await fs.pathExists(
+        path.join(coursePath, 'courseInstances', ciid, 'clientFilesCourseInstance')
+      );
+      if (!courseInstanceExists || !clientFilesCourseInstanceExists) {
         deletedChunks.push({
-          type: 'clientFilesAssessment',
+          type: 'clientFilesCourseInstance',
           courseInstanceName: ciid,
-          assessmentName: tid,
         });
-      });
-    }
+      }
 
-    if ()
-  });
+      await Promise.all(
+        [...courseInstanceInfo.assessments].map(async (tid) => {
+          const assessmentExists = !!courseData.courseInstances[ciid].assessments[tid];
+          const clientFilesAssessmentExists = await fs.pathExists(
+            path.join(
+              coursePath,
+              'courseInstances',
+              ciid,
+              'assessments',
+              tid,
+              'clientFilesAssessment'
+            )
+          );
+          if (!courseInstanceExists || !assessmentExists || !clientFilesAssessmentExists) {
+            deletedChunks.push({
+              type: 'clientFilesAssessment',
+              courseInstanceName: ciid,
+              assessmentName: tid,
+            });
+          }
+        })
+      );
+    })
+  );
 
   return { updatedChunks, deletedChunks };
 };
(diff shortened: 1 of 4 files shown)


lib/chunks.js
```diff
@@ -968,12 +968,6 @@ module.exports.ensureChunksForCourseAsync = async (courseId, chunks) => {
   // in the results of `select_metadata_for_chunks`. If it does not, we remove
   // the chunk from the course's runtime directory.
   //
-  // At this time, we only do this for `elements`, `elementExtensions`,
-  // `serverFilesCourse`, and `clientFilesCourse`, since it's critical that they
-  // not be present on disk if they're removed from a course. If it becomes
-  // necessary to handle additional chunks in the future, this code can easily
-  // be extended to support that.
-  //
```
nwalters512, line 976: I'm not sure why I added this comment, as I can't find anywhere that we're actually limiting to these chunks. Interestingly, I think that if we were limiting, things would work correctly, as the listed chunks don't depend on any of the joined columns that I changed in this PR.

mwest1066 (approved):

lib/chunks.js
mwest1066: I also can't see anywhere we actually implement this restriction. Perhaps this was originally intended as an optimization, because it doesn't actually matter if we have old question chunks on disk (for example). However, I think it's best to keep it like it currently is (and remove this comment).

nwalters512:
I started writing some tests for this and realized that the implementation isn't quite ready. I won't merge this yet.

lib/chunks.js
nwalters512: I figured out that this refers to the fact that we were only ever deleting chunks from the database, not that we had special-cased deletion from disk.

lib/chunks.js
```diff
@@ -475,6 +475,16 @@ module.exports.diffChunks = async ({ coursePath, courseId, courseData, changedFi
     }
   });
 
+  // Check for any deleted questions.
```
nwalters512, line 478: I ended up implementing deletion for all chunk types; this way everything behaves consistently and there's no special cases.

lib/chunks.js
```diff
@@ -525,6 +535,45 @@ module.exports.diffChunks = async ({ coursePath, courseId, courseData, changedFi
     }
   );
 
+  // Check for any deleted course instances or their assessments.
+  await Promise.all(
+    Object.entries(existingCourseChunks.courseInstances).map(async ([ciid, courseInstanceInfo]) => {
+      const courseInstanceExists = !!courseData.courseInstances[ciid];
+      const clientFilesCourseInstanceExists = await fs.pathExists(
```
nwalters512, line 542: Note that for both `clientFilesCourseInstance` and `clientFilesAssessment`, we need to actually hit the disk, as we can't tell from `courseData` whether or not these directories exist. I expect this to result in a very minor increase in sync time, but it should be dwarfed by the time spend actually doing the sync itself.

tests/chunks.test.js
```diff
@@ -304,55 +326,128 @@ describe('chunks', () => {
...
       ];
 
       // Generate chunks for the example course
-      let jobId = await chunksLib.generateAllChunksForCourseList(course_ids, authn_user_id);
-      await helperServer.waitForJobSequenceSuccessAsync(jobId);
+      await chunksLib.updateChunksForCourse({
+        coursePath: tempTestCourseDir.path,
+        courseId,
+        courseData: await courseDB.loadFullCourse(courseDir),
+      });
```
nwalters512, line 349: This is nicer than using a job, as we'll actually get a meaningful error message if something goes wrong.

lib/chunks.js
mwest1066: Makes sense.

lib/chunks.js
mwest1066: Ah!

lib/chunks.js
mwest1066: Excellent!

mwest1066 (approved):

PR #6313, opened 2022-09-12
pl-order-blocks: CSS and HTML cleanup

SethPoulsen:


SethPoulsen:
I know there is definitely more we could do here, but this is a good start. 

I also made it so that the headers are no longer lists, that should help with Accessibility (of course, I still need to actually make the blocks themselves keyboard accessible!)

elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -12,28 +12,26 @@ $(function() {
...
-                        <i class="fa fa-question-circle" aria-hidden="true" style="visibility: hidden;"></i>
-                    </a>
-                </li>
-            </ul>
-            <ul id="order-blocks-options-{{uuid}}" name="{{answer_name}}" class="pl-order-blocks-connected-sortable {{inline}}">
+            <div class="card-header list-group-item-primary">
+                {{source-header}}
+                <a style="visibility:hidden;" role="button" class="btn btn-light btn-sm" data-toggle="popover" data-html="true" title="Order Blocks" data-content="{{help_text}}" data-trigger="focus" tabindex="0">
+                    <i class="fa fa-question-circle" aria-hidden="true"></i>
+                </a>
```
SethPoulsen, line 19: Without this hidden button here, the headers of the source and target block dropzones are different sizes, which looks funny. But I'm sure there is a better way to do this than having a hidden button. Maybe we actually want to have the help button on both the source and target dropzones? When they are arranged vertically, it looks fine, but it looks a little odd when they are arranged horizontally.

elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -12,28 +12,26 @@ $(function() {
...
                 {{#options}}
                     <li class="{{block_formatting}} shadow" uuid="{{uuid}}" string="{{inner_html}}">{{{inner_html}}}</li>
                 {{/options}}
             </ul>
         </div>
         <div class="column {{dropzone_layout}}">
-            <ul class="list-group-item list-group-item-primary header">
-                <li><strong>{{solution-header}}</strong>
-                    <a role="button" class="btn btn-light btn-sm" data-toggle="popover" data-html="true" title="Order Blocks" data-content="{{help_text}}" data-trigger="focus" tabindex="0">
+            <div class="card-header list-group-item-primary">
```
SethPoulsen, line 28: I know its not ideal to use bootstrap's list colors (e.g. `list-group-item-primary`) for what are supposed to be headers, but its the middle of the semester so I don't feel like we should make a huge change to the existing color scheme at the moment.

elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -46,26 +44,26 @@ $(function() {
...
-            <li class="list-group-item"><strong>Your answer: </strong>
+    <div class="pl-order-blocks-answer-container">
+        <div class="card-header">
+            Your answer:
         {{#correct}}
             <span class="badge badge-success"><i class="fa fa-check" aria-hidden="true"></i> 100%</span>
         {{/correct}}
         {{#partially_correct}}
-            <span class="badge badge-warning"><i class="fa fa-times" aria-hidden="true"></i>{{partially_correct}}%</span>
+            <span class="badge badge-warning"><i class="fa fa-times" aria-hidden="true"></i> {{partially_correct}}%</span>
```
SethPoulsen, line 54: add some spacing that should have been there from the beginning

elements/pl-order-blocks/pl-order-blocks.css
```diff
@@ -35,12 +35,6 @@ li.ui-state-highlight {
   min-width: 3em !important;
 }
 
-ul.header {
-  color: black;
-  list-style-type: none;
-  margin-bottom: 0px;
-}
-
 ul.inline li {
```
nwalters512, line 38: It's a big red flag for these classes to not be scoped to this element, as they'd impact everything on the page. Can you change this to be scoped to an element-specific class like the other selectors? Ditto with everything else in this file.

elements/pl-order-blocks/pl-order-blocks.mustache
nwalters512: If you want them to be the same height, why not just set a specific height on the container, e.g. `height: 3rem`?

I agree it would look strange to have the same popover on both panels.

nwalters512:
I'm not sure if this is related to this PR or not, but the example question refers specifically to a yellow panel that does not seem to exist anymore:

> Drag all the prime numbers below to the yellow box.

elements/pl-order-blocks/pl-order-blocks.mustache
nwalters512: This seems like a reasonable tradeoff.

SethPoulsen:
> refers specifically to a yellow panel

definitely not related, but definitely still should be fixed

elements/pl-order-blocks/pl-order-blocks.css
```diff
@@ -28,61 +28,38 @@ ul.pl-order-blocks-connected-sortable li:hover{
...
-ul.header {
-  color: black;
-  list-style-type: none;
-  margin-bottom: 0px;
-}
-
-ul.inline li {
-  display: inline-block;
-  margin-right: 5px;
-}
```
SethPoulsen, line 47: Its time to get rid of this 'inline' option altogether. I'm almost positive that no one is using it because (1) its not documented anywhere, and (2) its been halfway broken for as long as I can remember and no one has complained. 

And actually, removing it won't completely break any questions that were using it, they will just be formatted slightly differently.

SethPoulsen:
@nwalters512 feel free to have a look at this again when you have a chance. I know you had some opinions about the borders before, I tried to make some a little bit less harsh, but happy to tweak them more if you like.

elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -10,30 +10,23 @@ $(function() {
...
-        <div class="column {{dropzone_layout}}">
-            <ul class="list-group-item list-group-item-primary header">
-                <li><strong>{{source-header}}</strong>
-                    <a role="button" class="btn btn-link btn-sm disabled" data-toggle="popover">
-                        <i class="fa fa-question-circle" aria-hidden="true" style="visibility: hidden;"></i>
-                    </a>
-                </li>
-            </ul>
-            <ul id="order-blocks-options-{{uuid}}" name="{{answer_name}}" class="pl-order-blocks-connected-sortable {{inline}}">
+    <div style="display:flex;flex-wrap:wrap;">
```
nwalters512, line 13: Nit: we generally err towards using Bootstrap classes instead of inline styles. e.g.:

```suggestion
    <div class="d-flex flex-wrap">
```

It doesn't make much difference in practice, but it's a convention nonetheless 🙂

elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -10,30 +10,23 @@ $(function() {
...
                 {{/options}}
             </ul>
         </div>
-        <div class="column {{dropzone_layout}}">
-            <ul class="list-group-item list-group-item-primary header">
-                <li><strong>{{solution-header}}</strong>
-                    <a role="button" class="btn btn-light btn-sm" data-toggle="popover" data-html="true" title="Order Blocks" data-content="{{help_text}}" data-trigger="focus" tabindex="0">
+        <div class="pl-order-blocks-container {{dropzone_layout}}">
+            <div class="card-header pl-order-blocks-header list-group-item-primary">
+                <span style="margin-right:4px;">{{solution-header}}</span>
```
nwalters512, line 24: Nit:

```suggestion
                <span class="mr-2">{{solution-header}}</span>
```

nwalters512 (approved):
LGTM! I'll leave it up to you if you want to accept the two suggestions.

PR #6316, opened 2022-09-12
pl-order-blocks: initial keyboard controls

SethPoulsen:
More work toward addressing #4912

- [ ] handle indenting/de-indenting
- [ ] add proper ARIA tags and announcements that indicate when things are moving around/are movable (possibly separate PR)


SethPoulsen:
I had a great conversation with the creator of the Quorum languge (https://quorumlanguage.com/), 

apparently they have block-based editing that works well for people who are blind of visually impaired. I'm going to have a closer look at what they are doing before I seriously revisit this feature.

PR #6320, opened 2022-09-13
Clean yum cache in executor image

nwalters512:
This shaves several hundred MB off the size of the executor image.

https://github.com/wagoodman/dive looks to be a great tool for identifying where the biggest parts of a Docker image come from.


mwest1066 (approved):
Good find!

PR #6321, opened 2022-09-13
Clean Conda caches after installation

nwalters512:
This shaves another ~80MB off our Docker images.

https://docs.conda.io/projects/conda/en/latest/commands/clean.html


mwest1066 (approved):
Now I'm wondering whether we have other caches that we should be clearing? yum, pip, conda, R, npm, yarn?

nwalters512:
I suspect we do; I'll check that out and open followup PRs if needed.

PR #6325, opened 2022-09-14
Remove mkdocs-material and extra R packages from the plbase image

nwalters512:
This PR makes a few optimizations to `prairielearn/plbase` that reduce the image size from 3.81GB to 3.57GB:

- Removes all R packages except for those in `r-essentials`. These extra packages were only used by two STAT courses, and those courses are no longer running (or at least aren't running with their PL questions that use these R dependencies).
- Removes the `mkdocs-material` Python dependency, which *may* have been used for local development of the PL docs, but should not be in our base image.

In addition to being smaller, the base image is now much quicker to build since we're not compiling dozens of R dependencies from source.


images/plbase/plbase-install.sh
```diff
@@ -27,7 +28,7 @@ yum -y install \
     tar \
     ImageMagick \
     fontconfig-devel \
-    texlive \
+    texlive-base \
```
mwest1066, line 31: While I agree this is probably safe, I'm a bit worried about how to test it. I guess it will get some testing on canary servers, but is this sufficient?

images/plbase/plbase-install.sh
```diff
@@ -61,18 +62,13 @@ arch=`uname -m`
...
     echo "installing R..."
-    conda install r-essentials
+    conda install --channel r r-essentials
 
     echo "installing Python packages..."
     python3 -m pip install --no-cache-dir -r /python-requirements.txt
-
-    echo "installing R packages..."
-    echo "set SKIP_R_PACKAGES=yes to skip this step"
-    Rscript /r-requirements.R
```
mwest1066, line 75: Should we reach out to CEE 202 and ask them to check their questions?

images/plbase/plbase-install.sh
echuber2: Installing only texlive-base and texlive-dvipng doesn't seem to install any actual tex/latex compiler. (The yum texlive-base seems very different from apt texlive-base...)

nwalters512: @mfsilva22 it looks like this was added for your course, can you provide an example question or two that uses LaTeX formatting in matplotlib labels?

images/plbase/plbase-install.sh
nwalters512: That sounds very reasonable, can you take that on?

images/plbase/plbase-install.sh
nwalters512: @echuber2 good catch, after testing with the `demo/annotated/LectureVelocity/1-Introduction` question, it looks like matplotlib does indeed literally just shell out to `latex`, which is a problem because `texlive-base` doesn't include that binary

I spent some time trying to figure out which subset of `texlive-*` packages are necessary, but I kept having to add more and more, and eventually I was almost back at the number of MB that `texlive` itself would install 😅

I'll back this change out; it's a shame that latex on RHEL is such a pain.

images/plbase/plbase-install.sh
mwest1066: I ran the command below in the `pl-cee202` repo and confirmed that none of these packages are being referenced by any of their questions, so I think are good to go.

```
for m in tidyverse remotes RcppArmadillo rmarkdown RSQLite nycflights13 fueleconomy babynames rbenchmark microbenchmark maps maptools mapproj mapdata ggmap fivethirtyeight caret e1071 factoextra gbm glmnet ISLR kernlab klaR mlbench nnet pROC randomForest rpart rpart.plot rsample kableExtra quantreg ; do echo $m ; git grep $m ; done
```

nwalters512: Great, thanks for checking!

PR #6326, opened 2022-09-14
Fix doc link in news item

echuber2:
The link in the news item has the wrong fragment (`#` part).

### DIFF
--- a/docs/course.md
+++ b/docs/course.md
@@ -216,7 +216,7 @@ The assessment set order in `infoCourse.json` is the order in which the assessme
 }
 ```
 
-## Assessment modules
+## Assessment modules<a id="modules"></a>
 
 Each assessment in the course belongs to a _module_ defined in `infoCourse.json`. Modules can represent course topics, chapters or sections, or in PrairieLearn terms, a collection of assessments related to one another, but not necessarily of the same type. This means you can have a module with two homeworks, one lab, and one exam, for example.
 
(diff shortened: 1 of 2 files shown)


echuber2:
At jjb's suggestion, I added a secondary heading ID on the actual "assessment modules" section of the course docs page. This is redundant with the first commit that changed the news link. In theory, adding the secondary permalink may be better for backwards-compatibility (if somebody else bookmarked the older link with `#modules`), but the hacky addition of `<a id="..."></a>` on a heading sticks out as unusual among the docs.

jonatanschroeder:
MkDocs and Python Markdown actually have an extension for permalink: https://www.mkdocs.org/user-guide/writing-your-docs/#writing-with-markdown and https://python-markdown.github.io/extensions/toc/#custom-labels.

nwalters512:
It's highly unlikely that anyone would have bookmarked a link that was *already* broken to begin with; I'm happy to just update the news item and leave the permalink alone.

docs/course.md
```diff
@@ -216,7 +216,7 @@ The assessment set order in `infoCourse.json` is the order in which the assessme
 }
 ```
 
-## Assessment modules
+## Assessment modules<a id="modules"></a>
```
nwalters512, line 219:
```suggestion
## Assessment modules
```

jonatanschroeder (approved):

PR #6330, opened 2022-09-14
Fix three.js element

nwalters512:
At some point, we installed a version of `three.js` that had a few backwards-incompatible changes:

- `FontLoader` and `TextGeometry` were removed from core and offered only as examples.
- `three` added an `exports` field to `package.json` that didn't expose `package.json`, which we relied on for retrieving the package version in order to facilitate caching.

This PR addresses both those issues:

- `FontLoader` and `TextGeometry` (and the functions/classes they rely on) were copied from the examples into the element script.
- `nodeModulesAssetPath` now directly reads `package.json` from disk instead of going through `require(...)`.


jonatanschroeder:
Does this PR (or should it) address #5540?

nwalters512:
No, it strictly gets the element back in the working state that it was previously in. It doesn't change the behavior at all.

mwest1066 (approved):

PR #6331, opened 2022-09-14
Use 403 status code when exam time limit has expired

nwalters512:
This will keep this event out of Sentry, and it also more accurately reflects what the error actually is.

### DIFF
--- a/pages/studentInstanceQuestionExam/studentInstanceQuestionExam.js
+++ b/pages/studentInstanceQuestionExam/studentInstanceQuestionExam.js
@@ -101,7 +101,9 @@ router.post('/', function (req, res, next) {
 
   if (req.body.__action === 'grade' || req.body.__action === 'save') {
     if (res.locals.authz_result.time_limit_expired) {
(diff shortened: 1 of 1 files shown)


pages/studentInstanceQuestionExam/studentInstanceQuestionExam.js
```diff
@@ -101,7 +101,9 @@ router.post('/', function (req, res, next) {
 
   if (req.body.__action === 'grade' || req.body.__action === 'save') {
     if (res.locals.authz_result.time_limit_expired) {
-      return next(new Error('time limit is expired, please go back and finish your assessment'));
+      return next(
+        error.make(403, 'time limit is expired, please go back and finish your assessment')
```
trombonekenny, line 105: The wording on that is a little strange, don't you think? How are they to finish it if the time limit is expired?

nwalters512: Maybe "close" is the right word? PRs welcome if you'd like to tweak it!

PR #6332, opened 2022-09-15
Ensure chunks are present when serving element extension client files

nwalters512:
This should fix the issue that @ffund reported on Slack.


pages/elementExtensionFiles/elementExtensionFiles.js
```diff
@@ -42,11 +43,16 @@ router.get('/*', function (req, res, next) {
     res.removeHeader('Cache-Control');
   }
 
-  let elementFilesDir = path.join(res.locals.course.path, 'elementExtensions');
-  res.sendFile(filename, {
-    root: elementFilesDir,
-    immutable: isCached,
-    maxAge: isCached ? '31536000s' : 0,
+  const coursePath = chunks.getRuntimeDirectoryForCourse(res.locals.course);
+  chunks.ensureChunksForCourse(res.locals.course.id, { type: 'elementExtensions' }, (err) => {
```
mwest1066, line 47: Should we be doing this only if we are a chunk consumer? I'm worried that this new code will only work on chunk servers (the reverse of the old code, which only worked on non-chunk-servers). I think we need this code to work on both chunk and non-chunk servers (e.g., for environments that are not running chunk servers at present).

nwalters512: This code is the same code we're running everywhere else we need to serve files, e.g. `clientFilesCourse`:

https://github.com/PrairieLearn/PrairieLearn/blob/1d6788d479d00b2e54bb6b526c06825dda87de7a/pages/clientFilesCourse/clientFilesCourse.js#L19-L25

`getRuntimeDirectoryForCourse` is aware of whether or not the current server is a chunk consumer:

https://github.com/PrairieLearn/PrairieLearn/blob/1d6788d479d00b2e54bb6b526c06825dda87de7a/lib/chunks.js#L694-L700

As is `ensureChunksForCourse`:

https://github.com/PrairieLearn/PrairieLearn/blob/1d6788d479d00b2e54bb6b526c06825dda87de7a/lib/chunks.js#L960-L964

mwest1066: Oh right, `ensureChunksForCourseAsync()` just returns if we are not a chunks consumer. All good! Sorry for the noise.

mwest1066 (approved):

PR #6334, opened 2022-09-15
Introduce pl-hidden-hints element

eliotwrobson:
Created a new element that gradually displays hints as students submit more answers. A port of an element we developed locally. Also, set up typechecking features for this question and locked pyright to run on python 3.8 (same version as live PL).

## Behavior
After each submission where students provided a complete (not containing format errors) answer to a question, the hidden hints element will advance a counter on the backend. Once the required number of submissions is met for a given hint, it will be displayed to the student. The order + display number of the hint can be controlled by the developer using the `show-after` attribute.

In addition, the hint element can be used in "backend mode", where the hints are all immediately loaded from the backend. This is useful in cases where the submission counter can't be advanced because of the use of an externally graded question.

## Usage
Have added example usage to a question from the QA course. Hints are set in the HTML file for a question and displayed progressively:

Screenshot_20220915_161648

## Screenshots:

Before submissions:
Screenshot_20220915_161821

After 2 submissions:
Screenshot_20220915_162026

At 3 submissions:
Screenshot_20220915_162052

At 4 submissions:
Screenshot_20220915_162125


testCourse/questions/addNumbers/question.html
```diff
@@ -4,3 +4,9 @@
 </pl-question-panel>
 
 <pl-number-input answers-name="c" comparison="sigfig" digits="3" label="$c=$"></pl-number-input>
+
+<pl-hidden-hints>
+  <pl-hint show-after=3>
+    Try counting on your fingers!
+  </pl-hint>
+</pl-hidden-hints>
```
SethPoulsen, line 12: It would be nice to have an example question in the `exampleCourse` too. Most people don't look in the test course, so they won't be able to discover the element.

You should also add it to the documentation in `docs/elements.md`. It would probably go in the "Conditional" elements category.

elements/pl-hidden-hints/pl-hidden-hints.css
```diff
@@ -0,0 +1,22 @@
+details.hidden-hint {
```
SethPoulsen, line 1: minor: it would be nice to have these classes prefixed with `pl-` too.

elements/pl-hidden-hints/pl-hidden-hints.py
```diff
@@ -0,0 +1,102 @@
...
+    html_hints: List[Tuple[int, str]]
+    submission_count: int
+
+
+def prepare(element_html: str, data: pl.QuestionData) -> None:
+    # Set submit count
+    if ('hidden-hints' in data['params']):
+        raise ValueError("Duplicate params key: 'hidden-hints'.")
+
+    hints_dict = data['params'].setdefault('hidden-hints', dict())
```
SethPoulsen, line 18: the convention we have for variable names (even dictionary keys) is to use `snake_case`, so `hidden_hints`.

elements/pl-hidden-hints/pl-hidden-hints.py
```diff
@@ -0,0 +1,102 @@
...
+
+        submission_count = hints_dict['submission_count']
+
+        all_correct = all_questions_correct(data)
+
+        # Parse through hint list to display hints.
+        for (idx, (priority, hint)) in enumerate(hint_list):
+            if priority <= submission_count:
+                # Close hints once all questions are correct
+                show_open = not all_correct and priority == submission_count
```
SethPoulsen, line 64: Should it be priority <= submission_count, so that the student retains access to all the hints they've seen?

SethPoulsen (changes requested):
This looks really cool!

@mwest1066 are we able to have Python type annotations in the code now? Or do will still have servers on too old of a python version?

Makefile
```diff
@@ -62,7 +62,7 @@ typecheck: typecheck-js typecheck-python
 typecheck-js:
 	@yarn tsc
 typecheck-python:
-	@yarn pyright
+	@yarn pyright elements/pl-hidden-hints/pl-hidden-hints.py
```
SethPoulsen, line 65: you shouldn't need to pass this as an arg here because its declared in the pyright config, right?

elements/pl-hidden-hints/pl-hidden-hints.py
SethPoulsen: or maybe in the future we can make it configurable to work either way.

pyrightconfig.json
```diff
@@ -1,4 +1,5 @@
 {
-    "include": ["./elements/pl-order-blocks"],
-    "extraPaths": ["./python"]
+    "include": ["./elements/pl-hidden-hints"],
+    "extraPaths": ["./python"],
+    "pythonVersion": "3.8"
```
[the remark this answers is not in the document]
eliotwrobson: This is what's on the live servers (as far as I know). We should double check with Nate.

elements/pl-hidden-hints/pl-hidden-hints.py
eliotwrobson: The `show_open` variable is just controlling whether the hint box is open when the page loads. The student always retains access to old hints.

elements/pl-hidden-hints/pl-hidden-hints.py
```diff
@@ -0,0 +1,102 @@
...
+        else:
+            raise Exception(f"Child tag type '{child.tag}' not supported by hidden-hints.")
+
+    hints_dict['html_hints'] = html_hints
+
+
+def render(element_html: str, data: pl.QuestionData) -> str:
+    # No need to display hints on answer / submission panels
+
+    if data['panel'] == 'question':
```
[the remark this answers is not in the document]
eliotwrobson: Should that be handled by wrapping in the existing PL panel tags, or with an option as an attribute on the element?

testCourse/questions/addNumbers/question.html
eliotwrobson: Will add examples + docs once some more comments come in and the behavior is finalized. I'm also wondering whether it's worth getting rid of the "backend mode", since client questions can always just generate the HTML through the mustache template.

elements/pl-hidden-hints/pl-hidden-hints.py
SethPoulsen: ah, ok. Sounds good.

pyrightconfig.json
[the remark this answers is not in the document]
nwalters512: Yep, we're still at a minimum of 3.8 right now.

elements/pl-hidden-hints/pl-hidden-hints.py
```diff
@@ -0,0 +1,102 @@
...
+    html_hints = []
+
+    # Read hints defined as child tags and load them into html_hints
+    for i, child in enumerate(element):
+        if child.tag == 'pl-hint':
+            pl.check_attribs(child, ['show-after'], [])
+            priority = pl.get_integer_attrib(child, 'show-after')
+            html_hints.append((priority, pl.inner_html(child)))
+        else:
+            raise Exception(f"Child tag type '{child.tag}' not supported by hidden-hints.")
```
nwalters512, line 33: Nit:
```suggestion
            raise Exception(f"Child tag type '{child.tag}' not supported by pl-hidden-hints.")
```

elements/pl-hidden-hints/pl-hidden-hints.py
```diff
@@ -0,0 +1,102 @@
...
+class HintsDict(TypedDict):
+    """A class with type signatures for the partial scores dict"""
+    html_hints: List[Tuple[int, str]]
+    submission_count: int
+
+
+def prepare(element_html: str, data: pl.QuestionData) -> None:
+    # Set submit count
+    if ('hidden_hints' in data['params']):
+        raise ValueError("Duplicate params key: 'hidden_hints'.")
```
nwalters512, line 16: I believe this means it's invalid to have multiple `<pl-hidden-hints>` on the same question?

elements/pl-hidden-hints/pl-hidden-hints.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class HintsDict(TypedDict):
+    """A class with type signatures for the partial scores dict"""
+    html_hints: List[Tuple[int, str]]
+    submission_count: int
+
+
+def prepare(element_html: str, data: pl.QuestionData) -> None:
+    # Set submit count
```
nwalters512, line 14: This comment doesn't line up with what's on line 15 - maybe move down to line 18?

elements/pl-hidden-hints/pl-hidden-hints.py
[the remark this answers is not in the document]
nwalters512: Yeah, this should be handled with the usual PL panel tags.

elements/pl-hidden-hints/pl-hidden-hints.py
```diff
@@ -0,0 +1,102 @@
...
+    return all(part['score'] == 1.0 for part in partial_scores.values())
+
+
+def grade(element_html: str, data: pl.QuestionData) -> None:
+    hints_dict: HintsDict = data['params']['hidden_hints']
+
+    # Only reveal next hint if the previous answer was valid
+    # (i.e. didn't contain any format errors) but incorrect.
+    if (len(data['format_errors']) == 0 and not all_questions_correct(data)):
+        hints_dict['submission_count'] += 1
```
nwalters512, line 97: I think this could have unexpected behavior if e.g. this element is placed between two graded elements: when `<pl-hidden-hints>` is graded, only the first one would be represented in `partial_scores`.

Separately, I'm a little hesitant to use calls to `grade()` as a proxy for "how many submissions have been made". I think that will generally hold true today, but improved regrading features in the future could result in us calling `grade()` multiple times per submission, even potentially out of order.

eliotwrobson: I think these concerns are valid. When we were working with our own course this was fine, but I think for this version it might be worth adding a "valid submissions so far" counter to the data dict.

elements/pl-hidden-hints/pl-hidden-hints.py
eliotwrobson: This does. It's probably worth removing, so then we just have to ask the client question for an dict key prefix.

elements/pl-hidden-hints/pl-hidden-hints.py
eliotwrobson: Got this working by exposing the internal PL submission counter to the data dict!

eliotwrobson:
@nwalters512 Now that most of the other items have been addressed, what type of testing do you think can be done for this element? If tests allow injection of an arbitrary `data` dict, then it should be straightforward to check that the HTML for this element appears in a straightforward way.

docs/devElements.md
```diff
@@ -79,6 +79,7 @@ The `data` dictionary has the following possible keys (not all keys will be pres
 | `data["manual_grading"]`        | boolean | Whether the question is being rendered in the manual grading view.                                                                                  |
 | `data["panel"]`                 | string  | Which panel is being rendered (`question`, `submission`, or `answer`).                                                                              |
 | `data["extensions"]`            | dict    | A list of extensions that are available to be loaded by this element. For more information see the [element extensions](elementExtensions.md) page. |
+| `data["num_valid_submissions"]` | int     | The number of valid (not containing format errors) submissions by the student for the current variant.                                              |
```
SethPoulsen, line 82: Is this correct that it is for the current variant? A lot of questions change variants after each attempt at them. In that case, I assume you still want to show hints after X number of tries, even if it is actually a different variant on each attempt, right?

docs/elements.md
```diff
@@ -1891,6 +1891,59 @@ element contents only in a specific panel.
 
 ---
 
+### `pl-hidden-hints` element
+
+Display progressive hints that reveal themselves as the number of student submissions increases for the current variant.
```
SethPoulsen, line 1896: same comment about 'for the current variant'

docs/elements.md
```diff
@@ -1891,6 +1891,59 @@ element contents only in a specific panel.
...
+Hints must be opened by the user until they are revealed.
```
SethPoulsen, line 1897: 'Hints must be opened by the user until they are revealed.' The meaning of this sentence is not clear to me. Are you trying to say something about when they are or aren't opened by default? It almost makes it sound like the user _must_ take some action on a hint, but they can just ignore if they want, right?

docs/elements.md
```diff
@@ -1891,6 +1891,59 @@ element contents only in a specific panel.
...
+For the inner `pl-hint` element:
+
+| Attribute    | Type | Default | Description                                                                                              |
+| ------------ | ---- | ------- | -------------------------------------------------------------------------------------------------------- |
+| `show-after` | int  | -       | Number of submissions needed before the current hint is shown. If not set, hint is shown from the start. |
+
+#### Details
+
+Add hints to a question that are revealed with more submissions using the `show-after` attribute. By default, hints without
+`show-after` set are shown at the beginning.
```
SethPoulsen, line 1932: Maybe 'are always shown' instead of 'are shown at the beginning'?

elements/pl-hidden-hints/pl-hidden-hints.py
```diff
@@ -0,0 +1,80 @@
...
+    # Read hints defined as child tags and load them into html_hints
+    for i, child in enumerate(element):
+        if child.tag == 'pl-hint':
+            pl.check_attribs(child, [], ['show-after'])
+
+            # Default show-after to -1 to automatically show hint (closed) at start
+            priority = pl.get_integer_attrib(child, 'show-after', -1)
+            html_hints.append((priority, pl.inner_html(child)))
+        else:
+            raise Exception(f"Child tag type '{child.tag}' not supported by pl-hidden-hints.")
```
SethPoulsen, line 41: Very minor: the error message would be a little more useful if it told the user what to do, instead of just telling them what not to do. Maybe "Tags inside of `pl-hiddent-hints` must be `pl-hint`. 

This is not required, but you could also maybe allow HTML comment nodes, and just do nothing with them. We do this with `pl-order-blocks` so people can leave themself comments in their HTML if they want.

exampleCourse/questions/element/hiddenHints/info.json
```diff
@@ -0,0 +1,7 @@
+{
+    "uuid": "28b2c61f-7808-4c22-a186-faec11920bb0",
+    "title": "Element pl-hidden-hints: Display progressive hits",
```
SethPoulsen, line 3: hits -> hints

python/prairielearn.py
```diff
@@ -15,7 +15,42 @@
...
 import collections
-
+from typing import Dict, Any, TypedDict, Literal, Optional
+from typing_extensions import NotRequired
+
+class PartialScore(TypedDict):
+    "A class with type signatures for the partial scores dict"
+    score: Optional[float]
+    weight: NotRequired[int]
+    feedback: NotRequired[str]
```
SethPoulsen, line 25: `Optional` means the key must exist in the dictionary, but its value can be `None`, but `NotRequired` means that the key doesn't even have to exist in the dictionary, is that right? Is there documentation that this is matches what is expected for this dictionary? I'm not sure myself.

exampleCourse/questions/element/hiddenHints/info.json
SethPoulsen, line 3:
```suggestion
    "title": "Element pl-hidden-hints: Display progressive hints",
```

docs/devElements.md
SethPoulsen: maybe instead of 'variant' it should be 'question instance'?

pyrightconfig.json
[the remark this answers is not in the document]
eliotwrobson: Declaring the python version explicitly is actually _super_ important. Without this, pyright could default to 3.10 (or some other more advanced version), and then typecheck allowing 3.10 syntax. If PL servers are on 3.8, pyright won't throw an error if some non-backwards-compatible 3.10 syntax is used in typechecked files.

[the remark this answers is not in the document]
SethPoulsen: That may be true of someone's local dev environment, but it would never reach production. During the automated CI testing that is done on all pull requests, the tests will be run in a docker container that has the right version for production.

python/prairielearn.py
eliotwrobson: That's correct. I actually don't know what the partial score dictionary is required to have or not have, I just copied this over from the type signature in our repo. @nwalters512 is there any documentation for what's required to be in this dictionary?

pyrightconfig.json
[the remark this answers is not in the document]
SethPoulsen: huh, actually I was just poking around the `plbase` docker container and it seems the python version isn't declared there anywhere that I can find, so I'm not sure how the version of python used for the CI testing is decided. 

In any case, if we are going to pin a particular version of python for CI tests, it should be done in one place, not in multiple places, and it seems like doing that in the container creation makes more sense than doing it in the type checker config. @nwalters512 do you have thoughts on this?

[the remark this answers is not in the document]
SethPoulsen: Also, last I checked with @mwest1066 there was still maybe one or two servers running something lower, like 3.7. Is that still true?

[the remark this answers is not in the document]
eliotwrobson: It looks like the CI version is running python 3.10 though? (see screenshot) I'm not even sure how much the installed python version matters for pyright since it uses node.

Screenshot_20221003_125825

docs/elements.md
```diff
@@ -1891,6 +1891,59 @@ element contents only in a specific panel.
...
+
+#### Example implementations
+
+- [element/hiddenHints]
+
+#### See also
+
+- [`pl-question-panel` for displaying the question prompt.](#pl-question-panel-element)
+- [`pl-submission-panel` for changing how a submitted answer is displayed.](#pl-submission-panel-element)
+- [`pl-answer-panel` for displaying the question's solution.](#pl-answer-panel-element)
```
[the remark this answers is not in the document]
eliotwrobson: Just to emphasize that the hints can be on any panel. In the example course one of the hints is used on the submission panel.

pyrightconfig.json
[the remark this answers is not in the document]
SethPoulsen: Ok, I just dug up some old messages and realized that I misremembered which version Matt had said before. 

So, I guess fixing this at 3.8 should be fine. 

On the other hand, it still seems problematic to me that our CI is running on 3.10 when we have servers on 3.8 for reasons aside from just type checking. That is probably something we should fix, but that's outside of the scope of this PR.

docs/devElements.md
SethPoulsen: hmm...I worry that this may be confusing for instructors. I can see instructors wanting to use this for questions where the student gets a new variant after every submission, but then being confused about why the hint isn't ever showing up after a few submissions. Unfortunately I don't really have a solution at the moment.

eliotwrobson: I think revealing hints across variants has the potential to be useful, but I think this could also make it difficult to test for instructors without the ability to manually reset the number of attempts for a given user on the current assessment. I think functionality like that would be good to include, but I think requires some more care to make it user friendly.

nwalters512: @SethPoulsen I think that we can simply document that this is how this element works: if you want to use hints, they'll only work within a certain variant. Can you think of a compelling pedagogical reason to give someone hints but only have those hints apply to different variants and not just multiple attempts to the same one?

Having question state that depends on previous variants instead of the current one is a precedent I really, really want to avoid setting. It just makes everything more complex, including implementation and conceptual models for students and faculty.

SethPoulsen (approved):

docs/devElements.md
nwalters512: I don't have statistics on that 😛 

I understand that's a thing that people _may_ do, but what's the pedagogical thinking for wanting hints _and_ multiple variants? Is the student experience or instructor's ability to measure mastery negatively impacted by having hints with only a single variant?

SethPoulsen: Suppose the instructor wants to give a student 3 attempts at a question on a test, but they want it to be a different variant each time.

The hint is general enough that it applies to any of the variants of the same question, but they only want the student to see it on their last attempt. 

So, the instructor makes a hint with `show-after:3`, so that the students will get the hint for their last attempt. But to the instructor's surprise, the students _never_ see the hint because they only ever do 1 submission for each variant.

nwalters512: Right, I'm not debating the fact that an instructor may want that; I'm trying to understand *why* they want multiple variants. Is it just personal preference? Or is there a specific reason they need that? Put differently, if I were to tell an instructor "sorry, hints can't span multiple variants", would they come back and tell me "I can't use a single variant because XYZ", or would they just happily switch to a single variant?

SethPoulsen: I think people like using multiple variants in a situation like that because the student can be shown the correct answer and then learn from their mistake. If you do a single variant, it is harder for the student to learn from their mistake because you can't show them the correct answer.

eliotwrobson: > If you do a single variant, it is harder for the student to learn from their mistake because you can't show them the correct answer.

This is really the core need this element is addressing and also why I'm skeptical of revealing hints across variants. There are questions where multiple variants are hard to generate or not appropriate, and the progressive hints is a good way of pushing students in the right direction without outright revealing the answer. If you're showing a new variant after each submission (possibly with a correct answer), this gives implicit feedback after each submission in a way that's not possible with single variant questions.

mwest1066: I think that a common use case for hints would be with questions that generate new variants on every submission. I think instructors would naturally want to give a hint after (say) the third variant where the student got it wrong, to nudge them towards the correct method.

I'm not saying anything here about whether this is possible (or even should be possible) within an element --- I'm only saying what I think many instructors will want from a pedagogy point of view.

SethPoulsen: > There are questions where multiple variants are hard to generate or not appropriate, and the progressive hints is a good way of pushing students in the right direction without outright revealing the answer.

Right, I'm totally with you here. I'm just saying that by the very fact that this element exists, people are going to try to use it for questions that give a new variant every submission (which is a totally reasonable thing to do), and then they're going to be frustrated when it doesn't work as expected. 

Maybe the way to make this shipp-able is just to be very clear about its limitations in the documentation. For example, change the summary of the element to include the fact that it won't work across variants, all the places in the documention that say 'the current question' should instead say 'the current variant', and the documentation for the `show-after` attribute should expclicitly state that `show-after` should only be non-zero with questions that have `singleVariant: true`.

Still, I'm sure we'll have feature requests from people who want it to work across variants.

mwest1066: Yeah, I think this is going to be a support headache if it doesn't work as expected across variants.

mwest1066: > Having question state that depends on previous variants instead of the current one is a precedent I really, really want to avoid setting. It just makes everything more complex, including implementation and conceptual models for students and faculty.

I get this (oh so much) but I think people's mental model is such that they will want this for hints. I think we should think hard about how to provide cross-variant functionality here, either by pulling hints functionality out into PL core, or by providing more data to elements so they can implement it.

eliotwrobson: Going to go ahead and update the documentation as @SethPoulsen suggested. I think it's worth making a separate issue / discussion for how any type of cross-variant functionality might work. It should be easy enough to update this element to take advantage of that if or when those features get implemented.

eliotwrobson: Resolving this discussion and created an issue for if/how we want to support this type of functionality in the future: https://github.com/PrairieLearn/PrairieLearn/issues/6465

eliotwrobson:
Are there any more items that need to be handled for this to be merged? All the discussions have been resolved.

SethPoulsen:
I don't know if @mwest1066's concerns about supporting this without cross-variant functionality were ever fully resolved.

elements/pl-hidden-hints/pl-hidden-hints.py
```diff
@@ -0,0 +1,77 @@
+import chevron
+import lxml
+import prairielearn as pl
+from typing import TypedDict, Tuple, List
+
+
+class HintsDict(TypedDict):
+    """A class with type signatures for the partial scores dict"""
+    html_hints: List[Tuple[int, str]]
```
nwalters512, line 9: Can you confirm that this should all work with Python 3.8+? I remember in the past we ran into issues adding Python typings, but I don't recall what version we had deployed in production at the time. Presumably pyright would yell at us if we use things that aren't available in the version declared in `pyrightconfig.json`?

elements/pl-hidden-hints/pl-hidden-hints.py
```diff
@@ -0,0 +1,77 @@
...
+
+            # Default show-after to -1 to automatically show hint (closed) at start
+            priority = pl.get_integer_attrib(child, 'show-after', -1)
+            html_hints.append((priority, pl.inner_html(child)))
+        elif child.tag is lxml.etree.Comment:
+            continue
+        else:
+            raise Exception(f"Tags inside of pl-hidden-hints must be pl-hint, not '{child.tag}'.")
+
+    hints_dict['html_hints'] = html_hints
```
nwalters512, line 45: Is it necessary to store anything in `params` at all? I'd almost expect this element to be completely stateless - that is, it looks like this element will always render deterministically based on just the question HTML and `num_valid_submissions`.

The advantage to always treating the HTML as the source of truth is that if a hint is adjusted by an instructor, students will be able to see it the next time they reload the page; they won't have to wait until they generate a new variant (if they can at all).

docs/elements.md
```diff
@@ -1891,6 +1891,61 @@ element contents only in a specific panel.
 
 ---
 
+### `pl-hidden-hints` element
+
+Display progressive hints that become accessible as the number of student submissions increases for the current variant.
+Hints are only open on page load when they are first revealed (when first reaching the desired submission count).
+Otherwise hints start closed and must be opened by the user. The submission counter is reset when new variants are
+generated. In particular, **this element does not reveal new hints across variants.**
```
nwalters512, line 1899:
```suggestion
generated. Note that **this element does not reveal new hints across variants.**
```

nwalters512, line 1899: I haven't fully thought this through, but I imagine that hints work best when there's a penalty for retrying; otherwise a student could just submit the same wrong answer a bunch of times to reveal all the hints. Is that how you use it in practice? Do you think it's worth mentioning that in the docs here?

elements/pl-hidden-hints/pl-hidden-hints.py
eliotwrobson: Pyright will indeed yell at you for using 3.10 constructs since the version is locked at 3.8 in the config. Will double check, but the code we deploy from our repo uses type hints in the same style and we haven't had any issues.

docs/elements.md
```diff
@@ -1891,6 +1891,61 @@ element contents only in a specific panel.
...
+
+#### Sample element
+
+```html
+<pl-hidden-hints>
+  <pl-hint> This is a hint that will be accessible immediately. </pl-hint>
+
+  <pl-hint show-after="3">
```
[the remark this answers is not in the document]
eliotwrobson: That's a little more explicit, will change.

elements/pl-hidden-hints/pl-hidden-hints.py
eliotwrobson: I think doing this made more sense for our internal version, but you're right that it will cause less confusion if we recompute this on the fly (using HTML as groundtruth). Will make those changes.

docs/elements.md
eliotwrobson: I think that it's worth mentioning. In our use case, the student is allowed full credit on all attempts, but there's a 1min delay between submissions. Since the hints are hidden, a student would have to sit there for a few minutes on each question to see if any were going to be revealed, and we've found this is enough of a deterrent to prevent the situation you're describing.

elements/pl-hidden-hints/pl-hidden-hints.py
eliotwrobson: Just to double check, copied it over to our course repo and our tooling had no issues. Should be good to go.

docs/elements.md
eliotwrobson: Added some of the discussion from here and addressed all other comments @nwalters512.

PR #6338, opened 2022-09-16
Improve workspace host error handling

nwalters512:
This PR makes a few improvements to workspace host error handling based on the errors we've seen in Sentry:

- Instead of writing the graded files archive to disk, we now directly pipe it to the response. This incidentally fixes an error where we were a) not awaiting the `fsPromises.unlink` return value and b) sometimes passing `undefined` to `fsPromises.unlink`
- The return value of `_getWorkspaceSettingsAsync` was previously unawaited; now it is.
- We weren't using the `new` keyword with `Stream.Transform`.

I also wired up Sentry request tracking for workspace host commands.

### DIFF
--- a/workspace_host/interface.js
+++ b/workspace_host/interface.js
@@ -2,7 +2,6 @@ const ERR = require('async-stacktrace');
 const _ = require('lodash');
 const util = require('util');
 const express = require('express');
-const app = express();
 const http = require('http');
 const request = require('request');
 const path = require('path');
@@ -67,6 +66,8 @@ setInterval(() => {
 const bodyParser = require('body-parser');
 const docker = new Docker();
 
+const app = express();
+app.use(Sentry.Handlers.requestHandler());
 app.use(bodyParser.urlencoded({ extended: false }));
 app.use(bodyParser.json());
 
@@ -97,7 +98,7 @@ app.get(
     } else {
       res.status(200);
     }
-    res.jsonp({
+    res.json({
       docker: containers,
       postgres: db_status,
     });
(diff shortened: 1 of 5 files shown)


workspace_host/interface.js
```diff
@@ -97,7 +98,7 @@ app.get(
     } else {
       res.status(200);
     }
-    res.jsonp({
+    res.json({
```
nwalters512, line 101: AFAICT there's no legitimate reason to be using `jsonp` here, as this is strictly server-to-server communication.

workspace_host/interface.js
```diff
@@ -392,7 +395,7 @@ async function pushContainerContentsToS3(workspace) {
 
   const workspacePath = path.join(workspacePrefix, `workspace-${workspace.launch_uuid}`);
   const s3Path = `workspace-${workspace.id}-${workspace.version}/current`;
-  const settings = _getWorkspaceSettingsAsync(workspace.id);
+  const settings = await _getWorkspaceSettingsAsync(workspace.id);
```
nwalters512, line 398: I briefly enabled TypeScript error checking in this file, which is how I discovered this. There are too many other errors to leave it enabled permanently (mostly things with `util.callbackify`/`util.promisify`), but this was an easy one to fix.

workspace_host/interface.js
```diff
@@ -1387,89 +1391,39 @@ function resetSequence(workspace_id, res) {
...
+      archive.file(filePath, { name: file });
+      debug(`Sending ${file}`);
+    } catch (err) {
+      logger.warn(`Graded file ${file} does not exist.`);
+      continue;
     }
-  );
+  }
+
+  await archive.finalize();
```
mwest1066, line 1428: Wow, that's a lot less code! 🎉

mwest1066 (approved):
Thanks for starting to bring some sanity to the code in `interface.js`!

PR #6348, opened 2022-09-19
fix: pl-dropdown proper use of inline tags

### LINKED ISSUE #5898: pl-dropdown does not inline properly inside a paragraph
Maybe this is a dupe: I remember finding it a year ago, but not sure I ever filed it...

Code A: a `pl-downdown` in a paragraph:
```
<p>
    This set forms a <pl-dropdown sort="random" answers-name="figure_b">
        <pl-answer correct="false">Plane</pl-answer>
        ...
        <pl-answer correct="false">Line</pl-answer>
    </pl-dropdown> in space.
</p>
```
This has some forced newlines that do not appear in the above source:

- - - -

Screenshot from 2022-06-11 14-15-02

- - - -

Code B: same `pl-dropdown` in NOT in a paragraph:
```
    This set forms a <pl-dropdown sort="random" answers-name="figure_b">
        <pl-answer correct="false">Plane</pl-answer>
        ...
        <pl-answer correct="false">Line</pl-answer>
    </pl-dropdown> in space.
```
This one looks good:

- - - -

Screenshot from 2022-06-11 14-15-22

- - - -

jonatanschroeder:
Fixes #5898. Dropdown was using div instead of span for an object intended to be inline. Also removed the dropdown id for parts that didn't use them.

### DIFF
--- a/elements/pl-dropdown/pl-dropdown.mustache
+++ b/elements/pl-dropdown/pl-dropdown.mustache
@@ -40,9 +40,9 @@
               });
             </script>
             <span class="badge text-danger badge-invalid"><i class="fa fa-exclamation-triangle" aria-hidden="true"></i> Invalid</span>
-            <button id="pl-dropdown-popover-button-{{uuid}}" class="btn btn-sm btn-secondary small border"
+            <button id="pl-dropdown-popover-button-{{uuid}}" class="btn btn-sm btn-secondary small border ml-1"
                     data-placement="auto" data-trigger="focus" data-toggle="popover"
-                    data-html="true" style="margin-left: 5px" title="Format Error" data-container="body"
+                    data-html="true" title="Format Error" data-container="body"
                     data-content="{{parse-error}}">
               Why <i class="fa fa-question-circle" aria-hidden="true"></i>
             </button>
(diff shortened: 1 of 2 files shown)


nwalters512 (approved):
Thinking through this out loud: I originally had backwards-compatibility concerns (what if someone was using this as a standalone element outside of a `<p>` tag, would this cause issues). But this should be safe because we're still using `display: inline-block`, right?

I guess the root cause is that `<p>` cannot contain `<div>` tags, and so the parser terminates the opening `<p>` when it encounters the div?

jonatanschroeder:
> Thinking through this out loud: I originally had backwards-compatibility concerns (what if someone was using this as a standalone element outside of a `<p>` tag, would this cause issues). But this should be safe because we're still using `display: inline-block`, right?
> 
> I guess the root cause is that `<p>` cannot contain `<div>` tags, and so the parser terminates the opening `<p>` when it encounters the div?

Yes, that is correct. I checked a browser's DOM inspector with a question using `div` inside `p` and it showed that the `p` block was implicitly closed by the browser before the `div`. Just replacing a `div` with a `span` fixed the issue, and from a display perspective, the `inline-block` style from the class overlaps the typically difference between `div` and `span`. 

`<rant>`The misuse of `p` tags is actually a pet-peeve of mine, particularly when someone uses `<p></p>` on its own (as was originally done in this example question) as a way to introduce margin between paragraphs instead of using `p` tags for what they were actually intended for. And I'm even ignoring the accessibility aspect here.`</rant>`

PR #6352, opened 2022-09-20
Remove legacy mode for nodemon invocation

jonatanschroeder:
The legacy mode (`-L` flag) of `nodemon` is causing 100% CPU usage. Given it was (presumably) originally used to address cases of mounted drives inside containers, and those problems don't seem to be an issue anymore (tested on Windows at least), I propose removing the flag, which fixes the CPU usage issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ dev:
 start: start-support
 	@node server.js
 start-nodemon: start-support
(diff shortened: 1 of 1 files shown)


nwalters512 (approved):
I tested this on macOS as well and it's working as expected!

PR #6356, opened 2022-09-22
Add X-PrairieLearn-Revision header to all responses

nwalters512:
This should help admins identify what revision of PrairieLearn is responding to each request. It'll be possible to reuse `getCurrentRevision` in other places, e.g. to add this information to the course instance admin page.

### DIFF
--- a/lib/revision.js
+++ b/lib/revision.js
@@ -1,32 +1,31 @@
 const execa = require('execa');
-const { createCache } = require('async-cache-dedupe');
 
-// We're using this module because it allows us to deduplicate concurrent requests
-// for the revision; we don't actually rely on it caching the result (we'll do that
-// ourselves, as the revision should never change for the life of a given process).
-const cache = createCache({
-  ttl: 0, // don't cache the result
-  storage: { type: 'memory' },
-});
+function memoize(fn) {
+  let hasResult = false;
+  let result = null;
+  let promise = null;
 
-cache.define('getCurrentRevision', async () => {
-  return (await execa('git', ['rev-parse', 'HEAD'])).stdout.trim();
-});
-
-let currentRevision;
-
-module.exports.getCurrentRevision = async function () {
-  // Note that we treat `null` and `undefined` differently here. Undefined
-  // means we haven't yet tried to retrieve the revision, whereas null means
-  // that we encountered an error while fetching the revision. This lets us
-  // avoid spamming subprocesses if e.g. git isn't installed.
-  if (currentRevision === undefined) {
-    try {
-      currentRevision = await cache.getCurrentRevision();
-    } catch {
-      currentRevision = null;
+  return async function () {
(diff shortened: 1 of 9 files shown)


lib/revision.js
```diff
@@ -0,0 +1,32 @@
+const execa = require('execa');
+const { createCache } = require('async-cache-dedupe');
+
+// We're using this module because it allows us to deduplicate concurrent requests
+// for the revision; we don't actually rely on it caching the result (we'll do that
+// ourselves, as the revision should never change for the life of a given process).
```
mwest1066, line 6: This all seems fine, but do we really need to pull in another dependency just for this? I think it would be simple to roll our own asynchronous zero-argument `memoize()`, which is what we want here? I remember that I also implemented one of these recently (can't find it now on my phone, might have been in PT). I initially used `async.memoize()` but then switched to a manual implementation. I didn't implement the optimization to ensure only one copy would ever be executed, but I would have like to have that.

I'm thinking we have a separate variable to store an in-progress promise if there is one?

Again, this all seems fine, but I'm pushing back against our tendency to have an ever-growing list of deps. I don't mind when the dependency is providing substantial functionality, but when it's only saving a couple of lines of code (and maybe not even that) then I'm increasingly thinking we should re-implement ourselves.

mwest1066 (approved):

lib/revision.js
nwalters512: That's a fair point! I replaced it with a custom `memoize` function in f0af70c. Let me know if that looks correct to you (the whole promise-caching thing can be a little error-prone).

trombonekenny: Do you even need the memoization? Could this be simplified to creating an empty variable, running some code once when the `require` is loaded that updates the variable, and have the exported route simply report the variable? I mean, I know that's what your code is doing but it seems like there's a layer of abstraction there that's not needed?

nwalters512: It's not just memoization, it's waiting for the result and deduplicating requests for the revision. If we just ran some code at the top level of the module when it's required, I suppose that'd solve duplication, but we'd still need the "get revision" function to be aware of when that work has finished and wait for it if it's not done yet. That'd end up looking roughly like what we have here anyways.

trombonekenny: Right. I'm not even sure we need the "be aware of when that work has finished and wait" part if we're tolerant to page loads not having the revision in them for that small window during startup when things are initializing. It would make the route middleware synchronous and simple without a few layers of async handling?

nwalters512: 0e3944d updates `lib/code-caller-container.js` to use the `getCurrentRevision` function, that that *needs* to have immediate access to the revision.

mwest1066: This `memoize()` implementation looks correct to me.

mwest1066 (approved):

PR #6357, opened 2022-09-22
Improve time limit sorting behaviour in assessment instances list

### LINKED ISSUE #6341: Suggestion: sorting by Remaining should distinguish Closed from Open
On the Students tab of an assessment (e.g. https://www.prairielearn.org/pl/course_instance/129465/instructor/assessment/2321131/instances), when you sort the table using the "Remaining" column header, "Open (no time limit)" and "Closed" have the same sort value: rows with either do not move relative to each other. While I don't have an opinion on which should be 'higher' than the other, I think that they should probably have distinct values, so that sorting by that column actually separates out open from closed.

jonatanschroeder:
Resolves #6341. Sorts closed assessments first (somewhat equivalent to expired/negative time limit), then assessments with a time limit, then assessments with no time limit (equivalent to infinite time limit).

### DIFF
--- a/pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
+++ b/pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
@@ -273,7 +273,7 @@ function detailsLinkSorter(valueA, valueB, rowA, rowB) {
 function timeRemainingLimitSorter(valueA, valueB, rowA, rowB) {
     // Closed assessments are listed first, followed by time limits
     // ascending, followed by open without a time limit
-    return (rowA.open - rowB.open) || ((valueA ?? Infinity) - (valueB ?? Infinity));
+    return (Number(rowA.open) - Number(rowB.open)) || ((valueA ?? Infinity) - (valueB ?? Infinity));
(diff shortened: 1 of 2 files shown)


pages/instructorAssessmentInstances/instructorAssessmentInstancesClientJS.ejs
```diff
@@ -270,6 +270,12 @@ function detailsLinkSorter(valueA, valueB, rowA, rowB) {
     return compare;
 }
 
+function timeRemainingLimitSorter(valueA, valueB, rowA, rowB) {
+    // Closed assessments are listed first, followed by time limits
+    // ascending, followed by open without a time limit
+    return (rowA.open - rowB.open) || ((valueA ?? Infinity) - (valueB ?? Infinity));
```
nwalters512, line 276: Subtracting booleans is neat, bordering on too clever 😅 this is a little more explicit:
```suggestion
    return (Number(rowA.open) - Number(rowB.open)) || ((valueA ?? Infinity) - (valueB ?? Infinity));
```

nwalters512 (approved):

PR #6358, opened 2022-09-22
Enable MathJax contextual menu

### LINKED ISSUE #6345: Feature request: hovering over LaTeX should display the source code
I'd like it to be easier for students to use LaTeX in their text answers. Often a question uses a symbol they don't know the TeX code for; it would be nice if they could just mouse over the symbol and see the code. It would be even nicer if there were also a quick way to _copy_ the code (e.g. by select-copy, right-click, etc). This feature could be toggleable if it is not universally desirable.

nwalters512:
Closes #6345!

### DIFF
--- a/pages/partials/mathjax.ejs
+++ b/pages/partials/mathjax.ejs
@@ -17,34 +17,34 @@
  <% } %>
 
  var MathJax = {
-     tex: {
-         inlineMath: [['$', '$'], ['\\(', '\\)']]
-     },
-     svg: {
-         fontCache: 'global'
-     },
-     loader: {
-         load: ['input/tex', outputComponent]
-     },
-     // Register a callback to be run when MathJax is loaded, use MathJax.config.onReady().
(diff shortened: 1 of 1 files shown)


mwest1066 (approved):

pages/partials/mathjax.ejs
```diff
@@ -1,50 +1,50 @@
...
- };
+var MathJax = {
+    tex: {
+        inlineMath: [['$', '$'], ['\\(', '\\)']]
+    },
+    svg: {
+        fontCache: 'global'
+    },
+    loader: {
+        load: ['input/tex', 'ui/menu', outputComponent]
```
nwalters512, line 27: This is the only meaningful change here; the rest of this diff is just fixing indentation.

mwest1066: Yep, I used the very helpful "no whitespace" option in GH when reviewing this.

PR #6359, opened 2022-09-22
Run git clean before syncing from disk after performing an edit

nwalters512:
This PR fixes the issues that @ffund reported on Slack this morning: https://prairielearn.slack.com/archives/C266KEH9A/p1663865172180029

For posterity, Fraida renamed an assessment from something like `foo/bar/infoAssessment.json` to `bar/baz/infoAssessment.json`. This left behind a `foo` directory that didn't contain any `infoAssessment.json` file in itself or its descendants, so that was flagged as a sync error.

After this PR, we'll now run `git clean -dfx` a second time after committing the edits to disk. This will ensure that no empty directories remain when we read in the course state from disk.

### DIFF
--- a/pages/instructorFileBrowser/instructorFileBrowser.js
+++ b/pages/instructorFileBrowser/instructorFileBrowser.js
@@ -424,11 +424,7 @@ router.get('/*', function (req, res, next) {
     (err) => {
       if (err) {
         if (err.code === 'ENOENT' && file_browser.paths.branch.length > 1) {
-          res.redirect(
-            `${res.locals.urlPrefix}/${res.locals.navPage}/file_view/${encodePath(
-              file_browser.paths.branch.slice(-2)[0].path
-            )}`
-          );
+          res.redirect(`${req.baseUrl}/${encodePath(file_browser.paths.branch.slice(-2)[0].path)}`);
           return;
         } else {
           return ERR(err, next);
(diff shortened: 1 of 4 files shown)


pages/instructorFileBrowser/instructorFileBrowser.js
```diff
@@ -424,11 +424,7 @@ router.get('/*', function (req, res, next) {
     (err) => {
       if (err) {
         if (err.code === 'ENOENT' && file_browser.paths.branch.length > 1) {
-          res.redirect(
-            `${res.locals.urlPrefix}/${res.locals.navPage}/file_view/${encodePath(
-              file_browser.paths.branch.slice(-2)[0].path
-            )}`
-          );
+          res.redirect(`${req.baseUrl}/${encodePath(file_browser.paths.branch.slice(-2)[0].path)}`);
```
nwalters512, line 427: The tests involving this code started failing once I made the change to run `git clean` again:

```
invalid input syntax for integer: "file_view"
```

I'm actually really glad they did, because I've been seeing a low volume of these errors on Sentry (e.g. https://sentry.io/organizations/prairielearn-inc/issues/3583272833/?project=6704263) but didn't actually know how to fix them!

Using `req.baseUrl` will be much less prone to failures compared to trying to manually construct the appropriate URL.

mwest1066: Nice!

mwest1066 (approved):

PR #6360, opened 2022-09-22
Use appropriate HTTP status codes for SQL exceptions

nwalters512:
This should clean up reports of 500 errors from Sentry and our load balancer logs, which should improve the signal-to-noise ratio there.

I don't *love* inspecting the text of error messages here, but it's the simplest option considering that the sprocs that are raising these exceptions are themselves called from inside *other* sprocs.

### DIFF
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -37,7 +37,7 @@ module.exports = {
     };
     sqldb.query(sql.check_belongs, params, (err, result) => {
       if (ERR(err, callback)) return;
-      if (result.rowCount !== 1) return callback(new Error('access denied'));
+      if (result.rowCount !== 1) return callback(error.make(403, 'access denied'));
       callback(null);
     });
   },
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -228,30 +228,18 @@ module.exports = {
     async.series(
       [
         async () => {
-          try {
-            if (requireOpen) {
-              await sqldb.callAsync('assessment_instances_ensure_open', [assessment_instance_id]);
-            }
+          if (requireOpen) {
+            await sqldb.callAsync('assessment_instances_ensure_open', [assessment_instance_id]);
+          }
 
-            if (close) {
-              // If we're supposed to close the assessment, do it *before* we
-              // we start grading. This avoids a race condition where the student
-              // makes an additional submission while grading is already in progress.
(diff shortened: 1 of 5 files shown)


jonatanschroeder:
> I don't _love_ inspecting the text of error messages here, but it's the simplest option considering that the sprocs that are raising these exceptions are themselves called from inside _other_ sprocs.

Suggestion (from https://www.postgresql.org/docs/current/plpgsql-errors-and-messages.html) add `DETAIL` to the `RAISE EXCEPTION` command like:

```sql
RAISE EXCEPTION 'assessment instance is not open: %', assessment_instance_id USING DETAIL '403';
```

Or something else in detail that helps determine the exception type (a JSON-formatted object?) then use `err?.detail` to determine the error code.

mwest1066 (approved):

mwest1066:
> Suggestion (from https://www.postgresql.org/docs/current/plpgsql-errors-and-messages.html) add `DETAIL` to the `RAISE EXCEPTION` command like

This is awesome! I think we should definitely do this.

nwalters512:
@jonatanschroeder great suggestion! I took a slightly different approach: we can specify a custom error code, which has pretty much exactly the semantics we want. So, we can raise an exception with `ERRCODE = 'ST404'`, and our error-handling middleware will detect that and extract a response code from it. Let me know if that implementation sounds reasonable to you and @mwest1066.

jonatanschroeder:
> @jonatanschroeder great suggestion! I took a slightly different approach: we can specify a custom error code, which has pretty much exactly the semantics we want. So, we can raise an exception with `ERRCODE = 'ST404'`, and our error-handling middleware will detect that and extract a response code from it. Let me know if that implementation sounds reasonable to you and @mwest1066.

I saw the errcode but thought it would be limited to SQLSTATE error codes. The documentation, though, mentions that your approach is valid, and it indeed looks better.

mwest1066:
> I saw the errcode but thought it would be limited to SQLSTATE error codes. The documentation, though, mentions that your approach is valid, and it indeed looks better.

I also initially thought that we wouldn't be allowed to use custom error codes, but since we are I think this is excellent! So much clearer to be able to throw a 404 directly from within SQL!

mwest1066 (approved):

PR #6362, opened 2022-09-23
Fix chunk generation when deleted course instances exist

### LINKED ISSUE #6361: Sync course failed when copying a question-- TypeError: Cannot read properties of undefined (reading 'assessments')
Hello, I got one issue from course ME330, Fall 2022. When I tried to copy one question to modified into a new question, the error message pops up: 
      **Sync course failed with this result:
      TypeError: Cannot read properties of undefined (reading 'assessments')
      Please go back and try again.**

The webpage is https://www.prairielearn.org/pl/course_instance/129462/instructor/edit_error/317783
image

nwalters512:
Closes #6361.

AFAICT this occurred because there were existing chunks in the DB for course instances that no longer existed. Those course instances would have appeared in `existingCourseChunks`, but not in `courseData`, hence the error that we've seen.

### DIFF
--- a/lib/chunks.js
+++ b/lib/chunks.js
@@ -494,10 +494,8 @@ module.exports.diffChunks = async ({ coursePath, courseId, courseData, changedFi
       );
       if (
         hasClientFilesCourseInstanceDirectory &&
-        (!existingCourseChunks.courseInstances[ciid] ||
-          !existingCourseChunks.courseInstances[ciid].clientFilesCourseInstance ||
-          (changedCourseChunks.courseInstances[ciid] &&
-            changedCourseChunks.courseInstances[ciid].clientFilesCourseInstance))
+        (!existingCourseChunks.courseInstances[ciid]?.clientFilesCourseInstance ||
+          changedCourseChunks.courseInstances[ciid]?.clientFilesCourseInstance)
       ) {
         updatedChunks.push({
           type: 'clientFilesCourseInstance',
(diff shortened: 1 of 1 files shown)


lib/chunks.js
```diff
@@ -494,10 +494,8 @@ module.exports.diffChunks = async ({ coursePath, courseId, courseData, changedFi
       );
       if (
         hasClientFilesCourseInstanceDirectory &&
-        (!existingCourseChunks.courseInstances[ciid] ||
-          !existingCourseChunks.courseInstances[ciid].clientFilesCourseInstance ||
-          (changedCourseChunks.courseInstances[ciid] &&
-            changedCourseChunks.courseInstances[ciid].clientFilesCourseInstance))
+        (!existingCourseChunks.courseInstances[ciid]?.clientFilesCourseInstance ||
+          changedCourseChunks.courseInstances[ciid]?.clientFilesCourseInstance)
```
mwest1066, line 498: These are just cleanups, right? Not connected to the bug itself?

lib/chunks.js
```diff
@@ -551,7 +545,7 @@ module.exports.diffChunks = async ({ coursePath, courseId, courseData, changedFi
 
       await Promise.all(
         [...courseInstanceInfo.assessments].map(async (tid) => {
-          const assessmentExists = !!courseData.courseInstances[ciid].assessments[tid];
+          const assessmentExists = !!courseData.courseInstances[ciid]?.assessments[tid];
```
mwest1066, line 548: Thanks for the explanation in the PR -- this makes sense to me.

mwest1066 (approved):

lib/chunks.js
nwalters512: Correct - just cleanup!

mwest1066: Thanks!

PR #6364, opened 2022-09-23
Remove remaining instances of self-closing elements

### LINKED ISSUE #1244: self-closing PrarieLearn elements in the new renderer
There is a gap / inconsistency in the description of how self-closing PrairieLearn elements will be handled in the documentation going forward with new renderer compared with the examples in the description of how to write a question.html file.  In the question.html file description examples, self-closing pl-elements , e.g. <pl-number-input answers_name="F" comparison="sigfig" digits="2" />, are used, and yet further down the page it seems to indicate that no self-closing PrairieLearn elements will be allowed by the new renderer.  It would be helpful if all examples given were compliant with the new renderer, and if some self-closing PrairieLearn elements are in fact allowed, the wording changed to reflect that, and a list or tagging of allowed self-closing PrairieLearn elements were added to the HTML Element list page.

nwalters512:
Closes #1244.

### DIFF
--- a/docs/question.md
+++ b/docs/question.md
@@ -116,7 +116,13 @@ The `question.html` is a template used to render the question to the student. A 
 </pl-question-panel>
 
 <p>
-  <pl-number-input answers_name="F" comparison="sigfig" digits="2" label="$F =$" suffix="$\rm m/s^2$"></pl-number-input>
+  <pl-number-input
+    answers_name="F"
+    comparison="sigfig"
+    digits="2"
+    label="$F =$"
+    suffix="$\rm m/s^2$"
+  ></pl-number-input>
 </p>
 ```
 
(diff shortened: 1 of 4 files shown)


docs/question.md
```diff
@@ -117,7 +117,7 @@ The `question.html` is a template used to render the question to the student. A
 
 <p>
   $F = $
-  <pl-number-input answers_name="F" comparison="sigfig" digits="2" /> $\rm m/s^2$
+  <pl-number-input answers_name="F" comparison="sigfig" digits="2"></pl-number-input> $\rm m/s^2$
```
jonatanschroeder, line 120:
```suggestion
  <pl-number-input answers_name="F" comparison="sigfig" digits="2" label="$F =$" suffix="$\rm m/s^2$"></pl-number-input>
```

mwest1066 (approved):

docs/question.md
nwalters512: Good suggestion - done in 25e4af3, I also removed the now-extraneous `$F =$` from the previous line.

PR #6369, opened 2022-09-26
Add time limit expiry events to assessment instance log

### LINKED ISSUE #6368: Add assessment log entry when exam timer expires
At the moment PL shows the "end of exam" as the last time the student took an action (e.g., viewed a page). The assessment log does not show when the exam time expired, and thus it can look like a student got less than the full time for the exam. To fix this, we should add a log entry for timer expiration.

jonatanschroeder:
Resolves #6368. ~~Still pending is the discussion if a new score log should be created on instance creation (https://github.com/PrairieLearn/PrairieLearn/issues/6368#issuecomment-1258224793).~~


sprocs/assessment_instances_select_log.sql
```diff
@@ -342,6 +342,68 @@ BEGIN
...
+                    JOIN assessment_instances AS ai ON (ai.id = ai_id)
+                    LEFT JOIN users AS u ON (u.user_id = asl.auth_user_id)
+                WHERE
+                    asl.assessment_instance_id = ai_id
+                    AND asl.open
+                    AND asl.date_limit IS NOT NULL
+                    -- Only list as expired if it already happened
+                    AND asl.date_limit < CURRENT_TIMESTAMP
+                    -- Only list the expiry date if the assessment was not closed or extended after this time limit was set but before it expired
+                    AND NOT EXISTS (SELECT 1 FROM assessment_state_logs aslc WHERE aslc.date > asl.date AND aslc.date <= asl.date_limit)
```
nwalters512, line 374: Do we need a `WHERE` in the `SELECT 1 FROM...` query here? Ditto for the other part of the `UNION` below.

sprocs/assessment_instances_select_log.sql
```diff
@@ -342,6 +342,68 @@ BEGIN
...
             )
             UNION
+            (
+                SELECT
+                    7.5 AS event_order,
+                    'Time limit expiry'::TEXT AS event_name,
+                    'gray3'::TEXT AS event_color,
+                    asl.date_limit AS date,
+                    u.user_id AS auth_user_id,
+                    u.uid AS auth_user_uid,
```
nwalters512, line 352: What does it mean for a time limit expiry to have an associated user?

nwalters512 (commented):
Can you explain why two separate queries are necessary? I would expect `assessment_state_logs` to contain all state transitions, making the second one unnecessary. I'd appreciate inline comments explaining this.

jonatanschroeder:
> Can you explain why two separate queries are necessary? I would expect `assessment_state_logs` to contain all state transitions, making the second one unnecessary. I'd appreciate inline comments explaining this.

`assessment_state_logs` is not created when the assessment instance is created, so the time limit set upon instance creation wouldn't be selected in the first query. Unless I change the code to create such a log, which I just implemented and I'm testing as we speak (see https://github.com/PrairieLearn/PrairieLearn/issues/6368#issuecomment-1258224793). In that case the second query would become unnecessary, unless we want to keep it for logs of assessments created before this is deployed.

sprocs/assessment_instances_select_log.sql
jonatanschroeder: Yes. For `aslc.date <= asl.date_limit`, if the assessment is closed or an instructor extends the time limit after the timer expires, we want to see the expiration event in the log. For `aslc.date > asl.date`, if the time limit was modified before this event, it has no impact on the current one.

sprocs/assessment_instances_select_log.sql
jonatanschroeder: The logs require a `auth_user_id` and `auth_user_uid` columns for all queries, in this case the query is set to whoever set the time limit.

sprocs/assessment_instances_select_log.sql
nwalters512: Sorry, I should have been more clear with my question (I agree that the existing `WHERE` clause is necessary): do we need something like `WHERE aslc.assessment_instance_id = ai_id`?

jonatanschroeder: Ah, right missed that. Will fix in a minute. Dangers of testing locally...

mwest1066 (approved):

PR #6370, opened 2022-09-26
Improve syntax highlighting for instructor file browser

nwalters512:
This makes a few improvements to syntax highlighting in the instructor file browser:

- When checking if a file is binary and determining its MIME type, use methods that read a limited number of bytes directly from disk. This limits the memory consumption and IO usage for especially large files.
- Treat any text file larger than 1MB as a binary file to avoid reading it into memory or trying to highlight it.
- Instead of using `highlightAuto` to guess the language, we now first guess based on file extension. If that guess fails, we'll use `highlightAuto` to try to guess based on just the first 2000 bytes of the file. This improves the speed of guessing and ensures we don't run out of memory if we try to highlight a particularly large file.

### DIFF
--- a/pages/instructorFileBrowser/instructorFileBrowser.js
+++ b/pages/instructorFileBrowser/instructorFileBrowser.js
@@ -297,47 +297,65 @@ function browseDirectory(file_browser, callback) {
 function browseFile(file_browser, callback) {
   async.waterfall(
     [
-      (callback) => {
-        fs.readFile(file_browser.paths.workingPath, (err, contents) => {
-          if (ERR(err, callback)) return;
-          callback(null, contents);
-        });
-      },
-      (contents, callback) => {
-        isBinaryFile(contents).then(
-          (result) => {
-            debug(`isBinaryFile: ${result}`);
-            file_browser.isBinary = result;
-            if (result) {
-              util.callbackify(async () => {
-                const type = await FileType.fromBuffer(contents);
-                if (type) {
-                  debug(`file type:\n ext = ${type.ext}\n mime = ${type.mime}`);
-                  if (type.mime.startsWith('image')) {
-                    file_browser.isImage = true;
-                  } else if (type.mime === 'application/pdf') {
-                    file_browser.isPDF = true;
-                  }
-                } else {
-                  debug(`could not get file type`);
-                }
-              })(callback);
-            } else {
-              debug(`found a text file`);
-              file_browser.isText = true;
-              file_browser.contents = hljs.highlightAuto(contents.toString('utf8')).value;
-              callback(null);
+      async () => {
+        // If the file is larger that 10MB, don't attempt to read it; treat
+        // it like an opaque binary file.
+        const { size } = await fs.stat(file_browser.paths.workingPath);
+        if (size > 10 * 1024 * 1024) {
+          file_browser.isBinary = true;
+          return;
+        }
+
+        const contents = await fs.readFile(file_browser.paths.workingPath);
+
+        const result = await isBinaryFile(contents);
+        debug(`isBinaryFile: ${result}`);
+        file_browser.isBinary = result;
+        if (result) {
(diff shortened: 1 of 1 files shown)


pages/instructorFileBrowser/instructorFileBrowser.js
```diff
@@ -297,47 +297,59 @@ function browseDirectory(file_browser, callback) {
...
-          (err) => {
-            if (ERR(err, callback)) return;
-            callback(null); // should never get here
           }
-        );
+        } else {
+          // This is probably a text file. If it's is larger that 10MB, don't
+          // attempt to read it; treat it like an opaque binary file.
+          const { size } = await fs.stat(file_browser.paths.workingPath);
+          if (size > 10 * 1024 * 1024) {
```
nwalters512, line 316: I picked this completely arbitrarily.

mwest1066: I'd also be fine with something lower. Maybe 1 MB? That's still a huge text file.

pages/instructorFileBrowser/instructorFileBrowser.js
```diff
@@ -297,47 +297,59 @@ function browseDirectory(file_browser, callback) {
...
+
+          file_browser.isText = true;
+
+          const contents = await fs.readFile(file_browser.paths.workingPath);
+          const stringifiedContents = contents.toString('utf8');
+
+          // Try to guess the language from the file extension. This takes
+          // advantage of the fact that Highlight.js includes common file extensions
+          // as aliases for each supported language, and `getLanguage()` allows
+          // us to look up a language by its alias.
```
mwest1066, line 329: Nice!

mwest1066 (approved):

pages/instructorFileBrowser/instructorFileBrowser.js
nwalters512: Good call - I originally had this as the first step of this function before we even know if we're dealing with a binary file or not, but it makes perfect sense to lower this now that we're reasonably confident that this is a text file!

nwalters512: Lowered to 1MB in aad67da.

pages/instructorFileBrowser/instructorFileBrowser.js
```diff
@@ -297,47 +297,59 @@ function browseDirectory(file_browser, callback) {
...
+          // https://highlightjs.readthedocs.io/en/latest/supported-languages.html
+          let language = null;
+          const extension = path.extname(file_browser.paths.workingPath).substring(1);
+          if (!['ml', 'ls'].includes(extension) && hljs.getLanguage(extension)) {
+            language = extension;
+          } else {
+            const result = hljs.highlightAuto(stringifiedContents.slice(0, 2000));
+            language = result.language;
+          }
+          file_browser.contents = hljs.highlight(stringifiedContents, { language }).value;
```
jonatanschroeder, line 346: This is causing an error for empty files, since in this case hljs sets the language to undefined, which is not accepted as input to highlight. Given that we're calling highlight on the whole string anyway, doesn't it make more sense to just call `highlightAuto` on the full string as well?

nwalters512: That's exactly the original problem we're trying to avoid: the implementation for `highlightAuto` is roughly "pass the string through every single grammar we have, compute a relevance score for each one, and pick the best one". For large-ish files, that results in a ton of synchronous compute and actually makes the node process run out of memory since it's buffering the results for each of the 300+ grammars in memory. That's why we only run `highlightAuto` on the first 2000 bytes.

Great find with an `undefined` language though. I think the correct fix there is to just always fall back to `plaintext` if we can't auto-detect a language

jonatanschroeder: Fixed in #6270, commit 21ea4ab.

PR #6371, opened 2022-09-27
Remove aws-config.json support

nwalters512:
This is a small bit of low-impact cleanup.

If someone wants to use AWS credentials when running locally, they should use the standard `AWS_*` environment variables (https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html, e.g. `AWS_PROFILE`). This is the standard way that AWS clients work; there's no need for our own idiosyncratic layer on top of that.

If we find ourselves needed to do more than configure credentials, we should add support for e.g. and `aws.config` property in our standard `config.json` file.

### DIFF
--- a/grader_host/lib/config.js
+++ b/grader_host/lib/config.js
@@ -31,47 +31,19 @@ config.loadConfig = function (callback) {
         });
       },
       (callback) => {
-        // Try to grab AWS config from a file; assume Metadata Service will
-        // provide credentials if the file is missing
-        fs.readFile('./aws-config.json', (err, awsConfig) => {
-          if (err) {
-            // we don't have the AWS config file, assume we are inside
-            // EC2 and that we can get all config info from the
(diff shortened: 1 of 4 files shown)


grader_host/lib/config.js
```diff
@@ -31,47 +31,18 @@ config.loadConfig = function (callback) {
...
-            );
-            MetadataService.request(
-              '/latest/dynamic/instance-identity/document',
-              (err, document) => {
-                if (ERR(err, callback)) return;
-                try {
-                  const data = JSON.parse(document);
-                  logger.info('instance-identity', data);
-                  AWS.config.update({ region: data.region });
-                  exportedConfig.instanceIdentity = data;
```
nwalters512, line 52: This property was unused.

mwest1066 (approved):

PR #6373, opened 2022-09-27
Improve creation of Docker images

nwalters512:
A few minor optimizations to the creation of our Docker image:

- We now run `nvm cache clear` to remove cached Node downloads
- `psutil` is now listed in `python-requirements.txt` (this should be slightly faster, and will allow Dependabot to update it)
- `procps-ng` is installed in `plbase` instead of in the `executor` image (this should be faster, as we don't have to download the RPM repository metadata again)

### DIFF
--- a/images/plbase/plbase-install.sh
+++ b/images/plbase/plbase-install.sh
@@ -10,31 +10,32 @@ amazon-linux-extras install -y \
     redis4.0
 
 # Notes:
-# - `libjpeg-devel` is needed by the Pillow package
 # - `gcc-c++` is needed to build the native bindings in `packages/bind-mount`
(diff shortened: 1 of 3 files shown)


images/plbase/plbase-install.sh
```diff
@@ -10,31 +10,32 @@ amazon-linux-extras install -y \
...
+    ImageMagick \
+    libjpeg-devel \
     lsof \
+    make \
+    man \
     openssl \
-    dos2unix \
+    postgresql-contrib \
+    postgresql-server \
+    procps-ng \
```
nwalters512, line 33: This is the only addition; the rest of the entries in this list were just alphabetized.

mwest1066 (approved):

PR #6379, opened 2022-09-29
Report Postgres idle client errors to Sentry

nwalters512:
This should give us slightly more visibility into what's happening when servers die.

@mwest1066 I'm wondering if there's some way we can get more useful information for `idle-in-transaction timeout` errors - maybe we should have clients store the last query that was executed with them and include that when errors are reported? It wouldn't be perfect, but it'd at least let us have something to grep for.


server.js
```diff
@@ -1865,8 +1865,11 @@ if (config.startServer) {
         };
         function idleErrorHandler(err) {
           logger.error('idle client error', err);
-          // https://github.com/PrairieLearn/PrairieLearn/issues/2396
-          process.exit(1);
+          Sentry.captureException(err);
+          Sentry.close().finally(() => {
+            // https://github.com/PrairieLearn/PrairieLearn/issues/2396
```
mwest1066, line 1870: Looks like this issue is no longer relevant?

mwest1066 (approved):

nwalters512:
@mwest1066 I believe the idle-in-transaction timeout errors are emitted per-client. In other words, they hit this:

https://github.com/PrairieLearn/PrairieLearn/blob/723b0783c1fe955134350e221891c3cc0934aa3a/prairielib/lib/sql-db.js#L152-L154

not this:

https://github.com/PrairieLearn/PrairieLearn/blob/723b0783c1fe955134350e221891c3cc0934aa3a/prairielib/lib/sql-db.js#L148-L150

So we should always know exactly which client had the error, and thus which query (assuming we store a mapping of clients to queries).

mwest1066:
> So we should always know exactly which client had the error, and thus which query (assuming we store a mapping of clients to queries).

Oh nice! This is awesome.

mwest1066:
Before I approve again, did you want to fix that one outdated comment?

And did you want to implement the last-query-saving as part of this? I don't think it's critical right now, but going forward it would be massively more helpful to know which query triggered this.

server.js
nwalters512: True! Let me remove this.

nwalters512:
> And did you want to implement the last-query-saving as part of this? I don't think it's critical right now, but going forward it would be massively more helpful to know which query triggered this.

Not yet; I'm going to open another PR for that shortly. Want to take a final look at this one?

mwest1066 (approved):

mwest1066:
> Not yet; I'm going to open another PR for that shortly. Want to take a final look at this one?

👍 approved again!

PR #6381, opened 2022-09-30
Implement graceful startup and shutdown

nwalters512:
This PR implements graceful startup and shutdown. There are two goals driving this:

- We want to be able to confidently deploy at any time. Currently, `SIGTERM` will immediately shut down the process, losing any in-progress requests, cron jobs, server jobs, etc.
- We want to be able to enable auto-scaling for the PrairieLearn server fleet.

Both of these are essentially the same problem: being able to destroy and recreate servers at any time without disrupting end users.

This PR implements graceful shutdown in response to both `SIGTERM` (for the case of host deploys) and the instance entering the `Terminating:Wait` state (for the case of instances deployed to ASGs). In each case, we'll stop accepting new connections, force all web socket connections to reconnect to a new host, and wait for all server jobs and cron jobs to complete.

In the case of `SIGTERM`, we'll exit cleanly with a zero exit code.

In the case of a deployment to an ASG, we'll use the `CompleteLifecycleActionCommand` API to indicate to the ASG that we can safely be terminated.

For good measure, we also introduce support for a launching lifecycle action that will run at boot to take us out of the `Pending:Wait` state.


server.js
```diff
@@ -2056,14 +2047,57 @@ if (config.startServer) {
...
+          .handleAndCompleteInstanceTermination(prepareForTermination)
+          .catch((err) => {
+            logger.error('Error handling instance termination', err);
+            Sentry.captureException(err);
+          })
+          .finally(terminate);
+
+        // Also listen for and handle SIGTERM, which can be sent to gracefully
+        // shutdown a process.
+        process.on('SIGTERM', () => prepareForTermination().finally(terminate));
```
nwalters512, line 2100: In theory, this will allow us to handle `systemctl stop`/`systemctl restart`, though we should validate that we've correctly configured it to send `SIGTERM`.

nwalters512:
@mwest1066 I still need to test this in staging, but the general concepts here should be ready for review when you have a chance!

server.js
```diff
@@ -2056,14 +2047,57 @@ if (config.startServer) {
...
+          try {
+            await util.promisify(server.close).call(server);
+
+            await socketServer.close();
+
+            // Wait for all currently-executing cron jobs to finish.
+            await cron.stop();
+
+            // Wait for all currently-executing server jobs to finish.
+            await serverJobs.stop();
```
nwalters512, line 2062: Did I miss anything else that we should be gracefully stopping? Does this order look correct? As I'm typing this, I'm thinking that maybe we should shut down server jobs before we stop the server and websockets?

nwalters512: Ah, external grader results queue processing. I need to implement that.

nwalters512: I'm working on this over in https://github.com/PrairieLearn/PrairieLearn/pull/6497.

lib/server-jobs.js
```diff
@@ -208,10 +208,21 @@ module.exports.init = function () {
...
+module.exports.stop = async function () {
+  await new Promise((resolve) => {
+    // Poll until all jobs are finished.
+    const pollIntervalId = setInterval(() => {
+      if (Object.keys(module.exports.liveJobs).length === 0) {
+        clearInterval(pollIntervalId);
+        resolve();
+      }
+    }, 100);
+  });
```
mwest1066, line 220: Similar to the above, can we do this with a while loop and sleeps?

server.js
```diff
@@ -2056,14 +2047,57 @@ if (config.startServer) {
...
+        // to enter the `Terminating:Wait` state.
+        //
+        // If we're not running in EC2, this will be a no-op.
+        lifecycleHooks
+          .handleAndCompleteInstanceTermination(prepareForTermination)
+          .catch((err) => {
+            logger.error('Error handling instance termination', err);
+            Sentry.captureException(err);
+          })
+          .finally(terminate);
```
mwest1066, line 2096: Not that I particularly object to the promise syntax, but I'm wondering whether we should try and use async/await syntax in cases like this? Mainly just to standardize on a/a to avoid having multiple equivalent syntaxes. Of course I don't mind using promises explicitly where it's much simpler to do so.

server.js
```diff
@@ -2056,14 +2047,57 @@ if (config.startServer) {
...
+            process.exit(0);
+          }
+        };
+
+        // If we're running in EC2 auto scaling, this will wait for our instance
+        // to enter the `Terminating:Wait` state.
+        //
+        // If we're not running in EC2, this will be a no-op.
+        lifecycleHooks
+          .handleAndCompleteInstanceTermination(prepareForTermination)
```
mwest1066, line 2091: If I'm understanding correctly, this will poll every 10 seconds forever? Should we instead set up an SNS notification or a lambda that will call a webhook or something like that?

mwest1066 (approved):

server.js
nwalters512: I thought about the webhook approach, but I opted to go the simpler route for now.

To think about what that might look like though: when we get notified about an instance being shut down, we'd somehow discover its internal hostname and post directly to port 80 on the host with some kind of secret?

Alternatively... I'm really curious if we could just rely on systemd SIGTERMing us when the instance is being shut down? Presumably shutdown is graceful and not just "whoops we killed power to your box".

mwest1066: > I thought about the webhook approach, but I opted to go the simpler route for now.

Understandable. I just can't help but feel like polling is horrible here 😄

> To think about what that might look like though: when we get notified about an instance being shut down, we'd somehow discover its internal hostname and post directly to port 80 on the host with some kind of secret?

Yep, the Event Bridge notification that's sent to the lambda contains the instance id, and from that we can look up the instance public hostname from the instance metadata: https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-event-reference.html

> Alternatively... I'm really curious if we could just rely on systemd SIGTERMing us when the instance is being shut down? Presumably shutdown is graceful and not just "whoops we killed power to your box".

Yep, it's graceful, but [the docs say](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Stop_Start.html#what-happens-stop):
> If the instance OS does not shut down cleanly within a few minutes, a hard shutdown is performed.

I couldn't find anything that specifies what "a few minutes" means.

server.js
nwalters512: I added external grader results processing shutdown in c0dc2f9.

cron/index.js
```diff
@@ -143,16 +142,17 @@ module.exports = {
...
+      const intervalId = setInterval(() => {
+        if (_.isEmpty(jobTimeouts)) {
+          debug(`stop(): all jobs stopped`);
+          clearInterval(intervalId);
+          resolve(null);
+        } else {
+          debug(`stop(): waiting for ${_.size(jobTimeouts)} jobs to stop`);
+        }
+      }, 100);
+    });
```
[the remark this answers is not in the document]
nwalters512: Done in f0e1503.

lib/lifecycle-hooks.js
```diff
@@ -0,0 +1,112 @@
...
+  DescribeAutoScalingInstancesCommand,
+  CompleteLifecycleActionCommand,
+} = require('@aws-sdk/client-auto-scaling');
+
+const config = require('./config');
+const logger = require('./logger');
+
+function sleep(ms) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
```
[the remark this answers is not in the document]
nwalters512: Moved to `lib/sleep` in f0e1503!

lib/server-jobs.js
nwalters512: Yep, done in 120a0c9.

server.js
nwalters512: I typically use `.then()`/`.catch()`/`.finally()` only when I don't want to actually wait for something to complete before proceeding to the next statement. In this case, we need to keep executing so that we could attach the `SIGTERM` handler. The alternative here would be an IIFE, which IMO is worse:

```js
(async () => {
  try {
    await lifecycleHooks.handleAndCompleteInstanceTermination(prepareForTermination);
  } catch (err) {
    logger.error('Error handling instance termination', err);
    Sentry.captureException(err);
  }

  await terminate();
})();
```

server.js
nwalters512: Alright, relying on `SIGTERM` seems like a bad idea.

In the interest of landing something that we can iterate on, are you ok with sticking with polling for now? EventBridge + Lambda seems like an objectively better option, and it'll be important once we get large numbers of chunk servers in operation, but I don't think it's necessary at the moment.

server.js
mwest1066: Yeah, I agree that an IIFE is pretty nasty and promises are better here.

server.js
mwest1066: No worries, I'm happy to go with whatever you want here. I'm slightly nervous that we'll never implement the EventBridge code if we deploy with polling, because it will be a scary change to launch into prod. But I also agree that setting up EV+Lambda is a pain in the neck (although terraform should make it much better than when I last did this by hand).

server.js
mwest1066: (and thanks for the detailed explanation)

server.js
```diff
@@ -2022,22 +2020,15 @@ if (config.startServer) {
...
         await externalGraderResults.init();
       },
-      function (callback) {
-        cron.init(function (err) {
-          if (ERR(err, callback)) return;
-          callback(null);
-        });
-      },
+      async () => cron.init(),
+      async () => lifecycleHooks.completeInstanceLaunch(),
```
mwest1066, line 2031: Nice!

mwest1066 (approved):

lib/aws.js
```diff
@@ -32,6 +32,25 @@ module.exports.init = async function () {
...
+async function getInstanceMetadata() {
+  if (!config.runningInEc2) return null;
+
+  if (!cachedInstanceMetadata) {
+    const document = await metadataServiceRequest('/latest/dynamic/instance-identity/document');
+    cachedInstanceMetadata = JSON.parse(document);
+  }
+
+  return cachedInstanceMetadata;
+}
```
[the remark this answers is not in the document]
mwest1066: Ping in case you want to do this (no worries if not).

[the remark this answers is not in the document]
nwalters512: I'll skip this for now 👍

PR #6383, opened 2022-09-30
Initialize freeform code before taking requests

nwalters512:
Followup to #6335. I don't think we caught this at the time, but we were actually waiting until *after* we had started the server before we would init the freeform code (basically, constructing the element cache). We need that in place before we can serve requests.

This shouldn't have mattered in practice, as we'd need to pass a couple of health check intervals before we actually received traffic. However, this is still more correct.

### DIFF
--- a/server.js
+++ b/server.js
@@ -1948,9 +1948,26 @@ if (config.startServer) {
           callback(null);
         });
       },
+      async () => await freeformServer.init(),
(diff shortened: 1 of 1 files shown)


mwest1066 (approved):
Ugh, how did I miss that?! Good find.

PR #6384, opened 2022-09-30
Simplify instance IDs

nwalters512:
As discussed with @mwest1066, this PR eliminates the special-cased `instanceIdEc2Override` config value in favor of always uniquely identifying a server as a combination of EC2 instance ID and a port number. The driver here is to allow `config.instanceId` to unconditionally represent the actual EC2 instance ID.

As I was implementing this, I started wondering if perhaps we actually want two (maybe three) separate pieces of config:

- `instanceId`: Always read from the EC2 metadata service; not overridable.
- `serverId`: Conceptually represents a running instance of PrairieLearn. Defaults to `${instanceId}:${serverPort}`, but overridable by actually setting `serverId` in the config.

Rationale: it might be nice to be able to differentiate between instructor servers and chunk servers? Or maybe we just don't care about this.


mwest1066:
> As I was implementing this, I started wondering if perhaps we actually want two (maybe three) separate pieces of config:
> 
> * `instanceId`: Always read from the EC2 metadata service; not overridable.
> * `serverId`: Conceptually represents a running instance of PrairieLearn. Defaults to `${instanceId}:${serverPort}`, but overridable by actually setting `serverId` in the config.
> 
> Rationale: it might be nice to be able to differentiate between instructor servers and chunk servers? Or maybe we just don't care about this.

I'd be fine with this scheme, but I don't actually see any use for it at the moment. The override was purely in place to deal with the fact that we have multiple servers on a single EC2 instance, but adding the port takes care of that, so I think we can just remove the capacity to override the instanceId/serverId.

nwalters512:
@mwest1066 sounds good! The only value I saw was that we could immediately differentiate between instructor servers and chunk servers, but I think that distinction is quickly losing meaning anyways. At any rate, it's not too hard to look up an instance ID in the EC2 instance list.

mwest1066 (approved):

PR #6442, opened 2022-10-03
Fix MathJax typesetting

### LINKED ISSUE #6441: LaTeX labels for pl-drawing are not rendering
As reported on Slack: https://prairielearn.slack.com/archives/C266KEH9A/p1664826166586559

Here are two examples of impacted questions:

- https://www.prairielearn.org/pl/course_instance/129504/instructor/question/9067610/preview
- https://www.prairielearn.org/pl/course_instance/129489/instructor/question/9074959/preview

In both cases, I see the following JavaScript error in the client-side console:

```
Uncaught (in promise) Error: MathJax retry
    at e.retryAfter (core.js:1:202736)
    at t.checkLoading (menu.js:1:31010)
    at t.checkLoading (menu.js:1:30260)
    at Object.renderMath (core.js:1:15655)
    at e.renderConvert (core.js:1:16468)
    at t.convert (core.js:1:25546)
    at t.convert (core.js:1:19581)
    at e.MathJax.<computed> [as tex2svg] (startup.js:formatted:793:36)
    at callback (mechanicsObjects.js:886:25)
    at preview:63:53
```

nwalters512:
Fixes #6441. This was inadvertently introduced by #6358: the inclusion of the contextual menu unexpectedly introduced async operations into MathJax startup.

As part of this PR, I introduced a new ESLint rule to warn us about usage of synchronous typesetting functions, as well as `MathJax.Hub.Queue`, which is deprecated.


elements/pl-rich-text-editor/pl-rich-text-editor.js
```diff
@@ -62,24 +62,15 @@ class MathFormula extends Embed {
   }
 
   static updateNode(node, value) {
-    // Wait for the page to be fully loaded. Mathjax will not work unless the page is fully loaded.
-    if (document.readyState === 'complete') {
-      let html = MathJax.tex2chtml(value);
+    MathJax.startup.promise.then(async () => {
+      let html = await MathJax.tex2chtmlPromise(value);
```
jonatanschroeder, line 66: Just did some tests locally. Adding formulas works fine, but when loading an answer that already includes some formulas, the console log includes an error:

image

I'm playing around to see if I can find the source, but just letting you know. This seems to only affect the question (editable) components, not the submission (read-only) components.

nwalters512: @jonatanschroeder I noticed that too, but the problem already exists on master. As it's not a regression, I don't think that should stand in the way of merging this.

The underlying problem is that `quilljs-markdown` assumes that `Quill#getSelection()` will always return a non-null value, but it's explicitly documented as sometimes returning null.

https://github.com/cloverhearts/quilljs-markdown/blob/84c80fee7d94bc2b027660e258137db7252e8f07/src/app.js#L167

jonatanschroeder: This seems to be unrelated to this issue. I'll open a new issue for it. It seems to be related to quilljs-markdown, I'll see if I can either fix it or replace it with another equivalent module (there seem to be other more reliable options).

jonatanschroeder:
I was tagged for review as code owner for pl-rich-text-editor, which seems to be working fine (other than #6444). I won't approve formally as I'm unfamiliar with the other issues, but that part of the code is ok.

elements/pl-drawing/mechanicsObjects.js
```diff
@@ -879,77 +879,62 @@ mechanicsObjects.LatexText = fabric.util.createClass(fabric.Object, {
...
-      } catch (err) {
-        // If we're erroring, it's because the MathJax glyph cache hasn't loaded yet.
-        //  We can retry again in a little bit to check if it's ready.
-        if (++num_tries < 10) {
-          setTimeout(callback, 100);
-          return;
-        } else {
-          throw err;
+  gen_text: async function (str, options) {
+    await MathJax.startup.promise;
```
mwest1066, line 883: Oh wow, this is much neater than the old hack to spin until MathJax has loaded.

mwest1066 (approved):

nwalters512:
For completeness, I tested on macOS Safari, Chrome, and Firefox, and things are looking good everywhere 🤞

nbalepur:
Hello, current in PrarieLearn we are experiencing issues with labels loading on <pl-drawing> elements. I'm wondering if it's related to this change

nwalters512:
@nbalepur you're most likely experiencing https://github.com/PrairieLearn/PrairieLearn/issues/6441, which was fixed by this PR. Let us know if you're still experiencing issues once this is deployed (look for an announcement on Slack or GitHub Discussions).

PR #6443, opened 2022-10-03
Silence matplotlib logging during startup

nwalters512:
Shortly after deploying new code, we sometimes see issues like the following occurring:

Screen Shot 2022-10-03 at 14 29 44

These originate from matplotlib's `FontManager` class:

https://github.com/matplotlib/matplotlib/blob/0517187b9c91061d2ec87e70442615cf4f47b6f3/lib/matplotlib/font_manager.py#L1098

This PR silences that warning by disabling the `Logger` instead that `FontManager` uses.

However, the fact that `FontManager` can take >5s to finish loading its cache is itself concerning. To attempt to alleviate that, we now import `matplotlib.font_manager` during zygote startup to hopefully ensure that it's available by the time someone tries to use it in their code.

Note that [EBS lazy initialization](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-initialize.html) is probably partly responsible for the slow loading + construction of the `FontManager` cache. Once we switch to executing code inside containers, this should improve significantly.

### DIFF
--- a/python/zygote.py
+++ b/python/zygote.py
@@ -15,7 +15,7 @@
 # Errors are signaled by exiting with non-zero exit code
 # Exceptions are not caught and so will trigger a process exit with non-zero exit code (signaling an error)
 
-import sys, os, json, importlib, copy, base64, io, matplotlib, signal, sklearn, nltk
+import sys, os, json, copy, base64, io, signal
 from inspect import signature
 
 saved_path = copy.copy(sys.path)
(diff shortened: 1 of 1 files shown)


python/zygote.py
```diff
@@ -15,7 +15,7 @@
 # Errors are signaled by exiting with non-zero exit code
 # Exceptions are not caught and so will trigger a process exit with non-zero exit code (signaling an error)
 
-import sys, os, json, importlib, copy, base64, io, matplotlib, signal, sklearn, nltk
+import sys, os, json, copy, base64, io, signal
```
nwalters512, line 18: `importlib` was unused; I moved the rest down to the pre-loading section below.

mwest1066 (approved):

PR #6445, opened 2022-10-03
Improve idle in transaction timeout logging

nwalters512:
I tested this locally by adding the following to my Postgres config at `/opt/homebrew/var/postgres/postgresql.conf`:

```
idle_in_transaction_session_timeout = 10000
```

I then added the following code to `server.js` to trigger an idle client error:

```js
namedLocks.tryLockAsync('foobar');
```

Sentry now shows a `last_query` tag:

Screen Shot 2022-10-03 at 17 10 00


prairielib/lib/sql-db.js
```diff
@@ -288,6 +293,7 @@ class PostgresPool {
     debug('queryWithClient()', 'params:', debugParams(params));
     const { processedSql, paramsArray } = paramsToArray(sql, params);
     try {
+      lastQueryMap.set(client, processedSql);
```
mwest1066, line 296: Would it be worth also including at least some of `paramsArray` here? I'm thinking we could truncate any params over 100 character in length, or something like that, to at least get ID params.

mwest1066 (approved):

prairielib/lib/sql-db.js
nwalters512: The trouble is that `params` could have completely arbitrary data in in, specifically PII. I agree that it would be very valuable for debugging, but from a data privacy standpoint we should do what we can to keep PII out of logs and third-party monitoring systems.

mwest1066: Yep, that's fair. Could we use AsyncLocalStorage to associate `res.locals.response_id` or `req.url` with the error? This is definitely not critical and I'm very happy to have this PR merged as is, but I'm just trying to think about other info that would be helpful to let us tie different logs together. Maybe we could even include a Honeycomb trace ID (if we have one in the current context)?

nwalters512: I'm actually not sure if that would work, specifically because I don't know if any of our code would actually be considered "upstream" of this error handler with regards to async-local storage. That is, the error is triggered asynchronously outside of any existing stream of work, so there wouldn't be any async contexts to reference.

What we *could* do is capture those things in `lastQueryMap`, but I'd rather not do that until we don't have any other choice, as I don't want that kind of weird coupling between DB internals and the rest of our app. If we find ourselves facing a problem that we absolutely cannot debug otherwise, we could make the pragmatic choice to add additional info then.

I'll merge this as-is!

mwest1066: Totally reasonable!

PR #6446, opened 2022-10-04
Set error status code before Sentry error middleware

nwalters512:
Followup to #6360. Our error handler runs after the Sentry error handler, which means `err.status` wasn't set in time for Sentry to read it.

### DIFF
--- a/server.js
+++ b/server.js
@@ -85,9 +85,7 @@ module.exports.initExpress = function () {
   app.set('trust proxy', config.trustProxy);
 
   // If we're set up with Sentry, use its middleware to record requests.
-  if (config.sentryDsn) {
-    app.use(Sentry.Handlers.requestHandler());
-  }
+  app.use(Sentry.Handlers.requestHandler());
 
   // Set res.locals variables first, so they will be available on
   // all pages including the error page (which we could jump to at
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -85,9 +85,7 @@ module.exports.initExpress = function () {
   app.set('trust proxy', config.trustProxy);
 
   // If we're set up with Sentry, use its middleware to record requests.
-  if (config.sentryDsn) {
-    app.use(Sentry.Handlers.requestHandler());
-  }
+  app.use(Sentry.Handlers.requestHandler());
```
mwest1066, line 88: The comment implies we only load this middleware if we are using Sentry, but the code now seems to load it unconditionally?

server.js
```diff
@@ -1624,29 +1622,51 @@ module.exports.initExpress = function () {
...
+  // We need to add our own handler here to ensure that we add the appropriate
+  // information to the current Sentry scope.
+  //
+  // Note that this is typically done by our `logResponse` middleware, but
+  // that doesn't run soon enough for the Sentry error handler to pick up.
+  app.use((err, req, res, next) => {
+    enrichSentryScope(req, res);
+    next(err);
+  });
+  app.use(Sentry.Handlers.errorHandler());
```
mwest1066, line 1669: Similarly here, I assume it's ok to be loading this middleware unconditionally?

mwest1066 (approved):

server.js
nwalters512: Yeah, this will be a no-op if we never initialized Sentry. Originally I was thinking it'd be nice to always check `config.sentryDsn` before trying to do something with Sentry, but Sentry is already doing the "am I initialized" check internally, so it's cleaner to just unconditionally call into Sentry and allow it to no-op when appropriate.

server.js
nwalters512: Good catch, let me clean up the comments.

server.js
mwest1066: Cool, thanks for confirming.

PR #6449, opened 2022-10-05
Fix workspace race condition

nwalters512:
This PR attempts to fix a workspace where multiple PrairieLearn servers try to create the same workspace at the same time.

We achieve this by having any PL host that tries to initialize a workspace construct the workspace's home directory in a temporary directory. We then grab and hold a lock while we move the temporary directory into place and update the workspace state. Then, only the host that manages to actually transition the state will try to assign the workspace to a host.

Original text retained below for posterity.

~The native solution would be to lock (via the database) a workspace as we're initializing it. However, locks consume a database connection, so we'd like to avoid long-lived locks wherever possible. My alternative solution was to allow file operations (copying files to S3 or EFS) to run in parallel, but then lock for a short time to transition the workspace state. This should also allow us to ensure that only one PrairieLearn instance tries to assign a workspace to a particular host.~

~This should be correct if the following assumptions hold:~

- ~For EFS-backed workspaces, things won't break if two hosts try to "initialize" the workspace contents on disk at the same time.~
- ~For S3-based workspaces, S3 must be resilient to the same file being written to concurrently.~

~I'm pretty confident about the latter, but less so about the former. Let's discuss.~


lib/workspace.js
```diff
@@ -190,43 +197,79 @@ module.exports = {
...
+    });
+
+    // Bail out if needed; this should only ever occur if another host is
+    // already trying to assign this host to a workspace.
+    if (!shouldAssignHost) return;
+
+    let workspace_host_id = null;
+    let attempt = 0;
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
```
nwalters512, line 250: This loop is probably going to cause problems during deploys: if we kill a server, no one else will ever try to assign a workspace to a host. I think there are two ways we could improve this:

- Add a cronjob to periodically try assigning workspaces to hosts. Then if host assignment doesn't immediately succeed after initializing the workspace, allow that cronjob to try performing host assignment instead.
- Switch from a push to a pull-based model, which could take a few forms:
  - Have workspace hosts long-poll PrairieLearn hosts for new workspaces.
  - Do what we do for external graders and push "launch requests" into SQS.

nwalters512:
@mwest1066 and I talked on Zoom and think we have a solution:

- For S3, we assume that concurrent writes are fine, as stated above.
- For EFS, we'll construct the initial directory in a temporary directory and then move it into the correct place. Note that on NFS (which EFS uses), renames are not atomic, so we'll likely have to hold a DB lock while performing the rename. This should be fast, however, so it's not as concerning.
  - We should most likely construct the initial directory on the EFS volume and not, say, in `/tmp/...`. I'd expect that renaming within EFS will be faster than renaming from local to EFS.

mwest1066 (approved):

PR #6451, opened 2022-10-05
chore: Normalize capitalization of sqldb across all files

SethPoulsen:


SethPoulsen:
I was annoyed some files having `sqldb` and some having `sqlDb`, so I normalized it to `sqldb` because that was the majority. I'm not too bothered about which capitalization is chose in the end, I mostly just want consistency.

tools/replace_prairielib_paths.py
```diff
@@ -40,22 +40,22 @@ def walk_directory(dirname, old_string, new_prefix, new_string=None):
 walk_directory('.', "require('@prairielearn/prairielib').error", "require('@prairielearn/prairielib/error')")
```
nwalters512, line 40: We should just remove this whole Python file entirely, it's not useful anymore.

prairielib/index.js
```diff
@@ -1,4 +1,3 @@
-module.exports.sqlDb = require('./lib/sql-db');
 module.exports.sqldb = require('./lib/sql-db');
```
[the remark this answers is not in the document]
nwalters512: I don't think anything is using this index file anymore; it should be safe to remove.

nwalters512 (approved):

tools/replace_prairielib_paths.py
SethPoulsen: done

PR #6457, opened 2022-10-06
Handle uncaught rejections when fetching chunks

nwalters512:
This started from a Sentry report of an unhandled rejection that, as far as I can tell, should have been completely handled. For folks with access: https://sentry.io/organizations/prairielearn-inc/issues/3651073375

I was able to reproduce an unhandled rejection locally by setting `chunksGenerator: true` and `chunksConsumer: true` in the config, syncing, and then trying to render the question. I saw something weird in the logs:

```
Emitted 'error' event on Domain instance at:
    at emit (node:internal/process/promises:138:33)
    at processPromiseRejections (node:internal/process/promises:274:27)
    at processTicksAndRejections (node:internal/process/task_queues:97:32) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'mkdir',
  path: '/chunks'
}
```

I didn't expect to see domains involved here, but we do have some of our instrumentation (Sentry + OpenTelemetry) that are using domains, so in retrospect it's not too surprising.

After adding some logs, I noticed that surprisingly, none of our Express error handlers were receiving this error, even though I verified that the error was being passed to `next()` in the route handler.

I dropped into a debugger right before the error was thrown and walked it all the way through. I indeed never saw it enter Express error handlers. As it turns out, this is because Express delays error handlers by a tick of the event loop:

https://github.com/expressjs/express/blob/631ada0c645dc84c6df8788f5a7eb2b8100acea5/lib/router/index.js#L203

I started commenting things out, and it turned out that removing the `.finally()` call from the promise made the error go away. Strange!

I originally reached for [`async-cache-dedupe`](https://github.com/mcollina/async-cache-dedupe), as I've seen the author of that package talking about the various pitfalls of caching promises. After discovering that that fixed the error, I looked at the source for that package and on a hunch, tried changing from `.finally(...)` to a `try` / `finally` block. That also worked!

---

After talking with @mwest1066 on Zoom and looking at this more closely, it's clearer what was wrong with the original code: when we called `chunkPromise.finally(...)`, that created a new promise chain. We never await or attach a `.catch()` to the end of it, so we naturally end up with an unhandled rejection!

### DIFF
--- a/lib/chunks.js
+++ b/lib/chunks.js
@@ -988,17 +988,19 @@ module.exports.ensureChunksForCourseAsync = async (courseId, chunks) => {
       // If this chunk is already being loaded, reuse the existing promise
       return pendingChunkPromise;
     }
+
     const chunkPromise = ensureChunk(courseId, chunk);
     pendingChunksMap.set(pendingChunkKey, chunkPromise);
-    chunkPromise.finally(() => {
+    try {
+      await chunkPromise;
+    } finally {
       // Once the promise has settled, remove it from our collection of
       // pending promises. This helps prevent memory leaks and, more
       // importantly, ensures we don't cache rejected promises - if loading
       // a chunk fails for some reason, this will ensure we try to load it
       // again the next time it's requested.
       pendingChunksMap.delete(pendingChunkKey);
-    });
-    return chunkPromise;
+    }
   });
 
   // We also need to take care to remove any chunks that are no longer valid.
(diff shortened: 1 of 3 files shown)


mwest1066 (approved):

lib/chunks.js
```diff
@@ -988,17 +988,19 @@ module.exports.ensureChunksForCourseAsync = async (courseId, chunks) => {
       // If this chunk is already being loaded, reuse the existing promise
       return pendingChunkPromise;
```
mwest1066, line 989: We could also add an `await` here for clarity if we want, although it doesn't matter either way.

PR #6461, opened 2022-10-07
Make errorAbandonedJobs safe for multiple servers

### LINKED ISSUE #3160: Cron errorAbandonedJobs is not multi-server safe

### LINKED ISSUE #6454: Error: Another user is already syncing or modifying the course:
I cant seem to finds a way to resolve the following issue after pressing Pull from remote Git Repository:
image

nwalters512:
This PR adds heartbeating to servers jobs so that we can properly detect when a job has been "abandoned", most likely by a server crash. Previously, we relied on an in-process list of "live" jobs, but that was only aware of jobs on that particular instance of PL; it didn't account for jobs that were running on other servers. This has caused a few unfortunate bugs, most notably stuck course locks during syncs or file edits.

Closes #3160.
Closes #6454.
Closes https://github.com/PrairieLearnInc/sysconf/issues/317.


cron/errorAbandonedJobs.js
```diff
@@ -1,12 +1,7 @@
-var ERR = require('async-stacktrace');
+const util = require('util');
 
-var serverJobs = require('../lib/server-jobs');
-
-module.exports = {};
+const serverJobs = require('../lib/server-jobs');
 
 module.exports.run = function (callback) {
```
nwalters512, line 5: Should we introduce `runAsync` support for cron jobs?

lib/server-jobs.js
```diff
@@ -189,19 +192,43 @@ Job.prototype.failAsync = util.promisify(Job.prototype.fail);
...
+  // Start a periodic task to heartbeat all live jobs. We don't use a cronjob
+  // for this because we want this to run for every host.
+  intervalId = setInterval(() => {
+    const jobIds = Object.keys(module.exports.liveJobs);
+    if (jobIds.length === 0) return;
+
+    sqlDb.queryAsync(sql.update_heartbeats, { job_ids: jobIds }).catch((err) => {
+      logger.error('Error updating heartbeats for live server jobs', err);
+    });
+  }, 10 * 1000);
```
nwalters512, line 209: Should this period be configurable?

lib/server-jobs.sql
```diff
@@ -144,6 +144,13 @@ FROM
 WHERE
     j.status = 'Running';
 
+-- BLOCK select_abandoned_jobs
+SELECT id, job_sequence_id
+FROM jobs AS j
+WHERE
+    j.status = 'Running'
+    AND j.heartbeat_at < (CURRENT_TIMESTAMP - interval '30 seconds');
```
nwalters512, line 152: Should this period be configurable?

lib/server-jobs.sql
```diff
@@ -160,7 +167,7 @@ WHERE
         ) -- no running jobs and no recently finished jobs
     )
 RETURNING
-    js.*;
+    js.id;
```
nwalters512, line 170: I suspect that in general, things will be slightly more performant if we're only returning columns we actually need?

pages/partials/jobSequenceResults.ejs
```diff
@@ -23,24 +23,36 @@
...
         // Join the rooms for the job_sequence and for each individual job.
         // Check return values for a change that happened since we rendered the page
         // and reload if anything changed. This avoids the race condition where
         // jobs change between page render and socket activation.
 
-        socket.emit('joinJobSequence', {job_sequence_id: <%= job_sequence.id %>}, function(msg) {
-            if (msg.job_count != '<%= job_sequence.job_count %>') {window.location.reload(true);}
+        socket.emit('joinJobSequence', {
+            job_sequence_id: '<%= job_sequence.id %>',
+            token: '<%= job_sequence.token %>',
```
nwalters512, line 40: Most changes here were just reformatting; the addition of the `token` properties here and below should be the only functional changes.

cron/errorAbandonedJobs.js
mwest1066: I'm definitely on board if you want to add it.

lib/server-jobs.js
```diff
@@ -9,6 +9,9 @@ const logger = require('./logger');
 const socketServer = require('./socket-server');
 const sqldb = require('../prairielib/lib/sql-db');
 const sqlLoader = require('../prairielib/lib/sql-loader');
+const sqlDb = require('../prairielib/sql-db');
```
mwest1066, line 12: Consistency with #6451
```suggestion
const sqldb = require('../prairielib/sql-db');
```

lib/server-jobs.js
mwest1066: I think so -- it's easy to do and can't hurt?

lib/server-jobs.sql
mwest1066: Yep again.

lib/server-jobs.js
```diff
@@ -492,15 +498,27 @@ module.exports.getJobSequenceAsync = util.promisify(module.exports.getJobSequenc
...
+    const jobSequenceTokenData = {
+      jobSequenceId: job_sequence_id.toString(),
+    };
+    jobSequence.token = csrf.generateToken(jobSequenceTokenData, config.secretKey);
+
+    (jobSequence.jobs ?? []).forEach((job) => {
+      // Also generate a token for each job.
+      const jobTokenData = { jobId: job.id.toString() };
+      job.token = csrf.generateToken(jobTokenData, config.secretKey);
+
```
mwest1066, line 512: Why are we doing this here and not inside `getJobSequence()`?

mwest1066 (approved):

lib/server-jobs.js
nwalters512: Great question! Because I started at the pages that render the live output and walked back to this callsite and no further 😂 I'll move it to that other function.

cron/errorAbandonedJobs.js
nwalters512: I do, but I'll save this for a separate PR.

lib/server-jobs.js
nwalters512: Done in cc1e319.

lib/server-jobs.js
nwalters512: Done in 5346f32.

lib/server-jobs.sql
nwalters512: Done in 5346f32.

lib/server-jobs.js
```diff
@@ -476,7 +483,27 @@ module.exports.getJobSequence = function (job_sequence_id, course_id, callback)
...
+      if (job.output) {
+        const ansiup = new AnsiUp();
+        job.output = ansiup.ansi_to_html(job.output);
```
mwest1066, line 502: I think these lines should not have been copied from `getJobSequenceWithFormattedOutput()` to this location in `getJobSequence()`?

nwalters512: Indeed! Removed in 7afbe51.

PR #6466, opened 2022-10-10
Pull prairielearn/prairielearn image in CI

nwalters512:


nwalters512:
I was thinking that this might allow us to skip Yarn dependency installation if Docker would be clever enough to reuse those layers from the existing image. Unfortunately, it doesn't look like it will!

PR #6469, opened 2022-10-11
Configure HTTP keep-alive timeout

nwalters512:
This should resolve the occasional 502s we see in the load balancer logs.

See the following for more details:

- https://www.tessian.com/blog/how-to-fix-http-502-errors/
- https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout

Closes https://github.com/PrairieLearnInc/sysconf/issues/398

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -45,6 +45,14 @@ config.authUin = '000000000'; // to specify the user uin on a local server (with
 config.authnCookieMaxAgeMilliseconds = 30 * 24 * 60 * 60 * 1000;
 config.serverType = 'http';
 config.serverPort = '3000';
+config.serverTimeout = 10 * 60 * 1000; // 10 minutes
+/**
+ * How many milliseconds to wait before destroying a socket that is being
+ * kept alive. This should always be greater than the timeout at the load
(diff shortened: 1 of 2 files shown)


lib/config.js
```diff
@@ -45,6 +45,14 @@ config.authUin = '000000000'; // to specify the user uin on a local server (with
 config.authnCookieMaxAgeMilliseconds = 30 * 24 * 60 * 60 * 1000;
 config.serverType = 'http';
 config.serverPort = '3000';
+config.serverTimeout = 10 * 60 * 1000; // 10 minutes
```
nwalters512, line 48: It wasn't strictly necessary to make this configurable, but I figure it doesn't hurt!

mwest1066: Want to default this to something a bit shorter? Like maybe 90 seconds? Just so it's not so extreme a change from the current 5 second default.

mwest1066: Oh sorry, I misread what this was, please ignore this comment!

mwest1066: I'm not really sure what the impact of this is going to be in practice, but I agree that it doesn't seem like it will hurt anything.

PR #891, opened 2017-11-01
Enhanced question statistics

mwest1066:

PR #6472, opened 2022-10-11
Remove calculateAssessmentMode cron job

nwalters512:
The `assessments.mode` column was introduced in https://github.com/PrairieLearn/PrairieLearn/pull/891, but is currently unused (as far as I can tell - @mwest1066 did you confirm this?). The `calculateAssessmentMode` cron job that computes it does a very expensive table scan over `assessment_instances` and `assessments`. It only runs once a day, but it puts relatively high load on the database, which we want to avoid. So, this PR removes that cron job.

For folks with access, see https://github.com/PrairieLearnInc/sysconf/issues/397#issuecomment-1274827340 for more details.

The `assessments.mode` column can be removed entirely too. I was thinking of saving that for a separate PR for after this code has been deployed, but honestly we can probably just remove it in this PR and allow the cron job to fail if it happens to run in the middle of a deployment. @mwest1066 thoughts?


mwest1066:
> The `assessments.mode` column was introduced in #891, but is currently unused (as far as I can tell - @mwest1066 did you confirm this?).

Yep, I confirmed.

> The `assessments.mode` column can be removed entirely too. I was thinking of saving that for a separate PR for after this code has been deployed, but honestly we can probably just remove it in this PR and allow the cron job to fail if it happens to run in the middle of a deployment. @mwest1066 thoughts?

I agree that it's fine to just remove the column in this PR.

mwest1066 (approved):

mwest1066 (approved):

PR #6474, opened 2022-10-11
Remove stoped workspaces and terminated hosts from CloudWatch

nwalters512:
The number of stopped workspaces and terminated workspace hosts will grow without bound, so there's no value in including them in CloudWatch. They also trigger sequential scans of `workspaces` and `workspace_hosts`, which is extra load on the database for no good reason.

This PR rewrites some of the queries in `workspace_loads_current` so that they can use indexes on the `workspace` table instead of doing sequential scans all the time.

@mwest1066 called out that "we might need some new indexes to help with this". However, I think these two indexes will be used for these queries:

- https://github.com/PrairieLearn/PrairieLearn/blob/9c3cfb7f110bf1e9780a636b4d978f558ad625c2/database/tables/workspaces.pg#L26
- https://github.com/PrairieLearn/PrairieLearn/blob/9c3cfb7f110bf1e9780a636b4d978f558ad625c2/database/tables/workspace_hosts.pg#L17

Does that sound correct to you?


mwest1066:
I tested the new queries and it looks like they are still doing full table scans -- see (1) below. In (2) below I tried adding an explicit `WHERE` clause and it helped a bit (360 ms down to 95 ms) but it still does a full scan. In (3) below I tried just doing one state and that was really fast (1 ms) and correctly used the index.

I think it might be best to break up all these queries to just do each state one at a time?

## (1) Queries as they are in this PR

```
EXPLAIN ANALYZE
SELECT
    count(*) FILTER (WHERE w.state = 'uninitialized'),
    count(*) FILTER (WHERE w.state = 'launching'),
    count(*) FILTER (WHERE w.state = 'launching' AND num_nonnulls(w.rebooted_at, w.reset_at) > 0),
    count(*) FILTER (WHERE w.state = 'running')
FROM
    workspaces AS w;

                                                                    QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------
 Finalize Aggregate  (cost=46603.27..46603.28 rows=1 width=32) (actual time=362.620..364.527 rows=1 loops=1)
   ->  Gather  (cost=46603.04..46603.25 rows=2 width=32) (actual time=362.529..364.513 rows=3 loops=1)
         Workers Planned: 2
         Workers Launched: 2
         ->  Partial Aggregate  (cost=45603.04..45603.05 rows=1 width=32) (actual time=359.031..359.038 rows=1 loops=3)
               ->  Parallel Seq Scan on workspaces w  (cost=0.00..39764.44 rows=233544 width=20) (actual time=0.008..200.092 rows=187147 loops=3)
 Planning Time: 0.061 ms
 Execution Time: 364.584 ms
(8 rows)
```

## (2) Queries modified with extra `WHERE` clauses

```
EXPLAIN ANALYZE
SELECT
    count(*) FILTER (WHERE w.state = 'uninitialized'),
    count(*) FILTER (WHERE w.state = 'launching'),
    count(*) FILTER (WHERE w.state = 'launching' AND num_nonnulls(w.rebooted_at, w.reset_at) > 0),
    count(*) FILTER (WHERE w.state = 'running')
FROM
    workspaces AS w
WHERE
    w.state = 'uninitialized'
    OR w.state = 'launching'
    OR w.state = 'running';

                                                                                QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Finalize Aggregate  (cost=43535.68..43535.69 rows=1 width=32) (actual time=92.704..94.517 rows=1 loops=1)
   ->  Gather  (cost=43535.45..43535.66 rows=2 width=32) (actual time=92.609..94.503 rows=3 loops=1)
         Workers Planned: 2
         Workers Launched: 2
         ->  Partial Aggregate  (cost=42535.45..42535.46 rows=1 width=32) (actual time=89.596..89.601 rows=1 loops=3)
               ->  Parallel Seq Scan on workspaces w  (cost=0.00..41516.02 rows=40777 width=20) (actual time=0.013..61.811 rows=33123 loops=3)
                     Filter: ((state = 'uninitialized'::enum_workspace_state) OR (state = 'launching'::enum_workspace_state) OR (state = 'running'::enum_workspace_state))
                     Rows Removed by Filter: 154026
 Planning Time: 0.181 ms
 Execution Time: 94.583 ms
```

## (3) Individual query for one state

```
EXPLAIN ANALYZE
SELECT
    count(*)
FROM
    workspaces AS w
WHERE
    w.state = 'running';

                                                                           QUERY PLAN
----------------------------------------------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=181.43..181.44 rows=1 width=8) (actual time=1.358..1.361 rows=1 loops=1)
   ->  Index Only Scan using workspaces_state_launched_at_key on workspaces w  (cost=0.42..180.36 rows=430 width=0) (actual time=0.020..1.068 rows=358 loops=1)
         Index Cond: (state = 'running'::enum_workspace_state)
         Heap Fetches: 363
 Planning Time: 0.083 ms
 Execution Time: 1.385 ms
```

echuber2:
What happens if you use `WHERE ... OR` in version 3, but without using `FILTER`?

nwalters512:
In 7f181fd, I changed to use separate queries for all data points. I spot-checked a few of them and they're now using indexes instead of sequential scans!

mwest1066 (approved):

PR #6475, opened 2022-10-11
Introduce new API for server jobs

nwalters512:
This PR introduces a simplified API for server jobs that makes use of async/await instead of explicitly-chained callbacks. While the code produced is generally more concise, it's also much easier to reason about, as one can use standard `if`/`else`/`try`/`catch` control flow!

The new API looks like this:

```ts
import { createServerJob } from './server-jobs-2';

const serverJob = await createServerJob({
  type: 'type',
  description: 'description',
});

serverJob.executeInBackground(async (context) => {
  context.info('Show messages!');
  await context.exec('git', ['status']);
});
```

From the consumer's point of view, there is no longer a need to deal explicitly in job sequences and jobs; there's just a single job. Internally, it's still represented as a job sequence with a single job associated with it.

I intend for the old and new APIs to coexist for some time as we slowly migrate from old to new. I migrated a few in this PR as a proof-of-concept:

- Course syncing
- Course git status
- Question testing


apps/prairielearn/src/lib/question.js
```diff
@@ -1215,84 +1214,55 @@ module.exports = {
...
-                success = success && ret_success;
-                callback(null);
-              }
-            );
-          },
-          (err) => {
-            if (ERR(err, () => {})) return job.fail(err);
-            if (!success) return job.fail('Some tests failed. See the "Errors" page for details.');
-            job.succeed();
+    serverJob.executeInBackground(async (context) => {
```
nwalters512, line 1239: An alternative API would have been to pass the execution function as the second argument to `createServerJob`. However, that makes this more difficult to test: in tests, I want to wait for everything to finish (`serverJob.execute(...)`), and in real-world code, I don't want to *ever* wait (`serverJob.executeInBackground(...)`). Hence, the separate functions.

apps/prairielearn/src/lib/server-jobs-2.ts
```diff
@@ -0,0 +1,184 @@
...
+  description: string;
+}
+
+interface ServerJobExecOptions {
+  cwd?: string;
+  env?: NodeJS.ProcessEnv;
+}
+
+type ServerJobContext = Omit<ServerJob, 'execute' | 'executeInBackground'>;
+type JobExecutionFunction = (context: ServerJobContext) => Promise<void>;
```
nwalters512, line 31: Thoughts on what else to name the object that the execution function receives? I considered `job` too.

apps/prairielearn/src/lib/server-jobs-2.ts
```diff
@@ -0,0 +1,184 @@
...
+
+class ServerJob {
+  public jobSequenceId: string;
+  public jobId: string;
+  private started = false;
+  private finished = false;
+  public output = '';
```
nwalters512, line 38: This has to be public for `lib/server-jobs.js` to read the output in its socket handling code.

apps/prairielearn/src/lib/server-jobs-2.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  constructor(jobSequenceId: string, jobId: string) {
+    this.jobSequenceId = jobSequenceId;
+    this.jobId = jobId;
+  }
```
nwalters512, line 43: I chose to model a non-legacy server job as a job sequence containing a single job. This allows us to reuse most of the code from `lib/server-jobs.js` w.r.t. sockets, abandoning jobs, etc.

apps/prairielearn/src/lib/server-jobs-2.ts
```diff
@@ -0,0 +1,184 @@
...
+    this.addToOutput(msg + '\n');
+  }
+
+  verbose(msg: string) {
+    this.addToOutput(msg + '\n');
+  }
+
+  async exec(file: string, args: string[] = [], options: ServerJobExecOptions = {}): Promise<void> {
+    const msg = `$ ${file} ${args.join(' ')}\n`;
+    this.addToOutput(chalk.greenBright(msg));
```
nwalters512, line 63: `greenBright` looked nice, so I went with it without much thought:

Screenshot 2023-05-02 at 17 32 38

Open to other suggestions though!

apps/prairielearn/src/lib/server-jobs-2.ts
```diff
@@ -0,0 +1,184 @@
...
+    });
+
+    proc2.all?.setEncoding('utf-8');
+    proc2.all?.on('data', (data) => {
+      this.addToOutput(data);
+    });
+
+    const result = await proc2;
+    if (result.exitCode !== 0) {
+      throw new Error(`Process exited with code ${result.exitCode}`);
```
nwalters512, line 77: This code doesn't log `result.signal` - should it?

apps/prairielearn/src/lib/server-jobs-2.ts
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  /**
+   * Identical to {@link execute}, except it doesn't return a Promise. This is
+   * just used as a hint to the caller that they don't need to wait for the job
+   * to finish. Useful for tools like TypeScript and ESLint that ensure that
+   * promises are awaited.
+   */
+  executeInBackground(fn: JobExecutionFunction): void {
+    this.execute(fn);
```
nwalters512, line 112: Maybe we should do the `this.started` check here too? Otherwise we'd throw an error that would never be caught since nothing awaits this promise.

apps/prairielearn/src/lib/server-jobs-2.ts
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  private addToOutput(msg: string) {
+    this.output += msg;
+    const ansiUp = new AnsiUp();
+    const ansifiedOutput = ansiUp.ansi_to_html(this.output);
+    socketServer.io
+      .to('job-' + this.jobId)
+      .emit('change:output', { job_id: this.jobId, output: ansifiedOutput });
```
nwalters512, line 121: I haven't yet implemented the thing where we only send deltas. Thankfully, it seems like `ansi_up` can work in a streaming manner, so this shouldn't be too hard to do in another PR! We just need to retain a single `AnsiUp` instead for each job.

apps/prairielearn/src/lib/server-jobs.js
```diff
@@ -46,9 +47,15 @@ const sql = sqldb.loadSqlEquiv(__filename);
...
-// to accumulate stderr/stdout, which are only written to the DB
-// once the job is finished.
+/** @typedef {{ output: string }} AbstractJob */
+
+/**
+ * Store currently active job information in memory. This is used
+ * to accumulate stderr/stdout, which are only written to the DB
+ * once the job is finished.
+ *
+ * @type {Record<string, AbstractJob>}
```
nwalters512, line 57: This type ensures that anything reading from this object doesn't rely on this being either a new or legacy server job. As long as it has an `output` property, this code is happy!

apps/prairielearn/src/lib/server-jobs.js
```diff
@@ -335,33 +342,33 @@ module.exports.spawnJob = function (options, callback) {
       cwd: job.options.working_directory,
       env: job.options.env,
     };
-    job.proc = child_process.spawn(job.options.command, job.options.arguments, spawnOptions);
+    const proc = child_process.spawn(job.options.command, job.options.arguments, spawnOptions);
```
nwalters512, line 345: Moving from the job object to a local made TypeScript happy. Nothing but this function accessed `job.proc`, so this is safe.

apps/prairielearn/src/pages/shared/syncHelpers.js
```diff
@@ -21,310 +22,121 @@ const error = require('@prairielearn/error');
 
 const docker = new Docker();
 
-module.exports.pullAndUpdate = function (locals, callback) {
-  const options = {
-    course_id: locals.course.id,
-    user_id: locals.user.user_id,
-    authn_user_id: locals.authz_data.authn_user.user_id,
+module.exports.pullAndUpdate = async function (locals) {
```
nwalters512, line 25: This function specifically serves as a great example of how much more concise and readable new-style server jobs can be! Careful review here would be appreciated since this is such a foundational bit of functionality.

apps/prairielearn/src/tests/serverJobs2.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    });
+
+    const finishedJobSequence = await serverJobs.getJobSequenceAsync(serverJob.jobSequenceId, null);
+
+    assert.equal(finishedJobSequence.status, 'Error');
+    assert.lengthOf(finishedJobSequence.jobs, 1);
+
+    const job = finishedJobSequence.jobs[0];
+    assert.equal(job.status, 'Error');
+    assert.match(job.output, /^testing info\nError: failing job\n\s+at/);
```
nwalters512, line 68: TODO: test `job.error_message` too.

apps/prairielearn/src/pages/partials/jobSequenceResults.ejs
```diff
@@ -68,12 +68,14 @@
...
-                <% if (job.working_directory) { %> run in <code><%= job.working_directory %></code><% } %>
-            </p>
+            <% if (job_sequence.legacy) { %>
+                <h4 class="list-group-item-heading">Stage <%= job.number_in_sequence %>: <%= job.description %></h4>
+                <% if (job.command) { %>
+                <p class="mb-1">
+                    Command <code><%= job.command %><% job.arguments.forEach(function(arg) { %> <%= arg %><% }); %></code>
+                    <% if (job.working_directory) { %> run in <code><%= job.working_directory %></code><% } %>
+                </p>
+                <% } %>
```
nwalters512, line 78: I turned off printing of the command, working directory, status, and signal since those should never be set on the `job` itself for non-legacy sequences.

apps/prairielearn/src/lib/server-jobs-2.ts
nwalters512, line 1: Is there a better name for this file? Maybe make this one `server-jobs.ts` and the other one `server-jobs-legacy.js`?

apps/prairielearn/src/lib/server-jobs-2.ts
```diff
@@ -0,0 +1,184 @@
...
+  info(msg: string) {
+    this.addToOutput(msg + '\n');
+  }
+
+  verbose(msg: string) {
+    this.addToOutput(msg + '\n');
+  }
+
+  async exec(file: string, args: string[] = [], options: ServerJobExecOptions = {}): Promise<void> {
+    const msg = `$ ${file} ${args.join(' ')}\n`;
```
nwalters512, line 62: Should we be logging the working directory somehow? We previously persisted that information in `jobs.working_directory`.

apps/prairielearn/src/lib/server-jobs-2.ts
```diff
@@ -0,0 +1,184 @@
...
+  courseRequestId?: string;
+  assessmentId?: string;
+  userId?: string;
+  authnUserId?: string;
+  type: string;
+  description: string;
+}
+
+interface ServerJobExecOptions {
+  cwd?: string;
```
nwalters512, line 26: This should probably be a mandatory property - I can't imagine we'd ever want to just run commands in the working directory (the default)?

apps/prairielearn/src/lib/server-jobs-2.ts
mwest1066: Looks good to me!

apps/prairielearn/src/lib/server-jobs-2.ts
mwest1066: Seems like it wouldn't hurt?

apps/prairielearn/src/lib/server-jobs-2.ts
mwest1066: I'm in favor of anything that can make sure we never throw errors that will be lost.

apps/prairielearn/src/lib/server-jobs-2.ts
mwest1066: Oh cool, that will be very nice to have!

apps/prairielearn/src/lib/server-jobs.js
mwest1066: Sneaky!

apps/prairielearn/src/lib/server-jobs-2.ts
mwest1066: Yep, I like the use of the `-legacy` name and it fits with the new `legacy` column.

apps/prairielearn/src/lib/server-jobs-2.ts
mwest1066: I don't know that we want to display this publicly, but it feels like it could be helpful for debugging (like the environment for the job). Is it worth adding a secret log for these that we persist into the DB?

apps/prairielearn/src/lib/server-jobs-2.ts
mwest1066: I can't think of any reason that we wouldn't specify the working directory explicitly so I'm on board with making this argument mandatory.

apps/prairielearn/src/lib/server-jobs-2.ts
mwest1066: Tbh I'm not really in favor of `private` at all because it makes `printf` debugging more painful. I generally prefer to have everything public and use a leading underscore (or similar convention) to indicate internal variables that should only be messed with at your own peril. That being said, I don't feel very strongly about this.

apps/prairielearn/src/lib/server-jobs-2.ts
```diff
@@ -0,0 +1,184 @@
...
+        Sentry.captureException(err);
+      }
+    }
+  }
+
+  /**
+   * Identical to {@link execute}, except it doesn't return a Promise. This is
+   * just used as a hint to the caller that they don't need to wait for the job
+   * to finish. Useful for tools like TypeScript and ESLint that ensure that
+   * promises are awaited.
```
mwest1066, line 109: Ah! I was wondering why we don't just call `execute()` and not await it. Do we actually have tooling in place to catch non-awaited promises? I don't think my PT setup is detecting this?

apps/prairielearn/src/lib/server-jobs-2.ts
```diff
@@ -0,0 +1,184 @@
...
+    this.addToOutput(msg + '\n');
+  }
+
+  info(msg: string) {
+    this.addToOutput(msg + '\n');
+  }
+
+  verbose(msg: string) {
+    this.addToOutput(msg + '\n');
+  }
```
mwest1066, line 59: Do we want to add some colors to these various outputs to mirror what happens if you call `logger.warn()` with console output? And should we have some kind of section headers (a `header()` method?) that can both color headers and in the future perhaps provide collapsible sections like GitHub actions?

apps/prairielearn/src/pages/shared/syncHelpers.js
mwest1066: It's crazy how much shorter and clearer this now is!! 🚀 

I checked it carefully and it looks good to me.

apps/prairielearn/src/lib/server-jobs-2.ts
mwest1066: Actually I think we used to display it publicly so we probably could still do so if that's the easiest option?

apps/prairielearn/src/lib/server-jobs-2.ts
mwest1066: I like `job` better for this.

apps/prairielearn/src/lib/question.js
mwest1066: I think this is easy enough to use!

mwest1066 (approved):

apps/prairielearn/src/lib/server-jobs-2.ts
nwalters512: In Alma I render commands as `cd /path/to/working/directory && git command --goes-here`. I could also do something like `/path/to/working/directory $ git command --goes-here`. Or even this:

```
Working directory: /path/to/working/directory
Command: git command --goes-here
```

Any preference?

apps/prairielearn/src/lib/server-jobs-2.ts
nwalters512: I like the idea of colors for warnings/errors - I can add that.

For collapsible sections, I'm envisioning an API like this:

```ts
await job.group('Name of group', async () => {
  // Do things in group
  job.info('here is a log');
});
```

If we borrow [GitHub's syntax](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines) directly, that would output something like this:

```txt
::group::Name of group
here is a log
::endgroup::
```

I'm inclined to avoid implementing this for now - this will most likely require shifting the rendering of logs to the client in order to support groups inside streaming logs.

apps/prairielearn/src/lib/server-jobs-2.ts
nwalters512: Changed to `job` in 90f94d6.

apps/prairielearn/src/lib/server-jobs-2.ts
nwalters512: In 90f94d6 I changed it to use Execa's own error, which properly logs both exit codes and signals.

apps/prairielearn/src/lib/server-jobs-2.ts
nwalters512: Changed in 90f94d6.

apps/prairielearn/src/tests/serverJobs2.test.ts
nwalters512: Test added in 90f94d6.

apps/prairielearn/src/lib/server-jobs-2.ts
nwalters512: Changed in 89f64ab / b90d568.

apps/prairielearn/src/lib/server-jobs-2.ts
nwalters512: When trying to make this change, TypeScript pointed out a case where we don't _technically_ need an explicit directory: `git clone`.

apps/prairielearn/src/lib/server-jobs-2.ts
nwalters512: Unfortunately we don't have that for TypeScript yet. `@typescript-eslint` does have rules we can use for that (https://typescript-eslint.io/rules/no-floating-promises), but they require type checking information to be made available to ESLint, and getting that to work in monorepos is a headache (https://github.com/typescript-eslint/typescript-eslint/issues/1718 - we OOM).

I think we can make it work if we add an `.eslintrc.js` to each package/app? That's a large change for another day though 🙂

apps/prairielearn/src/lib/server-jobs-2.ts
nwalters512: I added red/yellow/dim colors for error/warn/verbose, respectively, in e4e2748.

apps/prairielearn/src/lib/server-jobs-2.ts
mwest1066: Either the `/dir $ cmd` or the more verbose syntax look good to me. If I was forced to pick then I might pick the second one to avoid super long lines, but it's honestly a toss-up.

apps/prairielearn/src/lib/server-jobs-2.ts
mwest1066: Ha, nice.

mwest1066: I wouldn't object to forcing us to use `/` in this case, just for the certainty that it provides.

apps/prairielearn/src/lib/server-jobs-2.ts
mwest1066: Yep, no worries, I was just hoping 😄

apps/prairielearn/src/lib/server-jobs-2.ts
mwest1066: Nice! I'm on board with deferring groups to a later time.

apps/prairielearn/src/lib/server-jobs-2.ts
mwest1066: Ah! That's good to know. I was conflating the two types of `private`.

apps/prairielearn/src/lib/server-jobs.ts
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    delete serverJobs.liveJobs[this.jobId];
+
+    await queryAsync(sql.update_job_on_finish, {
+      job_sequence_id: this.jobSequenceId,
+      job_id: this.jobId,
+      output: this.output,
+      status: err ? 'Error' : 'Success',
+      error_message: err?.toString() ?? null,
```
nwalters512, line 161: @mwest1066 taking a second look at this, do we actually want `error_message` to be a part of non-legacy server jobs? This just repeats information that's already visible in the logs, so I'm leaning towards removing it.

apps/prairielearn/src/lib/server-jobs-2.ts
nwalters512: Sounds reasonable. I made that change in 35f7710.

apps/prairielearn/src/lib/server-jobs-2.ts
nwalters512: I went with the verbose one in 35f7710.

apps/prairielearn/src/lib/server-jobs.ts
mwest1066: Yeah, I agree it's not needed

nwalters512: Removed in 02cb07d 🚀

apps/prairielearn/test.mjs
```diff
@@ -0,0 +1,2 @@
+console.warn('testing')
+console.error('error')
```
nwalters512, line 2: Whoops, need to delete this file.

nwalters512: Removed in 0b3493c.

PR #6476, opened 2022-10-12
Allow theme changes for pl-file-editor

louisunlimited:
Finally got modal working, now everything should work as desired. The themes are persisted through localStorage and will be loaded once a new FileEditor instance is created. I was wondering if I need to let jquery handle modal dismiss or is the current implementation okay (data-dismiss="modal").


louisunlimited:
recheck

louisunlimited:

elements/pl-file-editor/pl-file-editor.mustache
```diff
@@ -21,8 +21,83 @@ $(function() {
...
+             <button type="button" class="btn btn-sm btn-secondary settings-button" tabindex="-1">
+                 <div class="fa fa-cog"></div>
+             </button>
+         </div>
         <div class="editor"></div>
+        <div class="modal fade" tabindex="-1" aria-hidden="true">
+        <div class="modal-dialog modal-dialog-centered" role="document">
+            <div class="modal-content">
+                <div class="modal-header">
+                    <h5 class="modal-title" id="exampleModalLabel">Editor Settings</h5>
```
nwalters512, line 35: This is not an appropriate `id` for this modal. You should incorporate the `uuid` in case there are multiple file editors on the page.

elements/pl-file-editor/pl-file-editor.mustache
```diff
@@ -21,8 +21,83 @@ $(function() {
 <div id="file-editor-{{uuid}}" class="mathjax_ignore">
     <input type="hidden" name="{{name}}">
     <div class="card">
-        <div class="card-header">{{file_name}}</div>
+        <div class="card-header d-flex justify-content-between align-items-center">
+             {{file_name}}
+             <button type="button" class="btn btn-sm btn-secondary settings-button" tabindex="-1">
```
nwalters512, line 26: Is there a reason you set `tabindex` here? A `<button>` should already be tabbable.

nwalters512 (commented):
Glad you got the modal issue figured out! I gave this a spin and functionally it's looking nice.

One thought: it might be nice to update the theme in real-time as the user selects different options in the modal. That way, they can at least get a sense of what the theme they chose looks like, even if it's partly hidden behind the modal.

elements/pl-file-editor/pl-file-editor.mustache
louisunlimited: By uuid do you mean the uuid passed in when creating editor instance?

louisunlimited: Actually I was thinking we should add `id` to line 32 instead. Shouldn't it be on the entire modal itself instead of on modal-title?

elements/pl-file-editor/pl-file-editor.js
```diff
@@ -100,6 +107,25 @@ window.PLFileEditor.prototype.updatePreview = function (html_contents) {
   }
 };
 
+window.PLFileEditor.prototype.initSettingsButton = function () {
+  var that = this;
```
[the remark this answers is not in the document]
louisunlimited: I see what's the problem here, should I keep this consistent with the rest of the functions and write these as non-arrow function instead? 
```javascript
this.settingsButton.click(function () {
    that.modal.modal('show');
  });
```

elements/pl-file-editor/pl-file-editor.mustache
louisunlimited: Yeah this should be removed, I was thinking that some weird behaviors might show up when modal is present, now it seems not necessary.

louisunlimited:
@nwalters512 I believe I have finished most of the desired features. This code might still need a bit of refactoring but should be functionally good.

elements/pl-file-editor/pl-file-editor.js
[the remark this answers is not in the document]
nwalters512: I don't think there's a problem with arrow functions, I'd just encourage you to not alias `this` if you do.

elements/pl-file-editor/pl-file-editor.mustache
nwalters512: No, the ID should be on the title, and the modal itself should reference that ID via `aria-labelledby="..."`. See https://getbootstrap.com/docs/4.6/components/modal/#accessibility for more details.

elements/pl-file-editor/pl-file-editor.mustache
```diff
@@ -21,8 +21,36 @@ $(function() {
...
-        <div class="card-header">{{file_name}}</div>
+        <div class="card-header d-flex justify-content-between align-items-center">
+             {{file_name}}
+             <button type="button" class="btn btn-sm btn-secondary settings-button">
+                 <div class="fa fa-cog"></div>
+             </button>
+         </div>
         <div class="editor"></div>
+        <div class="modal fade" tabindex="-1" aria-hidden="true" id="{{uuid}}">
+            <div class="modal-dialog modal-dialog-centered" role="document">
```
nwalters512, line 32: This shouldn't be necessary.

```suggestion
            <div class="modal-dialog modal-dialog-centered">
```

elements/pl-file-editor/pl-file-editor.mustache
```diff
@@ -21,8 +21,36 @@ $(function() {
...
+                    <div class="modal-header">
+                        <h5 class="modal-title">Editor Settings</h5>
+                        <button type="button" class="close close-settings-button" data-dismiss="modal" aria-label="Close">
+                            <span aria-hidden="true">&times;</span>
+                        </button>
+                    </div>
+                    <div class="modal-body">
+                        <div class="form-group">
+                            <label for="{{uuid}}-themes">Editor Theme</label>
+                            <select class="form-control" id="{{uuid}}-themes">
```
nwalters512, line 43: See https://getbootstrap.com/docs/4.6/components/forms/#select-menu.

```suggestion
                            <select class="custom-select" id="{{uuid}}-themes">
```

elements/pl-file-editor/pl-file-editor.mustache
```diff
@@ -21,8 +21,36 @@ $(function() {
 <div id="file-editor-{{uuid}}" class="mathjax_ignore">
     <input type="hidden" name="{{name}}">
     <div class="card">
-        <div class="card-header">{{file_name}}</div>
+        <div class="card-header d-flex justify-content-between align-items-center">
+             {{file_name}}
+             <button type="button" class="btn btn-sm btn-secondary settings-button">
+                 <div class="fa fa-cog"></div>
```
nwalters512, line 27:
```suggestion
                 <i class="fa fa-cog"></i>
```

elements/pl-file-editor/pl-file-editor.mustache
```diff
@@ -21,8 +21,36 @@ $(function() {
 <div id="file-editor-{{uuid}}" class="mathjax_ignore">
     <input type="hidden" name="{{name}}">
     <div class="card">
-        <div class="card-header">{{file_name}}</div>
+        <div class="card-header d-flex justify-content-between align-items-center">
+             {{file_name}}
+             <button type="button" class="btn btn-sm btn-secondary settings-button">
```
nwalters512, line 26: This is important for accessibility:

```suggestion
             <button type="button" class="btn btn-sm btn-secondary settings-button" title="Editor settings" aria-label="Editor settings">
```

elements/pl-file-editor/pl-file-editor.js
```diff
@@ -100,6 +108,56 @@ window.PLFileEditor.prototype.updatePreview = function (html_contents) {
...
+
+  this.saveSettingsButton.click(function () {
+    var theme = that.modal.find('#' + uuid + '-themes').val();
+    that.editor.setTheme(theme);
+    localStorage.setItem('pl-file-editor-theme', theme);
+    that.modal.modal('hide');
+  });
+
+  this.closeSettingsButton.click(function () {
+    that.editor.setTheme(localStorage.getItem('pl-file-editor-theme'));
```
nwalters512, line 157: This value isn't guaranteed to be present in local storage if the user never changed the theme, correct? We should be reverting back to whatever the theme was before the modal was opened, not necessarily what was in local storage.

elements/pl-file-editor/pl-file-editor.js
```diff
@@ -100,6 +108,56 @@ window.PLFileEditor.prototype.updatePreview = function (html_contents) {
...
+      var theme = $(this).val();
+      that.editor.setTheme(theme);
+      console.log('theme', theme);
+    });
+  });
+
+  this.saveSettingsButton.click(function () {
+    var theme = that.modal.find('#' + uuid + '-themes').val();
+    that.editor.setTheme(theme);
+    localStorage.setItem('pl-file-editor-theme', theme);
```
nwalters512, line 152: Something to consider: if a question has multiple file editors, we should try to sync the settings between them so that the user doesn't have to reload the page for the change to take effect. It'd also be nice to have this sync immediately across tabs. For the latter, there's a [`storage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event) event that can be listened to. For the former... I'm not totally sure. Maybe we can use custom events for this?

elements/pl-file-editor/pl-file-editor.js
```diff
@@ -100,6 +108,56 @@ window.PLFileEditor.prototype.updatePreview = function (html_contents) {
...
+            })
+          );
+        } else {
+          themeSelect.append(
+            $('<option>', {
+              value: theme,
+              text: caption,
+            })
+          );
+        }
```
nwalters512, line 137: I think you can refactor to this:

```suggestion
        themeSelect.append(
          $('<option>', {
            value: theme,
            text: caption,
            selected: localStorage.getItem('pl-file-editor-theme') === theme,
          })
        );
```

louisunlimited:
@nwalters512 Hi! I have updated the 'sync settings' part, now every time the 'save' button is clicked, a custom event will be fired and can be listened to by other editors on the same page to update their theme. I might change the event name to `settingsSync` for readability purposes. However, when updating editor theme real-time when selecting themes(preview), should we make this sync across editors as well? Now only the selected editor changes theme on selection. Personally I don;t think this is much of an issue, but I would love to hear your thoughts!

nwalters512:
@louisunlimited Awesome! I think only changing the current editor as a sort of "preview" is fine. Let me review your latest changes.

elements/pl-file-editor/pl-file-editor.js
```diff
@@ -77,9 +83,25 @@ window.PLFileEditor = function (uuid, options) {
...
 };
 
+window.PLFileEditor.prototype.syncSettings = function () {
+  window.addEventListener('storage', (event) => {
+    if (event.key === 'pl-file-editor-theme') {
+      this.editor.setTheme(event.newValue);
+    }
+  });
+
+  window.addEventListener('storageSync', () => {
```
nwalters512, line 100: Let's namespace this event to this specific element, i.e. name this something like `pl-file-editor-settings-changed`.

elements/pl-file-editor/pl-file-editor.mustache
```diff
@@ -21,8 +21,36 @@ $(function() {
...
+                    <div class="modal-header">
+                        <h5 class="modal-title"  id="modal-{{uuid}}">Editor Settings</h5>
+                        <button type="button" class="close close-settings-button" data-dismiss="modal" aria-label="Close">
+                            <span aria-hidden="true">&times;</span>
+                        </button>
+                    </div>
+                    <div class="modal-body">
+                        <div class="form-group">
+                            <label for="modal-{{uuid}}-themes">Editor Theme</label>
+                            <select class="form-control" id="modal-{{uuid}}-themes">
```
nwalters512, line 43: See https://getbootstrap.com/docs/4.6/components/forms/#select-menu.

```suggestion
                            <select class="custom-select" id="modal-{{uuid}}-themes">
```

nwalters512 (commented):
This is looking great! A few more small things and this will be good to go.

louisunlimited:
All done!

nwalters512 (approved):
Looks great - thanks for the contribution!

PR #6481, opened 2022-10-13
Disallow @ at the beginning of qids

SethPoulsen:
We're doing this now to ensure that we can safely use `@` to indicate using shared questions from outside a course once we get around to implementing question sharing.


SethPoulsen:
Wait, this is in the wrong place. I should be validating when processing the questions, not when processing the assessments.

echuber2:
I think `checkAndRecordQid` is still only used to check the questions mentioned in assessments. Perhaps instead you'd want to edit `loadQuestions`?

SethPoulsen:
Ya, that's what I did, thanks.

sync/course-db.js
```diff
@@ -1273,6 +1273,11 @@ module.exports.loadQuestions = async function (coursePath) {
     validate: validateQuestion,
     recursive: true,
   });
+  for (let qid in questions) {
+    if (qid[0] === '@') {
+      infofile.addError(questions[qid], `Question IDs are not allowed to begin with '@'`);
```
nwalters512, line 1278: For posterity, I think it's important to note that this will still result in a DB row for the given QID. I think this is fine, as a user will never be able to actually *point* to that question from an `infoAssessment.json` file since we'll special-case the handling of QIDs that start with `@` to never refer to questions within the same course.

sync/course-db.js
```diff
@@ -1273,6 +1273,11 @@ module.exports.loadQuestions = async function (coursePath) {
     validate: validateQuestion,
     recursive: true,
   });
+  for (let qid in questions) {
```
nwalters512, line 1276: Can you add a comment explaining why we're forbidding `@...`?

sync/course-db.js
SethPoulsen: huh, will this cause problems then, if the user then edits the QID but not the UUID?

sync/course-db.js
SethPoulsen: sure, it doesn't seem clear what tense it should be in though. 
> Don't allow questions to start with '@', because it will be used to reference questions outside the course.
or
> Don't allow questions to start with '@', because it is used to reference questions outside the course.

If I do the first, we have to come back and edit the comment later, if we do the second then its technically a lie until we push the rest.

nwalters512: I think "will be" is fine, as that's an accurate description at this time.

nwalters512 (approved):

PR #6489, opened 2022-10-14
Attempt to fix R installation

nwalters512:
Before this change, Conda would take ages (and sometimes get OOM-killed) when trying to "solve" the environment. I don't know why this fixes things, but coupled with a larger instance type, we can now build AMIs again!

Inspired by a recommended way of installing documented at https://docs.anaconda.com/anaconda/user-guide/tasks/using-r-language/.

### DIFF
--- a/images/plbase/plbase-install.sh
+++ b/images/plbase/plbase-install.sh
@@ -64,7 +64,7 @@ bash Miniforge3-Linux-${arch}.sh -b -p /usr/local -f
 # If R package installation is specifically disabled, we'll avoid installing anything R-related.
 if [[ "${SKIP_R_PACKAGES}" != "yes" ]]; then
     echo "installing R..."
-    conda install --channel r r-essentials
+    conda install --channel r r-base r-essentials
 
     echo "installing Python packages..."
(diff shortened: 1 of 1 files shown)


mwest1066 (approved):
I don't understand why this works, but sure!

trombonekenny:
R is not to be understood, it is to be experienced.

coatless:
@nwalters512 please use:

```sh
conda install --channel conda-forge r-base 
```

The `conda` channel should be `conda-forge` and not anything with `anaconda`.

Please ping me on the _R_ install hiccups.

nwalters512:
@coatless ack, let me give that a go! Is that documented anywhere in the Conda docs?

coatless:
@nwalters512 nope! It's a community-maintained effort as the support on the `anaconda` default package channels is lacking/problematic. In particular, the binary you were trying to use is stalled out on a lot of platforms:

https://anaconda.org/r/r-base 

(Note: 3.6 badges vs. 4.2 badge.)

`conda-forge` channel is nearly always "up-to-date" and has _many_ more _R_ specific packages in binary form. 

https://conda-forge.org/docs/

PR #5909, opened 2022-06-15
Manual grading: separate points/score for manual and automated grading

jonatanschroeder:
This PR enables the separation of points between manual grading and automated grading. The main purpose of this separation is to enable instructors to have questions with automated results, but with a component that is graded manually, such as:

* calculations that include a section to show work;
* code quality (comments, style, etc.);
* answers that require an explanation.

Tasks:

* [x] database;
* [x] schema;
* [x] sync;
* [x] sync tests;
* [x] sample question in exampleCourse;
* [x] review sprocs/libs that update scores;
* [x] review sprocs/libs that retrieve scores;
* [x] UI for instructor:
    * [x] list of questions;
    * [x] manual grading overview;
    * [x] assessment question;
    * [x] instance question;
    * [x] downloads;
    * [x] file upload;
    * [x] question stats;
    * [x] assessment instance log;
* [x] UI for student:
    * [x] assessment overview;
    * [x] instance question;
* [x] UI tests;
* [x] API entrypoints;
* [x] remove all old references to _in_grading;
* [x] Resolve #5110 
* [x] Resolve #6126
* [x] Resolve #5915 
* [x] Resolve #3191
* [x] Docs
* [x] News item

PR #6494, opened 2022-10-17
Introduce stop-gap for points computation when servers run different versions

jonatanschroeder:
The issue being resolved here is that there are servers running separately with different code, interacting with the same data. This is interfering with the assumptions made in #5909 regarding the values of `manual_points` and `auto_points` (and their derivatives) in multiple tables. This PR introduces UI changes to ensure that any updates made by one version are still consistent with the view of another version, particularly in questions that only use one grading method. In particular, `manual_points` and `auto_points` columns should be COALESCED with a suitable alternative as appropriate.

Once all production servers are running the new version, it should be possible to revert this PR to simplify the code.

### DIFF
--- a/sprocs/instance_questions_points_exam.sql
+++ b/sprocs/instance_questions_points_exam.sql
@@ -24,7 +24,7 @@ BEGIN
     -- exams don't use this, so just copy whatever was there before
     variants_points_list := iq.variants_points_list;
 
-    max_auto_points := COALESCE(aq.max_auto_points, 0);
+    max_auto_points := COALESCE(aq.max_auto_points, aq.max_points);
     max_manual_points := COALESCE(aq.max_manual_points, 0);
 
     -- Update points (instance_question will be closed when number_attempts exceeds bound,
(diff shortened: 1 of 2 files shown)


jonatanschroeder:
One of the main issues happened in this case because, for a few questions running in production (pre-5909) code, the grader was running in new code. This means the grader was assuming that auto/manual points were non-NULL, and making computations on their assumed values. In one particular case, this caused the computed points to be set to the result of the grader (as auto points) plus or minus NULL (as manual points), since the manual points were still null, causing the total points to be set to NULL.

This discrepancy could cause one of two issues:
* The assessment could only have one question, which meant that the code in https://github.com/PrairieLearn/PrairieLearn/blob/0641f0eac6082dd4521d9d818d3bf6fffaf7d053/sprocs/assessment_instances_grade.sql#L77 would just give the student full marks (since `LEAST(NULL, x) = x`);
* The assessment could have multiple questions, which meant that this question would not be part of the assessment points, based on https://github.com/PrairieLearn/PrairieLearn/blob/0641f0eac6082dd4521d9d818d3bf6fffaf7d053/sprocs/assessment_instances_points.sql#L80

jonatanschroeder:
Closing as all servers are now running the new version.

Issue #6495, opened 2022-10-17
MathJax SVG math renderer broken on Windows Chrome


trombonekenny:
A student reported that when he accidentally set the MathJax Math Renderer to SVG the rendering no longer happens and the raw MathML is posted to the screen (with https://www.prairielearn.org/pl/course_instance/4970/instructor/question/1793638/preview):

![image](https://user-images.githubusercontent.com/10765199/196295748-91abb920-969c-411d-8791-1cd82f7fafae.png)

In Windows 10 Chrome, after changing that to SVG, this error is in the Console and nothing is displayed for the math:

```
TypeError: Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'.
    at e.append (core.js:1:2475)
    at t.node (core.js:1:7882)
    at e.html (svg.js:1:20057)
    at r.svg (svg.js:1:118040)
    at r.pageElements (svg.js:1:115678)
    at e.documentPageElements (core.js:1:169324)
    at e.addPageElements (core.js:1:168225)
    at e.updateDocument (core.js:1:168019)
    at t.updateDocument (menu.js:1:31573)
    at Object.renderDoc (core.js:1:15620)
```
On page reload, with LocalStorage `MathJax-Menu-Settings` set to `{"renderer":"SVG"}` shows the raw mathML:

![image](https://user-images.githubusercontent.com/10765199/196296072-79dcbd20-22e7-4b4c-bace-d69ec019586c.png)

and these errors on the console:

```
TypeError: Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'.
    at e.append (core.js:1:2475)
    at t.node (core.js:1:7882)
    at e.html (svg.js:1:20057)
    at r.svg (svg.js:1:118040)
    at r.pageElements (svg.js:1:115678)
    at e.documentPageElements (core.js:1:169324)
    at e.addPageElements (core.js:1:168225)
    at e.updateDocument (core.js:1:168019)
    at t.updateDocument (menu.js:1:31573)
    at Object.renderDoc (core.js:1:15620)
core.js:1 Uncaught (in promise) TypeError: Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'.
    at e.append (core.js:1:2475)
    at t.node (core.js:1:7882)
    at e.html (svg.js:1:20057)
    at r.svg (svg.js:1:118040)
    at r.pageElements (svg.js:1:115678)
[...]

PR #6496, opened 2022-10-18
Change MathJax fontCache from global to local

echuber2:
Maybe closes https://github.com/PrairieLearn/PrairieLearn/issues/6495 (untested on Windows yet -- could be barking up the wrong tree)

This changes the `fontCache` setting for MathJax to `local`, which should create one or several cache(s) within each SVG object instead of one per page.

For background: I'm wondering if the issues @trombonekenny describes in https://github.com/PrairieLearn/PrairieLearn/issues/6495 might be due to the other recent changes to MathJax handling (https://github.com/PrairieLearn/PrairieLearn/pull/6358, https://github.com/PrairieLearn/PrairieLearn/pull/6442) which load MathJax asynchronously, combined with the fact that we have MathJax set to use the _global_ font cache injected on the page. ([See the MathJax developer explanation at the comment linked here.](https://github.com/mathjax/MathJax/issues/2897#issuecomment-1180218087))

On a related note, I see we have this code that tries to manually copy the globally cached parts for `pl-drawing`:

https://github.com/PrairieLearn/PrairieLearn/blob/master/elements/pl-drawing/mechanicsObjects.js#L887

Keeping in mind other weirdness that font rendering has exhibited in `pl-drawing`, and in light of what the code linked in `mechanicsObjects.js` does, maybe we should just set `fontCache` to `'none'` instead, which I think does the same thing natively, inlining each shape. ([MathJax doc reference](https://docs.mathjax.org/en/latest/options/output/svg.html#option-descriptions); see also the MathJax developer's explanation linked above.)

### DIFF
--- a/pages/partials/mathjax.ejs
+++ b/pages/partials/mathjax.ejs
@@ -21,7 +21,7 @@ var MathJax = {
         inlineMath: [['$', '$'], ['\\(', '\\)']]
     },
     svg: {
-        fontCache: 'global'
+        fontCache: 'local'
     },
     loader: {
         load: ['input/tex', 'ui/menu', outputComponent]


nwalters512:
I can confirm that this fixes things for me on macOS Chrome!

Have you or @trombonekenny made any effort to reproduce this in isolation? This very much feels like a bug to me; regardless of how we have the font cache configured, I'd expect switching the rendering method to still work. It'd be great if one of you could get an isolated reproduction and submit an issue to the MathJax folks.

> maybe we should just set fontCache to 'none' instead, which I think does the same thing natively, inlining each shape

I think this is reasonable as long as we continue rendering things only on the client; I don't think we need to worry about the complexity or number of DOM nodes that MathJax creates. And at any rate, if we move to rendering on the server, we'd probably not use MathJax anyways.

nwalters512:
> What do you mean by isolation?

As in, in an HTML document outside of PrairieLearn. See the MathJax bug report template: https://github.com/mathjax/MathJax/blob/master/.github/ISSUE_TEMPLATE/bug_report.md

echuber2:
I'm trying to put that together. I think indeed there may be some bugs in how MathJax handles multiple loaders and the global cache option. But for our case the fastest thing to do may be to just use `none`.

nwalters512:
Relevant: https://github.com/PrairieLearn/PrairieLearn/issues/6495#issuecomment-1286303715

echuber2:
Superseded by https://github.com/PrairieLearn/PrairieLearn/pull/7280 and/or https://github.com/PrairieLearn/PrairieLearn/pull/6974

PR #6500, opened 2022-10-18
Update status from saved to incorrect in homeworks upon grading

jonatanschroeder:
The instance question status indicates how the grading of the instance question is applied. In particular, I believe the intention is:

* unanswered: no submission;
* saved: submission saved, but not graded;
* invalid: submission saved, but parsing failed;
* grading: submission saved and grading requested but still waiting for completion (external grading);
* correct/incorrect: submission saved and graded, and more submissions allowed;
* complete: submission saved and graded, and no more submissions allowed (exams).

These are not properly handled right now. A homework question has the status "saved" until it is correct (i.e., "incorrect" is not used). Exams handle this properly.

Also, the "grading" status is never used (from what I could tell), the status in place while grading is actually "saved".

This PR updates the homework processing to set the status to "incorrect" if it was graded and not correct. It does not change the grading status.

### DIFF
--- a/migrations/20221018160700_instance_questions__status__update.sql
+++ b/migrations/20221018160700_instance_questions__status__update.sql
@@ -1,0 +1,10 @@
+UPDATE instance_questions iq
+SET status = 'incorrect'
+WHERE iq.status = 'saved'
+      AND CARDINALITY(iq.variants_points_list) > 0
+      AND (SELECT s.graded_at
+           FROM variants v LEFT JOIN submissions s ON (s.variant_id = v.id)
+           WHERE v.instance_question_id = iq.id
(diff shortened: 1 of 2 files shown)


jonatanschroeder:
This of course only affects new grading events after this is deployed. One solution for code that uses this status (e.g., the points component in the grading panel, which shows "pending" if the status is "saved") is to use "`variants_points_list`" as a heuristic for this value, since it's updated on grading of homeworks, though it's not 100% reliable: if a new submission is saved but not graded after a previously graded submission, the `variants_points_list` will have a value, but the status is expected to be "saved".

mwest1066 (approved):

PR #6506, opened 2022-10-19
Improve load testing functionality

nwalters512:
This PR allows us to accurately simulate large numbers of different users by requiring the `load_test_token` cookie to specify a UUID. This will be used in a UID (`loadtest+UUID@prairielearn.com`) and UIN (`loadtest+UUID`)

Part of https://github.com/PrairieLearnInc/sysconf/issues/433.


lib/config.js
```diff
@@ -17,7 +17,7 @@ config.postgresqlUser = 'postgres';
 config.postgresqlPassword = null;
 config.postgresqlDatabase = 'postgres';
 config.postgresqlHost = 'localhost';
-config.postgresqlPoolSize = 100;
+config.postgresqlPoolSize = 50;
```
nwalters512, line 20: My local Postgres was erroring with something like "sorry, there are already too many clients" when I saturated PL with requests. This fixed it. Idk if we actually want to change the default though? This mostly got committed on accident 😅

middlewares/authn.js
```diff
@@ -32,46 +35,42 @@ module.exports = function (req, res, next) {
 
   // look for load-testing override cookie
   if (req.cookies.load_test_token) {
-    if (
-      !csrf.checkToken(req.cookies.load_test_token, 'load_test', config.secretKey, {
-        maxAge: 24 * 60 * 60 * 1000,
-      })
-    ) {
+    const data = csrf.getCheckedData(req.cookies.load_test_token, config.secretKey, {
```
nwalters512, line 38: As usual, I'd appreciate an extra-close look at auth code.

middlewares/authn.js
```diff
@@ -32,46 +35,42 @@ module.exports = function (req, res, next) {
...
-            if (ERR(err, callback)) return;
-            callback(null);
+          res.locals.authn_user = userResult.rows[0].user;
+          res.locals.authn_institution = userResult.rows[0].institution;
+          res.locals.authn_provider_name = 'LoadTest';
+          res.locals.authn_is_administrator = userResult.rows[0].is_administrator;
+          checkAdministratorAccess(req, res);
+
+          // Enroll the load test user in the example course.
+          await sqldb.queryAsync(sql.enroll_user_in_example_course, {
```
nwalters512, line 72: I'm hoping this extra query won't actually cause load tests to behave differently than real users 🤞

middlewares/authn.js
```diff
@@ -80,6 +79,8 @@ module.exports = function (req, res, next) {
         return next();
       }
     );
+
+    return;
```
nwalters512, line 83: I spent like an hour trying to figure out why I was getting "header already sent" errors 🤦

middlewares/authn.sql
```diff
@@ -23,22 +23,8 @@ INSERT INTO enrollments
         course_instances AS ci
         JOIN pl_courses AS c ON (c.id = ci.course_id)
     WHERE
-        u.uid = $uid
-        AND c.short_name = $course_short_name
+        u.user_id = $user_id
+        AND c.example_course IS TRUE
```
nwalters512, line 27: This feels safer than relying on a course name?

pages/home/home.sql
```diff
@@ -120,7 +120,7 @@ WITH
             JOIN pl_courses AS c ON (
                 c.id = ci.course_id
                 AND c.deleted_at IS NULL
-                AND c.example_course IS FALSE
+                -- AND c.example_course IS FALSE
```
nwalters512, line 123: For discussion: I wanted to be able to simulate a user clicking on their course from the home page, but this filter prevented me from doing that (at least without hardcoding a course instance ID). I believe removing this will be safe, as we have code on the enrollment page that should prevent anyone from actually enrolling in the example course. Only my forceful backend enrollments in the authn middleware should ever be enrolling users in the example course.

tools/load-test.js
```diff
@@ -1,336 +0,0 @@
-/* eslint no-console: 0 */ // allow console.log() in this file
```
nwalters512, line 1: I opted for a fresh start in our internal sysconf repo (mainly so I can easily get all our nice TS tooling). I'd be happy to move that over to the main repo at some point, but for now I'm going to remove this, as it won't work without changes.

middlewares/authn.js
```diff
@@ -32,46 +35,42 @@ module.exports = function (req, res, next) {
...
-    if (
-      !csrf.checkToken(req.cookies.load_test_token, 'load_test', config.secretKey, {
-        maxAge: 24 * 60 * 60 * 1000,
-      })
-    ) {
+    const data = csrf.getCheckedData(req.cookies.load_test_token, config.secretKey, {
+      maxAge: 24 * 60 * 60 * 1000,
+    });
+
+    if (!data || !data.uuid || typeof data.uuid !== 'string' || !data.uuid.match(UUID_REGEXP)) {
```
nwalters512, line 42: I opted not to retain backwards compatibility with the previous use case where we default to `loadtest@prairielearn.org`.

middlewares/authn.js
```diff
@@ -8,6 +8,9 @@ var sqlLoader = require('../prairielib/lib/sql-loader');
 
 var sql = sqlLoader.loadSqlEquiv(__filename);
 
+const UUID_REGEXP =
+  /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/;
```
mwest1066, line 12: Nit: I don't think the `\b` entries are doing anything here, given that we've precisely specified which characters must come before and after the `\b`.

middlewares/authn.js
mwest1066: Looking carefully...

middlewares/authn.js
mwest1066: This was an existing error? Sorry!!

middlewares/authn.sql
mwest1066: Yep, although it has the downside that it will enroll into all example courses if there is more than one (there is only one at the moment). I don't think this is a problem.

mwest1066: And of course the `IS TRUE` is not needed, but it's not doing any harm.

middlewares/authn.js
mwest1066: It seems very lightweight and I can't see that it will have any noticeable impact.

middlewares/authn.sql
```diff
@@ -23,22 +23,8 @@ INSERT INTO enrollments
...
-    FROM
-        users AS u,
-        pl_courses AS c
-    WHERE
-        u.uid = $uid
-        AND c.short_name = $course_short_name
-)
-ON CONFLICT DO UPDATE SET course_role = 'Previewer'::enum_course_role;
+ON CONFLICT DO NOTHING
+RETURNING *;
```
mwest1066, line 30: Not that it matters, but we don't appear to actually use the return value so we could skip this final line.

pages/home/home.sql
mwest1066: Unfortunately we have a lot of users who are already enrolled in the example course, because they accessed it as an instructor but then did "View as student". Showing XC 101 on their homepage will just be clutter and they won't be able to remove it (because of the enrollment page restrictions).

As a slightly different horrible hack, how about adding a query param to the homepage to request viewing of the example course (if we are enrolled in it). That is, we'd change the above to:
```
AND (c.example_course IS FALSE OR $show_example_course)
```

mwest1066 (approved):

lib/config.js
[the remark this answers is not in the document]
nwalters512: I added a self-comment above:

> My local Postgres was erroring with something like "sorry, there are already too many clients" when I saturated PL with requests. This fixed it. Idk if we actually want to change the default though? This mostly got committed on accident 😅

middlewares/authn.js
nwalters512: Fair enough, I just copied this from a "uuid regexp" Google search result 😂

middlewares/authn.sql
nwalters512: Good catch, I added this just to do some debugging. I'll remove it.

pages/home/home.sql
nwalters512: Oh huh - "view as student" actually inserts an enrollment?

I can certainly do this with a query param, yeah.

middlewares/authn.js
nwalters512: Sadly, yes. This was a regression introduced way back in #3020:

https://github.com/PrairieLearn/PrairieLearn/pull/3020/files#diff-bfa663cd4ce127397d92741afe999bdfe7f72f95a4cbbaaae20d907ea6522449L60

pages/home/home.sql
nwalters512: I implemented support for an `?include_example_course_enrollments=true` in 8aced71. It's kinda hacky, but it's a small and well-documented amount of code.

lib/config.js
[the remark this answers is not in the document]
nwalters512: Reverted in 3f127d7.

middlewares/authn.js
mwest1066: Oh dear. Ah well 😄

tools/load-test.js
mwest1066: Yep, I'm happy to have this deleted.

middlewares/authn.sql
nwalters512: Removed in 24ff048.

middlewares/authn.js
nwalters512: Simplified in ca49a3a.

PR #6516, opened 2022-10-21
Add raw submitted answer to assessment instance log

jonatanschroeder:
Based on discussion on Slack. Given the raw submitted answer for questions with files will contain the base64 content of the file for a key like `_file_upload_XXX` or `_file_editor_XXX`, keys that start with a `_` are removed for cases where files should not be included.

### DIFF
--- a/sprocs/assessment_instances_select_log.sql
+++ b/sprocs/assessment_instances_select_log.sql
@@ -125,7 +125,8 @@ BEGIN
                     jsonb_build_object(
                       'submitted_answer', CASE WHEN include_files THEN s.submitted_answer ELSE (s.submitted_answer - '_files') END,
                       'raw_submitted_answer', CASE WHEN include_files THEN s.raw_submitted_answer
-                      ELSE (SELECT JSONB_OBJECT_AGG(key, value) FROM JSONB_EACH(s.raw_submitted_answer) WHERE key !~ '^_') END,
+                      -- Elements that produce files (upload, editor, etc.) will use keys like '_file_upload_XXX' or equivalent
+                      ELSE (SELECT JSONB_OBJECT_AGG(key, value) FROM JSONB_EACH(s.raw_submitted_answer) WHERE STARTS_WITH(key, '_')) END,
                       'correct', s.correct
                     ) AS data
                 FROM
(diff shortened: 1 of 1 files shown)


sprocs/assessment_instances_select_log.sql
```diff
@@ -124,6 +124,8 @@ BEGIN
                     s.id AS log_id,
                     jsonb_build_object(
                       'submitted_answer', CASE WHEN include_files THEN s.submitted_answer ELSE (s.submitted_answer - '_files') END,
+                      'raw_submitted_answer', CASE WHEN include_files THEN s.raw_submitted_answer
+                      ELSE (SELECT JSONB_OBJECT_AGG(key, value) FROM JSONB_EACH(s.raw_submitted_answer) WHERE key !~ '^_') END,
```
mwest1066, line 128: Nit: I think it would be more efficient to use `starts_with()` (or the completely unintuitive `^@` operator) instead of a regex here.

mwest1066 (approved):

sprocs/assessment_instances_select_log.sql
jonatanschroeder: Done. Do you think I should do the same treatment for `submitted_answer`, since it will (for elements like `pl-file-editor`) retain that same raw input?

mwest1066: Sorry, I'm not quite following here. What extra raw input is retained in `submitted_answer`?

jonatanschroeder: image

Note that this is `submitted_answer`. `pl-file-editor` (and `pl-rich-text-editor` for that matter) transfer this content to the `_files` object, but also keep it in the original raw form. `pl-file-upload` removes this entry.

https://github.com/PrairieLearn/PrairieLearn/blob/1ab3a6e7accfc4ff4245198b794a3cf19cff6e3b/elements/pl-file-upload/pl-file-upload.py#L82-L85

Though another approach would be, of course, to remove this entry in these other elements as well...

jonatanschroeder: Sounds good. The elements rely on this for rendering now, I'll add an issue about it.

PR #6533, opened 2022-10-25
Fix error handling in ecrUpdate sync helper

nwalters512:
Without this, servers crash when they fail to sync images. Best reviewed with whitespace changes ignored..

Yet another reason to move towards `async`/`await`!

### DIFF
--- a/pages/shared/syncHelpers.js
+++ b/pages/shared/syncHelpers.js
@@ -326,8 +326,6 @@ module.exports.ecrUpdate = function (locals, callback) {
       async.eachOfSeries(
         locals.images || [],
         (image, index, callback) => {
-          if (ERR(err, callback)) return;
-
           var jobOptions = {
             course_id: locals.course ? locals.course.id : null,
             type: 'image_sync',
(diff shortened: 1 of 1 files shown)


mwest1066 (approved):

pages/shared/syncHelpers.js
```diff
@@ -323,40 +323,47 @@ module.exports.ecrUpdate = function (locals, callback) {
...
-
-        serverJobs.createJob(jobOptions, (err, job) => {
-          if (err) {
-            logger.error('Error in createJob()', err);
-            serverJobs.failJobSequence(job_sequence_id);
-            return callback(err);
+      async.eachOfSeries(
+        locals.images || [],
+        (image, index, callback) => {
+          if (ERR(err, callback)) return;
```
mwest1066, line 329: This seems unnecessary?

mwest1066 (approved):
Good catch on the other `failJobSequence()`

pages/shared/syncHelpers.js
nwalters512: Good catch, addressed in 78057a8.

mwest1066: lgtm

PR #6538, opened 2022-10-27
fix: pl-order-blocks crashing in rare circumstances

### LINKED ISSUE #6537: pl-order-blocks: Crash in rare cases
I've just discovered that the pl-order-blocks element grader will crash under the very rare circumstances in which: 

- the problem is using the `dag` grading mode, with at least one `pl-block-group` in it
- the problem has disctractor blocks
- the problem is using `lcs` based partial credit
- the student submits an answer in which they placed a distractor block _in between_ two blocks that are in the same block group in a correct solution

This bug has been in production for ~9 Months, ever since lcs partial credit was implemented for block groups (https://github.com/PrairieLearn/PrairieLearn/commit/39ceb5267182c5f9ffdf5f3012922b209e98dc6a). 

In that time, I have never seen it reported, ever (its possible that some student hit the error, but the instructor never forwarded the report on to me). 

The only reason I found the issue is because I was benchmarking the partial credit algorithm over thousands of old student submissions for the next version of the grading algorithm paper (https://arxiv.org/abs/2204.04196)

this is the error message: 
```
  File "/PrairieLearn/python/zygote.py", line 211, in <module>
    worker_loop()
  File "/PrairieLearn/python/zygote.py", line 142, in worker_loop
    val = method(*args)
  File "/PrairieLearn/elements/pl-order-blocks/pl-order-blocks.py", line 484, in grade
    edit_distance = lcs_partial_credit(submission, depends_graph, group_belonging)
  File "/PrairieLearn/elements/pl-order-blocks/dag_checker.py", line 197, in lcs_partial_credit
    if not all([group_belonging[x] == group_belonging[node1] for x in submission[i:j + 1]]):
  File "/PrairieLearn/elements/pl-order-blocks/dag_checker.py", line 197, in <listcomp>
    if not all([group_belonging[x] == group_belonging[node1] for x in submission[i:j + 1]]):
KeyError: None
  File "/PrairieLearn/python/zygote.py", line 248, in <module>
```

There is an easy fix, I'll open a PR shortly.

SethPoulsen:
fix #6537

### DIFF
--- a/elements/pl-order-blocks/dag_checker_test.py
+++ b/elements/pl-order-blocks/dag_checker_test.py
@@ -32,6 +32,7 @@ problem_2_submissions = [
     ['1', '2', '3', '7'],
     ['1', '5', '6', '2', '3'],
     ['1', '6', '5', '2', '3'],
+    ['2', None, '3']
 ]
 problem_2_expected = [
     7,
@@ -40,7 +41,8 @@ problem_2_expected = [
     2,
     3,
     1,
-    1
+    1,
+    0
 ]
 problem_2_expected_ed_groups = [
     0,
@@ -50,6 +52,7 @@ problem_2_expected_ed_groups = [
     3,
     4,
     6,
+    6
 ]
 problem_2_dag_no_groups = {'1': [], '2': ['1'], '3': ['2'], '4': [], '5': ['4'], '6': ['5'], '7': ['3', '6']}
 problem_2_expected_ed_no_groups = [
@@ -60,6 +63,7 @@ problem_2_expected_ed_no_groups = [
     3,
     2,
     4,
+    6
 ]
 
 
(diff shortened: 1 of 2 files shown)


elements/pl-order-blocks/dag_checker_test.py
```diff
@@ -32,6 +32,7 @@
     ['1', '2', '3', '7'],
     ['1', '5', '6', '2', '3'],
     ['1', '6', '5', '2', '3'],
+    ['2', None, '3']
```
nwalters512, line 35: Thanks for adding a test! I assume this would have failed before this PR?

nwalters512 (approved):
I don't know that I understand either the problem or the solution, but if the tests pass and you're happy with it, :shipit:

elements/pl-order-blocks/dag_checker_test.py
SethPoulsen: Right.

Issue #6539, opened 2022-10-27
Bug report: message for duplicate UUID is an error


firasm:
In the past when I duplicate a question locally and it has a duplicate UUID, PL just gently warns me about it at every sync.

This time I had a duplicate UUID I got an error message when I tried to sync"

[Here's a link to the error message](https://ca.prairielearn.com/pl/course_instance/2478/instructor/jobSequence/66849) and the useful part (`"detail": "Key (course_id, uuid)=(54, 0e8097aa-b554-4908-9eac-d46a78d6c249) already exists.",`):

```
Trying lock coursedir:/data1/courses/pl-ubc-data301
Acquired lock coursedir:/data1/courses/pl-ubc-data301
Loading info.json files from course repository
Syncing info to database
error: duplicate key value violates unique constraint "questions_course_id_number_nondeleted_key"
Async Stacktrace:
    at Object.<anonymous> (/srv/sysconf/ca/main/revisions/0641f0eac6082dd4521d9d818d3bf6fffaf7d053/sync/syncFromDisk.js:136:15)

error: duplicate key value violates unique constraint "questions_course_id_number_nondeleted_key"
    at Parser.parseErrorMessage (/srv/sysconf/ca/main/revisions/0641f0eac6082dd4521d9d818d3bf6fffaf7d053/node_modules/pg-protocol/dist/parser.js:287:98)
    at Parser.handlePacket (/srv/sysconf/ca/main/revisions/0641f0eac6082dd4521d9d818d3bf6fffaf7d053/node_modules/pg-protocol/dist/parser.js:126:29)
    at Parser.parse (/srv/sysconf/ca/main/revisions/0641f0eac6082dd4521d9d818d3bf6fffaf7d053/node_modules/pg-protocol/dist/parser.js:39:38)
    at Socket.<anonymous> (/srv/sysconf/ca/main/revisions/0641f0eac6082dd4521d9d818d3bf6fffaf7d053/node_modules/pg-protocol/dist/index.js:11:42)
    at Socket.emit (node:events:527:28)
    at Socket.emit (node:domain:475:12)
    at addChunk (node:internal/streams/readable:315:12)
    at readableAddChunk (node:internal/streams/readable:289:9)
    at Socket.Readable.push (node:internal/streams/readable:228:10)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
    at TCP.callbackTrampoline (node:internal/async_hooks:130:17)
{
    "sqlError": {
[...]

PR #6540, opened 2022-10-27
Fix sync error in the face of renames and duplicate UUIDs

nwalters512:
Fixes https://github.com/PrairieLearn/PrairieLearn/issues/6539. See inline comments for details on the fix itself! Since we use essentially the same syncing algorithm for questions, assessments, and course instances, I had to make the fix in all three places.

### DIFF
--- a/tests/sync/questionsSync.test.js
+++ b/tests/sync/questionsSync.test.js
@@ -369,18 +369,18 @@ describe('Question syncing', () => {
     // Write and sync a single question.
     const originalQuestion = makeQuestion(courseData);
     originalQuestion.uuid = '0e8097aa-b554-4908-9eac-d46a78d6c249';
-    courseData.questions['templates/simple_randomized_mcq'] = originalQuestion;
+    courseData.questions['a'] = originalQuestion;
     const courseDir = await util.writeAndSyncCourseData(courseData);
 
     // Now "move" the above question to a new directory AND add another with the
     // same UUID.
-    delete courseData.questions['templates/simple_randomized_mcq'];
-    courseData.questions['firas_templates/simple_randomized_mcq'] = originalQuestion;
-    courseData.questions['firas_test/modulo_variable'] = originalQuestion;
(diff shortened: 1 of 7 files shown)


tests/sync/questionsSync.test.js
```diff
@@ -358,4 +358,26 @@ describe('Question syncing', () => {
     assert.equal(deletedQuestion.uuid, originalQuestion.uuid);
     assert.equal(deletedQuestion.sync_errors, null);
   });
+
+  it.only('reproduces problem Firas reported', async () => {
```
nwalters512, line 362: 👀 drop `.only` before merging.

tests/sync/util.js
```diff
@@ -269,6 +269,7 @@ module.exports.writeCourseToDirectory = async function (courseData, coursePath)
 module.exports.QUESTION_ID = 'test';
 module.exports.ALTERNATIVE_QUESTION_ID = 'test2';
 module.exports.MANUAL_GRADING_QUESTION_ID = 'test_manual';
+module.exports.WORKSPACE_QUESTION_ID = 'workspace';
```
nwalters512, line 272: This was missing; I only noticed after I saw a question with the qid `undefined` while debugging some tests.

tests/sync/util.js
```diff
@@ -72,6 +72,16 @@ const syncFromDisk = require('../../sync/syncFromDisk');
  * @property {"Set" | "Module"} [groupAssessmentsBy]
  */
 
+/**
```
nwalters512, line 75: I turned on type checking for `tests/sync/assessmentsSync.test.js`, which necessitated updating a lot of things here.

mwest1066: Nice!

mwest1066 (approved):
Wow, such a sneaky bug! Thanks for the clear explanations in the linked issue and in the new comments in the code. This makes total sense with the explanation.

PR #6544, opened 2022-10-28
Block file attachment submission if no file selected

nwalters512:
This PR solves the following Sentry error: https://sentry.io/organizations/prairielearn-inc/issues/3581270096/?project=6704263

The specific error is `Cannot read properties of undefined (reading 'originalname')`, coming from here:

https://github.com/PrairieLearn/PrairieLearn/blob/f2ee0e4280da6a83e36f271df6db5342c3154408/pages/shared/studentInstanceQuestion.js#L35

For good measure, I'm validating both on the client (don't allow submission if a file hasn't been selected) and on the server (respond with a 400 if no file was attached).

### DIFF
--- a/pages/partials/attachFilePanel.ejs
+++ b/pages/partials/attachFilePanel.ejs
@@ -51,52 +51,67 @@
     </div>
     <% } else { /* assessment_instance is open and authorized_edit is true */ %>
     <!-- UPLOAD FILE -->
-    <button class="btn btn-xs btn-secondary" type="button" data-toggle="collapse" data-target="#attachFileCollapse" aria-expanded="false" aria-controls="attachFileCollapse">
-      Attach a file <i class="far fa-caret-square-down"></i>
-    </button>
-    <div class="collapse" id="attachFileCollapse">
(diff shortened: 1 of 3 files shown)


pages/partials/attachFilePanel.ejs
```diff
@@ -51,52 +51,67 @@
...
-      <form class="attach-file-form" name="attach-file-form" method="POST" enctype="multipart/form-data" class="mb-3">
-        <p class="small mt-3">
-          Attached files will be saved here for your reference.
-        </p>
-        <div class="form-group">
-          <div class="custom-file">
-            <input type="file" name="file" class="custom-file-input" id="attachFileInput">
-            <label class="custom-file-label" for="attachFileInput">Choose file</label>
-            <small id="emailHelp" class="form-text text-muted">Max file size: <%= config.fileUploadMaxBytesFormatted %></small>
+    <div>
```
nwalters512, line 54: Recommended to review with whitespace changes hidden. While I was here, I changed from a `<br>` to wrapping each section in a `<div>` to avoid content jumping around when the panels are expanding and collapsing.

mwest1066 (approved):

PR #6545, opened 2022-10-31
Wait for code callers to boot up before sending calls to them

nwalters512:
We've observed intermittent timeouts under periods of high load. When a timeout occurs, we kill the child process and start a new one. However, because we preload a number of large Python libraries, starting a new worker can take a long time. Previously, our calling code wasn't aware of this. We believe this was capable of causing cascading failures where we'd send a call to a worker while it was still booting up, and that call would time out, leading to another restart, and so on.

This PR allows the host process to "ping" the child when it's booting up for the first time to ensure that it's ready to accept calls. This ping is modeled as a normal call; this means that when a worker is able to respond to a ping, it'll be able to respond to question/element calls too.

I originally started going down the path of having the worker reach out when it finished initializing, but getting that working with our current state machine turned out to be really complex. By modeling this check as a call into the worker, we reuse all the existing code, and it ends up being a very simple implementation!

### DIFF
--- a/executor.js
+++ b/executor.js
@@ -98,55 +98,42 @@ async function handleInput(line, codeCaller) {
   };
 }
 
-// Our overall loop looks like this: read a line of input from stdin, spin
-// off a python worker to handle it, and write the results back to stdout.
-const rl = readline.createInterface({
-  input: process.stdin,
-  output: process.stdout,
-  terminal: false,
-});
-
 let questionTimeoutMilliseconds = Number.parseInt(process.env.QUESTION_TIMEOUT_MILLISECONDS);
 if (Number.isNaN(questionTimeoutMilliseconds)) {
   questionTimeoutMilliseconds = 10000;
 }
 
-let codeCaller = new CodeCallerNative({
-  dropPrivileges: true,
-  questionTimeoutMilliseconds,
-  errorLogger: console.error,
-});
-codeCaller.ensureChild();
+(async () => {
+  let codeCaller = new CodeCallerNative({
+    dropPrivileges: true,
+    questionTimeoutMilliseconds,
+    errorLogger: console.error,
+  });
+  await codeCaller.ensureChild();
 
-// Safety check: if we receive more input while handling another request,
-// discard it.
-let processingRequest = false;
-rl.on('line', (line) => {
-  if (processingRequest) {
-    // Someone else messed up - fail fast.
-    process.exit(1);
+  // Our overall loop looks like this: read a line of input from stdin, spin
+  // off a python worker to handle it, and write the results back to stdout.
+  const rl = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout,
+    terminal: false,
+  });
+
+  // Once the readline interface closes, we can't get any more input; die
+  // immediately to allow our container to be removed.
+  rl.on('close', () => process.exit(0));
+
+  for await (const line of rl) {
+    const results = await handleInput(line, codeCaller);
+    const { needsFullRestart, ...rest } = results;
+    console.log(JSON.stringify(rest));
+    if (needsFullRestart) {
+      codeCaller.done();
+      codeCaller = new CodeCallerNative();
+      await codeCaller.ensureChild();
+    }
   }
-
-  processingRequest = true;
-  handleInput(line, codeCaller)
-    .then((results) => {
-      const { needsFullRestart, ...rest } = results;
-      if (needsFullRestart) {
-        codeCaller.done();
-        codeCaller = new CodeCallerNative();
-        codeCaller.ensureChild();
-      }
-      console.log(JSON.stringify(rest));
-      processingRequest = false;
-    })
-    .catch((err) => {
-      console.error(err);
-      processingRequest = false;
-    });
-});
-
-rl.on('close', () => {
-  // We can't get any more input; die immediately to allow our container
-  // to be removed.
-  process.exit(0);
+})().catch((err) => {
+  console.error(err);
+  process.exit(1);
 });
--- a/lib/code-caller/code-caller-container.js
+++ b/lib/code-caller/code-caller-container.js
@@ -195,7 +195,6 @@ class CodeCallerContainer {
 
     this.coursePath = coursePath;
 
-    await this._ensureChild();
     // If this Docker code caller was used before, we might have an existing
     // bind mount that we need to remove first. This should have been cleaned
     // up before, but if it wasn't, we'll do so here to avoid a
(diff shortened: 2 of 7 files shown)


nwalters512:
Just realized I'll also have to make changes to `code-caller-container` to ping at startup. I'll get to that tomorrow.

mwest1066 (approved):
This all looks good (with the coming implementation in `code-caller-container`).

lib/code-caller/code-caller-container.js
```diff
@@ -319,6 +316,25 @@ class CodeCallerContainer {
...
+    await this.call('ping', null, null, 'ping', [], { timeout: 60 * 1000 });
+
+    this._checkState();
+    this.debug('exit ensureChild()');
+  }
+
+  /**
+   * Ensures that this code caller has a container.
+   */
+  async _ensureChild() {
```
mwest1066, line 334: Rather than having `_ensureChild()` and `ensureChild()` could we just have a single `ensureChild()` function and do the ping after `_createAndAttachContainer()`?

lib/code-caller/code-caller-native.js
```diff
@@ -280,6 +283,12 @@ class CodeCallerNative {
       this._startChild();
     }
 
+    // Give the child a reasonably long amount of time to start up. Under
+    // periods of high load, it should be cheaper to wait for a child to
+    // start slowly than it would be to time out too quickly and try to
+    // launch a new one to replace it.
+    await this.call('ping', null, null, 'ping', [], { timeout: 30 * 1000 });
```
mwest1066, line 290: Could we shift this up to be inside the `this.state === CREATED` block, just after `_startChild()`?

executor.js
```diff
@@ -98,55 +98,42 @@ async function handleInput(line, codeCaller) {
...
+  // immediately to allow our container to be removed.
+  rl.on('close', () => process.exit(0));
 
-rl.on('close', () => {
-  // We can't get any more input; die immediately to allow our container
-  // to be removed.
-  process.exit(0);
+  for await (const line of rl) {
+    const results = await handleInput(line, codeCaller);
+    const { needsFullRestart, ...rest } = results;
```
mwest1066, line 128: Did you deliberately move this above the `needsFullRestart` handler? Any chance of a race condition here?

mwest1066 (approved):

mwest1066:
> I originally started going down the path of having the worker reach out when it finished initializing, but getting that working with our current state machine turned out to be really complex. By modeling this check as a call into the worker, we reuse all the existing code, and it ends up being a very simple implementation!

I like this a lot!

lib/code-caller/code-caller-container.js
nwalters512: I don't think so, unfortunately. Because `call()` needs to call some form of `ensureChild()` as it's booting up, we need to differentiate "just create the container" and "ensure the container and health-check it". Otherwise, `ensureChild()` would call `call('ping')`, which would then call `ensureChild()` again 😬 

I suppose we could have `call()` skip the call to `ensureChild()` if it's invoked with `ping`?

nwalters512: Alternatively, now that the pool of code callers always invokes `ensureChild()` after it creates a worker and before it leases it out, we could probably drop the internal calls to `ensureChild`? For that matter, I think `prepareForCourse` always calls `ensureChild()`, so a call to `ensureChild()` inside `call()` shouldn't be necessary for at least those two reasons.

mwest1066: Oh right, I was forgetting that `ping` calls `call()`! I would prefer a `doNotPing` argument to `ensureChild()` rather than having two versions (`_ensureChild()` and `ensureChild()`).

Remind me how the state machine works here -- will `ensureChild()` ever recreate a new child (maybe the child died?), or is this a one-and-done thing the first time we create a child? If the latter then I agree that dropping all the internal `ensureChild()` calls would be the cleanest (it would actually clean up the code in general) and we could just unconditionally create a child on initialization.

nwalters512: I changed `code-caller-container` to no longer use an internal call to `ensureChild()` in b828127. I made a similar change to `code-caller-native` in 1334a5e; `_startChild()` is no longer called from `call()`, and the caller must be in state `WAITING` when a call is made. This is consistent with how `code-caller-container` works.

lib/code-caller/code-caller-native.js
nwalters512: Done in 79d4ed1.

executor.js
nwalters512: This was deliberate; writing data back to stdout helps ensure that the parent process gets useful information instead of a potential timeout due to `ensureChild()` taking a long time.

I don't believe there's a chance for a race condition; is there a specific scenario you're asking about?

lib/code-caller/code-caller-container.js
nwalters512: As for how the state machine works, we can never transition from `EXITING` or `EXITED` back to `CREATED` or `WAITING`; as you put it, we'll never recreate the container or child process. If we need a new container or child process, the pool will detect that and create an entirely new `CodeCaller{Native,Container}` object.

mwest1066: Great! This feels much cleaner.

executor.js
mwest1066: I was thinking that after we write to stdout this signals the caller that the call has completed, so they might write a new line to our stdin, which might then trigger us to do something while the full restart is only partly completed? I have not thought this through with any care so this was just a general question to make sure you had thought about it 😄

executor.js
```diff
@@ -98,55 +98,42 @@ async function handleInput(line, codeCaller) {
...
-  errorLogger: console.error,
-});
-codeCaller.ensureChild();
+(async () => {
+  let codeCaller = new CodeCallerNative({
+    dropPrivileges: true,
+    questionTimeoutMilliseconds,
+    errorLogger: console.error,
+  });
+  await codeCaller.ensureChild();
```
mwest1066, line 112: Not that it really matters, but under the new scheme could we move `ensureChild()` inside the constructor to make sure we never forget it? Can we even have async constructors?

mwest1066 (approved):

executor.js
nwalters512: The rewrite to use an async iterator should prevent against that; we'll only repeat the loop once the `ensureChild()` call completes. Async iterators are really nice in that they have normal loop semantics, i.e. it'll never run the body of your loop multiple times concurrently 😛

executor.js
nwalters512: We cannot have async constructors 😢 it would be very nice if we could (that could simplify some of the code caller pool code), but I think this is the best we can do.

mwest1066: Yep, I've just been reading about all the hideous hacks that people have used to partially emulate async constructors and I agree we don't want to go there!

executor.js
mwest1066: Excellent point, thanks for clarifying!

PR #6601, opened 2022-11-02
Print sync warnings/errors after writing to database

### LINKED ISSUE #6546: Sync warnings are not listed during sync process
Some sync warnings are not shown during the sync process, even though they trigger a sync warning shown on the exam setting.

image
https://ca.prairielearn.com/pl/course_instance/2450/instructor/jobSequence/67500

In the example above the sync warning was for an access rule with an exam UUID that didn't exist on PrairieTest. Finding the offending exam was not trivial as it was in an assessment for a previous term, and this warning is not shown in the local environment.

nwalters512:
Fixes #6546. Note that if writing to the database itself fails, we won't print warnings and errors because an error will have been thrown. I think this is acceptable, as syncing is designed to be extremely resilient and shouldn't ever actually fail. If it does (like in #6540), we have a bigger problem.

Here's how the syncing output looks now:

Screenshot 2022-11-02 at 16 25 11

### DIFF
--- a/sync/syncFromDisk.js
+++ b/sync/syncFromDisk.js
@@ -45,10 +45,6 @@ async function syncDiskToSqlWithLock(courseDir, courseId, logger) {
   const courseData = await perf.timedAsync('loadCourseData', () =>
     courseDB.loadFullCourse(courseDir)
   );
-  // Write any errors and warnings to sync log
-  courseDB.writeErrorsAndWarningsForCourseData(courseId, courseData, (line) =>
-    logger.info(line || '')
-  );
   logger.info('Syncing info to database');
   await perf.timedAsync('syncCourseInfo', () => syncCourseInfo.sync(courseData, courseId));
   const courseInstanceIds = await perf.timedAsync('syncCourseInstances', () =>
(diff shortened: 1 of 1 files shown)


sync/syncFromDisk.js
```diff
@@ -93,6 +89,15 @@ async function syncDiskToSqlWithLock(courseDir, courseId, logger) {
...
     logger.info(chalk.green('✓ Course sync successful'));
   }
+
+  // Note that we deliberately log warnings/errors after syncing to the database
+  // since in some cases we actually discover new warnings/errors during the
+  // sync process. For instance, we don't actually validate exam UUIDs until
+  // the database sync step.
+  courseDB.writeErrorsAndWarningsForCourseData(courseId, courseData, (line) =>
+    logger.info(line || '')
+  );
```
mwest1066, line 99: Just to be extra pedantic, but did you test it to verify that `examUUID` warnings are in fact now shown correctly on sync? (The code all looks fine to me and I agree that it _should_ display `examUUID` warnings, so I'm happy to merge as is.)

mwest1066 (approved):

sync/syncFromDisk.js
nwalters512: 😂 thanks for asking, I hadn't specifically tested that. After setting `checkAccessRulesExamUuid` to `true` in my local config, I do indeed see UUID warnings being printed:

Screenshot 2022-11-03 at 12 25 09

mwest1066: Nice, thanks for humoring me! 😄

PR #6602, opened 2022-11-03
Replace showdown with remark in manual feedback rendering

### LINKED ISSUE #6600: Adding "bad html" in manual feedback in the box through manual grading can break rendering of the page
If you add in "bad" html in the manual feedback box while Manual Grading, it can cause the rendering of the page to break.

Here's the "result" of this bug:

I was able to fix it by just removing the comment that was added.

Screen Shot 2022-10-27 at 1 27 29 AM

jonatanschroeder:
Fixes #6600. Remark (via remark-raw) is better suited to handle badly formatted html.

### DIFF
--- a/lib/markdown.js
+++ b/lib/markdown.js
@@ -79,7 +79,7 @@ const makeHandler = (visitor) => {
 const handleCode = makeHandler(visitCodeBlock);
 const handleMath = makeHandler(visitMathBlock);
 
-const regularProcessor = unified()
+const defaultProcessor = unified()
   .use(markdown)
   .use(math)
   .use(handleMath)
@@ -110,6 +110,6 @@ module.exports.processQuestion = function (html) {
   });
 };
 
-module.exports.processContent = function (original) {
-  return regularProcessor.processSync(original).contents;
+module.exports.processContent = async function (original) {
+  return (await defaultProcessor.process(original)).contents;
 };
--- a/lib/markdown.js
+++ b/lib/markdown.js
@@ -79,7 +79,17 @@ const makeHandler = (visitor) => {
 const handleCode = makeHandler(visitCodeBlock);
 const handleMath = makeHandler(visitMathBlock);
 
-const processor = unified()
+const regularProcessor = unified()
+  .use(markdown)
+  .use(math)
+  .use(handleMath)
+  .use(gfm)
+  .use(remark2rehype, { allowDangerousHtml: true })
+  .use(raw)
+  .use(stringify);
+
+// The question processor also includes the use of pl-code instead of pre
+const questionProcessor = unified()
   .use(markdown)
   .use(math)
   .use(handleCode)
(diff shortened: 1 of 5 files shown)


lib/markdown.js
```diff
@@ -79,7 +79,17 @@ const makeHandler = (visitor) => {
 const handleCode = makeHandler(visitCodeBlock);
 const handleMath = makeHandler(visitMathBlock);
 
-const processor = unified()
+const regularProcessor = unified()
+  .use(markdown)
+  .use(math)
+  .use(handleMath)
+  .use(gfm)
+  .use(remark2rehype, { allowDangerousHtml: true })
```
nwalters512, line 87: Why is `allowDangerousHtml` set here?

lib/markdown.js
```diff
@@ -95,7 +105,11 @@ module.exports.processQuestion = function (html) {
...
       return `${prefix}${'#'.repeat(hashes.length - 1)}>`;
     });
-    const res = processor.processSync(decodedContents);
+    const res = questionProcessor.processSync(decodedContents);
     return res.contents;
   });
 };
+
+module.exports.processContent = function (original) {
+  return regularProcessor.processSync(original).contents;
```
nwalters512, line 114: Can we use an async processing function here instead? I'm unsure of exactly what might be async, but in general we try to avoid anything "sync" when an async version is available. (I acknowledge that I used `processSync` above - that was a mistake 😓)

lib/markdown.js
```diff
@@ -79,7 +79,17 @@ const makeHandler = (visitor) => {
 const handleCode = makeHandler(visitCodeBlock);
 const handleMath = makeHandler(visitMathBlock);
 
-const processor = unified()
+const regularProcessor = unified()
```
nwalters512, line 82: Nit: `defaultProcessor` instead?

lib/question.js
```diff
@@ -10,8 +10,9 @@ const moment = require('moment');
 const util = require('util');
 const fs = require('fs');
 const unzipper = require('unzipper');
-const showdown = require('showdown');
+const xss = require('xss');
```
nwalters512, line 13: Could we have remark/rehype filter out anything to do with scripting instead? IIRC you can configure it to do so, though I don't know exactly how off the top of my head.

lib/markdown.js
jonatanschroeder: remark-raw requires it: https://github.com/rehypejs/rehype-raw/blob/main/readme.md?plain=1#L44-L46

lib/question.js
jonatanschroeder: Yup: https://github.com/rehypejs/rehype-sanitize. I'll check it out.

lib/markdown.js
jonatanschroeder: Done.

nwalters512 (approved):
Code looks good to me! I'd greatly appreciate it if you could add some tests for this to https://github.com/PrairieLearn/PrairieLearn/blob/master/tests/markdown.test.js, specifically for things like ensuring that we're correctly sanitizing the resulting HTML.

nwalters512 (approved):

PR #6604, opened 2022-11-03
Finite State Machine Element

eliotwrobson:
A draft of the finite state machine element. Python code hasn't been reworked yet, but the main thing right now is getting comments on the existing javascript and quality of feedback. @SethPoulsen per discussion.

A running current version of the element can be seen [here](https://www.prairielearn.org/pl/course_instance/129595/instructor/question/9056546/preview).


elements/fsm-builder/fsm-builder.js
```diff
@@ -0,0 +1,646 @@
...
+//var caretTimer;
+//var caretVisible = true;
+
+//var nodes = [];
+//var links = [];
+
+//var stateNamesToHighlight = null
+//var transitionsToHighlight = null
+
+const snapToPadding = 6; // pixels
```
nwalters512, line 13: My main comment at this point: let's keep as much as possible out of the global scope. That goes for these constants, as well as all the classes in the JS files in `fsm-links.js`. You can do this by moving everything to one file, wrapping it in an IIFE, and then manually exporting only the bits that should be externally-accessible to the `window` object.

SethPoulsen: @eliotwrobson and I had a good chat the other day about different ways to address this. I should have commented here to note, sorry.

eliotwrobson: Just did this refactor (only took a few minutes). I _think_ I did it right 😅

elements/fsm-builder/fsm-builder.js
```diff
@@ -0,0 +1,1005 @@
...
+      this.transitionsToHighlight = null;
+
+      this.currentLink = null;
+      this.selectedObject = null;
+      this.movingObject = false;
+      this.shift = false;
+      this.caretVisible = true;
+      this.nodeRadius = largeStateSize;
+
+      this.canvas = document.getElementById(this.answersName + '-fsm-canvas');
```
eliotwrobson, line 29: The linter currently complains about this. Do I have to wait for the page to load (or something like that) before accessing the global document?

eliotwrobson:
Going to go ahead and close this for now, since we want to add some other features to the FSM builder as a push towards a publication. Once we get feedback from that (and integrate some more changes into the automata library to help support this element), will remake this PR using some initial changes from here.

PR #6608, opened 2022-11-03
Implement webhook for automated termination during deploys

nwalters512:
As discussed in #6381, this PR implements a webhook that can be used to gracefully terminate a server. This will be useful primarily for autoscaling chunk servers, as it will allow us to have a Lambda consume a lifecycle transition from EventBridge and notify the relevant instance that it should shut down and complete its lifecycle hook.

As part of this, I refactored how termination works. There's now a single codepath that starts with a `SIGTERM` handler. If we decide we want to shut down for whatever reason (namely this webhook), we can send `SIGTERM` to ourselves. This makes the code a bit easier to reason about.

I've left the polling code there for now, but I made it such that polling can be disabled entirely by setting `autoScalingLifecycleStatePollIntervalSec` to a falsy value.


webhooks/terminate.js
```diff
@@ -0,0 +1,65 @@
...
+    // If a client uses a keepalive connection, we'd be temporarily deadlocked
+    // if we waited for the client to close the connection, as `server.close()`
+    // in `server.js` won't finish until all connections have closed. To work
+    // around this, we'll explicitly close the underlying socket once we've sent
+    // the response.
+    //
+    // TODO: Once we have a minimum Node version of 18.2, we should use
+    // `server.closeIdleConnections` in `server.js` instead of this.
+    const socket = res.socket;
+    res.on('close', () => {
```
mwest1066, line 50: Tricky!

server.js
```diff
@@ -2087,21 +2093,30 @@ if (config.startServer) {
...
+          try {
+            await lifecycleHooks.completeInstanceTermination();
+          } catch (err) {
+            logger.error('Error completing instance termination', err);
             Sentry.captureException(err);
-          })
-          .finally(terminate);
+          }
+
+          await terminate();
```
mwest1066, line 2110: I guess we really don't want `terminate()` to throw an exception? It all looks safe with the possible exception of `Sentry.captureException()` in the catch -- maybe this should also be inside a `try` so that it can't possibly throw?

lib/lifecycle-hooks.js
```diff
@@ -66,50 +66,45 @@ module.exports.completeInstanceLaunch = async function () {
...
+      InstanceId: config.instanceId,
+    })
+  );
+  logger.info('Completed Auto Scaling lifecycle action for instance termination');
+};
 
-    // Return a promise that never settles.
+/**
+ * Resolves once the instance has entered the `Terminating:Wait` state. If
+ * not running in EC2 or not configured with an ASG or lifecycle hook name,
```
mwest1066, line 94:
```suggestion
 * not running in EC2, or not configured with an ASG or lifecycle hook name, or with a falsy poll interval,
```

mwest1066 (approved):

server.js
nwalters512: That's probably safest, yeah. Though something is probably extremely broken if that ever throws!

webhooks/terminate.js
```diff
@@ -0,0 +1,65 @@
...
+ * gracefully.
+ *
+ * This app requires a signed `PrairieLearn-Signature` JWT header to be present,
+ * so it's safe to publicly expose this endpoint if needed. However, it's
+ * recommended that this endpoint be blocked at your load balancer or firewall
+ * for extra security.
+ */
+router.post('/', async (req, res) => {
+  try {
+    const jwt = req.headers['prairielearn-signature'];
```
nwalters512, line 22: I know we usually use our CSRF token machinery when we want some kind of signed data, but I'm opting to use a standard JWT here. It's easier to generate/verify them from outside PL without needing to copy/paste around our custom code. Long term, I think we should broadly replace our custom CSRF tokens with JWTs.

mwest1066: Yep, I'm on board with this. I assume that the standard JWT libraries no longer have the [trivial vulnerabilities](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/) that were rife in the early versions.

lib/lifecycle-hooks.js
nwalters512: Done in 7a23bd2.

webhooks/terminate.js
```diff
@@ -0,0 +1,65 @@
...
+      return;
+    }
+    if (Array.isArray(jwt)) {
+      res.status(403).send('Multiple PrairieLearn-Signature headers');
+      return;
+    }
+
+    try {
+      const key = crypto.createSecretKey(config.secretKey, 'utf-8');
+      await jose.jwtVerify(jwt, key, { maxTokenAge: 60 });
```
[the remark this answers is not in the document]
nwalters512: We now check for a `PrairieLearn` issuer and a `terminate` subject. I don't think it's necessary to specify a specific signing algorithm, as we use symmetric cryptography.

webhooks/terminate.js
mwest1066: Yeah, I think the fact we are using symmetric key signing makes all these attacks irrelevant.

mwest1066 (approved):

PR #6610, opened 2022-11-04
Show later tests in python grader if quit early

### LINKED ISSUE #6385: Early stopping in Python Autograder grading scheme yielding point total issues.
Consider an auto-graded question with three tests:

- 1 point first test
- 2 points second test, and;
- 9 points for the third test

Total: 12 points

If we stop the grading suite early, say when only 2 of 3 tests are completed, the outcome reported to students is 1/3 points. However, the percentage is shown as 8.3% as we have 1/12 points earned from the total: 

sample-test-complete-with-early-stopping

Effectively the student sees 1/3. Thus, we would assume 33.3% of the grade would be given. However, the “full” portion seems to be what PrairieLearn is using to calculate their grade.

python-autograder-points-returned-early-stopping-test

nicknytko:
Should fix #6385 

image

Before, if we stopped grading early the later tests would not show up and would give weird point values in the grading results (though the score itself would be correct) Now, we show later tests but don't run them.


mwest1066:
This didn't seem to work for me in local testing. My test code was:
image
And the output was:
image
I assume that I'm supposed to be seeing the test for `z` as well?

For testing this locally I built a new image:
```
git checkout grader-python-fix-points-early-quit
cd graders/python
docker build . -t prairielearn/grader-python:latest
```

Can you spot something that I'm doing wrong here for testing?

nicknytko:
> This didn't seem to work for me in local testing. My test code was: image And the output was: image I assume that I'm supposed to be seeing the test for `z` as well?
> 
> For testing this locally I built a new image:
> 
> ```
> git checkout grader-python-fix-points-early-quit
> cd graders/python
> docker build . -t prairielearn/grader-python:latest
> ```
> 
> Can you spot something that I'm doing wrong here for testing?

Is PL pulling the latest version of the grader from dockerhub? I have
`"externalGradingPullImagesFromDockerHub": false` in my config.json file to turn that off

mwest1066:
> "externalGradingPullImagesFromDockerHub": false

🤦 Thank you.

mwest1066 (approved):

nicknytko:
@coatless would you mind trying this on one of your questions to see if it works?

coatless:
@nicknytko the fix works beautifully! Thanks again for patching the grader. 

Screenshot 2022-11-07 at 10 41 09 AM

PR #6612, opened 2022-11-04
Allow none-of-the-above to always be incorrect in multiple choice element

### LINKED ISSUE #3380: pl-multiple-choice does not support none-of-the-above option that is always incorrect.
The current implementation of `pl-multiple-choice` allows an option for `none-of-the-above`, which shows a final option "None of the above". This setting currently allows for this option to be always correct (when no answer is set to `correct=true`) or randomly set to true (by hiding the correct answer/s). However there is no way to set this option to never be correct. Please change the field to allow this to happen. In this case, the only difference between a "None of the above" answer listed in the answer list and set by `none-of-the-above` is that the latter is always the last option.

jonatanschroeder:
Resolves #3380. This is implemented by replacing how `none-of-the-above` (and `all-of-the-above`) are read as attributes, from boolean to string. Each can receive four values: `false` (the current default), `random` (the current behaviour for true), `correct` (always correct) and `incorrect` (always incorrect). True-like values still default to `random` for backwards-compatibility.

Note that the `none-of-the-above="correct"` behaviour could still be achieved before by having it set to `true` and no correct answers, which would cause the randomization of this option to always be true, however the option was added for completeness. The same is true for `all-of-the-above="correct"` and no incorrect answers.

### DIFF
--- a/elements/pl-multiple-choice/pl-multiple-choice.py
+++ b/elements/pl-multiple-choice/pl-multiple-choice.py
@@ -11,8 +11,8 @@ SCORE_CORRECT_DEFAULT = 1.0
 WEIGHT_DEFAULT = 1
 FIXED_ORDER_DEFAULT = False
 INLINE_DEFAULT = False
-NONE_OF_THE_ABOVE_DEFAULT = False
-ALL_OF_THE_ABOVE_DEFAULT = False
+NONE_OF_THE_ABOVE_DEFAULT = 'false'
+ALL_OF_THE_ABOVE_DEFAULT = 'false'
 EXTERNAL_JSON_DEFAULT = None
 HIDE_LETTER_KEYS_DEFAULT = False
 EXTERNAL_JSON_CORRECT_KEY_DEFAULT = 'correct'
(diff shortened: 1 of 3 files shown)


elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -85,52 +101,52 @@ def prepare(element_html, data):
...
-        # with probability 1/(number_correct + enable-nota + enable-aota).
-        # However, if len_correct is 0, nota_correct is guaranteed to be True.
-        # Thus, if no correct option is provided, 'None of the above' will always
+    if aota in ['correct', 'random'] and len_correct < 2:
+        # To prevent confusion on the client side
+        raise Exception('pl-multiple-choice element must have at least 2 correct answers when all-of-the-above is set to "correct" or "random"')
+
+    if len_correct == 0:
+        if nota == 'false':
+            # This means the code needs to handle the special case when len_correct == 0
```
nwalters512, line 113: I don't understand what this comment is saying - is there other code in this element that handles this special case?

elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -142,12 +158,22 @@ def prepare(element_html, data):
     # - nota and aota disabled
     number_answers = min(min(1, len_correct) + len_incorrect, number_answers)
 
-    if aota_correct:
+    if nota == 'random' or aota == 'random':
+        # Either 'None of the above' or 'All of the above' is correct
+        # with probability 1/(number_correct + nota + aota).
+        prob_space = len_correct + (1 if nota == 'random' else 0) + (1 if aota == 'random' else 0)
```
nwalters512, line 164: this `1 if ... else 0` is clearer to me than the implicit conversion from boolean to int up above; would you mind making the change there too?

nwalters512 (approved):

elements/pl-multiple-choice/pl-multiple-choice.py
jonatanschroeder: This was indeed confusing. I separated the logic to a separate check before this point.

elements/pl-multiple-choice/pl-multiple-choice.py
jonatanschroeder: Done in eb2d5b3. I did some rearranging as well to make it easier to follow.

PR #6613, opened 2022-11-05
fix: Initialization of socket server in workspace host

jonatanschroeder:
#6381 changed the interface to initialize a socket server, but a reference to this interface was not updated in the workspace host call. This caused the workspace host not to be initialized properly, as a callback function that was expected to run was indeed not running anymore.

### DIFF
--- a/workspace_host/interface.js
+++ b/workspace_host/interface.js
@@ -217,6 +217,12 @@ async.series(
         callback(null);
       });
     },
+    (callback) => {
+      server = http.createServer(app);
+      server.listen(workspace_server_settings.port);
+      logger.info(`Workspace server listening on port ${workspace_server_settings.port}`);
+      callback(null);
+    },
     async () => {
       socketServer.init(server);
     },
@@ -225,12 +231,6 @@ async.series(
         if (ERR(err, callback)) return;
         callback(null);
       });
-    },
-    (callback) => {
-      server = http.createServer(app);
-      server.listen(workspace_server_settings.port);
-      logger.verbose(`Workspace server listening on port ${workspace_server_settings.port}`);
-      callback(null);
     },
     async () => {
       // Set up file watching with chokidar
--- a/workspace_host/interface.js
+++ b/workspace_host/interface.js
@@ -218,10 +218,8 @@ async.series(
       });
     },
     (callback) => {
-      socketServer.init(server, function (err) {
-        if (ERR(err, callback)) return;
-        callback(null);
-      });
+      socketServer.init(server);
+      callback(null);
     },
     (callback) => {
       util.callbackify(workspaceHelper.init)((err) => {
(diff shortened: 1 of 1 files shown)


nwalters512 (approved):
Good catch! Approving because this is obviously broken, but I'm also curious if you know how the workspace host actually uses websockets. As far as I can tell, it's still not being initialized correctly, as the `server` that's being passed as an argument here isn't initialized until a few lines later. I'm wondering if we can just remove this code entirely?

jonatanschroeder:
> Good catch! Approving because this is obviously broken, but I'm also curious if you know how the workspace host actually uses websockets. As far as I can tell, it's still not being initialized correctly, as the `server` that's being passed as an argument here isn't initialized until a few lines later. I'm wondering if we can just remove this code entirely?

That's actually a pretty good question, but I have no answer for it. I know very little of the functionality of the workspace code. Maybe @mwest1066 or @nicknytko can confirm? A scan of the code seems to indicate it is used in lib/workspace.js, but from a quick read of the code it's not clear to me what it is used for.

jonatanschroeder:
Quick test: removing this code causes errors. Changing it to use `null` as the argument works (which makes sense, given your observation that `server` had not been initialized at this point). Moving the creation of the server to before this point also works, and may have been the intended result. My understanding of the difference here is that with a server null, the socket server would create its own server in a random port, instead of using the port from the workspace settings, but I may be wrong.

mwest1066:
Workspace hosts use websockets to communicate workspace state to the client. In particular, messages like "Pulling image: 56%" are sent to the client via websockets. For this reason the workspace hosts connect to redis to participate in the global websocket routing pool.

As an example, the message generated by the workspace host here: https://github.com/PrairieLearn/PrairieLearn/blob/3720df73a1b3c7a1bd5108f0cfb37dd6f9f4ab30/workspace_host/interface.js#L1045 results in a call to: https://github.com/PrairieLearn/PrairieLearn/blob/3720df73a1b3c7a1bd5108f0cfb37dd6f9f4ab30/lib/workspace.js#L127 which finally uses websockets: https://github.com/PrairieLearn/PrairieLearn/blob/3720df73a1b3c7a1bd5108f0cfb37dd6f9f4ab30/lib/workspace.js#L114

nwalters512:
Ah! That makes sense. So workspace hosts don't actually talk directly to clients, they just emit things to Redis and let the internet-facing servers handle them. That explains why having a null server isn't a problem.

mwest1066 (approved):

mwest1066:
> Ah! That makes sense. So workspace hosts don't actually talk directly to clients, they just emit things to Redis and let the internet-facing servers handle them. That explains why having a null server isn't a problem.

Yep, exactly!

At some point this global websocket communication layer will presumably become a problem, but I don't have an immediately good idea of how to do away with it.

PR #6617, opened 2022-11-05
Fix null-discrimination in question stats

mwest1066:
If all students receive the same score for a question then the correlation between question scores and assessments scores is undefined and the posgresl `corr()` function returns NULL. This breaks the plotting of our question statistics. This PR fixes this by setting the discrimination to zero in this case. I think that a discrimination of zero makes sense here because we are learning nothing about a student's ability by their score on such a question, so it's discrimination should be zero.

Bug reported by @BenjaminCosman


sprocs/assessment_questions_calculate_stats.sql
```diff
@@ -75,7 +75,7 @@ aq_stats AS (
     SELECT
         least(100, greatest(0, avg(question_stats_by_user_or_group.score_perc)))                     AS mean_question_score,
         sqrt(var_pop(question_stats_by_user_or_group.score_perc))                                    AS question_score_variance,
-        corr(question_stats_by_user_or_group.score_perc, assessment_scores_by_user_or_group.score_perc) * 100 AS discrimination,
+        coalesce(corr(question_stats_by_user_or_group.score_perc, assessment_scores_by_user_or_group.score_perc) * 100, 0) AS discrimination,
```
jonatanschroeder, line 78: There is a minor issue with this approach, which is the fact that, if there are no submissions, all other stats will be null, but the discrimination will be 0. This is by no means major, and I'd be fine if you were to ignore this case, but if consistency is to be considered it may be a good idea to have some way to address that, maybe like this?
```suggestion
        coalesce(corr(question_stats_by_user_or_group.score_perc, assessment_scores_by_user_or_group.score_perc) * 100, CASE WHEN count(question_stats_by_user_or_group.score_perc) > 0 THEN 0 ELSE NULL END) AS discrimination,
```

jonatanschroeder:
Another issue I have with this: if the problem is in the plotting, wouldn't it make more sense to move the handling of null values to the plot itself? In other words, change 
https://github.com/PrairieLearn/PrairieLearn/blob/20fea5e6e49631b53a21db4b00d936440b67f553/pages/instructorAssessmentQuestionStatistics/instructorAssessmentQuestionStatistics.ejs#L69

to use 0 if the value is null?

mwest1066:
> Another issue I have with this: if the problem is in the plotting, wouldn't it make more sense to move the handling of null values to the plot itself?

I want to have some useful value stored for the discrimination so that when we do things like compute averages across assessments or download CSVs we also get a useful value. In this case a value of "zero" makes sense (the question is not discriminating).

sprocs/assessment_questions_calculate_stats.sql
mwest1066: I'm on board with this.

jonatanschroeder (approved):

PR #6620, opened 2022-11-06
Maintain HTML ordering of pl-hidden-hint hints

eliotwrobson:
Fixes issue where hints with the same `show-after-submission` can appear out of the order they appear in the original HTML.


exampleCourse/questions/element/hiddenHints/question.html
```diff
@@ -13,8 +13,13 @@
 
 <pl-question-panel>
   <pl-hidden-hints>
+    <!-- Hints with the same show-after-submission appear in the order they're written -->
     <pl-hint>
-      This is a hint that will be shown immediately.
+      Another hint that will be shown immediately.
```
nwalters512, line 18: Nit: the "another" one should probably come second?

elements/pl-hidden-hints/pl-hidden-hints.py
```diff
@@ -29,14 +34,14 @@ def render(element_html: str, data: pl.QuestionData) -> str:
 
     hints_to_display = []
 
-    for idx, (priority, hint) in enumerate(sorted(hints)):
+    for idx, (priority, _, hint) in enumerate(sorted(hints), 1):
```
nwalters512, line 37: For my own understanding: when `sorted` is applied to a list of tuples, it sorts by the first element of each tuple, and considers the second, third, etc. elements for tie-breaking?

nwalters512 (approved):

elements/pl-hidden-hints/pl-hidden-hints.py
eliotwrobson: Yes, that's correct. That was the issue before, it would break ties by alphabetically sorting the contents of the hints. That won't be an issue anymore because the second elements are all distinct (no ties possible).

exampleCourse/questions/element/hiddenHints/question.html
eliotwrobson: This is actually on purpose to test the behavior is as-expected. "Another" is before "A " lexicographically (and I couldn't think of a generic hint message that started with the letter "B" 😅 )

nwalters512: Hah, fair! If you'd like to test behavior, consider writing a unit test instead.

eliotwrobson: I think I probably will if I need to make any more complex changes to this element. Unfortunately hard to strip out the ordering logic from parsing the element itself (and I don't want to move in all of our test fixtures just for testing something this simple, they're pretty huge).

nwalters512: You can and should write Python tests for elements (at least for the `server.py` code). And while we don't have any tests that test a rull render yet, it shouldn't be hard to add: you can just pass in a blob of HTML to the `render` function and make assertions about the returned HTML.

Feel free to save that for another PR thought - I'll merge this as-is!

PR #6621, opened 2022-11-07
Manual grading: hide student information by default (blind grading)

### LINKED ISSUE #6528: Feature request: double-blind grading
Currently when I'm manually grading on PL, there is a “Staff information” box which names the student I’m grading. I’d like the **option** to see this info, but most of the time I’d rather it be hidden. (For example, hidden by default + an 'expand' button.)

jonatanschroeder:
Resolves #6528. The manual grading assessment question page hides the student information columns by default, with a button that toggles its visibility.

Also, all instance question and assessment instance pages will hide this information by default as well. Showing the information involves expanding a details tag.

### DIFF
--- a/docs/elements.md
+++ b/docs/elements.md
@@ -1930,7 +1930,8 @@ For the inner `pl-hint` element:
 #### Details
 
 Add hints to a variant that are revealed with more submissions using the `show-after-submission` attribute. By default, hints without
-`show-after-submission` set are always shown.
+`show-after-submission` set are always shown. Hints with the same `show-after-submission` appear in the order they're written in the
+question HTML.
 
 #### Example implementations
 
--- a/elements/pl-hidden-hints/pl-hidden-hints.py
+++ b/elements/pl-hidden-hints/pl-hidden-hints.py
@@ -3,6 +3,9 @@ import lxml
 import prairielearn as pl
 
 
+# Based on the original hidden-hint element by Jason Xia
+
+
 def render(element_html: str, data: pl.QuestionData) -> str:
     element = lxml.html.fragment_fromstring(element_html)
     pl.check_attribs(element, [], [])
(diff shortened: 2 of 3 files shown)


pages/partials/instructorInfoPanel.ejs
```diff
@@ -16,20 +16,24 @@
...
-    <div class="d-flex flex-wrap pb-2">
-      <div class="pr-1"><%= instance_user.name %></div>
-      <div class="pr-1"><%= instance_user.uid %></div>
-    </div>
-    <% } %>
+  <hr>
+  <div>
+    <details>
+      <% if (instance_group) { %>
+      <summary><h5 class="card-title">Group:</h5></summary>
```
nwalters512, line 23:
```suggestion
      <summary><h5 class="card-title">Group details</h5></summary>
```

pages/partials/instructorInfoPanel.ejs
```diff
@@ -16,20 +16,24 @@
...
+      <div class="d-flex flex-wrap pb-2">
+        <div class="pr-1"><%= instance_group.name %></div>
+        <div class="pr-1">(<%= instance_group_uid_list.join(', ') %>)</div>
+      </div>
+      <% } else if (instance_user) { %>
+      <summary><h5 class="card-title d-inline-block">Student:</h5></summary>
```
nwalters512, line 29:
```suggestion
      <summary><h5 class="card-title d-inline-block">Student details</h5></summary>
```

pages/partials/instructorInfoPanel.ejs
```diff
@@ -16,20 +16,24 @@
   </div>
 
   <% if (locals.authz_result && (!authz_result.authorized_edit)) { %>
```
nwalters512, line 18: While trying to test this locally, `authz_result.authorized_edit` was true. It's not clear to me why that should prevent this information from showing? I acknowledge that this is unlikely in practice, I think this is only true if I (the person viewing the question) was also the one that created the instance.

We should consider changing this conditional, but if we don't, what's your recommendation for testing this locally?

jonatanschroeder: > While trying to test this locally, authz_result.authorized_edit was true. It's not clear to me why that should prevent this information from showing? I acknowledge that this is unlikely in practice, I think this is only true if I (the person viewing the question) was also the one that created the instance.

That's an interesting point. I didn't play with this conditional, but I believe this will also be the case in the question preview, as well as for all students in a group.

> We should consider changing this conditional, but if we don't, what's your recommendation for testing this locally?

You can use the admin query that generates students for a course, then change your effective user to one of the generated students and create a submission. Alternatively, you can use the tools/generate_submissions.py script instead of changing the effective user, that has the advantage of creating submissions for all questions, but is not really necessary in this context.

nwalters512 (approved):

PR #6622, opened 2022-11-07
Fix termination sequence

### LINKED ISSUE #6619: PUBLISH can't be processed
When rolling back an installation of 20fea5e6e49631b53a21db4b00d936440b67f553 on the canary main servers we got this error: https://sentry.io/organizations/prairielearn-inc/issues/3721855226/?project=6704263

This came from a canary main server and I'm guessing it was in the process of being shut down. I think this is a redis error, so we might be shutting down the redis connection while we are still processing external grading jobs, or something like that?

nwalters512:
Closes #6619. There should never be a need for us to "gracefully" shut down the HTTP or socket.io servers. If they still have connections when we start terminating, something else has gone wrong and we can allow those connections to die unceremoniously.

### DIFF
--- a/lib/socket-server.js
+++ b/lib/socket-server.js
@@ -33,7 +33,9 @@ module.exports.close = async function () {
 
   // Note that we don't use `io.close()` here, as that actually tries to close
   // the underlying HTTP server. In our desired shutdown sequence, we first
-  // close the HTTP server and then later disconnect all sockets.
+  // close the HTTP server and then later disconnect all sockets. There's some
+  // discussion about this behavior here:
+  // https://github.com/socketio/socket.io/discussions/4002#discussioncomment-4080748
   //
   // Note the use of `io.local`, which prevents the server from attempting to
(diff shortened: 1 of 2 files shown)


mwest1066 (approved):

lib/socket-server.js
```diff
@@ -33,7 +33,9 @@ module.exports.close = async function () {
 
   // Note that we don't use `io.close()` here, as that actually tries to close
   // the underlying HTTP server. In our desired shutdown sequence, we first
-  // close the HTTP server and then later disconnect all sockets.
+  // close the HTTP server and then later disconnect all sockets. There's some
+  // discussion about this behavior here:
+  // https://github.com/socketio/socket.io/discussions/4002#discussioncomment-4080748
```
mwest1066, line 38: Huh, that's pretty evil behavior on the part of socket.io.

mwest1066 (approved):

PR #6624, opened 2022-11-08
Don't treat external grader results JSON as binary-encoded

### LINKED ISSUE #6605: Ambiguous error message to students: Could not parse the grading results.
It turns out students were getting this message (Could not parse the grading results) when they submitted code where a test failed.  This did not appear when all of the tests passed.

This question was related to File I/O and I had 2 tests that used a fairly large file (thousands of lines).  I reduced the size of this input file and the error message went away.  I could not replicate this in my dev environment. 
This is the question: https://ca.prairielearn.com/pl/course_instance/2396/instructor/question/245234/preview

Funny thing is, I am using a much bigger file in this question and I not encountering the same problem.
https://ca.prairielearn.com/pl/course_instance/2396/instructor/question/243128/preview

This question was generating too much output and got the message 'result is too large' and so I fixed this by reducing the output but did not reduce the size of the input file.  

Looking more closely at the difference between these questions, perhaps it is call to copy of a REALLY large list that is causing the problem.

nwalters512:
Closes #6605.

This explains why I'm unable to reproduce the error in #6605, and why the issue was fixed for the author when they reduced the amount of output: the data is now small enough to fix in the SQS results message, so we now never try to read this file in PL. When the results file was originally larger, it *didn't* fit in the SQS results, so we had to read it from S3. We don't use this code locally, which explains why it can't be reproduced locally either. This *also* explains why this particular error went away

The `'binary'` has been present for as long as the external grader host code has existed. I don't know why I used that in the first place, but I can't see any reason why we should use it - everything should be UTF8.

### DIFF
--- a/grader_host/index.js
+++ b/grader_host/index.js
@@ -622,7 +622,7 @@ function uploadResults(info, callback) {
         const params = {
           Bucket: s3Bucket,
           Key: `${s3RootKey}/results.json`,
-          Body: Buffer.from(JSON.stringify(results, null, '  '), 'binary'),
+          Body: Buffer.from(JSON.stringify(results, null, 2)),
         };
         s3.putObject(params, (err) => {
           if (ERR(err, callback)) return;


grader_host/index.js
```diff
@@ -622,7 +622,7 @@ function uploadResults(info, callback) {
         const params = {
           Bucket: s3Bucket,
           Key: `${s3RootKey}/results.json`,
-          Body: Buffer.from(JSON.stringify(results, null, '  '), 'binary'),
+          Body: Buffer.from(JSON.stringify(results, null, 2)),
```
mwest1066, line 625: I'm impressed that you tracked down the issue to this line (which is clearly wrong, to be fair).

nwalters512: I ultimately compared `results.json` from the job tarball to the raw `results.json` that we upload to S3 and noticed that they had different encodings 🙃

PR #6625, opened 2022-11-08
Allow pl-file-editor font size to be customized

TwinIsland:
Hi, I add a font-adjust function in pl-file-editor. Similar to the theme changing function, the font-size adjust also support sync and live view

image


TwinIsland:

nwalters512:
Thanks for the PR! I'll take a look at it. In the meantime, I'm curious what the driver behind this feature is. My expectation is that folks looking for a larger font size would change their browser's zoom level or default font size.

elements/pl-file-editor/pl-file-editor.js
```diff
@@ -95,10 +101,14 @@ window.PLFileEditor.prototype.syncSettings = function () {
...
       this.editor.setTheme(event.newValue);
     }
+    if (event.key === 'pl-file-editor-textsize') {
+      this.editor.setFontSize(event.newValue);
+    }
   });
 
   window.addEventListener('pl-file-editor-settings-changed', () => {
     this.editor.setTheme(localStorage.getItem('pl-file-editor-theme'));
+    this.editor.setFontSize(localStorage.getItem('pl-file-editor-textsize'));
```
nwalters512, line 111: Whenever retrieving values from local storage, you'll need to take care to convert them back into an integer, as you do in the constructor. The fact that that's not currently being done is causing sync across multiple editors to fail. You'll need to make the same change in the `storage` event listener above.

Alternatively, you can store the value as a pixel string (e.g. `16px`). That might be a cleaner option so we're not constantly converting back and forwards between options.

elements/pl-file-editor/pl-file-editor.js
```diff
@@ -11,6 +11,8 @@ window.PLFileEditor = function (uuid, options) {
 
   this.inputElement = this.element.find('input');
   this.editorElement = this.element.find('.editor');
+  this.textsizeIncButton = this.element.find(".textsize-inc");
```
nwalters512, line 14: Since Ace uses "font size", let's use that same terminology instead of "text size".

elements/pl-file-editor/pl-file-editor.mustache
```diff
@@ -42,6 +42,19 @@ $(function() {
...
+                            <br>
+                            <button type="button" class="btn btn-sm btn-secondary textsize-dec" aria-label="fontsize decrease">
+                                <i class="fa fa-font"></i>
+                                -
+                            </button>
+                            <span class="textsize-current" id="modal-{{uuid}}-textsize" style="padding: 3px;"></span>
+                            <button type="button" class="btn btn-sm btn-secondary textsize-inc" aria-label="fontsize increase">
+                                <i class="fa fa-font"></i>
+                                +
+                            </button>
```
nwalters512, line 57: The increase/decrease buttons are nice, but I think it may be easier to make this accessible if we use a `<select>` element with predefined text sizes. I'm thinking maybe the following (based on https://github.com/ajaxorg/ace/blob/23208f2f19020d1f69b90bc3b02460bda8422072/src/ext/textarea.js#L390, with a larger max value):

- 10px
- 11px
- 12px
- 14px
- 16px
- 18px
- 20px
- 22px
- 24px

elements/pl-file-editor/pl-file-editor.mustache
```diff
@@ -42,6 +42,19 @@ $(function() {
                             <label for="modal-{{uuid}}-themes">Editor Theme</label>
                             <select class="custom-select" id="modal-{{uuid}}-themes">
                             </select>
+
+                            <div id="modal-split" style="padding-top: 1rem"></div>
```
nwalters512, line 46: Instead of a spacer div, you should wrap your new controls in a div with the `form-group` class. See https://getbootstrap.com/docs/4.6/components/forms/ for an example.

elements/pl-file-editor/pl-file-editor.mustache
```diff
@@ -42,6 +42,19 @@ $(function() {
...
+                            <label for="modal-{{uuid}}-font-adjust">Font Size</label>
```
nwalters512, line 47: There's no element in this modal with this ID, so this isn't a useful label. (This will be solved if you use `<select>` instead)

elements/pl-file-editor/pl-file-editor.js
```diff
@@ -218,4 +268,4 @@ window.PLFileEditor.prototype.b64EncodeUnicode = function (str) {
       return String.fromCharCode('0x' + p1);
     })
   );
-};
+};
```
nwalters512, line 271: Please retain the trailing newline.

TwinIsland:
> Thanks for the PR! I'll take a look at it. In the meantime, I'm curious what the driver behind this feature is. My expectation is that folks looking for a larger font size would change their browser's zoom level or default font size.

thanks for your reply! The reason drives me to write this function is that I find the default font size of the editor is small compared with the description text, so evereytime I will spend some time to adjust the font size before writting by zomming in the screen (the cost will be loss some readable area because it makes everything LARGER). I also find many of my friends have a similar experience as me. Hence, I think it might be a good idea to add a font adjust function in the seting modal bracket not only for conveniencing (set it once and enjoy the effect later), but also makes the setting icon more functional instead of only play a role in setting the theme.

TwinIsland:
@nwalters512 btw, Really thanks for your suggestions! It's actually the first time I try to contribute to an open source project, and before that, I only tired to implement my web knowledges into building my own blog. This experience gives me a hint on how an general open-source project works, and what should consider when coding for a huge project. I'll try to take some formal front-end programming course later, and hope I can contribute more to this project in the future :)

nwalters512:
@TwinIsland fantastic, we're honored that your first open-source contribution was to PL!

elements/pl-file-editor/pl-file-editor.js
```diff
@@ -141,27 +149,51 @@ window.PLFileEditor.prototype.initSettingsButton = function (uuid) {
...
     });
+    that.modal.find('#modal-' + uuid + '-fontsize').change(function () {
+      var fontSize = $(this).val();
+      that.editor.setFontSize(fontSize);
+    });
   });
 
   this.saveSettingsButton.click(function () {
     var theme = that.modal.find('#modal-' + uuid + '-themes').val();
+    var fontsize = that.editor.getFontSize();
```
nwalters512, line 181: I believe should use the value from the `<select>` element, just like `theme` does on the previous line.

nwalters512 (commented):
You'll need to auto-format the files to pass CI; run `make format` locally. After that, this PR should be good to go!

elements/pl-file-editor/pl-file-editor.js
nwalters512: This was resolved in db966ae.

nwalters512 (approved):

nwalters512:
Thanks for the contribution! We'll take this out to prod with the next deploy 🚀

TwinIsland:
> Thanks for the contribution! We'll take this out to prod with the next deploy 🚀

good to hear, thanks!!!

PR #6631, opened 2022-11-12
Add manual grader columns for assessment downloads

### LINKED ISSUE #6623: Feature request: "grader" column in some csv download(s)
I'd ultimately like to be able to e.g.

- see which TAs have graded how many problems total (for workload rebalancing)
- see if different TAs are assigning similar average grades (to maintain consistency)

So it would be nice if one or more of the standard csv downloads (possibly *_instance_questions.csv?) included a column for who graded each question instance.

(Currently the web interface can show you who the grader is for one student at a time, but this method is not good for producing aggregate statistics.)

jonatanschroeder:
Resolves #6623. Adds columns for assigned grader and last grader to the CSV downloads for instance questions and submissions (all/best/last).

#6623 refers to a column for when the assessment was last graded. This column already exists as "Grading date" in the submissions files.

### DIFF
--- a/pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql
+++ b/pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql
@@ -66,8 +66,8 @@ SELECT
     DATE_PART('epoch', iq.duration) AS duration_seconds,
     g.name AS group_name,
     groups_uid_list(g.id) AS uid_list,
-    COALESCE(agu.name, agu.uid) AS assigned_grader_name,
-    COALESCE(lgu.name, lgu.uid) AS last_grader_name
+    agu.uid AS assigned_grader,
+    lgu.uid AS last_grader
 FROM
     instance_questions AS iq
     JOIN assessment_questions AS aq ON (aq.id = iq.assessment_question_id)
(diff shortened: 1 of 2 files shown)


pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql
```diff
@@ -65,7 +65,9 @@ SELECT
     iq.number_attempts,
     DATE_PART('epoch', iq.duration) AS duration_seconds,
     g.name AS group_name,
-    groups_uid_list(g.id) AS uid_list
+    groups_uid_list(g.id) AS uid_list,
+    COALESCE(agu.name, agu.uid) AS assigned_grader_name,
+    COALESCE(lgu.name, lgu.uid) AS last_grader_name
```
nwalters512, line 70: It may be useful to return both the name and UID as separate columns? UIDs will be unique, whereas names may not be.

nwalters512 (approved):

pages/instructorAssessmentDownloads/instructorAssessmentDownloads.sql
jonatanschroeder: I'm trying to avoid crowding the files with too many columns. I can replace them with UIDs only, if that is a concern...

nwalters512: In that case, it may be better to just use UIDs - I believe that's what we do for group members (UIDs only, no names).

jonatanschroeder: Done in 0204092.

PR #6632, opened 2022-11-14
pl-order-blocks: raise exception if question has non-unique tags

SethPoulsen:
We've had a few people confused why problems aren't being graded properly, when they shouldn't be using duplicate tags.

Throw an error so the question won't even render.

This could potentially break some questions in the wild. Though arguably they are already broken if they are using duplicate tags. Making them break loudly might be better than having them be broken silently, which is what currently happens.

### DIFF
--- a/elements/pl-order-blocks/pl-order-blocks.py
+++ b/elements/pl-order-blocks/pl-order-blocks.py
@@ -134,7 +134,7 @@ def prepare(element_html, data):
             tag = str(index)
 
         if tag in used_tags:
-            raise Exception('tag "' + tag + '" used on multiple <pl-answer>. The tag attribute for each <pl-answer> must be unique.')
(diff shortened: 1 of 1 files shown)


mwest1066 (approved):

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -132,6 +133,11 @@ def prepare_tag(html_tags, index, group_info={'tag': None, 'depends': None}):
         if grading_method == 'ranking':
             tag = str(index)
 
+        if tag in used_tags:
+            raise Exception('tag "' + tag + '" used on multiple <pl-answer>. The tag attribute for each <pl-answer> must be unique.')
```
eliotwrobson, line 137:
```suggestion
            raise Exception(f'tag "{tag}" used on multiple <pl-answer>. The tag attribute for each <pl-answer> must be unique.')
```

nwalters512 (approved):

PR #6633, opened 2022-11-14
Soft-delete assessment instances

### LINKED ISSUE #3516: Soft-delete assessment instances.
Based on Slack discussion, we should evaluate the use of soft-delete for assessment instances. Potentially there should also be a way to see deleted instances, and to undelete an instance deleted by mistake. All associated elements should probably be kept as well (e.g., submissions, logs).

Also, it may be good to soft-delete an instance when a group associated to it is deleted.

### LINKED ISSUE #6541: Better listing of assessment instances with deleted groups
The assessment instances page (Students tab) for group assessments lists all instances, including those for (soft-deleted) groups. However, since the group doesn't "exist", the name of the group is "null" which is confusing, and it's hard to distinguish one list from another:

image

This list should either not list these instances, or list them with a clearer indication that they are related to a deleted group.

Related: #3516

jonatanschroeder:
Resolves #3516. Indirectly also fixes #6541.

* [x] Database migration/description;
* [x] Change sproc to update instead of delete assessment instance;
* [x] Update handling of assessment number (one possible implementation done, review if there is a better approach);
* [x] Update relevant queries to filter out deleted assessment instances;
* [ ] Update relevant queries to filter out instance questions, variants and submissions associated to deleted assessment instances;
* [ ] Add tests for handling of deleted assessment instances;
* [x] Group delete causes the corresponding assessment instance to be deleted;

For future PR (issue to be created later): option to list deleted assessment instances. In that case, there should also be a review of the associated pages (assessment log, instance question, etc.) to ensure they properly convey relevant information, block further actions and clearly indicate the assessment is deleted. Undelete may be possible, but should be done with care, to ensure another instance was not created for the same student.


jonatanschroeder:
This PR is really outdated and would require basically starting from scratch to update. I'll close it.

PR #6635, opened 2022-11-15
Use correct protocol for student links

### LINKED ISSUE #6628: Update Student Link generation in dev container
When working in a development container, viewing the Student View link for an assessment uses https even in the case where the dev container serves the instructor view with http. 

Suggestion by Nathan Walters on [linked slack thread](https://prairielearn.slack.com/archives/C266KEH9A/p1667968115903609?thread_ts=1667888643.153889&cid=C266KEH9A): "We should probably use the protocol of the request when constructing the link."

nwalters512:
Fixes #6628. Because we set [`trust proxy`](https://expressjs.com/en/guide/behind-proxies.html) to 1 when running behind a load balancer, it will respect the value of `X-Forwarded-Proto` and use it for `req.protocol`.

### DIFF
--- a/pages/instructorAssessmentSettings/instructorAssessmentSettings.js
+++ b/pages/instructorAssessmentSettings/instructorAssessmentSettings.js
@@ -40,15 +40,15 @@ router.get('/', function (req, res, next) {
     ],
     function (err) {
       if (ERR(err, next)) return;
-      debug('render page');
-      let host = config.serverCanonicalHost || 'https://' + req.headers.host;
-      res.locals.studentLink =
-        host +
+      const host = config.serverCanonicalHost || `${req.protocol}://${req.headers.host}`;
+      res.locals.studentLink = new URL(
         res.locals.plainUrlPrefix +
-        '/course_instance/' +
-        res.locals.course_instance.id +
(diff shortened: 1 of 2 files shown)


pages/instructorAssessmentSettings/instructorAssessmentSettings.js
```diff
@@ -40,15 +40,15 @@ router.get('/', function (req, res, next) {
     ],
     function (err) {
       if (ERR(err, next)) return;
-      debug('render page');
-      let host = config.serverCanonicalHost || 'https://' + req.headers.host;
-      res.locals.studentLink =
-        host +
+      const host = config.serverCanonicalHost || `${req.protocol}://${req.headers.host}`;
```
mwest1066, line 43: Just to make sure I'm understanding, this use of `req.protocol` won't affect us in prod because we always set `serverCanonicalHost`, right?

mwest1066 (approved):

pages/instructorAssessmentSettings/instructorAssessmentSettings.js
nwalters512: That's true, yes!

PR #6637, opened 2022-11-15
Increase default font size of pl-python-variable dataframes

ffund:
Previously, the font size of a `pandas` dataframe displayed using a `pl-python-variable` was smaller than the surrounding text.

### DIFF
--- a/elements/pl-python-variable/pl-python-variable.css
+++ b/elements/pl-python-variable/pl-python-variable.css
@@ -2,7 +2,7 @@
     display: block;
     overflow-x: scroll;
     font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
-    font-size: 12px;
+    font-size: 1rem;
     line-height: 20px;
 
     border-collapse: collapse;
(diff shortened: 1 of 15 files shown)


ffund:
Before:

image

After:

image

elements/pl-python-variable/pl-python-variable.css
```diff
@@ -2,7 +2,7 @@
     display: block;
     overflow-x: scroll;
     font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
-    font-size: 12px;
+    font-size: 1rem;
```
nwalters512, line 5: Could we just delete the `font-size` property entirely and use the Bootstrap defaults?

mwest1066: Good thinking! @ffund does it look sensible if we just don't set a `font-size` at all? I'm also tempted to remove `font-family` unless there's a good reason we are specifying it?

ffund: Sounds fine. if we do that, we should also remove `font-family` for the 'dimensions' line at the bottom.

PR #6638, opened 2022-11-15
Expose workspace container logs and history to instructors

### LINKED ISSUE #6530: Display workspace logs and state transitions to instructors
Once https://github.com/PrairieLearn/PrairieLearn/pull/6527 lands, we should build an interface to allow instructors to view logs and state transitions for a workspace. 

Considerations:

- We'll need to separately expose logs for different "versions" of workspaces.
- We should persist logs for only 90 days; after that, we should consider them inaccessible. The underlying S3 bucket can then be configured to e.g. delete logs after that or shift them to Glacier-tier storage.

nwalters512:
This PR exposes workspace logs (specifically, container logs and "history", the latter of which we confusingly refer to in the database as `workspace_logs`). This should make it easier for instructors to debug issues with their workspaces.

Logs are accessible from a new "Logs" button on the workspace itself:

Screenshot 2022-11-15 at 15 40 37

This feature is split into two pages to account for workspace versions. The first page shows the list of versions and all the "history":

Screenshot 2022-11-15 at 15 37 43

The second page shows the container logs and history for a specific version:
Screenshot 2022-11-15 at 15 37 54

I implemented container log expiration, which is configurable. Here's what the version page looks like after the container logs have expired (thanks to GitHub for the design inspiration!):

Screenshot 2022-11-15 at 15 38 17

Closes #6530.


pages/workspace/workspace.ejs
```diff
@@ -84,17 +90,25 @@
           <span class="nav-item badge badge-light"><%= navTitle %></span>
         </li>
         <li class="nav-item ml-2 my-1">
-          <a id="reboot" class="nav-item btn btn-light" data-toggle="modal" data-target="#rebootModal">
+          <button id="reboot" class="nav-item btn btn-light" data-toggle="modal" data-target="#rebootModal">
```
nwalters512, line 93: Minor accessibility improvement (these are semantically buttons, not anchors).

pages/workspace/workspace.ejs
```diff
@@ -84,17 +90,25 @@
...
+          <button id="reset" class="nav-item btn btn-light" data-toggle="modal" data-target="#resetModal">
             <i class="fas fa-trash text-danger" aria-hidden="true"></i>
             Reset
-          </a>
+          </button>
         </li>
+        <% if (showLogs) { %>
+          <li class="nav-item ml-2 my-1">
+            <a class="nav-item btn btn-light" href="<%= urlPrefix %>/workspace/<%= workspace_id %>/logs" target="_blank">
+              <i class="fas fa-arrow-up-right-from-square" aria-hidden="true"></i>
```
nwalters512, line 107: Any idea for a better icon? I couldn't find anything that screamed "logs" to me 🤔

pages/workspaceLogs/workspaceLogs.html.js
```diff
@@ -0,0 +1,239 @@
...
+  containerLogsExpired,
+  resLocals,
+}) => {
+  return html`
+    <!DOCTYPE html>
+    <html lang="en">
+      <head>
+        ${renderEjs(__filename, "<%- include('../partials/head'); %>", {
+          ...resLocals,
+          pageTitle: 'Workspace logs',
```
nwalters512, line 85:
```suggestion
          pageTitle: 'Workspace version logs',
```

pages/workspaceLogs/workspaceLogs.html.js
```diff
@@ -0,0 +1,239 @@
...
+                      ></span>
+                      <span class="js-button-message">Load more logs</span>
+                    </button>
+                  </div>
+                </div>
+              `
+            : html`
+                <div class="bg-dark py-5 px-2 mb-3 rounded text-white text-center text-monospace">
+                  <div class="mb-2">
+                    <i class="fa fa-calendar fa-2xl" aria-hidden="true"></i>
```
nwalters512, line 124: TODO: pick a better icon for this when container logs are disabled - maybe [`fa-ban`](https://fontawesome.com/icons/ban?s=solid&f=classic)?

pages/workspaceLogs/workspaceLogs.html.js
```diff
@@ -0,0 +1,239 @@
...
+          <h2>History</h2>
+          ${WorkspaceLogsTable({ workspaceLogs, includeVersion: false })}
+        </div>
+
+        <script>
+          $(() => {
+            const LOAD_MORE_LOGS_MESSAGE = 'Load more logs';
+            const LOADING_LOGS_MESSAGE = 'Loading logs...';
+
+            // This element will only be present if the logs haven't expired.
```
nwalters512, line 143:
```suggestion
            // This element will only be present if logs are enabled and haven't expired.
```

pages/workspaceLogs/workspaceLogs.html.js
```diff
@@ -0,0 +1,239 @@
...
+
+              const start = Date.now();
+              try {
+                let url = logsEndpoint;
+                if (startAfter) {
+                  url += '?' + new URLSearchParams({ start_after: startAfter });
+                }
+                const res = await fetch(url);
+                if (!res.ok) {
+                  // TODO: better error handling
```
nwalters512, line 170: 👀

pages/workspaceLogs/workspaceLogs.html.js
```diff
@@ -0,0 +1,239 @@
...
+              buttonMessage.textContent = LOADING_LOGS_MESSAGE;
+              loadingSpinner.hidden = false;
+
+              const start = Date.now();
+              try {
+                let url = logsEndpoint;
+                if (startAfter) {
+                  url += '?' + new URLSearchParams({ start_after: startAfter });
+                }
+                const res = await fetch(url);
```
nwalters512, line 168: Native fetch instead of `$.get` 👀

pages/workspaceLogs/workspaceLogs.html.js
```diff
@@ -0,0 +1,239 @@
...
+
+                  if (res.headers.has('x-next-start-after')) {
+                    startAfter = res.headers.get('x-next-start-after');
+                  }
+                }
+              } catch (err) {
+              } finally {
+                // Add artificial delay so that the spinner is visible for at
+                // least a second. This will also keep people from spamming the
+                // button as fast as possible.
```
nwalters512, line 184: This was an attempt to make things feel better locally when everything is really really fast and the spinner pops in and out for a fraction of a second.

pages/workspaceLogs/workspaceLogs.js
```diff
@@ -0,0 +1,168 @@
...
+ * returns null.
+ *
+ * @param {string | number} workspaceId
+ * @param {string | number} version
+ * @param {string} [startAfter]
+ *
+ * @returns {Promise<{ logs: string, startAfter: string} | null>}
+ */
+async function loadLogParts(workspaceId, version, startAfter) {
+  // TODO: handle errors and missing data.
```
nwalters512, line 28: I think this is done - have I missed anything?
```suggestion
```

pages/workspaceLogs/workspaceLogs.js
```diff
@@ -0,0 +1,168 @@
...
+
+  if (logItems.Contents.length === 0) {
+    return {
+      logs: '',
+      // If we couldn't load any more logs, reuse the same startAfter key.
+      startAfter,
+    };
+  }
+
+  const logParts = await Promise.all(
```
nwalters512, line 47: This will load up to 10 logs in parallel - should we limit the concurrency more?

Aside: I'm realizing that while "only include up to a minute's worth of logs in one object" is great for writing, it's suboptimal for reading, as to fully load the logs you'd have to make potentially hundreds of `GET` requests to reassemble the full logs. That's what makes this whole async-loading thing necessary.

I'm not sure how much better we can make this. If S3 had a way to append to files, this would all be really easy, but it doesn't. I think our options are:

- Lots of small writes (current implementation)
- One file that we periodically overwrite with the entire log history (a lot of wasted bandwidth)
- One big write as a container is shutting down (makes real-time debugging impossible)
- One big write as a container is shutting down, and *also* a way to request logs directly from live containers (API to workspace hosts on the fly)

The last option might actually be the best of both worlds - thoughts?

pages/workspaceLogs/workspaceLogs.js
```diff
@@ -0,0 +1,168 @@
...
+
+function areContainerLogsEnabled() {
+  return config.workspaceLogsS3Bucket !== null;
+}
+
+// Only instructors and admins can access these routes. We don't need to check
+// if the instructor has access to the workspace; that's already been checked
+// by the workspace authorization middleware.
+router.use((req, res, next) => {
+  // TODO: is `authn_is_instructor` the right permission to check?
```
nwalters512, line 92: 👀

pages/workspaceLogs/workspaceLogs.js
```diff
@@ -0,0 +1,168 @@
...
+    const workspaceLogs = await sqldb.queryAsync(sql.select_workspace_logs, {
+      workspace_id: res.locals.workspace_id,
+      display_timezone:
+        res.locals.course_instance.display_timezone ?? res.locals.course.display_timezone,
+    });
+    res.send(WorkspaceLogs({ workspaceLogs: workspaceLogs.rows, resLocals: res.locals }));
+  })
+);
+
+// All state transitions for a single workspace version, as wel as the container
```
nwalters512, line 114:
```suggestion
// All state transitions for a single workspace version, as well as the container
```

pages/workspaceLogs/workspaceLogs.html.js
mwest1066: Ideas:
https://icons.getbootstrap.com/icons/body-text/
https://icons.getbootstrap.com/icons/card-text/
https://icons.getbootstrap.com/icons/file-text/
https://icons.getbootstrap.com/icons/list-columns-reverse/
https://icons.getbootstrap.com/icons/receipt/

pages/workspace/workspace.ejs
jonatanschroeder: Maybe https://fontawesome.com/icons/bars-staggered?s=solid&f=classic?

pages/workspaceLogs/workspaceLogs.html.js
mwest1066: I like `fa-ban` for this.

pages/workspaceLogs/workspaceLogs.sql
```diff
@@ -0,0 +1,23 @@
...
+SELECT
+    date,
+    format_date_full_compact(date, $display_timezone) AS date_formatted,
+    message,
+    version,
+    state
+FROM workspace_logs
+WHERE
+    workspace_id = $workspace_id
+    AND version = $version
```
mwest1066, line 22: If you want to combine these the two queries in this file you can do `AND ($version IS NULL OR version = $version)` and then pass `version: null` for the case where you want all the logs.

pages/workspaceLogs/workspaceLogs.js
mwest1066: Should be fine, I think.

pages/workspaceLogs/workspaceLogs.js
mwest1066: Hopefully all this complexity will go away if you shift to the "pull from workspace_host, fall back to S3" scheme.

mwest1066 (approved):

lib/config.js
```diff
@@ -175,6 +175,15 @@ config.trustProxy = false; // Set this to a value from (https://expressjs.com/en
...
 config.workspaceLogsFlushIntervalSec = 60;
+/**
+ * The number of days after which a workspace version's logs should no longer
+ * be available. Set to `null` to disable log expiration.
+ *
+ * This useful when you want to configure the underlying S3 bucket to move
+ * logs to cheaper storage tiers or evict them entirely after a certain
+ * amount of time.
+ */
+config.workspaceLogsExpirationDays = 90;
```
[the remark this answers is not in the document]
nwalters512: Sure thing - I'll update the rules for the S3 buckets too.

pages/workspace/workspace.ejs
nwalters512: I like `bars-staggered`!

pages/workspaceLogs/workspaceLogs.html.js
nwalters512: Done in 8c73f53.

pages/workspace/workspace.ejs
nwalters512: Changed in 8c73f53.

pages/workspaceLogs/workspaceLogs.js
nwalters512: Yes, it should!

pages/workspaceLogs/workspaceLogs.sql
nwalters512: Done in 3ec08ea!

pages/workspaceLogs/workspaceLogs.js
nwalters512: I made that change in the most recent few commits (https://github.com/PrairieLearn/PrairieLearn/pull/6638/files/8c73f53f90d30db4e25c1abc179276971f4ca4ee..2de19e9cde7eb9b2be54263047ab20f8e9dc6406) - it feels overall much simpler and more robust.

pages/workspaceLogs/workspaceLogs.js
nwalters512: And indeed it did 🎉

pages/workspaceLogs/workspaceLogs.html.js
nwalters512: This code no longer exists.

pages/workspaceLogs/workspaceLogs.html.js
nwalters512: This code no longer exists.

pages/workspace/workspace.js
```diff
@@ -24,34 +24,29 @@ router.get('/', (_req, res, _next) => {
...
       await sqldb.queryAsync(sql.increment_workspace_version, { workspace_id });
       res.redirect(`/pl/workspace/${workspace_id}`);
     } else {
-      return next(
-        error.make(400, 'unknown action', {
-          locals: res.locals,
-          body: req.body,
-        })
-      );
+      return next(error.make(400, 'unknown __action'));
```
mwest1066, line 49:
```suggestion
      return next(error.make(400, `unknown __action: ${req.body.__action}`));
```

pages/workspaceLogs/workspaceLogs.js
```diff
@@ -0,0 +1,166 @@
...
+    }
+  }
+
+  // If the current workspace version matches the requested version, we can
+  // reach out to the workspace host directly to get the remaining logs. Otherwise,
+  // they should have been flushed to S3 already.
+  //
+  // To avoid a race condition where a workspace shuts down by the time we reach
+  // out to the host, we'll fall back to attempting to load the logs from S3 if
+  // the host doesn't have any logs.
```
mwest1066, line 90: I don't see anything below where we are explicitly falling back to reading from S3? It looks like we just "read everything from S3" and then "read from host". I think the code is fine like it is, but this comment was just a bit confusing to me.

workspace_host/interface.js
```diff
@@ -1440,3 +1418,66 @@ async function sendGradedFilesArchive(workspace_id, res) {
...
+}
+
+/**
+ * @param {import('dockerode').Container} container
+ */
+async function flushLogsToS3(container) {
+  if (!config.workspaceLogsS3Bucket) return;
+
+  // Read all data from the container's labels instead of trying to look it up
+  // in the database, as some things (namely the version) may have changed.
```
mwest1066, line 1447: Good point on the version change!

workspace_host/interface.js
```diff
@@ -1440,3 +1418,66 @@ async function sendGradedFilesArchive(workspace_id, res) {
...
+  const institutionId = containerInfo.Config.Labels['prairielearn.institution-id'];
+  // We use the version recorded in the container labels instead of in the
+  // workspace row from the database. We might be flushing these logs sometime
+  // after the workspace has been relaunched in a new version, and we want to
+  // associate the logs with the correct (older) version.
+  const version = containerInfo.Config.Labels['prairielearn.workspace-version'];
+
+  // For any container versions A and B, if A < B, then the `StartedAt` date for A
+  // should be before the `StartedAt` date for B. This means that we can use the
+  // date for ordering of logs from different versions.
```
mwest1066, line 1461: Nice.

mwest1066 (approved):
This Is awesome! I like the new scheme of pulling from host+S3 a lot. Much cleaner code.

pages/workspaceLogs/workspaceLogs.js
mwest1066: Agreed, this new code is a definite improvement.

pages/workspaceLogs/workspaceLogs.js
mwest1066: So nice!

pages/workspaceLogs/workspaceLogs.js
nwalters512: Good catch! This comment is just out of date. Originally I had it read only from the host _or_ S3, but then I realized that I always need to do both for multi-version workspaces. I'll update this comment to reflect what's actually happening in the latest code.

nwalters512: Fixed in 252279f.

PR #6640, opened 2022-11-16
Improvements in Java grader reliability

### LINKED ISSUE #6639: Java grader does not return an appropriate result if it crashes due to OS issues
As seen in https://ca.prairielearn.com/pl/course_instance/2450/instance_question/35797995/, if the Java autograder fails due to issues outside Java (in this case, the student created too many threads and the process ran out of threads allowed by the OS), the result file will contain a partial set of tests, causing an invalid percentage result. This issue seems to be similar to #6610, but in Java.

jonatanschroeder:
Resolves #6639. This PR causes the list of tests to be retrieved before all tests are run. This ensures that, if any of the tests fails to report a result due to errors like out-of-memory or thread interruption, they are still accounted for in the points total and list of tests.

### DIFF
--- a/graders/java/JUnitAutograder.java
+++ b/graders/java/JUnitAutograder.java
@@ -1,5 +1,6 @@
 import org.json.simple.JSONArray;
 import org.json.simple.JSONObject;
+import org.junit.platform.engine.DiscoverySelector;
 import org.junit.platform.engine.TestExecutionResult;
 import org.junit.platform.engine.TestTag;
 import org.junit.platform.engine.support.descriptor.MethodSource;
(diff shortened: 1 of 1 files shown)


jonatanschroeder:
Simple way to test it: in the example course, for the Java autograder test, include a line like:
```java
int[] a = new int[1000000000];
```

graders/java/JUnitAutograder.java
```diff
@@ -61,78 +71,129 @@ public static void main(String args[]) {
...
+                                                "failing early tests, or by an early autograder crash.")
+                             .toJson());
+        }
+
         JSONObject results = new JSONObject();
-        results.put("score", this.maxPoints > 0 ? this.points / this.maxPoints : 0);
-        results.put("points", this.points);
-        results.put("max_points", this.maxPoints);
+        results.put("score", maxPoints > 0 ? Math.min(this.points / maxPoints, 1) : 0);
+        results.put("points", Math.max(this.points, maxPoints));
```
mwest1066, line 195: Should this be a `Math.min`?

jonatanschroeder: Good catch. This field is usually ignored in the results, which is why it didn't affect the results. Fixed in 9585c3d.

PR #6641, opened 2022-11-17
Introduce pl-big-o-input element 

eliotwrobson:
BigO (family) input element that gives detailed feedback about asymptotic input and allows instructors to ask questions about this. Similar semantically to the `pl-symbolic-input` element (which it shares a lot of backend with).


SethPoulsen:
This is awesome. 

Do we think it would make sense to have this just be extra functionality on the `pl-symbolic-input` element? Maybe not, I guess the semantics are a bit different. 

Either way, we're going to want to merge `python_helper_sympy.py` with parts of your `big_o_utils.py` somehow, we aren't going to want to have duplicated code. Maybe just add the functionality that you added into `python_helper_sympy.py`, then just call functions from there?

eliotwrobson:
> Do we think it would make sense to have this just be extra functionality on the `pl-symbolic-input` element? Maybe not, I guess the semantics are a bit different.

I think the semantics are different enough that it makes sense to have a different element altogether. This element also has a different display and custom partial credit scheme, which would make merging these elements tricky.

> Either way, we're going to want to merge `python_helper_sympy.py` with parts of your `big_o_utils.py` somehow, we aren't going to want to have duplicated code. Maybe just add the functionality that you added into `python_helper_sympy.py`, then just call functions from there?

This actually might be quite difficult, and may even be beyond the scope of this PR. The issue is that for this element, we want to disallow certain functions (like the trig functions, for example), but in `pl-symbolic-input`, these functions are all allowed and this is hardcoded. This suggests that the best approach would be to refactor the backend for `pl-symbolic-input` to allow for custom functions, but refactoring this could cause problems because there aren't any tests to make sure that backend functions as expected (and, from a cursory glance, it looks like the js tests for the symbolic input element have been disabled, see below).

image

In short, I think that refactoring this code to work for both use cases is actually pretty easy, but I really don't want to risk breaking something important as part of this PR (which is hard to ensure since there are few safeguards against this).

SethPoulsen:
I think the refactor should be done now, rather than merging duplicate code into the repository and then refactoring later.

Given that its just a matter of enabling/disabling certain functions, it seems like a pretty low-risk refactor. (We should add tests too though!).

SethPoulsen:
If you really think that this should be done in separate PRs, then I would vote for fixing the Symbolic input tests and merging that first, then merging this one.

eliotwrobson:
> If you really think that this should be done in separate PRs, then I would vote for fixing the Symbolic input tests and merging that first, then merging this one.

I'm very much in favor of this, since that simplifies the code needed for this element considerably (we would just need to call the sympy evaluation with a different set of allowable functions).

eliotwrobson:
Per discussion above, tests will be updated on this PR: https://github.com/PrairieLearn/PrairieLearn/pull/6646

docs/elements.md
```diff
@@ -620,6 +622,52 @@ Do not include `i` or `j` in the list of `variables` if `allow-complex="true"`.
 
 ---
 
+### `pl-symbolic-input` element
```
SethPoulsen, line 625: copy-paste error?

python/prairielearn_test.py
```diff
@@ -1,15 +1,124 @@
...
+)
+def test_grade_question_parametrized_correct(
+    question_data: pl.QuestionData,
+    question_name: str,
+    student_answer: str,
+    weight: int,
+    expected_grade: bool,
+) -> None:
+
+    question_data["submitted_answers"] = {question_name: student_answer}
```
SethPoulsen, line 43: why are the changes in this file part of this PR?

python/conftest.py
```diff
@@ -0,0 +1,22 @@
+
+import pytest
+from prairielearn import QuestionData
+
+@pytest.fixture
+def question_data() -> QuestionData:
+    return {
+        "params": dict(),
```
SethPoulsen, line 8: whats this for?

python/prairielearn_test.py
SethPoulsen: oh, I see, you added a new function to `prairielearn.py`.

You should be really careful what type annotations are added to this file, because the typechecking is not running on this file, and the unit tests are being run on a different python version. So we could break production without realizing it. 

Alternatively, enable type checking on this file. But that may require some out-of-scope changes elsewhere in the file. A third option could be to leave the type annotations out of this file for now and then add in a seperate PR, while fixing up the rest of the file for type checking.

python/prairielearn.py
```diff
@@ -52,6 +53,78 @@ class QuestionData(TypedDict):
 class ElementTestData(QuestionData):
     test_type: Literal['correct', 'incorrect', 'invalid']
 
+def grade_question_parameterized(data: QuestionData,
+                                 question_name: str,
+                                 grade_function: Callable[[Any], Tuple[Union[bool, float], Optional[str]]],
+                                 weight: int = 1) -> None:
+    '''
+    Grade question question_name, marked correct if grade_function(student_answer) returns True in
+    its first argument. grade_function should take in a single parameter (which will be the submitted
```
SethPoulsen, line 62: the 'in its first argument' part here is confusing. don't you mean, 'in the first element of the tuple that is returned?'

python/prairielearn.py
```diff
@@ -52,6 +53,78 @@ class QuestionData(TypedDict):
...
+        - The second element of the 2-tuple should either be:
+            - a string containing feedback
+            - None, if there is no feedback (usually this should only occur if the answer is correct)
+    '''
+
+    # Create the data dictionary at first
+    data['partial_scores'][question_name] = {
+            'score': 0.0,
+            'weight': weight
+    }
```
SethPoulsen, line 76: weird indentation here

python/prairielearn.py
```diff
@@ -52,6 +53,78 @@ class QuestionData(TypedDict):
...
+        data["format_errors"][question_name] = str(err)
+        return
+
+
+    # Set question score if grading succeeded
+    data['partial_scores'][question_name]['score'] = partial_score
+
+    # Put all feedback here
+    if feedback_content:
+        data['partial_scores'][question_name]['feedback'] = feedback_content
```
SethPoulsen, line 109: I think all these comments should be removed, the code is clear enough that the comments don't add anything, they just get in the way.

python/prairielearn.py
```diff
@@ -52,6 +53,78 @@ class QuestionData(TypedDict):
...
+
+def determine_score_params(score: Optional[float]) -> Tuple[str, float]:
```
SethPoulsen, line 111: Why do we want to allow score to be `None`?

python/python_helper_sympy.py
```diff
@@ -1,30 +1,70 @@
...
+    _value: str
+    _variables: List[str]
+
+class LocalsForEval(TypedDict):
+    "A class with type signatures for the locals_for_eval dict"
+
+    functions: SympyMapT
+    variables: SympyMapT
+    helpers: SympyMapT
+
```
SethPoulsen, line 22: inconsistent spacing between these two class definitions.

python/python_helper_sympy.py
```diff
@@ -1,30 +1,70 @@
 import sympy
 import ast
-import sys
+from dataclasses import dataclass
+from itertools import chain
+from typing import Any, cast, Dict, List, Tuple, TypedDict, Literal, Optional
+
+SympyMapT = Dict[str, Any]
```
SethPoulsen, line 7: could the `Any` here be `Union[Callable, Float]`? or am I missing something?

python/python_helper_sympy.py
```diff
@@ -1,30 +1,70 @@
...
+ASTWhitelistT = Tuple[Any, ...]
```
SethPoulsen, line 8: All AST nodes probably inherit from some ast node type, right? Should be able to do `Tuple[ast.AST, ...]` or something? Actually, wouldn't it make more sense for it to be `List[ast.AST]` anyway?

python/python_helper_sympy.py
```diff
@@ -85,116 +122,130 @@ def __init__(self):
...
+            raise HasComplexError(node.col_offset, str(node.n))
         return node
 
 class CheckWhiteList(ast.NodeVisitor):
-    def __init__(self, whitelist):
+    def __init__(self, whitelist: ASTWhitelistT) -> None:
         self.whitelist = whitelist
 
-    def visit(self, node):
+    def visit(self, node: Any) -> None:
```
SethPoulsen, line 184: should be `node: ast.AST`?

python/python_helper_sympy.py
```diff
@@ -85,116 +122,130 @@ def __init__(self):
...
+    def visit_Name(self, node: Any) -> None:
         if isinstance(node.ctx, ast.Load):
             if not is_name_of_function(node):
                 if node.id not in self.variables:
                     node = get_parent_with_location(node)
                     raise HasInvalidVariableError(node.col_offset, node.id)
         self.generic_visit(node)
 
-def is_name_of_function(node):
+def is_name_of_function(node: Any) -> bool:
```
SethPoulsen, line 213: should be `node: ast.AST`?

python/python_helper_sympy.py
```diff
@@ -85,116 +122,130 @@ def __init__(self):
...
     # 3) it is not in the list of parent's args
-    return isinstance(node, ast.Name) and isinstance(node.parent, ast.Call) and (node not in node.parent.args)
+    return isinstance(node, ast.Name) and isinstance(node.parent, ast.Call) and (node not in node.parent.args) # type: ignore
 
-def get_parent_with_location(node):
+def get_parent_with_location(node: ast.AST) -> ast.AST:
     if hasattr(node, 'col_offset'):
         return node
-    else:
-        return get_parent_with_location(node.parent)
```
SethPoulsen, line 176: seems like an unnecessary change to make

python/python_helper_sympy.py
```diff
@@ -85,116 +122,130 @@ def __init__(self):
...
+    ind = expr.find('\\')
+    if ind != -1:
+        raise HasEscapeError(ind)
 
     # Disallow comment character
-    if '#' in expr:
-        raise HasCommentError(expr.find('#'))
+    ind = expr.find('#')
+    if ind != -1:
+        raise HasCommentError(ind)
```
SethPoulsen, line 236: why change this?

python/python_helper_sympy.py
```diff
@@ -85,116 +122,130 @@ def __init__(self):
...
 
     # Parse (convert string to AST)
     try:
         root = ast.parse(expr, mode='eval')
-    except Exception as err:
-        raise HasParseError(err.offset)
+    except SyntaxError as err:
+        offset = err.offset if err.offset is not None else -1
+        raise HasParseError(offset)
```
SethPoulsen, line 244: could we miss catching certain Exceptions with this change?

python/python_helper_sympy.py
```diff
@@ -85,116 +122,130 @@ def __init__(self):
...
 
     # Link each node to its parent
     for node in ast.walk(root):
         for child in ast.iter_child_nodes(node):
-            child.parent = node
+            child.parent = node  # type: ignore
```
SethPoulsen, line 248: why ignore here?

python/python_helper_sympy.py
```diff
@@ -221,64 +273,70 @@ def evaluate(expr, locals_for_eval={}):
 
     # Convert AST to code and evaluate it with no global expressions and with
     # a whitelist of local expressions
-    locals = {}
-    for key in locals_for_eval:
-        locals = {**locals, **locals_for_eval[key]}
+    locals = dict(chain.from_iterable(
+        cast(SympyMapT, local_expressions).items() for local_expressions in locals_for_eval.values()
+    ))
```
SethPoulsen, line 278: why this change? I think the original was more readable.

python/python_helper_sympy.py
```diff
@@ -221,64 +273,70 @@ def evaluate(expr, locals_for_eval={}):
...
+
     return eval(compile(root, '<ast>', 'eval'), {'__builtins__': None}, locals)
 
-def convert_string_to_sympy(a, variables, allow_hidden=False, allow_complex=False):
+def convert_string_to_sympy(expr: str, variables: Optional[List[str]], *, allow_hidden: bool=False, allow_complex: bool=False, allow_trig_functions: bool=True) -> sympy.Expr:
```
SethPoulsen, line 282: do you need the `bool` annotations here, or does having the default be `False` make the typechecker assume bool? I think typescript does this, IDK about python.

python/python_helper_sympy.py
SethPoulsen: Actually, I don't know if we should be bothering to define these at all. They are only used once, and they are defined by a dict literal, so a type inferencer should be able to do all the type checking just fine without this extra clutter.

python/python_helper_sympy.py
```diff
@@ -1,30 +1,70 @@
...
 # is to avoid accidentally modifying these dictionaries.
 class _Constants:
-    def __init__(self):
+    helpers: SympyMapT
+    variables: SympyMapT
+    hidden_variables: SympyMapT
+    complex_variables: SympyMapT
+    hidden_complex_variables: SympyMapT
+    functions: SympyMapT
+    trig_functions: SympyMapT
```
SethPoulsen, line 33: do these need to be here?

SethPoulsen (commented):
No big issues, just lots of small comments and questions.

python/python_helper_sympy.py
eliotwrobson: You do need the bool. Anything without annotations is automatically treated as an `Any`.

python/python_helper_sympy.py
eliotwrobson: The original is wasteful (recreates the dictionary several times). There's a more readable way to write this though, so I'll use that.

python/python_helper_sympy.py
eliotwrobson: The `type: ignore` statements are here because `parent` is not part of the declared class members for an ast Node, so the typechecker will get mad trying to access those. Avoiding the ignores requires making everything an inherited class that has this parameter (or something else weird like that).

python/python_helper_sympy.py
eliotwrobson: Technically yes but a generic exception isn't guaranteed to have an `offset`, so trying to access that will just throw another exception. Either way, in Python you should always avoid catching a generic exception if possible (since this will catch sys calls and stuff).

python/python_helper_sympy.py
eliotwrobson: The old code read through the string twice needlessly. Since we're changing this file anyway, might as well fix all of the small issues like this.

python/python_helper_sympy.py
eliotwrobson: We have to change this line anyway to add the type ignore comment.

python/python_helper_sympy.py
SethPoulsen: I disagree that it was an issue. 

The performance difference of reading the string again is negligible, only happens in the error case, and the old code was much more readable.

python/python_helper_sympy.py
SethPoulsen: that's unfortunate.

python/python_helper_sympy.py
eliotwrobson: Type inferences are always treated as generously as possible. Without this, the only type hint we can give is for a `Dict[str, Any]` (in fact, I think the inferred type may just be a `Dict`). This is actually why the typechecker will usually ask you for an explicit type hint on every dict that gets created.

python/prairielearn.py
eliotwrobson: This is used to read the score directly from the data dict, which may be None. Not allowing the none here means we just have to check for it at the call site for this function anyway.

python/python_helper_sympy.py
SethPoulsen: this is very annoying compared to TypeScript, where the the typechecker would just infer the type from the literal.

python/conftest.py
eliotwrobson: This gets used by some of the new tests. Having an argument named `question_data` in a test function automatically creates a new dict by calling this function. See https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session

docs/elements.md
```diff
@@ -620,6 +622,52 @@ Do not include `i` or `j` in the list of `variables` if `allow-complex="true"`.
...
+| `correct-answer` | string                                                | -        | Correct answer for grading.                                                                                 |
+| `display`        | "block" or "inline"                                   | "inline" | How to display the input field.                                                                             |
+| `variables`      | string                                                | —        | A comma-delimited list of symbols that can be used in the symbolic expression. Only one variable supported. |
+| `size`           | integer                                               | 35       | Size of the input box.                                                                                      |
+| `show-help-text` | boolean                                               | true     | Show the question mark at the end of the input displaying required input parameters.                        |
+
+#### Details
+
+Correct answers must be specified as strings, e.g., `n**2`. Only one variable is supported at the moment.
```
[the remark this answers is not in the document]
eliotwrobson: This actually doesn't use Python syntax (so `n^2` is allowed, the same as `pl-symbolic-input`).

python/python_helper_sympy.py
eliotwrobson: The tuple enforces immutability. I will try and get the base ast node to work though.

python/python_helper_sympy.py
eliotwrobson: Without these the types of the instance variables cannot be inferred.

python/prairielearn_test.py
eliotwrobson: Added typechecking to this file, although I actually don't think it matters in this case (unless test files are running in production for some reason).

python/python_helper_sympy.py
eliotwrobson: Yes, all Python typecheckers are written to make interop with non-type-hinted code possible without changes. The downside of this is having to be very explicit with everything.

python/prairielearn_test.py
SethPoulsen: oh, right. I guess this comment of mine should have been applied to `prairielearn.py`, which is run in production.

eliotwrobson: Yes. The type signatures from that are taken from our codebase, which is already run in production.

Makefile
```diff
@@ -64,7 +64,7 @@ typecheck: typecheck-js typecheck-python
 typecheck-js:
 	@yarn tsc
 typecheck-python:
-	@yarn pyright
+	@yarn pyright --skipunannotated
```
eliotwrobson, line 67: To clarify in case there are any questions, this flag is set so that the `prairielearn.py` file can have typechecking enabled while not having to fix all of the type issues in that file. Annoyingly, this flag cannot be set in the config file.

pyrightconfig.json
```diff
@@ -1,9 +1,12 @@
 {
     "include": [
       "./elements/pl-hidden-hints",
-      "./elements/pl-order-blocks",
```
eliotwrobson, line 4: Removed since functions without annotations are now ignored (and without them, the typechecking does almost nothing anyway).

exampleCourse/questions/element/bigOInput/question.html
```diff
@@ -0,0 +1,139 @@
...
+
+<div class="card my-2">
+    <div class="card-header">
+        An example with polynomial complextiy.
+    </div>
+    <div class="card-body">
+        <pl-question-panel>
+            <p>Submit $O(n^2)$.</p>
+        </pl-question-panel>
+        <pl-big-o-input answers-name="q5" variables="n" correct-answer="n**2" ></pl-big-o-input>
```
nwalters512, line 89:
```suggestion
        <pl-big-o-input answers-name="q5" variables="n" correct-answer="n**2"></pl-big-o-input>
```

exampleCourse/questions/element/bigOInput/question.html
```diff
@@ -0,0 +1,139 @@
...
+
+<div class="card my-2">
+    <div class="card-header">
+        An example with exponential complextiy.
+    </div>
+    <div class="card-body">
+        <pl-question-panel>
+            <p>Submit $O(2^n)$.</p>
+        </pl-question-panel>
+        <pl-big-o-input answers-name="q6" variables="n" correct-answer="2**n" ></pl-big-o-input>
```
nwalters512, line 109:
```suggestion
        <pl-big-o-input answers-name="q6" variables="n" correct-answer="2**n"></pl-big-o-input>
```

exampleCourse/questions/element/bigOInput/question.html
```diff
@@ -0,0 +1,139 @@
...
+
+<div class="card my-2">
+    <div class="card-header">
+        An example with factorial complextiy.
+    </div>
+    <div class="card-body">
+        <pl-question-panel>
+            <p>Submit $O(n!)$.</p>
+        </pl-question-panel>
+        <pl-big-o-input answers-name="q7" variables="n" correct-answer="factorial(n)" ></pl-big-o-input>
```
nwalters512, line 129:
```suggestion
        <pl-big-o-input answers-name="q7" variables="n" correct-answer="factorial(n)"></pl-big-o-input>
```

images/plbase/python-requirements.txt
```diff
@@ -28,3 +27,4 @@ openpyxl==3.0.10
 pytest==7.2.0
 typing-extensions==4.4.0
 psutil==5.9.4
+black==22.10.0
```
nwalters512, line 30: 👀 did you mean to invoke Black somewhere, perhaps in the Makefile?

pyrightconfig.json
```diff
@@ -1,9 +1,12 @@
...
       "./elements/pl-symbolic-input",
-      "python/python_helper_sympy_test.py"
+      "./elements/pl-big-o-input",
+      "python/python_helper_sympy.py",
```
nwalters512, line 6: Let's be consistent with using a leading dot-slash (or not).

python/conftest.py
```diff
@@ -0,0 +1,22 @@
+
```
nwalters512, line 1:
```suggestion
```

python/prairielearn.py
```diff
@@ -52,6 +54,74 @@ class QuestionData(TypedDict):
 class ElementTestData(QuestionData):
     test_type: Literal['correct', 'incorrect', 'invalid']
 
+class DisplayType(Enum):
```
nwalters512, line 57: Does this really belong here?

python/prairielearn.py
```diff
@@ -52,6 +54,74 @@ class QuestionData(TypedDict):
...
+    INLINE = "inline"
+    BLOCK = "block"
+
+def grade_question_parameterized(data: QuestionData,
```
nwalters512, line 61: Is this used outside of this new element? If not, let's keep it there for now - I want to avoid adding things to this public API unless there's value in doing so.

python/prairielearn.py
```diff
@@ -52,6 +54,74 @@ class QuestionData(TypedDict):
...
+        data["format_errors"][question_name] = str(err)
+        return
+
+
+    data['partial_scores'][question_name]['score'] = partial_score
+
+    if feedback_content:
+        data['partial_scores'][question_name]['feedback'] = feedback_content
+
+def determine_score_params(score: Optional[float]) -> Tuple[str, float]:
```
nwalters512, line 110: Same question for this function: if it's only used in `pl-big-o-input`, let's not add it here.

elements/pl-big-o-input/pl-big-o-input.py
```diff
@@ -0,0 +1,317 @@
...
+            "size": size,
+            "show_info": pl.get_boolean_attrib(
+                element, "show-help-text", SHOW_HELP_TEXT_DEFAULT
+            ),
+            "uuid": pl.get_uuid(),
+            display.value: True,
+            "show_placeholder": size >= PLACEHOLDER_TEXT_THRESHOLD,
+            "raw_submitted_answer": raw_submitted_answer,
+            "type": bigo_type,
+            score_type: True,
```
nwalters512, line 140: 👀 What's this? Ditto for the submission panel below.

docs/elements.md
```diff
@@ -620,6 +622,48 @@ Do not include `i` or `j` in the list of `variables` if `allow-complex="true"`.
...
+| `answers-name`   | string                                                | —        | Variable name to store data in.                                                                             |
+| `type`           | "big-o", "theta", "omega", "little-o", "little-omega" | "big-o"  | Type of asymptotic answer required.                                                                         |
+| `weight`         | integer                                               | 1        | Weight to use when computing a weighted average score over elements.                                        |
+| `correct-answer` | string                                                | -        | Correct answer for grading.                                                                                 |
+| `display`        | "block" or "inline"                                   | "inline" | How to display the input field.                                                                             |
+| `variables`      | string                                                | —        | A comma-delimited list of symbols that can be used in the symbolic expression. Only one variable supported. |
```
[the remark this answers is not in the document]
eliotwrobson: Ah good catch. We _might_ support other variables in the future (which is why the internal name for this is `variables`), but we don't want to expose that to the user. Will change.

images/plbase/python-requirements.txt
eliotwrobson: So the more I think about it, I should probably remove this line until the backend has been black-ified. My original intent was to have it installed so people can run it if they choose and know what they're doing, but this actually doesn't seem like a good idea.

I have another PR where I actually added this to the makefile. The issue is, adding that command to the makefile runs it over the whole codebase, and that leads to a huge number of changes. I don't think we want people going out of their way to run it like that until it's been run over the whole backend, and running it doesn't lead to a giant diff.

nwalters512: Yep - let's hold off on installing Black or running it on any files until that other PR where we install + run it on everything.

elements/pl-big-o-input/pl-big-o-input.py
eliotwrobson: Setting the score type? Basically, that's getting the output of the corresponding helper function `determine_score_params` and adding it to the display dictionary. It's serving the same purpose as the code block below (that appears duplicated in the code for basically every element):

image

When writing this, we thought it was a good idea to consolidate this logic. As an aside, the exception catch and re-throw in the above block is unnecessary. We think that using this helper function can really be used by every element to replace the block above (since that just contributes to code bloat).

python/prairielearn.py
eliotwrobson: Ah I did this by mistake. This can actually be used in an arbitrary element for grading (including some of the others we have in PR, which is why this is here). But I'll wait to move this here as part of those later PRs.

images/plbase/python-requirements.txt
eliotwrobson: Sounds good. I'll run it on the files that I edit heavily but for everything else, it makes sense to wait and make sure to run it manually in safe mode.

docs/elements.md
```diff
@@ -24,6 +24,8 @@ PrairieLearn presently provides the following templated **input field** elements
   such as -71, 0, 5, 21, and so on.
 - [`pl-symbolic-input`](#pl-symbolic-input-element): Fill in a **symbolic** value
   such as `x^2`, `sin(z)`, `mc^2`, and so on.
+- [`pl-big-o-input`](#pl-big-o-input-element): Fill in a **symbolic** value
+  representing asymptotic input.
```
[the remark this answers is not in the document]
eliotwrobson: Not necessarily true. Big-O family notation gets used in combinatorics (like bounding the size of a graph) and differential equations.

elements/pl-big-o-input/big_o_utils.py
```diff
@@ -0,0 +1,190 @@
...
+
+    sym_true = phs.convert_string_to_sympy(
+        a_true, variables, allow_complex=False, allow_trig_functions=False
+    )
+    sym_sub = phs.convert_string_to_sympy(
+        a_sub, variables, allow_complex=False, allow_trig_functions=False
+    )
+
+    if sym_true.equals(sym_sub):
+        return (1, CORRECT_COMPLEX_FEEDBACK)
```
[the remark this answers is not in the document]
eliotwrobson: Currently yes. Looking at this again, we can actually check for structural equality above instead of string equality (since structural doesn't evaluate expressions like `1+1` as equal to `2`). Will change.

elements/pl-big-o-input/pl-big-o-input.py
```diff
@@ -0,0 +1,317 @@
...
+GRADE_FUNCTION_DICT: Dict[BigOType, bou.BigOGradingFunctionT] = {
+    BigOType.BIG_O: bou.grade_big_o_expression,
+    BigOType.THETA: bou.grade_theta_expression,
+    BigOType.OMEGA: bou.grade_omega_expression,
+    BigOType.LITTLE_O: bou.grade_little_o_expression,
+    BigOType.LITTLE_OMEGA: bou.grade_little_omega_expression,
+}
+
+
+VARIABLES_DEFAULT = None
```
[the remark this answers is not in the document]
eliotwrobson: If asking for a constant time algorithm (i.e. answer is `O(1)`).

elements/pl-big-o-input/pl-big-o-input.py
```diff
@@ -0,0 +1,317 @@
...
+
+    info_params = {
+        "format": True,
+        "variables": variables,
+        "operators": operators,
+        "constants": constants,
+    }
+
+    PARTIAL_SCORE_DEFAULT: pl.PartialScore = {"score": None}
+    score = data["partial_scores"].get(name, PARTIAL_SCORE_DEFAULT).get("score")
```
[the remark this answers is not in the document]
eliotwrobson: This seems to be ok. Depends on how the typechecking is configured (the way ours is set up, it complains without this).

docs/elements.md
```diff
@@ -620,6 +622,48 @@ Do not include `i` or `j` in the list of `variables` if `allow-complex="true"`.
...
+| `show-help-text` | boolean                                               | true     | Show the question mark at the end of the input displaying required input parameters.                        |
+
+#### Details
+
+Correct answers must be specified as strings with Python syntax (e.g., `n**2`, `2**n`, `n * log(n)`), with
+the same syntax as [`pl-symbolic-input`](#pl-symbolic-input-element). Only one variable is supported.
+
+#### Example implementations
+
+- [element/bigOInput]
```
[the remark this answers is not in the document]
eliotwrobson: Done, also added missing link for hidden hints question.

elements/pl-big-o-input/big_o_utils.py
```diff
@@ -0,0 +1,190 @@
...
+
+    if sympy.limit(sym_sub, sympy.Symbol(variables[0]), sympy.oo) < sympy.sympify(0):
```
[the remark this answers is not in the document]
eliotwrobson: They got a bit too clever with this 😅

docs/elements.md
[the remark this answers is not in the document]
nwalters512: Ah, fair point!

docs/elements.md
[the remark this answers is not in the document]
nwalters512: Gotcha - `variable` makes more sense for now. Is there anything standing in the way of easily supporting multiple variables at the moment?

elements/pl-big-o-input/big_o_utils.py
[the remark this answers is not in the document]
nwalters512: If we can check for structural equality, is there a reason we wouldn't just do that instead of checking for string equality?

docs/elements.md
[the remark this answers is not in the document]
eliotwrobson: The issue is the grading algorithm. For example, given n and m for n <= m <= n**2, we haven't figured out a nice way of delivering some kind of feedback in this case.

elements/pl-big-o-input/pl-big-o-input.py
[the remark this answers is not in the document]
nwalters512: Doh - of course.

elements/pl-big-o-input/pl-big-o-input.py
nwalters512: Oh man, sorry, this is my JavaScript brain not differentiating between `"object_key"` and `object_key`. I see what's happening now, there's not a literal `"score_type"` key in the dict.

elements/pl-big-o-input/big_o_utils.py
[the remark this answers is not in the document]
eliotwrobson: So the issue is that structural equality simplifies literals. For example, if you input `(1+1)**n` and the reference solution is `2**n`, these would be treated as equal under structural equality using `==`. We haven't found a way of comparison that avoids this simplification.

However, if we find something that works in this case, it's very straightforward to change the comparison to this and add tests.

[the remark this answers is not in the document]
nwalters512: Ah, ok - that's unfortunate. I've spent approximately zero time checking how feasible this is, but I imagine you could improve over simple string equality by walking the AST and comparing nodes, which would in theory avoid the problem of `n**2` and `n**(2)` not being equivalent. Using `parse_expr('..', evaluate=False)` and a traversal like the one described in https://stackoverflow.com/questions/59816730/run-through-the-ast-of-an-expression-in-sympy might get you pretty close. (Not requesting that you implement here, just food for thought.)

docs/elements.md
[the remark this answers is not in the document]
nwalters512: Gotcha, that makes sense. Do you have any questions that ask about asymptotic behavior with multiple variables? I'm thinking about e.g. graph algorithms where you often have $V$ and $E$ terms. I suppose you could do that just with the standard `pl-symbolic-input` element, just without the fancier feedback?

elements/pl-big-o-input/big_o_utils.py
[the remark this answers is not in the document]
eliotwrobson: I think that sounds reasonable as a feature to add later on. Right now, most runtimes are only a single term (something like `n**3`), so removing just spaces gets us pretty close already.

docs/elements.md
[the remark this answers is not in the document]
eliotwrobson: Yes, using just symbolic input is what we're doing currently.

elements/pl-big-o-input/big_o_utils.py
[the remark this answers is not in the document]
nwalters512: Right, makes sense! As long as students get full credit for a correct answer (which they currently do), that's all that really matters 🙂

docs/elements.md
```diff
@@ -620,6 +622,48 @@ Do not include `i` or `j` in the list of `variables` if `allow-complex="true"`.
...
 
+### `pl-big-o-input` element
+
+Fill in the blank field that allows for asymptotic mathematical input (i.e. big O, big Theta, etc.).
+Gives automated feedback in the case of improper asymptotic input.
+
+**question.html**
+
+```html
+<pl-big-o-input answers-name="ans" variables="n" correct-answer="n**2" size="10"></pl-big-o-input>
```
nwalters512, line 633:
```suggestion
<pl-big-o-input answers-name="ans" variable="n" correct-answer="n**2" size="10"></pl-big-o-input>
```

eliotwrobson:
Note to self: Remove the optional check from the score params function.

elements/pl-big-o-input/big_o_utils.py
```diff
@@ -0,0 +1,252 @@
...
+        return (0.25, TOO_LOOSE_FEEDBACK)
+    elif L == sympy.sympify(0):
+        return (0, INCORRECT_FEEDBACK)
+    elif L == sympy.sympify(1):
+        return (0.5, LOWER_ORDER_TERMS_FEEDBACK)
+
+    return (0.5, CONSTANT_FACTORS_FEEDBACK)
+
+
+# Start of generic utilities. Move these to prairieleran.py as-needed
```
stevenxpan, line 151:
```suggestion
# Start of generic utilities. Move these to prairielearn.py as-needed
```

elements/pl-big-o-input/pl-big-o-input.py
```diff
@@ -0,0 +1,310 @@
...
+    constants = list(constants_class.variables.keys())
+
+    info_params = {
+        "format": True,
+        "variable": variables,
+        "operators": operators,
+        "constants": constants,
+    }
+
+    score = data["partial_scores"].get(name, {"score": None}).get("score")
```
stevenxpan, line 104:
```suggestion
    score = data["partial_scores"].get(name, {}).get("score")
```

python/python_helper_sympy.py
```diff
@@ -221,71 +325,206 @@ def evaluate(expr, locals_for_eval={}):
 
     # Convert AST to code and evaluate it with no global expressions and with
     # a whitelist of local expressions
-    locals = {}
-    for key in locals_for_eval:
-        locals = {**locals, **locals_for_eval[key]}
-    return eval(compile(root, '<ast>', 'eval'), {'__builtins__': None}, locals)
+    locals = {
```
stevenxpan, line 328: I think it would be good to separate this out into a `for` loop, similar to the original code.

python/python_helper_sympy.py
```diff
@@ -221,71 +325,206 @@ def evaluate(expr, locals_for_eval={}):
...
     # Do the conversion
-    return evaluate(a, locals_for_eval)
+    return evaluate(expr, locals_for_eval)
+
 
-def point_to_error(s, ind, w = 5):
-    w_left = ind - max(0, ind-w)
-    w_right = min(ind+w, len(s)) - ind
-    return s[ind-w_left:ind+w_right] + '\n' + ' '*w_left + '^' + ' '*w_right
+def point_to_error(s: str, ind: int, w: int = 5) -> str:
```
stevenxpan, line 374: I'm not super happy with the readability of this function -- what does `w` mean, and where do all the shifts come from?

python/python_helper_sympy.py
```diff
@@ -221,71 +325,206 @@ def evaluate(expr, locals_for_eval={}):
...
-        for v in variables:
-            if k == v:
-                raise ValueError('sympy expression has a variable with a reserved name: {:s}'.format(k))
+        reserved |= (
+            const.complex_variables.keys() | const.hidden_complex_variables.keys()
+        )
+    if allow_trig_functions:
+        reserved |= const.trig_functions.keys()
+
+    for v in variables:
```
stevenxpan, line 403: Similar to above, I think you can replace this with an `any()` call.

python/python_helper_sympy.py
```diff
@@ -221,71 +325,206 @@ def evaluate(expr, locals_for_eval={}):
...
+            f"<br><br><pre>{point_to_error(expr, err.offset)}</pre>"
+        )
+    except Exception:
+        return "Invalid format."
+
+    return None
+
+
+def get_variables_list(variables_string: Optional[str]) -> List[str]:
+    if variables_string is not None:
```
stevenxpan, line 515:
```suggestion
    if variables_string is None:
        return []
    ...
```

python/python_helper_sympy.py
```diff
@@ -221,71 +325,206 @@ def evaluate(expr, locals_for_eval={}):
...
+        return [variable.strip() for variable in variables_string.split(",")]
+
+    return []
+
+
+def process_student_input(student_input: str) -> str:
+    # Replace '^' with '**' wherever it appears. In MATLAB, either can be used
+    # for exponentiation. In python, only the latter can be used.
```
stevenxpan, line 523:
```suggestion
    # for exponentiation. In Python, only the latter can be used.
```

python/python_helper_sympy.py
```diff
@@ -221,71 +325,206 @@ def evaluate(expr, locals_for_eval={}):
...
+    a_sub = student_input.replace("^", "**")
+
+    # Replace unicode minus with hyphen minus wherever it occurs
```
stevenxpan, line 526:
```suggestion
    # Replace Unicode minus with hyphen minus wherever it occurs
```

python/python_helper_sympy.py
eliotwrobson: This is one of those kinda magic functions that was in this file for controlling something in the display. I'll add a comment about what this does, but something I don't really want to alter _too_ much.

python/python_helper_sympy.py
eliotwrobson: I tend to prefer comprehensions in cases like this, since it compartmentalizes the logic that's used for object creation and makes it clear this dict doesn't need to be modified.  Unfortunately we're not on 3.9, where the dict union syntax makes this much cleaner...

elements/pl-big-o-input/big_o_utils.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+def get_enum_attrib(
+    enum_type: Type[EnumT],
+    element: lxml.html.HtmlElement,
+    name: str,
+    default: Optional[EnumT] = None,
+) -> EnumT:
+    """value = get_enum_attrib(enum_type, element, name, default)
+
```
nwalters512, line 225: I think the signature should be immediately clear from the types?
```suggestion
    """
```

elements/pl-big-o-input/big_o_utils.py
```diff
@@ -0,0 +1,252 @@
...
+    elif score > 0:
+        return ("partial", math.floor(score * 100))
+
+    return ("incorrect", True)
+
+
+EnumT = TypeVar("EnumT", bound=Enum)
+
+
+def get_enum_attrib(
```
nwalters512, line 218: Consider adding a test for this function, although I don't think that's super-critical until we lift it into `prairielearn.py`.

setup.cfg
```diff
@@ -2,3 +2,6 @@
 filename = ./elements/**/*.py,./question-servers/**/*.py
 ignore = E501
 inline-quotes = single
+# Black compatibility.
```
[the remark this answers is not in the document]
nwalters512: We expect this diff to disappear once https://github.com/PrairieLearn/PrairieLearn/pull/6793 is merged, correct?

python/conftest.py
```diff
@@ -0,0 +1,4 @@
+"""
+Test file containing shared test fixtures. Needed for now for test runner to pick up this folder.
+Will add fixtures as needed
```
nwalters512, line 3: 👀 are there supposed to be shared test fixtures in here? Is this file really necessary for pytest to find this directory? We already have files like `prairielearn_test.py` in this directory, and I believe they've been working fine.

eliotwrobson: I tried removing it and had some issues, but I might have been doing something wrong. Will try again.

elements/pl-big-o-input/big_o_utils.py
eliotwrobson: I think I'd prefer to wait until this gets lifted.

docs/elements.md
```diff
@@ -621,6 +623,48 @@ Do not include `i` or `j` in the list of `variables` if `allow-complex="true"`.
 
 ---
 
+### `pl-big-o-input` element
+
+Fill in the blank field that allows for asymptotic mathematical input (i.e. big O, big Theta, etc.).
```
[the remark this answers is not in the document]
eliotwrobson: I actually think that big-o is the best name. The majority of the time, instructors (and students) will be using the big-o (default) input mode, so some ambiguity is removed by having this name. In addition, most people (including myself) refer to this asymptotic notation as "big-O family notation" (this is even the case on the Wikipedia article https://en.wikipedia.org/wiki/Big_O_notation). Having this name just reduces ambiguity for people in most cases.

Also, for purely selfish reasons, I don't really want to update the docs and all the filenames 😅

elements/pl-big-o-input/big_o_utils.py
eliotwrobson: This is just to match the docs for the other `get_*_attrib` functions.

eliotwrobson: Tangentially related, @nwalters512 do you think the order of the arguments be changed for this function?  I think it feels a little different than the other getters since the element isn't the first argument.

python/conftest.py
eliotwrobson: So double checked, removing this file causes issues with the test runner being able to pick up imports from `prairielearn.py` when testing `big_o_utils.py`. Although kinda weird, I don't think this is an issue because (1) all the problematic imports will be gone once the desired functions get promoted to `prairielearn.py` and (2) this file is actually used to hold test fixtures in some of my other PRs (see https://github.com/PrairieLearn/PrairieLearn/pull/6764 for example).

As for the `conftest.py` semantics themselves, this is standard practice in pytest: https://docs.pytest.org/en/6.2.x/fixture.html

docs/elements.md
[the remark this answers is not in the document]
jeffgerickson: I agree that `big-o` is the right name.  Yes, even for questions about Ω or Θ notation.  

(I have a slight preference for `big-O`, but that's not worth arguing about.)

I grew up in a part of the country that calls _all_ soft drinks "Coke".  ("Would you like a Coke with that, darlin'?"  "What kind do you have?"  "We have Co-cola, Dr. Pepper, root bear, and 7-Up.")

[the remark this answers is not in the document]
nwalters512: I'm good with this! Just wanted to make sure we're sure about the name, as changing an element name once it exists is difficult to impossible.

elements/pl-big-o-input/big_o_utils.py
nwalters512: > This is just to match the docs for the other `get_*_attrib` functions.

Ah, gotcha, the other functions didn't show up in the diff. Especially with type info, I think we should remove that repeat of the function signatures from _all_ these functions, though I won't ask you to do that here.

I might consider putting `enum_type` after `name`?

eliotwrobson: I think that's a reasonable course of action. There's a PR we can get in a bit later that adds type info for these functions (best done after typechecking is set for a few elements first), and we can change the docs there.

And sounds good, will change as a part of rebasing this.

python/python_helper_sympy.py
```diff
@@ -264,98 +323,105 @@ def evaluate(expr, locals_for_eval={}):
...
-        }
-    for k in reserved.keys():
-        for v in variables:
-            if k == v:
-                raise ValueError(
-                    "sympy expression has a variable with a reserved name: {:s}".format(
-                        k
-                    )
-                )
+        reserved |= (
```
nwalters512, line 398: 👀 fancy new syntax!

python/python_helper_sympy.py
```diff
@@ -264,98 +323,105 @@ def evaluate(expr, locals_for_eval={}):
...
+        )
+    if allow_trig_functions:
+        reserved |= const.trig_functions.keys()
+
+    # Check if reserved variables conflict, raise an error if they do
+    conflicting_reserved_variables = reserved & set(variables)
+
+    if conflicting_reserved_variables:
+        raise ValueError(
+            f"sympy expression has a variable with a reserved name: {conflicting_reserved_variables}"
```
nwalters512, line 409: Nit: reflecting the fact that there may be multiple conflicts.

```suggestion
            f"sympy expression has variables with reserved names: {conflicting_reserved_variables}"
```

python/python_helper_sympy.py
```diff
@@ -264,98 +323,105 @@ def evaluate(expr, locals_for_eval={}):
...
+            [(val, key) for key, val in const.hidden_complex_variables.items()]
         )
 
-    return {"_type": "sympy", "_value": str(a), "_variables": variables}
+    return {"_type": "sympy", "_value": str(a_sub), "_variables": variables}
 
 
-def json_to_sympy(a, allow_complex=True):
+def json_to_sympy(
+    a: SympyJson, *, allow_complex: bool = True, allow_trig_functions: bool = True
```
nwalters512, line 423: For my own knowledge, what's the `*` here?

nwalters512 (approved):

python/python_helper_sympy.py
tdy: It enforces that subsequent keywords must be named, i.e. original allows `json_to_numpy(a, False)` while latter requires `json_to_numpy(a, allow_complex=False)`

nwalters512: neat, thanks for explaining!

python/python_helper_sympy.py
eliotwrobson: I think this syntax for sets has actually been around for a while (same as the `.update` method), but for dicts it was introduced in 3.9 (see https://peps.python.org/pep-0584/)

python/python_helper_sympy.py
eliotwrobson: Yep! The `*` is usually used for additional options that get defaulted which don't interact with core data.

PR #6643, opened 2022-11-18
Move Azure auth to Enterprise Edition

nwalters512:
This PR makes Azure auth an Enterprise Edition feature.

As part of this PR, I simplified the config options available for Azure. Most of them are unused in production, and some of them actually *can't* be configurable for auth to work correctly.

## TODO

- [x] Test in `uiuc-staging` environment to ensure that auth still works correctly.


nwalters512:
I set up Azure in `us-staging`, confirmed it worked on master, and then confirmed that it still worked with this PR. We should be good to go!

lib/config.js
```diff
@@ -318,88 +320,41 @@ config.serverJobsAbandonedTimeoutSec = 30;
...
-  // If app key contains '\', replace it with '\\'.
-  azureClientSecret: '<your_client_secret>',
-
-  // Required to set to false if you don't want to validate issuer
-  azureValidateIssuer: false,
-
-  // Required if you want to provide the issuer(s) you want to validate instead of using the issuer from metadata
-  azureIssuer: null,
+/** The client ID of your app in AAD; required. */
+config.azureClientID = '<your_client_id>';
```
mwest1066, line 324: We'll have to remember to reconfigure our existing prod setups (for all envs).

mwest1066 (approved):

lib/config.js
nwalters512: What exactly do you think we'll have to change configuration for? I only removed unused options; I didn't change any of the ones we're using in existing configs (unless I changed something else on accident).

mwest1066: Oh sorry! I misread the config changes and thought there was a change in the level of nesting for the Azure options. But this all looks fine!

PR #6646, opened 2022-11-20
Improve tests for SymPy helpers and pl-symbolic-input

eliotwrobson:
Fix test for sympy element. Adds unit tests to internal sympy helper functions to ensure functionality is preserved before a refactor in a later pull request.

### DIFF
--- a/elements/pl-symbolic-input/pl-symbolic-input.py
+++ b/elements/pl-symbolic-input/pl-symbolic-input.py
@@ -370,14 +370,29 @@ def grade(element_html, data):
 def test(element_html, data):
     element = lxml.html.fragment_fromstring(element_html)
     name = pl.get_string_attrib(element, 'answers-name')
+    variables = get_variables_list(pl.get_string_attrib(element, 'variables', VARIABLES_DEFAULT))
+    allow_complex = pl.get_boolean_attrib(element, 'allow-complex', ALLOW_COMPLEX_DEFAULT)
     weight = pl.get_integer_attrib(element, 'weight', WEIGHT_DEFAULT)
+    imaginary_unit = pl.get_string_attrib(element, 'imaginary-unit-for-display', IMAGINARY_UNIT_FOR_DISPLAY_DEFAULT)
+
+    # Get raw correct answer
+    a_tru = data['correct_answers'][name]
+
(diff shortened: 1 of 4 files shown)


elements/pl-symbolic-input/pl-symbolic-input.py
```diff
@@ -367,63 +369,100 @@ def grade(element_html, data):
...
+
         if invalid_type == 'float':
-            data['raw_submitted_answers'][name] = 'x + 1.234'
-            s = 'Your answer contains the floating-point number ' + str(1.234) + '. '
-            s += 'All numbers must be expressed as integers (or ratios of integers). '
-            s += '<br><br><pre>' + phs.point_to_error('x + 1.234', 4) + '</pre>'
-            data['format_errors'][name] = s
+            invalid_input = 'x + 1.234'
+            data['raw_submitted_answers'][name] = invalid_input
+            data['format_errors'][name] = (
```
eliotwrobson, line 410: @nwalters512 When I was going over this test function, I noticed that changing this format error message didn't cause any test failures. Are the format errors produced not checked against the reference ones here? If so, I can remove these from the tests (since they're not necessary).

eliotwrobson:
Going to change this PR to open. Not sure if there are enough tests on the backend to ensure expected default behavior, so open to suggestions on other tests there. The goal is to have enough tests that we feel comfortable refactoring to add custom functions as part of the Big-O element PR.

elements/pl-symbolic-input/pl-symbolic-input.py
nwalters512: It looks like we only compare the keys, not their values:

https://github.com/PrairieLearn/PrairieLearn/blob/1957d08a566f18c29519562ed716e5c820b8988a/lib/question.js#L901-L905

@mwest1066 do you know why that is?

nwalters512 (approved):

elements/pl-symbolic-input/pl-symbolic-input.py
eliotwrobson: That makes sense if we don't care about the error message (as long as their is one). It might be nice in this case to check the values as well, but could cause issues testing other elements.

PR #6649, opened 2022-11-21
Periodically clean old rows from time_series table

nwalters512:
Closes #6645.

PR #6651, opened 2022-11-22
Create index for the date column of time_series

nwalters512:
This will need to be merged and deployed before #6649.


migrations/20221122001508_time_series__date__add_index.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX CONCURRENTLY time_series_date_idx ON time_series (date);
```
nwalters512, line 1: This table has many rows in production (>600 million); `CREATE INDEX CONCURRENTLY` will avoid a lengthy table lock as this index is being created.

mwest1066 (approved):

PR #6652, opened 2022-11-22
Report desired chunk instance count to CloudWatch

nwalters512:
This PR adds a new EE cron job that runs once every 10 seconds to compute a desired instance count for chunk servers. The following CloudWatch metrics are used:

- Page views per second
- Number of active Python workers per second
- Load balancer requests per minute

Each metric has a desired per-server count - that is, we aim to have X page views per second per server. Based on that, we can compute the number of desired servers per metric. We take the highest desired server count across all metrics.

To ensure that we can scale up quickly but avoid scaling in too rapidly, we also take the max over the past 15 minutes.

Note that I'm not using the `time_series` table to store historical data points. We _already_ have to hit CloudWatch to get these metrics; we don't pay more for fetching more data points. Given that, I don't see a reason to fetch current data from CloudWatch but use the DB to determine historical data.

This is a part of https://github.com/PrairieLearnInc/sysconf/issues/159.


ee/cron/chunksHostLoads.js
```diff
@@ -0,0 +1,189 @@
...
+          Unit: 'Count',
+          Timestamp: new Date(now),
+          Value: desiredInstances,
+        },
+      ],
+      Namespace: 'Chunks',
+    })
+    .promise();
+
+  await setAutoScalingGroupCapacity(config.chunksAutoScalingGroupName, desiredInstances);
```
nwalters512, line 170: Note that I've added code that will set the desired capacity for the ASG, but it won't be enabled until we actually set `chunksAutoScalingGroupName` in the config. The intent is that we'll let this code run in production for a while and see if it makes accurate predictions about how many chunk servers we'll need under varying amounts of load.

ee/cron/chunksHostLoads.js
```diff
@@ -0,0 +1,189 @@
...
+    return;
+  }
+
+  const now = Date.now();
+  const cloudwatch = new AWS.CloudWatch();
+
+  const metrics = await cloudwatch
+    .getMetricData({
+      // Use the last 15 minutes of data.
+      // TODO: make this configurable?
```
nwalters512, line 36: 👀

ee/cron/chunksHostLoads.js
```diff
@@ -0,0 +1,189 @@
...
+// a lowercase letter and contain only numbers, letters, and underscores.
+const PAGE_VIEWS_PER_SECOND = 'pageViewsPerSecond';
+const ACTIVE_WORKERS_PER_SECOND = 'activeWorkersPerSecond';
+const LOAD_BALANCER_REQUESTS_PER_MINUTE = 'loadBalancerRequestsPerMinute';
+
+/**
+ * Finds the maximum value in the given array.
+ * @param {number[]} arr
+ */
+function arrayMax(arr) {
```
mwest1066, line 17: I generally use [`_.max()`](https://lodash.com/docs/4.17.15#max).

ee/cron/chunksHostLoads.js
mwest1066: Yes please! The equivalent of https://github.com/PrairieLearn/PrairieLearn/blob/5526b70f2792354bdd4c52cc2cd3ce2bac0e035f/lib/config.js#L108

cron/index.js
```diff
@@ -113,6 +113,12 @@ module.exports = {
         module: require('../ee/cron/workspaceHostLoads'),
         intervalSec: config.cronOverrideAllIntervalsSec || config.cronIntervalWorkspaceHostLoadsSec,
       });
+
+      module.exports.jobs.push({
+        name: 'chunksHostLoads',
+        module: require('../ee/cron/chunksHostLoads'),
+        intervalSec: config.cronOverrideAllIntervalsSec || config.cronIntervalChunksHostLoadsSec,
+      });
```
mwest1066, line 121: Should we wrap this in `isEnterprise()`?

ee/cron/chunksHostLoads.js
```diff
@@ -0,0 +1,189 @@
...
+  );
+  const loadBalancerRequestsPerMinuteMetric = metrics.MetricDataResults.find(
+    (m) => m.Id === LOAD_BALANCER_REQUESTS_PER_MINUTE
+  );
+
+  const maxPageViewsPerSecond = Math.ceil(arrayMax(pageViewsPerSecondMetric.Values));
+  const maxActiveWorkersPerSecond = Math.ceil(arrayMax(activeWorkersPerSecondMetric.Values));
+  const maxLoadBalancerRequestsPerMinute = Math.ceil(
+    arrayMax(loadBalancerRequestsPerMinuteMetric.Values)
+  );
```
mwest1066, line 113: Why do we want `ceil()` here? I don't think we care if these are not integers.

Note that if we use `_.max()` instead of `arrayMax()` then we should add `?? 0` to the end to handle the zero-length array case.

ee/cron/chunksHostLoads.js
```diff
@@ -0,0 +1,189 @@
...
+  const desiredInstancesByActiveWorkers =
+    maxActiveWorkersPerSecond / config.chunksActiveWorkersCapacityFactor;
+  const desiredInstancesByLoadBalancerRequests =
+    maxLoadBalancerRequestsPerMinute / config.chunksLoadBalancerRequestsCapacityFactor;
+  const desiredInstances = Math.max(
+    desiredInstancesByPageViews,
+    desiredInstancesByActiveWorkers,
+    desiredInstancesByLoadBalancerRequests,
+    1
+  );
```
mwest1066, line 125: Maybe we want a `ceil()` wrapped around this?

mwest1066 (approved):

ee/cron/chunksHostLoads.js
mwest1066: I can't wait to see these metrics in prod!

cron/index.js
nwalters512: Look a few lines up 😉

ee/cron/chunksHostLoads.js
nwalters512: Fair point - I can move the `Math.ceil` call around the division expressions that get assigned to the `desiredInstancesBy...` variables?

ee/cron/chunksHostLoads.js
nwalters512: I'm going to move the `ceil` to the individual `desiredInstancesBy...` values, so that's most likely unnecessary.

ee/cron/chunksHostLoads.js
nwalters512: Added in [96ac1ce](https://github.com/PrairieLearn/PrairieLearn/pull/6652/commits/96ac1ce0c7073339f5260861308ed04fe2ed6771)! The config option is named `chunksHostAutoScalingHistoryIntervalSec`, seem reasonable? (Note that I also renamed the cron job from `chunksHostLoads` to `chunksHostAutoScaling`.)

ee/cron/chunksHostLoads.js
nwalters512: Changed in 2f3c116.

ee/cron/chunksHostLoads.js
nwalters512: Changed in 2f3c116, does that look better?

ee/cron/chunksHostLoads.js
mwest1066: Sounds good to me

cron/index.js
mwest1066: Ah... Serves me right for just looking at the diff in GitHub 😂

ee/cron/chunksHostLoads.js
mwest1066: Yep! This looks good. I would slightly lean towards computing the individual metrics without the ceil (and applying one ceil at the very end) so that we can see when the individual metrics might be close to an integer boundary. E.g., if we aren't scaling above 1 instance, I'd like to know whether the individual metrics are at 0.8 or 0.008 to see whether we need a small or large change to the capacity factors. However, in practice this probably won't be a huge deal so I'm also fine to leave the code as is.

mwest1066 (approved):

ee/cron/chunksHostLoads.js
mwest1066: I made this change in 250cd68a88d2db5a8b9e3f419b9165b69212f3cc

PR #6653, opened 2022-11-22
Add type checking to cronjobs index

nwalters512:
This caught a few issues with our OTel instrumentation! 🎉


mwest1066 (approved):
Nice! Also good to keep the date usage clean.

PR #6656, opened 2022-11-23
Fix slow instructor assessments page

### LINKED ISSUE #6488: Assessment Overview Page has Multi-Second Load Times in Large Classes
We use a PL assessment for every class period in a large class (n=600 students).  Loading the assessment overview page (`/course_instance/129485/instructor/instance_admin/assessments`) currently takes multiple seconds to load every time the page is visited.  When bouncing around multiple assignments, this page gets visited a lot.

The page has always been slow in my large classes (particularly near the end of the semester), but this semester is slower than normal and I've never had it be this slow so early.  It appears it may be related to #4915, but thought this report might be useful. :)

### LINKED ISSUE #6647: Plan to speed up instructor assessments page
At the moment the instructor assessments page can take many seconds to load in large classes (see #6488). This is because we are calling `assessments_stats()` and `assessments_duration_stats()` for each assessment to calculate statistics in real time: https://github.com/PrairieLearn/PrairieLearn/blob/9f7c2b7a3b3f7d57c83dc4d5b781e06214447110/pages/instructorAssessments/instructorAssessments.sql#L58-L59 Each of these calls can take a few hundred milliseconds for an assessment so with hundreds of assessments the page render time can be tens of seconds.

My plan for fixing this is twofold:
1. Cache the results of `assessments_stats()` and `assessments_duration_stats()` in the database (probably a separate table), together with timestamps for when the cache was last updated. Also add an index on `(assessment_id, modified_at)` for the `assessment_instances` table. On the instructor assessments page, for each assessment check whether there are any assessment instances where `modified_at` is greater than the cache update time and recalculate those statistics (and update the cache).
2. On the instructor assessments page, if there are assessments where the cache is not up to date, simply mark those assessments as requiring a recalculation but don't actually do it. Send all the data to the client, and then have the client lazy-load the missing statistics via ajax calls, which will recalculate/re-cache as needed on the server.

The reason for step (2) above is that, if students are actively working on a few assessments, recalculating the statistics could still take a second or two, which is undesirably slow for a page load.

mwest1066:
Speed up the instructor assessments page by caching the assessment statistics. See #6647 for an explanation of the strategy used.

Resolves #6647
Fixes #6488

* [x] Add safety factor for time comparisons.
* [x] Don't show old stats when refreshing, to avoid table redraw jumping.
* [x] Unconditionally check whether a recompute is needed in `assessment.updateAssessmentStatistics()` (remove the argument).
* [x] Use `eachLimit` inside `assessment.updateAssessmentStatisticsForCourseInstance()`.
* [x] Call `assessment.updateAssessmentStatistics()` before we render `instructorQuestionStats`. Also use the cached data in the `assessments` row rather than calling `assessments_stats()` and `assessments_duration_stats()`.
* [x] Check we don't have any other direct calls to `assessments_stats()` or `assessments_duration_stats()`.
* [x] Consider renaming `assessments_stats()` to `assessments_score_stats()`.
* [x] Show error symbol instead of loading animation on error.
* [x] `BLOCK update_assessment_statisics` can have `duration_stat_hist` being NULL -- see [this course](https://staging.prairielearn.org/pl/course_instance/129074/instructor/instance_admin/assessments/)


lib/paginate.js
```diff
@@ -55,3 +55,31 @@ module.exports = {
...
+ * param that will be used to paginate the query.
+ *
+ * Should be used like this:
+ *
+ * for await (const row of paginateQueryAsync(sql, params)) {
+ *  // do something with row
+ * }
+ *
+ */
+module.exports.paginateQueryAsync = async function* (sql, params) {
```
mwest1066, line 71: I'm experimenting with async generators for the first time, so extra eyes on this would be appreciated!

pages/instructorAssessments/instructorAssessments.js
```diff
@@ -35,40 +37,51 @@ const fileSubmissionsName = (locals) => {
...
-    course_instance_id: res.locals.course_instance.id,
-    authz_data: res.locals.authz_data,
-    req_date: res.locals.req_date,
-    assessments_group_by: res.locals.course_instance.assessments_group_by,
-  };
-  sqldb.query(sql.select_assessments, params, function (err, result) {
-    if (ERR(err, next)) return;
+router.get(
+  '/',
+  asyncHandler(async (req, res) => {
```
mwest1066, line 42: This is best viewed with "Hide whitespace" turned on.

migrations/20221122004832_assessments__statistics__add.sql
```diff
@@ -0,0 +1,22 @@
+ALTER TABLE assessments ADD COLUMN OUT statistics_last_updated_at timestamptz NOT NULL DEFAULT now() - interval '100 years';
```
mwest1066, line 1: I elected to cache the statistics directly in the `assessments` table, to avoid the mess when we use a second caching table but it doesn't yet have a cached row for our assessment. To hopefully keep the `assessments` table organized, I used a common prefix for all the cached rows.

lib/assessment.js
```diff
@@ -523,4 +523,44 @@ module.exports = {
...
+   *
+   * @param {number} course_instance_id - The course instance ID.
+   */
+  async updateAssessmentStatisticsForCourseInstance(course_instance_id) {
+    const result = await sqldb.queryAsync(sql.select_assessments_for_statistics_update, {
+      course_instance_id,
+    });
+    for (const row of result.rows) {
+      await module.exports.updateAssessmentStatistics(row.assessment_id);
+    }
```
jonatanschroeder, line 539: Might be better to use [`async.eachLimit`](https://caolan.github.io/async/v3/docs.html#eachLimit) since these operations are independent and could potentially run in parallel.

pages/instructorAssessments/instructorAssessments.js
```diff
@@ -35,40 +37,51 @@ const fileSubmissionsName = (locals) => {
...
+      // update assessment statistics if needed
+      assessment.updateAssessmentStatistics(res.locals.course_instance.id);
+
+      var params = {
+        course_instance_id: res.locals.course_instance.id,
+        authz_data: res.locals.authz_data,
+        req_date: res.locals.req_date,
+        assessments_group_by: res.locals.course_instance.assessments_group_by,
+      };
+      const result = sqldb.query(sql.select_assessments, params);
```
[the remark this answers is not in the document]
mwest1066: Thanks! Fixed in d523a9c5ee0d3aff4c57caa1486f797b7f7ff77a

pages/instructorAssessments/instructorAssessments.js
```diff
@@ -35,40 +37,51 @@ const fileSubmissionsName = (locals) => {
...
+      // update assessment statistics if needed
+      assessment.updateAssessmentStatistics(res.locals.course_instance.id);
+
+      var params = {
+        course_instance_id: res.locals.course_instance.id,
+        authz_data: res.locals.authz_data,
+        req_date: res.locals.req_date,
+        assessments_group_by: res.locals.course_instance.assessments_group_by,
+      };
+      const result = sqldb.queryAsync(sql.select_assessments, params);
```
jonatanschroeder, line 83: Shouldn't it be await? 🙂

mwest1066: Hahaha, this has to the most inefficient method of debugging code ever 😄

Fixed in 3dc5fa7396b0585d69f6809972b8e800ea44a4f8 (but still not actually tested 😅)

pages/instructorAssessments/instructorAssessments.ejs
```diff
@@ -15,7 +16,41 @@
...
+              var newContent = document.importNode(template.content, true);
+
+              // update the page content
+              var row = document.getElementById('row-' + assessment_id);
+              row.querySelector('.score-stat-number').innerHTML = newContent.querySelector('.score-stat-number').innerHTML;
+              row.querySelector('.score-stat-score-hist').innerHTML = newContent.querySelector('.score-stat-score-hist').innerHTML;
+              row.querySelector('.score-stat-mean').innerHTML = newContent.querySelector('.score-stat-mean').innerHTML;
+              row.querySelector('.duration-stat-mean').innerHTML = newContent.querySelector('.duration-stat-mean').innerHTML;
+
+              // re-evaluate scripts in the row to render statistics
```
mwest1066, line 42: This is kind of horrible. Rather than having scripts in each row I could have encoded the row-specific data into the HTML and then pulled the rendering code out into a top-level function. However, we don't yet have good support for client-side JS in PL (need to bring this over from PT) so it felt like it was getting out of control. This script-rehydrating code here is ugly, but at least it's fairly short and contained.

pages/instructorAssessments/instructorAssessments.ejs
```diff
@@ -15,7 +16,41 @@
             trigger: 'hover',
         }).on("show.bs.popover", function() {
             $($(this).data("bs.popover").getTipElement()).css("max-width", "80%");
-        });;
+        });
+
+        var assessment_ids_needing_stats_update = <%- JSON.stringify(assessment_ids_needing_stats_update) %>;
+        async.eachLimit(assessment_ids_needing_stats_update, 3, function(assessment_id, callback) {
+          fetch('<%= urlPrefix %>/instance_admin/assessments/stats/' + assessment_id)
```
mwest1066, line 23: I experimented with using [htmx](https://htmx.org) to do this lazy-loading (see b22954fc8bd877bf2b5502f405be2836c7955f4d). It was super short and clean, but there was no good way to both fetch the updated rows incrementally and also limit the concurrency of fetches. The code in b22954fc8bd877bf2b5502f405be2836c7955f4d fetched the entire table at once, which allowed us to limit concurrency server-side. This was nice, but in a really big class could still lead to timeouts if the number of assessments became very large. In the end I opted to write the fetching code by hand, which let me limit concurrency client-side and incrementally fetch each row, so no single fetch will take very long.

pages/instructorAssessmentQuestionStatistics/instructorAssessmentQuestionStatistics.js
```diff
@@ -29,20 +30,27 @@ router.get('/', function (req, res, next) {
...
         });
+        res.locals.assessment = result.rows[0].assessment;
       },
       function (callback) {
-        debug('query assessment_stats');
+        // Fetch assessments.stats_last_updated (the time when we last updated
+        // the _question_ statistics for this assessment). Note that this is
+        // different to assessments.statistics_last_updated_at (the time we last
+        // updated the assessment instance statistics stored in the assessments
+        // row itself).
```
mwest1066, line 49: The preexisting `assessments.stats_last_updated` column is very confusingly named, now that we have added `assessments.statistics_last_updated_at`. Ideally we'd rename `stats_last_updated` to `question_stats_last_updated`, but column renames are a bit unpleasant to manage and I want to eliminate `stats_last_updated` by rewriting question statistic caching using a similar scheme to the current PR. However I'll defer that to a future PR and so I'm just leaving `stats_last_updated` as it is for the moment. Is there somewhere else that we should add a comment about this confusing naming situation?

lib/assessment.js
mwest1066: Done in https://github.com/PrairieLearn/PrairieLearn/pull/6656/commits/f8d0e83fc17685fa326617f0f777b1c15410034e

lib/assessment.sql
```diff
@@ -56,3 +56,65 @@ SET
...
+    duration_stat_median = duration_stats.median,
+    duration_stat_thresholds = duration_stats.thresholds,
+    duration_stat_threshold_seconds = duration_stats.threshold_seconds,
+    duration_stat_threshold_labels = duration_stats.threshold_labels,
+    duration_stat_hist = duration_stats.hist
+FROM
+    assessments_score_stats($assessment_id) AS score_stats,
+    assessments_duration_stats($assessment_id) AS duration_stats
+WHERE
+    a.id = $assessment_id;
```
jonatanschroeder, line 120: Given this is the only place that these sprocs are used, wouldn't it be better to move this update to the sprocs themselves? Maybe even the check if the update is needed, to reduce the back-and-forth between Postgres and JS?

pages/instructorAssessmentStatistics/instructorAssessmentStatistics.js
```diff
@@ -31,24 +32,23 @@ router.get('/', function (req, res, next) {
...
-        // FIXME: change to assessment_instance_duration_stats and show all instances
-        var params = { assessment_id: res.locals.assessment.id };
-        sqldb.queryOneRow(sql.assessment_duration_stats, params, function (err, result) {
-          if (ERR(err, callback)) return;
-          res.locals.duration_stat = result.rows[0];
-          callback(null);
+      async () => {
+        // get formatted duration statistics
+        const result = await sqldb.queryOneRowAsync(sql.select_duration_stats, {
+          assessment_id: res.locals.assessment.id,
```
jonatanschroeder, line 49: Any reason why these queries are separate? They could very easily be merged.

pages/instructorAssessmentStatistics/instructorAssessmentStatistics.sql
```diff
@@ -1,20 +1,23 @@
--- BLOCK assessment_stats
```
jonatanschroeder, line 1: This query is still in use in line 84 of instructorAssessmentStatistics.js (the score stats CSV).

lib/paginate.js
```diff
@@ -55,3 +55,31 @@ module.exports = {
...
+  const _params = Object.assign({}, params);
+  let offset = 0;
+  while (true) {
+    _params.offset = offset;
+    const result = await sqldb.queryAsync(sql, _params);
+    for (const row of result.rows) {
+      yield row;
+    }
```
jonatanschroeder, line 79: I don't have experience with `yield` in JS, but I was reading about it and this code may be simplified to (if I'm reading [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield*) correctly):
```suggestion
    yield* result.rows;
```

pages/instructorAssessmentQuestionStatistics/instructorAssessmentQuestionStatistics.js
```diff
@@ -29,20 +30,27 @@ router.get('/', function (req, res, next) {
...
   async.series(
     [
-      function (callback) {
-        var params = { assessment_id: res.locals.assessment.id };
-        sqldb.queryOneRow(sql.assessment_stats_last_updated, params, function (err, result) {
-          if (ERR(err, callback)) return;
-          res.locals.stats_last_updated = result.rows[0].stats_last_updated;
-          callback(null);
+      async () => {
+        // make sure statistics are up to date
```
nwalters512, line 34: Is it actually necessary to re-compute the statistics for this page? AFAICT it only shows *question* statistics, not stats for the assessment as a whole.

pages/instructorAssessmentQuestionStatistics/instructorAssessmentQuestionStatistics.ejs
```diff
@@ -60,7 +60,7 @@
           </div>
         </div>
 
-        <% if (assessment_stat.number > 0) { %>
+        <% if (assessment.score_stat_number > 0) { %>
```
nwalters512, line 63: It's not clear to me why this is part of the conditional - isn't this unrelated to question stats?

migrations/20221122004832_assessments__statistics__add.sql
```diff
@@ -0,0 +1,22 @@
...
+ALTER TABLE assessments ADD COLUMN score_stat_min double precision NOT NULL DEFAULT 0;
+ALTER TABLE assessments ADD COLUMN score_stat_max double precision NOT NULL DEFAULT 0;
+ALTER TABLE assessments ADD COLUMN score_stat_mean double precision NOT NULL DEFAULT 0;
+ALTER TABLE assessments ADD COLUMN score_stat_std double precision NOT NULL DEFAULT 0;
+ALTER TABLE assessments ADD COLUMN score_stat_median double precision NOT NULL DEFAULT 0;
+ALTER TABLE assessments ADD COLUMN score_stat_n_zero integer NOT NULL DEFAULT 0;
+ALTER TABLE assessments ADD COLUMN score_stat_n_hundred integer NOT NULL DEFAULT 0;
+ALTER TABLE assessments ADD COLUMN score_stat_n_zero_perc double precision NOT NULL DEFAULT 0;
+ALTER TABLE assessments ADD COLUMN score_stat_n_hundred_perc double precision NOT NULL DEFAULT 0;
+ALTER TABLE assessments ADD COLUMN score_stat_score_hist integer[] NOT NULL DEFAULT ARRAY[]::integer[];
```
nwalters512, line 13: Nit: `score_stat_hist` might better match other `score_*` columns here (e.g. they aren't `store_stat_score_median`) and the duration below (`duration_stat_hist`, not `duration_stat_duration_hist`).

pages/instructorAssessments/instructorAssessments.ejs
nwalters512: Makes sense. Note that if we're using `fetch`, we can safely use `async`/`await`/`const`/etc.

pages/instructorAssessments/instructorAssessments.ejs
nwalters512: +1 to getting better infra for client-side JS in this repo!

pages/instructorAssessmentQuestionStatistics/instructorAssessmentQuestionStatistics.js
mwest1066: Good point. https://github.com/PrairieLearn/PrairieLearn/issues/6743

lib/paginate.js
mwest1066: Very nice, thanks!

pages/instructorAssessmentStatistics/instructorAssessmentStatistics.sql
mwest1066: Good catch, thanks! Fixed in a83518ec7a0ab8627258a3e39e5e288856223925

pages/instructorAssessmentStatistics/instructorAssessmentStatistics.js
mwest1066: They are needed separately in the `GET(:filename)` handler (see a83518ec7a0ab8627258a3e39e5e288856223925).

pages/instructorAssessmentQuestionStatistics/instructorAssessmentQuestionStatistics.js
```diff
@@ -29,20 +30,27 @@ router.get('/', function (req, res, next) {
...
+        await assessment.updateAssessmentStatistics(res.locals.assessment.id);
+      },
+      async () => {
```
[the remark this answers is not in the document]
mwest1066: Fixed in b0335484004925996f4ee15afc338ca1c3f1577d

pages/instructorAssessmentQuestionStatistics/instructorAssessmentQuestionStatistics.js
mwest1066: That's true, but it uses the student count from the overall assessment statistics to determine whether to show the question statistics at all. I'm making the minimal possible changes to the question statistics pages in this PR (deferring to #6743) so I'm just preserving the use of `assessment.score_stat_number` as a flag.

pages/instructorAssessmentQuestionStatistics/instructorAssessmentQuestionStatistics.ejs
mwest1066: Sort of. As mentioned above, this page is using the _assessment_ statistics value `score_stat_number` (the number of students who have taken the assessment) as a flag to decide whether to show any _question_ statistics. I'm definitely not defending this choice but I'm also trying to make the minimal necessary changes to the question statistics pages in this PR.

migrations/20221122004832_assessments__statistics__add.sql
mwest1066: Good call. `score_stat_score_hist` renamed in 1a72ee079f526c60e1950d168b4f8d516a074622. I believe `duration_stat_hist` is already called that (there is no `duration_stat_duration_hist `).

lib/assessment.sql
mwest1066: Over the long term I'm trying to slowly move away from sprocs and towards regular queries, with typescript to check the returned data types. In this case I would ideally remove the two `_stats()` sprocs in favor of queries in `assessment.sql`. I didn't do this because I wanted to minimize the risk of this PR breaking the statistics calculation, so I left the sprocs untouched.

migrations/20221122004832_assessments__statistics__add.sql
nwalters512: Understood about the duration stats - was just pointing to that as an analogous column that was named how I wanted the score one to be!

mwest1066: Ah, got it! Apologies for the confusion.

pages/instructorAssessmentQuestionStatistics/instructorAssessmentQuestionStatistics.js
```diff
@@ -29,20 +30,26 @@ router.get('/', function (req, res, next) {
...
       function (callback) {
-        debug('query assessment_stats');
+        // Fetch assessments.stats_last_updated (the time when we last updated
+        // the _question_ statistics for this assessment). Note that this is
+        // different to assessments.statistics_last_updated_at (the time we last
+        // updated the assessment instance statistics stored in the assessments
+        // row itself).
         var params = { assessment_id: res.locals.assessment.id };
-        sqldb.queryOneRow(sql.assessment_stats, params, function (err, result) {
+        sqldb.queryOneRow(sql.assessment_stats_last_updated, params, function (err, result) {
```
nwalters512, line 50: Nit: while you're here, you could convert this to an async call.

pages/instructorAssessmentStatistics/instructorAssessmentStatistics.js
```diff
@@ -1,16 +1,14 @@
-const ERR = require('async-stacktrace');
 const _ = require('lodash');
-const async = require('async');
 const express = require('express');
+const asyncHandler = require('express-async-handler');
 const router = express.Router();
-const csvStringify = require('../../lib/nonblocking-csv-stringify');
-const path = require('path');
-const debug = require('debug')('prairielearn:' + path.basename(__filename, '.js'));
+const csvStringifyAsync = require('../../lib/nonblocking-csv-stringify').nonblockingStringifyAsync;
```
nwalters512, line 5: Very minor nit:

```suggestion
const { nonblockingStringifyAsync } = require('../../lib/nonblocking-csv-stringify');
```

pages/instructorAssessmentStatistics/instructorAssessmentStatistics.js
```diff
@@ -26,65 +24,54 @@ const setFilenames = function (locals) {
...
-        var params = { assessment_id: res.locals.assessment.id };
-        sqldb.queryOneRow(sql.assessment_stats, params, function (err, result) {
-          if (ERR(err, callback)) return;
-          res.locals.assessment_stat = result.rows[0];
-          callback(null);
-        });
-      },
-      function (callback) {
-        debug('query assessment_duration_stats');
-        // FIXME: change to assessment_instance_duration_stats and show all instances
```
nwalters512, line 45: Is this `FIXME` indeed no longer relevant?

nwalters512 (approved):

pages/instructorAssessmentQuestionStatistics/instructorAssessmentQuestionStatistics.js
mwest1066: I converted the entire file to async/await in 49ea0f5003de058775a03834c00a05546602240a

pages/instructorAssessmentStatistics/instructorAssessmentStatistics.js
mwest1066: I think that it's not really clear whether we actually want to see all instances rather than just the highest scoring instance per student (which is what we do currently). I added a comment (not a FIXME) about this in afebb922cd7eb2df8faf424e5cb1540032d806b0

Issue #4965, opened 2021-10-12
Use pep8-naming plugin for flake8


nwalters512:
We should add the [`pep8-naming` plugin](https://pypi.org/project/pep8-naming/) to our setup so that we can automatically enforce consistent naming conventions. Sister issue to https://github.com/PrairieLearn/PrairieLearn/issues/4964.

eliotwrobson:
Worth including in #6447 if the Python workflow gets reworked.

eliotwrobson:
Was playing around with this plugin, and most of the violations come from `pl-drawing/elements.py`. It looks like the changes required to bring that file into compliance require a substantial reworking of some of the logic, so it may be worth skipping that file when setting this up. Alternatively, it may be worth a refactor (since there are other issues that go beyond style there).

nwalters512:
FWIW, we now use Ruff, which also supports the naming rules: https://docs.astral.sh/ruff/rules/#pep8-naming-n

eliotwrobson:
> FWIW, we now use Ruff, which also supports the naming rules: https://docs.astral.sh/ruff/rules/#pep8-naming-n

I think there is a way to have per-file ignores for individual rules, so this should be pretty straightforward to set up. I'll go ahead and self-assign.

PR #6665, opened 2022-11-29
build: Add black and isort to Python build

### LINKED ISSUE #5027: Use Black to format Python code
Just like we now use Prettier for JavaScript, we should use [Black](https://github.com/psf/black) to automatically format our Python code.

eliotwrobson:
Resolve #5027, by adding black and isort autoformatting and enforcement to the makefile and linter. Makes minor adjustments to CI jobs (Python parts run before Javascript, since they're just backend for this).

Could not include plugin for #4965, since this required some substantial name changes to resolve. Also, does not extend linting to work on files in the "python" directory, since these also required some substantial changes (and some of these files are getting refactored anyway as part of #6641).


eliotwrobson:
So I don't forget: Part of this PR should be to move all python config files into a single `pyproject.toml` file. That way all python related configs can be done in one place (and we condense the `setup.cfg` and `pyrightconfig.json` files).

eliotwrobson:
Going to remake, running into a very weird bug.

PR #6740, opened 2022-12-01
Remove unused PL Peek functionality

mwest1066:
We used to support a "PL Peek" QR code that would link CBTF proctors to exam data in the scheduler. This isn't supported in PT so this link doesn't resolve. It's apparently been causing some confusion for students and proctors so it seems best to remove it.

### DIFF
--- a/.changeset/smooth-knives-remember.md
+++ b/.changeset/smooth-knives-remember.md
@@ -1,6 +1,0 @@
----
-'@prairielearn/opentelemetry': patch
-'@prairielearn/sentry': patch
----
-
-Upgrade all dependencies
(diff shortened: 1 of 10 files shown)


trombonekenny:
Can probably remove even more, like the library we used to make the QR code:

https://github.com/PrairieLearn/PrairieLearn/blob/4c33d18ddeee88ee0703ccb74499de68fe21ffb6/lib/question.js#L8

mwest1066:
> Can probably remove even more, like the library we used to make the QR code:

👍 done in 98aff0ce061113bf583dea879e8d0ef943b2cf12

mwest1066:
> > Can probably remove even more, like the library we used to make the QR code:
> 
> 👍 done in [98aff0c](https://github.com/PrairieLearn/PrairieLearn/commit/98aff0ce061113bf583dea879e8d0ef943b2cf12)

Actually we use the QR library to make links for student pages so we need to keep it.

nwalters512 (approved):

trombonekenny (approved):

PR #6741, opened 2022-12-02
feat: Add score manipulation functions

### LINKED ISSUE #4381: helper functions for manipulating `score`
Letting your average idiot (raises hand) shoot themselves in the foot with `def grade(data)` is wonderful but some help would be appreciated.

In particular, the recalculation of `score` after one manipulates `partial_scores` seems like pain point.

We did this:
```py
    score = 0
    variables = data['partial_scores'].keys()
    for name in variables:
        score += data['partial_scores'][name]['score']
    data['score'] = score / len(variables)
```

@jonatanschroeder said "something like this":
```py
data['score'] = sum(p.get('score', 0) * p.get('weight', 1) for p in data['partial_scores']) / sum(p.get('weight', 1) for p in data['partial_scores'])
```

But basically whatever it is, I don't want to reinvent it every time.  I humbly request a little helper function like this:
```py
import prairielearn as pl

def grade(data):
    ...
    # carefully aim at foot
    # make randomish changes to data['partial_scores']
    ...
    pl._recompute_the_total_score(data)
```

eliotwrobson:
Resolves #4381 by adding some score manipulation functions. Leaving as draft since I want to wait for the BigO element to get merged before merging this (since that PR modifies the way the test files are set up).


python/prairielearn.py
```diff
@@ -52,6 +52,39 @@ class QuestionData(TypedDict):
...
 
+
+def set_weighted_score_data(data: QuestionData) -> None:
+    "Sets main question score to be weighted average of all partial scores"
+    # TODO change to check that all weights are unset if using an equal weighting
+    weight_total = 0
+    score_total = 0.0
+    for part in data["partial_scores"].values():
+        score = part["score"]
+        weight = part["weight"]
```
jonatanschroeder, line 63: There are questions that skip the weight in the `grade` function, it would probably make sense to default it to 1 for compatibility in these cases.
```suggestion
        weight = part.get("weight", 1)
```

eliotwrobson:
Flipping to open since already got some initial feedback, so should be ready for review. @cbm755 Please take a look (as you submitted the original issue).

python/prairielearn.py
jonatanschroeder: I think it makes more sense to have it default to 1 instead of a parameter. This is also done in other parts of the code, e.g.:
https://github.com/PrairieLearn/PrairieLearn/blob/8ced3f982a733fc3ff2c879cce65f3e222fbb511/sprocs/instance_questions_update_score.sql#L138
https://github.com/PrairieLearn/PrairieLearn/blob/8ced3f982a733fc3ff2c879cce65f3e222fbb511/elements/pl-integer-input/pl-integer-input.py#L10

eliotwrobson: I think it's reasonable to be consistent then. Will change the parameter default to 1.

jonatanschroeder:
I'd prefer not to perform a formal review of it as it includes some Python features I'm not familiar with.

python/prairielearn.py
```diff
@@ -52,6 +53,45 @@ class QuestionData(TypedDict):
 class ElementTestData(QuestionData):
     test_type: Literal['correct', 'incorrect', 'invalid']
 
+
+def set_weighted_score_data(data: QuestionData, weight_default: int=1) -> None:
+    """
+    Sets main question score to be weighted average of all partial scores. Uses
+    weight_default to fill in a default weight for a question if one is missing.
```
nwalters512, line 60: It looks like you're using "question" to refer to both "a question" and "a scored element within a question". This is a bit confusing and isn't used to refer to pieces of questions anywhere else, so I'd like to avoid introducing it.

```suggestion
    Sets overall question score to be weighted average of all partial scores. Uses
    weight_default to fill in a default weight for a score if one is missing.
```

python/prairielearn.py
```diff
@@ -52,6 +53,45 @@ class QuestionData(TypedDict):
...
+
+        score_total += score * weight
+        weight_total += weight
+
+    data['score'] = score_total / weight_total
+
+def set_all_or_nothing_score_data(data: QuestionData) -> None:
+    "Gives points to main question score if all partial scores are correct"
+
+    data['score'] = 1. if all_questions_correct(data) else 0.
```
nwalters512, line 80: Nit (I just did a double-take when I didn't see any digits after the decimal 🙃)

```suggestion
    data['score'] = 1.0 if all_questions_correct(data) else 0.0
```

python/prairielearn.py
```diff
@@ -52,6 +53,45 @@ class QuestionData(TypedDict):
...
+        if score is None:
+            raise ValueError("Can't set weighted score data if score is None.")
+
+        score_total += score * weight
+        weight_total += weight
+
+    data['score'] = score_total / weight_total
+
+def set_all_or_nothing_score_data(data: QuestionData) -> None:
+    "Gives points to main question score if all partial scores are correct"
```
nwalters512, line 78: Per [PEP 257](https://peps.python.org/pep-0257/), we should still use triple-quotes even if a docstring fits on one line.

```suggestion
    """Gives points to main question score if all partial scores are correct."""
```

python/prairielearn_test.py
```diff
@@ -1,15 +1,66 @@
 # lol
 import sys
 import os
-sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '../../lib')))
+
+sys.path.insert(
+    0, os.path.abspath(os.path.join(os.path.dirname(__file__), "../../lib"))
+)
```
nwalters512, line 7: I know this didn't actually change in this PR, but the diff drew my eyes to this. Is this actually necessary? AFAICT, `../../lib` is outside of the PL repo, but the tests still seen to pass just fine. This git-blames to me; I'm inclined to say we can remove this if you want to do an extra bit of cleanup.

python/prairielearn_test.py
```diff
@@ -1,15 +1,66 @@
...
+    expected_score: float,
+) -> None:
 
-def test_inner_html():
-    e = lxml.html.fragment_fromstring('<div>test</div>')
-    assert pl.inner_html(e) == 'test'
+    question_data["partial_scores"] = {
+        "p1": prepare_partial_score_dict(score_1, 2),
+        "p2": prepare_partial_score_dict(score_2, 4),
+        "p3": prepare_partial_score_dict(score_3),  # No weight tests default behavior
```
nwalters512, line 61: I'd be inclined to just inline this function; IMO it just obscures what this test is doing.

```suggestion
        "p1": {"score": score_1, "weight": 2},
        "p2": {"score": score_2, "weight": 4},
        "p3": {"score": score_3},  # No weight tests default behavior
```

python/prairielearn.py
```diff
@@ -52,6 +53,45 @@ class QuestionData(TypedDict):
...
+        weight_total += weight
+
+    data['score'] = score_total / weight_total
+
+def set_all_or_nothing_score_data(data: QuestionData) -> None:
+    """Gives points to main question score if all partial scores are correct."""
+
+    data['score'] = 1.0 if all_questions_correct(data) else 0.0
+
+def all_questions_correct(data: QuestionData) -> bool:
```
nwalters512, line 82: Another use of the word "question" to refer to something that's not actually the question as a whole. Maybe name this `all_partial_scores_correct`?

nwalters512 (approved):
I won't require any changes here, but I left a few suggestions for your consideration.

python/prairielearn_test.py
eliotwrobson: I think I asked about this somewhere else and had the same question. As far as I can tell it's not needed, will remove.

python/prairielearn.py
eliotwrobson: Renamed! I think merging this should wait until we have the linter running over these files (which should be resolved after the black PR is done).

pyproject.toml
```diff
@@ -16,8 +16,10 @@ skip_glob = ["tools/*", "exampleCourse/*", "testCourse/*", "node_modules/*"]
 [tool.pyright]
 include = [
     "./elements/pl-hidden-hints",
-    "./elements/pl-order-blocks",
```
nwalters512, line 19: IIRC we should either add `--skipunannotated` or remove this, but not both - is that correct?

nwalters512 (approved):

pyproject.toml
eliotwrobson: Not quite; the issue is that if we add ` --skipunannotated` (which we should to have typechecking for annotated functions in `prairielearn.py`), then order blocks won't have any typechecking done (since there aren't any annotations). So it's actually ok to leave it in the include, but pyright won't do anything.

Although this technically disables a dev tool for order blocks, those annotations need to be added back anyway, and the type inference that can be done without annotations is super limited.

Issue #6347, opened 2022-09-19
Feature Request: line breaks in `pl-python-variable` for large dictionaries


coatless:
The `pl-python-variable` is great for displaying python variable contents if the amount of material shown is small. However, if there is lots of data, say for a web API response, then it quickly becomes an unfeasible view.

The goal would be to allow `pl-python-variable` to go from: 

```python
{ "key1": "value1",  "key2": "value2",  "key3": "value3"}
```

To something along the lines of: 

```python
{ 
 "key1": "value1", 
 "key2": "value2",  
 "key3": "value3"
}
```

The temporary solution proposed by @echuber2 on [Slack](https://prairielearn.slack.com/archives/C266KEH9A/p1663615555186629?thread_ts=1663551225.140169&cid=C266KEH9A):

```python
import json
o = {'name': 'bob'}
json.dumps(o, indent=1)
'{\n "name": "bob"\n}'
```

echuber2:
Until an option is added to `pl-python-variable` to do this automatically, you can use a drop-in wrapper around basic types like this:

```python3
#!/usr/bin/env python3

import json

class IndentWrapper:
    def __init__(self, obj, indent=1):
        self.obj = obj
        self.indent = indent
    def __repr__(self):
        try:
            return json.dumps(self.obj, indent=self.indent)
        except TypeError:
            return repr(self.obj)

o = {'name': 'bob'}
w = IndentWrapper(o)
# Now display variable w
```

If you try to wrap something that isn't JSON serializable like a set, then the exception handler will use the default repr.

Issue #6609, opened 2022-11-04
Feature: `pl-collapse` decorative bootstrap element for collapsing content


coatless:
When working with a _large_ data table, there needs to be the ability to collapse large tables. 

One option would be to use the [`collapse`](https://getbootstrap.com/docs/4.3/components/collapse/) bootstrap styling:
 
<img width="800" alt="df-sales-example" src="https://user-images.githubusercontent.com/833642/199861248-3a5a5109-527f-4a73-a5bc-f5ce66f217bc.png">

```html
<p>
  <button class="btn btn-primary" type="button" data-toggle="collapse" data-target="#collapse-df-sales" aria-expanded="false" aria-controls="collapseExample">
    Data Frame: df_sales (click to view)
  </button>
</p>
<div class="collapse" id="collapse-df-sales">
  <div class="card card-body">
      <pl-python-variable params-name="df_sales"></pl-python-variable>
  </div>
</div>
```

Another possible option would be to use the [`details` HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) tag:

<img width="279" alt="df-sales-details-example" src="https://user-images.githubusercontent.com/833642/199861876-942b5eb2-bee2-45e8-89ec-4f6d6e17e258.png">

```html
<details>
  <summary>Data Frame: df_sales (click to view)</summary>
  <pl-python-variable params-name="df_sales"></pl-python-variable>
</details>
```

If `collapse` could be added to `pl-python-variable`, there could be some nice truncation and a decrease in custom styling:

```html
<pl-python-variable params-name="df_sales" collapse="true"></pl-python-variable>
```

nwalters512:
Have you considered using the [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) element? It's not nearly as much markup as what you have, and it allows us to avoid setting a precedent of building this into arbitrary elements. IMO we should keep purely presentational things like this out of elements wherever possible.

coatless:
@nwalters512 I did this after chatting with @echuber2 on Slack and updated the example above, c.f. 

https://prairielearn.slack.com/archives/C266KEH9A/p1667519694812719 

I think it might be time to consider more decorative elements if the goal is to not build design functionality into submission elements. c.f. #1753

nwalters512:
[...]

PR #6744, opened 2022-12-03
`pl-python-variable` overhaul

### LINKED ISSUE #2706: Add function to clean numeric data in dataframes
(From discussion in https://github.com/PrairieLearn/PrairieLearn/pull/2702)

Provide a `df_clean` function which

1. if data is string, Latex etc - just leave it alone and pass through
2. if all column data is float format (except header), check magnitude of ALL numbers in a columns and use some simple logic like (really big or really small use e format else use d format), sort of like
```
if -3 < log_10 (max mag) < 3 convert to string in e format with desired # of figs
else convert to string in d format with desired # of figs
```

We should add this to the freeform python lib

### LINKED ISSUE #6131: Add option to display object types for Pandas DF in pl-python-variable
With [{pandas} v1.5.0](https://pandas.pydata.org/pandas-docs/dev/whatsnew/v1.5.0.html#styler) right around the corner, it would be great to incorporate the [`styler` enhancements](https://pandas.pydata.org/pandas-docs/dev/reference/api/pandas.io.formats.style.Styler.concat.html) that allow for information to be placed in the footer of a rendered pandas data frame. In particular, the `dtype` for each data frame column could be incorporated in the footer.

Example: 

```python 
descriptors = df.agg([lambda s: s.dtype])
#descriptors.index = ["dtype"]
other = (descriptors.style
         .applymap(lambda v: "font-weight: bold;"))
styler = (df.style
            .set_table_styles([{"selector": ".foot_row0",
                                "props": "border-top: 1px solid black;"}]))
styler.concat(other)  
```

Without index on final row:

styler-no-final-row-index-pandas-df

With index on final row:

styler-final-row-with-index-pandas-df

--- 

The `styler` component would need to be placed _before_ the `.to_html()` call in `pl-python-variable`:

https://github.com/PrairieLearn/PrairieLearn/blob/f7ad14591fa639ed37c53fa214b0153c7da7fd6f/elements/pl-python-variable/pl-python-variable.py#L41

eliotwrobson:
EDIT: Moved back to draft because there have been questions about how to proceed with respect to the creation of a new dataframe element.

ORIGINAL TEXT:
Resolves all of the following:

- https://github.com/PrairieLearn/PrairieLearn/issues/6501
- https://github.com/PrairieLearn/PrairieLearn/issues/6131
- https://github.com/PrairieLearn/PrairieLearn/issues/2706
- https://github.com/PrairieLearn/PrairieLearn/issues/6347

@coatless all issues opened by you. Disables linting dependency that is incompatible with black autoformatter (in preparation for #5027), and uses black to format the backend code for this element. Also adds dependency for type signatures for pandas to aid in development.

EDIT: Screenshot of table with datatype at the bottom and new default integer formatting:

image

EDIT 2: Now incorporates floating point and integer formatting, so now resolves #2706 as well. Screenshot of this behavior is below. Open to more discussion on whether commas in integers should be the default behavior:

image


elements/pl-python-variable/pl-python-variable.py
```diff
@@ -1,52 +1,119 @@
...
+    )
+    show_dtype = pl.get_boolean_attrib(element, "show-dtype", SHOW_DATATYPE_DEFAULT)
+    add_line_breaks = pl.get_boolean_attrib(
+        element, "add-line-breaks", ADD_LINE_BREAKS_DEFAULT
+    )
 
-    if varname not in data['params']:
-        raise Exception('Could not find {} in params!'.format(varname))
+    if varname not in data["params"]:
+        raise ValueError(f"Could not find {varname} in params!")
```
coatless, line 51:
```suggestion
        raise ValueError(f"Could not find {varname} in params! Please make sure to set params-name='variable' in the element.")
```

elements/pl-python-variable/pl-python-variable.py
```diff
@@ -1,52 +1,119 @@
...
-            var_type = var_out['_type']
+    if isinstance(var_out, dict) and "_type" in var_out:
         var_out = pl.from_json(var_out)
 
+    html_list: List[str] = []
+
     # render the output variable
-    if var_type == 'dataframe':
-        html += var_out.to_html(header=show_header, index=show_index, classes=['pl-python-variable-table'])
+    if isinstance(var_out, pandas.DataFrame) and not force_text:
```
coatless, line 62: The instance check probably needs to happen on `var_type` if `dataframe-v2` is included.

elements/pl-python-variable/pl-python-variable.py
```diff
@@ -1,52 +1,119 @@
...
+        frame = cast(pandas.DataFrame, var_out)
+        frame_style = frame.style
+
+        if show_dtype:
```
coatless, line 66: Excellent to see this incorporated!

elements/pl-python-variable/pl-python-variable.py
```diff
@@ -1,52 +1,119 @@
...
+            )
+            return f"{{\n{inner_string}\n}}"
+
+        elif isinstance(var, list):
+            inner_string = ",\n".join(f"    {repr(elem)}" for elem in var)
+            return f"[\n{inner_string}\n]"
 
-        text = prefix + repr(var_out) + suffix
-        html += '<pl-code language="python" no-highlight="{}">{}</pl-code>'.format(no_highlight, text)
+        raise ValueError(f"Line breaks not supported for type {type(var).__name__}")
```
coatless, line 117: Should this be promoted to `python/prairielearn.py`?

coatless (commented):
Minor notes related to the proposed DF v2 format and the re-structuring of `var_type`.

elements/pl-python-variable/pl-python-variable.py
eliotwrobson: As I understand it, `dataframe-v2` will be an alternative encoding for `pl.from_json`, meaning that the conversion to `var_out` will still happen first (on the lines above).

elements/pl-python-variable/pl-python-variable.py
eliotwrobson: I'm not so sure. I actually don't know of many places where we have to produce a nicer string corresponding to an internal collection (i.e. set or dict).

elements/pl-python-variable/pl-python-variable.py
eliotwrobson: Actually, if your main use case is using dataframes, I don't think it's unreasonable to bundle that in with these changes here. I can can close that other PR later. Though this means that this PR is going to have to wait for the BigO element to go in, since that one modifies the backend Python code too (and I don't want to have to rebase that on top of changes made here).

elements/pl-python-variable/pl-python-variable.py
nwalters512: Unless this code is used in at least one other element, it should not be moved there.

docs/elements.md
```diff
@@ -1200,6 +1200,8 @@ def generate(data):
 | `show-header`     | boolean | true    | Show the header row of a DataFrame in default mode. (No effect in `text` mode.)                                                                             |
 | `show-index`      | boolean | true    | Show the index column of a DataFrame in default mode. (No effect in `text` mode.)                                                                           |
 | `show-dimensions` | boolean | true    | Show a footer with the dimensions of a DataFrame in default mode. (No effect in `text` mode.)                                                               |
+| `show-dtype`      | boolean | false   | Append a row to a DataFrame containing column data types for display. (No effect in `text` mode.)                                                           |
+| `add-line-breaks` | boolean | false   | Add line breaks for Python dictionaries or lists.                                                                                                           |
```
nwalters512, line 1204: Does this need to be configurable? In other words, is there any reason folks might *not* want to show line breaks?

elements/pl-python-variable/pl-python-variable.py
nwalters512, line 51: `KeyError` is probably more appropriate here.

```suggestion
        raise KeyError(f"Could not find {varname} in params!")
```

docs/elements.md
eliotwrobson: Yes. For short dictionaries like `{1:2, 2:4, 3:6}` it may be more convenient than having multiple lines. Also having the ability to default to false means this feature won't change the way people's existing questions look.

elements/pl-python-variable/pl-python-variable.py
eliotwrobson: @coatless Added new dataframe encoding and corresponding tests. Might be worth adding in more tests with a few more test frames but that can be done on top of the existing framework. Now resolves #6501 as well.

python/prairielearn_test.py
```diff
@@ -1,15 +1,138 @@
 # lol
 import sys
 import os
-sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '../../lib')))
 
+sys.path.insert(
```
eliotwrobson, line 5: @nwalters512 Does this line need to be here? When I removed it I had no issues running pytest.

eliotwrobson:
Another example dataframe (using an example @coatless provided):

image

python/prairielearn.py
```diff
@@ -93,7 +94,15 @@ def to_json(v):
...
-        return {'_type': 'dataframe', '_value': {'index': list(v.index), 'columns': list(v.columns), 'data': v.values.tolist()}}
+        df_modified_names = v.copy()
+
+        indexing_dtype = df_modified_names.columns.dtype
+        if indexing_dtype == np.float64 or indexing_dtype == np.int64:
+            df_modified_names.columns = df_modified_names.columns.astype('string')
+
+        encoded_json_df = df_modified_names.to_json(orient="table", date_format="iso")
+
+        return {'_type': 'dataframe-v2', '_value': encoded_json_df}
```
coatless, line 105: Added some comments to discuss the workaround on export. 

```suggestion
        # The next lines of code are required to address the JSON table-orient 
        # generating numeric keys instead of strings for an index sequence with
        # only numeric values (c.f. pandas-dev/pandas#46392)
        df_modified_names = v.copy()

        indexing_dtype = df_modified_names.columns.dtype
        if indexing_dtype == np.float64 or indexing_dtype == np.int64:
            df_modified_names.columns = df_modified_names.columns.astype('string')

        # For version 2 storing a data frame, we use the table orientation alongside of
        # enforcing a date format to allow for numeric values
        # Details: https://pandas.pydata.org/docs/reference/api/pandas.read_json.html
        encoded_json_df = df_modified_names.to_json(orient="table", date_format="iso")

        return {'_type': 'dataframe-v2', '_value': encoded_json_df}
```

coatless:
@eliotwrobson awesome! Thanks for adding in the new converters. I think there's only a backward compatibility issue that @mwest1066 would flag (e.g. we removed the v1 data frame export).

I'm proposing adding the `df_version` parameter into `pl.to_json()` to allow for a version specification. The default can be set to version `1` with a graceful swap coming later.

python/prairielearn.py
```diff
@@ -93,7 +94,15 @@ def to_json(v):
             M.append(row)
         return {'_type': 'sympy_matrix', '_value': M, '_variables': s, '_shape': [num_rows, num_cols]}
     elif isinstance(v, pandas.DataFrame):
```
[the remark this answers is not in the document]
eliotwrobson: I think this should be a boolean flag, but it's easy enough to set this to be the default and modify the test cases. correspondingly.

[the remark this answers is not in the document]
eliotwrobson: @coatless actually, is the new encoding incompatible with the old? Does it erase data in some cases that might be useful? Not opposed to adding it or anything just curious.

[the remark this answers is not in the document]
coatless: I think the issue might be that some courses are expecting the data frame to be passed in the same way if the have custom elements. 

The v1 format stores data in a very different format as columns are separated out from values.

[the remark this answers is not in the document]
eliotwrobson: Made the legacy encoding the default as you suggested

coatless:
@eliotwrobson maybe include a news article describing the refreshed `pl-python-variable` features and new data frame structure? 

@nwalters512 & @mwest1066 any objections to adding a news article?

eliotwrobson:
@coatless what do you mean by a "news article"? Do you have any examples?

python/prairielearn.py
```diff
@@ -162,6 +181,8 @@ def from_json(v):
                     return pandas.DataFrame(index=val['index'], columns=val['columns'], data=val['data'])
                 else:
                     raise Exception('variable of type dataframe should have value with index, columns, and data')
+            elif v['_type'] == 'dataframe-v2':
+                return pandas.read_json(v['_value'], orient="table")
```
coatless, line 185:
```suggestion


                # Convert native JSON back to a string representation so that 
                # pandas read_json() can process it.
                value_str = json.dumps(v['_value'])
                return pandas.read_json(value_str, orient="table")
```

python/prairielearn.py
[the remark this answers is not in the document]
coatless: @eliotwrobson I probably would stick with a versioning number, e.g. 1, 2, 3 instead of True/False, since there may be a need to modify further if the upstream pandas bug is resolved.

python/prairielearn.py
```diff
@@ -93,7 +94,25 @@ def to_json(v):
...
+            indexing_dtype = df_modified_names.columns.dtype
+            if indexing_dtype == np.float64 or indexing_dtype == np.int64:
+                df_modified_names.columns = df_modified_names.columns.astype('string')
+
+            # For version 2 storing a data frame, we use the table orientation alongside of
+            # enforcing a date format to allow for numeric values
+            # Details: https://pandas.pydata.org/docs/reference/api/pandas.read_json.html
+            encoded_json_df = df_modified_names.to_json(orient="table", date_format="iso")
+
+            return {'_type': 'dataframe-v2', '_value': encoded_json_df}
```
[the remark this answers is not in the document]
coatless: @eliotwrobson this came up from a comment @mwest1066 made in #6501 as the first go around was string encoded with escape sequences instead of native JSON 😢

coatless:
@eliotwrobson 

Regarding news items, see: 
https://github.com/PrairieLearn/PrairieLearn/tree/master/news_items

Or on PrairieLearn:

https://www.prairielearn.org/pl/news_items

python/prairielearn.py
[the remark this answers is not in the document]
eliotwrobson: Ah that makes sense. My main question is really if there's a good way of checking this encoding is done properly in the test suite (as opposed to what's there now).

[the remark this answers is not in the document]
coatless: @eliotwrobson maybe use `jsonschema` to validate the pandas output? 

I'm not sure we need that to be further scrutinized though. The check now covers the most important part: re-creating the original Pandas data frame.

eliotwrobson:
> Regarding news items, see: 
> https://github.com/PrairieLearn/PrairieLearn/tree/master/news_items
> 
> Or on PrairieLearn:
> 
> https://www.prairielearn.org/pl/news_items

Ah I see. Based on the other news items, I don't think this is quite big enough to warrant that (since this is really just affecting a single element). Though we should find a way of letting people that use this know about the new capabilities.

python/prairielearn.py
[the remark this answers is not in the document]
eliotwrobson: That makes sense. I'll play with the new encoding when I get the chance.

python/prairielearn.py
eliotwrobson: @coatless Added the changes you suggested and tests + element seem to be working!

python/prairielearn.py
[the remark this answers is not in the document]
eliotwrobson: Makes sense. Have changed to the version number and defaulted to 1.

eliotwrobson:
@coatless another question, should we incorporate any ideas from https://github.com/PrairieLearn/PrairieLearn/issues/6609 in this? It seems like the consensus there was to not include this functionality in the PL element.

I'm inclined to agree with the decision there, I don't think there's any benefit to having it built into the PL element if instructors can add it themselves by wrapping some boilerplate bootstrap elements (in contrast to some of the tricky styling needed for the dataframes).

coatless:
@eliotwrobson could we please use `digits` for the `num-sig-fig` to bring it into parity with other pl elements? c.f.
https://prairielearn.readthedocs.io/en/latest/elements/#pl-number-input-element 

re: #6609 we probably should discuss in real-time with @nwalters512 in #pl-dev a way forward.

eliotwrobson:
@coatless Changed to digits and simplified the backend formatter code some more.

coatless:
Thinking about the changes more, this feels like we created two different elements: 

1. Scalar/Dictionary/List/Tuple
2. Pandas data frame

Should we potentially split off the data frame components into a `pl-dataframe` element? We could add a redirect to the new element in the existing code similar to the `pl-code` example:

https://github.com/PrairieLearn/PrairieLearn/pull/6744/files#diff-11f4bbc30d18502211c5e62d481280c79d48997be67a7ec226b2152e30c8cf13R118-R120

eliotwrobson:
@coatless I actually agree that it feels like we've created two different elements here. However, separating these could create breaking changes for existing questions, as the dataframe display functionality was already present prior to the changes we've made here. I think that it probably wasn't a great idea to merge these two somewhat disparate display functionalities, but that decision has kinda already been made.

eliotwrobson:
Going to close this for now in light of discussion in #6789, seems like the way to proceed is to split off into the dataframe element and then add legacy support for the python variable element (with possible new features).

Issue #5017, opened 2021-10-21
Negative weights for pl-graph adjacency matrix


stevenxpan:
It doesn't seem that the pl-graph element is able to handle an adjacency matrix that has edges with negative weights:

![image](https://user-images.githubusercontent.com/52120379/138191754-f783c9d7-3aa5-4ae6-a26d-188b6e63dbfb.png)

This issue can be reproduced by using the sample question [here](https://github.com/PrairieLearn/PrairieLearn/tree/master/exampleCourse/questions/element/graph) and then changing line 11 of server.py to `mat = -1 * mat / la.norm(mat, 1, axis=0)`.

Is there a setting that can be used to enable displaying of negative weight edges?

PR #6745, opened 2022-12-03
Add negative weight and undirected display to `pl-graph`

eliotwrobson:
See title. Also added type hints and cleaned up some rendering code. Resolves https://github.com/PrairieLearn/PrairieLearn/issues/5017

image

image


elements/pl-graph/pl-graph.py
```diff
@@ -33,35 +35,30 @@ def graphviz_from_adj_matrix(element, data):
...
-    for node in mat_label:
-        G.add_node(node)
-
-    for i, out_node in enumerate(mat_label):
-        for j, in_node in enumerate(mat_label):
-            x = mat[j, i]
-            if (x > 0):
+    for in_node, row in zip(mat_label, mat):
+        for out_node, x in zip(mat_label, row):
+            if (show_negative_weights) or x > 0:
```
stevenxpan, line 61:
```suggestion
            if show_negative_weights or x > 0:
```

stevenxpan (approved):

elements/pl-graph/pl-graph.py
eliotwrobson: @stevenxpan I'm realizing that there should be a way to hide edges even when there are negative weights. Do you think that hiding edges when the matrix entries are equal to `nan` is a good idea?

docs/elements.md
```diff
@@ -1592,6 +1592,7 @@ def generate(data):
...
 | `weights`                   | boolean | `None`             | When using an adjacency matrix, whether or not to show the edge weights. By default will automatically show weights for stochastic matrices (when they are not binary `0`/`1`).                                                                                         |
 | `weights-digits`            | integer | `"2"`              | When using an adjacency matrix, how many digits to show for the weights.                                                                                                                                                                                                |
+| `parse-negative-weights`    | boolean | false              | Whether to recognize negative weights in an adjacency matrix. If set to false, then all weights at most 0 are ignored (not counted as an edge). If set to true, then all weights that are not `None` are recognized.                                                    |
```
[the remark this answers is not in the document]
eliotwrobson: Will change the name to `negative-weights`.

So the issue is that, the current behavior (equivalent to setting this to false) uses anything nonpositive (0 or negative) to indicate that an edge should not be displayed. However, with this set to true, any negative or 0 value will be displayed as part of the graph. So if this is defaulted to true, then existing uses of this element will display the complete graph with some edges set to weight 0. This issue is actually why I had to use `None` to indicate no edge in the negative edge weight case.

Reading over this again, it may also be worth changing the default behavior to allow `None` to indicate no edge in the non-negative case for consistency.

nwalters512 (approved):

Issue #5533, opened 2022-03-23
Hide "Description" column if no description is provided


jasonxia17:
In some problems, we don't write a description in this table because the descriptions are already provided earlier in the problem statement. When this occurs, it would be nice if the "Description" column was not displayed at all instead of being left blank.

![image](https://user-images.githubusercontent.com/8812440/159773418-e0833d63-65ef-4969-ae28-b8e66b8444c0.png)

PR #6746, opened 2022-12-03
feat: `pl-external-grader-variables` description hiding and inline definitions

eliotwrobson:
Resolves https://github.com/PrairieLearn/PrairieLearn/issues/5533 and adds a frontend mode to the `pl-external-grader-variables` element, along with reworking the backend to use type hints + add formatting.


nicknytko:
Neat!  I like this idea of being able to specify variables either in the HTML or the `server.py`.  That being said, for variables defined in the question html do they saved to `data` somewhere?  I ask because at least the Python autograder will use `data['names_for_user']` and `data['names_from_user']` to control what gets passed to and from the student code, so I worry that we might break the grader if that's not being sent over.

eliotwrobson:
@nicknytko They don't get saved to data (I set it up so that they could be changed without having to regenerate the question code). But this is pretty easy to change back.

eliotwrobson:
Tried this locally on an external grader and it seems to work! Defining variables on the frontend seems like a good way to reduce the number of files that people need to create; our course has a ton of `server.py` files that are just defining variables for the user in an external grader.

elements/pl-external-grader-variables/pl-external-grader-variables.py
```diff
@@ -1,33 +1,66 @@
...
 
-    params_name = pl.get_string_attrib(element, 'params-name', PARAMS_NAME_DEFAULT)
-    if params_name is None:
-        raise Exception('Attribute "params-name" is not defined.')
     if params_name not in data['params']:
-        raise Exception(f"Variable name {params_name} does not exist in data['params'].")
+        data["params"][params_name] = frontend_variables
+    else:
+        raise ValueError(
+            "Cannot define variables from both backend and frontend for variable name '{params_name}'"
```
nicknytko, line 41:
```suggestion
            "Cannot define variables from both 'question.html' and 'server.py' for variable name '{params_name}'"
```

elements/pl-external-grader-variables/pl-external-grader-variables.py
```diff
@@ -1,33 +1,66 @@
...
+            raise ValueError(
+                f"Tags inside of pl-external-grader-variables must be pl-variable, not '{child.tag}'."
+            )
 
-    params_name = pl.get_string_attrib(element, 'params-name', PARAMS_NAME_DEFAULT)
-    if params_name is None:
-        raise Exception('Attribute "params-name" is not defined.')
     if params_name not in data['params']:
-        raise Exception(f"Variable name {params_name} does not exist in data['params'].")
+        data["params"][params_name] = frontend_variables
```
nicknytko, line 38: Do you think we should throw an error if we don't have anything define either in the question html or the server file?  I think it could be nice to have a small sanity check, I don't think we would run in the case where we don't want *anything* from the user?

images/plbase/python-requirements.txt
```diff
@@ -1,6 +1,5 @@
 chevron==0.14.0
 flake8==5.0.4
-flake8-quotes==3.3.1
```
nicknytko, line 3: Does this need to be removed?

nicknytko (commented):
Few minor comments, this is generally looking pretty good imo

images/plbase/python-requirements.txt
eliotwrobson: Discussed in some other PRs, but this is inconsistent with the black autoformatter, which we eventually want to run over the whole codebase (see https://github.com/PrairieLearn/PrairieLearn/issues/5027). Unfortunately that means a developer can't run black while this is installed (and I used this to format the code in the element file that I modified, which made development much easier).

elements/pl-external-grader-variables/pl-external-grader-variables.py
eliotwrobson: I was actually wondering about this too. There's an example of this use case in this PR in the Pandas autograder. Basically, there's actually many instances where either the `names_from_user` or `names_for_user` are empty, and it's nice have an element fill that part of the dictionary without having to explicitly make a `server.py` just to set an empty list.

You raise an important point that not raising an error in this case could make it harder for the end user. Should we add an `empty` boolean flag as an attribute to this element? That flag would be set in the case that the user explicitly wants to declare that parameter as the empty list. That way, we can still raise an error message in the case of typos on the `params-name` attribute.

eliotwrobson: Just added the empty flag. So to define a parameter as the empty list, you need to set the empty flag explicitly, and without this flag set, the element will throw an error if there are no backend variables defined.

nicknytko: I think that checks out. I would sort of assume if there weren't any variables defined anyway then whoever is writing the question could omit the `pl-external-grader-variables` tag altogether.

images/plbase/python-requirements.txt
nicknytko: Ah, gotcha, makes sense.  I think we should avoid making unrelated changes in PRs (though, cc @nwalters512 for opinion here)

eliotwrobson: Agreed, I only made this change because I personally find it a convenient tool to use.

elements/pl-external-grader-variables/pl-external-grader-variables.py
eliotwrobson: Moved the description to be the inner part of the nested `<pl-variable>` tag. To omit, the tag is left empty (but name and type can still be set).

elements/pl-external-grader-variables/pl-external-grader-variables.py
```diff
@@ -1,33 +1,90 @@
...
+        else:
+            raise ValueError(
+                f"Tags inside of pl-external-grader-variables must be pl-variable, not '{child.tag}'."
+            )
 
-def render(element_html, element_index, data):
+    declared_empty = pl.get_boolean_attrib(element, "empty", EMPTY_DEFAULT)
+
+    if declared_empty:
+        if len(frontend_variables) > 0:
```
stevenxpan, line 45:
```suggestion
        if frontend_variables:
```

elements/pl-external-grader-variables/pl-external-grader-variables.py
```diff
@@ -1,33 +1,90 @@
...
+                f"Variable name '{params_name}' was declared empty, but has variables defined in 'question.html'."
+            )
+        elif params_name in data["params"]:
+            raise ValueError(
+                f"Variable name '{params_name}' was declared empty, but has variables defined in 'server.py'."
+            )
+
+        data["params"][params_name] = []
+    elif params_name not in data["params"]:
+        if len(frontend_variables) == 0:
```
stevenxpan, line 56:
```suggestion
        if not frontend_variables:
```

docs/elements.md
```diff
@@ -1697,7 +1703,18 @@ data["params"]["names_from_user"] = [
...
+  <pl-variable name="x" type="numpy array (length $n$)"
+    >Solution to $\mathbf{Ax}=\mathbf{b}$.</pl-variable
+  >
+</pl-external-grader-variables>
+
+<!--
+  The following tag defines an empty list for the given params-name.
+  This is useful for some cases where a parameter must be set to empty to run the external grader.
+  Nothing will be displayed from this tag.
+-->
```
nwalters512, line 1716: It's not clear to me what cases this would be useful in. Do you have an example? It honestly feels super weird to me to have an element that just sets a param to `[]`.

elements/pl-external-grader-variables/pl-external-grader-variables.py
```diff
@@ -1,33 +1,90 @@
...
-def prepare(element_html, element_index, data):
+def prepare(element_html: str, data: pl.QuestionData) -> None:
     element = lxml.html.fragment_fromstring(element_html)
-    pl.check_attribs(element, ['params-name'], [])
+    pl.check_attribs(element, ["params-name"], ["empty"])
+
+    params_name = pl.get_string_attrib(element, "params-name")
+
+    # Get frontend variables
+    frontend_variables = []
```
nwalters512, line 15: What does "frontend" mean in this context? Is this referring to the fact that this data comes from `question.html`? AFAIK we don't use that terminology elsewhere; maybe name this `html_variables` instead?

docs/elements.md
eliotwrobson: This is actually super common, take a look at the Pandas example question that got refactored in this PR. It comes up in cases where you want a student to write a function that the grader code just calls, so there's no need to pass any parameters directly to the student code. I can double check, but I think the external grader requires that be set (even if it's just the empty list).

Just searching our own course, I found 19 examples of setting the empty list in server.py. It would also feel weird not to have this common case covered by HTML definitions, while other cases are.

elements/pl-external-grader-variables/pl-external-grader-variables.py
eliotwrobson: Yes, taking variables from HTML is the intended meaning. Will change.

docs/elements.md
nwalters512: Would it make more sense to update the grader(s) to treat missing or `None` param the same as an empty list?

eliotwrobson: I'm actually not sure. There's some other discussion on this PR about explicitly declaring empty vs silently doing something different with the default. Like if there's a typo in the variable name declaration, the external grader would just pick up None for the variables to be used, which could be frustrating and hard to debug.

I'm not sure who wrote the Python autograder either, but I think @jonatanschroeder wrote some of the other autograders; I'm curious what his thoughts are.

jonatanschroeder: I have never worked with Python questions in PL, so I don't have an opinion here. The other autograders I've worked with (including C, Java and some custom graders) don't use global variables for this element to make sense in those contexts.

nwalters512: Yeah, the Python autograder is the only one I know of that actually does anything with `names_for_user` or `names_from_user`. I believe @nicknytko himself wrote the Python autograder. In retrospect, I would have been inclined to name this thing `<pl-python-external-grader-variables>` to make it clear that this element *only* does anything useful for the Python autograder. Maybe we can make that clear in the docs?

eliotwrobson: Yeah, I think we can make that a little more clear in the docs. With this change actually, you could use it to display information for other autograders, but they don't use globals. So it's worth being explicit about recommended use cases.

nwalters512: I think this change is actually sub-optimal for non-Python graders. If an instructor interprets this element as "oh, this is just a nice way to render a table of variables", they might get bitten by the fact that any updates to the variables in `question.html` won't be updated for students who have already started a question because the data is snapshotted into `params` when the question is generated. This runs counter to how other display elements work. So we might want to call that out in the docs too.

nwalters512: I suppose part of my hesitation here stems from the fact that I view elements in `question.html` as being fundamentally related to rendering. Having a specific configuration of an element that doesn't render anything and only adds something to `params` feels very idiosyncratic to me. Frankly, an element that modifies `params` but doesn't also do any submission-handling/validation/grading feels idiosyncratic too, though I acknowledge it's useful for this specific external grader.

eliotwrobson: Yeah, I'm kindof in agreement actually. The way this grader works is a bit cumbersome compared to the others, which is why having this element becomes convenient. If there was some better way to configure the grader outside of setting these lists, I probably wouldn't advocate for this change. But as-written, this change can alleviate a lot of the bloat from having tons of server files that are just defining variables. Our course repo has a _ton_ of these, and being able to remove them will make maintenance significantly easier.

nwalters512: Yeah, I gotcha. I acknowledge the convenience, so I won't block this change.

Not to further derail this conversation, but spitballing for future development ideas: if it were possible to populate the initial params statically via a `params` key in `question.json`, or perhaps via YAML frontmatter in a hypothetical `question.md` file, would that be an attractive alternative to you?

eliotwrobson: I think that could be interesting, though the main use case for populating the params dict is to have things that you can use in later templating / question creation, so this edge case benefits heavily from not having to create the extra file (since there's no extra custom logic). It seems like a more limited version of some of the other proposed ways of having a question take parameters from the corresponding assessment.

Maybe not a bad idea but I'm struggling to come up with use cases right now.

docs/elements.md
```diff
@@ -1703,7 +1703,15 @@ An overlay is pre-defined as a "overlay area" with a static size. By default, el
 
 ### `pl-external-grader-variables` element
 
-Displays variables that are given to the student, or expected for the student to define in externally-graded questions. The list of variables should be stored in `data['params']` and has the following format:
+Displays variables that are given to the student, or expected for the student to define in externally-graded Python questions.
+In particular, **we advise against using this element for any non-Python externally graded question**, since the way this element
+stores parameters is specifically for use with the Python autograder.
+
+The list of variables can be defined in in `data['params']` or the question HTML itself (but not both!). If defined in the
```
nwalters512, line 1710: Nit for consistency with https://github.com/PrairieLearn/PrairieLearn/pull/6794:

```suggestion
The list of variables can be defined in in `data["params"]` or the question HTML itself (but not both!). If defined in the
```

docs/elements.md
```diff
@@ -1703,7 +1703,15 @@ An overlay is pre-defined as a "overlay area" with a static size. By default, el
...
+question HTML itself, the variable information is added to `data['params']` for use by the external grader. If no descriptions
```
nwalters512, line 1711:
```suggestion
question HTML itself, the variable information is added to `data["params"]` for use by the external grader. If no descriptions
```

docs/elements.md
```diff
@@ -1703,7 +1703,15 @@ An overlay is pre-defined as a "overlay area" with a static size. By default, el
...
+are present, this colmun is hidden in the table shown to the student.
+
+If stored in `data['params']`, the variables list has the following format:
```
nwalters512, line 1714:
```suggestion
If stored in `data["params"]`, the variables list has the following format:
```

docs/elements.md
```diff
@@ -1738,16 +1757,21 @@ def generate(data):
...
 ```
 
 #### Customizations
 
-| Attribute     | Type   | Default | Description                                                                                         |
-| ------------- | ------ | ------- | --------------------------------------------------------------------------------------------------- |
-| `params-name` | string | `None`  | Name of variable specification in `data['params']` to display, the format for which is given above. |
+| Attribute     | Type    | Default | Description                                                                                                                                 |
+| ------------- | ------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------- |
+| `params-name` | string  | -       | Name of variable specification in `data['params']` to display, the format for which is given above.                                         |
```
nwalters512, line 1766:
```suggestion
| `params-name` | string  | -       | Name of variable specification in `data["params"]` to display, the format for which is given above.                                         |
```

docs/elements.md
```diff
@@ -1703,7 +1703,15 @@ An overlay is pre-defined as a "overlay area" with a static size. By default, el
 
 ### `pl-external-grader-variables` element
 
-Displays variables that are given to the student, or expected for the student to define in externally-graded questions. The list of variables should be stored in `data['params']` and has the following format:
+Displays variables that are given to the student, or expected for the student to define in externally-graded Python questions.
+In particular, **we advise against using this element for any non-Python externally graded question**, since the way this element
+stores parameters is specifically for use with the Python autograder.
```
nwalters512, line 1708: It'd be nice to provide an alternative:

```suggestion
In particular, **we advise against using this element for any non-Python externally graded question**, since the way this element stores parameters is specifically for use with the Python autograder. If you want to display a table of variables for a question using a different autograder, consider using a standard HTML `<table>` instead.
```

exampleCourse/questions/demo/autograder/python/pandas/question.html
```diff
@@ -8,9 +8,11 @@
         Create a Pandas DataFrame, <code>df</code>, that matches the following format:
         <pl-python-variable params-name="df"></pl-python-variable>
     </p>
-
+    <pl-external-grader-variables params-name="names_for_user" empty="true"></pl-external-grader-variables>
```
nwalters512, line 11: This still looks so strange to me 🙈

nwalters512 (approved):

docs/elements.md
```diff
@@ -1783,7 +1783,14 @@ An overlay is pre-defined as a "overlay area" with a static size. By default, el
 
 ### `pl-external-grader-variables` element
 
-Displays variables that are given to the student, or expected for the student to define in externally-graded questions. The list of variables should be stored in `data["params"]` and has the following format:
+Displays variables that are given to the student, or expected for the student to define in externally-graded Python questions.
+In particular, **we advise against using this element for any non-Python externally graded question**, since the way this element stores parameters is specifically for use with the Python autograder. If you want to display a table of variables for a question using a different autograder, consider using a standard HTML `<table>` instead.
```
nwalters512, line 1787: Folks may have their own Python graders; we should be more specific here.

```suggestion
Displays variables that are given to the student, or expected for the student to define in externally-graded Python questions.

**We advise against using this element for any question not using the `prairielearn/grader-python` Docker image**, as the way this element stores parameters is specifically for use with that autograder. If you want to display a table of variables for a question using a different autograder, consider using a standard HTML `<table>` instead.
```

elements/pl-external-grader-variables/pl-external-grader-variables.py
```diff
@@ -1,34 +1,90 @@
...
-    params_name = pl.get_string_attrib(element, "params-name", PARAMS_NAME_DEFAULT)
-    if params_name is None:
-        raise Exception('Attribute "params-name" is not defined.')
-    if params_name not in data["params"]:
-        raise Exception(
-            f"Variable name {params_name} does not exist in data['params']."
-        )
+    params_name = pl.get_string_attrib(element, "params-name")
 
+    # Get frontend variables
```
nwalters512, line 14: I get what you're going for with "frontend", but we don't use that terminology elsewhere (I'm trying hard to avoid introducing new terminology for things):

```suggestion
    # Get any variables defined in HTML
```

elements/pl-external-grader-variables/pl-external-grader-variables.py
```diff
@@ -1,34 +1,90 @@
...
+                "type": pl.get_string_attrib(child, "type"),
+            }
+
+            description_html = pl.inner_html(child)
+
+            # Check for empty string, inner_html doesn't return None
+            if description_html:
+                var_dict["description"] = description_html
+
+            html_variables.append(var_dict)
```
nwalters512, line 32: Nit: I believe this will be equivalent, and is a little terser?

```suggestion
            var_dict = {
                "name": pl.get_string_attrib(child, "name"),
                "type": pl.get_string_attrib(child, "type"),
                # Handle empty string, inner_html doesn't return None
                "description": pl.inner_html(child) or None,
            }

            html_variables.append(var_dict)
```

elements/pl-external-grader-variables/pl-external-grader-variables.py
```diff
@@ -1,34 +1,90 @@
...
+
+    params_name = pl.get_string_attrib(element, "params-name")
+
+    # Get final variables list
     names_user_description = data["params"][params_name]
     has_names_user_description = len(names_user_description) > 0
 
+    # Show descriptions if any variable has them set
+    has_descriptions = any(
+        "description" in var_description for var_description in names_user_description
```
nwalters512, line 81: To go along with my other suggested change:
```suggestion
        d.get("description", None) for d in names_user_description
```

Issue #1695, opened 2019-09-07
Optimize conditional rendering in questions


nwalters512:
Right now, all "conditional" elements (`<pl-answer-panel>`, etc.) are implemented as normal PL elements in Python. This means that for every such panel, we have to serialize the relevant portion of the tree to a string, send to to the Python element code, parse it, reserialize it, send it back to JavaScript, and re-parse it. This could be avoided by making these elements "pseudoelements" that don't have an implementation in Python, but rather are implemented as a core part of the renderer. Thus, to "render" such an element, we just check if the element name is `pl-answer-panel`, and if we aren't rendering that part of the tree, just throw it away; if we are, we can just lift the children elements up to be children of the parent.

Once https://github.com/PrairieLearn/PrairieLearn/issues/1694 is implemented, we could easily implement that as a "pseudoelement" too.

Currently, a course could theoretically create their own version of `pl-answer-panel` or the like by putting it in their own `elements` directory. Although I don't think any course has done that yet, we should be consistent and continue offering that same level of flexibility. That is, we should use a course's implementation of these elements even if they are implemented as "pseudoelements" in the renderer.

echuber2:
Has #1694 been implemented in the necessary way now?

coatless:
@echuber2 added as [`<pl-hide-in-panel>`](https://prairielearn.readthedocs.io/en/latest/elements/#pl-hide-in-panel-element). c.f.  #1710

eliotwrobson:
@nwalters512 is this no longer relevant in light of https://github.com/PrairieLearn/PrairieLearn/pull/7128? I think all of the savings in efficiency from this issue are done more comprehensively in that PR.

nwalters512:
Good call, that's going to be a much better solution to this minor problem. I'll close this!

PR #6747, opened 2022-12-04
Introduce pl-template element

### LINKED ISSUE #4701: Implement new <pl-template> element

eliotwrobson:
First draft of the `pl-template` element, resolves #4701. @mwest1066 A few questions I had while implementing:

1. Should there be render warnings for missing variables? I feel like this makes things easier for the end user (since it helps catch annoying things like typos in variable names), but then variables that are set to false need to be declared explicitly. I'm not sure if enabling this is the right way to go.
2. Will all of the template files be in `serverFilesCourse`? This is true for all of the use cases in our course, but I'm not sure if there are situations where files from different directories are needed (or if there is the need to have variables containing things from multiple different directories).
3. Should we allow people to explicitly declare boolean variables? There's technically no need, since mustache can treat the presence / absence of a string as true/false, but I'm not really a fan of the way mustache silently fails. This can produce unexpected behavior and cause frustration when writing these templates.

Aside from those questions about edge cases, the core feature and implementation are working. Also, adds in type hints for `chevron` that were missing (as this element benefits from having those) and disables `flake8-quotes` to allow for running black autoformatter.


mfbutner:
@eliotwrobson 
1. Maybe make turning on warnings for missing variables an optional parameter that defaults to on? Then those who are expecting the mustache default of just nothing happening can turn it off and get their "normal mustache behavior"
2. I think different directories should be supported. One use case I could think of is if you had a very strict question structure that you wanted to maintain across multiple questions, especially when you have multiple people developing questions for a single course. You could place the general template structure with all of the nice styling in server files course but then have the files that have the actual content in the question folder. Let me know if this isn't clear enough and you want an example.
3. I'm not sure. When I want to optionally render something in Mustache I use a [Section or Inverted Section](https://mustache.github.io/mustache.5.html).


I tried downloading and playing with your element and some other things I noticed were
1. It would be nice to have access to the data dictionary implicitly in the templates as I think it will be very common to pull variables from there. You could of course pass the specific values you want into the template one variable at a time but that seems like an annoying amount of extra work for what I think will be a very common use case. If you added this feature you would need to disallow the user from naming a variables after what is in data (params, submitted_answers, etc)
2. I don't think that there is currently a way to pass more complex data types like lists, dictionaries, and functions to a template. Or at least I couldn't figure out how to. This would prohibit users from using some Mustache features in their templates such as Sections for iteration or Lambdas. I'm not sure how big of a deal it is because these are more advanced features that most users won't be using but it might surprise some to find that the Mustache in templates is not feature complete. This can be worked around by having the user generate the desired text in `server.py` and then pass the finalized text in as the value to to the template variable. One potential drawback of this strategy is the text generated will likely be done in the `prepare` or `generate` functions which are only run when a new variant of the question is created and not every time it is rendered. So if an instructor makes a typo in the text generated but doesn't notice it until after the question is released, once they fix the mistake the fix won't be visible until the student generates a new variant of the question.
3. This is probably a me only thing but I think it might be a good idea to remove the option to specify a variable's value as the contents of a file. I can totally see why this would be helpful but for some reason having it there really tripped me up when thinking about nested templates. After sleeping on it, I realized that this was not at all the point of this feature but for some reason my brain got it mixed up. If this option was removed then the there is only a single way to load in the contents of a file, using a nested `pl-template`. Here is what your example would look like if you couldn't specify a variable's value to be the contents of a file

question.html
```html
<pl-template file-name="outerTemplate.mustache" subdirectory="templates">
  <pl-variable name="is-open">True</pl-variable>
  <pl-variable name="problem-statement">statement.html</pl-variable>
</pl-template>
```

outerTemplate.mustche (only the modified section is shown)
```html
<details class="statement rounded" {{#is-open}}open{{/is-open}}>
    <summary> Problem Statement</summary>
    <pl-template file-name={{problem-statement}} subdirectory="templates">
    </pl-template>
</details>
```

Of course you could still leave this feature in but in the documentation note that loading in a variable's value from a file is kind of like a macro where it does direct text substitution and provide an example of using a nested template. 

Hopefully I'm not coming across as being overly negative. I think what you have already is great and would satisfy the needs of 95%+ of users but I think there will be some that try and use your element in a way slightly different way and they won't be able to do what they want without some interesting workarounds.

eliotwrobson:
>     1. Maybe make turning on warnings for missing variables an optional parameter that defaults to on? Then those who are expecting the mustache default of just nothing happening can turn it off and get their "normal mustache behavior"

I think that makes sense. A lot of the mustache semantics revolve around leaving out certain keys to control rendering behavior, but it's nice to just have this as a "debug mode" while writing a question or figuring out complex rendering logic.

>     2. I think different directories should be supported...

That makes sense. I didn't have that as part of the initial implementation because I wasn't sure if it was something that was needed; will add.

>     3. I'm not sure. When I want to optionally render something in Mustache I use a [Section or Inverted Section](https://mustache.github.io/mustache.5.html).

Likewise, I think it's enough to allow people to leave the body of a variable declaration empty and have that count as `None`. Although this feels a little weird, it's closer to pure mustache semantics, and explicit boolean declaration would just be a convenience anyway.

>     1. It would be nice to have access to the data dictionary implicitly in the templates ...

Makes sense, will add.

>     2. I don't think that there is currently a way to pass more complex data types like lists, dictionaries, and functions to a template.

This seems like a reasonable concern. I think that passing in entries from the data dictionaries will mitigate this somewhat; however, I'm not sure what syntax on the frontend would be best for defining these more complex data structures. It may be more inconvenient to define these on the frontend rather than the backend, so I'm quite sure what the best resolution to this is right now.

>     3. This is probably a me only thing but I think it might be a good idea to remove the option to specify a variable's value as the contents of a file. 
> ...
> Of course you could still leave this feature in but in the documentation note that loading in a variable's value from a file is kind of like a macro where it does direct text substitution and provide an example of using a nested template.

I'm leaning towards keeping this in and adding more detailed examples and documentation. I'd like the element to handle most of the common use cases without the developer having to use recursive rendering, with the option to dig into that for more complex needs. 

> Hopefully I'm not coming across as being overly negative...

Not at all! I greatly appreciate your detailed comments and feedback. This initial implementation was intentionally a little barebones, since I wanted to field some community feedback before moving ahead on final semantics and features. This discussion has been really helpful and gives us a way forward on delivering this element.

mfbutner:
Hi @eliotwrobson . I tried playing around some more with your pl-template and I can't get the `<markdown>` tag to work with it. I first tried the following.

```html
<markdown>

<pl-template file-name="prompt.md" parent-directory="question">
</pl-template>

</markdown>
```

But the markdown wasn't converted at all. If prompt.md contained
```markdown
### Problem 

Multiply two complex numbers together. Complex numbers have the form A*i + B where

A is the imaginary part of the number
B is the real part of the number
i is the imaginary unit and i * i = -1

Your program should ask the user for 2 complex numbers, multiply them together, and then display the result.

### Requirements

- Your Program must compile with the -Wall, -Werror, and -Wextra options enabled

```

Then it rendered as
```html
### Problem Multiply two complex numbers together. Complex numbers have the form A*i + B where A is the imaginary part of the number B is the real part of the number i is the imaginary unit and i * i = -1 Your program should ask the user for 2 complex numbers, multiply them together, and then display the result. ### Requirements - Your Program must compile with the -Wall, -Werror, and -Wextra options enabled
```

I then tried moving the markdown tags into the prompt directly but that also didn't work. I know that that markdown tag behaves very different from the other pl tags but I was wondering if you knew what was going on and how to fix it. 



Thanks for the help

nwalters512:
@mfbutner this is because the `<markdown>` "tag" is processed before any elements are. This is by design, as `<markdown>` emits element code itself (e.g. in the case of fenced code, that gets transformed to `<pl-code>`).

The upshot is that I believe `<pl-template>` won't be compatible with `<markdown>`, at least with the current implementation. It's possible that we could do a second pass of Markdown rendering after all elements have been rendered, though I haven't fully thought through the implications of that.

I'm also thinking about how this could work in a world where we support Markdown more natively via a `question.md` file. Even in that case, I think we'd need a multi-pass approach to make this work.

Alternatively, we could do a single Markdown rendering pass at the _end_, once all elements have executed. I'm not 100% sure if this would have the same semantics as doing it first (e.g., maybe something about the element rendering process is lossy when it comes to Markdown syntax).

mfbutner:
@nwalters512 Thanks for the explanation on what is happening. For now I'll just run my markdown prompts through an html converter and then load that html through `pl-template` instead.

eliotwrobson:
@mfbutner out of curiosity, are you currently using this element in any substantial way?

mfbutner:
@eliotwrobson Right now I was using it just to load my problem prompt into question.html.

 Another use case I was thinking of trying out was making a "Collapsible Section" template so that I could avoid having to both see and write the boostrap that is needed to create one. 

Haven't had any ideas beyond that yet.

elements/pl-template/pl-template.py
```diff
@@ -0,0 +1,108 @@
...
+        "parent-directory",
+        ParentDirectoryEnum,
+        PARENT_DIRECTORY_CHOICE_DEFAULT,
+    )
+    parent_directory = options_dict[parent_directory_choice.value]
+
+    subdirectory = pl.get_string_attrib(element, "subdirectory", SUBDIRECTORY_DEFAULT)
+
+    file_name = pl.get_string_attrib(element, "file-name")
+    warn_undefined = pl.get_string_attrib(
```
mfbutner, line 101: This should be `get_boolean_attrib` so that that the string value gets correctly converted to `True`/`Fasle`

eliotwrobson: Good catch! I think I was on cruise control when I wrote this

mfbutner: Sorry I didn't have time to make it a proper pull request. I notice it just before I had to go to work.

mfbutner: Another thing I was thinking about this morning was whether the warnings should be on by default or not. The reason I was wondering this was because of how "defaults" are implemented in Mustache. You'd do something like
```html
Your lowest {{num_dropped_assignments}}{{^num_dropped_assignments}}3{{/num_dropped_assignments}} will be dropped in the course. 
```

This relies on the behavior of missing variables being expanded to nothing. I don't know if enough people will be doing this to warrant the warnings to be defaulted to off or not but I thought I should mention it.

mfbutner:
One option that we might want to think about adding to the `<pl-variable>` is the ability to ignore leading/trailing whitespace. I'm not sure if it should be on by default or not.

 As an example of why this might be useful I was trying to pass the path to a file through a variable as follows

```
<pl-variable name="programming_problem_prompt_path">
        if_statements_and_functions/quadratic_formula/prompt.html
</pl-variable>
```
because the line was getting long but then it couldn't find the file because it ended up being `\nif_statements_and_functions/quadratic_formula/prompt.html\n`

So I had to go back to writing it as 
```
<pl-variable name="programming_problem_prompt_path">if_statements_and_functions/quadratic_formula/prompt.html</pl-variable>
```
which is a pretty long line as you can see.

mfbutner:
@nwalters512 or @eliotwrobson I'm encountering some weird error interactions between `<pl-template>` and other `pl` elements.  As of right now I've found issues with
-  `<pl-multiple-choice>`: If it is in a template it does not render. It still seems to kind of exist though because on submission it 
   will raise an error about no true answer existing.
- `<pl-file-editor>`/`<pl-file-preview>`:  They will still both render if they are in a template but if `<pl-file-editor>` is in the template any files uploaded to it are "lost". They aren't saved when submitting.

I'm assuming that there are going to be other issues with other submission elements but I haven't tried it out yet.

I was wondering if either of you had an idea of what was happening? I did just pull the latest updates for my docker image of prairie learn this morning (1/29/2023) so it is up to date. I thought that the issue might be that `<pl-template>` uses a copy of the data dictionary but when I switched it to just reference the original it did not fix things.

Here is a an example for `<pl-multiple-choice>` that doesn't render for me.

question.html
```html
<pl-template file-name="simple_multiple_choice.html" parent-directory="question">
</pl-template>
```

simple_multiple_choice.html
```
<p>
    Is the number -3 positive, negative, or 0?
</p>
<pl-multiple-choice answers-name="acc" weight="1">
    <pl-answer correct="false">positive</pl-answer>
    <pl-answer correct="true">negative</pl-answer>
    <pl-answer correct="false">zero</pl-answer>
</pl-multiple-choice>
```

And here is an example of what I have with `<pl-file-editor>` not working correctly. 
question.html
```html
<p> Explain how your code works in English </p>
<pl-template file-name="simple_file_editor.html" parent-directory="question">
</pl-template>
```

simple_file_editor.html
```html
<pl-file-editor
        file-name="report.md"
        ace-theme="ace/theme/monokai"
        preview="markdown" ,
        normalize-to-ascii=true>
# My Algorithm (steps to solve the problem)
</pl-file-editor>

<pl-submission-panel>
    <pl-file-preview></pl-file-preview>
</pl-submission-panel>
```

elements/pl-template/pl-template.py
eliotwrobson: I was wondering about this. The above template doesn't necessarily rely on the behavior of missing variables, but will perform equivalently if the variables are present and set to false. However, this might create an extra burden on the developer (having to define useless variables), so maybe it's worth defaulting to false and just having it as an option that can be enabled for debugging typos and stuff.

eliotwrobson: @mfbutner in the last push, I change this option to be a boolean and also added an option to trim whitespace (defaulted to True).

As to the discussion below, I'll investigate later if I have time. I'm not quite sure why that would be happening (I'm not an expert on the renderer). Strange that the multiple choice question seems to run its backend code.

nwalters512:
@mfbutner I think I independently discovered that issue the other day as I was thinking through our rendering process. I'm pretty sure that addressing it will require us to pretty fundamentally rethink how this element works (or use a different approach for templating entirely).

The `render` phase is the only phase during which we're actually able to know about elements that are rendered by _other_ elements. During parsing, grading, and so on, there is no way we can possibly know that element A "contains" element B.

For a concrete example, if I have `<pl-template file-name="foo.mustache">` and `foo.mustache` contains a `<pl-number-input>`, there's no way that we can see into the template when parsing or grading.

To me, this feels like we need to implement templates outside of Python as part of the core code that processes `question.html` functions. The alternative is to always `render()` before we parse/grade/etc., but that would significantly increase (double, roughly) the time it takes to perform grades, which would be unacceptable.

I just remembered that this is actually something I've thought of in a separate context already: https://github.com/PrairieLearn/PrairieLearn/issues/1695

eliotwrobson:
@nwalters512 do you think that rendering issue is a dealbreaker for this element? I think the purpose of this is really to allow displaying different types of boilerplate text, not to generate entire questions. Would it be enough to discuss the limitations in the documents? I don't know if people are going to try to do very complex things.

nwalters512:
I don't know that it's necessarily a dealbreaker, but I do think that there's some shareable boilerplate that still involves other elements, for instance the very common `pl-file-upload`/`pl-file-preview`/`pl-external-grader-results` combo.

eliotwrobson:
@nwalters512 that's a good point, definitely would be nice to support rendering other elements through templates here in the future. All of the use cases in our course just involve basic HTML, so I think this element still has significant value with the current feature set. Though I think it's important the documentation and examples emphasize the current limitations of the element.

@mfbutner would you be willing to donate some examples (especially recursive ones) to the documentation of this element? I haven't written any docs yet and that would speed things up substantially for me if you have any laying around.

mfbutner:
@eliotwrobson I think I was able to submit a pull request to your repo with an example from my own course where I ask students to explain in English how they plan on solving a programming problem. Let me know if you have any questions about it and feel free to make any changes to it. This is also my first time making a pull request another repo other than my own so chances are good I messed something up >.<

I do think that until other elements are supported inside of templates we should actively check if there are elements other than pl-template and pl-variable inside of a template and error out if there is one. I suggest this because I think inserting elements into the template would be something that would happen fairly often. Even if it is explicitly stated not to do this in the documentation someone is going to miss it and the behavior if you do have them in is extremely odd.

eliotwrobson:
@mfbutner I think that's a great suggestion, even with discussion in the docs we should have some safeguards for this element. Could this be as easy as checking for any HTML tags starting with `<pl-*`? We could probably just use `lxml` to search through strings to do this.

mfbutner:
@eliotwrobson The only concern I have for just doing `<pl-*` is what if the course has custom elements? The better check would likely be to look for all nonstandard html elements aside from `pl-template` and `pl-variable` but that is going to be more painful to implement.

eliotwrobson:
@mfbutner yeah, I think that's going to be tricky. I also don't know if there's an easy way to have a definitive list of "standard" HTML elements, and I'm leaning towards the element being more lenient about what it can process with the disclaimer that it may not work for some situations.

mfbutner:
@eliotwrobson Your proposal sounds good to me.

 The only idea I would have if we did want to subject ourself to the pain of a standard list is to not only include the standard tags but also have an option on the template tag itself about what other tags are permissible.
```html
<pl-template ... allowable_tags="my_custom_element my_awesome_element">
</pl-template>
```

eliotwrobson:
@mfbutner I've updated this PR so the template element now displays a warning (which can be disabled) whenever the output contains another PL element that isn't part of a given whitelist. I think this is the best way to retain flexibility while still trying to make things easier for the end user.

eliotwrobson:
@mfbutner I've added documentation and the examples you put up to the latest push. I think this is in a state where we can start thinking about reviews + merging pretty soon. There might be a few more usability details to iron out, but I think those need input from maintainers (cc @mwest1066).

CODEOWNERS
```diff
@@ -4,6 +4,7 @@
 /elements/pl-external-grader-variables @nicknytko @eliotwrobson
 /elements/pl-graph @nicknytko @eliotwrobson
 /elements/pl-hidden-hints @eliotwrobson
+/elements/pl-template @eliotwrobson
```
nwalters512, line 7: Nit: alphabetical order please (so after the `/elements/pl-rich-text-editor` line).

docs/elements.md
```diff
@@ -1992,6 +1993,60 @@ The `pl-card` attributes mirror the options of [Bootstrap 4 cards](https://getbo
...
+    file-name="statement.html"
+  ></pl-variable>
+</pl-template>
+```
+
+    QUESTION = "question_path"
+    CLIENT_FILES_QUESTION = "client_files_question_path"
+    CLIENT_FILES_COURSE = "client_files_course_path"
+    SERVER_FILES_COURSE = "server_files_course_path"
+    COURSE_EXTENSIONS = "course_extensions_path"
```
nwalters512, line 2018: Is this supposed to be here?

exampleCourse/questions/element/template/question.html
```diff
@@ -0,0 +1,38 @@
...
+
+<hr>
+
+<pl-template file-name="cardTemplate.mustache" parent-directory="question" subdirectory="serverFilesQuestion">
+  <pl-variable name="header"><b>A bolded card header</b></pl-variable>
+</pl-template>
+
+<hr>
+
+<pl-template file-name="prompt_in_collapsed_section.mustache.html" parent-directory="server-files-course"
```
nwalters512, line 21: Nit: while we can't safely run every single HTML file through Prettier yet (thanks mostly to my stupid `<markdown>` tag), it'd be nice if we could format new files with Prettier. https://prettier.io/playground is a lightweight way to do that.

nwalters512 (commented):
I haven't fully reviewed yet, but here's a first round of comments.

docs/elements.md
```diff
@@ -1992,6 +1993,60 @@ The `pl-card` attributes mirror the options of [Bootstrap 4 cards](https://getbo
...
+| ------------------ | ------------------------------------------------------------------------------------------------------ | --------------------- | -------------------------------------------------------------------------------------------------------------- |
+| `file-name`        | string                                                                                                 | -                     | File name of the outer template to use.                                                                        |
+| `parent-directory` | `question`, `client-files-question`, `client-files-course`, `server-files-course`, `course-extensions` | `server-files-course` | Parent directory to locate `file-name`.                                                                        |
+| `subdirectory`     | string                                                                                                 | -                     | Subdirectory within parent directory to locate `file-name`.                                                    |
```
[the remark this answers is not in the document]
eliotwrobson: The only one I have possible questions about is camel casing; is there any desire to change this convention for consistency with other attributes later? If so, it's easy enough to bundle changes to those other elements in here or in another PR before this one.

docs/elements.md
eliotwrobson: Whoops 😅

exampleCourse/questions/element/template/question.html
eliotwrobson: I've actually been using the VSCode plugin to do HTML formatting. I think that `pl-code` is another element that has issues with autoformatting.

docs/elements.md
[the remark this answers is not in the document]
nwalters512: I'm not _opposed_ to changing them, but I don't think there's a need for that at the moment. I acknowledge that it's slightly idiosyncratic, but I'd argue that that's ok since it matches the actual names of the directories on disk.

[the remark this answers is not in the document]
eliotwrobson: Updated in https://github.com/PrairieLearn/PrairieLearn/pull/6747/commits/a0ef247641b9e185eb39ff7003b71d181e447609

elements/pl-template/pl-template.py
```diff
@@ -0,0 +1,144 @@
...
+
+            if inner_html and has_template_file:
+                raise ValueError(
+                    f'pl-variable "{name}" must have at most one of file-name or its inner html defined'
+                )
+
+            elif has_template_file:
+                variable_file_name = pl.get_string_attrib(child, "file-name")
+                # Default parent directory and subdirectory to what's defined in the outer template.
+                # TODO maybe change the defaults? Could be confusing, and may not save that much writing on the frontend
```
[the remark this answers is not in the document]
eliotwrobson: Sounds good, won't change the current behavior.

elements/pl-template/pl-template.py
```diff
@@ -0,0 +1,128 @@
...
+    element_list = cast(list, lxml.html.fragments_fromstring(element_html))
+
+    # First element can be a string, remove since there's nothing to check.
+    if isinstance(element_list[0], str):
+        element_list.pop(0)
+
+    for e in chain.from_iterable(element.iter() for element in element_list):
+        if (
+            e.tag is not lxml.etree.Comment
+            and e.tag.startswith("pl-")
```
eliotwrobson, line 29: Note to self: Add check for `<markdown>` tag.

docs/elements.md
```diff
@@ -1994,6 +1995,51 @@ The `pl-card` attributes mirror the options of [Bootstrap 4 cards](https://getbo
...
+| `file-name`      | string                                                                                          | -                   | File name of the outer template to use.                                                                      |
+| `directory`      | `question`, `clientFilesQuestion`, `clientFilesCourse`, `serverFilesCourse`, `courseExtensions` | `serverFilesCourse` | Parent directory to locate `file-name`.                                                                      |
+| `warn-undefined` | boolean                                                                                         | false               | Whether to print a warning when rendering templates with undefined variables. Useful for debugging.          |
+| `log-warnings`   | boolean                                                                                         | true                | Whether to log warnings if a rendered template contains elements which are not guaranteed to work correctly. |
```
eliotwrobson, line 2022: Note to self: Make these names consistent with other warning logging.

docs/elements.md
```diff
@@ -1994,6 +1995,51 @@ The `pl-card` attributes mirror the options of [Bootstrap 4 cards](https://getbo
...
+
+#### Sample element
+
+```html
+<pl-template file-name="outerTemplate.mustache" subdirectory="templates">
+  <pl-variable name="is-open">True</pl-variable>
+  <pl-variable
+    name="problem-statement"
+    directory="question"
+    file-name="serverFilesQuestion/statement.html"
```
SethPoulsen, line 2010: Looking at the docs below, this is confusing. Shouldn't the directory name `serverfilesQuestion` _not_ be in the `file-name`, since the directory should be specified in `directory`?

docs/elements.md
```diff
@@ -1994,6 +1995,51 @@ The `pl-card` attributes mirror the options of [Bootstrap 4 cards](https://getbo
...
+| `directory`      | `question`, `clientFilesQuestion`, `clientFilesCourse`, `serverFilesCourse`, `courseExtensions` | `serverFilesCourse` | Parent directory to locate `file-name`.                                                                      |
+| `warn-undefined` | boolean                                                                                         | false               | Whether to print a warning when rendering templates with undefined variables. Useful for debugging.          |
+| `log-warnings`   | boolean                                                                                         | true                | Whether to log warnings if a rendered template contains elements which are not guaranteed to work correctly. |
+
+Inside the `pl-template` element, variables for use in rendering the template may be specified with a `pl-variable` tag. Each `pl-variable` tag can be used to define a variable from a file or with the contents of the tag (but not both). Note that recursion is **not** applied to data in `pl-variable`. The `pl-variable` tag supports the following attributes:
```
SethPoulsen, line 2024: > recursion is **not** applied to data in `pl-variable`

What is this supposed to mean?

docs/elements.md
```diff
@@ -1994,6 +1995,51 @@ The `pl-card` attributes mirror the options of [Bootstrap 4 cards](https://getbo
...
+| `file-name`       | string                                                                                          | -                   | File name to use if variable data is being taken from a file. |
+| `directory`       | `question`, `clientFilesQuestion`, `clientFilesCourse`, `serverFilesCourse`, `courseExtensions` | `serverFilesCourse` | Parent directory to locate `file-name`.                       |
+| `trim-whitespace` | boolean                                                                                         | true                | Whether to trim whitespace of data specified by this tag.     |
+
+#### Details
+
+Because of the way that elements are rendered in PrairieLearn, templates should only contain other decorative elements. In particular, **this element will not work correctly with elements that accept and/or grade student input.** When rendering a template, all entries from `data["params"]` are included as available variables and may be used when the template is rendered. Templates may also be used recursively.
```
SethPoulsen, line 2035: Is the problem that _this_ element won't work correctly, or is the problem that the graded element that is inside of it won't work correctly?

I guess it doesn't matter in the sense that in either case, you just don't want to put graded elements inside of this one. But it seems like we might as well say whatever is most correct if we can.

exampleCourse/questions/element/template/pre_code_prompt.html
SethPoulsen, line 1: This is really long and detailed to be putting in the example course. 

It would be nicer to have a really short example that shows the features of the element with as little other distraction as possible.

exampleCourse/questions/element/template/quadratic_formula_prompt.html
SethPoulsen, line 1: This example also seems far longer than necessary.

exampleCourse/questions/element/template/question.html
```diff
@@ -0,0 +1,48 @@
+<p>
+  This page shows has example uses of the <code>pl-template</code>
+  element, including uses of recursive templates.
+</p>
```
SethPoulsen, line 4: To me a recursive template would seem to denote a template that imports _itself_. I don't see that happening here, but I do see templates importing _other_ templates. Is that right, or am I missing something?

Either way, I don't think "recursive" is the best thing to call this in documentation. I think it would be much more clear to just say "...including uses of templates that import other templates."

docs/elements.md
eliotwrobson: This just means that there's no variable substitution applied to the html provided through a `pl-variable` (that HTML is used as-is). I'll change the phrasing here to be clearer.

docs/elements.md
eliotwrobson: The overall output just looks really weird, but you're correct that it's more precise to say that the graded element won't work as expected. Will change.

exampleCourse/questions/element/template/pre_code_prompt.html
eliotwrobson: This was a donated example, I'll cut down some of the longer pieces of text. This element is tricky in that you need examples that are complicated enough to demonstrate intricate use cases, but don't have too much else to them.

exampleCourse/questions/element/template/question.html
eliotwrobson: That's correct, templates don't call themselves. I'll change the language to say "nested" templates.

SethPoulsen: I don't think that be obvious to all users what that is supposed to mean. Why can't you just say "templates that import/use other templates"?

exampleCourse/questions/element/template/pre_code_prompt.html
SethPoulsen: The question itself doesn't have to be interesting to demonstrate the use of the element. 

It could literally say something like "This is in example of how you use a template in a question..."

eliotwrobson: Simplified here: https://github.com/PrairieLearn/PrairieLearn/pull/6747/commits/6952a2bed7adb19fd8e09233d3afdcb6e7f9d14f

docs/elements.md
eliotwrobson: Changed here: https://github.com/PrairieLearn/PrairieLearn/pull/6747/commits/6698f6249a1b8802227ccb4e8ea79618901306c1

docs/elements.md
eliotwrobson: https://github.com/PrairieLearn/PrairieLearn/pull/6747/commits/6698f6249a1b8802227ccb4e8ea79618901306c1

docs/elements.md
eliotwrobson: Done here: https://github.com/PrairieLearn/PrairieLearn/pull/6747/commits/6698f6249a1b8802227ccb4e8ea79618901306c1

exampleCourse/questions/element/template/question.html
```diff
@@ -0,0 +1,48 @@
+<p>
+  This page shows has example uses of the <code>pl-template</code>
+  element, including uses of templates within other templates.
+</p>
+
+<hr />
+
+<pl-template file-name="templates/outerTemplate.mustache">
```
nwalters512, line 8: Let's be consistent with `camelCase` vs `snake_case` in template names.

exampleCourse/questions/element/template/pre_code_prompt.html
nwalters512, line 1: Should we establish a convention that templates should be located in `serverFilesCourse`? I don't understand why they're currently split between that and the root of the question directory.

exampleCourse/serverFilesCourse/templates/prompt_in_collapsed_section.mustache.html
```diff
@@ -0,0 +1,22 @@
+Here is some text at the top of the template.
```
nwalters512, line 1: Is this testing an edge case? I'd err on the side of removing this.

exampleCourse/serverFilesCourse/templates/outerTemplate.mustache
nwalters512, line 1: `outerTemplate` is not descriptive of what this template actually is.

exampleCourse/serverFilesCourse/templates/collapsible_section.mustache.html
```diff
@@ -0,0 +1,18 @@
...
+      class="card-title d-flex justify-content-between"
+      data-toggle="collapse"
+      data-target="#{{section_name}}"
+    >
+      <div>{{section_header}}</div>
+      <div class="fa fa-angle-down"></div>
+    </h4>
+  </div>
+
+  <div class="collapse{{#show}} show{{/show}}" id="{{section_name}}">
```
nwalters512, line 13: I get that you're just trying to illustrate using templates, but this feels like in the "real world" this would make much more sense as an element that can automatically assign an ID here. As written, this question has a very strange UX: as soon as someone submits an answer to this question, there will be two panels with identical IDs, which is a) not allowed and b) means that every instance of this panel will expand/collapse whenever any instance is clicked.

Is there a more realistic use case for templates that we could use instead?

exampleCourse/serverFilesCourse/templates/collapsible_section.mustache.html
```diff
@@ -0,0 +1,18 @@
+<div class="card mb-1 mt-1">
+  <div class="card-header" style="cursor: pointer">
+    <h4
+      class="card-title d-flex justify-content-between"
+      data-toggle="collapse"
+      data-target="#{{section_name}}"
+    >
+      <div>{{section_header}}</div>
+      <div class="fa fa-angle-down"></div>
+    </h4>
```
nwalters512, line 10: If we're going to stick with this example, I'd rather not use an `<h4>` here since this isn't underneath an `<h3>`. Plain text is fine.

nwalters512 (commented):
Most of my concerns are about the example question. We want our examples to feature high-quality code that one could use in their own questions, and I don't think we've reached that bar yet.

exampleCourse/questions/element/template/pre_code_prompt.html
eliotwrobson: There are some templates that are located in the question folder, but I did this just as a test. I'll go ahead and move all templates to server files course.

In the case of this file however, this is not actually a template. This is content that is pasted within a template, which is a reasonable use case (there's instances where the template is defining some formatting, but in each question, you want to populate that outer formatting with different contents).

exampleCourse/serverFilesCourse/templates/prompt_in_collapsed_section.mustache.html
eliotwrobson: This is testing an edge case for the `check_tags` function, because the list returned by `lxml.html.fragments_fromstring` is of mixed type (and unfortunately the type annotations for that library don't give great guarantees for that function).

exampleCourse/serverFilesCourse/templates/collapsible_section.mustache.html
eliotwrobson: So this actually does come from a real world example (donated by @mfbutner), but I think you're right that this would make a bit more sense as an element. The `pl-template` element lives in an unfortunate middle-ground where it's very useful for people that just want to have boilerplate text copied, but for those that don't want to write their own elements, it's not _quite_ flexible enough that it's easy to get yourself into trouble with stuff like this.

Thinking about it some, there's enough HTML like this that relies on having a unique id that it's worth having the `pl-template` always contain a `uuid` generated for use just by the contents of this template. I'll update this example and the documentation accordingly.

exampleCourse/serverFilesCourse/templates/collapsible_section.mustache.html
eliotwrobson: Switched this to a div (not 1000% sure this is correct syntax): https://github.com/PrairieLearn/PrairieLearn/pull/6747/commits/c34c7251e8702c9594174441044a5d108bf85466

exampleCourse/questions/element/template/question.html
eliotwrobson: Removed all camelCase in latest push

exampleCourse/serverFilesCourse/templates/collapsible_section.mustache.html
mfbutner: Sorry for being somewhat absent from this conversation. Regarding the example I submitted. 
1. I am very new to writing stuff for the front end so I probably made quite a few mistakes or did things a weird/wrong way.
2. I didn't notice the multiple id issue because I had always been wrapping my templates at the highest level in a `<pl-question-panel>`. So since it rendered only in the question panel there weren't multiple ids. I shouldn't be expecting everyone else to be using it  this same way so I'll probably have to update the example to make it more universal
3. I didn't consider writing my own custom element because up until reading your message I only considered writing an element if I needed some custom grading logic along with custom displaying. I forgot you could write elements that were decorative only even though there are already a ton of them. I noticed what eliot had made and it seemed like exactly what I needed: a way to easily copy and paste around the same block of text with some minor variations across multiple questions. It has been really helpful and helped to cut down on a lot of code bloat.

eliotwrobson: So in the latest pushes, I added the `uuid` generation for each separate `<pl-template>` and updated the examples, so the duplicate IDs should no longer be an issue @mfbutner.

To your third item: This is actually what I like most about the `pl-template` as a concept; instructors can essentially write their own decorative elements with this without the need to define a custom element each time, which cuts down heavily on code duplication.

elements/pl-template/pl-template.py
```diff
@@ -0,0 +1,136 @@
...
+                e.tag.startswith("pl-") and e.tag not in ALLOWED_PL_TAGS
+            ) or e.tag == "markdown"
+
+            if is_tag_invald:
+                warnings.warn(
+                    f'Element "{e.tag}" may not work correctly when used inside of "pl-template" element.'
+                )
+
+
+def get_file_path(data: pl.QuestionData, element: lxml.html.HtmlElement) -> str:
```
eliotwrobson, line 43: Note to self: switch order of these arguments.

eliotwrobson:
@nwalters512 TL;DR You raise some very good points, but resolving those issues requires some careful design and is out of scope for this element, and we include messages telling people not to do those things.

## The Long Version

I think it's perfectly reasonable to have a discussion about this, especially since most of the issues you point out are structural and not tied to the current implementation. As far as scope, the purpose of this element is really to prevent people from having to write a custom element every time they want to copy and paste some generic HTML / text for display, and not a full-on framework for question writing (i.e., including uses of graded elements). A lot of the limitations you describe are inherent to all custom elements, and I think that the solutions to some of these issues fall outside of the stated scope (but may be worth addressing).

> Revisiting this a few months later, I'm slightly concerned with the current implementation, primarily with the limitations and quirks it has:
> 
>     * `<markdown>` doesn't work.

I agree this is pretty annoying, but this is an issue people will encounter with any custom element that outputs HTML that includes `<markdown>`. The main difference is that, with this element vs a custom one, users at least get a warning that markdown won't work as expected as opposed to a silent failure. I really think the best solution is to scrap `<markdown>` and replace it with a `<pl-markdown>` element that does processing during the `render()` phase (or some other more generic solution, unless it's really slow or something).

>     * Using elements in templates sometimes doesn't work as expected.
>     * The hardcoded list of "allowed" elements inside templates is a maintenance burden (see recent conversation in Slack - we missed `pl-figure` as an allowed element) and doesn't support custom elements in courses.

As above, this is always the behavior with any element outputting HTML. There actually _is_ a way of handling these more involved use cases (statically generate the HTML during `generate()` and add it to `data["params"]`, then include this in `question.html` using standard mustache), but as stated above, handling this use case is really not the goal here.

> I'm wondering if perhaps we need another phase to complement `render()`, like `template()`. This can have slightly different semantics: it will run before all phases to support templates with elements that require generate/parse/grade, and it will run before `<markdown>` is processed. I think this would solve the current problems in a generic and flexible way.

Isn't this effectively an extension of the mustache substitution that already happens? I think something like this could be cool, but I'm not so sure that it really simplifies things for people. I'm not opposed to it, but I don't necessarily think adding more phases to the rendering (especially if questions will now have both dynamic + static rendering phases) simplifies the mental model for course content creation. This change would also mean that this element would have a harder time supporting any kind of dynamically changing content based on question code (at least, I think, if it's only generating things during the proposed `template()` phase).

After writing out my thoughts, I think these are all great points and it's absolutely worth considering how it will affect the question writing experience in general and with respect to this element. But I think the solution to a lot of the above issues is really to rethink and simplify some aspects of the phases and question writing process (especially if we want to allow for more generic methods of reusing parts of questions) instead of just tacking on carveouts that make this element work more intuitively in some cases.

The solution to the above problems in the existing implementation is basically to print a warning that says "What you're doing may not work as expected, proceed at your own risk", and very plainly state that this element is intended to solve one very simple use case in the documentation. I think this is ultimately the best solution, since attacking the other problems requires unpacking some fundamental aspects of PL and the question writing process, and the current implementation is already a significant convenience.

I think the discussion here is reminiscent of the `pl-hidden-hint` element, where the (simple) need the element is trying to fill rubs up against some sharp corners inherent to the way PL is currently built, and has the potential to confuse some people. But similar to that element, I think the _vast_ majority of people are going to use this element for the simple (intended) cases and won't rub up against these problems if we plainly state what this is for (I don't know if anyone has asked about some of the corner cases we were worried about for the hints element).

mwest1066 (approved):

PR #6750, opened 2022-12-07
Add tooling for compiling and serving client-side assets

nwalters512:
This PR lifts over our tooling for client-side assets from PrairieTest into a package that can be used by both PL and PT. It's designed to be as simple as possible and forgo the complexities of tools like Webpack. Tl;dr:

- All files located within the root of `assets/scripts` are transpiled and processed into bundles that can be served to clients.
- During development, these are always bundled on-the-fly at request time. [`esbuild`](https://esbuild.github.io/) is extremely fast, which makes this feasible.
- In production, the bundles are prebuild and written to `public/build/scripts`. A JSON manifest file in that directory maps from a bundle name (e.g. `foo.ts`) to the name of the compiled bundle (e.g. `foo-ABCD1234.js`).
- The functions `compiledScriptPath` and `compiledScriptTag` are used to obtain a path/tag for a given bundle, taking into account whether we're operating in dev or prod mode.


server.js
```diff
@@ -1944,6 +1948,14 @@ if (config.startServer) {
           callback(null);
         });
       },
+      async () => {
+        compiledAssets.init({
+          dev: config.devMode,
```
nwalters512, line 1953: The fact that the `prairielearn/prairielearn` Docker image runs in dev mode is unfortunate, as it means we won't be taking advantage of prebuilt assets. esbuild should be fast enough where this shouldn't be a big issue, but there's room for improvement here. A potential option is that we should produce two images, e.g. `prairielearn/prairielearn` and `prairielearn/prairielearn-dev`?

assets/scripts/test.js
```diff
@@ -0,0 +1 @@
+console.log('hello');
```
nwalters512, line 1: TODO: revert before merging.

pages/partials/head.ejs
```diff
@@ -6,6 +6,7 @@
 <link href="<%= node_modules_asset_path('bootstrap-icons/font/bootstrap-icons.css') %>" rel="stylesheet" />
 <link href="<%= asset_path('stylesheets/colors.css') %>" rel="stylesheet" />
 <link href="<%= asset_path('stylesheets/local.css') %>" rel="stylesheet" />
+<%- compiled_script_tag('test.js') %>
```
nwalters512, line 9: TODO: revert before merging.

server.js
mwest1066: I'm on board with a separate `-dev` image for people doing development. It would certainly be nice to keep `prairielearn/prairielearn` as fast as possible.

mwest1066 (approved):

server.js
nwalters512: Absolutely. This would also help when we deploy `prairielearn/executor`, which doesn't need to include many of the tools that a development-focused image would.

pages/partials/head.ejs
nwalters512: Removed in 9b9c438.

packages/compiled-assets/README.md
```diff
@@ -0,0 +1,82 @@
...
+Create a directory of assets that you wish to bundle, e.g. `assets/`. Within that directory, create another directory `scripts/`. Any JavaScript or TypeScript files in the root of the `scripts/` directory will become a bundle that can be loaded on a page. For example, the following directory structure would produce bundles named `foo` and `bar`:
+
+```
+├── assets/
+│   ├── scripts/
+│   │   ├── foo.ts
+│   │   └── bar.ts
+```
+
+You can locate shared code in directories inside this directory. As long as those files aren't in the root of the assets directory, they won't become separate bundles.
```
jonatanschroeder, line 20:
```suggestion
You can locate shared code in directories inside this directory. As long as those files aren't in the root of the assets/scripts directory, they won't become separate bundles.
```

packages/compiled-assets/README.md
```diff
@@ -0,0 +1,82 @@
...
+        <script src="${compiledScriptPath('bar.ts')}"></script>
+      </head>
+      </body>
+        Hello, world.
+      </body>
+    </html>
+  `;
+});
+```
+
```
jonatanschroeder, line 75: Given we're using EJS instead of TS in PrairieLearn would it be appropriate to have a PL-specific example?

PR #6751, opened 2022-12-07
Enable stricter settings in the TypeScript config for packages

nwalters512:


packages/tsconfig/tsconfig.json
```diff
@@ -7,5 +7,8 @@
     // We don't yet make extensive use of ES Modules, so compile
     // `import`/`export` down to CommonJS.
     "module": "CommonJS",
+    "strict": true,
```
mwest1066, line 10: 👍

mwest1066 (approved):

PR #6764, opened 2022-12-10
style: Miscellaneous type annotations and test cases

eliotwrobson:
Added some type hints and optimized inner html function (as the old implementation used an inefficient string join method). Won't flip to open until some other things are merged and test cases are added to make sure there's not backwards compatibility issues with the old implementation.

Partially handles #6771 for some core functions.

EDIT: Modified this and tested locally, should be good to go now.


stevenxpan (approved):

elements/pl-hidden-hints/pl-hidden-hints.py
```diff
@@ -15,22 +18,26 @@ def render(element_html: str, data: pl.QuestionData) -> str:
 
     # Use position so that hints appear in order if show-after-submission values are equal.
     for position, child in enumerate(element):
-        if child.tag == 'pl-hint':
-            pl.check_attribs(child, [], ['show-after-submission'])
+        if child.tag == "pl-hint":
```
nwalters512, line 21: We expect these diffs to go away once https://github.com/PrairieLearn/PrairieLearn/pull/6793 is merged, correct?

python/prairielearn.py
```diff
@@ -167,14 +168,19 @@ def from_json(v):
...
-        inner = ''
-    inner = html.escape(str(inner))
-    for child in element:
-        inner += lxml.html.tostring(child, method='html').decode('utf-8')
-    return inner
+def inner_html(element: lxml.html.HtmlElement) -> str:
+    """
+    Gets the inner HTML of an element. A bit ugly, but hacked together to be as fast as possible
+    (since this is a function that gets used in a ton of places). The performance increase from
+    the old version of this function could have a substantial impact on question render times.
```
nwalters512, line 175: What kind of benchmarking have you done to validate that this is indeed faster? I ran a test here with a small HTML document and it looks like the old implementation was actually faster on average:

```python
>>> import prairielearn
>>> import timeit
>>> import lxml.html
>>> element = lxml.html.fragment_fromstring('<div>test<p>hello</p>test2</div>')
>>> timeit.timeit(lambda: prairielearn.inner_html_new(element), number = 10000)
0.04604645899962634
>>> timeit.timeit(lambda: prairielearn.inner_html_new(element), number = 10000)
0.04699804197298363
>>> timeit.timeit(lambda: prairielearn.inner_html_new(element), number = 10000)
0.047380250005517155
>>> timeit.timeit(lambda: prairielearn.inner_html(element), number = 10000)
0.02854979201219976
>>> timeit.timeit(lambda: prairielearn.inner_html(element), number = 10000)
0.03979900002013892
>>> timeit.timeit(lambda: prairielearn.inner_html(element), number = 10000)
0.03487775003304705
```

elements/pl-hidden-hints/pl-hidden-hints.py
eliotwrobson: Yep! (modulo a merge patch)

python/prairielearn.py
eliotwrobson: Ran this with a large test case:

image

And a smaller one:
image


The perf impact is definitely on the subtle side (😅 ), this was more pronounced on my other machine (the one that ran these benchmarks is a very fast workstation). Looking at this again, the main benefit of this PR is really the uplifted test cases for this function and adding in the missing type hints. I think this can probably wait until some of the other element / utility PRs are in. That way there will be a few different files that use the new types before adding this one on top.

python/prairielearn.py
```diff
@@ -449,14 +452,20 @@ def from_json(v):
...
 
-def inner_html(element):
-    inner = element.text
-    if inner is None:
-        inner = ""
-    inner = html.escape(str(inner))
-    for child in element:
-        inner += lxml.html.tostring(child, method="html").decode("utf-8")
-    return inner
+def inner_html(element: lxml.html.HtmlElement) -> str:
```
eliotwrobson, line 455: Benchmarks of new vs old:

image

Not a huge difference but essentially free performance for rewriting a very short function.

elements/pl-graph/pl-graph.py
```diff
@@ -47,10 +47,14 @@ def graphviz_from_adj_matrix(
     )
     input_param_name = pl.get_string_attrib(element, "params-name", input_param_matrix)
 
+    # Exception to make typechecker happy.
+    if input_param_name is None:
+        raise ValueError('"params-name" is a required attribute.')
+
```
nwalters512, line 53: Is this still necessary? It looks like you defined overloads for the case where a default is provided, so I'd expect the type checker to know that `input_param_name` is not in fact `None` here.

elements/pl-hide-in-panel/pl-hide-in-panel.py
```diff
@@ -25,7 +25,6 @@ def render(element_html, data):
         or (data["panel"] == "submission" and not hide_in_submission)
         or (data["panel"] == "answer" and not hide_in_answer)
     ):
-        element = lxml.html.fragment_fromstring(element_html)
```
nwalters512, line 28: Nice catch!

elements/pl-variable-output/pl-variable-output.py
```diff
@@ -132,7 +132,7 @@ def render(element_html: str, data: pl.QuestionData) -> str:
             if not pl.has_attrib(child, "digits"):
                 var_digits = digits
             else:
-                var_digits = pl.get_string_attrib(child, "digits")
+                var_digits = pl.get_integer_attrib(child, "digits")
```
nwalters512, line 135: Nice! Was this genuinely broken, or did it happen to still work?

nwalters512 (approved):
If the type checker is happy, I'm happy. Just one question about whether or not an explicit `None` check is needed.

elements/pl-graph/pl-graph.py
eliotwrobson: Unfortunately, it does not. The reason is because this uses a passthrough, and the old default value of `None` is still getting used. I think this wasn't here originally because the `None` check was happening outside of this function.

nwalters512: Ah, ok, I see that now when looking at the broader context here.

PR #6767, opened 2022-12-13
Allow redisUrl to be null

nwalters512:
We've assigned a meaning to `redisUrl: null`:

https://github.com/PrairieLearn/PrairieLearn/blob/310ff716139f576b20280456a64b3b5fb34f5a93/lib/redis.js#L16

This PR ensures that the JSON schema for our config allows Redis to be disabled.

### DIFF
--- a/schemas/schemas/serverConfig.json
+++ b/schemas/schemas/serverConfig.json
@@ -34,7 +34,7 @@
         },
         "redisUrl": {
             "description": "The url for connecting to Redis.",
-            "type": "string"
+            "type": ["string", "null"]
         },
         "logFilename": {
             "description": "Filename to use for server logging.",


nwalters512:
Tangentially related: I'm thinking we should switch to using Zod to parse+validate our config 👀

mwest1066 (approved):

mwest1066:
> Tangentially related: I'm thinking we should switch to using Zod to parse+validate our config 👀

I'm generally in favor of zod, it's been great in PT. But I would like to retain the ability to have unknown properties in the config so we don't need to synchronize code deploys with config updates (either upgrading or downgrading versions).

nwalters512:
Agreed - don't limit to only known properties, but ensure that we're getting the values we think we are!

I files #6772 to track this.

PR #6773, opened 2022-12-14
Clarify docs for `showClosedAssessment` access restriction

mwest1066:
The `showClosedAssessment` and `showClosedAssessmentScore` access restrictions are easy to misuse because they only apply to specific access rules. Hopefully these docs changes help.

### DIFF
--- a/docs/accessControl.md
+++ b/docs/accessControl.md
@@ -164,9 +164,14 @@ To block students from viewing closed assessment details, set `"showClosedAssess
         "timeLimitMin": 50,
         "showClosedAssessment": false,
         "credit": 100
+    },
+    {
+        "showClosedAssessment": false
     }
 ]
 ```
+
+Note that the `showClosedAssessment` statement is listed twice above, once while the assessment is in progress (`startDate` to `endDate`) and once in a second rule that will take effect after the `endDate`.
 
 The `showClosedAssessment` access rule restriction is only really useful in conjuction with [time limits](#time-limits). It is common to pair `"showClosedAssessment": false` with [disabled real-time grading](assessment.md#disabling-real-time-grading).
 
@@ -185,9 +190,15 @@ To block students from viewing closed assessment scores, set `"showClosedAssessm
         "showClosedAssessment": false,
         "showClosedAssessmentScore": false,
         "credit": 100
+    },
+    {
+        "showClosedAssessment": false,
+        "showClosedAssessmentScore": false
     }
 ]
 ```
+
+Note that the `showClosedAssessment` and `showClosedAssessmentScore` statements are listed twice above, once while the assessment is in progress (`startDate` to `endDate`) and once in a second rule that will take effect after the `endDate`.
 
 The `showClosedAssessmentScore` access rule restriction is only useful in conjunction with [disabling real-time grading](assessment.md#disabling-real-time-grading) and setting `"showClosedAssessment": false`.
 


docs/accessControl.md
```diff
@@ -164,10 +164,15 @@ To block students from viewing closed assessment details, set `"showClosedAssess
...
         "showClosedAssessment": false,
         "credit": 100
+    },
+    {
+        "showClosedAssessment": false
     }
 ]
 ```
 
+Note that the `showClosedAssessment` statement is listed twice above, once while the assessment is in progress (`startDate` to `endDate`) and once in a second rule that will take effect after the `endDate`.
```
jonatanschroeder, line 174: May want to include a comment that new submissions are allowed after this point if this new rule is added, so if they don't want that they should set `"active": false` (maybe a link to the corresponding section).

docs/accessControl.md
```diff
@@ -185,10 +190,16 @@ To block students from viewing closed assessment scores, set `"showClosedAssessm
...
+    {
+        "showClosedAssessment": false,
+        "showClosedAssessmentScore": false
     }
 ]
 ```
 
+Note that the `showClosedAssessment` and `showClosedAssessmentScore` statements are listed twice above, once while the assessment is in progress (`startDate` to `endDate`) and once in a second rule that will take effect after the `endDate`.
+
 The `showClosedAssessmentScore` access rule restriction is only useful in conjunction with [disabling real-time grading](assessment.md#disabling-real-time-grading) and setting `"showClosedAssessment": false`.
```
jonatanschroeder, line 203: This is not your change, but this may not be 100% correct in this case, so piggybacking. Here's a suggestion:
```suggestion
The `showClosedAssessmentScore` access rule restriction is often used in conjunction with [disabling real-time grading](assessment.md#disabling-real-time-grading) and setting `"showClosedAssessment": false`. It may also be used in assessments with [manual grading](manualGrading.md), to hide the assessment grade until all students have been graded.
```

docs/accessControl.md
mwest1066: But won't the assessments be closed by this point, so that they won't even be viewable by the student?

I can certainly add `"active": false` to the final rule just to be on the safe side, if you think it would help?

jonatanschroeder: > But won't the assessments be closed by this point, so that they won't even be viewable by the student?

Not if they didn't start.

docs/remoteExams.md
```diff
@@ -98,14 +100,15 @@ Some notes about this configuration:
...
+- This configuration sets `"showClosedAssessment": false` to prevent students from seeing the details of their exam after it is over. This can help to mitigate cheating with students taking conflict exams. The final access rule containing only `"showClosedAssessment": false` is necessary because the earlier rules will only apply up until their `endDate` times.
 
 ## Asynchronous, timed exams
 
-We do **NOT** recommend exams to be run using an asynchronous, timed configuration. While giving exams asynchronously will simplify exam administration and provide students with more flexibility, it comes at the expense of making it easier to cheat. We recommend [synchronous, timed exams](#synchronous-timed-exams).
+We do **NOT** recommend exams to be run using this configuration (asynchronous with time limit). While giving exams asynchronously will simplify exam administration and provide students with more flexibility, it comes at the expense of making it easier to cheat. We recommend [synchronous, timed exams](#synchronous-timed-exams).
```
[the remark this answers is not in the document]
mwest1066: Updated the recommendations in 86b32dd4c2cabfc072b5b4351cdb304a66de1328 (also updated the no-realtime-grading recommendation).

docs/accessControl.md
mwest1066: Good point. Added in 90d0092d3d1cab14c60e1ebc60380658562587a9

jonatanschroeder (approved):

Issue #6774, opened 2022-12-15
`pl-dropdown` semantics and possible deprecation 


eliotwrobson:
Mainly a discussion item for now, but might be worth pursuing. Currently, the `pl-dropdown` and `pl-multiple-choice` elements seem to have identical semantics in the sense that they only allow the student to input one correct answer from a given ordered list. However, they are separate elements with different semantics for the developer (identified by a distinct tag and don't have all the same options, for example https://github.com/PrairieLearn/PrairieLearn/issues/2668).

Instead of trying to make sure they behave close to identically, wouldn't it be better to deprecate `pl-dropdown` and make using a dropdown a display option for `pl-multiple-choice`? That way, the missing options from `pl-dropdown` will all get added along with reducing the amount of code that needs to be maintained on the backend. This might also make things easier for course content developers, since they can try different display options without having to change as much code or semantics. Thoughts?

nwalters512:
This makes a lot of sense to me. This could be as easy as `display="dropdown"`.

mwest1066:
Yep, I'm on board.

firasm:
I'd love this! it would simplify my scripts a lot too 😁

jonatanschroeder:
> That's a very good point. Getting random HTML to display within a `<select>` is pretty much impossible without reimplementing the dropdown in JS; not even Bootstrap styles will fix that. Agreed that we shouldn't try to support that; we can just document it.

I was thinking replacing the select with a hidden input controlled by this: https://getbootstrap.com/docs/4.0/components/dropdowns/

We've had instructors with issues showing formulas in dropdown, this would fix that too.

eliotwrobson:
Bootstrap dropdowns seem like a nice way to push things forward. Then we could even allow for dividers and other logic.

nwalters512:
Ah, good call! We'll have to be careful to ensure that it behaves like a `<select>` from an accessibility standpoint. The docs make it seem like most of that is already taken care of; we may just have to add an appropriate `aria-label`.

jonatanschroeder:
Related: #3145

eliotwrobson:
Sounds like this refactor is the direction things are going. I'll bundle this with the duplicate checking and an uplifting of the backend logic. I think I'll be able to get started next week.

eliotwrobson:
[...]

Issue #5369, opened 2022-02-08
One standard way to specify order


ffund:
There are currently three ways that I know of to specify order, across different elements:

* `fixed-order` for pl-matching, pl-multiple-choice, pl-checkbox (can be `true` or `false`)
* `sort` for pl-dropdown (can be `fixed`, `random`, `ascend`, or `descend`)
* `source-blocks-order` for pl-order-blocks (can be `random` or `ordered`)

It would be cool if I only had to remember one way! and if they all supported the same attribute values.

Related: #4920

Issue #3145, opened 2020-09-16
Dropdown element does not support latex


tdy:
`pl-dropdown` currently uses html `<select>`/`<option>`, which is not supported by mathjax.

One possible workaround:

> @nicknytko 2:43 PM
> we could rework the dropdown to use the bootstrap one that is a bunch of `<a>`s and a `<div>`

See **https://getbootstrap.com/docs/4.5/components/dropdowns/**

```html
<div class="dropdown">
  <button class="btn btn-secondary dropdown-toggle" type="button" id="dropdownMenuButton" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
    Dropdown button
  </button>
  <div class="dropdown-menu" aria-labelledby="dropdownMenuButton">
    <a class="dropdown-item" href="#">Action</a>
    <a class="dropdown-item" href="#">Another action</a>
    <a class="dropdown-item" href="#">Something else here</a>
  </div>
</div>
```

andrewstec:
CSS is insufficient in the area of styling Select Dropdown Options. Where there is limited success in styling, it cannot be guaranteed by cross-browser support.

I do not know much about Mathjax, but what are the requirements for Mathjax to work with the element? What does it have to accomplish?

Cinda has experienced a similar issue recently involving simple styling of the font. The lack of CSS support means on the Select Dropdown Options means that the font will not change to something that is friendlier for programming languages.

nicknytko:
> CSS is insufficient in the area of styling Select Dropdown Options. Where there is limited success in styling, it cannot be guaranteed by cross-browser support.
> 
> I do not know much about Mathjax, but what are the requirements for Mathjax to work with the element? What does it have to accomplish?
> 
> Cinda has experienced a similar issue recently involving simple styling of the font. The lack of CSS support means that the font will not change to something that is friendlier for programming languages.

MathJax will insert rendered math as SVG elements on the page, so I think we need some HTML element that will render children.  I think the `<option>` will only display plain text (https://html.spec.whatwg.org/multipage/form-elements.html#the-option-element), which is why I was suggesting we check out the bootstrap dropdowns as those create a fake "dropdown" out of a bunch of `div` tags.

andrewstec:
I think that your `DIV` and `A` approach sounds perfect then.

Surg-Dev:
Is there anyone else looking into this?

[...]

Issue #2846, opened 2020-08-06
Add option to disable vertical alignment on checkbox and multiple choice elements


nicknytko:
Perhaps a `vertical-align-options`, default True

eliotwrobson:
Would this be similar to the display option of having `inline=True`? If not, and if this functionality does not exist yet, it would be easy to bundle in https://github.com/PrairieLearn/PrairieLearn/pull/6775, as that PR is already reworking a significant part of the frontend for the multiple choice element.

PR #6775, opened 2022-12-16
`pl-multiple-choice` dropdown refactor

eliotwrobson:
Currently in draft state. Addresses the following issues with `pl-multiple-choice`:

- Resolves https://github.com/PrairieLearn/PrairieLearn/issues/1228
- Resolves https://github.com/PrairieLearn/PrairieLearn/issues/6774
- Resolves https://github.com/PrairieLearn/PrairieLearn/issues/3145
- Resolves https://github.com/PrairieLearn/PrairieLearn/issues/2846
- Partially addresses https://github.com/PrairieLearn/PrairieLearn/issues/5369 by consolidating pl-dropdown and pl-multiple-choice sorting options.
- Add hide score badge, per https://github.com/PrairieLearn/PrairieLearn/issues/878

Planned: https://github.com/PrairieLearn/PrairieLearn/issues/4391

Also, refactor uses utility functions from https://github.com/PrairieLearn/PrairieLearn/pull/6641 and somewhat addresses the following issue for the element, as using an enum makes sense rather than true or false: https://github.com/PrairieLearn/PrairieLearn/issues/1671

EDIT: Screenshots of current implementation, open to change colors / style:

With blue answer choice selected:
image

After submitting:
image

Another submission without feedback:
image

With the options selector open:
image

### DIFF
--- a/docs/elements.md
+++ b/docs/elements.md
@@ -105,7 +105,10 @@ compatibility, but they should not be used in new questions.
 ### `pl-multiple-choice` element
 
 A `pl-multiple-choice` element selects **one** correct answer and zero or more
-incorrect answers and displays them in a random order as radio buttons.
+incorrect answers and displays them in a random order as radio buttons or a dropdown.
+Duplicate answer choices (string equivalents) are not permitted in the
+`pl-multiple-choice` element, and an exception will be raised upon question
+generation if two (or more) choices are identical.
 
 #### Sample element
 
@@ -136,6 +139,7 @@ incorrect answers and displays them in a random order as radio buttons.
 | `external-json`               | string  | special | Optional path to a JSON file to load external answer choices from. Answer choices are stored as lists under "correct" and "incorrect" key names.                     |
 | `external-json-correct-key`   | string  | special | Optionally override default json "correct" attribute name when using `external-json` file.                                                                           |
 | `external-json-incorrect-key` | string  | special | Optionally override default json "incorrect" attribute name when using `external-json` file.                                                                         |
+| `allow-blank`                 | boolean | false   | Whether or not an empty submission is allowed. If `allow-blank` is set to `true`, no response will be marked as incorrect instead of invalid.                        |
 
 The attributes `none-of-the-above` and `all-of-the-above` can be set to one of these values:
 
--- a/apps/prairielearn/elements/pl-multiple-choice/info.json
+++ b/apps/prairielearn/elements/pl-multiple-choice/info.json
@@ -1,8 +1,8 @@
 {
-    "controller": "pl-multiple-choice.py",
-    "dependencies": {
-        "elementStyles": [
-            "pl-multiple-choice.css"
-        ]
-    }
+  "controller": "pl-multiple-choice.py",
+  "dependencies": {
+    "elementStyles": ["pl-multiple-choice.css"],
+    "nodeModulesStyles": ["bootstrap-select/dist/css/bootstrap-select.min.css"],
+    "nodeModulesScripts": ["bootstrap-select/dist/js/bootstrap-select.min.js"]
+  }
 }
--- a/apps/prairielearn/elements/pl-multiple-choice/pl-multiple-choice.mustache
+++ b/apps/prairielearn/elements/pl-multiple-choice/pl-multiple-choice.mustache
@@ -1,41 +1,41 @@
 {{#question}}
 
+{{#radio}}
 {{#inline}}
 <span>
 {{/inline}}
-{{#answers}}
-    <div class="form-check {{#inline}}form-check-inline d-inline-flex{{/inline}}{{^inline}}d-flex{{/inline}} align-items-center py-1">
-        <input class="form-check-input mt-0" type="radio"
-               name="{{name}}" value="{{key}}" {{^editable}}disabled{{/editable}}
-               {{#checked}}checked{{/checked}} id="{{name}}-{{key}}">
+    {{#answers}}
+        <div class="form-check {{#inline}}form-check-inline d-inline-flex{{/inline}}{{^inline}}d-flex{{/inline}} align-items-center py-1">
+            <input class="form-check-input mt-0" type="radio"
+                name="{{name}}" value="{{key}}" {{^editable}}disabled{{/editable}}
+                {{#selected}}checked{{/selected}} id="{{name}}-{{key}}">
 
-        <label class="form-check-label d-flex align-items-center" for="{{name}}-{{key}}">
-            {{^hide_letter_keys}}<div class="pl-multiple-choice-key-label">({{key}})</div>{{/hide_letter_keys}}
-            <div class="ml-1 mr-1">{{{html}}}</div>
-        </label>
+            <label class="form-check-label d-flex align-items-center" for="{{name}}-{{key}}">
+                {{^hide_letter_keys}}<div class="pl-multiple-choice-key-label">({{key}})</div>{{/hide_letter_keys}}
+                <div class="ml-1 mr-1">{{{html}}}</div>
+            </label>
 
-        {{#display_feedback}}
-        <div class="pl-multiple-choice-feedback">
-            {{{feedback}}}
+            {{#display_feedback}}
+            <div class="pl-multiple-choice-feedback">
+                {{{feedback}}}
+            </div>
+            {{/display_feedback}}
+
+            {{#display_score_badge}}
+                {{#correct}}
+                    <span class="badge badge-success"><i class="fa fa-check" aria-hidden="true"></i></span>
+                {{/correct}}
+                {{#incorrect}}
+                    <span class="badge badge-danger"><i class="fa fa-times" aria-hidden="true"></i></span>
+                {{/incorrect}}
+                {{#partial}}
+                    <span class="badge badge-warning"><i class="far fa-circle" aria-hidden="true"></i></span>
+                {{/partial}}
+            {{/display_score_badge}}
         </div>
-        {{/display_feedback}}
-
-        {{#display_score_badge}}
-            {{#correct}}
-                <span class="badge badge-success"><i class="fa fa-check" aria-hidden="true"></i></span>
-            {{/correct}}
-            {{#incorrect}}
-                <span class="badge badge-danger"><i class="fa fa-times" aria-hidden="true"></i></span>
-            {{/incorrect}}
-            {{#partial}}
-                <span class="badge badge-warning"><i class="far fa-circle" aria-hidden="true"></i></span>
-            {{/partial}}
-        {{/display_score_badge}}
-    </div>
-{{/answers}}
+    {{/answers}}
 
 {{#display_score_badge}}
-    &nbsp;
     {{#correct}}
         <span class="badge badge-success"><i class="fas fa-check" aria-hidden="true"></i> 100%</span>
     {{/correct}}
@@ -50,6 +50,52 @@
 {{#inline}}
 </span>
 {{/inline}}
+{{/radio}}
+{{^radio}}
+<!-- Weirdly enough putting this in a div causes it to display without the inline -->
+<span class="py-1 d-inline-flex align-items-center">
+    <script>
+        $(() => {
+            $('#pl-multiple-choice-select-{{uuid}}').selectpicker({})
+                .on('rendered.bs.select loaded.bs.select shown.bs.select', (e) => {
+                    // If any of the options has math, render the math as needed
+                    // Events: loaded (when the page opens), rendered (when the value changes),
+                    // shown (when the dropdown extends)
+                    MathJax.startup.promise.then(async () => MathJax.typesetPromise($(e.target).parent()));
+                });
+        });
+    </script>
+    <select name="{{name}}" data-style="btn-outline-secondary"
+            id="pl-multiple-choice-select-{{uuid}}" {{^editable}}disabled{{/editable}}>
+        <option disabled selected value></option>
+        <!-- TODO disable blank option to use if blank is expected -->
+        {{#answers}}
+            <option {{#selected}}selected{{/selected}}
+                value="{{key}}"
+                data-content="{{^hide_letter_keys}}({{key}}){{/hide_letter_keys}} {{{html}}}"></option>
+        {{/answers}}
+    </select>
+
+    {{#feedback}}
+    <div class="pl-multiple-choice-feedback">
+        {{{feedback}}}
+    </div>
+    {{/feedback}}
+
+    {{#display_score_badge}}
+        &nbsp;
+        {{#correct}}
+            <span class="badge badge-success"><i class="fas fa-check" aria-hidden="true"></i> 100%</span>
+        {{/correct}}
+        {{#incorrect}}
+            <span class="badge badge-danger"><i class="fas fa-times" aria-hidden="true"></i> 0%</span>
+        {{/incorrect}}
+        {{#partial}}
+            <span class="badge badge-warning"><i class="far fa-circle" aria-hidden="true"></i> {{partial}}%</span>
+        {{/partial}}
+    {{/display_score_badge}}
+</span>
+{{/radio}}
 
 {{/question}}
 {{#submission}}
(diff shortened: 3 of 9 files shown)


eliotwrobson:
One thing I haven't figured out how to do effectively with this is make the width of the button for always be the same as that of the corresponding dropdown menu. Still some styling work to go, but the core submission logic works.

elements/pl-multiple-choice/pl-multiple-choice.mustache
```diff
@@ -3,36 +3,81 @@
...
+            });
+            $(this).addClass("active");
+
+            // Finally, set input value
+            $("#pl-dropdown-input-{{uuid}}").val($(this).attr("value"));
+        });
+    });
+</script>
+<div class="dropdown py-1 d-inline-flex">
+    <button class="btn btn-secondary dropdown-toggle" type="button" id="pl-multiple-choice-dropdown-button-{{uuid}}" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
```
eliotwrobson, line 64: @nwalters512 do you know if there's a way to force this button to have the same width as the corresponding dropdown menu? I tried setting the widths to be equal using some jQuery but it wound up looking weird. Not sure if this is something weird related to the way this works in bootstrap.

docs/elements.md
```diff
@@ -133,6 +136,7 @@ incorrect answers and displays them in a random order as radio buttons.
 | `external-json`               | string  | special | Optional path to a JSON file to load external answer choices from. Answer choices are stored as lists under "correct" and "incorrect" key names.            |
 | `external-json-correct-key`   | string  | special | Optionally override default json "correct" attribute name when using `external-json` file.                                                                  |
 | `external-json-incorrect-key` | string  | special | Optionally override default json "incorrect" attribute name when using `external-json` file.                                                                |
+| `allow-blank`                 | boolean | false   | Whether or not an empty submission is allowed. If `allow-blank` is set to `true`, no response will be marked as incorrect instead of invalid.               |
```
SethPoulsen, line 139: What is the use case for this? I'm mostly just curious, I don't see why I would ever allow a student to click submit without answering the question.

eliotwrobson: The issue https://github.com/PrairieLearn/PrairieLearn/issues/6350 is for adding `allow-blank` in every element. I actually had some issues trying to include this functionality and wound up getting rid of it, but forgot to remove it from the docs.

eliotwrobson:
> For clarity: this only _partially_ resolves #5369, right? So it should remain open after this is merged.

Correct

SethPoulsen (commented):
I would love to see, as part of this PR, almost all of the `pl-dropdown` code get deleted, and simply be replaced by generating the equivalent `pl-multiple-choice` code. e.g. If someone's question has:

```html
<pl-dropdown answers-name="x">
    <pl-answer correct="true">1</pl-answer>
    <pl-answer correct="false">0</pl-answer>
  </pl-dropdown>
```

then the `render` function of `pl-dropdown` should output:

```html
<pl-multiple-choice display="dropdown" answers-name="x">
    <pl-answer correct="true">1</pl-answer>
    <pl-answer correct="false">0</pl-answer>
  </pl-multiple-choice>
```

Then `pl-multiple-choice` will do all the heavy lifting, and we have the code reuse that we are looking for.

This sort of thing works automagically because PL uses a pre-order when it traverses the element tree to render elements. See also, how we insert `pl-code` inside of `pl-order-blocks` when `format="code"`.

SethPoulsen:
I just remembered this may not work if people are still using the old question renderer in their course. @nwalters512  do we now how many courses are using the legacy renderer and if they have dropdown questions? I know this was a problem for CS 105 when we rolled out the new order-blocks code formatting, but it was really easy to get them switched to the new question renderer.

eliotwrobson:
> I would love to see, as part of this PR, almost all of the `pl-dropdown` code get deleted, and simply be replaced by generating the equivalent `pl-multiple-choice` code.

I would personally love this too, but I think we could run into incompatibility issues because of different ways of storing data internally in these elements.

SethPoulsen:
> we could run into incompatibility issues because of different ways of storing data internally in these elements.
ugh, true. I guess we should do it as a separate PR between semesters or something.

elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -1,33 +1,69 @@
...
+
+
+class AotaNotaType(Enum):
+    FALSE = 1
+    RANDOM = 2
+    INCORRECT = 3
+    CORRECT = 4
+
+
+class SortType(Enum):
```
eliotwrobson, line 28: There has been some discussion about which types of sorting should be supported. We should decide on that before this gets merged.

jonatanschroeder:
I see a commit adding the `hide-score-badge` attribute. I just want to make sure that there is an understanding on what this attempts to hide. To the best of my knowledge (I haven't tested it, though), in `pl-checkbox`, this option hides the badges beside each item, but not the badge for the whole element. I believe that the functionality of this option is valid, but it doesn't address #878 directly. In a multiple choice question where a single item is correct this difference is less relevant, but it's probably a good idea to keep a consistent name for these attributes.

eliotwrobson:
@jonatanschroeder Yes, in `pl-checkbox` the `hide-score-badge` only hides the badges next to each time, whereas in the latest commit, this attribute hides the badge _and_ the partial score percentage (I don't think people need the granularity of being able to hide just one or the other since only one answer can be selected). This probably isn't the best name for this attribute though, maybe something like `hide-partial-score`?

elements/pl-multiple-choice/pl-multiple-choice.mustache
```diff
@@ -3,36 +3,81 @@
...
+                $(this).removeClass("active");
+            });
+            $(this).addClass("active");
+
+            // Finally, set input value
+            $("#pl-dropdown-input-{{uuid}}").val($(this).attr("value"));
+        });
+    });
+</script>
+<div class="dropdown py-1 d-inline-flex">
```
jonatanschroeder, line 63:
```suggestion
<div class="dropdown py-1 d-inline-flex" role="listbox">
```
I'm not an expert on accessibility, but I believe this is necessary for screen readers. See: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role#examples. Also add `role="option"` to the options.

elements/pl-multiple-choice/pl-multiple-choice.mustache
```diff
@@ -3,36 +3,81 @@
...
+            // Next, change active choice to be what the user just selected
+            $(".pl-dropdown-choice-{{uuid}}").each(function(){
+                $(this).removeClass("active");
+            });
+            $(this).addClass("active");
+
+            // Finally, set input value
+            $("#pl-dropdown-input-{{uuid}}").val($(this).attr("value"));
+        });
+    });
```
jonatanschroeder, line 61: I wonder if it may be a good idea to move this code to a separate js file. You can add a reference to it from the json file.

elements/pl-multiple-choice/pl-multiple-choice.mustache
```diff
@@ -3,36 +3,81 @@
...
-        {{/display_score_badge}}
+    {{/answers}}
+{{/radio}}
+{{^radio}}
+<script>
+    $(function(){
+        $(document).ready(function(){
+            $(".pl-dropdown-choice-{{uuid}}.active").each(function(){
+                $("#pl-multiple-choice-dropdown-button-{{uuid}}").text($(this).text());
+            });
```
jonatanschroeder, line 45: Any reason why this is done with Javascript instead of mustache?

elements/pl-multiple-choice/pl-multiple-choice.mustache
```diff
@@ -3,36 +3,81 @@
...
+</script>
+<div class="dropdown py-1 d-inline-flex">
+    <button class="btn btn-secondary dropdown-toggle" type="button" id="pl-multiple-choice-dropdown-button-{{uuid}}" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+    </button>
+    <input type="hidden" name="{{name}}" id="pl-dropdown-input-{{uuid}}" {{^editable}}disabled{{/editable}}>
+    <div class="dropdown-menu" id="pl-multiple-choice-dropdown-menu-{{uuid}}" aria-labelledby="dropdownMenuButton">
+        {{#answers}}
+            <li><a class="dropdown-item pl-dropdown-choice-{{uuid}} {{#checked}}active{{/checked}}" value="{{key}}" {{^editable}}disabled{{/editable}}>
+                <div>{{^hide_letter_keys}}({{key}}){{/hide_letter_keys}} {{{html}}}</div>
+            </a></li>
```
jonatanschroeder, line 71: Any reason why you're using `li` elements here?

docs/elements.md
```diff
@@ -102,6 +102,9 @@ compatibility, but they should not be used in new questions.
 
 A `pl-multiple-choice` element selects **one** correct answer and zero or more
 incorrect answers and displays them in a random order as radio buttons.
+Duplicate answer choices (string equivalents) are not permitted in the
+`pl-multiple-choice` element, and an exception will be raised upon question
+generation if two (or more) choices are identical.
```
jonatanschroeder, line 107: This may be a problem for dynamically generated options that happen to be identical. Can they be silently ignored?

docs/elements.md
jonatanschroeder: This is mostly for questions with several elements, including a `pl-multiple-choice`, where a student chooses to deliberately keep it blank (e.g., out of time) but still wants partial grades for previous elements. If one element gives an invalid parse result, the question is fully invalid.

elements/pl-multiple-choice/pl-multiple-choice.py
```diff
@@ -65,107 +111,118 @@ def categorize_options(element, data):
...
-        string_value = pl.get_string_attrib(element, name, default)
-    if string_value not in ["false", "random", "incorrect", "correct"]:
-        raise Exception(
-            f'pl-multiple-choice element has invalid value for {name} attribute, must be "false", "true", "random", "correct" or "incorrect"'
+        boolean_value = pl.get_boolean_attrib(
+            element, name, default != AotaNotaType.FALSE
         )
-    return string_value
+        return AotaNotaType.RANDOM if boolean_value else AotaNotaType.FALSE
+        # TODO add specific attrib not found exception and catch it here
```
jonatanschroeder, line 154: Just making sure this is handled.

jonatanschroeder (commented):
Just a few comments, not a formal review.

docs/elements.md
eliotwrobson: This is actually the intended behavior change (see https://github.com/PrairieLearn/PrairieLearn/issues/1228). We _could_ add a flag that enables silently ignoring identical options (defaulted to false), but the purpose of this is to loudly fail. I don't know why an instructor would want to allow identical options. Perhaps if there was a way to raise a warning that got displayed to the instructor only and not students?

elements/pl-multiple-choice/pl-multiple-choice.mustache
eliotwrobson: Whoops, was copying and pasting too quickly.

elements/pl-multiple-choice/pl-multiple-choice.mustache
eliotwrobson: If it gets much longer I probably will move it.

elements/pl-multiple-choice/pl-multiple-choice.mustache
eliotwrobson: For some reason I thought this had to be done with Javascript, but I think you're right that it can be done with mustache (see [here](https://stackoverflow.com/questions/33726202/how-to-use-bootstrap-dropdown-as-an-html-form-select)).

elements/pl-multiple-choice/pl-multiple-choice.mustache
eliotwrobson: @jonatanschroeder this is what I was having issues with. The width of the dropdown keeps changing based on the text inside (instead of remaining a constant width like the HTML `select` element does).

docs/elements.md
jonatanschroeder: I'm thinking something like:

```
<pl-answer>{{params.a}}</pl-answer>
<pl-answer>{{params.b}}</pl-answer>
...
```

Where `a` and `b` may randomly end up having the same value. I'd imagine they'd be deduplicated (i.e., only one would appear). I indeed see the discussion in the original issue, and will mark this conversation as resolved, if this comes up again we can have a new issue. It probably makes sense to an instructor that wants to create these random options to have some type of handling of duplicate options.

elements/pl-multiple-choice/pl-multiple-choice.mustache
jonatanschroeder: I was checking out some options and indeed it doesn't seem like there is an easy option here. That said, I think if you provide some `min-width` (maybe use `10rem`, which is [the default for bootstrap menu items](https://github.com/twbs/bootstrap/blob/v4.6.2/dist/css/bootstrap.css#L3241)) this may not be a big deal.

jonatanschroeder: Check this option instead: https://github.com/snapappointments/bootstrap-select/

eliotwrobson: @jonatanschroeder I started trying to integrate the bootstrap select in this commit: https://github.com/PrairieLearn/PrairieLearn/pull/6775/commits/145ec1bda04e2bc32f76ebd8cfcfe3ef3dd6400a

I thought it seemed really nice, but was having an issue where the LaTeX wouldn't render for individual options. That's really my only reservation to using this, really would be nice not to have to implement that functionality ourselves.

jonatanschroeder: Did you try disabling sanitizer?

jonatanschroeder: I'm playing around with a few options, I'll push a new change if I find something useful. I'll also add the yarn package.

jonatanschroeder: Latex math issue resolved in d2fadff32.

elements/pl-multiple-choice/pl-multiple-choice.mustache
jonatanschroeder: I have removed it from the latest version, as `select` already implicitly has this role.

jonatanschroeder:
@eliotwrobson I have done a few improvements on the support for selectpicker, including support for MathJax, and I've done a bit of cleanup there. I'll leave it to you to test it further, including in cases like inline dropdowns and similar issues. I've also not handled the case where the dropdown starts out blank (i.e., one of the options is blank on purpose). Let me know if you'd like another review.

eliotwrobson:
@jonatanschroeder thank you! I should be able to take it from here, this was really the only thing I got hung up on while I was working on this. This is one of the lower priority items for me now, but will let you know if things ramp up again.

eliotwrobson:
@jonatanschroeder do you know if there's a way to make LaTeX render once the given item has been selected (see attached screenshot)? I don't think it's a dealbreaker but would be nice to have that working if possible. It seems to work as expected when the selected answer has latex when the page initially loads.

Otherwise, the rest of the testing I've been doing has been working really well!

image

jonatanschroeder:
> @jonatanschroeder do you know if there's a way to make LaTeX render once the given item has been selected (see attached screenshot)? I don't think it's a dealbreaker but would be nice to have that working if possible. It seems to work as expected when the selected answer has latex when the page initially loads.

Seems like the problem is that `typesetPromise` is being called asynchronously, but the content of the elements is retrieved as soon as sanitize returns (which in this case may be before the typeset promise completes), so the HTML render is retrieved too early. I'm looking into alternatives. The easiest solution would be to run the MathJax typeset promise again when the change event happens, but I'm trying to find a lighter solution that wouldn't require a repeat of the process.

jonatanschroeder:
Just pushed a change that takes care of the MathJax issue.

eliotwrobson:
Going to close this for now and repoen later once I have more time to test / develop code here. In the interest of resolving some issues, I'll port over some of the shorter, more useful features (like duplicate checking and allow blank functionality) into a separate PR that we'll hopefully be able to merge before the fall semester starts.

PR #6779, opened 2022-12-19
Fix keyword capitalization in sync_assessments.sql

SethPoulsen:


nwalters512 (approved):

SethPoulsen:
makes me think we ought to add a SQL linter/prettifier too

nwalters512:
> makes me think we ought to add a SQL linter/prettifier too

We've talked about this before! I'm open to it if we can find one that works well for our needs.

PR #6780, opened 2022-12-19
Add configurable site-wide announcement

nwalters512:
This PR adds the ability to configure a site-wide announcement that will appear at the top of every page. This is primarily meant to be used to announce maintenance windows, including the upcoming downtime in January 2023.

Screenshot 2022-12-19 at 13 40 57

This also removes the old "advertisement" code in favor of this new configurable announcement.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -325,6 +325,17 @@ config.titleOverride = null;
 config.homepageFooterText = null;
 config.homepageFooterTextHref = null;
 
+/**
+ * HTML that will be displayed in a banner at the top of every page. Useful for
+ * announcing maintenance windows, etc.
+ */
+config.announcementText = null;
+/**
+ * A Bootstrap color (`primary`, `secondary`, `warning`, etc.) for the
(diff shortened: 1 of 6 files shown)


eliotwrobson:
Might this be a place to (temporarily) post about news items? I actually really like this feature, since there aren't many ways to broadcast information to PL users outside of the slack announcements (which I know some people don't follow very closely).

nwalters512:
I think there could be separate work to increase discoverability of news items, but I think a site-wide, non-dismissible banner is pretty intrusive for that kind of thing. Feel free to spin up another issue or discussion post if you want to discuss that further!

jonatanschroeder:
Would it make sense to have a configurable end date/time for the announcement?

nwalters512:
@jonatanschroeder at the moment, we'll be doing redeploys with new configs, so I don't think it's necessary for this specific use case. It would be relatively straightforward to change this into an array of announcements with start/end times in the future if we find that's something we need.

Issue #2099, opened 2020-02-28
pl.to_json Object of type ndarray is not JSON serializable


trombonekenny:
@tbretl Did something change since creating the https://github.com/PrairieLearn/pl-ece470/tree/master/questions/fromWedge question (see it live at https://prairielearn.engr.illinois.edu/pl/course/112/question/1244932/preview)? 

Trying to load it now gives an error:

`TypeError: Object of type 'ndarray' is not JSON serializable`

My guess is something about `pl.to_json()` used to work to serialize the data but now it doesn't (maybe with nested ndarrays)? Any insights?

eliotwrobson:
It looks like this is no longer an issue? Can probably close this issue.

Issue #1366, opened 2018-11-07
convert numpy integer types to/from json properly


tbretl:
If I do this in `server.py`:
`data['params']['a'] = np.int64(5)`

Then I see this error:
TypeError: Object of type 'int64' is not JSON serializable

This is because `json.dumps` can't handle numpy integer types. Either create a custom serializer or modify `pl.to_json` and `pl.from_json` to handle these types.

nicknytko:
Is it really important to keep the distinction that it is a numpy type here?  It wouldn't be difficult to add another case to the `pl.to_json()` and `pl.from_json()` methods but I'm wondering if it would be okay to just discretely convert to the base types?  My thoughts are that most people would just want to store _an_ integer or _a_ floating point number and using the conversion methods is an extra annoyance, but I might be totally wrong here.

eliotwrobson:
To do the conversion mentioned in the original issue, all that's needed is to convert the numpy integer to a python `int` and then back. For more complex conversions (i.e. json serializing an entire array), there's a dependency that can be added: https://github.com/mverleg/pyjson_tricks

PR #6783, opened 2022-12-20
feat: JSON conversion improvements

eliotwrobson:
Improvements to json conversion that have been requested. Based on changes from #6744. Addresses the following:

- Resolves https://github.com/PrairieLearn/PrairieLearn/issues/1366
- Resolves https://github.com/PrairieLearn/PrairieLearn/issues/2099
- Resolves https://github.com/PrairieLearn/PrairieLearn/issues/6965

EDIT: Dependency on other PRs has been removed, should be good to review!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ max-line-length = 88
 extend-ignore = ["E203", "W503"]
 
 [tool.black]
-exclude = '^\/tools\/.*\.py$|^\/exampleCourse\/.*\.py$|^\/testCourse\/.*\.py$|^\/node_modules\/.*\.pyi?$'
+exclude = '^\/tools\/|^\/exampleCourse\/|^\/testCourse\/|\/node_modules\/$'
 
 [tool.isort]
 profile = "black"


coatless:
Eh, I'm a bit worried that we're now getting into the territory of a bunch of merge conflicts as #6744 changes have crept into this PR. :/

eliotwrobson:
@coatless this is why this is in draft, I want to wait for other things to get merged before this is even reviewed seriously 😅

python/prairielearn.py
```diff
@@ -125,23 +159,33 @@ def from_json(v):
                 if ('_value' in v) and ('real' in v['_value']) and ('imag' in v['_value']):
                     return complex(v['_value']['real'], v['_value']['imag'])
                 else:
-                    raise Exception('variable of type complex should have value with real and imaginary pair')
+                    raise ValueError('variable of type complex should have value with real and imaginary pair')
```
coatless, line 162: I don't think there is a way to incorporate a more specific error message here related to the problematic variable. e.g. 

```python
ValueError(f'Variable {variable_name} with type complex should have both with real and imaginary pair values inside of value.')
```

`variable_name` I think has to be passed in as there doesn't seem to be a good option to extract the information via `inspect`.

eliotwrobson: Yeah, this is kindof a persistent issue unfortunately. Debugging these json serialization issues is always a pain.

Regarding your other comment, since it seems like we might go in another direction with the dataframes, I'll remove those changes from this PR and flip it to open (should get merged sooner that way).

python/prairielearn.py
```diff
@@ -759,7 +777,7 @@ def string_fraction_to_number(a_sub, allow_fractions=True, allow_complex=True):
             if not np.isfinite(a_sub_parsed):
                 raise ValueError('The submitted answer is not a finite number.')
             value = a_sub_parsed
-            data['submitted_answers'] = to_json(value)
+            data['submitted_answers'] = value
```
coatless, line 780: Why was the `to_json()` call removed here? Is `value` already in JSON format?

eliotwrobson: It's actually the opposite. This element was saving numpy floats natively because this to_json call wasn't doing anything. So adding the new encoding to this function actually broke some tests that were using custom grader code.

python/prairielearn.py
```diff
@@ -75,6 +77,10 @@ def to_json(v):
     """
     if np.isscalar(v) and np.iscomplexobj(v):
```
coatless, line 78: Let's add a quick variable name extractor: 

```suggestion
    # Retrieve the name of the `v` variable passed in by _inspecting_ the frame. 
    # Note, this approach does _not_ scale to multiple frame calls. Hence, we're opting not to functionalize it.
    variable_name = [k for k,x in inspect.currentframe().f_back.f_locals.items() if x is v][0]
    if np.isscalar(v) and np.iscomplexobj(v):
```

python/prairielearn.py
```diff
@@ -75,6 +77,10 @@ def to_json(v):
...
         return {'_type': 'complex', '_value': {'real': v.real, 'imag': v.imag}}
+    elif np.issubdtype(type(v), np.integer) and type(v).__module__ == 'numpy':
+        return {'_type': 'np_integer', '_concrete_type': type(v).__name__, '_value': int(v)}
```
nwalters512, line 81: Are we doing something sensible if a number can't be represented as a 32-bit integer? IIRC Numpy ints are 64-bit but JSON can't represent a full 64-bit int.

python/prairielearn.py
nwalters512: Given this, it looks like the docs for this function are now out of date:

> On successful parsing, "data" will contain a 'submitted_answers' key that is the JSON encoded parsed answer.

python/prairielearn.py
eliotwrobson: Good catch! We can just always use a string to represent the integer value. I'll also add a test for this.

python/prairielearn.py
coatless: I thought about this a bit over break. I've surmised my thoughts here: 

https://github.com/PrairieLearn/PrairieLearn/issues/6883

Maybe we could have this as a _separate_ PR to follow once the updated JSON conversions are included.

python/prairielearn.py
eliotwrobson: Actually, looking at this function I think that the input string can be a complex number as well. I might try and see if there's a way to avoid changing this function. A little weird that sometimes it saves the submitted answer as a JSON dict with more detailed information and sometimes just an integer. Maybe the best way is to just update the custom grader tests?

eliotwrobson: @coatless after thinking about this some, do you think we should have the numpy encodings have an opt-out option (similar to the way the dataframe element is set up), to avoid issues with breaking peoples custom grader code relying on the way that numpy integers are stored? That way, the option can be applied to the `string_fraction_to_number` function, which will avoid breaking people's custom grader code.

python/prairielearn.py
```diff
@@ -117,6 +122,26 @@ def to_json(v):
     If v can be json serialized or does not have a standard type, then it is
     returned without change.
     """
+    if new_np_scalar_encoding and type(v).__module__ == "numpy":
```
coatless, line 125: Numbers are better than bools for new versions. 

```suggestion
    if np_encoding == 2 and type(v).__module__ == "numpy":
```

python/prairielearn.py
```diff
@@ -96,13 +96,19 @@ def all_partial_scores_correct(data: QuestionData) -> bool:
     )
 
 
-def to_json(v):
+def to_json(v, *, np_encoding=1):
```
nwalters512, line 99: What's the migration path to the new encoding? Will we update all elements and example questions to use it? How do we communicate this change to courses? Is there a plan for deprecating and removing the legacy encoding entirely someday?

Is it strictly necessary to put this behind a flag? AFAICT the only case where this would cause issues is if someone was using `to_json` to store data but not using `from_json` to retrieve it, which feels like a question bug that we shouldn't feel the need to support.

If we decide we definitely need to flag this, the name `np_encoding` feels weird to me. I'd be inclined to make this either `np_encoding_version=1` or `legacy_np_encoding=True`.

eliotwrobson: Just to clarify, in a previous comment, I mistakenly said that numpy scalars were being auto-encoded as their corresponding Python types. No idea why I observed this particular behavior for a while, but PL just throws an error in this case as was noted in the original issue.

> What's the migration path to the new encoding? Will we update all elements and example questions to use it?

This encoding type is mainly for use by instructors writing custom questions (this seems to be the use case mentioned in the original issue #1366), so I don't think that there really needs to be a migration for core PL code. In particular, using this for the `pl-number-input` would change the way that some internal data is stored, which in turn would break custom grader code (including custom grader code used in internal test questions). So I think the best way to introduce this is by having it be opt-in for those that want it. 

> How do we communicate this change to courses?

Since this change is opt-in, I think we can communicate it through a news item of some type. There has been some discussion about a news item for the recent Python changes, and I'd be happy to whip up a combined one once we merge in some more of these instructor-facing changes.

> Is there a plan for deprecating and removing the legacy encoding entirely someday?

So the current legacy encoding is to do nothing (not sure why I thought that numpy types were being auto-encoded as their Python counterparts, testing again reveals that it just raises an error, as in the original issue), so I don't think there's much of a point to deprecating this. Alternatively, we could just make this particular encoding a different function altogether, but I think that could cause more confusion than it's worth.

> Is it strictly necessary to put this behind a flag? AFAICT the only case where this would cause issues is if someone was using `to_json` to store data but not using `from_json` to retrieve it, which feels like a question bug that we shouldn't feel the need to support.

As discussed above, there are cases where people (including core-PL code) run `to_json` and expect it to do nothing (don't know why these were the semantics that were decided upon, but that's how it works), and so the flag prevents breaking this in those cases.

> If we decide we definitely need to flag this, the name `np_encoding` feels weird to me. I'd be inclined to make this either `np_encoding_version=1` or `legacy_np_encoding=True`.

I think I'm leaning towards `np_encoding_version`, as @coatless has a preference against using a boolean flag.

eliotwrobson: Ok so I figured it out. Weirdly, the `float64` numpy type is json serializable (but get its type erased when decoding), and the other types are not. See [here](https://stackoverflow.com/questions/44459168/why-are-some-numpy-datatypes-json-serializable-and-others-not) for details. Either way, this means that the new conversion is technically backwards incompatible for some cases, so I think should be behind a flag (since some code relies on the above default behavior, including code in the number input element).

python/prairielearn_test.py
```diff
@@ -88,26 +89,82 @@ def test_set_score_data(
...
+def test_legacy_serialization(object_to_encode: Any, expected_result: Any) -> None:
+    """Test that nothing happens under the old encoding for numpy scalars."""
+
+    json_object = json.dumps(pl.to_json(object_to_encode), allow_nan=False)
+    decoded_json_object = pl.from_json(json.loads(json_object))
+
+    assert decoded_json_object == expected_result
+
+
+class DummyEnum(Enum):
```
eliotwrobson, line 139: Had to change this because pytest was trying to pick this up as a test class before, whoops.

docs/question.md
```diff
@@ -219,6 +219,29 @@ def grade(data):
...
+def generate(data):
+    data["params"]["numpy_array"] = pl.to_json(np.array([1.2, 3.5, 5.1]))
+
+def grade(data):
+    pl.from_json(data["params"]["numpy_array"])
+```
+
+In particular, in the `to_json` function, there are keyword-only options for different types of encodings. Encoding a DataFrame `df` by setting `to_json(df, df_encoding_version=2)` allows for missing and date time values whereas `pl.to_json(df, df_encoding_version=1)` (default) does not. However, `df_encoding_version=1` has support for complex numbers, while `df_encoding_version=2` does not.
+
+Similarly, if encoding numpy scalars by setting `np_encoding_version=1`, then only `np.float64` and `np.complex128` can be serialized by `to_json`, and their types will be erased after deserialization (will beccome native Python `float` and `complex` respectively). It is recommended to set `np_encoding_version=2`, which supports serialization for all numpy scalars and does not result in type erasure on deserialization.
```
nwalters512, line 243: Minor wordsmithing:

```suggestion
The `to_json` function supports keyword-only options for different types of encodings (e.g. `to_json(var, df_encoding_version=2)`). These options have been added to allow for new encoding behavior while still retaining backwards compatibility with existing usage.

- `df_encoding_version` controls the encoding of Pandas DataFrames. Encoding a DataFrame `df` by setting `to_json(df, df_encoding_version=2)` allows for missing and date time values whereas `pl.to_json(df, df_encoding_version=1)` (default) does not. However, `df_encoding_version=1` has support for complex numbers, while `df_encoding_version=2` does not.

- `np_encoding_version` controls the encoding of Numpy values. When using `np_encoding_version=1`, then only `np.float64` and `np.complex128` can be serialized by `to_json`, and their types will be erased after deserialization (will become native Python `float` and `complex` respectively). It is recommended to set `np_encoding_version=2`, which supports serialization for all numpy scalars and does not result in type erasure on deserialization.
```

docs/question.md
```diff
@@ -219,6 +219,29 @@ def grade(data):
         data["feedback"]["y"] = "Your value for $y$ is larger than $x$, but incorrect."
 ```
 
+## Question Data Storage
+
+All persistent related to a question is stored under different entries in the `data` dictionary. This dictionary is stored in JSON format by PrairieLearn, and as a result, everything in `data` must be JSON serializable. Some types in Python are natively JSON serializable, such as strings, lists, and dicts, while others are not, such as complex numbers, numpy ndarrays, and pandas DataFrames.
```
nwalters512, line 224:
```suggestion
All persistent data related to a question variant is stored under different entries in the `data` dictionary. This dictionary is stored in JSON format by PrairieLearn, and as a result, everything in `data` must be JSON serializable. Some types in Python are natively JSON serializable, such as strings, lists, and dicts, while others are not, such as complex numbers, numpy ndarrays, and pandas DataFrames.
```

nwalters512 (approved):

docs/question.md
```diff
@@ -219,6 +219,31 @@ def grade(data):
         data["feedback"]["y"] = "Your value for $y$ is larger than $x$, but incorrect."
 ```
 
+## Question Data Storage
+
+All persistent data related to a question variant is stored under different entries in the `data` dictionary. This dictionary is stored in JSON format by PrairieLearn, and as a result, everything in `data` must be JSON serializable. Some types in Python are natively JSON serializable, such as strings, lists, and dicts, while others are not, such as complex numbers, numpy ndarrays, and pandas DataFrames.
+
+To account for this, the `prairielearn` Python library from [`prairielearn.py`](https://github.com/PrairieLearn/PrairieLearn/blob/master/python/prairielearn.py) provides the functions `to_json` and `from_json`, which can respectively serialize and deserialize various objects for storage as part of question data. Please refer to the docstrings on those functions for more information. Here is a simple example:
```
coatless, line 226:
```suggestion
To account for this, the `prairielearn` Python library from [`prairielearn.py`](https://github.com/PrairieLearn/PrairieLearn/blob/master/python/prairielearn.py), usually aliased and used as `pl`, provides the functions `to_json` and `from_json`, which can respectively serialize and deserialize various objects for storage as part of question data. Please refer to the docstrings on those functions for more information. Here is a simple example:
```

docs/question.md
```diff
@@ -219,6 +219,31 @@ def grade(data):
...
+import prairielearn as pl
+
+def generate(data):
+    data["params"]["numpy_array"] = pl.to_json(np.array([1.2, 3.5, 5.1]))
+
+def grade(data):
+    pl.from_json(data["params"]["numpy_array"])
+```
+
+The `to_json` function supports keyword-only options for different types of encodings (e.g. `to_json(var, df_encoding_version=2)`). These options have been added to allow for new encoding behavior while still retaining backwards compatibility with existing usage.
```
coatless, line 241:
```suggestion
The `pl.to_json` function supports keyword-only options for different types of encodings (e.g. `pl.to_json(var, df_encoding_version=2)`). These options have been added to allow for new encoding behavior while still retaining backwards compatibility with existing usage.
```

docs/question.md
```diff
@@ -219,6 +219,31 @@ def grade(data):
...
+
+- `df_encoding_version` controls the encoding of Pandas DataFrames. Encoding a DataFrame `df` by setting `to_json(df, df_encoding_version=2)` allows for missing and date time values whereas `pl.to_json(df, df_encoding_version=1)` (default) does not. However, `df_encoding_version=1` has support for complex numbers, while `df_encoding_version=2` does not.
```
coatless, line 243:
```suggestion
- `df_encoding_version` controls the encoding of Pandas DataFrames. Encoding a DataFrame `df` by setting `pl.to_json(df, df_encoding_version=2)` allows for missing and date time values whereas `pl.to_json(df, df_encoding_version=1)` (default) does not. However, `df_encoding_version=1` has support for complex numbers, while `df_encoding_version=2` does not.
```

docs/question.md
```diff
@@ -219,6 +219,31 @@ def grade(data):
...
+
+- `np_encoding_version` controls the encoding of Numpy values. When using `np_encoding_version=1`, then only `np.float64` and `np.complex128` can be serialized by `to_json`, and their types will be erased after deserialization (will become native Python `float` and `complex` respectively). It is recommended to set `np_encoding_version=2`, which supports serialization for all numpy scalars and does not result in type erasure on deserialization.
```
coatless, line 245:
```suggestion
- `np_encoding_version` controls the encoding of Numpy values. When using `np_encoding_version=1`, then only `np.float64` and `np.complex128` can be serialized by `pl.to_json`, and their types will be erased after deserialization (will become native Python `float` and `complex` respectively). It is recommended to set `np_encoding_version=2`, which supports serialization for all numpy scalars and does not result in type erasure on deserialization.
```

Issue #712, opened 2017-07-27
Add multiple-choice suffixes to numeric input elements


mwest1066:
Going beyond #711, it would be nice to be able to specify a list of possible suffixes and a correct option, so the student would have to pick the correct suffix. For example:

```text
<number_input label="$v_x=$" suffixes="$m$,$m/s$,$m/s^2$" correct-suffix="$m/s$">
```

The list of choices could be rendered with a dropdown:

```text
$v_x=$ [INPUT] [Select \/]
```

where the `[Select \/]` element is a dropdown with the list of possible suffixes.

eliotwrobson:
Isn't this covered by the introduction of the dropdown element? The units use case will be covered by the introduction of the units element, but I think this original issue can be closed.

PR #6790, opened 2022-12-22
Introduce pl-units-input element

eliotwrobson:
Introduces the units element, a refresh of #4801 with cleaner backend. Still rough around the edges (needs more extensive testing), but core functionality is mostly working. In response to #6784. Will update more detailed tests  and documentation once behavior is finalized. Also resolves https://github.com/PrairieLearn/PrairieLearn/issues/712 by satisfying the use case outlined there.

Screenshots:
image
image
image


elements/pl-units-input/unit_utils.py
```diff
@@ -0,0 +1,130 @@
...
+            )
+
+        magnitudes_match = pl.is_correct_scalar_ra(
+            parsed_sub_base_unit.magnitude,
+            correct_ans_base_unit.magnitude,
+            rtol,
+            parsed_atol.magnitude,
+        )
+
+        if not magnitudes_match:
```
stevenxpan, line 121:
```suggestion
        if magnitudes_match:
            ...
```

stevenxpan: (and then switch the two return statements)

SethPoulsen:
From my point of view, the first priority with this is for someone understand what the pluses and minuses are of existing Units implementations and why we should use one over another (among the code in this PR, along with the multiple existing python packages). After we get that sorted out, we can move on to actual code review.

eliotwrobson:
> From my point of view, the first priority with this is for someone understand what the pluses and minuses are of existing Units implementations and why we should use one over another (among the code in this PR, along with the multiple existing python packages). After we get that sorted out, we can move on to actual code review.

I actually spent some time looking through different unit packages before working on this (as this dependency was replacing the included units code in the old PR). The main benefits of Pint over others is that this was the only package I found that included a parser (that doesn't call eval or any other unsafe functions, as far as I can tell) for parsing strings of numerical quantities with units. This substantially simplifies the code needed on the backend for reading student input. Among standalone packages, Pint also seems to be the most popular (see [here](https://socialcompare.com/en/comparison/python-units-quantities-packages)) and most well maintained.

The only other real contender would be sympy's units module (since this wouldn't require adding another dependency), but this doesn't include the aforementioned parser and has more convoluted syntax for many operations that get used in grading (converting to a unit of a standard type, for example).

I'm open to changing the package used on the backend but I think we would need a strong reason to switch from Pint to something else.

SethPoulsen:
Thanks, that's exactly the kind of info I wanted to know.

> more convoluted syntax for many operations that get used in grading 

does this leak into the student user interface? or just for people writing questions? It does seem nice to just use something sympy already has if possible, but if it is t very good then I guess we just add Pint.

eliotwrobson:
> does this leak into the student user interface? or just for people writing questions? It does seem nice to just use something sympy already has if possible, but if it is t very good then I guess we just add Pint.

Yes, from what I'm reading in the sympy documentation, you have to submit quantities with multiplication (so `4 * m` instead of `4m`), and this would need to be run through an eval call to make this work, so it would involve writing a ton of additional code to sanitize the input. Some units are also missing (I don't think they have temperature). Avoiding adding Pint as a dependency would basically boil down to having to rewrite a bunch of the feature ourselves.

elements/pl-units-input/unit_utils.py
coatless: @stevenxpan could you do a multi-line suggest comment? 

c.f. https://stackoverflow.com/a/69044609/1345455

python/prairielearn.py
```diff
@@ -52,6 +54,99 @@ class QuestionData(TypedDict):
 class ElementTestData(QuestionData):
     test_type: Literal['correct', 'incorrect', 'invalid']
 
+def grade_question_parameterized(
```
coatless, line 57: @eliotwrobson I think this is leaking from another PR?

eliotwrobson: This is intentional. This is a shared function that elements can use to cut down on boilerplate, and this is needed to make this element work. It'll disappear from the diff once it gets into master at least once.

eliotwrobson:
Per discussion, added a popover to the submission panel showing what the student input was parsed from:

image

Not sure if it's worth adding one to the correct answer panel (since there's no reason to display the raw correct answer in addition to the parsed one).

elements/pl-units-input/pl-units-input.py
```diff
@@ -0,0 +1,462 @@
...
+        else:
+            data["format_errors"][
+                name
+            ] = "Invalid format. The submitted answer was left blank."
+            data["submitted_answers"][name] = None
+
+        return
+
+    # TODO double check that doing this is ok. May need to switch to directory local to question / server
+    ureg = UnitRegistry(cache_folder=":auto:")
```
eliotwrobson, line 324: @nwalters512 I think these cache files are the last real item that might be of concern in this PR (other pending changes are minor cosmetic). There's documentation about what's going on in this cache folder here: https://pint.readthedocs.io/en/develop/advanced/performance.html#speed-up-registry-instantiation

Should this be made local to the assessment? Not sure if using the default across courses could create potential issues, or whether old cache files need to be cleaned out. Maybe there's a way to store them in the DB local to the question? Without setting up a cache like this, this element is really really slow.

nwalters512: This should specifically be in `/tmp/`, as that's the only directory guaranteed to be writable in prod. There's precedent for this with the `matplotlib` cache:

https://github.com/PrairieLearn/PrairieLearn/blob/4b07deac3d386a1843a1b2423df07be1c0dd9562/python/zygote.py#L30-L38

If this is truly really slow to load, it should be preloaded in the zygote that we fork from:

https://github.com/PrairieLearn/PrairieLearn/blob/4b07deac3d386a1843a1b2423df07be1c0dd9562/python/zygote.py#L46-L59

eliotwrobson: Gotcha. Using the a directory in `/tmp/` should be fine. The issue isn't that the module is slow to load (though it's probably also worth preloading in `zygote.py`), the issue is that the `UnitRegistry` object is expensive to construct, so using a cached version pickles the object for subsequent constructions. But if there's a folder where that caching can happen, we shouldn't run into any issues.

docs/elements.md
```diff
@@ -721,6 +723,52 @@ def generate(data):
...
+| `answers-name`             | string                                       | &mdash;      | Variable name to store data in.                                                                                                                                                                                                                                                         |
+| `weight`                   | integer                                      | 1            | Weight to use when computing a weighted average score over elements.                                                                                                                                                                                                                    |
+| `correct-answer`           | string                                       | special      | Correct answer for grading. Defaults to `data["correct_answers"][answers-name]`.                                                                                                                                                                                                        |
```
SethPoulsen, line 744: I think you are missing quotes around "answers-name" in the description field here.

docs/elements.md
```diff
@@ -721,6 +723,52 @@ def generate(data):
...
+| `suffix`                   | text                                         | &mdash;      | A suffix to display after the input box (e.g., `suffix="$\rm m/s^2$"`).                                                                                                                                                                                                                 |
+| `display`                  | "block" or "inline"                          | "inline"     | How to display the input field.                                                                                                                                                                                                                                                         |
+| `grading-mode`             | "with-units", "exact-units", or "only-units" | "with-units" | How to grade student submission. "only-units" only checks for the units input by the student. "exact-units" asks for a quantity with a specified unit. "with-units" is similar to "units-fixed", but will automatically convert the units used by the given answer if possible.         |
```
SethPoulsen, line 748: What is "units-fixed"? It is referenced in the description but not given as an option.

docs/elements.md
```diff
@@ -721,6 +723,52 @@ def generate(data):
...
+| `comparison`               | "exact", "sigfig", "decdig" or "relabs"      | "sigfig"     | How to grade. "relabs" uses relative ("rtol") and absolute ("atol") tolerances. "sigfig" use "digits" significant digits. "decdig" uses "digits" after decimal place, "exact" uses `==` and should only be used for integers. Attribute can only be set if `grading-mode=exact-units`   |
```
SethPoulsen, line 749: > relabs" uses relative ("rtol") and absolute ("atol") tolerances

I have no idea what this means. I don't know if the explanation needs to be rewritten or if I am just missing some context.

docs/elements.md
```diff
@@ -721,6 +723,52 @@ def generate(data):
...
+| `placeholder`              | string                                       | -            | String to override default placeholder text. The default placeholder gives information about the comparison type used.                                                                                                                                                                  |
+| `magnitude-partial-credit` | float                                        | -            | Fraction of partial credit given to answers of correct magnitude and incorrect units when `grading-mode=exact-units`. Remaining fraction of credit given when units are correct but magnitude is incorrect. Must be between 0.0 and 1.0. Partial credit is disabled if this is not set. |
+| `show-feedback`            | boolean                                      | true         | Whether to show feedback for incorrect answers.                                                                                                                                                                                                                                         |
```
SethPoulsen, line 759: What is this controlling exactly? If this option has the exact same effect of using `<pl-hide-in-panel>`, we should remove this option and just have people use that.

elements/pl-units-input/pl-units-input.mustache
```diff
@@ -0,0 +1,188 @@
...
+        style="margin-left: 5px" title="Format Error" tabindex="0" data-content="{{parse_error}}">
+        Why
+        <i class="fa fa-question-circle" aria-hidden="true"></i>
+    </a>
+    {{/parse_error}}
+
+    {{#missing_input}}
+    {{#label}}<span>{{{label}}}</span>{{/label}}
+    <span class="badge text-dark badge-missing-input">
+        <i class="fa fa-question-circle" aria-hiden="true"></i> Missing Input
```
SethPoulsen, line 107: aria-hidden is misspelled here

elements/pl-units-input/pl-units-input.mustache
```diff
@@ -0,0 +1,188 @@
...
+<span id="pl-units-input-submission-{{uuid}}">
+    {{#parse_error}}
+    {{#label}}<span>{{{label}}}</span>{{/label}}
+    {{#raw_submitted_answer}}<code class="user-output-invalid">{{raw_submitted_answer}}</code>{{/raw_submitted_answer}}
+    <span class="badge text-danger badge-invalid">
+        <i class="fa fa-exclamation-triangle" aria-hidden="true"></i> Invalid
+    </span>
+    <a href="javascript:void(0);" role="button" class="btn btn-sm btn-secondary small border"
+        data-placement="auto" data-trigger="focus" data-toggle="popover" data-html="true"
+        style="margin-left: 5px" title="Format Error" tabindex="0" data-content="{{parse_error}}">
```
SethPoulsen, line 98: why are we messing with tabindex here?

elements/pl-units-input/pl-units-input.mustache
```diff
@@ -0,0 +1,188 @@
...
+    {{/parse_error}}
+
+    {{#missing_input}}
+    {{#label}}<span>{{{label}}}</span>{{/label}}
+    <span class="badge text-dark badge-missing-input">
+        <i class="fa fa-question-circle" aria-hiden="true"></i> Missing Input
+    </span>
+    <a href="javascript:void(0);" role="button" class="btn btn-sm btn-secondary small border"
+        data-placement="auto" data-trigger="focus" data-toggle="popover" data-html="true"
+        style="margin-left: 5px" title="Format Error" tabindex="0"
```
SethPoulsen, line 111: Again, why mess with tabindex?

elements/pl-units-input/pl-units-input.py
```diff
@@ -0,0 +1,468 @@
...
+            placeholder_text = pl.get_string_attrib(element, "placeholder")
+        elif grading_mode is GradingMode.ONLY_UNITS:
+            placeholder_text = "Unit"
+        elif grading_mode is GradingMode.WITH_UNITS:
+            atol = pl.get_string_attrib(element, "atol", ATOL_DEFAULT)
+            placeholder_text = f"Number (atol={atol}) + Unit"
+
+        # Grading mode must be units fixed at this point,
+        # now can case based on comparison
+        elif comparison is uu.ComparisonType.RELABS:
```
SethPoulsen, line 180: I would prefer a nested if statement to case based on a different variable, with no comments, rather than having a complicated comment to explain the odd programming pattern. I think its more readable that way.

elements/pl-units-input/pl-units-input.py
```diff
@@ -0,0 +1,468 @@
...
+        parse_error = data["format_errors"].get(name, None)
+        html_params = {
+            "submission": True,
+            "label": label,
+            "parse_error": parse_error,
+            "uuid": pl.get_uuid(),
+        }
+
+        if parse_error is None and name in data["submitted_answers"]:
+            # Get submitted answer, raising an exception if it does not exist
```
SethPoulsen, line 232: this comment doesn't seem necessary

SethPoulsen (commented):
Looks good overall, just some minor comments.

elements/pl-units-input/pl-units-input.py
```diff
@@ -0,0 +1,468 @@
...
+    correct_answer = data["correct_answers"].get(name)
+
+    digits = pl.get_integer_attrib(element, "digits", None)
+    if digits is not None and digits <= 0:
+        raise ValueError(
+            f"Number of digits specified must be at least 1, not {digits}."
+        )
+
+    atol = pl.get_string_attrib(element, "atol", ATOL_DEFAULT)
+    ureg = pl.get_unit_registry()
```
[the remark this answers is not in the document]
eliotwrobson: This is just how the library is set up. It always requires a registry, and the reason it's slow and needs caching is because the default constructor reads from a file.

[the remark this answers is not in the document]
SethPoulsen: Oh, I see from the docs it has something to do with speeding up operations on units. That's annoying but I guess we just deal with it.

docs/elements.md
eliotwrobson: Whether or not detailed feedback is shown (so like telling people whether their specific unit or magnitude is off). This doesn't do the same thing as hide in panel, as that will hide the correct answer along with the feedback, and instructors should have the granularity.

SethPoulsen: Sounds good, maybe add a little bit more of that info to the description here so its more clear.

elements/pl-units-input/pl-units-input.mustache
eliotwrobson: This is copied from another element (not sure which one, actually). Looking through the repo, virtually all of the elements set `tabindex="0"` in their format error popover. Fine with avoiding removing this but it's probably better to be consistent.

docs/elements.md
eliotwrobson: This is intentional (treating `answers-name` as a variable) and consistent with the docs for other elements.

docs/elements.md
eliotwrobson: Meant to put `"exact-units"`, got lost while I was changing the name here.

docs/elements.md
SethPoulsen: `answers-name` is not a valid python variable. It parses as an expression where you are subtracting the variable `answers` from the variable `name`

elements/pl-units-input/pl-units-input.py
eliotwrobson: Done!

elements/pl-units-input/pl-units-input.py
eliotwrobson: Removed in [4d5b245](https://github.com/PrairieLearn/PrairieLearn/pull/6790/commits/4d5b245479e0260867e69994f65d9d5b8dd71cdb).

elements/pl-units-input/pl-units-input.py
```diff
@@ -0,0 +1,468 @@
...
+            html_params["suffix"] = suffix
+            html_params["a_sub"] = a_sub
+            html_params["raw_submitted_answer"] = raw_submitted_answer
+
+        elif name not in data["submitted_answers"]:
+            html_params["missing_input"] = True
+            html_params["parse_error"] = None
+
+        else:
+            # Getting data from submitted answers means that
```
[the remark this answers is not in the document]
eliotwrobson: Removed with other comment.

elements/pl-units-input/pl-units-input.mustache
eliotwrobson: Fixed in [4d5b245](https://github.com/PrairieLearn/PrairieLearn/pull/6790/commits/4d5b245479e0260867e69994f65d9d5b8dd71cdb).

docs/elements.md
eliotwrobson: I'm in favor of going back and addressing this later; all of the element docs do this (which I admit is kinda weird).

elements/pl-units-input/pl-units-input.mustache
SethPoulsen: @nwalters512 is there a good explanation for this?

docs/elements.md
eliotwrobson: Added some discussion in https://github.com/PrairieLearn/PrairieLearn/pull/6790/commits/4d5b245479e0260867e69994f65d9d5b8dd71cdb. Could add more discussion but the feedback given is actually pretty rudimentary.

SethPoulsen (approved):

elements/pl-units-input/unit_utils.py
```diff
@@ -0,0 +1,152 @@
...
+            return pl.is_correct_scalar_ra(
+                a_sub=submitted_magnitude_parsed,
+                a_tru=correct_magnitude_parsed,
+                rtol=rtol,
+                atol=parsed_atol.magnitude,
+            )
+
+        assert_never(comparison)
+
+    magnitudes_match_credit = (
```
stevenxpan, line 90: I would just use an `if` statement if the ternary is going to be split across multiple lines.

stevenxpan (approved):

elements/pl-units-input/unit_utils.py
eliotwrobson: I think this was a one-liner before, but the autoformatter broke it up.

elements/pl-units-input/pl-units-input.py
```diff
@@ -0,0 +1,467 @@
...
+        "grading-mode",
+        "comparison",
+        "rtol",
+        "atol",
+        "digits",
+        "size",
+        "show-help-text",
+        "placeholder",
+        "magnitude-partial-credit",
+        "show-feedback",
```
eliotwrobson, line 62: Note to self: change to allow feedback for consistency with other elements

elements/pl-units-input/pl-units-input.py
```diff
@@ -0,0 +1,467 @@
...
+
+        ureg = pl.get_unit_registry()
+        a_tru_parsed = ureg.Quantity(a_tru)
+
+        if grading_mode is GradingMode.ONLY_UNITS:
+            a_tru_string = f"{a_tru_parsed.units}"
+        else:
+            # Display reference solution with the given number of digits
+            # TODO double check if this is confusing or not
+            a_tru_string = f"{a_tru_parsed:.{digits}g}"
```
eliotwrobson, line 282: Note: should change this to just take in a custom format string instead of using digits at all.

exampleCourse/questions/element/unitsInput/server.py
```diff
@@ -0,0 +1,8 @@
+import prairielearn as pl
+
+
+def generate(data):
+    ureg = pl.get_unit_registry()
```
nwalters512, line 5: How fast is this in the cached case? Is there a reason we're explicitly choosing to make `pint` part of our public interface? Unless there's a good reason not to, I'd rather just tell instructors to do something like `f"{num_here} ft"`.

eliotwrobson: The speed difference was significant enough that I couldn't make progress on writing the test question without setting this up first as part of the element (it almost times out on every grade call). This is actually why I had to set some custom options for the test fixture to avoid creating this registry too many times:

https://github.com/eliotwrobson/PrairieLearn/blob/8ec7cd581560f49f7a34b1945ebabfa80d2781ef/elements/pl-units-input/units_utils_test.py#L9-L11

I haven't actually timed it, but removing `scope="module"` from the above decorator and running tests locally, the slowdown is extremely noticeable.

Regarding making `pint` part of the public interface, the reasons are twofold. Primarily, now that it's added as a dependency, I think any question code can import + call from `pint`, so adding this helper function avoids people running into slowdown because of creating a unit registry without using a cache file (and not knowing where a writeable directory might be).

Second, `pint` actually has uses beyond just this parsing for question writing. In particular, it allows for "unitful" computations, where operating on quantities with units will be automatically expressed in derived units, see https://pint.readthedocs.io/en/stable/getting/tutorial.html#simplifying-units.

elements/pl-units-input/pl-units-input.mustache
nwalters512: See https://getbootstrap.com/docs/4.6/components/popovers/#dismiss-on-next-click - this is necessary when using the `focus` popover trigger.

elements/pl-units-input/pl-units-input.mustache
```diff
@@ -0,0 +1,186 @@
...
+<samp>{{a_tru}}</samp>
+{{#suffix}}<span>{{suffix}}</span>{{/suffix}}
+{{/answer}}
+
+
+{{#format}}
+<!--
+    Might need separate format errors for first and last inputs.
+    TODO: more detailed format error messages
+-->
```
nwalters512, line 176: Is this still a valid TODO?

eliotwrobson: Nah, playing with it now the format errors are pretty detailed. Will remove. There's another TODO about more test cases, but that's handled by backend testing. Will remove that too.

nwalters512 (approved):

PR #6791, opened 2022-12-22
fix: Proper handling of pending grading jobs for socket updates

jonatanschroeder:
Based on discussion on Slack. There are three related bugs:

* An unnecessary call to `updateStatus` with incorrect parameters;
* Save-only submissions are considered pending;
* The last grading job counts manual grading jobs instead of external only.

These will cause unnecessary submissions to be reloaded via socket, and may prevent submissions from loading as expected.

A separate issue is that socket loading does not parse the manual grading feedback. While this is unlikely to cause an issue in usual circumstances, since it would require manual grading to finish while the student has the question open but where external grading has not yet completed, it was an issue in the case above where a submission was considered pending unnecessarily. It is being fixed for completeness as well.


lib/externalGradingSocket.sql
```diff
@@ -20,16 +20,11 @@ SELECT
     grading_job_status(gj.id) AS grading_job_status,
     s.variant_id AS variant_id
 FROM
-    grading_jobs AS gj
+    grading_jobs AS gj_orig
+    JOIN grading_jobs AS gj ON (gj.submission_id = gj_orig.submission_id)
```
nwalters512, line 24: Checking my understanding here: we're finding the submissions associated with a particular grading job, then all the grading jobs associated with that submission, and then picking the most recent one?

pages/partials/externalGradingLiveUpdate.ejs
```diff
@@ -19,8 +19,8 @@
 
       var status = $(elem).data('grading-job-status');
       var submissionId = $(elem).attr('id').replace('submission-', '');
-      updateStatus(submissionId, status);
```
nwalters512, line 22: I believe this call is still necessary to show the correct status immediately after page load, though I do see that it's clearly being called with incorrect parameters. Maybe this would work instead?

```js
updateStatus({ id: submissionId, grading_job_status: status });
```

lib/externalGradingSocket.sql
```diff
@@ -20,16 +20,11 @@ SELECT
...
-            SELECT s2.id
-            FROM submissions AS s2
-            RIGHT JOIN grading_jobs AS gj3 ON (gj3.submission_id = s2.id)
-            WHERE gj3.id = $grading_job_id
-        )
-    );
+    gj_orig.id = $grading_job_id
+    AND gj.grading_method != 'Manual'
+ORDER BY gj.date DESC, gj.id DESC
+LIMIT 1;
```
nwalters512, line 30: It looks like this query is invoked with `queryOneRow` - is there a risk of ending up with zero rows? Given that the function that uses this query should only be called in response to an external grading job completing, I think we're safe, I just want to double check.

lib/externalGradingSocket.sql
jonatanschroeder: Yes. I'm just rewriting the same query that was there before to make it clearer. If the associated submission has a new grading job, the latest one is selected instead.

pages/partials/externalGradingLiveUpdate.ejs
jonatanschroeder: Note that this function does not retrieve the new status, it only uses the status from the data attributes. So presumably the data being updated here was already used in rendering. It would be updated in the init event, later. So this call is not actually necessary.

lib/externalGradingSocket.sql
jonatanschroeder: I see this query is called with a known grading job number, so presumably it should return the original grading job itself if no other grading job exists. I checked all references to this function, and all of them seem to work with the assumption that the job exists.

pages/partials/externalGradingLiveUpdate.ejs
nwalters512: We don't actually render the status messages or spinners on the server; that's done strictly client-side. We could make the change to render them on the server if you want.

lib/externalGradingSocket.sql
nwalters512: Good point; I'm satisfied that this is good to go!

pages/partials/externalGradingLiveUpdate.ejs
nwalters512: Of course, things were working well enough before with the obviously incorrect call, so I think it would be ok to delete it. But my intention in writing this was definitely to ensure that we can show something useful even before we've gotten the initial state back from the websocket.

jonatanschroeder: Things were working because, if this display was expected to be shown, then grading is also pending, so the socket would be immediately initialized and the status would be updated. So this call, if it were correct, would only serve to set the status between when the page is loaded and when the socket returned the new status. This time is too short to be noticed in the vast majority of cases. In any case, I've added it back with your suggestion.

lib/question.js
```diff
@@ -1888,29 +1888,35 @@ module.exports = {
...
-                  'partials',
-                  'submission.ejs'
-                );
-                ejs.renderFile(templatePath, renderParams, (err, html) => {
-                  if (ERR(err, callback)) return;
-                  panels.submissionPanel = html;
-                  callback(null);
-                });
+                markdown
+                  .processContent(submission.feedback?.manual?.toString() || '')
```
jonatanschroeder, line 1892: Also being done (in asyncified version) in #6271.

lib/question.js
```diff
@@ -1888,29 +1888,35 @@ module.exports = {
...
+                      'pages',
+                      'partials',
+                      'submission.ejs'
+                    );
+                    ejs.renderFile(templatePath, renderParams, (err, html) => {
+                      if (ERR(err, callback)) return;
+                      panels.submissionPanel = html;
+                      callback(null);
+                    });
+                  });
```
nwalters512, line 1919: We should handle any errors encountered while rendering Markdown:

```suggestion
                  }, (err) => ERR(err, callback));
```

nwalters512 (approved):

PR #6793, opened 2022-12-27
build: Add black and isort to Python build

### LINKED ISSUE #5027: Use Black to format Python code

eliotwrobson:
Add black and isort to Python workflow and as part of `make format` and lint checks. Resolves #5027. Also, condenses all Python tooling configuration to a single `pyproject.toml` file for easier editing and less clutter.

Note that this doesn't make any changes to Python files in the `python` folder. This is because the old linting was not configured to run on these files, and they require some more substantial changes before this can be enabled. I think that should be delegated to a later PR (as those code changes are more than just running autoformatting, as has been done here).

The intention of this PR is not to squash and merge, but merge both commits together. Reviews should focus on the changes to the config files made in the second commit, as the first commit is just a result of running the autoformatting as configured.


pyproject.toml
```diff
@@ -0,0 +1,23 @@
+[tool.flake8]
+# TODO add python directory here in later PR
+filename = ["./elements/**/*.py", "./question-servers/**/*.py"]
+ignore = ["E501"]
+# Black compatibility
+max-line-length = 88
+extend-ignore = ["E203", "W503"]
+
+[tool.black]
+include = '^\/elements\/.*\.py$|^\/question-servers\/.*\.py$'
```
eliotwrobson, line 10: Note that (for some reason) black uses a pure regex to select files for the include, which differs from the way that flake8 and isort are configured.

.github/workflows/main.yml
```diff
@@ -65,26 +65,26 @@ jobs:
...
       run: docker exec test_container /PrairieLearn/docker/lint_html.sh
-    - name: Run the Python linter
-      run: docker exec test_container /PrairieLearn/docker/lint_python.sh
-    - name: Run the JavaScript typechecker
-      run: docker exec test_container /PrairieLearn/docker/typecheck_js.sh
     - name: Run the Python typechecker
       run: docker exec test_container /PrairieLearn/docker/typecheck_python.sh
+    - name: Run the JavaScript typechecker
+      run: docker exec test_container /PrairieLearn/docker/typecheck_js.sh
+    - name: Run the Python tests
```
eliotwrobson, line 78: Moved the Python testing to run before the javascript. Although a bit off-issue, I think this is a good idea since the Python is just backend for the javascript anyway, and it's annoying to have to sometimes wait 10 minutes for javascript tests to run just to have a Python test fail.

Makefile
```diff
@@ -50,16 +50,20 @@ lint-js:
 	@yarn eslint --ext js --report-unused-disable-directives "**/*.js"
 	@yarn prettier --check "**/*.{js,ts,md}"
 lint-python:
-	@python3 -m flake8 ./
+	@python3 -m flake8p ./
```
eliotwrobson, line 53: Note the plugin for using a `pyproject.toml` file requires a different command. (see [here](https://pypi.org/project/Flake8-pyproject/)).

SethPoulsen:
oooh isort looks cool, I didn't even know that existed.

I haven't had a chance to look in detail yet, but my one comment so far is: can we not change the quotation convention? Is there a good reason to change? It will just make git diffs messier than necessary and force everyone to relearn.

eliotwrobson:
> Is there a good reason to change? It will just make git diffs messier than necessary and force everyone to relearn.

Using black effectively forces you to use double quotes instead of single ones. There's no option to set single quotes for it (see detailed discussion [here](https://github.com/psf/black/issues/118)), and the flake8 plugin doesn't have an option to disable enforcement of just this rule (even though black has an option to skip reformatting strings). The inability to automatically reformat strings

pyproject.toml
```diff
@@ -0,0 +1,23 @@
+[tool.flake8]
+# TODO add python directory here in later PR
```
jonatanschroeder, line 2: Any reason why `./python` is not included now? Maybe `./exampleCourse` and `./graders` as well. It's such a disruptive PR, may as well do it at once. Also note that `./question-servers` is not necessary, it used to be for when `./python` was called `./question-servers/freeformPythonLib`.

pyproject.toml
```diff
@@ -0,0 +1,23 @@
...
+filename = ["./elements/**/*.py", "./question-servers/**/*.py"]
+ignore = ["E501"]
+# Black compatibility
+max-line-length = 88
```
jonatanschroeder, line 6: `.prettierrc.json` uses 100. Should we use 100 for consistency with that? I don't feel strongly about it, though.

jonatanschroeder (commented):
This looks great. 

I'm not sure, but it may be good to also check some of the docs that have non-formatted Python docs that may benefit from an update for consistency. I won't be offended if you don't though 😄. A quick check brings up:

* `docs/elementExtensions.md` (single-quotes)
* `docs/elements.md` (single-quotes and indentation)
* `docs/getStarted.md` (indentation)
* `docs/pl-drawing/index.md` (single-quotes)
* `docs/python-grader/index.md` (single-quotes and indentation, maybe line length)
* `docs/c-grader/index.md` (single-quotes)

pyproject.toml
eliotwrobson: `./python` was not included before (you can check the old deleted config). I'm fine running black / isort over everything at once, but there's changes outside of running those required for `./python` to pass a lint check (there are issues like unused imports, improper comparisons, etc). So we can run them over those files, but can't set up enforcement yet.

I didn't know about the question servers thing though, I'll go ahead and remove.

pyproject.toml
eliotwrobson: Black does have the ability to change this (one of the few configurable things), so I think changing to 100 lines isn't unreasonable.

pyproject.toml
eliotwrobson: @jonatanschroeder changes to docs are here: https://github.com/PrairieLearn/PrairieLearn/pull/6794

.git-blame-ignore-revs
```diff
@@ -4,3 +4,6 @@
 
 # Format all files with Prettier
 2f3b359987bebf00d45a8c32d58e3b302d117c4a
+
+# Format element Python files with black
+a0f42e54dea00d363b2a7e443d6085d5fd3923ea
```
eliotwrobson, line 9: Will switch this out once we decide which files to run this on / before this is merged.

pyproject.toml
nwalters512: Let's keep this at 88 for now.

pyproject.toml
nwalters512: @eliotwrobson I'd be fine to leave formatting + linting `./python` to another PR, but can we do that as a very quick followup? I'd like for all Python to be treated the same.

Makefile
nwalters512: > For compatibility with earlier versions of this package, and perhaps extra reliability in terms of possible future breakage of the plug-in hook, the package also provides a flake8p command that could be called alternatively to lint the code.

The way I read this, we can keep using the normal `flake8` command. Is that not the case?

Makefile
```diff
@@ -50,16 +50,20 @@ lint-js:
...
 	@node tools/validate-links.mjs
 
-format: format-js
+format: format-js format-python
 format-js:
 	@yarn eslint --ext js --fix "**/*.js"
 	@yarn prettier --write "**/*.{js,ts,md}"
+format-python:
+# Very annoying that isort doesn't allow this to be set in the config
+	@python3 -m isort ./elements/**/*.py
```
nwalters512, line 65: Rather than doing a positive match, could we instead run `isort ./` and set `skip_glob` in the config file? I'm pretty sure that this is the semantics we want anyways; IMO, linting + formatting should be opt-out, not opt-in. Ditto for the other tools, e.g. Black, though I think we can leave Pyright as opt-in for now.

nwalters512 (approved):
This looks great! I'd like to discuss opt-in vs. opt-out before merging. Let's make sure that `.git-blame-ignore-revs` has the correct SHA too.

pyproject.toml
eliotwrobson: Sounds like a plan! There are a couple litter carveouts that are needed, but not too bad otherwise. I just wanted to avoid making this PR more disruptive than necessary.

I'm traveling right now, but once this is merged I should be able to bang out that next one in a couple hours.

Makefile
eliotwrobson: That makes sense I think. The linting might need to be opt-in until more files can be made compliant, but we can just always run isort and black over everything

eliotwrobson: @nwalters512 set up negative matching for autoformatting in latest push! I think we can change the linting to use this once a few more files are brought into compliance. For now, the autoformatter is run over the `elements`, `python`, and `graders` directories, so the later PR to turn on linting for those should be shorter (since it won't include running the autoformatting itself).

.git-blame-ignore-revs
```diff
@@ -4,3 +4,6 @@
 
 # Format all files with Prettier
 2f3b359987bebf00d45a8c32d58e3b302d117c4a
+
+# Format Python files with black + isort
+61dea371246bb7e14e42649746187c9e0628ba1b
```
eliotwrobson, line 9: @nwalters512 can confirm, this is the latest commit hash for running the formatter:

image

Makefile
```diff
@@ -56,10 +56,14 @@ lint-html:
 lint-links:
 	@node tools/validate-links.mjs
 
-format: format-js
+format: format-js format-python
 format-js:
 	@yarn eslint --ext js --fix "**/*.js"
 	@yarn prettier --write "**/*.{js,ts,md}"
+format-python:
+# Very annoying that isort doesn't allow this to be set in the config
```
nwalters512, line 64: I believe this comment is no longer relevant?
```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,24 @@
+[tool.flake8]
+# TODO add python directory here in later PR
+filename = ["./elements/**/*.py"]
+ignore = ["E501"]
+# Black compatibility
+max-line-length = 88
+extend-ignore = ["E203", "W503"]
+
+[tool.black]
+exclude = '^\/tools\/.*\.py$|^\/exampleCourse\/.*\.py$|^\/testCourse\/.*\.py$|^\/node_modules\/.*\.pyi?$'
```
nwalters512, line 10: IMO it'd be great to include all of these (except `node_modules`) in the future. Can we do that as a relatively quick followup?

nwalters512 (approved):
A final nit: could you rename the "Ran formatting" commit to something more specific like "Run black and isort on Python files"?

pyproject.toml
eliotwrobson: Sure! The only thing that has to stay out is some of the tools, there's some Python 2 in there.

Issue #2067, opened 2020-02-07
Document best practices for custom grading functions


mwest1066:
* Use build-in elements for input (not raw HTML `<input>`) to get input validation and format help
* Set `data["score"]`
* Unset `data["partial_scores"]` to avoid badges showing partial scores

Issue #2056, opened 2020-02-05
Document best practices for rounding and displaying values


mwest1066:
Document the following somewhere.

It's important to not display rounded values to students but compute with full precision. E.g., this is bad:
```
def generate(data):
    a = 33.33337
    b = 33.33333
    data["params"]["a_for_student"] = f'{a:.2f}'
    data["params"]["b_for_student"] = f'{a:.2f}'
    data["correct_answer"] = a - b
```

Instead, we should round first and use the rounded values to compute the true answer:
```
def generate(data):
    a = np.round(33.33337, 2)
    b = np.round(33.33333, 2)
    data["params"]["a_for_student"] = f'{a:.2f}'
    data["params"]["b_for_student"] = f'{b:.2f}'
    c = a - b
    data["correct_answer"] = c
```

Issue #2745, opened 2020-07-17
Add docs for Mustache syntax in question.html


mwest1066:
In the docs we should link to https://mustache.github.io/mustache.5.html and also provide an easy-to-read summary of the syntax in our documentation. Maybe a few more example questions could also show common use cases.

eliotwrobson:
There are now a number of different mentions of mustache syntax in the documentation and custom example questions: https://prairielearn--11380.org.readthedocs.build/en/11380/question/#question-questionhtml

https://github.com/PrairieLearn/PrairieLearn/tree/master/exampleCourse/questions/demo/randomZippedFile

PR #6796, opened 2022-12-29
docs: Improve documentation for question writing

eliotwrobson:
Resolves https://github.com/PrairieLearn/PrairieLearn/issues/2056, resolves https://github.com/PrairieLearn/PrairieLearn/issues/2067. Corrects an error which stated that functions should return the data dictionary instead of modifying it in-place. Handles the documentation part of #5881.

Related: It looks like https://github.com/PrairieLearn/PrairieLearn/issues/2745 has already been mentioned in the docs, so I think this can be closed.


docs/question.md
```diff
@@ -438,12 +443,46 @@ Example of valid HTML:
...
+
+### Custom Question Best Practices
+
+- Although custom questions usually don't use the grading functions from individual elements, it is _highly_ recommended that built-in elements are used for student input. Parsed student answers are present in the `data["submitted_answers"]` dictionary.
+
+- Any custom grading function for the whole question should set `data["score"]` as a value between 0.0 and 1.0, which will be the final score for the given question.
```
jonatanschroeder, line 460: I'd argue for a paragraph-style text instead of a list, for consistency with the rest of the document.

docs/question.md
```diff
@@ -438,12 +443,46 @@ Example of valid HTML:
...
+
+- If a custom grading function is only grading a specific part of a question, the grading function should set the corresponding dictionary entry in `data["partial_scores"]` and then recompute the final `data["score"]` value for the whole question. To avoid showing score badges for individual parts, the dictionary entries in `data["partial_scores"]` should be unset.
+
+- To set custom feedback, the grading function should set the corresponding entry in the `data["feedback"]` dictionary. These feedback entries are passed in when rendering the `question.html`, which can be accessed by using the mustache prefix `{{feedback.}}`.
```
jonatanschroeder, line 464: This may benefit from a code block with an example (either move the feedback code above here, or point to the code above).

docs/question.md
```diff
@@ -438,12 +443,46 @@ Example of valid HTML:
...
+  ```python
+  def generate(data):
+    a = np.round(33.33337, 2)
+    b = np.round(33.33333, 2)
+    data["params"]["a_for_student"] = f'{a:.2f}'
+    data["params"]["b_for_student"] = f'{b:.2f}'
+    data["correct_answers"]["c"] = a - b
+  ```
+
+- Similarly, for grading functions involving floating point numbers, _avoid exact comparisons with `==`._ Floating point calculations in Python introduce error, and comparisons with `==` might unexpectedly fail. Instead, the function [`math.isclose`](https://docs.python.org/3/library/math.html#math.isclose) should be used, as it performs comparisons with tolerance values.
```
jonatanschroeder, line 488: You can also point them to `pl.is_correct_scalar_ra`, `pl.is_correct_scalar_sf` and `pl.is_correct_scalar_dd` for more fine-tuned comparison options.

docs/question.md
```diff
@@ -438,12 +443,46 @@ Example of valid HTML:
 
 For most [elements] there are four different ways of auto-grading the student answer. This applies to elements like [`pl-number-input`](elements/#pl-number-input-element) and [`pl-string-input`](elements/#pl-string-input-element) that allow students to input an answer of their choosing, but not [`pl-multiple-choice`](elements/#pl-multiple-choice-element) or [`pl-checkbox`](elements/#pl-checkbox-element) that are much more constrained. The four ways are:
 
-1. Set the correct answer using the correct-answer attribute in `question.html`. This is for hard-coded, fixed answers. We normally want some degree of randomization of the question, so this is the least-used method.
+1. Set the correct answer using the correct-answer attributes for each element in `question.html`. This will use the built-in grading methods for each element. We normally want some degree of randomization of the question, so this is the least-used method.
```
jonatanschroeder, line 446: I find the statement that "this is the least-used method" a bit too strong. Maybe:
```suggestion
1. Set the correct answer using the correct-answer attributes for each element in `question.html`. This will use the built-in grading methods for each element. Given that this option is typically used for answers with a hard-coded value, without randomization, it is not expected to be used frequently.
```

docs/question.md
```diff
@@ -207,11 +209,12 @@ def grade(data):
     # grade() can also set `data['format_errors'][NAME]` if there is any reason to mark the question
     # invalid during grading time.  This will cause the question to not use up one of the student's attempts' on exams.
```
jonatanschroeder, line 210:
```suggestion
    # grade() can also set `data['format_errors'][NAME]` if there is any reason to mark the question
    # invalid during grading time.  This will cause the question to not use up one of the student's attempts' on exams.
    # You are encouraged, though, to do any checks for invalid data that can be done in `parse(data)` there instead,
    # since that method is also called when the student hits "Save only" or in assessments without real-time grading.
```

jonatanschroeder:
> Resolves #2056 and #2067.

If you want Github to link these two issues to this PR, you should write the keyword once for each issue, like this: Resolves #2056, resolves #2067.

docs/question.md
eliotwrobson: I think I'll refer to the code above. Custom feedback appears in a lot of custom questions, so if we add more examples, they should all use it.

docs/question.md
```diff
@@ -204,14 +206,17 @@ def grade(data):
...
-            data["score"] = 0.5
+    # You are encouraged, though, to do any checks for invalid data that can be done in `parse(data)` there instead,
+    # since that method is also called when the student hits "Save only" or in assessments without real-time grading.
+
+    # As an example, we will give half points for incorrect answers larger than "x",
+    # only if not already correct. Use math.isclose to avoid possible floating point errors.
+    if math.isclose(data["score"], 0.0) and data["submitted_answers"]["y"] > data["params"]["x"]:
+        data["partial_scores"]["y"]["score"] = 0.5
+        data["score"] = 0.5
+        data["feedback"]["y"] = "Your value for $y$ is larger than $x$, but incorrect."
```
jonatanschroeder, line 219: With #6798 in mind this may need to be moved to partial scores. Same for the question.html code above.

docs/question.md
```diff
@@ -438,12 +445,46 @@ Example of valid HTML:
...
+
+### Custom Question Best Practices
+
+Although custom questions usually don't use the grading functions from individual elements, it is _highly_ recommended that built-in elements are used for student input. Parsed student answers are present in the `data["submitted_answers"]` dictionary.
+
+Any custom grading function for the whole question should set `data["score"]` as a value between 0.0 and 1.0, which will be the final score for the given question.
+
+If a custom grading function is only grading a specific part of a question, the grading function should set the corresponding dictionary entry in `data["partial_scores"]` and then recompute the final `data["score"]` value for the whole question. To avoid showing score badges for individual parts, the dictionary entries in `data["partial_scores"]` should be unset.
+
+To set custom feedback, the grading function should set the corresponding entry in the `data["feedback"]` dictionary. These feedback entries are passed in when rendering the `question.html`, which can be accessed by using the mustache prefix `{{feedback.}}`. See the [above custom question](#Question-server.py) for an example of this.
```
jonatanschroeder, line 466: Same as above.

docs/question.md
```diff
@@ -438,12 +445,46 @@ Example of valid HTML:
...
+
+### Custom Question Best Practices
+
+Although custom questions usually don't use the grading functions from individual elements, it is _highly_ recommended that built-in elements are used for student input. Parsed student answers are present in the `data["submitted_answers"]` dictionary.
+
+Any custom grading function for the whole question should set `data["score"]` as a value between 0.0 and 1.0, which will be the final score for the given question.
+
+If a custom grading function is only grading a specific part of a question, the grading function should set the corresponding dictionary entry in `data["partial_scores"]` and then recompute the final `data["score"]` value for the whole question. To avoid showing score badges for individual parts, the dictionary entries in `data["partial_scores"]` should be unset.
```
jonatanschroeder, line 464: I think these two paragraphs can be combined. Also, the last sentence seems to suggest this is always desirable, rather than a preference. Here's a suggestion, feel free to tweak it.
```suggestion
Any custom grading function for the whole question should set `data["score"]` as a value between 0.0 and 1.0, which will be the final score for the given question. If a custom grading function is only grading a specific part of a question, the grading function should set the corresponding dictionary entry in `data["partial_scores"]` and then recompute the final `data["score"]` value for the whole question. If you would prefer not to show score badges for individual parts, you may unset the dictionary entries in `data["partial_scores"]`.
```

docs/question.md
eliotwrobson: #6798 only applies to element code, not custom `server.py` files. I believe that the `feedback` entry in the data dictionary is specifically for use by custom question grader code, and the feedback in partial scores is for element grader code.

This is pretty confusing though, so I'll add some discussion about this somewhere.

jonatanschroeder:
Given that @mwest1066 was the one that raised the issues above I'll let him do the final approval, just in case there was something else he had in mind, but I don't see any issues here.

docs/question.md
```diff
@@ -438,12 +445,44 @@ Example of valid HTML:
 
 For most [elements] there are four different ways of auto-grading the student answer. This applies to elements like [`pl-number-input`](elements/#pl-number-input-element) and [`pl-string-input`](elements/#pl-string-input-element) that allow students to input an answer of their choosing, but not [`pl-multiple-choice`](elements/#pl-multiple-choice-element) or [`pl-checkbox`](elements/#pl-checkbox-element) that are much more constrained. The four ways are:
 
-1. Set the correct answer using the correct-answer attribute in `question.html`. This is for hard-coded, fixed answers. We normally want some degree of randomization of the question, so this is the least-used method.
+1. Set the correct answer using the correct-answer attributes for each element in `question.html`. This will use the built-in grading methods for each element. Given that this option is typically used for answers with a hard-coded value, without randomization, it is not expected to be used frequently.
 
 2. Set `data["correct_answers"][VAR_NAME]` in server.py. This is for questions where you can pre-compute a single correct answer based on the (randomized) parameters.
```
nwalters512, line 450:
```suggestion
2. Set `data["correct_answers"][VAR_NAME]` in `server.py`. This is for questions where you can pre-compute a single correct answer based on the (randomized) parameters.
```

docs/question.md
```diff
@@ -438,12 +445,44 @@ Example of valid HTML:
...
+If a question has more than one of the above options, each of them overrides the one before it. Even if options 3 (custom grade function) or 4 (external grader) are used, then it can still be helpful to set a correct answer so that it is shown to students as a sample of what would be accepted. If there are multiple correct answers then it's probably a good idea to add a note with [`pl-answer-panel`](elements/#pl-answer-panel-element) that any correct answer would be accepted and the displayed answer is only an example. Moreover, if there is no relevant information to display on the correct answer panel (i.e., a question has multiple correct answers and is meant to be attempted until a full score is achieved), then the panel can be hidden by setting `showCorrectAnswer: false` in `info.json`.
+
+### Custom Question Best Practices
```
nwalters512, line 458: I'd rather call this "custom grading" instead of "custom questions", as *all* questions are technically custom.

```suggestion
### Custom grading best practices
```

docs/question.md
```diff
@@ -438,12 +445,44 @@ Example of valid HTML:
...
+
+Although custom questions usually don't use the grading functions from individual elements, it is _highly_ recommended that built-in elements are used for student input. Parsed student answers are present in the `data["submitted_answers"]` dictionary.
```
nwalters512, line 460: Adding a little rationale about *why* this is recommended.
```suggestion
Although questions with custom grading usually don't use the grading functions from individual elements, it is _highly_ recommended that built-in elements are used for student input, as these elements include helpful parsing and feedback by default. Parsed student answers are present in the `data["submitted_answers"]` dictionary.
```

docs/question.md
```diff
@@ -438,12 +445,44 @@ Example of valid HTML:
...
+
+To set custom feedback, the grading function should set the corresponding entry in the `data["feedback"]` dictionary. These feedback entries are passed in when rendering the `question.html`, which can be accessed by using the mustache prefix `{{feedback.}}`. See the [above custom question](#Question-server.py) or [this demo question](https://github.com/PrairieLearn/PrairieLearn/tree/master/exampleCourse/questions/demo/custom/gradeFunction) for examples of this. Note that the feeback set in the `data["feedback"]` dictionary is meant for use by custom grader code in a `server.py` file, while the feedback set in `data["partial_scores"]` is meant for use by element grader code.
+
+For generated floating point answers, it's important to round values before displaying them to the student, as this could cause unexpected behavior when students perform calculations. For example, the following is problematic:
```
nwalters512, line 466: I _think_ this is a more specific description of what's important?

```suggestion
For generated floating point answers, it's important to use consistent rounding when displaying numbers to students _and_ when computing the correct answer. For example, the following is problematic:
```

docs/question.md
```diff
@@ -438,12 +445,44 @@ Example of valid HTML:
...
+
+```python
+def generate(data):
+    a = 33.33337
+    b = 33.33333
+    data["params"]["a_for_student"] = f'{a:.2f}'
+    data["params"]["b_for_student"] = f'{a:.2f}'
+    data["correct_answers"]["c"] = a - b
```
nwalters512, line 474:
```suggestion
    # Note how the correct answer is computed with full precision,
    # but the parameters displayed to students are rounded.
    data["correct_answers"]["c"] = a - b
```

docs/question.md
```diff
@@ -438,12 +445,44 @@ Example of valid HTML:
...
+```
+
+The rounding should be done before the format strings are set, as in the following example:
```
nwalters512, line 477:
```suggestion
Instead, the numbers should be rounded at the beginning:
```

docs/question.md
```diff
@@ -438,12 +445,44 @@ Example of valid HTML:
...
+```python
+def generate(data):
+  a = np.round(33.33337, 2)
+  b = np.round(33.33333, 2)
+  data["params"]["a_for_student"] = f'{a:.2f}'
+  data["params"]["b_for_student"] = f'{b:.2f}'
+  data["correct_answers"]["c"] = a - b
+```
+
+Similarly, for grading functions involving floating point numbers, _avoid exact comparisons with `==`._ Floating point calculations in Python introduce error, and comparisons with `==` might unexpectedly fail. Instead, the function [`math.isclose`](https://docs.python.org/3/library/math.html#math.isclose) can be used, as it performs comparisons within given tolerance values. The functions `pl.is_correct_scalar_ra`, `pl.is_correct_scalar_sf`, and `pl.is_correct_scalar_dd` can be used to perform more specialized comparisons.
```
nwalters512, line 488: Consider linking to the `prairielearn` module or providing inline examples here, or maybe even both. If I didn't know about `import pl from prairielearn`, I wouldn't know where to find these functions or how to use them.

eliotwrobson: I think that makes sense. There are some example usages I can link to with these functions, and each of them is documented so that should be enough for people to get started. I'll probably also drop the `pl.*` import prefix and just state where they can be found.

eliotwrobson: Cleaned this description up some and included links to the functions in `prairielearn.py`. Also, added docstrings to those functions describing what they do.

nwalters512: Yeah, let's just link to the file. The specific line numbers are likely to become out-of-date almost immediately.

docs/question.md
```diff
@@ -485,4 +487,9 @@ def generate(data):
...
+
+Specifically, the function `is_correct_scalar_ra` compares floats using relative and absolute tolerances. The function
+`is_correct_scalar_sf` compares floats up to the number of significant figures given by the `digits` parameter. The function
+`is_correct_scalar_dd` compares floats up to the number of digits after the decimal place given by the `digits` parameter.
+More detailed information can be found in the docstrings for these functions.
```
nwalters512, line 495: Minor wordsmithing:

```suggestion
Similarly, for grading functions involving floating point numbers, _avoid exact comparisons with `==`._ Floating point calculations in Python introduce error, and comparisons with `==` might unexpectedly fail. Instead, the function [`math.isclose`](https://docs.python.org/3/library/math.html#math.isclose) can be used, as it performs comparisons within given tolerance values. The `prairielearn` Python library from [`prairielearn.py`](https://github.com/PrairieLearn/PrairieLearn/blob/master/python/prairielearn.py) also offers several functions to perform more specialized comparisons:

- `is_correct_scalar_ra` compares floats using relative and absolute tolerances
- `is_correct_scalar_sf` compares floats up to a specified number of significant figures
- `is_correct_scalar_dd` compares floats up to a specified number of digits

More detailed information can be found in the docstrings for these functions.
```

docs/question.md
```diff
@@ -438,12 +445,52 @@ Example of valid HTML:
...
+
+### Custom grading best practices
+
+Although questions with custom grading usually don't use the grading functions from individual elements, it is _highly_ recommended that built-in elements are used for student input, as these elements include helpful parsing and feedback by default. Parsed student answers are present in the `data["submitted_answers"]` dictionary.
+
+Any custom grading function for the whole question should set `data["score"]` as a value between 0.0 and 1.0, which will be the final score for the given question. If a custom grading function is only grading a specific part of a question, the grading function should set the corresponding dictionary entry in `data["partial_scores"]` and then recompute the final `data["score"]` value for the whole question. Alternatively, if you would prefer not to show score badges for individual parts, you may unset the dictionary entries in `data["partial_scores"]`.
```
eliotwrobson, line 462: Now that there are helper functions to recompute the score after changing partial_scores, I'll add some mention of them here.

eliotwrobson: Discussion has been added!

PR #6798, opened 2022-12-30
docs: Fix error in element writing docs

### LINKED ISSUE #5523: Custom element feedback issue
In writing a custom element, we're seeing an issue where, after modifying `data['feedback']` in the `grade` function, that dictionary entry is cleared out when the `render` function is called next. Specifically, our code looks something like this
```
    print(data['feedback']) # prints the empty dictionary, other operations behave as if this is empty

def grade(element_html, data):
    ...
    data['feedback'][question_name] = 'feedback for question'
```
The specific pull request where we're having this issue can be found [here](https://github.com/PrairieLearn/pl-uiuc-cs374/pull/195).

eliotwrobson:
Resolves #5523.


jonatanschroeder:
I'm not sure this is an error per se, but some sort of inconsistency in the code. I have feedback working in one of my questions, so it may be a matter of figuring out exactly when feedback can be used and when not.

nwalters512 (approved):

Issue #6787, opened 2022-12-21
Feature Request: Incorporate _R_ dataframe specific types and indexes into `pl-python-variable`/`pl-dataframe`


coatless:
The `pl-python-variable` element is constrained to only interpreting a _pandas_ data frame as an object present in _Python_.  That is, the `pl-python-variable` applies a _Python_ specific interpretation of a data frame (e.g. 0-based indices). With the incorporation of https://github.com/PrairieLearn/PrairieLearn/issues/6131, the element now becomes aware of _Pandas_ specific data types. 

However, the nature of a DataFrame structure differs only slightly between _R_ and _Python_. It would be nice to take advantage of the tabular display of data by merely modifying the output to be _R_ specific by starting row indices at 1 and allowing for `dtype` to be specific to _R_'s atomic types (e.g. `numeric`, `integer`, `character`, `logical`, ...)

eliotwrobson:
Is there a builtin way of doing this in Pandas? This just seems like a relabeling of data types, but unless people find R data types easier to read than their Python equivalents, I'm not sure who this might help.

coatless:
@eliotwrobson this is just adding a mapping of Python's Pandas data types to _R_ data types via a dictionary lookup. The goal here would be to re-use the `pl-dataframe` element for either Python or _R_ problems. As it stands now, the element mainly focuses on a _Python_ orientation of a data frame.  

At a later time, this could be expanded to Mathematica's [Dataset](https://reference.wolfram.com/language/ref/Dataset.html) or Julia's [Dataframes.jl](https://dataframes.juliadata.org/stable/). Glancing at the docs, I'm not seeing an immediate mapping table, but I can look more over next week/write one.

https://dataframes.juliadata.org/stable/man/comparisons/
https://pandas.pydata.org/docs/getting_started/comparison/comparison_with_r.html

coatless:
Returning back, I think the following covers a majority of the Pandas <-> R data type translation:

| R                                                                                       | Pandas     |
| --------------------------------------------------------------------------------------- | --------- |
[...]

Issue #6786, opened 2022-12-21
Feature Request: Incorporate copy code mechanics from pl-variable-output to pl-python-variable output


coatless:
I think the `pl-python-variable` feature could benefit from being able to generate output code for the underlying object being shown if desired. This feature exists already for the `pl-variable-output` that is constrained to display matrix/vector outputs.

c.f. 
https://prairielearn.readthedocs.io/en/latest/elements/#pl-variable-output-element 

My thought to provide an output format for the _dataframe_ part would be:

```python
import pandas as pd
df = pd.DataFrame({'col1': [1, 2],
                   'col2': [0.5, 0.75]},
                  index=['row1', 'row2'])

out_records = df.to_dict('split')

df_variable_name = "my_df"

code_string = f"""import pandas as pd
{df_variable_name} = pd.dataframe({out_records})
"""

print(code_string)
```

This should give: 

```python
import pandas as pd
my_df = pd.dataframe({'index': ['row1', 'row2'], 'columns': ['col1', 'col2'], 'data': [[1, 0.5], [2, 0.75]]})
```

Potentially, the element could receive a "tab" feature with the default tab showing the Pandas dataframe in a similar manner. In the _second_ tab, we could incorporate the above reconstruction code for students to be able to re-create the data frame.

eliotwrobson:
I think this would be well suited for a dataframe specific element. For other Python objects, I think that this is consistent with the current default behavior.

coatless:
@eliotwrobson well, the current behavior focuses on _displaying_ the actual variable instead of displaying the contents while assigning it to a variable.  Maybe incorporating a variable assignment option overall would be desirable? Many of the elements grew organically based on course needs.

PR #6800, opened 2022-12-31
Introduce pl-dataframe element

eliotwrobson:
@coatless new version of #6744. Addresses the following:

- First step in https://github.com/PrairieLearn/PrairieLearn/issues/6789 (before deprecating that functionality in the python variable element).
- Resolves https://github.com/PrairieLearn/PrairieLearn/issues/6786
- Resolves https://github.com/PrairieLearn/PrairieLearn/issues/6501
- Resolves https://github.com/PrairieLearn/PrairieLearn/issues/6131
- Resolves https://github.com/PrairieLearn/PrairieLearn/issues/2706
- Resolves https://github.com/PrairieLearn/PrairieLearn/issues/6787

Planning on updating documentation still Will also remove dataframe examples from the `pythonVariable` question as part of deprecation, once we verify that the old behavior is captured by the use of the new element. However, this deprecation may be broken off into its own PR later, since there are other changes that have been requested for the `pl-python-variable` element.

Not quite sure about how to incorporate https://github.com/PrairieLearn/PrairieLearn/issues/6788, since I'm not sure if it's worth adding the Datatable dependency and getting rendering to work just for this single use case (plus I don't know how to do it).

Screenshots:

image

image

image

R-types dataframe:

image

image


python/prairielearn.py
```diff
@@ -93,7 +94,32 @@ def to_json(v):
...
+            # generating numeric keys instead of strings for an index sequence with
+            # only numeric values (c.f. pandas-dev/pandas#46392)
+            df_modified_names = v.copy()
+
+            indexing_dtype = df_modified_names.columns.dtype
+            if indexing_dtype == np.float64 or indexing_dtype == np.int64:
+                df_modified_names.columns = df_modified_names.columns.astype('string')
+
+            # For version 2 storing a data frame, we use the table orientation alongside of
+            # enforcing a date format to allow for numeric values
```
coatless, line 111:
```suggestion
            # enforcing a date format to allow for passing datetime and missing (`pd.NA`/`np.nan`) values
```

exampleCourse/questions/element/dataframe/question.html
```diff
@@ -0,0 +1,25 @@
+<pl-question-panel>
+  <p>
+    The following examples below were created to show the capabilities of the <a href="http://prairielearn.readthedocs.io/en/latest/elements/#pl-dataframe"><code>pl-dataframe</code></a> element, which allows formatting and rendering various Python variables defined in the <code>server.py</code> on a question page.  For example, Pandas DataFrames can be set to a parameter and then rendered on the page a la Jupyter Notebook, or ordinary Python variables can be displayed in text format.
```
coatless, line 3: This feels a bit like left over from the `pl-python-variable` component. I cleaned it a bit.

```suggestion
    The following examples below were created to show the capabilities of the <a href="http://prairielearn.readthedocs.io/en/latest/elements/#pl-dataframe"><code>pl-dataframe</code></a> element, which allows formatting and rendering of pandas data frames, e.g. <a href="https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html"><code>pd.DataFrame()</code>, inside the <code>server.py</code> to be displayed on the question page. 
```

elements/pl-dataframe/pl-dataframe.py
```diff
@@ -0,0 +1,110 @@
...
+    element = lxml.html.fragment_fromstring(element_html)
+    pl.check_attribs(
+        element,
+        required_attribs=["params-name"],
+        optional_attribs=[
+            "show-index",
+            "show-header",
+            "show-dimensions",
+            "show-dtype",
+            "show-code",
```
coatless, line 32: Let's keep in line with the `pl-variable-output` options. So, instead of `show-code`, let's use:

```suggestion
            "show-python",
```

So, in the _odd_ case we want to support MATLAB/Mathematica/R we could add the language tabs in.

elements/pl-dataframe/pl-dataframe.py
```diff
@@ -0,0 +1,110 @@
...
+NO_HIGHLIGHT_DEFAULT = False
+PREFIX_DEFAULT = ""
+SUFFIX_DEFAULT = ""
+TEXT_DEFAULT = False
+SHOW_HEADER_DEFAULT = True
+SHOW_INDEX_DEFAULT = True
+SHOW_DIMENSIONS_DEFAULT = True
+SHOW_DATATYPE_DEFAULT = False
+ADD_LINE_BREAKS_DEFAULT = False
+INTERACTIVE_DEFAULT = True
```
coatless, line 17: I think the interactive DF table was paused due to the external dependency part. This might need to get deleted as a result.

eliotwrobson: Ah this was left in by mistake, will remove.

elements/pl-dataframe/pl-dataframe.py
eliotwrobson: That makes sense, best to use similar attributes to other elements.

elements/pl-dataframe/pl-dataframe.py
```diff
@@ -0,0 +1,104 @@
...
+        # Get headers for all floating point columns and style them to use the desired number of sig figs.
+        float_column_names = frame.select_dtypes(include="float").columns
+
+        # This format string displays the desired number of digits, as given by the instructor
+        # Switches between exponential and decimal notation as needed
+        frame_style.format(subset=float_column_names, formatter=f"{{:.{num_digits}g}}")
+
+    if show_dtype:
+        descriptors = frame.agg([lambda s: s.dtype])
+        descriptors.index = pandas.Index(["dtype"])
```
tdy, line 75:
```suggestion
        descriptors = frame.agg([lambda s: s.dtype]).set_axis(["dtype"])
```

elements/pl-dataframe/pl-dataframe.py
```diff
@@ -0,0 +1,104 @@
+from typing import cast
+
+import chevron
+import lxml.html
+import pandas
```
tdy, line 5: Any reason not to use the ubiquitous `pd` alias?

elements/pl-dataframe/pl-dataframe.py
```diff
@@ -0,0 +1,104 @@
...
+
+    if varname not in data["params"]:
+        raise KeyError(
+            f'Could not find {varname} in params. Please make sure to set params-name="{varname}" in the element.'
+        )
+
+    # Always assume that entry in params dict is serialized dataframe
+    frame = pl.from_json(data["params"][varname])
+
+    if not isinstance(frame, pandas.DataFrame):
```
tdy, line 54:
```suggestion
    if not isinstance(frame, pd.DataFrame):
```

elements/pl-dataframe/pl-dataframe.py
```diff
@@ -0,0 +1,104 @@
...
+        raise ValueError(f"Parameter name '{varname}' does not encode a dataframe.")
+
+    frame = cast(pandas.DataFrame, frame)
```
tdy, line 57:
```suggestion
    frame = cast(pd.DataFrame, frame)
```

elements/pl-dataframe/pl-dataframe.py
tdy, line 5:
```suggestion
import pandas as pd
```

eliotwrobson: Used the full name because that's what was done in the python file this was originally based on. Will use the alias since that seems to be more generally accepted.

elements/pl-dataframe/pl-dataframe.py
```diff
@@ -0,0 +1,105 @@
...
+    if num_digits is not None:
+        # Get headers for all floating point columns and style them to use the desired number of sig figs.
+        float_column_names = frame.select_dtypes(include="float").columns
+
+        # This format string displays the desired number of digits, as given by the instructor
+        # Switches between exponential and decimal notation as needed
+        frame_style.format(subset=float_column_names, formatter=f"{{:.{num_digits}g}}")
+
+    if show_dtype:
+        descriptors: pd.DataFrame = frame.agg([lambda s: s.dtype]).set_axis(
```
coatless, line 74: RIP find & replace

```suggestion
        descriptors = frame.agg([lambda s: s.dtype]).set_axis(
```

elements/pl-dataframe/pl-dataframe.py
```diff
@@ -0,0 +1,105 @@
...
+            ["dtype"], copy=False
+        )
```
coatless, line 76: RIP find & replace and a fix for style added the `axis="index"` argument in [`set_axis()`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.set_axis.html)

```suggestion
        descriptors = frame.agg([lambda s: s.dtype]).set_axis(
            ["dtype"], axis="index", copy=False
        )
```

eliotwrobson: So _that's_ why the type checker wasn't liking this line. Drove me nuts for like a half hour.

CODEOWNERS
```diff
@@ -1,6 +1,7 @@
 /elements/pl-hidden-hints @eliotwrobson
 /elements/pl-order-blocks @SethPoulsen
 /elements/pl-rich-text-editor @jonatanschroeder
+/elements/pl-dataframe @coatless @eliotwrobson
```
nwalters512, line 4: Should we name this `pl-pandas-dataframe` instead? This currently seems to be very closely tied to Pandas, so I'd like to reflect this in the name if this isn't generalizable to _other_ types of dataframes (whatever those may be).

coatless: The goal here would be to move away from being bound to just Python's `pandas` and focus on how best to display tabular data for any language. For example, in the current languages supported in `pl-variable-output`, we have `DataFrame` equivalents of:

- R's [`dataframe()`](https://stat.ethz.ch/R-manual/R-devel/library/base/html/data.frame.html) or [`tibble()`](https://tibble.tidyverse.org/)
- MATLAB's community-led [dataframe approach](https://www.mathworks.com/matlabcentral/fileexchange/95258-dataframe-timeframe)
- Mathematica's [`Dataset`](https://reference.wolfram.com/language/ref/Dataset.html)

If Julia was added, then we would have: 

- Julia's [`DataFrame.jl`](https://dataframes.juliadata.org/stable/)

nwalters512: `pl-variable-output` displays data _in_ different formats, but AFAICT this element is about displaying data _from_ a specific format. That feels like an important distinction. Folks can't write questions in MATLAB or Mathematica or Julia, so how would they get a `DataFrame.jl` to pass to this element in the first place?

exampleCourse/questions/element/dataframe/server.py
```diff
@@ -0,0 +1,64 @@
...
+
+    dft = pd.DataFrame(
+        {
+            # Scalars
+            "integer": 1,
+            "numeric": 3.14,
+            "logical": False,
+            "character": "foo",
+            # TODO adding in complex numbers won't deserialize correctly, fix this (somehow?)
+            # "complex": complex(1, 2),
```
eliotwrobson, line 40: @coatless do you know why the complex numbers won't reserialize correctly? It looks like the output has a dict with real and imaginary parts, but this isn't correctly converted back during the deserialization step (just calls float constructor on this dict).

coatless: @eliotwrobson thanks for the heads up. It looks like Pandas' complex number support has a few hiccups. 

I've raised the issue upstream at: https://github.com/pandas-dev/pandas/issues/50782

That said, I'm not sure if this is a show-stopper or not. In particular, the original `dataframe-v1` is able to be _reconstructed_ with complex numbers but lacks `missing value`/`time` support. Over time, I'm sure that issue can get resolved.

Some experimentations with complex numbers: 

https://colab.research.google.com/github/coatless/colab-notes/blob/main/03-dataframe-json-export-and-import-with-missing-and-date-values-prairielearn.ipynb#scrollTo=lqZFltSTHsAx

CODEOWNERS
coatless: @nwalters512 I think `pl-dataframe` is similar to `pl-variable-output`'s goal of displaying data in different formats. Not necessarily a _specific_ format, e.g. `pandas`. This element has the same limitation as `pl-variable-output`, we need to use _Python_  and its libraries to generate the underlying object. 

For instance, with the incorporation of https://github.com/PrairieLearn/PrairieLearn/issues/6787, we now have the ability to highlight _R_ data types and indices. We can further add re-construction code in _R_ to re-build the data frame or provide a file download option using a language agnostic file format (e.g. [`arrow`](https://arrow.apache.org/docs/python/pandas.html)) containing the data frame.

If you have a few seconds to address external dependency, we can even allow for the _visual_ table to gain interactive sorting features described in https://github.com/PrairieLearn/PrairieLearn/issues/6788 (@eliotwrobson and myself are not sure how to resolve the external dependency).

So, I view this element as being a bit more than just a _pandas_ specific implementation, but more so first-class support for displaying and using tabular data. We're just slowly moving to incorporate other export formats.

docs/elements.md
```diff
@@ -1244,6 +1244,55 @@ As of right now, the element supports displaying either Pandas DataFrames as an
...
+| `params-name`     | string                | —       | The name of the key in `data['params']` to get a value from.                                                                                                            |
+| `show-header`     | boolean               | true    | Show the header row of a DataFrame.                                                                                                                                     |
+| `show-index`      | boolean               | true    | Show the index column of a DataFrame.                                                                                                                                   |
+| `show-dimensions` | boolean               | true    | Show a footer with the dimensions of a DataFrame.                                                                                                                       |
+| `show-dtype`      | "none", "python", "r" | "none"  | Show the data types contained in each column of the DataFrame at the bottom of each column corresponding to the language given, where "none" doesn't display the types. |
```
coatless, line 1278: @eliotwrobson I think we need to keep the `show-dtype` as a bool here for consistency with other options.

How about we introduce maybe `language` that stores `python` or `r` with the default being `python`?

elements/pl-dataframe/pl-dataframe.py
```diff
@@ -0,0 +1,155 @@
...
+        frame_style.set_table_styles(
+            [{"selector": ".foot_row0", "props": "border-top: 1px solid black;"}]
+        )
+        frame_style.concat(other)
+
+    if not show_header:
+        frame_style.hide(axis="columns")
+
+    if not show_index:
+        frame_style.hide()
```
coatless, line 137:
```suggestion
    if show_index:
        # Switch row indices to being 1-based.
        if show_dtype is DtypeLanguage.R:
           frame_style.index = list(range(1, len(frame_style.index) + 1))
    else:
        frame_style.hide()
```

docs/elements.md
eliotwrobson: I think that makes sense, if the language will also affect the indexing. I'll add the parameter then.

exampleCourse/questions/element/dataframe/server.py
eliotwrobson: I don't think this is a show stopper then. If people need complex numbers they can use the old encoding, and we'll pick up any fixes that get deployed in the main Pandas repo too.

elements/pl-dataframe/pl-dataframe.py
```diff
@@ -0,0 +1,166 @@
...
+        )
+        frame_style.concat(other)
+
+    if not show_header:
+        frame_style.hide(axis="columns")
+
+    if show_index:
+        # Switch row indices to being 1-based if supplied indices are integer. Otherwise, leave untouched.
+        # TODO double check if this will cause unexpected behavior if these aren't just indices.
+        if display_language is DisplayLanguage.R and pd.api.types.is_integer_dtype(
```
eliotwrobson, line 143: @coatless I had to change your code slightly for this to work (not sure why the styler didn't like having its index set directly, kinda weird). But I was wondering, won't this cause issues for non-default indexing? Should we instead check for the index being the numbers 0 to n?

elements/pl-dataframe/pl-dataframe.py
```diff
@@ -159,8 +159,7 @@ def render(element_html: str, data: pl.QuestionData) -> str:
     }
 
     if show_dimensions:
-        html_params["num_rows"] = frame.shape[0]
-        html_params["num_cols"] = frame.shape[1]
+        html_params["num_rows"], html_params["num_cols"] = frame.shape
```
coatless, line 162: Huzzah for unpacking!

exampleCourse/questions/element/dataframe/server.py
coatless: Aye, that was my thought as well.

docs/elements.md
coatless: Excellent! I think that'll help out greatly.

elements/pl-dataframe/pl-dataframe.py
coatless: @eliotwrobson thanks for the question. I can understand why the linter wouldn't like the shift to 1-based indices. The other way around it would be to switch away from integer indices to string indices: 

```python
frame_style.index = [str(idx) for idx in range(1, len(frame_style.index) + 1)]
```

With that being said, the index shift is 100% cosmetic. Why it's cosmetic is we're not shifting indices and, then, giving the data frame to another element. Everything is happening in a localized and contained manner. Moreover, we can _rename_ and change indices without worrying about subsetting (under the `.loc` case as we're dealing with labels, the `.iloc` case would still be 0-based as it is positional).

```python
import pandas as pd 

## Instantiate a dataframe 
df = pd.DataFrame(
    [[1, "tall"], [2, "small"], [3, "normal"]],
    columns=["id", "height"]
)

print(df)
#    id  height
# 0   1    tall
# 1   2   small
# 2   3  normal

df.loc[1, :]
# id            2
# height    small
# Name: 1, dtype: object

## Create a copy of the dataframe
df_shift_index = df.copy()

## Shift indices
df_shift_index.index = list(range(1, df.shape[0] + 1))
print(df_shift_index)
#   id  height
# 1   1    tall
# 2   2   small
# 3   3  normal

df_shift_index.loc[1,:]
# id           1
# height    tall
# Name: 1, dtype: object
```

So, no issues should arise as the data is localized to the element itself.

docs/elements.md
```diff
@@ -1244,6 +1244,56 @@ As of right now, the element supports displaying either Pandas DataFrames as an
...
+| `show-dimensions`  | boolean       | true     | Show a footer with the dimensions of a DataFrame.                                                                                                      |
+| `show-dtype`       | boolean       | false    | Show the data types contained in each column of the DataFrame at the bottom of each column. Types used correspond to the `display-language` parameter. |
+| `display-language` | "python", "r" | "python" | Language to use for displaying data types and indices.                                                                                                 |
+| `show-python`      | boolean       | true     | Show code that can be used to recreate the DataFrame in Python in a separate tab.                                                                      |
+| `digits`           | integer       | -        | Number of digits to display for floating point entries.                                                                                                |
+
+#### Details
+
+When setting a parameter, use PrairieLearn's built in `pl.to_json()` on the DataFrame to display.
```
coatless, line 1285: Should there be a separate documentation page somewhere that describes `pl.to_json()` and `pl.from_json()` serialization/unserialization efforts? 

We probably should mention here that `pl.to_json(df, df_encoding_version=2)` allows for missing and date time values whereas `pl.to_json(df, df_encoding_version=1)` (default) does not. However, it has support for complex numbers.

coatless (commented):
I think everything should be good to go for a review by @mwest1066 & @nwalters512. 

I made a few quick comments answering @eliotwrobson your comments.

CODEOWNERS
eliotwrobson: I also don't think it's unreasonable to think we could support different types of Python dataframes in the future, for example, [Polars](https://www.pola.rs/) is becoming more popular.

elements/pl-dataframe/pl-dataframe.py
eliotwrobson: Right, I understand that this code here is making a cosmetic change, the issue I was wondering about is whether, if using a dataframe that has an integer index which is not just numbering the rows from 0 to n-1, we still want to add 1 to all the values. I think we can make a more robust check on what the index values are if necessary though.

coatless: Gotcha, how about adding: 

```python
type(frame_style.index) == pd.RangeIndex
```

where `frame_style.index` should have type `pandas.core.indexes.range.RangeIndex`.

If an index is supplied, then `type(frame_style.index)` should have:

```
pandas.core.indexes.base.Index
```

eliotwrobson: Unfortunately, the re-serialization erases the RangeIndex type and recodes it as an Int64. Maybe we can add a note about this to make sure nobody using R runs into this problem?

docs/elements.md
eliotwrobson: Docs added!

elements/pl-dataframe/pl-dataframe.py
```diff
@@ -0,0 +1,164 @@
...
+            "display-language",
+            "digits",
+        ],
+    )
+
+
+def render(element_html: str, data: pl.QuestionData) -> str:
+    element = lxml.html.fragment_fromstring(element_html)
+    # TODO should there be a different display varname?
+    varname = pl.get_string_attrib(element, "params-name")
```
nwalters512, line 76: Any reason not to just render this as `df = pd.DataFrame(...)`? I recognize there's precedence for reusing the parmas name for a user-facing variable name (e.g. in `pl-matrix-output`), but I don't necessarily agree that that's a good idea. A key in `params` may not necessarily be a valid Python identifier and it also may have no relation to what students should see.

elements/pl-dataframe/pl-dataframe.py
```diff
@@ -0,0 +1,164 @@
...
+
+    display_language = pl.get_enum_attrib(
+        element, "display-language", DisplayLanguage, DISPLAY_LANGUAGE_DEFAULT
+    )
+
+    num_digits = pl.get_integer_attrib(element, "digits", NUM_DIGITS_DEFAULT)
+
+    if varname not in data["params"]:
+        raise KeyError(
+            f'Could not find {varname} in params. Please make sure to set params-name="{varname}" in the element.'
```
nwalters512, line 93: Not sure if this error message is useful. If we reach this point in the code, then `params-name` *must* have been set in the HTML for this element.

elements/pl-dataframe/pl-dataframe.py
```diff
@@ -0,0 +1,164 @@
...
+
+    if show_dtype:
+        if display_language is DisplayLanguage.PYTHON:
+            get_dtype_function = get_pandas_dtype
+        elif display_language is DisplayLanguage.R:
+            get_dtype_function = convert_pandas_dtype_to_r
+        else:
+            assert_never(display_language)
+
+        # old agg function:
```
nwalters512, line 126: What does this comment mean?

elements/pl-dataframe/pl-dataframe.py
```diff
@@ -0,0 +1,164 @@
...
+            [{"selector": ".foot_row0", "props": "border-top: 1px solid black;"}]
+        )
+        frame_style.concat(other)
+
+    if not show_header:
+        frame_style.hide(axis="columns")
+
+    if show_index:
+        # Switch row indices to being 1-based if supplied indices are integer. Otherwise, leave untouched.
+        # TODO double check if this will cause unexpected behavior if these aren't just indices.
```
nwalters512, line 141: 👀 do you intend to resolve this TODO in this PR?

exampleCourse/questions/element/dataframe/info.json
```diff
@@ -0,0 +1,7 @@
+{
+    "uuid": "53af648a-68a3-49c8-b062-b329a2ca7842",
+    "title": "Element pl-dataframe: Display of Pandas Dataframes",
```
nwalters512, line 3: Nit:
```suggestion
    "title": "Element pl-dataframe: Display of Pandas DataFrames",
```

exampleCourse/questions/element/dataframe/question.html
```diff
@@ -0,0 +1,40 @@
+<pl-question-panel>
+  <p>
+    The following examples below were created to show the capabilities of the <a
+      href="http://prairielearn.readthedocs.io/en/latest/elements/#pl-dataframe"><code>pl-dataframe</code></a> element,
+    which allows formatting and rendering of pandas data frames, e.g. <a
```
nwalters512, line 5:
```suggestion
    which allows formatting and rendering of pandas DataFrames, e.g. <a
```

exampleCourse/questions/element/dataframe/question.html
```diff
@@ -0,0 +1,40 @@
...
+      href="https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html"><code>pd.DataFrame()</code></a>, inside
+    the <code>server.py</code> to be displayed on the question page.
+  </p>
+</pl-question-panel>
+
+<div class="card my-2">
+  <div class="card-body">
```
nwalters512, line 12: There's no need for this to be in a card (#2911).

exampleCourse/questions/element/dataframe/server.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+
+    dft = pd.DataFrame(
+        {
+            # Scalars
+            "integer": 1,
+            "numeric": 3.14,
+            "logical": False,
+            "character": "foo",
+            # TODO adding in complex numbers won't deserialize correctly, fix this (somehow?)
```
nwalters512, line 39: 👀 do you plan to address this TODO in this PR?

python/prairielearn.py
```diff
@@ -96,17 +141,28 @@ def all_partial_scores_correct(data: QuestionData) -> bool:
     )
 
 
-def to_json(v):
+def to_json(v, *, df_encoding_version=1):
```
nwalters512, line 144: As with #6783, I have my reservations about this. Is there a reason we can't use the new encoding logic in all scenarios?

elements/pl-dataframe/pl-dataframe.py
eliotwrobson: Yes, turns out I messed up the config file a little bit and the typechecker wasn't running over this file before... Will resolve shortly.

exampleCourse/questions/element/dataframe/server.py
eliotwrobson: Unfortunately, this probably can't be resolved in this PR (see https://github.com/PrairieLearn/PrairieLearn/pull/6800#discussion_r1070474330)

python/prairielearn.py
eliotwrobson: See (https://github.com/PrairieLearn/PrairieLearn/pull/6800#discussion_r1070474330) for discussion, unfortunately there's some issues with the way that Pandas natively encodes complex numbers. Though I understand the desire to use fewer encoding versions for things.

elements/pl-dataframe/pl-dataframe.py
eliotwrobson: Will defer to @coatless on what they think, but always using `df` seems reasonable to me.

elements/pl-dataframe/pl-dataframe.py
eliotwrobson: That's a good point, will remove the extra text.

elements/pl-dataframe/pl-dataframe.py
eliotwrobson: @coatless ?

elements/pl-dataframe/pl-dataframe.py
coatless: I think the `varname` component came into play to mimic being able to set a variable name in `pl-variable-output`, c.f. `<variable params-name=""></variable>` tag:

https://prairielearn.readthedocs.io/en/latest/elements/#pl-variable-output-element

I'm not opposed to defaulting to `df` as the internal variable. 

Though, it would be nice to provide an option to specify a data frame output name. This would help with providing flavor text to the problem. For example:

```
"Please use the `employees` data frame to find the employee with the highest salary" 
```

vs.

```
"Please use the `df` data frame
```

elements/pl-dataframe/pl-dataframe.py
coatless: This was a note that emphasized in the POC that the lambda function originally being used in the `.agg()` could be swapped with the new `convert_pandas_dtype_to_r()` function. The note can be removed.

elements/pl-dataframe/pl-dataframe.py
coatless: @eliotwrobson how about the following to detect non-standard indices:

```python
def is_default_index_type(df):
  return df.index.is_integer()

def is_default_numeric_index_values(df):
  return pd.Index(list(range(0, len(df)))).equals(df.index)

def is_default_index(df):
  return is_default_index_type(df) and is_default_numeric_index_values(df)

# Sample call
is_default_index(df_zero_based)
```

Experiments:

https://colab.research.google.com/github/coatless/colab-notes/blob/main/04-pandas-index-default-checks.ipynb#scrollTo=5eQr9mCSWwBu

exampleCourse/questions/element/dataframe/server.py
coatless: As @eliotwrobson mentioned, this is an upstream pandas bug with the JSON format, c.f.  

https://github.com/pandas-dev/pandas/issues/50782

python/prairielearn.py
coatless: We've opted to go the version route for consistency with how serialization formats change (e.g. `.pkl`). In the odd case that we need to further extend the format, this gives a bit of breathing room while also allowing for graceful backward compatibility to the community who develops custom implementations.

coatless (commented):
@eliotwrobson I prototyped some new code for detecting standard Pandas indices. Can you see if that would work?

elements/pl-dataframe/pl-dataframe.py
eliotwrobson: I'll go ahead and add an option that defaults to `df` then, since it's pretty straightforward.

elements/pl-dataframe/pl-dataframe.py
eliotwrobson: This has now been resolved!

elements/pl-dataframe/pl-dataframe.py
eliotwrobson: Added the ability to change this with the `display-variable-name` attribute.

elements/pl-dataframe/pl-dataframe.py
eliotwrobson: Added this in and seems to work correctly. Thanks!

exampleCourse/questions/element/dataframe/server.py
coatless: There was some movement in: https://github.com/pandas-dev/pandas/issues/50782

Not sure if the person responding, @dicristina, is an official member of pandas core; but, they confirmed that the `"table"` orientation being employed is missing handling code for complex numbers.
  
> Unlike in the previous case in the `orient="table"` case we have the data type of each column so in theory we should be able to reconstruct the values without doing any work at all. The problem here is that there is no special handling for complex numbers and when the values are read they are passed to the `float` function. The relevant code is in `pandas/io/json/_table_schema.py`.

docs/elements.md
```diff
@@ -1288,6 +1288,57 @@ As of right now, the element supports displaying either Pandas DataFrames as an
...
+| `display-variable-name` | string        | "df"     | Variable name to display in code to recreate DataFrame.                                                                                                |
+| `show-python`           | boolean       | true     | Show code that can be used to recreate the DataFrame in Python in a separate tab.                                                                      |
+| `digits`                | integer       | -        | Number of digits to display for floating point entries.                                                                                                |
+
+#### Details
+
+When setting a parameter, use PrairieLearn's built in `pl.to_json()` on the DataFrame to display. Note that there are multiple serialization options for Pandas DataFrames. Encoding a DataFrame `df` by setting `pl.to_json(df, df_encoding_version=2)` allows for missing and date time values whereas `pl.to_json(df, df_encoding_version=1)` (default) does not. However, it `df_encoding_version=1` has support for complex numbers, while `df_encoding_version=2` does not.
```
nwalters512, line 1330:
```suggestion
When setting a parameter, use PrairieLearn's built in `pl.to_json()` on the DataFrame to display. Note that there are multiple serialization options for Pandas DataFrames. Encoding a DataFrame `df` by setting `pl.to_json(df, df_encoding_version=2)` allows for missing and date time values whereas `pl.to_json(df, df_encoding_version=1)` (default) does not. However, `df_encoding_version=1` has support for complex numbers, while `df_encoding_version=2` does not.
```

nwalters512 (approved):

exampleCourse/questions/element/dataframe/server.py
nwalters512: Nice! Seems like it would be relatively straightforward to add a `"complex"` type to go alongside `integer` and `number`.

python/prairielearn.py
eliotwrobson: @nwalters512 I think that's perfectly reasonable, maybe as part of the question writing documentation? I think that we can could include those changes in https://github.com/PrairieLearn/PrairieLearn/pull/6783, since that will need a rebase on top of this anyway, and the code changes there are very short.

docs/elements.md
nwalters512: FWIW, I filed https://github.com/PrairieLearn/PrairieLearn/issues/6965 to actually put these into the user-facing docs.

Issue #4857, opened 2021-10-01
Groups page renders page from POST instead of redirecting


jonatanschroeder:
Most pages, after a POST is completed, will redirect to a GET request. This makes sense since, if the page is reloaded for some reason (refresh, browser closes/reopens, etc.), you want the current view of the page, you don't want to re-take the POST action. This is not done in the instructor assessment groups page. In there the POST actions actually render the same page again from the POST itself. It seems this is done to retain any error messages to be shown on the page in case of errors, but that can be achieved through other means (e.g., query strings).

nwalters512:
This was fixed in #7905.

PR #6801, opened 2022-12-31
chore: Lint rule for no-floating-promise

jonatanschroeder:
It's very easy to incur in race conditions by forgetting an await when calling an async function. This PR installs the no-floating-promise rule in lint, which forces async function calls to be handled either with an await, or with a return, or with a `.then()`.

Initial draft version, some of the code may need to be handled in different ways.


.eslintrc.js
```diff
@@ -48,5 +50,12 @@ module.exports = {
         mocha: true,
       },
     },
+    {
+      files: ['assets/scripts/*.js'],
+      env: {
+        browser: true,
+        jquery: true,
+      },
+    },
```
jonatanschroeder, line 59: I added this to the rubrics PR, so it ended up here as well in the mix. Since it's there, may as well piggyback it here.

pages/instructorAssessmentGroups/instructorAssessmentGroups.js
```diff
@@ -198,8 +199,9 @@ router.post('/', function (req, res, next) {
         res.locals.errormsg +=
           'Failed to remove ' + failedUids + '. Please check if the uid exist.\n';
       }
+    })().then(() => {
       obtainInfo(req, res, next);
-    })();
+    });
```
jonatanschroeder, line 204: This page is already a mess, and ideally should be rewritten from scratch (check #4857 among other issues). I updated it only enough to make it work.

lib/group-update.js
```diff
@@ -331,7 +331,7 @@ module.exports = {
                   logger.error('Error while creating groups', err);
                   serverJobs.failJobSequence(job_sequence_id);
                 }
-              })();
+              })().then(() => {});
```
jonatanschroeder, line 334: This one can be replaced with a `tryLockAsync`, I'm working on it.

tests/localLock.test.js
```diff
@@ -85,7 +85,7 @@ describe('local locks', function () {
                 resolve();
               }
             }, 100);
-          })();
+          })().then(() => {});
```
jonatanschroeder, line 88: I'm not that familiar with this particular code, so I'd appreciate a second set of eyes here.

workspace_host/interface.js
```diff
@@ -367,11 +367,11 @@ async.series(
       });
     },
   ],
-  function (err, data) {
+  async function (err, data) {
     if (err) {
       Sentry.captureException(err);
       logger.error('Error initializing workspace host:', err, data);
-      markSelfUnhealthy(err);
+      await markSelfUnhealthy(err);
```
nwalters512, line 374: This function will never throw, so I'm inclined to just disable the rule here. If you do indeed want the await, I don't think it's valid to pass an `async` function as the last argument to `async.series`, so you could write it like this instead, as `async.series` will return a Promise if no callback is provided:

```js
async.series([
  // ...
  async () => logger.info('Workspace host ready'),
]).catch(async (err) => {
  Sentry.captureException(err);
  logger.error('Error initializing workspace host:', err, data);
  await markSelfUnhealthy(err);
});
```

nwalters512 (approved):

Issue #4260, opened 2021-05-27
depluralize "1 significant figures" in pl-number-input help text


cbm755:
Here's a minor papercut issue:

![Screenshot from 2021-05-24 09-29-55](https://user-images.githubusercontent.com/818622/119757992-9faa3c00-be5a-11eb-971c-0ed18b20899e.png)

pl-number-input and friends do that too.

I looked as far as finding the `.moustache` file and the `shortformat` flag.  Maybe the word `figures` could be taken out of the `.moustache` file and the appropriate word put into  into say `info_params["figures_word"]` or somesuch.  Does that sound right?  I might try this but will obviously not be offended if someone does it first!

Issue #2344, opened 2020-04-05
Element inputs should have autocomplete="off"


mwest1066:
All answer inputs should have `autocomplete="off"` to stop them suggesting previous values.

See https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

eliotwrobson:
This has now been implemented for all of the most commonly used input elements. Follow-up issues will be created for refactoring elements that are not as actively maintained.

PR #6854, opened 2023-01-03
Depluralize significant figures in pl-number-input

eliotwrobson:
See description. Resolves https://github.com/PrairieLearn/PrairieLearn/issues/4260 and https://github.com/PrairieLearn/PrairieLearn/issues/2344 for the number input element.

Screenshot:
Screenshot_20230103_175205

### DIFF
--- a/elements/pl-number-input/pl-number-input.mustache
+++ b/elements/pl-number-input/pl-number-input.mustache
@@ -16,7 +16,7 @@
             <span class="input-group-text">{{{label}}}</span>
         </span>
         {{/label}}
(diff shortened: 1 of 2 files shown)


jonatanschroeder:
Do you want to piggyback this and fix the wrong parameter example in here?
https://github.com/PrairieLearn/PrairieLearn/blob/52a0642f790947fb92e5a4cdd42592864e4fba04/exampleCourse/questions/element/numberInput/question.html#L34

It shows `sigfigs` instead of `sigfig`.

jonatanschroeder (approved):

Issue #6523, opened 2022-10-21
Use streaming CSV generation


nwalters512:
Currently, our internal [`nonblocking-csv-stringify,js`](https://github.com/PrairieLearn/PrairieLearn/blob/968fd86821e056d1ddd02c231cfcf483267a7c9b/lib/nonblocking-csv-stringify.js) lib that we use to generate CSVs is actually buffering the entire thing in memory. This causes the process to run out of memory and crash when someone tries to export a CSV that's too large, as the size of the CSV exceeds the max size of a Node string (~1GB).

To fix this, we should pipe the `csv-stringify` stream directly to the Express response. This will avoid the need to construct massive strings in memory.

We may want to look into using a cursor to read from the database (https://node-postgres.com/api/cursor).

Making these changes should allow us to close https://github.com/PrairieLearn/PrairieLearn/issues/1155.

PR #6855, opened 2023-01-04
Writing csv-stringified data to http res stream instead of buffering

potatokim:
#### Issue
#6523 

#### Description
Partially addressed #6523 by piping `stringifier` stream in `lib/nonblocking-csv-stringify` directly to the Express response. This should fix the issue of buffered csv strings hitting the 1GB Node string limit. 
TODO after this PR: use Cursor or pagination when reading from db.

#### Tests
- New sanity unit tests for `lib/nonblocking-csv-stringify` and the dependent `lib/csv-maker` to ensure existing functionalities weren't broken
- Manual tests
- Existing [`tests/instructorAssessmentDownloads.test.js`](https://github.com/PrairieLearn/PrairieLearn/blob/master/tests/instructorAssessmentDownloads.test.js) should already cover the changes too

### DIFF
--- a/lib/nonblocking-csv-stringify.js
+++ b/lib/nonblocking-csv-stringify.js
@@ -30,18 +30,18 @@ function nonblockingStringify(data, callback, options) {
     }
     loop();
   });
-  let chunks = [];
+  
   stringifier.on('readable', function () {
     let chunk;
     while ((chunk = stringifier.read()) !== null) {
-      chunks.push(chunk);
+      callback(null, chunk);
     }
   });
   stringifier.on('error', function (err) {
     return callback(err);
   });
   stringifier.on('end', function () {
-    return callback(null, chunks.join(''));
+    return callback(null, null);
   });
 }
 
--- a/lib/nonblocking-csv-stringify.js
+++ b/lib/nonblocking-csv-stringify.js
@@ -30,7 +30,7 @@ function nonblockingStringify(data, callback, options) {
     }
     loop();
   });
-  
+
   stringifier.on('readable', function () {
     let chunk;
     while ((chunk = stringifier.read()) !== null) {
@@ -45,13 +45,18 @@ function nonblockingStringify(data, callback, options) {
   });
 }
 
-function nonblockingStringifyAsync(data, options) {
+function nonblockingStringifyAsync(data, callback, options) {
   return new Promise((resolve, reject) => {
     nonblockingStringify(
       data,
-      (err, csv) => {
-        if (err) return reject(err);
-        return resolve(csv);
+      (err, chunk) => {
+        if (err) {
+          return reject(err);
+        } else if (chunk) {
+          callback(chunk);
+        } else {
+          return resolve();
+        }
       },
       options
     );
--- a/pages/instructorAssessmentInstance/instructorAssessmentInstance.js
+++ b/pages/instructorAssessmentInstance/instructorAssessmentInstance.js
@@ -92,13 +92,13 @@ router.get('/:filename', (req, res, next) => {
         ];
       });
       csvData.splice(0, 0, csvHeaders);
-      csvStringify(csvData, (err, csv) => {
+      res.attachment(req.params.filename);
+      csvStringify(csvData, function (err, chunk) {
         if (err) {
           throw Error('Error formatting CSV', err);
-        } else if (csv) {
-          res.write(csv);
+        } else if (chunk) {
+          res.write(chunk);
         } else {
-          res.attachment(req.params.filename);
           res.end();
         }
       });
--- a/pages/instructorAssessmentInstance/instructorAssessmentInstance.js
+++ b/pages/instructorAssessmentInstance/instructorAssessmentInstance.js
@@ -93,9 +93,14 @@ router.get('/:filename', (req, res, next) => {
       });
       csvData.splice(0, 0, csvHeaders);
       csvStringify(csvData, (err, csv) => {
-        if (err) throw Error('Error formatting CSV', err);
-        res.attachment(req.params.filename);
-        res.send(csv);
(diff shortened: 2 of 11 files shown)


potatokim:

potatokim:
recheck

eliotwrobson:
@potatokim is this addressing https://github.com/PrairieLearn/PrairieLearn/issues/6523? If so, you should add "Resolves https://github.com/PrairieLearn/PrairieLearn/issues/6523" to you initial comment. That way, when this is merged, the issue will be closed automatically.

pages/instructorAssessmentInstance/instructorAssessmentInstance.js
```diff
@@ -93,9 +93,14 @@ router.get('/:filename', (req, res, next) => {
...
       csvStringify(csvData, (err, csv) => {
-        if (err) throw Error('Error formatting CSV', err);
-        res.attachment(req.params.filename);
-        res.send(csv);
+        if (err) {
+          throw Error('Error formatting CSV', err);
+        } else if (csv) {
+          res.write(csv);
+        } else {
+          res.attachment(req.params.filename);
```
jonatanschroeder, line 101: My understanding is that `res.attachment` needs to be sent before any write, since (to the best of my knowledge) if affects the HTTP headers.

jonatanschroeder:
> @potatokim is this addressing #6523? If so, you should add "Resolves #6523" to you initial comment. That way, when this is merged, the issue will be closed automatically.

I don't think that this is a full resolution to that issue. This still requires the entire database query to be completely loaded before the data is sent back. It is a great step in the right direction, but further work needs to be done before that issue is resolved.

pages/instructorAssessmentInstance/instructorAssessmentInstance.js
potatokim: You're right! I just caught this with some quick test.

potatokim:
> > @potatokim is this addressing #6523? If so, you should add "Resolves #6523" to you initial comment. That way, when this is merged, the issue will be closed automatically.
> 
> I don't think that this is a full resolution to that issue. This still requires the entire database query to be completely loaded before the data is sent back. It is a great step in the right direction, but further work needs to be done before that issue is resolved.

I'm looking to address #6523, or at least the first part of it ie. piping `stringifier` stream directly to the Express response, before jumping to the second part of using `Cursor` to read from the db. If I understand correctly, part 1 should solve the case of hitting the 1GB Node string limit?

Sorry for the lack of PR description. I'll add it when this draft PR is getting more complete. For now, I'm stuck on how to rewriting the async version of `nonblockingStringify`. Do you have any pointer?
https://github.com/PrairieLearn/PrairieLearn/blob/52a0642f790947fb92e5a4cdd42592864e4fba04/lib/nonblocking-csv-stringify.js#L48-L59
If I were to move `res.attachment` above `res.write`, this current callback version works but it requires access to `res`, which I don't think `nonblockingStringifyAsyns` has.

jonatanschroeder:
> For now, I'm stuck on how to rewriting the async version of `nonblockingStringify`. Do you have any pointer?

One option for the async version would be to have a callback that is called only for partial results, and then only resolve when you have complete results. This way you could have something like this in the call:

```js
try {
  res.attachment(...);
  await nonblockingStringifyAsync(csvData, (chunk) => { res.write(chunk); });
  // or: await nonblockingStringifyAsync(csvData, res.write);
  res.end();
} catch (err) {
  ...
}
```

Another option, if you want to avoid callbacks, is to use [generators and the yield operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators), and [a `for await` loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncGenerator).

jonatanschroeder:
This may be both a good example in PL of using yield and a good lead for paginating the query (an alternative to using cursors).

https://github.com/PrairieLearn/PrairieLearn/blob/52a0642f790947fb92e5a4cdd42592864e4fba04/lib/paginate.js#L59-L83

potatokim:
Generator looks elegant but I went with callback. I tried some variations but none worked, particularly placing `yield chunk` where `chunks.push(chunk)` is here:
https://github.com/PrairieLearn/PrairieLearn/blob/968fd86821e056d1ddd02c231cfcf483267a7c9b/lib/nonblocking-csv-stringify.js#L34-L39

lib/nonblocking-csv-stringify.js
```diff
@@ -30,28 +30,33 @@ function nonblockingStringify(data, callback, options) {
     }
     loop();
   });
-  let chunks = [];
+
   stringifier.on('readable', function () {
     let chunk;
     while ((chunk = stringifier.read()) !== null) {
-      chunks.push(chunk);
+      callback(null, chunk);
```
nwalters512, line 37: Rather than relying on the consumer needing to use a callback, did you consider relying on the fact that `stringifier` follows the standard Node Stream API? That, is if this function returned `stringifier` itself, we could pipe it directly to an Express response:

```js
// In an Express handler:
const stringifier = nonblockingStringify(data, options);
stringifier.pipe(res)
```

In fact, at that point, `nonblockingStringify` doesn't actually do anything useful, so we can just use `csv-stringify` directly:

```js
const { Stringifier } = require('csv-stringify');

app.get('/', (res) => {
  const data = getDataSomehow();
  const stringifier = new Stringifier(data);
  stringifier.pipe(res);
});

nwalters512:
@potatokim do you intend to see this PR through? If not, I'm going to close it and implement this change in another PR.

jonatanschroeder:
Closing this PR in favour of #7284.

PR #6856, opened 2023-01-04
Remove unused opsbot code

nwalters512:
This PR removes some outdated code and terminology from our opsbot utils.

### DIFF
--- a/lib/opsbot.js
+++ b/lib/opsbot.js
@@ -1,3 +1,4 @@
+// @ts-check
 const ERR = require('async-stacktrace');
 const request = require('request');
 const util = require('util');
@@ -48,8 +49,7 @@ module.exports.sendMessageAsync = util.promisify((msg, callback) =>
  * Called with callback(err, response, body)
  */
 module.exports.sendSlackAlmaMessage = (msg, channel, options, callback) => {
-  // Fall-back to the old token config if it exists
-  const token = config.secretSlackToken || config.secretSlackProctorToken;
+  const token = config.secretSlackToken;
 
   // Log the message if there's no token specified
   if (!token || !channel) {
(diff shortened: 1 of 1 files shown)


lib/opsbot.js
```diff
@@ -84,45 +84,28 @@ module.exports.sendSlackAlmaMessage = (msg, channel, options, callback) => {
   });
 };
 
-module.exports.sendSlackAlmaMessageAsync = util.promisify((msg, channel, options, callback) =>
-  module.exports.sendSlackAlmaMessage(msg, (err, res, body) => callback(err, { res, body }))
```
nwalters512, line 88: This would have never worked to begin with - it's missing two arguments (channel and options).

lib/opsbot.js
```diff
@@ -47,13 +48,12 @@ module.exports.sendMessageAsync = util.promisify((msg, callback) =>
  * @param callback Function that is called after the message is sent.
  * Called with callback(err, response, body)
  */
-module.exports.sendSlackAlmaMessage = (msg, channel, options, callback) => {
-  // Fall-back to the old token config if it exists
-  const token = config.secretSlackToken || config.secretSlackProctorToken;
+module.exports.sendSlackMessage = (msg, channel, options, callback) => {
```
nwalters512, line 51: The word "Alma" is a remnant of the early days of this Slack code, when it wrote messages as a user named "Alma". It's not necessary to include that in function names.

lib/opsbot.js
```diff
@@ -84,45 +84,28 @@ module.exports.sendSlackAlmaMessage = (msg, channel, options, callback) => {
   });
 };
 
-module.exports.sendSlackAlmaMessageAsync = util.promisify((msg, channel, options, callback) =>
```
nwalters512, line 87: All these `*Async` functions are unused.

lib/opsbot.js
nicknytko: Checked the git blame to see who put this in...

https://pbs.twimg.com/media/FEz4HC5XsAMztm3?format=jpg&name=large

nicknytko (approved):
looks generally fine to me

mwest1066 (approved):

Issue #6792, opened 2022-12-24
More MathJax issues with disappearing lines in CHTML rendering mode


echuber2:
Previously, some flags in `pages/partials/mathjax.ejs` were set to ensure MathJax would use CHTML rendering on Windows instead of SVG, due to some issues where thin horizontal or vertical lines would disappear. This caused problems with glyphs like $\lceil x\rceil$ vs. $| x |$, $\leq$ vs. $\lt$, overlines like $\overline{x}$, etc. (https://github.com/PrairieLearn/PrairieLearn/issues/3086)

Now, as reported by Audrey Lin on Slack, similar problems are occurring again for Windows users **even with CHTML mode**. Some overlines and underlines disappear at different font sizes (e.g. display style vs. inline or with `\LARGE`). Zooming in or out with the browser controls can affect which lines appear or disappear. Although Windows is currently the only platform where PL renders MathJax with CHTML instead of SVG, if I force CHTML rendering on Linux Firefox, I can reproduce the issue there too. So, I think the underlying issue may be with the fonts MathJax uses for CHTML (perhaps [font hinting](https://en.wikipedia.org/wiki/Font_hinting) problems).

I'm drafting a PR that may address this issue.

Here are a few specimens from a 1080p screen. The bugs are marked in red boxes.

# Windows Chrome (CHTML by default)

Missing overlines and/or underlines on Windows Chrome at 100% zoom, 1080p: (Windows Edge also exhibits the same.)
![win chrome bug 1](https://user-images.githubusercontent.com/5503940/209420049-e471bc99-43b0-4fd3-9162-77aba63b41e8.png)
![win chrome bug 2](https://user-images.githubusercontent.com/5503940/209420052-14d91653-08cc-4e2c-b62b-824f14ee4ac9.png)

# Windows Firefox (CHTML by default)

Missing lines on Windows Firefox at 80% zoom, 1080p:
![win ff REAL 80pct bug 1](https://user-images.githubusercontent.com/5503940/209420056-3dd44ce5-5f3c-4a75-b206-a292800ef686.png)
![win ff REAL 80pct bug 2](https://user-images.githubusercontent.com/5503940/209420059-1ebdf603-c281-4ee4-b976-701226ab544f.png)

# Linux Firefox (CHTML mode forced)

Missing underlines on the $\overline{\underline{x}}$ glyph here on Linux Firefox with CHTML mode forced, 90% zoom, 1080p:
[...]

PR #6881, opened 2023-01-10
fix: Use appropriate MathJax conversion in pl-rich-text-editor

### LINKED ISSUE #6880: pl-rich-text-editor: Math does not get inserted in Safari
Reported by a student: The math/formula button does not insert anything when used in Safari. The student reported that this works in Chrome. I have tried this in Firefox (works) and Safari (does not work).

Screen Shot 2023-01-09 at 15 13 48

Safari, blank space highlighted:
Screen Shot 2023-01-09 at 15 14 37

Expected, in Firefox:
Screenshot 2023-01-09 at 15-15-38 admin_reflection1 (Preview) - CAPP 30271 wi23 PrairieLearn

jonatanschroeder:
Resolves #6880. Uses the appropriate format of conversion for MathJax depending on which output is currently in use.

I don't have access to Safari to test this there, but by changing manually the format in pages/partials/mathjax.js I was able to reproduce the original error and make it run properly there. I'd appreciate if someone can actually test it in Safari.

Also, keeping in mind #6792, this may be reviewed soon as well, so it may be good to ensure that any changes that address that also test this functionality.


elements/pl-rich-text-editor/pl-rich-text-editor.js
```diff
@@ -63,7 +63,10 @@ class MathFormula extends Embed {
 
   static updateNode(node, value) {
     MathJax.startup.promise.then(async () => {
-      let html = await MathJax.tex2chtmlPromise(value);
+      let html =
+        outputComponent === 'output/chtml'
```
nwalters512, line 67: I reallllly don't want to rely on a bespoke global variable here. Could we instead inspect MathJax's configuration? `MathJax` is at least a standard and well-known global, and it looks like `MathJax.config.loader.load` can be checked for `output/chtml` or `output/svg`.

jonatanschroeder: Sure, I can do that. I realize this was completed in a rush when I was tired, so it can certainly use a bit more care.

echuber2:
How is CHTML on Safari (vs. SVG)? I'm going to need other people's help checking some rendering stuff soon.

elements/pl-rich-text-editor/pl-rich-text-editor.js
jonatanschroeder: Resolved in 4c7a53c in a more elegant way, using the existence of the function itself as a signal.

elements/pl-rich-text-editor/pl-rich-text-editor.js
```diff
@@ -63,8 +63,8 @@ class MathFormula extends Embed {
 
   static updateNode(node, value) {
     MathJax.startup.promise.then(async () => {
-      let html = await MathJax.tex2chtmlPromise(value);
-      let formatted = html.innerHTML;
+      const html = await (MathJax.tex2chtmlPromise || MathJax.tex2svgPromise)(value);
```
nwalters512, line 66: Very nice!

nwalters512 (approved):

PR #6882, opened 2023-01-10
Upgrade to Yarn 3

nwalters512:
This PR upgrades for Yarn v1 to v3. The primary motivation is a version of `yarn upgrade-interactive` that actually updates version ranges in `package.json` files, but I believe this may also help with some of the package installation timeout issues that we've seen in our ARM Docker image builds in CI.


packages/bind-mount/package.json
```diff
@@ -5,7 +5,8 @@
   "types": "index.d.ts",
   "dependencies": {
     "bindings": "^1.5.0",
-    "nan": "^2.17.0"
+    "nan": "^2.17.0",
+    "node-gyp": "^9.3.1"
```
nwalters512, line 9: This was previously an implicit dependency, but Yarn now requires it to be explicit. This is a best practice anyways.

.yarnrc.yml
```diff
@@ -0,0 +1,7 @@
+nodeLinker: node-modules
```
nwalters512, line 1: This keeps Yarn's behavior compatible with the `node_modules` layout used by yarn v1 and all versions of npm.

.yarnrc.yml
```diff
@@ -0,0 +1,7 @@
...
+
+plugins:
+  - path: .yarn/plugins/@yarnpkg/plugin-interactive-tools.cjs
+    spec: "@yarnpkg/plugin-interactive-tools"
```
nwalters512, line 5: This makes `yarn upgrade-interactive` available.

mwest1066 (approved):
Works for me!

PR #6884, opened 2023-01-11
R-autograder multiple tests per file

barbehenna:
Updating the merge of file documentation and file tests to support multiple tests per file. A corresponding PR is made in to the pl-r-helpers repository to help avoid errors (https://github.com/PrairieLearn/pl-r-helpers/pull/1).


nwalters512:
@eddelbuettel or @coatless, could one of you give this a 👍/👎 as an R expert? Would one/both of you be willing to volunteer to be added as [`CODEOWNERS`](https://github.com/PrairieLearn/PrairieLearn/blob/master/CODEOWNERS) of the R grader?

coatless:
@nwalters512 I would probably add @barbehenna, @eddelbuettel . You can add me if you want. But, both Alton and Dirk should be listed.

On the topic of multi-test, my only comment/issue is the use of a naive order joining over the merge. I've been burned before in that aspect. Any way a numbered index alongside testfile id could be incorporated instead?

barbehenna:
@coatless I totally agree on the risk of a naive order joining over a merge. We were actually doing the merge by file (assuming one test per file) before this PR. Unfortunately, we currently lack an easy/clean way to tag both the question details and the test results with a unique ID. I have a few ideas and would be happy to talk more about them, but I this PR (along with https://github.com/PrairieLearn/pl-r-helpers/pull/1), should avoid merging issues.

barbehenna:
I should add, we _do_ have the ability to manually add extra tags to the test results, but that approach will likely introduce overhead and redundancy for the question writer to deal with.

daviddalpiaz:
Since this feature request came from me, I did some testing and it looks good to me. I'd suggest merging, but I think @eddelbuettel should probably give it a look first.

There isn't really a rush to merge this because nothing currently depends on it, but we will likely use it as soon as it is available. (Beyond the quality-of-life benefits to question authors, this can also reduce the amount of compute used by the R autograder, assuming new questions utilize this strategy.)

eddelbuettel:
I appear to have been a little removed from the discussion(s) motivating and leading this, and I should not be one holding up merges in PL.  

But if it works with our autograder for R (and there is no reason to think it won't) it's fine by me (as in: creates no headaches for me).

daviddalpiaz:
In that case, I recommend merging this!

nwalters512 (approved):

coatless:
@barbehenna thanks for the description! I raised the point because I saw the `LC_COLLATE` option get set to `NA`. 

Glancing at [`get_question_details()`](https://github.com/PrairieLearn/pl-r-helpers/pull/1/files#diff-6711694efbe7bd938c308000af5375141f3340f4b6f066802c5528a9486adc61R17-R25) in the `pl-r-helpers` package, could we potentially structure the question data frame to include a sequence:

```r
        # Add a check to ensure balanced amounts of `title` and `scores` usage:
        stopifnot(paste0("Unbalanced amount of `title` vs `scores` in file:", f) = title == score)
        data.frame(
                   name=title,
                   file=f,
                   max_points=as.numeric(score),
                   position = seq_along(title), # create an index per each test entry, e.g. 1:n
                   stringsAsFactors = FALSE
       )
```

I think from there, we could potential have a group by option in `test_results`.

PR #6896, opened 2023-01-11
Admin query: list instructors by institution

trombonekenny:
New query to see which `uids` have course Editor or Owner access. Includes course instance aggregate information too.


admin_queries/instructors_by_institution.sql
```diff
@@ -0,0 +1,24 @@
+SELECT
+    i.short_name AS institution,
+    u.uid, u.name, plc.short_name AS course, plc.id AS course_id, cp.course_role,
```
mwest1066, line 3: Nit: write each of these on separate lines

mwest1066: I also suggest reordering the columns to be: institution, course, course_id, uid, name, ...

This will make the leading columns the ones that are used for GROUP BY, which I find clearer to understand.

mwest1066 (approved):

PR #6931, opened 2023-01-11
Enable lint checks for files in /python directory

eliotwrobson:
See title. Minor changes to files in `python` directory so that linting can be enabled.


pyproject.toml
```diff
@@ -1,7 +1,14 @@
 [tool.flake8]
 # TODO add python directory here in later PR
```
nwalters512, line 2: TODO should be resolved now.
```suggestion
```

python/zygote.py
```diff
@@ -68,8 +69,8 @@
 def try_dumps(obj, sort_keys=False, allow_nan=False):
     try:
         return json.dumps(obj, sort_keys=sort_keys, allow_nan=allow_nan)
-    except:
-        print("Error converting this object to json:\n{:s}\n".format(str(obj)))
+    except TypeError:
```
eliotwrobson, line 72: Although this is a "less general" exception catch, this function just prints the error and re-raises an exception, so nothing should break from this (not that I expect it to anyway).

nwalters512 (approved):

python/zygote.py
nwalters512: I think I'd rather keep this as general as possible so that *any* serialization error results in the object being displayed. I think that's pretty crucial for being able to debug errors.

eliotwrobson: Done!

nwalters512 (approved):

PR #6937, opened 2023-01-12
style: Enable linting for grader scripts

eliotwrobson:
Add linting to grader scripts with changes. May have introduced issues, so should review _carefully_ before merging.


graders/python/python_autograder/pl_main.py
```diff
@@ -135,7 +134,7 @@ def add_files(results):
 
         with open(output_fname, mode="w", encoding="utf-8") as out:
             json.dump(grading_result, out)
-    except:
+    except:  # noqa: E722
```
eliotwrobson, line 137: Is the bare except needed here? Not sure if it is intended to catch system interrupts.

graders/c/cgrader/cgrader.py
```diff
@@ -17,10 +15,10 @@
...
     "\n\nTIMEOUT! Typically this means the program took too long,"
-    + "\nrequested more inputs than provided, or an infinite loop was found."
-    + "\nIf your program is reading data using scanf inside a loop, this "
-    + "\ncould also mean that scanf does not support the input provided "
-    + "\n(e.g., reading an int if the input is a double).\n"
+    "\nrequested more inputs than provided, or an infinite loop was found."
+    "\nIf your program is reading data using scanf inside a loop, this "
+    "\ncould also mean that scanf does not support the input provided "
+    "\n(e.g., reading an int if the input is a double).\n"
 )
```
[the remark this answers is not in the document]
eliotwrobson: It probably is. No need to include the \n's

graders/c/cgrader/cgrader.py
```diff
@@ -115,7 +113,7 @@ def compile_file(
         std_obj_files = []
         objs = []
         for std_c_file in c_file if isinstance(c_file, list) else [c_file]:
-            obj_file = re.sub("\.[^.]*$", "", std_c_file) + ".o"
+            obj_file = re.sub(r"\.[^.]*$", "", std_c_file) + ".o"
```
coatless, line 116: Isn't using the `r` triggering a raw encoding of the string making the `\` added in the pattern string redundant?

eliotwrobson: I think that the `\.` is supposed to be part of the pattern (not escaping out the period). The issue is that the linter was complaining about `\.` being an invalid escape sequence.

jonatanschroeder: Actually the raw encoding is necessary (and the original should have had a `\\`, my mistake). In this case the regular expression is looking for a literal `.`, not its interpretation as "any character".

graders/c/cgrader/cgrader.py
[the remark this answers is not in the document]
eliotwrobson: @jonatanschroeder Switched to triple quotes in latest push

graders/c/cgrader/cgrader.py
```diff
@@ -5,23 +5,21 @@
...
 
-TIMEOUT_MESSAGE = (
-    "\n\nTIMEOUT! Typically this means the program took too long,"
-    + "\nrequested more inputs than provided, or an infinite loop was found."
-    + "\nIf your program is reading data using scanf inside a loop, this "
-    + "\ncould also mean that scanf does not support the input provided "
-    + "\n(e.g., reading an int if the input is a double).\n"
-)
+TIMEOUT_MESSAGE = """
+TIMEOUT! Typically this means the program took too long,
```
jonatanschroeder, line 17: The original had two line breaks at the start. There is a reason for this: if there is a message printed before timeout happens, you want a space between the message and the timeout error.
```suggestion
TIMEOUT_MESSAGE = """

TIMEOUT! Typically this means the program took too long,
```

graders/python/python_autograder/pl_main.py
nicknytko: Off the top of my head, I'm actually not sure how often this branch gets taken, if at all.  I'd be inclined to leave it in for now just in case something breaks inside of the grader.

nicknytko (approved):
This looks fine on the Python side

graders/python/python_autograder/pl_main.py
eliotwrobson: Well the issue isn't quite whether this branch gets taken; we would change this to `except Exception`, since otherwise the bare `except` this will catch system interrupts (like a user hitting Ctrl+C). This is why this is a linter issue rather than a style one, see [here](https://www.flake8rules.com/rules/E722.html) for more information.

That was what I was getting at with the original question, it doesn't seem like this code needs to catch signals from the OS to get errors from the grader code.

eliotwrobson: Looking at this again, I think it's _probably_ not going to cause issues to leave in this bare except (since all that will happen is marking a grade from the autograder before exiting), and it's probably best to limit this PR to changes just to make the linter pass.

Taking a look through the autograder code, there's enough little hiccups that I think it would be worth doing a style pass to enable typechecking, and we can revisit this exception then.

PR #6939, opened 2023-01-12
Fix chunk autoscaling metric queries

nwalters512:
Metrics have the dimension `Server Group`, not `ServerGroup`. For consistency, I also updated the dimension for the of the metrics that chunk autoscaling emits. This will require us to update CloudWatch dashboards once this is deployed.

### DIFF
--- a/ee/cron/chunksHostAutoScaling.js
+++ b/ee/cron/chunksHostAutoScaling.js
@@ -36,7 +36,7 @@ module.exports.run = callbackify(async () => {
               MetricName: 'PageViewsPerSecond',
               Dimensions: [
                 {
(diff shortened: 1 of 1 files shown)


ee/cron/chunksHostAutoScaling.js
```diff
@@ -36,7 +36,7 @@ module.exports.run = callbackify(async () => {
               MetricName: 'PageViewsPerSecond',
               Dimensions: [
                 {
-                  Name: 'ServerGroup',
+                  Name: 'Server Group',
```
mwest1066, line 39: Oh yuck! Good find.

mwest1066 (approved):

PR #6944, opened 2023-01-13
General functionality improvements to xterm.js workspace

jonatanschroeder:
This PR fixes a few issues related to the xterm.js workspace, and provides some general improvements:

* In production the workspace runs as user 1000, but this user does not exist by default. So the prompt by default is shown with "I have no name". This PR creates a user called "student" and starts the terminal in its home directory.
* The web socket used by the terminal to transmit data to the server would timeout after being idle for a couple of minutes. This PR sends a heartbeat message to keep the socket alive.
* Dependencies are updated to latest versions.


nwalters512 (approved):
LGTM!

Issue #6102, opened 2022-08-02
Share files between grader and workspace


ffund:
Currently, if I need some data file to be available to both external grader and student workspace, I copy it to both `tests` and `workspace`. The duplication seems non-ideal.

One workaround suggested on slack is to add it as a graded file, so it gets copied into the external grader automatically. But then if the student modifies it in the workspace, the grader will get the modified version instead of a "clean" copy.

Perhaps both `externalGradingOptions` and `workspaceOptions` should have config variables `serverFilesCourse` and `serverFilesQuestion`, to specify files that would be copied to the grader and/or workspace. (currently only the `externalGradingOptions` + `serverFilesCourse` combination exists.)

That way, if I have a file that should be shared between workspace and grader, I could put it in `serverFilesQuestion` and it will be copied to both. and if the file should further be shared between multiple questions, I could put it in `serverFilesCourse`  and copy it to workspace and/or grader in many questions.

(Generally, files in `serverFiles` are not supposed to be student-accessible. But Matt said: "I think it would be fine to copy files from `serverFiles` into workspaces, so long as it’s being done explicitly under instructor control.")

(Summarized from a Slack discussion.)

PR #6947, opened 2023-01-13
Add support for dynamic files in workspaces

jonatanschroeder:
Allows questions using workspaces to have dynamic files. Files are created in `data['params']['_files']` using the same format as `data['submitted_answers']['_files']`, i.e., an array of objects where each object has a name and a base64-encoded content.

I have no way of testing the S3 component of this implementation, so I'd appreciate it if someone could work on it for me.

* [x] Basic implementation
* [ ] Error handling (e.g., incorrect format for the parameter)
* [ ] Example course question
* [ ] Tests
* [ ] Documentation

### DIFF
--- a/lib/workspace.js
+++ b/lib/workspace.js
@@ -22,16 +22,21 @@ const sqlLoader = require('../prairielib/lib/sql-loader');
 const ERR = require('async-stacktrace');
 const sql = sqlLoader.loadSqlEquiv(__filename);
 
-const zipDirectory = async function (source, zip) {
+const zipDirectory = async function (source, dynamic, zip) {
   const stream = fs.createWriteStream(zip);
   const archive = archiver('zip');
 
   await new Promise((resolve, reject) => {
     stream
-      .on('open', () => {
+      .on('open', async () => {
         archive.pipe(stream);
         if (source) {
           archive.directory(source, false);
+        }
+        if (dynamic) {
+          async.eachSeries(dynamic, ({ name, contents }) => {
+            archive.append(Buffer.from(contents, 'base64'), { name });
+          });
         }
         archive.on('error', (err) => {
           throw err;
@@ -327,7 +332,7 @@ module.exports = {
    * @returns {Promise<InitializeResult | null>}
    */
   async initialize(workspace_id) {
-    const { workspace, question, course } = (
+    const { workspace, variant, question, course } = (
       await sqldb.queryOneRowAsync(sql.select_workspace_data, { workspace_id })
     ).rows[0];
     const course_path = chunks.getRuntimeDirectoryForCourse(course);
@@ -368,16 +373,31 @@ module.exports = {
     } catch (err) {
       localPathExists = false;
     }
+    const paramsFilesExist = Array.isArray(variant.params?._files);
 
     // upload files to s3/efs depending on how the workspace is configured
     if (workspace.homedir_location === 'S3') {
-      await zipDirectory(localPathExists ? localPath : null, localZipPath);
+      await zipDirectory(localPathExists ? localPath : null, variant.params?._files, localZipPath);
       const remoteZipPath = `${remoteDirName}/initial.zip`;
       await awsHelper.uploadToS3Async(config.workspaceS3Bucket, remoteZipPath, localZipPath, false);
 
       if (localPathExists) {
         debug(`Syncing ${localPath} to ${remotePath}`);
         await awsHelper.uploadDirectoryToS3Async(config.workspaceS3Bucket, remotePath, localPath);
+      }
+
+      if (paramsFilesExist) {
+        async.eachSeries(variant.params._files, async ({ name, contents }) => {
+          // TODO Handle errors or incorrect format
+          const buffer = Buffer.from(contents, 'base64');
+          const result = await awsHelper.uploadToS3Async(
+            config.workspaceS3Bucket,
+            path.join(remotePath, name),
+            false, // localPath
+            false, // isDirectory
+            buffer
+          );
+        });
       }
 
       return null;
@@ -396,7 +416,17 @@ module.exports = {
       if (localPathExists) {
         debug(`Syncing ${localPath} to ${remotePath}`);
         await fse.copy(localPath, sourcePath);
+      }
 
+      if (paramsFilesExist) {
+        async.eachSeries(variant.params._files, async ({ name, contents }) => {
+          // TODO Handle errors or incorrect format
+          const buffer = Buffer.from(contents, 'base64');
+          const result = await fse.writeFile(path.join(sourcePath, name), buffer);
+        });
+      }
+
+      if (localPathExists || paramsFilesExist) {
         // Update permissions so that the directory and all contents are owned by the workspace user
         for await (const file of klaw(sourcePath)) {
           await fsPromises.chown(
--- a/lib/workspace.sql
+++ b/lib/workspace.sql
@@ -19,6 +19,7 @@ FOR UPDATE;
 -- BLOCK select_workspace_data
 SELECT
     to_jsonb(w.*) AS workspace,
+    to_jsonb(v.*) AS variant,
     to_jsonb(q.*) AS question,
     to_jsonb(c.*) AS course
 FROM


jonatanschroeder:
I haven't created CI tests because there are currently no tests that even check the static files, so I wouldn't really know how to check these.

Also, I'm almost sure the S3 version of the code is correct, but I don't have a way to test it. I have tested the zip version manually and it creates the archive as expected, but I can't test if the upload method works in the same way. All that needs to be done is to check if the demo/workspace/dynamicFiles question in the exampleCourse, when the workspace is created, has the combined files of the workspace directory (original.txt) and the ones in server.py, with appropriate content.

lib/workspace.js
```diff
@@ -380,6 +406,18 @@ module.exports = {
...
+          await awsHelper.uploadToS3Async(
+            config.workspaceS3Bucket,
+            path.join(remotePath, name),
+            null, // localPath
+            false, // isDirectory
+            buffer
+          );
+        });
+      }
+
```
jonatanschroeder, line 420: This is the section I'm unable to test.

jonatanschroeder:
I can also use similar functionality to resolve #6102 and related issues, by having the item point to another file inside the question directory (e.g., tests, clientFilesQuestion, etc.).

lib/workspace.js
```diff
@@ -22,7 +22,7 @@ const sqlLoader = require('../prairielib/lib/sql-loader');
 const ERR = require('async-stacktrace');
 const sql = sqlLoader.loadSqlEquiv(__filename);
 
-const zipDirectory = async function (source, zip) {
+const zipDirectory = async function (source, dynamic, zip) {
```
nwalters512, line 25: I'd encourage you to add a JSDoc comment to this function with types.

lib/workspace.js
```diff
@@ -369,9 +379,38 @@ module.exports = {
       localPathExists = false;
     }
 
+    const dynamicFiles = Array.isArray(variant_params?._files)
+      ? variant_params._files
+          .map((file) => {
+            try {
+              if (!file.name) {
+                logger.warn(`Dynamic file included without a name. Ignoring file.`);
```
nwalters512, line 387: I'd be inclined to remove these logs, as they won't ever be seen by users and will just serve to clutter up our normal logs. We should think of some way to surface these to instructors - perhaps via a course issue? We don't currently have any great way to show an error that was produced when a workspace was being initialized.

lib/workspace.js
```diff
@@ -369,9 +379,38 @@ module.exports = {
...
+                return null;
+              }
+              if (file.questionFile) {
+                return {
+                  name: file.name,
+                  localPath: path.join(questionBasePath, file.questionFile),
```
nwalters512, line 393: I believe this is vulnerable to directory traversal, e.g. one could specify `../../../../../etc/passwd`. We should validate the `file.questionFile` is located within `questionBasePath`.

lib/workspace.js
```diff
@@ -396,7 +447,20 @@ module.exports = {
...
+      }
+
+      if (dynamicFiles) {
+        await async.eachSeries(dynamicFiles, async ({ name, localPath, buffer }) => {
+          await fse.ensureDir(path.join(sourcePath, path.dirname(name)));
+          if (localPath) {
+            await fse.copy(localPath, path.join(sourcePath, name));
+          } else {
+            await fse.writeFile(path.join(sourcePath, name), buffer);
+          }
```
nwalters512, line 459: As above, we should ensure that users can't create or write to arbitrary files, which in this case means validating that the results of `path.join(sourcePath, ...)` don't resolve to somewhere outside of `sourcePath`.

jonatanschroeder: I believe this is resolved in 4f38c71.

lib/workspace.js
jonatanschroeder: I believe this is resolved in 4f38c71.

lib/workspace.js
jonatanschroeder: I took a look at the option of creating an issue, but it would require a few changes that I'm not comfortable doing now. One example is the fact that the initialization is done in the context of a socket connection, where the authn_user_id is not readily available, and that field is required in the issue. I changed it to a comment for now. I can add a GitHub issue to track this.

nwalters512: A comment is fine for now, yeah.

lib/workspace.js
```diff
@@ -369,9 +388,49 @@ module.exports = {
...
+            if (!file) return false;
+            // Discard names with directory traversal outside the home directory
+            const relative = path.relative(remotePath, path.join(remotePath, file.name));
+            if (relative === '.' || relative === '..' || relative.startsWith('../')) return false;
+            if (file.localPath) {
+              // Discard paths with directory traversal outside the question
+              const relative = path.relative(questionBasePath, file.localPath);
+              if (relative === '.' || relative === '..' || relative.startsWith('../')) return false;
+              // Discard paths that do not exist
+              if (!fse.existsSync(file.localPath)) return false;
```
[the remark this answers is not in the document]
jonatanschroeder: Moved this logic to the map function, now using `async.mapSeries`.

lib/workspace.js
```diff
@@ -369,9 +388,49 @@ module.exports = {
...
+              // Error retrieving contents of dynamic file. Ignoring file.
+              return null;
+            }
+          })
+          .filter((file) => {
+            // Discard files ignored in previous step
+            if (!file) return false;
+            // Discard names with directory traversal outside the home directory
+            const relative = path.relative(remotePath, path.join(remotePath, file.name));
+            if (relative === '.' || relative === '..' || relative.startsWith('../')) return false;
```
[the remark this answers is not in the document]
jonatanschroeder: Done in 683cdf2. Also changed a few other locations doing a similar check to use the lib function. Pending: tests.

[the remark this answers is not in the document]
jonatanschroeder: Tests done in 722bce3eb.

lib/file-paths.js
```diff
@@ -123,3 +123,17 @@ module.exports.questionFilePath = function (
     }
   });
 };
+
+/**
+ * Returs true if the parent path contains the child path. Used to allow code to make checks that
+ * prevent directory traversal attacks.
+ * @param {string} parentPath - the path of the parent directory.
+ * @param {string} childPath - the path of the child directory.
+ * @param {boolean=true} includeSelf - return valud if both paths point to the same directory.
```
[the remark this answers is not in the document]
jonatanschroeder: It's actually supposed to be return value... Fixed.

lib/workspace.js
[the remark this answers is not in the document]
jonatanschroeder: @nwalters512 Seems like we've done concurrent work on the same feature (see #2674). I've adapted the code to the new function there and moved tests and other calls to the new library.

docs/workspaces/index.md
```diff
@@ -154,6 +152,45 @@ A minimal `question.html` for an externally graded workspace should look somethi
 </pl-submission-panel>
 ```
 
+### Creating files in the workspace home directory
+
+Workspace questions can optionally include a `workspace/` subdirectory within the regular [PrairieLearn question directory structure](../question.md#directory-structure). If this `workspace/` subdirectory exists, its contents will be copied into the home directory of the student's workspace container, as configured in the `home` setting in `info.json`.
```
mwest1066, line 157: Can we specify which file is used if a file appears in both the `workspace/` directory and the `_files` parameter? I think the dynamic files will overwrite files in `workspace/`?

docs/workspaces/index.md
```diff
@@ -154,6 +152,45 @@ A minimal `question.html` for an externally graded workspace should look somethi
...
+
+Questions using workspaces can also be randomized, i.e., include files that contain random and dynamic content. This is done using [the `server.py` file in the question directory](../question.md#question-serverpy). In addition to other random parameters that can be created for the question page itself, the `_files` parameter can also be set, containing an array of files to be created. Each element of the array must include a `name` property, containing the file name (which can include a path with directories), and either a `contents` property, containing the contents of the file, or a `questionFile` property, pointing to an existing file in a different location in the question directory. For example:
```
mwest1066, line 159: I like the flexibility of this mechanism, but I'm a bit worried that it has a lot of overhead for most users. Could we also support a `workspaceTemplates/` directory that contains files which we will process with mustache (just like `question.html`, with the same `data` available) and copy into the workspace? I'm thinking the precedence would be (1) files from `workspace/`, (2) files files from `workspaceTemplates/`, and finally (3) files from `_files`.

lib/workspace.js
```diff
@@ -318,6 +320,39 @@ module.exports = {
...
+            // Discard paths that do not exist
+            if (!(await fse.exists(localPath))) return null;
+            return {
+              name: file.name,
+              localPath,
+            };
+          }
+          return {
+            name: file.name,
+            buffer: Buffer.from(file.contents ?? '', file.encoding ?? 'utf-8'),
```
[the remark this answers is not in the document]
jonatanschroeder: I'm ok with that but I'd add hex.

docs/workspaces/index.md
jonatanschroeder: I'm fine with that. Is it ok if I move it to a future PR?

docs/workspaces/index.md
jonatanschroeder: Done in 2119568e0.

docs/workspaces/index.md
```diff
@@ -154,6 +152,45 @@ A minimal `question.html` for an externally graded workspace should look somethi
...
+
+```py
+def generate(data):
+
+    # Generate 1000 random bytes
+    random_binary = os.urandom(1000)
+    # Generate 1000 random printable ASCII characters, ending with a line break
+    random_text = "".join(random.choices(string.printable, k=1000)) + "\n"
+
+    data["params"]["_files"] = [
```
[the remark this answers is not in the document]
jonatanschroeder: Done in d7ce83739.

lib/workspace.js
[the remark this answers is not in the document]
jonatanschroeder: Done in d7ce83739.

docs/workspaces/index.md
mwest1066: Well, yeah, I guess, but I'm worried it will never happen 😄. Could I lean on you to rush it into this PR? I know I've already delayed this PR a lot so I won't block merging over this request, but it would be so nice to see this feature added (and a News item!).

jonatanschroeder: I'll start on it soon, as soon as this is merged. See #7089. I assume the news item should go to that PR, and include both dynamic files and wildcards in #6982?

mwest1066 (approved):

PR #6948, opened 2023-01-13
Add config to enable or disable specific cron jobs

nwalters512:
This PR gives us the ability to finely control which cron jobs run on which hosts, which is particularly useful in heterogeneous deployments like those with chunk servers and canary servers. Before this change, running cron jobs was all-or-nothing.


mwest1066 (approved):
Short and sweet!

PR #6952, opened 2023-01-14
Display institution on course settings page

### LINKED ISSUE #6950: Show institution on course settings page
At the moment there is no way for an instructor to see which institution their course is in. We should display this on the course "Settings" page.

mwest1066:
Resolves #6950


nwalters512 (approved):
We should consider if we need to fetch the institution for *all* pages using `authzCourseOrInstance`. I'd be inclined to fetch the institution only for the `instructorCourseAdminSettings` page.

nwalters512 (approved):

Issue #6789, opened 2022-12-21
Migrate Pandas Dataframe code from pl-python-variable into new element pl-dataframe


coatless:
@eliotwrobson has done a fantastic job in https://github.com/PrairieLearn/PrairieLearn/pull/6744.

Though thinking about the new options proposed, there has been a bunch of _dataframe_ specific components that don't really make sense being merged with a standardized Python structure, e.g. we have _two_ distinct tracks:

- Scalar/Dictionary/List/Tuple (part of base Python)
- Pandas data frame (found in an external package)

Should we potentially split off the data frame components into a `pl-dataframe` element? This element could be used more generally to display data tables. My thought on this relates to incorporating even _more_ features that are data frame specific. For example, we might want to mirror the capabilities of `pl-variable-output` to embed a _dataframe_ export for students to use (c.f. https://github.com/PrairieLearn/PrairieLearn/issues/6786), incorporate _R_-specific `dtypes` and indices (c.f. https://github.com/PrairieLearn/PrairieLearn/issues/6787), or allow for sorting _within_ the output table (c.f. https://github.com/PrairieLearn/PrairieLearn/issues/6788).

For backwards compatibility, we could retain the existing dataframe options but add a redirect to `pl-dataframe` similar to the existing `pl-code` usage within the element.

https://github.com/PrairieLearn/PrairieLearn/blob/7a3934c5e6853a5019b57320a16a6bae59677a8d/elements/pl-python-variable/pl-python-variable.py#L50

So, something along the lines of:

```python
 html += f'<pl-dataframe show-header="{show_header}" show-index="{show_index}" show-dimensions="{show_dimensions}">
</pl-dataframe>'
```

eliotwrobson:
Thanks for the shoutout! I think these features all seem really cool! If we go this far (especially adding some level of interactivity), then it makes sense to split this off into a different element. All of the changes thus far have really been extensions of the existing functionality of the `pl-python-variable` element (since these are just minor modifications to the display). I'm curious what others think about this.

PR #6953, opened 2023-01-15
feat: robust printing for `pl-python-variable`

eliotwrobson:
@coatless Resolves https://github.com/PrairieLearn/PrairieLearn/issues/6789 and resolves https://github.com/PrairieLearn/PrairieLearn/issues/6347 in greater generality. Adapts code from https://github.com/PrairieLearn/PrairieLearn/pull/6744. Takes advantage of the built-in pprint module, allowing instructors to customize the way that Python objects are printed. Significantly more general than the previous code, since now recursive objects can be printed using highly customizable options.

Screenshot:
image


docs/elements.md
```diff
@@ -1193,46 +1193,24 @@ def generate(data):
...
+| `underscore-numbers` | boolean | false   | If true, integers will be formatted with the `_` character for a thousands separator, otherwise underscores are not displayed.                                                                                                                                              |
+| `sort-dicts`         | boolean | true    | If true, dictionaries will be formatted with their keys sorted, otherwise they will display in insertion order.                                                                                                                                                             |
+| `no-highlight`       | boolean | false   | Disable syntax highlighting.                                                                                                                                                                                                                                                |
 
 #### Details
 
-As of right now, the element supports displaying either Pandas DataFrames as an HTML table or Python objects via `repr()`. When setting a parameter to a DataFrame, use PrairieLearn's built in `pl.to_json()`.
```
eliotwrobson, line 1235: Realizing that this isn't quite accurate, but I think that for all cases this was being used for, the `__repr__` was equivalent to the printed output. Will double check this though.

elements/pl-python-variable/pl-python-variable.py
```diff
@@ -1,74 +1,122 @@
...
-        suffix = pl.get_string_attrib(element, "suffix", SUFFIX_DEFAULT)
+        if isinstance(json_object, str):
+            var_string = repr(json_object)
 
-        text = prefix + repr(var_out) + suffix
-        html += '<pl-code language="python" no-highlight="{}">{}</pl-code>'.format(
-            no_highlight, text
-        )
+        else:
+            var_string = pprint.pformat(
```
eliotwrobson, line 97: Maybe change pretty print to only work with lists and dicts and then have everything else use repr? I think those are the only cases we care about being able to print complex nested things anyway.

coatless (commented):
See comments.

elements/pl-python-variable/pl-python-variable.py
eliotwrobson: I think Canada is on 3.8, but the only thing missing is the `underscore_numbers` parameter. That's supposed to be in the library in 3.10, but weirdly had issues getting that working. I guess we can try including it later.

exampleCourse/questions/element/pythonVariable/question.html
```diff
@@ -38,3 +35,31 @@
   <pl-python-variable params-name="my_dictionary" prefix="my_dictionary = "></pl-python-variable>
   </div>
 </div>
+
+<div class="card my-2">
```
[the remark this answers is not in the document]
eliotwrobson: I would agree if  `pl-card` were already introduced. I think that's going to have to be done at once (and I'm following the format of the rest of this file).

elements/pl-python-variable/pl-python-variable.py
```diff
@@ -1,74 +1,122 @@
...
-    if isinstance(var_out, dict) and "_type" in var_out:
-        if not force_text:
-            var_type = var_out["_type"]
-        var_out = pl.from_json(var_out)
-
-    # render the output variable
-    if var_type == "dataframe":
-        html += var_out.to_html(
-            header=show_header, index=show_index, classes=["pl-python-variable-table"]
+        raise KeyError(
```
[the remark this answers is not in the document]
eliotwrobson: I don't think so. The attribute is required, but this error gets thrown if the corresponding entry can't be found in the data dict.

docs/elements.md
eliotwrobson: Is there a way to reference the `question.md` file from here?

exampleCourse/questions/element/pythonVariable/question.html
```diff
@@ -5,14 +5,11 @@
 </pl-question-panel>
 
 <div class="card my-2">
+  <!-- TODO remove this DataFrame stuff once we test that it looks as it did before, so we can fully deprecate the pandas part of this element. -->
```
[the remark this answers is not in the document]
eliotwrobson: Will remove before a merge, but leaving this in just to test that the passthrough works and output looks the same.

elements/pl-python-variable/pl-python-variable.py
eliotwrobson: @coatless the one issue is that tuples and sets can't be json serialized, so they'll never be output types here. It's easy enough to add them in pl.to_json though. I'll still add the types to the instance check.

exampleCourse/questions/element/pythonVariable/question.html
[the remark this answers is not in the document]
coatless: @eliotwrobson if you have cycles maybe look at:


https://github.com/PrairieLearn/PrairieLearn/issues/6971#issuecomment-1399106468

elements/pl-python-variable/pl-python-variable.py
coatless: @eliotwrobson effort is worth it to serialize these structures. 

Not a fun experience when trying to write questions about Python structures. The element gets a bit invalidated if you need to write a custom string pass to `pl-code` with printing. 

Let me think about serialization. I think this PR is on the bottom of your todo list? I probably can have a solution on Tuesday of next week.

docs/elements.md
coatless: Something like:

```md
See [Question Writing](question.md#section-name-here)
```

Many examples of cross-referencing documentation pages can be found in the [FAQ Source](https://raw.githubusercontent.com/PrairieLearn/PrairieLearn/master/docs/faq.md) (the [FAQ page](https://prairielearn.readthedocs.io/en/latest/faq/) should show how the link appears.)

elements/pl-python-variable/pl-python-variable.py
eliotwrobson: Gotcha. No worries on waiting, I can knock off another one in the interim.

exampleCourse/questions/demo/randomDataFrame/question.html
```diff
@@ -2,7 +2,7 @@
 <markdown>
 Consider the following probability distribution $P(X)$ and its event space $x$.
 </markdown>
-<pl-python-variable params-name="df"></pl-python-variable>
+<pl-dataframe params-name="df"  show-python="false"></pl-dataframe>
```
nwalters512, line 5:
```suggestion
<pl-dataframe params-name="df" show-python="false"></pl-dataframe>
```

exampleCourse/questions/element/pythonVariable/question.html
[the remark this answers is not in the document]
nwalters512: Has this been thoroughly tested now?

elements/pl-python-variable/pl-python-variable.py
```diff
@@ -1,74 +1,115 @@
...
-        text = prefix + repr(var_out) + suffix
-        html += '<pl-code language="python" no-highlight="{}">{}</pl-code>'.format(
-            no_highlight, text
+    var_out = pl.from_json(data["params"][varname])
+
+    # Passthrough legacy support for pl-dataframe
+    if isinstance(var_out, pandas.DataFrame) and not force_text:
+        return (
+            f'<pl-dataframe params-name="{varname}" show-header="{show_header}" show-index="{show_index}" '
+            f'show-dimensions="{show_dimensions}" show-python="false"></pl-dataframe>'
```
nwalters512, line 87: Heh, this is clever!

nwalters512 (approved):
Looks reasonable!

exampleCourse/questions/element/pythonVariable/question.html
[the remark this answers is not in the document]
eliotwrobson: Yes, my previous tests all actually worked, I just left this in case there was a large code change and I wanted to make sure the pass-through still worked. Since the interface is fine, I'll go ahead and remove.

docs/elements.md
eliotwrobson: Added this reference!

elements/pl-python-variable/pl-python-variable.py
eliotwrobson: Since this has been approved with the current semantics and features, I think it's fine to add the ability to print sets / tuples in later.

exampleCourse/questions/element/pythonVariable/question.html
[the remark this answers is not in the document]
eliotwrobson: Tested again and is working!

elements/pl-python-variable/pl-python-variable.py
coatless: Added as an issue #7000

PR #6955, opened 2023-01-18
refactor: pl-symbolic-input element code cleanup

### LINKED ISSUE #3469: Improve Symbolic Input Error Messages
The following syntax error, presumably provided by Python, is not very illuminating for students, since the error is actually a missing `*` between the `9` and the `n`, and not at the exponent.

Consider adding a line of information that says that the error might occur before the caret, or something to dissuade the student from thinking that the issue happens *at the caret*.

image

### LINKED ISSUE #4268: Feature request to add more built-in reserved function names for symbolic input parsing
@jonatanschroeder @echuber2 @cheeren 

I am following up from a discussion about several work arounds and in-progress features for PL recently on the Slack forum. I have a very strong use case for adding a few more built-in reserved function names that are parsed when invoking the input element tag ``pl-symbolic-input``. These functions are stock in intro calculus classes so that they should most likely be integrated by default. The list of actively accepted functions given on my course workspace on [PL.org](https://www.prairielearn.org/) (Illinois / UIUC version of the software) is:

Screen Shot 2021-05-29 at 6 26 19 PM

Minimally, I will need support for the following functions: ``ln / sec / cot / csc / atan / asin / acos`` and perhaps even some hyperbolic trig function names are nice as well with ``atanh / acosh / asinh / etc``. I realize that support for a subset of my requested functions is already in the works. Can one of the developers please try to add the remaining function names sometime soon?

I also mentioned that I am interested in hacking around the PL source to eventually file a pull request to offer more robust support for reserved function names in the symbolic input elements. The tag already supports on-the-fly definitions of custom variable names to be used in the symbolic parsing. I aim to do the same to add on-the-fly support for custom function names in the ``pl-symbolic-input`` element tag. I do not know how long this will take me, but I will start looking at the task tonight. E.g., eventually I would like the following to work:
```html
<pl-symbolic-input 
     answers-name="du" 
     variables="x, dx" 
     extra-functions="ln, sec, asinh, atanh, atan"
     label="$14x^6dx$"
>
</pl-symbolic-input>
```
Many thanks for the helpful discussions on learning and using PL!

### LINKED ISSUE #4463: natural logarithm
It would be really nice if PrairieLearn symbolic-input question accepted ln (the natural logarithm).  As far as I can tell, symbolic-input questions will only accept log.  It would be nice if log was assumed to be base 10 and ln to be base e.

eliotwrobson:
Uplifts the `pl-symbolic-input` element to use more shared backend code with the `pl-big-o-input` and adds type hints. Also addresses the following issues:

- Resolves #4463
- Resolves #3469
- Resolves #4268
- Addresses #2344 for `pl-symbolic-input` only.


python/prairielearn.py
```diff
@@ -56,6 +68,105 @@ class ElementTestData(QuestionData):
...
+        if default is None
+        else _get_attrib(element, name, default)
+    )
+
+    # Default doesn't need to be converted, already a value of the enum
+    if is_default:
+        return enum_val
+
+    upper_enum_str = enum_val.upper()
+    accepted_names = {member.name.replace("_", "-") for member in enum_type}
```
SethPoulsen, line 162: do we really need/want to accept the underscored version of the name here? I know this is down with the other ones for backwards compatibility, but we shouldn't need backwards compatibility since this is new functionality.

python/python_helper_sympy.py
```diff
@@ -477,40 +486,59 @@ def validate_string_as_sympy(
...
         return "Invalid format."
 
+    # If complex numbers are not allowed, raise error if expression has the imaginary unit
+    if (
+        (not allow_complex)
+        and (imaginary_unit is not None)
+        and (expr_parsed.has(sympy.I))
+    ):
+        expr_parsed = expr_parsed.subs(sympy.I, sympy.Symbol(imaginary_unit))
+        return (
```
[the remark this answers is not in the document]
SethPoulsen: Meaning, the cases of allowing/not allowing the imaginary unit.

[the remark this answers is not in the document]
eliotwrobson: So after taking a look, I'm actually not sure how this error could come up (since if allow complex is False, then the imaginary unit won't be added to the list of symbols that can be parsed). I'm kindof in favor of keeping this in just because it was part of the original code and don't want to break anything, but it maybe can be removed?

Not sure if there's anyone around who knows more about how this element was written we could ask.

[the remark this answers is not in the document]
SethPoulsen: Oh sorry, I didn't realize this was preexisting functionality. Should be fine.

python/prairielearn.py
SethPoulsen: ok cool

python/python_helper_sympy.py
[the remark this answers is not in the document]
eliotwrobson: Yeah, it's preexisting, just moved from the element code to this backend helper function.

SethPoulsen (approved):

elements/pl-big-o-input/pl-big-o-input.py
```diff
@@ -33,7 +33,7 @@ class BigOType(Enum):
 PLACEHOLDER_TEXT_THRESHOLD = 20
 SHOW_HELP_TEXT_DEFAULT = True
 WEIGHT_DEFAULT = 1
-DISPLAY_DEFAULT = bou.DisplayType.INLINE
+DISPLAY_DEFAULT = pl.DisplayType.INLINE
```
nwalters512, line 36: I'm a little hesitant about making this globally shared. This is small enough (3 lines) where I really think this should just be duplicated for each element.

python/prairielearn.py
```diff
@@ -58,6 +69,67 @@ class ElementTestData(QuestionData):
...
+
+    # Set corresponding partial score and feedback
+    data["partial_scores"][question_name]["score"] = partial_score
+
+    if feedback_content:
+        data["partial_scores"][question_name]["feedback"] = feedback_content
+
+
+def determine_score_params(score: float) -> Tuple[str, Union[bool, float]]:
+    """Determine score params taken from data dict"""
```
nwalters512, line 123: It's not immediately obvious to me what this function does. Why is the second element of the tuple either a boolean or a float?

Consider typing the tuple as `Tuple[Literal["correct", "partial", "incorrect"], Union[bool, float]]` instead to reflect the fact that only specific strings will ever be present in the first element of the tuple. (I just learned about `Literal`: https://mypy.readthedocs.io/en/latest/literal_types.html)

elements/pl-big-o-input/pl-big-o-input.py
eliotwrobson: I had second thoughts about this after I did it actually, will revert.

python/prairielearn.py
eliotwrobson: I'll add the literal change, that seems very reasonable.

The main purpose of this function is to replace this weird block of code that appears everywhere:

https://github.com/PrairieLearn/PrairieLearn/blob/ae347976c4d2c1946d5417fdc28811565b42b341/elements/pl-multiple-choice/pl-multiple-choice.py#L361-L372

This is just setting the appropriate parameter to make the score badge and percentage render for different elements. Incorrect has to be set to True, since setting it to 0 won't force mustache to render the score, but the mustache templates use the "partial" key to store the percentage of partial credit given. I'll add a note about this in the docstring though.

python/prairielearn.py
```diff
@@ -58,6 +69,68 @@ class ElementTestData(QuestionData):
...
+        data["partial_scores"][question_name]["feedback"] = feedback_content
+
+
+def determine_score_params(
+    score: float,
+) -> Tuple[Literal["correct", "partial", "incorrect"], Union[bool, float]]:
+    """
+    Determine appropriate key and value for display on the frontend given the
+    score for a particular question. Setting the given key-value pair in a
+    dict passed to a mustache template will make the score badge appear.
```
nwalters512, line 123: Nit (IMO the previous version implied there was something magic about this key/value pair; I want to make it clear that this is just a convention that many elements follow)

```suggestion
    score for a particular question. For elements following PrairieLearn
    conventions, the return value can be used as a key/value pair in the
    dictionary passed to an element's Mustache template to display a score badge.
```

PR #6956, opened 2023-01-18
feat: networkx support for pl-graph

eliotwrobson:
See title. Adds networkx encoding to the `to_json` and `from_json` functions. Still needs examples but core functionality seems to work. Screenshots of examples:

image
image


coatless:
100% on board with `networkx` support. Simplifies so many options when designing custom DAGs.

eliotwrobson:
@coatless do you have any more ideas for parameters that might be useful in the new networkx display mode? I added a relabeling function but not sure what else people might find useful, or whether you have anything specific in mind.

coatless:
@eliotwrobson oh... I have a list. You might not like me anymore though. :/

eliotwrobson:
@coatless no worries 😅 if anything, it might give us some cool ways to customize the graphs we use in our course

docs/elements.md
```diff
@@ -1639,6 +1639,8 @@ def generate(data):
...
 | `directed`                  | boolean | true               | Whether to treat edges in an adjacency matrix as directed or undirected. If set to false, then edges will be rendered as undirected. _The input adjacency matrix must be symmetric if this is set to false._                                                            |
 | `weights-presentation-type` | string  | `'f'`              | Number display format for the weights when using an adjacency matrix. If presentation-type is 'sigfig', each number is formatted using the to_precision module to digits significant figures. Otherwise, each number is formatted as `{:.{digits}{presentation-type}}`. |
+| `params-name-networkx`      | string  | `None`             | The the name of a parameter containing a networkx graph to use as input for the graph.                                                                                                                                                                                  |
```
coatless, line 1642:
```suggestion
| `params-name-networkx`      | string  | `None`             | The name of a parameter containing a networkx graph to use as input for the graph.                                                                                                                                                                                  |
```

coatless:
@eliotwrobson yup, there is native support for colors; but, coloring a specific node required some hacking of the element. 

Maybe see if the following feedforward network graph could be visualized with color: 
```
import itertools
import matplotlib.pyplot as plt
import networkx as nx

subset_sizes = [5, 5, 4, 3, 2, 4, 4, 3]
subset_color = [
    "gold",
    "violet",
    "violet",
    "violet",
    "violet",
    "limegreen",
    "limegreen",
    "darkorange",
]


def multilayered_graph(*subset_sizes):
    extents = nx.utils.pairwise(itertools.accumulate((0,) + subset_sizes))
    layers = [range(start, end) for start, end in extents]
    G = nx.Graph()
    for (i, layer) in enumerate(layers):
        G.add_nodes_from(layer, layer=i)
    for layer1, layer2 in nx.utils.pairwise(layers):
        G.add_edges_from(itertools.product(layer1, layer2))
    return G


G = multilayered_graph(*subset_sizes)
color = [subset_color[data["layer"]] for v, data in G.nodes(data=True)]
pos = nx.multipartite_layout(G, subset_key="layer")

nx.draw(G, pos)
```

eliotwrobson:
@coatless Was able to convert the example you provided, seems to be as simple as setting the "color" attribute for the associated object:

image

coatless:
@eliotwrobson mmm, what happens with the `node_color` parameter? That's supposed to generate a filled circle instead of a border.

eliotwrobson:
@coatless the node color parameter didn't work, but this is turning these into graphviz, so I think the attributes may be different than pyplot? Not 100% sure.

CODEOWNERS
```diff
@@ -3,6 +3,7 @@
 /elements/pl-order-blocks @SethPoulsen
 /elements/pl-rich-text-editor @jonatanschroeder
 /elements/pl-dataframe @coatless @eliotwrobson
+/elements/pl-graph @nicknytko @eliotwrobson
```
nwalters512, line 6: Alphabetized please 🙏

docs/elements.md
```diff
@@ -1665,6 +1688,12 @@ def generate(data):
...
 | `weights-presentation-type` | string  | `'f'`              | Number display format for the weights when using an adjacency matrix. If presentation-type is 'sigfig', each number is formatted using the to_precision module to digits significant figures. Otherwise, each number is formatted as `{:.{digits}{presentation-type}}`. |
+| `params-name-networkx`      | string  | `None`             | The name of a parameter containing a networkx graph to use as input for the graph.                                                                                                                                                                                      |
+| `networkx-edge-label`       | string  | `None`             | The key from the edge data dictionary to set to the `"label"` entry for display. By default, the entry stored in `"label"` will be displayed.                                                                                                                           |
```
nwalters512, line 1692: Is this what you were referring to here?

> I'm even wondering now whether the edge relabeling is needless abstraction.

If so, I agree; let's remove this unless there's a really compelling reason for this to be an attribute we support.

elements/pl-graph/pl-graph.py
```diff
@@ -137,17 +175,27 @@ def render(element_html: str, data: pl.QuestionData) -> str:
...
+
+    input_param_matrix = pl.get_string_attrib(
         element, "params-name-matrix", PARAMS_NAME_MATRIX_DEFAULT
     )
+
+    input_param_graph = pl.get_string_attrib(
+        element, "params-name-networkx", PARAMS_NAME_GRAPH_DEFAULT
+    )
+
     input_type = pl.get_string_attrib(element, "params-type", PARAMS_TYPE_DEFAULT)
```
nwalters512, line 187: You should update the documentation for the `params-type` attribute to clarify that it can now be `networkx` as well.

python/prairielearn_test.py
```diff
@@ -45,6 +46,33 @@ def test_encoding_legacy(df: pd.DataFrame) -> None:
...
+        edge_data["weight"] = i
+        edge_data["label"] = chr(ord("a") + i)
+
+    json_object = json.dumps(pl.to_json(networkx_graph), allow_nan=False)
+    decoded_json_object = pl.from_json(json.loads(json_object))
+
+    assert type(networkx_graph) == type(decoded_json_object)
+
+    assert nx.utils.nodes_equal(networkx_graph.nodes(), decoded_json_object.nodes())
+    assert nx.utils.edges_equal(networkx_graph.edges(), decoded_json_object.edges())
```
nwalters512, line 73: 🙌 really appreciate the comprehensive tests here!

docs/elements.md
eliotwrobson: Yes, I was referring to this edge label parameter. It's convenient in some cases but can be replaced by a two-liner on the side of the instructor.

docs/elements.md
```diff
@@ -1652,19 +1651,47 @@ def generate(data):
...
 | `engine`                    | string  | dot                | The rendering engine to use; supports `circo`, `dot`, `fdp`, `neato`, `osage`, and `twopi`.                                                                                                                                                                             |
-| `params-name-matrix`        | string  | `None`             | The the name of a parameter containing the adjacency matrix to use as input for the graph.                                                                                                                                                                              |
+| `params-name-matrix`        | string  | `None`             | The the name of a parameter containing the adjacency matrix to use as input if `params-type` is set to `adjacency-matrix`.                                                                                                                                              |
```
eliotwrobson, line 1682: @nwalters512 looking at this again, would you be open to consolidating the `params-name-matrix` and `params-name-graph` into a single attribute, and having legacy support for `params-name-matrix` (via a passthrough)? Just simplifies things and makes the parameter name less confusing.

nwalters512: I think that'd be reasonable, yeah. Are you thinking that we'd just check the type of the (JSON-decoded) param and switch behavior based on that?

eliotwrobson: Not even that complex, `params-type` is already explicitly given, which specifies the backend to use. Then `params-name` is just giving the entry in the data dictionary to pull from (to be interpreted by the backend however the backend wants to).

docs/elements.md
```diff
@@ -1652,20 +1651,47 @@ def generate(data):
     data["params"]["matrix"] = pl.to_json(mat)
 ```
 
+---
+
+**question.html**
+
+```html
+<pl-graph params-name="random-graph"></pl-graph>
```
nwalters512, line 1659: This needs a `params-type` attribute, correct?

eliotwrobson: Ah whoops, good catch

eliotwrobson: Fixed!

nwalters512 (approved):

PR #6957, opened 2023-01-18
Don't track render cache stats in the database

nwalters512:
These stats are better tracked by Honeycomb and CloudWatch.

I've left the `panel_render_cache_hit_count` and `panel_render_count` columns in the DB so that we can safely deploy this. Once we've done that, I'll remove them: #6958


mwest1066 (approved):
Excellent! We've only ever used the CloudWatch data for this anyway, so this is great cleanup.

PR #5805, opened 2022-05-29
Start out pl-order-blocks in alphabetical order.

roloza7:
Adds the `source-blocks-order='alphabetized'` option to `pl-order-blocks`. 

Also adds a new test in the relevant question for the test course. Hopefully resolves #4920!

PR #6959, opened 2023-01-18
pl-order-blocks: Add option to alphabetize source blocks

### LINKED ISSUE #4920: Start out pl-order-blocks in alphabetical order
Add an option to alphabetized lines in shuffle-options (motivated by https://dl.acm.org/doi/pdf/10.1145/3017680.3017704). It makes the lines more easily findable for students but likely doesn't give anything away compared to pure shuffling.

Originally suggested by @nathanielweinman in https://github.com/PrairieLearn/PrairieLearn/issues/3476

nwalters512:
This PR adds support for a `source-blocks-order="alphabetized"` attribute on the `<pl-order-blocks>` element. Includes documentation and an example question.

Closes #4920.

Continues the work started by @roloza7 in #5805.


ffund:
With @eliotwrobson working to maybe deprecate `pl-dropdown` eventually, maybe this is a good opportunity to think about standardizing the way we specify order across elements? So that users don't have to remember different syntax per element. (See #5369 )

SethPoulsen:
Looks good overall! Just 2 things I care to mention: 

First of all, I just want to call out again (though @nwalters512 already mentioned it in the other PR), that this is doing Ordering based on the HTML, which leads to some counterintuitive results. For example, this:

```html
<pl-order-blocks answers-name="example" source-blocks-order="alphabetized" solution-placement="bottom" >
  <pl-answer correct="false">$\pi$</pl-answer>
  <pl-answer correct="true">a</pl-answer>
  <pl-answer correct="false"><pl-code>stuff and things</pl-code></pl-answer>
  <pl-answer correct="false"><pl-integer-input answers-name="stuff"></pl-integer-input> banana </pl-answer>
  <pl-answer correct="false"><pl-integer-input answers-name="things"></pl-integer-input> apple </pl-answer>
</pl-order-blocks>
```

ends up ordered as: 

image

this shouldn't really be a big deal, because often when people do have HTML formatting in a block, it means that _all_ of the blocks have the same HTML at the begining (e.g. they are all wrapped in a `pl-code` or something). If people care about it at some point later, we could actual parse the HTML then take the inner text of it or something. That being said, we should probably have it be consistent across elements. 

Which brings me to the second thing I care to mention, which is that I agree with @ffund that we should have a conversation about standardization across elements before merging this.

SethPoulsen:
A slight wrinkle in ordering based on the inner text of the HTML might be that not all the PL elements have been expanded yet at time of ordering, but that shouldn't be a big deal either.

eliotwrobson:
To the discussion about standardization, I think it's easy enough to add new ways of ordering things. As long as we have a comprehensive list of ways, we can make a shared enum with all of the types.

nwalters512:
> Which brings me to the second thing I care to mention, which is that I agree with @ffund that we should have a conversation about standardization across elements before merging this.

I'm not sure if that conversation is necessary right now. This PR adds an additional value to an existing attribute; whether or not we someday rename that attribute for standardization seems orthogonal to this PR.

That said, @SethPoulsen is there a specific change you have in mind that could be applied in this PR?

SethPoulsen:
is there any value to the ascend/descend that pl-dropdown had? Does anyone actually use 'descend'?

nwalters512:
@ffund very much agreed re: the value of standardization!

> Is there any value to the ascend/descend that pl-dropdown had? Does anyone actually use 'descend'?

I see 6 questions the use them. Four are in Portuguese, so I can't tell if it's relevant to the question, but for the other two, it doesn't seem to be relevant. I can't think of a pedagogical reason one would want to explicitly choose between ascending and descending alphabetical order either...

mwest1066 (approved):

jonatanschroeder:
> I see 6 questions the use them. Four are in Portuguese, so I can't tell if it's relevant to the question, 

Let me know if you need a review of the Portuguese content (assuming @mfsilva22 is busy).

PR #6960, opened 2023-01-18
Convert questionFilePath function to async/await

nwalters512:
Lifting this out of #4934 so it can be reviewed and merged independently.


nwalters512:
Reverting this to a draft; this change may actually be unnecessary.

nwalters512:
Closing in favor of #4934.

PR #6964, opened 2023-01-19
Upgrade to Postgres 14 in plbase image

nwalters512:
This matches what we're now using in production.

Postgres 12 changed `extra_float_digits` from 0 to 1, which subtly changed some of the calculations involved in the tests:

```
pltest_1=# SET extra_float_digits = 0;
SET
pltest_1=# SELECT ((30::double precision) / 100 * 6::double precision);
 ?column? 
----------
      1.8
(1 row)

pltest_1=# SET extra_float_digits = 1;
SET
pltest_1=# SELECT ((30::double precision) / 100 * 6::double precision);
      ?column?      
--------------------
 1.7999999999999998
(1 row)
```

I'm working around that by doing an approximate comparison for the time being. Jonatan intends to add explicit rounding in another PR.


jonatanschroeder (approved):
It's not ideal that the values are showing unrounded in the interface, but given this is being addressed in #6271 and that this behaviour is already in place in production, it makes sense to have it as is for now. Of course this is extra incentive to get #6271 reviewed/merged/deployed to address this. @mwest1066

PR #6972, opened 2023-01-20
Fix position of pl-file-preview in element docs

### LINKED ISSUE #6969: Fix positioning of pl-file-preview in docs
In the docs, `pl-file-preview` is listed as a submission element. It should move to the decorative elements section instead.

eliotwrobson:
Resolves #6969. Also fixed broken links and ordering inconsistency with hidden hints documentation.

### DIFF
--- a/docs/elements.md
+++ b/docs/elements.md
@@ -1024,33 +1024,6 @@ The `focus` attribute defaults to `"false"`. Setting this to true will cause the
 
 ---
 
-### `pl-file-preview` element
-
-Provides an in-browser preview of pure-text or image files submitted by a student as part of an external grading system.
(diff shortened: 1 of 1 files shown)


jonatanschroeder:
I don't think this is not a conditional element. It is more of a "decorative" element, in the same vein as `pl-code` or `pl-external-grader-variables`. That said, I'm not sure if the term "decorative" here is the right word... Maybe a bit of rethinking of the titles?

Do you want to piggyback and also fix the broken links at the start of these files? The links to `pl-file-preview` and `pl-hidden-hints` are broken.

eliotwrobson:
Sure, will fix.

jonatanschroeder (approved):

Issue #6968, opened 2023-01-20
Add custom placeholder in more elements


jonatanschroeder:
Elements like `pl-string-input` and `pl-rich-text-editor` allow a custom placeholder to be placed. This would be useful in other elements, like:

* [x] `pl-number-input`
* [x] `pl-integer-input`
* [x] `pl-symbolic-input`
* [x] `pl-big-o-input`
* [ ] `pl-matrix-input`

eliotwrobson:
This has now been done for all of the most commonly used text input elements:

- pl-string-input
- pl-symbolic-input
- pl-big-o-input
- pl-integer-input
- pl-number-input
- pl-units-input

Any text input elements that aren't as actively maintained and don't have this feature will have this listed in follow-up refactor PRs.

PR #6973, opened 2023-01-20
feat: Customizable placeholder text for pl-symbolic-input and pl-big-o-input elements

eliotwrobson:
See title, in relation to https://github.com/PrairieLearn/PrairieLearn/issues/6968. Also cleans up mustache templates and adds a variable to reduce duplication of string literal name of the mustache template.

### DIFF
--- a/elements/pl-big-o-input/pl-big-o-input.py
+++ b/elements/pl-big-o-input/pl-big-o-input.py
@@ -35,7 +35,6 @@ GRADE_FUNCTION_DICT: Dict[BigOType, bou.BigOGradingFunctionT] = {
 
 VARIABLES_DEFAULT = ""
 SIZE_DEFAULT = 35
-PLACEHOLDER_TEXT_THRESHOLD = 20
 SHOW_HELP_TEXT_DEFAULT = True
 WEIGHT_DEFAULT = 1
 DISPLAY_DEFAULT = DisplayType.INLINE
@@ -133,7 +132,6 @@ def render(element_html: str, data: pl.QuestionData) -> str:
             ),
             "uuid": pl.get_uuid(),
             display.value: True,
-            "show_placeholder": size >= PLACEHOLDER_TEXT_THRESHOLD,
             "placeholder": placeholder,
             "raw_submitted_answer": raw_submitted_answer,
             "type": bigo_type,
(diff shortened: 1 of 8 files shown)


elements/pl-big-o-input/pl-big-o-input.py
```diff
@@ -130,6 +134,7 @@ def render(element_html: str, data: pl.QuestionData) -> str:
             "uuid": pl.get_uuid(),
             display.value: True,
             "show_placeholder": size >= PLACEHOLDER_TEXT_THRESHOLD,
```
jonatanschroeder, line 136: Should a custom placeholder override this check?

jonatanschroeder: Same for pl-symbolic-input.

jonatanschroeder (commented):
This is preliminary, I haven't had a chance to test it. I'll do it later (though other than the comments it all looks fine).

elements/pl-big-o-input/pl-big-o-input.py
eliotwrobson: It may not be worth having this check at all, since placeholder text can be manually set to the empty string. How does pl-string-input do it?

jonatanschroeder: Good point. `pl-string-input` doesn't have a default placeholder, so this check is not necessary there. `pl-number-input` has a default placeholder and an option for `show-placeholder`, but no check based on size (I'd be in favour of deprecating that option in favour of a blank placeholder). Feel free to remove this check.

jonatanschroeder: Don't remove existing options, for backwards compatibility.

eliotwrobson: As far as I can tell, there's no option for `show-placeholder` outside of this check in symbolic or big-o input (i.e. just setting the size). So the proposed change is just to remove the check entirely and always show whatever placeholder is given (which can be emptied out).

jonatanschroeder: Ah, ok, got it. I though you meant in other elements.

eliotwrobson: Oh no, just these two.

jonatanschroeder (approved):
All looks good, though as I'm testing this I see some issues with the example question on big-O, which (a) has a few typos (e.g., complextiy), (b) does not show some of the available options (e.g., size and inline display) and (c) is not linked from the assessment containing a collection of element examples (`gallery/elements`). Are you ok with fixing these before I merge?

eliotwrobson:
@jonatanschroeder yeah, I can do a quick pass through the big-O.

PR #6974, opened 2023-01-20
Prefer SVG MathJax on all platforms; use precision rendering mode

### LINKED ISSUE #6792: More MathJax issues with disappearing lines in CHTML rendering mode

echuber2:
Fixes #6792
Fixes https://github.com/PrairieLearn/PrairieLearn/issues/6495 

Since it appears that CHTML renders erratically compared to SVG lately, and SVG appears to be rendering accurately on all platforms now, let's prefer SVG over CHTML in the MathJax startup. Also, add CSS for the MathJax SVG to request that "geometricPrecision" [rendering mode](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering) be used, rather than a "crisp" or "speed" mode. (In the past, SVG antialiasing had similar bugs on Windows to what CHTML is exhibiting now; it appears to be okay now, but the precision mode should help to avoid that, I believe.)

One caveat: When I was double-checking the results yesterday, I did have trouble consistently reproducing the CHTML bugs on Windows Chrome, whereas in previous weeks it was happening very consistently. However, I can still see the bugs happening in Firefox with CHTML mode forced (on Windows _and_ Linux), so I think there's still an underlying, sporadic issue with the CHTML mode. I can't identify _any_ missing lines in SVG mode now, on any platform. I hope someone will point out any such case if they can find it in testing.

Regarding the other issues mentioned in https://github.com/PrairieLearn/PrairieLearn/issues/6495 : I did not disable the right-click menu option that lets you switch the renderer, because I think that's useful for testing this PR. I did switch the cache from global to local to avoid the issue (which MathJax won't fix until version 4), as well as resetting any local storage override for the renderer that the user had previously set. This allows the menu option toggle to work without issue during testing of the PR, and to make sure that if the PR is merged, the resulting switch will go smoothly and automatically for users.

The comments in the PR could be cleaned up a little bit if reviewers agree to the edits I've made overall.

One caveat about using `\LARGE` with SVG mode: The SVG rendering mode does strange things if you use `\LARGE` at all. See the examples further below where it causes `\overline` and `\hat` to render with extra vertical space. This is not a dealbreaker, but I also found a workaround. Instead of using `\LARGE`, wrap display style Latex markup in old-fashioned html `<big>` tags, like this:

```html
<big><big><big>$$x+y$$</big></big></big>
```

That results in large-size rendering that has the right proportions, even in SVG mode.

For the screenshots below, I'm using this nonsense chosen to highlight the potential issues:

```html
<p><center>inline:<br>$x=+y\frac{+w}{-z}\leq \lceil a \rceil \lfloor b \rfloor \left\lceil c \right\rceil \left\lfloor d \right\rfloor || \underline{\overline{x}} \bar{x}_{i}\hat{n}$</center></p>
<p><center>display style:</center>$$x=+y\frac{+w}{-z}\leq \lceil a \rceil \lfloor b \rfloor \left\lceil c \right\rceil \left\lfloor d \right\rfloor || \underline{\overline{x}} \bar{x}_{i}\hat{n}$$</p>
<p><center>inline LARGE:<br>$\LARGE{x=+y\frac{+w}{-z}\leq \lceil a \rceil \lfloor b \rfloor \left\lceil c \right\rceil \left\lfloor d \right\rfloor || \underline{\overline{x}} \bar{x}_{i}\hat{n}}$</center></p>
<p><center>display style LARGE:</center>$$\LARGE{x=+y\frac{+w}{-z}\leq \lceil a \rceil \lfloor b \rfloor \left\lceil c \right\rceil \left\lfloor d \right\rfloor || \underline{\overline{x}} \bar{x}_{i}\hat{n}}$$</p>
<p><center>inline big tags:<br><big><big><big>$x=+y\frac{+w}{-z}\leq \lceil a \rceil \lfloor b \rfloor \left\lceil c \right\rceil \left\lfloor d \right\rfloor || \underline{\overline{x}} \bar{x}_{i}\hat{n}$</big></big></big></center></p>
<p><center>display style big tags:</center><big><big><big>$$x=+y\frac{+w}{-z}\leq \lceil a \rceil \lfloor b \rfloor \left\lceil c \right\rceil \left\lfloor d \right\rfloor || \underline{\overline{x}} \bar{x}_{i}\hat{n}$$</big></big></big></p>
<!-- part 2 -->
<p><center>inline:<br>$s = \sqrt{\frac{\sum{(x_i + \overline{x})^2\frac{+w}{-z}\leq \hat{v}}}{n+w}}$</center></p>
<p><center>display style:</center>$$s = \sqrt{\frac{\sum{(x_i + \overline{x})^2\frac{+w}{-z}\leq \hat{v}}}{n+w}}$$</p>
<p><center>inline LARGE:<br>$\LARGE{s = \sqrt{\frac{\sum{(x_i + \overline{x})^2\frac{+w}{-z}\leq \hat{v}}}{n+w}}}$</center></p>
<p><center>display style LARGE:</center>$$\LARGE{s = \sqrt{\frac{\sum{(x_i + \overline{x})^2\frac{+w}{-z}\leq \hat{v}}}{n+w}}}$$</p>
<p><center>inline big tags:<br><big><big><big>$s = \sqrt{\frac{\sum{(x_i + \overline{x})^2\frac{+w}{-z}\leq \hat{v}}}{n+w}}$</big></big></big></center></p>
<p><center>display style big tags:</center><big><big><big>$$s = \sqrt{\frac{\sum{(x_i + \overline{x})^2\frac{+w}{-z}\leq \hat{v}}}{n+w}}$$</big></big></big></p>
```

You can easily test with the example course XC 101, question `demo/markdownEditorLivePreview`. I hope someone will try it on macOS, especially Safari. Windows or Linux users with 4K screens would also be good data points. Whatever the platform, try zooming in and out in your browser to make sure that horizontal or vertical lines don't disappear spontaneously.

Here are screenshots showing the accurate results in SVG mode (default) after applying the PR. (Contrast with buggy images shown at https://github.com/PrairieLearn/PrairieLearn/issues/6792)

## Windows Chrome SVG rendering

win chrome svg fixed 1

win chrome svg fixed 2

## Windows Edge SVG rendering

win edge svg fixed 1

win edge svg fixed 2

### At 80% zoom

This is smaller than you would typically use, but I show it here because in CHTML, it had been explicitly buggy (missing lines). Now with SVG mode and precision rendering, the lines are all there.

win ff svg 80pct fixed 1

win ff svg 80pct fixed 2

### At 100% zoom

win ff svg 100pct fixed 1

win ff svg 100pct fixed 2


echuber2:
Linux Firefox was already defaulting to SVG before this change, but here is how it looks now with the precision rendering mode. (There is no big difference:)

linux ff svg 100pct better 1

linux ff svg 100pct better 2

jonatanschroeder:
I just tested on Windows with Edge, Chrome and Firefox on master, and none gave me an issue originally, so I can't tell you (at least on my environment) if this will fix any issue because I can't reproduce the original issue. Is there something specific I need to do to test it? Specific resolution? Specific zoom level? Specific display scale settings?

echuber2:
Thanks! If you have a high-DPI monitor, it would be good to know what your Windows scaling level is. My tests were done with a normal-DPI monitor with default 100% scaling at the OS level and 100% zoom in the browser. If you zoom in and out with the browser controls by small amounts, the font shouldn't have any missing horizontal lines at any zoom level, but it's hard to predict which zoom levels _might_ have the issue.

echuber2:
This might conflict with the approach in https://github.com/PrairieLearn/PrairieLearn/pull/7280

pages/partials/mathjax.ejs
```diff
@@ -1,7 +1,15 @@
 <script>
 
 // default to CHTML because of https://github.com/PrairieLearn/PrairieLearn/pull/3520
-var outputComponent = 'output/chtml';
+// var outputComponent = 'output/chtml';
```
nwalters512, line 4: No point keeping this around.

pages/partials/mathjax.ejs
```diff
@@ -1,7 +1,15 @@
...
+
+// Default to SVG again: Windows+Chrome was the primary case where the CHTML
+// mode was active, but now thin lines are disappearing on that platform too.
```
nwalters512, line 7:
```suggestion
// Default to SVG, as lines were sometimes disappearing when using the CHTML renderer.
```

pages/partials/mathjax.ejs
```diff
@@ -1,7 +1,15 @@
...
+var outputComponent = 'output/svg';
+
+// With 'output/svg' now being the default, the following checks may be
+// redundant, but they could also be useful reference if new issues seem to
+// arise.
```
nwalters512, line 12: Let's just remove all this; it will pretty much immediately become debt. We have `git blame` and history if we ever want to know what this looked like previously.

pages/partials/mathjax.ejs
```diff
@@ -16,12 +24,32 @@ var outputComponent = 'output/chtml';
...
+    const settingsKey = 'MathJax-Menu-Settings';
+    const settings = JSON.parse(localStorage.getItem(settingsKey)) || {};
+    delete settings['renderer'];
+    const settingsJson = JSON.stringify(settings);
+    if (settingsJson === '{}') {
+        localStorage.removeItem(settingsKey);
+    } else {
+        localStorage.setItem(settingsKey, settingsJson);
+    }
+} catch(e) { }
```
nwalters512, line 41: Going by what @trombonekenny said, we should allow users to change this if they must (maybe it allows them to circumvent *another* issue with rendering). I like his suggestion of adding a "Reset MathJax settings" button to the user settings page.

pages/partials/mathjax.ejs
```diff
@@ -16,12 +24,32 @@ var outputComponent = 'output/chtml';
...
+
 var MathJax = {
     tex: {
         inlineMath: [['$', '$'], ['\\(', '\\)']]
     },
     svg: {
-        fontCache: 'global'
+        // Using local instead of global prevents a MathJax bug if the user
```
nwalters512, line 48: Let's link to the bug for future reference.

public/stylesheets/local.css
```diff
@@ -114,6 +114,14 @@ pre {
     overflow-wrap: normal;
 }
 
+/* MathJax SVG shape rendering: Prefer accurate antialiased font shapes
+   over "crisp" pixel-aligned ones. This may be less sharp in some cases,
+   but it can prevent thin lines from disappearing. */
```
nwalters512, line 119: Is there a MathJax issue tracking this? If so, let's link to it so we can identify when this specific CSS won't be necessary in the future. If not, please file one with them!

echuber2: I can try to raise it to them as an issue, but it seems like it's a limitation of the renderer on some platforms that they don't have control over. I could suggest that they try adding this CSS to their own styles, or maybe PR that if I can find the right place.

pages/partials/mathjax.ejs
echuber2: I can look into putting the reset on the settings page instead of here. However, I should note that with this as-is, the page will load successfully even if the rendering looks bad (thanks to the local, non-global cache), and then the user can use the popup menu to change the renderer for the duration of the page view. It will reset if they reload the page though.

For the sake of merging this PR sooner, should we just remove this chunk here?

nwalters512: Yeah, I'd err towards removing this. If the user _can_ fix rendering by changing the renderer, I'd rather they have the ability to persist that.

public/stylesheets/local.css
nwalters512: In case it _is_ something they have control over, I still think it would be a good idea to report this upstream. If they really can't do anything about it, they may still want to document this limitation and encourage folks to use the SVG renderer. (I acknowledge that this doesn't help us, but it feels like the right thing to do.)

pages/partials/mathjax.ejs
echuber2: Looks like @trombonekenny has covered this here: https://github.com/PrairieLearn/PrairieLearn/pull/7286

nwalters512 (approved):

PR #6976, opened 2023-01-20
Update PrairieLearn and PrairieTest hostnames

nwalters512:
This is necessary cleanup after we moved from `.org` to `.com` domains.

Closes https://github.com/PrairieLearnInc/sysconf/issues/694


jonatanschroeder:
Maybe this?

https://github.com/PrairieLearn/PrairieLearn/blob/master/docs/whatToTellStudents.md#L14

nwalters512:
Oh man - it didn't even occur to me to grep for the _old_ old hostname! Let me fix those too.

mwest1066 (approved):

PR #6979, opened 2023-01-21
Add binary attribute to .yarn files

jonatanschroeder:
This is to ensure files in `.yarn` are not included in results of `git grep` or `git diff`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,5 +6,6 @@
 
 # Treat all files inside .yarn/ as binary files. This ensures they do
 # not make a mess in results for git grep or git diff. See
-# https://stackoverflow.com/a/51564689
-.yarn/**/* binary
+# https://stackoverflow.com/a/51564689 and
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -3,3 +3,8 @@
 * text=auto
 
 *.sh text eol=lf
+
+# Treat all files inside .yarn/ as binary files. This ensures they do
+# not make a mess in results for git grep or git diff. See
+# https://stackoverflow.com/a/51564689
```
nwalters512, line 9: In fact this is exactly what Yarn recommends! See "Tip" under https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored.

nwalters512 (approved):

.gitattributes
jonatanschroeder: Want me to change the link in the comments?

Issue #878, opened 2017-10-24
Add option to suppress partial score labels for input elements


tbretl:
Input elements now display their own partial scores. Sometimes, doing so doesn't make sense. For example consider the question `partialCredit1` in `PrairieLearn/exampleCourse`. It looks strange to have the input element say "0%" and to have the submission panel header say something like "23%".

eliotwrobson:
@nwalters512 can this be closed? I think this is already possible by just emptying out/manually changing the partial scores dictionary for the corresponding input element.

mwest1066:
It's true that we can do this via Python code, but this is pretty obscure for many instructors. I think it would be better to add a `"show-partial-score"="false"` attribute to (all?) elements.

eliotwrobson:
@mwest1066 I'm not sure what the use cases would be outside of custom graded questions, where emptying out the partial scores dictionary is pretty straightforward in the grader code. I think it's a fine option to add, but doesn't seem useful outside of this specific context. Would anyone want to hide the partial scores for a question that doesn't use custom grading?

jonatanschroeder:
This is a bit of a niche problem, but assume a question where the instructor wants to give an all-or-nothing set of items (https://prairielearn.readthedocs.io/en/latest/question/#the-partialcredit-option). In this case, the instructor may expect the student to provide a correct answer to all items before awarding points, and may want to reduce changes that students stumble upon the answer through trial and error.

Another option where this may be useful is in manually-graded questions using gradable elements (e.g., `pl-string-input`), where the grading of the question as a whole means individual items may not receive individual partial score tags.

I understand your point of view, and I agree that this option would not be used often outside of custom grading scenarios, but forcing an instructor to use a programmatic solution when a simple argument can be used instead may not always be a best approach, even when custom functions are already in place. There are instructors that may be trained to use very targeted pieces of code, often borrowed from others, and may not have enough expertise to expand on that code to provide such functionality.

So I'm in favour of keeping this issue open, though I understand that it can be kept a low priority. Note that elements like `pl-checkbox` and `pl-matching` already have options (`hide-score-badge`) that provide such functionality.

eliotwrobson:
@jonatanschroeder that all sounds reasonable. I asked about closing this because a lot of these old issues describe features that have already been implemented, but if there you see a reasonable use case then this is a pretty simple thing to add to existing elements, so it makes sense to keep it open.

eliotwrobson:
[...]

PR #6980, opened 2023-01-22
refactor: pl-integer-input code cleanup and feature addition

eliotwrobson:
Resolves https://github.com/PrairieLearn/PrairieLearn/issues/3708, adds custom placeholder text as described in https://github.com/PrairieLearn/PrairieLearn/issues/6968 and the ability to hide the score badge as described in https://github.com/PrairieLearn/PrairieLearn/issues/878 for this element. Cleans up the code and adds typechecking.

The code for this element is a bit confusing, did my best to keep the changes down while adding the desired features. Basically, if the correct answer is too large to store as a javascript int, just stores it as a string. This condition is needed to avoid breaking existing custom grader code (if we don't care about this backward compatibility, then this can be changed to be less confusing). Also, has to support weird `to_json` and `from_json` conversions, which made adding this feature a bit trickier.

New score badge display:
image


jonatanschroeder:
I believe there is an issue here if the user sets the value of the correct answer in `generate` to a value of type int above the limit. This would involve a round-trip via JS which would likely lose the int value. I believe an appropriate handling of this issue would require converting the value to string not in `pl-integer-input` code, but in the code that serializes `data` before/after a Python call.

eliotwrobson:
Yes, an oversized raw int won't be converted correctly, but converting it to a string will work with the changes here.

jonatanschroeder:
> Yes, an oversized raw int won't be converted correctly, but converting it to a string will work with the changes here.

What I'm saying is that, if an instructor sets the correct answer using an integer that is larger than the limit, it would fail silently. If you expect the instructor to use strings, it should fail if an integer is provided.

eliotwrobson:
@jonatanschroeder Ah, I see what you mean now. Yes, we should eliminate silent failures somehow, maybe have a warning displayed if someone tries to serialize oversized integers. This would involve a change to the core serialization command. I wonder if @nwalters512 has any preference.

nwalters512:
By "core serialization", do you mean this code?

https://github.com/PrairieLearn/PrairieLearn/blob/92d0202ef93897abc0f7ed6fe10f6f997c70858c/python/zygote.py#L73-L78

If so, I'm very much in favor of making that error instead of failing silently. We can just recursively check every list/dict/etc. for `int`s that are too large.

eliotwrobson:
> By "core serialization", do you mean this code?

Yes, I guess it's not a huge deal to just have a little function that checks all the dict values / keys and lists. This should arguably be the existing behavior right now anyway, since the truncation is a problem beyond just this element.

elements/pl-integer-input/pl-integer-input.py
```diff
@@ -34,46 +44,47 @@ def prepare(element_html, data):
...
+        "show-score",
     ]
 
     pl.check_attribs(element, required_attribs, optional_attribs)
     name = pl.get_string_attrib(element, "answers-name")
     base = pl.get_integer_attrib(element, "base", BASE_DEFAULT)
 
     if base != 0 and (base < 2 or base > 36):
-        raise Exception("Base must be either 0, or between 2 and 36")
+        raise ValueError(f"Base must be either 0, or between 2 and 36, not {base}")
```
nwalters512, line 56: I know this didn't change here, but what exactly is base 0? 👀

python/prairielearn.py
```diff
@@ -1776,3 +1776,28 @@ def index2key(i):
...
+                raise ValueError(
+                    f"Data structure contains oversized integer: {next_item}"
+                )
+
+        elif isinstance(next_item, list):
+            item_stack.extend(next_item)
+
+        elif isinstance(next_item, dict):
+            item_stack.extend(next_item.keys())
+            item_stack.extend(next_item.values())
```
nwalters512, line 1803: IMO we should inline these into the zygote (or maybe `zygote_utils.py` so that we can import it in tests); I don't want these to be part of our public API if we can avoid it.

python/zygote.py
```diff
@@ -72,9 +72,10 @@
 # debug the problem.
 def try_dumps(obj, sort_keys=False, allow_nan=False):
     try:
+        prairielearn.all_integers_within_limits(obj)
         return json.dumps(obj, sort_keys=sort_keys, allow_nan=allow_nan)
-    except Exception:
-        print(f"Error converting this object to json:\n{obj}\n")
+    except Exception as err:
+        print(f"Error: {err}\n When converting this object to json:\n{obj}\n")
```
nwalters512, line 78: Is printing the error actually necessary? I believe this will happen automatically when the exception bubbles up and the process dies, and I don't want to print it twice.

elements/pl-integer-input/pl-integer-input.py
jonatanschroeder: https://prairielearn.readthedocs.io/en/latest/elements/#specifying-a-non-trivial-base

python/prairielearn.py
eliotwrobson: Good call! I think having a `zygote_utils.py` makes the most sense.

python/zygote.py
eliotwrobson: I'll play around by intentionally triggering this. I think you're right though, not sure why this just catches + re-throws. Though frankly this file is a little weird overall.

nwalters512: The catch + rethrow is so we can log the entire JSON object for instructor debugging.

python/prairielearn.py
eliotwrobson: Moved here: https://github.com/PrairieLearn/PrairieLearn/pull/6980/commits/164b0ac219e30f8bcc2eb5d1730b4aabdf792c90

python/zygote.py
eliotwrobson: Ahhh I see what you mean now, will remove.

docs/elements.md
```diff
@@ -530,6 +530,8 @@ def generate(data):
 | `display`        | "block" or "inline" | "inline" | How to display the input field.                                                                                                                              |
 | `size`           | integer             | 35       | Size of the input box.                                                                                                                                       |
 | `show-help-text` | boolean             | true     | Show the question mark at the end of the input displaying required input parameters.                                                                         |
+| `placeholder`    | string              | None     | Custom placeholder text. If not set, defaults to "integer" if `base` is 10, otherwise "integer in base `base`".                                              |
+| `show-score`     | boolean             | true     | Whether to show the score badge next to this element.                                                                                                        |
```
eliotwrobson, line 534: Note: This name should be generic enough that all graded elements should support a similarly named option in the future.

python/prairielearn.py
```diff
@@ -1776,3 +1776,7 @@ def index2key(i):
         key = chr(ord("a") + i)
 
     return key
+
+
+def is_within_limits(n: int) -> bool:
```
nwalters512, line 1781: Nit: this feels a bit generic when read from a callsite (which limits? is *what* within those limits?). Consider a name like `is_int_json_serializable` instead?

python/zygote.py
```diff
@@ -72,6 +74,7 @@
 # debug the problem.
 def try_dumps(obj, sort_keys=False, allow_nan=False):
     try:
+        zu.all_integers_within_limits(obj)
```
nwalters512, line 77: Micro-nit: maybe a slightly-more descriptive name like `assert_all_integers_within_limits`?

nwalters512 (approved):
Looks good! Just a few nits that you may or may not wish to address.

exampleCourse/questions/element/integerInput/question.html
```diff
@@ -40,7 +53,11 @@
 <div class="card my-2">
   <div class="card-body">
     <pl-question-panel>
-      <p> Within this variant, the different display options for <a href="http://prairielearn.readthedocs.io/en/latest/elements/#pl-integer-input-element"><code>pl-integer-input</code></a>  on the page are displayed. Specifically, this question uses the block option (<code>display="block"</code>) to display the input field on its own line instead of being inline with the question. Moreover, the weight for answering this question is higher than its predecessors.</p>
+      <p> Within this variant, the different display options for <a
```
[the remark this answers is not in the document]
eliotwrobson: That is correct, this is the case with the version of this question that's currently in master. Must have slipped in by mistake before, so I'll go ahead and remove.

python/zygote.py
eliotwrobson: Done in https://github.com/PrairieLearn/PrairieLearn/pull/6980/commits/1d2f92670d5dc5699c47a66f85c111f7f5cae091

python/prairielearn.py
eliotwrobson: Done in https://github.com/PrairieLearn/PrairieLearn/pull/6980/commits/1d2f92670d5dc5699c47a66f85c111f7f5cae091

exampleCourse/questions/element/integerInput/question.html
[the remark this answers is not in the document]
nwalters512: Thanks for cleaning that up!

exampleCourse/questions/element/integerInput/question.html
```diff
@@ -40,39 +69,74 @@
...
-      <p>This question uses the <code>blank-answer</code> to specify a value to be used if the user leaves it blank.</p>
-      <p>Solve for the value of $c$ given $c = 0^n$ for $n = 1$ (leave the question blank if the value is zero). </p>
+      <p>The integer can also be expected to be provided in a different base.</p>
+      <p>
+        Solve for the value of $c$ given $c = {{params.a16}}_{16} + {{params.b16}}_{16}$ (provide
+        your answer in base $16$).
+      </p>
     </pl-question-panel>
-    <pl-integer-input answers-name="c_6" allow-blank="true" blank-value="0" correct-answer="0"></pl-integer-input>
+    <pl-integer-input answers-name="c_6" base="16"></pl-integer-input>
```
nwalters512, line 113: Changing the `answers-name` here will have unexpected results for any existing variants of this question. This is _probably_ ok since this is the example course, but I'd still rather err on the side of caution.

(aside: this is why semantic answer names are valuable: if you reorder the inputs, you don't have to break existing variants!)

eliotwrobson: Ah yep. I think it'll be fine but I'll change the names back to be on the safe side.

eliotwrobson: Updated in https://github.com/PrairieLearn/PrairieLearn/pull/6980/commits/3e748594029889b02fdf0a4bf2c8e392f5f2d348

python/prairielearn_test.py
```diff
@@ -304,3 +305,33 @@ def grading_function(_: str) -> tuple[bool, Optional[str]]:
...
+@pytest.mark.parametrize(
+    "item",
+    [
+        1,
+        [1, 2, [3, 4, {"thing": 5}]],
+        [1, 2, "999999999999999999999"],
+        [{"stuff": 999999}, {"other stuff": "key"}],
+    ],
+)
+def test_all_integers_within_limits_no_exception(item: Any) -> None:
```
nwalters512, line 319: Nit: can we move these tests to a `zygote_utils_test.py` file?

eliotwrobson: Done here: https://github.com/PrairieLearn/PrairieLearn/pull/6980/commits/ffb03392d1bdf6150c1774ef643d6df62263f33c

elements/pl-integer-input/pl-integer-input.mustache
```diff
@@ -121,8 +116,8 @@ integer{{^default_base}} in base {{base}}{{/default_base}}
             the submitted answer could not be interpreted as an integer{{^default_base}} in base {{base}}{{/default_base}}.  No symbolic input, fractions, or scientific notation are allowed.
         {{/format_error_message}}
     </p>
-    <p> Example <i>valid</i> inputs: <code class="user-output">1</code> &nbsp; <code class="user-output">-100</code>{{#zero_base}}; <code>0x1a</code>; <code>-0b0110</code>{{/zero_base}} </p>
-    <p> Example <i>invalid</i> inputs: <code class="user-output">6+9</code> &nbsp; <code class="user-output">3.7</code> &nbsp; <code class="user-output">3*pi</code> &nbsp; <code class="user-output">6e19</code>{{^zero_base}}; <code>0x1a</code>; <code>-0b0110</code>{{/zero_base}} </p>
+    <p> Example <i>valid</i> inputs: <code class="user-output">1</code> &nbsp; <code class="user-output">-100</code>{{#zero_base}} &nbsp; <code class="user-output">0x1a</code> &nbsp; <code class="user-output">-0b0110</code>{{/zero_base}} </p>
```
eliotwrobson, line 119: Note: Add an example using underscores in here.

eliotwrobson:
Even though this PR was already approved, I'm going to flip this back to a draft since we want to use this element as one of the examples for the changes outlined in the element consistency discussion. The changes will probably be pretty minor compared to what's already here, but I want to minimize the total number of PRs (since there's no shortage of those on my end 😅 ).

eliotwrobson:
@firasm @nwalters512 I've updated this PR with all of the items from https://github.com/PrairieLearn/PrairieLearn/discussions/7353 except for an updated version of the invalid input button (see discussion on that issue for more). It looks like there's already an `arialabel` on the input here, but a custom one can be added.

I think this is far enough along that it's ready for initial review. The main thing is that the other input elements are going to be modeled after the consistency and visual changes in this one, so it's important that we get everything settled before merging this so we don't have to go back after making the changes here.

eliotwrobson:
Screenshot of the new invalid button on the submission panel:
image

On the question panel:
image

The icon on the question panel looks a little weird with respect to the text, but I think this is a solid starting point.

firasm:
Definitely a solid starting point!!!

I think the alert symbol before the invalid text will probably take away the aesthetic weirdness

eliotwrobson:
GitHub is being weird and not picking up new commits on this branch, going to remake this PR.

docs/elements.md
```diff
@@ -545,7 +547,7 @@ The `base` argument can also accept a special value of 0. In this case, the valu
 
 #### Integer range
 
-The valid range of values accepted by pl-integer-input is between -9007199254740991 and +9007199254740991 (between -(2^53 - 1) and +(2^53 - 1)). If you need a larger input, one option you can consider is a [`pl-string-input`](#pl-string-input-element) with a [custom grade function](question.md#question-serverpy).
+pl-integer-input can accept integers of unbounded size, however the correct answer will only be stored as the Python `int` if it is between -9007199254740991 and +9007199254740991 (between -(2^53 - 1) and +(2^53 - 1)). Otherwise, the correct answer will be stored as a string.
```
jonatanschroeder, line 550: Probably worth mentioning where this is important: in the server.py scripts for parse and grade, as well as in downloaded assessment results.

exampleCourse/questions/element/integerInput/question.html
```diff
@@ -30,49 +50,93 @@
...
 <div class="card my-2">
   <div class="card-body">
     <pl-question-panel>
-      <p>The integer can also be expected to be provided in a different base.</p>
-      <p> Solve for the value of $c$ given $c = {{params.a16}}_{16} + {{params.b16}}_{16}$ (provide your answer in base $16$). </p>
+      <p>
+        The submitted integer can also be very large. This question changes the placeholder text by
+        setting the <code>placeholder</code> attribute.
+      </p>
+      <p>Submit <code>{{params.c_large}}</code>.</p>
```
jonatanschroeder, line 124:
```suggestion
      <p>Submit <code>{{correct_answers.c_large}}</code>.</p>
```
This way you avoid the need for a duplicate param.

jonatanschroeder (commented):
These are just comments from a quick skim over the code. I haven't tested this.

apps/prairielearn/elements/pl-integer-input/pl-integer-input.mustache
```diff
@@ -16,30 +16,55 @@
             <span class="input-group-text" id="pl-integer-input-{{uuid}}-label">{{{label}}}</span>
         </span>
         {{/label}}
-        <input name="{{name}}" type="text" class="form-control pl-integer-input-input {{^display_append_span}}rounded-right{{/display_append_span}}" size="{{size}}" {{^editable}}disabled{{/editable}} {{#raw_submitted_answer}}value="{{raw_submitted_answer}}"{{/raw_submitted_answer}} aria-describedby="pl-integer-input-{{uuid}}-label pl-integer-input-{{uuid}}-suffix" {{#show_placeholder}}placeholder="{{shortinfo}}"{{/show_placeholder}} />
+        <input
+            name="{{name}}"
+            type="text"
```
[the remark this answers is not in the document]
jonatanschroeder: I'm not saying you should do it, I'm just raising a discussion.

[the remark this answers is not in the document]
eliotwrobson: I think most of the time this will be an integer in base 10, but we could add a check for this if necessary.

docs/elements.md
eliotwrobson: I also forgot to mention that answers are allowed to contain underscores 🤦‍♂️

apps/prairielearn/elements/pl-integer-input/pl-integer-input.mustache
```diff
@@ -121,8 +139,8 @@ integer{{^default_base}} in base {{base}}{{/default_base}}
...
         {{/format_error_message}}
     </p>
-    <p> Example <i>valid</i> inputs: <code class="user-output">1</code> &nbsp; <code class="user-output">-100</code>{{#zero_base}}; <code>0x1a</code>; <code>-0b0110</code>{{/zero_base}} </p>
-    <p> Example <i>invalid</i> inputs: <code class="user-output">6+9</code> &nbsp; <code class="user-output">3.7</code> &nbsp; <code class="user-output">3*pi</code> &nbsp; <code class="user-output">6e19</code>{{^zero_base}}; <code>0x1a</code>; <code>-0b0110</code>{{/zero_base}} </p>
+    <p> Example <i>valid</i> inputs: <code class="user-output">1</code> &nbsp; <code class="user-output">-100</code>{{#zero_base}} &nbsp; <code class="user-output">0x1a</code> &nbsp; <code class="user-output">-0b0110</code>{{/zero_base}} </p>
+    <p> Example <i>invalid</i> inputs: <code class="user-output">6+9</code> &nbsp; <code class="user-output">3.7</code> &nbsp; <code class="user-output">3*pi</code> &nbsp; <code class="user-output">6e19</code>{{^zero_base}} &nbsp; <code class="user-output">0x1a</code> &nbsp; <code class="user-output">-0b0110</code>{{/zero_base}} </p>
```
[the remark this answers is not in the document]
eliotwrobson: Should we add word spacing to a particular CSS class then and use it here? Not sure what best practices are for this.

docs/elements.md
eliotwrobson: Discussion added here: https://github.com/PrairieLearn/PrairieLearn/pull/6980/commits/736944fa7557c6023bfd70b202cbe332ea6a0c3b

exampleCourse/questions/element/integerInput/question.html
eliotwrobson: Done here: https://github.com/PrairieLearn/PrairieLearn/pull/6980/commits/f440f210aacb2dfff8fd45f558ab3770ee08f1d5

apps/prairielearn/elements/pl-integer-input/pl-integer-input.mustache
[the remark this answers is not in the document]
nwalters512: Not relevant for accessibility, but it is a nice UX enhancement for folks working from a mobile device. I'm in favor of `inputmode="numeric"` here, of course with a check for an appropriate base.

apps/prairielearn/elements/pl-integer-input/pl-integer-input.mustache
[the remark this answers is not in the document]
nwalters512: Consider a Bootstrap margin class on the `<code>` elements, e.g. `mr-2` or `mr-3`, depending on what looks nice?

apps/prairielearn/elements/pl-integer-input/pl-integer-input.mustache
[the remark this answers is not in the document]
jonatanschroeder: I think a combination of inputmode with a [`pattern="[0-9a-fA-F_]+"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern) for base 16 (or equivalents in other bases or for special cases like allow-blank and base=0) would be a great addition.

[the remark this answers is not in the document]
eliotwrobson: Tried doing this here: https://github.com/PrairieLearn/PrairieLearn/pull/6980/commits/bba0fca7d61e325bed2f7f28bba7bb9b9b78e919

apps/prairielearn/elements/pl-integer-input/pl-integer-input.mustache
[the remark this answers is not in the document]
eliotwrobson: Done here: https://github.com/PrairieLearn/PrairieLearn/pull/6980/commits/4140536acb7036aa8d63a449cac8cb827ae32b93

Had issues getting screenshots but it looks good.

apps/prairielearn/elements/pl-integer-input/pl-integer-input.mustache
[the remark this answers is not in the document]
eliotwrobson: @jonatanschroeder I'm actually not totally sure if the pattern is a good idea. It prevents submission of invalid characters, but I think we actually want to allow student submissions in that case and then just show the invalid message. What are your thoughts on this?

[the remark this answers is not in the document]
jonatanschroeder: I think the type can remain `text` (which would allow text to get an invalid input), only `inputmode` (and possibly `pattern`) can be set. In that case, students will have the option of writing something invalid and still get feedback. The input mode is mostly a way to allow students to see a suitable keyboard in onscreen keyboard environments (e.g., mobile, tablets).

The pattern is a way to provide client-side validation, but not sure if and how we want to handle that (a warning, [maybe via Bootstrap classes?](https://getbootstrap.com/docs/4.6/components/forms/#validation) @nwalters512 thoughts?).

About the code: you use numeric for bases 1-9, it should be 1-10. As for the pattern (assuming we keep using it), I kinda liked your idea of using `string` constants. Just make sure your pattern allows `_` characters (plus whatever else is acceptable, spaces?). Also, use `*` instead of `+` if allow blank is set. About base 0, I'd suggest something like: `([0-9_]+|0[xX][0-9a-fA-F_]+|0[oO][0-7_]+|0[bB][01_]+)`, which would allow any valid combination of numbers acceptable in that base option.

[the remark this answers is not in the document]
nwalters512: IIRC the browser itself will warn if input doesn't match the pattern, so we shouldn't require any custom CSS/markup for that (🤞).

Even if we don't use `pattern` (or don't want to decide that at the moment), `inputmode="numeric"` for appropriate bases is a great starting point.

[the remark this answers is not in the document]
eliotwrobson: > IIRC the browser itself will warn if input doesn't match the pattern, so we shouldn't require any custom CSS/markup for that (🤞).

This is actually what I was concerned about. My browser wouldn't let me submit something (or at least, I wasn't sure how to dismiss the warning) that didn't match the pattern when I was playing with this. @nwalters512 I agree with your comment about allowing unconditional submission of invalid-ish stuff, so I think we should get rid of the pattern if this is the behavior.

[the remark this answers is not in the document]
eliotwrobson: @nwalters512 updated here https://github.com/PrairieLearn/PrairieLearn/pull/6980/commits/41881e12afdd1b6c405a9d56ba8a9aaa28a144d3

Removed the pattern checking since this prevented submission of arbitrary text (which is the goal of always using `type="text"`).

nwalters512 (approved):

PR #6982, opened 2023-01-23
Add support for wildcards in workspace graded files

### LINKED ISSUE #3168: Workspace syncIgnore and gradedFiles should accept globs

jonatanschroeder:
Resolves #3168 (gradedFiles only, syncIgnore is not needed since #6983 is getting rid of S3 code and the need for syncIgnore).


jonatanschroeder:
Assuming #6983 is being deployed, I'm changing this PR to mark #3168 as resolved, since `sync_ignore` is only used in S3 code.

workspace_host/interface.js
```diff
@@ -1397,19 +1398,31 @@ async function sendGradedFilesArchive(workspace_id, res) {
...
+    cwd: workspaceDir,
+    stats: true,
+  });
+
+  if (gradedFiles.length > config.workspaceMaxGradedFilesCount) {
+    return res.status(500).send({ message: 'Too many files submitted in workspace.' });
+  }
+  if (_.sumBy(gradedFiles, (file) => file.stats.size) > config.workspaceMaxGradedFilesSize) {
+    return res.status(500).send({ message: 'Workspace files are larger than allowed.' });
+  }
```
nwalters512, line 1411: I'd consider including the actual limits here, e.g. "Cannot submit more than ### files from workspace" and "Workspace files total size exceeded ###". We've used https://www.npmjs.com/package/filesize elsewhere to get a human-readable string for a number of bytes.

lib/workspace.js
```diff
@@ -528,7 +542,7 @@ module.exports = {
     const archive = archiver('zip');
     const s3Name = `workspace-${workspace_id}-${workspace_version}`;
     await async.eachLimit(
-      workspace_graded_files,
+      workspace_graded_files, // wildcards are not supported in S3, as it is not in use in production
```
nwalters512, line 545: Ideally I'd like to merge https://github.com/PrairieLearn/PrairieLearn/pull/6983 first so that we don't need to have this at all.

lib/workspace.js
```diff
@@ -473,22 +479,30 @@ module.exports = {
 
     const archive = archiver('zip');
     const remoteName = `workspace-${workspace_id}-${workspace_version}`;
+    const remoteDir = path.join(config.workspaceHomeDirRoot, remoteName, 'current');
+    const gradedFiles = await fg(workspace_graded_files, { cwd: remoteDir, stats: true });
+
+    if (gradedFiles.length > config.workspaceMaxGradedFilesCount) {
+      throw { message: 'Too many files submitted in workspace.', isFormatError: true };
```
nwalters512, line 486: Throwing a non-error object is inadvisable. If you want to be able to identify a specific kind of error, I'd recommend subclassing `Error` (e.g. `SubmissionFormatError`) and then checking if an error is an instance of that.

workspace_host/interface.js
jonatanschroeder: Fixed in 8425f6166.

lib/workspace.js
jonatanschroeder: Fixed in https://github.com/PrairieLearn/PrairieLearn/commit/8425f6166fb906eeb0287fb35eff2266737c1793.

lib/workspace.js
jonatanschroeder: I have removed the comment and the related S3 code in #6947 so a conflict resolution in either PR should be easier, depending on which one is merged first.

exampleCourse/questions/demo/workspace/xtermjs/info.json
```diff
@@ -18,8 +18,9 @@
         "port": 8080,
         "home": "/home/student",
         "gradedFiles": [
-            "starter_code.h",
-            "starter_code.c"
+            "**/*.h",
+          "**/*.c",
+          "../**/*"
```
nwalters512, line 23: Was this supposed to be committed?

lib/config.js
```diff
@@ -252,6 +252,8 @@ config.workspaceDockerDiskQuota = 1 << 30; // 1 GiB
 config.workspaceDockerCpuPeriod = 100000; // microseconds
 config.workspaceDockerCpuQuota = 90000; // portion of the CpuPeriod for this container
 config.workspaceDockerPidsLimit = 1024;
+config.workspaceMaxGradedFilesCount = 100; // maximum number of graded files
+config.workspaceMaxGradedFilesSize = 100 * 1024 * 1024; // maximum size for graded files
```
nwalters512, line 256:
```suggestion
config.workspaceMaxGradedFilesSize = 100 * 1024 * 1024; // maximum total size for graded files
```

nwalters512 (approved):

exampleCourse/questions/demo/workspace/xtermjs/info.json
jonatanschroeder: The first two yes, the last one was added for testing. Fixed in 45be3985a.

lib/question.js
```diff
@@ -6,9 +6,9 @@ const jsonStringifySafe = require('json-stringify-safe');
 const debug = require('debug')('prairielearn:' + path.basename(__filename, '.js'));
 const ejs = require('ejs');
 const moment = require('moment');
-const util = require('util');
 const fs = require('fs');
 const unzipper = require('unzipper');
+const fg = require('fast-glob');
```
mwest1066, line 11: I'm not in love with the fact that we are implicitly supporting all glob syntaxes that `fast-glob` supports. Would could at least turn off all the optional extended syntaxes. That is, set `braceExpansion: false` and `extglob: false`.

lib/question.js
```diff
@@ -379,22 +381,22 @@ module.exports = {
             callback(null);
           }
         },
-        (callback) => {
+        async () => {
```
mwest1066, line 384: I'm loving these conversions to async/await!

workspace_host/interface.js
```diff
@@ -918,19 +921,39 @@ async function sendGradedFilesArchive(workspace_id, res) {
...
   res.attachment(zipName).status(200);
   const archive = archiver('zip');
   archive.pipe(res);
 
-  for (const file of workspaceSettings.workspace_graded_files) {
+  for (const file of gradedFiles) {
     try {
-      const filePath = path.join(workspaceDir, file);
+      const filePath = path.join(workspaceDir, file.path);
       await fsPromises.lstat(filePath);
```
mwest1066, line 952: We don't seem to be using the result of this call?

workspace_host/interface.js
```diff
@@ -918,19 +921,39 @@ async function sendGradedFilesArchive(workspace_id, res) {
...
-      archive.file(filePath, { name: file });
-      debug(`Sending ${file}`);
+      archive.file(filePath, { name: file.path });
```
mwest1066, line 953: Should we be checking the `stat` result for `isFile()` to match the code in `lib/workspace.js`? I think we want to explicitly skip anything that is not a regular file?

mwest1066 (approved):

workspace_host/interface.js
jonatanschroeder: Actually that is not necessary, either here or in lib/workspace.js. By default `fast-glob` only returns regular files (https://www.npmjs.com/package/fast-glob#onlyfiles). I'll remove the test from lib/workspace.js.

mwest1066: Oh, very good.

lib/question.js
jonatanschroeder: I can see how `extglob` may potentially be confusing, but I have been using brace expansions occasionally in bash and I don't see how that could be an issue.

workspace_host/interface.js
jonatanschroeder: Done in [797286c](https://github.com/PrairieLearn/PrairieLearn/pull/6982/commits/797286c5c8c4a2a443b383d4685654d74bc473b6).

workspace_host/interface.js
jonatanschroeder: Removed in 797286c.

lib/question.js
mwest1066: I'm not worried about confusion, I'm worried about the fact that we are implicitly adding a lot of features that we'll need to support going forward. For example, if we need to switch to a different glob library in the future, we'll need to maintain backwards compatibility with everything we currently support. I'd like that list to be as short as possible 😄.

I'm ok to support brace expansions if you think they are useful -- hopefully they are standard enough that most glob libraries will support them.

jonatanschroeder: Done in 2afeeed23. I have set this to a lib-level object to avoid having redundant information in multiple places, so only one place can be changed if these options must be reviewed. I also added an abridged list of allowed patterns in the docs.

mwest1066: Excellent, thanks!

PR #6983, opened 2023-01-23
Remove workspaces S3 code

### LINKED ISSUE #4024: Remove _pushAllRunningContainersToS3()
The `_pushAllRunningContainersToS3()` function is no longer needed and just causes problems when running locally and sleeping. We should remove it.

nwalters512:
This PR removes the majority of the code that was used for running workspaces with S3 as a backend. This code is now unused and has just been a confusing maintenance burden as we work on adding workspace features.

We do eventually want to move back to S3 from EFS, but we intend on taking a different approach (uploading the entire home directory as a zip instead of separately syncing individual files).

I also took the opportunity to do a bit of general cleanup. I'll leave inline comments with more details.

I opened #6984 to track the work of removing unused columns once we've gotten a change to deploy this.

Resolves #4024.


lib/config.js
```diff
@@ -211,12 +211,9 @@ config.workspaceDevHostHostname = 'localhost';
 config.workspaceHostPort = 8081;
 config.workspaceDevContainerHostname = 'host.docker.internal';
 config.workspaceAuthzCookieMaxAgeMilliseconds = 60 * 1000;
-config.workspaceJobsDirectory = '/jobs/workspaces';
 config.workspaceJobsDirectoryOwnerUid = 0;
 config.workspaceJobsDirectoryOwnerGid = 0;
 config.workspaceJobsParallelLimit = 5;
-config.workspaceMainZipsDirectory = '/workspace_main_zips';
-config.workspaceHostZipsDirectory = '/workspace_host_zips';
```
nwalters512, line 219: AFAICT there's no reason we can't just use temp files for these, so I opted to do that and eliminate these two config options.

lib/workspace.js
```diff
@@ -336,13 +302,10 @@ module.exports = {
       questionId: question.id,
     });
 
-    // Update home directory location. We do this now to catch both first
-    // time startups and when the workspace is being reset.
-    const homedir_location = config.workspaceHomeDirLocation;
-    workspace.homedir_location = homedir_location;
+    // TODO: remove this code once nothing is reading from the `homedir_location` columns.
```
nwalters512, line 305: This is necessary to get us through a deployment of this code; we'll need to continue writing this column for as long as there are hosts that could read it.

workspace_host/interface.js
```diff
@@ -119,8 +93,6 @@ app.post(
     } else if (action === 'init') {
       const useInitialZip = _.get(req.body.options, 'useInitialZip', false);
       await initSequenceAsync(workspace_id, useInitialZip, res);
-    } else if (action === 'reset') {
-      await resetSequence(workspace_id, res);
```
nwalters512, line 123: This action was unused.

workspace_host/interface.js
```diff
@@ -545,15 +376,6 @@ async function dockerAttemptKillAndRemove(container) {
...
   }
-
-  if (name) {
-    const workspaceJobPath = path.join(workspacePrefix, name);
-    try {
-      await fsPromises.rmdir(workspaceJobPath, { recursive: true });
-    } catch (err) {
-      logger.error(`Error removing directory ${workspaceJobPath}`, err);
-    }
-  }
```
nwalters512, line 556: This was broken code to begin with; we've been seeing errors like the following in prod for quite some time now:

```
Error removing directory /jobs/workspace-a80b4e08-a522-45d5-9d7b-03cc51207ac9 ENOENT: no such file or directory, stat '/jobs/workspace-a80b4e08-a522-45d5-9d7...'
```

workspace_host/interface.js
```diff
@@ -36,35 +33,12 @@ const sqlLoader = require('../prairielib/lib/sql-loader');
...
-        )}`
-      );
-    }
-  }, 1000);
-}
-
-setInterval(() => {
-  const elapsedSec = (Date.now() - lastPushAllTime) / 1000;
-  if (elapsedSec > 900) {
-    logger.error(`_pushAllRunningContainersToS3() has not run for ${elapsedSec} seconds`);
```
jonatanschroeder, line 64: Resolves #4024

jonatanschroeder:
As pointed out in #6982 , this also means `syncIgnore` is no longer needed. Should the sync/docs/example code remove it? I imagine the schema can stay for compatibility reasons, and the database entry remains for a bit, but the functionality is not needed.

nwalters512:
@jonatanschroeder good call; I removed `syncIgnore` in 9d1433f.

schemas/schemas/infoQuestion.json
```diff
@@ -227,7 +227,16 @@
...
-            "additionalProperties": false,
+            "additionalProperties": {
+                "syncIgnore": {
+                    "description": "(deprecated) The list of files or directories that will be excluded from sync.",
+                    "type": "array",
+                    "items": {
+                        "description": "A single file or directory that will be excluded from sync.",
+                        "type": "string"
+                    }
+                }
```
nwalters512, line 238: I moved the property here so that it wouldn't produce a sync error. Is it work emitting a sync warning for this? I'm erring towards no, as it could be quite noisy for courses.

jonatanschroeder: I'd also suggest no warning. I believe a lot of instructors may have added this to their VSCode-based workspaces based on the existing examples in docs and the example code.

PR #6986, opened 2023-01-24
Rewrite some tests and database utils with async/await

nwalters512:
Before this PR, there were 159 files using `ERR\(...\)`. With that as a proxy for usage of callback functions, I figure if I can rewrite 1-3 files a day to use async/await instead of callbacks, we're at most 6 months away from getting rid of callback code entirely 🎉

### DIFF
--- a/lib/databaseDescribe.js
+++ b/lib/databaseDescribe.js
@@ -54,32 +54,21 @@ const sql = sqlLoader.loadSqlEquiv(__filename);
  */
 
 /**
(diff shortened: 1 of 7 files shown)


mwest1066 (approved):
I love the goal here!

Remind me why we can't just convert these fully to TS as we go?

nwalters512:
We'll need some build tooling for TS stuff. That's on my list, but it's gonna require maybe a week of sustained work on my part.

PR #6988, opened 2023-01-24
Don't duplicate file contents in pl-file-preview HTML

nwalters512:
Previously, we were including the contents of each submitted file three times in the HTML: two base-64-encoded strings, and one decoded one. For questions where students submit especially large files, this make page renders and loads unnecessarily slow because of all the extra bytes we had to push around.

This PR changes it to use a single base-64-encoded `data-contents` attribute with the contents. This is then decoded on the client.

### DIFF
--- a/elements/pl-file-preview/pl-file-preview.mustache
+++ b/elements/pl-file-preview/pl-file-preview.mustache
@@ -8,7 +8,7 @@
       {{#has_files}}
       <ul class="list-group list-group-flush">
         {{#files}}
-        <li class="list-group-item" data-file="{{name}}" data-contents="{{contentsb64}}" data-decode-error="{{decode_error}}">
+        <li class="list-group-item" data-name="{{name}}" data-contents="{{contentsb64}}" data-decode-error="{{decode_error}}">
           <div class="file-status-container has-preview collapsed" data-toggle="collapse" data-target="#file-preview-contents-{{uuid}}-{{index}}">
             <div class="file-status-container-left">
               <i class="file-status-icon fa fa-check-circle" style="color: #4CAF50;" aria-hidden="true"></i>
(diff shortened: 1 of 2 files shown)


elements/pl-file-preview/pl-file-preview.py
```diff
@@ -34,14 +34,15 @@ def render(element_html, data):
         for idx, file in enumerate(submitted_files):
             b64contents = file["contents"] or ""
             try:
-                contents = base64.b64decode(b64contents).decode()
+                base64.b64decode(b64contents).decode()
+                decode_error = False
```
eliotwrobson, line 38: Instead of using a boolean variable, you can just set `decode_error = "false"` and then assign it in the dictionary directly.

eliotwrobson (commented):
Might be piggybacking a little bit, but would be nice to add type hints + enable typechecking to `pl-file-preview.py`, since I think it's unlikely to result in needing any other changes.

elements/pl-file-preview/pl-file-preview.mustache
```diff
@@ -41,3 +39,49 @@
...
+    aElement.setAttribute('download', name);
+    aElement.setAttribute('target', '_blank');
+    aElement.click();
+  }
+
+  document.querySelectorAll('#file-preview-{{uuid}}').forEach((el) => {
+    el.querySelectorAll('.list-group-item').forEach((li) => {
+      const name = li.getAttribute('data-name');
+      const contents = li.getAttribute('data-contents')
+      const hasDecodeError = li.getAttribute('data-decode-error') === 'true';
```
jonatanschroeder, line 63: Personally I prefer using [`dataset`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset) instead, it's clearer:
```suggestion
      const name = li.dataset.name;
      const contents = li.dataset.contents;
      const hasDecodeError = li.dataset.decodeError === 'true';
```

nwalters512:
Closing in favor of #6995.

PR #6990, opened 2023-01-25
Java grader: replace command line arguments with parameters file

jonatanschroeder:
As discussed on Slack. Further details withheld from public discussion.

### DIFF
--- a/graders/java/autograder.sh
+++ b/graders/java/autograder.sh
@@ -17,6 +17,7 @@ exception() {
 
 mkdir -p /grade/results
 mkdir -p /grade/classpath
+mkdir -p /grade/params
 
 cp -r /javagrader/*.class /javagrader/org /javagrader/libs/* /grade/classpath
 cp -r /grade/tests/libs/*                                    /grade/classpath 2> /dev/null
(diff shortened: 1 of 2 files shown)


graders/java/autograder.sh
```diff
@@ -42,20 +43,36 @@ fi
 
 RESULTS_TEMP_DIR=$(mktemp -d -p /grade/results)
 RESULTS_TEMP_FILE="$RESULTS_TEMP_DIR/$RANDOM.json"
```
nwalters512, line 45: TIL about `$RANDOM` 🤯

nwalters512 (approved):
It's been years since I've written any Java so I can't comment on that, but the overall strategy looks good to me.

graders/java/autograder.sh
```diff
@@ -42,20 +43,36 @@ fi
...
+java -cp "$CLASSPATH" JUnitAutograder
 EOF
 
 if [ -f $RESULTS_TEMP_FILE ] ; then
-    mv $RESULTS_TEMP_FILE $RESULTS_FILE
+    RESULT_SIGNATURE=$(jq -r '.signature' $RESULTS_TEMP_FILE)
+    if [ "$RESULT_SIGNATURE" != "$SIGNATURE" ] ; then
+        echo Expected signature: $SIGNATURE
+        echo Provided signature: $RESULT_SIGNATURE
+        exception "Results did not contain correct signature. Exploit suspected. Please contact the instructor."
```
tdy, line 72: 😆 Should we deliberately make this more vague to encourage actual reporting? "Unable to process submission. Please contact the instructor."

echuber2: Along those lines, I'd recommend making the error message completely generic (like 0% grade, compilation error, or server error) with an explanation only appearing on the instructor log side.

echuber2: By generic I just mean indistinguishable from an error that would be emitted by some other code path. Indicating on the frontend whether the signature was the issue or not could be used for fuzzing or something, I imagine.

jonatanschroeder: Considering this is open source, I don't think it's that useful to make it too obscure, at the risk of instructors unable to understand why a student would get this error. I'm merging as is, future PRs can address that if necessary.

PR #6991, opened 2023-01-25
pl-order-blocks: Only disallow duplicate tags for correct blocks

### LINKED ISSUE #6962: pl-order-blocks fails with multiple empty tags
We added a check a while ago to ensure that the `tag` property was unique between `pl-answer` tags inside of a `pl-order-blocks`. However, now questions that leave `tag=""` for incorrect blocks (that don't really need to have a tag) will fail to render. 

Maybe we should allow multiple blocks to have the empty string as their tag, using the same semantics as when the `tag` attribute is left out completely.

### LINKED ISSUE #6963: pl-order-blocks only enforce unique tags among correct blocks
I like to use blocks like this -

```
<pl-order-blocks answers-name="x" grading-method="dag" >
  <pl-answer correct="true" tag="1" depends=""  > First block always </pl-answer>

  <pl-answer correct="{{params.A}}" tag="2" depends="1" > Second block in scenario A. </pl-answer>
  <pl-answer correct="{{params.A}}" tag="3" depends="2" > Third block in scenario A. </pl-answer>

  <pl-answer correct="{{params.B}}" tag="2" depends="1" > Second block in scenario B. </pl-answer>
  <pl-answer correct="{{params.B}}" tag="3" depends="2" > Third block in scenario B. </pl-answer>

  <pl-answer correct="true" tag="4" depends="3"  > Fourth block always </pl-answer>
</pl-order-blocks>
```

but now that `pl-order-blocks` enforces unique tags, it's a little bit more complicated.

Per discussion with @SethPoulsen - we might consider only enforcing unique tags among **correct** blocks. (This would also solve #6962 )

SethPoulsen:
fix #6962
fix #6963


SethPoulsen:
@ffund please review if you have a minute (GitHub won't let me request a review from you officially, maybe because you aren't a member of the PrairieLearn organization, I don't know).

nwalters512 (approved):
Looks reasonable!

PR #6994, opened 2023-01-25
Optimize Python calls with large amounts of data

nwalters512:
This reduces a Python call from ~3s to about 200ms 🎉 

This was especially bad because Node feeds us data in chunks of 8kB. For elements that output megabytes of data, we'd do the `indexOf` operation 100s or 1000s of times. Each check only took a few milliseconds, but that was really expensive across so many operations. If I recall from undergrad correctly, that would make this $O(n^2)$ on the length of the data 😭


mwest1066 (approved):
Nice catch!

PR #6995, opened 2023-01-26
Download files instead of inlining into HTML

nwalters512:
This PR aims to reduce server load when students are uploading files that are many megabytes large.

## Background

File uploading was kinda hacked into PL back in the day by yours truly. It worked via base-64-encoded files in the submitted answer. This worked fine when the only thing folks were uploading were small source files (think 10s of kilobytes at most).

However, since then, folks have started uploading very large files, e.g. 5MB Jupyter notebooks with embedded images, or 20MB PDFs. When that data is base-64-encoded and inlined into HTML, it presents a real burden for our HTML-processing machinery. Specifically, getting that data out of the Python process and then running it through an HTML parser takes several seconds for large files, of which a good deal of that is actually *synchronous* parsing that blocks the event loop.

## Improvements

This PR adds a new API route for downloading a file associated with a submission. So, instead of inlining the file's contents in the HTML, we can just pass along a name and rely on the client to lazily fetch the files when they are needed (e.g. to show a preview or to download the file).

It also introduces changes to the elements that are most often used with very large files: `pl-file-upload` and `pl-file-preview`. These elements will use the above API to fetch file contents.


echuber2:
I don't know if it's possible to get away from base64 completely, but it also bloats the size by ~30%.

nwalters512:
@echuber2 we'll most likely do that eventually via an overhaul of the way that files are handled. That's a much larger change though, and something we'll have to be very careful about to avoid breaking existing questions. I wouldn't expect to see any movement there for a while.

elements/pl-file-upload/pl-file-upload.mustache
```diff
@@ -1,15 +1,13 @@
...
-$(function() {
-  var UUID = "{{uuid}}";
-
-  new window.PLFileUpload(UUID, {
-    acceptedFiles: {{&file_names}}
-    {{#has_files}}, files: {{&files}}{{/has_files}}
+$(() => {
+  new window.PLFileUpload("{{uuid}}", {
+    acceptedFiles: {{&file_names}},
+    {{#has_submitted_files}}submittedFileNames: {{&submitted_file_names}},{{/has_submitted_files}}
```
eliotwrobson, line 5: Is the `has_submitted_files` key necessary? Can't you just case on whether `submitted_file_names` is empty?

elements/pl-file-upload/pl-file-upload.py
```diff
@@ -65,16 +65,21 @@ def render(element_html, data):
...
+        # asynchronously once the page loads.
+        submitted_file_names = [x.get("name") for x in filtered_files]
+        html_params["has_submitted_files"] = True
+        html_params["submitted_file_names"] = json.dumps(submitted_file_names, allow_nan=False)
     else:
-        html_params["has_files"] = False
+        html_params["has_submitted_files"] = False
 
     with open("pl-file-upload.mustache", "r", encoding="utf-8") as f:
         html = chevron.render(f, html_params).strip()
```
eliotwrobson, line 85:
```suggestion
        return chevron.render(f, html_params).strip()
```

elements/pl-file-preview/pl-file-preview.py
```diff
@@ -28,20 +29,13 @@ def render(element_html, data):
     # Pass through format errors from the file input elements
     html_params["errors"] = data["format_errors"].get("_files", [])
 
-    # Decode and reshape files into a useful form
+    # Reshape files into a useful form
     if len(submitted_files) > 0:
         files = []
```
eliotwrobson, line 34: Files can now be assembled in one line (and possibly some of the logic above can be simplified?)
```suggestion
        files = [
            {
                    "name": file["name"],
                    "index": idx,
            }
            for idx, file in enumerate(submitted_files)
        ]
```

eliotwrobson (commented):
Is this an improvement on https://github.com/PrairieLearn/PrairieLearn/pull/6988? If so, I think that PR can be closed.

As I mentioned on that PR, this is a good opportunity to add type hints + checking to these Python files.

pages/submissionFiles/submissionFiles.sql
```diff
@@ -0,0 +1,21 @@
...
+        submissions AS s
+        JOIN variants AS v ON (v.id = s.variant_id)
+    WHERE
+        s.id = $submission_id
+        -- We check both `question_id` and `instance_question_id` to make sure that
+        -- this variant is actually accessible by the user. Both will have been
+        -- validated by middleware. For the instructor preview page, we don't have
+        -- an instance question, so we ignore it in that case.
+        AND v.question_id = $question_id
+        AND (($has_instance_question IS FALSE) OR (v.instance_question_id = $instance_question_id))
```
mwest1066, line 16: Elsewhere we've used the pattern that `$instance_question_id` is passed as NULL to indicate not to check it, which would then give
```suggestion
        AND (($instance_question_id::bigint IS NULL) OR (v.instance_question_id = $instance_question_id))
```

mwest1066: (but I'm also fine with the explicit boolean flag if you prefer that)

nwalters512: I like that better! I copied this from #5500 which git-blames back to me 🙃

elements/pl-file-upload/pl-file-upload.mustache
nwalters512: I sure can! This was based on existing code.

pages/submissionFiles/submissionFiles.sql
nwalters512: Changed in 3aa81c1.

elements/pl-file-upload/pl-file-upload.py
```diff
@@ -2,42 +2,40 @@
...
 # file_names attribute
-def get_answer_name(file_names):
+def get_answer_name(file_names: str) -> str:
     return "_file_upload_{0}".format(
         hashlib.sha1(file_names.encode("utf-8")).hexdigest()
     )
 
 
-def add_format_error(data, error_string):
+def add_format_error(data: pl.QuestionData, error_string: str):
```
eliotwrobson, line 32: Functions that don't return anything need a `None` return annotation. Otherwise, the typechecker assumes the return type to be `Any` (in other words, doesn't enforce any behavior).
```suggestion
def add_format_error(data: pl.QuestionData, error_string: str) -> None:
```

elements/pl-file-upload/pl-file-upload.py
```diff
@@ -65,24 +63,25 @@ def render(element_html, data):
...
-        html_params["files"] = json.dumps(filtered_files, allow_nan=False)
-    else:
-        html_params["has_files"] = False
+    files = data["submitted_answers"].get("_files", [])
+    # Filter out any files not part of this element's file_names
+    filtered_files = [x for x in files if x.get("name", "") in file_names]
+    # Only send the file names to the client. We don't include the contents
+    # to avoid bloating the HTML. The client will fetch any submitted files
+    # asynchronously once the page loads.
+    submitted_file_names = [x.get("name") for x in filtered_files]
```
eliotwrobson, line 75:
```suggestion
    submitted_file_names = {x.get("name") for x in files} & set(file_names)
```

elements/pl-file-upload/pl-file-upload.py
```diff
@@ -103,6 +102,7 @@ def parse(element_html, data):
         parsed_files = json.loads(files)
     except ValueError:
```
eliotwrobson, line 103: Probably want to catch if anything happens here (even though it will be a `ValueError` most of the time).
```suggestion
    except Exception:
```

elements/pl-file-upload/pl-file-upload.py
```diff
@@ -49,7 +47,7 @@ def prepare(element_html, data):
     data["params"]["_required_file_names"].extend(file_names)
 
 
-def render(element_html, data):
+def render(element_html: str, data: pl.QuestionData):
```
eliotwrobson, line 50:
```suggestion
def render(element_html: str, data: pl.QuestionData) -> str:
```

elements/pl-file-preview/pl-file-preview.py
```diff
@@ -1,22 +1,15 @@
...
         return ""
 
-    html_params = {"uuid": pl.get_uuid()}
+    html_params = {
+        "uuid": pl.get_uuid(),
+        "submission_files_url": data["options"]["submission_files_url"],
+    }
 
     # Fetch the list of required files for this question
     required_file_names = data["params"].get("_required_file_names", [])
```
eliotwrobson, line 15: No need to make this a variable if used once.
```suggestion
    html_params["required_files"] = data["params"].get("_required_file_names", [])
```

elements/pl-file-preview/pl-file-preview.py
```diff
@@ -1,22 +1,15 @@
-import base64
-
 import chevron
-import lxml.html
 import prairielearn as pl
 
 
-def prepare(element_html, data):
```
eliotwrobson, line 8: Shouldn't we keep this `check_attribs` call to enforce there are no attributes?

nwalters512: I suppose, though I don't think we really care if there are attributes or not.

exampleCourse/questions/demo/autograder/codeUpload/question.html
```diff
@@ -17,7 +17,7 @@
   <p>Your code snippet should define the following variables:</p>
   <pl-external-grader-variables params-name="names_from_user"></pl-external-grader-variables>
 
-  <pl-file-upload file-names="fib.py"></pl-file-upload>
+  <pl-file-upload file-names="fib.py,fib-py.py,img.png,foo.pdf"></pl-file-upload>
```
nwalters512, line 20: TODO: revert before merging.

lib/question.js
```diff
@@ -1392,6 +1393,7 @@ module.exports = {
       urls.clientFilesCourseUrl = urlPrefix + '/clientFilesCourse';
       urls.clientFilesQuestionGeneratedFileUrl =
         questionUrl + 'generatedFilesQuestion/variant/' + variant.id;
+      urls.submissionFilesUrl = questionUrl + 'submissionFiles';
```
nwalters512, line 1396:
```suggestion
```

pages/submissionFiles/submissionFiles.js
```diff
@@ -0,0 +1,78 @@
...
+      return;
+    }
+
+    const buffer = Buffer.from(contents, 'base64');
+
+    // res.sendStatus(404);
+    // return;
+
+    // Artificial delay for testing.
+    // await new Promise((resolve) => setTimeout(resolve, 3000));
```
nwalters512, line 67: TODO: delete before merging.

pages/submissionFiles/submissionFiles.sql
```diff
@@ -0,0 +1,21 @@
...
+    SELECT
+        s.id,
+        jsonb_array_elements(s.submitted_answer->'_files') AS file
+    FROM
+        submissions AS s
+        JOIN variants AS v ON (v.id = s.variant_id)
+    WHERE
+        s.id = $submission_id
+        -- We check both `question_id` and `instance_question_id` to make sure that
+        -- this variant is actually accessible by the user. Both will have been
```
nwalters512, line 12:
```suggestion
        -- this submission is actually accessible by the user. Both will have been
```

elements/pl-file-preview/pl-file-preview.js
```diff
@@ -0,0 +1,124 @@
...
+        downloadButton.addEventListener('click', (event) => {
+          event.stopPropagation();
+          downloadFile(path, file)
+            .then(() => {
+              hideErrorMessage();
+            })
+            .catch((err) => {
+              console.error(err);
+              showErrorMessage();
+            });
```
nwalters512, line 75: The error experience is kind of janky; here's hoping it doesn't happen much!

elements/pl-file-upload/pl-file-upload.js
```diff
@@ -1,278 +1,356 @@
...
-
-  $dropTarget.dropzone({
-    url: '/none',
-    autoProcessQueue: false,
-    accept: function (file, done) {
-      // fuzzy case:
-      const fileNameLowerCase = file.name.toLowerCase();
-      if (_.includes(that.acceptedFilesLowerCase, fileNameLowerCase)) {
-        return done();
+  class PLFileUpload {
```
nwalters512, line 30: I refactored this file into a class, which unfortunately made this difficult to review. I can refactor the existing behavior in another PR to make the diff cleaner if folks would like that.

mwest1066: I think we can probably just test the functionality rather than needing a careful diff? Or are there particular code changes in this file that you'd like eyes on?

nwalters512: I only propose this so that I/we could reason about what actually _did_ change - the diff right now has a lot of false positives.

nwalters512: Good call - that definitely helped a bunch!

server.js
```diff
@@ -928,6 +928,16 @@ module.exports.initExpress = function () {
       require('./pages/generatedFilesQuestion/generatedFilesQuestion'),
     ]
   );
+
+  // Submission files
+  app.use(
+    '/pl/course_instance/:course_instance_id/instructor/instance_question/:instance_question_id/submission/:submission_id/files',
```
nwalters512, line 934: I think `.../file` might be more idiomatic than `/files`? We tend to use singular nouns in most of our routes.

lib/question.js
```diff
@@ -1378,6 +1378,7 @@ module.exports = {
     if (!assessment) {
       // instructor question pages
       const questionUrl = urlPrefix + '/question/' + question.id + '/';
+      urls.questionUrl = questionUrl;
```
mwest1066, line 1381: I think we also need this in the `else` branch of the `if (!assessment)` statement so that student questions will also have this defined. At the moment we get this on student pages:
image

elements/pl-file-preview/pl-file-preview.py
mwest1066: I'd prefer to enforce strict compliance here (don't allow attributes if none are permitted) as a matter of policy. Being strict now can only help us with backwards compatibility in the future.

elements/pl-file-preview/pl-file-preview.js
```diff
@@ -0,0 +1,106 @@
...
+      .map((part) => encodeURIComponent(part))
+      .join('/');
+  }
+
+  class PLFilePreview {
+    constructor(uuid) {
+      const filePreview = document.querySelector('#file-preview-' + uuid);
+      const submissionFilesUrl = filePreview.dataset.submissionFilesUrl;
+
+      filePreview.querySelectorAll('li').forEach((li) => {
```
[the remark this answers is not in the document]
nwalters512: I don't believe this is a regression; `pl-file-upload` displays files in the order that they appear in the `file-names` attribute, whereas `pl-file-preview` displays the files in the order in which they appear in `params._files`. Perhaps we could update the code to display the `pl-file-preview` files in the order in which they appear in `params._required_file_names`?

lib/question.js
nwalters512: Good call! Will add.

elements/pl-file-upload/pl-file-upload.css
```diff
@@ -1,10 +1,7 @@
 [id^=file-upload] .upload-dropzone {
```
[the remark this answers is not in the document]
nwalters512: Good catch, I'll look into this.

nwalters512:
Apologies for the codeowners tag, I messed up a rebase and it appears as though GitHub doesn't remove the review requests after fixing that 🤦

lib/question.js
nwalters512: Fixed in af78b02.

server.js
nwalters512: Changed to `/file` in ac775c4.

elements/pl-file-preview/pl-file-preview.py
nwalters512: I added attribute checking back in fa7326a.

elements/pl-file-upload/pl-file-upload.css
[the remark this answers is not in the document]
nwalters512: Resolved in 3ce14fd - turns out that `dz-message` class was important!

elements/pl-file-preview/pl-file-preview.js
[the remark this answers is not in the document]
nwalters512: This was pretty easy to implement: 7174621. I wrote some tests for this too.

elements/pl-file-preview/pl_file_preview_utils.py
```diff
@@ -0,0 +1,21 @@
+from typing import List
+
+
+def order_files(
+    submitted_files: List[dict], required_file_names: List[str]
```
eliotwrobson, line 5: Lowercase dict annotations ara python 3.10 only feature! Also missing type of this dict.

elements/pl-file-preview/test_pl_file_preview_utils.py
```diff
@@ -0,0 +1,20 @@
+from pl_file_preview_utils import order_files
+
+
+def test_order_files():
```
eliotwrobson, line 4: Missing annotations

eliotwrobson (dismissed):
See comments.

elements/pl-file-preview/pl-file-preview.js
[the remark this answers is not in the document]
mwest1066: Nice!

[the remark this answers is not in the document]
jonatanschroeder: I see that files that are not required are not sorted in any meaningful order. Should they be? I'm thinking of workspaces with wildcards, which don't appear as required files.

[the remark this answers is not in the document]
mwest1066: Good point. Would it be cleaner to always alphabetize?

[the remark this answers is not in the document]
nwalters512: Good point - plain old alphabetization is almost certainly better. I'll make that change.

elements/pl-file-preview/pl-file-preview.js
```diff
@@ -0,0 +1,106 @@
...
+              } else if (type.startsWith('image/')) {
+                const url = URL.createObjectURL(blob);
+                img.src = url;
+                img.onload = () => {
+                  URL.revokeObjectURL(url);
+                };
+                img.classList.remove('d-none');
+              } else {
+                // We can't preview this file.
+                code.textContent = 'Content preview is not available for this type of file.';
```
mwest1066, line 90: In this case can we write this text not inside the code element? For example, in the files below it looks like `foo.pdf` has rendered some content until you read the text. I'm thinking it might be better to just have plain text for the message.

image

elements/pl-file-preview/pl-file-preview.js
```diff
@@ -0,0 +1,106 @@
...
+                pre.classList.remove('d-none');
+              }
+              wasOpened = true;
+              hideErrorMessage();
+            })
+            .catch((err) => {
+              console.error(err);
+              showErrorMessage('An error occurred while downloading the file.');
```
mwest1066, line 98: `showErrorMessage()` doesn't seem to take an argument?

mwest1066 (approved):

elements/pl-file-preview/pl-file-preview.js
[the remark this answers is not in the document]
nwalters512: Alphabetization was implemented in 754f260.

elements/pl-file-preview/pl-file-preview.js
nwalters512: Good catch, will remove.

nwalters512: Actually, I might repurpose the error display for the "Content preview is not available for this type of file" message, in which case I'll actually change `showErrorMessage` to accept a string.

elements/pl-file-preview/pl-file-preview.js
nwalters512: Yep - I tweaked this in 19a115b. How does this look?

Screenshot 2023-02-07 at 15 28 52

exampleCourse/questions/demo/autograder/codeUpload/question.html
nwalters512:
```suggestion
  <pl-file-upload file-names="fib.py"></pl-file-upload>
```

elements/pl-file-preview/pl-file-preview.mustache
```diff
@@ -1,43 +1,45 @@
...
           </div>
-          <div class="file-preview collapse" id="file-preview-contents-{{uuid}}-{{index}}">
-            <pre class="bg-dark text-white rounded p-3 mb-0"><img class="w-100" src="data:application/octet-stream;base64,{{contentsb64}}" onerror="$(this).hide();$(this).parent().find('code').removeClass('d-none');"><code class="d-none">{{contents}}</code></pre>
+          <div class="align-self-center">
+            <button class="btn btn-outline-secondary btn-sm file-preview-download-file">
+              Download
+            </button>
+            <button type="button" class="btn btn-outline-secondary btn-sm file-preview-button">
+              <i class="file-preview-icon fa fa-angle-down"></i>
+            </button>
```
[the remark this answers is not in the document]
nwalters512: Hmm, I can't reproduce that with a text file submission on the latest commit to this branch. Is there a particular file type (e.g. image or binary file) that you tested this with?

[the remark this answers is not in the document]
nwalters512: Chatted with Matt offline; he was unable to reproduce this.

mwest1066 (approved):

PR #6996, opened 2023-01-27
Add fixed overhead to external grader timeout

nwalters512:
When instructors specified a really short timeout for their external grading jobs, our naive approach of computing an overall timeout by multiplying the question timeout by 2 fell apart, as that couldn't account for the overhead of pulling the image and so on. So, we'll instead just add a generous amount of time (default 5 minutes) to the instructor timeout to account for this.

Closes https://github.com/PrairieLearnInc/sysconf/issues/781.


grader_host/config/base.yaml
```diff
@@ -74,9 +74,14 @@ properties:
   resultsQueueUrl:
     default: null
     envVar: RESULTS_QUEUE_URL
+
+  # Timeouts
```
nwalters512, line 78: Man, who introduced this bespoke config format 😉

mwest1066: 🤣

grader_host/index.js
```diff
@@ -406,8 +411,8 @@ function runJob(info, callback) {
   const globalJobTimeoutId = setTimeout(() => {
```
mwest1066, line 411: Rename this variable to `jobTimeoutId` for consistency with the fact that we aren't using `globalJobTimeout` anymore?

mwest1066 (approved):

grader_host/index.js
nwalters512: Done in 0ec0f34.

mwest1066 (approved):

PR #7001, opened 2023-01-27
Remove unused database functions

nwalters512:
These callback database functions are unused, so we can safely kill them off.


docs/dev-guide.md
```diff
@@ -458,48 +458,6 @@ sqldb.queryOneRow(sql.block_name, params, function (err, result) {
...
-  sqldb.queryWithClient(client, sql.block_name, params, function (err, result) {
-    if (ERR(err, callback)) return; // THIS IS WRONG, it may exit without endTransaction()
-    sqldb.endTransaction(client, done, err, function (err) {
-      if (ERR(err, callback)) return;
-      callback(null);
-    });
-  });
-});
-```
-
```
mwest1066, line 502: Should we replace this with documentation of how to use `runInTransactionAsync()`, or are we moving to a documentation model where we'll document this in the `sqldb` package and expect developers to find it there?

lib/assessment.js
```diff
@@ -108,72 +108,27 @@ module.exports = {
...
-                        debug('updated:', updated);
-                        callback(null);
-                      }
-                    );
-                  },
-                  (callback) => {
-                    if (!updated) return callback(null); // skip if not updated
+        async () => {
+          await sqldb.runInTransactionAsync(async (client) => {
+            await sqldb.callWithClientAsync(client, 'assessment_instances_lock', [
```
mwest1066, line 113: In PT I've been relying on your code that uses AsyncLocalStorage to ensure that `callAsync()` inside `runInTransactionAsync` will automatically use the correct client, without needing to explicitly get the client and pass it down. Should we be promoting this as the preferred pattern?

mwest1066 (approved):

docs/dev-guide.md
nwalters512: Over on https://github.com/PrairieLearn/PrairieLearn/pull/7003#discussion_r1089378055 I brought up basically the same question. I do think there's space for an overarching dev guide (although I think that should be multiple pages and not one monolithic document), but I don't know that `docs/dev-guide.md` will be a useful starting point for that. I'm leaning towards just deleting for now and worrying about improved dev docs later, as I don't think this giant doc is particularly useful in its current form.

lib/assessment.js
nwalters512: Yes! Good call; I'll make that change.

nwalters512: Done in 6bfd6ae.

docs/dev-guide.md
mwest1066: I agree the giant doc is not very useful. Per-package docs seem like the highest priority to me (as we pull `sqldb` and similar things out into standalone packages). But this can wait.

PR #7003, opened 2023-01-27
Introduce @prairielearn/postgres package

### LINKED ISSUE #6997: Move Postgres code to @prairielearn/postgres package
This will allow us to use this code within PrairieTest without relying on a PrairieLearn submodule. This will help unblock work to start writing portions of PrairieLearn in TypeScript.

nwalters512:
Closes #6997.


docs/dev-guide.md
```diff
@@ -149,9 +149,9 @@ var ERR = require('async-stacktrace');
 var _ = require('lodash');
 var express = require('express');
 var router = express.Router();
-var sqldb = require('../prairielib/sql-db'); // adjust path as needed
+var sqldb = require('@prairielearn/postgres') // adjust path as needed
```
nwalters512, line 152: I'm wondering if we should significantly trim back the dev guide? A lot of the advice is outdated, incorrect, or doesn't reflect the current patterns we're following.

packages/postgres/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export { loadSql, loadSqlEquiv } from './loader';
```
mwest1066, line 1: TS!!!! 🚀

docs/dev-guide.md
mwest1066: I'm happy to have it significantly trimmed back. Most of the code-style stuff is horribly outdated at this point.

packages/postgres/src/pool.ts
```diff
@@ -254,40 +229,37 @@ class PostgresPool {
     // It does NOT support clearing the existing search schema - e.g.,
     // `setSearchSchema(null)` would not work as you expect. This is fine, as
     // that's not something we ever do in practice.
-    if (this.searchSchema != null && client[SEARCH_SCHEMA] !== this.searchSchema) {
+    if (this.searchSchema != null && (client as any)[SEARCH_SCHEMA] !== this.searchSchema) {
```
mwest1066, line 232: Can we say it's a `pg.PoolClient`?

mwest1066 (approved):

packages/postgres/src/pool.ts
nwalters512: This `any` cast is necessary because we try to access a property (`SEARCH_SCHEMA`) that isn't declared in the types.

If we wanted to avoid this, we could probably instead maintain a `WeakMap` from client to the search schema that it's using?

mwest1066: Oh right, we've (probably me) done a horrible thing here. I agree the clean thing (for TS) is to use a WeakMap. I'm ok if you don't want to make that change in this PR.

nwalters512: I went ahead and made this change in 5ae6ac2. I'd appreciate one final look to make sure I didn't mess anything up!

mwest1066 (approved):

PR #7007, opened 2023-01-27
Format exampleCourse and testCourse with Black and isort

nwalters512:
This PR formats all Python files within `exampleCourse/` and `testCourse/` with Black and isort.

Note that I have deliberately not added those directories to the flake8 config yet; I'll do that in another PR. Splitting them up should make review easier:

- This first PR is 100% automated formatting changes.
- The next PR will include manual changes that should be more closely reviewed.


eliotwrobson (approved):
Didn't look super close but if this is just running black and isort then should be fine.

PR #7008, opened 2023-01-28
Lint exampleCourse and testCourse with flake8

nwalters512:
Second part of #7007.

The vast majority of the changes here were just unused imports. There were one or two actual bugs.

I had to carve out some exceptions for files related to the Python autograder, as many of those files are by design not valid Python in isolation. See `pyproject.toml` for more details.


exampleCourse/questions/demo/computeRotationMatrix/server.py
```diff
@@ -14,11 +14,11 @@ def generate(data):
     M[1, 0] = math.sin(math.pi / a)
     M[1, 1] = math.cos(math.pi / a)
     if a < 0:
-        data["params"]["a"] = "- \\frac{\pi}{" + str(-a) + "}"
-        data["params"]["b"] = "R_{-\\frac{\pi}{" + str(-a) + "}}"
+        data["params"]["a"] = "- \\frac{\\pi}{" + str(-a) + "}"
```
eliotwrobson, line 17: Can we make these raw strings instead of using escape characters? Maybe also worth using a format string here.

testCourse/serverFilesCourse/python_autograder/pltest.py
```diff
@@ -91,7 +91,7 @@ def getResults(self):
 
         with open("results.json", mode="w") as out:
             json.dump(grading_result, out)
-    except:
+    except:  # noqa: E722
```
eliotwrobson, line 94: Maybe just use an Exception here? This looks like it doesn't need to catch system calls.

Might also be worth assembling some of the dicts in this file with one liners.

exampleCourse/questions/demo/autograder/python/square/server.py
```diff
@@ -1,6 +1,3 @@
-import numpy as np
-
-
 def generate(data):
     data["params"]["names_for_user"] = [
```
eliotwrobson, line 2: Might be worth inlining some of these descriptions with the new element feature? Not strictly necessary but some of these files are getting cleaned out anyway

eliotwrobson (approved):
Overall looks ok. A few items worth considering for further cleanup but not strictly necessary.

testCourse/serverFilesCourse/python_autograder/pltest.py
nwalters512: I'm actually wondering if this file is even needed - AFAICT it's not referenced or used at all within the `testCourse`?

eliotwrobson: This might have been used as an example custom autograder? Not sure if that's necessary though.

exampleCourse/questions/demo/autograder/python/square/server.py
nwalters512: I'll let that wait for another PR to keep this PR focused on linting changes.

exampleCourse/questions/demo/computeRotationMatrix/server.py
nwalters512: With a format string we'd need to use `{{{` for the literal curly braces, which IMO is less readable than just doing string concatenation. And I'm also inclined to not use r-strings for the potentially stupid reason that my editor formats them in such a way as to still make `\f` seem like an escaped character:

Screenshot 2023-01-27 at 16 58 34

testCourse/serverFilesCourse/python_autograder/pltest.py
nwalters512: Maybe, though that would belong in `exampleCourse` if that were true. Let me open another PR to remove that and see if the tests still pass 🤷

exampleCourse/questions/demo/computeRotationMatrix/server.py
eliotwrobson: That's very weird, what editor are you using? VS code seems to be ok with raw strings.

testCourse/serverFilesCourse/python_autograder/pltest.py
nwalters512: #7010

exampleCourse/questions/demo/computeRotationMatrix/server.py
nwalters512: This is VSCode too 👀 maybe it's a rogue extension?

eliotwrobson: Do you have pylance installed?

nwalters512: I do!

eliotwrobson: May be related to this: https://github.com/microsoft/vscode-python/issues/13702

nwalters512: Yep, alas, it's this issue: https://github.com/MagicStack/MagicPython/issues/114

eliotwrobson: Ahhh, yeah I guess go with capital R

pyproject.toml
```diff
@@ -1,19 +1,51 @@
...
+exclude = [
+    # These files are intentionally not parseable as Python.
+    "./exampleCourse/questions/demo/autograder/codeUpload/submitted_code_samples/fib_error_binary_file/fib.py",
+    "./exampleCourse/questions/demo/autograder/codeUpload/submitted_code_samples/fib_error_large_text_file/fib.py",
+]
+per-file-ignores = [
+    # Files related to the Python autograder will often intentionally appear
+    # broken in isolation. We'll allow specific errors in these files to
+    # account for that.
+    "exampleCourse/questions/**/tests/setup_code.py: F401, F821",
```
eliotwrobson, line 18: Looking again, does `setup_code` need to have F821? That's an error for undefined variable names, but the `setup_code.py` should be setting all variable names that get used later (I think).

eliotwrobson: After looking, realized that `setup_code.py` can use stuff from the `data` dict.

PR #7009, opened 2023-01-28
Add @types/lodash and fix type issues

nwalters512:
The types for lodash will be pulled in by #7003. Since they require small code changes to make type checking succeed, I'm opting to make those changes in isolation.


cron/index.js
```diff
@@ -151,9 +151,12 @@ module.exports = {
 
     const jobsByPeriodSec = _.groupBy(module.exports.jobs, 'intervalSec');
     _.forEach(jobsByPeriodSec, (jobsList, intervalSec) => {
+      const intervalSecNum = Number.parseInt(intervalSec);
       if (intervalSec === 'daily') {
         this.queueDailyJobs(jobsList);
-      } else if (intervalSec > 0) {
+      } else if (Number.isNaN()) {
```
nwalters512, line 157:
```suggestion
      } else if (Number.isNaN(intervalSecNum)) {
```

PR #7016, opened 2023-01-30
Fix bug when question is renamed to a subdirectory of itself

nwalters512:
See inline descriptions of the problem and the fix.

### DIFF
--- a/lib/chunks.js
+++ b/lib/chunks.js
@@ -940,6 +940,8 @@ const ensureChunk = async (courseId, chunk) => {
       const stat = await fs.lstat(parentPath);
       if (stat.isSymbolicLink()) {
         await fs.remove(parentPath);
+      } else if (!stat.isDirectory()) {
+        throw new Error(`${parentPath} exists but is not a directory`);
       }
     } catch (err) {
       if (err.code !== 'ENOENT') throw err;
--- a/lib/chunks.js
+++ b/lib/chunks.js
@@ -925,6 +925,27 @@ const ensureChunk = async (courseId, chunk) => {
     cwd: unpackPath,
   });
 
+  // Before we configure the symlink, we need to check if there are any
+  // outdated symlinks that need to be removed. Those can occur when a question
+  // is renamed into a directory nested inside of its former directory, e.g.
+  // if `questions/a/b/info.json` is moved to `questions/a/b/c/info.json`.
+  //
+  // We'll handle this by checking if any parent directory of the `targetPath`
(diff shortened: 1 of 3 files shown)


lib/chunks.js
```diff
@@ -925,6 +925,27 @@ const ensureChunk = async (courseId, chunk) => {
...
+  // exists and is a symlink. If so, we'll remove it. This should always be
+  // safe because we should never have nested symlinks.
+  const pathSegments = relativeTargetPath.split(path.sep);
+  for (let i = 1; i < pathSegments.length; i++) {
+    const parentPath = path.join(courseChunksDirs.course, ...pathSegments.slice(0, i));
+    try {
+      const stat = await fs.lstat(parentPath);
+      if (stat.isSymbolicLink()) {
+        await fs.remove(parentPath);
+      }
```
mwest1066, line 943: For an added check we could do something like:
```suggestion
      } else if (!stat.isDirectory()) {
        throw new Error(`${parentPath} exists but is not a directory`, stat);
      }
```

tests/chunks.test.js
```diff
@@ -311,7 +313,60 @@ describe('chunks', () => {
...
     });
 
-    it('deletes chunks that are no longer needed', async () => {
+    it('handles question nesting after a rename', async () => {
+      // Scenario: there's a question named `foo/bar` (that is,
+      // `foo/bar/info.json` exists). We load the chunk for that
+      // question. We then move that `info.json` file to
+      // `foo/bar/baz/info.json`. We then try to load that chunk again. In the
+      // past, the new chunk would be written to an invalid location. This test
+      // ensures that it's written correctly.
```
mwest1066, line 322: Such a sneaky scenario!

mwest1066 (approved):

lib/chunks.js
nwalters512: Ah, interesting. That might actually be a scenario in which we should do something useful? Consider the following course, where `bar` is an extensionless file:

```
├── questions
│   ├── foo
│   │   ├── info.json
│   │   ├── bar
```

At some point that's changed to the following:

```
├── questions
│   ├── foo
│   │   ├── bar
│   │   │   ├── info.json
```

I'm pretty sure our code won't handle this case; maybe it should? This is admittedly a pretty contrived situation.

nwalters512: Actually it turns out that this new code handles this just fine! I'll accept your suggestion then.

Issue #5542, opened 2022-03-26
symbolic-input questions with square roots


jake-bobowski:
When using SymPy in Python, the following block of code will return False when executed:

import sympy as sp
b = sp.symbols('b')
b == sp.sqrt(b**2)  

This happens because if b < 0, then sp.sqrt(b**2) will be positive and not equal to b.  If it is known that b must be positive, SymPy allows for assumptions such that the following block of code will return True when executed:  

import sympy as sp
b = sp.symbols('b', positive = True)
b == sp.sqrt(b**2)  

(One can also specify that a variable is Real.)

I've run into an issue in one of my symbolic PrairieLearn problems.  The problem requires students to calculate the radius of the resulting circular motion when a charged particle is accelerated through potential V before entering a uniform magnetic field B.  The solution is:

r = (1/B)*(2*V*m/q)**(1/2)

The issue is that PrairieLearn considers the equivalent expression:

r = (2*V*m/(q*B**2))**(1/2)

to be incorrect.  This response is marked incorrect even when explicitly including positive = True when defining the B symbol in the server.py file.  Is there away to assume conditions on variables used in symbolic PrairieLearn problems?

firasm:
I had a look at this, and this seems like a sympy limitation.

```
import sympy as sp

B, V, m, q = sp.symbols('B V m q', positive = True, real = True)

r1 = (1/B)*(2*V*m/q)**(1/2)
r2 = (2*V*m/(q*B**2))**(1/2)
```

I've tried `.expand()` and `.simplify()` and I cannot convince `sympy` to think they're the same.

<img width="642" alt="Screen Shot 2022-03-27 at 11 52 34 AM" src="https://user-images.githubusercontent.com/2507459/160296233-e4bbe40c-b972-4997-bebe-1a2a8494f63c.png">

## A workaround

A workaround within PrairieLearn could be to award points by accepting both versions of the answer as an override to the `grade()` function.

See [this section](https://prairielearn.readthedocs.io/en/latest/question/#question-serverpy) in the docs.

You want something like this (pseudo-code):

```
def grade(data):

    if data["score"] == 0: # only if not already correct
            if data["submitted_answers"]["XXXX"] == "second sympy expresssion":
                data["score"] = 1.0
```

firasm:
[...]

PR #7020, opened 2023-01-31
feat: Added basic assumption preservation support

eliotwrobson:
See title. Hasn't been tested extensively yet (and not yet with any live questions), but the local test case written seems to work. Resolves https://github.com/PrairieLearn/PrairieLearn/issues/1844 and resolves https://github.com/PrairieLearn/PrairieLearn/issues/5542. Should also include a test for backwards compatibility (but I don't expect this to be an issue).


eliotwrobson:
@jake-bobowski does this satisfy your use case on its own? Not sure if the issue you ran into would require changes to the symbolic input element's grader code.

jake-bobowski:
@eliotwrobson Yes, I think that preserving the sympy assumptions will resolve the issues I was running into with symbolic questions.  Thanks!

eliotwrobson:
@jake-bobowski great! Just wanted to be sure. It would be nice to be able to integrate this into the symbolic input element, but I don't see any immediate way to do so and if it resolves the issues you're running into, then i don't think there's really a need.

eliotwrobson:
@jake-bobowski do you happen to have a question that would benefit from this enhancement lying around you'd be willing to donate? If so, I'd like to add it to the example course to test to make sure this fixes the issue you were running into (since I'm not 100% sure how exactly this serialization was intended interacting with the element grader code).

jake-bobowski:
Yes, here is the question.html file:

```python
<pl-question-panel>
A particle of charge $q$ and mass $m$ is accelerated from rest through a potential difference $V$, after which it encounters a uniform magnetic field $B$.
If the particle moves in a plane perpendicular to $B$, what is the radius of its circular orbit?
</pl-question-panel>
<p></p>

<p>
<pl-symbolic-input answers-name="part1_ans" label = "$r = $" variables = "m, B, q, V" weight = "1" allow-blank = "False" ></pl-symbolic-input>
</p>
```

and here is the server.py file:
```python
import sympy as sp
import prairielearn as pl

def generate(data):

    # Declare math symbols to be used by sympy
    m, q, V = sp.symbols('m q V')
    
    # Assume that B is positive
    B = sp.symbols('B', positive = True)
    
    # Describe the solution equation
    r = (1/B)*sp.sqrt((2*V*m)/q)
    
    # Answer to fill in the blank input stored as JSON.
    data['correct_answers']['part1_ans'] = pl.to_json(r)

def prepare(data):
    pass
    
def parse(data):
    pass
    
def grade(data):
    pass
```

A solution that gets graded as correct is:
```math
r = \frac{1}{B}\sqrt{\frac{2Vm}{q}}
```

Here's a screenshot of an answer being marked correct:

image

Last year, if the equivalent expression:
```math
r = \sqrt{\frac{2Vm}{q B^2}}
```
was entered it would be marked as incorrect.  I don't know why, but at the moment entering this alternative solution now causes the following error that I don't understand:
Git

I assume that this is a bug that was somehow introduced to PL and not an issue with my simple problem.

Thanks!

Jake

eliotwrobson:
@jake-bobowski yes, that was a bug from some recent refactors. The fix actually got merged (but not yet deployed), but that error effectively means that the equality comparison was false (would be treated as an incorrect answer). Thank you for the question code! I'll play around with it and see if the element will accept the answer. It may require adding the symbol assumptions dictionary to the parsing logic for student answers, but that should be a straightforward addition.

eliotwrobson:
@jake-bobowski So I tried a simple example of this locally and still having the issue where `sqrt(B**2)` is not equal to B when submitted as an answer by the student. It looks like the fix is to apply assumptions when parsing the student input. However, I'm not sure then what the best way is to incorporate this system into the existing element.

I think the system is a bit too complicated to integrate into the HTML attributes of the element. We could just use all the assumptions set for variables in the correct answer, but this may not work in all situations (i.e. the correct answer is not set but we still want to apply assumptions while parsing student input). Should using the assumptions in the correct answer be the default behavior, then have the option to have these overridden by setting a specific parameter? I think that it would be great if we could incorporate this into the element somehow, but I think there's enough layers to it that we should be careful about how this gets implemented.

eliotwrobson:
Open to wording changes, but added popover showing student's original input.

exampleCourse/questions/element/symbolicInput/question.html
```diff
@@ -85,3 +85,21 @@
...
+        <pl-question-panel>
+            <p>Sometimes, certain expressions can only be treated as equivalent by sympy if additional assumptions are
+                made. For example, consider that $\sqrt{B^2} = B$ is only true if $B \geq 0$. To use these,
+                assumptions must be set in the <code>server.py</code> file when variables are declared.
+            </p>
+            <p>
+                Submit an expression equivalent to $\sqrt{B^2}$ when assuming $B > 0$.
+            </p>
+        </pl-question-panel>
+        <pl-symbolic-input answers-name="assumptions" variables="B"></pl-symbolic-input>
```
cbm755, line 103: Another nice example here might be:

> Simplify $|C|$ while assuming $C \leq 0$.

```
C = sympy.symbols("C", nonpositive=True)
```

Where the correct answer is "-C".

python/python_helper_sympy_test.py
```diff
@@ -91,13 +95,29 @@ def test_string_conversion_no_complex(
     def test_json_conversion(self, a_sub: str) -> None:
         sympy_expr = phs.convert_string_to_sympy(a_sub, ["n", "m"], allow_complex=True)
         # Check that json serialization works
-        json_expr = phs.sympy_to_json(sympy_expr)
-        assert type(json.dumps(json_expr)) == str
+        json_expr = json.dumps(phs.sympy_to_json(sympy_expr), allow_nan=False)
```
cbm755, line 98: (drive by comment, ignore if not helpful)

`json.dumps(sympy_to_json(...))` seems a bit like "boil the boiling water" to me.  Is this b/c of no-NaN-JSON-thing?  (e.g., https://stackoverflow.com/questions/6601812/sending-nan-in-json)

Maybe a comment here would clarify?

eliotwrobson: So the dumps turns the `sympy_to_json(...)` into a serialized string, simulating what PL does on the backend for a question. `sympy_to_json` just turns a sympy expression into a format that can be JSON serialized, not doing the serialization itself. This part of the test is just asserting that operation all worked as expected.

cbm755: Got it!  (`sympy_prep_for_json` perhaps a more accurate name then---definitely not suggesting you change it!)

exampleCourse/questions/element/symbolicInput/question.html
eliotwrobson: Added here! https://github.com/PrairieLearn/PrairieLearn/pull/7020/commits/54332483444b7161f8889e14432df20a2bcc06bc#diff-1ffb6b7efae669fdf09b47d0a0fe77a316394be961705a07c34f37db40e4ef0e

jonatanschroeder:
You may want to clarify in the PR title (which usually is used as basis for the master commit and then in the deployment announcement) that this is for symbolic inputs.

elements/pl-symbolic-input/pl-symbolic-input.py
```diff
@@ -147,13 +154,17 @@ def render(element_html: str, data: pl.QuestionData) -> str:
...
+                        variables,
+                        allow_complex=allow_complex,
+                        custom_functions=custom_functions,
                     )
                 else:
-                    a_sub = phs.json_to_sympy(a_sub, allow_complex=allow_complex)
-                a_sub = a_sub.subs(sympy.I, sympy.Symbol(imaginary_unit))
-                html_params["a_sub"] = sympy.latex(a_sub)
+                    a_sub_parsed = phs.json_to_sympy(a_sub, allow_complex=allow_complex)
+                a_sub_parsed = a_sub_parsed.subs(sympy.I, sympy.Symbol(imaginary_unit))
```
eliotwrobson, line 165: Note: maybe remove this line?

python/python_helper_sympy_test.py
SethPoulsen: or maybe `sympy_to_pyobj`? its just converting it to a python object/dict or something like that, right?

eliotwrobson: `sympy_to_pyobj` is more accurate, but I'm against changing the name of this for backwards compatibility and consistency with the `to_json` and `from_json` functions.

SethPoulsen (commented):
I don't have too much to say. This looks good. If you want to be sure about backward compatibility it would be good in local dev to submit answers to some problems with the old version of the element, then try re-opening them with the new version of the element.

elements/pl-symbolic-input/pl-symbolic-input.py
```diff
@@ -70,8 +72,12 @@ def render(element_html: str, data: pl.QuestionData) -> str:
...
     name = pl.get_string_attrib(element, "answers-name")
     label = pl.get_string_attrib(element, "label", LABEL_DEFAULT)
-    variables_string = pl.get_string_attrib(element, "variables", VARIABLES_DEFAULT)
-    variables = phs.get_variables_list(variables_string)
+    variables = phs.get_items_list(
+        pl.get_string_attrib(element, "variables", VARIABLES_DEFAULT)
+    )
+    custom_functions = phs.get_items_list(
+        pl.get_string_attrib(element, "custom-functions", CUSTOM_FUNCTIONS_DEFAULT)
+    )
```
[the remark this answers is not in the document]
eliotwrobson: Yes; the changes here are pretty far-reaching and there could still be changes to how this is all set up based on feedback. I'm holding off on adding docs until we're sure this will be the final behavior. There still needs to be documentation on how the assumptions system is set up and some discussion of what's allowed by the new parser.

elements/pl-big-o-input/big_o_utils_test.py
```diff
@@ -120,9 +120,11 @@ def test_negative_submission(
             ("n**2", "n**3"),
             ("n**2", "n**4"),
             ("n**2", "factorial(n)"),
+            ("n**2", "n!"),
             ("n**2", "n**2*log(n)"),
             ("2**n", "n**n"),
             ("2**n", "factorial(n)"),
+            ("n**2", "n!"),
```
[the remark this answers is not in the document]
eliotwrobson: Fixed in https://github.com/PrairieLearn/PrairieLearn/pull/7020/commits/89878fcad87ecdae918dcfeb669a7853e2d5f9be

elements/pl-symbolic-input/pl-symbolic-input.py
eliotwrobson: Realizing this is just to make the display more friendly.

apps/prairielearn/python/python_helper_sympy.py
```diff
@@ -529,20 +667,16 @@ def validate_string_as_sympy(
...
-    if variables_string is None:
+def get_items_list(items_string: Optional[str]) -> list[str]:
+    if items_string is None:
         return []
 
-    return list(map(str.strip, variables_string.split(",")))
+    return list(map(str.strip, items_string.split(",")))
 
 
 def process_student_input(student_input: str) -> str:
```
eliotwrobson, line 677: Note to self: Remove this function and have this be part of the evaluation function.

elements/pl-symbolic-input/pl-symbolic-input.py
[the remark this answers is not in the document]
eliotwrobson: Docs have been added as of latest commit.

eliotwrobson:
@nwalters512 @SethPoulsen After going through all of the partially finished items in this PR, this is ready for final review (cc @cbm755 ).

Makefile
```diff
@@ -46,7 +46,7 @@ test-python:
 # `pl_unit_test.py` has an unfortunate file name - it matches the pattern that
 # pytest uses to discover tests, but it isn't actually a test file itself. We
 # explicitly exclude it here.
-	@python3 -m pytest --ignore graders/python/python_autograder/pl_unit_test.py
+	@python3 -m pytest --ignore graders/python/python_autograder/pl_unit_test.py --cov=apps
```
SethPoulsen, line 49: I feel like adding coverage reports should be a separate PR, but that's up to @nwalters512 really.

apps/prairielearn/elements/pl-big-o-input/pl-big-o-input.mustache
```diff
@@ -109,18 +118,28 @@ ${{type}}({{a_tru}})$
...
+</p>
+{{/constants.0}}
+{{#operators.0}}
+<p>
+    <strong>Allowable operators:</strong><br>
+    {{#operators}} <code class="user-output">{{.}}</code> &nbsp; {{/operators}}
+</p>
+{{/operators.0}}
+
+<p>Note that either <code class="user-output">^</code> or <code class="user-output">**</code> can be used for exponentiation.</p>
```
SethPoulsen, line 142: is there actually a code change here or just formatting? I'd prefer to not have that in this PR.

apps/prairielearn/python/python_helper_sympy.py
```diff
@@ -288,9 +298,9 @@ def evaluate(expr: str, locals_for_eval: LocalsForEval) -> sympy.Expr:
     # https://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html
     # http://blog.delroth.net/2013/03/escaping-a-python-sandbox-ndh-2013-quals-writeup/
     #
-    whitelist: ASTWhitelistT = (
-        ast.Module,
-        ast.Expr,
+    # Note some whitelist items were removed. If there are compatibility issues, try
+    # adding those items back.
+    whitelist: ASTWhiteListT = (
```
SethPoulsen, line 303: Who is this comment supposed to be here for? 
Most people who use this element are not going to see this is they have problems, they are going to come to pl-help on slack. 

Is there a reason to remove some Whitelist items? If there is a new security problem with some of them, then I'm all for it, but aside from that I don't know that we should remove things.

SethPoulsen (approved):
Looks great, cool new features. 

Just a few minor comments.

Makefile
eliotwrobson: I thought about splitting it out, but this was a 2 line change anyway.

apps/prairielearn/elements/pl-big-o-input/pl-big-o-input.mustache
eliotwrobson: The text that is displayed was actually changed.

apps/prairielearn/python/python_helper_sympy.py
eliotwrobson: The whitelist for this should be as short as possible. This PR changed the way that the parser works so that certain things that were allowed before no longer need to be for test cases to pass, so I removed the unneeded items.

This comment is basically just a reminder that the removal of items could be the culprit if people run into weird errors. I'm happy to remove the comment if it seems redundant.

SethPoulsen: Ideally ya, the whitelist should be as short as possible, but we are trying to avoid breaking people's questions. 
I don't think we have reason to believe the test cases we have are covering all the cases people are using this element for. 

But anyway, if you want to try to shrink the whitelist, right now is not a bad time since it is summer, just be ready to revert if we have people complain that we broke their questions.

Makefile
nwalters512: Yeah, I'm fine to add this here. I guess we should revisit JS test coverage reports as well?

apps/prairielearn/python/python_helper_sympy.py
nwalters512: I've made comments like these in the past, but going back to them even six months in the future, they've almost always had very little utility. If we're going to keep this comment, I'd like to at least link to this PR so we can easily discover what was removed and when.

```suggestion
    # Note some whitelist items were removed in this PR:
    # https://github.com/PrairieLearn/PrairieLearn/pull/7020
    # If there are compatibility issues, try adding those items back.
    whitelist: ASTWhiteListT = (
```

nwalters512 (approved):
My eyes glazed over a bit on the `python_helper_sympy` changes; thanks for the tests to give us confidence in our implementation!

PR #7083, opened 2023-02-01
Upgrade plbase Python dependencies

nwalters512:
Combining various Dependabot PRs into one.

The only major version bump here is Black, which introduced some new changes to how code is formatted. Those changes make up the bulk of this diff.


elements/pl-matching/pl-matching.py
```diff
@@ -210,15 +210,15 @@ def prepare(element_html, data):
     # Build the options to display to the student.
     chosen_option_names = []
     display_options = []
-    for (i, opt) in enumerate(options):
+    for i, opt in enumerate(options):
```
eliotwrobson, line 213: This is actually a pretty cool addition IMO, I always try to shoot for not having the parens here when it's not necessary.

eliotwrobson (approved):
LGTM!

PR #7084, opened 2023-02-01
Upgrade all Python grader dependencies

nwalters512:
Combining various Dependabot PRs into one.


eliotwrobson (approved):
Seems fine, these don't even seem like major version bumps.

PR #7086, opened 2023-02-02
Alphabetize Python dependency files

nwalters512:


eliotwrobson (commented):
Was this done with a script? There are apparently ones that do this: https://pypi.org/project/sort-requirements/

 Line numbers seem to be equal so seems ok.

nwalters512:
I have a VSCode extension that can alphabetically sort lines; I used that!

PR #7090, opened 2023-02-02
Fix use of additionalProperties in schema

jonatanschroeder:
Seems like questions that still include `syncIgnore` will fail sync because of incorrect type. This seems to fix it.


nwalters512:
I think I misunderstood what `additionalProperties` does. I think your previous recommendation of moving the `syncIgnore` property back outside of `additionalProperties` is actually the correct fix, if you'd like to make that change instead.

nwalters512 (approved):

PR #7091, opened 2023-02-02
Add support for dynamic workspace files based on mustache templates

### LINKED ISSUE #7089: Support for workspace templates using mustache
> I like the flexibility of this mechanism, but I'm a bit worried that it has a lot of overhead for most users. Could we also support a `workspaceTemplates/` directory that contains files which we will process with mustache (just like `question.html`, with the same `data` available) and copy into the workspace? I'm thinking the precedence would be (1) files from `workspace/`, (2) files files from `workspaceTemplates/`, and finally (3) files from `_files`.

_Originally posted by @mwest1066 in https://github.com/PrairieLearn/PrairieLearn/pull/6947#discussion_r1094651296_

jonatanschroeder:
Resolves #7089 .


docs/workspaces/index.md
```diff
@@ -156,11 +156,27 @@ A minimal `question.html` for an externally graded workspace should look somethi
...
+Questions using workspaces can also be randomized, i.e., include files that contain random and dynamic content. This may be done in two ways: using mustache-based template files or using [the `server.py` file in the question directory](../question.md#question-serverpy). For template files, a workspace question can optionally include a `workspaceTemplates/` subdirectory within the regular question directory structure. The contents will be copied into the home directory of the student's workspace container, as with the `workspace/` directory. However, all files within this directory may include mustache tags (e.g., `{{params.value}}`), which will be replaced with the equivalent values set by `server.py`. For example, if `server.py` sets `data["params"]["starting_value"]` to `17`, then if the file `main.py` inside `workspaceTemplates` has the following content:
```
nwalters512, line 163: Recommendation: support an optional `.mustache` extension on files, and strip off that suffix if it exists. This will help folks like me who get endlessly annoyed when my editor complains about invalid syntax in templated files. In other words, the below file could be `main.py.mustache`, and it would appear to the workspace as `main.py`.

If someone really wants a templated mustache file that is itself a mustache file (that hurts my brain), they could us `main.mustache.mustache`.

lib/workspace.js
```diff
@@ -336,17 +337,53 @@ module.exports = {
...
-    const localPathExists = await fse.pathExists(localPath);
+    const staticFiles = (
+      await async
+        .mapSeries(klaw(localPath), async (file) => {
+          return file.stats.isFile()
+            ? { name: path.relative(localPath, file.path), localPath: file.path }
+            : null;
+        })
+        .catch(() => {
+          // Template path does not exist or is not accessible, do nothing
```
mwest1066, line 354:
```suggestion
          // Path does not exist or is not accessible, do nothing
```

docs/workspaces/index.md
```diff
@@ -156,11 +156,27 @@ A minimal `question.html` for an externally graded workspace should look somethi
...
+
+```py
```
nwalters512, line 165:
```suggestion
```txt
```

lib/workspace.js
```diff
@@ -336,17 +337,53 @@ module.exports = {
...
+            if (!file.stats.isFile()) return null;
+            return {
+              name: path.relative(templatePath, file.path),
+              buffer: mustache.render(
+                await fsPromises.readFile(file.path, { encoding: 'utf-8' }),
+                mustacheParams
+              ),
+            };
+          } catch (_err) {
+            // File cannot be rendered, treat file as static file
```
mwest1066, line 374: I'm thinking we should throw an error instead here? It feels better to break the question if something is not working correctly rather than to continue and hope for the best.

mwest1066 (approved):

lib/workspace.js
jonatanschroeder: I'm considering cases here of binary files that may be included in the templates by the instructor for some reason, but I'm happy to remove this if necessary. In any case, none of the errors so far cause an issue to be created, though there was discussion in the previous PR to do this in the future.

lib/workspace.js
```diff
@@ -336,17 +337,53 @@ module.exports = {
...
+    const mustacheParams = { params: variant.params, correct_answers: variant.true_answer };
+
+    const templateFiles = (
+      await async
+        .mapSeries(klaw(templatePath), async (file) => {
+          try {
+            if (!file.stats.isFile()) return null;
+            return {
+              name: path.relative(templatePath, file.path),
+              buffer: mustache.render(
```
nwalters512, line 368: Per the [Mustache docs](https://github.com/janl/mustache.js/#pre-parsing-and-caching-templates), Mustache maintains an internal cache of parsed templates. For long-running servers, this is not ideal, as this is essentially a memory leak. I think we should disable that by setting `mustache.templateCache = undefined` as described here: https://github.com/janl/mustache.js/pull/731/files. We can probably do this in `server.js` as part of the initialization code.

We probably should have been doing this the whole time we've used mustache - oops!

Related to mustache rendering, I've recently been hyper-aware of the ability of poorly-written or malicious user input to trigger denial-of-service attacks. In the future, we may consider moving potentially-expensive synchronous operations like this to worker threads, or to write a custom renderer based on `mustache.Writer` that can enforce a max size and abort rendering if we reach it. That's all for the future though.

docs/workspaces/index.md
```diff
@@ -193,7 +209,9 @@ def generate(data):
     ]
 ```
 
-By default, `contents` is expected to be a string in UTF-8 format. To provide binary content, the value must be encoded using base64 or hex, as shown in the example above. In this case, the `encoding` property must also be provided. Only one of `questionFile` and `contents` may be provided. If neither `questionFile` nor `contents` are provided, an empty file is created. If a file name is found in both the dynamic list above and in the `workspace/` directory, the dynamic content is used.
+By default, `contents` is expected to be a string in UTF-8 format. To provide binary content, the value must be encoded using base64 or hex, as shown in the example above. In this case, the `encoding` property must also be provided. Only one of `questionFile` and `contents` may be provided. If neither `questionFile` nor `contents` are provided, an empty file is created.
+
+If a file name is found in both the dynamic list from `_workspace_files` and in the `workspace/` or `workspaceTemplates/` directory, the dynamic content is used. For file names found in both `workspace/` and `workspaceTemplates/`, the template version is used.
```
nwalters512, line 214: Minor suggestion; I find an ordered, bulleted list easier to parse.

```suggestion
If a file appears in multiple locations, the following precedence takes effect (higher precedence first):

- dynamic content from `_workspace_files`
- `workspaceTemplate/` directory
- `workspace/` directory
```

lib/workspace.js
```diff
@@ -336,17 +337,53 @@ module.exports = {
...
+            ? { name: path.relative(localPath, file.path), localPath: file.path }
+            : null;
+        })
+        .catch(() => {
+          // Path does not exist or is not accessible, do nothing
+          return [];
+        })
+    ).filter((file) => file);
+
+    const mustacheParams = { params: variant.params, correct_answers: variant.true_answer };
```
nwalters512, line 359: Is there a specific use case you have in mind for making `correct_answers` available to the template?

news_items/042_workspace_dynamic/index.html
```diff
@@ -0,0 +1,72 @@
...
+     background: #f0f0f0;
+     padding: 1em;
+     padding-bottom: 0em;
+ }
+</style>
+
+<p>
+  In an effort to enhance the experience of PrairieLearn workspaces,
+  two new features have been introduced: randomized workspace files,
+  and wildcards for graded files.
```
nwalters512, line 25: Edit for concision:
```suggestion
  Two new features have been introduced for PrairieLearn workspaces:
  randomized workspace files, and wildcards for graded files.
```

news_items/042_workspace_dynamic/index.html
```diff
@@ -0,0 +1,72 @@
...
+    name and content of each file.</li>
+</ul>
+<p>
+  You can find more details on how to set up these files
+  in <a href="https://prairielearn.readthedocs.io/en/latest/workspaces/#creating-files-in-the-workspace-home-directory"
+  target="_blank">the PrairieLearn documentation</a>.
+</p>
+<p>
+  Additionally, you can now set up your question to allow a more
+  generic set of files, by using wildcard patterns to specify graded
```
nwalters512, line 46: Edit for concision:
```suggestion
  Additionally, you can now set up your question to use wildcard patterns to specify graded
```

lib/workspace.js
```diff
@@ -393,23 +432,21 @@ module.exports = {
...
-    }
-
-    if (dynamicFiles) {
-      await async.eachSeries(dynamicFiles, async ({ name, localPath, buffer }) => {
-        await fse.ensureDir(path.join(sourcePath, path.dirname(name)));
-        if (localPath) {
-          await fse.copy(localPath, path.join(sourcePath, name));
-        } else {
-          await fse.writeFile(path.join(sourcePath, name), buffer);
+    if (allWorkspaceFiles) {
```
nwalters512, line 435: I don't think this will do what you want it to; perhaps you want this?

```suggestion
    if (allWorkspaceFiles.length > 0) {
```

jonatanschroeder: Yeah, used to Python where an empty array is falsy...

lib/workspace.js
jonatanschroeder: Moved to #7098 .

lib/workspace.js
jonatanschroeder: I mostly just added for completeness, but it could be useful for cases like a workspace-side preliminary check. I just copied the processing from the rubrics code, which adds these two plus the submission (which doesn't apply here). I don't mind removing it.

lib/workspace.js
jonatanschroeder: > In any case, none of the errors so far cause an issue to be created, though there was discussion in the previous PR to do this in the future.

See #7095.

mwest1066: Excellent, I agree we should be creating issues in these cases.

mwest1066: I also think that it's fine for binary files to create errors -- they should be moved to `workspace/`. I really don't want to have silent failures during question creation.

lib/workspace.js
mwest1066: I do like consistency here.

lib/workspace.js
nwalters512: These errors actually occur during *workspace* creation, which happen outside of the question generation flow. As there's no request/response flow, there's also no way for us to surface meaningful errors to students or instructors. Ideally we'd both create an issue *and* show a meaningful error directly to the user (I know that's outside the scope of this PR though).

mwest1066: Good point! One option would be to create the issue and then set the workspace state to `stopped` with a message that reflects the error. This will be immediately shown to the student.

docs/workspaces/index.md
jonatanschroeder: Done in 9b3a54b9b

lib/workspace.js
jonatanschroeder: Moved these comments to #7095 so they're not lost.

lib/workspace.js
jonatanschroeder: I removed them in c8b8b6a but I can add them back. @mwest1066 what's your preference?

docs/workspaces/index.md
```diff
@@ -156,11 +156,27 @@ A minimal `question.html` for an externally graded workspace should look somethi
...
+Questions using workspaces can also be randomized, i.e., include files that contain random and dynamic content. This may be done in two ways: using mustache-based template files or using [the `server.py` file in the question directory](../question.md#question-serverpy). For template files, a workspace question can optionally include a `workspaceTemplates/` subdirectory within the regular question directory structure. The contents will be copied into the home directory of the student's workspace container, as with the `workspace/` directory. However, files within this directory may include mustache tags (e.g., `{{params.value}}`), which will be replaced with the equivalent values set by `server.py`. File names may optionally include the `.mustache` extension, and the file will be renamed before being presented to the student. For example, if `server.py` sets `data["params"]["starting_value"]` to `17`, then if the file `main.py,mustache` inside `workspaceTemplates` has the following content:
```
nwalters512, line 163:
```suggestion
Questions using workspaces can also be randomized, i.e., include files that contain random and dynamic content. This may be done in two ways: using mustache-based template files or using [the `server.py` file in the question directory](../question.md#question-serverpy). For template files, a workspace question can optionally include a `workspaceTemplates/` subdirectory within the regular question directory structure. The contents will be copied into the home directory of the student's workspace container, as with the `workspace/` directory. However, files within this directory may include mustache tags (e.g., `{{params.value}}`), which will be replaced with the equivalent values set by `server.py`. File names may optionally include the `.mustache` extension, and the file will be renamed before being presented to the student. For example, if `server.py` sets `data["params"]["starting_value"]` to `17`, then if the file `main.py.mustache` inside `workspaceTemplates` has the following content:
```

exampleCourse/questions/demo/workspace/dynamicFiles/workspaceTemplates/code.py.mustache
```diff
@@ -0,0 +1,2 @@
+value = {{params.random_value}}
+print(value)
```
nwalters512, line 2: Nit, trailing newline:
```suggestion
print(value)

```

PR #7092, opened 2023-02-02
Attempt to fix code caller memory leak

nwalters512:
This is a huge kludge to work around the issue described here: https://github.com/getsentry/sentry-javascript/issues/7031. Tl;dr: we need to ensure that code callers are created outside the ["domain"](https://nodejs.org/api/domain.html) of a request. Otherwise, that domain will indefinitely hold on to the Express `res` object, which in turn means that any large values in `res.locals` will be kept alive (prevented from being garbage-collected) for the lifespan of the worker.

I tested this in isolation on my reproduction repo: https://github.com/nwalters512/sentry-domain-memory-leak/pull/1. After refreshing 6 times, memory usage falls back to normal levels, so I believe this fix should work. However, it's difficult to test in PrairieLearn itself, as it's difficult to reproduce the failure condition. I think we'll just have to deploy it and ensure that it has the desired impact on memory usage.

### DIFF
--- a/lib/code-caller/index.js
+++ b/lib/code-caller/index.js
@@ -12,6 +12,7 @@ const load = require('../load');
 const { CodeCallerContainer, init: initCodeCallerDocker } = require('./code-caller-container');
 const { CodeCallerNative } = require('./code-caller-native');
 const { FunctionMissingError } = require('./code-caller-shared');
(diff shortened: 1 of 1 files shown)


mwest1066 (approved):
World's most horrible hack, but looks good!

PR #7094, opened 2023-02-02
Add devLogin option; refactor authn code

### LINKED ISSUE #1509: dev_user login mode

trombonekenny:
The development mode is currently inflexible with authentication. When you're running in development, it takes over as a default admin without many toggles to change that. It especially makes developing new authentication sources challenging because then you don't have the devMode options alongside something else. 

This is a more modern and modest rewrite which will replace PR #1724 and its branch. It doesn't try to change `config.authType = 'none'` but adds in a duplicated devMode login as a full login type available in devMode. That enables you to create a `pl_authn` cookie that gives devMode privileges, but also logout and login with a different mode. When I develop in that mode, I use the `authType: x-auth` config (or something that's not `none`).

(Changing `authType: none` is more complicated because that's the method the automated tests use and getting it to use cookie auth is non-trivial. This compromise leaves `authType: none` as-is but offers devMode logins along side other options when `config.devMode: true`)

Closes #1509


middlewares/authn.js
```diff
@@ -83,6 +83,7 @@ module.exports = function (req, res, next) {
   }
 
   // bypass auth for local /pl/ serving
+  // keeping for automated testing, see pages/authLoginDev for another way to dev signin
```
nwalters512, line 86: Nit:
```suggestion
  // Allow auth to be bypassed for local dev mode; also used by tests.
  // See `pages/authLoginDev` for another way to authenticate in dev mode.
```

pages/authLoginDev/authLoginDev.js
```diff
@@ -0,0 +1,55 @@
...
+var router = express.Router();
+
+router.get('/', (req, res, next) => {
+  if (!config.devMode) {
+    return next(new Error('devMode logins are not enabled'));
+  }
+
+  var authUid = config.authUid;
+  var authName = config.authName;
+  var authUin = config.authUin;
```
nwalters512, line 17: Presumably this could be enhanced in the future with a form where someone could type these in on the client?

pages/authLoginDev/authLoginDev.js
```diff
@@ -0,0 +1,55 @@
...
+    authName = 'Student User';
+    authUin = '000000001';
+  } else if (req.cookies.pl_test_user === 'test_instructor') {
+    authUid = 'instructor@illinois.edu';
+    authName = 'Instructor User';
+    authUin = '100000000';
+  }
+
+  let params = [authUid, authName, authUin, 'dev'];
+  sqldb.call('users_select_or_insert', params, (err, result) => {
```
nwalters512, line 33: Let's use `async`/`await` in new code.

pages/authLoginDev/authLoginDev.js
trombonekenny: Yeah, depending on how much of a front-end we want to put on authLoginDev, sure.

trombonekenny:
OK, so I took @nwalters512's "do async with the new code" comment to heart. And also looked at all the code duplication that exists in calls to `users_select_or_insert` or cookie handling across all the end points. I decided to bite the bullet and library-ize the commonality with more modern code.

It will also help if we decide to refactor the login cookie or sessions later, as all of that is captured in the new `lib/authn` file and called from all the various endpoints.

Locals tests pass with this, but we know we don't have automated testing for too many of the auth methods. We should run it in staging with the different authentication sources available to confirm the refactor worked out.

The only thing I didn't change was LTI 1.1's endpoint, since it does a different sproc for adding users and is deprecated code anyway, but if it's desired we can refactor that in too.

lib/authn.js
```diff
@@ -0,0 +1,86 @@
...
+// authn_provider_name: short name for the provider
+//
+// See optionsDefault:
+const optionsDefault = {
+  pl_authn_cookie: true,
+  redirect: false,
+  locals_authn: true,
+};
+
+module.exports.load_user_profile = async (
```
nwalters512, line 21: Let's add some JSDoc types to this function. You can add `// @ts-check` to the top of this file (and ideally the other files you modified) to get TypeScript to check things for you. Here's a starting point (this is using my other suggestion of a single `options` argument at the end):

```js
/**
 * @typedef {Object} LoadUserOptions
 * @property {'Google' | 'SAML'} provider
 */

/**
 * @param {import('express').Request} req
 * @param {import('express').Response} res
 * @param {LoadUserOptions} options
 */
module.exports.loadUser = async (
```

You'll need to add any other necessary properties (and additional `provider` types) to the `LoadUserOptions` type.

lib/authn.js
```diff
@@ -0,0 +1,86 @@
...
+  req,
+  res,
+  authnParams,
+  authn_provider_name,
+  optionsParams = {}
```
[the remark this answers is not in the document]
trombonekenny: I think it works better with two Objects at the end: `(req, res, AuthnParams, optionsParams)`  AuthnParams has the user-specific info, including the provider, and the `optionsParams` control how `loadUser` works with it: Does it make a cookie? Does it redirect? Does it populate `res.locals.authn*` etc.

The first object looks like what we pass into `users_select_or_insert` anyway, but now with a simple Object wrapper as an API, and the second option has defaults that can be overwritten for specific auth implementations.

trombonekenny:
I think this is ready to test on staging?

lib/authn.js
```diff
@@ -0,0 +1,90 @@
...
+const config = require('../lib/config');
+const csrf = require('../lib/csrf');
+const sqldb = require('@prairielearn/postgres');
+
+const sql = sqldb.loadSqlEquiv(__filename);
+
+/**
+ * @typedef {Object} LoadUserOptions
+ * @property {boolean} [LoadUserOptions.pl_authn_cookie=true] - Create the cookie?
+ * @property {boolean} [LoadUserOptions.redirect=false] - Redirect after processing?
```
nwalters512, line 11: No need for the `LoadUserOptions.` prefix in the property name, and type information for an object can't capture defaults, so the `=...` is unnecessary.

```suggestion
 * @property {boolean} [pl_authn_cookie] - Create the cookie?
 * @property {boolean} [redirect] - Redirect after processing?
```

lib/authn.js
```diff
@@ -0,0 +1,90 @@
...
+ * @property {boolean} [LoadUserOptions.locals_authn=true] - Populate res.locals.authn*?
```
nwalters512, line 12: AFAICT this property is never read; let's remove it.

```suggestion
```

lib/authn.js
```diff
@@ -0,0 +1,90 @@
...
+ */
+const optionsDefault = {
+  pl_authn_cookie: true,
+  redirect: false,
+  locals_authn: true,
```
nwalters512, line 17:
```suggestion
```

lib/authn.js
```diff
@@ -0,0 +1,90 @@
...
+
+const sql = sqldb.loadSqlEquiv(__filename);
+
+/**
+ * @typedef {Object} LoadUserOptions
+ * @property {boolean} [LoadUserOptions.pl_authn_cookie=true] - Create the cookie?
+ * @property {boolean} [LoadUserOptions.redirect=false] - Redirect after processing?
+ * @property {boolean} [LoadUserOptions.locals_authn=true] - Populate res.locals.authn*?
+ */
+const optionsDefault = {
```
nwalters512, line 14: Nit: this object is only used in one place; I'd be inclined to just inline these properties into the first line of `loadUser`.

server.js
```diff
@@ -451,6 +451,9 @@ module.exports.initExpress = function () {
 
   app.use('/pl/lti', require('./pages/authCallbackLti/authCallbackLti'));
   app.use('/pl/login', require('./pages/authLogin/authLogin'));
+  if (config.devMode) {
+    app.use('/pl/devLogin', require('./pages/authLoginDev/authLoginDev'));
```
nwalters512, line 455: Micro-nit: consider this for consistency with Azure?

```suggestion
    app.use('/pl/dev_login', require('./pages/authLoginDev/authLoginDev'));
```

lib/authn.js
```diff
@@ -0,0 +1,90 @@
...
+ * @property {string} [LoadUserAuth.name]
+ * @property {string} [LoadUserAuth.provider]
+ * @property {number} [LoadUserAuth.user_id] - If present, skip the users_select_or_insert call
+ */
+/**
+ *
+ * @param {import('express').Request} req
+ * @param {import('express').Response} res
+ * @param {LoadUserAuth} authnParams
+ * @param {LoadUserOptions} [optionsParams={}]
```
nwalters512, line 34:
```suggestion
 * @param {LoadUserOptions} [optionsParams]
```

lib/authn.js
```diff
@@ -0,0 +1,90 @@
...
+};
+
+/**
+ *
+ * @typedef {Object} LoadUserAuth
+ * @property {string} [LoadUserAuth.uid]
+ * @property {string} [LoadUserAuth.uin]
+ * @property {string} [LoadUserAuth.name]
+ * @property {string} [LoadUserAuth.provider]
+ * @property {number} [LoadUserAuth.user_id] - If present, skip the users_select_or_insert call
```
nwalters512, line 27:
```suggestion
 * @property {string} [uid]
 * @property {string} [uin]
 * @property {string} [name]
 * @property {string} [provider]
 * @property {number} [user_id] - If present, skip the users_select_or_insert call
```

nwalters512 (approved):
Won't block on any of these comments, but they should all be pretty easy to address.

lib/authn.js
trombonekenny: I prefer to give the library two objects keeping the purpose separate. The first has all of the user data. The second tells the library what actions to take with that data.

nwalters512: Sorry, I should have been more explicit.

```diff
- let options = { ...optionsDefault, ...optionsParams };
+ let options = { pl_authn_cookie: true, redirect: false, ...optionsParams };
```

trombonekenny:
@nwalters512 OK, took your feedback nits and did them. I made one more large-ish change before finishing this branch which was renaming the AuthLoginOauth2 to AuthLoginGoogle.

The two things I'm not sure about are patterns for deprecating `config.hasOauth` (do we warn on that somehow if it exists? I have the code accepting both `config.hasGoogleAuth` and `config.hasOauth` in most places but it's not bullet proof).

And if this is a considered a breaking change and how we mark that so that people who run their own servers know to update the config as well? Thoughts?

server.js
```diff
@@ -436,8 +436,9 @@ module.exports.initExpress = function () {
   app.use(require('./middlewares/cors'));
   app.use(require('./middlewares/date'));
   app.use(require('./middlewares/effectiveRequestChanged'));
-  app.use('/pl/oauth2login', require('./pages/authLoginOAuth2/authLoginOAuth2'));
-  app.use('/pl/oauth2callback', require('./pages/authCallbackOAuth2/authCallbackOAuth2'));
+  app.use('/pl/google_login', require('./pages/authLoginGoogle/authLoginGoogle'));
+  app.use('/pl/google_callback', require('./pages/authCallbackGoogle/authCallbackGoogle'));
```
nwalters512, line 440: This change will require a lot of effort and careful coordination when deploying, and AFAICT will result in zero tangible benefit to us. Unless there's a very good reason for making this change, let's revert this and leave it as is. Ditto for `hasGoogleAuth`.

trombonekenny: It was in the spirit of resolving tech debt while touching these old files. I don't think the impact is as grave as you're making it out to be, but we can also explore that in a different PR. No sense in making this one harder, I guess.

Plus, like I said in my comments, it uncovered a few things we don't really have in the project yet like deprecating config options or highlighting breaking changes.

nwalters512: Yep; we generally avoid those types of changes precisely *because* they're a hassle.

> It was in the spirit of resolving tech debt while touching these old files. I don't think the impact is as grave as you're making it out to be, but we can also explore that in a different PR.

To get very specific: this would require adding an additional redirect URI to each environment's OAuth config in Google's console and updating the redirect URI in our own configs. Certainly not insurmountable or particularly grave, but again, this is toil with no benefit for us. As the person deploying and operating this code, I don't like those kind of changes.

trombonekenny:
OK, I removed the "do not merge" tag since this was tested in staging. I'm good to see it merged and deploy. Thanks!

PR #7099, opened 2023-02-03
fix: Tests on instructorFiles rely on specific working directory

jonatanschroeder:
As discussed on Slack. These tests only worked if the CWD was /PrairieLearn.


tests/instructorFiles.test.js
```diff
@@ -17,9 +18,9 @@ describe('File paths', () => {
     });
 
     it('works with valid absolute over relative paths', async () => {
-      assert.ok(contains('/PrairieLearn', 'tests'));
-      assert.ok(contains('/PrairieLearn', 'exampleCourse/questions'));
-      assert.ok(contains('/PrairieLearn/exampleCourse', 'exampleCourse/questions'));
+      assert.ok(contains(process.cwd(), 'tests'));
+      assert.ok(contains(process.cwd(), 'exampleCourse/questions'));
+      assert.ok(contains(path.join(process.cwd(), 'exampleCourse'), 'exampleCourse/questions'));
```
nwalters512, line 23: Maybe I'm overthinking this, but if relative paths are interpreted relative to the current directory, that seems like undesirable behavior. For instance, say my working directory is `/foo/bar`. To check for something like `/bar/baz/biz`, I call `contains('/bar/baz', 'biz')`  which returns false.

Perhaps we want an implementation like this?

```ts
function contains(parentPath, childPath, includeSelf = true) {
  const absolutePath = path.resolve(parentPath, childPath);
  const relPath = path.relative(parentPath, absolutePath);
  if (!relPath) return includeSelf;
  return !(relPath.split(path.sep)[0] === '..' || path.isAbsolute(relPath));
}
```

This passes all tests except for the second test in "works with relative paths that are not contained":

```ts
assert.notOk(contains('PrairieLearn', 'tmp'));
```

That said, I wonder if that test is either wrong or meaningless. If `/foo/bar` can contain `baz`, why can't `bar` contain `baz`?

echuber2: In the case of that last test, is the intention that the items are './PrairieLearn' and './tmp', i.e. they are side by side?

jonatanschroeder: Maybe the issue is that `contains` is indeed not properly defined when the child path is relative and the parent path is absolute. I don't think this function is ever called with relative paths at all (I haven't checked thoroughly, though).

nwalters512: Yeah, perhaps `contains` only has meaning when the `parentPath` is absolute. I think `childPath` can be relative, but I think we should always resolve it relative to `parentPath`.

jonatanschroeder: I checked all references of `contains`, and all of them have absolute paths in the `parentPath` and all are expected to have an absolute path in `childPath` (kinda, see below). So it is safe to say that, whatever we decide the behaviour should be if either `parentPath` or `childPath` is relative, it wouldn't affect the behaviour of any (well-behaved) call to this function.

About the kinda: there is a bit of an exception here. The "upload" feature in the file browser obtains the file path (for re-uploads of existing files) or the base directory (for uploads of new files) from the query. The default values for the query parameters in the hidden elements of the page are absolute addresses. I don't think this is exposing anything sensitive, since the course settings already expose the absolute path of the course directory anyway, and it is not actually directly susceptible to naming attacks, since the `contains` check in this function does what it needs to do to ensure that this upload is only allowed if the path is within the root of the specific context (e.g., question directory, assessment directory, etc.). However, if someone where to send a custom POST request that modifies these parameters to a relative path, that would be the only situation that I found where this `contains` would be called with a potentially relative path in `childPath`. So if we were to define a specific behaviour for the function, this kind of inadvertent "attack" would have a specific behaviour instead of an unexpected behaviour based on whatever the CWD is.

With that in mind, I think your suggestion of resolving relative paths based on the parent path makes sense, and would give the unconventional call above a well-defined behaviour.

I will change the code with your suggestion and update the tests accordingly.

echuber2: Maybe a weird edge case:
```
> contains('/', '..')
true
```
If the function is used everywhere with an absolute path for both parent and child, would it do to just raise an exception if that isn't the case going forward?

jonatanschroeder: > Maybe a weird edge case:
> ```
> > contains('/', '..')
> true
> ```

The parent of the root directory in POSIX systems is always the root itself. This test makes sense. If the third argument were false is expect this to be false as well.

> If the function is used everywhere with an absolute path for both parent and child, would it do to just raise an exception if that isn't the case going forward?

Not sure if it's worth it.

lib/editors.js
```diff
@@ -1751,7 +1751,7 @@ class FileUploadEditor extends Editor {
       let err = new Error('Invalid file path');
       err.info =
         `<p>The file path</p>` +
-        `<div class="container"><pre class="bg-dark text-white rounded p-2">${this.oldPath}</pre></div>` +
+        `<div class="container"><pre class="bg-dark text-white rounded p-2">${this.filePath}</pre></div>` +
```
nwalters512, line 1754: 👀 was this a bug?

nwalters512 (approved):

lib/editors.js
jonatanschroeder: Yup. I was testing contains with an invalid path and this showed as undefined. I didn't do an extensive search for more bugs, only fixed the one I found. That said, this error is not expected to show during regular use, only if invalid paths are specified, so it's not a bug that regular users would experience.

Issue #7100, opened 2023-02-03
Fix `pl-dataframe` alignment on table header/dtype


coatless:
From a user in #random, we have:

<img width="461" alt="Screenshot 2023-02-02 at 5 37 35 PM" src="https://user-images.githubusercontent.com/833642/216490646-374d402f-e0c3-4586-81ae-501844cd0d18.png">

Issue #7026, opened 2023-02-01
pl-dataframe copy to clipboard button


eliotwrobson:
See title. The `pl-variable-output` has a "copy to clipboard" button, while the `pl-dataframe` element does not. Add this functionality to the dataframe element. It may be worth copying the HTML from the `pl-variable-output` element directly, since these elements use a similar tab structure and it might be best to make them look similar. cc @coatless

PR #7101, opened 2023-02-03
feat: pl-dataframe alignment and copy code

eliotwrobson:
Resolves https://github.com/PrairieLearn/PrairieLearn/issues/7100
Resolves https://github.com/PrairieLearn/PrairieLearn/issues/7026

Makes the display for this element more closely resemble others that use tabs and a copy button.

Table headers are now right aligned, see screenshot:
image

image

image

@coatless not sure if people have any other preferences for the table styling (i.e. adding column separators).


tdy:
IMO the `thead` names should be padded like the `tbody` cells. Otherwise there's still enough misalignment to be confusing IMO.

elements/pl-dataframe/pl-dataframe.css
```diff
@@ -17,6 +17,7 @@
     box-sizing: border-box;
     color: black;
     vertical-align: bottom;
+    text-align: right;
```
tdy, line 20: IMO the `thead th` names should be padded like the `td` cells below. Otherwise there's still enough misalignment to be confusing IMO.

eliotwrobson: Is there a way to share that padding in this CSS somehow? I'm no expert unfortunately.

echuber2: Maybe you could use CSS variables or just specify the padding for several things at once, like the `border: none` is written below.

elements/pl-dataframe/pl-dataframe.mustache
```diff
@@ -1,32 +1,43 @@
 {{#show_python}}
-<ul class="nav nav-tabs">
-  <li class="nav-item"><a class="nav-link active" data-toggle="tab" href="#table-{{uuid}}">Table</a></li>
-  <li class="nav-item"><a class="nav-link" data-toggle="tab" href="#code-{{uuid}}">Code</a></li>
-</ul>
+<script>
```
coatless, line 2: Great to see this functionality incorporated from `pl-variable-output`.

elements/pl-dataframe/pl-dataframe.py
```diff
@@ -23,6 +24,12 @@ class DisplayLanguage(Enum):
 SHOW_PYTHON_DEFAULT = True
 
 
+def get_python_string(varname: str, frame: pd.DataFrame) -> str:
+    # FIXME this doesn't work with certain data types (an issue with repr as well).
+    code_string = pprint.pformat(frame.to_dict(), width=500, indent=4)
```
coatless, line 29: I'm a little bit worried about this. If `pprint` isn't supporting what was added in `serialization 2`, that might be a bit problematic. Can you describe a little bit about the data type/edge cases that are problematic?

coatless (commented):
Thanks for tackling the improvements so quickly. Just a few quick comments.

elements/pl-dataframe/pl-dataframe.py
eliotwrobson: So the issue is that the `repr` values for types like `Timestamp` and `float('nan')` don't output valid Python. I'm not quite sure how to fix this without writing a custom serializer of some type.

elements/pl-code/pl-code.py
```diff
@@ -231,6 +232,9 @@ def render(element_html, data):
         "prevent_select": prevent_select,
     }
 
+    if pl.has_attrib(element, "id"):
+        html_params["id"] = pl.get_string_attrib(element, "id")
```
eliotwrobson, line 236: Looking at this again, would it be better to just build the copy text button as an option directly into pl-code? That way, we don't need this id passthrough, and it seems like a common thing to do with a code block anyway.

elements/pl-dataframe/pl-dataframe.css
eliotwrobson: Updated the CSS:

image

elements/pl-dataframe/pl-dataframe.mustache
eliotwrobson: Have now moved this functionality to `pl-code`. With this change, I think that `pl-variable-output` should be reworked to use `pl-code` (since this will add syntax highlighting and simplify the code on their end).

docs/elements.md
```diff
@@ -1166,6 +1166,7 @@ line callouts.
...
 | `highlight-lines-color` | text    | `#b3d7ff` | Specifies the color of highlighted lines of code.                                                                                                                                                                                                                                                                                                                                                                                                     |
+| `copy-code-button`      | boolean | false     | Whether to include a button to copy the code displayed by this element.                                                                                                                                                                                                                                                                                                                                                                               |
```
eliotwrobson, line 1169: Worth handling in a later PR, we should add this option to `pl-python-variable` and switch `pl-variable-output` to use this instead of the current copy code system.

elements/pl-dataframe/pl-dataframe.css
nwalters512: Is there a reason we're using right-aligned columns? It looks kinda weird for text. Is this because we don't have the ability to differentiate between numerical and text columns in the markup?

elements/pl-dataframe/pl-dataframe.py
nwalters512: Do you see a path forward to fixing this? If not, I'm inclined to drop the `FIXME` and just document that limitation.

elements/pl-dataframe/pl-dataframe.css
tdy: Now it looks properly aligned 🚀 The css could be refactored, but not really urgent.

nwalters512 (approved):

elements/pl-dataframe/pl-dataframe.css
tdy: Oops, nathan's comment hadn't loaded for me. I'll unresolve this chain.

elements/pl-dataframe/pl-dataframe.py
eliotwrobson: Yeah, I'm not really sure how to fix this, even using just repr has this same problem. It's not a deal breaker because it's fixable by the end user, so just adding a note in the docs is probably the way to go.

elements/pl-dataframe/pl-dataframe.css
eliotwrobson: Yeah, not sure if there's a way to distinguish between them using the styler. The right alignment was used because that was the styling before, but could definitely be changed if there's a better way.

tdy: That's how pandas chooses aligns everything (both in the REPL and HTML), so pandas users will be used to everything being right-aligned. I'm not sure about other libraries though. Personally I think the consistency looks better.

eliotwrobson: I think displaying things in a way people expect is probably the way to go then. If Pandas veterans expect it to be right aligned, we should probably stick with that.

elements/pl-dataframe/pl-dataframe.py
eliotwrobson: Added to the docs that some types may not be deserialized correctly.

coatless: Eh, I would say it's a deal breaker as those were two reasons for `v2`. I can't dig into this at the moment as I'm fighting other issues with workspaces and lecture slides. So, it'll be a couple of weeks at the earliest.

eliotwrobson: Oh I didn't mean we shouldn't still attempt to get this working for the edge cases, but this same issue is present in the current version of the code, and there's no immediate fix that should block this PR.

coatless: Ah, didn't realize that. Thought the dump worked correctly. Ewk! Alrighty, on with the merging on conquest.

eliotwrobson: Yep, the version in this PR works in more cases than the current code from the testing I did.

coatless: Should we just disable the `show-python` code tab if we detect a problematic type in the interim?

eliotwrobson: I'm not sure if there's a nice way to get a comprehensive list of everything that causes problems. I only found 2, but they can be fixed by setting certain variables.

eliotwrobson: I managed to fix some of the edge cases by adding some extra imports into the template! We will probably run into others but the current version worked for everything in the examples.

elements/pl-dataframe/pl-dataframe.mustache
```diff
@@ -26,7 +26,15 @@
       </div>
 
       <div id="code-{{uuid}}" role="tabpanel" class="tab-pane">
-        <pl-code language="python" copy-code-button="True">{{{code_string}}}</pl-code>
+<!-- Don't unindent this, will cause formatting issues -->
```
coatless, line 29: Yeah, I thought there was a reason for some of these elements were always left-aligned.

elements/pl-dataframe/pl-dataframe.py
```diff
@@ -163,7 +158,10 @@ def render(element_html: str, data: pl.QuestionData) -> str:
     html_params = {
         "uuid": pl.get_uuid(),
         "frame_html": frame_style.to_html(),
-        "code_string": get_python_string(display_variable_name, frame),
+        "code_string": pprint.pformat(
+            frame.to_dict(), width=500, indent=4, sort_dicts=False
```
coatless, line 162: Width seems a bit arbitrary. Did we want to set that as a possible option? 

The `indent` I think matches with the current linters?

eliotwrobson: It's easy enough to set that I can add it as an option. 

Yeah, 4 space indent matches what Black uses.

nwalters512 (approved):

PR #7104, opened 2023-02-03
Add demo question for RStudio workspace

echuber2:
As suggested by @coatless, a demo question for the RStudio workspace. This has a note about the persistence in particular, as requested, so I guess the idea is to just let students use it as a sandbox? Should there be an example of how to use it with a grader as well? I'm not sure which grader you'd want to use for that, but feel free to add a second example question if appropriate.

@nwalters512 Please let me know if the info in the question note isn't accurate. It seems like something we could move to the docs, if it isn't there already, but maybe it doesn't hurt to mention it in the demo too.


coatless:
Let's avoid adding a grader on this. The goal here is a persistent workspace similar to the xterm.js example: 
https://github.com/PrairieLearn/PrairieLearn/tree/master/testCourse/questions/workspace

If need be, we can also update it to use `jamesbalamuta/workspace-adv-rstudio` due to the lack of `tidyverse` packages & an out-of-date version of R/RStudio. (Plus, that image is building on `aarch64` (M1/M2) + `arm64` without `duckdb`)

echuber2:
OK, it has the note included about how the persistence works. Maybe it's better to base this one off of the "default" PL version of the workspace (trying to avoid external dependencies in the example course), but your advanced spin might be worth adopting as an official PL build too. You could also put your own fork of the example in your course as a sandbox with whatever variation of the workspace image you prefer.

coatless:
@echuber2 already doing that :) I just noticed there wasn't any demo question with the RStudio workspace. Hence, the gentle nudge to submit a PR on it.

Needed a stable version of RStudio w/ Quarto on it.

exampleCourse/questions/demo/workspace/rstudio/workspace/starter_file.md
```diff
@@ -0,0 +1,6 @@
+<!-- prettier-ignore -->
```
nwalters512, line 1: Why not let Prettier format this file?

exampleCourse/questions/demo/workspace/rstudio/question.html
```diff
@@ -0,0 +1,8 @@
+<pl-question-panel>
+<markdown>
+You can press the button below to launch an RStudio workspace that will persist for students, even if they refresh this page and click "Open workspace" again, because the question has been configured as "single variant" in `info.json`.
+
+**Note for instructors**: If you are an instructor previewing the question *in staff mode*, then refreshing this page and clicking the "Open workspace" button *will* take you to a fresh workspace container. Preview in student mode to test the persistence behavior. You can do this by going to the most recent instance of the example course, changing your user mode in the upper right to "Student view without access restrictions," and opening the "Questions using workspaces" assessment.
```
nwalters512, line 5: Other demo workspace questions don't have this information; I'd be inclined to omit it for consistency.

This information is also not 100% correct; if you click "save" or "save and grade", the variant ID is stored in the URL, and the same workspace instance will be used.

exampleCourse/questions/demo/workspace/rstudio/workspace/starter_file.md
echuber2: All right. At first it failed for me, before I edited it, and I thought maybe course staff would fork it and get thrown by that too.

coatless (approved):

PR #7106, opened 2023-02-03
Remove workspace sprocs

nwalters512:
This PR removes the `workspaces_message_update` and `workspaces_state_update` sprocs, both of which have been replaced by CTEs. This will allow us to avoid the creation of a random schema and subsequent creation of sprocs on every workspace host initialization, which puts a lot of load on the database when launching a lot of workspace hosts at the same time.

Fixes https://github.com/PrairieLearnInc/sysconf/issues/805.


pages/workspace/workspace.ejs
```diff
@@ -171,7 +171,13 @@
...
         if (state == 'running') {
           showWorkspaceFrame();
-          workspaceFrame.src = `${window.location.href}/container/`;
+          
+          // Avoid unnecessarily reassigning the src attribute, which causes the
+          // iframe to reload.
+          const workspaceFrameSrc = `${window.location.href}/container/`;
+          if (workspaceFrame.src != workspaceFrameSrc) {
+            workspaceFrame.src = workspaceFrameSrc;
+          }
```
nwalters512, line 180: I meant to include this in #7093, but I never committed it 🤦

mwest1066 (approved):

PR #7107, opened 2023-02-03
Exclude unhealthy hosts from workspace load computations

nwalters512:
See inline comment for description.

Fixes https://github.com/PrairieLearnInc/sysconf/issues/804.

### DIFF
--- a/sprocs/workspace_loads_current.sql
+++ b/sprocs/workspace_loads_current.sql
@@ -109,7 +109,7 @@ BEGIN
     -- Compute desired number of workspace hosts.
     --
     -- We only consider workspaces that are launching OR that are running on
-    -- an active host. In the case where a workspace is running on a host that
+    -- a healthy host. In the case where a workspace is running on a host that
     -- is unhealthy, that workspace will continue to run there until it times
     -- out and is shut down. We don't want to include those workspaces in the
     -- computation to avoid overprovisioning after deploys when all existing
--- a/sprocs/workspace_loads_current.sql
+++ b/sprocs/workspace_loads_current.sql
@@ -25,6 +25,7 @@ CREATE FUNCTION
         OUT workspace_running_count integer,
         OUT workspace_running_on_healthy_hosts_count integer,
         OUT workspace_active_count integer,
+        OUT workspace_active_on_healthy_hosts_count integer,
         -- max time in each workspace state
         OUT workspace_longest_launching_sec double precision,
         OUT workspace_longest_running_sec double precision,
@@ -49,6 +50,7 @@ BEGIN
         AND wh.state != 'unhealthy';
 
     workspace_active_count := workspace_running_count + workspace_launching_count;
+    workspace_active_on_healthy_hosts_count := workspace_running_on_healthy_hosts_count + workspace_launching_count;
 
     -- Longest running workspace in launching state.
     SELECT COALESCE(max(DATE_PART('epoch', now() - state_updated_at)), 0)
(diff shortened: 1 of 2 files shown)


sprocs/workspace_loads_current.sql
```diff
@@ -39,6 +40,13 @@ BEGIN
     SELECT count(*) INTO workspace_launching_count FROM workspaces AS w WHERE w.state = 'launching';
     SELECT count(*) INTO workspace_relaunching_count FROM workspaces AS w WHERE w.state = 'launching' AND num_nonnulls(w.rebooted_at, w.reset_at) > 0;
     SELECT count(*) INTO workspace_running_count FROM workspaces AS w WHERE w.state = 'running';
+    SELECT count(*) INTO workspace_running_on_healthy_hosts_count
+    FROM
+        workspaces AS w
+        JOIN workspace_hosts AS wh ON (wh.id = w.workspace_host_id)
+    WHERE
+        w.state = 'running'
+        AND wh.state != 'unhealthy';
```
nwalters512, line 49: I _think_ this is correct. The other option would have been to use `wh.state = 'ready'`, but I'm relatively confident that we want to include workspaces running on `draining` hosts too. There should never be any workspaces running on a `launching`, `terminating`, or `terminated` host, so those shouldn't matter either way.

sprocs/workspace_loads_current.sql
```diff
@@ -98,8 +106,15 @@ BEGIN
     FROM workspace_hosts AS wh
     WHERE wh.state = 'terminating';
 
-    -- Compute desired number of workspace hosts
-    workspace_jobs_capacity_desired := workspace_active_count * workspace_capacity_factor;
+    -- Compute desired number of workspace hosts.
+    --
+    -- We only consider workspaces that are launching OR that are running on
+    -- an active host. In the case where a workspace is running on a host that
```
nwalters512, line 112:
```suggestion
    -- a healthy host. In the case where a workspace is running on a host that
```

sprocs/workspace_loads_current.sql
mwest1066: I agree that we want to include workspaces on `draining` hosts, because we'll convert `draining` hosts back to `ready` hosts: https://github.com/PrairieLearn/PrairieLearn/blob/57208cb117c7ecf22bbaa910c2797069454321ca/ee/cron/workspaceHostLoads.js#L151-L152

I think it doesn't matter in practice whether we write `wh.state != 'unhealthy'` or `(wh.state = 'ready' OR wh.state = 'draining')`. I might slightly prefer the latter to make the intent clear, but I could go either way.

sprocs/workspace_loads_current.sql
```diff
@@ -98,8 +106,15 @@ BEGIN
...
-    workspace_jobs_capacity_desired := workspace_active_count * workspace_capacity_factor;
+    -- Compute desired number of workspace hosts.
+    --
+    -- We only consider workspaces that are launching OR that are running on
+    -- a healthy host. In the case where a workspace is running on a host that
+    -- is unhealthy, that workspace will continue to run there until it times
+    -- out and is shut down. We don't want to include those workspaces in the
+    -- computation to avoid overprovisioning after deploys when all existing
+    -- hosts are marked as unhealthy.
+    workspace_jobs_capacity_desired := (workspace_running_on_healthy_hosts_count + workspace_launching_count) * workspace_capacity_factor;
```
mwest1066, line 117: Do we want to compute
```
workspace_active_on_healthy_host_count := workspace_running_on_healthy_hosts_count + workspace_launching_count
```
and output this to CloudWatch? That would allow us to better visualize the metric that is actually being used for autoscaling.

mwest1066 (approved):

sprocs/workspace_loads_current.sql
nwalters512: Sure, that sounds reasonable!

sprocs/workspace_loads_current.sql
nwalters512: I'll make that change 👍

sprocs/workspace_loads_current.sql
nwalters512: Done in 64288c2.

sprocs/workspace_loads_current.sql
nwalters512: Changed in 08898ef.

mwest1066 (approved):

PR #7108, opened 2023-02-03
Disallow skipped or exclusive Mocha tests

nwalters512:
@trombonekenny pointed out on Slack that we had a skipped chunks test. This PR unskips it and adds a few new ESLing rules so that this doesn't happen again.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -7,7 +7,7 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 13,
   },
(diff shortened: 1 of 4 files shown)


trombonekenny:
So that's at the linting level... would we need something that failed the CI if a test was skipped or pending? Or the lint enough?

nwalters512:
CI already fails if a linter fails; there's not any additional plumbing we need there.

trombonekenny (approved):

PR #6375, opened 2022-09-28
On demand rendering of older submissions

jonatanschroeder:
Resolves #6365. Only the rendering of the submission block is lazy, the page is still rendered as usual. The code takes advantage of the fact that submissions are collapsed individually by default (so there is an easy trigger for retrieving the rendered submission), and that the external grading process already uses sockets for retrieving submission renders after grading is complete (so there is a path for retrieving this data).

PR #7111, opened 2023-02-03
Load only rendered submissions from database

nwalters512:
#6375 changed v3/Freeform questions to lazily render all old submissions when their panels are expanded. However, we were still actually loading all the data for *every* submission to a variant, which could be quite a lot. We believe this is partially responsible for some memory issues we've been seeing in production recently.

This PR changes that to only load the detailed information (the JSONB columns) for the submissions that will actually be rendered. This does mean that we'll need 2 database round-trips instead of just one, but the cost of that should be low.

### DIFF
--- a/lib/question.js
+++ b/lib/question.js
@@ -26,9 +26,10 @@ const error = require('../prairielib/lib/error');
 const sql = sqlLoader.loadSqlEquiv(__filename);
 
 /**
- * Question module.
- * @module question
+ * To improve performance, we'll only render at most three submissions on page
+ * load. If the user requests more, we'll render them on the fly.
  */
+const MAX_RECENT_SUBMISSIONS = 3;
 
 /**
  * Internal error type for tracking lack of submission.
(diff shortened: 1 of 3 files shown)


pages/partials/submission.ejs
```diff
@@ -43,7 +43,7 @@
     </button>
   </div>
 
-  <div class="collapse js-submission-body <% if (submission.submission_number == submissionCount || locals.expanded) { %>show<% } %> <% if (submissionHtml === undefined) { %>render-pending<% } %>" data-submission-id="<%= submission.id %>" id="submission-<%= submission.id %>-body">
+  <div class="collapse js-submission-body <% if (submission.submission_number == submissionCount || locals.expanded) { %>show<% } %> <% if (submissionHtml === undefined && question.type === 'Freeform') { %>render-pending<% } %>" data-submission-id="<%= submission.id %>" id="submission-<%= submission.id %>-body">
```
nwalters512, line 46: I happened to discover this while testing - we've been unable to display older submissions for v2 questions for a while now! I suppose it's a promising sign that no one noticed 😂

lib/question.js
```diff
@@ -1641,16 +1642,43 @@ module.exports = {
...
+
             locals.submissions = result.rows.map((s, idx) => ({
               grading_job_stats: module.exports._buildGradingJobStats(s.grading_job),
               submission_number: result.rowCount - idx,
               ...s,
+              // Both queries order results consistently, so we can just use
+              // the array index to match up the basic and detailed results.
+              ...(idx < detailedSubmissionResult.rowCount
+                ? detailedSubmissionResult.rows[idx]
+                : {}),
```
mwest1066, line 1681: This is some sneaky object spreading!

pages/partials/submission.ejs
mwest1066: Ha, nice! 😄

mwest1066 (approved):

PR #7112, opened 2023-02-03
Improve issue queries

nwalters512:
This PR makes two improvements to the issues queries used in `lib/question.js`:

- When only a count is needed, only load the count instead of all the data.
- If the `course_data` will not be displayed, don't load it.

These should both reduce load on the database and decrease memory consumption.

### DIFF
--- a/lib/question.js
+++ b/lib/question.js
@@ -1709,7 +1709,6 @@ module.exports = {
           // page render. Specifically, we'll only load the "course data" if
           // the user has the appropriate permissions. This should match the
           // check used in `pages/partials/question.ejs`.
-          console.log(locals.devMode || locals.authz_data.has_course_permission_view);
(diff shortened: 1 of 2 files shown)


lib/question.js
```diff
@@ -1192,20 +1192,17 @@ module.exports = {
...
-
-            if (result.rowCount > 0) {
-              success = false;
-              logger.verbose(`ERROR: ${result.rowCount} issues encountered during test.`);
-            } else {
-              logger.verbose('Success: no issues during test');
-            }
-            callback(null);
+        async () => {
+          const result = await sqldb.queryOneRowAsync(sql.select_issue_count_for_variant, {
```
mwest1066, line 1196: Ugh, I can't believe we were fetching them all just to get `rowCount`. I'm not going to `git blame` because I'm sure it's me 🤦

mwest1066 (approved):

PR #7113, opened 2023-02-04
Truncate issue course data

nwalters512:
This PR recursively truncates all strings in the "course data" object that are over 1000 characters. See inline comment in `lib/issues.js` for more details.


sprocs/index.js
```diff
@@ -117,7 +117,6 @@ module.exports.init = function (callback) {
       'files_insert.sql',
       'files_delete.sql',
       'issues_insert_for_variant.sql',
-      'issues_insert_for_assessment.sql',
```
nwalters512, line 120: This sproc was unused!

mwest1066: Good catch!

mwest1066 (approved):

PR #7118, opened 2023-02-06
style: Migrate `pl-variable-output` to use `pl-code`

### LINKED ISSUE #7116: Switch `pl-variable-output` to use `pl-code` for code output.
See title. Now that `pl-code` has the copy code button introduced in https://github.com/PrairieLearn/PrairieLearn/pull/7101, `pl-variable-output` can take advantage of this to reduce the amount of code on the backend for that element and add syntax highlighting.

eliotwrobson:
See title. Removes a lot of redundant code from `pl-variable-output` and cleans up the backend for that element. Resolves #7116. 

Also, tried to do something similar for pl-matrix-output, but it doesn't look like there's documentation or an example question for that element. @coatless do you happen to know if that element is deprecated? 

Screenshot:
image


coatless:
@eliotwrobson definitely not deprecated. I think the element was just overlooked due to `pl-variable-output` capturing the `pl-matrix-output` style.

coatless (approved):
Looks good to me! Simplifies the underlying components greatly.

eliotwrobson:
@coatless huh, that's very odd then. Matrix output doesn't appear in docs or any example questions. When I try this uplifting for that element, I'll see if the functionality can be folded into this one.

coatless:
@eliotwrobson not sure if I would spend time on `pl-matrix-output` as this seems to be an element from early `v3` days and likely isn't heavily used outside of core folks.

eliotwrobson:
@coatless makes sense, doesn't seem like anybody would use matrix output over this. Maybe worth removing the internals and using a passthrough like the way that we did for the `pl-python-variable` when we introduced the `pl-dataframe` element? @nwalters512 any thoughts?

nwalters512 (approved):

PR #7121, opened 2023-02-07
Add submission file endpoint

nwalters512:
Lifting this out of https://github.com/PrairieLearn/PrairieLearn/pull/6995 so that we can deploy it first. This should help avoid any issues where a request is routed to a server that doesn't have this API yet.


pages/submissionFile/submissionFile.js
```diff
@@ -0,0 +1,76 @@
...
+    if (fileRes.rowCount === 0) {
+      res.sendStatus(404);
+      return;
+    }
+
+    const contents = fileRes.rows[0].contents;
+    if (!contents) {
+      res.sendStatus(404);
+      return;
+    }
```
jonatanschroeder, line 57: Wouldn't this cause an empty file to cause an error? I know `pl-file-upload` does not allow for blank files, but other elements (or files submitted via workspaces) may be empty.

pages/submissionFile/submissionFile.js
```diff
@@ -0,0 +1,76 @@
...
+const express = require('express');
+const asyncHandler = require('express-async-handler');
+const { isBinaryFile } = require('isbinaryfile');
+const mime = require('mime');
+const sqldb = require('@prairielearn/postgres');
+
+const sql = sqldb.loadSqlEquiv(__filename);
+const router = express.Router({ mergeParams: true });
+
+const MEDIA_PREFIXES = ['image/', 'audio/', 'video/'];
```
jonatanschroeder, line 11: I think it may be a good idea to include `application/pdf`, even if `pl-file-preview` (and other elements) may not yet support it.

coatless:
Awesome to see work being done on this!

pages/submissionFile/submissionFile.js
nwalters512: Good call - I believe it would. Checking explicitly for `null` is probably better.

pages/submissionFile/submissionFile.js
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    const buffer = Buffer.from(contents, 'base64');
+
+    // res.sendStatus(404);
+    // return;
+
+    // Artificial delay for testing.
+    // await new Promise((resolve) => setTimeout(resolve, 3000));
+
```
nwalters512, line 66:
```suggestion
```

pages/submissionFile/submissionFile.js
jonatanschroeder: I just created two empty files in a worksheet, here's the result:

```
postgres=# select submitted_answer from submissions where variant_id = 286;
                                                                                                                                                                  submitted_answer

----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 {"_files": [{"name": "empty.c", "contents": ""}, {"name": "empty.h", "contents": ""}, {"name": "starter_code.c", "contents": "Ly8gc3RhcnRlcl9jb2RlLmMKCiNpbmNsdWRlIDxzdGRpby5oPgojaW5jbHVkZSAic3RhcnRlcl9jb2RlLmgiCgppbnQgbWFpbigpCnsKICAgIC8vIGVudGVyIHlvdXIgY29kZQp9Cg=="}, {"name": "starter_code.h", "contents": "Ly8gc3RhcnRlcl9jb2RlLmgK"}]}
(1 row)
```

nwalters512: Changed to check for `contents == null` in ca1ebda.

pages/submissionFile/submissionFile.js
nwalters512: Added in 703283f.

pages/submissionFile/submissionFile.js
```diff
@@ -0,0 +1,70 @@
...
+ * @param {string} name The file's name
+ * @param {Buffer} buffer The file's contents
+ * @returns {Promise<string>} The guessed mime type
+ */
+async function guessMimeType(name, buffer) {
+  const mimeType = mime.getType(name);
+  const isBinary = await isBinaryFile(buffer);
+
+  if (!isBinary) {
+    return 'text/plain';
```
mwest1066, line 25: Is this actually the right thing to do here? For example, a `.svg` file would detect as `text/plain` with this check, I think?

No need to hold up this this PR over this concern -- we can always address this later.

mwest1066 (approved):

pages/submissionFile/submissionFile.js
nwalters512: Good comment! I think I should actually swap the order here. That is, first check if it's a "media" mime (which would catch `image/svg+xml`), and only if it's not check if it's binary or not.

mwest1066: Yep, agreed.

nwalters512: Changed in c47f83b.

mwest1066 (approved):

PR #7127, opened 2023-02-07
Correctly capture course error data from error object

nwalters512:
This was a regression introduced accidentally in #7113.


jonatanschroeder (approved):
Tested locally, this fixed the issue.

Issue #2279, opened 2020-03-29
Throw an error on unrecognized pl-* elements


mwest1066:
In `traverseQuestionAndExecuteFunctions()` inside `freeform.js`, if `visitNode()` finds an element with `tagName` starting with `pl-` that we don't recognize, then it should throw an error rather than silently proceeding.

eliotwrobson:
I think that implementing this is actually somewhat problematic. The traversal function traverses the inner `pl` tags as well as the outer ones, but the platform doesn't have information about which inner tags should be recognized. It could retrieve these from a hardcoded list somewhere, but this seems like overcomplicating things for no _huge_ benefit.

echuber2:
I agree this may be a problem if course staff had intentionally been using `pl-` for course elements (or something similar). Although the server could conceivably include course elements in the list of "known" tag names, perhaps course staff are doing something unusual. Would it make sense to make this optional with a "strict" setting on the course? Or have a way to explicitly add tag names to ignore to the course?

echuber2:
On second thought, I haven't looked at the renderer in detail recently. If the expectation is that every `pl-` tag should be recognized and converted at that stage (so the other action would be to discard it silently), then it seems better to error than to discard it. If the current action is to leave it alone (to be converted in another pass), then erroring out seems overzealous.

nwalters512:
IIRC the traversal algorithm always processes parents before children, so even if you have a `pl-answer` inside a `pl-checkbox`, the renderer itself should never see that child element as it will have been turned into something else when `pl-checkbox` renders. This should mean that we don't have to know about any "inner" elements like that.

echuber2:
Maybe I'm not reading your comment correctly, but did you mean it processes children before parents?

nwalters512:
Nope, I meant parents first. Studying the implementation is definitely recommended!

eliotwrobson:
@nwalters512 is this the type of warning that would be better suited to being behind a feature flag? Looking at this again, I'm wondering if bundling this with the Python renderer is the way to go (as I did in #7983), since the new renderer is still opt-in only. Having a feature flag for just this warning feels like a very small change to put behind something like that, and enabling it for all courses at once is probably going to throw a ton of errors.

PR #7128, opened 2023-02-08
Experimental processing of question HTML entirely with Python

nwalters512:
This is an experimental change to the way that we process question HTML (`generate`, `render`, `parse`, `grader`, etc.).

Historically, we've done all HTML parsing + traversing in Node and only called into Python for individual elements. However, this is suboptimal for questions that have either large `data` objects, large question HTML, lots of elements, or some combination of these three. Every time we cross the process boundary, we have to serialize and parse that data (either JSON or HTML, sometimes both), which can result in high memory consumption (in the case of serializing/parsing lots of data) and contention for the main thread (in the case of parsing a large HTML document).

This PR introduces an HTML "processor" implemented entirely in Python. This means we'll only need to do a single IPC round trip for the generate/render/parse/grade/etc. for a question. This should improve tail latency quite a bit, as we only incur a single serialize/parse in both Node and Python.

A few good questions to test this with:

- https://us.prairielearn.com/pl/course_instance/130110/instructor/question/8510993/preview (~4MB of data in `params`)
- https://ca.prairielearn.com/pl/course_instance/2494/instructor/question/248192/file_view/questions/B12345Review/ClimateImpact/question.html (hundreds? of dropdowns)


python/phases.py
```diff
@@ -0,0 +1,107 @@
...
+SAVED_PATH = copy.copy(sys.path)
+
+
+# def resolve_element(element_name: str, context: dict) -> dict:
+#     if element_name in context["course_elements"]:
+#         return context["course_elements"][element_name]
+#     elif element_name in context["c_elements"]:
+
+
+def render(data: dict, context: dict) -> Tuple[str, Set[str]]:
```
eliotwrobson, line 23: Data and context should use TypedDicts for their annotations, which will make the comments below unnecessary.

python/phases.py
```diff
@@ -0,0 +1,107 @@
...
+    # This will track which elements have been rendered.
+    rendered_elements: Set[str] = set()
+
+    total_time = 0
+
+    def render_element(element: lxml.html.HtmlElement) -> Union[str, None]:
+        if element.tag not in elements:
+            return None
+
+        start = time.time()
```
eliotwrobson, line 42: Saw this somewhere else too, I'm assuming that this is to time the renderer to get performance data? Should use time.perf_counter to get more accurate results.

python/traverse.py
```diff
@@ -0,0 +1,63 @@
+from typing import Callable, Union
+
+import lxml.html
+
+
+def traverse_and_execute(
+    html: str, fn: Callable[[lxml.html.HtmlElement], None]
```
eliotwrobson, line 7: If we're doing more intricate things with lxml elements now, we might want to include type stubs of some type (either externally or our own). Right now everything from lxml is treated as  Any because the stubs are missing.

python/traverse.py
```diff
@@ -0,0 +1,63 @@
...
+    return root_parent.getchildren()
+
+
+def traverse_and_execute_impl(
+    element: lxml.html.HtmlElement, fn: Callable[[lxml.html.HtmlElement], None]
+) -> None:
+    new_element = fn(element)
+    has_new_element = (
+        new_element is not None
+        and new_element is not element
```
eliotwrobson, line 22: Why use the "is not" comparison here instead of "!="? I don't think we want to check whether these are equivalent objects in memory.

eliotwrobson (commented):
A few comments, looks really good so far!

python/traverse.py
nwalters512: That is actually exactly what I want to check; this supports returning the element that was passed to this function because I thought that might be something I need. I don't know if the current code relies on that though; I can reevaluate.

python/traverse.py
nwalters512: I ran headfirst into two issues with the stubs:

- https://github.com/lxml/lxml-stubs/issues/81
- https://github.com/lxml/lxml-stubs/issues/68

Given that, I'll opt to omit the stubs and forge ahead with `Any`s 😢

eliotwrobson: Yeah, I ran into similar problems when I tried both of the public stubs packages. We'll probably have to figure out how to add our own if we want to use them. We only use a small portion of the library so it shouldn't be a huge deal, but I ran into issues getting the directory right.

python/check_data.py
```diff
@@ -0,0 +1,174 @@
...
+
+    if phase not in edit_phases and old_value != new_value:
+        raise ValueError(f"data.{prop} has been illegally modified")
+
+
+def check_data(old_data: dict, new_data: dict, phase: Phase) -> None:
+    # First, check for extra keys on `new_data`.
+    extra_keys = set(new_data.keys()) - set(PROPS.keys())
+    if extra_keys:
+        extra_keys_str = ", ".join(key for key in extra_keys)
```
eliotwrobson, line 153: The join method can take in any iterable.
```suggestion
        extra_keys_str = ", ".join(extra_keys)
```

python/check_data.py
```diff
@@ -0,0 +1,174 @@
+from typing import Any, Dict, List, Literal, TypedDict, Union
+
+Phase = Union[
```
eliotwrobson, line 3:
```suggestion
Phase = Literal["generate", "prepare", "render", "parse", "grade", "test", "file"]
```

python/check_data.py
```diff
@@ -0,0 +1,174 @@
...
+        "edit_phases": [],
+    },
+}
+
+
+def check_prop(
+    prop: str,
+    old_value: Any,
+    new_value: Any,
+    value_type: str,
```
eliotwrobson, line 128: Should this be a literal as well?

nwalters512: It probably can be, this is a well-known set of values. I'll add that.

eliotwrobson:
Don't want this to get too bloated, but if we're rewriting the renderer already, would it be a good idea to incorporate this as well?: https://github.com/PrairieLearn/PrairieLearn/issues/2279

nwalters512:
@eliotwrobson I'd rather keep this PR limited to strictly re-implementing existing render functionality. As mentioned in that issue, it'd be pretty straightforward to implement this independently, and the way that the experimental renderer is implemented wouldn't get in the way of adding that check (that is, this new renderer would know all the element names, just like the old one).

python/question_phases.py
```diff
@@ -0,0 +1,189 @@
...
+
+            element_info = elements[element.tag]
+            element_type = element_info["type"]
+            element_name = element_info["name"]
+            if element_type == "core":
+                element_path = CORE_ELEMENTS_PATH / element_name
+            elif element_type == "course":
+                element_path = pathlib.Path(course_path) / "elements" / element_name
+            else:
+                raise Exception(f"Unknown element type: {element_type}")
```
eliotwrobson, line 94: If you make element_type a literal, then you can replace this with an assert_never.

nwalters512: Ahh, very cool! I made that change in 8ff06f2.

python/traverse.py
```diff
@@ -0,0 +1,227 @@
...
+    )
+    traverse_element = (
+        new_element if has_new_element and new_element is not None else element
+    )
+    for child in traverse_element:
+        if not isinstance(child, str):
+            traverse_and_execute_impl(child, fn)
+
+
+def traverse_and_replace(
```
eliotwrobson, line 94: Managed to write an iterative version of this that clocks in at about half the lines, I think it's conceptually simpler as well:
```python
def get_source_definition(element):
    #TODO reconstruct including attributes somehow
    return f"<{element.tag}>"


def traverse_and_replace(
    html: str,
    replace: Callable[
        [lxml.html.HtmlElement],
        ElementReplacement
    ]) -> str:
    import copy

    # Initialize result and work data structures
    result: List[Union[str, lxml.html.HtmlElement]] = []

    initial_list = lxml.html.fragments_fromstring(html)
    count_stack = [len(initial_list)]
    work_stack: List[Union[str, lxml.html.HtmlElement]] = copy.copy(initial_list[::-1])
    tail_stack = []

    while work_stack:
        element = work_stack.pop()

        # For just a string, append to final result
        if isinstance(element, str):
            result.append(element)

        else:
            new_elements = replace(element)

            # Turn new_elements into a list containing we can process
            if new_elements is None:
                new_elements = []
            elif isinstance(new_elements, str):
                fragments = lxml.html.fragments_fromstring(new_elements)
                new_elements = fragments if fragments is not None else []

            if isinstance(new_elements, list):
                # Modify count stack for new elements and decrement for element that was replaced
                count_stack[-1] += len(new_elements)-1

                # Add element tail before processing replaced element
                if element.tail is not None:
                    count_stack[-1] += 1
                    work_stack.append(element.tail)

                # Extend and go to the next iteration
                work_stack.extend(reversed(new_elements))
                continue

            # Add opening tag and text
            result.append(get_source_definition(new_elements))
            if new_elements.text is not None:
                result.append(new_elements.text)

            # Add all children to the work stack
            children = new_elements.getchildren()

            work_stack.extend(reversed(children))
            count_stack.append(len(children))
            tail_stack.append((new_elements.tag, element.tail))

        # Close all closable tags
        while count_stack[-1] == 0:
            count_stack.pop()
            tail_tag, tail_text = tail_stack.pop()

            result.append(f"</{tail_tag}>")
            if tail_text is not None:
                result.append(tail_text)
        else:
            count_stack[-1] -= 1

    # Empty out rest of tail tags.
    while tail_stack:
        tail_tag, tail_text = tail_stack.pop()

        result.append(f"</{tail_tag}>")
        if tail_text is not None:
            result.append(tail_text)

    return "".join(result)
```
@nwalters512 Passes all but one of the test cases (there was a convention for extra whitespace in the one that fails, wasn't sure how that should be so I chaged the test case locally). Haven't tested / benchmarked beyond the unit tests but seemed to be somewhat faster.

python/traverse.py
abelcheung: Sorry to chip in the discussion without any pretext. I have been maintaining [an enhanced lxml stub](https://github.com/abelcheung/types-lxml), does it solve the first issue?

For 2nd issue, actually `_Element.getchildren()` has been deprecated for a long time, that's why it doesn't appear in any stub. Current way is to use `list(_Element)` to get all children. Hope this helps!

python/traverse.py
nwalters512: Nice! I added a small benchmark:

```python
if __name__ == "__main__":
    html = """
    <pl-question-panel>
    <p> Consider two numbers $a = {{params.a}}$ and $b = {{params.b}}$.</p>
    <p> What is the sum $c = a + b$?</p>
    </pl-question-panel>

    <pl-integer-input answers-name="c" label="$c=$"></pl-integer-input>

    <pl-answer-panel>
      <pl-code-tabs filename="foo.py"></pl-code-tabs>
    </pl-answer-panel>
    """

    def replace(e: lxml.html.Element) -> str:
        if e.tag == "<pl-integer-input>":
            return "<pl-number-input></pl-number-input>"
        if e.tag == "pl-number-input":
            return '<input type="number" />'
        if e.tag == "pl-code-tabs":
            return "<div><pl-code></pl-code><pl-code></pl-code></div>"
        if e.tag == "pl-code":
            return "<pre><code>foo</code></pre>"

    original = timeit.timeit(lambda: traverse_and_replace(html, replace), number=1000)

    print("original", original)

    new = timeit.timeit(lambda: traverse_and_replace_new(html, replace), number=1000)

    print("new", new)
```

Your new version is pretty consistently faster:

```
original 0.0625948749948293
new 0.020077666005818173
```

eliotwrobson: @nwalters512 great to see! I actually had some trouble getting consistent benchmarks for this because I wrote it on a really slow laptop 😅

python/traverse.py
```diff
@@ -0,0 +1,179 @@
...
+        f'''{k}="{v.replace('"', "&quot;")}"''' for k, v in element.attrib.items()
+    )
+    return f"<{' '.join((element.tag, *attributes))}>"
+
+
+def traverse_and_replace(
+    html: str, replace: Callable[[lxml.html.HtmlElement], ElementReplacement]
+) -> str:
+    # Initialize result and work data structures
+    result: Deque[Union[str, lxml.html.HtmlElement]] = deque()
```
eliotwrobson, line 71: My mistake on this one:
```suggestion
    result: Deque[str] = deque()
```

python/traverse.py
```diff
@@ -0,0 +1,179 @@
...
+
+                # Extend and go to the next iteration
+                if new_elements:
+                    work_stack.extend(reversed(new_elements))
+
+                continue
+
+            if isinstance(new_elements, lxml.html.HtmlComment):
+                result.append(lxml.html.tostring(new_elements, encoding="unicode"))
+                continue
```
eliotwrobson, line 112: I don't think there should be a `continue` here. If comments are allowed to be items in `new_elements` when it is a list, then the algorithm still needs to modify `count_stack` after appending this to the result. The continue in the above case is only used because it modifies the count stack separately. I'll admit that I was kinda going for speed when I wrote this and didn't do that great of a job cleaning up the cases 😅

nwalters512: If we don't `continue` here, we try to do a bunch of things that don't make sense for comments (access its `.tag`, count its children, etc). With enough conditionals after this, we *could* forgo the `continue`, but IMO it's much easier to understand this if we just special-case comments and immediately stringify them. But maybe I'm misunderstanding the point about `count_stack` - can you write a test case that would fail with the current code?

eliotwrobson: I think immediately stringifying is the way to go, if we have a comment, but we just need to wrap that in an if-else case with the stuff that comes after this. The following test case fails (and this should fail on any test case where you have to close a tag after inserting a comment):
```python

def test_traverse_and_replace_comment() -> None:
    def replace(e) -> ElementReplacement:
        if e.tag == "p":
            return "<!-- Goodbye --><i>world</i>"
        return e

    html = traverse_and_replace("<i><p>Hello</p></i>", replace)
    assert html == "<i><!-- Goodbye --><i>world</i></i>"
```

nwalters512: I found another test case that didn't work and committed both it and your to the repo. I tried a few variations of `if`/`else` here, but it wasn't immediately obvious to me how to get the tests passing. If you understand the fix, feel free to add it as a suggestion!

nwalters512: > Ran into another error trying to trace through this test case, it seems like the fragments_fromstring function sometimes doesn't return the top-level list as I would have expected? Very strange.

Can you elaborate?

eliotwrobson: Sounds good, I'll play with it more a little later. There's another bug that comes out of some weird behavior from lxml, but I think the comment fix should be straightforward.

python/traverse.py
```diff
@@ -0,0 +1,164 @@
...
+                result.append(new_elements.text)
+
+            # Add all children to the work stack
+            children = list(new_elements)
+
+            if children:
+                work_stack.extend(reversed(children))
+
+            count_stack.append(len(children))
+            tail_stack.append((new_elements.tag, element.tail))
```
eliotwrobson, line 116:
```suggestion
            if isinstance(new_elements, lxml.html.HtmlComment):
                result.append(lxml.html.tostring(new_elements, encoding="unicode"))
            else:
                # Add opening tag and text
                result.append(get_source_definition(new_elements))
    
                if new_elements.text is not None:
                    result.append(new_elements.text)
    
                # Add all children to the work stack
                children = list(new_elements)
    
                if children:
                    work_stack.extend(reversed(children))
    
                count_stack.append(len(children))
                tail_stack.append((new_elements.tag, element.tail))
```
I think this is all that's needed to make it play nice with comments (tried locally). There's another issue going on with how the list of elements is returned by the `fragments_fromstring` function, but that should be handled with a different fix.

nwalters512: Nice, I adopted this change in 77e207b.

nwalters512:
Note to self: this is a particularly bad question that would likely perform much better under the new renderer: https://ca.prairielearn.com/pl/course_instance/2494/instructor/question/248192/file_view/questions/B12345Review/ClimateImpact/question.html (note the hundreds? of dropdowns in a single question)

python/traverse.py
```diff
@@ -0,0 +1,169 @@
...
+    html: str, fn: Callable[[lxml.html.HtmlElement], None]
+) -> None:
+    elements = lxml.html.fragments_fromstring(html)
+
+    # If there's leading text, the first element of the array will be a string.
+    # We can just discard that.
+    if isinstance(elements[0], str):
+        del elements[0]
+
+    for element in elements:
```
eliotwrobson, line 39: Have to `from itertools import chain` at the top.
```suggestion
    for e in chain.from_iterable(element.iter() for element in element_list)
```

apps/prairielearn/src/lib/question.js
```diff
@@ -2053,4 +2053,17 @@ module.exports = {
...
+   * which renderer was used for a given request.
+   *
+   * @param {import('express').Response} res
+   */
+  setRendererHeader(res) {
+    const renderer = res.locals.question_renderer;
+    if (renderer) {
+      res.set('X-PrairieLearn-Question-Renderer', renderer);
+    }
+  },
```
nwalters512, line 2068: I felt like this would be useful information to have to verify whether a given request is flowing through the new renderer or not? We should be able to tell this without question from the state of the feature flag, but still...

apps/prairielearn/src/question-servers/freeform.js
```diff
@@ -1260,6 +1348,15 @@ module.exports = {
...
       const context = await module.exports.getContext(question, course);
 
+      // Hack: we need to propagate this back up to the original caller so
+      // they can expose the selected renderer to the client via a header, but
+      // parent functions don't actually return things. So we'll just stick it
+      // in the `locals` object that the parent will be able to read from.
+      //
+      // See the `setRendererHeader` function in `lib/question` for where this
+      // is actually used.
+      locals.question_renderer = context.renderer;
```
nwalters512, line 1357: 🤮 but eventually we'll rewrite all this code to explicitly propagate things both down and up the call stack.

apps/prairielearn/src/lib/features/index.ts
```diff
@@ -1,5 +1,8 @@
 import { FeatureManager, FeatureGrantType } from './manager';
 
-const features = new FeatureManager([]);
+const features = new FeatureManager([
+  // Can only be applied to courses/institutions.
+  'process-questions-in-worker',
+]);
```
nwalters512, line 6: 🎉 first feature flag!

We'll see how well inline documentation of what this flag can be granted to holds up relative to enforcing this programmatically.

apps/prairielearn/python/traverse.py
```diff
@@ -0,0 +1,183 @@
...
+            return "<pre><code>foo</code></pre>"
+        return e
+
+    number = 1000
+    total_time = timeit.timeit(
+        lambda: traverse_and_replace(html, replace), number=number
+    )
+    time_per_iter = total_time / number
+
+    print(f"${number} loops, {time_per_iter * 1000}ms per loop")
```
nwalters512, line 183: TODO: can probably remove this?

eliotwrobson: I think this can be removed. More robust benchmarking code for this can be added back later if needed.

apps/prairielearn/python/check_data.py
```diff
@@ -0,0 +1,168 @@
+from typing import Any, Dict, List, Literal, TypedDict
```
eliotwrobson, line 1: Now that we're on Python 3.10, no need for uppercase `Dict`, `List`, or `Set` imports (lowercase versions work).

apps/prairielearn/python/check_data.py
```diff
@@ -0,0 +1,168 @@
...
+    "parse",
+    "grade",
+    "test",
+    "file",
+]
+
+
+class PropInfo(TypedDict):
+    type: ValueType
+    present_phases: List[Phase]
```
eliotwrobson, line 20: Nit: This list is meant to never change and contain unique elements, not likely to be an issue but nice to express this with the data structure choice.

```suggestion
    present_phases: FrozenSet[Phase]
```

apps/prairielearn/python/check_data.py
```diff
@@ -0,0 +1,168 @@
...
+    # Then, check for missing keys on `new_data`.
+    missing_keys = set(old_data.keys()) - set(new_data.keys())
+    if missing_keys:
+        missing_keys_str = ", ".join(missing_keys)
+        raise ValueError(f"data is missing keys: {missing_keys_str}")
+
+    # Validate each entry in `new_data`.
+    for key, new_value in new_data.items():
+        prop_info = PROPS[key]
+        old_value = old_data.get(key, None)
```
eliotwrobson, line 159: Micro nit: `None` is the default with no second argument.
```suggestion
        old_value = old_data.get(key)
```

apps/prairielearn/python/question_phases.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # This will track which elements have been processed.
+    processed_elements: Set[str] = set()
+
+    # If we're in the `render` phase, we'll eventually capture the HTML here.
+    # If we're in the `file` phase, we'll capture file data here.
+    # Otherwise, this will remain `None`.
+    result = None
+
+    total_time = 0
```
eliotwrobson, line 75: `total_time` and other timing variables aren't used anymore. They should be removed along with the `time` import.

apps/prairielearn/python/question_phases.py
```diff
@@ -0,0 +1,187 @@
...
+            os.chdir(element_path)
+
+            # Update the path to include the appropriate directories.
+            sys.path = copy.copy(SAVED_PATH)
+            sys.path.insert(0, str(PYTHON_PATH))
+            if element_info["type"] == "course":
+                sys.path.insert(0, str(pathlib.Path(course_path) / "serverFilesCourse"))
+            sys.path.insert(0, str(element_path))
+
+            mod = {}
```
eliotwrobson, line 109: Can this dict have a type annotation? Something like `dict[Phase, Callable[[str, dict], Optiona[str]]]`.

apps/prairielearn/python/traverse.py
```diff
@@ -0,0 +1,183 @@
...
+    html: str, fn: Callable[[lxml.html.HtmlElement], None]
+) -> None:
+    elements = lxml.html.fragments_fromstring(html)
+
+    # If there's leading text, the first element of the array will be a string.
+    # We can just discard that.
+    if isinstance(elements[0], str):
+        del elements[0]
+
+    for element in elements:
```
eliotwrobson, line 40: Put this in a previous comment, but this nested loop can be done in a single line with a `chain.from_iterable` call.

apps/prairielearn/python/traverse.py
```diff
@@ -0,0 +1,183 @@
...
+                # Handling processing instructions is necessary for elements like `<pl-graph>`
+                # that produce SVG documents.
+                #
+                # We transform these into comments to match the behavior of the HTML spec,
+                # as well as the `parse5` npm package that we use for parsing in JavaScript.
+                tail = new_elements.tail
+                new_elements.tail = None
+                instruction = lxml.html.tostring(new_elements, encoding="unicode")
+                instruction = instruction.removeprefix("<?").removesuffix("?>")
+                result.append("<!--?" + instruction + "?-->")
```
eliotwrobson, line 112:
```suggestion
                result.append(f"<!--?{instruction}?-->")
```

apps/prairielearn/python/traverse.py
eliotwrobson, line 1: I think I mentioned this in passing elsewhere, but for files containing important code, it would be nice to have `pytest` coverage reports included in the current CI job.

eliotwrobson (approved):
Some minor nitpicks, but looks great!

apps/prairielearn/python/traverse.py
```diff
@@ -0,0 +1,183 @@
...
+from typing import Callable, Deque, List, Optional, Tuple, Union
+
+import lxml.html
+
+ElementReplacement = Optional[
+    Union[str, lxml.html.HtmlElement, List[lxml.html.HtmlElement]]
+]
+
+# https://developer.mozilla.org/en-US/docs/Glossary/Void_element
+VOID_ELEMENTS = {
```
eliotwrobson, line 13:
```suggestion
VOID_ELEMENTS = frozenset({
```
Similar comment as somewhere else, makes sense to declare this immutable.

apps/prairielearn/python/traverse.py
nwalters512: That's a change for another PR; I'd accept that if you added it.

apps/prairielearn/python/traverse.py
nwalters512: Done in 8e121e4. It's a little unfortunate how Black formats this!

apps/prairielearn/python/traverse.py
nwalters512: Done in 8e121e4.

apps/prairielearn/python/traverse.py
nwalters512: Done in 8e121e4.

apps/prairielearn/python/question_phases.py
nwalters512: Done in 8e121e4.

apps/prairielearn/python/check_data.py
nwalters512: Done in 8e121e4.

apps/prairielearn/python/check_data.py
nwalters512: Done in 8e121e4.

apps/prairielearn/python/check_data.py
nwalters512: Done in 8e121e4 / 0ec9590.

python/traverse.py
nwalters512: Done in 8e121e4.

apps/prairielearn/python/traverse.py
nwalters512: Removed in c43d770.

apps/prairielearn/python/question_phases.py
```diff
@@ -0,0 +1,178 @@
...
+    if isinstance(filelike, str):
+        filelike = bytes(filelike, "utf-8")
+
+    # if this next call does not work, it will throw an error, because
+    # the thing returned by file() does not have the correct format
+    return base64.b64encode(filelike).decode()
+
+
+def process(
+    phase: Phase, data: dict, context: RenderContext
```
eliotwrobson, line 59:
```suggestion
    phase: Phase, data: dict[str, Any], context: RenderContext
```
Slipped my mind earlier that we basically always use strings as dict keys.

apps/prairielearn/python/check_data.py
```diff
@@ -0,0 +1,179 @@
...
+        raise ValueError(f'Expected data["{prop}"] to be a boolean')
+    elif value_type == "object" and not isinstance(new_value, dict):
+        raise ValueError(f'Expected data["{prop}"] to be an object')
+
+    # Check the value.
+    if phase not in edit_phases and old_value != new_value:
+        raise ValueError(f'data["{prop}"] has been illegally modified')
+
+
+def check_data(old_data: dict, new_data: dict, phase: Phase) -> None:
```
eliotwrobson, line 154:
```suggestion
def check_data(old_data: dict[str, Any], new_data: dict[str, Any], phase: Phase) -> None:
```
See other comment. I actually think this function will throw an unintended exception if keys are not all strings.

nwalters512: Can you elaborate on the exception-throwing thing?

FWIW, this change feels like lying to the typechecker - `data` is modifiable by users and there's absolutely nothing stopping them from using a non-string as a key.

eliotwrobson: > FWIW, this change feels like lying to the typechecker - `data` is modifiable by users and there's absolutely nothing stopping them from using a non-string as a key.

Ahhh that's a good point, I wasn't thinking about this. Especially since validating these dictionaries is the purpose of this function.

As for the exception throwing, the iterable used in the string `.join()` method needs to only consist of strings, or the method will throw a type error. So this will throw an exception:
```python
items = ['a', 'b', 1]
res_str = ''.join(items)
```
So if the dict we're checking has a non-string key, an exception will get thrown once we try to make the error message for the exceptions in this function.

nwalters512: What's the right way to handle that, something like this?

```python
''.join(str(val) for val in [1, '2', 'three'])
```

tools/snapshot-question-html.js
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Usage:
+ *
+ * ```sh
+ * node tools/snapshot-question-html.js ./old-renderer
+ * node tools/snapshot-question-html.js ./new-renderer
+ * git diff --no-index ./old-renderer ./new-renderer
+ * ```
+ */
```
nwalters512, line 42: Do folks think this would actually be useful in the future? Otherwise I can just remove this from the repo.

apps/prairielearn/python/check_data.py
eliotwrobson: Yeah, I would do `''.join(map(str, iterable))`.

nwalters512: Changed in b47aeab; I also added a test for good measure.

apps/prairielearn/python/question_phases.py
```diff
@@ -0,0 +1,178 @@
...
+            return element
+
+        try:
+            processed_elements.add(element.tag)
+
+            element_info = elements[element.tag]
+            element_type = element_info["type"]
+            element_name = element_info["name"]
+            if element_type == "core":
+                element_path = CORE_ELEMENTS_PATH / element_name
```
mwest1066, line 87: TIL about `/` and `pathlib`, which has apparently been there since Python 3.4 😲

apps/prairielearn/src/lib/question.js
mwest1066: Can't hurt!

tools/snapshot-question-html.js
mwest1066: Seems useful during dev, but hopefully we'll switch completely and delete these scripts?

apps/prairielearn/python/traverse.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def traverse_and_replace(
+    html: str, replace: Callable[[lxml.html.HtmlElement], ElementReplacement]
+) -> str:
+    # Initialize result and work data structures
+    result: Deque[str] = deque()
+
+    initial_list = lxml.html.fragments_fromstring(html)
+    count_stack: Deque[int] = deque([len(initial_list)])
```
mwest1066, line 65: Why do we need the `count_stack`? I'm not really following how this code works. Maybe a brief overview in a comment would help?

mwest1066 (approved):

mwest1066:
It looks like there are still a few unresolved comments from @eliotwrobson -- I didn't check whether they had actually been taken care of.

apps/prairielearn/python/traverse.py
eliotwrobson: Basically, the traversal algorithm needs this to track where to insert closing tags (since a single element can have multiple direct children). If you think of the HTML that's being parsed as a tree, this count stack just tracks where in that tree the algorithm is currently processing.

This algorithm is actually really similar to one solving the [balanced brackets](https://medium.com/@eric_lum/the-balanced-brackets-problem-ffc01acee8ee) problem, with the main difference being a slight change in how closing brackets (in this case closing tags) are handled.

nwalters512: cc @eliotwrobson want to add a description for your code here?

tools/snapshot-question-html.js
nwalters512: Yeah, this was mostly a sanity-check thing for myself during development, I don't think it'll be useful in the long term.

apps/prairielearn/python/question_phases.py
nwalters512: This again feels like lying to the type-checker; we don't actually know *what* the module will be.

python/traverse.py
nwalters512: This was ultimately integrated into this PR ✅

python/traverse.py
nwalters512: @eliotwrobson I don't actually remember what the resolution here was. Did this bug become a test case? Is it fixed?

apps/prairielearn/python/traverse.py
```diff
@@ -0,0 +1,153 @@
...
+    attributes = (
+        f'''{k}="{format_attrib_value(v)}"''' for k, v in element.attrib.items()
+    )
+    return f"<{' '.join((element.tag, *attributes))}>"
+
+
+def traverse_and_replace(
+    html: str, replace: Callable[[lxml.html.HtmlElement], ElementReplacement]
+) -> str:
+    # Initialize result and work data structures
```
eliotwrobson, line 61:
```suggestion
   """
   Perform traversal and element replacement on HTML with the given replace function.
   In short, uses stacks to track what has been parsed already and what still needs to be parsed.
   The count_stack tracks how many children each unclosed tag (contained in the tail_stack) has.
   The top entry in count_stack is decremented every time something is moved onto result,
   and when an entry hits zero, the corresponding tag from tail_stack is moved onto result as well.
   """
   
    # Initialize result and work data structures
```

apps/prairielearn/python/traverse.py
eliotwrobson: @nwalters512 a brief description suggested here https://github.com/PrairieLearn/PrairieLearn/pull/7128/files#r1222267135

python/traverse.py
charmoniumQ: I noticed this was linked from [lxml-stubs](https://github.com/lxml/lxml-stubs/issues/68) and recognized some folks from UIUC :wave:

That particular issue is about lack of stubs for `getchildren()`, but `getchildren()` is deprecated anyway (TIL).

After removing the `getchildren()`s, the rest of the `lxml-stubs` were mostly fine/usable with Mypy for my project, but also I'm not parsing HTML, so grains of salt.

PR #7134, opened 2023-02-08
Truncate and conditionally load issue system data

nwalters512:
Applies the same changes from #7112 and #7113 to the `system_data` of issues. This should further improve memory usage and performance in the average case when errors are encountered (that is, the user is a student).

### DIFF
--- a/lib/issues.js
+++ b/lib/issues.js
@@ -38,7 +38,9 @@ module.exports.insertIssue = async ({
   // if needed. The worst data is submission data, which can be very large;
   // this is stored on each individual submission.
   const truncatedCourseData = recursivelyTruncateStrings(courseData, 1000);
-  const truncatedSystemData = recursivelyTruncateStrings(systemData, 1000);
+  // Allow for a higher limit on the system data. This object contains output
+  // from the Python subprocess, which can be especially useful for debugging.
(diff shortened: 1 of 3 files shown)


lib/issues.js
```diff
@@ -31,21 +31,24 @@ module.exports.insertIssue = async ({
...
+  // Truncate all strings in the data objects to 1000 characters. This ensures
   // that we don't store too much unnecessary data. This data is here for
   // convenience, but it's not the source of truth: pretty much all of it
   // is also stored elsewhere in the database, so we can always retrieve it
   // if needed. The worst data is submission data, which can be very large;
   // this is stored on each individual submission.
   const truncatedCourseData = recursivelyTruncateStrings(courseData, 1000);
+  // Allow for a higher limit on the system data. This object contains output
+  // from the Python subprocess, which can be especially useful for debugging.
+  const truncatedSystemData = recursivelyTruncateStrings(systemData, 10000);
```
nwalters512, line 43: I picked 10000 bytes somewhat at random; open to suggestions for other limits!

mwest1066: I don't want to read more than 10k of debug output 😅

mwest1066 (approved):

PR #7135, opened 2023-02-08
Allow lastCallData to be quickly garbage collected

nwalters512:
After spending some time with the Node inspector, I think this may have been contributing to our memory issues. We'd be holding this object in memory until `this.lastCallData` gets reassigned in the next `call()` call. If we get unlucky and have a lot of failed calls in a row *and* have a low request rate (read: we're not cycling through the Python code callers very quickly), we might have a lot of these objects hanging around in memory. By dropping the reference to it, we allow it to be garbage collected ASAP.


mwest1066 (approved):
Excellent discovery! Now I'm wondering whether there are more cases where we are holding references to large objects.

PR #7136, opened 2023-02-09
Remove architecture-dependent NumPy tests

nwalters512:
`np.float128` and `np.complex256` are unavailable on certain CPUs/platforms, including Windows and Apple Silicon. This PR removes tests that use those types so that our tests can run on any platform.


eliotwrobson (approved):
Lgtm!

PR #7138, opened 2023-02-09
Force Google auth to prompt user for account

jonatanschroeder:
Addresses part of https://github.com/PrairieLearn/PrairieTest-feedback/issues/20. The issue is that, when using Google Auth, often the user is already logged in to a single Google account (e.g., their Gmail), but they may need to login with an institution-specific account. Causing the account page to always be shown when logging in ensures the user is aware of which account is being used, even in the specific cases where the Gmail account (or whatever they're already logged in with) is the expected one.

Unfortunately I don't have the means to test this.

### DIFF
--- a/pages/authLoginOAuth2/authLoginOAuth2.js
+++ b/pages/authLoginOAuth2/authLoginOAuth2.js
@@ -20,6 +20,7 @@ router.get('/', function (req, res, next) {
     url = oauth2Client.generateAuthUrl({
       access_type: 'online',
       scope: scopes,
+      prompt: 'select_account',
       // FIXME: should add some state here to avoid CSRF
     });
   } catch (err) {


jonatanschroeder:
Additional context: currently, if you are logged in with a Google account and you need to switch to a different one, and you are only logged in to that single account on the device/browser, you need to go to a Google page and add a new login there. Without this step, and assuming consent has already been given, Google will not even prompt for which account to use. So it is pretty laborious to change accounts.

jonatanschroeder:
I tested locally by adding `&prompt=select_account` to the URL generated by OAuth, and it works exactly as expected: if a single Google account is currently logged in, without it the account is used without prompt, with it a prompt is shown asking if the account should be used or a different account.

image

The behaviour is the same as usual if no account is currently logged in (ask for a username), or if more than one account is logged in (show prompt above regardless).

nwalters512 (approved):

Issue #7137, opened 2023-02-09
Passthrough and possibly deprecate `pl-matrix-output` element


eliotwrobson:
See discussion here: https://github.com/PrairieLearn/PrairieLearn/pull/7118#issuecomment-1424670154

The `pl-matrix-output` element has overlapping (and possibly redundant) functionality with the `pl-variable-output` element. Change the internals of this element to use a passthrough. There also isn't any documentation for this element, so it may be worth listing it as deprecated.

PR #7140, opened 2023-02-09
style: pl-matrix-output passthrough.

eliotwrobson:
Resolves https://github.com/PrairieLearn/PrairieLearn/issues/7137. Still needs more extensive testing and documentation but seems to work as expected.


eliotwrobson:
Looking at this again, turns out the code for this element would have to be significantly more complex to make this passthrough work correctly. I think the best way to deprecate this element is probably to throw some type of error message to the instructor ever time it gets used, then delete the element code itself once it is no longer in use.

PR #7143, opened 2023-02-10
feat: add express-session support

trombonekenny:
Implements `express-session` and a PL-database specific `session-store`. However, nothing uses the session (yet) so this PR is more of a no-op that adds the functionality for future features, like LTI 1.3.

See the RFC for usage and notes.


database/tables/pl_sessions.pg
```diff
@@ -0,0 +1,7 @@
+columns
```
nwalters512, line 1: I'd be inclined to add an indexed `user_id` column to allow ourselves to easily implement things like "destroy all sessions for user X" or "show all sessions for user Y".

database/tables/pl_sessions.pg
```diff
@@ -0,0 +1,7 @@
...
+    session: jsonb
+    sid: text not null
+    updated_at: timestamp with time zone not null default CURRENT_TIMESTAMP
```
nwalters512, line 4: We should index this column to minimize the likelihood of doing a table scan for operations like `all`, `length`, or the expiration cronjob.

tests/session-store.test.js
```diff
@@ -0,0 +1,153 @@
...
+  describe('basic functions', () => {
+    it('set() 5 sessions', async () => {
+      await promisify(store.set)('sid1', { value: 1 });
+      await promisify(store.set)('sid2', { value: 2 });
+      await promisify(store.set)('sid3', { value: 3 });
+      await promisify(store.set)('sid4', { value: 4 });
+      await promisify(store.set)('sid5', { value: 5 });
+    });
+
+    it('all() returns list of 5', async () => {
```
nwalters512, line 26: You should either rewrite these tests to not share state between them or use [`mocha-steps`](https://github.com/rprieto/mocha-steps) to ensure that we abort on the first error. You can find some examples of `mocha-steps` usage in other PL tests.

lib/session-store.js
```diff
@@ -0,0 +1,74 @@
+/*
+// https://github.com/expressjs/session#session-store-implementation
+//
+// Uses PL database as a express-session store
+*/
```
nwalters512, line 5: We don't need to do this immediately, but we should eventually move this to a package like `@prairielearn/session` or `@prairielearn/session-store` so that we can share this with PrairieTest.

lib/session-store.js
```diff
@@ -0,0 +1,74 @@
...
+const session = require('express-session');
+const _ = require('lodash');
+const util = require('util');
+
+const config = require('./config');
+const sqldb = require('@prairielearn/postgres');
+const sql = sqldb.loadSqlEquiv(__filename);
+
+class plSessionStore extends session.Store {
```
nwalters512, line 14:
```suggestion
class SessionStore extends session.Store {
```

Alternatively, `PLSessionStore` (class names should be capitalized), but I don't think we really need the "PL" branding on things.

lib/session-store.js
```diff
@@ -0,0 +1,74 @@
...
+      sid,
+      session: JSON.stringify(session),
+    };
+    await sqldb.queryOneRowAsync(sql.upsert, params);
+  });
+
+  get = util.callbackify(async (sid) => {
+    let params = {
+      sid,
+      expirationInSeconds: config.sessionStoreExpireSeconds,
```
nwalters512, line 29: Let's pass the expiration in the constructor so it's very straightforward to eventually lift this into a module that doesn't directly depend on `./config`.

lib/session-store.js
```diff
@@ -0,0 +1,74 @@
...
+  clear = util.callbackify(async () => {
+    await sqldb.queryAsync(sql.clear, {});
+  });
+
+  all = util.callbackify(async () => {
+    let result = await sqldb.queryAsync(sql.allsessions, {
+      expirationInSeconds: config.sessionStoreExpireSeconds,
+    });
+    let sessions = _.map(result.rows, 'session');
+    return sessions;
```
nwalters512, line 70:
```suggestion
    return result.rows.map(r => r.session);
```

lib/session-store.js
```diff
@@ -0,0 +1,74 @@
+/*
```
nwalters512, line 1: Let's get TypeScript checking us!

```suggestion
// @ts-check
/*
```

cron/sessionStoreExpire.sql
```diff
@@ -0,0 +1,4 @@
+-- BLOCK expire
+DELETE FROM pl_sessions
+WHERE EXTRACT(EPOCH FROM (now() - updated_at)) >= $expirationInSeconds
+;
```
nwalters512, line 4: Nit: we conventionally don't put SQL semicolons on a separate line.

database/tables/pl_sessions.pg
trombonekenny: We don't get user data from the session API. But, if we know the structure of how information is stored in the JSONB blob in the database, we can good queries off of that. We can add more columns if we need it, but I'm not sure would.

lib/session-store.js
trombonekenny: I was thinking about that, it's one of the reasons the database name is PL-specific. I think the code very much could be shared as a package/library.

database/tables/pl_sessions.pg
nwalters512: We can most likely set `req.session.user_id` in our login code and teach our store to pull that out and write it to a separate column. I agree that it's likely that performance will be good enough (and that features that require a full table scan are likely to be used infrequently), but I'd still be inclined to add this from the start, or at least in a quick followup PR.

cron/sessionStoreExpire.sql
trombonekenny: On a multiline query, I prefer it alone/left-justified. It's easier to skim and know that it's there. If it's on the end of a long last line, like in the query above, it's not always as easy to verify that it's missing (until you get an SQL error).

nwalters512: I can understand that, but I believe consistency is more important than personal preference. Eventually we'll have tooling enforcing consistent SQL formatting for us; for now, I unfortunately have to do it manually.

database/tables/pl_sessions.pg
trombonekenny: Sure. I didn't realize we had to index `tstz` tables; I thought postgres did that for us?

nwalters512: Nope; indexes have to be explicitly created if they're desired.

database/tables/pl_sessions.pg
trombonekenny: I didn't want to make columns we didn't use, so maybe we wait and make that a follow up PR if a JSONB query isn't good enough and we have a structured way of pulling the `user_id` out of the session data before the upsert?

trombonekenny: What if we add an [index](https://www.postgresql.org/docs/current/datatype-json.html#JSON-INDEXING)?

> GIN indexes can be used to efficiently search for keys or key/value pairs occurring within a large number of jsonb documents (datums).

cron/sessionStoreExpire.sql
jonatanschroeder: #7151, so you don't need to do it manually...

database/tables/pl_sessions.pg
nwalters512: Hmm, I don't think we've used GIN indexes before. I'm happy to try this, but I'd also be fine to forgo an index for now until we have a use case.

trombonekenny:
I'm going to need some TS help fixing this one:

```
lib/session-store.js:52:19 - error TS2554: Expected 1 arguments, but got 3.

52     this.set(sid, session, callback);
                     ~~~~~~~~~~~~~~~~~


Found 1 error in lib/session-store.js:52
```

What incantation will forgive my transgresses here, @nwalters512 ?

trombonekenny:
I think this is good to merge. Two things we might want to consider (or fix later):

- This will make a second cookie for all visitors of the default `connect.sid` naming. It's not used yet but will be there. (Worth it help the other dev work that depends on this, I think.)
- We might want to think about the default settings for the cookie. The `sameSite: none` setting is needed for the iframe bits under Canvas LTI but we can add that line in for the branches that test iframes/LTI. Since we aren't using the cookie for anything serious yet, we can leave it at `none` and revisit our security choices when we change the usage in the future.

rfcs/005-session-usage.md
nwalters512, line 1: If this is meant to be long-term developer documentation, let's put this somewhere else. Otherwise, let's remove this file entirely.

apps/prairielearn/src/lib/session-store.js
```diff
@@ -0,0 +1,92 @@
+// @ts-check
```
nwalters512, line 1: This file could be converted to TypeScript! Since it's already being type-checked, that conversion should be straightforward. If you'd rather not do it yourself, let me know and I'll do it in a followup PR, though I'd recommend getting acquainted with TS because we'll mandate that all new code be TS in the near future.

apps/prairielearn/src/lib/config.js
```diff
@@ -50,6 +50,7 @@ const ConfigSchema = z.object({
   /** Overrides the user UIN in development with `authType: 'none'` */
   authUin: z.string().nullable().default('000000000'),
   authnCookieMaxAgeMilliseconds: z.number().default(30 * 24 * 60 * 60 * 1000),
+  sessionStoreExpireSeconds: z.number().default(84600),
```
nwalters512, line 53: Nit, you probably meant this (one day):

```suggestion
  sessionStoreExpireSeconds: z.number().default(86400),
```

nwalters512 (commented):
🙌 excited for this to land soon! A few more comments and then this should be good to go.

rfcs/005-session-usage.md
trombonekenny: Where do you suggest? Wouldn't an RFC be the right place to set standards like this?

apps/prairielearn/src/lib/session-store.js
trombonekenny: I'd rather get this merged and work on rewriting later. (Or you can hop on the branch and do it.) The long running PRs are problematic to leave open when the code base keeps changing and I'm tired of merging without landing.

I'm still waiting for guidance to the PL dev community about how to "get acquainted with TS". You've done a ton of work, but relatively little bringing us along with you.

rfcs/005-session-usage.md
nwalters512: The way I've used RFCs in the past (both in PL and elsewhere), they're used to capture discussions, designs, constraints, etc. for particular features or functionality. They don't necessarily serve the function of long-term documentation/specification, but they're great for working asynchronously on a design and, in the future, answering questions like "what constraints did we consider when planning the implementation of XYZ".

This is different from how e.g the IETF uses RFCs.

apps/prairielearn/src/lib/session-store.js
nwalters512: That's fair, we can land as-is. What kind of guidance are you looking for - links to the TS docs, examples of conversion PRs, a syntax 101, something else?

rfcs/005-session-usage.md
trombonekenny: I'm not sure we have a convention yet on sessions. I know we have conventions for things like `res.locals` but I'm not sure those are documented either so there's no precedent there. 

I know you liked sessions for flash results and the LTI stuff isn't hard baked yet. I'd be inclined to leave it as an RFC (even as your interpretation of them vs IETF's) since things are fluid and emerging and we can permanently home them later.

apps/prairielearn/src/lib/session-store.js
trombonekenny: I'm not sure, I just feel very ill prepared to do any dev in PrairieLearn these days and I'm not sure the path to fixing that.

nwalters512: I know there's been a fair amount of rapid change and that can require some adjustment, but it will pay off in the long run - just ask @mwest1066 about his experience in PT before/after similar changes there! I'm happy to do whatever I can to get you and others better prepared, just let me know how I can help. Feel free to take this conversation to #pl-dev, this PR is probably not the best place to have it.

nwalters512 (approved):

PR #7144, opened 2023-02-10
Fix slow question statistics calculation

### LINKED ISSUE #4915: Slow assessment_questions_calculate_stats()
Running `assessment_questions_calculate_stats(125174104)` takes about 5 seconds. It can be reduced to 0.1 seconds by:

1. Save 2 seconds by removing the `users_is_instructor_in_course_instance()` call: https://github.com/PrairieLearn/PrairieLearn/blob/5b997eabb69a0b5707f6b0e2ab2d615069455756/sprocs/assessment_questions_calculate_stats.sql#L18
2. Save 1 second by removing the `LEFT JOIN groups LEFT JOIN group_users JOIN enrollments`: https://github.com/PrairieLearn/PrairieLearn/blob/5b997eabb69a0b5707f6b0e2ab2d615069455756/sprocs/assessment_questions_calculate_stats.sql#L13
3. Save 2 seconds by changing `(ai.user_id, ai.group_id) AS u_gr_id` to `coalesce(ai.user_id, ai.group_id) AS u_gr_id` in all three locations: https://github.com/PrairieLearn/PrairieLearn/blob/5b997eabb69a0b5707f6b0e2ab2d615069455756/sprocs/assessment_questions_calculate_stats.sql#L23

We can't apply the above changes immediately because they will change behavior, but we should try and restructure things so that we can do this.

mwest1066:
See #4915 for a description of the problem and https://github.com/PrairieLearn/PrairieLearn/issues/4915#issuecomment-990060717 for various options to fix it. I went with option 1.iii (added the column `assessment_instances.include_in_statistics`) and 3.iii (the coalesce trick with negative group IDs). As pointed out in item 2 in https://github.com/PrairieLearn/PrairieLearn/issues/4915#issuecomment-990060717, adding the `include_in_statistics` automatically removed the need to JOIN with `group_users`.

* [x] display `include_in_statistics` on the `instructorsAssessmentInstance` page in the top card

Fixes #4915


migrations/20230210165254_assessment_instances__include_in_statistics__update.sql
```diff
@@ -0,0 +1,6 @@
+UPDATE assessment_instances AS ai
```
mwest1066, line 1: I split this out into a separate migration for more flexibility during deploy.

sprocs/assessment_instances_insert.sql
```diff
@@ -103,8 +112,7 @@ BEGIN
     VALUES (true, assessment_instance_id, date_limit, authn_user_id);
 
     -- ######################################################################
-    -- start a record of the last access time
-    -- After code review I will delete those two lines of comment
```
mwest1066, line 107: Left over from the original group-work PR.

sprocs/assessment_questions_calculate_stats.sql
```diff
@@ -10,17 +10,15 @@ relevant_assessment_instances AS (
         assessment_questions AS aq
         JOIN assessments AS a ON (a.id = aq.assessment_id)
         JOIN assessment_instances AS ai ON (ai.assessment_id = a.id)
-        LEFT JOIN groups AS g ON (g.id = ai.group_id AND g.deleted_at IS NULL)
-        LEFT JOIN group_users AS gu ON (gu.group_id = g.id)
-        JOIN enrollments AS e ON ((e.user_id = ai.user_id OR e.user_id = gu.user_id) AND e.course_instance_id = a.course_instance_id)
```
mwest1066, line 15: These were only being used to call `users_is_instructor_in_course_instance()` so we can simply drop them now that we are storing `include_in_statistics` in the assessment instance.

migrations/20230210165254_assessment_instances__include_in_statistics__update.sql
```diff
@@ -0,0 +1,6 @@
...
+SET include_in_statistics = NOT users_is_instructor_in_course_instance(ai.user_id, a.course_instance_id)
+FROM assessments AS a
+WHERE
+    ai.assessment_id = a.id
+    AND ai.date > now() - interval '4 months';
```
nwalters512, line 6: Checking my understanding: this line plus the fact that `include_in_statistics` defaults to `TRUE` means that we'll have slightly inaccurate stats for old course instances? I think that's a reasonable tradeoff.

nwalters512 (approved):
Part of your comment from https://github.com/PrairieLearn/PrairieLearn/issues/4915:

> We should probably display this on the instructorsAssessmentInstance page in the top card, just for visibility/debugging.

Did you still want to do that?

migrations/20230210165254_assessment_instances__include_in_statistics__update.sql
mwest1066: Yep, that's right. I intend to run the migration manually on older course instances, but I didn't want to put that into the migration. I might shorten the migration to 2 months, actually.

mwest1066:
> Part of your comment from #4915:
> 
> > We should probably display this on the instructorsAssessmentInstance page in the top card, just for visibility/debugging.
> 
> Did you still want to do that?

Good point, I'll add it.

migrations/20230210165254_assessment_instances__include_in_statistics__update.sql
mwest1066: Actually I timed it and it looks like doing all `assessment_instances` at once will only take about 15 seconds, so in 1b5299e2e3578171ad8218d7d5e4fbc3d8591f09 I removed the date WHERE clause.

pages/instructorAssessmentInstance/instructorAssessmentInstance.ejs
```diff
@@ -78,6 +78,26 @@
                 <% } %>
               </td>
             </tr>
+            <tr>
+              <th>Statistics</th>
+              <td colspan="2">
+                <% if (assessment_instance.include_in_statistics) { %>
+                    Included
+                    <a tabindex="0" class="btn btn-xs" role="button"
+                        data-toggle="popover" data-trigger="focus" data-container="body"
```
jonatanschroeder, line 87: I'm wondering if a tooltip is better for this kind of scenario, both in terms of simplicity and accessibility.

mwest1066: What's the accessibility benefit here? (not disagreeing, just curious)

jonatanschroeder: I don't have a lot of expertise here, but it would make sense that an element that has a tooltip role would more clearly be parsed by a screen reader than a popover without a clear indication of its intended purpose. Though I was trying to read about it, and I may be wrong: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role

mwest1066: Thanks for clarifying. I'll leave this as is for now because I'm also not super sure what the best option is here.

migrations/20230210165254_assessment_instances__include_in_statistics__update.sql
```diff
@@ -0,0 +1,44 @@
...
+        JOIN assessments AS a ON (a.id = ai.assessment_id)
+        JOIN course_instances AS ci ON (ci.id = a.course_instance_id)
+        LEFT JOIN groups AS g ON (g.id = ai.group_id AND g.deleted_at IS NULL)
+        LEFT JOIN group_users AS gu ON (gu.group_id = g.id)
+),
+instructor_assessment_instances AS (
+    -- For each of the previous assessment instances with users, select only the
+    -- ones with users who are instructors in the course, copying the logic from
+    -- sprocs/users_is_instructor_in_course_instance.sql. Then only keep
+    -- distinct assessment instances.
```
nwalters512, line 20: Checking my understanding: this means that if a group consists of 5 students and an instructor, that assessment will be marked as `include_in_statistics = FALSE`, correct?

nwalters512 (approved):

migrations/20230210165254_assessment_instances__include_in_statistics__update.sql
mwest1066: That's correct. This is not necessarily the same as what the new code will set for `include_in_statistics` because the new code using the instructor-status of whichever user creates the assessment instance to decide. I think in practice that this shouldn't make much difference either way.

PR #7147, opened 2023-02-10
Use google-auth-library instead of googleapis

nwalters512:
`googleapis` is a massive package, weighing in at ~120MB and taking ~250ms to load on my machine. `google-auth-library` is comparably much smaller: it's only 680kB and takes ~20ms to load on my machine.

This change should decrease the size of our Docker image, decrease application boot time, and make our production AMI builds complete more quickly.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -183,6 +183,9 @@ config.shibLinkColors = {
 };
 config.hasAzure = false;
 config.hasOauth = false;
+config.googleClientId = null;
+config.googleClientSecret = null;
+config.googleRedirectUrl = null;
 config.syncExamIdAccessRules = false;
 config.ptHost = 'http://localhost:4000';
 config.checkAccessRulesExamUuid = false;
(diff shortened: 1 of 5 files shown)


trombonekenny:
OK that's really weird because I was running into type checking problems with the `authCallbackOAuth2.js` file and was looking at `googleapis` vs something else that type checked better.... at the same time you were writing this PR.

trombonekenny:
> OK that's really weird because I was running into type checking problems with the `authCallbackOAuth2.js` file and was looking at `googleapis` vs something else that type checked better.... at the same time you were writing this PR.

I'll leave that file alone right now and wait for a merge. :)

mwest1066 (approved):
Awesome! Hopefully this will also cut down on the number of versions bumps we get from dependabot for `googleapis`.

trombonekenny:
How do you test these things? Throw them up on a staging server?

mwest1066:
> How do you test these things? Throw them up on a staging server?

Yep!

PR #7148, opened 2023-02-11
[WIP] Communicate with Python process over TCP

nwalters512:
Questions to test this with to see if things improve:

- https://us.prairielearn.com/pl/course_instance/138322/instructor/question/9164957/preview
- https://us.prairielearn.com/pl/course_instance/144858/instructor/question/9217339/preview


python/zygote.py
```diff
@@ -74,166 +75,177 @@ def try_dumps(obj, sort_keys=False, allow_nan=False):
         raise
 
 
-def worker_loop():
+def worker_loop(s: socket.socket):
```
eliotwrobson, line 78: A good opportunity to add type hints to this whole file. Don't forget to add zygote.py to the pyproject file.

python/zygote.py
```diff
@@ -74,166 +75,177 @@ def try_dumps(obj, sort_keys=False, allow_nan=False):
...
-                                + str(cwd)
-                                + "/"
-                                + str(file)
-                                + ".py.\n\n passed:\n  "
-                                + str(args[-1])
-                                + "\n\n returned:\n  "
-                                + str(val)
-                                + '\n\nThere is no need to be returning "data" at all (it is mutable, i.e., passed by reference). In future, this code will throw a fatal error. For now, the returned value of "data" was used and the passed value was discarded.'
-                            )
+            data += s.recv(256 * 1024)
```
eliotwrobson, line 90: So not sure what the purpose of this is, but I think data is a bytes object, which is immutable. The += and slice operators below will always make copies, which could get really slow. Better to use a bytearray.

python/zygote.py
```diff
@@ -74,166 +75,177 @@ def try_dumps(obj, sort_keys=False, allow_nan=False):
...
+                # if val is a string, treat it as utf-8
+                if isinstance(val, str):
+                    val = bytes(val, "utf-8")
+                # if this next call does not work, it will throw an error, because
+                # the thing returned by file() does not have the correct format
+                val = base64.b64encode(val).decode()
+
+            # Any function that is not 'file' or 'render' will modify 'data' and
+            # should not be returning anything (because 'data' is mutable).
+            if (fcn != "file") and (fcn != "render"):
```
eliotwrobson, line 208: Can do `fcn not in {"file", "render"}`

python/zygote.py
```diff
@@ -74,166 +75,177 @@ def try_dumps(obj, sort_keys=False, allow_nan=False):
...
+                    length = int.from_bytes(data[0:4], byteorder="big")
+                    data = data[4:]
+                if len(data) >= length:
+                    json_inp = data[0:length]
+                    data = data[length:]
+                    length = None
+                    return json.loads(json_inp)
+
+    def write_json_to_socket(data: str):
+        encoded_data = bytes(try_dumps(data, allow_nan=False), "utf-8")
```
eliotwrobson, line 102: Not sure if the Json encoding itself is a bottleneck, but if performance is a sticking point there are drop-in replacement Json libraries that are supposedly much faster. Just a thought.

eliotwrobson (commented):
Early comments. Looks cool so far!

python/zygote.py
nwalters512: `asyncio` only really helps if you're trying to do multiple things in parallel, right? All this code is intentionally very sequential, so I'm not sure what improvements we'd expect to see.

python/zygote.py
nwalters512: FYI this is all existing code. I can make a few changes, but I'm not looking to revamp the overall code here (minimizing things I could break and all).

python/zygote.py
nwalters512: I very much doubt it is a bottleneck; the objects in play here are generally quite small. May still be worth measuring though.

python/zygote.py
nwalters512: Would that just look like this?

```python
data = bytearray()
bytearray.extend(new_bytes)
```

python/zygote.py
eliotwrobson: Makes sense. I noticed that this was the same after taking a closer look, this just comes up in the diff because of the indentation.

python/zygote.py
eliotwrobson: Yep, more or less a drop-in replacement with just a couple of different methods.

nwalters512: Sweet - done in 1e196b5!

python/zygote.py
eliotwrobson: It's helpful in situations where you have blocking calls (like writing or reading data to a socket here). If you want to keep the changes to this file minimal, it's probably not worth pursuing, but the framework has builtin constructs for an event loop and signal handling that simplify some of the other code in this file https://docs.python.org/3/library/asyncio-stream.html#register-an-open-socket-to-wait-for-data-using-streams

nwalters512: But does it actually change anything if there's not other work that could be done while waiting for those operations to complete? i.e., if I was writing a Node server that only handled one request at a time, it wouldn't matter if I used sync vs async filesystem operations, as I already wouldn't be using the main thread for other things while waiting for those FS operations to complete. Does `async`/`await` somehow make async operations faster in Python? I wouldn't expect that to be the case.

python/zygote.py
```diff
@@ -74,166 +75,177 @@ def try_dumps(obj, sort_keys=False, allow_nan=False):
...
-                                + str(args[-1])
-                                + "\n\n returned:\n  "
-                                + str(val)
-                                + '\n\nThere is no need to be returning "data" at all (it is mutable, i.e., passed by reference). In future, this code will throw a fatal error. For now, the returned value of "data" was used and the passed value was discarded.'
-                            )
+            data.extend(s.recv(256 * 1024))
+            if len(data) >= 4:
+                if length is None:
+                    length = int.from_bytes(data[0:4], byteorder="big")
+                    data = data[4:]
```
eliotwrobson, line 94: Using `del` avoids making a copy. Should do the same in the if case below as well.
```suggestion
                    del data[:4]
```

python/zygote.py
eliotwrobson: Does PL expect requests sent to this file to be served in the order they were received? If that's the case then I think you're right that there's no real way to speed this up. `async`/`await` in Python basically just allows you to have your code do other work while waiting for blocking calls to finish (i.e. serialize JSON while waiting to read something from the file system). But if you want to force requests to be handled in-order there's not much that can be done.

nwalters512: PL only sends a new request once the previous one is done; there will never be multiple pending or in-flight IPCs at the time time. This is very much intentional: elements + question code _must_ be processed sequentially (since they mutate the shared `data` object), and we can't have a single Python worker doing work for multiple questions at once because we only a expose a single course directory to the worker at a time to prevent course A from accessing course B's files.

python/zygote.py
nwalters512: Changed in db4e174.

python/zygote.py
eliotwrobson: Gotcha, makes sense. Yeah then the main benefit of using async is not having to manually fork the process and cleaner handling of the event loop. Something to consider if this file gets reworked but to minimize changes then shouldn't be included here.

python/zygote.py
```diff
@@ -74,166 +75,177 @@ def try_dumps(obj, sort_keys=False, allow_nan=False):
...
-                                + str(args[-1])
-                                + "\n\n returned:\n  "
-                                + str(val)
-                                + '\n\nThere is no need to be returning "data" at all (it is mutable, i.e., passed by reference). In future, this code will throw a fatal error. For now, the returned value of "data" was used and the passed value was discarded.'
-                            )
+            data.extend(s.recv(256 * 1024))
+            if len(data) >= 4:
+                if length is None:
+                    length = int.from_bytes(data[0:4], byteorder="big")
+                    del data[4:]
```
eliotwrobson, line 94:
```suggestion
                    del data[:4]
```

python/zygote.py
```diff
@@ -74,166 +75,177 @@ def try_dumps(obj, sort_keys=False, allow_nan=False):
...
+                if len(data) >= length:
+                    json_inp = data[0:length]
+                    del data[length:]
```
eliotwrobson, line 97:
```suggestion
                    del data[:length]
```

eliotwrobson (commented):
Mistake in https://github.com/PrairieLearn/PrairieLearn/pull/7148/commits/db4e17404b6544cf60fda95e3bb917930449e1c2

nwalters512:
I resolved conflicts as best I could, but things are very much not functioning on this branch at the moment. I'll take a look soon to see if I can figure out what's going wrong.

python/zygote.py
nwalters512: Thanks for pointing this out! Addressed in fbfe401.

python/zygote.py
nwalters512: Addressed in fbfe401.

python/zygote.py
nwalters512: Manual forking is still necessary for isolation, right?

eliotwrobson: I think so, but I'm not even sure the zygote relies on this? The main benefit would be simplifying the `worker_loop` function, the external fork creation can still happen.

reteps:
I think this can be restored to it's previous state with
```
git reset --hard ec286700db97f1a31a2e2c91da8a90261b1ab3a0
git push -f
```

but I will let someone else do this so I don't mess up the history further.

nwalters512:
Done, specifically the following:

```sh
git reset --hard HEAD^1
git push -f
```

nwalters512:
@reteps I'm glad you're excited about landing this! I'll work on resolving conflicts tomorrow and take another look to see what has to happen before this is ready for review.

reteps:
Hopefully the merge conflicts from the other PR provide some value tomorrow. I am excited to land this! I had done some research on ZMQ, shared memory, message queues.

Some thoughts I had:
- ZMQ adds an abstraction layer that we don't need yet, and the JS library for it is barely used by others.
- Shared memory `/dev/shm/...` or a shared file seems too low level, while it would give us performance gains, I'm not sure we want to deal with the complexity at this moment. It also locks us further into the model where the question execution and question submission happen on the same machine (not that this is a bad model, but I could see this model changing in the future)
- Message queues: this makes sense in a different architecture to the one that is setup.

nwalters512:
I don't think this is going anywhere at the moment, nor is it necessary. We already had a big performance win from moving the entire rendering process into Python.

PR #7152, opened 2023-02-11
Format all SQL queries using prettier

jonatanschroeder:
Format all queries using the rules in #7151. Separate PR to allow this one to be ignored in git blame.


admin_queries/course_instance_compute_time.sql
```diff
@@ -1,83 +1,97 @@
...
     WHERE
-        ($institution_short_name = '' OR i.short_name = $institution_short_name)
-        AND gj.date BETWEEN $start_date AND $end_date
-        AND gj.grading_method = 'External'
+      (
+        $institution_short_name = ''
+        OR i.short_name = $institution_short_name
+      )
+      AND gj.date >= $start_date
+      AND gj.date <= $end_date
```
jonatanschroeder, line 58: For some reason I had to replace this BETWEEN with comparisons. I believe it has do to with a corner case in sql-formatter since `date` is a keyword, since other cases in this same query that use BETWEEN are still ok.

mwest1066 (approved):

admin_queries/course_instance_compute_time.sql
jonatanschroeder: Ok, I think I found the reason: https://github.com/sql-formatter-org/sql-formatter/issues/534

jonatanschroeder: TLDR to anyone reading this and not the Slack discussion: any constructs like `a BETWEEN b AND c AND d` need to be rewritten as `(a BETWEEN b AND c) AND d` to avoid ambiguity. It had nothing to do with the column name.

docs/running-in-production/admin-user.md
```diff
@@ -11,19 +11,29 @@ psql postgres
...
-SELECT * FROM users;
+SELECT
+  *
+FROM
+  users;
 ```
 
 Which will display a table of users in the database:
 
 ```sql
```
nwalters512, line 22: I'd revert the below change and make this code block `txt` instead?

lib/chunks.sql
```diff
@@ -1,119 +1,222 @@
...
+course_instance_id = EXCLUDED.course_instance_id,
+assessment_id = EXCLUDED.assessment_id,
+uuid = EXCLUDED.uuid;
 
 -- BLOCK delete_chunks
-WITH chunks_metadata AS (
+WITH
+  chunks_metadata AS (
+    SELECT
+      (cm ->> 'type')::enum_chunk_type AS
```
nwalters512, line 169: It'll take me some time to adjust to spaces on either side of JSON operators 😅

nwalters512 (approved):
As with JS, I don't always agree on the formatting decisions Prettier makes, but I'll have to trust that I'll get used to them! Consistency is king 🙌

docs/running-in-production/admin-user.md
jonatanschroeder: Done in b2cbf6a32.

lib/chunks.sql
jonatanschroeder: `sql-formatter` has an option to remove those spaces, but it's all-or-nothing, and setting it would also change operators like `+`.

nwalters512: Understood - I think spaces around operators is on the whole the way to go.

PR #7154, opened 2023-02-12
Java grader: fix handling of disabled Java management

jonatanschroeder:
Previous fix actually made the autograder stop working altogether.

### DIFF
--- a/graders/java/autograder.sh
+++ b/graders/java/autograder.sh
@@ -63,7 +63,7 @@ chmod 777 /grade/params/params.json
 # Disable Java management options to hinder student's ability to dump
 # the heap. The port is not a valid int on purpose, to trigger an
 # error on attempt to launch it.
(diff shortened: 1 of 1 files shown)


mwest1066 (approved):
I have no idea what's going on here but I trust you on this 😅

graders/java/autograder.sh
```diff
@@ -63,7 +63,7 @@ chmod 777 /grade/params/params.json
 # Disable Java management options to hinder student's ability to dump
 # the heap. The port is not a valid int on purpose, to trigger an
```
echuber2, line 64: Should the comment here about the port be removed? (I wonder if the incidents reported on Slack related to this were students trying to do something...)

jonatanschroeder: It was not. The grader was failing even with regular submissions (it was pretty easy to reproduce). But yes, the comment can go. PR has been merged for now, but I'll likely revisit this and make the protection stronger when I have some time.

PR #7162, opened 2023-02-14
[WIP] Prometheus metrics

nwalters512:
In my ongoing effort to better understand what's happening in PrairieLearn's internals in production, I'm experimenting with exposing Prometheus metrics that can be scraped from each server.

### DIFF
--- a/middlewares/prometheus.js
+++ b/middlewares/prometheus.js
@@ -1,0 +1,3 @@
+const client = require('prom-client');
+
+client.collectDefaultMetrics();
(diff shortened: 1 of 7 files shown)


nwalters512:
After talking with @mwest1066, we decided to go with [OpenTelemetry metrics](https://opentelemetry.io/docs/instrumentation/js/instrumentation/#metrics) instead. We can still expose a Prometheus endpoint to scrape them if we want, but they can also be natively exported to services like Honeycomb.

PR #7164, opened 2023-02-15
Add support for OpenTelemetry metrics

nwalters512:
This PR introduces the infrastructure to support emitting [OpenTelemetry metrics](https://opentelemetry.io/docs/reference/specification/metrics/). In production, we'll send these to Honeycomb, but in the future we could also expose them to be scraped by Prometheus via something like [`@opentelemetry/exporter-prometheus`](https://www.npmjs.com/package/@opentelemetry/exporter-prometheus).

I added a single instrumentation as a proof-of-concept: we now record the duration and success/failure count of question operations (generate, render, etc).


lib/config.js
```diff
@@ -302,9 +302,12 @@ config.openTelemetryEnabled = false;
  * Note that the `console` exporter should almost definitely NEVER be used in
  * production environments.
  *
- * @type {'console' | 'honeycomb'}
+ * @type {'console' | 'honeycomb' | 'jaeger'}
```
mwest1066, line 305: TIL about Jaeger!

packages/opentelemetry/src/index.ts
```diff
@@ -291,5 +340,30 @@ export async function instrumented<T>(
...
+  meter: Meter,
+  name: string,
+  fn: () => Promise<T> | T
+): Promise<T> {
+  const success = meter.createCounter(`${name}.success`, { valueType: ValueType.INT });
+  const error = meter.createCounter(`${name}.error`, { valueType: ValueType.INT });
+  const histogram = meter.createHistogram(`${name}.duration`, {
+    unit: 'milliseconds',
+    valueType: ValueType.DOUBLE,
+  });
```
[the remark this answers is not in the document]
mwest1066: I can't see any downside to using a Map for caching the metrics. Slight additional code complexity here, but that's ok.

packages/opentelemetry/src/index.ts
```diff
@@ -291,5 +340,30 @@ export async function instrumented<T>(
     });
 }
 
-export { trace, context, SpanStatusCode } from '@opentelemetry/api';
+export async function instrumentedWithMetrics<T>(
+  meter: Meter,
+  name: string,
+  fn: () => Promise<T> | T
+): Promise<T> {
+  const success = meter.createCounter(`${name}.success`, { valueType: ValueType.INT });
```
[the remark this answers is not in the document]
mwest1066: Depending on how things look in HC I'm leaning towards maintaining a separate `error` counter, even if it's redundant, just to make it more natural to use the metrics. I feel like having a clearly named `error` metric is way simpler than needing to remember how to compute it on the fly whenever we need it.

packages/opentelemetry/README.md
```diff
@@ -59,4 +65,18 @@ await tracer.startActiveSpan('span.name', async (span) => {
...
+import { instrumentedWithMetrics } from '@prairielearn/opentelemetry';
+
+await instrumentedWithMetrics('operation.name', async () => {
+  const random = Math.random() * 1000;
+  await new Promise((resolve) => setTimeout(resolve, random));
+  if (random > 900) {
+    throw new Error('Failed!');
+  }
+});
+```
```
mwest1066, line 82: I really like the `instrumentedWithMetrics()` wrapper. How are the [buckets](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#explicit-bucket-histogram-aggregation) being specified for the underlying histogram?

packages/opentelemetry/README.md
```diff
@@ -23,6 +25,10 @@ await init({
 
 This will automatically instrument a variety of commonly-used Node packages.
 
+When using code from the OpenTelemetry libraries, make sure you import it via `@prairielearn/opentelemetry` instead of installing it separately to ensure that there is only one version of each OpenTelemetry package in use at once. If the desired functionality is not yet exported, please add it!
```
mwest1066, line 28: What would it look like if we wanted to directly report some other metric in our application code, for example using a plain Counter?

packages/opentelemetry/src/index.ts
```diff
@@ -291,5 +340,30 @@ export async function instrumented<T>(
...
+  const error = meter.createCounter(`${name}.error`, { valueType: ValueType.INT });
+  const histogram = meter.createHistogram(`${name}.duration`, {
+    unit: 'milliseconds',
+    valueType: ValueType.DOUBLE,
+  });
+
+  const start = Date.now();
```
mwest1066, line 355: Should we be using [`performance.now()`](https://nodejs.org/api/perf_hooks.html#performancenow) instead?

mwest1066 (approved):

packages/opentelemetry/src/index.ts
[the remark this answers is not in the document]
nwalters512: Sorry, want to check if we're talking about the same thing. `error` is not redundant; `success` is. Would you still like to keep that one?

packages/opentelemetry/src/index.ts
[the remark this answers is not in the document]
nwalters512: Sounds good; I'll make the change.

lib/config.js
nwalters512: Yeah, it's really nice for locally exploring traces without needing to export them all the way to Honeycomb! We've supported it since #6514, this type just needed to be updated.

packages/opentelemetry/README.md
nwalters512: They default to the following ([source](https://github.com/open-telemetry/opentelemetry-js/blob/47444f2a4c13a80f3328ffafc5b4c8db9e9a0886/packages/sdk-metrics/src/view/Aggregation.ts#L105)):

```ts
[0, 5, 10, 25, 50, 75, 100, 250, 500, 1000]
```

Unfortunately, configuring them is kind of a pain; see https://github.com/open-telemetry/opentelemetry-js/issues/2999#issuecomment-1432337849. We can still do it, of course; it's just super unfortunate that the boundaries have to be described in a completely different part of the app.

packages/opentelemetry/README.md
nwalters512: Heh, I was waiting to document that until we decided whether or not we'd use a `Map` to store metrics, as that would change both the API and what ends up being documented. I'll add that before merging.

packages/opentelemetry/src/index.ts
nwalters512: Yes, I keep forgetting this exists! Will change.

packages/opentelemetry/src/index.ts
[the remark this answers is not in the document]
mwest1066: Oh sorry, I did misread that. I don't think we need a separate `success` counter. I normally want to plot the total count and the error count, so having an explicit success count feels unnecessary. And as you say, if we really want it then we can always infer it from the other two.

packages/opentelemetry/README.md
mwest1066: Oh yuck! In practice we'll probably be ok with those default boundaries for most of the timing we are doing, but this is really clunky.

packages/opentelemetry/README.md
mwest1066: Awesome, thanks!

packages/opentelemetry/README.md
nwalters512: Yeah, it's not great. I opened https://github.com/open-telemetry/opentelemetry-js/issues/3619 to discuss this problem specifically.

packages/opentelemetry/src/index.ts
nwalters512: Switched to `performance.now()` in af91b32.

packages/opentelemetry/src/index.ts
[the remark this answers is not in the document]
nwalters512: I dropped the separate `success` counter in af91b32.

packages/opentelemetry/src/index.ts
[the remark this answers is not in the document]
nwalters512: I switched to caching in a map in fe99a61.

packages/opentelemetry/README.md
nwalters512: I added some more documentation in fe99a61, how does that look?

mwest1066: Looks super easy to use. Nice!

PR #7166, opened 2023-02-15
fix: tools/build-image-if-needed.sh under bash

trombonekenny:
I noticed this error in the CI builds for the docker images:
```
./tools/build-image-if-needed.sh: 18: [[: not found
```
Which is a hint the script is running under `sh` and not `bash`. This fix changes the way CI launches the script to make sure it's running under bash. I'm not sure if the error was breaking or not but I guess we'll see on this CI run?


trombonekenny:
I think we could take the `set -ex` out of the scripts too and cut down on the verbosity of the CI report, but it seems like several of the CI scripts run in verbose mode, so I didn't change it.

nwalters512 (approved):

PR #7173, opened 2023-02-16
Use object-hash to speed up cache key computation

nwalters512:
This should reduce synchronous thread-blocking work in prod. Note that the value of the cache key will change, so we'll temporarily get zero cache hits.

Benchmark:

```
cpu: Apple M1 Max
runtime: node v18.12.0 (arm64-darwin)

benchmark                                  time (avg)             (min … max)       p75       p99      p995
----------------------------------------------------------------------------- -----------------------------
• big strings
----------------------------------------------------------------------------- -----------------------------
JSON.stringify + crypto.subtle.digest   68.53 ms/iter   (66.05 ms … 71.24 ms)  70.27 ms  71.24 ms  71.24 ms
JSON.stringify + crypto.createHash      61.59 ms/iter   (59.49 ms … 66.89 ms)  62.03 ms  66.89 ms  66.89 ms
object-hash                                11 ms/iter   (10.78 ms … 11.75 ms)  11.02 ms  11.75 ms  11.75 ms

summary for big strings
  object-hash
   5.6x faster than JSON.stringify + crypto.createHash
   6.23x faster than JSON.stringify + crypto.subtle.digest

• big object
----------------------------------------------------------------------------- -----------------------------
JSON.stringify + crypto.subtle.digest   63.03 ms/iter    (62.2 ms … 64.34 ms)  63.43 ms  64.34 ms  64.34 ms
JSON.stringify + crypto.createHash      58.03 ms/iter   (56.34 ms … 61.51 ms)  58.21 ms  61.51 ms  61.51 ms
object-hash                             10.08 ms/iter    (9.58 ms … 13.03 ms)  10.21 ms  13.03 ms  13.03 ms

summary for big object
  object-hash
   5.76x faster than JSON.stringify + crypto.createHash
   6.25x faster than JSON.stringify + crypto.subtle.digest
```

Benchmark source:

```ts
// @ts-check
const crypto = require('node:crypto');
const objectHash = require('object-hash');

const bigString = crypto.randomBytes(5 * 1024 * 1024).toString('hex');
const bigStrings = { foo: bigString, bar: bigString };

// Create a variable named `bigObject` that hsa 1000 keys, each of which is a
// string of 1000 bytes of random data.
const bigObject = {};
for (let i = 0; i < 1000; i++) {
  bigObject[i] = crypto.randomBytes(10000).toString('hex');
}

async function hashJsonDigest(obj) {
  const data = JSON.stringify(obj);
  const encoder = new TextEncoder();
  const digest = await crypto.subtle.digest('SHA-1', encoder.encode(data));
  Buffer.from(digest).toString('base64');
}

function hashJsonCreateHash(obj) {
  const data = JSON.stringify(obj);
  crypto.createHash('sha1').update(data).digest('base64');
}

function hashObjectHash(obj) {
  objectHash(obj, { algorithm: 'sha1', encoding: 'base64' });
}

async function main() {
  const { group, bench, run } = await import('mitata');

  group('big strings', () => {
    bench('JSON.stringify + crypto.subtle.digest', async () => hashJsonDigest(bigStrings));
    bench('JSON.stringify + crypto.createHash', () => hashJsonCreateHash(bigStrings));
    bench('object-hash', () => hashObjectHash(bigStrings));
  });

  group('big object', () => {
    bench('JSON.stringify + crypto.subtle.digest', async () => hashJsonDigest(bigObject));
    bench('JSON.stringify + crypto.createHash', () => hashJsonCreateHash(bigObject));
    bench('object-hash', () => hashObjectHash(bigObject));
  });

  await run();
}

main().catch((err) => {
  console.error(err);
  process.exit(1);
});
```

### DIFF
--- a/question-servers/freeform.js
+++ b/question-servers/freeform.js
@@ -9,6 +9,7 @@ const hash = require('crypto').createHash;
 const parse5 = require('parse5');
 const debug = require('debug')('prairielearn:' + path.basename(__filename, '.js'));
 const { instrumented } = require('@prairielearn/opentelemetry');
+const objectHash = require('object-hash');
 
 const schemas = require('../schemas');
 const config = require('../lib/config');
@@ -1854,7 +1855,7 @@ module.exports = {
   async getCacheKey(course, data, context) {
     try {
       const commitHash = await courseUtil.getOrUpdateCourseCommitHashAsync(course);
-      const dataHash = hash('sha1').update(JSON.stringify({ data, context })).digest('base64');
+      const dataHash = objectHash({ data, context }, { algorithm: 'sha1', encoding: 'base64' });
       return `${commitHash}-${dataHash}`;
     } catch (err) {
       return null;
--- a/question-servers/freeform.js
+++ b/question-servers/freeform.js
@@ -5,7 +5,6 @@ const fs = require('fs-extra');
 const path = require('path');
 const mustache = require('mustache');
 const cheerio = require('cheerio');
-const hash = require('crypto').createHash;
 const parse5 = require('parse5');
 const debug = require('debug')('prairielearn:' + path.basename(__filename, '.js'));
 const { instrumented } = require('@prairielearn/opentelemetry');
--- a/package.json
+++ b/package.json
@@ -98,6 +98,7 @@
         "nodemon": "^2.0.20",
         "numeric": "^1.2.6",
         "oauth-signature": "^1.5.0",
+        "object-hash": "^3.0.0",
         "on-finished": "^2.4.1",
         "parse5": "^6.0.1",
         "passport": "^0.6.0",
--- a/package.json
+++ b/package.json
@@ -166,6 +166,7 @@
         "@types/mocha": "^10.0.1",
         "@types/node": "^18.11.18",
         "@types/node-fetch": "^2.6.2",
+        "@types/object-hash": "^3.0.2",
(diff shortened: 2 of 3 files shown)


question-servers/freeform.js
```diff
@@ -1854,7 +1854,7 @@ module.exports = {
   async getCacheKey(course, data, context) {
     try {
       const commitHash = await courseUtil.getOrUpdateCourseCommitHashAsync(course);
-      const dataHash = hash('sha1').update(JSON.stringify({ data, context })).digest('base64');
+      const dataHash = objectHash({ data, context }, { algorithm: 'sha1', encoding: 'base64' });
```
mwest1066, line 1857: Is this still running on the main thread? You mentioned offloading computation via the web crypto API?

mwest1066 (approved):

question-servers/freeform.js
nwalters512: It is indeed still running on the main thread; for better or worse `object-hash` doesn't utilize the `crypto.subtle` API. TBH I think the fact that the `createHash` API supports incremental `.update()` calls with new data is pretty critical to the way that `object-hash` works.

I think that this should virtually always be fast enough on the main thread to not notice the fact that this is synchronous.

mwest1066: Ok, sounds good, the speedup is a huge win in any case.

PR #7178, opened 2023-02-17
Don't query for question render stats on admins page

nwalters512:
I forgot to do this as part of #2378.


mwest1066 (approved):
Excellent!

PR #7179, opened 2023-02-17
Better Python worker parser

eliotwrobson:
Switch to use the selectolax parser.

### DIFF
--- a/python/traverse.py
+++ b/python/traverse.py
@@ -1,7 +1,9 @@
 import timeit
 from collections import deque
 from typing import Callable, Deque, List, Optional, Tuple, Union
-
+import selectolax as sl
+import codecs
+from itertools import chain
 import lxml.html
 
 ElementReplacement = Optional[
@@ -41,14 +43,25 @@ def traverse_and_execute(
             fn(e)
 
 
-def format_attrib_value(v: str) -> str:
-    # https://html.spec.whatwg.org/multipage/parsing.html#escapingString
-    return v.replace("&", "&amp;").replace('"', "&quot;").replace("\xa0", "&nbsp;")
+def get_elements_list(html: str):
+    parser = sl.parser.HTMLParser(html)
 
+    return list(chain(
+        parser.head.iter(include_text=True),
+        parser.body.iter(include_text=True),
+    ))
+
+def format_attrib_value(v: Optional[str]) -> str:
+    # https://html.spec.whatwg.org/multipage/parsing.html#escapingString
+
+    if v is None:
+        return ""
+
+    return v.replace("&", "&amp;").replace('"', "&quot;").replace("\xa0", "&nbsp;")
 
 def get_source_definition(element: lxml.html.HtmlElement) -> str:
     attributes = (
-        f'''{k}="{format_attrib_value(v)}"''' for k, v in element.attrib.items()
+        f'{k}="{format_attrib_value(v)}"' for k, v in element.attributes.items()
     )
     return f"<{' '.join((element.tag, *attributes))}>"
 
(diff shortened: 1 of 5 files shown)


python/traverse.py
```diff
@@ -26,50 +26,70 @@
...
+    return res
 
-def format_attrib_value(v: str) -> str:
+
+def traverse_and_execute(html: str, fn: Callable[[slp.Node], None]) -> None:
+    # elements = get_elements_list(html)
+
+    page = slp.HTMLParser(html)
+
+    if page.head is None:
```
nwalters512, line 47: Does `head` correspond to `<head>`, or something else? If the former, why are we iterating over the nodes in head?

python/traverse.py
```diff
@@ -26,50 +26,70 @@
...
+    if parser.body is not None:
+        res.extend(parser.body.iter(include_text=True))
 
+    return res
 
-def format_attrib_value(v: str) -> str:
+
+def traverse_and_execute(html: str, fn: Callable[[slp.Node], None]) -> None:
+    # elements = get_elements_list(html)
+
```
nwalters512, line 44: Can we remove this?

```suggestion
```

python/traverse.py
```diff
@@ -26,50 +26,70 @@
...
+def get_elements_list(html: str) -> List[slp.Node]:
+    parser = slp.HTMLParser(html)
+    res = []
 
-    # If there's leading text, the first element of the array will be a string.
-    # We can just discard that.
-    if isinstance(elements[0], str):
-        del elements[0]
+    if parser.head is not None:
+        res.extend(parser.head.iter(include_text=True))
```
nwalters512, line 34: Same question as below, why are we looking at `head` at all?

python/traverse_test.py
```diff
@@ -29,35 +33,35 @@ def test_traverse_and_replace_none() -> None:
...
 def test_traverse_and_replace_comment() -> None:
-    def replace(e) -> ElementReplacement:
+    def replace(e: slp.Node) -> ElementReplacement:
         if e.tag == "p":
             return "<!-- Goodbye --><i>world</i>"
         return e
 
     html = traverse_and_replace("<p>Hello</p>", replace)
-    assert html == "<!-- Goodbye --><i>world</i>"
+    assert html == "<i>world</i>"
```
nwalters512, line 42: 👀 why this change?

python/traverse_test.py
```diff
@@ -287,15 +291,15 @@ def replace(e) -> ElementReplacement:
 
 
 def test_traverse_and_replace_void_elements() -> None:
-    def replace(e) -> ElementReplacement:
+    def replace(e: slp.Node) -> ElementReplacement:
         return e
 
-    html = traverse_and_replace('<div><br><input name="input"</div>', replace)
+    html = traverse_and_replace('<div><br><input name="input"></div>', replace)
```
nwalters512, line 297: Heh, neat that the old parser was resilient enough to handle this!

python/traverse.py
eliotwrobson: It's a quirk of the parser. When `HTMLParser(html)` is called on an HTML fragment, it treats the whole thing as it's own HTML page, putting some leading tags into the `head` automatically, and some into the `body` (which it creates). Checking for `None` and looking at the iter, we read into those tags and just get the parsed version of the input.

python/traverse.py
eliotwrobson: Discussed in https://github.com/PrairieLearn/PrairieLearn/pull/7179/files/1cbd286caa0eec6a789665879de5a1d186dba96b#r1117968904. Basically, the parser treats `html` as a complete page and automatically creates `head` and `body` tags. In this case, the body is treated as one of the children of the head tag, so calling `traverse` in head goes through all the nodes in the page.

python/traverse_test.py
eliotwrobson: Another quirk of the parser, leading comments that aren't wrapped in a tag get stripped. I think the way to fix this would be, in `get_elements_list`, wrap the html input in a `<body>` tag, then run the parser on that. That way, nothing should get stripped and we can iterate all the nodes more easily. Though I'm not sure if any `script` or similar tags will get placed into the head automatically. I didn't have time to experiment with it yet, so I just opted to remove some of the comments from the test cases.

python/traverse_test.py
eliotwrobson: Yes, sadly the new one isn't quite as robust (the closing `</div>` tags got turned into attributes).

nwalters512: Hmm, this actually seems highly undesirable. `lxml` handled this in a somewhat-sensible way, as would our current parser. If selectolax isn't as forgiving, I'm not sure if it's a good idea to move to it, as there are undoubtedly *tons* of existing questions with invalid HTML.

eliotwrobson: I will double check, there might be a way to coax more resilient behavior out of this.

eliotwrobson: So I couldn't get selectolax to be more lenient about this, though I'm not sure how negatively this might affect some questions. Can discuss on Slack, but `lxml` runs into other issues that might be more problematic. Maybe it's worth trying to use the builtin Python HTML parser? Kinda unfortunate all of the libraries I've seen for this seem to be more geared towards web scraping than what we're trying to do.

PR #7180, opened 2023-02-17
C Grader: support for regular expressions and option to highlight matches

jonatanschroeder:
In programs with relatively long outputs it is hard for students to identify what was missing in their program. This PR uses ANSI colors to highlight what was expected and what was missing in the program.

Some of the comparison functionality was also changed to simplify the implementation of highlighting. In particular, instead of replacing the output and string with a lower-case and space-only output for comparison, regular expressions are used instead.

image

### DIFF
--- a/graders/c/cgrader/cgrader.py
+++ b/graders/c/cgrader/cgrader.py
@@ -338,9 +338,9 @@ class CGrader:
         outcmp = out
         if highlight_matches and out:
             for _, r in exp_output:
-                out = r.sub(f"\\033[32m\\g<0>\\033[0m", out)
+                out = r.sub(r"\033[32m\g<0>\033[0m", out)
             for _, r in reject_output:
-                out = r.sub(f"\\033[31m\\g<0>\\033[0m", out)
+                out = r.sub(r"\033[31m\g<0>\033[0m", out)
         if not out:
             out = "(NO OUTPUT)"
         elif not out.endswith("\n"):
@@ -355,21 +355,17 @@ class CGrader:
                 else " one of"
             )
             msg = f"Expected{comment}:\n\t" + "\n\t".join(
-                [
-                    f"\033[32m{t}\033[0m"
-                    if highlight_matches and r.search(outcmp)
-                    else t
-                    for t, r in exp_output
-                ]
+                f"\033[32m{t}\033[0m"
+                if highlight_matches and r.search(outcmp) is not None
+                else t
+                for t, r in exp_output
             )
             if reject_output:
                 msg += "\nBut not:\n\t" + "\n\t".join(
-                    [
-                        f"\033[31m{t}\033[0m"
-                        if highlight_matches and r.search(outcmp)
-                        else t
-                        for t, r in reject_output
-                    ]
+                    f"\033[31m{t}\033[0m"
+                    if highlight_matches and r.search(outcmp) is not None
+                    else t
+                    for t, r in reject_output
                 )
         elif msg is None:
             msg = ""
@@ -381,10 +377,10 @@ class CGrader:
             out = out[0:size_limit] + "\nTRUNCATED: Output too long."
             points = False
         elif not (all if must_match_all_outputs else any)(
-            [r.search(outcmp) for _, r in exp_output]
+            r.search(outcmp) is not None for _, r in exp_output
         ):
             points = False
-        elif any([r.search(outcmp) for _, r in reject_output]):
+        elif any(r.search(outcmp) is not None for _, r in reject_output):
             points = False
 
         return self.add_test_result(
@@ -576,4 +572,5 @@ class CPPGrader(CGrader):
         super(CPPGrader, self).__init__(compiler)
 
 
-CGrader().start()
+if __name__ == "__main":
+    CGrader().start()
--- a/graders/c/cgrader/cgrader.py
+++ b/graders/c/cgrader/cgrader.py
@@ -313,15 +313,20 @@ class CGrader:
             reject_output = [reject_output]
 
         def compile_re(t):
-            return re.compile(
-                "\\s+".join(map(re.escape, re.split("\\s+", str(t))))
-                if ignore_consec_spaces
-                else re.escape(str(t)),
-                re.I if ignore_case else 0,
+            if isinstance(t, re.Pattern):
+                return (t.pattern, t)
+            return (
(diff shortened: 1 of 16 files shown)


graders/c/cgrader/cgrader.py
```diff
@@ -302,69 +301,90 @@ def test_run(
...
             command if args is None else ([command] + args),
             input,
             sandboxed=True,
             timeout=timeout,
         )
+
         outcmp = out
+        if highlight_matches and out:
+            for _, r in exp_output:
+                out = r.sub(f"\\033[32m\\g<0>\\033[0m", out)
```
eliotwrobson, line 341: Why are these f-strings? It doesn't look like there's anything being substituted here. Would be better to use r-strings, since this means not having to use `\\` as an escape character.

graders/c/cgrader/cgrader.py
```diff
@@ -302,69 +301,90 @@ def test_run(
...
-            and must_match_all_outputs
-            and [t for t in exp_output if str(t) not in outcmp]
-        ):
-            points = False
-        elif (
-            exp_output is not None
-            and not must_match_all_outputs
-            and not [t for t in exp_output if str(t) in outcmp]
+        elif not (all if must_match_all_outputs else any)(
+            [r.search(outcmp) for _, r in exp_output]
```
eliotwrobson, line 384: I don't think you need to have these in a list, `all` and `any` can take in arbitrary generators. Also, better to be explicit about checking for `None` vs a match object.
```suggestion
            r.search(outcmp) is not None for _, r in exp_output
```

graders/c/cgrader/cgrader.py
```diff
@@ -302,69 +301,90 @@ def test_run(
...
-                exp_output = [re.sub(r"\s+", " ", str(t)) for t in exp_output]
+        if msg is None and exp_output:
+            comment = (
+                ""
+                if len(exp_output) == 1
+                else " all of"
+                if must_match_all_outputs
+                else " one of"
+            )
+            msg = f"Expected{comment}:\n\t" + "\n\t".join(
```
eliotwrobson, line 357: I don't think these need to be in a list (can delete the `[]` brackets) because join can take in any generator expression.

graders/c/cgrader/cgrader.py
```diff
@@ -302,69 +301,90 @@ def test_run(
...
-                [f"\n{t}\n" if "\n" in str(t) else f'"{t}"' for t in exp_output]
-            ) + (
-                ' but not "%s"' % '"/"'.join([str(t) for t in reject_output])
-                if reject_output
-                else ""
+
+        def compile_re(t):
+            if isinstance(t, re.Pattern):
+                return (t.pattern, t)
+            return (
```
eliotwrobson, line 318: Maybe put `str(t)` in a variable? Probably doesn't matter for efficiency but makes things more readable.

eliotwrobson (changes requested):
See comments. Have some minor additional comments about this file (would be nice to have type annotations enabled, and the last line should be wrapped with a `if __name__ == "__main__":` check), but these are independent of the new feature.

graders/c/cgrader/cgrader.py
jonatanschroeder: Resolved in ea794457e. These are just copied from the `msg` processing where f-strings are needed, and I forgot to replace them. I thought the linter (or black) would take care of this?

graders/c/cgrader/cgrader.py
jonatanschroeder: Resolved in ea794457e.

graders/c/cgrader/cgrader.py
jonatanschroeder: Resolved in ea794457e.

graders/c/cgrader/cgrader.py
jonatanschroeder: Done in de0103b9d.

graders/c/cgrader/cgrader.py
eliotwrobson: @jonatanschroeder I thought so too, enabling linting for the grader files required some changes, which included getting rid of unused f-strings in some places. Could you manually run `flake8` on this file and see if any errors get thrown? I'm wondering if the linter is running correctly now.

jonatanschroeder: I checked out the commit before your comment, and flake8 does indeed complain about it. `make lint-python` also does, as does `docker/lint_python.sh`. Not sure why this would fail locally, but not in Github CI. @nwalters512 any thoughts?

graders/c/cgrader/cgrader.py
```diff
@@ -556,4 +574,5 @@ def __init__(self, compiler="g++"):
         super(CPPGrader, self).__init__(compiler)
 
 
-CGrader().start()
+if __name__ == "__main":
```
eliotwrobson, line 577:
```suggestion
if __name__ == "__main__":
```

graders/c/cgrader/cgrader.py
jonatanschroeder: Ok, I believe the issue is that the `graders` directory is not added to the Docker container (see .dockerignore), so the Python linter doesn't find any files in those directories.

graders/c/cgrader/cgrader.py
```diff
@@ -302,69 +301,88 @@ def test_run(
...
-            and not must_match_all_outputs
-            and not [t for t in exp_output if str(t) in outcmp]
+        elif not (all if must_match_all_outputs else any)(
+            r.search(outcmp) is not None for _, r in exp_output
         ):
             points = False
-        elif reject_output is not None and [
-            t for t in reject_output if str(t) in outcmp
-        ]:
+        elif any(r.search(outcmp) is not None for _, r in reject_output):
```
eliotwrobson, line 385: Could condense some of these cases because they all set `points = False`, but I understand if you think this makes the code harder to read.

eliotwrobson (approved):
Looks good! I think that typechecking + fixing the `.dockerignore` can be deferred to later PRs as long as running `make lint-python` works locally.

graders/c/cgrader/cgrader.py
jonatanschroeder: Combined the last two in 96c9303, but not the rest since the order of those and the additional change in output is deliberate.

PR #7185, opened 2023-02-20
feat: Support for style and ansi-color in pl-code

jonatanschroeder:
* [x] Element
* [x] Docs
* [x] Example


elements/pl-code/pl-code.py
```diff
@@ -21,6 +24,22 @@
 COPY_CODE_BUTTON_DEFAULT = False
 
 
+class AnsiColorStyle(Style):
+    styles = dict(get_style_by_name("friendly").styles)
+    # These are the same colors used in pl-external-grader-result
+    ansi_colors = {
```
eliotwrobson, line 30: If these colors are duplicated, might be a good idea to put these in a shared file somewhere?

jonatanschroeder: Yes, I thought about that. There is a minor issue that ansi2html expects an ordered array while pygments requires a dict, so there is a need to adjust from one to the other, though this is doable.

Another option is to have pl-external-grader-result defer to pl-code for this functionality, so all colors would be found here.

eliotwrobson: Having external grader result defer is probably the best option imo, since it should mean getting to simplify the backend of that element, similar to what we did for pl-variable-output.

jonatanschroeder: Unfortunately `pygments-ansi-color` doesn't currently support bright colors (https://github.com/chriskuehl/pygments-ansi-color/issues/8). So we should either keep both elements separate, or lose access to bright colors in pl-external-grader-result.

jonatanschroeder: Submitted https://github.com/chriskuehl/pygments-ansi-color/pull/27, if it is pushed and deployed relatively soon we can use it.

jonatanschroeder: The update was merged and deployed to pip, so 8eb0187ad updates the code to use bright colors as well.

elements/pl-code/pl-code.py
```diff
@@ -20,6 +23,26 @@
 DIRECTORY_DEFAULT = "."
 COPY_CODE_BUTTON_DEFAULT = False
 
+# These are the same colors used in pl-external-grader-result
+ansi_colors = {
```
eliotwrobson, line 27:
```suggestion
ANSI_COLORS = {
```
Convention for class constants is to have them in all caps.

elements/pl-code/pl-code.py
```diff
@@ -131,6 +155,14 @@ def prepare(element_html, data):
                 f'Unknown language: "{language}". Must be one of {", ".join(allowed_languages)}'
             )
 
+    style = pl.get_string_attrib(element, "style", STYLE_DEFAULT)
+    pygments_style = get_style_by_name(style)
+    if pygments_style is None:
+        allowed_styles = STYLE_MAP.keys()
+        raise Exception(
```
eliotwrobson, line 162:
```suggestion
        raise KeyError(
```

eliotwrobson (approved):
Minor style items but looks good overall!

elements/pl-code/pl-code.py
jonatanschroeder: Replaced all generic Exceptions with `KeyError` or `ValueError` in 118f90313.

elements/pl-code/pl-code.py
jonatanschroeder: Done in 118f90313.

elements/pl-code/pl-code.py
jonatanschroeder: Moved this discussion to #7191.

elements/pl-code/pl-code.py
```diff
@@ -178,7 +212,7 @@ def render(element_html, data):
             base_path = os.path.join(data["options"]["question_path"], directory)
         file_path = os.path.join(base_path, source_file_name)
         if not os.path.exists(file_path):
-            raise Exception(f'Unknown file path: "{file_path}".')
+            raise ValueError(f'Unknown file path: "{file_path}".')
```
eliotwrobson, line 215: Somewhat tangential (but we're already rearranging this file): two lines below this, I think the ending newline removal is broken, and the file doesn't have to be manually closed.

eliotwrobson (changes requested):
Noticed something weird in the pl-code controller. Given the volume of changes, I think adding type hints as part of this is the way to go.

elements/pl-code/pl-code.py
jonatanschroeder: Replaced with builtins in 17aa3b722.

eliotwrobson: Are the Canadian servers on Python >= 3.9? Those builtins are from Python 3.9, but if no versioning issues I think the new code looks much cleaner?

jonatanschroeder: 3.10.6 https://prairielearn.slack.com/archives/C0XBTPZMZ/p1673396644523079?thread_ts=1673392551.982729&cid=C0XBTPZMZ

jonatanschroeder: It seems you're on top of it, but apparently pyright doesn't support it yet...

```
pyright 1.1.295
/home/jonatan/pl/elements/pl-code/pl-code.py
  /home/jonatan/pl/elements/pl-code/pl-code.py:221:29 - error: Cannot access member "removesuffix" for type "str"
    Member "removesuffix" is unknown (reportGeneralTypeIssues)
  /home/jonatan/pl/elements/pl-code/pl-code.py:235:39 - error: Cannot access member "removeprefix" for type "str"
    Member "removeprefix" is unknown (reportGeneralTypeIssues)
```

eliotwrobson: Yeah, no idea why pyright doesn't pick that up, weird.

elements/pl-code/pl-code.py
```diff
@@ -202,19 +232,18 @@ def render(element_html, data):
...
 
-    if specify_language:
-        lexer = get_lexer_by_name(language)
-    else:
-        lexer = NoHighlightingLexer()
+    lexer = NoHighlightingLexer() if language is None else get_lexer_by_name(language)
+
+    pygments_style = get_style_by_name(style)
+
+    class CustomStyleWithAnsiColors(pygments_style):  # type: ignore
```
nwalters512, line 241: I opened https://github.com/python/typeshed/pull/9803 to remove the need for `# type: ignore` here.

elements/pl-code/pl-code.py
```diff
@@ -202,19 +232,18 @@ def render(element_html, data):
         #
         # which technically starts with a newline, but we probably
         # don't want a blank line at the start of the code block.
-        code = pl.inner_html(element)
-        if len(code) > 1 and code[0] == "\r" and code[1] == "\n":
-            code = code[2:]
-        elif len(code) > 0 and (code[0] == "\n" or code[0] == "\r"):
-            code = code[1:]
+        code = pl.inner_html(element).removeprefix("\r").removeprefix("\n")  # type: ignore
```
nwalters512, line 235: Consider annotating `pl.inner_html` with a return type to remove the need for `# type: ignore` here; same with above.

jonatanschroeder:
Probably worth adding here as well (https://github.com/PrairieLearn/PrairieLearn/issues/7191#issuecomment-1442529183):

> One issue with this is that pygments styles with dark background have typically no proper support for the default (Text) token (see https://github.com/pygments/pygments/issues/870 and https://github.com/pygments/pygments/issues/1526). I couldn't find any style with dark background and support for text (the default, non-formatted token) with a light color. So at this point I wouldn't recommend using pl-code (or anything using pygments) with a dark background unless there is an expectation that the user (instructor?) has reviewed the visibility to ensure it's good.

nwalters512 (approved):
Looks reasonable to me!

elements/pl-code/pl-code.py
eliotwrobson: I think this is actually unavoidable, Python typechecking doesn't like dynamic class creation, see https://github.com/microsoft/pyright/issues/2058

elements/pl-code/pl-code.py
eliotwrobson: That's not the issue, the `removeprefix` and `removesuffix` functions aren't getting recognized by pyright for some reason.

elements/pl-code/pl-code.py
nwalters512: I tested this change locally and it worked; is there something else that you're thinking will be broken?

elements/pl-code/pl-code.py
jonatanschroeder: The issue is not `pl.inner_html`, it's `removeprefix`, which is a feature introduced to `str` in Python 3.9 that apparently is not yet supported by pyright for some reason.

elements/pl-code/pl-code.py
eliotwrobson: Oh interesting, I guess pyright got updated to work in these situations or something.

elements/pl-code/pl-code.py
eliotwrobson: I think i know what's up actually, give me a sec to push.

nwalters512: We pinned pyright to 3.8, so it makes sense that 3.9 features wouldn't be available: https://github.com/PrairieLearn/PrairieLearn/blob/c50f199cb622b2721bd64cb7d3b7ac55aeadb7d5/pyproject.toml#L78. Let's bump that to 3.9? That's the oldest version currently in production.

eliotwrobson: @nwalters512 read my mind https://github.com/PrairieLearn/PrairieLearn/pull/7185/commits/93647d71907ba271d27938dffd9685d473b789c2

Going to go ahead and change to the simpler type annotations as well. I thought we were on 3.10 though?

nwalters512: Not everywhere; I still need to find the bandwidth to get us updated in Canada.

jonatanschroeder:
@eliotwrobson I believe I can't merge this until you make a new review, since your last review requested changes.

eliotwrobson (approved):

PR #7187, opened 2023-02-21
Allow multiple rows for administrator query

nwalters512:
Local testing of #7178 didn't reveal any issues because there was only one admin locally 🤦


mwest1066 (approved):

mwest1066:
> Local testing of #7178 didn't reveal any issues because there was only one admin locally 🤦

😂

PR #7194, opened 2023-02-24
Move assessment instance log from sproc to lib

jonatanschroeder:
This is a pretty clear candidate to move to a lib.

I took the opportunity to change the handlers for get in the instance log page to async.


lib/assessment.js
```diff
@@ -510,4 +512,32 @@ module.exports = {
...
     });
   },
+
+  /**
+   * Selects a log of all events associated to an assessment instance.
+   *
+   * @param {number} assessment_instance_id - The ID of the assessment instance.
+   * @param {boolean} include_files - A flag indicating if submitted files should be included in the
+   * log.
+   * @returns {Promise<Array<{event_name: string,
```
nwalters512, line 522: If you're going to define a type here, let's define a schema with [zod](https://github.com/PrairieLearn/PrairieLearn/pull/7119) and use `queryValidatedRows` instead. For an example, see the following:

https://github.com/PrairieLearn/PrairieLearn/blob/daa897834aaf09d0bae6845733ec282bff37a848/pages/administratorWorkspaces/administratorWorkspaces.html.js#L6-L26

https://github.com/PrairieLearn/PrairieLearn/blob/daa897834aaf09d0bae6845733ec282bff37a848/pages/administratorWorkspaces/administratorWorkspaces.js#L14

jonatanschroeder: Done in 8c78bc0 and f115be6.

nwalters512 (approved):

PR #7196, opened 2023-02-24
Stop workspaces after being in background for a while

### LINKED ISSUE #6599: Time out workspaces after tab in background for certain period
If a workspace tab is in the background for more than X minutes, we should prompt the user to see if they're still there. After another timeout, we should unload the iframe entirely and stop the socket heartbeat so that the workspace container can be killed.

Relevant API: https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API

nwalters512:
This PR uses the [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) to shut down workspaces that have been hidden in a background tab for too long.

Closes #6599.


pages/workspace/workspace.html.js
```diff
@@ -288,11 +312,28 @@ function Workspace({ navTitle, showLogs, resLocals }) {
...
             }, heartbeatInterval * 1000);
+
+            let timeoutId = null;
+            document.addEventListener('visibilitychange', () => {
+              if (document.visibilityState == 'hidden') {
+                timeoutId = setTimeout(() => {
+                  clearInterval(intervalId);
+                  workspaceFrame.src = 'abount:blank';
+                  showStoppedFrame();
+                }, 10 * 1000);
```
nwalters512, line 328: This is set artificially low for testing. This should probably be configurable too.

pages/workspace/workspace.html.js
```diff
@@ -288,11 +312,28 @@ function Workspace({ navTitle, showLogs, resLocals }) {
...
               socket.emit('heartbeat', { workspace_id: workspaceId }, (msg) => {
                 console.log('heartbeat, msg =', msg);
               });
             }, heartbeatInterval * 1000);
+
+            let timeoutId = null;
+            document.addEventListener('visibilitychange', () => {
+              if (document.visibilityState == 'hidden') {
+                timeoutId = setTimeout(() => {
+                  clearInterval(intervalId);
```
nwalters512, line 325: Currently, we rely on disabling the heartbeat to shut down the container. Maybe we should use an explicit message to stop it? That would let us track a metric like `workspaces.stopped.visibility.count` if we were interested to know how many workspaces we're shutting down like this.

pages/workspace/workspace.html.js
```diff
@@ -206,11 +206,19 @@ function Workspace({ navTitle, showLogs, resLocals }) {
...
           >
             <i class="d-block fa fa-10x fa-circle-notch fa-spin text-info" aria-hidden="true"></i>
             <span class="sr-only">Loading workspace &hellip;</span>
           </div>
+          <div
+            id="stopped"
+            class="d-none h-100 flex-grow flex-column justify-content-center align-items-center"
+          >
+            <h2>Workspace stopped due to inactivity</h2>
+            <p>Your data was automatically saved. Reload the page to restart the workspace.</p>
```
nwalters512, line 219: Does this copy sound ok? I kind of just threw it together. Here's how it looks:

Screenshot 2023-02-23 at 17 15 43

nwalters512, line 219: @mwest1066 I know we had talked about communicating the specific time limits involved, but I wasn't sure how to communicate the intricacies of "your tab was in the background for too long" vs. "your client didn't heartbeat for a while" and the fact that each of those use different timeouts. Hopefully this generic message is sufficient?

pages/workspace/workspace.html.js
nwalters512: Made configurable in 2323f70. In theory, this is something that we could make vary per course/institution to allow for different "tiers" of service in the future.

pages/workspace/workspace.html.js
mwest1066: I like this idea. The current behavior is due to the fact that we were assuming we wouldn't be able to detect why we went offline, but in this new situation it would be great to get more metrics. We could also register with [`beforeunload`](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) and do a similar thing then?

pages/workspace/workspace.html.js
mwest1066: Sounds good! Of course we are relying on the workspace to have actually automatically saved their data, but I think it's good to be reassuring about this.

mwest1066: I think this is fine. No need to get more specific.

nwalters512: Yeah, I would consider any workspace that *didn't* automatically save data as fundamentally broken for our use case. Might be worth documenting that expectation? I don't think we have docs for building custom workspace images, FWIW.

pages/workspace/workspace.html.js
```diff
@@ -16,7 +16,8 @@ function Workspace({ navTitle, showLogs, resLocals }) {
       <body
         class="d-flex flex-column h-100"
         data-workspace-id="${resLocals.workspace_id}"
-        data-heartbeat-interval="${resLocals.workspaceHeartbeatIntervalSec}"
+        data-heartbeat-interval="${heartbeatIntervalSec}"
+        data-visibility-timeout="${visibilityTimeoutSec}"
```
mwest1066, line 20: When possible I'd like to carry the units along to make later code easier to read (do I need a `*1000` or not?).
```suggestion
        data-heartbeat-interval-sec="${heartbeatIntervalSec}"
        data-visibility-timeout-sec="${visibilityTimeoutSec}"
```

pages/workspace/workspace.html.js
```diff
@@ -224,20 +233,33 @@ function Workspace({ navTitle, showLogs, resLocals }) {
             const heartbeatInterval = Number.parseFloat(
               document.body.getAttribute('data-heartbeat-interval')
             );
+            const visibilityTimeout = Number.parseFloat(
+              document.body.getAttribute('data-visibility-timeout')
+            );
```
mwest1066, line 238:
```suggestion
            const visibilityTimeoutSec = Number.parseFloat(
              document.body.getAttribute('data-visibility-timeout-sec')
            );
```

mwest1066 (approved):

pages/workspace/workspace.html.js
mwest1066: Good point, it would be great to explicitly document this somewhere in https://prairielearn.readthedocs.io/en/latest/workspaces/

pages/workspace/workspace.html.js
nwalters512: Nice, I'll make this change and add a metric.

Not sure I understand what we'd do in an unload handler, can you elaborate?

pages/workspace/workspace.html.js
nwalters512: Done in c8b640d.

pages/workspace/workspace.html.js
nwalters512: Done in c8b640d.

pages/workspace/workspace.html.js
mwest1066: I was thinking that we could register a handler with `beforeunload()` that would also send a websocket message to stop the workspace. This would trigger in the case that the user closes the tab or navigates away. But maybe these actions will already trigger a `visibilityState` change?

HOWEVER, thinking about this whole idea further I now think that we should _not_ send any explicit "stop workspace" events and instead we should only rely on heartbeat timeouts. This is because of group workspaces where we have multiple students connected to the same backend workspace (e.g., the collaborative Jupyter workspace). We don't want to shut it down just because one student goes away because the others might still be working.

mwest1066: I guess we could send the visibility timeout message so long as the server doesn't blindly shut the workspace down, but instead uses this as a trigger to check for recent heartbeats (basically running the usual workspace-timeout-check cron job, but just for this workspace). Maybe this is worth it for the metrics? But it's a fair bit of additional code and complexity that we'll be keeping around.

nwalters512: Ahhhh, great point about group workspaces! That makes a lot of sense; I'll remove the explicit stop.

nwalters512: Given this, it probably also makes sense to *not* immediately show a "stopped due to inactivity" message and instead rely on receiving a `change:state` message if when it ultimately stops. What do you think? I think in that case we'd also want to restart the heartbeat if the tab becomes visible again.

workspace_host/interface.sql
```diff
@@ -155,6 +155,9 @@ FROM
   JOIN workspace_hosts AS wh ON (w.workspace_host_id = wh.id)
 WHERE
   w.state = 'stopped'::enum_workspace_state
+  -- We only want workspaces that have been marked as stopped but still have a
+  -- `launch_uuid`, which means it's probably still running on this host.
+  AND w.launch_uuid IS NOT NULL
```
nwalters512, line 160: This is unrelated to the broader change here; I just noticed this while reading the code related to stopping workspaces.

pages/workspace/workspace.html.js
mwest1066: Yes! So you are saying that the client will just do the following two things?

1. Send heartbeats whenever it's visible.
2. Respond to server messages like `change:state` by displaying the message and changing which frame is visible.

This seems very clean!

pages/workspace/workspace.html.js
```diff
@@ -288,11 +316,39 @@ function Workspace({ navTitle, showLogs, resLocals }) {
...
+            let heartbeatIntervalId = null;
+
+            function startHeartbeat() {
+              if (heartbeatIntervalId) return;
+
+              heartbeatIntervalId = setInterval(() => {
+                socket.emit('heartbeat', { workspace_id: workspaceId }, (msg) => {
+                  console.log('heartbeat, msg =', msg);
+                });
+              }, heartbeatIntervalSec * 1000);
```
mwest1066, line 328: Say you have a hidden workspace page that you switch to for 30 seconds and then switch away. When we switch to the page we will start the heartbeat timer, but `setInterval` will wait 60 seconds (`heartbeatIntervalSec`) before it sends the first heartbeat so it won't send one before we stop the timer again. This means we will only send a heartbeat if we actually stay on the workspace page for at least 60 seconds continuously.

```suggestion
              function sendHeartbeat() {
                socket.emit('heartbeat', { workspace_id: workspaceId });
              }
              sendHeartbeat(); // send immediately
              heartbeatIntervalId = setInterval(sendHeartbeat, heartbeatIntervalSec * 1000);
```

The above solution is simple but it has the downside that we'll send lots of heartbeats if the user is constantly toggling back and forwards between a workspace page and another page.

Another possible solution would be something like:
```
let pageLastVisibleTime = Date.now();

setInterval(() => {
  if (Document.visibilityState == 'visible') {
    // Update our page visibility record if we are currently visible.
    pageLastVisibleTime = Date.now();
  }
  // Only send a heartbeat if this page was recently visible.
  if (Date.now() - pageLastVisibleTime < visibilityTimeoutSec * 1000) {
    socket.emit('heartbeat', { workspace_id: workspaceId });
  }
}, heartbeatIntervalSec * 1000);

document.addEventListener('visibilitychange', () => {
  // Every time we switch to or from this page, record that it was visible.
  // This is needed to capture the visibility when we switch to this page
  // and then quickly switch away again.
  pageLastVisibleTime = Date.now();
});
```

The above solution has the disadvantage that the client will still use resources to wake up every minute and check whether the page is visible, but I think this is pretty minimal usage on the client. The main advantages of this solution are that the code is simple with no start/stop of timers, and that it minimizes load on the server because clients will only send at most one heartbeat every 60 seconds.

workspace_host/interface.sql
mwest1066: Good find.

pages/workspace/workspace.html.js
nwalters512: That's what I'm thinking, yeah!

pages/workspace/workspace.html.js
nwalters512: I like the second solution better - I'm all about minimizing server load! I'll make that change.

nwalters512: Done in bac0689.

mwest1066: Seems to be missing the `visibilitychange` handler from the code above?

pages/workspace/workspace.html.js
nwalters512: I documented this expectation in 743cdc6.

PR #7197, opened 2023-02-25
Introduce @prairielearn/logger package

nwalters512:
This gets us one step closer to being able to convert the whole project into TypeScript workspaces 🚀


packages/logger/src/index.ts
```diff
@@ -1,25 +1,21 @@
...
     new winston.transports.Console({
-      name: 'console',
       level: 'info',
       format: format.combine(format.colorize(), format.simple()),
     }),
   ],
 });
 
-logger.addFileLogging = function (filename) {
+export function enableFileLogging(filename: string) {
```
mwest1066, line 13: While we are editing this code, could we add `level` as an argument to this function? I would like to be able to log `debug` and higher to `/var/log/prairielearn.log` and `error` and higher to `/var/log/prairielearn_error.log`, so that we can ship the error log to CloudWatch.

server.js
```diff
@@ -1894,7 +1894,7 @@ if (config.startServer) {
         }
 
         if (config.logFilename) {
-          logger.addFileLogging(config.logFilename);
+          enableFileLogging(config.logFilename);
```
mwest1066, line 1897: Here I'd love to have:
```suggestion
          enableFileLogging(config.logFilename, 'debug');
          enableFileLogging(config.logErrorFilename, 'error');
```

mwest1066 (approved):
Feel free to merge without implementing my suggestions -- we can certainly do those at a later time. (Although it would be nice to at least add the `level` argument to `enableFileLogging()`.)

packages/logger/src/index.ts
nwalters512: Interesting idea! I make this change in 7fe6ffc. I ultimately added an options object to mirror the underlying `winston.transports.File` constructor.

PR #7198, opened 2023-02-26
pl-matrix-component-input simplification

eliotwrobson:
See title. Part of https://github.com/PrairieLearn/PrairieLearn/discussions/7027.


eliotwrobson:
Going to go ahead and close this and extract some of the more useful components into a smaller PR. On further evaluation, the way that the `pl-matrix-component-input` element generates the HTML is kinda messy, and rather than extracting the functions doing this, it should be baked into a mustache template for a new element (perhaps just `pl-matrix-input`). Then, this element can have `symbolic` and `numeric` grading modes which grade relative to a reference matrix correct answer.

PR #7199, opened 2023-02-26
Improve performance of workspace logs parsing

nwalters512:
This PR improves the performance of workspace logs parsing by avoiding the allocation of a new `Buffer` on each iteration of the loop. For an artificial log with 100000 "chunks", this improved performance from ~14 seconds to ~100ms.

This also changes the log handling to read at most 50k lines of logs, which should also help ensure that we don't process/transmit/store ridiculous amounts of information.

### DIFF
--- a/workspace_host/interface.js
+++ b/workspace_host/interface.js
@@ -957,7 +957,13 @@ async function sendLogs(workspaceId, res) {
   try {
     const workspace = await _getWorkspaceAsync(workspaceId);
     const container = await _getDockerContainerByLaunchUuid(workspace.launch_uuid);
-    const logs = await container.logs({ stdout: true, stderr: true, timestamps: true });
+    const logs = await container.logs({
+      stdout: true,
+      stderr: true,
+      timestamps: true,
+      // This should give us a reasonable bound on the worst-case performance.
+      tail: 50000,
+    });
     const parsedLogs = parseDockerLogs(logs);
     res.status(200).send(parsedLogs);
   } catch (err) {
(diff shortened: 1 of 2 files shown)


workspace_host/interface.js
```diff
@@ -957,7 +957,13 @@ async function sendLogs(workspaceId, res) {
   try {
     const workspace = await _getWorkspaceAsync(workspaceId);
     const container = await _getDockerContainerByLaunchUuid(workspace.launch_uuid);
-    const logs = await container.logs({ stdout: true, stderr: true, timestamps: true });
+    const logs = await container.logs({
+      stdout: true,
+      stderr: true,
+      timestamps: true,
+      // This should give us a reasonable bound on the worst-case performance.
+      tail: 50000,
```
mwest1066, line 965: Good catch! I need to get in the habit of always adding limits to calls like this. Basically any call that's going to process data, tbh.

workspace_host/lib/docker.js
```diff
@@ -34,7 +34,7 @@ function parseDockerLogs(buffer) {
     return out;
   }
 
-  return output;
+  return Buffer.concat(outputChunks);
```
mwest1066, line 37: Classic O(n^2) to O(n) fix. Nice!

mwest1066 (approved):

workspace_host/lib/docker.js
nwalters512: Yep! I almost want ESLint to yell at me if I call `Buffer.concat` in any kind of loop, this is the second or third time this has really bitten me!

Issue #6655, opened 2022-11-22
Add Python-only workflow 


eliotwrobson:
It would be nice if the current CI job outputted a Python code coverage report from the tests that were run, see https://github.com/marketplace/actions/pytest-coverage-commentator.

Since there has been some discussion of uplifting the Python part of the CI job, this would be nice to include as part of those changes (see #6447). It may also be beneficial to have a separate CI job for just Python linting / typecheck / testing, as this means the overall CI job will take less time and those working just on Python aspects of the codebase don't have to wait for everything to run to get feedback.

eliotwrobson:
This has effectively been completed when the CI steps were split up a while back. Feedback loop for development is now much better!

PR #7202, opened 2023-02-27
Remove "notify course staff" from Python autograder error messages

### LINKED ISSUE #6491: Review autograder notify the course staff messaging
In the testing center, we often encounter students seeking help with their question tells them to contact the course staff, i.e. 

```
The grading code was not able to run.
Please notify the course staff and include this entire message,
including the traceback below.
```

When the error is most typically due to bad student input, and the error is solvable by the student. I think the cautiousness around "notify the course staff" was when the autograder was new and maybe the course hadn't thought about all the edge cases that students might throw at it. But in practice, I can't see the system giving advice to "ask your TA" every time they code something badly.

Could we change that language in our graders to something more descriptive and immediately helpful? Like:

```
Something unexpected happened. Please read and consider the error message below 
that happened when the autograder tried to run your code.
```

Or if there is helpful feedback that channels back to the course staff from these kinds of messages, perhaps we can toggle the language to do different things when `"mode": "Exam"` that's more appropriate for how to get help in different contexts?

mwest1066:
Hopefully this PR can help reduce @trombonekenny's blood pressure 😄 (and also help the CBTF proctors!)

While the original error message was well-intentioned in suggesting that the failure _could_ be due to an error in the autograder code, in virtually all cases that we've ever seen the failures are caused by the student code and the suggestion to "notify course staff" just confuses the student (and proctors) and prevents the student from debugging their code.

I don't quite understand the intention behind checking `test_id.startswith("test")` so it might be better to fix this somehow rather than just removing it. @nicknytko please feel free to tell me what should be done here.

Fixes #6491


graders/python/python_autograder/pl_result.py
```diff
@@ -13,20 +13,8 @@ class PLTestResult(unittest.TestResult):
...
-        "The error traceback is below:\n"
-    )
-
-    major_error_message = (
-        "The grading code was not able to run.\n"
-        "Please notify the course staff "
-        "and include this entire message,\n"
-        "including the traceback below.\n\n"
-        "The error traceback is:\n"
+        "Thre was an error while grading your code.\n\n"
```
jonatanschroeder, line 16:
```suggestion
        "There was an error while grading your code.\n\n"
```

graders/python/python_autograder/pl_result.py
```diff
@@ -13,20 +13,8 @@ class PLTestResult(unittest.TestResult):
...
+        "Look at the traceback below to help debug your code:\n"
```
jonatanschroeder, line 17: I'd suggest add a note encouraging students to "review the question text to ensure your code matches the expected requirements, such as variable names, function names and parameters" or something in this regard.

trombonekenny: Yeah, I like language like that. Softens it a little, even if they don't know how to read the error. "Incorrect syntax or not matching what the grader is expecting might cause these errors. Review the traceback below to help debug your code." Could add a "Use the Report an Issue with this question button on the right if you think the grader is malfunctioning"

eliotwrobson:
Maybe piggybacking a bit, but the CI job isn't linting the files changed here because the `graders` folder isn't included in the `.dockerignore` (if I remember discussion with @jonatanschroeder correctly). Could this be added as part of this PR? Would be nice to have that check in place if the grader is being modified.

nicknytko:
I pushed code that will now correctly make the distinction between errors thrown in a test case, or arbitrary errors (like in the setup code, for example).  I like this new error message for errors in the test cases because more often than not it _is_ because of something weird in the student code.  I'm wondering if we should have a separate error message for the setup code: if that's broken I don't think there is much the student can do with their code there.

eliotwrobson:
> I don't think adding the `graders` to the docker image is desirable, so such a change would require some extra consideration as to how to perform this kind of check.

Ahh that makes sense, since it's actually adding files to the image and not just running tooling over them. Then it might make sense to have some of this tooling run via a separate workflow. I mentioned some of this on a discussion here: https://github.com/PrairieLearn/PrairieLearn/issues/6655 (sorry for taking things in this PR a bit off track).

graders/python/python_autograder/pl_result.py
mwest1066: Added in da7c1105428e8f0a91b4e27395af67615a83e9fd. I opted to not include the "Report an issue" text because I fear that it will cause the same problem as we currently have with the proctors, but for course staff 😄

mwest1066:
> I pushed code that will now correctly make the distinction between errors thrown in a test case, or arbitrary errors (like in the setup code, for example). I like this new error message for errors in the test cases because more often than not it _is_ because of something weird in the student code. I'm wondering if we should have a separate error message for the setup code: if that's broken I don't think there is much the student can do with their code there.

Thanks!

I agree that there isn't much the student can do if there are errors in the setup code. I'm on board to give a different error if we are _really_ confident that the error is in question code and can't possibly have been caused by student code. Do you think this is possible? If there is any doubt then I'd rather just keep the generic error message -- there will always be some good students who can understand that their code is not the problem and report it to course staff. Even if not, if everyone fails a certain question then the course staff will eventually realize 😄

graders/python/python_autograder/pl_result.py
```diff
@@ -13,20 +13,9 @@ class PLTestResult(unittest.TestResult):
...
-    )
-
-    major_error_message = (
-        "The grading code was not able to run.\n"
-        "Please notify the course staff "
-        "and include this entire message,\n"
-        "including the traceback below.\n\n"
-        "The error traceback is:\n"
+        "There was an error while grading your code.\n\n"
+        "Review the question text to ensure your code matches the expected requirements, such as variable names, function names, and parameters.\n\n"
```
jonatanschroeder, line 17: I believe this line may be too long and may cause the output box to scroll unnecessarily (it doesn't wrap).

mwest1066: Good call. Wrapped in af83690ca472d0d7d9fd78d59b9b6cd26806e41a

nicknytko:
> > I pushed code that will now correctly make the distinction between errors thrown in a test case, or arbitrary errors (like in the setup code, for example). I like this new error message for errors in the test cases because more often than not it _is_ because of something weird in the student code. I'm wondering if we should have a separate error message for the setup code: if that's broken I don't think there is much the student can do with their code there.
> 
> Thanks!
> 
> I agree that there isn't much the student can do if there are errors in the setup code. I'm on board to give a different error if we are _really_ confident that the error is in question code and can't possibly have been caused by student code. Do you think this is possible? If there is any doubt then I'd rather just keep the generic error message -- there will always be some good students who can understand that their code is not the problem and report it to course staff. Even if not, if everyone fails a certain question then the course staff will eventually realize 😄

I feel reasonably confident that these ErrorHolder objects are used only outside of test case runs (https://github.com/python/cpython/blob/b5ff38243355c06d665ba8245d461a0d82504581/Lib/unittest/suite.py#L328)

mwest1066:
> I feel reasonably confident that these ErrorHolder objects are used only outside of test case runs (https://github.com/python/cpython/blob/b5ff38243355c06d665ba8245d461a0d82504581/Lib/unittest/suite.py#L328)

Ok, sounds good.

How about we treat this case differently by _not_ telling the student to check their code, but also _not_ telling them to contact course staff. That is, we'll just neutrally report it as an error. I pushed some sample text in 5a918ed8a813eac6a8bb08b251c3f8f5026381d5

My thinking here is that we won't be misleading anyone but we also won't be promoting any particular course of action. For anyone in the know, the phrase "internal error" is a way to differentiate this type of error.

Thoughts?

trombonekenny:
> > I feel reasonably confident that these ErrorHolder objects are used only outside of test case runs (https://github.com/python/cpython/blob/b5ff38243355c06d665ba8245d461a0d82504581/Lib/unittest/suite.py#L328)
> 
> Ok, sounds good.
> 
> How about we treat this case differently by _not_ telling the student to check their code, but also _not_ telling them to contact course staff. That is, we'll just neutrally report it as an error. I pushed some sample text in [5a918ed](https://github.com/PrairieLearn/PrairieLearn/commit/5a918ed8a813eac6a8bb08b251c3f8f5026381d5)
> 
> My thinking here is that we won't be misleading anyone but we also won't be promoting any particular course of action. For anyone in the know, the phrase "internal error" is a way to differentiate this type of error.

I think that's an edge case where not promoting any particular course of action makes sense, but I would be more direct with the error wording - I don't know what the obfuscation gets us. "grader error" or even "external grader error" might be enough to lead course staff in the right direction. Internal error sounds like they're going to blame us on Slack. :P

nicknytko:
> > > I feel reasonably confident that these ErrorHolder objects are used only outside of test case runs (https://github.com/python/cpython/blob/b5ff38243355c06d665ba8245d461a0d82504581/Lib/unittest/suite.py#L328)
> > 
> > 
> > Ok, sounds good.
> > How about we treat this case differently by _not_ telling the student to check their code, but also _not_ telling them to contact course staff. That is, we'll just neutrally report it as an error. I pushed some sample text in [5a918ed](https://github.com/PrairieLearn/PrairieLearn/commit/5a918ed8a813eac6a8bb08b251c3f8f5026381d5)
> > My thinking here is that we won't be misleading anyone but we also won't be promoting any particular course of action. For anyone in the know, the phrase "internal error" is a way to differentiate this type of error.
> 
> I think that's an edge case where not promoting any particular course of action makes sense, but I would be more direct with the error wording - I don't know what the obfuscation gets us. "grader error" or even "external grader error" might be enough to lead course staff in the right direction. Internal error sounds like they're going to blame us on Slack. :P

➕ for subtly pushing the blame onto course staff

mwest1066:
> > I think that's an edge case where not promoting any particular course of action makes sense, but I would be more direct with the error wording - I don't know what the obfuscation gets us. "grader error" or even "external grader error" might be enough to lead course staff in the right direction. Internal error sounds like they're going to blame us on Slack. :P
> 
> ➕ for subtly pushing the blame onto course staff

Ok, I went with `grader error` in eb9ce044c40015356da9e32238af164959d6546a (I think that `external grader error` is a bit too much PL-specific language for something that students will read).

graders/python/python_autograder/pl_result.py
```diff
@@ -13,20 +13,13 @@ class PLTestResult(unittest.TestResult):
...
     )
-
-    major_error_message = (
-        "The grading code was not able to run.\n"
-        "Please notify the course staff "
-        "and include this entire message,\n"
-        "including the traceback below.\n\n"
-        "The error traceback is:\n"
+    grader_error_message = (
+        "There was a grader error while grading your code.\n\n"
```
nwalters512, line 21: Continuing to bikeshed, this reads slightly better to me?

```suggestion
        "The grader encountered an error while grading your code.\n\n"
```

nicknytko (approved):
lgtm (pending Nathan's edit to the text)

PR #7205, opened 2023-02-27
Add timeout to workspace host health check

### LINKED ISSUE #7200: Add timeout to workspace host health check
We check for workspace host health by pinging the `/status` endpoint on each host:

https://github.com/PrairieLearn/PrairieLearn/blob/ce0d20ba002b2e970554e5979e4223607b9a5201/cron/workspaceHostTransitions.js#L109

However, this doesn't include a timeout, which means that it will hang indefinitely if a host is a particular kind of unhealthy (for instance, the event loop is stuck doing some really long synchronous task). If that's the case, we should probably just mark the host as unhealthy instead of waiting indefinitely for a response that may never come. We can achieve that by adding a timeout to this request.

nwalters512:
Closes #7200.

### DIFF
--- a/cron/workspaceHostTransitions.js
+++ b/cron/workspaceHostTransitions.js
@@ -106,7 +106,7 @@ async function checkHealth() {
       healthy = false;
     } else {
       try {
-        await request({ uri: url, resolveWithFullResponse: true });
+        await request({ uri: url, timeout: 30_000 });
       } catch (err) {
         healthy = false;
         logger.info(`Could not reach host ${host.hostname}: ${err}`);
--- a/cron/workspaceHostTransitions.js
+++ b/cron/workspaceHostTransitions.js
@@ -1,10 +1,11 @@
+// @ts-check
 const AWS = require('aws-sdk');
-const { callbackify } = require('util');
-
-const { logger } = require('@prairielearn/logger');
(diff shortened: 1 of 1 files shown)


cron/workspaceHostTransitions.js
```diff
@@ -106,7 +107,7 @@ async function checkHealth() {
       healthy = false;
     } else {
       try {
-        await request({ uri: url, resolveWithFullResponse: true });
+        await fetch(url, { signal: AbortSignal.timeout(30_000) });
```
mwest1066, line 110: TIL about `AbortControllers` -- neat!

mwest1066 (approved):

cron/workspaceHostTransitions.js
mwest1066: Just to make sure I'm understanding, this is using `fetch()` from `node-fetch`, not the built-in `fetch()` (currently experimental?), but we are using the built-in `AbortSignal` (added in 16.14, I think)? I assume this all plays together nicely?

nwalters512: Yeah, `node-fetch` has supported `signal` for a long time, and we're running Node 16.19, so we should be good!

PR #7206, opened 2023-02-27
Remove request-promise-native

nwalters512:
This PR replaces `request-promise-native` with `node-fetch`. `request` and all derivative libraries haven't been supported for years now.


cron/workspaceHostTransitions.js
```diff
@@ -106,10 +106,11 @@ async function checkHealth() {
       healthy = false;
     } else {
       try {
-        await request({ uri: url, resolveWithFullResponse: true });
+        const res = await fetch(url);
```
mwest1066, line 109: Does this conflict with #7205?

mwest1066 (approved):

cron/workspaceHostTransitions.js
nwalters512: Yeah, possibly, I'm going to merge that one first and then update here.

PR #7207, opened 2023-02-27
Split out more code into packages

nwalters512:
This PR introduces 2 new packages:

- `@prairielearn/workspace-utils`: Shared code for update workspace state/message and getting the graded files for a particular workspace directory.
- `@prairielearn/path-utils`: Currently just includes the `contains()` function. I had to split this into a separate package so that it could be used in `@prairielearn/workspace-utils`.


workspace_host/interface.js
```diff
@@ -550,7 +540,7 @@ async function _getWorkspaceSettingsAsync(workspace_id) {
 }
 
 function _pullImage(workspace, callback) {
-  workspaceHelper.updateMessage(workspace.id, 'Checking image');
+  workspaceUtils.updateWorkspaceMessage(workspace.id, 'Checking image');
```
mwest1066, line 543: I think it was probably an error to not be awaiting these calls. In practice it likely works anyway and the only real harm could be out-of-order message updates. For correctness I think it's better to await them, however.

There are a lot more of these in this file but I only marked this one to reduce commenting noise.

```suggestion
  await workspaceUtils.updateWorkspaceMessage(workspace.id, 'Checking image');
```

mwest1066 (approved):

workspace_host/interface.js
nwalters512: Good catch. Hopefully it's not too bad to convert everything upstream of here to `async`!

workspace_host/interface.js
```diff
@@ -909,25 +903,19 @@ async function sendGradedFilesArchive(workspace_id, res) {
...
-  }
-  if (_.sumBy(gradedFiles, (file) => file.stats.size) > config.workspaceMaxGradedFilesSize) {
-    return res.status(500).send({
-      message: `Workspace files exceed limit of ${filesize(config.workspaceMaxGradedFilesSize, {
-        base: 2,
-      })}.`,
-    });
+  let gradedFiles;
+  try {
+    gradedFiles = workspaceUtils.getWorkspaceGradedFiles(
```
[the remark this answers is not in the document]
nwalters512: Done in 81b9807.

workspace_host/interface.js
nwalters512: Done in 6a3f608 and 3ab471e. I touched some hairy callback code, so I'd appreciate a close look at that!

workspace_host/interface.js
```diff
@@ -539,116 +539,125 @@ async function _getWorkspaceSettingsAsync(workspace_id) {
...
+        }
+
+        const toDatabase = false;
+        workspaceUtils
+          .updateWorkspaceMessage(workspace.id, `Pulling image (100%)`, toDatabase)
+          .catch((err) => {
+            logger.error('Error updating workspace message', err);
+            Sentry.captureException(err);
+          })
+          .then(resolve);
```
mwest1066, line 608: Should this be `.finally()`?

mwest1066 (approved):

workspace_host/interface.js
nwalters512: I don't believe so, as the previous line is a `catch` - no errors should make its way past that. Does that sound right?

mwest1066: Agreed. I don't know what I was thinking there, sorry!

mwest1066 (approved):

workspace_host/interface.js
mwest1066: These look good to me (read the code, did not test).

PR #7210, opened 2023-02-28
Don't destructure require of grader host logger

nwalters512:
#7208 was still not quite right. We need TypeScript so bad here!


mwest1066:
> #7208 was still not quite right. We need TypeScript so bad here!

Yes!!!!!!!!

PR #7272, opened 2023-03-01
Join to course based on variant.course_id, not question.course_id

SethPoulsen:
This refactor is in preparation for question sharing, when the `course_id` that is associated with a question is not necessarily the `course_id` that should be associated with a generated variant of that question, since a variant of the question could be created in any course that the question is shared with.

Must be deployed with or after #7283


grader_host/tools/syncECR.js
```diff
@@ -88,7 +88,7 @@ async.series(
       var sql = `
         SELECT DISTINCT external_grading_image
         FROM questions AS q
-        JOIN pl_courses AS plc ON (q.course_id = plc.id)
+        JOIN pl_courses AS plc ON (q.course_id = plc.id) -- does it matter if I change this?
```
nwalters512, line 91: This whole file has since been deleted; this should just go away when you rebase.

lib/chunks.sql
```diff
@@ -77,7 +77,7 @@ WITH RECURSIVE
     SELECT
       tq.id,
       tq.qid,
-      tq.course_id,
+      tq.course_id, -- TODO: should be left the same, right?
```
nwalters512, line 80: Correct - this is getting data (chunks) tied to the question itself.

lib/editors.sql
```diff
@@ -36,5 +36,5 @@ SELECT
 FROM
   questions AS q
 WHERE
-  q.course_id = $course_id
+  q.course_id = $course_id -- should be left the same?
```
nwalters512, line 39: Yep. This is for file editing, which will always be within the question's course.

lib/externalGrader.sql
```diff
@@ -10,7 +10,7 @@ FROM
   LEFT JOIN submissions AS s ON (s.id = gj.submission_id)
   LEFT JOIN variants AS v ON (v.id = s.variant_id)
   LEFT JOIN questions AS q ON (q.id = v.question_id)
-  LEFT JOIN pl_courses AS c ON (c.id = q.course_id)
+  LEFT JOIN pl_courses AS c ON (c.id = v.course_id)
```
nwalters512, line 13: LGTM. I followed the course out of this query, and it's used to copy over `serverFilesCourse` into the grading job, which is what we want.

lib/file-paths.sql
```diff
@@ -4,6 +4,6 @@ SELECT
 FROM
   questions AS q
 WHERE
-  q.course_id = $course_id
+  q.course_id = $course_id -- leave the same?
```
nwalters512, line 7: Yep. This is used to read files for a specific question, which may include "template" questions in the same course.

lib/question.sql
```diff
@@ -85,7 +85,7 @@ FROM
   LEFT JOIN assessments AS a ON (a.id = ai.assessment_id)
   LEFT JOIN course_instances AS ci ON (ci.id = a.course_instance_id)
   JOIN questions AS q ON (q.id = v.question_id)
-  JOIN pl_courses AS c ON (c.id = q.course_id)
+  JOIN pl_courses AS c ON (c.id = v.course_id)
```
nwalters512, line 88: LGTM. AFAICT, this is only used to get the course's display timezone.

lib/workspace.sql
```diff
@@ -34,7 +34,7 @@ FROM
   workspaces AS w
   JOIN variants AS v ON (v.workspace_id = w.id)
   JOIN questions AS q ON (q.id = v.question_id)
-  JOIN pl_courses AS c ON (c.id = q.course_id)
+  JOIN pl_courses AS c ON (c.id = v.course_id)
```
nwalters512, line 37: I actually don't think we want this change: the `course` returned from this query is used to fetch chunks and read workspace files, which will always be tied to the question itself.

middlewares/authzWorkspace.sql
```diff
@@ -3,7 +3,7 @@ SELECT
   v.question_id,
   v.instance_question_id,
   v.course_instance_id,
-  q.course_id,
+  v.course_id,
```
nwalters512, line 6: I believe this is correct, as we want to authorize based on the consuming course in the case of a shared question.

middlewares/selectAndAuthzInstructorQuestion.sql
```diff
@@ -21,7 +21,9 @@ FROM
   issue_count
 WHERE
   q.id = $question_id
-  AND q.course_id = $course_id
+  AND q.course_id = $course_id -- I'm not really sure what to do here. This authorization is used for TONS of stuff.
```
nwalters512, line 24: Agreed. I think we keep this for now and add an `OR (however we determine question sharing access)` when we implement sharing.

middlewares/selectAndAuthzInstructorQuestion.sql
```diff
@@ -50,5 +52,5 @@ FROM
 WHERE
   q.id = $question_id
   AND ci.id = $course_instance_id
-  AND q.course_id = ci.course_id
+  AND q.course_id = ci.course_id -- if the course_instance is included in the URL, this one is used. Otherwise, the one above is used.
```
nwalters512, line 55: I think we'll want the same `OR (...)` eventually here too?

pages/instructorFileTransfer/instructorFileTransfer.sql
```diff
@@ -14,7 +14,8 @@ FROM
   questions AS q
 WHERE
   q.uuid = $uuid
-  AND q.course_id = $course_id
+  AND q.course_id = $course_id -- leave this the same, right? In the long run, we want to allow people to make a copy
```
nwalters512, line 17: Correct, this should stay the same for now.

pages/instructorGradingJob/instructorGradingJob.sql
```diff
@@ -49,7 +49,7 @@ FROM
   LEFT JOIN assessment_questions AS aq ON (aq.id = iq.assessment_question_id)
   LEFT JOIN assessment_instances AS ai ON (ai.id = iq.assessment_instance_id)
   LEFT JOIN assessments AS a ON (a.id = ai.assessment_id)
-  LEFT JOIN pl_courses AS c ON (c.id = q.course_id)
+  LEFT JOIN pl_courses AS c ON (c.id = v.course_id)
```
nwalters512, line 52: LGTM, only used for the display timezone AFAICT.

pages/instructorQuestions/instructorQuestions.sql
```diff
@@ -34,7 +34,7 @@ FROM
   JOIN topics AS top ON (top.id = q.topic_id)
   LEFT JOIN issue_count ON (issue_count.question_id = q.id)
 WHERE
-  q.course_id = $course_id
+  q.course_id = $course_id -- leave the same for now? eventually we will have a way to view shared questions but not now
```
nwalters512, line 37: SGTM.

sprocs/issues_insert_for_variant.sql
```diff
@@ -25,7 +25,7 @@ BEGIN
     FROM
         variants AS v
         JOIN questions AS q ON (q.id = v.question_id)
-        JOIN pl_courses AS c ON (c.id = q.course_id)
+        JOIN pl_courses AS c ON (c.id = v.course_id)
```
nwalters512, line 28: LGTM - we want the issue to be associated with the variant's course, not the question's course.

sprocs/variants_select_for_assessment_instance_grading.sql
```diff
@@ -18,7 +18,7 @@ BEGIN
         JOIN instance_questions AS iq ON (iq.id = v.instance_question_id)
         JOIN assessment_instances AS ai ON (ai.id = iq.assessment_instance_id)
         JOIN questions AS q ON (q.id = v.question_id)
-        JOIN pl_courses AS c ON (c.id = q.course_id)
+        JOIN pl_courses AS c ON (c.id = v.course_id)
```
nwalters512, line 21: Ooooh, this is a tricky one. We need `course` to correspond to the question's course so that we know where to find it on disk, but we also add the entire `course` object to any reported issues. I think we might need to return both `question_course` and `variant_course` here, and pipe both of them around?

tools/paths-lookup.sql
```diff
@@ -7,6 +7,6 @@ from
   instance_questions iq
   join assessment_questions aq on (aq.id = iq.assessment_question_id)
   join questions q on (q.id = aq.question_id)
-  join pl_courses plc on (plc.id = q.course_id)
+  join pl_courses plc on (plc.id = q.course_id) -- leave it? I think this script is interested in the host course of the question
```
nwalters512, line 10: This script was removed in https://github.com/PrairieLearn/PrairieLearn/pull/7521, it should go away once you get up to date with master.

workspace_host/interface.sql
```diff
@@ -7,7 +7,7 @@ FROM
   workspaces AS w
   JOIN variants AS v ON (v.workspace_id = w.id)
   JOIN questions AS q ON (q.id = v.question_id)
-  JOIN pl_courses AS c ON (c.id = q.course_id)
+  JOIN pl_courses AS c ON (c.id = v.course_id)
```
nwalters512, line 10: I think this is good too. AFAICT this is only used to get a course ID that we can label S3 objects with, and we do want them to be associated with the variant's course.

middlewares/selectAndAuthzInstructorQuestion.sql
SethPoulsen: I will add an explicit TODO comment here that can be merged here, then changed in the main question sharing PR.

pages/instructorFileTransfer/instructorFileTransfer.sql
SethPoulsen: Should I leave a TODO comment here? It seems like a really long off feature at this point given everything else we are trying to do.

pages/instructorQuestions/instructorQuestions.sql
SethPoulsen: Same question, should I leave a TODO comment here in the codebase or no?

pages/instructorFileTransfer/instructorFileTransfer.sql
SethPoulsen: I would lean toward not bothering. When we get to implementing it, it should be apparent what needs to change.

SethPoulsen: On the other hand, maybe it would help whoever is doing the implementing avoid some weird bugs. I don't know.

lib/workspace.sql
SethPoulsen: ok, reverted. 

What's going to be really fun about merging this PR is that if we mess up, some of those mistakes would show up immediately...but if we make a mistake like this one, then we won't know there is a problem until the question sharing PR is merged and we start having the situation where q.course_id and v.course_id are actually different.

SethPoulsen: I double checked and I agree, the `course` object returned by this query is only used to fetch the question/workspace content.

sprocs/variants_select_for_assessment_instance_grading.sql
SethPoulsen: Nice catch.

SethPoulsen: See edits.

SethPoulsen:
In the process of the refactor of `question.js` in this PR, I made some mistakes that should have been caught by typechecking, only to realize that typechecking had not been enabled for the file. For my sanity, I enabled it, but had to make a few small unrelated changes in order to do so.

apps/prairielearn/src/lib/question.js
```diff
@@ -1862,6 +1918,8 @@ module.exports = {
         )
       );
 
+      /** @type function */
+      let renderFileAsync = util.promisify(ejs.renderFile);
```
SethPoulsen, line 1922: For some reason, type checking on this expression was failing without this explicit declaration. 

see https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33340

apps/prairielearn/src/lib/question.js
```diff
@@ -498,15 +511,31 @@ module.exports = {
...
+              sql.select_question_course,
+              { question_course_id: question.course_id },
+              (err, result) => {
+                if (ERR(err, callback)) return;
+                question_course = result.rows[0].course;
+                callback(null);
+              }
+            );
+          }
+        },
```
nwalters512, line 538: Let's make this an `async` function.

apps/prairielearn/src/lib/question.js
```diff
@@ -1035,19 +1064,44 @@ module.exports = {
...
```
nwalters512, line 1104: Ditto re: `async` function.

apps/prairielearn/src/lib/question.js
```diff
@@ -1522,6 +1577,22 @@ module.exports = {
...
+              sql.select_question_course,
+              { question_course_id: locals.question.course_id },
+              (err, result) => {
+                if (ERR(err, callback)) return;
+                locals.question_course = result.rows[0].course;
+                callback(null);
+              }
+            );
+          }
+        },
```
nwalters512, line 1595: Ditto re: `async` function. Having seen this three times, maybe this should be a little reusable function?

nwalters512 (approved):
LGTM! @mwest1066 I'd appreciate a review from you too given that these changes touch such core functionality.

mwest1066 (approved):
I checked the `variant_course` versus `question_course` distinction as well as I could and they all look correct to me.

apps/prairielearn/src/lib/question.js
SethPoulsen: I thought about it. This is annoying to do because if I make it an async function because the calling function is not async, so I would need to go make a bunch of things async, and I don't want that sort of refactor to be part of this. 

If I make it an old school asynchronous function with a callback, then calling such a function is not pretty, but I guess that's still better than this because the code is only in one place. I can go ahead and do that.

SethPoulsen: done

nwalters512: It's actually pretty straightforward to mix async + callback functions when using `async.series` (or really any `async.*`) function: just make the individual function async and the rest will work as normal. We use this all over the codebase to ease the transition to async functions. Here's an example from `grader_host`:

https://github.com/PrairieLearn/PrairieLearn/blob/0d62b3d6c622a1886d1f9985a119d206f0538ba2/apps/grader-host/src/index.js#L40-L58

SethPoulsen: done

PR #7273, opened 2023-03-01
Add course_id to variants table

### LINKED ISSUE #7176: Add course_id column to variants table
Variants are currently associated with "parent" entities via the following columns:

- `question_id`
- `course_instance_id`
- `instance_question_id`

However, this poses an issue in a world where question sharing exists. If an instructor creates a variant from the course question preview page (e.g. http://localhost:3000/pl/course/2/question/446/preview), only `question_id` will be populated, and that's not enough information for us to know whether information about a given variant (for instance, issues created from it) should be visible in the course from which a question is shared.

If we ensure that variants are always associated with a course, this can provide pretty foolproof logical isolation for variants generated from shared questions.

cc @SethPoulsen this is related to our conversation yesterday.

SethPoulsen:
Resolves #7176

In preparation for question sharing, when a variant of a question may be created under a different course_id than the course_id associated with the question.


migrations/20240210165254_variants__reference_course_id.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE variants
+ADD COLUMN IF NOT EXISTS course_id bigint;
+
+ALTER TABLE variants
+ADD CONSTRAINT variants_course_id_fkey FOREIGN KEY (course_id) REFERENCES pl_courses (id) ON UPDATE CASCADE ON DELETE CASCADE;
+
+UPDATE variants AS v
+SET course_id = q.course_id
+FROM questions AS q
+WHERE v.question_id = q.id;
```
SethPoulsen, line 10: @mwest1066 is there any way to do this that isn't going to completely lock the variants table and thus take down PL? Or is there no way around it and we're just going to have to deploy during a low/no usage time?

migrations/20240210165254_variants__reference_course_id.sql
```diff
@@ -0,0 +1,13 @@
...
+ALTER TABLE variants
+ADD CONSTRAINT variants_course_id_fkey FOREIGN KEY (course_id) REFERENCES pl_courses (id) ON UPDATE CASCADE ON DELETE CASCADE;
+
+UPDATE variants AS v
+SET
+  course_id = q.course_id
+FROM
+  questions AS q
+WHERE
+  v.question_id = q.id;
```
SethPoulsen, line 13: @mwest1066 is there any way to do this that isn't going to completely lock the variants table and thus take down PL? Or is there no way around it and we're just going to have to deploy during a low/no usage time?

nwalters512: I'll either need to manually run this in chunks, or (much preferred by me) hold off on merging this until I can implement something like this: https://docs.gitlab.com/ee/development/database/batched_background_migrations.html.

We'll want to add a `NOT NULL` constraint once this has been deployed. We can use the method described here to add it without locking the entire table: https://dba.stackexchange.com/a/268128

SethPoulsen: I'm happy to make another PR to add the `NOT NULL` constraint, to be deployed after this one.

sprocs/variants_insert.sql
```diff
@@ -115,11 +115,11 @@ BEGIN
     INSERT INTO variants
         (instance_question_id, question_id,      course_instance_id, user_id, group_id,
         number,     variant_seed, params, true_answer, options, broken, authn_user_id,
-        workspace_id)
+        workspace_id, course_id)
     VALUES
         (instance_question_id, real_question_id, real_course_instance_id, real_user_id, real_group_id,
         new_number, variant_seed, params, true_answer, options, broken, authn_user_id,
-        workspace_id)
+        workspace_id, course_id)
```
SethPoulsen, line 122: I've realized there are some cases where `course_id` could be null here. I'm working on fully understanding this sproc so that I can make sure it ends up non-null in all cases.

SethPoulsen: Should be fixed now.

migrations/20240210165254_variants__reference_course_id.sql
nwalters512: I've started working on infrastructure for automatically executing batched migrations: #7292. It borrows the basic idea of GitLab's system (a function that is repeatedly invoked with different ranges of a table's ID space), but without quite as much of the complexity.

SethPoulsen: Cool! Feel free to push directly to this PR branch if you want to use it for testing out that infra. Or I'm happy to work on it with you. just let me know.

mwest1066: > We'll want to add a `NOT NULL` constraint once this has been deployed. We can use the method described here to add it without locking the entire table: https://dba.stackexchange.com/a/268128

Oh wow, I hadn't known that PG 12 added this way to add a NOT NULL without a full table lock and scan. So awesome!

It would be great to use this to add a lot more NOT NULL constraints to our tables. We have many places (especially for foreign keys) where we are assuming that columns are not null but we don't actually have a NOT NULL constraint in place. I was better about this for PT but in the original PL schema I hadn't appreciated how useful it is to to add NOT NULL. This will be especially helpful when we add zod types corresponding to the DB schema.

SethPoulsen: Ya, it looks pretty awesome. Open some issues for the other ones you want to do?

mwest1066: https://github.com/PrairieLearn/PrairieLearn/issues/7320

database/tables/variants.pg
```diff
@@ -1,6 +1,7 @@
 columns
     authn_user_id: bigint
     broken: boolean default false
+    course_id: bigint
```
mwest1066, line 4: Add a NOT NULL constraint.

migrations/20240210165254_variants__reference_course_id.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE variants
+ADD COLUMN IF NOT EXISTS course_id bigint;
+
+ALTER TABLE variants
+ADD CONSTRAINT variants_course_id_fkey FOREIGN KEY (course_id) REFERENCES pl_courses (id) ON UPDATE CASCADE ON DELETE CASCADE;
```
mwest1066, line 5: Not that it really matters, but you can add this directly in the ADD COLUMN query (and also add NOT NULL).
```suggestion
ALTER TABLE variants
ADD COLUMN IF NOT EXISTS course_id bigint NOT NULL REFERENCES pl_courses (id) ON UPDATE CASCADE ON DELETE CASCADE;
```

database/tables/variants.pg
nwalters512: We can't do that yet because this column won't be populated yet; that's the whole point of the batched migration machinery I've been working on!

mwest1066: Ugh, right, thanks for the reminder yet again! 🤦

SethPoulsen: to be fair, this PR still has the code to populate the column, so it was easy to be confused (I'll be ripping that out today).

migrations/20240210165254_variants__reference_course_id.sql
SethPoulsen: Oh, that's cool! When you do it that way, does it automatically generate the constraint name for you?

nwalters512 (approved):
Sweet! I won't block on anything, but I'd apply Matt's suggestion about adding the foreign key constraint, and let's talk about the value (or lack thereof?) of checking for `course_instance_id` consistency.

sprocs/variants_insert.sql
```diff
@@ -83,20 +83,17 @@ BEGIN
...
-            pl_courses AS c
-            JOIN course_instances AS ci ON (ci.course_id = c.id)
-            JOIN questions AS q ON (q.course_id = c.id)
-        WHERE
-            ci.id = real_course_instance_id
-            AND q.id = real_question_id;
-
-        IF course_id IS NULL THEN RAISE EXCEPTION 'inconsistent course for question_id and course_instance_id'; END IF;
-    END IF;
+    -- check consistency of question_id and course_id
```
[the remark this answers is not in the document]
SethPoulsen: I may have had a good reason why I thought this was fine when I did it, but it was a month ago now and I honestly can't remember. Let me just try adding a simple check back in and see how it goes.

SethPoulsen:
@nwalters512 should I merge?

SethPoulsen:
Wait a minute, they way I rewrote the SQL is going to cause an issue if you create a variant in a course that has no course instances. Let me tweak it.

sprocs/variants_insert.sql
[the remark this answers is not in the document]
SethPoulsen: Done. What do you think?

sprocs/variants_insert.sql
```diff
@@ -83,19 +83,30 @@ BEGIN
...
-    IF real_question_id IS NOT NULL AND real_course_instance_id IS NOT NULL THEN
-        SELECT c.id
-        INTO course_id
+    -- check consistency of question_id and course_id
+    SELECT q.id
+    INTO real_question_id
+    FROM
+        questions AS q
+    WHERE
+        q.id = real_question_id
```
nwalters512, line 92: I am ever-so-slightly weirded out by the same variable being both an input to and output from a query, I did a double take when I saw this. Perhaps it would be slightly clearer to do something like this?

```sql
SELECT FROM questions ...;
IF NOT FOUND THEN RAISE ...;
```

cc @mwest1066 I'm not sure if this is a smell I should be worried about.

SethPoulsen: I agree that looks better, I'll change it.

I'm pretty inexperienced with PL/pgSQL so I didn't think to use that, even though I think I may have seen it before.

mwest1066: Yeah, I saw that and was also slightly unhappy but I didn't flag it. I'd certainly prefer to avoid this double-use of the same variable.

SethPoulsen: huh, what you are suggesting is actually not allowed, I get a SQL error:

> If you want to discard the results of a SELECT, use PERFORM instead

SethPoulsen: @mwest1066 what do recommend the best thing to do here stylistically, given the error I got?

mwest1066: This is a small quirk of plpgsql -- if you don't want to directly save the results of the SELECT then you need to write it as PERFORM instead. We already do this elsewhere. For example, here we do a perform: https://github.com/PrairieLearn/PrairieLearn/blob/236d6813961354faa213e15e050197ba2564b52a/sprocs/ip_to_mode.sql#L25 just so we can test for FOUND: https://github.com/PrairieLearn/PrairieLearn/blob/236d6813961354faa213e15e050197ba2564b52a/sprocs/ip_to_mode.sql#L37

SethPoulsen: Thanks for the code pointer! I couldn't figure out the syntax. 

All fixed now.

mwest1066: lgtm

nwalters512 (approved):

PR #7279, opened 2023-03-01
Only fetch workspace logs from host if running

nwalters512:
Before this change, trying to get logs from a stopped workspace would cause an error:

Screenshot 2023-03-01 at 13 51 41

### DIFF
--- a/pages/workspaceLogs/workspaceLogs.js
+++ b/pages/workspaceLogs/workspaceLogs.js
@@ -83,7 +83,7 @@ async function loadLogsForWorkspaceVersion(workspaceId, version) {
   // If the current workspace version matches the requested version, we can
   // reach out to the workspace host directly to get the remaining logs. Otherwise,
   // they should have been flushed to S3 already.
-  if (workspace.state === 'running' && workspace.is_current_version) {
+  if (workspace.state === 'running' && workspace.is_current_version && workspace.hostname) {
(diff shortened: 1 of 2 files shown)


pages/workspaceLogs/workspaceLogs.js
```diff
@@ -83,7 +83,7 @@ async function loadLogsForWorkspaceVersion(workspaceId, version) {
   // If the current workspace version matches the requested version, we can
   // reach out to the workspace host directly to get the remaining logs. Otherwise,
   // they should have been flushed to S3 already.
-  if (workspace.is_current_version) {
+  if (workspace.state === 'running' && workspace.is_current_version && workspace.hostname) {
```
nwalters512, line 86: The first and last condition here are probably redundant, but it doesn't hurt 🤷

mwest1066 (approved):
Should we also add a fallback to fetch from S3 if the fetch-from-host fails? Should we have a timeout on this fetch? I'm thinking about the case were we start to fetch from a host just as it's being terminated, but this might be enough of an edge case that it's not worth the complexity to handle it. If the host terminates mid-fetch we can always just reload the logs page and then it will get the logs from S3.

nwalters512:
Note that we unconditionally load from S3 (read the first part of this function); we just _might_ also load from the host if its in a state where the logs are there.

I'll add a timeout 👍

Issue #7189, opened 2023-02-22
Add user toggle to reset MathJax browser settings


trombonekenny:
There's some support load for users who misconfigure their MathJax settings (presumably by picking a renderer that doesn't work in their environment from the right-click menu) that disables the context menu that would let them fix it. The typical symptom presents as MathJax not rendering so it shows up as non-clickable `$a = 8$` text in the question content.

We should add a button or toggle, perhaps under the user settings area of PL, that resets MathJax to a default state. For me on Windows 10 Chrome, that's under Local Storage / MathJax-Menu-Settings. Having a button to reset it would be so much easier than guiding people how to reset those settings in the browser.

trombonekenny:
The student I was working with confirmed that removing the mathjax setting made her browser work again.

Client JS to do that could be
`<button onClick='localStorage.removeItem("MathJax-Menu-Settings");'>Reset MathJax local settings</button>`

PR #7280, opened 2023-03-02
Disallow user from changing the MathJax renderer

### LINKED ISSUE #6495: MathJax SVG math renderer broken on Windows Chrome

nwalters512:
This should help resolve #6495. It might mean we don't have to implement #7189 (cc @trombonekenny). If we have a snippet of JS that can clear the user settings, I wouldn't mind adding that here and just making it always automatically run on every page load.

### DIFF
--- a/pages/partials/mathjax.ejs
+++ b/pages/partials/mathjax.ejs
@@ -38,6 +38,11 @@ var MathJax = {
                     MathJax.typesetPromise();
                 }
             };
+
+            // Disallow the user from changing the MathJax renderer. This often
+            // lead to bugs in the past.
+            const renderer = window.MathJax.startup.document.menu.menu.find('Renderer');
+            if (renderer) renderer.disable();
         },
         pageReady: () => {
             MathJax.startup.defaultPageReady();


mwest1066 (approved):
Works for me.

trombonekenny:
I'm torn because I want the rendering to "just work" but I also don't want to take choices away from people. Noodling around on different configs/browsers/platforms can be very helpful for troubleshooting. Behind just making it work, my preference would be to leave MathJax user configs as open as possible, maybe fix up what the caching situation should look like for us, and give them a settings page somewhere where they can reset if needed.

nwalters512:
If you can identify the correct bit of code to reset the preferences, we can add that to the user settings page instead. I couldn't tell exactly what we'd need from the docs.

echuber2:
The way I implemented https://github.com/PrairieLearn/PrairieLearn/pull/6974 , the user can temporarily change the preference, but reloading the page always resets it.

nwalters512:
Closing in favor of #6974.

PR #7281, opened 2023-03-02
Fix Azure callback error handling

nwalters512:
Looks like this was a regression from #7094. The callback to `passport.authenticate` isn't meant to be an async function. If one is passed, errors won't be handled correctly, resulting in a server crash.

### DIFF
--- a/ee/auth/azure/callback.js
+++ b/ee/auth/azure/callback.js
@@ -1,36 +1,40 @@
 // @ts-check
-const ERR = require('async-stacktrace');
 const passport = require('passport');
 const express = require('express');
-const router = express.Router();
 const asyncHandler = require('express-async-handler');
+const util = require('util');
 
 const authnLib = require('../../../lib/authn');
 
+const router = express.Router();
+
 router.post(
   '/',
-  asyncHandler(async (req, res, next) => {
-    const authData = {
-      response: res,
-      failureRedirect: '/pl',
-      session: false,
+  asyncHandler(async (req, res) => {
+    // This will write the resolved user to `req.user`.
+    await util.promisify(
+      passport.authenticate('azuread-openidconnect', {
+        failureRedirect: '/pl',
+        session: false,
+      })
+    )(res, res);
+
+    const user = req.user;
+    if (!user) throw new Error('Login failed');
+
+    const authnParams = {
+      // @ts-expect-error `upn` is not defined on the type.
+      uid: user.upn,
+      // @ts-expect-error `displayName` is not defined on the type.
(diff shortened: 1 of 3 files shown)


ee/auth/azure/callback.js
```diff
@@ -1,36 +1,40 @@
...
-      if (ERR(err, next)) return;
-      if (!user) return next(new Error('Login failed'));
+  asyncHandler(async (req, res) => {
+    // This will write the resolved user to `req.user`.
+    await util.promisify(
+      passport.authenticate('azuread-openidconnect', {
+        failureRedirect: '/pl',
+        session: false,
+      })
+    )(req, res);
```
mwest1066, line 20: Ugh, I just attempted to read the passportjs docs for the first time and it was not a pleasant experience. I'm now inclined to avoid passport as much as possible!

If I'm understanding correctly, `passport.authenticate()` will return an express middleware? That is, a function that accepts arguments `(req, res, next)`? If so, how can we call `promisify()` on this? I thought `promisify` only works on functions with callbacks that use the `callback(err, ...)` pattern, but I don't think the `next` function is a callback in this sense?

trombonekenny: Yeah, I've found their code is better than their docs. (And their approach to 'standardizing' authentication is really hard to wrap my head around, for some reason.)

> Note that if a callback is supplied, it becomes the application's
responsibility to log-in the user, establish a session, and otherwise perform
the desired operations.

Which is fine for our stuff, and shortcuts their standardization.

nwalters512: This is exactly the same pattern we've been using in the SAML auth code:

https://github.com/PrairieLearn/PrairieLearn/blob/6b3b64f27ed21bf9bc9b34c3d109d7f9ce987c3d/ee/auth/saml/router.js#L36-L42

I tested this on staging and I'm able to authenticate. If you're not comfortable with this, we could instead use the callback form and manually wrap it in a Promise?

```ts
await new Promise((resolve, reject) => {
  passport.authenticate('foo', data, (err, user) => {
    if (err) {
      reject(err);
    } else {
      resolve(user);
    }
  });
});
```

nwalters512: In 2aef281 I changed to using the callback form wrapped explicitly with a Promise. It's still quite messy IMO, but the `authenticate()` wrapper better matches how I want to think about auth ("give me a request, I'll give you a user").

trombonekenny (approved):
I didn't run the code, but the refactoring makes sense to me and looks OK.

PR #7283, opened 2023-03-02
Delcare variants.course_id not null

SethPoulsen:
After adding the course_id column to variants and then filling it out, 
we want to enforce that it is never null. 

Must be deployed after #7273


jonatanschroeder:
To avoid confusion I'd suggest converting this to Draft, so that it's clear this is not ready for review.

Also, you set it up to merge into `variants-course-id`, I imagine that, when this is to be merged, it will be merged to master, right?

SethPoulsen:
Sure. And ya, that's right.

I don't actually intend to merge this into `variants-course-id` directly, I just set it that way so you can read the diff properly. This will be merged after that one is merged to master.

nwalters512:
I'm adding the "do not merge" label for now since we'll want to wait for the batched migration in #7513 to be deployed and finished.

SethPoulsen:
After this PR is successfully deployed, we should merge https://github.com/PrairieLearn/PrairieLearn/pull/7272

(technically they could go out in the same deployment? as long as the migration failing means that the deployment doesn't actually happen.)

SethPoulsen:
The only reason this CI build is failing right now is because #7272 isn't yet merged to master yet.

apps/prairielearn/src/migrations/20230523200759_variants__course_id__nonnull.sql
```diff
@@ -0,0 +1,25 @@
+-- declare the new course_id column of the variants table to be not null
+-- using the approach suggested by a postgres dev here: https://dba.stackexchange.com/questions/267947/how-can-i-set-a-column-to-not-null-without-locking-the-table-during-a-table-scan/268128#268128
```
mwest1066, line 2: Awesome find! 🚀

mwest1066 (approved):

mwest1066:
Looks good to me but I'll let @nwalters512 give the final sign-off. He'll coordinate deployment.

apps/prairielearn/src/migrations/20230523200759_variants__course_id__nonnull.sql
SethPoulsen: Credit to @nwalters512

nwalters512 (approved):

Issue #1155, opened 2018-05-10
Files download fails


mwest1066:
As an example, CS 125 final exam file download fails. Running the query by hand takes about 80 s and returns 5 MB of data, both of which should be manageable.

gchallen:
Anecdotally I also think that this query can cause serious problems with ongoing assessments—at minimum within the same course and maybe across the whole PL instance? (No evidence for that.) Evidence: two attempts to run this query both created clustered of "Rendering failed" issues during our CS 125 exam, with timestamps that matched well with the time that the downloads were attempted. I didn't try to reproduce 🙁.

trombonekenny:
gwa: I think the PrairieLearn front end had python concurrency problems due to finals load around the same time these database calls were done. Lacking any more definitive evidence, I can't see how a long database pull would have caused other v3 question rendering problems. (Unless something is sync that shouldn't be and caused other timeouts.)

matt: Do we know if it's the query time or the data returned that's the problem? Any logs on why it fails?

gchallen:
I'm assuming you have the logs to run this down. If it's a coincidence, it's a very very strange one—particularly given that we saw so more similar problems all day despite having heavier exam periods later that Wednesday (IIRC).

mwest1066:
Long-delayed followup to this:

1. There was an issue with large CSV downloads blocking student question code. This was fixed by #1364 

2. The API can be used as a work-around for large CSV downloads. Downloading from the API does not have any size restrictions because it's done object-by-object and the client is in control.

3. We should in any case still fix large CSV downloads to stream results from the DB query, which would allow very large CSVs to be downloaded. This should use https://node-postgres.com/api/cursor

nwalters512:
This was resolved as of #7284. Query results are now streamed from the database using cursors, and the CSVs themselves are constructed via streaming instead of concatenation in memory.

PR #7284, opened 2023-03-02
Use streaming CSV stringification

### LINKED ISSUE #6523: Use streaming CSV generation

nwalters512:
I originally thought that we'd be able to use `require('csv-stringify').stringify` directly, but AFAICT it will still try to write every item synchronously, which will block the event loop. So, I kept our existing `nonblockingStringify` function, but instead of immediately collecting the chunks, it now returns the `Stringifier` instance for the caller to `.pipe(...)` as needed.

I did not update any code to user cursors, If folks still think that would be of high value, I'll split that out into another issue.

Closes #6523; closes #6855.


mwest1066 (dismissed):
This PR is clearly a benefit over the current state of the code and I'm happy to have it merged as is. However:
* At some point we need to pull this out into a separate package (PT also wants to consume this). Maybe this is a good chance to do this? I'm fine to leave this for a separate PR of course.
* I'm not convinced that this is going to solve #1155 without also adding streaming for both the DB query (using cursors) and `rowsToCsv()`. I think most of our CSV generation is fine (things like Gradebook downloads) but the bad large queries come in `instructorAssessmentDownloads.js` from the queries `submissions_for_manual_grading` and `assessment_instance_submissions`. These are the queries that will return submission data as a CSV, which is where our huge CSVs are coming from (afaik). Again, it's fine to leave a fully-streaming implementation to a future PR.

jonatanschroeder:
> * I'm not convinced that this is going to solve #1155 without also adding streaming for both the DB query (using cursors) and `rowsToCsv()`. I think most of our CSV generation is fine (things like Gradebook downloads) but the bad large queries come in `instructorAssessmentDownloads.js` from the queries `submissions_for_manual_grading` and `assessment_instance_submissions`. These are the queries that will return submission data as a CSV, which is where our huge CSVs are coming from (afaik). Again, it's fine to leave a fully-streaming implementation to a future PR.

https://github.com/PrairieLearn/PrairieLearn/pull/6855#issuecomment-1371202288 (and following comments)

pages/instructorQuestionStatistics/instructorQuestionStatistics.js
```diff
@@ -1,138 +1,106 @@
...
-router.get('/:filename', function (req, res, next) {
-  _.assign(res.locals, filenames(res.locals));
+    res.locals.question_attempts_histogram = null;
+    res.locals.question_attempts_before_giving_up_histogram = null;
+    res.locals.question_attempts_histogram_hw = null;
+    res.locals.question_attempts_before_giving_up_histogram_hw = null;
+    // res.locals.question_attempts_histogram = res.locals.result.question_attempts_histogram;
+    // res.locals.question_attempts_before_giving_up_histogram = res.locals.result.question_attempts_before_giving_up_histogram;
+    // res.locals.question_attempts_histogram_hw = res.locals.result.question_attempts_histogram_hw;
+    // res.locals.question_attempts_before_giving_up_histogram_hw = res.locals.result.question_attempts_before_giving_up_histogram_hw;
```
nwalters512, line 34: 👀 any idea what's up with this?

packages/postgres/src/pool.ts
```diff
@@ -924,7 +927,23 @@ export class PostgresPool {
...
+          writableObjectMode: true,
+          transform(chunk, _encoding, callback) {
+            for (const row of chunk) {
+              this.push(row);
+            }
+            callback();
+          },
+        });
+
+        // TODO: use native `node:stream#compose` once it's stable.
```
nwalters512, line 942: It's currently listed as experimental and isn't present in the TypeScript types for Node, so I opted to avoid it for now.

pages/instructorAssessmentDownloads/instructorAssessmentDownloads.js
```diff
@@ -298,139 +291,105 @@ router.get('/:filename', function (req, res, next) {
...
+      const archive = archiver('zip');
+      const dirname = (res.locals.assessment_set.name + res.locals.assessment.number).replace(
+        ' ',
+        ''
+      );
+      const prefix = `${dirname}/`;
+      archive.append(null, { name: prefix });
+      archive.pipe(res);
+
+      for await (const rows of cursor.iterate(100)) {
```
mwest1066, line 314: ❤️

tests/postgres.test.js
mwest1066, line 1: I feel like we discussed this recently, but what's stopping us from moving (most of) this file into the `@prairielearn/postgres` package?

mwest1066 (approved):

packages/postgres/src/pool.ts
```diff
@@ -924,7 +927,23 @@ export class PostgresPool {
...
+        return multipipe(Readable.from(iterator.iterate(batchSize)), transform);
```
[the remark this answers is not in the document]
nwalters512: `pipeline` is best suited for when you have an initial source and a final destination; `multipipe` (or `compose`, which it mirrors) lets you construct a stream that in turn will be piped elsewhere. I think the [`compose` docs](https://nodejs.org/api/stream.html#streamcomposestreams) say this best:

> Because stream.compose returns a new stream that in turn can (and should) be piped into other streams, it enables composition. In contrast, when passing streams to stream.pipeline, typically the first stream is a readable stream and the last a writable stream, forming a closed circuit.

packages/csv/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Streaming transform from an array of objects to a CSV that doesn't
+ * block the event loop.
+ */
+export function stringifyNonblocking(data: any[], options: StringifierOptions = {}): Stringifier {
+  const stringifier = new Stringifier(options);
+
+  process.nextTick(function () {
+    let j = 0;
+    function loop() {
+      for (let i = 0; i < 100; i++) {
```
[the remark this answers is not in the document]
nwalters512: I added a `batchSize` option in a33f506,

tests/postgres.test.js
nwalters512: I moved things around in #7376, so this is now resolved!

pages/instructorQuestionStatistics/instructorQuestionStatistics.js
mwest1066: I guess this is some leftover thing from the student who originally implemented this? I guess it was never implemented, although they do look like interesting things to be computing.

packages/postgres/src/pool.ts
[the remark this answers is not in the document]
mwest1066: Cool, thanks!

pages/instructorQuestionStatistics/instructorQuestionStatistics.js
nwalters512: Gotcha - I'll rip this out in another PR!

PR #7286, opened 2023-03-02
feat: user settings page: add user info, mathjax reset button

### LINKED ISSUE #7189: Add user toggle to reset MathJax browser settings

trombonekenny:
Closes #7189 
Part of the effort to resolve the Mathjax rendering complexities and support, issues, and PRs

Also adds some logged-in user data (like institution and authentication source) readonly to the user settings page. We don't show that information anywhere and it could be helpful for LTI development

### DIFF
--- a/pages/userSettings/userSettings.ejs
+++ b/pages/userSettings/userSettings.ejs
@@ -51,10 +51,14 @@
           Browser configuration
         </div>
         <div class="card-body">
-          <p>This section will let you reset browser settings related to technology inside PrairieLearn.</p>
-           <p>
+          <p>
+            This section will let you reset browser settings related to technology inside PrairieLearn.
+          </p>
+          <p>
             If math formulas shows up as code like <strong>$ x = \frac {-b \pm \sqrt {b^2 - 4ac}}{2a} $</strong>
-            <br>
+            resetting the MathJax menu settings might help.
+          </p>
+          <p>
             <button class="btn btn-md btn-info"
               onClick="localStorage.removeItem('MathJax-Menu-Settings');alert('MathJax menu settings have been reset');">
               Reset MathJax menu settings
--- a/pages/userSettings/userSettings.ejs
+++ b/pages/userSettings/userSettings.ejs
@@ -15,6 +15,59 @@
 
       <div class="card mb-4">
         <div class="card-header bg-primary text-white d-flex align-items-center">
+          Logged-in user profile information
+          <a href="/pl/logout" class="btn btn-danger btn-sm ml-auto">Logout</a>
+        </div>
+          <table class="table table-sm two-column-description">
+            <tbody>
+              <tr>
+                <th>User ID (UID)</th>
+                <td>
+                  <%= authn_user.uid %>
+                </td>
+              </tr>
+              <tr>
+                <th>User Name</th>
+                <td><%= authn_user.name %></td>
+              </tr>
+              <tr>
+                <th>Unique Id Number (UIN)</th>
+                <td><%= authn_user.uin %></td>
+              </tr>
+              <tr>
+                <th>Institution</th>
+                <td><%= authn_institution.long_name %> (<%= authn_institution.short_name %>)</td>
+              </tr>
+              <tr>
+                <th>Authentication method</th>
+                <td>
+                  <%= authn_provider_name %>
+                </td>
+              </tr>
+            </tbody>
+          </table>
+      </div>
+
+      <div class="card mb-4">
+        <div class="card-header bg-primary text-white d-flex">
+          Browser configuration
+        </div>
+        <div class="card-body">
+          <p>These options will let you reset some browser settings related to technology inside PrairieLearn.
+            Use these buttons to reset things if things are broken for you.
+          </p>
+          <p>
+          <button class="btn btn-md btn-info"
+            onClick="localStorage.removeItem('MathJax-Menu-Settings');alert('MathJax menu settings have been reset');">
+            Reset MathJax menu settings
+          </button> Can be helpful if your math shows up as
+          <strong>$ x = \frac {-b \pm \sqrt {b^2 - 4ac}}{2a} $</strong>
+          </p>
+        </div>
+      </div>
+
+      <div class="card mb-4">
+        <div class="card-header bg-primary text-white d-flex align-items-center">
           Personal Access Tokens
           <button id="generateTokenButton"
             type="button" class="btn btn-light btn-sm ml-auto"
--- a/pages/userSettings/userSettings.ejs
+++ b/pages/userSettings/userSettings.ejs
@@ -12,40 +12,38 @@
     <%- include('../partials/navbar', {navPage: 'user_settings'}); %>
     <div id="content" class="container">
       <h1 class="mb-4">Settings</h1>
-
       <div class="card mb-4">
         <div class="card-header bg-primary text-white d-flex align-items-center">
-          Logged-in user profile information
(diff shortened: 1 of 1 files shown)


pages/userSettings/userSettings.ejs
```diff
@@ -13,6 +13,59 @@
     <div id="content" class="container">
       <h1 class="mb-4">Settings</h1>
 
+      <div class="card mb-4">
+        <div class="card-header bg-primary text-white d-flex align-items-center">
+          Logged-in user profile information
+          <a href="/pl/logout" class="btn btn-danger btn-sm ml-auto">Logout</a>
```
nwalters512, line 19: Unless there's a specific reason to have a logout button here, let's remove this in favor of the one in the navbar dropdown.

pages/userSettings/userSettings.ejs
```diff
@@ -13,6 +13,59 @@
...
+                <td>
+                  <%= authn_user.uid %>
+                </td>
+              </tr>
+              <tr>
+                <th>User Name</th>
+                <td><%= authn_user.name %></td>
+              </tr>
+              <tr>
+                <th>Unique Id Number (UIN)</th>
```
nwalters512, line 34: This is not guaranteed to be a number; let's just call this `UIN`.

```suggestion
                <th>UIN</th>
```

pages/userSettings/userSettings.ejs
```diff
@@ -13,6 +13,59 @@
...
 
+      <div class="card mb-4">
+        <div class="card-header bg-primary text-white d-flex align-items-center">
+          Logged-in user profile information
+          <a href="/pl/logout" class="btn btn-danger btn-sm ml-auto">Logout</a>
+        </div>
+          <table class="table table-sm two-column-description">
+            <tbody>
+              <tr>
+                <th>User ID (UID)</th>
```
nwalters512, line 24: "User ID" actually means something different to us. I'd be inclined to just call this `UID` - that has a well-known meaning to most people who would need to interpret this information.

```suggestion
                <th>UID</th>
```

pages/userSettings/userSettings.ejs
```diff
@@ -13,6 +13,59 @@
     <div id="content" class="container">
       <h1 class="mb-4">Settings</h1>
 
+      <div class="card mb-4">
+        <div class="card-header bg-primary text-white d-flex align-items-center">
+          Logged-in user profile information
```
nwalters512, line 18:
```suggestion
          User profile
```

pages/userSettings/userSettings.ejs
```diff
@@ -13,6 +13,59 @@
...
+          <a href="/pl/logout" class="btn btn-danger btn-sm ml-auto">Logout</a>
+        </div>
+          <table class="table table-sm two-column-description">
```
nwalters512, line 21: Can you fix up the indentation here? This should be at the same level as the above `<div>`.

pages/userSettings/userSettings.ejs
```diff
@@ -13,6 +13,59 @@
...
+          Browser configuration
+        </div>
+        <div class="card-body">
+          <p>These options will let you reset some browser settings related to technology inside PrairieLearn.
+            Use these buttons to reset things if things are broken for you.
+          </p>
+          <p>
+          <button class="btn btn-md btn-info"
+            onClick="localStorage.removeItem('MathJax-Menu-Settings');alert('MathJax menu settings have been reset');">
+            Reset MathJax menu settings
```
nwalters512, line 62:
```suggestion
            Reset MathJax settings
```

pages/userSettings/userSettings.ejs
```diff
@@ -13,6 +13,59 @@
...
+          </button> Can be helpful if your math shows up as
+          <strong>$ x = \frac {-b \pm \sqrt {b^2 - 4ac}}{2a} $</strong>
```
nwalters512, line 64: Let's move the explanatory text into a `<p>` that precedes the button. Given that there's only one button, I'd be inclined to just describe the option instead of referring to "these options". Maybe copy like this?

> Resetting the MathJax settings may help if math formulas are showing as [bad example here].

nwalters512 (commented):
As usual, I have design + formatting thoughts 😛 this is looking good though! I'm excited to close a bunch of the related issues.

pages/userSettings/userSettings.ejs
trombonekenny: I would like to put something there in addition to a techy acronym. What words would you use to describe UID and UIN?

pages/userSettings/userSettings.ejs
trombonekenny: Wouldn't the linter / formatter catch that?

pages/userSettings/userSettings.ejs
trombonekenny: I'm going to leave this one as is, as the menu name hints where the settings got set from. There could be other reset buttons in the future.

pages/userSettings/userSettings.ejs
nwalters512: Nope; this is a legacy EJS file.

pages/userSettings/userSettings.ejs
nwalters512: We already use the term "UID" throughout PL (think access rules), so I don't think it'll be confusing to folks. If you want to give UIN a human-readable name, I might go with "unique identifier"?

pages/userSettings/userSettings.ejs
trombonekenny: I tried it a few ways and I feel the left justified button works best.

nwalters512: "precedes" as in on the preceding line as opposed to on the same line as the button; I agree that a left-justified button is the way to go.

trombonekenny: image

I'm OK with putting the text on the line above the button. That degrades better with long text than inline with the button.

pages/userSettings/userSettings.ejs
```diff
@@ -12,6 +12,56 @@
...
+        <div class="card-body">
+          <p>This section will let you reset browser settings related to technology inside PrairieLearn.</p>
+           <p>
+            If math formulas shows up as code like <strong>$ x = \frac {-b \pm \sqrt {b^2 - 4ac}}{2a} $</strong>
+            <br>
+            <button class="btn btn-md btn-info"
+              onClick="localStorage.removeItem('MathJax-Menu-Settings');alert('MathJax menu settings have been reset');">
+              Reset MathJax menu settings
+            </button>
+          </p>
```
nwalters512, line 62: Nit:

```suggestion
          <p>
            If math formulas shows up as code like <strong>$ x = \frac {-b \pm \sqrt {b^2 - 4ac}}{2a} $</strong>,
            resetting the MathJax menu settings may help.
          </p>
          <button class="btn btn-md btn-info"
            onClick="localStorage.removeItem('MathJax-Menu-Settings');alert('MathJax menu settings have been reset');">
            Reset MathJax menu settings
          </button>
```

nwalters512 (approved):

PR #7289, opened 2023-03-02
Reformat exampleCourse documentation comments

nwalters512:
While reviewing another PR, I noticed that the documentation note in `exampleCourse/questions/demo/annotated/mathNumberInput/question.html` wasn't actually a comment. I changed that and took the opportunity to find/replace to remove an unnecessary `<p>` tag.


eliotwrobson (approved):
Seems fine to me!

mwest1066 (approved):

PR #7291, opened 2023-03-03
Move Docker utilities into new @prairielearn/docker-utils package

nwalters512:


pages/shared/syncHelpers.js
```diff
@@ -303,12 +310,164 @@ module.exports.gitStatus = function (locals, callback) {
   });
 };
 
+function locateImage(image, callback) {
```
nwalters512, line 313: All of the following code was only ever used in this file. Rather than trying to convert it to TypeScript and modernize it, I opted to just inline it here.

pages/shared/syncHelpers.js
```diff
@@ -303,12 +310,164 @@ module.exports.gitStatus = function (locals, callback) {
...
+              // Create a new docker image instance with the new registry name
+              // localImage isn't specific enough to the ECR repo
+              const pushImageName = repository.getCombined();
+              var pushImage = new Docker.Image(docker.modem, pushImageName);
+
+              job.info(`Pushing image: ${repository.getCombined()}`);
+              pushImage.push(
+                {
+                  authconfig: dockerAuth,
+                },
```
nwalters512, line 436: The only change I made to this copy/pasted code was moving the auth information from the third argument to this `authconfig` property, which is actually reflected in the TypeScript types.

packages/docker-utils/src/index.test.ts
```diff
@@ -0,0 +1,37 @@
+import { assert } from 'chai';
+
+import { DockerName } from './index';
+
+describe('DockerName', () => {
```
nwalters512, line 5: I added some new tests!

mwest1066: Awesome!

pages/shared/syncHelpers.js
mwest1066: Reasonable.

mwest1066 (approved):

PR #7292, opened 2023-03-03
Create infrastructure for batched migrations

nwalters512:
This PR creates infrastructure for running batched migrations, which are migrations that execute in small chunks of a table's ID space to avoid locking every row in the table.

### DIFF
--- a/packages/migrations/schema-migrations/20230303193423_batched_migrations__create.sql
+++ b/packages/migrations/schema-migrations/20230303193423_batched_migrations__create.sql
@@ -1,0 +1,12 @@
+CREATE TABLE IF NOT EXISTS batched_migrations (
+  current BIGINT NOT NULL,
+  finished_at TIMESTAMP WITH TIME ZONE,
+  id BIGSERIAL PRIMARY KEY,
+  max BIGINT NOT NULL,
+  min BIGINT NOT NULL,
+  name TEXT,
+  project TEXT DEFAULT 'prairielearn',
+  started_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+  timestamp TEXT,
+  UNIQUE (project, timestamp)
+)
--- a/packages/migrations/package.json
+++ b/packages/migrations/package.json
@@ -1,0 +1,19 @@
+{
+  "name": "@prairielearn/migrations",
+  "version": "1.0.0",
+  "main": "./dist/index.js",
+  "scripts": {
+    "build": "tsc",
+    "dev": "tsc --watch --preserveWatchOutput"
+  },
+  "devDependencies": {
+    "@prairielearn/tsconfig": "*",
+    "@types/node": "^18.14.2",
+    "ts-node": "^10.9.1",
+    "typescript": "^4.9.5"
+  },
+  "dependencies": {
+    "@prairielearn/postgres": "^1.2.0",
+    "zod": "^3.20.6"
+  }
+}
--- a/packages/migrations/package.json
+++ b/packages/migrations/package.json
@@ -4,11 +4,14 @@
   "main": "./dist/index.js",
   "scripts": {
     "build": "tsc",
-    "dev": "tsc --watch --preserveWatchOutput"
+    "dev": "tsc --watch --preserveWatchOutput",
+    "test": "mocha --no-config --require ts-node/register src/*.test.ts"
   },
   "devDependencies": {
     "@prairielearn/tsconfig": "*",
+    "@types/mocha": "^10.0.1",
     "@types/node": "^18.14.2",
+    "mocha": "^10.2.0",
     "ts-node": "^10.9.1",
     "typescript": "^4.9.5"
   },
--- a/packages/migrations/src/batched-migration.test.ts
+++ b/packages/migrations/src/batched-migration.test.ts
@@ -1,0 +1,7 @@
+import { assert } from 'chai';
+
+describe('BatchedMigrationExecutor', () => {
+  it('works', () => {
+    assert.equal(true, true);
+  });
+});
--- a/packages/migrations/src/batched-migration.test.ts
+++ b/packages/migrations/src/batched-migration.test.ts
@@ -23,7 +23,6 @@ async function makeTestDatabase() {
   await client.query(`CREATE DATABASE ${POSTGRES_DATABASE};`);
   await client.end();
 
-  console.log('here');
   await sqldb.initAsync(
     {
       user: POSTGRES_USER,
@@ -36,7 +35,6 @@ async function makeTestDatabase() {
       throw err;
     }
   );
-  console.log('now here');
 
   // Run local migrations to set up the database. Once the real migration-running
   // code is lifted into this package, we can use that instead.
@@ -71,9 +69,7 @@ class TestBatchMigration extends BatchedMigration {
     return '10000';
   }
 
-  async execute(start: BigInt, end: BigInt) {
-    console.log('running for range', start, end);
-  }
+  async execute(_start: BigInt, _end: BigInt) {}
 }
 
 async function getBatchedMigrationState() {
--- a/packages/migrations/src/batched-migration.test.ts
+++ b/packages/migrations/src/batched-migration.test.ts
@@ -68,7 +68,7 @@ class TestBatchMigration extends BatchedMigration {
   }
 
   async getMax() {
-    return '0';
+    return '10000';
   }
 
   async execute(start: BigInt, end: BigInt) {
@@ -76,15 +76,35 @@ class TestBatchMigration extends BatchedMigration {
   }
 }
 
+async function getBatchedMigrationState() {
+  const migration = await sqldb.queryOneRowAsync(
+    'SELECT * FROM batched_migrations WHERE name = $name;',
+    { name: 'test_batch_migration' }
+  );
+  return migration.rows[0];
+}
+
 describe('BatchedMigrationExecutor', () => {
   beforeEach(async () => makeTestDatabase());
   afterEach(async () => destroyTestDatabase());
 
-  it('works', async () => {
+  it('runs one iteration of a batched migration', async () => {
     assert.equal(true, true);
 
     const migration = new TestBatchMigration();
     const executor = new BatchedMigrationExecutor('test_batch_migration', migration);
+    await executor.run({ iterations: 1 });
+
+    const migrationState = await getBatchedMigrationState();
+    assert.equal(migrationState.current, '1000');
+  });
+
+  it('runs an entire batched migration', async () => {
+    const migration = new TestBatchMigration();
+    const executor = new BatchedMigrationExecutor('test_batch_migration', migration);
     await executor.run();
+
+    const migrationState = await getBatchedMigrationState();
+    assert.equal(migrationState.current, '10000');
   });
 });
--- a/packages/migrations/src/batched-migration.test.ts
+++ b/packages/migrations/src/batched-migration.test.ts
@@ -1,7 +1,90 @@
 import { assert } from 'chai';
+import * as sqldb from '@prairielearn/postgres';
+import pg from 'pg';
+import fs from 'fs';
+import path from 'path';
+
+import { BatchedMigration, BatchedMigrationExecutor } from './batched-migration';
+import { SCHEMA_MIGRATIONS_PATH } from './index';
+
+const WORKER_ID = process.env.MOCHA_WORKER_ID || '1';
+const POSTGRES_DATABASE = `prairielearn_migrations_${WORKER_ID}`;
+const POSTGRES_INIT_CONNECTION_STRING = 'postgres://postgres@localhost/postgres';
+
+// This was borrowed from `tests/helperDb.js` in the main database. When that's
+// moved to a shared package, we should be able to reuse that functionality.
+async function makeTestDatabase() {
+  const POSTGRES_USER = 'postgres';
+  const POSTGRES_HOST = 'localhost';
+
+  const client = new pg.Client(POSTGRES_INIT_CONNECTION_STRING);
+  await client.connect();
+  await client.query(`DROP DATABASE IF EXISTS ${POSTGRES_DATABASE};`);
+  await client.query(`CREATE DATABASE ${POSTGRES_DATABASE};`);
+  await client.end();
+
+  console.log('here');
+  await sqldb.initAsync(
+    {
+      user: POSTGRES_USER,
+      database: POSTGRES_DATABASE,
+      host: POSTGRES_HOST,
+      max: 10,
+      idleTimeoutMillis: 30000,
+    },
+    (err) => {
+      throw err;
+    }
+  );
+  console.log('now here');
+
+  // Run local migrations to set up the database. Once the real migration-running
+  // code is lifted into this package, we can use that instead.
+  const migrations = fs.readdirSync(SCHEMA_MIGRATIONS_PATH);
+  const sortedMigrations = migrations.sort((a, b) => {
+    const aTimestamp = parseInt(a.split('_')[0]);
+    const bTimestamp = parseInt(b.split('_')[0]);
+    return aTimestamp - bTimestamp;
+  });
+  for (const migrationName of sortedMigrations) {
+    const migrationPath = path.join(SCHEMA_MIGRATIONS_PATH, migrationName);
+    const migration = fs.readFileSync(migrationPath, 'utf8');
+    await sqldb.queryAsync(migration, {});
+  }
+}
+
+async function destroyTestDatabase() {
+  await sqldb.closeAsync();
+
+  const client = new pg.Client(POSTGRES_INIT_CONNECTION_STRING);
+  await client.connect();
+  await client.query(`DROP DATABASE IF EXISTS ${POSTGRES_DATABASE}`);
+  await client.end();
+}
+
+class TestBatchMigration extends BatchedMigration {
+  async getMin() {
+    return '0';
+  }
+
+  async getMax() {
+    return '0';
+  }
+
+  async execute(start: BigInt, end: BigInt) {
+    console.log('running for range', start, end);
+  }
+}
 
 describe('BatchedMigrationExecutor', () => {
-  it('works', () => {
+  beforeEach(async () => makeTestDatabase());
+  afterEach(async () => destroyTestDatabase());
+
+  it('works', async () => {
     assert.equal(true, true);
+
+    const migration = new TestBatchMigration();
+    const executor = new BatchedMigrationExecutor('test_batch_migration', migration);
+    await executor.run();
   });
 });
--- a/packages/migrations/src/batched-migration.ts
+++ b/packages/migrations/src/batched-migration.ts
@@ -1,0 +1,94 @@
+import { z } from 'zod';
+import {
+  queryAsync,
+  queryValidatedOneRow,
+  queryValidatedZeroOrOneRow,
+} from '@prairielearn/postgres';
+
+const BatchedMigrationSchema = z.object({
+  id: z.string(),
+  name: z.string(),
+  min: z.bigint({ coerce: true }),
+  max: z.bigint({ coerce: true }),
+  current: z.bigint({ coerce: true }),
+});
+type BatchedMigrationRow = z.infer<typeof BatchedMigrationSchema>;
+
+/**
+ * A batched migration operates on a range of IDs in batches. It's designed
+ * for migrations that are too large to run in a single transaction.
+ *
+ * Migrations should be idempotent, as they may be run multiple times in case
+ * of unexpected failure.
+ */
+export abstract class BatchedMigration {
+  public async getMin(): Promise<string> {
+    return '0';
+  }
+  public abstract getMax(): Promise<string>;
+  public abstract execute(start: BigInt, end: BigInt): Promise<void>;
+}
+
+export class BatchedMigrationExecutor {
+  private name: string;
+  private migration: BatchedMigration;
+
+  constructor(name: string, migration: BatchedMigration) {
+    this.name = name;
+    this.migration = migration;
+  }
+
+  private async getMigrationState(): Promise<BatchedMigrationRow> {
+    const migration = await queryValidatedZeroOrOneRow(
+      'SELECT * FROM batched_migrations WHERE name = $name;',
+      { name: this.name },
+      BatchedMigrationSchema
+    );
+    if (migration) return migration;
+
+    return this.createMigrationState();
+  }
+
+  private async createMigrationState(): Promise<BatchedMigrationRow> {
+    const min = await this.migration.getMin();
+    const max = await this.migration.getMax();
+    return await queryValidatedOneRow(
+      'INSERT INTO batched_migrations (name, min, max, current) VALUES ($name, $min, $max, $min) ON CONFLICT DO NOTHING RETURNING *;',
+      { name: this.name, min, max },
+      BatchedMigrationSchema
+    );
+  }
+
+  private async updateMigrationState(current: BigInt): Promise<void> {
+    await queryAsync('UPDATE batched_migrations SET current = $current WHERE name = $name;', {
+      name: this.name,
+      current,
+    });
+  }
+
+  /**
+   * Should be used when the migration is being run with a short-lived lock
+   * held. Allows the migration to be picked up by different machines and
+   * be automatically restarted in case of failure.
+   *
+   * @param duration How many seconds to execute for.
+   */
+  async runForDuration(duration: number): Promise<void> {
+    await this.run({ signal: AbortSignal.timeout(duration * 1000) });
+  }
+
+  async run({ signal }: { signal?: AbortSignal }): Promise<void> {
+    const migrationState = await this.getMigrationState();
+
+    let current = migrationState.current;
+    while (!signal?.aborted && current < migrationState.max) {
+      let min = current;
+      let max = current + 1000n;
+
+      this.migration.execute(min, max);
+
+      current = max;
+      await this.updateMigrationState(current);
+    }
+  }
+}
--- a/packages/migrations/src/batched-migration.ts
+++ b/packages/migrations/src/batched-migration.ts
@@ -77,19 +77,28 @@ export class BatchedMigrationExecutor {
     await this.run({ signal: AbortSignal.timeout(duration * 1000) });
   }
 
-  async run({ signal }: { signal?: AbortSignal } = {}): Promise<void> {
+  async run({
+    signal,
+    iterations,
+  }: { signal?: AbortSignal; iterations?: number } = {}): Promise<void> {
     const migrationState = await this.getMigrationState();
 
     let current = migrationState.current;
-    while (!signal?.aborted && current <= migrationState.max) {
-      let min = current;
+    let iterationCount = 0;
+    while (
+      !signal?.aborted &&
+      (iterations ? iterationCount < iterations : true) &&
+      current < migrationState.max
+    ) {
       // TODO: configurable batch size
       let max = current + 1000n;
+      if (max > migrationState.max) max = migrationState.max;
 
-      this.migration.execute(min, max);
+      this.migration.execute(current, max);
 
       current = max;
       await this.updateMigrationState(current);
+      iterationCount += 1;
     }
   }
 }
--- a/packages/migrations/src/batched-migration.ts
+++ b/packages/migrations/src/batched-migration.ts
@@ -77,12 +77,13 @@ export class BatchedMigrationExecutor {
     await this.run({ signal: AbortSignal.timeout(duration * 1000) });
   }
 
-  async run({ signal }: { signal?: AbortSignal }): Promise<void> {
+  async run({ signal }: { signal?: AbortSignal } = {}): Promise<void> {
     const migrationState = await this.getMigrationState();
 
     let current = migrationState.current;
-    while (!signal?.aborted && current < migrationState.max) {
+    while (!signal?.aborted && current <= migrationState.max) {
       let min = current;
+      // TODO: configurable batch size
       let max = current + 1000n;
 
       this.migration.execute(min, max);
--- a/packages/migrations/src/index.ts
+++ b/packages/migrations/src/index.ts
@@ -1,1 +1,1 @@
-export { BatchedMigration } from './batched-migration';
+export { BatchedMigration, BatchedMigrationExecutor } from './batched-migration';
--- a/packages/migrations/src/index.ts
+++ b/packages/migrations/src/index.ts
@@ -1,1 +1,5 @@
+import path from 'path';
+
 export { BatchedMigration, BatchedMigrationExecutor } from './batched-migration';
+
+export const SCHEMA_MIGRATIONS_PATH = path.resolve(__dirname, '..', 'schema-migrations');
--- a/packages/migrations/src/index.ts
+++ b/packages/migrations/src/index.ts
@@ -1,0 +1,1 @@
+export { BatchedMigration } from './batched-migration';
--- a/packages/migrations/tsconfig.json
+++ b/packages/migrations/tsconfig.json
@@ -1,0 +1,8 @@
+{
+  "extends": "@prairielearn/tsconfig",
+  "compilerOptions": {
+    "outDir": "./dist",
+    "rootDir": "./src",
+    "types": ["node"],
+  }
+}
--- a/packages/migrations/tsconfig.json
+++ b/packages/migrations/tsconfig.json
@@ -3,6 +3,6 @@
   "compilerOptions": {
     "outDir": "./dist",
     "rootDir": "./src",
-    "types": ["node"],
+    "types": ["node", "mocha"],
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -2481,7 +2481,9 @@ __metadata:
   dependencies:
     "@prairielearn/postgres": ^1.2.0
     "@prairielearn/tsconfig": "*"
+    "@types/mocha": ^10.0.1
     "@types/node": ^18.14.2
+    mocha: ^10.2.0
     ts-node: ^10.9.1
     typescript: ^4.9.5
     zod: ^3.20.6
--- a/yarn.lock
+++ b/yarn.lock
@@ -2475,6 +2475,19 @@ __metadata:
   languageName: unknown
   linkType: soft
 
+"@prairielearn/migrations@workspace:packages/migrations":
+  version: 0.0.0-use.local
+  resolution: "@prairielearn/migrations@workspace:packages/migrations"
+  dependencies:
+    "@prairielearn/postgres": ^1.2.0
+    "@prairielearn/tsconfig": "*"
+    "@types/node": ^18.14.2
+    ts-node: ^10.9.1
+    typescript: ^4.9.5
+    zod: ^3.20.6
+  languageName: unknown
+  linkType: soft
+
 "@prairielearn/opentelemetry@^1.0.0, @prairielearn/opentelemetry@workspace:packages/opentelemetry":
   version: 0.0.0-use.local
   resolution: "@prairielearn/opentelemetry@workspace:packages/opentelemetry"
@@ -2511,7 +2524,7 @@ __metadata:
   languageName: unknown
   linkType: soft
 
-"@prairielearn/postgres@^1.0.0, @prairielearn/postgres@workspace:packages/postgres":
+"@prairielearn/postgres@^1.0.0, @prairielearn/postgres@^1.2.0, @prairielearn/postgres@workspace:packages/postgres":
   version: 0.0.0-use.local
   resolution: "@prairielearn/postgres@workspace:packages/postgres"
   dependencies:


packages/migrations/src/batched-migrations/batched-migration-job.ts
```diff
@@ -0,0 +1,18 @@
+import { z } from 'zod';
+
+export const BatchedMigrationJobStatusSchema = z.enum(['pending', 'failed', 'succeeded']);
+export type BatchedMigrationJobStatus = z.infer<typeof BatchedMigrationJobStatusSchema>;
+
+export const BatchedMigrationJobRowSchema = z.object({
+  id: z.string(),
+  batched_migration_id: z.string(),
+  min_value: z.bigint({ coerce: true }),
+  max_value: z.bigint({ coerce: true }),
```
mwest1066, line 10: I assume you are using `Bigint` to allow us to calculate with values from postgres BIGINT `id` columns? I’ve been wondering what to do with this because of the `table_name_to_jsonb()` sprocs we have in PT. Ideally I’d like to remove these sprocs and use `to_jsonb()` instead. The main issue will be that `to_jsonb()` translates postgres BIGINT into a JSON `Number`, so we can’t safely handle `id` columns larger than 2^53 (the biggest integer you can store in a JS `Number`). I was thinking that this might be fine in practice, because that feels like a safe number of rows to me. What do you think about just mapping postgres BIGINT to JS `Number` and not worrying about overflow?

batched-migrations/20230407165410_test.js
```diff
@@ -0,0 +1,26 @@
+// @ts-check
+const { BatchedMigration } = require('@prairielearn/migrations');
+const { setTimeout } = require('node:timers/promises');
+
+// TODO: delete before merging
```
nwalters512, line 5: TODO: remove this file.

migrations/20230410205430_batched_migration__test.js
```diff
@@ -0,0 +1,6 @@
+const { enqueueBatchedMigration } = require('@prairielearn/migrations');
+
+// TODO: delete this file before merging.
```
nwalters512, line 3: TODO: remove this file.

packages/migrations/src/batched-migrations/batched-migrations-runner.ts
```diff
@@ -0,0 +1,288 @@
...
+      throw new Error(`No migration found for identifier ${identifier}`);
+    }
+    const MigrationClass = await this.loadMigrationClass(migrationFile);
+    const migrationInstance = new MigrationClass();
+
+    // TODO: we should still lock this migration. Probably what we want to do
+    // is to have the normal run loop check for a runnable migration *before*
+    // acquiring a lock. If it finds a running one, it should grab and hold a
+    // lock just for that specific migration. This allows the instance that's
+    // performing finalization to run without anyone else trying to grab a global lock.
```
nwalters512, line 135: TODO: grab and hold a lock here.

packages/migrations/src/batched-migrations/batched-migrations-runner.ts
```diff
@@ -0,0 +1,288 @@
...
+        }
+      }
+    }
+  }
+
+  private async getOrStartMigration(): Promise<BatchedMigrationRow | null> {
+    // This query will lock all existing rows in the `batched_migrations` table
+    // to guard against race conditions.
+    //
+    // TODO: should this use `LOCK TABLE ...` instead?
```
nwalters512, line 193: 👀 thoughts? This table will probably always be small enough that we can just use lots of row-level locks. Alternatively, we can just use a named lock here.

packages/migrations/src/batched-migrations/batched-migrations-runner.ts
```diff
@@ -0,0 +1,288 @@
...
+
+    // This server may not yet know about the current running migration. If
+    // that's the case, we'll just skip it for now.
+    const migrationFile = await this.getMigrationForIdentifier(migration.timestamp);
+    if (!migrationFile) {
+      return false;
+    }
+
+    let didWork = false;
+    await tryWithLock(this.lockNameForTimestamp(migrationFile.timestamp), async () => {
```
nwalters512, line 216: We should use an auto-renewing lock here.

pages/administratorBatchedMigrations/administratorBatchedMigrations.html.js
```diff
@@ -0,0 +1,223 @@
...
+ * @returns {string}
+ */
+function AdministratorBatchedMigration({
+  batchedMigration,
+  recentSucceededJobs,
+  recentFailedJobs,
+  resLocals,
+}) {
+  // TODO: directly interpolate BigInt values once this PR lands:
+  // https://github.com/PrairieLearn/PrairieLearn/pull/7492
```
nwalters512, line 84: 👀 this should be safe to do now.

nwalters512: Done in f45f585.

packages/migrations/src/batched-migrations/batched-migrations-runner.ts
nwalters512: I changed this to use a named lock in 46af675.

packages/migrations/src/batched-migrations/batched-migrations-runner.ts
nwalters512: Done in 36c0750.

server.js
```diff
@@ -1979,12 +1988,34 @@ if (config.startServer) {
 
         logger.verbose('Successfully connected to database');
       },
+      async () => {
+        // We need to do this before we run migrations, as some migrations will
+        // call `enqueueBatchedMigration` which requires this to be initialized.
+        //
+        // TODO: add an attempts counter to each job. Increment the counter before
+        // trying to run the job. This is just for our own recordkeeping.
```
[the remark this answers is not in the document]
nwalters512: Done in 5aaa762.

server.js
```diff
@@ -1979,12 +1988,34 @@ if (config.startServer) {
...
+        //
+        // TODO: log error details on job row.
```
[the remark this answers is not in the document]
nwalters512: Done in bac814d.

packages/migrations/src/batched-migrations/batched-migration.ts
```diff
@@ -0,0 +1,117 @@
...
+ * A batched migration operates on a range of IDs in batches. It's designed
+ * for migrations that are too large to run in a single transaction.
+ *
+ * Migrations should be idempotent, as they may be run multiple times in case
+ * of unexpected failure.
+ */
+export abstract class BatchedMigration {
+  public abstract getParameters(): Promise<BatchedMigrationParameters>;
+  public abstract execute(start: bigint, end: bigint): Promise<void>;
+}
```
nwalters512, line 53: This class doesn't actually do any class-y things. I'm thinking maybe we should "make" migrations like this instead?

```ts
import { makeBatchedMigration } from '@prairielearn/migration';

export default makeBatchedMigration({
  getParameters() {
    return {}
  },
  async execute() {
    // ...
  },
});
```

Not having to export something new-able gives us slightly more flexibility in the future. (This is probably bikeshedding.)

.eslintrc.js
```diff
@@ -3,7 +3,12 @@ module.exports = {
     node: true,
     es2021: true,
   },
-  extends: ['eslint:recommended', 'plugin:import/recommended', 'prettier'],
+  extends: [
+    'eslint:recommended',
+    'plugin:import/recommended',
+    'plugin:import/typescript',
+    'prettier',
+  ],
```
nwalters512, line 11: This diff should go away once #7483 lands.

packages/migrations/src/batched-migrations/batched-migration-job.ts
mwest1066: I've been thinking about adding the equivalent of PT's `db-types.ts` file to PL to give us zod and TS types that exactly (as close as possible) mirror the DB tables. I've found this to be super helpful in converting everything from `query()` to `queryValidated()` (as well as adding strict typing everywhere). Part of this requires deciding how to get data out of SQL queries, especially for nested data structures. For PT we've been using the `table_name_to_jsonb` sprocs, but this ties us into a long-term requirement for using sprocs. It would be much nicer to use `to_jsonb()` for this. The main blockers in doing this in PT right now are:
* treatment of `id` columns, where we currently export the BIGINTs as `text` to map to `z.string()`
* conversion of INTERVALs, where we currently convert them to milliseconds to map to `z.number()`

For intervals, `to_jsonb()` will convert to a string like `"00:32:20.32"` (HH:MM:SS.ssss). I was thinking to use a custom zod schema to parse this into a numeric value, like we do with our custom `DateFromISOString` schema for dates.

So that really only leaves `id` columns (and BIGINTs in general). The problem here is that `to_jsonb()` _doesn't_ convert these to strings so we can't use a custom zod schema to transform them to JavaScript `Bigint`. Instead, `to_jsonb()` converts them to JSON numbers, which is where we get the 2^53 restriction from.

This is all a long background to say that I think we should start with right now with a `db-types.ts` (or `.js`) file in PL containing a `BatchedMigrationJobCodec` zod type (except swap `Codec` for whatever we are going to use in PL) and this should have `min_value: z.number()` (not `z.bigint()`).

mwest1066: And yes, this is for consistency, so that we can know that we can always use `to_jsonb()` in SQL to construct nested data and this will correctly validate with zod on the other side.

mwest1066: Although, now I remembered that this is actually in a package, not PL, so it would actually be fine to use `z.bigint()` here. Sorry for all the noise!

SethPoulsen (commented):
I have not read through everything yet so some of my questions may be dumb.

packages/migrations/src/batched-migrations/batched-migrations-runner.ts
```diff
@@ -0,0 +1,309 @@
...
+    this.running = true;
+    while (true) {
+      if (this.abortController.signal.aborted) {
+        this.running = false;
+        return;
+      }
+
+      let didWork = false;
+      try {
+        didWork = await this.maybePerformWork(60 * 1000);
```
mwest1066, line 168: Add config variable for this value?

packages/migrations/src/batched-migrations/batched-migrations-runner.ts
```diff
@@ -0,0 +1,309 @@
...
+      }
+
+      // If we did work, we'll immediately try again since there's probably more
+      // work to be done. If not, we'll sleep for a while - maybe some more work
+      // will become available!
+      if (!didWork) {
+        // We provide the signal here so that we can more quickly stop things
+        // when we're shutting down.
+        try {
+          await sleep(30_000, null, { ref: false, signal: this.abortController.signal });
```
mwest1066, line 180: Another config value?

database/enums/enum_batched_migration_status.pg
```diff
@@ -0,0 +1 @@
+pending, paused, running, finalizing, failed, succeeded
```
mwest1066, line 1: Can you help me understand the use of the `running` and `finalizing` states? Are these purely informational or do they actually contain useful information for the state machine?

database/tables/batched_migrations.pg
```diff
@@ -0,0 +1,14 @@
+columns
+    current: bigint not null
+    finished_at: timestamp with time zone
+    id: bigint not null default nextval('batched_migrations_id_seq'::regclass)
+    max: bigint not null
+    min: bigint not null
+    name: text
+    project: text default 'prairielearn'::text
+    started_at: timestamp with time zone default CURRENT_TIMESTAMP
+    timestamp: text
```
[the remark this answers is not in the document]
nwalters512: `timestamp` is something like `20230411211407` that we pull directly from the migration filename. In the case of batched migrations, it's really more of a unique identifier for a specific migration file; we don't actually rely on it for ordering (although we did in intermediate implementations), since ordering is determined by the order of the regular migrations that actually enqueue the batched migrations. Maybe for batched migrations specifically, we can get away without using a timestamp and just rely on the filename (minus extension) as a unique identifier? That might be preferable so that folks don't wrongly assume that the `timestamp` portion of the filename will influence ordering.

packages/migrations/src/batched-migrations/batched-migration-runner.test.ts
```diff
@@ -0,0 +1,193 @@
...
+import { SCHEMA_MIGRATIONS_PATH, init } from '../index';
+import { BatchedMigrationJobRowSchema, BatchedMigrationRowSchema } from './schemas';
+
+const postgresTestUtils = makePostgresTestUtils({
+  database: 'prairielearn_migrations',
+});
+
+class TestBatchMigration extends BatchedMigration {
+  private failingIds: bigint[] = [];
+  private _executionCount = 0;
```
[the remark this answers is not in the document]
nwalters512: `TestBatchMigration` also has an `executionCount` getter; a class can't contain both a property and a getter by the same name.

lib/config.js
```diff
@@ -302,6 +302,11 @@ config.chunksLoadBalancerRequestsCapacityFactor = 1000;
  */
 config.isEnterprise = false;
 
+/**
+ * How long we'll hold a lock and execute batched migration jobs.
+ */
+config.batchedMigrationsRunDurationMs = 60_000;
+
```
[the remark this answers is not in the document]
nwalters512: Why? Remember that this isn't literally locking tables with actual data, it's just using our named locks functionality.

packages/migrations/src/batched-migrations/fixtures/20230406184103_test_migration_1.ts
```diff
@@ -0,0 +1,18 @@
+import { BatchedMigration } from '../batched-migration';
+
+class TestMigration1 extends BatchedMigration {
+  async getParameters() {
+    return {
+      min: 1n,
+      max: 100n,
+      batchSize: 10,
+    };
```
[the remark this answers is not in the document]
nwalters512: In real-world migrations, you'd do something like `SELECT MAX(id) FROM my_table`; the bounds should never be hardcoded (except for maybe the default, which should almost always be `1`).

There's not a great story for what happens if you pick a batch size that's too large at the moment. In theory, one could manually/automatically split failed jobs into smaller batches, but that's not implemented yet. I think one would have to choose a _really_ large batch size to get anything to timeout.

packages/migrations/schema-migrations/20230303193423_batched_migrations__create.sql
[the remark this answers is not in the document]
nwalters512: This is part of a package, so it must be self-contained (that is, cannot be in the root `migrations/` directory.

packages/migrations/src/batched-migrations/batched-migration-runner.ts
```diff
@@ -0,0 +1,178 @@
...
+    this.migrationStatus = migration.status;
+  }
+
+  private async hasIncompleteJobs(migration: BatchedMigrationRow) {
+    const res = await queryValidatedOneRow(
+      sql.batched_migration_has_incomplete_jobs,
+      { batched_migration_id: migration.id },
+      z.object({ exists: z.boolean() })
+    );
+    return res.exists;
```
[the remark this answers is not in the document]
nwalters512: I think one could even use `queryValidatedSingleColumnOneRow` to directly return a boolean?

database/enums/enum_batched_migration_status.pg
nwalters512: Good question! `running` says "this batched migration is running asynchronously and can safely be worked on by any machine". `finalizing` says "this batched migration is being run synchronously to completion by a specific instance and other machines should not attempt to work on it".

packages/migrations/src/batched-migrations/batched-migration.ts
mwest1066: I'm also generally in favor of avoiding classes wherever possible.

database/tables/batched_migrations.pg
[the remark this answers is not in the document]
mwest1066: I would prefer to just use the "timestamp-like" prefix of the filename as the identifier, to allow us to rename the rest of the filename without breaking everything (I think we've done this in the past for regular migrations).

packages/migrations/src/batched-migrations/batched-migration-runner.ts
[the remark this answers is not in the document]
mwest1066: In this case, absolutely! I didn't think about what it was actually returning 😄

database/enums/enum_batched_migration_status.pg
mwest1066: How is `running` different than `pending`? It seems that as soon as we are in `pending` state it should be good to go?

What happens if a machine dies while it is doing the finalization step? I would prefer to not have a `finalizing` state (just be in `running`/`pending`) and instead just have the finalizing machine take a lock and then run to completion. I can't see the value in blocking other machines from processing blocks during this time?

database/tables/batched_migration_jobs.pg
```diff
@@ -0,0 +1,20 @@
+columns
+    attempts: integer default 0
+    batched_migration_id: bigint not null
+    created_at: timestamp with time zone default CURRENT_TIMESTAMP
+    data: jsonb
+    finished_at: timestamp with time zone
+    id: bigint not null default nextval('batched_migration_jobs_id_seq'::regclass)
+    max_value: bigint not null
+    min_value: bigint not null
```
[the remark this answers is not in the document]
nwalters512: It can't hurt!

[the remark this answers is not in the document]
nwalters512: Actually, it already exists (check the migration file), my db-description tool just apparently can't handle `CHECK` constraints. I'll have to add support for that.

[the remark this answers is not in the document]
nwalters512: Once #7502 lands, the existing constraints will be captured here.

database/enums/enum_batched_migration_status.pg
nwalters512: `pending` represents the fact that batched migrations are technically queued. Say there's a migration `foo` that's half complete and we deploy code that calls `enqueueBatchedMigration('bar')`. `bar` won't immediately start running; it will only start being processed once `foo` completes. This isn't strictly necessary, but I figured it would be useful to specifically define this behavior *and* I think it's generally true that we'll want to spend our resources running older migrations first.

If a machine dies while it's doing finalization, the corresponding non-batch migration won't be finished, so the next instance to try to run that migration will see that the batched migration is already in the `finalizing` state and pick up where the last one left off.

I didn't implement this yet, but the value in "stealing" the work from other machines is that we can then print progress as we go for anyone who is monitoring the migration/deployment. We don't get that if we allow other random hosts to work on the jobs.

database/tables/batched_migration_jobs.pg
```diff
@@ -0,0 +1,20 @@
+columns
+    attempts: integer default 0
```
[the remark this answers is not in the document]
nwalters512: Done in 728e0f6.

database/tables/batched_migration_jobs.pg
```diff
@@ -0,0 +1,20 @@
...
+    batched_migration_id: bigint not null
+    created_at: timestamp with time zone default CURRENT_TIMESTAMP
```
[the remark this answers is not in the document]
nwalters512: Done in 728e0f6.

database/tables/batched_migration_jobs.pg
```diff
@@ -0,0 +1,20 @@
...
+    data: jsonb
+    finished_at: timestamp with time zone
+    id: bigint not null default nextval('batched_migration_jobs_id_seq'::regclass)
+    max_value: bigint not null
+    min_value: bigint not null
+    started_at: timestamp with time zone
+    status: enum_batched_migration_job_status default 'pending'::enum_batched_migration_job_status
```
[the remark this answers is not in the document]
nwalters512: Done in 728e0f6.

database/tables/batched_migration_jobs.pg
```diff
@@ -0,0 +1,20 @@
...
+    updated_at: timestamp with time zone default CURRENT_TIMESTAMP
```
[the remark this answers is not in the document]
nwalters512: Done in 728e0f6.

database/tables/batched_migrations.pg
```diff
@@ -0,0 +1,19 @@
+columns
+    batch_size: integer not null
+    created_at: timestamp with time zone default CURRENT_TIMESTAMP
+    filename: text not null
+    id: bigint not null default nextval('batched_migrations_id_seq'::regclass)
+    max_value: bigint not null
+    min_value: bigint not null
+    project: text default 'prairielearn'::text
+    started_at: timestamp with time zone
+    status: enum_batched_migration_status default 'pending'::enum_batched_migration_status
```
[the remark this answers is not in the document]
nwalters512: Done in 728e0f6.

database/tables/batched_migrations.pg
```diff
@@ -0,0 +1,19 @@
...
+    timestamp: text not null
+    updated_at: timestamp with time zone default CURRENT_TIMESTAMP
```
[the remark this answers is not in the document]
nwalters512: Done in 728e0f6.

packages/migrations/src/batched-migrations/batched-migration-runner.ts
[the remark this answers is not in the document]
nwalters512: Changed in 7a79a18.

database/tables/batched_migration_jobs.pg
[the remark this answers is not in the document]
nwalters512: These were captured in the database descriptions in 19c6d4c.

database/enums/enum_batched_migration_status.pg
mwest1066: Ok, I'm convinced! Thanks for the clear explanations!

server.js
```diff
@@ -2050,6 +2077,11 @@ if (config.startServer) {
           });
         });
       },
+      async () => {
+        // Now that all migrations have been run, we can start executing any
+        // batched migrations that may have been enqueued by migrations.
+        startBatchedMigrations();
```
nwalters512, line 2083: We'll want to have a config option like `config.runBatchedMigrations` that we can use to disable execution on chunk servers, just like we do with `config.runMigrations`. Or honestly, we could even use the same option - what do you think @mwest1066?

packages/migrations/src/batched-migrations/batched-migration-runner.test.ts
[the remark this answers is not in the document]
nwalters512: This was resolved in 2504585 where I stopped using classes entirely.

packages/migrations/src/batched-migrations/batched-migrations-runner.ts
nwalters512: Done in b81ea6d.

packages/migrations/src/batched-migrations/batched-migrations-runner.ts
nwalters512: Done in b81ea6d.

database/enums/enum_batched_migration_status.pg
nwalters512: Yay! I just implemented logging during finalization, here's what that will look like:

Screenshot 2023-04-12 at 09 46 58

Look reasonable?

batched-migrations/20230407165410_test.js
nwalters512: Removed in 3719949.

migrations/20230410205430_batched_migration__test.js
nwalters512: Removed in 3719949.

database/tables/batched_migrations.pg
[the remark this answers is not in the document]
SethPoulsen: Maybe we could rename the `timestamp` column to makes that more clear?

packages/migrations/src/batched-migrations/fixtures/20230406184103_test_migration_1.ts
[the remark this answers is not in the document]
SethPoulsen: ok, cool. I figured as much and I don't think its a problem for now, I mostly wanted to make sure I was understanding things properly.

packages/migrations/schema-migrations/20230303193423_batched_migrations__create.sql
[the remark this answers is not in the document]
SethPoulsen: I'm not sure I understand. This code is only needs to be run once to set up the tables for batched migrations right? Like, its only a part of getting the production environment ready for the package to be used, not actually part of the ongoing functionality of the package? I don't see why create a whole new folder for it instead of putting it in with all the other migrations.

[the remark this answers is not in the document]
SethPoulsen: Ok cool, that makes sense I guess. I still feel sort of like there is trade-off in that now we have PL migrations in multiple places which will not be obvious to a newcomer to the codebase, but that's a minor issue.

database/tables/batched_migrations.pg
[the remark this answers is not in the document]
nwalters512: This nomenclature is copied directly from the existing `migrations` table. I'd rather keep consistent terminology for the same concept.

database/tables/batched_migration_jobs.pg
```diff
@@ -0,0 +1,20 @@
...
+
+indexes
+    batched_migration_jobs_pkey: PRIMARY KEY (id) USING btree (id)
+    batched_migration_jobs_batched_migration_id_max_value_idx: USING btree (batched_migration_id, max_value)
+    batched_migration_jobs_batched_migration_id_status_idx: USING btree (batched_migration_id, status)
```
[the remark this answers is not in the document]
SethPoulsen: This could be used to help make sure two servers don't attempt to do a job on the same range at the same time. I suppose this would be a rare race condition, and I know the operation must be idempotent so it won't lead to inconsistency issues, its just nice to not duplicate work.

database/enums/enum_batched_migration_status.pg
SethPoulsen: Why do we want the feature of being able to 'finalize' a migration though?

Isn't it good enough to just be able to check if its completed? I don't see why we want to be able to force a machine to synchronously finish a migration.

packages/migrations/src/batched-migrations/batched-migration-runner.ts
```diff
@@ -0,0 +1,200 @@
...
+    // Safety check: if there are any pending jobs, don't mark this
+    // migration as finished.
+    if (await this.hasIncompleteJobs(migration)) {
+      this.log(`Incomplete jobs found, not marking as finished`);
+    }
+
+    const hasFailedJobs = await this.hasFailedJobs(migration);
+    const finalStatus = hasFailedJobs ? 'failed' : 'succeeded';
+    await updateBatchedMigrationStatus(migration.id, finalStatus);
+    this.log(`Finished with status '${finalStatus}'`);
```
SethPoulsen, line 90: Should the code below be in an 'else' block?

SethPoulsen (commented):
I've read almost all of it now. Looks great overall, a few questions.

packages/migrations/src/batched-migrations/batched-migrations-runner.ts
```diff
@@ -0,0 +1,318 @@
...
+
+      const runner = new BatchedMigrationRunner(migration, migrationImplementation, {
+        logProgress: true,
+      });
+
+      // Because we don't give any arguments to `run()`, it will run until it
+      // has attempted every job.
+      await runner.run();
+    });
+
```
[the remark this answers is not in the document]
SethPoulsen: Oh actually I guess there is still no risk of overloading the database since its all being done in batches still anyway. 

I still don't see the point of forcing a single machine to synchronously do all the jobs though.

packages/migrations/README.md
```diff
@@ -0,0 +1,130 @@
...
+  await enqueueBatchedMigration('20230411002409_example_migration');
+}
+```
+
+This will queue the batched migration for execution.
+
+You may need to ensure that a given batched migration has succeeded before running a subsequent regular migration. For instance, you might have a batched migration that copies a column from one table to another, and you want to ensure that all data has been copied before you delete the original column. You can achieve this by "finalizing" the migration with `finalizeBatchedMigration()`. This will synchronously execute any remaining batches, and will error if the migration ends up in a failed state. This gives you a chance to fix any errors and retry the failed jobs.
+
+```ts
+// migrations/20230411002409_finalize_batched_migration__example_migration.sql
```
SethPoulsen, line 122:
```suggestion
// migrations/20230411002409_finalize_batched_migration__example_migration.ts
```

packages/migrations/README.md
```diff
@@ -0,0 +1,130 @@
...
+startBatchedMigrations({
+  workDurationMs: 60_000,
+  sleepDurationMs: 30_000,
+});
+```
+
+Unlike regular migrations, batched migrations aren't automatically started. Instead, you must write a regular migration to call `enqueueBatchedMigration()` to explicitly start a given batched migration. This provides precise control over execution order.
+
+```ts
+// migrations/20230411002409_start_batched_migration__example_migration.sql
```
SethPoulsen, line 109:
```suggestion
// migrations/20230411002409_start_batched_migration__example_migration.ts
```

packages/migrations/README.md
```diff
@@ -0,0 +1,130 @@
...
+  // Handle error, e.g. by reporting to Sentry.
+});
+
+await init();
+
+startBatchedMigrations({
+  workDurationMs: 60_000,
+  sleepDurationMs: 30_000,
+});
+```
```
SethPoulsen, line 104: I'm not sure if this example is needed/valuable here. I guess it is a question of the difference between documentation for someone wanting to write a batched migration within PrairieLearn, and documentation for someone wanting to use this package in a different system.

As someone that wants to implement a batched migration within PrairieLearn, having this here confused me a little bit because I was wondering if I should be doing this, only to realize I didn't need to do this part because it is already done by PrairieLearn. 

Maybe instead of having the code example here, you could just end the previous paragraph with "for an example, see how this is implemented in PrairieLearn in `server.js`"

nwalters512: I can move this to the end of the document if that would help surface the information that's going to be useful most often.

packages/migrations/src/batched-migrations/batched-migration-runner.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  private async refreshMigrationStatus(migration: BatchedMigrationRow) {
+    this.migrationStatus = await queryValidatedSingleColumnOneRow(
+      sql.get_migration_status,
+      {
+        id: migration.id,
+      },
+      BatchedMigrationStatusSchema
+    );
+  }
```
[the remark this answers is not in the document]
nwalters512: `get*` to me implies that it returns something, which this does not.

packages/migrations/src/batched-migrations/batched-migrations-runner.ts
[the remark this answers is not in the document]
nwalters512: Automatically finishing them is extremely useful for anyone doing local development or self-hosting a single-server installation. If we *didn't* do that, you'd have to very carefully plan a list of revisions to slowly step through, which is highly undesirable. Introducing finalization allows us to maintain the very nice property that one can generally just do an upgrade via `git pull && make start` and have a functioning server with zero intervention.

packages/migrations/src/batched-migrations/batched-migration.ts
```diff
@@ -0,0 +1,129 @@
...
+export interface BatchedMigrationParameters {
+  min?: bigint | null;
+  max: bigint | null;
+  batchSize?: number;
+}
+
+export interface BatchedMigrationImplementation {
+  getParameters(): Promise<BatchedMigrationParameters>;
+  execute(start: bigint, end: bigint): Promise<void>;
+}
```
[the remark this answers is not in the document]
nwalters512: In short: because "batched migration" throughout this code means both "a row in the `batched_migrations` table" and "the code for a specific batched migration" and I wanted the naming to reflect that fact. I'm open to changing that, but maybe there's a better name than just `BatchedMigration`?

packages/migrations/src/batched-migrations/batched-migration-runner.ts
nwalters512: Yep! I accidentally dropped a `return` during a refactor. Restored in `c709898`.

packages/migrations/README.md
nwalters512: Resolved in 2f0f361.

packages/migrations/README.md
nwalters512: Resolved in 2f0f361.

packages/migrations/README.md
nwalters512: How do the changes from 85c31ca look?

server.js
mwest1066: I'm slightly tempted to have a separate config option just to give us more control, but I will admit I don't have a specific use case in mind at the moment. So either way is fine by me.

packages/migrations/src/batched-migrations/batched-migration.ts
mwest1066: I'm fine with whatever you prefer here, although my bias is certainly against classes 😄

packages/migrations/src/batched-migrations/batched-migration-runner.ts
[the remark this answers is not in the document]
SethPoulsen: oh right, its doing a class state thing

packages/migrations/src/batched-migrations/batched-migration.ts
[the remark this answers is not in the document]
SethPoulsen: That makes enough sense to me.

packages/migrations/src/batched-migrations/batched-migrations-runner.ts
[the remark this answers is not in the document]
SethPoulsen: Ok this makes sense. I think if the local dev was the only use case for it, there would be a simpler solution, but I agree this seems very desirable for running with a single instance in production set up.

packages/migrations/README.md
SethPoulsen: Looks good! I still sort of think it would be nice to reference the implementation in "server.js" so people know where to look if they want to see it actually implemented but thats not critical.

database/enums/enum_batched_migration_status.pg
nwalters512: @SethPoulsen is there still an outstanding question here, or did we clear this up in the other thread?

server.js
nwalters512: `config.runBatchedMigrations` added in b0e45f0.

mwest1066: Excellent point.

pages/administratorBatchedMigrations/administratorBatchedMigrations.html.js
```diff
@@ -0,0 +1,238 @@
...
+                  <th>Created at</th>
+                  <td>${batchedMigration.created_at.toUTCString()}</td>
+                </tr>
+                <tr>
+                  <th>Updated at</th>
+                  <td>${batchedMigration.updated_at.toUTCString()}</td>
+                </tr>
+                <tr>
+                  <th>Started at</th>
+                  <td>${batchedMigration.started_at.toUTCString()}</td>
```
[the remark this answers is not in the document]
SethPoulsen: now I must debug why my migration didn't generate any jobs...

SethPoulsen:
It seems slightly odd that the only reason one can make migrations with JS is to start or finalize batched migrations. If this is the case, is there simpler solution? or are there other intended use cases?

SethPoulsen:
While testing I had a SQL error in my batched migration, and I had the following error come out in my console:

```
error: Batched migration runner error Do not know how to serialize a BigInt {"data":{"sql":"-- BLOCK finish_batched_migration_job\nUPDATE batched_migration_jobs\nSET\n  status = $status::enum_batched_migration_job_status,\n  updated_at = CURRENT_TIMESTAMP,\n  finished_at = CURRENT_TIMESTAMP,\n  data = $data\nWHERE\n  id = $id;\n","sqlError":{"message":"Do not know how to serialize a BigInt"},"sqlParams":{"data":{"error":{"code":"42703","data":{"sql":"\n      UPDATE variants AS v\n      SET\n        v.course_id = q.course_id\n      FROM\n        questions AS q\n      WHERE\n        v.course_id IS NULL AND\n        v.question_id = q.id AND\n        v.id >= $min AND\n        v.id <= $max","sqlError":{"code":"42703","file":"analyze.c","length":124,"line":"2425","message":"column \"v\" of relation \"variants\" does not exist","name":"error","position":"47","routine":"transformUpdateTargetList","severity":"ERROR"},"sqlParams":{"max":80,"min":1}},"file":"analyze.c","length":124,"line":"2425","message":"column \"v\" of relation \"variants\" does not exist","name":"error","position":"47","routine":"transformUpdateTargetList","severity":"ERROR","stack":"error: column \"v\" of relation \"variants\" does not exist\n    at Parser.parseErrorMessage (/PrairieLearn/node_modules/pg-protocol/dist/parser.js:287:98)\n    at Parser.handlePacket (/PrairieLearn/node_modules/pg-protocol/dist/parser.js:126:29)\n    at Parser.parse (/PrairieLearn/node_modules/pg-protocol/dist/parser.js:39:38)\n    at Socket.<anonymous> (/PrairieLearn/node_modules/pg-protocol/dist/index.js:11:42)\n    at Socket.emit (node:events:513:28)\n    at Socket.emit (node:domain:489:12)\n    at addChunk (node:internal/streams/readable:315:12)\n    at readableAddChunk (node:internal/streams/readable:289:9)\n    at Socket.Readable.push (node:internal/streams/readable:228:10)\n    at TCP.onStreamRead (node:internal/stream_base_commons:190:23)"}},"id":"1","status":"failed"}},"stack":"TypeError: Do not know how to serialize a BigInt\n    at JSON.stringify (<anonymous>)\n    at prepareObject (/PrairieLearn/node_modules/pg/lib/utils.js:81:15)\n    at prepareValue (/PrairieLearn/node_modules/pg/lib/utils.js:66:12)\n    at prepareValueWrapper (/PrairieLearn/node_modules/pg/lib/utils.js:182:12)\n    at writeValues (/PrairieLearn/node_modules/pg-protocol/dist/serializer.js:67:41)\n    at Object.bind (/PrairieLearn/node_modules/pg-protocol/dist/serializer.js:98:5)\n    at Connection.bind (/PrairieLearn/node_modules/pg/lib/connection.js:162:26)\n    at Query.prepare (/PrairieLearn/node_modules/pg/lib/query.js:211:18)\n    at Query.submit (/PrairieLearn/node_modules/pg/lib/query.js:162:12)\n    at Client._pulseQueryQueue (/PrairieLearn/node_modules/pg/lib/client.js:497:45)\n    at Client._handleReadyForQuery (/PrairieLearn/node_modules/pg/lib/client.js:305:10)\n    at Connection.emit (node:events:513:28)\n    at Connection.emit (node:domain:489:12)\n    at /PrairieLearn/node_modules/pg/lib/connection.js:116:12\n    at Parser.parse (/PrairieLearn/node_modules/pg-protocol/dist/parser.js:40:17)\n    at Socket.<anonymous> (/PrairieLearn/node_modules/pg-protocol/dist/index.js:11:42)"}
error: Batched migration runner error invalid message format {"code":"08P01","data":{"sql":"COMMIT","sqlError":{"code":"08P01","file":"pqformat.c","length":81,"line":"642","message":"invalid message format","name":"error","routine":"pq_getmsgend","severity":"ERROR"},"sqlParams":{}},"file":"pqformat.c","length":81,"line":"642","name":"error","routine":"pq_getmsgend","severity":"ERROR","stack":"error: invalid message format\n    at Parser.parseErrorMessage (/PrairieLearn/node_modules/pg-protocol/dist/parser.js:287:98)\n    at Parser.handlePacket (/PrairieLearn/node_modules/pg-protocol/dist/parser.js:126:29)\n    at Parser.parse (/PrairieLearn/node_modules/pg-protocol/dist/parser.js:39:38)\n    at Socket.<anonymous> (/PrairieLearn/node_modules/pg-protocol/dist/index.js:11:42)\n    at Socket.emit (node:events:513:28)\n    at Socket.emit (node:domain:489:12)\n    at addChunk (node:internal/streams/readable:315:12)\n    at readableAddChunk (node:internal/streams/readable:289:9)\n    at Socket.Readable.push (node:internal/streams/readable:228:10)\n    at TCP.onStreamRead (node:internal/stream_base_commons:190:23)"}
error: Error initializing PrairieLearn server: current transaction is aborted, commands ignored until end of transaction block {"code":"25P02","data":{"sql":"INSERT INTO named_locks (name) VALUES ($name) ON CONFLICT (name) DO NOTHING;","sqlError":{"code":"25P02","file":"postgres.c","length":145,"line":"1432","message":"current transaction is aborted, commands ignored until end of transaction block","name":"error","routine":"exec_parse_message","severity":"ERROR"},"sqlParams":{"name":"news_items"}},"file":"postgres.c","length":145,"line":"1432","name":"error","routine":"exec_parse_message","severity":"ERROR","stack":"Async Stacktrace:\n    at Object.<anonymous> (/PrairieLearn/server.js:2125:15)\n\nerror: current transaction is aborted, commands ignored until end of transaction block\n    at Parser.parseErrorMessage (/PrairieLearn/node_modules/pg-protocol/dist/parser.js:287:98)\n    at Parser.handlePacket (/PrairieLearn/node_modules/pg-protocol/dist/parser.js:126:29)\n    at Parser.parse (/PrairieLearn/node_modules/pg-protocol/dist/parser.js:39:38)\n    at Socket.<anonymous> (/PrairieLearn/node_modules/pg-protocol/dist/index.js:11:42)\n    at Socket.emit (node:events:513:28)\n    at Socket.emit (node:domain:489:12)\n    at addChunk (node:internal/streams/readable:315:12)\n    at readableAddChunk (node:internal/streams/readable:289:9)\n    at Socket.Readable.push (node:internal/streams/readable:228:10)\n    at TCP.onStreamRead (node:internal/stream_base_commons:190:23)"}
```

The error reported at the top level there is "Do not know how to serialize a BigInt", which was an error in the migration runner, but the error that I actually cared about is the one that said "column \"v\" of relation \"variants\" does not exist", which is the one that my SQL query generated. Does it make more sense to have the more 'inner' error be the one quoted at the top level in this case? Or will that be too confusing and its better to just leave it with the way PL handles errors in general?

Regardless of what we decide, I definitely don't think this is a blocker for this PR.

nwalters512:
This is a good find! The fact that the "real" error appeared here was a happy accident: `finish_batched_migration_job` will try to record information about the failure, but the failure message itself contained a `BigInt`, which can't be serialized to JSON. This is a flaw in our error-recording machinery, I can and should fix this.

nwalters512:
c662832 adds a replacer to our `JSON.stringify()` call so that we can serialize BigInt values to strings. @SethPoulsen I added a test, but could you rebase on this branch and try the broken migration again to see if you get a more useful error?

SethPoulsen:
No errors in the terminal, but I get this on the batched migration page: 
image

I suppose this was the intended behavior?

nwalters512:
> I suppose this was the intended behavior?

Yep, that's exactly what's supposed to happen! That'll let us review and fix failed jobs.

> It seems slightly odd that the only reason one can make migrations with JS is to start or finalize batched migrations. If this is the case, is there simpler solution? or are there other intended use cases?

I don't have any other specific use cases in mind at the moment.

SethPoulsen (approved):
I think this is all ready to go.

packages/migrations/src/batched-migrations/batched-migrations-runner.ts
mwest1066: Sounds fine to me.

database/tables/batched_migration_jobs.pg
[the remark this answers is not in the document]
mwest1066: Ok, I'm fine without them.

mwest1066 (approved):

PR #7294, opened 2023-03-04
Introduce @prairielearn/named-locks package

nwalters512:
Continuing my efforts to break things apart into independent packages that can be shared across other packages and projects. Prerequisite for #7292.


packages/named-locks/src/index.ts
```diff
@@ -1,17 +1,24 @@
...
+  /** How many milliseconds to wait (anything other than a positive number means forever) */
+  timeout?: number;
+}
 
-/**
- * @typedef {Object} LockOptions
- * @property {number} [timeout] - How many milliseconds to wait (anything other than a positive number means forever)
- */
+type InternalLockOptions =
+  | {
```
mwest1066, line 15: What does the leading `|` do here?

mwest1066 (approved):
I ❤️ ❤️ ❤️ these new packages! (and in TS 🎉)

packages/named-locks/src/index.ts
nwalters512: This is just a Prettier thing - I know it looks a little strange! It's equivalent to if there was no leading `|`.

PR #7298, opened 2023-03-06
Fix pint cache issue

eliotwrobson:
Fixes the pint cache issue, discussed on slack. Summary: If multiple `zygote.py` processes are running on the same server, then this creates a race condition with `pint` cache creation, where a process might try and read a cache file that's still being created. To solve this, we use a separate cache directory for each process, that way, the cache file creation won't interfere with other processes.


jonatanschroeder (approved):
Tested locally, seems to resolve the issue at hand.

jonatanschroeder:
I think CI will only pass once master's image is fully build...

PR #7300, opened 2023-03-06
Introduce @prairielearn/migrations package

nwalters512:
Depends on #7293 and #7294.


packages/migrations/src/index.ts
```diff
@@ -0,0 +1,177 @@
+import fs from 'fs-extra';
+import path from 'path';
+
+// @ts-expect-error -- Update import once https://github.com/PrairieLearn/PrairieLearn/pull/7294 lands
+import namedLocks from '../../../lib/named-locks';
+import { logger } from '@prairielearn/logger';
+import * as sqldb from '@prairielearn/postgres';
+// @ts-expect-error -- Update import once https://github.com/PrairieLearn/PrairieLearn/pull/7293 lands
+import error from '../../../prairielib/lib/error';
```
nwalters512, line 9: TODO: update before merging.

packages/migrations/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@prairielearn/migrations",
+  "version": "1.0.0",
+  "main": "./dist/index.js",
+  "scripts": {
+    "build": "tsc && copyfiles -u 1 \"./src/**/*.sql\" dist",
```
nwalters512, line 6: `tsc` doesn't natively support copying non-TS files, see https://github.com/microsoft/TypeScript/issues/30835. This `copyfiles` util was the best option I could find.

mwest1066: Yuck, but I agree with this solution.

mwest1066 (approved):

packages/migrations/src/index.ts
nwalters512: Done in cb276be.

PR #7304, opened 2023-03-07
Allow unmounting errors to propagate up

nwalters512:
I made two bad assumptions when writing this code:

1. Unmounting will only ever fail if the directory wasn't already mounted.
2. If an error occurred while unmounting, we could safely proceed with the subsequent mount.

In practice, we seem to have a really weird failure case where transient issues outside our control result in a bind mount being unable to be removed (e.g. `EBUSY` errors). If we try to re-mount on top of an existing mount, weird things that I don't fully understand happen, and the kernel locks up for a while.

It's vastly preferable for us to fail to grade/render a question (this can be retried by the student) than it is to get stuck in a state where the kernel is unhappy, so I'm changing this code so that we'll only try to remove a bind mount if we know we created one before, and if it errors, we'll allow that error to bubble up and kill this particular worker as a whole.

### DIFF
--- a/lib/code-caller/code-caller-container.js
+++ b/lib/code-caller/code-caller-container.js
@@ -154,8 +154,6 @@ class CodeCallerContainer {
   }
 
   /**
-   * Wrapper around `createBindMount` that includes instance-specific logs.
-   *
    * @param {string} directory
    * @param {string} mountpoint
    */
@@ -175,7 +173,9 @@ class CodeCallerContainer {
    *
    * @param {string} mountpoint
    */
-  async removeBindMount(mountpoint) {
+  async removeBindMountIfNeeded(mountpoint) {
+    if (!this.hasBindMount) return;
+
     this.debug(`removing bind mount at ${mountpoint}`);
     await instrumented('removeBindMount', async (span) => {
       span.setAttribute('mountpoint', mountpoint);
@@ -198,10 +198,7 @@ class CodeCallerContainer {
 
     this.coursePath = coursePath;
 
-    if (this.hasBindMount) {
-      await this.removeBindMount(this.hostDirectory.path);
-    }
-
+    await this.removeBindMountIfNeeded(this.hostDirectory.path);
     await this.createBindMount(coursePath, this.hostDirectory.path);
   }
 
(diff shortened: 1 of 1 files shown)


lib/code-caller/code-caller-container.js
```diff
@@ -566,7 +562,8 @@ class CodeCallerContainer {
       try {
         await this.removeBindMount(this.hostDirectory.path);
       } catch (e) {
-        // Probably not mounted; swallow error
+        // Probably not mounted; swallow error.
+        logger.error('Error unmounting host directory', e);
```
mwest1066, line 566: Would it be safer to only attempt `removeBindMount` here if `hasBindMount` is true? We could even move the check for `hasBindMount` into the start of `removeBindMout()` so that we can call it unconditionally.

The above code is clearly no worse than it currently is, so feel free to ignore this suggestion and merge.

mwest1066 (approved):

mwest1066:
> It's vastly preferable for us to fail to grade/render a question (this can be retried by the student) than it is to get stuck in a state where the kernel is unhappy, so I'm changing this code so that we'll only try to remove a bind mount if we know we created one before, and if it errors, we'll allow that error to bubble up and kill this particular worker as a whole.

I very much agree that this PR is making the correct tradeoff here.

lib/code-caller/code-caller-container.js
```diff
@@ -195,17 +198,10 @@ class CodeCallerContainer {
 
     this.coursePath = coursePath;
 
-    // If this Docker code caller was used before, we might have an existing
-    // bind mount that we need to remove first. This should have been cleaned
-    // up before, but if it wasn't, we'll do so here to avoid a
-    // `mount point XXX is itself on a OSXFUSE volume` error.
-    try {
+    if (this.hasBindMount) {
       await this.removeBindMount(this.hostDirectory.path);
```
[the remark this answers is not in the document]
nwalters512: Can't hurt! Added in 1bf2db5.

lib/code-caller/code-caller-container.js
nwalters512: Yep, I made this change in ff9cac7.

PR #7308, opened 2023-03-07
Add chevron and lxml type stubs

eliotwrobson:
See title. Changes are to comply with the newly added types.


elements/pl-external-grader-variables/pl-external-grader-variables.py
```diff
@@ -1,5 +1,6 @@
 import chevron
-import lxml
+import lxml.etree
+import lxml.html
```
nwalters512, line 3: 👀 is this an error in the stubs?

elements/pl-hidden-hints/pl-hidden-hints.py
```diff
@@ -16,7 +17,7 @@ def render(element_html: str, data: pl.QuestionData) -> str:
     pl.check_attribs(element, [], [])
 
     # Parse hints from frontend
-    hints: List[Tuple[int, int, Optional[str], str]] = []
+    hints: list[tuple[int, int, Optional[str], str]] = []
```
nwalters512, line 20: Are we going to switch entirely to lowercase types? IMO we should do that all in one go for consistency.

elements/pl-external-grader-variables/pl-external-grader-variables.py
eliotwrobson: I'll have to double check, but I actually think this is correct and the `import lxml` was working before because of some weird Python thing. Most of the imports in PL actually use `import lxml.html` already.

elements/pl-hidden-hints/pl-hidden-hints.py
eliotwrobson: I can bundle that here. There aren't a ton of uses.

elements/pl-external-grader-variables/pl-external-grader-variables.py
eliotwrobson: Looked again, yeah this has to do with the way that `html` and `etree` are configured within `lxml`. Frankly, it's just kindof a weird package.

nwalters512: Gotcha, thanks for checking!

nwalters512 (approved):

python/python_helper_sympy_test.py
```diff
@@ -37,7 +36,7 @@ class TestSympy:
         "acosh(m)",
     ]
 
-    EXPR_LIST: List = [
+    EXPR_LIST: list = [
```
nwalters512, line 39: micro-nit: is declaring this type actually necessary? Seems like this is something that should be easily inferred by type checkers.

eliotwrobson: I think in this instance it isn't, but in general it's a good idea to be _very_ explicit with annotations in Python. Most of the inferences that get made can be overriden really easily by calling a function incorrectly.

I don't want to make a ton of changes to this file here though, it's already being reworked in https://github.com/PrairieLearn/PrairieLearn/pull/7020 and the version there is much easier to read.

PR #7309, opened 2023-03-07
Improve performance of code caller output handling

nwalters512:
When I implemented similar changes in #6994, I neglected to make them for the container code caller. I also realized that we're doing a lot of unnecessary string allocation here, so I changed things to collect an array of data chunks and only join them all together at the end.

### DIFF
--- a/lib/code-caller/code-caller-container.js
+++ b/lib/code-caller/code-caller-container.js
@@ -129,8 +129,8 @@ class CodeCallerContainer {
     this.options = options;
 
     // These will accumulate output from the container.
-    this.outputStdout = '';
-    this.outputStderr = '';
+    this.outputStdout = [];
+    this.outputStderr = [];
     this.outputBoth = '';
 
     // for error logging
@@ -230,8 +230,8 @@ class CodeCallerContainer {
     const callDataString = JSON.stringify(callData);
 
     // Reset output accumulators.
-    this.outputStdout = '';
-    this.outputStderr = '';
+    this.outputStdout = [];
+    this.outputStderr = [];
     this.outputBoth = '';
 
     const deferred = deferredPromise();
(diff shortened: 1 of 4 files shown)


lib/code-caller/code-caller-container.js
```diff
@@ -418,15 +418,15 @@ class CodeCallerContainer {
...
+    this.outputStdout.push(data);
+    if (data.indexOf('\n') >= 0) {
       this._callIsFinished();
     }
     this.debug('exit _handleStdout()');
   }
   _handleStderr(data) {
     this.debug('enter _handleStderr()');
-    this.outputStderr += data;
+    this.outputStderr.push(data);
```
mwest1066, line 429: Where do we do `this.outputStderr.join('')`? I can't see it anywhere.

mwest1066 (approved):

lib/code-caller/code-caller-container.js
nwalters512: We actually never use it at all (I had to double-check). I guess we could remove it entirely, though I like having it around for potential debugging (I can just add `console.log(this.outputStderr.join(''))` to finish handler).

mwest1066: Ok, sounds good, I was just worried we were missing it.

lib/code-caller/code-caller-container.js
```diff
@@ -401,6 +403,8 @@ class CodeCallerContainer {
     this.stderrStream = new MemoryStream();
     this.stdinStream.pipe(stream);
     this.container.modem.demuxStream(stream, this.stdoutStream, this.stderrStream);
+    this.stdoutStream.setEncoding('utf8');
+    this.stderrStream.setEncoding('utf8');
```
nwalters512, line 407: This was necessary to ensure we get strings in the handlers, which is consistent with how the native code caller works.

PR #7319, opened 2023-03-09
fix: navbar always show logout option

trombonekenny:
This small change removes the conditional on showing the logout link in the navbar so that it is never disabled.

I ran into a case in development where my auth cookie was no longer valid and I noticed the error page didn't give me the logout link to easily remove the cookie. When I traced through the partials, I found I had a case for the only conditional in that partial (maybe there were originally more?), so I simplified it just to always show the link in `navbar.ejs`

### DIFF
--- a/pages/partials/logout.ejs
+++ b/pages/partials/logout.ejs
@@ -1,5 +1,0 @@
-<% if (typeof authn_user != 'undefined') { %>
-   <a class="dropdown-item" href="<%= plainUrlPrefix %>/logout">Log out</a>
-<% } else { %>
-   <a class="dropdown-item disabled">Log out</a>
-<% } %>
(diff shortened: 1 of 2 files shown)


nwalters512 (approved):
AFAIK the navbar is generally only shown on pages where the user is authenticated (with the exception of an auth error page), so this should be fine.

Perhaps we should automatically remove any invalid auth cookie if we find one? I haven't fully thought through the implications of that.

trombonekenny:
> Perhaps we should automatically remove any invalid auth cookie if we find one? I haven't fully thought through the implications of that.

This was a pretty rare edge case. I don't think it's worth designing too much into as long as there are easy ways to reset.

PR #7321, opened 2023-03-09
Make data argument optional in @prairielearn/error make()

nwalters512:
This only came to my attention when I tried to install this package into PrairieTest, which is fully written in TypeScript.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,24 +1,14 @@
 name: Release
 
-# We want this workflow to run on master, but only after we've finished the
-# main CI workflow. This ensures that, for instance, if integration tests
-# fail on master, we don't accidentally publish broken versions of pacakges.
-#
-# See the `if` condition on the `release` job for the second half of this logic.
 on:
-  workflow_run:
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -1,24 +1,14 @@
 name: Release
 
-# We want this workflow to run on master, but only after we've finished the
-# main CI workflow. This ensures that, for instance, if integration tests
-# fail on master, we don't accidentally publish broken versions of pacakges.
-#
-# See the `if` condition on the `release` job for the second half of this logic.
```
nwalters512, line 7: This wasn't providing much additional safety, as the PR that this action opens also runs through CI again before publishing. Unless we're doing something super bad like pushing straight to master or merging before CI finishes, this check only slows us down, and if we _do_ do those bad things, we deserve any bad consequences.

mwest1066: I agree with this. Especially with the new merge queue we shouldn't every have skipped CI.

mwest1066 (approved):

PR #7323, opened 2023-03-09
Disable lib checking in tsconfig.json

nwalters512:
Locally, this cut the time for a `yarn turbo build --force` in half from ~8 seconds to ~4 seconds. This should help a lot on CI where we have fewer CPUs to parallelize things.


nwalters512:
Looks like this resulted in a 30s decrease in build time!

mwest1066 (approved):

PR #7328, opened 2023-03-09
Capture Postgres pool metrics

nwalters512:
This should allow us to better understand how our queries are performing under load. Specifically, this will give us insight into how queries are queueing.

Here's what these look line on Honeycomb. Once I finally limited my local connection pool to 2 clients, you can see that we start measuring waiting clients!

Screenshot 2023-03-09 at 16 12 23

### DIFF
--- a/graders/python/python_autograder/pl_result.py
+++ b/graders/python/python_autograder/pl_result.py
@@ -93,7 +93,7 @@ class PLTestResult(unittest.TestResult):
                 ipynb_key=Feedback.test.ipynb_key,
             )
         else:
-            tr_message = ''.join(traceback.format_exception(*err))
+            tr_message = "".join(traceback.format_exception(*err))
 
             if isinstance(test, unittest.suite._ErrorHolder):
                 # Error occurred outside of a test case, like in setup code for example
--- a/packages/opentelemetry/src/metrics.ts
+++ b/packages/opentelemetry/src/metrics.ts
@@ -92,7 +92,7 @@ export async function instrumentedWithMetrics<T>(
   }
 }
 
-export interface CreateObservableRangeGaugeOptions extends MetricOptions {
+export interface createObservableValueGaugesOptions extends MetricOptions {
   interval: number;
 }
 
@@ -107,10 +107,10 @@ export interface CreateObservableRangeGaugeOptions extends MetricOptions {
  * - `${name}.max`
  * - `${name}.avg`
  */
-export async function createObservableRangeGauge(
+export async function createObservableValueGauges(
   meter: Meter,
   name: string,
-  options: CreateObservableRangeGaugeOptions,
+  options: createObservableValueGaugesOptions,
   observe: () => number
 ) {
   const { interval, ...metricOptions } = options;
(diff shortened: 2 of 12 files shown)


packages/opentelemetry/src/metrics.ts
```diff
@@ -91,3 +91,73 @@ export async function instrumentedWithMetrics<T>(
...
+ * time. The value is observed on a regular interval.
+ *
+ * The provided {@link name} is used as the base name for the three gauges. The
+ * names of the individual gauges are:
+ *
+ * - `${name}.min`
+ * - `${name}.max`
+ * - `${name}.avg`
+ */
+export async function createObservableRangeGauge(
```
nwalters512, line 110: Bikeshedding: any ideas for a better name?

graders/python/python_autograder/pl_result.py
```diff
@@ -93,7 +93,7 @@ def addError(self, test, err):
                 ipynb_key=Feedback.test.ipynb_key,
             )
         else:
-            tr_message = ''.join(traceback.format_exception(*err))
+            tr_message = "".join(traceback.format_exception(*err))
```
nwalters512, line 96: This was changed when I ran `make format`. It wasn't caught previously as we don't mount `images` into the Docker container, so our Python linters never run on them.

packages/opentelemetry/src/metrics.ts
```diff
@@ -91,3 +91,77 @@ export async function instrumentedWithMetrics<T>(
...
+  let min = 0;
+  let max = 0;
+  let sum = 0;
+  let count = 0;
+
+  // Observe the value on a regular interval. Black-hole any errors.
+  const intervalId = setInterval(() => {
+    Promise.resolve(observe())
+      .then((value) => {
+        min = Math.min(min, value);
```
mwest1066, line 127: With the initialized value `min = 0` we'll never update to any other value.
```suggestion
        min = (count == 0 ? value : Math.min(min, value));
```
Alternatively we could keep the existing update rule but initialize to `min = Number.MAX_VALUE` or `min = Infinity`. However, I worry how this will interact with gauges where the underlying type is `ValueType.INT` and so I think it's safer to keep the initialization as `min = 0` but change the update rule as above.

mwest1066 (approved):

packages/opentelemetry/src/metrics.ts
mwest1066: These both sound fine to me (or at least, equally unclear 😄) and I can't think of anything better.

packages/opentelemetry/src/metrics.ts
nwalters512: Good catch! I like your solution here.

packages/opentelemetry/src/metrics.ts
nwalters512: `observeValueStats` maybe? 🤔

mwest1066: Honestly I'm fine with your current name in the code.

PR #7329, opened 2023-03-10
Add workspace log when terminating hosts

nwalters512:
This may give us better visibility into cases where workspaces launch twice.

### DIFF
--- a/cron/workspaceHostTransitions.sql
+++ b/cron/workspaceHostTransitions.sql
@@ -79,4 +79,8 @@ SELECT
   tw.state,
   'Host instance was not found'
 FROM
-  terminated_workspaces AS tw;
+  terminated_workspaces AS tw
+RETURNING
+  workspace_id,
+  state,
+  message;
--- a/cron/workspaceHostTransitions.sql
+++ b/cron/workspaceHostTransitions.sql
@@ -72,12 +72,11 @@ WITH
       w.*
   )
 INSERT INTO
-  workspace_logs (date, workspace_id, version, state, message)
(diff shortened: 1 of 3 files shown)


cron/workspaceHostTransitions.sql
```diff
@@ -56,14 +56,28 @@ WITH
...
   )
-UPDATE workspaces AS w
-SET
-  state = 'stopped',
-  stopped_at = NOW(),
-  state_updated_at = NOW()
+INSERT INTO
+  workspace_logs (date, workspace_id, version, state, message)
+SELECT
+  now(),
```
mwest1066, line 77: Not that it matters, but all of our `_logs` tables have `DEFAULT now()` for the `date` so we don't have to actually set it.

cron/workspaceHostTransitions.sql
```diff
@@ -56,14 +56,28 @@ WITH
...
+  tw.id,
+  tw.version,
+  tw.state,
+  'Host was terminated'
```
mwest1066, line 81:
```suggestion
  'Host was terminated because instance was not found'
```

mwest1066 (approved):

mwest1066 (approved):

cron/workspaceHostTransitions.sql
nwalters512: Tweaked in 0ecabcc.

cron/workspaceHostTransitions.sql
nwalters512: Resolved in 0ecabcc.

PR #7330, opened 2023-03-10
Log actual originalUrl after workspace proxy errors

nwalters512:


middlewares/selectAndValidateWorkspace.sql
```diff
@@ -1,8 +0,0 @@
--- BLOCK select_workspace_hostname
```
nwalters512, line 1: This file should have been removed with #6260.

server.js
```diff
@@ -338,8 +338,13 @@ module.exports.initExpress = function () {
   app.use('/pl/workspace/:workspace_id/container', [
     cookieParser(),
     (req, res, next) => {
-      // Needed for workspaceAuthRouter and `selectAndValidateWorkspace` middleware.
+      // Needed for workspaceAuthRouter.
       res.locals.workspace_id = req.params.workspace_id;
+
+      // Needed for logs; this will retain the original URL if we end up
+      // rewriting the URL in `pathRewrite` above.
+      req.originalUrl = req.url;
```
mwest1066, line 346: I thought express automatically set `req.originalUrl` for us? (see [the docs](https://expressjs.com/en/api.html#req.originalUrl))

server.js
```diff
@@ -304,7 +304,7 @@ module.exports.initExpress = function () {
       logger.error(`Error proxying workspace request: ${err}`, {
         err,
         url: req.url,
-        originalUrl: req.url,
+        originalUrl: req.originalUrl,
```
mwest1066, line 307: I just read this line in the code and somehow I never saw this!!!

mwest1066 (approved):

server.js
nwalters512: Ah, true! Good catch.

nwalters512: Resolved in 2fe82a9.

PR #7334, opened 2023-03-11
Update yarn.lock after changeset versioning

nwalters512:
A recent merge to master failed because `yarn.lock` wasn't in sync after running `changeset version` (https://github.com/changesets/action/issues/170). We now run `yarn` after `changeset version` to make any necessary changes to `yarn.lock` automatically.


mwest1066 (approved):
Good find!

PR #7335, opened 2023-03-11
Update sql-formatter to get FOR UPDATE support

nwalters512:
Pulls in my changes from https://github.com/sql-formatter-org/sql-formatter/pull/574.


jonatanschroeder (approved):
I guess not many queries are updated because most of the "FOR UPDATE" cases are in sprocs, but this makes sense to me.

PR #7336, opened 2023-03-11
Add support for Postgres cursors

nwalters512:
I had some time to kill at airports and on flights last night! This will be used to reduce memory requirements for large queries in #7284 or a followup PR.


packages/postgres/README.md
```diff
@@ -156,6 +156,41 @@ const { user, course } = await sqldb.runInTransactionAsync(async () => {
...
 
+### Cursors
+
+For very large queries that don't need to fit in memory all at once, it's possible to use a cursor to read a limited number of rows at a time.
+
+```ts
+import { queryCursor } from '@prairielearn/postgres';
+
+const cursor = await queryCursor(sql.select_all_users, {});
+for await (const users of cursor.iterate(100)) {
```
nwalters512, line 167: I played around with the API for this for a while, and this is the best thing I landed on. Using async iterators means we can free the client from having to worry about careful error handling and client releasing.

Rather than return one row per iteration, I opted to return the full batch. Since every `await` of the loop would need a tick of the event loop, we can make things a little more performant by processing larger batches synchronously.

packages/postgres/README.md
```diff
@@ -156,6 +156,41 @@ const { user, course } = await sqldb.runInTransactionAsync(async () => {
...
+```ts
+import { z } from 'zod';
+import { queryCursor } from '@prairielearn/postgres';
+
+const UserSchema = z.object({
+  id: z.string(),
+  name: z.string(),
+});
+
+const cursor = await queryCursor(sql.select_all_users, {}, UserSchema);
```
nwalters512, line 186: Rather than a separate `queryValidatedCursor` function, I opted to just make the model an optional third argument. This made sense to me because the function will yield pure arrays instead of a results object, so we can treat the validated case as basically the same.

packages/postgres/package.json
```diff
@@ -18,6 +18,10 @@
   "dependencies": {
     "@types/debug": "^4.1.7",
     "@types/lodash": "^4.14.191",
+    "@types/pg-cursor": "^2.7.0",
+    "pg": "^8.10.0",
+    "pg-cursor": "^2.9.0",
+    "pg-pool": "^3.6.0",
```
nwalters512, line 24: `pg` and `pg-pool` should have been dependencies all along.

tests/postgres.test.js
```diff
@@ -0,0 +1,123 @@
...
+const { queryAsync, queryCursor } = require('@prairielearn/postgres');
+const { z, ZodError } = require('zod');
+
+chai.use(chaiAsPromised);
+const { assert } = chai;
+
+const helperDb = require('./helperDb');
+
+// TODO: move this into the `@prairielearn/postgres` package once the database
+// helpers are in a shared package.
```
nwalters512, line 13: I need to bite the bullet and do this soon.

mwest1066 (approved):

packages/postgres/README.md
nwalters512: Ahh, I buy that rationale (though in theory we can *also* write a lint rule to detect 2-argument calls to `queryCursor`. I'll make this change regardless.

packages/postgres/src/pool.ts
```diff
@@ -831,6 +840,52 @@ export class PostgresPool {
...
+    params: Params,
+    model?: Model
+  ): Promise<CursorIterator<z.infer<Model>>> {
+    const client = await this.getClientAsync();
+    const cursor = await this.queryCursorWithClient(client, sql, params);
+
+    let iterateCalled = false;
+    return {
+      iterate: async function* (batchSize: number) {
+        if (iterateCalled) {
```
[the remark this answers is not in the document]
nwalters512: It's just a safety check, and it's also just an oversight on my part that we never set it to true. Will fix!

packages/postgres/src/pool.ts
```diff
@@ -114,6 +119,17 @@ function escapeIdentifier(identifier: string): string {
   return pg.Client.prototype.escapeIdentifier(identifier);
 }
 
+function enhanceError(err: Error, sql: string, params: Params): Error {
+  // TODO: why do we do this?
```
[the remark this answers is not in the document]
nwalters512: I found out why we do this: it's basically to make a copy of the object. Otherwise, we end up with a circular reference in the final `err` object. I'm going to change this to just use `{ ...err }` instead.

packages/postgres/README.md
nwalters512: Changed in 8e5d612.

packages/postgres/src/pool.ts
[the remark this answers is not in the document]
nwalters512: Changed in 8e5d612.

packages/postgres/src/pool.ts
```diff
@@ -1,12 +1,17 @@
...
 import pg, { QueryResult } from 'pg';
+import Cursor from 'pg-cursor';
 import path from 'node:path';
 import debugFactory from 'debug';
 import { callbackify } from 'node:util';
 import { AsyncLocalStorage } from 'node:async_hooks';
 import { z } from 'zod';
 
-type Params = Record<string, any> | any[];
+export type Params = Record<string, any> | any[];
```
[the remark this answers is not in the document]
nwalters512: Good call, done in 8e5d612.

packages/postgres/src/pool.ts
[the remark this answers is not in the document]
nwalters512: Fixed in 09703a9.

PR #7338, opened 2023-03-11
Review student view of question status and score

### LINKED ISSUE #7311: Manually-graded instance question shows 0% despite having full points
@ffund reported this issue on Slack: https://prairielearn.slack.com/archives/C266KEH9A/p1678296976199299

Relevant assessment instance: https://us.prairielearn.com/pl/course_instance/129723/assessment_instance/5355820

On the assessment overview page, question 15 is shown as having 0%, even though the exam as a whole has 100/100 points:

image

There are inconsistencies on the instance question page as well: the question is shown as having 10/10 points, and the submission itself is shown as 100% correct, but the "submission status" percentage is still 0:

Screenshot 2023-03-08 at 10 26 53
Screenshot 2023-03-08 at 10 26 58

jonatanschroeder:
Resolves #7311. Separates the submission status view from the best submission view for clarity particularly in assessments with a mix of manual and auto-grading points.

Exam with mix of auto-graded and manual graded, overview:

Before:
image

After:
image

---

Assessment question manually graded:

Before:
image

After:
image


pages/partials/examQuestionStatus.ejs
```diff
@@ -0,0 +1,20 @@
...
+    unanswered: 'danger',
+    invalid: 'danger',
+    grading: 'default',
+    saved: 'info',
+    complete: 'success',
+    correct: 'success',
+    incorrect: 'danger',
+  };
+  %>
+  <span class="badge badge-<%= badge_color[instance_question.status] ?? 'default' %>"><%= instance_question.status %></span>
```
jonatanschroeder, line 13: One aspect that may benefit from a review here is the distinction between incorrect, correct and complete in these cases. Given these are not often used in other contexts, they have been loosely updated. In essence:

* For exams: complete is a question that is correct and has no manual points, or with exhausted number of attempts. Correct is used for a question that has 100% auto-points but still has manual points, while incorrect is used for incorrect answers with attempts left.
* For homeworks: complete is not used. Correct is used if at least one submission got 100% auto-grading score. Incorrect is used otherwise.
* Manual grading right now always updates the status to complete.

We may want to revisit how these statuses are updated, or group these three into a different status, at least for student display. Other than this case, there is no other code that I could find (though I'm still looking a bit more thoroughly where this distinction is used, though there are cases where the other status values are important.

jonatanschroeder:
I may still include in this PR an update of the screenshots shown in the docs (https://prairielearn.readthedocs.io/en/latest/assessment/#disabling-real-time-grading).

pages/partials/examQuestionStatus.ejs
jonatanschroeder: After meeting with Matt, here's the decision:
* for homeworks, complete designates a question where max points have been achieved, or where manual grading has been done.
* everything else remains as is.

jonatanschroeder: Done in 49051d632.

mwest1066 (approved):
Many apologies for taking far too long to review this! I didn't exhaustively test the new functionality but the code looks good,  I remember the discussion about this issue, and I agree with the new UI to separate out the highest submission score from the question status.

PR #7340, opened 2023-03-13
Replace moment with date-fns

### LINKED ISSUE #4684: Change effective date - use course/instance timezone
In the page to change effective user/role/etc, there is an option to change the effective date. This date defaults to being formatted in UTC. I think it may be easier for instructors if it defaults to the course instance timezone instead.

### LINKED ISSUE #7163: Replace use of moment.js with other libraries
Some parts of the code use moment.js for date processing. However, even [the library's own docs](https://momentjs.com/docs/) recommend using a different library instead, given the relatively heavy library and some limitations related to garbage collection. Since `date-fns` is already being used in other parts of the code, it is probably a good alternative.

jonatanschroeder:
Resolves #7163, replaces the use of moment.js with equivalent functions in date-fns.js.

Also resolves #4684.

### DIFF
--- a/lib/question.js
+++ b/lib/question.js
@@ -5,7 +5,7 @@ const path = require('path');
 const jsonStringifySafe = require('json-stringify-safe');
 const debug = require('debug')('prairielearn:' + path.basename(__filename, '.js'));
 const ejs = require('ejs');
-const moment = require('moment');
+const { differenceInMilliseconds, parseISO } = require('date-fns');
 const fs = require('fs');
 const unzipper = require('unzipper');
 const fg = require('fast-glob');
@@ -1772,17 +1772,18 @@ module.exports = {
   _buildGradingJobStats(job) {
     if (job) {
       const phases = [];
-      const totalDuration = moment
-        .duration(moment(job.grading_requested_at).diff(job.graded_at))
-        .asMilliseconds();
+      const totalDuration = differenceInMilliseconds(
+        parseISO(job.graded_at),
+        parseISO(job.grading_requested_at)
+      );
       const formatDiff = (start, end, addToPhases = true) => {
-        const duration = moment.duration(moment(end).diff(start)).asMilliseconds();
+        const duration = differenceInMilliseconds(parseISO(end), parseISO(start));
         if (addToPhases) {
-          const percentage = -1 * (duration / totalDuration);
+          const percentage = duration / totalDuration;
           // Round down to avoid width being greater than 100% with floating point errors
           phases.push(Math.floor(percentage * 1000) / 10);
         }
-        return moment.utc(duration).format('s.SSS');
+        return (duration / 1000).toFixed(3).replace(/\.?0+$/, '');
       };
 
       return {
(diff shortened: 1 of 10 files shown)


jonatanschroeder:
I might just go through all references of moment...

lib/question.js
```diff
@@ -1772,17 +1772,18 @@ module.exports = {
...
       const formatDiff = (start, end, addToPhases = true) => {
-        const duration = moment.duration(moment(end).diff(start)).asMilliseconds();
+        const duration = differenceInMilliseconds(parseISO(end), parseISO(start));
         if (addToPhases) {
-          const percentage = -1 * (duration / totalDuration);
+          const percentage = duration / totalDuration;
           // Round down to avoid width being greater than 100% with floating point errors
           phases.push(Math.floor(percentage * 1000) / 10);
         }
-        return moment.utc(duration).format('s.SSS');
```
jonatanschroeder, line 1785: I believe this was actually a bug. If the duration was longer than a minute, this would return only the seconds component of the time, not the minutes.

pages/instructorEffectiveUser/instructorEffectiveUser.ejs
```diff
@@ -20,7 +20,7 @@
           <p><strong>Authenticated course instance role:</strong> <%= authz_data.authn_course_instance_role %></p>
           <% } %>
           <p><strong>Authenticated mode:</strong> <%= authz_data.authn_mode %></p>
-          <p><strong>Authenticated date:</strong> <%= req_date_for_display %></p>
+          <p><strong>Authenticated date:</strong> <%= true_req_date_for_display %></p>
```
jonatanschroeder, line 23: This was also a bug, the "authenticated date" was showing the overridden date.

lib/question.js
nwalters512: Nice catch!

pages/instructorEffectiveUser/instructorEffectiveUser.js
```diff
@@ -43,7 +44,20 @@ router.get('/', function (req, res, next) {
     if (res.locals.ipaddress.substr(0, 7) === '::ffff:') {
       res.locals.ipaddress = res.locals.ipaddress.substr(7);
     }
-    res.locals.req_date_for_display = moment(res.locals.req_date).toISOString(true);
+    res.locals.true_req_date_for_display = format(
+      utcToZonedTime(res.locals.true_req_date, res.locals.course_instance.display_timezone),
```
nwalters512, line 48: Was about to ask if we should be coalescing with `course.display_timezone`, but then I remembered that you discovered in https://github.com/PrairieLearn/PrairieLearn/issues/7344 that `course_instance_.display_timezone` is effectively non-null!

yarn.lock
```diff
@@ -10291,7 +10300,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"moment@npm:^2.19.3, moment@npm:^2.29.4":
+"moment@npm:^2.19.3":
```
nwalters512, line 10303: Dang, looks like this is still transitively required through the logging library that [`passport-azure-ad`](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/maintenance/passport-azure-ad) uses.

nwalters512 (approved):

Issue #7000, opened 2023-01-27
Feature: Serialization of `sets` and `tuples`


coatless:
Need to think about ways to incorporate `sets` and `tuples` in JSON serialization. 

Related to https://github.com/PrairieLearn/PrairieLearn/pull/6953#discussion_r1088479325

echuber2:
Does `dict` work already? Sets could be stored as dict keys with placeholder values, or stored as a list with a set tag. Tuples could be handled as a list too.

eliotwrobson:
Dict already works, but there might be issues with recursive definitions

eliotwrobson:
Looking at this again, I'm wondering whether it's worth doing this as opposed to including instructions on how to format + store a formatted string for an object using pprint.

In fact, the existing python variable element is really just a thin wrapper around the pl-code element and the pprint library. It may be worth deprecating this element in favor of providing instructions on how to format Python variables for output to the user. Thinking about some use cases, it may actually be less verbose to use pl-code than pl-python-variable.

echuber2:
It may be useful to have serialization for these other types in general. I know general-case serialization (e.g. of functions, classes, environment state, etc.) is a bag of worms for security.

eliotwrobson:
@echuber2 I'm definitely not against some generic socialization if other elements would use it and instructors find it painful. But implementing that for just this element increases the maintenance burden for something that may be worth deprecating to begin with.

PR #7341, opened 2023-03-14
feat: Add complex object display example and copy code button to `pl-python-variable`

eliotwrobson:
See title. Affirmatively resolves https://github.com/PrairieLearn/PrairieLearn/issues/7000 by creating an example of how to display sets and tuples without having to serialize them as JSON (which is it's own can of worms). Also, adds the copy code button to `pl-python-variable` via a passthrough. Updates documentation accordingly.


elements/pl-python-variable/pl-python-variable.py
```diff
@@ -114,4 +119,4 @@ def render(element_html: str, data: pl.QuestionData) -> str:
 
     language = "" if no_highlight else ' language="python"'
 
-    return f"<pl-code{language}>{prefix}{var_string}{suffix}</pl-code>"
+    return f'<pl-code{language} copy_code_button="{copy_code_button}">{prefix}{var_string}{suffix}</pl-code>'
```
jonatanschroeder, line 122:
```suggestion
    return f'<pl-code{language} copy-code-button="{copy_code_button}">{prefix}{var_string}{suffix}</pl-code>'
```

exampleCourse/questions/element/code/server.py
```diff
@@ -0,0 +1,12 @@
...
+
+def generate(data):
+    complex_object = {
+        (1, 2): "hi",
+        (0, 0): "origin",
+        (2, 2): "another point",
+        (3, 3): "yet another point",
+    }
+
+    data["params"]["my_object_string"] = pprint.pformat(complex_object)
```
coatless, line 12: Could we build the conversion to a string into the serialization part for the tuple & set object? This would negate needing to give a "custom" example.

eliotwrobson: I think doing that has the potential to cause more problems than it's worth. For example, people would need some way to put in custom string conversion options (like indentation size), and could be confusing if objects are being silently converted behind the scenes.

docs/elements.md
```diff
@@ -1264,23 +1264,24 @@ def generate(data):
...
 
 #### Details
 
-The element supports displaying Python objects via `repr()`, with support for more complex display options similar to the built-in pprint library. Objects to be displayed must be JSON serialized. For details about what objects can be serialized and how to do this with the provided `to_json` and `from_json` functions, see the [Question Writing documentation](question.md#question-data-storage).
+The element supports displaying Python objects via `repr()`, with support for more complex display options similar to the built-in `pprint` library. **Objects to be displayed must be JSON serialized.** For details about what objects can be serialized and how to do this with the provided `to_json` and `from_json` functions, see the [Question Writing documentation](question.md#question-data-storage). To display objects that cannot be easily JSON serialized, please refer to the `pl-code` example question [element/code].
```
nwalters512, line 1284: Nit, I think this is subtly more correct?

```suggestion
The element supports displaying Python objects via `repr()`, with support for more complex display options similar to the built-in `pprint` library. **Objects to be displayed must be serializable to JSON.** For details about what objects can be serialized and how to do this with the provided `to_json` and `from_json` functions, see the [Question Writing documentation](question.md#question-data-storage). To display objects that cannot be easily JSON serialized, please refer to the `pl-code` example question [element/code].
```

Aside: should we refer to the thing to be displayed as "python values" instead of "python objects"? I guess "objects" is still correct, but colloquially I think "object" isn't often used to refer to things like lists or primitives.

nwalters512 (approved):

docs/elements.md
eliotwrobson: I always refer to lists or sets as either collections or objects, since they have methods. In this case however, there are some types like dataframes or numpy arrays that can get displayed, so I think it makes sense to use very general language.

PR #7342, opened 2023-03-14
C Grader: better support for AddressSanitizer

jonatanschroeder:
This PR introduces simpler support for the use of AddressSanitizer (and by consequence LeakSanitizer) to detect memory problems. While the main functionality was possible before using flags, this code introduces a simpler interface for it, as well as documentation and some controls to make it harder for students to disable these checks.


jonatanschroeder:
Upon reviewing this code, it seems it might be possible for students to bypass some of these checks using ASAN calls themselves (e.g., [by adding attribute annotations](https://clang.llvm.org/docs/AddressSanitizer.html#disabling-instrumentation-with-attribute-no-sanitize-address)). I'll take a look to see if I can provide some way to disable these calls in student code. Marking this PR as draft until I can get this working reliably.

graders/c/cgrader/cgrader.py
```diff
@@ -113,10 +142,23 @@ def compile_file(
...
             out += self.run_command(
                 [compiler, "-c", std_c_file, "-o", obj_file] + cflags, sandboxed=False
             )
-            std_obj_files.append(obj_file)
+            # Identify references to functions intended to disable sanitizers from object file
+            if os.path.isfile(obj_file):
+                symbols = self.run_command(
+                    ["nm", "-j", obj_file], sandboxed=False
+                ).splitlines()
+                found_symbols = [s for s in INVALID_SYMBOLS if s in symbols]
```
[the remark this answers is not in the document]
jonatanschroeder: Done in c61325073. This issue of empty lists/sets is always confusing, since JS and Python assign different boolean interpretations for them (in JS an empty list/dict is true).

jonatanschroeder:
The solution used here is not the cleanest, but should take care of the most common cases of student attempts to disable sanitizers.

docs/c-grader/index.md
```diff
@@ -330,6 +330,46 @@ If your application explicitly needs to keep any of the restricted environments
...
+```
+
+If you are using the check-based workflow, note that while the setup above will cause the tests to fail in these scenarios, it may not provide a useful message to students. To provide a more detailed feedback to students in this case, you are strongly encouraged to add a call to `pl_setup_asan_hooks()` at the start of your main function, like this:
+
+```c
+int main(int argc, char *argv[]) {
+
+  pl_setup_asan_hooks();
+  Suite *s = suite_create(...
+
```
nwalters512, line 356: Nit (GitHub syntax highlighting will hopefully be happier with this?)
```suggestion
  Suite *s = suite_create(...);
```

exampleCourse/questions/demo/autograder/c/linkedListDelete/tests/main.c
```diff
@@ -0,0 +1,83 @@
...
+    // Frees node to avoid a memory leak detected in the test
+    free(nodes[i]);
+  }
+  // Frees list to avoid a memory leak detected in the test
+  free(list);
+}
+
+int main(int argc, char *argv[]) {
+
+  __asan_set_error_report_callback(asan_abort_hook);
```
nwalters512, line 65: Can/should this use `pl_setup_asan_hooks` instead?

exampleCourse/questions/demo/autograder/c/linkedListDelete/tests/test.py
```diff
@@ -0,0 +1,20 @@
...
+import cgrader
+
+
+class TestGrader(cgrader.CGrader):
+    def tests(self):
+        self.test_compile_file(
+            "deletefirst.c",
+            "main",
+            add_c_file="/grade/tests/main.c",
+            flags="-I/grade/tests -fsanitize=address -static-libasan -g",
```
nwalters512, line 12: Any reason you're manually passing the flags instead of using `enable_asan=True`?

graders/c/cgrader/check.h
```diff
@@ -73,10 +78,17 @@ static inline void pl_fixture_sandbox_setup(void) {
   // Clear all environment variables, including those used to configure libcheck
   clearenv();
 #endif
+
+  if (*sanitizer_output) {
+    mkstemp(sanitizer_output);
```
nwalters512, line 83: Doesn't the `template` argument to `mkstemp` need to include `XXXXXX` as the last six characters? AFAICT `sanitizer_output` is the empty string.

jonatanschroeder: This function is called if `*sanitizer_output` has been initialized in `pl_setup_asan_hooks`, in which case this string will have the required XXXXXX. The idea is that this call is only done if ASAN hooks are set up, and the value of `*sanitizer_output` is a way to check if that is the case.

I originally tried to initialize the temp directory there, but the call to `closefrom` a few lines above interferes with that.

exampleCourse/questions/demo/autograder/c/linkedListDelete/tests/test.py
jonatanschroeder: I updated the interface and forgot to update the example question 😬 ...

jonatanschroeder: Done in ac4c7a83c (also the call to `pl_setup_asan_hooks`).

graders/c/cgrader/check.h
jonatanschroeder: Technically because `*sanitizer_output == '\0'`, but yes, that's it.

docs/c-grader/index.md
```diff
@@ -330,6 +330,46 @@ If your application explicitly needs to keep any of the restricted environments
...
+```
+
+If you are using the check-based workflow, note that while the setup above will cause the tests to fail in these scenarios, it may not provide a useful message to students. To provide a more detailed feedback to students in this case, you are strongly encouraged to add a call to `pl_setup_asan_hooks()` at the start of your main function, like this:
+
+```c
+int main(int argc, char *argv[]) {
+
+  pl_setup_asan_hooks();
+  Suite *s = suite_create(...);
+
```
nwalters512, line 356: One more missing character:

```suggestion
}
```

nwalters512 (approved):

PR #7357, opened 2023-03-22
Log all created/updated/deleted chunks after sync

### LINKED ISSUE #7022: Log which chunks were updated during a sync
We've periodically had weird issues with chunk generation that are difficult to debug after the fact. We should log which chunks were generated during a sync, and perhaps which files were identified as changed.

nwalters512:
Closes #7022.

### DIFF
--- a/lib/chunks.js
+++ b/lib/chunks.js
@@ -195,28 +195,25 @@ module.exports.chunkMetadataFromDatabaseChunk = function (chunk) {
  * Returns the absolute path for a course's chunk within the course's runtime
  * directory.
  *
- * @param {string} coursePath
  * @param {ChunkMetadata} chunkMetadata
  */
-module.exports.coursePathForChunk = function (coursePath, chunkMetadata) {
+module.exports.pathForChunk = function (chunkMetadata) {
   switch (chunkMetadata.type) {
     case 'elements':
     case 'elementExtensions':
     case 'clientFilesCourse':
     case 'serverFilesCourse':
-      return path.join(coursePath, chunkMetadata.type);
+      return chunkMetadata.type;
     case 'question':
-      return path.join(coursePath, 'questions', chunkMetadata.questionName);
+      return path.join('questions', chunkMetadata.questionName);
     case 'clientFilesCourseInstance':
       return path.join(
-        coursePath,
         'courseInstances',
         chunkMetadata.courseInstanceName,
         'clientFilesCourseInstance'
       );
     case 'clientFilesAssessment':
       return path.join(
-        coursePath,
         'courseInstances',
         chunkMetadata.courseInstanceName,
         'assessments',
@@ -224,6 +221,17 @@ module.exports.coursePathForChunk = function (coursePath, chunkMetadata) {
         'clientFilesAssessment'
       );
   }
+};
+
+/**
+ * Returns the absolute path for a course's chunk within the course's runtime
+ * directory.
+ *
+ * @param {string} coursePath
+ * @param {ChunkMetadata} chunkMetadata
+ */
+module.exports.coursePathForChunk = function (coursePath, chunkMetadata) {
+  return path.join(coursePath, module.exports.pathForChunk(chunkMetadata));
 };
 
 /**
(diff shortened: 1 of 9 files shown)


lib/server-jobs.js
```diff
@@ -55,137 +55,129 @@ module.exports.liveJobs = {};
...
-  };
-  sqldb.queryZeroOrOneRow(sql.update_job_on_close, params, (err, result) => {
-    delete module.exports.liveJobs[this.id];
-
-    // Notify sockets.
-    socketServer.io.to('job-' + this.id).emit('update');
-    if (this.options.job_sequence_id != null) {
-      socketServer.io.to(`jobSequence-${this.options.job_sequence_id}`).emit('update');
-    }
+class Job {
```
nwalters512, line 58: Refactored to a class so I could more easily use this as a type elsewhere.

lib/chunks.js
```diff
@@ -458,7 +465,7 @@ module.exports.diffChunks = async ({ coursePath, courseId, courseData, changedFi
         (!existingCourseChunks[chunkType] || changedCourseChunks[chunkType])
       ) {
         updatedChunks.push({ type: chunkType });
-      } else if (!hasChunkDirectory) {
+      } else if (!hasChunkDirectory && existingCourseChunks[chunkType]) {
```
nwalters512, line 468: This was an interesting one: we were marking these chunks as needing deletion even if they didn't exist! In practice, this wasn't actually a bug, as Postgres silently ignores these (as it should). However, it would be surprising for instructors if these showed up in logs, so I fixed this.

mwest1066: Good catch!

lib/chunks.js
```diff
@@ -1065,3 +1077,34 @@ module.exports.getTemplateQuestionIdsAsync = async (question) => {
...
+/**
+ * Logs the changes to chunks for a given job.
+ *
+ * @param {{ updatedChunks: ChunkMetadata[], deletedChunks: ChunkMetadata[] }} chunkMetadata
+ * @param {import('./server-jobs').Job} job
+ * @returns
+ */
+module.exports.logChunkChangesToJob = ({ updatedChunks, deletedChunks }, job) => {
+  if (updatedChunks.length === 0 && deletedChunks.length === 0) {
+    return;
```
mwest1066, line 1090: Perhaps it would be useful to explicitly record that we did not update/delete chunks, just to be extra clear (see below)?

lib/chunks.js
```diff
@@ -1065,3 +1077,34 @@ module.exports.getTemplateQuestionIdsAsync = async (question) => {
...
+  }
+
+  const lines = [];
+
+  if (updatedChunks.length > 0) {
+    lines.push('Created or updated the following chunks:');
+    updatedChunks.forEach((chunk) => {
+      lines.push(`  ${module.exports.pathForChunk(chunk)}`);
+    });
+  }
```
mwest1066, line 1100:
```suggestion
  } else {
    lines.push('No chunks required creation or deletion.');
  }
```

lib/chunks.js
```diff
@@ -1065,3 +1077,34 @@ module.exports.getTemplateQuestionIdsAsync = async (question) => {
...
+
+  if (deletedChunks.length > 0) {
+    lines.push('Deleted the following chunks:');
+    deletedChunks.forEach((chunk) => {
+      lines.push(`  ${module.exports.pathForChunk(chunk)}`);
+    });
+  }
```
mwest1066, line 1107:
```suggestion
  } else {
    lines.push('No chunks required deletion.');
  }
```

mwest1066 (approved):

lib/chunks.js
nwalters512: I find myself reaching for a word like "updation" here (this branch doesn't cover deletions) - what's the right way to phrase this? Maybe instead of "created or updated" we just say "generated", then we can say "No changes required chunk generation"?

nwalters512: I think I landed on something I liked in 7583c53 - how does that look?

mwest1066: Looks great!

PR #7363, opened 2023-03-23
Use Amazon Linux 2023 for plbase image

nwalters512:
Unfortunately this new image is almost a full gigabyte larger than before, possibly due to `texlive` including many more packages now?

```
$ docker images --filter=reference='prairielearn/plbase:*'
REPOSITORY            TAG              IMAGE ID       CREATED          SIZE
prairielearn/plbase   nathan-testing   bc214209b6d0   28 minutes ago   4.65GB
prairielearn/plbase   latest           b1328997485b   13 days ago      3.66GB
```


nwalters512:
With the changes in 4589a69, I was able to get this image down to `3.78GB`. Still not ideal, but 100MB more is tolerable for now.

Removing Docker means I had to remove the `kill_running_workspaces.sh` script. If we really need that, we should either rewrite that as a JS script (because then we can use dockerode, which will already be installed), or make it a part of the workspace host itself, possibly enabled via an environment variable?

PR #7364, opened 2023-03-23
Create Alpine Linux base image

nwalters512:
Naturally, R and its packages was the only piece of this that didn't work correctly. I'm putting this here for me to reference for the time being, but realistically I'm probably going to have to build our image on Ubuntu instead.


nwalters512:
This is probably not feasible at the moment.

PR #7368, opened 2023-03-24
Clean up R workspace and workspace element

### LINKED ISSUE #7352: Clean up R workspace question
From https://github.com/PrairieLearn/PrairieLearn/pull/4826#discussion_r1143793526:

Need to remove from `question.html`:

https://github.com/PrairieLearn/PrairieLearn/blob/6c4ae75662b3bfb2997ed4fa61258b8881271465/exampleCourse/questions/demo/workspace/rstudio/question.html#L18

Modify `info.json`:

https://github.com/PrairieLearn/PrairieLearn/blob/6c4ae75662b3bfb2997ed4fa61258b8881271465/exampleCourse/questions/demo/workspace/rstudio/info.json#L17-L23

by adding the key of: 

```json
        "gradedFiles": [
            "student.R",
        ]
```

Move the `demo.R` code into the `workspace/` folder and rename it to `student.R`.

This should give the directory structure of:

```sh
workspace
`-- rstudio                      # for an externally graded workspace question
    +-- info.json               # metadata for rstudio
    +-- question.html      # HTML template for rstudio
    |
    +-- tests                    # external grading files (see other doc)
    |   +-- ans.R
    |   +-- test_01_exists.R
    |   +-- test_02_type.R
    |   +-- test_03_length.R
    |   +-- test_04_components_exist.R
    |   +-- test_05_types.R
    |   `-- test_06_correct_components.R
    |
    `-- workspace          # copied into the student's workspace container home dir (optional)
        `-- student.R
```

eliotwrobson:
Resolves #7352. Haven't tested locally and couldn't get workspaces working, so need someone else to take a look at this (or some help in general). cc @coatless.

Also @jonatanschroeder I tried removing the question + workspace in https://github.com/PrairieLearn/PrairieLearn/blob/master/exampleCourse/questions/demo/autograder/r/createRObject/question.html but this caused syncing issues for some reason.


jonatanschroeder:
> Also @jonatanschroeder I tried removing the question + workspace in https://github.com/PrairieLearn/PrairieLearn/blob/master/exampleCourse/questions/demo/autograder/r/createRObject/question.html but this caused syncing issues for some reason.

I'm not sure what you mean, but all I suggested there was to remove the comment "You can press _Open workspace_ above to launch an RStudio instance", since this question does not use workspaces.

exampleCourse/questions/demo/workspace/rstudio/workspace/student.R
jonatanschroeder, line 1: Can you review the comments at the start of this file? There seem to be some grammar issues, including an unclosed parenthesis.

exampleCourse/questions/demo/workspace/rstudio/question.html
```diff
@@ -10,13 +10,12 @@
 - a numeric `nn` with value 3.1415
 - a text variable `cc` with text `hello, world!` including the comma an exclamation
 
-You can press the button below to launch a demo RStudio workspace.
+You can press the button below to launch a demo RStudio workspace. The file must be
+completed in the workspace.
```
jonatanschroeder, line 14:
```suggestion
Press the button below to launch an RStudio workspace. The object above must be created in the `student.R` file inside the workspace.
```

I'm ok with other ways to phrase it, but (1) it needs to be clearer that opening the workspace is required, and (2) the file name is important in this context.

jonatanschroeder (commented):
I tested this locally, and while the `student.R` file is added to the studio directory, it is not open by default:

image

Do you know if there is any setting that can be used to do that in R studio?

Also, you can delete the starter_file.md file in the workspace directory, I don't think it's needed.

jonatanschroeder:
> I tested this locally, and while the `student.R` file is added to the studio directory, it is not open by default.
> Do you know if there is any setting that can be used to do that in R studio?

I was playing around with this (while in a meeting where my input was not needed 😬), you can create a `.Rprofile` file inside the workspace directory with the following content (thanks to https://stackoverflow.com/a/70070152):

```r
setHook("rstudio.sessionInit", function(newSession) {
  file.edit("student.R")
}, action = "append")
```

Of course I know virtually nothing about R, so there may be better options. Maybe @coatless or @eddelbuettel can opine?

jonatanschroeder:
> couldn't get workspaces working, so need someone else to take a look at this (or some help in general).

Did you check this? https://prairielearn.readthedocs.io/en/latest/workspaces/#developing-with-workspaces-in-docker

nwalters512:
FWIW, I'm pretty sure the VSCode and JupyterLab workspaces don't open any files by default either, so I don't think we should spend too much time trying to make that work.

coatless:
@jonatanschroeder that's perfect actually! We actually used that hook when CBTF computers needed to be imaged with RStudio and we had to override the installation command:

https://github.com/illinois-r/r-centos7-cbtf/blob/d7459d3251d9e3522035e3bb0de2136a6b0e621a/.Rprofile#L121-L131

Official documentation on the startup script trick:
https://docs.posit.co/ide/server-pro/rstudio_pro_sessions/session_startup_scripts.html

That said, we might want to use `rstudioapi::navigateToFile()` instead of `file.edit()` as the later is just being overridden by RStudio's context handler.

If we want to open _all_ the script files by default, we could add:

`~/.Rprofile` -> `/home/rstudio/.Rprofile`

```r
setHook("rstudio.sessionInit", function(newSession) {
  script_files = list.files("~/", "\\.R$", full.names = TRUE)
  for(script in script_files) {
       rstudioapi::navigateToFile(script)
  }
}, action = "append")
```

The `list.files()` search can be further extended if need be. Though, I'm not sure if we want to open more than _just_ the R scripts though... 🤷 

Otherwise, we would need to access the file information stored in `info.json` for `gradedFiles`.

exampleCourse/questions/demo/workspace/rstudio/question.html
coatless: I like this wording a bit more.

echuber2:
Whatever you choose to do about opening a file by default, please take into account what the Dockerfile is already trying to do. There's a comment about using `.Rprofile` vs. `rsession.conf`: I tried to get it to open all panes to the expected directory but not to open a file. If you rely on a `.Rprofile` hack per-question, you also need to take into account whether the question home for the workspace matches the user home in the container: I expected that they would be different, keeping the question home deeper in `/home/rstudio/workspace`.

echuber2:
It seems that it's okay to keep `.Rprofile` in the base of the _project_ directory rather than the user home, so it would probably be okay to customize that per-question by including the file in the starter files for the workspace.

exampleCourse/questions/demo/workspace/rstudio/workspace/student.R
eliotwrobson: Updated here https://github.com/PrairieLearn/PrairieLearn/pull/7368/commits/e0e9564b60d2d3d71a6bfdd8b779206741d81c00

eliotwrobson:
@coatless @jonatanschroeder In the latest push, I think I've addressed all the comments here. Still having weird issues getting the workspaces to launch (along with other technical issues I've been having getting PL to work right on my machine), so flying a little bit blind. But I think things are starting to look more reasonable.

eliotwrobson:
Edited this PR to include uplifting the workspace element, should be ready for review. For the life of me, I could not get workspaces running locally, so whoever reviews this should test it briefly to make sure everything is working as expected (cc @coatless).

exampleCourse/questions/demo/workspace/rstudio/question.html
jonatanschroeder, line 1: I can't comment on line 4 since it's not changed, but is that horizontal line needed there for some reason?

exampleCourse/questions/demo/workspace/rstudio/question.html
```diff
@@ -10,13 +10,11 @@
 - a numeric `nn` with value 3.1415
 - a text variable `cc` with text `hello, world!` including the comma an exclamation
 
-You can press the button below to launch a demo RStudio workspace.
+Press the button below to launch an RStudio workspace. The object above must be created in the `student.R` file inside the workspace.
```
jonatanschroeder, line 13: It seems RStudio does not auto-save, so a reminder to save the file in RStudio may be a good idea.

jonatanschroeder (approved):
There are some minor question text issues, but otherwise it looks fine.

exampleCourse/questions/demo/workspace/rstudio/question.html
eliotwrobson: I don't think it's needed. When I open the page it doesn't appear for some reason? Either way, I'll go ahead and remove it.

exampleCourse/questions/demo/workspace/rstudio/question.html
eliotwrobson: Sure, will add that here.

exampleCourse/questions/demo/workspace/rstudio/question.html
eliotwrobson: Removed here https://github.com/PrairieLearn/PrairieLearn/pull/7368/commits/e24fa2667373b1d5f305b159a179d02ed7e799b8

exampleCourse/questions/demo/workspace/rstudio/question.html
eliotwrobson: Removed here https://github.com/PrairieLearn/PrairieLearn/pull/7368/commits/e24fa2667373b1d5f305b159a179d02ed7e799b8

exampleCourse/questions/demo/workspace/rstudio/question.html
```diff
@@ -11,6 +9,8 @@
 - a text variable `cc` with text `hello, world!` including the comma an exclamation
```
jonatanschroeder, line 9: typo: "comma and exclamation mark"

eliotwrobson: https://github.com/PrairieLearn/PrairieLearn/pull/7368/commits/5f05a563a42dfd66953711bb727161fc7e2ae416

PR #7369, opened 2023-03-24
C Grader: fix handling of questions using Check without ASAN

jonatanschroeder:
This ensures ASAN calls in check.h are only used if ASAN is actually enabled. As discussed in Slack.


nwalters512 (approved):
Haven't personally tested this, but it looks reasonable!

PR #7371, opened 2023-03-27
Use FOR NO KEY UPDATE instead of FOR UPDATE

nwalters512:
This should help reduce lock contention by allowing `page_view_logs` to insert records even when locks are held on referenced tables.

Though this is probably only necessary for `assessment`/`assessment_instance`/`variant` tables, I opted to make the change for all tables. This should be safe, as we're neither deleting a row or changing its primary key after acquiring a lock on it.

Closes https://github.com/PrairieLearnInc/sysconf/issues/870.

### DIFF
--- a/lib/assessment.sql
+++ b/lib/assessment.sql
@@ -81,7 +81,7 @@ FROM
   assessments AS a
 WHERE
   a.id = $assessment_id
-FOR UPDATE;
+FOR NO KEY UPDATE;
 
 -- BLOCK select_assessment_needs_statisics_update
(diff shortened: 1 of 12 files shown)


mwest1066 (approved):
Do we also want to convert the FOR UPDATE statements in `packages/named-locks`? It probably doesn't matter either way because nothing else should have foreign key constraints on named lock rows.

nwalters512:
Yeah, that should never matter - if for some reason it does, we've done something horrible.

PR #7372, opened 2023-03-27
Don't start nested transactions in runInTransactionAsync

nwalters512:
Consider the following code:

```ts
await runInTransactionAsync(async () => {
  await runInTransactionAsync(async () => {
    queryAsync('SELECT ...', {});
  });
});
```

Before this change, that would have resulted in the following warnings from Postgres:

```
2023-03-27 01:08:49 UTC:10.0.49.218(39398):pdb@prairielearn:[4822]:WARNING:  there is already a transaction in progress
2023-03-27 01:08:50 UTC:10.0.49.218(39398):pdb@prairielearn:[4822]:WARNING:  there is no transaction in progress
```

The first one is fine, but the second one is potentially a problem, as it means we're prematurely committing a transaction.

With this change, any nested transactions won't execute `START TRANSACTION` and either `COMMIT` or `ROLLBACK`; only the root transaction will be able to do that.

### DIFF
--- a/packages/postgres/src/pool.ts
+++ b/packages/postgres/src/pool.ts
@@ -470,19 +470,39 @@ export class PostgresPool {
    * will be committed otherwise.
    */
   async runInTransactionAsync<T>(fn: (client: pg.PoolClient) => Promise<T>): Promise<T> {
-    const client = await this.beginTransactionAsync();
+    // Check if we're already inside a transaction. If so, we won't start another one,
+    // as Postgres doesn't support nested transactions.
+    let client = this.alsClient.getStore();
+    const isNestedTransaction = client !== undefined;
+    if (!client) {
+      client = await this.beginTransactionAsync();
+    }
+
+    // We reassign to a separate variable here so that TypeScript can reason
+    // about the fact that `client` will always be defined by this point.
+    const transactionClient = client;
+
     let result: T;
     try {
-      result = await this.alsClient.run(client, () => fn(client));
+      result = await this.alsClient.run(transactionClient, () => fn(transactionClient));
     } catch (err: any) {
-      await this.endTransactionAsync(client, err);
+      if (!isNestedTransaction) {
+        // If we're inside another transaction, we assume that the root transaction
+        // will catch this error and roll back the transaction.
+        await this.endTransactionAsync(transactionClient, err);
+      }
       throw err;
     }
 
-    // Note that we don't invoke `endTransactionAsync` inside the `try` block
-    // because we don't want an error thrown by it to trigger *another* call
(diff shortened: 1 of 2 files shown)


packages/postgres/src/pool.ts
```diff
@@ -470,19 +470,39 @@ export class PostgresPool {
...
+    // as Postgres doesn't support nested transactions.
+    let client = this.alsClient.getStore();
+    const isNestedTransaction = client !== undefined;
+    if (!client) {
+      client = await this.beginTransactionAsync();
+    }
+
+    // We reassign to a separate variable here so that TypeScript can reason
+    // about the fact that `client` will always be defined by this point.
+    const transactionClient = client;
```
jonatanschroeder, line 483:
```suggestion
    const transactionClient = client || await this.beginTransactionAsync();
```

This is nit, of course, but I think it makes the code slightly more readable. Also, this allows `client` to be a `const`.

jonatanschroeder (commented):
I haven't tested, but it looks good.

packages/postgres/src/pool.ts
nwalters512: That is indeed much nicer - thanks! I'll tweak it to use `??` instead of `||`.

mwest1066 (approved):

packages/postgres/src/pool.ts
```diff
@@ -470,19 +470,33 @@ export class PostgresPool {
    * will be committed otherwise.
    */
   async runInTransactionAsync<T>(fn: (client: pg.PoolClient) => Promise<T>): Promise<T> {
-    const client = await this.beginTransactionAsync();
+    // Check if we're already inside a transaction. If so, we won't start another one,
+    // as Postgres doesn't support nested transactions.
+    let client = this.alsClient.getStore();
+    const isNestedTransaction = client !== undefined;
```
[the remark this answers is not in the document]
nwalters512: The last one: any code running outside of `alsClient.run()` will see `alsClient.getStore() === undefined`.

nwalters512:
> Does this mean that we are currently committing transactions prematurely in cases where we have nested `runInTransactionAsync()` calls?

I believe this to be the case, based on the fact that we have `there is no transaction in progress` logs.

mwest1066:
> > The first one is fine, but the second one is potentially a problem, as it means we're prematurely committing a transaction.
> 
> Does this mean that we are currently committing transactions prematurely in cases where we have nested `runInTransactionAsync()` calls?

Ah sorry, I misunderstood you. It's the _second_ message that's the problem, coming from the fact that we _committed_ the transaction prematurely (from the nested `runInTransactionAsync()`, rather than the parent `runInTransactionAsync()`).

PR #7376, opened 2023-03-27
Move database testing utilities into @prairielearn/postgres

nwalters512:
This allows us to reuse our utilities for spinning up and tearing down test databases across packages, including in `@prairielearn/postgres` itself.

Once sproc handling is moved into a shared package, I plan to introduce a package like `@prairielearn/server-test-utils` that encapsulates migrations and sproc creation as well; we can then reuse that across PrairieLearn and PrairieTest (and any other servers we end up with in the future).

### DIFF
--- a/packages/postgres/src/test-utils.ts
+++ b/packages/postgres/src/test-utils.ts
@@ -7,10 +7,11 @@ const POSTGRES_HOST = 'localhost';
 const POSTGRES_DATABASE = 'postgres';
 
 export interface PostgresTestUtilsOptions {
+  database: string;
   user?: string;
   host?: string;
+  poolConfig?: Pick<pg.PoolConfig, 'max' | 'idleTimeoutMillis'>;
   defaultDatabase?: string;
-  database: string;
   prepareAfterReset?: (client: pg.Client) => Promise<void>;
 }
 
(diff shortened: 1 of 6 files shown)


packages/postgres/src/test-utils.ts
```diff
@@ -0,0 +1,153 @@
...
+
+interface DropDatabaseOptions {
+  database?: string;
+  force?: boolean;
+  closePool?: boolean;
+}
+
+async function createDatabase(
+  options: PostgresTestUtilsOptions,
+  createOptions: CreateDatabaseOptions = {}
```
mwest1066, line 33: To help me understand the patterns we are using here, could you have kept all the defaults together (and simplified the later code) with something like the following?
```suggestion
  {dropExistingDatabase = true, configurePool = true, ...more options... }: CreateDatabaseOptions
```

packages/postgres/src/test-utils.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  await createOptions.prepare?.(client);
+
+  if (configurePool) {
+    await defaultPool.initAsync(
+      {
+        user: options.user ?? POSTGRES_USER,
+        host: options.host ?? POSTGRES_HOST,
+        database: getDatabaseNameForCurrentMochaWorker(options.database),
+        // TODO: make these configurable?
```
mwest1066, line 68: Seems like we may as well do it now, given that it's annoying to change a package and do the whole publish dance if we need to change this in the future?

mwest1066 (approved):

packages/postgres/src/test-utils.ts
nwalters512: Yep, good call - I meant to do this before requesting review but forgot, sorry!

packages/postgres/src/test-utils.ts
nwalters512: Yep, that would be equivalent. I'll make that change!

mwest1066 (approved):

PR #7377, opened 2023-03-28
Remove unnecessary auth checks

nwalters512:
These are all covered by the middleware that runs for the entire `/pl/course_instance/:course_instance_id/instance_question/:instance_question_id` route prefix.

https://github.com/PrairieLearn/PrairieLearn/blob/ed27bcfa732e8a8e1ff6833a06499a34fa85d7bf/server.js#L1335-L1341

### DIFF
--- a/server.js
+++ b/server.js
@@ -946,7 +946,10 @@ module.exports.initExpress = function () {
   // Submission files
   app.use(
     '/pl/course_instance/:course_instance_id/instructor/instance_question/:instance_question_id/submission/:submission_id/file',
-    [require('./pages/submissionFile/submissionFile')]
+    [
+      require('./middlewares/selectAndAuthzInstanceQuestion'),
+      require('./pages/submissionFile/submissionFile'),
+    ]
   );
 
   app.use(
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -1369,42 +1366,26 @@ module.exports.initExpress = function () {
   );
   app.use(
     '/pl/course_instance/:course_instance_id/instance_question/:instance_question_id/clientFilesQuestion',
-    [
-      require('./middlewares/selectAndAuthzInstanceQuestion'),
-      require('./middlewares/studentAssessmentAccess'),
-      require('./pages/clientFilesQuestion/clientFilesQuestion'),
-    ]
+    [require('./pages/clientFilesQuestion/clientFilesQuestion')]
```
mwest1066, line 1369: We can get rid of the `[]` here if we want (and similarly elsewhere).
```suggestion
    require('./pages/clientFilesQuestion/clientFilesQuestion')
```

server.js
```diff
@@ -946,10 +946,7 @@ module.exports.initExpress = function () {
   // Submission files
   app.use(
     '/pl/course_instance/:course_instance_id/instructor/instance_question/:instance_question_id/submission/:submission_id/file',
-    [
-      require('./middlewares/selectAndAuthzInstanceQuestion'),
```
mwest1066, line 950: I think this one is actually needed?

nwalters512: Good catch! Resolved in 2aaec9c.

server.js
nwalters512: Done in 2aaec9c.

mwest1066: Not that it really matters, but it looks like you missed the ones for `legacyQuestionFile` and `legacyQuestionText`.

mwest1066 (approved):
Looks good. See https://github.com/PrairieLearn/PrairieLearn/pull/7377#discussion_r1152648245 for a minor tweak if you want to do it.

server.js
nwalters512: Whoops, indeed - addressed in 5e59a34.

PR #7381, opened 2023-03-30
Fix bad studentAssessments query plan

mwest1066:
@nwalters512 discovered that the `studentAssessments` page query had recently switched to a seq scan over `assessment_instances` for a small number of course instances. The specific problem was in the CTE: https://github.com/PrairieLearn/PrairieLearn/blob/a2fde2c64ad291eba85b692d8eaceed3c0feaf6a/pages/studentAssessments/studentAssessments.sql#L80 and even more specifically in the join condition: https://github.com/PrairieLearn/PrairieLearn/blob/a2fde2c64ad291eba85b692d8eaceed3c0feaf6a/pages/studentAssessments/studentAssessments.sql#L125-L128

To help postgres use indexes for both `assessment_instances.user_id` and `assessment_instances.group_id`, this PR uses two joins on `assessment_instances`, one for `user_id` and one for `group_id`. In testing this allowed postgres to use index scans and gave approximately a 1000x speed up in the cases when the old query used a seq scan.

### DIFF
--- a/pages/studentAssessments/studentAssessments.sql
+++ b/pages/studentAssessments/studentAssessments.sql
@@ -98,10 +98,10 @@ WITH
       aa.active AS active,
       aa.access_rules,
       aa.show_closed_assessment_score,
-      coalesce(ai1.id, ai2.id) AS assessment_instance_id,
-      coalesce(ai1.number, ai2.number) AS assessment_instance_number,
-      coalesce(ai1.score_perc, ai2.score_perc) AS assessment_instance_score_perc,
-      coalesce(ai1.open, ai2.open) AS assessment_instance_open,
+      ai.id AS assessment_instance_id,
+      ai.number AS assessment_instance_number,
+      ai.score_perc AS assessment_instance_score_perc,
+      ai.open AS assessment_instance_open,
       am.id AS assessment_module_id,
       am.name AS assessment_module_name,
       am.heading AS assessment_module_heading,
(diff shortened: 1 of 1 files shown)


pages/studentAssessments/studentAssessments.sql
```diff
@@ -120,12 +120,17 @@ WITH
...
-      LEFT JOIN assessment_instances AS ai ON (
-        ai.assessment_id = a.id
-        AND (
-          ai.user_id = $user_id
-          OR ai.group_id = gu.group_id
-        )
+      -- We join assessment_instances twice, once on user_id and once on
+      -- group_id. We used to have these combined with AND (ai.user_id =
+      -- $user_id OR ai.group_id = gu.group_id) but this was triggering a bad
+      -- query plan for some course instances.
```
nwalters512, line 126: Maybe add slightly more detail about the fact that we have two indexes we want to utilize?

```suggestion
      -- We join assessment_instances twice, once on user_id and once on
      -- group_id. We used to have these combined with AND (ai.user_id =
      -- $user_id OR ai.group_id = gu.group_id) but this was triggering a bad
      -- query plan for some course instances. By joining twice, we encourage
      -- the query planner to utilize the two separate indexes we have.
```

pages/studentAssessments/studentAssessments.sql
```diff
@@ -120,12 +120,17 @@ WITH
...
+      LEFT JOIN assessment_instances AS ai1 ON (
+        ai1.assessment_id = a.id
+        AND ai1.user_id = $user_id
+      )
+      LEFT JOIN assessment_instances AS ai2 ON (
+        ai2.assessment_id = a.id
+        AND ai2.group_id = gu.group_id
```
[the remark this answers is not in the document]
jonatanschroeder: Because the condition of `ai2` does not use fields in `ai1`, they are independent, if that is what you are referring to.

pages/studentAssessments/studentAssessments.sql
```diff
@@ -120,12 +120,17 @@ WITH
...
       )
```
jonatanschroeder, line 134:
```suggestion
      LEFT JOIN (
        SELECT * FROM assessment_instances AS ai1 WHERE
        ai1.assessment_id = a.id
        AND ai1.user_id = $user_id
        UNION ALL
        SELECT * FROM assessment_instances AS ai2 WHERE
        ai2.assessment_id = a.id
        AND ai2.group_id = gu.group_id
      ) AS ai ON (TRUE)
```

Not sure how Postgres would react to this, but I believe this may make the query a bit clearer.

jonatanschroeder: https://www.cybertec-postgresql.com/en/avoid-or-for-better-performance/

mwest1066: Good thinking! I switched to this in 7bdee2d54007fc661ffbe98b90bb2792aebbc34e. I tested this and it still avoids seq scans. This modified query should behave identically to the old code, avoiding the edge case I identified above: https://github.com/PrairieLearn/PrairieLearn/pull/7381#discussion_r1153512721

pages/studentAssessments/studentAssessments.sql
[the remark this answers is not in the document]
mwest1066: @jonatanschroeder had a better idea to use a subquery with a UNION, which is clearer and also avoids the edge case I described above. I switched in 7bdee2d54007fc661ffbe98b90bb2792aebbc34e

pages/studentAssessments/studentAssessments.sql
mwest1066: @jonatanschroeder that's an excellent blog post, thanks!

Issue #7344, opened 2023-03-15
Ensure consistency in handling of timezone columns


jonatanschroeder:
For the `display_timezone` column in `course_instances`, the sync process always sets this column to a default value based on the course (or a fixed value if the course doesn't have it), so it could be potentially tagged as `NOT NULL` as well.

https://github.com/PrairieLearn/PrairieLearn/blob/c2dbb6219bb52ae0bbe54aedf2ed355e0934b71a/sync/fromDisk/courseInstances.js#L34

Curiously, the same field in `pl_courses` may be set to null, though it may be an interesting idea to change that too for consistency.

https://github.com/PrairieLearn/PrairieLearn/blob/c2dbb6219bb52ae0bbe54aedf2ed355e0934b71a/sync/fromDisk/courseInfo.js#L29

Additionally, this should also allow a lot of queries that currently coalesce the course instance value with the course value to simplify and use `ci.display_timezone` only (whenever the course instance is expected to exist, of course). For example:

https://github.com/PrairieLearn/PrairieLearn/blob/c2dbb6219bb52ae0bbe54aedf2ed355e0934b71a/middlewares/selectAndAuthzAssessmentInstance.sql#L27
https://github.com/PrairieLearn/PrairieLearn/blob/c2dbb6219bb52ae0bbe54aedf2ed355e0934b71a/pages/instructorAssessmentManualGrading/instanceQuestion/instanceQuestion.sql#L8

Related: #7320.

mwest1066:
Note that there is also `institutions.display_timezone` which is NOT NULL. This is the correct fallback for `pl_courses.display_timezone`. I think we should make both `pl_courses.display_timezone` and `course_instances.display_timezone` NOT NULL and on sync we should use:
* For course instances: `display_timezone: courseInstance.timezone ?? courseTimezone`
* For courses: `display_timezone: courseInfo.timezone ?? institutionTimezone`
I 100% agree that having all `display_timezone` columns be NOT NULL would significantly simply lots of queries and keep everything consistent.

mwest1066:
Also see https://github.com/PrairieLearn/PrairieLearn/pull/7382 and especially https://github.com/PrairieLearn/PrairieLearn/pull/7382#discussion_r1153871876

jonatanschroeder:
Summarizing the relevant discussion from #7382 (@nwalters512 and @mwest1066 please feel free to add):

[...]

PR #7382, opened 2023-03-30
Define fallback timezones for effective user page

nwalters512:
#7340 didn't account for the fact that the effective user page is also mounted to `/pl/course/...`, where we won't have access to a course instance from which we can fetch a timezone.

### DIFF
--- a/pages/instructorEffectiveUser/instructorEffectiveUser.js
+++ b/pages/instructorEffectiveUser/instructorEffectiveUser.js
@@ -47,11 +47,12 @@ router.get('/', function (req, res, next) {
 
     // This page can be mounted under `/pl/course/...`, in which case we won't
     // have a course instance to get a display timezone from. In that case, we'll
-    // fall back to the course, and then to `UTC`. Note that in some other places
-    // we fall back to `America/Chicago` if neither the course nor the course
-    // instance has a timezone; UTC seems to make more sense here.
+    // fall back to the course, and then to the institution. All institutions must
+    // have a display timezone, so we're always guaranteed to have one.
     const displayTimezone =
-      res.locals.course_instance?.display_timezone ?? res.locals.course.display_timezone ?? 'UTC';
+      res.locals.course_instance?.display_timezone ??
+      res.locals.course.display_timezone ??
+      res.locals.institution.display_timezone;
 
     res.locals.true_req_date_for_display = format(
       utcToZonedTime(res.locals.true_req_date, displayTimezone),
--- a/pages/instructorEffectiveUser/instructorEffectiveUser.js
+++ b/pages/instructorEffectiveUser/instructorEffectiveUser.js
@@ -44,18 +44,27 @@ router.get('/', function (req, res, next) {
     if (res.locals.ipaddress.substr(0, 7) === '::ffff:') {
       res.locals.ipaddress = res.locals.ipaddress.substr(7);
     }
+
+    // This page can be mounted under `/pl/course/...`, in which case we won't
+    // have a course instance to get a display timezone from. In that case, we'll
+    // fall back to the course, and then to `UTC`. Note that in some other places
+    // we fall back to `America/Chicago` if neither the course nor the course
+    // instance has a timezone; UTC seems to make more sense here.
+    const displayTimezone =
+      res.locals.course_instance?.display_timezone ?? res.locals.course.display_timezone ?? 'UTC';
+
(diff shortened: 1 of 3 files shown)


pages/instructorEffectiveUser/instructorEffectiveUser.js
```diff
@@ -44,18 +44,27 @@ router.get('/', function (req, res, next) {
     if (res.locals.ipaddress.substr(0, 7) === '::ffff:') {
       res.locals.ipaddress = res.locals.ipaddress.substr(7);
     }
+
+    // This page can be mounted under `/pl/course/...`, in which case we won't
+    // have a course instance to get a display timezone from. In that case, we'll
+    // fall back to the course, and then to `UTC`. Note that in some other places
+    // we fall back to `America/Chicago` if neither the course nor the course
```
nwalters512, line 51: I think defaulting to `America/Chicago` made sense when PL was UIUC only, but now I imagine that using UTC is a better choice. I don't know how safe it is to change this at the moment. Maybe the right thing to do is to not even have a default and error out during sync if at least a course timezone isn't specified? We'd have to make that change carefully, of course.

mwest1066: I think the ultimate fallback should be to `institutions.display_timezone`, which is NOT NULL. My original intention here was to always use:
```
coalesce(course_instance.display_timezone, course.display_timezone, institution.display_timezone)
```

In retrospect I think this is a bad design and we should change all three tables to have the `display_timezone` column be NOT NULL. At sync time we should fill in the value from either JSON (if provided) or from the "parent" table.

nwalters512: `display_timezone` for institutions is currently a PT concept; `database/tables/institutions.pg` doesn't contain it. Even then, the default institution is set to `America/Chicago`, which would be confusing to anyone not in the Midwest. Fixing that is of course outside the scope of this PR, and I agree that defaulting to `institution.display_timezone` is ultimately the right thing to do. We should just be more cognizant of being, uh, non-Illinois-centric when picking defaults for things in the future.

All I have to do is add a migration that adds `institutions.display_timezone` such that the definition matches that from PT, correct?

mwest1066: Ah, excellent point that PL doesn't know about `institutions.display_timezone` 😅 

Just make sure the migration uses `ADD COLUMN IF NOT EXISTS`.

I fully agree that we want to get away from defaulting to `America/Chicago`. I'm not sure what the best way is to manage this for courses in the default institution? Error during sync if a course doesn't have a timezone set and is in the default institution?

nwalters512: The problem is that _all_ courses are in the default institution during local dev 😓 I can't see a way to make this check consistent between local dev and prod if "is timezone required for course" is dependent upon state that's not in the course repo.

mwest1066: Dang, good point. Next idea:
* When we create new courses (copied from the template course), set the timezone in `infoCourse.json` from `institutions.display_timezone`.
* On sync, set `course.display_timezone = courseInfo.timezone ?? 'America/Chicago'`
* On sync, set `course_instance.display_timezone = courseInstance.timezone ?? courseTimezone`

This will be no worse than what we currently have, and will let us make all timezones NOT NULL.

jonatanschroeder: Can the default institution be created with a default timezone? This can be set to `America/Chicago` for now to maintain compatibility. We can then have some call of action asking all courses to assign a timezone in their `infoCourse.json` (e.g., a sync warning, a Slack/GitHub announcement, and a news item), and after a few months switch the default to `UTC`.

jonatanschroeder (approved):

pages/instructorEffectiveUser/instructorEffectiveUser.js
jonatanschroeder: Moved this discussion to #7344, for future action.

nwalters512: I added the migration in 516b490 and updated this page to fall back to `institutions.display_timezone` in 537b211.

mwest1066 (approved):

PR #7389, opened 2023-04-01
Introduce @prairielearn/config package

nwalters512:
This will allow us to reload config loading logic, specifically loading from AWS Secrets Manager, in PrairieTest and other applications we introduce.

I took the opportunity to introduce config validation with Zod. This will replace the JSON Schema validation that's already been disabled for a while now.

## TODO

- [x] Load instance metadata for hostname and instance ID
  - [x] Maybe introduce a new package to share the instance metadata fetching logic?
- [x] Add support for loading from AWS Secrets Manager


packages/aws-imds/src/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  privateIp: z.string(),
+  version: z.string(),
+  instanceId: z.string(),
+  instanceType: z.string(),
+  accountId: z.string(),
+  imageId: z.string(),
+  pendingTime: z.string(),
+  architecture: z.string(),
+  region: z.string(),
+});
```
mwest1066, line 19: Just to confirm my understanding: if AWS adds more properties to the instance metadata in the future we'll still correctly parse it, because those extra properties will only be narrowing the type, correct? That is, we won't suddenly break because AWS adds a property?

mwest1066 (approved):
This approach looks great!

packages/aws-imds/src/index.ts
nwalters512: Correct! We'll just silently drop them and they'll be inaccessible to anything downstream of this package.

mwest1066: Perfect!

tools/validate-configs.mjs
nwalters512, line 1: I wrote this little script to validate that all our production config files parse correctly - and they do! I'll remove this before merging.

lib/config-new.js
nwalters512, line 1: I named this `config-new` so that I could keep both config systems around to check for consistency (see `tests/config.test.js`). I'll replace `lib/config.js` with the new implementation before merging this.

workspace_host/interface.js
```diff
@@ -26,17 +26,11 @@ const socketServer = require('../lib/socket-server'); // must load socket server
 const { logger } = require('@prairielearn/logger');
 const LocalLock = require('../lib/local-lock');
 
-const config = require('../lib/config');
+const { config, loadConfigAsync } = require('../lib/config-new');
```
nwalters512: In a future PR, I plan on separating the bits of the workspace config schema from the primary config schema, just like we did with the graders.

mwest1066 (approved):

lib/config-new.js
```diff
@@ -0,0 +1,524 @@
...
+module.exports.setLocalsFromConfig = (locals) => {
+  locals.homeUrl = module.exports.config.homeUrl;
+  locals.urlPrefix = module.exports.config.urlPrefix;
+  locals.plainUrlPrefix = module.exports.config.urlPrefix;
+  locals.navbarType = 'plain';
+  locals.devMode = module.exports.config.devMode;
+  locals.is_administrator = false;
+};
+
+module.exports.getDBConfigValue = async function (key, defaultVal) {
```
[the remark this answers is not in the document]
nwalters512: I *think* in favor of that, yeah.

`display_timezone` will be *slightly* annoying since we use the `config_select` sproc to get that value inside of some other queries; we'll have to switch to manually piping that around. But even then, honestly, I think we could get away with displaying these sort of global dates (those not attached to something in an institution/course/course instance) in UTC. We're already in a situation where I (PT) have to do math to go from CT to either PT or UTC; I think it'd be fine for *all* admins to have to do UTC math since there's no guarantee on what timezones we'll be in. Do you buy that?

`workspaceAutoscalingEnabled` would be a great use case for "negative" feature flags - I think there's value in having a kill switch for this that doesn't require a redeploy to take effect. I suppose we could always disable the auto scaling process on the ASG itself if we really need to stop scaling...

The only other thing we reference from this table is `workspaceDesiredHostCount`. I don't think we've ever had to use that to pin the ASG to a certain size?

packages/config/README.md
```diff
@@ -0,0 +1,52 @@
...
+### Loading config from AWS
+
+If you're running in AWS, you can use `makeImdsConfig()` and `makeSecretsManagerConfig()` to load config from IMDS and Secrets Manager, respectively:
+
+- `makeImdsConfig()` will load `hostname`, `instanceId`, and `region`, which will be available if you config schema contains these values.
+- `makeSecretsManagerConfig()` will look for a `ConfSecret` tag on the instance. If found, the value of that tag will be used treated as a Secrets Manager secret ID, and that secret's value will be parsed as JSON and merged into the config.
+
+Note that both of these config sources are no-ops by default. To active them, you must do one of the following:
+
+- Set `CONFIG_LOAD_FROM_AWS=1` in the process environment.
```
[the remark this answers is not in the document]
mwest1066: I'm totally on board.

lib/config-new.js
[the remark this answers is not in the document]
mwest1066: I'm totally fine with switching global admin to be UTC. The server-level `display_timezone` was a leftover from when we only had one institution, I think.

For the ASG, I think we can stop autoscaling manually by setting max instances = min instances in the ASG console, so it presumably doesn't really matter what the code is doing? I don't think we've ever actually used this kill switch in practice. Similarly, I think `workspaceDesiredHostCount` can be overridden by ASG instance settings, should we ever need it.

nwalters512:
I tested all these changes in our staging environment, and all services (PL, workspaces, graders) are working correctly 🎉

PR #7466, opened 2023-04-01
Bump package versions for NodeJS

mwest1066:
@nwalters512 what happens with building new versions of packages that have had their dependencies upgraded? Does that happen automatically or do I need to do something to trigger it?


package.json
```diff
@@ -54,28 +54,29 @@
         "clipboard": "^2.0.11",
         "cookie-parser": "^1.4.6",
         "crypto-js": "^4.1.1",
-        "csv-parse": "^5.3.5",
+        "csv-parse": "^5.3.6",
+        "csv-stringify": "^6.2.4",
```
nwalters512, line 58: 👀 is this a new dep?

nwalters512 (approved):
LGTM! If you want to release new versions of packages, run `make changeset` and follow the prompts. Skimming the packages, I think `prettier-plugin-sql` and `opentelemetry` are the only ones for which we should publish new versions.

Before merging, could you run `yarn dedupe` and verify that `yarn.lock` doesn't change?

package.json
jonatanschroeder: Seems like an old dependency removed in #7284 . Should probably be removed here.
```suggestion
```

mwest1066: Good catch, thanks, applied.

mwest1066:
> Before merging, could you run `yarn dedupe` and verify that `yarn.lock` doesn't change?

Done in https://github.com/PrairieLearn/PrairieLearn/pull/7466/commits/2e9d7ccd7e636e89498b8b7028e5fa311e3dce31 (there were some changes).

PR #7468, opened 2023-04-03
Add repository field to package.json files

nwalters512:
This should allow Dependabot to surface inline changelogs in PrairieTest when updating these dependencies.


mwest1066 (approved):
Neat!

PR #7477, opened 2023-04-04
Only update course commit hash after successful sync

nwalters512:
This fixes a syncing issue that came up recently. If a sync failed before chunks could be generated, we'd be left with new course code on disk and the latest commit hash in the DB. However, on the next sync, we'd use that new commit hash when deciding what chunks to regenerate, which means we'd "lose" any changes that had been made in the last sync. This same kind of issue could occur if a server died in the middle of performing a sync.

By only updating the commit hash in the database _after_ chunks have been generated, we ensure that a future successful sync will include the previous changes.

This change has a potential side effect: the fact that we now don't update the hash means that the question render cache may serve state entries until a successful sync occurs. I think this is fine; we don't guarantee that things will behave sensibly after a failed sync. If we want things to behave more predictably, we could change the "pull and sync" processes to reset the on-disk revision to the previous revision if a sync fails.

### DIFF
--- a/lib/editors.js
+++ b/lib/editors.js
@@ -260,16 +260,16 @@ class Editor {
           arguments: ['push'],
           working_directory: this.course.path,
           env: gitEnv,
-          on_success: _updateCommitHash,
+          on_success: _getEndCommitHash,
           on_error: _cleanupAfterCommit,
           no_job_sequence_update: true,
         };
         serverJobs.spawnJob(jobOptions);
       };
 
-      const _updateCommitHash = () => {
+      const _getEndCommitHash = () => {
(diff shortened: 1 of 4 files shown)


mwest1066:
> This change has a potential side effect: the fact that we now don't update the hash means that the question render cache may serve state entries until a successful sync occurs. I think this is fine; we don't guarantee that things will behave sensibly after a failed sync. If we want things to behave more predictably, we could change the "pull and sync" processes to reset the on-disk revision to the previous revision if a sync fails.

I'm generally ok with this, given that we will be in an undefined state in any case if a sync fails. I wouldn't object to trying to clean up better by reverting the on-disk state, although that won't help in cases where a server dies during sync.

PR #7480, opened 2023-04-04
Fix require of lru-cache

nwalters512:
I broke this accidentally in #7428. TypeScript lied to me, so I didn't notice; see https://github.com/isaacs/node-lru-cache/issues/289.

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -2,7 +2,7 @@
 const ERR = require('async-stacktrace');
 const redis = require('redis');
 const redisLru = require('redis-lru');
-const { LRUCache } = require('lru-cache');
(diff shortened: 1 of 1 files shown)


mwest1066 (approved):

mwest1066:
> I broke this accidentally in #7428. TypeScript lied to me, so I didn't notice; see [isaacs/node-lru-cache#289](https://github.com/isaacs/node-lru-cache/issues/289).

Very annoying! This is opposite of what we want from types!

isaacs:
You can also just do `const LRUCache = require('lru-cache')`, if upgrading to the fixed version doesn't resolve it.

PR #7481, opened 2023-04-05
Use LRUCache named export from lru-cache

nwalters512:
Followup to #7480; the named export is more elegant than tacking a `.default` on the end. The change from https://github.com/isaacs/node-lru-cache/commit/798fef3902e4a1a9d7ba106b285394bd93476022 allows us to do this!


mwest1066 (approved):
Nice! I'm glad you followed up with upstream!

PR #7486, opened 2023-04-07
Allow named locks to automatically renew themselves

nwalters512:
This will be useful for migrations (and our upcoming batched migrations), which might take a long time to execute.

### DIFF
--- a/packages/migrations/src/index.ts
+++ b/packages/migrations/src/index.ts
@@ -11,10 +11,23 @@ const sql = sqldb.loadSqlEquiv(__filename);
 export async function init(migrationDir: string, project: string) {
   const lockName = 'migrations';
   logger.verbose(`Waiting for lock ${lockName}`);
-  await namedLocks.doWithLock(lockName, {}, async () => {
-    logger.verbose(`Acquired lock ${lockName}`);
-    await initWithLock(migrationDir, project);
-  });
+  await namedLocks.doWithLock(
+    lockName,
+    {
+      // Migrations *might* take a long time to run, so we'll periodically
+      // renew the lock so that our lock doesn't get killed by the Postgres
+      // idle session timeout.
+      //
+      // That said, we should generally try to keep migrations executing as
+      // quickly as possible. A long-running migration likely means that
+      // Postgres is locking a whole table, which is unacceptable in production.
+      renewPeriodMs: 60_000,
+    },
+    async () => {
+      logger.verbose(`Acquired lock ${lockName}`);
+      await initWithLock(migrationDir, project);
+    }
+  );
   logger.verbose(`Released lock ${lockName}`);
 }
 
--- a/packages/named-locks/src/index.ts
+++ b/packages/named-locks/src/index.ts
@@ -4,21 +4,22 @@ import { PoolConfig } from 'pg';
 
 interface Lock {
   client: PoolClient;
+  intervalId: NodeJS.Timeout | null;
 }
 
 interface LockOptions {
   /** How many milliseconds to wait (anything other than a positive number means forever) */
   timeout?: number;
+  /**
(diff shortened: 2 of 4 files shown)


packages/named-locks/src/index.ts
```diff
@@ -95,7 +96,7 @@ export async function close() {
  * @param name The name of the lock to acquire.
  */
 export async function tryLockAsync(name: string): Promise<Lock | null> {
-  return getLock(name, { wait: false });
+  return getLock(name, { timeout: 0 });
```
nwalters512, line 99: I simplified this a bit by making `timeout: 0` imply "don't wait"; any non-zero timeout means we should in fact wait.

packages/migrations/src/index.ts
```diff
@@ -11,10 +11,23 @@ const sql = sqldb.loadSqlEquiv(__filename);
...
+    lockName,
+    {
+      // Migrations *might* take a long time to run, so we'll periodically
+      // renew the lock so that our lock doesn't get killed by the Postgres
+      // idle session timeout.
+      //
+      // That said, we should generally try to keep migrations executing as
+      // quickly as possible. A long-running migration likely means that
+      // Postgres is locking a whole table, which is unacceptable in production.
+      renewPeriodMs: 60_000,
```
mwest1066, line 24: Rather than having to specify the period here, I'm wondering whether we should just have `autoRenew` be a boolean option. I can't see why different callers would ever want to use different renew periods -- we will just always want the renew period to be longer than the DB idle in transaction timeout, which is a global setting. If we ever change the DB timeout then we'll need to go through and change all the lock renew timeouts to be less than the new DB value.

I thus think we should have a global config option for lock renew timeout, and lock callers can request whether they want auto-renew or not.

nwalters512: That's a fair point! I made this configurable in 9af56c7 via `init()` and added a `namedLocksRenewIntervalMs` config option that defaults to a minute.

packages/named-locks/README.md
```diff
@@ -0,0 +1,79 @@
+# `@prairielearn/named-locks`
```
mwest1066, line 1: Thanks for adding docs! 🙏

server.js
```diff
@@ -1973,7 +1973,9 @@ if (config.startServer) {
 
         // Our named locks code maintains a separate pool of database connections.
         // This ensures that we avoid deadlocks.
-        await namedLocks.init(pgConfig, idleErrorHandler);
+        await namedLocks.init(pgConfig, idleErrorHandler, {
+          renewIntervalMs: config.namedLocksRenewIntervalMs,
```
mwest1066, line 1977: I like this method of passing config variables into packages during `init()`

packages/named-locks/src/index.ts
```diff
@@ -65,14 +74,17 @@ interface LockOptions {
...
 /**
  * Initializes a new {@link PostgresPool} that will be used to acquire named locks.
  */
 export async function init(
   pgConfig: PoolConfig,
-  idleErrorHandler: (error: Error, client: PoolClient) => void
+  idleErrorHandler: (error: Error, client: PoolClient) => void,
+  namedLocksConfig: NamedLocksConfig = {}
 ) {
+  renewIntervalMs = namedLocksConfig.renewIntervalMs ?? renewIntervalMs;
```
mwest1066, line 87:
```suggestion
  renewIntervalMs ??= namedLocksConfig.renewIntervalMs;
```

mwest1066 (approved):

packages/named-locks/src/index.ts
nwalters512: I don't think this operator does what we want. This will assign `renewIntervalMs` only if `renewIntervalMs` is `null`/`undefined`, but that will never be true because we always initialize it with a default value.

PR #7490, opened 2023-04-08
Replace xml2js with fast-xml-parser

jonatanschroeder:
Given https://github.com/Leonidas-from-XIV/node-xml2js/issues/663, this PR replaces the use of xml2js with fast-xml-parser. No testing has been performed yet (other than ensuring that a call to the function doesn't crash the application if LTI is not enabled), I'd appreciate some pointers as to how this could be tested from others.

If the original issue is fixed and a new version of xml2js is deployed this PR may be closed, unless the change in libraries provides some other advantage (e.g., fast-xml-parser has a smaller footprint, and is maintained more frequently).

Note that xml2js is still a dependency of aws-sdk-js, so this PR won't replace all uses of the deprecated library.


mwest1066:
@trombonekenny can you look at this?

jonatanschroeder:
Closing in favor of #7494.

PR #7492, opened 2023-04-10
Allow interpolating BigInt values in @prairielearn/html

nwalters512:
I'm making use of `BigInt` values over in #7292; this will allow me to easily interpolate them into HTML templates.

### DIFF
--- a/packages/html/src/index.test.ts
+++ b/packages/html/src/index.test.ts
@@ -11,6 +11,14 @@ describe('html', () => {
     assert.equal(html`<p>${'cats'} and ${'dogs'}</p>`.toString(), '<p>cats and dogs</p>');
   });
 
+  it('interpolates a number', () => {
+    assert.equal(html`<p>${123}</p>`.toString(), '<p>123</p>');
+  });
+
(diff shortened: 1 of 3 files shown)


packages/html/src/index.test.ts
```diff
@@ -11,6 +11,14 @@ describe('html', () => {
     assert.equal(html`<p>${'cats'} and ${'dogs'}</p>`.toString(), '<p>cats and dogs</p>');
   });
 
+  it('interpolates a number', () => {
+    assert.equal(html`<p>${123}</p>`.toString(), '<p>123</p>');
+  });
+
+  it('interpolates a bigint', () => {
+    assert.equal(html`<p>${123n}</p>`.toString(), '<p>123</p>');
```
jonatanschroeder, line 19: May be safer to use a value not supported by regular Number, like a 16+ digit number? Just to be safe?

jonatanschroeder: Please ignore. I didn't notice it had been merged already, and this is not that important.

PR #7493, opened 2023-04-10
Add Cross-Origin Isolation configuration support per question

Nydauron:
This change adds an optional field to the question `info.json` named `crossOriginIsolated` of type bool. It defaults to `false` as Cross-Origin Isolation (COI) was not configured on PL universally and ensures existing questions continue to work. The purpose of making a page Cross-Origin isolated allows for the use of `SharedArrayBuffer`, `performance.measureUserAgentSpecificMemory()`, and high-resolution timers with better precision.

When the value is set, the routes:
- `/pl/course_instance/:course_instance_id/instructor/question/:question_id/preview`
- `/pl/course_instance/:course_instance_id/instance_question/:instance_question_id`
- `/pl/course/:course_id/question/:question_id/preview`

all have the COI middleware applied (i.e. `Cross-Origin-Opener-Policy` is set to `same-origin` and `Cross-Origin-Embedder-Policy` is set to `require-corp`).

All other resources that have the same site/origin (i.e. that come from PL like in `clientFilesQuestion/` and `clientFilesCourse/`) will not be affected by COI as they already pass CORS (`Access-Control-Allow-Origin: *`).

However, if a question were to request documents from a cross-origin site, the resource needs to either have the appropriate `CORS` header set or the `CORP` header set to `cross-origin`, otherwise the browser will fail to load the resource.

Something that might be considered is adding an option to send developer reports with [Chrome's reporting API](https://developer.chrome.com/articles/reporting-api/) to help developers figure out why certain documents might be being blocked, but this is more of a smaller QoL feature.


Nydauron:

Nydauron:
recheck

middlewares/coi.js
nwalters512, line 1: `coi` doesn't have any meaning to someone coming across a require of this file. Consider `crossOriginIsolatedQuestion` instead?

migrations/20230410200930_questions__cross_origin_isolated__add.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE questions
+ADD COLUMN IF NOT EXISTS cross_origin_isolated boolean DEFAULT false;
```
nwalters512, line 2: Needs a `NOT NULL` constraint.

sprocs/sync_questions.sql
```diff
@@ -163,7 +163,8 @@ BEGIN
         workspace_enable_networking = (src.data->>'workspace_enable_networking')::boolean,
         workspace_environment = (src.data->>'workspace_environment')::jsonb,
         sync_errors = NULL,
-        sync_warnings = src.warnings
+        sync_warnings = src.warnings,
+        cross_origin_isolated = (src.data->>'cross_origin_isolated')::boolean
```
nwalters512, line 167: This should come before `sync_errors` (it doesn't really matter, just convention within syncing code to have errors/warnings last).

sync/fromDisk/questions.js
```diff
@@ -48,6 +48,7 @@ function getParamsForQuestion(q) {
     workspace_url_rewrite: q.workspaceOptions && q.workspaceOptions.rewriteUrl,
     workspace_enable_networking: q.workspaceOptions && q.workspaceOptions.enableNetworking,
     workspace_environment: q.workspaceOptions?.environment ?? {},
+    cross_origin_isolated: q.crossOriginIsolated === undefined ? false : q.crossOriginIsolated,
```
nwalters512, line 51:
```suggestion
    cross_origin_isolated: q.crossOriginIsolated ?? false,
```

middlewares/crossOriginIsolation.js
```diff
@@ -0,0 +1,13 @@
+// For the given question, sets Cross-Origin-Opener-Policy to 'same-origin' and
+// Cross-Origin-Embedder-Policy: 'require-corp' if cross_origin_isolated is set to true
+//
+// Beneficial for questions that rely on SharedArrayBuffers.
```
nwalters512, line 4: We generally prefer `/** */` comments so that Intellisense and other IDE tools can pick up on it. Also editing to focus more on the "why" than the "what" since this is such a simple function.

```suggestion
/**
 * Allows questions to opt in to cross-origin isolation. This is useful for
 * questions that rely on advanced features like `SharedArrayBuffer`.
 *
 * @see https://developer.mozilla.org/en-US/docs/Web/API/crossOriginIsolated
 */
```

nwalters512:
@Nydauron would you still like to see this land? Sorry we've been slacking on this, I'm trying to get a handle on our oldest PRs.

Nydauron:
> @Nydauron would you still like to see this land? Sorry we've been slacking on this, I'm trying to get a handle on our oldest PRs.

As an update on my situation, it's been about a year since I've been on the Spimbot team, so I am unsure of the current state where the team is on porting Spimbot to PL.

I still think that it would be good to have this as an option in case the team makes further progress or if other courses are thinking about using multithreaded WASM, so I am fine with tying things up here and getting this merged.

nwalters512:
I talked to @glherman, it sounds like the effort to move spimbot onto PL is suspended as he isn't teaching CS 233 at the moment. Given that there's no immediate need for this functionality, I'm going to do the annoying PM thing and close this for now so that we aren't merging code that no one is using. If/when spimbot work picks up again, I'm happy to reconsider this.

PR #7496, opened 2023-04-11
Show external grader test results with zero points as gray

### LINKED ISSUE #7484: For non-points earning test cases, update the display for pl-external-grader-results to show [0/0] in a muted gray color next to the test case's name
In the grading results screen, non-points earning test cases show only the test case name. Meanwhile, when students make full credit on a test case, [x/x] is shown in green next to the test case's name; [y/x] is shown in red next to the test case's name. This issue proposes that non-points earning test cases show [0/0] in a muted gray color next to the test case's name.

jonatanschroeder:
Resolves #7484.

image

### DIFF
--- a/elements/pl-external-grader-results/pl-external-grader-results.py
+++ b/elements/pl-external-grader-results/pl-external-grader-results.py
@@ -139,17 +139,20 @@ def render(element_html, data):
                             test["max_points"] = round_value(
                                 results_test.get("max_points")
                             )
-                            correct = test["max_points"] == test["points"]
 
-                            # Don't show points for test cases that are 0/0.
-                            # Note that we compare with a string instead of a
-                            # number because we converted it to a string with
-                            # the above call to `round_value()`.
-                            if correct and test["max_points"] == "0":
-                                test["show_points"] = False
+                            if test["max_points"] != test["points"]:
+                                test["results_color"] = "#F44336"
+                                test["results_icon"] = "fa-times"
+                            # Don't consider points for test cases that are 0/0
+                            # to be correct. We compare with a string because
+                            # `round_value()` returns a string.
+                            elif test["max_points"] == "0":
+                                test["results_color"] = "rgba(0, 0, 0, 0.3)"
+                                test["results_icon"] = "fa-circle-info"
+                            else correct:
+                                test["results_color"] = "#4CAF50"
+                                test["results_icon"] = "fa-check"
 
-                            test["results_color"] = "#4CAF50" if correct else "#F44336"
-                            test["results_icon"] = "fa-check" if correct else "fa-times"
                         test["images"] = results_test.get("images", [])
                         test["has_images"] = len(test["images"]) > 0
                         tests.append(test)
--- a/elements/pl-external-grader-results/pl-external-grader-results.py
+++ b/elements/pl-external-grader-results/pl-external-grader-results.py
@@ -149,7 +149,7 @@ def render(element_html, data):
                             elif test["max_points"] == "0":
                                 test["results_color"] = "rgba(0, 0, 0, 0.3)"
                                 test["results_icon"] = "fa-circle-info"
-                            else correct:
+                            else:
                                 test["results_color"] = "#4CAF50"
                                 test["results_icon"] = "fa-check"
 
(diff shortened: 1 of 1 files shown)


elements/pl-external-grader-results/pl-external-grader-results.py
```diff
@@ -139,17 +139,20 @@ def render(element_html, data):
...
-                            correct = test["max_points"] == test["points"]
 
-                            # Don't show points for test cases that are 0/0.
-                            # Note that we compare with a string instead of a
-                            # number because we converted it to a string with
-                            # the above call to `round_value()`.
-                            if correct and test["max_points"] == "0":
-                                test["show_points"] = False
+                            if test["max_points"] != test["points"]:
+                                test["results_color"] = "#F44336"
```
eliotwrobson, line 144: Are these standard (named) colors? Would be nice to use the mapping in `colors.py` to avoid having the raw color hex value in these strings. Not a huge deal but just a thought.

eliotwrobson (approved):
Looks good!

elements/pl-external-grader-results/pl-external-grader-results.py
jonatanschroeder: These were the colors originally used here (see removed line 151 below). I did a quick script to find the closest colors from that list, and didn't get that close:
* for incorrect: tomato (ff6347)
* for correct: mediumseagreen (3cb371)

@mwest1066 or @nwalters512, should I switch these colors for named colors?

elements/pl-external-grader-results/pl-external-grader-results.py
```diff
@@ -139,17 +139,20 @@ def render(element_html, data):
...
+                                test["results_icon"] = "fa-times"
+                            # Don't consider points for test cases that are 0/0
+                            # to be correct. We compare with a string because
+                            # `round_value()` returns a string.
+                            elif test["max_points"] == "0":
+                                test["results_color"] = "rgba(0, 0, 0, 0.3)"
```
nwalters512, line 150: Does this meet the [WCAG color contrast requirements](https://webaim.org/resources/contrastchecker/)?

elements/pl-external-grader-results/pl-external-grader-results.py
nwalters512: We haven't touched this code in the five years since I wrote it. I'm not opposed to changing it, but I see extraordinarily little value in doing so.

elements/pl-external-grader-results/pl-external-grader-results.py
jonatanschroeder: It does not. I'll play a bit with some numbers. That said, the green color for correct doesn't pass it either (red passes only the WCAG AA test for large text, which given this is bold/16pt is ok here).

jonatanschroeder: Here's with CSS colors, crimson, darkgreen and dimgray. These all meet the standards.

image

nwalters512: IMO the red/green there are borderline too dark to be distinguishable from each other, and the green appears almost black (caveat: my optometrist has said I am slightly colorblind with blues/greens). Are there lighter shades that still pass contrast requirements?

jonatanschroeder: image

These are E70000, 008900 and 757575 for red, green and grey, respectively. These still pass the requirements. Not yet pushed, let me know (or maybe we should get someone else's opinion, since I'm also colorblind 😬 )?

trombonekenny: When we revisit this in earnest (it that this issue?), we will want to move away from red/green contrast explicitly because it's the most common form of colorblindness.

nwalters512: The good news is that we don't *solely* rely on color to convey pass/fail/info - we have icons and points as well.

@jonatanschroeder I think that looks improved - can you push that so I can play with it locally?

jonatanschroeder: It has been pushed in e728fd0.

nwalters512 (approved):

PR #7498, opened 2023-04-11
Deduplicate dependencies

nwalters512:
Followup to #7497. I really need to make `yarn dedupe` part of CI...


mwest1066 (approved):
It would be so nice if we could automatically run `yarn dedupe` somehow! I get caught by this all the time.

jonatanschroeder:
TIL about `yarn dedupe`...

PR #7502, opened 2023-04-11
Add support for CHECK constraints to database description tooling

nwalters512:
I discovered in #7292 that our tooling didn't support `CHECK` constraints in the descriptions it generates. This PR adds support for that. It turns out there were already some existing constraints that weren't captured yet!


mwest1066 (approved):
Very nice, thanks!

Issue #1356, opened 2018-10-27
Display deleted questions for exams


mwest1066:
On all assessment instances views (instructor and student), for exams we should show all `instance_questions`, even if the corresponding `assessment_question` has been deleted. This is to handle the case where the exam has been modified to delete a question, but some exams were already complete and we want to view them as they were, not with the new exam config.

mwest1066:
Actually, I can't see why this bug happens. Maybe it's been fixed? To test:

1. Start an Exam as a student
2. Edit the exam as an Instructor to remove one of the questions on the exam (don't delete the question itself, just take it off the exam)
3. As a student again, reload the exam page and see whether the question is still visible (we want it to be visible)
4. As an instructor, look at the "Details" page for the student assessment and see whether the question is still visible (we want it to be visible)

mwest1066:
From #952 it looks like the problem is actually in `sprocs/question_order.sql`

We need to patch:
```
-    AND aq.deleted_at IS NULL
+    AND (a.type = 'Exam' OR aq.deleted_at IS NULL)
```
We need to change studentAssessmentInstanceExam.sql (and maybe other files) so they don't use alternative groups, because these will be deleted for deleted questions. We should either add a deleted_at flag to alternative groups, so they can persist, or we should add a zone_id to assessment questions so they don't need to join over alternative groups to find the zone.

PR #7506, opened 2023-04-12
Forbid access to deleted assessments and questions

### LINKED ISSUE #7505: Forbid access to deleted assessments
If someone navigates to any of the instructor assessment pages by ID, we don't check for `deleted_at IS NULL` before loading the page. This can lead to other errors as various files or data that we expect to be present can't be found. We should 404 for deleted assessments.

nwalters512:
Closes #7505.

It's possible that more of our `selectAndAuthz*` middlewares need something similar? I spot-checked and it seems like they're mostly doing the right thing. A few I'm not sure about:

- `selectAndAuthzInstanceQuestion`: should this require `q.deleted_at IS NULL`?
- `selectAndAuthzAssessmentQuestion` - should this require `q.deleted_at IS NULL AND aq.deleted_at IS NULL`?


mwest1066:
> * `selectAndAuthzInstanceQuestion`: should this require `q.deleted_at IS NULL`?
> * `selectAndAuthzAssessmentQuestion` - should this require `q.deleted_at IS NULL AND aq.deleted_at IS NULL`?

Yes to all of these, I think.

middlewares/selectAndAuthzInstanceQuestion.sql
```diff
@@ -155,5 +155,6 @@ WHERE
     $assessment_id::BIGINT IS NULL
     OR a.id = $assessment_id
   )
+  AND q.deleted_at IS NULL
```
mwest1066, line 158: Do we also want checks on `aq.deleted_at` and `a.deleted_at` here? Conceivably also `c.deleted_at`? I'm not sure how high up the chain it's worth enforcing?

nwalters512: The first two make sense! I think `c.deleted_at` will (should) always be caught by `middlewares/authzCourseOrInstance`?

nwalters512: I added checks for `aq.deleted_at` and `a.deleted_at` in b0c6979.

mwest1066: Good point on course access being checked already.

mwest1066 (approved):

middlewares/selectAndAuthzInstanceQuestion.sql
jonatanschroeder: Is there a problem here with using `aq.deleted_at`? What happens if a student opens a question that was active when they took the assessment, but deleted afterwards?

nwalters512: We check for `aq.deleted_at` in many places already (grep for `aq.deleted_at IS NULL`), so this is just making it consistent with everywhere else. If an assessment question is deleted, I think the correct thing to do is to not show it to students and forbid access to it even if they manage to guess the URL. Does that sound right to you?

jonatanschroeder: That makes sense. I didn't check the list, I was just confirming that this is the intended behaviour, particularly for exams, since they don't regenerate instance questions when the assessment changes.

mwest1066: This has been a sore point for a long time, actually. What happens is that we are running an exam in the CBTF, the instructor realizes on day 2 that one question in an `alternative_group` is partially broken, and they want to remove this question from the assessment so that future students won't be assigned it. However, they want existing exams taken on day 1 to still show the question that the student did, especially because the question might have worked fine for these students (and we want to show how their score was computed).

I found the old issue about this: https://github.com/PrairieLearn/PrairieLearn/issues/1356

In summary, I think we want to allow access to instance questions even though the assessment question has been deleted. However, handling this cleanly will probably require some more code changes beyond just changing `deleted_at` filters, because we'll need to track the `alternative_group` and `zone` information that might have been deleted.

mwest1066: Yes, let's _not_ add `aq.deleted_at IS NULL` here.

nwalters512: Removed that check in fd5f67e.

nwalters512: Whoops, there was one more that I removed in 2815b39.

mwest1066 (approved):

Issue #3869, opened 2021-02-23
Reorganize the deployment of experimental features


jonatanschroeder:
Based on the current list of experimental features partially deployed (group work, manual grading, pl-order-blocks, etc.) and being deployed soon (force order, broken questions, gradebook, etc.), I propose that we include a method for allowing the partial release of these features in a controlled way.

A table of features should be created, with some internal identification (e.g., `manual_grading`) and a user-friendly name and description suitable for instructors/course owners. To avoid discrepancies between installations these features should be created via migration script.

A feature may be enabled or disabled for a course (or maybe course instance), see below. A locals variable (e.g., `features`) should be populated by the course/course instance middleware, and all feature implementations should check if their feature record is enabled before displaying links or data. For some features this may require an additional check with an exception if the user attempts an operation that can only be performed if the feature is enabled. Some features, even if enabled, could also show a warning to instructors (e.g., in question preview, if workspace is used, a "workspace is experimental" warning could be shown).

Features can be in various stages:

* Private beta: in this case admins have the ability (via a page in the Admins view) to enable/disable these features for individual courses. Once enabled, course owners can then disable the feature if they would like to remove it for some reason, but they can't enable it unless the admin has enabled it as well.
* Public beta: in this case all course owners have the ability (via a separate tab in the course or course instance) to enable the feature. All public features are listed (as well as private features enabled for that course), and owners can enable specific features. On enabling a feature, a warning is given that the feature may include bugs or be incomplete (wording to be discussed), and may be subject to frequent changes with minimal warning.
* Release candidate: same as the public beta, but the feature is enabled by default.
* Fully deployed: in this case the feature is enabled for all courses. The feature is maintained until a "garbage collection" PR is able to remove all checks for the feature in the implementation.

There should also be a way for admins to list all courses that have a feature enabled, so that they can track its use.

nwalters512:
As of https://github.com/PrairieLearn/PrairieLearn/pull/7507, we have feature flags. They aren't user-visible and can't be used by instructors to opt themselves into features, but this is still much better than we had before. If we decide we want to allow self-service opting-in of features, this can be tracked in a separate issue.

PR #7507, opened 2023-04-12
Add infrastructure for feature flags

nwalters512:
This PR creates some foundational infrastructure for feature flags, which we're calling "features". Features are granted based on a hierarchy of institution, course, and course instance ("entities"). Features may also target a user, either in isolation or in combination with an entity.

We're intentionally trying to keep a simple mental model where features can never be explicitly disabled for a certain entity/user. They can only be explicitly enabled; the lack of any specific enabled flag for a given user means it is implicitly disabled. This does mean that we can't currently express something like "enable feature X for an institution but disable it for specific course within that institution".

We're intending to build various bits of subscription functionality on top of this, which will require us to do things like enable a feature (e.g. workspaces) for some specific combinations of user and course instance. In the future, we could also implement non-user-specific subscriptions at other levels of this hierarchy as well.

So far I've only built basic read functionality to get a sense of how admins will be able to deal with this somewhat complex hierarchy.

Screenshot 2023-04-13 at 16 03 52

This is going to scale _incredibly_ poorly once we start actually filling out this table, especially once we start enabling features by default for all courses or getting large numbers of subscription-enabled features. I imagine we'll want a few more ways of interacting with features:

- Some kind of tree-like exploration of institution, click into course, click into course instance, and for each entity see what flags are enabled and maybe which are already enabled by parent entities.
- A way to look at all the features for a given user.
- For all the above, a way to manually apply features.

I didn't design this in a reusable package, as it's currently pretty tied to our specific hierarchy of entities (institution / course / course instance). Our indexes and constraints also mean we might be in trouble if we ever want to introduce a parallel hierarchy of entities to control features. One way around this would be to replace `institution_id`/`course_id`/`course_instance_id` with a single `key` column that has values like `course:1234` or `course_instance:57`. This is more flexible at the cost of losing referential integrity 😢


Makefile
```diff
@@ -35,7 +35,7 @@ start-s3rver:
 test: test-js test-python
 test-js: test-prairielearn test-prairielib test-grader-host test-workspace-host test-packages
 test-prairielearn: start-support
-	@yarn mocha --parallel "tests/**/*.test.{js,mjs}"
+	@yarn mocha --parallel "tests/**/*.test.{js,mjs}" "lib/**/*.test.{js,mjs}"
 test-prairielearn-serial: start-support
 	@yarn mocha "tests/**/*.test.{js,mjs}"
```
jonatanschroeder: You probably want these two targets to match, right?

lib/config.js
```diff
@@ -432,6 +432,8 @@ config.azureResourceURL = 'https://graph.windows.net';
 config.azureDestroySessionUrl =
   'https://login.microsoftonline.com/common/oauth2/logout?post_logout_redirect_uri=http://localhost:3000';
 
+config.enabledFeatures = ['course:manual-grading'];
```
nwalters512, line 435: TODO: remove this before merging.

Should this be an array or an object? An object would allow us to override specific features while still respecting any remaining defaults. For instance, if this were the following:

```ts
config.features = {
  'course:manual-grading': true,
};
```

We could add the following JSON config and keep the above defaults:

```json
{
  "features": {
    "course:lti": true
  }
}
```

lib/features/index.js
```diff
@@ -0,0 +1,6 @@
+// @ts-check
+const { FeatureManager, FeatureGrantType } = require('./manager');
+
+const features = new FeatureManager(['course:manual-grading']);
```
nwalters512, line 4: TODO: empty this out before merging.

```suggestion
const features = new FeatureManager([]);
```

lib/features/manager.js
```diff
@@ -0,0 +1,157 @@
...
+};
+
+/**
+ * Grants can have different types, which can be used to determine how they
+ * were applied. These are mostly meant to be consumed by administrators who
+ * need to be able to tell if a feature was enabled by default, set manually
+ * by another administrator, or enabled as part of a subscription plan.
+ *
+ * @enum {string}
+ */
```
nwalters512, line 28: Thoughts on this? Are there any other types of grants we should account for?

jonatanschroeder (commented):
Would it make sense to load these feature flags on a middleware and have them ready when needed instead of checking them individually? I guess it depends on how often these flags will be used, but just raising this as an option. Also, does this close #3869?

nwalters512:
> Would it make sense to load these feature flags on a middleware and have them ready when needed instead of checking them individually? I guess it depends on how often these flags will be used, but just raising this as an option.

We're generally trying to avoid the "spooky action at a distance" of middleware in new code, so I think we'll stick to loading them on-demand for now. If we start using/checking lots of them per request, we can possibly add some per-request caching with AsyncLocalStorage, or even longer-lived caching with Redis.

> Also, does this close https://github.com/PrairieLearn/PrairieLearn/issues/3869?

I think we can say it does. While it doesn't implement the tiered/staged feature rollout outlined there, I think what *is* implemented meets our needs at the moment. We can relatively easily build public betas with self-service opt-in on top of this code in the future if we want to.

migrations/20230412200134_feature_grants__create.sql
```diff
@@ -0,0 +1,45 @@
...
+  name,
+  type
+);
+
+-- Secondary index to allow looking up all users with a given feature flag.
+CREATE INDEX IF NOT EXISTS feature_grants_name_user_id_type_idx ON feature_grants (
+  name,
+  user_id,
+  type
+);
```
nwalters512, line 45: These probably aren't strictly necessary at this point, I'm just thinking ahead to various query patterns we'll want to have in the near future. I imagine this table could get quite large, though maybe not for a while. Some thoughts/comments:

- `type` is at the end because we might frequently want to filter out specific types (e.g. we might not want to show a list of all `subscription` feature grants on an admin page by default).
- We might be able to avoid the two user-specific indexes if we put `user_id` first in the other one?

pages/administratorFeatures/administratorFeatures.js
nwalters512, line 1: I deliberately did not build feature grant editing in this PR; that's gonna be somewhat complex and it'll hopefully be easier to review in an isolated followup PR.

tests/chunks.test.js
```diff
@@ -9,7 +9,7 @@ const sqldb = require('@prairielearn/postgres');
 const courseDB = require('../sync/course-db');
 const chunksLib = require('../lib/chunks');
 const config = require('../lib/config');
-const logger = require('./dummyLogger');
+const { makeMockLogger } = require('./mockLogger');
```
nwalters512, line 12: These changes and those in `helper*` were made in the interest of not duplicating our "sync the test course" logic again.

lib/features/manager.js
SethPoulsen: I'm not sure if 'grant' is the best word to use here, especially considering that most instructors using PL are probably primed to think about writing applications to the NSF for funding when they see the word "grant". Is there a different word we could use?

nwalters512: Instructors never see table names, or implementation details in general, so that shouldn't matter. I'm open to suggestions if you have any.

mwest1066: Just to pick a couple of semi-random examples, Hashicorp's RBAC Boundary product uses [grants](https://developer.hashicorp.com/boundary/docs/concepts/security/permissions/permission-grant-formats) for this concept. Kubernetes uses [RoleBinding](https://kubernetes.io/docs/reference/access-authn-authz/rbac/) for it's RBAC model.

apps/prairielearn/src/pages/administratorFeatures/administratorFeatures.html.ts
```diff
@@ -0,0 +1,169 @@
...
+            ${features.length > 0
+              ? html`<div class="list-group list-group-flush">
+                  ${features.map((feature) => {
+                    return html`
+                      <div class="list-group-item d-flex align-items-center">
+                        <a
+                          href="${resLocals.urlPrefix}/administrator/features/${feature}"
+                          class="mr-auto text-monospace"
+                        >
+                          ${feature}
```
mwest1066, line 47:
```suggestion
                          ${AdministratorFeature({feature, featureGrants, resLocals})}
```

apps/prairielearn/src/lib/features/index.ts
```diff
@@ -0,0 +1,5 @@
+import { FeatureManager, FeatureGrantType } from './manager';
+
+const features = new FeatureManager(['course:manual-grading']);
```
mwest1066, line 3: What does the `course:` mean here? Are we going to programmatically enforce that `course:` features can only be enabled for courses (and institutions)? If so, I'm worried about hiding `course:` inside the string where we aren't typechecking it. On the other hand, if we aren't planning to enforce something here, I wonder whether we should bother having the `course:` prefix at all?

mwest1066 (approved):

migrations/20230412200134_feature_grants__create.sql
mwest1066: One way around this would be to replace `institution_id`/`course_id`/`course_instance_id` with a single `key` column that has values like `course:1234` or `course_instance:57`. This is more flexible at the cost of losing referential integrity 😢

I'm in favor of keeping things tied to our particular hierarchy at the moment (and maintaining RI!).

mwest1066:
> One way around this would be to replace `institution_id`/`course_id`/`course_instance_id` with a single `key` column that has values like `course:1234` or `course_instance:57`. This is more flexible at the cost of losing referential integrity 😢

I'm in favor of keeping things tied to our particular hierarchy at the moment (and maintaining RI!).

apps/prairielearn/src/lib/features/index.ts
nwalters512: I was just playing around with naming conventions, I was going to drop this before this was merged anyways. We _had_ talked about programmatically annotating feature flags with a particular scope that they apply to (e.g. we could say rubrics are scoped to "course" and can thus only be applied to "course" or "institution"). I still think that might be a good idea for documentation and runtime safeguards?

lib/config.js
nwalters512: @mwest1066 thoughts on this? Both on data format, and more generally if it makes sense to be able to globally override feature flags via the config. I'm thinking primarily about local development, where folks may be regularly nuking their DB via container restarts; it'd be useful to be able to maintain flags in the persistent config file.

apps/prairielearn/src/pages/administratorFeatures/administratorFeatures.html.ts
nwalters512: Why? `AdministratorFeature` is a top-level page, we don't want to render it inline in _another_ page.

apps/prairielearn/src/lib/features/index.ts
nwalters512: Talked with @mwest1066 about this offline. We're going to avoid this for now in favor of just being careful. In the future, we can look at adding guardrails.

lib/config.js
mwest1066: I agree it would be nice to be able to override specific flags from `config.json` for local dev. The "merge objects" syntax seems nice for this.

lib/features/manager.js
mwest1066: This all sounds good to me. No further grant types that I can think of!

lib/config.js
nwalters512: This was changed in b918899.

PR #7508, opened 2023-04-12
Render booleans in html template strings

mwest1066:
Previously we silently discarded booleans in `html` template strings. This allowed the use of the pattern:
```
html`${bool_value && 'some string'}`
```
which would template in `some string` if `bool_value` was true and nothing otherwise. However, it also meant that it was easy to accidentally fail to render a boolean value when we actually wanted to do it. This PR fixes this, but it does mean that we'll need to replace the pattern above with:
```
html`${bool_value ? 'some string' : ''}`
```
I personally prefer the explicit nature of this anyway.

### DIFF
--- a/packages/html/src/index.test.ts
+++ b/packages/html/src/index.test.ts
@@ -19,6 +19,10 @@ describe('html', () => {
     assert.equal(html`<p>${123n}</p>`.toString(), '<p>123</p>');
   });
 
+  it('interpolates a boolean', () => {
+    assert.equal(html`<p>${false}</p>`.toString(), '<p>false</p>');
+  });
+
   it('escapes values when rendering array', () => {
     const arr = ['cats>', '<dogs'];
     assert.equal(
--- a/packages/html/src/index.test.ts
+++ b/packages/html/src/index.test.ts
@@ -50,10 +50,6 @@ describe('html', () => {
     );
   });
 
-  it('omits boolean values from template', () => {
-    assert.equal(html`<p>${true}${false}</p>`.toString(), '<p></p>');
-  });
-
   it('omits nullish values from template', () => {
     assert.equal(html`<p>${null}${undefined}</p>`.toString(), '<p></p>');
   });
--- a/packages/html/src/index.test.ts
+++ b/packages/html/src/index.test.ts
@@ -20,6 +20,7 @@ describe('html', () => {
   });
 
   it('interpolates a boolean', () => {
+    assert.equal(html`<p>${true}</p>`.toString(), '<p>true</p>');
     assert.equal(html`<p>${false}</p>`.toString(), '<p>false</p>');
   });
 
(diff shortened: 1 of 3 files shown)


packages/html/src/index.test.ts
```diff
@@ -19,6 +19,10 @@ describe('html', () => {
     assert.equal(html`<p>${123n}</p>`.toString(), '<p>123</p>');
   });
 
+  it('interpolates a boolean', () => {
+    assert.equal(html`<p>${false}</p>`.toString(), '<p>false</p>');
```
nwalters512, line 23:
```suggestion
    assert.equal(html`<p>${true}</p>`.toString(), '<p>true</p>');
    assert.equal(html`<p>${false}</p>`.toString(), '<p>false</p>');
```

nwalters512 (commented):
You should run `make changeset` and mark this as a breaking change.

Do you have plans for how to safely adopt this change in PL/PT? I'm thinking maybe we should temporarily (just in `node_modules`, not in the published package itself) remove `boolean` from the `HtmlValue` type; that should at least get TypeScript to yell at use everywhere we're currently using booleans. Then we can replace them with all with the `condition ? ... : ''` form. Then we upgrade to the latest version of this package and convert all `condition ? 'true' : 'false'` to just `condition`.

mwest1066:
> You should run `make changeset` and mark this as a breaking change.

Excellent, thanks, I was trying to remember what had to be done for this. Done now.

> Do you have plans for how to safely adopt this change in PL/PT? I'm thinking maybe we should temporarily (just in `node_modules`, not in the published package itself) remove `boolean` from the `HtmlValue` type; that should at least get TypeScript to yell at use everywhere we're currently using booleans. Then we can replace them with all with the `condition ? ... : ''` form. Then we upgrade to the latest version of this package and convert all `condition ? 'true' : 'false'` to just `condition`.

Good thinking. I did this in PL and as far as I'm aware it didn't show up any issues. I did both a `make typecheck` (which I think checks the main PL code?) and `make build` (which I think builds/checks all packages?) and both showed no issues (other than the use in the test cases for the `html` package itself). Is there anything else that should be checked here?

For PT we definitely still have cases of `&&` with booleans so I'll need to clean those up separately.

nwalters512:
> Is there anything else that should be checked here?

I don't think so! I've generally been good about adding `// @ts-check` + strong types in all new-style code, so fingers crossed that we've got everything. Even then, this has mostly been in admin-facing pages so far. I just wanted to make sure that you know to check this so that we don't end up with `false` scattered at random places in the UI 🙂

nwalters512 (approved):

mwest1066:
> I don't think so! I've generally been good about adding `// @ts-check` + strong types in all new-style code, so fingers crossed that we've got everything. Even then, this has mostly been in admin-facing pages so far. I just wanted to make sure that you know to check this so that we don't end up with `false` scattered at random places in the UI 🙂

Haha, right!

PR #7512, opened 2023-04-13
Downgrade bootstrap-table to 1.21.2

jonatanschroeder:
bootstrap-table 1.21.4 changed the way filters are applied in custom search functions, which broke some filter functionality in our pages. This downgrades it to the previous version until we have time to review these changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
         "body-parser": "^1.20.2",
         "bootstrap": "^4.6.2",
         "bootstrap-icons": "^1.10.3",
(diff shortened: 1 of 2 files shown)


jonatanschroeder:
Issue reported with bootstrap-table: https://github.com/wenzhixin/bootstrap-table/issues/6745

nwalters512 (approved):
Thanks for pinning this. Given the number of bugs we've seen introduced in patch releases of this packages, I think that's the way to go for now.

PR #7513, opened 2023-04-13
Backfill course_id column of variants table with batched migration

SethPoulsen:
Use a batched migration to backfill the new course_id column of the variants table.

This is in preparation for Question Sharing, when the course_id of variant is no longer determined by the course_id of the question, and so the course_id must be saved in the variants table.

This PR must be merged sometime after:
- the course_id column is created in the variants table: #7273
- the batched migration infrastructure is set up: #7292


batched-migrations/20240210165255_variants_course_id.ts
nwalters512, line 1: Despite the examples I show in the docs, PL does not yet support TypeScript :( we'll have to author these as a regular JavaScript file.

batched-migrations/20240210165255_variants_course_id.ts
```diff
@@ -0,0 +1,28 @@
...
+      max: result.rows[0].max,
+      batchSize: 1000,
+    };
+  },
+
+  async execute(min: bigint, max: bigint) {
+    await queryAsync(`
+      UPDATE variants AS v
+      SET
+        course_id = q.course_id
```
nwalters512, line 17:
```suggestion
        v.course_id = q.course_id
```

batched-migrations/20240210165255_variants_course_id.ts
```diff
@@ -0,0 +1,28 @@
...
+        questions AS q
+      WHERE
+        v.course_id = NULL AND
+        v.question_id = q.id AND
+        id >= $min AND
+        id <= $max`,
+      {min, max}
+    );
+  },
+});
```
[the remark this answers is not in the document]
SethPoulsen: done

migrations/20240210165255_variants__course_id__backfill.js
nwalters512, line 1: Seeing `2024` in the timestamp is a bit unexpected. Did you use the following command from https://github.com/PrairieLearn/PrairieLearn/blob/master/migrations/README.md#creating-a-migration?

```sh
node -e "console.log(new Date().toISOString().replace(/\D/g,'').slice(0,14))"
```

SethPoulsen: LOL no. When I have a PR that is WIP I sometimes just put the timestamp way ahead so that I don't have to keep editing it when new PRs get merged. Then when its closer to ready I put an actual timestamp on. I'll change these once done debugging.

nwalters512: Fair enough! FWIW, if you want to save yourself the trouble, we don't mandate that all new migrations have timestamps ahead of all the ones on master (that was a primary motivation for switching to timestamps from sequential integer indexes). As long as any migrations on which this one depends are ordered before it, you'll be good.

SethPoulsen: Ah, that's cool. I must admit (as is now apparent I suppose) I didn't spend enough time looking at the new migration system to understand it.

nwalters512:
Going to mark this as "do not merge". We need to ensure that #7273 has been fully deployed to all environments before landing this.

nwalters512 (approved):

PR #7514, opened 2023-04-13
Replace bootstrap-table with datatables in questions page

jonatanschroeder:
Also moves all relevant JS to an asset script.


nwalters512:
I haven't looked at the code yet, but I have a few initial UI/UX concerns.

Screenshot 2023-04-13 at 17 00 42

The padding/alignment/borders are all over the place:

- "Showing 0 to 0 of 0 entries" and the adjacent buttons aren't vertically centered and have inconsistent padding; ditto for the table footer.
- The up/down sorting arrows aren't aligned with the adjacent inputs.
- There are weird doubled-up borders to the left/right of various components.

Are these something we have control over, or are they baked into this library's styles/markup?

I can't use <kbd>Ctrl + A</kbd> in the filters to select all the text so I can hit Backspace to clear it. That's already sort of broken in `bootstrap-table`, but it's _completely_ broken here.

The "Columns" dropdown makes the container scroll instead of overflowing, and there's some weird horizontal escaping of the selected indicators:

Screenshot 2023-04-13 at 17 04 00

jonatanschroeder:
@nwalters512 @mwest1066 thinking about this more broadly, here are the options for bootstrap-table support:

* replace it with datatables as in this PR. This has a few kinks as listed above, which I think can be tweaked, but since filters here are custom I'm not sure it's worth the effort.
* update the bootstrap-table code to fix the issue introduced by the latest update. I was playing a bit with it and it should be possible. The problem is that, based on recent experience, the library has been known to add breaking changes, so I'd prefer to avoid it.
* consider another library (e.g., [tabulator](https://www.npmjs.com/package/tabulator-tables)). Tabulator looks interesting, seems to be very robust, has a good API, and has robust filter options embedded. It also has pretty good pagination controls,  One limitation I found is that it doesn't have a native UI for user-defined visible columns, but this should be relatively easy to implement in custom form, particularly since the API allows the ability to show/hide columns to be done really easily.

In any case, I'm closing this PR. I'll play a bit more with tabulator and see if it makes sense to switch. If that doesn't pan out, I'll move on with bootstrap-table.

PR #7517, opened 2023-04-14
style: pl-matrix-latex uplift

eliotwrobson:
See title. Lifted from #7198.


python/prairielearn.py
```diff
@@ -1515,21 +1520,21 @@ def latex_from_2darray(A, presentation_type="f", digits=2):
     Otherwise, each number is formatted as '{:.{digits}{presentation_type}}'.
     """
     # if A is a scalar
-    if np.isscalar(A):
```
eliotwrobson, line 1518: This is a minor mistake that appears in a bunch of places, see the notes on the docs for `isscalar`: https://numpy.org/doc/stable/reference/generated/numpy.isscalar.html

python/prairielearn.py
```diff
@@ -1547,8 +1552,8 @@ def latex_from_2darray(A, presentation_type="f", digits=2):
         .splitlines()
     )
     rv = [r"\begin{bmatrix}"]
-    rv += ["  " + " & ".join(line.split()) + r"\\" for line in lines]
-    rv += [r"\end{bmatrix}"]
+    rv.extend("  " + " & ".join(line.split()) + r"\\" for line in lines)
+    rv.append(r"\end{bmatrix}")
```
nwalters512, line 1556: Nit: why not just declare `rv` as a list with three element?

python/prairielearn.py
```diff
@@ -1515,21 +1520,21 @@ def latex_from_2darray(A, presentation_type="f", digits=2):
...
-    if np.isscalar(A):
+    if isinstance(A, numbers.Number):
         if presentation_type == "sigfig":
             return string_from_number_sigfig(A, digits=digits)
         else:
             return "{:.{digits}{presentation_type}}".format(
                 A, digits=digits, presentation_type=presentation_type
             )
-
+    # Using Any annotation here because of weird Pyright-isms.
```
nwalters512, line 1530: Oof. This feels similar to an issue I face sometimes in TypeScript code where I want to declare some variable of a type that's defined in a third-party library. If that library doesn't export the underlying type, you have to redeclare it or use [`Parameters<...>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#parameterstype) to extract the type from the function signature. Unfortunately it looks like the underlying `_FormatDict` type only exists in the `.pyi` file, and Python doesn't appear to have an equivalent of `Parameters` we could use with `np.array2string` 😢

nwalters512 (approved):

python/prairielearn.py
jonatanschroeder: It's not 3 elements, the middle one is a list comprehension.

nwalters512: Indeed - I missed that! Thanks for pointing that out.

python/prairielearn.py
eliotwrobson: Yeah, it has trouble because these formatters are different types but both work for the format dict. So it gets tripped up over this very minor thing.

PR #7522, opened 2023-04-17
Use workspace ranges for all monorepo packages

### LINKED ISSUE #7520: Change package versions to `workspace:^`
As suggested by @jonatanschroeder, we should change the inclusion of our own packages to `workspace:^` [1] to avoid versioning problems like that we recently had in the `@prairielearn/html` package. The timeline of that problem was:
1. Changed `packages/html` in #7508 
2. Bump `@prairielearn/html` to 3.0.0 and publish it to `npm` in #7515
3. `packages.json` still requested 2.0.0 leading to typechecking failures, which we fixed by manually updating to 3.0.0 in #7519 

[1] https://yarnpkg.com/features/workspaces#workspace-ranges-workspace

nwalters512:
Closes #7520. I opted to use `workspace:^` instead of `workspace:*` so that we consistently use the same one everywhere. While we could use `workspace:*` in the root `package.json`, we never want to use that in individual packages since that would be transformed into a pinned version range at publish time (see https://yarnpkg.com/features/workspaces#publishing-workspaces).


mwest1066 (approved):
Makes sense to me!

packages/html-ejs/package.json
```diff
@@ -13,7 +13,7 @@
     "test": "mocha --no-config --require ts-node/register src/index.test.ts"
   },
   "dependencies": {
-    "@prairielearn/html": "^3.0.0",
+    "@prairielearn/html": "workspace:^",
     "ejs": "^3.1.9"
   },
   "devDependencies": {
```
jonatanschroeder, line 19: The dev dependency for tsconfig wasn't updated here.

jonatanschroeder (approved):

jonatanschroeder:
I believe there are a couple of missing dependencies here. `@prairielearn/migrations` and `@prairielearn/html-ejs` are not listed as dependencies in package.json.

packages/html-ejs/package.json
nwalters512: That doesn't matter as much since it's just a dev dependency, but I'll change it anyways for consistency.

PR #7523, opened 2023-04-17
Make warnings and errors non-optional on infofiles

nwalters512:
This is part of a broader effort to enable TypeScript strict null checks. Having these properties be potentially undefined complicates a lot of code + error handling. It's easier to just always have the array be defined but treat an empty array as the "no warnings"/"no errors" case.

### DIFF
--- a/tests/sync/courseDb.test.js
+++ b/tests/sync/courseDb.test.js
@@ -213,8 +213,6 @@ describe('course database', () => {
         await writeQuestion(dir, 'question2', getQuestion());
         await writeQuestion(dir, 'question3', getAlternativeQuestion());
         const result = await courseDb.loadQuestions(dir);
-        assert.isFalse(infofile.hasErrors(result));
-        assert.isFalse(infofile.hasWarnings(result));
         assert.equal(Object.keys(result).length, 3);
         assert.match(
           infofile.stringifyWarnings(result['question1']),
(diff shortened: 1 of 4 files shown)


nwalters512:
Going to merge this so I can proceed with #7525. Comments still welcome; I'll address them in a followup PR!

tests/sync/courseDb.test.js
```diff
@@ -213,8 +213,6 @@ describe('course database', () => {
         await writeQuestion(dir, 'question2', getQuestion());
         await writeQuestion(dir, 'question3', getAlternativeQuestion());
         const result = await courseDb.loadQuestions(dir);
-        assert.isFalse(infofile.hasErrors(result));
-        assert.isFalse(infofile.hasWarnings(result));
```
mwest1066, line 217: Don't we still want zero warnings/errors here?

mwest1066 (approved):

tests/sync/courseDb.test.js
nwalters512: This was just straight up invalid code that type-checking caught once I added required properties to infofiles. `result` isn't an infofile-like thing that we can pass to `hasErrors` or `hasWarnings`. We check for errors/warnings a few lines down on individual objects within `result`.

mwest1066: Ha, nice 😄

PR #7525, opened 2023-04-17
Enable TypeScript strict null checks

nwalters512:
This PR sets `"strictNullChecks": true` in our root `tsconfig.json` file and makes all necessary changes for type checking to pass.

Depends on #7523.

### DIFF
--- a/cron/index.js
+++ b/cron/index.js
@@ -276,7 +276,7 @@ module.exports = {
                 }
 
                 // resolve no matter what so that we run all jobs even if one fails
-                resolve();
+                resolve(null);
               });
               debug(`runJobs(): completed ${job.name}`);
             });
--- a/lib/chunks.js
+++ b/lib/chunks.js
@@ -130,7 +130,7 @@ const sql = sqldb.loadSqlEquiv(__filename);
  * @typedef {Object} DatabaseChunk
  * @property {string | number | null} id
  * @property {ChunkType} type
- * @property {string} [uuid]
+ * @property {string} uuid
  * @property {string | number} course_id
  * @property {string | number} [course_instance_id]
  * @property {string} [course_instance_name]
@@ -173,17 +173,29 @@ module.exports.chunkMetadataFromDatabaseChunk = function (chunk) {
         type: chunk.type,
       };
     case 'clientFilesCourseInstance':
+      if (!chunk.course_instance_name) {
+        throw new Error(`course_instance_name is missing for chunk ${chunk.uuid}`);
+      }
       return {
         type: chunk.type,
         courseInstanceName: chunk.course_instance_name,
       };
     case 'clientFilesAssessment':
+      if (!chunk.course_instance_name) {
+        throw new Error(`course_instance_name is missing for chunk ${chunk.uuid}`);
+      }
+      if (!chunk.assessment_name) {
+        throw new Error(`assessment_name is missing for chunk ${chunk.uuid}`);
+      }
       return {
         type: chunk.type,
         courseInstanceName: chunk.course_instance_name,
         assessmentName: chunk.assessment_name,
       };
     case 'question':
+      if (!chunk.question_name) {
+        throw new Error(`question_name is missing for chunk ${chunk.uuid}`);
+      }
       return {
         type: chunk.type,
         questionName: chunk.question_name,
@@ -865,6 +877,9 @@ const ensureChunk = async (courseId, chunk) => {
       relativeTargetPath = chunk.type;
       break;
     case 'clientFilesCourseInstance':
+      if (!chunk.course_instance_name) {
+        throw new Error(`course_instance_name is missing for chunk ${chunk.uuid}`);
+      }
       relativeTargetPath = path.join(
         'courseInstances',
         chunk.course_instance_name,
@@ -872,6 +887,12 @@ const ensureChunk = async (courseId, chunk) => {
       );
       break;
     case 'clientFilesAssessment':
+      if (!chunk.course_instance_name) {
+        throw new Error(`course_instance_name is missing for chunk ${chunk.uuid}`);
+      }
+      if (!chunk.assessment_name) {
+        throw new Error(`assessment_name is missing for chunk ${chunk.uuid}`);
+      }
       relativeTargetPath = path.join(
         'courseInstances',
         chunk.course_instance_name,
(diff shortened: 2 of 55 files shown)


lib/code-caller/code-caller-container.js
```diff
@@ -348,11 +354,13 @@ class CodeCallerContainer {
   /**
    * Creates a container and attaches its stdin/stdout/stderr to streams
    * we can write to and read from.
+   *
+   * @param {string} hostDirectory
```
nwalters512, line 358: It was easier to make TS happy by just passing in a guaranteed non-null directory vs making redundant checks for `this.hostDirectory.path`.

lib/chunks.js
```diff
@@ -173,17 +173,29 @@ module.exports.chunkMetadataFromDatabaseChunk = function (chunk) {
         type: chunk.type,
       };
     case 'clientFilesCourseInstance':
+      if (!chunk.course_instance_name) {
+        throw new Error(`course_instance_name is missing for chunk ${chunk.uuid}`);
+      }
```
nwalters512, line 178: We could likely avoid all these checks by defining type-specific versions of `DatabaseChunk` (e.g. `QuestionDatabaseChunk`, `ElementsDatabaseChunk`, etc.) and validating with Zod when fetching from the DB. I went with the simpler option for now though.

lib/chunks.js
```diff
@@ -130,7 +130,7 @@ const sql = sqldb.loadSqlEquiv(__filename);
  * @typedef {Object} DatabaseChunk
  * @property {string | number | null} id
  * @property {ChunkType} type
- * @property {string} [uuid]
+ * @property {string} uuid
```
nwalters512, line 133: `uuid` is `NOT NULL` in the schema, so it should never be undefined here.

lib/code-caller/code-caller-native.js
```diff
@@ -212,8 +215,8 @@ class CodeCallerNative {
 
     this.lastCallData = callData;
 
-    this.child.stdin.write(callDataString);
-    this.child.stdin.write('\n');
+    this.child?.stdin?.write(callDataString);
+    this.child?.stdin?.write('\n');
```
nwalters512, line 219: Would it be better to assert that `this.child` is non-null here? `this._checkState([WAITING])` above does this check, but not in a way that TypeScript is aware of 😢

lib/code-caller/index.js
```diff
@@ -234,7 +244,10 @@ module.exports = {
     if (overallErr) {
       throw overallErr;
     } else {
-      return fnResult;
+      // TypeScript doesn't understand our error-handling logic above. If
+      // `overallErr` is falsy, `fnResult` will indeed have type `T`. We'll
+      // cast it to `T` to appease TypeScript.
+      return /** @type {T} */ (fnResult);
```
nwalters512, line 250: I'm once again tempted to reach for something like https://dev.to/karanpratapsingh/go-like-error-handling-in-typescript-4plm 😅

sync/course-db.js
```diff
@@ -680,26 +687,35 @@ module.exports.loadInfoFile = async function ({
...
+
+  if (maybeNullLoadedData && infofile.hasErrors(maybeNullLoadedData)) {
     // We'll only have an error if we couldn't parse JSON data; abort
-    return loadedData;
+    return maybeNullLoadedData;
+  }
+
+  if (!maybeNullLoadedData || !maybeNullLoadedData.data) {
+    throw new Error('Could not load infoCourse.json');
   }
```
nwalters512, line 704: I'm *pretty* sure that we just want to blow up in this case? We didn't previously handle the case where `loadedData` was null.

lib/chunks.js
mwest1066: I've been increasingly using Union types and really enjoying them, but I'm fine with the approach of deferring this for now.

nwalters512: Yeah, I love them too, and make extensive use of them in this file! I'm just trying to KISS a bit for now.

lib/code-caller/code-caller-native.js
mwest1066: I've used assertions for this sort of case and I find them a reasonable pattern when it substantially simplifies the following code. I'm not sure in this case that it will really make much difference either way, so I'm agnostic.

lib/code-caller/index.js
mwest1066: Yuck, but I have no better ideas.

sync/course-db.js
mwest1066: I can't think of any other useful action to take?

lib/code-caller/code-caller-container.js
```diff
@@ -453,7 +461,7 @@ class CodeCallerContainer {
 
   _clearTimeout() {
     this.debug('enter _clearTimeout()');
-    clearTimeout(this.timeoutID);
+    clearTimeout(this.timeoutID ?? undefined);
```
mwest1066, line 464: Could we avoid this change by setting timeoutID to `undefined` rather than `null` when we clear it?

mwest1066 (approved):
I skimmed the more repetitive changes, but things look good. Many thanks for this big effort! 🙏

cron/index.js
```diff
@@ -276,7 +276,7 @@ module.exports = {
                 }
 
                 // resolve no matter what so that we run all jobs even if one fails
-                resolve();
+                resolve(null);
```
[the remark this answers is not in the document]
nwalters512: See [this TypeScript playground](https://www.typescriptlang.org/play?#code/HYUw7gBACgTg9gWwJYGcQAp0xCuAbANxABoJsArEAYwBcBKCAXgD4IBvAKAjJ3yPToBuDgF86QA) for an example. Here's the error message:

Screenshot 2023-04-17 at 17 03 27

The error is a little different in a `.js` file, but it's the same general idea: you need to declare your promise as `Promise<void>` if you want to call `resolve()` with no arguments. In most of these cases, we don't actually care about the return type, so it was easier to just call it with `null` than to smear `/** @type {Promise<void>} */` everywhere.

lib/code-caller/code-caller-container.js
nwalters512: Yep; we could also just `if (this.timeoutID) { ... }`, which might be better? I do think `null` is a better "no value" sentinel than `undefined` in most cases.

cron/index.js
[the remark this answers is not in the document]
mwest1066: Ah cool, makes sense, thanks!

lib/code-caller/code-caller-container.js
mwest1066: I agree that I prefer `null` as a sentinel. I'm fine with the `?? undefined`, I was just checking to make sure I was understanding why we were doing it.

PR #7527, opened 2023-04-18
pl-order-blocks: fix respecting indentation grading settings

### LINKED ISSUE #7516: pl-order-blocks grading does not correctly respect indentation settings
using `pl-order-blocks` to create a question in which the student *may* indent their constructed solution for readability, but indentation *should not* be considered in grading. the docs say that specifying `indentation="true"` for the `pl-order-blocks` element, and `indent="-1"` for each answer, should do this; but when graded (using `ranking`) it still is marked wrong if i deliberately indent some lines in my solution. 
here is a minimal repro `question.html`.  if you drag the blocks in the correct order BUT add indentation to either of them, the grader will say it's incorrect.

```html
<pl-question-panel>Put the blocks in order.
  You may indent the elements for your own readability if you wish, but
  indentation won't affect the correctness of your solution.
</pl-question-panel>

<pl-order-blocks
  format="code"
  answers-name="answers"
  source-blocks-order="random"
  indentation="true"
  grading-method="ranking">

  <pl-answer ranking="10" indent="-1" correct="true">first block </pl-answer>
  <pl-answer ranking="20" indent="-1" correct="true">second block</pl-answer>
</pl-order-blocks>
```

SethPoulsen:
Fix #7516

While fixing the grading bug, I recognized and fixed a few other issues: 
-fixing negative indentation in the "Correct answer box"
-improved messaging in the answer box: it will now say "some" or "all" blocks need correct indentation, rather than just saying "correct indentation needed"

Something we could talk about before this gets merged:
Do we actually like the fact that "-1" is used as the sentinal value to denote the indentation won't be graded for a given line?
The reported bug has existed in the codebase for at least a year, and in that time no one complained until recently, suggesting that no one was using this feature. So if we want to change it to something else we probably could.


elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -603,7 +611,9 @@ def grade(element_html, data):
     if check_indentation:
         indentations = {ans["uuid"]: ans["indent"] for ans in true_answer_list}
         for ans in student_answer:
-            if ans["indent"] != indentations.get(ans["uuid"]):
+            if indentations.get(ans["uuid"]) != -1 and ans[
+                "indent"
+            ] != indentations.get(ans["uuid"]):
```
nwalters512, line 616: Nit, maybe a little more readable?

```suggestion
            indentation = indentations.get(ans["uuid"])
            if indentation != -1 and ans["indent"] != indentation:
```

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -477,14 +477,22 @@ def render(element_html, data):
...
+                indentation_message = ", correct indentation required"
+            elif len(required_indents) > 1:
+                indentation_message = ", some blocks require correct indentation"
 
         question_solution = [
             {
                 "inner_html": solution["inner_html"],
-                "indent": ((solution["indent"] or 0) * TAB_SIZE_PX) + INDENT_OFFSET,
+                "indent": max((solution["indent"] or 0) * TAB_SIZE_PX, 0)
+                + INDENT_OFFSET,
```
nwalters512, line 495: Possibly dumb question: I see `INDENT_OFFSET` is hardcoded to `0`. Can we just remove it entirely?

elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -477,14 +477,22 @@ def render(element_html, data):
...
+            if -1 not in required_indents:
+                indentation_message = ", correct indentation required"
+            elif len(required_indents) > 1:
+                indentation_message = ", some blocks require correct indentation"
 
         question_solution = [
             {
                 "inner_html": solution["inner_html"],
-                "indent": ((solution["indent"] or 0) * TAB_SIZE_PX) + INDENT_OFFSET,
+                "indent": max((solution["indent"] or 0) * TAB_SIZE_PX, 0)
```
nwalters512, line 494: Asking for my own understanding: this was the cause of negative indentation in the answer panel?

nwalters512 (approved):
I once again find myself wishing we had tests for this element!

SethPoulsen:
> I once again find myself wishing we had tests for this element!

We do have a lot! Just not enough apparently...

This sort of subtle bug would only be caught by fuzzing submissions, something that has been a TODO for a few years now.

elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: right

elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: sure, I don't see why not

elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: sure! it looked a lot more readable before it got auto formatted.

eliotwrobson (approved):
Changes seem fine. As in the other comment, little bugs like this (and the large number of feature requests) suggest the need for a more robust way to represent input and test this element.

SethPoulsen:
> suggest the need for a more robust way to represent input and test this element

That's exactly right. The problem with this sort of thing is that it requires a specific question configuration _and_ a specific input to show the bug. 

We can do specific question configurations, that's why I have a whole bunch of different instances of the element with different config options in the test question in `testCourse`. What we can't do well is test different inputs. 

With the current set up, the ways to do this would be:
- hard code some inputs to specific questions in the `test` function. This would look really hacky and could cause problems with users trying to test their questions.
- do some fuzzing (generate random inputs in the `test` function. This would be really cool but could lead to CI tests nondeterministically failing or succeeding.

I'm starting to think it would be better if we had a testing framework where we could explicitly declare which inputs will be tested. I imagine there are other elements that could also benefit from this.

Issue #7526, opened 2023-04-17
Remove the config database table


nwalters512:
As discussed here: https://github.com/PrairieLearn/PrairieLearn/pull/7389#discussion_r1169099000.

There are a few usages we'll have to replace:

- `display_timezone`: this was only useful when PL was a smaller tool operating at a single institution. Now that it's a large multi-institution application, there's really no need for a global timezone. We'll display timestamps in UTC if they don't have an associated institution/course/course instance.
- `workspaceLaunchTemplateId`: We'll move this to the standard config file; this doesn't need to be configured dynamically.
- `workspaceAutoscalingEnabled`/`workspaceDesiredHostCount`: These can be overridden on the ASG itself via the AWS console or Terraform. We'll remove these entirely.

PR #7528, opened 2023-04-18
Remove all references to config table

nwalters512:
Works towards #7526. We can't safely drop the table entirely until these changes have been deployed.


ee/cron/workspaceHostLoads.js
```diff
@@ -169,11 +158,11 @@ async function handleWorkspaceAutoscaling(stats) {
           MaxCount: needed,
           MinCount: 1,
           LaunchTemplate: {
-            LaunchTemplateId: launch_template_id,
+            LaunchTemplateId: config.workspaceLoadLaunchTemplateId,
```
nwalters512, line 161: Note to self: we'll have to add this to our configs before deploying this change.

mwest1066 (approved):

PR #7531, opened 2023-04-19
Delete chunks even for deleted entities

nwalters512:
We had previously been refusing to delete chunks for deleted questions/course instances/assessments. However, this means that once one of those was deleted, we would try again to delete those chunks every time the course was synced. These could build up into hundreds of chunks over time, slowing down syncs and crowding the sync logs.

We'll now delete chunks even for deleted entities. This should be completely safe, since we shouldn't ever try to do anything with deleted things.

### DIFF
--- a/lib/chunks.sql
+++ b/lib/chunks.sql
@@ -176,17 +176,14 @@ WITH
       JSON_ARRAY_ELEMENTS($chunks) AS cm
       LEFT JOIN questions AS q ON (
         q.qid = (cm ->> 'questionName')
-        AND q.deleted_at IS NULL
         AND q.course_id = $course_id
       )
       LEFT JOIN course_instances AS ci ON (
         ci.short_name = (cm ->> 'courseInstanceName')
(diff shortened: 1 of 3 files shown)


lib/chunks.sql
mwest1066, line 1: Can you remember what the original reason was for using `deleted_at IS NULL` here? Was it just a general sense of always filtering on `deleted_at IS NULL` or was there a specific reason? I agree this looks like a safe change, but I have no memory of why we did this originally.

mwest1066 (approved):

lib/chunks.sql
nwalters512: I went spelunking in the Git history and this was introduced in the very first commit with this query (originally a sproc). I don't know why I did this. The only thing I was thinking about was the syncing code: I couldn't remember if maybe we were setting `deleted_at` in weird situations? But AFAICT, we weren't, so I think we're good 🤞

PR #7532, opened 2023-04-19
Optimize chunk deletion query

nwalters512:
For a query that tries to delete 1000 question chunks, this PR reduces the execution time from ~18 seconds to 10 milliseconds.

Postgres was choking on the `chunks.type = cm.type` condition of the `chunks_to_delete` CTE. The execution order it took was to do that join first, and only *then* filter out rows where the type/IDs were mismatched. This resulted in essentially a cross product between the deleted questions and *all* questions, which was millions of rows. By using separate CTEs, we allow Postgres to work much more efficiently.

The only reason we were attempting to delete this many chunks on a regular basis is that we weren't *actually* deleting chunks for deleted questions/assessments/etc, meaning the number of questions we tried to delete each sync just increased over time. This will be resolved by #7531.

### DIFF
--- a/lib/chunks.sql
+++ b/lib/chunks.sql
@@ -190,83 +190,67 @@ WITH
         AND a.course_instance_id = ci.id
       )
   ),
-  -- We use separate CTEs for each chunk type so that Postgres doesn't tye to
+  -- We use separate queries for each chunk type so that Postgres doesn't try to
   -- compute a huge cross product of the chunks + metadata on `type` as an
   -- intermediate step. This was O(n^2) with the number of chunks to delete.
-  question_chunks_to_delete AS (
-    SELECT
-      id
-    FROM
-      chunks
-      JOIN chunks_metadata AS cm ON (
-        cm.type = chunks.type
-        AND cm.question_id = chunks.question_id
-      )
-    WHERE
-      chunks.course_id = $course_id
-      AND chunks.type = 'question'
-  ),
-  client_files_course_instance_chunks_to_delete AS (
-    SELECT
-      id
-    FROM
-      chunks
-      JOIN chunks_metadata AS cm ON (
-        cm.type = chunks.type
-        AND cm.course_instance_id = chunks.course_instance_id
-      )
-    WHERE
-      chunks.course_id = $course_id
-      AND chunks.type = 'clientFilesCourseInstance'
-  ),
-  client_files_assessment_chunks_to_delete AS (
-    SELECT
-      id
-    FROM
-      chunks
-      JOIN chunks_metadata AS cm ON (
-        cm.type = chunks.type
-        AND cm.assessment_id = chunks.assessment_id
-      )
-    WHERE
-      chunks.course_id = $course_id
-      AND chunks.type = 'clientFilesAssessment'
-  ),
-  other_chunks_to_delete AS (
-    SELECT
-      id
-    FROM
-      chunks
-      JOIN chunks_metadata AS cm ON (cm.type = chunks.type)
-    WHERE
-      chunks.course_id = $course_id
-      AND cm.type IN (
-        'elements',
-        'elementExtensions',
-        'clientFilesCourse',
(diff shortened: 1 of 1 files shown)


lib/chunks.sql
```diff
@@ -190,27 +190,83 @@ WITH
...
+    WHERE
+      chunks.type = 'question'
+  ),
+  client_files_course_instance_chunks_to_delete AS (
+    SELECT
+      id
+    FROM
+      chunks
+      JOIN chunks_metadata AS cm ON (
+        chunks.course_id = $course_id
```
jonatanschroeder, line 215: Is there a reason why this is on the join instead of the where?

lib/chunks.sql
```diff
@@ -190,27 +190,83 @@ WITH
...
+        AND chunks.type = cm.type
+        AND cm.course_instance_id = chunks.course_instance_id
```
jonatanschroeder, line 217: Nit:

```suggestion
        AND cm.type = chunks.type
        AND cm.course_instance_id = chunks.course_instance_id
```

lib/chunks.sql
nwalters512: Nope - I'll move it!

nwalters512: Addressed in a442019.

lib/chunks.sql
nwalters512: Addressed in a442019.

lib/chunks.sql
```diff
@@ -190,27 +190,83 @@ WITH
...
+      chunks
+      JOIN chunks_metadata AS cm ON (cm.type = chunks.type)
+    WHERE
+      chunks.course_id = $course_id
+      AND (
+        cm.type = 'elements'
+        OR cm.type = 'elementExtensions'
+        OR cm.type = 'clientFilesCourse'
+        OR cm.type = 'serverFilesCourse'
       )
```
jonatanschroeder, line 248:
```suggestion
      AND cm.type IN ('elements', 'elementExtensions', 'clientFilesCourse', 'serverFilesCourse')
```

`IN` has better use of indices than `OR`.

nwalters512: Changed in 63c5341.

lib/chunks.sql
```diff
@@ -190,27 +190,83 @@ WITH
         AND a.course_instance_id = ci.id
       )
   ),
-  chunks_to_delete AS (
+  -- We use separate CTEs for each chunk type so that Postgres doesn't tye to
+  -- compute a huge cross product of the chunks + metadata on `type` as an
+  -- intermediate step. This was O(n^2) with the number of chunks to delete.
```
mwest1066, line 195: Rather than having separate CTEs and then an extra one to UNION them all together, we could do the UNIONs directly within a single top-level CTE like this: https://github.com/PrairieLearn/PrairieLearn/blob/79081d269ead89e8a013399c0d72119ac62ad295/lib/assessment.sql#L171

This would give us something like:
```
chunks_to_delete AS (
  (
    -- question chunks
    SELECT id
    FROM ...
    WHERE chunks.type = 'question'
  )
  UNION
  (
    -- client_files_course_instance chunks
    SELECT id
    FROM ...
    WHERE chunks.type = 'clientFilesCourseInstance'
  )
  ... etc
)
```

I'm fine with the current approach if you prefer it, however.

mwest1066 (approved):

mwest1066:
> For a query that tries to delete 1000 question chunks, this PR reduces the execution time from ~18 seconds to 10 milliseconds.

Awesome! 🚀

lib/chunks.sql
mwest1066: This bad performance of OR is something I need to remind myself about. I need to keep remembering that excellent blog post that you found, @jonatanschroeder.

lib/chunks.sql
nwalters512: Done in 996c887.

PR #7533, opened 2023-04-19
Fix workspace versions before publish

nwalters512:
Changesets unfortunately doesn't support publishing with Yarn, which means that `workspace:` version ranges aren't replaced prior to publishing:

* https://github.com/changesets/changesets/issues/432
* https://github.com/changesets/action/issues/246

This meant that packages were published to npm with literal `workspace:^` versions, which obviously broke when we tried to install them in another project.

This PR introduces a small script that will rewrite the versions for us.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -8,6 +8,10 @@
     "allowSyntheticDefaultImports": true,
     "skipLibCheck": true,
     "strictNullChecks": true,
+    "module": "NodeNext",
+    // Corresponds to Node 16.x, which we currently run on.
+    // To check compatibility in the future, see https://node.green/.
+    "target": "ES2021",
   },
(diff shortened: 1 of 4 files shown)


tsconfig.json
```diff
@@ -8,6 +8,10 @@
     "allowSyntheticDefaultImports": true,
     "skipLibCheck": true,
     "strictNullChecks": true,
+    "module": "NodeNext",
+    // Corresponds to Node 16.x, which we currently run on.
+    // To check compatibility in the future, see https://node.green/.
+    "target": "ES2021",
```
nwalters512, line 14: These changes were necessary to support top-level await in ESM files.

jonatanschroeder:
Sorry, I'm not familiar enough with changesets to be able to review this with confidence.

PR #7535, opened 2023-04-19
Fix pl-code error message on invalid language

jonatanschroeder:
The original code relied on `get_all_lexers` returning a list of available lexers where all items have at least one alias, but there are lexers that don't have an alias (e.g., JSONBareObject or RawToken), so this was causing an error. Given the list was so extensive that it made no sense to list all elements, this PR replaces the list of lexers with a link to the table of Pygments languages.


eliotwrobson (approved):
Nice catch!

nwalters512 (approved):

PR #7537, opened 2023-04-20
C Grader: Add option to invoke objcopy for student code

jonatanschroeder:
While this was already possible with the separation of compiler and linker, the idea here is to simplify the use of this option.


nwalters512 (approved):
Seems reasonable! This is just a shortcut for `self.run_command("objcopy ...")`, right?

jonatanschroeder:
> Seems reasonable! This is just a shortcut for `self.run_command("objcopy ...")`, right?

Yes, but using `run_command` would require separating compilation and linking (since you'd need to do objcopy after compilation but before linking), with this option this can be done internally.

PR #7542, opened 2023-04-20
feat: `pl-figure` missing file error

### LINKED ISSUE #894: pl-figure element: throw error with missing file
Can we add a test (either to the sync or question load or something instructor visible) to see that images referenced in the pl_figure element actually exist on the disk?

Bonus points if that could check that dynamically generated images actually create a valid image too, but I understand that's a different test-case.

eliotwrobson:
See title, resolves #894. Not sure if it would be cool for `pl-figure` to render images stored server-side, that could be bundled with the changes here if it's a useful thing.


elements/pl-figure/pl-figure.py
```diff
@@ -1,36 +1,74 @@
...
     element = lxml.html.fragment_fromstring(element_html)
     pl.check_attribs(
         element,
         required_attribs=["file-name"],
         optional_attribs=["width", "type", "directory", "inline", "alt"],
     )
 
+    file_type = pl.get_enum_attrib(element, "type", FileType, FILE_TYPE_DEFAULT)
+
+    # Only check for file existence if type is static
```
nwalters512, line 42: Micro-nit:

```suggestion
    # If the file is static, validate that it exists on disk
```

nwalters512 (approved):
Seems reasonable. This has the potential to break questions, but I suppose they'd already be considered broken if they were referencing an image that doesn't exist.

PR #7544, opened 2023-04-20
Introduce @prairielearn/signed-token package

nwalters512:
For easy consumption in PrairieTest.

I was originally going to name this `@prairielearn/csrf`, but we use these things for much more than just CSRF protection, so the more generic `@prairielearn/signed-token` seemed appropriate.


packages/signed-token/src/index.ts
```diff
@@ -0,0 +1,120 @@
+import _ from 'lodash';
+import hmacSha256 from 'crypto-js/hmac-sha256';
```
nwalters512, line 2: Is there a reason we use this and not `createHmac` from the native `node:crypto` module?

lib/authn.js
```diff
@@ -1,6 +1,7 @@
 // @ts-check
 const { config } = require('../lib/config');
-const csrf = require('../lib/csrf');
```
mwest1066, line 3: I don't see a deletion of `lib/csrf` in this PR?

packages/signed-token/src/index.ts
mwest1066: Great question! I can only assume that I missed the fact that node includes HMAC support since v0.1.94 🤦

Can we convert to using node's hmac now? Ideally we'd want them to be bitwise compatible, but they might be if we specify sha256 as the algorithm?

mwest1066 (approved):

packages/signed-token/src/index.ts
nwalters512: they should be, I can write a quick script to verify that they match for a couple inputs.

nwalters512: I think I've convinced myself that they should be the same:

```ts
import hmacSha256 from 'crypto-js/hmac-sha256.js';
import crypto from 'node:crypto';

const KEY = '12345abcde';
const values = ['', 'foo', 'bar', '{"foo":"bar"}'];

for (const value of values) {
  const oldAlg = hmacSha256(value, KEY).toString();
  const newAlg = crypto.createHmac('sha256', KEY).update(value).digest('hex');
  if (oldAlg !== newAlg) {
    console.log(`Error: ${oldAlg} !== ${newAlg}`);
  }
}
```

mwest1066: Nice! It's also a good test of the algorithm implementations 😄

mwest1066 (approved):

PR #7559, opened 2023-04-24
Fix: manual feedback of questions with variants without submission

### LINKED ISSUE #7556: Manual grading not updating feedback in some cases
As reported by @firasm. In some situations it seems the manual grading feedback is not getting updated on the manual grading UI. Upload works as expected. One case:

https://ca.prairielearn.com/pl/course_instance/2575/instructor/assessment/26411/manual_grading/instance_question/53805670

From what I could gather, in these cases the grading job record is also not created. I was unable to reproduce the issue locally, but I'm 99% sure this is related to the test here failing:

https://github.com/PrairieLearn/PrairieLearn/blob/a878642fd27f17230fd2ce27c3d25fc419b75536/sprocs/instance_questions_update_score.sql#L188-L194

I'm not sure which of these is the culprit, since I expect `arg_feedback` to be not null in this case, and there is a submission for this instance question so I'm pretty sure `found_submission_id` should be not null as well. Investigating.

In #6271 this will likely be resolved since the sproc has been revamped into a lib function, but we should review to ensure that whatever is causing this is addressed there.

jonatanschroeder:
Fixes #7556. In essence it ensures that, if there is more than one variant, then a variant with submission is chosen.

### DIFF
--- a/sprocs/instance_questions_update_score.sql
+++ b/sprocs/instance_questions_update_score.sql
@@ -100,7 +100,11 @@ BEGIN
                 AND (g.name = arg_uid_or_group OR u.uid = arg_uid_or_group)
                 AND ai.number = arg_assessment_instance_number AND q.qid = arg_qid)
         )
-    ORDER BY s.date DESC, ai.number DESC;
+    -- If some variants have submissions and others don't, select one that has a
+    -- submission. Variants without submissions will include a NULL s.date. If none of the variants
+    -- have submissions, the process proceeds, but without submission and grading job changes.
+    ORDER BY s.date DESC NULLS LAST, ai.number DESC
+    LIMIT 1;
 
     IF NOT FOUND THEN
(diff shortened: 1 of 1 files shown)


nwalters512 (approved):
Continuing discussion from Slac:

> The issue is that the students clicked on "New variant" but didn't submit on the last variant.
>
> The manual grading assigns the feedback to the latest submission of the latest variant.

Does this mean that the manual grading page was showing an old variant (the one with the submission) instead of the latest, but the code was always assigning to the latest variant/submission regardless? I'm trying to decide if there's anything we should dig deeper into here.

jonatanschroeder:
> Does this mean that the manual grading page was showing an old variant (the one with the submission) instead of the latest, but the code was always assigning to the latest variant/submission regardless? I'm trying to decide if there's anything we should dig deeper into here.

The manual grading interface indeed shows the latest variant that has at least one submission. As for the manual grading sproc, it was selecting the latest submission if it happened to be assigned to the latest variant, or was updating no submission at all if the latest variant (or actually any variant) had no submission. In that case, it would behave in the same way as updating just the score in the assessment instance log.

To be on the safe side, in the rubrics PR, I'm adding the submission ID to the manual grading page, so that if there is a race condition and the student submits a new submission (to the same or a new variant), the manual grading feedback is applied to the submission that was shown to the grader.

PR #7565, opened 2023-04-25
Move code to @prairielearn/sanitize package

nwalters512:
After doing this, I realized that all these functions are actually only used in one file. Perhaps we just inline them in their respective files? 🤷


prairielib/README.md
nwalters512, line 1: PrairieLib is dead!

mwest1066: Yay!

mwest1066 (approved):

mwest1066:
> After doing this, I realized that all these functions are actually only used in one file. Perhaps we just inline them in their respective files? 🤷

Hmmm, I agree that would be very straightforward. But these seem like generically-useful functions that we probably _should_ be using in more places?

PR #7566, opened 2023-04-25
Replace local-lock with async-mutex

nwalters512:
Rather than moving `local-lock` to a package or into the `workspace_host` directory, I opted to just use the `async-mutex` package that we had already installed. This has the added benefit that it fixes a bug where we'd fail to unlock the `LocalLock` if there was an error acquiring a port!


mwest1066 (approved):
Nice! I like the `runExclusive()` helper (just like all the helpers you've been writing recently).

PR #7568, opened 2023-04-25
Typecheck workspace host

nwalters512:
Doing this so I can have more confidence when refactoring the workspace host config.

### DIFF
--- a/cron/workspaceHostTransitions.js
+++ b/cron/workspaceHostTransitions.js
@@ -113,7 +113,13 @@ async function checkHealth() {
       healthy = false;
     } else {
       try {
-        const res = await fetch(url, { signal: AbortSignal.timeout(30_000) });
+        const res = await fetch(url, {
+          // @ts-expect-error: `node-fetch@2` uses a custom type for `AbortController`
+          // which is incompatible with the `AbortController` type defined in TypeScript's
+          // Node types for ES2021. This is fixed in `node-fetch@3`, which we'll be able
+          // to consume once we can use ESM packages.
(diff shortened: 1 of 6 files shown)


cron/workspaceHostTransitions.js
```diff
@@ -113,7 +113,13 @@ async function checkHealth() {
       healthy = false;
     } else {
       try {
-        const res = await fetch(url, { signal: AbortSignal.timeout(30_000) });
+        const res = await fetch(url, {
+          // @ts-expect-error: `node-fetch@2` uses a custom type for `AbortController`
+          // which is incompatible with the `AbortController` type defined in TypeScript's
+          // Node types for ES2021. This is fixed in `node-fetch@3`, which we'll be able
+          // to consume once we can use ESM packages.
```
nwalters512, line 120: 😢 see https://github.com/node-fetch/node-fetch/issues/741.

workspace_host/interface.js
```diff
@@ -461,6 +462,10 @@ async function _allocateContainerPort(workspace) {
   });
 }
 
+/**
+ * @param {object} workspace
+ * @param {function} callback
```
nwalters512, line 467: I didn't define useful types for these just yet; I only needed this so that the promisified versions would have types that reflect the number of parameters.

cron/workspaceHostTransitions.js
mwest1066: 😭

mwest1066 (approved):

PR #7574, opened 2023-04-25
Use typescript-cp instead of copyfiles

nwalters512:
This requires slightly less fiddling with exactly which extensions we want to copy - it'll just copy everything that isn't already being transformed by TypeScript.

I'm once again wishing `tsc --copy-files` was a thing! Alas, https://github.com/microsoft/TypeScript/issues/30835.

### DIFF
--- a/packages/migrations/package.json
+++ b/packages/migrations/package.json
@@ -8,8 +8,8 @@
     "directory": "packages/migrations"
   },
   "scripts": {
-    "build": "tsc && copyfiles -u 1 \"./src/**/*.sql\" dist",
-    "dev": "tsc --watch --preserveWatchOutput",
+    "build": "tsc && tscp",
+    "dev": "tsc --watch --preserveWatchOutput & tscp --watch",
     "test": "mocha --no-config --require ts-node/register src/**/*.test.ts"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


packages/migrations/package.json
```diff
@@ -8,19 +8,19 @@
     "directory": "packages/migrations"
   },
   "scripts": {
-    "build": "tsc && copyfiles -u 1 \"./src/**/*.sql\" dist",
-    "dev": "tsc --watch --preserveWatchOutput",
+    "build": "tsc && tscp",
+    "dev": "tsc --watch --preserveWatchOutput & tscp --watch",
```
mwest1066, line 12: Does the backgrounded task die correctly when we kill `yarn dev`?

mwest1066 (approved):
Good that someone wrote `tscp` at least!

packages/migrations/package.json
nwalters512: It appears to!

ryanvu:
@nwalters512 How did you mimic the copyfiles config of what files to copy with tscp? I noticed no change .tscprc file.

nwalters512:
@ryanvu I believe `tscp` will by default copy all files that weren't already transformed by the TypeScript compiler. See https://github.com/body-builder/typescript-cp/blob/master/src/helpers.ts if you're curious about which files it chooses to ignore.

Please take any further questions to the [`typescript-cp` repository](https://github.com/body-builder/typescript-cp).

PR #7576, opened 2023-04-25
Upgrade typescript and @types/node

nwalters512:
I'm lifting these bumps out of #7572.

### DIFF
--- a/cron/workspaceHostTransitions.js
+++ b/cron/workspaceHostTransitions.js
@@ -113,13 +113,7 @@ async function checkHealth() {
       healthy = false;
     } else {
       try {
-        const res = await fetch(url, {
-          // @ts-expect-error: `node-fetch@2` uses a custom type for `AbortController`
-          // which is incompatible with the `AbortController` type defined in TypeScript's
-          // Node types for ES2021. This is fixed in `node-fetch@3`, which we'll be able
-          // to consume once we can use ESM packages.
-          signal: AbortSignal.timeout(30_000),
-        });
+        const res = await fetch(url, { signal: AbortSignal.timeout(30_000) });
         healthy = res.ok;
       } catch (err) {
         healthy = false;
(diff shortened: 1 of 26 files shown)


cron/workspaceHostTransitions.js
```diff
@@ -113,13 +113,7 @@ async function checkHealth() {
       healthy = false;
     } else {
       try {
-        const res = await fetch(url, {
-          // @ts-expect-error: `node-fetch@2` uses a custom type for `AbortController`
-          // which is incompatible with the `AbortController` type defined in TypeScript's
-          // Node types for ES2021. This is fixed in `node-fetch@3`, which we'll be able
-          // to consume once we can use ESM packages.
```
nwalters512, line 120: 🎉 the latest `@types/nodes` has an `AbortController` that matches the type expected by `node-fetch`!

tests/helperDb.js
```diff
@@ -175,13 +175,7 @@ module.exports.createTemplate = async function createTemplate() {
   });
 };
 
-/**
- * @this {import('mocha').Context}
- */
 module.exports.dropTemplate = async function dropTemplate() {
-  // long timeout because DROP DATABASE might take a long time to error
-  // if other processes have an open connection to that database
-  this.timeout?.(20000);
```
nwalters512, line 184: I pushed this up into the function that actually calls this one; that seems better than having Mocha-specific stuff in here? I had to do that because TypeScript started complaining about `this` being the incorrect type at the callsite of this function.

cron/workspaceHostTransitions.js
mwest1066: Excellent!

tests/helperDb.js
mwest1066: Yep, seems reasonable.

mwest1066 (approved):

tests/helperDb.js
nwalters512: As it turns out, `this.timeout` is straight up not defined in global fixtures! I just removed the timeout entirely.

PR #7578, opened 2023-04-26
Add simplify to big O grading

### LINKED ISSUE #7573: big-oh-input: simplify before taking limits
The `big-oh-input` element is timing out when evaluating expressions that grow exponentially.  There is an active assignment in CS 374, where the correct answer is `factorial(n)`.  Over the last week, several dozen students have experienced timeouts with inputs of the form `n*factorial(n-1)`, leading to the incorrect error message “invalid, not gradeable”.

image

See for example: https://us.prairielearn.com/pl/course/243/question/9071122/preview?variant_id=74394191

I believe the timeout is caused by the way sympy computes limits of quickly growing functions.  Executing `limit(n*factorial(n-1)/factorial(n), n, oo)` takes well over five minutes on my laptop, I assume because sympy is _actually computing factorials_ of large integers.  Cursory examination of the code suggests that sympy is converting limits involving the factorial function (and possibly other combinatorial functions) into limits involving infinite _series_.  I haven’t been able to reproduce the same time-out using simpler exponential functions.  Surprisingly, `limit( sqrt(n)*factorial(n-1/2)/factorial(n), n, oo)` immediately returns the correct answer `1`, presumably because the function values are no longer integers.

Fortunately, sympy is actually pretty good about symbolic simplification!  Executing `limit( simplify (n*factorial(n-1)/factorial(n)), n, oo)` immediately returns the correct answer `1`.

**I recommend inserting a call to `simplify` inside any call to `limit` in the `big-oh-input` code.**  The only reasonable alternative seems to be *never use factorials*, but that would really suck: Factorials show up naturally in running times of backtracking algorithms (for example, TSP).

eliotwrobson:
See title, resolves #7573. Tested locally and seems to work, but hard to tell if this will fix the problem without testing on a live server. Should merge after the semester ends.

### DIFF
--- a/apps/prairielearn/elements/pl-big-o-input/big_o_utils.py
+++ b/apps/prairielearn/elements/pl-big-o-input/big_o_utils.py
@@ -55,7 +55,9 @@ def grade_o_expression(
         ):
             return (0.0, NEGATIVE_FEEDBACK)
 
-        L = sympy.limit(sym_true / sym_sub, sympy.Symbol(variables[0]), sympy.oo)
+        L = sympy.limit(
+            sympy.simplify(sym_true / sym_sub), sympy.Symbol(variables[0]), sympy.oo
+        )
 
         if L < sympy.sympify(0):
             return (0.0, NEGATIVE_FEEDBACK)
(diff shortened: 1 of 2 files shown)


nwalters512 (approved):
Looks reasonable!

> Tested locally and seems to work, but hard to tell if this will fix the problem without testing on a live server.

Isn't it pretty easy to tell if something is timing out or not locally?

eliotwrobson:
> Isn't it pretty easy to tell if something is timing out or not locally?

It can be quite tricky. If something consistently times out, it's pretty easy to tell locally. The problem is that this comes up inconsistently on the production machines, so even if this seems ok for local runs it might be poorly behaved on a live server with other jobs running alongside it.

That being said, I think the fix here should be fine with most functions students will give as answers (see the original issue). But it might still run into issues if someone inputs something really exotic.

PR #7581, opened 2023-04-26
Rename nodemon scripts to dev

nwalters512:
This PR renames `start-nodemon` and `start-workspace-host-nodemon` to `dev` and `dev-workspace-host`, respectively. This is being done ahead of work to process all PrairieLearn source through a build tool. Once that happens, it'll be necessary to differentiate between essentially `node dist/server.js` and `ts-node src/server.ts`.


docker/init.sh
```diff
@@ -7,8 +7,8 @@ make -s start-support
 node server.js --migrate-and-exit >/dev/null
 
 if [[ $NODEMON == "true" ]]; then
```
nwalters512, line 9: I left the `NODEMON` environment variable alone for now. Any thoughts on better ways to handle this?

mwest1066 (approved):

docker/init.sh
mwest1066: What about adding support for a `DEV` environment variable and supporting both of them, but only describing `DEV` in the developer docs?

nwalters512: I like it! I'll make that change.

nwalters512: Done in de70d06 - want to give that a final look?

docs/docker.md
```diff
@@ -84,7 +84,8 @@ To run PrairieLearn with `docker-compose`, run `docker-compose up pl`. This will
 - Build the PL docker image, and tag it as `prairielearn/prairielearn:local`
 - Mount `./testCourse` as a volume for a test course
 - Set up the container to run [external grading jobs](externalGrading.md)
-- Mount the current directory as `/PrairieLearn` and enable `nodemon`, so the container live reloads
+- Mount the current directory as `/PrairieLearn`
+- Configure the server to automatically restart when file change
```
nwalters512, line 88:
```suggestion
- Configure the server to automatically restart when files are modified
```

docs/installingLocal.md
```diff
@@ -22,15 +22,34 @@ docker run -it --rm -p 3000:3000 -w /PrairieLearn -v /path/to/PrairieLearn:/Prai
 # If editing code in `packages/`, you should also repeat either this command or `make build`.
 make deps
 
+# Build and transpile all source files.
+make build
+
```
nwalters512, line 27: Turns out `make deps` already does this!

```suggestion
```

docker/init.sh
mwest1066: Looks great!

PR #7640, opened 2023-04-27
Get apps/prairielearn working and passing all checks

nwalters512:
These changes will be merged into #7582, which itself will then be merged to `master`.


apps/prairielearn/src/server.js
```diff
@@ -389,27 +389,51 @@ module.exports.initExpress = function () {
   // us to treat those files as immutable and cache them essentially forever.
   app.use(
     '/cacheable_node_modules/:cachebuster',
-    express.static(path.join(__dirname, 'node_modules'), {
+    express.static(path.resolve(__dirname, '..', 'node_modules'), {
```
nwalters512, line 392: TODO: dry this up with an abstraction that can handle the need to serve from two different `node_modules` directories.

apps/prairielearn/src/lib/assets.js
```diff
@@ -20,11 +20,11 @@ async function hashDirectory(dir) {
 }
 
 async function computeAssetsHash() {
-  assetsHash = await hashDirectory(path.join(__dirname, '..', 'public'));
+  assetsHash = await hashDirectory(path.join(__dirname, '..', '..', 'public'));
```
nwalters512, line 23: @mwest1066 I'm not planning on ever moving this code again, but I do wonder if it'd be nice to encapsulate this with something like `PACKAGE_ROOT` and `REPOSITORY_ROOT` constants that we can define in a single place? That way we don't have to carefully count the correct number of `..` every time we want to access something at a given path?

apps/prairielearn/src/server.js
```diff
@@ -1847,13 +1871,21 @@ if (config.startServer) {
       async () => {
         logger.verbose('PrairieLearn server start');
 
-        let configFilename = 'config.json';
+        // For backwards compatibility, we'll try loading a `config.json` file
+        // in both the repository root and in the `apps/prairielearn` directory.
+        let configFilenames = [
+          path.resolve(__dirname, '..', 'config.json'),
+          path.resolve(__dirname, '..', '..', '..', 'config.json'),
+        ];
```
nwalters512, line 1879: 👀 do we think this is actually a good idea? I think it's good to avoid breaking people's dev workflows. If we like this, I'll add `apps/prairielearn/config.json` to `.gitignore`.

nwalters512: I should probably do this for graders + workspaces too - especially workspaces, as they historically read from `config.json` in the root of the repo.

apps/prairielearn/src/lib/assets.js
mwest1066: That feels helpful even just from a code readability point of view.

apps/prairielearn/src/server.js
mwest1066: Yeah, I think preserving people's dev workflows is a very polite thing to do.

apps/prairielearn/src/lib/assets.js
nwalters512: This was done in #7665.

Makefile
```diff
@@ -64,8 +60,12 @@ format-python:
 	@python3 -m black ./
 
 typecheck: typecheck-js typecheck-python
+# This is just an alias to our build script, which will perform typechecking
+# as a side-effect.
+# TODO: Do we want to have a separate typecheck command for all packages/apps?
+# Maybe using TypeScript project references?
```
nwalters512, line 66: 👀

apps/prairielearn/src/server.js
nwalters512: This is being done in #7666.

apps/prairielearn/src/server.js
nwalters512: I'm doing this in #7667 - `.gitignore` change to come!

Makefile
mwest1066: I don't know what project references are but I'm in favor of separate commands if it's easy to do. If it's a mess then I don't think it's worth the effort.

mwest1066: Of course it would be good to retain a `make typecheck` or similar that covers everything.

.github/workflows/main.yml
```diff
@@ -76,8 +76,6 @@ jobs:
       run: docker exec test_container /PrairieLearn/docker/lint_html.sh
     - name: Run the Python typechecker
       run: docker exec test_container /PrairieLearn/docker/typecheck_python.sh
-    - name: Run the JavaScript typechecker
-      run: docker exec test_container /PrairieLearn/docker/typecheck_js.sh
```
mwest1066, line 80: Is this no longer needed because the transpilation step implicitly typechecks? I note we are running `yarn typecheck` as an explicit step in the PT CI so I don't quite understand why we don't need it here.

apps/prairielearn/src/tests/database.test.js
```diff
@@ -24,12 +25,14 @@ describe('database', function () {
       outputFormat: 'string',
       coloredOutput: process.stdout.isTTY,
     };
+    const dbDirectory = path.resolve(__dirname, '..', '..', '..', '..', 'database');
```
mwest1066, line 28:
```suggestion
    const dbDirectory = path.join(REPOSITORY_ROOT_PATH, 'database');
```

mwest1066 (approved):

apps/prairielearn/src/server.js
nwalters512: `.gitignore` adjusted in 994455e,

.github/workflows/main.yml
nwalters512: It's necessary in PT because we use Babel to transpile files. Over here, we're using `tsc` to transpile, which implicitly does a type check.

This will get a little hairy once we have TS files in `apps/prairielearn/assets/scripts` - those are built with ESBuild which _doesn't_ perform type-checking. Even then, I think it's better to have type-checking be part of the build. Building/type-checking is not inexpensive in CI, so I'd rather avoid double-processing all files (once for `typecheck`, once for `build`).

apps/prairielearn/src/tests/database.test.js
nwalters512: Done in 27a5837.

PR #7653, opened 2023-04-27
Don't rely on fixed node_modules path for assets

nwalters512:
Once we move the PrairieLearn code to `apps/prairielearn`, we won't be able to _necessarily_ rely on serving `node_modules` assets from exactly one of `node_modules` and `apps/prairielearn/node_modules` - Yarn could in theory place dependencies in either directory (though it will generally host them to the root `node_modules` directory).

This PR uses `require.resolve` to locate the correct version to serve. This is conceptually similar to what we used to do before #6330, except it's robust to the issue that caused us to change this in the first place.

Note that I'll be updating `app.use('/cacheable_node_modules/:cachebuster', ...)` in `server.js` in the PR that actually moves PrairieLearn. We'll need to serve from both `node_modules` directories to handle all cases.

### DIFF
--- a/lib/assets.js
+++ b/lib/assets.js
@@ -6,7 +6,6 @@ const { hashElement } = require('folder-hash');
 
 let assetsHash = 'NOT_SET';
 let elementsHash = 'NOT_SET';
-const cachedPackageVersions = {};
 const cachedPackageVersionHashes = {};
 
 /**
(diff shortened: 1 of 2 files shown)


mwest1066 (approved):
Oh wow, I hadn't even thought about this issue! I'm very glad that you know how to fix it!

PR #7655, opened 2023-04-27
Fix pl-threejs element

nwalters512:
This element has been broken in a few ways for a while now:

- The entire element didn't render because files were moved around in the `three` package.
- Once that was fixed, the motion toggles didn't work, and the motion buttons displayed a missing icon error.

This PR fixes both these issues. It also adds new tests for element assets to prevent future regressions.

### DIFF
--- a/elements/pl-threejs/OrbitControls.js
+++ b/elements/pl-threejs/OrbitControls.js
@@ -1,0 +1,1007 @@
+import { EventDispatcher, MOUSE, Quaternion, Spherical, TOUCH, Vector2, Vector3 } from 'three';
+
+// OrbitControls performs orbiting, dollying (zooming), and panning.
+// Unlike TrackballControls, it maintains the "up" direction object.up (+Y by default).
+//
+//    Orbit - left mouse / touch: one-finger move
+//    Zoom - middle mouse, or mousewheel / touch: two-finger spread or squish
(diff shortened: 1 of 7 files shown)


elements/pl-threejs/info.json
```diff
@@ -3,15 +3,15 @@
     "dependencies": {
         "nodeModulesScripts": [
             "three/build/three.min.js",
```
nwalters512, line 5: We're getting a warning that they'll be removing the pre-built threejs files soon in favor of ESM modules. This makes sense from a moving-the-ecosystem-forward perspective, but is going to cause us some pain 😢

jonatanschroeder:
Is it worth it to handle #5540 here?

nwalters512:
IMO no; I'm trying to do the smallest amount of work possible to proceed with https://github.com/PrairieLearn/PrairieLearn/pull/7582. This PR only aims to return the element to a working state, not make any other changes to it.

elements/pl-threejs/OrbitControls.js
mwest1066, line 1: So nasty to include these in our our source tree, but I agree it's the best bad option at the moment.

mwest1066 (approved):

elements/pl-threejs/OrbitControls.js
nwalters512: Yeah. Once Three forces our hand into consuming ESM, we can consider some little `@prairielearn/three` package that bundles Three + these other files into a single file that we can use in the element.

PR #7665, opened 2023-04-28
Use path constants to construct paths

nwalters512:
This PR introduces a number of constants from which we can construct paths instead of sprinkling `path.resolve(__dirname, ...)` all over the project. This will minimize the number of changes we have to make when we move the PrairieLearn code to the `apps/prairielearn` directory.

### DIFF
--- a/lib/paths.js
+++ b/lib/paths.js
@@ -1,0 +1,17 @@
+// @ts-check
+const path = require('node:path');
+
+const REPOSITORY_ROOT_PATH = path.resolve(__dirname, '..');
+
+const APP_ROOT_PATH = path.resolve(__dirname, '..');
+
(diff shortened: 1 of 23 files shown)


lib/paths.js
```diff
@@ -0,0 +1,17 @@
+// @ts-check
+const path = require('node:path');
+
+const REPOSITORY_ROOT_PATH = path.resolve(__dirname, '..');
+
+const APP_ROOT_PATH = path.resolve(__dirname, '..');
```
nwalters512, line 6: These currently have the same value. When we move PrairieLearn code to `apps/prairielearn`, `APP_ROOT_PATH` will become the absolute path to the `apps/prairielearn` directory. I've kept this future usage in mind when picking between these constants in this PR - bear with the seemingly-arbitrary choices for now!

mwest1066 (approved):
This makes all the path code so much clearer!

PR #7670, opened 2023-04-30
fix: Make `rtol` usage easier in `pl-units-input` element

eliotwrobson:
See title, expands functionality for `rtol` attribute and makes documentation on usage clearer.

@jonatanschroeder changes are per your comments on slack.


jonatanschroeder:
I'm wondering if it may be better to loosen the requirement for tolerance to allow either atol or rtol, instead of just atol, for `with-units`. If I say I'm ok with 0.01 relative tolerance, I don't necessarily care about absolute tolerance. This may be particularly interesting in cases with randomization.

And I'm adding @nwalters512 and @mwest1066 as reviewers in case they're able to do it before me.

eliotwrobson:
The only reason atol is always required in with-units grading is because this quantity needs a unit for grading to work. If no unit is provided or only the default is used, we could have the default behavior just take the unit from the correct answer, however this feels like too much silently happening under the hood for my taste.

Open to suggestions though.

jonatanschroeder:
A value of `rtol` of 0.01 means that the tolerance is 1% of the original value, units and all. So if the answer is 100 seconds, the tolerance is 1 second. I don't see a problem with using the same unit here.

eliotwrobson:
> A value of `rtol` of 0.01 means that the tolerance is 1% of the original value, units and all. So if the answer is 100 seconds, the tolerance is 1 second. I don't see a problem with using the same unit here.

Yes, this is exactly why, in this PR, `rtol` has only one default and doesn't have a unit (stealing the unit used for the correct answer).

The issue I mentioned in my original comment was about `atol`. Maybe if it's not set, we just default it to 0 and then the unit used doesn't matter? I'm still a little wary of changing the default in a non-trivial way depending on the grading mode. I think we should get an opinion from Matt and/or Nate before finalizing the changes here.

jonatanschroeder:
I'm not sure I understand your reservation. Are you talking about with-units? My suggestion was to require, for with-units, either rtol or atol to be present. In this case, if rtol is given and atol is not (in the with-units case), then atol is zero. This case, to the best of my knowledge, was not allowed before, so it would not break any existing question. Is that not what you were thinking?

eliotwrobson:
> My suggestion was to require, for with-units, either rtol or atol to be present. In this case, if rtol is given and atol is not (in the with-units case), then atol is zero. 

Ahhh sorry, I didn't pick up on the second part of this in your original comment. In this case then I'm in agreement, defaulting the atol to 0 if rtol is set shouldn't cause any issues with mismatched units. That logic isn't present in this PR right now but I'm happy to add it.

docs/elements.md
```diff
@@ -748,8 +748,8 @@ Fill in the blank field that allows for **numeric** input and accompanying **uni
...
+| `rtol`                     | number                                       | 1e-2         | Relative tolerance for `comparison="relabs"` and `grading-mode=with-units`.                                                                                                                                                                                                                                                   |
+| `atol`                     | string                                       | 1e-8         | Absolute tolerance for `comparison="relabs"` and `grading-mode=with-units`. In `grading-mode=with-units`, the default value copies the units used in the correct answer, and custom values must include units.                                                                                                                |
```
nwalters512, line 752: Nit: let's be consistent here and elsewhere between quoting and not quoting (I always prefer quoting):

```suggestion
| `rtol`                     | number                                       | 1e-2         | Relative tolerance for `comparison="relabs"` and `grading-mode="with-units"`.                                                                                                                                                                                                                                                   |
| `atol`                     | string                                       | 1e-8         | Absolute tolerance for `comparison="relabs"` and `grading-mode=with-units`. In `grading-mode="with-units"`, the default value copies the units used in the correct answer, and custom values must include units.                                                                                                                |
```

nwalters512 (approved):
Looks reasonable! Review from Matt would be good as well since he was more involved in this discussion than I was.

apps/prairielearn/elements/pl-units-input/units_utils_test.py
```diff
@@ -33,22 +33,34 @@ def test_only_units_grading_fn(
...
-        ("1m", "1.010001 meters", "1cm", False),
-        ("1m", "1 foot", "1cm", False),
-        ("1m", "1 m/s", "1cm", False),
-        ("1m", "1 second", "1cm", False),
-        ("1m", "1 us", "1cm", False),
+        ("1m", "1 meter", 0.0, "1cm", True),
+        ("1m", "1.009 meter", 0.0, "1cm", True),
+        ("1m", "3.28 feet", 0.0, "1cm", True),
+        ("1m", "1.010001 meters", 0.02, "0m", True),
+        ("1m", "1.2 meters", 0.1, "0.1m", True),
```
mwest1066, line 42: Why is this True? Both the `rtol` and `atol` should give us valid submission ranges of 0.9m to 1.1m? Same question for the first kilofoot test below.

eliotwrobson: The rtol and atol are applied together, so atol contributes 0.1 and rtol 0.1 * 1 = 0.1 for a total of 0.2. If you look at the grading function, it just does unit conversions and uses the same comparison function as pl-number-input, so I think this is identical behavior to that element.

mwest1066: Oh! Is this a change in `pl-number-input`? It should be `rtol` OR `atol`, not a sum of them.

eliotwrobson: Oh no, this has always been the pl-number-input behavior. We're just mimicking that here.

mwest1066: Oh right, we are using `np.allclose()` which adds them. No worries, all good.

mwest1066 (approved):

PR #7686, opened 2023-05-01
Fix all source code links in docs

nwalters512:
This wasn't caught in #7582 because the `link-links` target wasn't running in CI. This PR adds that command to our CI workflow and fixes all the broken links


.github/workflows/main.yml
```diff
@@ -74,6 +74,8 @@ jobs:
       run: docker exec test_container /PrairieLearn/docker/lint_js.sh
     - name: Run the HTML linter
       run: docker exec test_container /PrairieLearn/docker/lint_html.sh
+    - name: Run the links linter
+      run: docker exec test_container /PrairieLearn/docker/lint_links.sh
```
nwalters512, line 78: @mwest1066 this fails in CI because we don't mount `graders`/`workspaces`/`images`/`.gitattributes` into the container. Are you opposed to just running this specific step on the host instead of in the container?

mwest1066: I'm fine to run it outside the container. We are generally only running things inside the container to make sure we've got the right version of node and everything else.

PR #7688, opened 2023-05-01
Fix migration execution in Docker init script

nwalters512:
As reported by @jonatanschroeder on Slack: https://prairielearn.slack.com/archives/C0XBTPZMZ/p1682965741437919.


apps/prairielearn/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "build": "tsc && tscp && compiled-assets build ./assets ./public/build",
-    "dev:no-watch": "ts-node --transpile-only src/server.js",
+    "dev:no-watch": "ts-node src/server.js",
```
nwalters512, line 7: We specify `ts-node.transpileOnly` in `tsconfig.json`, so the flag is unnecessary.

docker/init.sh
```diff
@@ -1,14 +1,17 @@
 #!/bin/bash
 
+set -e
```
nwalters512, line 3: This ensures we fail fast if something goes wrong.

mwest1066 (approved):

docker/init.sh
```diff
@@ -1,14 +1,17 @@
...
+
 echo 'Starting PrairieLearn...'
 
 cd /PrairieLearn
 make -s start-support
-node server.js --migrate-and-exit >/dev/null
 
 if [[ $NODEMON == "true" || DEV == "true" ]]; then
+    make dev-migrate > /dev/null
```
jonatanschroeder, line 11:
```suggestion
    make migrate-dev > /dev/null
```

mwest1066: I agree that the syntax of "command verb noun" is good here (like `git clone repo`).

jonatanschroeder (commented):
Is there a need for separate `migrate` and `migrate-dev` tasks? I don't really see a need to allow the migration process to be interrupted by a file change during the initialization process, since it's not intended to be a long-running process.

docker/init.sh
jonatanschroeder: It's a bugfix, not a suggestion. The target here does not match the name in the Makefile.

nwalters512:
@jonatanschroeder `-dev` in this context is a modifier meaning "use `ts-node src/server.js` instead of `node dist/server.js`". I'm probably a bad person to judge if this makes a lot of sense or not; do you have suggestions for improving this terminology?

docker/init.sh
nwalters512: Thanks for pointing this out, will fix.

To Matt's point, maybe the terminology is lacking? `-dev` in this context is more of an adjective than a verb here as I explained below.

An alternative way of thinking about this: we could use a `start-dev` command instead of just `dev`, and then we'd have `start : start-dev :: migrate : migrate-dev`.

mwest1066: `start-dev` and `migrate-dev` feel natural to me.

nwalters512: I can add `start-dev` as an alias for `dev` in another PR if desired?

PR #7692, opened 2023-05-01
Convert ee directory to TypeScript

nwalters512:
These files were already annotated with `// @ts-check`, so the conversions were pretty straightforward! See a few inline comments for more details.


apps/prairielearn/src/ee/auth/azure/callback.ts
```diff
@@ -1,13 +1,12 @@
-// @ts-check
-const passport = require('passport');
-const express = require('express');
-const asyncHandler = require('express-async-handler');
+import * as passport from 'passport';
+import { Router } from 'express';
+import asyncHandler = require('express-async-handler');
```
nwalters512, line 3: This weird little construct is necessary to import modules that look like `module.exports =`, see https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require. We'll be able to just use `import asyncHandler from 'express-async-handler'` when we start using native ESM.

mwest1066: Ok, cool.

apps/prairielearn/src/ee/auth/azure/callback.ts
```diff
@@ -1,13 +1,12 @@
...
 
-const authnLib = require('../../../lib/authn');
+import * as authnLib from '../../../lib/authn';
 
-const router = express.Router();
+const router = Router();
 
-function authenticate(req, res) {
+function authenticate(req, res): Promise<any> {
```
mwest1066, line 9: We can't import `@types/passport` or something to get a better return type here?

apps/prairielearn/src/ee/auth/saml/router.ts
```diff
@@ -29,7 +28,7 @@ router.get('/login', (req, res, next) => {
   })(req, res, next);
 });
 
-function authenticate(req, res) {
+function authenticate(req, res): Promise<any> {
```
mwest1066, line 31: Similarly here for the `any`.

apps/prairielearn/src/ee/institution/admin/saml.ts
```diff
@@ -1,25 +1,22 @@
...
-const pem = require('pem');
+import { Router } from 'express';
+import asyncHandler = require('express-async-handler');
+import * as pem from 'pem';
 
-const error = require('@prairielearn/error');
-const sqldb = require('@prairielearn/postgres');
-const { InstitutionAdminSaml } = require('./saml.html');
-const {
+import * as error from '@prairielearn/error';
```
mwest1066, line 5: Side comment, but this syntax seems to be hard for VSCode intellisense to automatically discover. Whenever I have some code that's doing `error.make()`, it can't figure out how to add the import line for `error` by itself. Could we change `@prairielearn/error` so that we can `import { error } from ...` or whatever VSCode is expecting?

mwest1066 (approved):

apps/prairielearn/src/ee/auth/azure/callback.ts
nwalters512: We already have passport types; passport just has no way of tying the first argument (`'azuread-openidconnect'`) to a particular shape for a user object. The best we could do is to define that type ourselves and annotate this function with it (or maybe validate explicitly with zod).

apps/prairielearn/src/ee/institution/admin/saml.ts
nwalters512: One step ahead of you:

https://github.com/PrairieLearn/PrairieLearn/blob/b56f252e3ade31c49f6a41c4a98e28b82b4f1588/packages/error/src/index.ts#L13-L14

This is more "canonical" that exporting an object with functions on it - just export the functions directly!

I think the _best_ thing to do here is to just transform these from functions to `Error` subclasses so we can just do something like this:

```ts
import { ErrorWithStatus } from '@prairielearn/error';

throw new ErrorWithStatus(404, 'thing was not found');
```

This is much more friendly to intellisense *and* it will show the actual location of the error in the stacktrace (as opposed to having `@prairielearn/lib/dist/index.js` as the first line in the stack trace).

All that said: these are changes for another day and another PR.

PR #7698, opened 2023-05-03
feat: `get_uuid` always starts with an alpha

### LINKED ISSUE #7654: Possible `get_uuid()` changes
Per discussion in office hours with @mwest1066, vanilla uuids generated by this function may not be usable as ids in HTML because they start with numbers, not letters. To avert this problem, the `get_uuid()` function could be modified to only return uuids that start with a letter.

eliotwrobson:
See title. Changes the `get_uuid` function to return a uuid that always starts with a lowercase letter (as starting with a number can cause problems in certain situations). All returned uuids are valid, standard uuids, with a slight loss in randomness. Resolves #7654.


apps/prairielearn/python/prairielearn.py
```diff
@@ -1628,12 +1629,20 @@ def is_correct_scalar_sf(a_sub, a_tru, digits=2):
...
+
+    This is done because certain web components need identifiers to
+    start with letters and not numbers.
     """
-    return str(uuid.uuid4())
+
+    uuid_string = str(uuid.uuid4())
+    random_char = random.choice("abcdef")
+
+    return random_char + uuid_string[1:]
```
jonatanschroeder, line 1645: This loses a bit of randomness (not a big deal, but if you can make it, why not). How about:

```suggestion
    uuid_string = str(uuid.uuid4())

    # replaces 0-f range in first character with a-p range
    return chr(ord('a')+int(uuid_string[0], 16)) + uuid_string[1:]
```

eliotwrobson: I deliberately chose to make the uuid this generated only use valid hex characters in case any client code depends on this property, at the expense of a loss in randomness. I think the loss is inconsequential, and I _really_ want to avoid breaking course code (not that I think this would break anything either way).

nwalters512 (approved):

PR #7699, opened 2023-05-03
Improve formatting of postgres-tools diff

nwalters512:
Before:

Screenshot 2023-05-03 at 12 09 23

After:

Screenshot 2023-05-03 at 12 08 32

The key change is that this shows context instead of just added/removed lines. I could have used the patch format directly, but this is a bit more concise and skips over info that (IMO) is unnecessary, like file names (which may not even be defined for a diff) and line numbers.


mwest1066 (approved):
So much better! And it's less code 🧙

PR #7701, opened 2023-05-03
Convert chunks code to TypeScript

nwalters512:
This was a straightforward conversion, as the file was already checked with `// @ts-check`.


apps/prairielearn/src/lib/chunks.ts
```diff
@@ -685,7 +684,14 @@ module.exports.getChunksDirectoriesForCourseId = (courseId) => {
...
   };
-};
+}
+
+interface CourseWithRuntimeDirectory {
+  /** The database ID of the course. */
+  id: string;
+  /** The path to the course source (not the chunks) */
+  path: string;
+}
```
nwalters512, line 694: Eventually we'll want to pick these off a `Course` model, but this is good enough for now.

apps/prairielearn/src/lib/chunks.ts
```diff
@@ -1072,55 +1051,56 @@ module.exports.ensureChunksForCourseAsync = async (courseId, chunks) => {
...
   );
-};
-module.exports.ensureChunksForCourse = util.callbackify(module.exports.ensureChunksForCourseAsync);
+}
+export const ensureChunksForCourse = util.callbackify(ensureChunksForCourseAsync);
+
+interface QuestionWithTemplateDirectory {
+  id: string;
+  template_directory: null | string;
+}
```
nwalters512, line 1069: Ditto here - a stopgap until we have a `Question` model.

apps/prairielearn/src/lib/chunks.ts
mwest1066: Yep, I like this strategy of "first make it TS at all, then switch to clean DB-based models later".

mwest1066 (approved):

PR #7705, opened 2023-05-04
Remove /MathJax route

nwalters512:
I've reviewed the past 4 months of production traffic and there have been zero requests to this route. I also audited all courses, and the only references to this route are in TAM 251 formula sheets from 4+ years ago. I confirmed with the current instructor that there's no concern with impacting those.


testCourse/courseInstances/Sp15/assessments/exam1-automaticTestSuite/clientFilesAssessment/formulas.html
nwalters512, line 1: This file wasn't even references in the assessment text.

mwest1066 (approved):
Nice!

PR #7706, opened 2023-05-04
Redirect legacy node_modules to cacheable_node_modules

nwalters512:
This maintains backwards compatibility while also allowing us to have clients cache files. There's an extra network round-trip, which will be *ever so slightly* slower, but if that bothers instructors, they can upgrade to the standard `cacheable_node_modules` route via element/question dependencies.


apps/prairielearn/src/server.js
```diff
@@ -367,9 +367,13 @@ module.exports.initExpress = function () {
...
   app.use(express.static(path.join(APP_ROOT_PATH, 'public')));
 
-  // This is included for backwards-compatibility with pages that might still
-  // expect to be able to load files from the `/node_modules` route.
-  app.use('/node_modules', staticNodeModules('.'));
+  // For backwards compatibility, we redirect requests for the old `node_modules`
+  // route to the new `cacheable_node_modules` route.
+  app.use('/node_modules', (req, res) => {
+    // Strip the leading slash.
+    const assetPath = req.url.slice(1);
```
mwest1066, line 374: Just to check, will `req.url` already have the `/node_modules` stripped? We don't want `node_modules` in `assetPath`, I think?

mwest1066 (approved):

apps/prairielearn/src/server.js
nwalters512: Yeah, `req.url` excludes the route part of the path (`/node_modules`, in this case). `req.originalUrl` contains the full one, including `/node_modules`. This is documented here: https://expressjs.com/en/api.html#req.originalUrl

PR #7707, opened 2023-05-04
Improve usage of cacheable assets

nwalters512:
This PR replaces a few hardcoded URLs with ones that use the cacheable paths instead. It also removes an unused `issues.css` stylesheet from a few pages.


mwest1066 (approved):

apps/prairielearn/src/pages/instructorAssessmentManualGrading/instanceQuestion/instanceQuestion.ejs
```diff
@@ -8,10 +8,10 @@
       document.urlPrefix = '<%= urlPrefix %>';
     </script>
     <% if (question.type != 'Freeform') { %>
-    <script src="/javascripts/lodash.min.js"></script>
-    <script src="/javascripts/require.js"></script>
-    <script src="/localscripts/question.js"></script>
-    <script src="/localscripts/question<%= effectiveQuestionType %>.js"></script>
+    <script src="<%= asset_path('javascripts/lodash.min.js') %>"></script>
```
jonatanschroeder, line 11: I know this is already merged, but I only noticed this when merging it into #6271. In there I had already made this change, but I originally used the node_modules version of lodash instead of `/javascripts`. Any preference for node_modules vs javascripts? I believe node_modules may be better maintained in case of updates.

nwalters512: `javascripts` exists purely to support existing legacy code. It should not be used in any new work.

PR #7709, opened 2023-05-04
Serve static assets from a configurable prefix

### LINKED ISSUE #7704: Serve static assets from a prefixed path
To support blue/green, canary, and multi-service deploys, we should start serving assets from prefixed paths that we can scope per environment, service, and version. PrairieLearn will support an `assetPathPrefix` config option, which will determine what asset paths are served to clients in the form of `<script>`/`<style>` tags.

We'll need to account for the following existing assets that are served:

- `/assets/:cachebuster`: serves files from `public` with a hash based on directory contents
- `/`: all files from `public` are available at this path
- `/build`: compiled assets are served from here
- `/cacheable_node_modules/:cachebuster`: files from `node_modules` are available here

There are some other places we serve static assets from that we probably won't be able to use in this new system:

- Legacy asset paths (we don't make any guarantees about asset consistency):
  - `/node_modules`: serves files from `node_modules` without caching (improved in #7706 to redirect to the cacheable route)
  - `/MathJax`: serves MathJax files from `node_modules` without caching (removed in #7705)
  - `/localscripts/calculationQuestion/ace`: serves `ace-builds` files from `node_modules` without caching
  - `/javascripts/ace`: serves `ace-builds` files from `node_modules` without caching
- Any files served from courses: these require access control, so there's no easy way to put them behind a CDN. The good news is that these won't have consistency problems that the rest of our assets do, so we don't need to worry about them respecting `assetPathPrefix`.

nwalters512:
Closes #7704.

This implementation aims to do the right thing by default, while still allowing us to achieve the exact behavior we want in production:

- In local development, or in production environments with the default config, a hash will be computed based on the contents of `yarn.lock`, `apps/prairielearn/public`, and `apps/prairielearn/elements`. This will get added to the default asset prefix to produce a path like `/assets/abcd1234`. Our assets are then served based on that:
  - `/assets/abcd1234/stylesheets/colors.css`
  - `/assets/abcd1234/images/google_logo.svg`
  - `/assets/abcd1234/node_modules/bootstrap/dist/css/bootstrap.min.css`
  - `/assets/abcd1234/elements/pl-file-editor/pl-file-editor.js`
  - `/assets/abcd1234/build/application-ab12cd34.js`
- In our own production environment, we can set `assetsPrefix` to something like `/assets/us-prod/main/1a2b3c4d` and `assetsUseCachebuster` to `false`. Our assets will then be served like the following:
  - `/assets/us-prod/main/1a2b3c4d/stylesheets/colors.css`
  - `/assets/us-prod/main/1a2b3c4d/images/google_logo.svg`
  - `/assets/us-prod/main/1a2b3c4d/node_modules/bootstrap/dist/css/bootstrap.min.css`
  - `/assets/us-prod/main/1a2b3c4d/elements/pl-file-editor/pl-file-editor.js`
  - `/assets/us-prod/main/1a2b3c4d/build/application-ab12cd34.js`

An open question: after implementing this and writing this out, I'm not sure if we actually need the cachebuster to be configurable. I was working under the assumption that we'd want to use the revision as part of the asset prefix in prod. However, I think we can get away with just setting `assetsPrefix` to `/assets/us-prod/main` and relying on the default cachebuster behavior. Does that make sense?


apps/prairielearn/src/lib/assets.js
```diff
@@ -1,70 +1,90 @@
...
-    );
-
-    const pkgJson = JSON.parse(fs.readFileSync(pkgJsonPath, 'utf-8'));
-    return pkgJson.version;
-  }
+async function computeCachebuster() {
+  const hashes = await Promise.all([
+    hashPath(path.join(APP_ROOT_PATH, 'public')),
+    hashPath(path.join(APP_ROOT_PATH, 'elements')),
+    hashPath(path.join(REPOSITORY_ROOT_PATH, 'yarn.lock')),
```
nwalters512, line 31: The assumption here is that the contents of our `node_modules` directories change iff `yarn.lock` changes. I believe that should be a valid assumption in practice.

apps/prairielearn/src/lib/assets.js
```diff
@@ -83,25 +103,7 @@ module.exports.assetPath = (assetPath) => {
...
-  // relatively expensive; cache the hashes for each package.
-  let hash = cachedPackageVersionHashes[moduleName];
-  if (!hash) {
-    const version = getPackageVersion(moduleName);
-    hash = crypto.createHash('sha256').update(version).digest('hex').slice(0, 16);
-    cachedPackageVersionHashes[moduleName] = hash;
-  }
-
-  return `/cacheable_node_modules/${hash}/${assetPath}`;
+  return `${assetsPrefix}/node_modules/${assetPath}`;
```
nwalters512, line 106: Now that we have a "global" cachebuster, I removed the package specific version hash. The only downside to this is that if some package `X` isn't modified by another package is (and thus the cachebuster changes), we'll still get cache misses for `X`; in the old version, the hash was based on the package version, which meant that we'd only get misses when the package was actually updated. I think this is a reasonable tradeoff for now, and will be essentially a non-issue once we're behind a CDN that can cache this stuff.

apps/prairielearn/src/server.js
```diff
@@ -729,6 +699,8 @@ module.exports.initExpress = function () {
   // files.
   // TODO: if we can determine that these routes are no longer receiving
   // traffic in the future, we can delete these.
+  //
+  // TODO: the only internal usage of this is in the `pl-drawing` element. Fix that.
```
nwalters512, line 703: 👀 I think we need to start providing something like `client_files_element_url` to individual elements.

apps/prairielearn/src/server.js
```diff
@@ -2128,14 +2100,7 @@ if (require.main === module && config.startServer) {
...
       },
-      async () => {
-        compiledAssets.init({
-          dev: config.devMode,
-          sourceDirectory: path.resolve(APP_ROOT_PATH, 'assets'),
-          buildDirectory: path.resolve(APP_ROOT_PATH, 'public/build'),
-          publicPath: '/build',
-        });
-      },
+      async () => {},
```
nwalters512, line 2103:
```suggestion
```

apps/prairielearn/src/lib/config.js
```diff
@@ -36,6 +36,13 @@ const ConfigSchema = z.object({
   courseRepoDefaultBranch: z.string().default('master'),
   urlPrefix: z.string().default('/pl'),
   homeUrl: z.string().default('/'),
+  assetsPrefix: z
+    .string()
+    .default('/assets')
+    .refine((s) => s.startsWith('/'), {
+      message: 'must be an absolute path',
+    }),
```
nwalters512, line 44: A nice feature of Zod I hadn't used before: custom validations!

apps/prairielearn/src/lib/assets.js
```diff
@@ -1,70 +1,90 @@
...
  * Should be run at server startup before any responses are served.
+ *
+ * Also initializes the assets compiler.
  */
 module.exports.init = async () => {
-  await Promise.all([computeAssetsHash(), computeElementsHash()]);
+  const cachebuster = await computeCachebuster();
+  assetsPrefix = config.assetsUseCachebuster
+    ? `${config.assetsPrefix}/${cachebuster}`
+    : config.assetsPrefix;
```
nwalters512, line 48: See comment at the end of the PR description about this - maybe we don't need `assetsUseCachebuster` to be configurable?

mwest1066:
> An open question: after implementing this and writing this out, I'm not sure if we actually need the cachebuster to be configurable. I was working under the assumption that we'd want to use the revision as part of the asset prefix in prod. However, I think we can get away with just setting `assetsPrefix` to `/assets/us-prod/main` and relying on the default cachebuster behavior. Does that make sense?

Yes, and I broadly agree that the automatic cachebusting behavior is generally good enough for us in prod.

One slight tweak on the idea would be to also put an extra element in the path that has no meaning to the LB just in case we ever need to manually cache-bust. For example, we could use `/assets/us-prod/main/1/abcd1234` (note the extra `/1/`) where our LB acts on the `/assets/us-prod/main` part, the `abcd1234` is the auto-cachebuster part, and the `1` is a value we can set at deploy time (defaults to `1`) to force an extra cache bust if we have to. This is similar to the "release" tag in [rpm package versioning](https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/), which is used to distinguish otherwise identically-versioned releases. I don't have a specific use-case for this "release" variable right now in the path so it's more of a failsafe if the automatic scheme ever lets us down.

nwalters512:
> One slight tweak on the idea would be to also put an extra element in the path that has no meaning to the LB just in case we ever need to manually cache-bust.

Not a bad idea! Though I think this is something we can always do after the fact if we discover that need in the future. We can hash out those details when I add this config in our private repo.

apps/prairielearn/src/lib/assets.js
mwest1066: Do we also need `build` here? Or are we storing this in `public/build` now? It feels a bit weird to me to put `build` inside `public`?

apps/prairielearn/src/lib/assets.js
mwest1066: Excellent point! I agree that we don't need configurability here.

apps/prairielearn/src/lib/assets.js
mwest1066: That was a nice property to enable fine-grained hashing. Would it be much additional code complexity to handle `node_modules` using the per-package cache-busting and only use the "global cache-busting" for `public` and `elements`? I'm thinking that between dependabot runs we normally don't need to invalidate the `node_modules` caches because it's only our `build` code that's changing most of the time. For example, at the moment we'll be forcing clients to re-download all of mathjax (including fonts) every time we change one of our transpiled client files.

Going further down the fine-grained rabbit hole, I almost wonder whether could cache-bust each element separately?

apps/prairielearn/src/lib/config.js
mwest1066: I love zod!

apps/prairielearn/src/server.js
mwest1066: Makes sense!

mwest1066 (approved):

apps/prairielearn/src/lib/assets.js
nwalters512: We can certainly put `build` somewhere else, but conceptually it's public, so it seemed to me to be a reasonable place to put it. (Note that PT uses `public/build` too.)

FWIW, even if `build` was not located in `public`, we wouldn't need it to factor into the cachebuster, as each file in `build` is guaranteed to have a fingerprint embedded in its name.

apps/prairielearn/src/lib/assets.js
nwalters512: I don't think that's an unreasonable thing to do. I'm envisioning routes like these (using the default `/assets` prefix):

- `/assets/node_modules/CACHEBUSTER/bootstrap/...`
- `/assets/public/CACHEBUSTER/colors.css/...`
- `/assets/elements/pl-file-editor/CACHEBUSTER/pl-file-editor.js`
- `/assets/build/foo-CACHEBUSTER.js`

Of course, `CACHEBUSTER` would be scoped to each respective route "type".

If you like this, I can implement it!

nwalters512: I implemented the fine-grained approach in 1815fd7. It includes hashes for individual files within `elements` and `public`, and also includes package-version-based hashes for things within `node_modules`.

mwest1066: Awesome!

mwest1066 (approved):

Issue #7710, opened 2023-05-05
pl-drawing.py fails lint in container


trombonekenny:
Recently pulled master branch with a recently pulled prairielearn/prairielearn docker image fails the lint check:

```
bash-4.2# make lint-python
./apps/prairielearn/elements/pl-drawing/pl-drawing.py:7:1: I001 isort found an import in the wrong position
./apps/prairielearn/elements/pl-drawing/pl-drawing.py:11:1: I003 isort expected 1 blank line in imports, found 0
make: *** [Makefile:53: lint-python] Error 1
bash-4.2# 
```

Running `make format` fixes it with the following (benign) diff:

```
diff --git a/apps/prairielearn/elements/pl-drawing/pl-drawing.py b/apps/prairielearn/elements/pl-drawing/pl-drawing.py
index 26f89aa1d..21c73c59a 100644
--- a/apps/prairielearn/elements/pl-drawing/pl-drawing.py
+++ b/apps/prairielearn/elements/pl-drawing/pl-drawing.py
@@ -4,11 +4,12 @@ import warnings
 
 import chevron
 import defaults
-import elements
 import lxml.etree
 import lxml.html
 import prairielearn as pl
 
+import elements
+
 
 def union_drawing_items(e1, e2):
     # Union two sets of drawing items, prioritizing e2 in cases of duplicates.
```

I'm not sure why we're not seeing this failure in CI or more places. Is it something specific to my branch or environment?

PR #7711, opened 2023-05-05
Remove nonexistent paths from .dockerignore

nwalters512:


eliotwrobson:
Will this resolve #7710?

PR #7717, opened 2023-05-07
Make chai-as-promised available in all tests

nwalters512:
This removes the `chai-as-promised` boilerplate from all tests in favor of declaring it once in a Mocha root hook.


apps/prairielearn/src/tests/mocha-hooks.mjs
mwest1066, line 1: Just to make sure I'm understanding how hooks work, does this file still get loaded even when we run a single test by hand with something like `yarn mocha path/to/test.js`?

mwest1066 (approved):

apps/prairielearn/src/tests/mocha-hooks.mjs
nwalters512: Yes, because it's loaded explicitly here: https://github.com/PrairieLearn/PrairieLearn/blob/55f3fd30e9c110f5ba5a67f67553ad7abe47a1b9/apps/prairielearn/.mocharc.js#L2

PR #7721, opened 2023-05-09
Handle stylesheets in compiled-assets package

nwalters512:
This PR adds support for bundling stylesheets in `assets/stylesheets` with esbuild. It also switches to using esbuild's [serve API](https://esbuild.github.io/api/#serve) in dev mode, which a) avoids unnecessary recompilation and b) correctly handles font assets that are included in some stylesheets.


apps/prairielearn/src/lib/assets.ts
```diff
@@ -138,6 +138,13 @@ export async function init() {
   });
 }
 
+/**
+ * Shuts down the development assets compiler if it is running.
+ */
+export async function close() {
+  await compiledAssets.close();
```
jonatanschroeder, line 145:
```suggestion
  return compiledAssets.close();
```

No point in awaiting here instead of just returning the promise.

apps/prairielearn/assets/stylesheets/issues.css
nwalters512, line 1: I originally built an `application.css` bundle that included Bootstrap and friends and used that on every page via `partials/head.ejs`. However, I figure it'd be safer to start with the issues page so we can verify that everything is working correctly in production first; this is a low-risk page that will gracefully degrade if this stylesheet can't load.

apps/prairielearn/src/lib/assets.ts
nwalters512: I generally agree in performance-sensitive code (no point waiting for two promise resolutions), but this will only ever be run during termination of the dev server during tests, and I think it's clearer to have the explicit await here.

mwest1066 (approved):

PR #7728, opened 2023-05-11
Fix ts-node config for dev mode

nwalters512:
I accidentally broke `make dev`/`yarn dev` in #7723. This should get everything back into a good working state. I tested the following commands and they all work!

- `make dev` from repo root
- `yarn dev` from app root
- `yarn test` from app root
- `yarn mocha src/tests/assets.test.js` from app root

### DIFF
--- a/apps/prairielearn/src/tests/mocha-env.mjs
+++ b/apps/prairielearn/src/tests/mocha-env.mjs
@@ -3,9 +3,6 @@ import { fileURLToPath } from 'url';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
-// Don't perform typechecking, just transpile to keep tests fast.
-process.env.TS_NODE_TRANSPILE_ONLY = 'true';
-
 // Always use the tsconfig.json file in the `src` directory, not the one in the
-// application root. `ts-node` doesn't seem to like our usage of project references.
+// application root. `ts-node` doesn't seem to understand our usage of project references.
 process.env.TS_NODE_PROJECT = path.resolve(__dirname, '..', 'tsconfig.json');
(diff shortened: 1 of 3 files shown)


apps/prairielearn/src/tests/mocha-env.mjs
```diff
@@ -3,9 +3,6 @@ import { fileURLToPath } from 'url';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
-// Don't perform typechecking, just transpile to keep tests fast.
-process.env.TS_NODE_TRANSPILE_ONLY = 'true';
-
 // Always use the tsconfig.json file in the `src` directory, not the one in the
-// application root. `ts-node` doesn't seem to like our usage of project references.
+// application root. `ts-node` doesn't seem to understand our usage of project references.
```
nwalters512, line 7: This is the root issue here. I found this issue that I think corresponds to what we're facing? https://github.com/TypeStrong/ts-node/issues/1057

mwest1066: I *think* I understand what's going on here, but jeez, we are deep in the weeds 😅

mwest1066 (approved):

apps/prairielearn/src/tests/mocha-env.mjs
nwalters512: Yeah, it's super gross. LMK if you have questions, I'm happy to share context! I'm equally fine with us just moving on and acknowledging that we'll probably never touch this again.

mwest1066: Yeah, I'm happy to know that you know 😅

Issue #7727, opened 2023-05-10
Migrate to v3 of the AWS SDK


nwalters512:
v3 of the AWS SDK ships pser-service packages (e.g. `@aws-sdk/client-s3`) instead of a single monolithic `aws-sdk` package. This will significantly reduce our build times, as TypeScript will no longer have to load and parse types for the hundreds of services that we don't use.

For example, running `tsc --diagnostics` in `packages/docker-utils` with and without `aws-sdk` showed a reduction in memory from ~737MB to just 88MB, respectively.

PR #7729, opened 2023-05-11
Use v3 of AWS SDK in docker-utils package

nwalters512:
Part of #7727.

### DIFF
--- a/packages/docker-utils/package.json
+++ b/packages/docker-utils/package.json
@@ -24,7 +24,6 @@
   "dependencies": {
     "@aws-sdk/client-ecr": "^3.329.0",
     "@prairielearn/logger": "workspace:^",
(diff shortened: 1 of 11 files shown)


mwest1066 (approved):

packages/docker-utils/package.json
```diff
@@ -22,6 +22,7 @@
     "typescript": "^5.0.4"
   },
   "dependencies": {
+    "@aws-sdk/client-ecr": "^3.329.0",
     "@prairielearn/logger": "workspace:^",
     "aws-sdk": "^2.1369.0",
```
jonatanschroeder, line 27: Is this still necessary?

nwalters512: No! Good catch, I'll remove it.

PR #7732, opened 2023-05-11
Use v3 AWS SDK to interact with SQS

nwalters512:
I tested this end-to-end in our staging environment, and external grading is still working well!

Part of #7727.

### DIFF
--- a/apps/prairielearn/src/cron/sendExternalGraderDeadLetters.js
+++ b/apps/prairielearn/src/cron/sendExternalGraderDeadLetters.js
@@ -24,7 +24,11 @@ module.exports.run = (callback) => {
     return callback(null);
   }
 
-  const sqs = new SQSClient(config.awsServiceGlobalOptions);
+  const sqs = new SQSClient({
+    region: config.awsRegion,
+    ...config.awsServiceGlobalOptions,
+  });
+
   let msg;
   async.series(
     [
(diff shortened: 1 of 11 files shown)


apps/prairielearn/src/cron/sendExternalGraderDeadLetters.js
```diff
@@ -1,49 +1,43 @@
...
-  const sqs = new AWS.SQS(config.awsServiceGlobalOptions);
+  const jobsDeadLetterQueueName = config.externalGradingJobsDeadLetterQueueName;
+  const resultsDeadLetterQueueName = config.externalGradingResultsDeadLetterQueueName;
+  if (!config.externalGradingUseAws || !jobsDeadLetterQueueName || !resultsDeadLetterQueueName) {
+    return callback(null);
+  }
+
+  const sqs = new SQSClient({
+    region: config.awsRegion,
+    ...config.awsServiceGlobalOptions,
```
nwalters512, line 29: AFAICT this was added for [localstack](https://localstack.cloud/) compatibility? Do any of us use that?

mwest1066: Not any more. At some point we were trying it for local S3 mocking, I think?

mwest1066 (approved):

apps/prairielearn/src/cron/sendExternalGraderDeadLetters.js
nwalters512: Yeah, I guess [s3rver](https://github.com/jamhall/s3rver) is what we ultimately landed on?

PR #7738, opened 2023-05-12
Modernize authLogin page

nwalters512:
I'm doing this ahead of work on this page in #6151 so that the modernization can be reviewed in isolation.


apps/prairielearn/src/pages/authLogin/authLogin.ts
nwalters512, line 1: Git seems to have lost track of the fact that this was technically a rename :/ is it worth splitting this PR into pieces so that Git will track this file through history?

apps/prairielearn/src/pages/authLogin/authLogin.ts
```diff
@@ -0,0 +1,72 @@
...
+  '/',
+  asyncHandler(async (req, res, _next) => {
+    let institutionAuthnProviders: InstitutionAuthnProvider[] | null = null;
+
+    if (isEnterprise()) {
+      const institutionAuthnProvidersRes = await queryValidatedRows(
+        sql.select_institution_authn_providers,
+        {},
+        InstitutionAuthnProviderSchema
+      );
```
nwalters512, line 30: A thought I've had: we semi-frequently make queries without any parameters. Would it be worth it to teach this function how to work without the `params` argument? That is, in this case, we'd just have the following:

```suggestion
      const institutionAuthnProvidersRes = await queryValidatedRows(
        sql.select_institution_authn_providers,
        InstitutionAuthnProviderSchema
      );
```

mwest1066: I've had the same thought but wasn't sure whether there was a nice way to do this with TS. Would we need to re-order the arguments so the optional one (`params`) comes last?

nwalters512: Actually, we would not! With [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) TypeScript can be aware of the difference between the two-argument and three-argument behavior. Then we just check `arguments.length` or the types of the arguments to actually implement that.

apps/prairielearn/src/pages/authLogin/authLogin.ts
mwest1066: That's annoying. I worry that trying to split it up is going to be a mess, however. Do you think you can do it easily and cleanly? It's not worth going to some extreme effort just to preserve history. When using `git blame` manually we can always use `-M` and `-C` (maybe with `-w`) to better track the history of lines, even if GitHub doesn't do it.

mwest1066 (approved):

apps/prairielearn/src/pages/authLogin/authLogin.ts
mwest1066: Oh nice! I didn't think about TS overloads in this context (I haven't actually used them for anything yet). In that case I'm 100% on board. I just added https://github.com/PrairieLearn/PrairieLearn/issues/7739 to summarize our earlier discussion on modernizing the query functions -- that feels like a natural place to make this change as well?

apps/prairielearn/src/pages/authLogin/authLogin.ts
nwalters512: I think if I split up "convert to TypeScript" and "use HTML template literal" into 2 PRs, Git might be able to track the changes better?

nwalters512: https://github.com/PrairieLearn/PrairieLearn/pull/7740 - looks like that did the trick without any difficult work.

apps/prairielearn/src/pages/authLogin/authLogin.ts
jonatanschroeder: Would function overloads also be able to address #7739 in the sense of allowing a `Model` parameter to be optionally used, instead of a separate function for validated rows?

PR #7741, opened 2023-05-15
Introduce improved Postgres query functions

### LINKED ISSUE #7739: Clean up typed-checking `query*` functions
As discussed on Slack with @nwalters512 we could clean up our queries with some new functions:
* `queryRows()` -- the equivalent of the current `queryValidatedRows()`
* `queryOneRow()` -- the equivalent of the current `queryValidatedOneRow()`
* `queryOptionalRow()` -- the equivalent of the current `queryValidatedZeroOrOneRow()`

Each of these functions should support two new features:
* Optional `params` argument. All functions should work with three arguments as `query*(sql, params, type)` and two arguments as `query*(sql, type)`.
* Auto-detect single-column mode. At the moment we have separate `*SingleColumn*` versions of all functions, but this is very error-prone in practice. Instead we should check whether exactly one column was returned and treat this the same way the `SingleColumn` variants do at the moment. That is, apply the type-checking directly to the object in that column and return the column value (or a list of them) rather than the row or list of rows.

nwalters512:
Closes #7739.


packages/postgres/src/default-pool.ts
```diff
@@ -62,6 +62,9 @@ export const queryValidatedSingleColumnZeroOrOneRow =
 export const callValidatedRows = defaultPool.callValidatedRows.bind(defaultPool);
 export const callValidatedOneRow = defaultPool.callValidatedOneRow.bind(defaultPool);
 export const callValidatedZeroOrOneRow = defaultPool.callValidatedZeroOrOneRow.bind(defaultPool);
+export const queryRows = defaultPool.queryRows.bind(defaultPool);
+export const queryRow = defaultPool.queryRow.bind(defaultPool);
```
nwalters512, line 66: As mentioned on #7739, we can't use `queryOneRow` like we originally planned on since that's already allocated to legacy code. I went with this for now, but `querySingleRow` is another option if we want to emphasize the "singleness" of the result.

mwest1066: I'm really torn about this! I normally dislike having almost identical function names, but in this case I think that type hinting in the editor (and type checking in CI) will immediately make it clear if the wrong function is being used, so I'm leaning towards `queryRow`. I'd be happy with `querySingleRow` as well if you decide you prefer that.

mwest1066 (approved):

PR #7743, opened 2023-05-15
Fix instructor file editor client code

nwalters512:
When testing this after deploying to our staging environment, I discovered that our new compilation process (reasonably) doesn't play well with the pattern used here: serving a client-side script located in the `pages` directory. Now that we have support for client-side code bundles, I can use that instead.

I deliberately did not do any work to convert this to TypeScript or actually include ace or `crypto-js` in the bundle; I'm doing the least amount of work possible now to unblock a production deployment.

### DIFF
--- a/apps/prairielearn/src/pages/instructorFileEditor/instructorFileEditor.ejs
+++ b/apps/prairielearn/src/pages/instructorFileEditor/instructorFileEditor.ejs
@@ -5,8 +5,7 @@
     <%- include('../partials/head', {pageTitle: 'Edit ' + fileEdit.fileName}); %>
     <script src="<%= node_modules_asset_path('ace-builds/src-min-noconflict/ace.js') %>"></script>
     <script src="<%= node_modules_asset_path('crypto-js/crypto-js.js') %>"></script>
-    <script src="./file_edit?serve=client"></script>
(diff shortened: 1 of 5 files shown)


apps/prairielearn/src/pages/instructorFileEditor/instructorFileEditor.ejs
```diff
@@ -5,8 +5,7 @@
     <%- include('../partials/head', {pageTitle: 'Edit ' + fileEdit.fileName}); %>
     <script src="<%= node_modules_asset_path('ace-builds/src-min-noconflict/ace.js') %>"></script>
     <script src="<%= node_modules_asset_path('crypto-js/crypto-js.js') %>"></script>
-    <script src="./file_edit?serve=client"></script>
-    <script src="<%= node_modules_asset_path('socket.io-client/dist/socket.io.min.js') %>"></script>
```
nwalters512, line 9: AFAICT this was unused on the page and probably just blindly copy/pasted from another page, so I removed this.

mwest1066 (approved):

mwest1066:
> I deliberately did not do any work to convert this to TypeScript or actually include ace or `crypto-js` in the bundle; I'm doing the least amount of work possible now to unblock a production deployment.

Very reasonable!

PR #7745, opened 2023-05-15
Report cron errors to Sentry

nwalters512:
This will give us additional visibility into when cron jobs aren't behaving as expected.


mwest1066 (approved):
Excellent!

PR #7747, opened 2023-05-15
Use AWS SDK v3 in grader-host

nwalters512:
Part of #7727.


apps/grader-host/package.json
```diff
@@ -19,13 +21,12 @@
...
     "async": "^3.2.4",
     "async-stacktrace": "^0.0.2",
-    "aws-sdk": "^2.1369.0",
     "byline": "^5.0.0",
     "chai": "^4.3.7",
     "dockerode": "^3.3.5",
     "fs-extra": "^11.1.1",
     "logform": "^2.5.1",
-    "s3-streamlogger": "1.7.0",
+    "s3-streamlogger": "^1.9.0",
```
nwalters512, line 29: We previously pinned this to an earlier version because the internal v3 upgrade broke some things for us. However, now that we *want* the v3 SDK, I unpinned this and updated our usage to explicitly pass in a `region`.

apps/grader-host/src/lib/jobLogger.js
```diff
@@ -9,6 +9,9 @@ module.exports = function (options) {
 
   const s3StreamLoggerTransport = new winston.transports.Stream({
     stream: new S3StreamLogger({
+      config: {
+        region: config.awsRegion,
+      },
```
nwalters512, line 14: This is necessary because we no longer globally set the region.

mwest1066 (approved):

PR #7748, opened 2023-05-15
Use AWS SDK v3 in workspace-host

nwalters512:
Part of #7727.


apps/workspace-host/src/lib/aws.ts
nwalters512, line 1: I copied all the code in this file from `lib/aws.js` in `prairielearn`.

mwest1066 (approved):

PR #7749, opened 2023-05-15
Upgrade Python grader deps

nwalters512:
I'm deliberately not bumping Pandas to v2 just yet. I want to put *slightly* more effort into communicating that change.


eliotwrobson (approved):
Seems fine, no major version bumps.

PR #7750, opened 2023-05-15
Upgrade plbase Python deps

nwalters512:
I'm not upgrading to pandas 2 or sphinx 7 just yet.


eliotwrobson (approved):
Looks good, no major version bumps and type stubs are in sync with package versions

PR #7756, opened 2023-05-16
Use new server jobs in chunks lib

nwalters512:
This also includes a small refactor of the new server jobs code. See inline comment for more details.

### DIFF
--- a/apps/prairielearn/src/lib/chunks.ts
+++ b/apps/prairielearn/src/lib/chunks.ts
@@ -9,14 +9,12 @@ import tar = require('tar');
 import util = require('util');
 import { v4 as uuidv4 } from 'uuid';
 
-import { logger } from '@prairielearn/logger';
 import namedLocks = require('@prairielearn/named-locks');
 import sqldb = require('@prairielearn/postgres');
 
 import aws = require('./aws');
 import { chalk, chalkDim } from './chalk';
-import serverJobs = require('./server-jobs-legacy');
-import type { Job } from './server-jobs-legacy';
+import { createServerJob, ExecutionJob } from './server-jobs';
 import courseDB = require('../sync/course-db');
 import type { CourseData } from '../sync/course-db';
 import { config } from './config';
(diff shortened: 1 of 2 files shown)


apps/prairielearn/src/lib/server-jobs.ts
```diff
@@ -27,10 +27,23 @@ interface ServerJobExecOptions {
...
+  info(msg: string): void;
+  verbose(msg: string): void;
+  exec(file: string, args?: string[], options?: ServerJobExecOptions): Promise<void>;
+}
+
+export interface ServerJobExecutor {
+  jobSequenceId: string;
+  execute(fn: ServerJobExecutionFunction): Promise<void>;
+  executeInBackground(fn: ServerJobExecutionFunction): void;
+}
```
nwalters512, line 42: Defining these two interfaces and switching the class name to `ServerJobImpl` is an attempt to make sure that we're careful with how we use these properties. Specifically, we want to avoid calling `exec()`/`error()`/etc. on the thing returned from `createServerJob()` and we want to avoid calling `execute()` or `executeInBackground()` inside an already-executing job. These separate interfaces allow us to achieve that.

I don't _love_ the `*Impl` name 🤷 but the alternative is to name the interface `IServerJob` to avoid a name conflict, which I don't love either!

apps/prairielearn/src/lib/chunks.ts
```diff
@@ -754,73 +752,33 @@ export async function updateChunksForCourse({
...
-        job.succeed();
-      } catch (err) {
-        job.error(chalk.red(JSON.stringify(err)));
-        await job.failAsync(err);
-        throw err;
-      }
+  serverJob.executeInBackground(async (job) => {
+    for (const [i, courseId] of course_ids.entries()) {
+      job.info(`Generating chunks for course ${courseId} [${i + 1}/${course_ids.length}]`);
+      await _generateAllChunksForCourseWithJob(courseId, job);
```
nwalters512, line 765: Still loving how much more concise new-style job code is!

mwest1066: Sooooooo much better!

apps/prairielearn/src/lib/chunks.ts
```diff
@@ -1081,7 +1039,7 @@ export const getTemplateQuestionIds = util.callbackify(getTemplateQuestionIdsAsy
  */
 export function logChunkChangesToJob(
   { updatedChunks, deletedChunks }: ChunksDiff,
-  job: Pick<Job, 'verbose'>
+  job: Pick<ServerJob, 'verbose'>
```
mwest1066, line 1042: I don't strongly object to this, but I feel that it's adding complexity for an only minimal improvement in safety (assuming the purpose is to stop us calling functions other than `verbose()`).

mwest1066 (approved):

apps/prairielearn/src/lib/chunks.ts
nwalters512: The reason this looks like it does is that this function is being called with both legacy and modern jobs. While the types of both types of jobs are inherently incompatible, both job types *do* have a `verbose` function with an identical signature, and that's the only function that is used here. This type allows us to continue calling this function with type-safety until we completely eliminate all legacy job callers.

apps/prairielearn/src/lib/server-jobs.ts
nwalters512: Yeah, `ServerJobImpl` is deliberately not exported, so this is strictly an implementation detail.

apps/prairielearn/src/lib/chunks.ts
mwest1066: Ah! I completely misunderstood the purpose! Thanks for the explanation.

PR #7758, opened 2023-05-16
Use new server jobs for score uploading

nwalters512:


apps/prairielearn/src/lib/score-upload.js
```diff
@@ -1,316 +1,249 @@
...
+            json = _.mapKeys(json, (v, k) => {
+              return k.toLowerCase();
+            });
+            const msg = `Processing CSV record ${number}: ${JSON.stringify(json)}`;
+            if (output == null) {
+              output = msg;
+            } else {
+              output += '\n' + msg;
+            }
+            module.exports._updateInstanceQuestionFromJson(
```
nwalters512, line 88: I opted to not convert all this to async/await just yet, I'm only focusing on server jobs in this PR.

mwest1066 (approved):

apps/prairielearn/src/lib/score-upload.js
jonatanschroeder: This is actually done in #6271...

PR #7766, opened 2023-05-18
Fix v2 question execution in production

nwalters512:
Because `question-servers/default-calculation` was nested inside `src/`, it was being transformed by TypeScript, which ended up breaking things. By moving all v2 question servers to a separate directory, we avoid this.


apps/prairielearn/src/lib/file-paths.js
```diff
@@ -4,13 +4,10 @@ const path = require('path');
...
+const { APP_ROOT_PATH } = require('./paths');
+
 const sql = sqldb.loadSqlEquiv(__filename);
-const QUESTION_DEFAULTS_PATH = path.resolve(
-  __dirname,
-  '..',
-  'question-servers',
-  'default-calculation'
-);
+const QUESTION_DEFAULTS_PATH = path.resolve(APP_ROOT_PATH, 'v2-question-servers');
```
nwalters512, line 10: Bikeshed away on the directory name 🙃

mwest1066 (approved):

PR #7767, opened 2023-05-18
Set timezone in courses and instances to NOT NULL

### LINKED ISSUE #7157: feat: set an institution default timezone
We track a `pl_courses.display_timezone` configuration per course, but I realized there's no institutional default.

If we had a `institutions.display_timezone` we could have PL use that reference when it's creating a new course (one time copy), or make it a fall-through default if a `display_timezone` isn't set at the course level. (Similar to what we do for `course_instances.display_timezone` with COALESCE).

### LINKED ISSUE #7344: Ensure consistency in handling of timezone columns

jonatanschroeder:
Resolves #7344. Resolves #7157. Does not set the default timezone to UTC yet, as that can break existing functionality.

- [x] Database migration and description
- [x] ~~Client-side validation (require timezone)~~ there is no validation in those forms, and timezone has a default, so skipping
- [x] Default course timezone on creation to institution timezone
- [ ] Sync warning if course does not set a timezone
- [x] Course repo creation sets timezone in infoCourse.json
- [x] Replace unnecessary COALESCE cases with direct access to course or instance timezone


jonatanschroeder:
> * Sync warning if course does not set a timezone

@mwest1066 @nwalters512 this was proposed, but before I add it: this will cause most courses that use a default timezone before this change to start showing warnings. The impact of not adding this to the courses, though, is that any local development, where the courses are assigned to the "Default" institution, will cause these courses to switch to UTC locally when the default institution is also switched to UTC. So do we want to do the intrusive change of warning on sync if a course doesn't have it, or have the (easily ignorable) news/announcement for a timezone to be added? An alternative is to update infoCourse.json for all existing courses/repos, but that can be intrusive on its own as well.

nwalters512:
I think we can forego sync warnings for now. Many courses will never be edited locally, in which case they can inherit the institution's timezone in production, and if folks edit locally and are bothered by UTC, they can easily add a timezone to `infoCourse.json` to remedy that.

apps/prairielearn/src/sync/fromDisk/courseInstances.js
```diff
@@ -31,7 +30,7 @@ function getParamsForCourseInstance(courseInstance, courseTimezone) {
     long_name: courseInstance.longName,
     number: courseInstance.number,
     hide_in_enroll_page: courseInstance.hideInEnrollPage || false,
-    display_timezone: courseInstance.timezone || courseTimezone || 'America/Chicago',
+    display_timezone: courseInstance.timezone || null,
```
nwalters512, line 33: Nit, I generally reach for the nullish-coalescing operator for things like this:

```suggestion
    display_timezone: courseInstance.timezone ?? null,
```

nwalters512 (approved):

apps/prairielearn/src/pages/administratorCourseRequests/courseRequestsTable.ejs
```diff
@@ -55,7 +55,7 @@
                                     </button>
                                     <button type="button" class="btn btn-sm btn-success float-right mr-2" id="approve-request-button-<%= req.id %>"
                                             name="approve_deny_action" value="approve" tabindex="0"
-                                            data-toggle="popover" data-container="body"
+                                            data-toggle="popover" data-container="body" data-boundary="window"
```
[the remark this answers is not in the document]
jonatanschroeder: In some cases this popover would not respect window boundaries. Before: 
image

After: 
image

apps/prairielearn/src/sync/fromDisk/courseInstances.js
jonatanschroeder: The `||` would take care of blank timezones, though...

PR #7770, opened 2023-05-18
Update page titles for manual grading pages

### LINKED ISSUE #7552: Manual grading pages: review page titles
The page title for the instance question page of manual grading is based on the preview name for the question. The title for the assessment question is just a generic "Manual Grading" title. These can be confusing when using multiple tabs. We should use something more specific/sensible in this case.

jonatanschroeder:
Resolves #7552


nwalters512 (approved):
Seems fine 👍

PR #7772, opened 2023-05-18
Sessions fixes

trombonekenny:
- fixes cookie expiration treating seconds as milliseconds
- changes cookie security to use Express "trust proxy" for if it's secure or not
- adds config.sessionCookieSameSite which defaults to 'none' in production and 'lax' in development


apps/prairielearn/src/lib/config.js
```diff
@@ -57,6 +57,9 @@ const ConfigSchema = z.object({
   authUin: z.string().nullable().default('000000000'),
   authnCookieMaxAgeMilliseconds: z.number().default(30 * 24 * 60 * 60 * 1000),
   sessionStoreExpireSeconds: z.number().default(86400),
+  sessionCookieSameSite: z
+    .string()
+    .default(process.env.NODE_ENV === 'development' ? 'lax' : 'none'),
```
nwalters512, line 62: For anyone landing on this PR via `git blame`, we need `SameSite: None` for iframing in LTI tools like canvas. However, this breaks for local dev workflows that typically don't use TLS, as browsers will reject cookies with `SameSite: None` that don't also set `Secure`. This is our compromise: reasonable defaults with the ability to turn on `SameSite: None` locally if someone has TLS set up and wants to test iframes.

apps/prairielearn/src/server.js
```diff
@@ -108,6 +108,14 @@ module.exports.initExpress = function () {
     }
   }
 
+  //////////////////////////////////////////////////////////////////////
+  //////////////////////////////////////////////////////////////////////
+  //////////////////////////////////////////////////////////////////////
+  // Webhooks //////////////////////////////////////////////////////////
```
nwalters512, line 114:
```suggestion
  // This should come before the session middleware so that we don't
  // create a session every time we get a health check request.
```

nwalters512 (approved):

PR #7774, opened 2023-05-21
pl-order-blocks - distractor feedback 

CoffeePoweredComputers:
This pull request adds additional feedback mechanisms for distractor blocks in the pl-order-blocks element. It makes the following additions:

### new attribute for pl-answer-blocks - `distractor-feedback`

This attribute allows for feedback to be associated with a block where `correct=false`. For example:
```html
<pl-answer distractor-feedback="this block is incorrect as you should use parentheses rather than square brackets when attempting to call a function" correct="false">lst.pop[i]</pl-answer>  
```

### new attribute for pl-order-blocks - `distractor-info`

The distractor info attribute controls the information in the true_answer panel. It can take on the following three values:
* `"none"` (default): This shows the true answer panel as it has always been. Simply showing a correct answer and nothing more.
* `"show-distractors`": This shows the distractors in a new "incorrect blocks" pane that sits just below the existing "correct answer" pane with any feedback associated with each distractor via the `distractor-feedback` tag in pl-answer.
image

### new value for feedback attribute - first-wrong-verbose
If a student's first incorrect block is a distractor this displays any feedback included with the distractor below that block in the submission panel if any feedback is associated with that distractor.

With code formatting on:
image

Without code formatting on:
image


jonatanschroeder:
> new attribute for pl-answer-blocks - distractor-feedback

Nit: other elements like multiple choice or checkbox use only `feedback` as the attribute name. I'd argue for using that name for consistency.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -181,6 +184,12 @@ def prepare_tag(html_tags, index, group_info={"tag": None, "depends": None}):
         answer_indent = pl.get_integer_attrib(html_tags, "indent", None)
         inner_html = pl.inner_html(html_tags)
         ranking = pl.get_integer_attrib(html_tags, "ranking", -1)
+        distractor_info = pl.get_string_attrib(element, "distractor-info", DISTRACTOR_INFO_DEFAULT)
```
eliotwrobson, line 187: It looks like you're using the string `"none"` as a sentinel value. Instead, you should use the keyword `None` as the default sentinel, then your comparisons will be `distractor_info is None`.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -410,6 +423,9 @@ def render(element_html, data):
         return html
 
     elif data["panel"] == "submission":
+        
+        feedback_type = pl.get_string_attrib(element, 'feedback', FEEDBACK_DEFAULT)
```
eliotwrobson, line 427: Since you're using `feedback_type` as an enum with a fixed number of values, it would be best to do this explicitly along with the `get_enum_attrib` function.

eliotwrobson (changes requested):
See comments. Seems good so far but some parts of the backend can be cleaned up.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.css
```diff
@@ -72,3 +75,88 @@ div.pl-order-blocks-container {
...
+  padding: 5px !important;
+  border-radius: inherit;
+  border: 1px solid grey;
+}
+
+
+.feedback-tag{
+    color: #6c757d;
+    margin-right: 10px;
+}
```
SethPoulsen, line 162: Most of these classes are from old versions of the CSS of this element and should be deleted now. 

This is probably just an oops when trying to get this PR up to date with the latest version of the code.

See https://github.com/PrairieLearn/PrairieLearn/commit/63c6e11495926d3330e7573e8d0acee3ee081f41

The ones that are new as of this PR should be prefixed with `pl-order-blocks-` and done in kebab case like, `feedback-tag` should be something like `pl-order-blocks-feedback`, `feedback_badge_container` should probably be changed to something like `pl-order-blocks-feedback-badge`, etc.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -39,26 +39,50 @@ $(function() {
 
 {{#submission}}
     {{^parse-error}}
-    <div class="pl-order-blocks-answer-container">
-        <div class="card-header">
-            Your answer:
+    <div class="submission_container">
+        <ul class="list-group submission_header">
+            <li class="list-group-item"><strong>Your answer: </strong>
```
SethPoulsen, line 44: You shouldn't be reverting the order blocks CSS to the old version, see above comment.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -340,7 +351,7 @@ def render(element_html, data):
         mcq_options = data["params"][answer_name]
         mcq_options = filter_multiple_from_array(mcq_options, ["inner_html", "uuid"])
 
-        if answer_name in data["submitted_answers"]:
+        if answer_name in data["submitted_answers"]:
```
SethPoulsen, line 354: Make sure you run the black formatter over your code, you should be able to do this just with 

`make format-python`

docs/elements.md
```diff
@@ -447,9 +447,10 @@ Element to arrange given blocks of code or text that are displayed initially in
...
+| `distractor-info`       | "show-distractors" or "show-feedback"                                 | -                              | The level of feedback that is to be given to students with respect to the distractors. `show-distractors` will show students which blocks were incorrect after the problem has been solved or all attempts have been sued. `show-feedback` shows the same but includes the feedback strings associated with each distractors as we by the `distractor-feedback` attribute for each `<pl-answer>` block.                                                                                                                                                                                                                                                                                                          |
```
SethPoulsen, line 453: To make sure I understand, here we are using `feedback` to control the feedback the student is getting while working on the problem, and then `distractor-info` is used to control feedback they are shown once they have completed the problem or used all attempts?

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.mustache
SethPoulsen: done

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: done

docs/elements.md
SethPoulsen: @CoffeePoweredComputers I don't think there is any point in the "show-distractors" option here---if the student has completed the question, they can see the correct solution. So they are already shown (by them being omitted from the correct) which blocks are the distractor blocks, its just more work for no reason to figure it out if we don't directly show them.

With "show-feedback", it seems like if the question writer goes to the trouble of writing the feedback, they are going to want to show it. 

Given all the above, can we just get rid of this `distractor-info` property completely and just show it all by default?

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.css
SethPoulsen: Fixed

docs/elements.md
SethPoulsen: Re: the "first-wrong" and "first-wrong-verbose" options above:

There are at least 3 different ways that the line can be wrong:
- the current block group hasn't been finished
- the dependencies of the line haven't been met (based on the depends attribute"
- the line is a distractor

currently, the 'first-wrong' feedback doesn't say anything about why the block is wrong, just that its wrong. If we are going to make 'first-wrong-verbose' be an option, that is going to tell people about if the line is a distractor, should we also have it be telling them about the other 2 possible ways it could be wrong too? or do we want to configure these different things separately? I'm not completely sure myself.

SethPoulsen:
The displayed diff for this is currently sort of broken, because I merged this branch with the one from #7792 . 

That one will be merged before this one, so the diff will be fixed before we need to do another round of reviews on this one.

docs/elements.md
SethPoulsen: I have spend some more time thinking about this and I have decided what I want. 

I think that we should leave `feedback` exactly as @CoffeePoweredComputers implemented it. I think the differentiation between "first-wrong" and "first-wrong-verbose" will be useful for people, but I don't think anyone will care about having super granular control over feedback related to distractor vs. dependencies haven't been meet vs. block group not finished. If we do want to add more detailed feedback for dependencies not met and for block group not finished, we can add that in sometime in the future and just have it under the same umbrella as "first-wrong-verbose". 

I still think we should get rid of the `distractor-info` attribute altogether and just have it on all the time. If people don't want the feedback to be shown, they won't write any.

SethPoulsen: I talked to @CoffeePoweredComputers in person and we are on the same page about these options. I will go ahead and make the changes.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: resolved by the fact that we got rid of this options altogether

SethPoulsen:
> other elements like multiple choice or checkbox use only feedback as the attribute name. I'd argue for using that name for consistency.

@jonatanschroeder if we do that, then `pl-order-blocks` will have a `feedback` property which defines what type of feedback is given, and then the `pl-answer` inside can also each have a `feedback` property to actually define the feedback messages. Is this ok or is it confusing?

jonatanschroeder:
> I'm slightly worried that we would then have two attributes have the same name but the values they take are a different type--one of them is essentially an Enum and the other is a string.

That's a good point. A different name in this case makes sense, though I'm not sure about `distractor-feedback`, as it implies it's only used for distractors (is that the case, and will be the case in likely future improvements?).

SethPoulsen:
New UI screenshots after my tweaks:
image

image

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: this was an already existing attribute, not new to this PR, so I'm not going to refactor to use `get_enum_attrib` as part of this PR. But it should be done in future when the typechecking is put back in to this element, along with converting a few of the other string attributes to Enums.

SethPoulsen (approved):
As with #7792, this was a collaboration between myself and @CoffeePoweredComputers, so while I approve these changes, I am going to wait for approval from someone else before merging.

SethPoulsen:
> A different name in this case makes sense, though I'm not sure about distractor-feedback, as it implies it's only used for distractors (is that the case, and will be the case in likely future improvements?).

This is a good question. Currently yes, only used for distractors. I don't see myself giving a student targeted feedback about on of the _correct_ blocks,  I don't think students ever look at feedback when they get something right anyway. But I suppose it is possible that an instructor would want to do that. One nice thing about the current name is that it goes nicely with the `distractor-for` attribute added in #7792.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -555,6 +583,7 @@ def parse(element_html, data):
         element, "grading-method", GRADING_METHOD_DEFAULT
     )
     correct_answers = data["correct_answers"][answer_name]
+    mcq_options = data["params"][answer_name]
```
nwalters512, line 586: `mcq_options` - is this a bad copy/paste from some other element?

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -568,6 +597,18 @@ def parse(element_html, data):
             )
             answer["ranking"] = search["ranking"] if search is not None else None
             answer["tag"] = search["tag"] if search is not None else None
+            if search is None:
```
nwalters512, line 600: I acknowledge that this `search` variable wasn't introduced in this PR, but I don't fully understand what it is. Is there a better variable name that we can use here?

docs/elements.md
```diff
@@ -434,34 +434,35 @@ Element to arrange given blocks of code or text that are displayed initially in
...
+| `partial-credit`      | "none" or "lcs"                                 | Depends on `grading-method`    | For the `dag` and `ranking` grading methods, you may specify `none` for no partial credit or `lcs` (default) for partial credit based on the LCS edit-distance from the student solution to some correct solution. For the other grading methods, using this property is not yet supported. Grading method `unordered` will always assign partial credit, and grading method `ordered` will never do so.                                                       |
+| `feedback`            | "none", "first-wrong", or "first-wrong-verbose" | "none"                         | The level of feedback the student will recieve upon giving an incorrect answer. Available with the `dag` or `ranking` grading mode. `first-wrong` will tell the student which block in their answer was the first to be incorrect, `none` will give no feedback. If set to "first-wrong-verbose", if the first incorrect block is a distractor any feedback associated with that distractor will be shown as well (see "distractor-feedback" in `<pl-answer>`) | .   |
```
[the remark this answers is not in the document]
SethPoulsen: done

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.css
```diff
@@ -85,13 +91,28 @@ div.pl-order-blocks-container {
...
+  color: #6c757d;
+  margin-right: 10px;
+}
+
+.pl-order-blocks-correctness-badge-container {
+  display: flex;
+  align-items: center;
+}
+.pl-order-blocks-correctness-badge {
+  margin:4px;
```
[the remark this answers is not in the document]
SethPoulsen: fixed.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.css
```diff
@@ -85,13 +91,28 @@ div.pl-order-blocks-container {
...
 .pl-order-blocks-pairing-indicator {
   border-radius: 10px;
   border: 0px !important;
   list-style: none;
-  margin-left: -10px;
-  margin-right: -10px;
-  padding: 4px 10px 4px 10px;
+  margin-left: -6px;
+  margin-right: -6px;
+  padding: 4px 6px 4px 6px;
```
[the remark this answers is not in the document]
SethPoulsen: I'm not sure what you are getting at. Do you have something against the horizontal and vertical padding being different?

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.css
```diff
@@ -85,13 +91,28 @@ div.pl-order-blocks-container {
   align-items: center;
 }
 
+.pl-order-blocks-feedback-icon {
+  color: #6c757d;
```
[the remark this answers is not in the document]
SethPoulsen: fixed, thanks.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -67,7 +67,30 @@ $(function() {
...
-            	<li class="{{block_formatting}}" style="margin-left: {{indent}}px">{{{inner_html}}}</li>
+                {{#show_distractor_feedback}}
+                <li class="pl-order-blocks-pairing-indicator list-group-item-warning">
+                    <ul style="padding:0px;">
+                        <li class="" style="list-style-type: none; ">
+                            <div class="pl-order-blocks-correctness-badge-container">
+                                <span class="pl-order-blocks-correctness-badge badge {{badge_type}}"><i class="fa {{icon}}" aria-hidden="true"></i></span>
+                                <div class="{{block_formatting}} pl-order-block" style="margin-left: {{indent}}px;flex-grow:1;">{{{inner_html}}}</div>
+                            </div>
+                        </li>
```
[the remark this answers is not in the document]
SethPoulsen: nice catch! all fixed.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: ya sorry, all fixed now

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: This code goes through all blocks in the problem to find the one whose text matches a block in the student answer. 

I'll rename it to `matching_block`, that will be more useful.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -565,30 +596,70 @@ def parse(element_html, data):
...
                 (
                     item
                     for item in correct_answers
                     if item["inner_html"] == answer["inner_html"]
                 ),
                 None,
             )
-            answer["ranking"] = search["ranking"] if search is not None else None
-            answer["tag"] = search["tag"] if search is not None else None
+            answer["ranking"] = (
```
eliotwrobson, line 611: Instead of using inline assignment for these two keys, just wrap this block with a `if matching_block is None` check. You can even combine it with the one below, I think.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -565,30 +596,70 @@ def parse(element_html, data):
...
+            )
+            answer["tag"] = (
+                matching_block["tag"] if matching_block is not None else None
+            )
+            if matching_block is None:
+                matching_block = next(
+                    (
+                        item
+                        for item in blocks
+                        if item["inner_html"] == answer["inner_html"]
```
eliotwrobson, line 622: This block of code is almost identical to the one on line 603. This should be done with a helper function.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -565,30 +596,70 @@ def parse(element_html, data):
...
+            matching_block = next(
                 (
                     item
                     for item in correct_answers
                     if item["inner_html"] == answer["inner_html"]
                 ),
                 None,
             )
-            answer["tag"] = search["tag"] if search is not None else None
+            answer["tag"] = (
```
eliotwrobson, line 641: Same comments as above; this setting logic is very self-similar and can probably be cleaned up using some helper functions.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -702,6 +779,28 @@ def grade(element_html, data):
             -1 if num_initial_correct == len(submission) else num_initial_correct
         )
 
+        if feedback_type.startswith("first-wrong"):
+            for i, block in enumerate(student_answer):
+                if i >= num_initial_correct:
```
eliotwrobson, line 784: Instead of this break, use a list slice on `student_answer` (like `student_answer[:num_inital_correct]`) in the enumerate call.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -702,6 +779,28 @@ def grade(element_html, data):
...
+                student_answer[first_wrong]["icon"] = "fa-xmark"
+
+                for block in student_answer[first_wrong + 1 :]:
+                    block["badge_type"] = ""
+                    block["icon"] = ""
+
+        num_initial_correct, true_answer_length = grade_dag(
+            submission, depends_graph, group_belonging
+        )
+        first_wrong = (
```
eliotwrobson, line 800: Nit: Use `None` as the sentinel value instead of -1.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -718,9 +817,16 @@ def grade(element_html, data):
         if final_score < 1:
             if feedback_type == "none":
                 feedback = ""
-            elif feedback_type == "first-wrong":
+            elif feedback_type.startswith("first-wrong"):
                 if first_wrong == -1:
                     feedback = FIRST_WRONG_FEEDBACK["incomplete"]
+                elif (
+                    feedback_type.endswith("verbose")
```
eliotwrobson, line 824: Isn't there only one verbose feedback type? Should explicitly check for it in that case.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -775,9 +880,13 @@ def test(element_html, data):
     # TODO: The only wrong answer being tested is the correct answer with the first
     # block mising. We should instead do a random selection of correct and incorrect blocks.
     elif data["test_type"] == "incorrect":
-        answer = filter_multiple_from_array(
-            data["correct_answers"][answer_name], ["inner_html", "indent", "uuid"]
-        )
+        answer = deepcopy(data["correct_answers"][answer_name])
+        distractor_feedback = {
+            item["inner_html"]: item["distractor_feedback"]
```
eliotwrobson, line 885: Are these answers being uniquely identified by their inner HTML? Would be better to set a uuid as a unique identifier (or something similar).

eliotwrobson (changes requested):
See comments, in general I think a lot of the recently added code is repetitive and can be substantially simplified.

Unless this feature is urgent, I'm inclined to say that we should include some type annotations along with this. The code for this element is starting to get unruly and it makes it difficult to review, and enabling typechecking would make this substantially easier to organize.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: Ya, I agree with you its not cool. These actually have UUIDs on them now, but this code is a relic of the time years ago when they didn't. 

Changing this to use UUIDs now would break some questions from a few years ago. Maybe its ok if we break some questions from >2 years ago to improve code quality? What do you think @nwalters512 ?

SethPoulsen: What I mean is, there are some questions from courses run >2 years ago, that, if someone reopened them and tried to submit something else, this grading code wouldn't work.

SethPoulsen: We already recently changed the grading of the `unordered` grading mode to use the UUIDs anyway

SethPoulsen:
> enabling typechecking would make this substantially easier to organize

Here we go: #8013 

I did it as a separate PR so this one can be cleaner. We can merge that one first and then this one!

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: done

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: done

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: I fixed all these when I merged with the new typechecking.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -453,8 +462,21 @@ def render(element_html, data):
...
+                    "badge_type": attempt["badge_type"]
+                    if (feedback_type.startswith("first-wrong"))
+                    else "",
+                    "icon": attempt["icon"]
+                    if (feedback_type.startswith("first-wrong"))
+                    else "",
+                    "first_wrong": feedback_type.startswith("first-wrong"),
+                    "distractor_feedback": attempt.get("distractor_feedback"),
+                    "show_distractor_feedback": (
+                        attempt.get("distractor_feedback") is not None
```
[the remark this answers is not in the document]
SethPoulsen: done

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -453,8 +462,21 @@ def render(element_html, data):
                 {
                     "inner_html": attempt["inner_html"],
                     "indent": (attempt["indent"] or 0) * TAB_SIZE_PX,
+                    "badge_type": attempt["badge_type"]
+                    if (feedback_type.startswith("first-wrong"))
```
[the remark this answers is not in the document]
SethPoulsen: fixed when merging with new typechecking

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -498,29 +506,40 @@ def render(element_html: str, data: pl.QuestionData) -> str:
...
+        )
+        show_first_wrong = feedback_type in [
+            FeedbackType.FIRST_WRONG,
+            FeedbackType.FIRST_WRONG_VERBOSE,
+        ]
+        student_submission = [
+            {
+                "inner_html": attempt["inner_html"],
+                "indent": (attempt["indent"] or 0) * TAB_SIZE_PX,
+                "badge_type": attempt["badge_type"] if show_first_wrong else "",
```
eliotwrobson, line 520: Is the idea here to include this set of information only if `show_first_wrong` is set to true? If so, it would be better to add all these keys to the dictionary under a single `if show_first_wrong:` statement, instead of having a bunch of these inline if-else statements.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -498,29 +506,40 @@ def render(element_html: str, data: pl.QuestionData) -> str:
...
+                "icon": attempt["icon"] if show_first_wrong else "",
+                "first_wrong": show_first_wrong,
+                "distractor_feedback": attempt.get("distractor_feedback"),
+                "show_distractor_feedback": "distractor_feedback" in attempt
+                and attempt["badge_type"] == "badge-danger"
```
eliotwrobson, line 525: Is the idea here to check whether the `attempt` was marked correct or not? I think there are more expressive ways of doing this than checking whether the badge is equal to `"badge-danger"`.

In general, it's best if we can reduce the number of comparisons involving string literals, as this makes the code harder to read and usually indicates something can be simplified.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -498,29 +506,40 @@ def render(element_html: str, data: pl.QuestionData) -> str:
...
+                and feedback_type is FeedbackType.FIRST_WRONG_VERBOSE,
+            }
+            for i, attempt in enumerate(data["submitted_answers"][answer_name])
```
eliotwrobson, line 528: Looks like `i` is unused?

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -760,6 +831,25 @@ def grade(element_html: str, data: pl.QuestionData) -> None:
             None if num_initial_correct == len(submission) else num_initial_correct
         )
 
+        if feedback_type in [
```
eliotwrobson, line 834: An identical check to this appears in a couple places; should just set to a variable.

eliotwrobson (commented):
Some weird control flow (though I realized these files were still being edited while I was writing the comments).

SethPoulsen:
>  files were still being edited while I was writing the comments

Ya, I'm still in the middle of things. I'll go ahead and re-request review when I'm ready. Thanks!

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: I combined some if blocks instead to get rid of the code duplication.

SethPoulsen: I combined some if blocks instead to get rid of the code duplication.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: I combined some if blocks instead to get rid of the code duplication.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: I would also need an else block to assign none, I think it would just get more complicated that way.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: fixed

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: Its actually more subtle than this, its only if the block is the first one that was wrong. 

Anyway, I agree with you about string comparisons, it was just a hack while refactoring and I figured out a better way to do it.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: fixed

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: I refactored to handle all of this conditional stuff in `grade` as it was before, some of it was being handled in `grade` and some here in `render`. Now its all in one place which I think is much cleaner.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: Sorry, I just realized that this comment was on the `test` function, not on `grade`. 

I can fix this here, and I have done so. 

I'm going to leave it the way it is in `grade` for now unless @nwalters512 thinks we should change it even with the potential to break old (>2 year old) questions.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -837,25 +891,43 @@ def test(element_html: str, data: pl.ElementTestData) -> None:
     # TODO: The only wrong answer being tested is the correct answer with the first
     # block mising. We should instead do a random selection of correct and incorrect blocks.
     elif data["test_type"] == "incorrect":
-        answer = filter_multiple_from_array(
-            data["correct_answers"][answer_name], ["inner_html", "indent", "uuid"]
-        )
+        answer = deepcopy(data["correct_answers"][answer_name])
+        distractor_feedback = {
```
eliotwrobson, line 895: Looks like the distractor feedback dicts are using different keys when they're constructed in different places?

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -837,25 +891,43 @@ def test(element_html: str, data: pl.ElementTestData) -> None:
...
+            if not item["is_correct"]
+        }
+
         answer.pop(0)
         score = 0
         if grading_method is GradingMethodType.UNORDERED or (
             (
-                grading_method is GradingMethodType.DAG
-                or grading_method is GradingMethodType.RANKING
+                grading_method in [GradingMethodType.DAG, GradingMethodType.RANKING]
```
eliotwrobson, line 905: Similar to the feedback type list, this list can hardcoded once (it's constructed twice in this file).

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -695,6 +717,12 @@ def grade(element_html: str, data: pl.QuestionData) -> None:
 
     true_answer_list = data["correct_answers"][answer_name]
 
+    distractor_feedback = {
```
eliotwrobson, line 720: Not 1000% sure, but it does't look like the values in this dict are ever accessed (only the keys). Also uses the inner HTML instead of a UUID for the key.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -59,9 +60,12 @@ class OrderBlocksAnswerData(TypedDict):
     depends: list[str]  # only used with DAG grader
     group_info: GroupInfo  # only used with DAG grader
     distractor_bin: NotRequired[str]
+    distractor_feedback: Optional[str]
+    is_correct: bool
     uuid: str
 
 
+FIRST_WRONG_TYPES = [FeedbackType.FIRST_WRONG, FeedbackType.FIRST_WRONG_VERBOSE]
```
eliotwrobson, line 68:
```suggestion
FIRST_WRONG_TYPES = frozenset((FeedbackType.FIRST_WRONG, FeedbackType.FIRST_WRONG_VERBOSE))
```
Not completely sure this will work but nice to have an immutable type rather than a list.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -760,6 +788,26 @@ def grade(element_html: str, data: pl.QuestionData) -> None:
...
+            for block in student_answer[:num_initial_correct]:
+                block["badge_type"] = "badge-success"
+                block["icon"] = "fa-check"
+                block["distractor_feedback"] = ""
+
+            if first_wrong is not None:
+                student_answer[first_wrong]["badge_type"] = "badge-danger"
+                student_answer[first_wrong]["icon"] = "fa-xmark"
+                if feedback_type is not FeedbackType.FIRST_WRONG_VERBOSE:
+                    student_answer[first_wrong]["distractor_feedback"] = ""
```
eliotwrobson, line 801: Is this necessary? Seems strange to delete the data given in the student answer rather than just not display it later on somewhere.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -837,25 +891,43 @@ def test(element_html: str, data: pl.ElementTestData) -> None:
...
-            and feedback_type is FeedbackType.FIRST_WRONG
+            grading_method
+            in [
+                GradingMethodType.DAG,
+                GradingMethodType.RANKING,
+            ]
+            and feedback_type in FIRST_WRONG_TYPES
         ):
-            feedback = FIRST_WRONG_FEEDBACK["wrong-at-block"].format(1)
+            first_wrong = 0
```
eliotwrobson, line 919: This logic seems very similar to some feedback setting logic above. I know this is test code, but having that factored out into a function would be cleaner IMO.

eliotwrobson (changes requested):
Getting closer, but a few oddities that need to be hashed out.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: Whether or not it is displayed depends on the value of `first_wrong`, which (now that we removed it from the `partial_scores` dict), is only accessible here in the `grade` function and not later on in `render`. That's why I figured this might be the cleanest way to do it.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -521,6 +531,9 @@ def render(element_html: str, data: pl.QuestionData) -> str:
             "student_submission": student_submission,
             "feedback": feedback,
             "block_formatting": block_formatting,
+            "allow_feedback_badges": not all(
+                block.get("badge_type", "") == "" for block in student_submission
```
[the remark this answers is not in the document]
SethPoulsen: Yes, look the `grade` function.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: Its actually more places than two. I don't really want to change this right now though because its almost completely unrelated to the actual content of this PR. I also don't think there is an obvious thing to call this list as there was with the feedback list.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: ya, it works. Done.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: done.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: cleaned up

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -676,6 +708,35 @@ def parse(element_html: str, data: pl.QuestionData) -> None:
...
+
+    if first_wrong is None:
+        return FIRST_WRONG_FEEDBACK["incomplete"]
+    elif (
+        feedback_type is FeedbackType.FIRST_WRONG_VERBOSE and first_wrong_is_distractor
+    ):
+        return FIRST_WRONG_FEEDBACK["distractor-feedback"].format(str(first_wrong + 1))
+    else:
+        feedback = FIRST_WRONG_FEEDBACK["wrong-at-block"].format(str(first_wrong + 1))
+        has_block_groups = group_belonging != {} and set(group_belonging.values()) != {
```
eliotwrobson, line 729:
```suggestion
        has_block_groups = group_belonging and set(group_belonging.values()) != {
```
Nit: non-empty dicts evaluate to true.

eliotwrobson (approved):
Python looks good! One minor nitpick.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: To someone who has written a lot of JavaScript, this looks so wrong (in JS, the empty object `{}` evaluates to true). Most of the other PL devs have done a lot of JS too so I'm not sold on this change. Do we have a convention for this in the codebase more generally?

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -67,7 +67,34 @@ $(function() {
         </div>
         <ul class="card-body list-group">
             {{#student_submission}}
-            	<li class="{{block_formatting}}" style="margin-left: {{indent}}px">{{{inner_html}}}</li>
+                {{#distractor_feedback}}
+                <li class="pl-order-blocks-pairing-indicator list-group-item-warning">
+                    <ul style="padding:0px;">
```
nwalters512, line 72: Why are we using both inline styles and classes? Why not use `p-0` here?

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -67,7 +67,34 @@ $(function() {
...
+                        <li class="" style="list-style-type: none; ">
```
nwalters512, line 73: Why the empty class? Note that Bootstrap has a `list-unstyled` style that might fit your needs here.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.css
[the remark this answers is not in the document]
nwalters512: Yes. To me, this looks bad:

Screenshot 2023-07-05 at 10 12 50

And this looks better:

Screenshot 2023-07-05 at 10 13 42

nwalters512 (commented):
Someday we'll hire a designer to take care of the cards-in-cards-in-cards problem 😅

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
nwalters512: Yes; in Python code, we use idiomatic Python, which means using the form that Eliot is suggesting.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.mustache
SethPoulsen: sounds good

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.mustache
SethPoulsen: oh cool, I didn't know that existed. will fix.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.css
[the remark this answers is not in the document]
SethPoulsen: Did you mean to change all of them to 6px? in the suggested edit, you left one of them as 4px.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.mustache
```diff
@@ -67,7 +67,34 @@ $(function() {
         </div>
         <ul class="card-body list-group">
             {{#student_submission}}
-            	<li class="{{block_formatting}}" style="margin-left: {{indent}}px">{{{inner_html}}}</li>
+                {{#distractor_feedback}}
+                <li class="pl-order-blocks-pairing-indicator list-group-item-warning">
+                    <ul class="p-0">
+                        <li class="list-unstyled">
```
nwalters512, line 73: Nit: check out the Bootstrap docs. This class should be on the `<ul>` element, and it will automatically remove the padding (so no need for `p-0`).

https://getbootstrap.com/docs/4.6/content/typography/#unstyled

nwalters512 (approved):

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.mustache
SethPoulsen: done, thanks

PR #7775, opened 2023-05-21
token-based-extension-system for PrarieLearn

vikramrkini:
I wanted to share with you an initial Request for Comments (RFC) regarding a proposed late-day token-based extension system. This system is intended to be seamlessly integrated with Prarie Learn. I would greatly appreciate your valuable feedback and suggestions for improvement. Thanks.


vikramrkini:

-Vikram Kini

nwalters512:
Hey Vikram, you need to copy/paste the CLA agreement verbatim for our bot to pick it up (that is, don't append your name).

rfcs/006-token-based-extension-system.md
```diff
@@ -0,0 +1,44 @@
...
+2.2 Additional Tokens: Instructors will have the discretion to assign additional tokens to individual students on a case-by-case basis if deemed necessary. This flexibility allows instructors to accommodate special circumstances or provide extra support to specific students.
+
+## Token Usage
+
+3.1 Extension Types: Each token will represent a specific type of extension, and the following options are proposed as initial suggestions:
+a) One token = one day extension
+b) One token = one week extension
+c) One token = student selects the new due date
+
+3.2 Token Constraints: Instructors will define the assessment sets where tokens can be utilized at the course instance level. Assessment sets can include various types of assessments such as homework, quizzes, exams, machine problems, etc. By specifying token constraints, instructors can control which assessments are eligible for extension requests using tokens. This provides instructors with the flexibility to tailor the extension policy to suit the nature and requirements of each assessment.
```
nwalters512, line 20: Two questions here:

- I assume this is reusing PrairieLearn's existing notion of assessment sets? Or is this a separate token-specific notion of assessment sets? I _hope_ that assessment sets would be sufficient for instructors, but I do wonder if this would provide sufficient flexibility for students.
- You specifically mention quizzes and exams here. What would it mean for a student to use a token to extend a deadline for one of those? I'm imagining we wouldn't try to support extensions for assessments administered via a CBTF via PrairieTest. If so, we should call that out as an explicit non-goal.

rfcs/006-token-based-extension-system.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Requesting Extensions
+
+4.1 Student Request: When a student requires an extension on a specific assessment within an eligible assessment set, they can submit a request indicating the desired length of the extension.
+
+4.2 Token Consumption: Upon approval of the extension request, the system will deduct the appropriate number of tokens from the student's token balance based on the chosen extension type.
+
+4.3 Token Validation: The system will verify if the student has a sufficient number of tokens to cover the requested extension. If the student lacks the necessary tokens, the request will be automatically rejected.
+
+## Considerations and Future Enhancements
```
nwalters512, line 30: Is it fair to say that these are all outside the scope of an initial implementation of this?

nwalters512 (commented):
At a high level, this all sounds reasonable. Next step is hashing out design and implementation. Mariana mentioned you wanted to meet to discuss that sometime. Feel free to reach out on Slack and we can get some time on the calendar!

rfcs/006-token-based-extension-system.md
vikramrkini: Yes, these are future considerations and enhancements. It won't be a part of the initial implementation.

vikramrkini:
recheck

rfcs/006-token-based-extension-system.md
vikramrkini: 1) It will be reusing all the existing sets of assessments already present in PrarieLearn.
2) Yes, for the initial implementation we will only be working on the standard assessments such as group activities and pre-class quizzes etc. All the assessments administered through the CBTF via PrairieTest are not a part of the initial implementation.

mfsilva22: I think the tokens should be allowed to be applied to any PL set. If a instructor does not want it to be applied to a set that is called quizzes, they have the option to do that. I don't think we can pre-select the types of sets that can have this feature applied or not. 

Assessment that are controlled via PT will not appear on the assessment page outside the time the student is taking the assessment, so at least they won't be able to "request an extension" like the other assessments.

nwalters512:
@elisacarrillo is continuing with the implementation of this feature, starting with #9103.

PR #7777, opened 2023-05-23
Modernize enrollment page

nwalters512:


apps/prairielearn/src/pages/enroll/enroll.sql
```diff
@@ -1,7 +1,7 @@
 -- BLOCK select_course_instances
 SELECT
   c.short_name || ': ' || c.title || ', ' || ci.long_name AS label,
-  c.short_name || ', ' || ci.short_name AS short_label,
+  c.short_name || ', ' || ci.long_name AS short_label,
```
nwalters512, line 4: IMO we should never be exposing course instance short names to students.

apps/prairielearn/src/pages/enroll/enroll.html.ts
```diff
@@ -0,0 +1,282 @@
...
+                                      Add course
+                                    </button>
+                                    <div
+                                      class="modal fade"
+                                      id="addModal${idx}"
+                                      tabindex="-1"
+                                      role="dialog"
+                                      aria-labelledby="addModal${idx}Title"
+                                      aria-hidden="true"
+                                    >
```
nwalters512, line 74: We really shouldn't be doing the inline-modal-per-row thing. I can make the change to a templated modal in this PR if that's desired.

apps/prairielearn/src/pages/enroll/enroll.ts
nwalters512, line 1: I had a somewhat annoying time testing these changes since the "enroll" buttons are never shown for admins. I ended up just hardcoding the query to return `false` for the `instructor_access` column. In PT we have e.g. `dev-student@example.com` for testing things as non-admins. AFAIK, there's no equivalent in PL, is that correct?

jonatanschroeder: You can change the UID/UIN/Name in the config file to test a non admin locally

apps/prairielearn/src/pages/enroll/enroll.html.ts
mwest1066: As you like. The templated modal is definitely the way to go (including pulling `Modal()` and `templateFromAttributes()` over from PT).

apps/prairielearn/src/pages/enroll/enroll.sql
mwest1066: I'm on board with that.

mwest1066 (approved):

apps/prairielearn/src/pages/enroll/enroll.html.ts
nwalters512: I think I'll leave this as-is for now just so I can continue progress on enrollment limits and billing.

We should discuss our strategy for those sort of generic-ish bits of client-side functionality. Do we duplicate them, or put them in some kind of sharable package in PL?

PR #7783, opened 2023-05-25
Add admin pages for institutions/courses/instances

nwalters512:
This PR splits the global admin view of institutions out of the "Courses" tab:

Screenshot 2023-05-25 at 15 08 14

Screenshot 2023-05-25 at 15 08 27

It also scaffolds out some admin-specific pages for institutions, courses, and course instances:

Screenshot 2023-05-25 at 15 08 39

Screenshot 2023-05-25 at 15 08 48

Screenshot 2023-05-25 at 15 09 16

Screenshot 2023-05-25 at 15 09 25

They are currently very barebones, but provide us with a nice place to put some upcoming billing things. They also lay some groundwork for an eventual institution admin role.


apps/prairielearn/src/ee/institution/admin/course.html.ts
```diff
@@ -0,0 +1,73 @@
...
+        })}
+      </head>
+      <body>
+        ${renderEjs(__filename, "<%- include('../../../pages/partials/navbar') %>", {
+          ...resLocals,
+          institution,
+          navbarType: 'institution',
+          navPage: 'institution_admin',
+          navSubPage: 'courses',
+        })}
```
nwalters512, line 33: This little snippet gets really tedious copying around 😓 I wonder if we should have some notion of layout components that can handle this boilerplatey stuff?

apps/prairielearn/src/ee/institution/admin/course.html.ts
```diff
@@ -0,0 +1,73 @@
...
+  institution,
+  course,
+  courseInstances,
+  resLocals,
+}: {
+  institution: Institution;
+  course: Course;
+  courseInstances: CourseInstance[];
+  resLocals: Record<string, any>;
+}) {
```
nwalters512, line 15: I'm using inline types as previously discussed.

apps/prairielearn/src/lib/db-types.ts
nwalters512, line 1: `db-types` and `IdSchema` 🎉

apps/prairielearn/src/pages/administratorInstitutions/administratorInstitutions.html.ts
```diff
@@ -0,0 +1,92 @@
...
+                    <th>Short name</th>
+                    <th>Long name</th>
+                    <th>Timezone</th>
+                    <th>UID regexp</th>
+                    <th>Authn providers</th>
+                  </tr>
+                </thead>
+                <tbody>
+                  ${institutions.map(
+                    (inst) => html`
```
mwest1066, line 61: To match the suggested change to the schema:
```suggestion
                    ({institution, authn_providers}) => html`
```
The naming convention I've been using here is that whenever I have a variable called `table_name` (singular) then I know it's a row from the table, and `table_names` (plural) means it's an array of rows from the named table. I avoid using abbreviations unless really necessary, to cut down on the cognitive overhead. If I have something else (like an array of authn provider names) then I call it something else (e.g., `authn_provider_names`) but I try and always use the full types wherever possible -- I've found it makes it much easier/faster to modify code.

apps/prairielearn/src/ee/institution/utils.ts
```diff
@@ -1,38 +1,49 @@
...
 
 const sql = loadSqlEquiv(__filename);
 
-export async function getInstitution(institutionId) {
-  const institutionRes = await queryOneRowAsync(sql.select_institution, {
-    id: institutionId,
-  });
-  return institutionRes.rows[0];
+export async function getInstitution(institutionId: string): Promise<Institution> {
+  const institutionRes = await queryRow(
```
mwest1066, line 16: Using singular-name-of-table for the variable name:
```suggestion
  const institution = await queryRow(
```

apps/prairielearn/src/ee/institution/utils.ts
```diff
@@ -1,38 +1,49 @@
...
+} from '../../lib/db-types';
 
 const sql = loadSqlEquiv(__filename);
 
-export async function getInstitution(institutionId) {
-  const institutionRes = await queryOneRowAsync(sql.select_institution, {
-    id: institutionId,
-  });
-  return institutionRes.rows[0];
+export async function getInstitution(institutionId: string): Promise<Institution> {
```
mwest1066, line 15: I've been using `institution_id` for these variables to match what they look like in SQL (e.g., `courses. institution_id`).

apps/prairielearn/src/ee/institution/utils.ts
```diff
@@ -1,38 +1,49 @@
...
+    { institution_id: institutionId },
+    AuthnProviderSchema
+  );
 }
 
-export async function getSupportedAuthenticationProviders() {
-  const authProvidersRes = await queryAsync(sql.select_authentication_providers, {});
-  return authProvidersRes.rows.filter((row) => {
+export async function getSupportedAuthenticationProviders(): Promise<AuthnProvider[]> {
+  const authProviders = await queryRows(sql.select_authentication_providers, AuthnProviderSchema);
```
mwest1066, line 45: I would call this variable the plural table name.
```suggestion
  const auth_providers = await queryRows(sql.select_authentication_providers, AuthnProviderSchema);
```
As the variable names become more complex I'm less sure what to do. E.g., if we have new and old authn providers, are they `new_authn_provider`/`old_authn_provider` or `newAuthnProvider`/`oldAuthnProvider`? I haven't always been consistent with this.

apps/prairielearn/src/ee/institution/admin/course.html.ts
mwest1066: I'm on board, but I don't see how it will save much code? I'm mainly interested in getting rid of the EJS but that will take some time.

apps/prairielearn/src/ee/institution/admin/course.sql
```diff
@@ -0,0 +1,21 @@
+-- BLOCK select_course
+SELECT
+  *
```
mwest1066, line 3: For consistency I always convert to `jsonb` so that we get the same data back from the DB irrespective of whether it's a single row or a deeply-nested structure.
```suggestion
  to_jsonb(*)
```
I actually tend to do the `pl_courses AS c` and use `to_jsonb(c)` even in this case, just for consistency, but that's certainly not necessary.

apps/prairielearn/src/ee/institution/admin/course.html.ts
```diff
@@ -0,0 +1,73 @@
...
+        <main class="container mb-4">
+          <div class="table-responsive">
+            <table class="table table-hover table-striped table-bordered table">
+              <thead>
+                <tr>
+                  <th>Name</th>
+                </tr>
+              </thead>
+              <tbody>
+                ${courseInstances.map((courseInstance) => {
```
mwest1066, line 53: I would tend to call these:
```suggestion
                ${course_instances.map((course_instance) => {
```

apps/prairielearn/src/ee/institution/admin/course.ts
```diff
@@ -0,0 +1,37 @@
...
+import { InstitutionAdminCourse } from './course.html';
+import { getInstitution } from '../utils';
+
+const sql = loadSqlEquiv(__filename);
+const router = Router({ mergeParams: true });
+
+router.get(
+  '/',
+  asyncHandler(async (req, res) => {
+    const institution = await getInstitution(req.params.institution_id);
```
mwest1066, line 14: I've found that the `selectAndAuthz...` functions in PT do a nice job of combining authz with data fetching. There is certainly no need to shift to that model here, however, unless you want to. For this page both the `course` and `institution` would be fetched by `selectAndAuthz...` functions and only `courseInstances` would need a dedicated page query.

apps/prairielearn/src/ee/institution/admin/courseInstance.ts
```diff
@@ -0,0 +1,39 @@
...
+      {
+        institution_id: req.params.institution_id,
+        course_id: req.params.course_id,
+        course_instance_id: req.params.course_instance_id,
+      },
+      z.object({
+        course: CourseSchema,
+        course_instance: CourseInstanceSchema,
+      })
+    );
```
mwest1066, line 27: In this case we could avoid all SQL for the page by using `selectAndAuthz...` functions. Not pushing you to switch to this model, but rather just pointing out how they would work.

apps/prairielearn/src/ee/institution/admin/courseInstance.html.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export function InstitutionAdminCourseInstance({
+  institution,
+  course,
+  courseInstance,
+  resLocals,
+}: {
+  institution: Institution;
+  course: Course;
+  courseInstance: CourseInstance;
```
mwest1066, line 13: I love arguments like this that are all using the `db-types`!

apps/prairielearn/src/ee/institution/admin/courseInstance.sql
```diff
@@ -0,0 +1,12 @@
+-- BLOCK select_course_and_instance
+SELECT
+  to_jsonb(c.*) AS course,
+  to_jsonb(ci.*) AS course_instance
```
mwest1066, line 4: Nice!

apps/prairielearn/src/ee/institution/admin/courses.html.ts
```diff
@@ -0,0 +1,59 @@
...
+                  <th>Name</th>
+                </tr>
+              </thead>
+              <tbody>
+                ${courses.map((course) => {
+                  return html`
+                    <tr>
+                      <td>
+                        <a href="/pl/institution/${institution.id}/admin/course/${course.id}">
+                          ${course.short_name ?? '—'}: ${course.title ?? '—'}
```
mwest1066, line 46: I've been using mini-components like `CourseNameAndTitle(course)` for things like this, to maintain consistency across pages. Totally not necessary, of course.

apps/prairielearn/src/ee/institution/admin/general.ts
```diff
@@ -0,0 +1,39 @@
...
+      sql.select_institution_statistics,
+      { institution_id: req.params.institution_id },
+      InstitutionStatisticsSchema
+    );
+    res.send(
+      InstitutionAdminGeneral({
+        institution,
+        courseCount: institutionStatistics.course_count,
+        courseInstanceCount: institutionStatistics.course_instance_count,
+        enrollmentCount: institutionStatistics.enrollment_count,
```
mwest1066, line 32: I would generally pass the entire `institutionStatistics` object here, so that if we want to add more statistics in the future we only need to add it to the `.sql` and `.html.ts` and not pass all the individual items along. However, this is a judgement call in this case and for three numbers I'm also fine passing them individually.

mwest1066 (approved):
This was a very nitpicky and bike-shedding review, sorry! I'm sure you'll push back against any suggestions that are over the top so I just nitpicked everything I could see 😅

apps/prairielearn/src/ee/institution/admin/course.html.ts
nwalters512: Yeah, maybe a good stopgap would be to make a TypeScript component wrapper around the existing EJS?

apps/prairielearn/src/lib/db-types.ts
```diff
@@ -0,0 +1,71 @@
...
+  options: z.any(),
+  path: z.string().nullable(),
+  repository: z.string().nullable(),
+  short_name: z.string().nullable(),
+  sync_errors: z.string().nullable(),
+  sync_job_sequence_id: IdSchema.nullable(),
+  sync_warnings: z.string(),
+  title: z.string().nullable(),
+});
+export type Course = z.infer<typeof CourseSchema>;
```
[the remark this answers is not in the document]
nwalters512: I found that it even does a decent job if the schema is open in an adjacent tab!

apps/prairielearn/src/pages/administratorInstitutions/administratorInstitutions.html.ts
```diff
@@ -0,0 +1,92 @@
...
+import { isEnterprise } from '../../lib/license';
+
+export const InstitutionSchema = z.object({
+  id: z.string(),
+  short_name: z.string(),
+  long_name: z.string(),
+  display_timezone: z.string(),
+  uid_regexp: z.string().nullable(),
+  authn_providers: z.array(z.string()),
+});
```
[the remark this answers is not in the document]
nwalters512: Good call, this predated my introduction of `InstitutionSchema` in this PR! I'll tweak it.

apps/prairielearn/src/pages/administratorInstitutions/administratorInstitutions.html.ts
nwalters512: This is courtesy of a mindless copy/paste from EJS, I've definitely been avoiding abbreviations otherwise. I'll fix this.

apps/prairielearn/src/ee/institution/utils.ts
nwalters512: Yep, good call.

apps/prairielearn/src/ee/institution/admin/course.ts
nwalters512: I'll do that once we implement institution admins. For now, this entire set of pages is gated on the user being a global admin, so auth within pages isn't strictly necessary (though I've been including `institution_id` in queries for future-proofing and safety to make sure that entities do in fact belong to the institution being viewed).

apps/prairielearn/src/pages/administratorInstitutions/administratorInstitutions.html.ts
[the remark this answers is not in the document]
nwalters512: Done in d935d41.

apps/prairielearn/src/pages/administratorInstitutions/administratorInstitutions.html.ts
```diff
@@ -0,0 +1,92 @@
...
+type Institution = z.infer<typeof InstitutionSchema>;
```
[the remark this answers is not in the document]
nwalters512: Done in d935d41.

apps/prairielearn/src/pages/administratorInstitutions/administratorInstitutions.sql
```diff
@@ -0,0 +1,17 @@
...
+SELECT
+  i.*,
+  coalesce(
+    jsonb_agg(
+      ap.name
+      ORDER BY
+        ap.name
+    ),
+    '[]'::jsonb
+  ) AS authn_providers
```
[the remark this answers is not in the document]
nwalters512: Done in d935d41.

apps/prairielearn/src/pages/administratorInstitutions/administratorInstitutions.html.ts
nwalters512: Done in d935d41.

apps/prairielearn/src/ee/institution/admin/general.ts
nwalters512: Done in 508d20f.

apps/prairielearn/src/ee/institution/admin/course.html.ts
nwalters512: Done in 508d20f.

apps/prairielearn/src/ee/institution/utils.ts
nwalters512: Done in 508d20f.

apps/prairielearn/src/ee/institution/utils.ts
nwalters512: Done in 508d20f.

apps/prairielearn/src/ee/institution/admin/course.sql
nwalters512: > For consistency I always convert to jsonb so that we get the same data back from the DB irrespective of whether it's a single row or a deeply-nested structure.

Can you elaborate on this? I'm not sure I understand what you're getting at. With Zod, the data should always be the same?

apps/prairielearn/src/ee/auth/saml/router.ts
```diff
@@ -57,14 +58,18 @@ router.post(
     // Fetch this institution's attribute mappings.
     const institutionId = req.params.institution_id;
     const institutionSamlProvider = await getInstitutionSamlProvider(institutionId);
+    if (institutionSamlProvider === null) {
```
nwalters512, line 61:
```suggestion
    if (!institutionSamlProvider) {
```

apps/prairielearn/src/ee/institution/admin/course.html.ts
mwest1066: I'm totally fine with that!

apps/prairielearn/src/lib/db-types.ts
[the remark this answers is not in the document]
mwest1066: Sweet! ❤️

apps/prairielearn/src/ee/institution/admin/course.sql
mwest1066: I think I'm being overly cautious here.

I was initially worried about two different type conversion pathways:
1. Using `to_jsonb(c)` converts to JSONB using postgres' internal conversion rules, and is then parsed first by `node-postures` using `JSON.parse()` and then by `zod`.
2. Using `*` sends the data back to the client in postgres' wire format (preserving more type information) and is then parsed first by `node-postgres` using it's [type conversion rules](https://node-postgres.com/features/types) and then by `zod`.

In principle these two conversion paths could end up doing something different. For example, if we have an ID bigger than 2^52, like 12345678901234567 (approx 1.2e16), then pathway (1) will convert it to the JSON number `12345678901234567 ` in postgres and then `JSON.stringify()` will convert to the JS number `12345678901234568` (note the final digit is now `8`) and finally `IdSchema` will coerce it back to the JS string `"12345678901234568"` (again with an `8` at the end). In contrast, pathway (2) will convert to the string `"12345678901234567"` in postgres and then preserve this string exactly on the way back to JS.

However, reading more about the different types, I think BIGINTs might be the only type that suffers from this problem, and we've already decided not to worry about BIGINTs above 2^52.

apps/prairielearn/src/ee/institution/admin/course.ts
mwest1066: Yep, makes sense.

PR #7785, opened 2023-05-26
Template courses

mfsilva22:
Creating course templates as OER. 
As of now, I created some element extensions, and I was able to get them to work when including them in my own course repository. But they are not working when I add them to a template course. For example, the question
`/Users/marimac/GitRepo/PrairieLearn/templateCourses/physicsEM/questions/circuitDraw` uses element extensions, but the element is not rendered.


mfsilva22:
Closing this PR, since it was split into two others: #7973 and #7975

PR #7786, opened 2023-05-26
Speed up watch restarts

nwalters512:
This PR makes two key changes to shave ~3 seconds off application booting in development:

- Uses `swc` to transpile TypeScript instead of `tsc`. See https://typestrong.org/ts-node/docs/swc/. I can't wait for https://tc39.es/proposal-type-annotations/ to become accepted and for Node to implement it so we don't have to pay any tax at all for using TypeScript!
- Don't wait for the code callers to boot up in dev mode, where we don't particularly care if the code callers are ready by the time the application starts accepting requests.

After these changes, on my machine, it takes ~3.4 seconds to completely boot up the server in watch mode, down from ~6.2.

### DIFF
--- a/apps/prairielearn/src/lib/code-caller/index.js
+++ b/apps/prairielearn/src/lib/code-caller/index.js
@@ -144,11 +144,17 @@ module.exports = {
     // Ensure that the workers are ready; this will ensure that we're ready to
     // execute code as soon as we start processing requests.
     //
+    // We skip this if we're running in dev mode, as we want to prioritize the
+    // speed of starting up the server to ensure running in watch mode is as
+    // fast as possible.
+    //
     // Note: if resource creation fails for any reason, this will never resolve
     // or reject. This is unfortunate, but we'll still log and report the error
     // above, so it won't fail totally silently. If we fail to create workers,
     // we have a bigger problem.
-    await pool.ready();
+    if (!config.devMode) {
+      await pool.ready();
+    }
   },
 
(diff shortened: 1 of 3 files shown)


apps/prairielearn/src/lib/code-caller/index.js
```diff
@@ -144,11 +144,17 @@ module.exports = {
...
+    // fast as possible.
+    //
     // Note: if resource creation fails for any reason, this will never resolve
     // or reject. This is unfortunate, but we'll still log and report the error
     // above, so it won't fail totally silently. If we fail to create workers,
     // we have a bigger problem.
-    await pool.ready();
+    if (!config.devMode) {
+      await pool.ready();
+    }
```
mwest1066, line 157: Will this do the right thing if we try and reload a question page (for example) before the workers are spun up? That is, will it wait for a worker, or will it error?

mwest1066 (approved):

apps/prairielearn/src/lib/code-caller/index.js
nwalters512: It will indeed do the right thing: the underlying code caller resources managed by `generic-pool` aren't considered available until they have a child process. Even if we didn't implement that specifically, the code caller state machines handle this correctly.

mwest1066: Excellent, thanks for confirming! 🙏

PR #7787, opened 2023-05-26
Implement enrollment limits

nwalters512:
This PR implements enrollment limits that site admins can set. There are three types of enrollment limits in place:

- `institutions.course_instance_enrollment_limit`: set on the institution and apply to all course instances at the institution.
- `institutions.yearly_enrollment_limit`: places a cap on the number of enrollments that can occur within a sliding 365-day window.
- `course_instances.enrollment_limit`: set on individual course instances. In general, this should be left null in favor of the institution-level enrollment limits, but it can be set in case we need to fine-tune limits for particular course instances.

## TODO

- [ ] @mwest1066 and I had discussed allowing some number of course staff (5?) to enroll in their own course instances without being counted towards enrollment limits.


apps/prairielearn/src/ee/institution/admin/courseInstance.sql
```diff
@@ -10,3 +10,17 @@ WHERE
...
+  enrollment_limit = $enrollment_limit
+FROM
+  pl_courses AS c
+  JOIN institutions AS i ON (i.id = c.institution_id)
+WHERE
+  ci.id = $course_instance_id
+  AND ci.course_id = c.id
+  AND i.id = $institution_id
+RETURNING
+  *;
```
nwalters512, line 26:
```suggestion
  AND i.id = $institution_id;
```

apps/prairielearn/src/migrations/20230523200508_enrollment_limits.sql
```diff
@@ -0,0 +1,17 @@
...
+SET DEFAULT CURRENT_TIMESTAMP;
+
+ALTER TABLE institutions
+ADD COLUMN IF NOT EXISTS yearly_enrollment_limit INTEGER;
+
+ALTER TABLE institutions
+ADD COLUMN IF NOT EXISTS course_instance_enrollment_limit INTEGER;
+
+ALTER TABLE course_instances
+ADD COLUMN IF NOT EXISTS enrollment_limit INTEGER;
```
mwest1066, line 17: Can we make all these limits NOT NULL? I'm ok with a high DEFAULT but I want to move towards having limits on everything.

apps/prairielearn/src/pages/enroll/enroll.html.ts
```diff
@@ -280,3 +287,33 @@ export function EnrollLtiMessage({ ltiInfo, resLocals }: EnrollLtiMessageProps)
...
+        ${renderEjs(__filename, "<%- include('../partials/navbar'); %>", {
+          ...resLocals,
+          navPage: 'enroll',
+        })}
+        <main id="content" class="container">
+          <div class="card mb-4">
+            <div class="card-header bg-danger text-white">Enrollment limit exceeded</div>
+            <div class="card-body">
+              This course has reached its enrollment limit. Please contact the course staff for more
+              information.
```
mwest1066, line 312: Do we want to give their emails? Or at least UIDs until we have emails.

apps/prairielearn/src/pages/enroll/enroll.sql
```diff
@@ -44,23 +44,93 @@ ORDER BY
...
+      ) AS course_instance_enrollment_limit
+    FROM
+      institutions AS i
+      JOIN pl_courses AS c ON (c.institution_id = i.id)
+      JOIN course_instances AS ci ON (ci.course_id = c.id)
+    WHERE
+      ci.id = $course_instance_id
+      -- Lock institution to prevent concurrent enrollments that would exceed
+      -- the enrollment limit.
+    FOR UPDATE OF
```
mwest1066, line 66: NO KEY UPDATE to prevent blocking SELECTs?

apps/prairielearn/src/pages/enroll/enroll.sql
```diff
@@ -44,23 +44,93 @@ ORDER BY
...
-  u.user_id = $user_id
-  AND check_course_instance_access (
-    $course_instance_id,
-    u.uid,
-    u.institution_id,
-    $req_date
-  )
-RETURNING
-  e.id;
+  limits_exceeded;
```
mwest1066, line 133: Fwiw I've been moving towards doing this kind of operation in TS, wrapping it in a `runTransactionAsync`, starting with a `lockCourseInstance()`, then doing a query to get the various enrollment counts, checking limits in TS, then doing the actual INSERT. I've found this to be easier to read debug. This style is facilitated by using `selectAndAuthz...` functions (sorry to keep harping on about these, but they really are nice 😀) and passing whole objects around rather than row IDs.

mwest1066 (approved):

apps/prairielearn/src/migrations/20230523200508_enrollment_limits.sql
nwalters512: For `course_instances.enrollment_limit` specifically, no, I do not want to make this `NOT NULL` - the default should be that this is inherited from the institution (or in the future, the division), which is represented by `NULL`. I _do_ think it makes sense to place `NOT NULL` constraints on the top-level limits on `institutions`. I'll do that. I checked prod data, and I think 10k is a comfortable default limit; that's several times larger than the largest course instance we've ever had.

Tangential but related thought: should limits only be enforced for enterprise servers? The UI to view/edit these limits is only available for enterprise servers, and this seems like an enterprise feature anyways, so this might make sense?

apps/prairielearn/src/pages/enroll/enroll.html.ts
nwalters512: IMO, no. Given that this is a student-initiated action, they should already know who is running their course? And a few more reasons:

- In practice, we see people adding "random" (trusted, but not actually on course staff) people as owners. We don't want them getting random emails.
- In a similar vein, this is different from the recent changes in PT that we're probably both thinking about, where identity is deliberately revealed by the person making invites.

apps/prairielearn/src/pages/enroll/enroll.sql
nwalters512: Yes!! Good call. I need to get in the habit of doing that.

apps/prairielearn/src/pages/enroll/enroll.sql
nwalters512: That's fair! I can move to doing that. I'm still in "minimize DB roundtrips" mindset 😅

apps/prairielearn/src/migrations/20230523200508_enrollment_limits.sql
mwest1066: Good point about `course_instances.enrollment_limit`. I agree that 10k seems like a fine default and I agree that this limit should only be enforced for EE.

apps/prairielearn/src/pages/enroll/enroll.html.ts
mwest1066: Yeah, I agree that in practice students will know where to complain 😄

apps/prairielearn/src/pages/enroll/enroll.sql
mwest1066: Me too! But I've been getting over it and the resulting code has been much easier to deal with (plus typechecking!).

mwest1066 (approved):

PR #7788, opened 2023-05-26
style: Use inlined variable definition in more places

eliotwrobson:
Changes multiple questions in the example and test courses to use inlined variable definition with the external python grader. Hasn't been extensively tested yet.


exampleCourse/questions/demo/autograder/python/plots/question.html
```diff
@@ -11,10 +11,12 @@
     </p>
 
     <p>The setup code gives the following variables:</p>
-    <p><pl-external-grader-variables params-name="names_for_user"></pl-external-grader-variables></p>
+    <p><pl-external-grader-variables params-name="names_for_user" empty="true"></pl-external-grader-variables></p>
```
nwalters512, line 14: Nit: can this be removed entirely?

nwalters512 (approved):
LGTM, one nit though.

exampleCourse/questions/demo/autograder/python/plots/question.html
eliotwrobson: Will double check, but I don't think so. The Python grader requires certain parameters to be declared, even if they're just an empty list.

nwalters512: Ah, sorry, should have been more specific. We can probably remove the `<p>` specifically, since I think that would render... nothing? an empty table? regardless, probably nothing useful. In all the other examples, the `<pl-external-grader-variables params-name="names_for_user" empty="true">` doesn't have any student-visible labels or anything.

eliotwrobson: Removed here https://github.com/PrairieLearn/PrairieLearn/pull/7788/commits/db00a1677b7287220f6c6095f4bfc9b88d940fd3

PR #7789, opened 2023-05-26
Configure ts-node to use SWC in tsconfig.json

nwalters512:
This means that our Mocha tests will use SWC as well, which should result in a slight performance improvement.


mwest1066 (approved):

apps/prairielearn/src/lib/file-store.js
```diff
@@ -117,7 +117,7 @@ module.exports.delete = async (file_id, authn_user_id) => {
  */
 module.exports.getStream = async (file_id) => {
   debug(`getStream(): file_id=${file_id}`);
-  const file = await this.get(file_id, 'stream');
+  const file = await module.exports.get(file_id, 'stream');
```
nwalters512, line 120: Using `this` in place of `module.exports` feels like a perpetual footgun: https://github.com/swc-project/swc/issues/7476. The sooner we switch to ESM, the better!

PR #7790, opened 2023-05-27
Fix use of percentage-based input in manual grading

### LINKED ISSUE #7791: Manual grading issues for questions without points
As reported on Slack, the manual grading page is unable to submit a new grade if the question has no points. This is likely related to the handling of percentage, with division of zero involved in that case.

jonatanschroeder:
Based on issue mentioned on Slack. Resolves two issues:

* Recent rubrics update broke the way manual input of points (particularly percentage) is handled, since the function relies on `this`, for which the function is no longer bound to. For safety and clarity this was replaced with an explicit argument.
* Resolves #7791, i.e., properly handles manual grading cases when the question is worth zero in total (i.e., no manual or auto points).

### DIFF
--- a/apps/prairielearn/assets/scripts/instructorAssessmentManualGradingInstanceQuestion.js
+++ b/apps/prairielearn/assets/scripts/instructorAssessmentManualGradingInstanceQuestion.js
@@ -123,7 +123,7 @@ function resetInstructorGradingPanel() {
     .forEach((item) => item.addEventListener('change', computePointsFromRubric));
   document
     .querySelectorAll('.js-grading-score-input')
-    .forEach((input) => input.addEventListener('input', computePointsFromRubric));
+    .forEach((input) => input.addEventListener('input', () => computePointsFromRubric(input)));
 
   document.querySelectorAll('.js-adjust-points-enable').forEach((link) =>
     link.addEventListener('click', function () {
@@ -390,7 +390,7 @@ function updatePointsView(sourceInput) {
   });
 }
 
-function computePointsFromRubric() {
+function computePointsFromRubric(sourceInput = null) {
   document.querySelectorAll('form[name=manual-grading-form]').forEach((form) => {
     if (form.dataset.rubricActive === 'true') {
       const manualInput = form.querySelector('.js-manual-score-value-input-points');
@@ -414,7 +414,7 @@ function computePointsFromRubric() {
       manualInput.value = manualPoints;
     }
   });
-  updatePointsView(this);
+  updatePointsView(sourceInput);
 }
 
 function enableRubricItemLongTextField(event) {
--- a/apps/prairielearn/assets/scripts/instructorAssessmentManualGradingInstanceQuestion.js
+++ b/apps/prairielearn/assets/scripts/instructorAssessmentManualGradingInstanceQuestion.js
@@ -76,7 +76,7 @@ function resetInstructorGradingPanel() {
         element.style.display = use_percentage ? '' : 'none';
       });
       window.localStorage.manual_grading_score_use = use_percentage ? 'percentage' : 'points';
-      updatePointsView();
+      updatePointsView(null);
     });
     toggle.checked = window.localStorage.manual_grading_score_use === 'percentage';
     toggle.dispatchEvent(new Event('change'));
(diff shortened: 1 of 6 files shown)


jonatanschroeder:
Converting to draft to allow this to also handle #7791.

jonatanschroeder:
This is now ready for review. The change also gets rid of percentage-based grading for questions with no points, since in this case percentage makes no sense. The use of local storage to handle percentage or points is ignored in this kind of question, since the processing is done in an event handler for the selector, and the selector does not exist.

apps/prairielearn/src/pages/instructorAssessmentManualGrading/instanceQuestion/gradingPointsSection.ejs
```diff
@@ -3,21 +3,25 @@
...
           <span class="js-<%= type %>-score-value-info">
-            <span class="js-value-<%= type %>-points"><%= Math.round(points * 100) / 100 %></span> / <%= max_points %>
+            <span class="js-value-<%= type %>-points"></span> / <%= max_points %>
           </span>
         </span>
+        <% if (assessment_question.max_points) { %>
         <span class="js-manual-grading-percentage">
           <span class="js-<%= type %>-score-value-info">
-            <span class="js-value-<%= type %>-percentage"><%= Math.round(points * 10000 / (max_points || assessment_question.max_points)) / 100 %></span>%
+            <span class="js-value-<%= type %>-percentage"></span>%
```
jonatanschroeder, line 21: The initialization of these values is unnecessary, since JS already updates them when the page is loaded. Keeping it blank for simplicity and to avoid inconsistencies between the formatting done here and in JS.

apps/prairielearn/assets/scripts/instructorAssessmentManualGradingInstanceQuestion.js
```diff
@@ -414,7 +414,7 @@ function computePointsFromRubric() {
       manualInput.value = manualPoints;
     }
   });
-  updatePointsView();
+  updatePointsView(this);
```
mwest1066, line 417: I personally find the use of `this` to be hard to follow in code that isn't a member function of a `class`. I like the fact that arrow functions have dropped this-binding, making it harder to write code that uses this-binding via `call`/`apply`/`bind`.

Could we instead pass the `item`/`input` (hopefully with a better name) as an argument to `computePointsFromRubric()` (using an arrow function in the `addEventListener()` calls that set up the handlers)? Then we can explicitly pass this `item` down to `updatePointsView()`.

apps/prairielearn/src/pages/instructorAssessmentManualGrading/instanceQuestion/gradingPointsSection.ejs
mwest1066: Sounds good.

Unrelated to the current PR: I find this pattern of dynamically constructing class names (or attribute names, etc) to make the code harder to follow because we can't grep for names. I generally prefer to either have all the code written out multiple times, once for each possible class name, or to store the entire class name as a variable with the full class names in the code:
```
if (type == 'auto') {
    valuePercentageClass = 'js-value-auto-percentage';
} else {
    valuePercentageClass = 'js-value-manual-percentage';
}
```

mwest1066 (approved):

apps/prairielearn/assets/scripts/instructorAssessmentManualGradingInstanceQuestion.js
jonatanschroeder: Done in d9df5cbf2.

apps/prairielearn/src/pages/instructorAssessmentManualGrading/instanceQuestion/gradingPointsSection.ejs
jonatanschroeder: > Unrelated to the current PR: I find this pattern of dynamically constructing class names (or attribute names, etc) to make the code harder to follow because we can't grep for names. I generally prefer to either have all the code written out multiple times, once for each possible class name, or to store the entire class name as a variable with the full class names in the code:

Noted. Since this is not directly related to this PR I'll keep this as is and keep it in mind for future dev.

PR #6269, opened 2022-09-05
feat: Add visual indicator for paired distractor blocks in pl-order-blocks

SethPoulsen:
image

PR #7792, opened 2023-05-29
pl-order-blocks - Adding Visual Distractor Grouping

CoffeePoweredComputers:
This feature adds the `distractor-for` tag to `<pl-answer>` elements for the `ranking` and `dag` grading modes. This allows a distractor (i.e., an incorrect block) to be visually associated with it's correct alternative based on the correct alternative's `tag` value. An example is provided below.

image

```html
<pl-order-blocks
        answers-name="answers"
        indentation=true
        grading-method="ranking"
        format="code"
        code-language="python">
    <pl-answer correct="true" indent="0" ranking="1">def remove_element_even_pos(lst, i):</pl-answer>  
    <pl-answer correct="true" indent="1" ranking="2">if i % 2 == 0:</pl-answer>  
    <pl-answer tag="a" correct="true" indent="2" ranking="3">lst.pop(i)</pl-answer>  
    <pl-answer correct="false" distractor-for="a">lst.pop[i]</pl-answer>  
    <pl-answer correct="true" indent="1" ranking="4">else:</pl-answer>  
    <pl-answer correct="true" indent="2" ranking="5">print("Index is not even")</pl-answer>  
</pl-order-blocks
```
This feature was originally created by @SethPoulsen for his research on proof blocks.


eliotwrobson:
@CoffeePoweredComputers Haven't taken a detailed look yet, but is this related to your changes in your other PR here?: https://github.com/PrairieLearn/PrairieLearn/pull/7774

If these are related features, then it would be best to combine these PRs (easier to review + simplify common changes). It looks like this is a remake of https://github.com/PrairieLearn/PrairieLearn/pull/6269, in which case that PR should be closed in favor of a combined one.

SethPoulsen:
This PR supersedes #6269, I will close that one. 

This PR has little overlap with #7774. That one changes the display of the Feedback/output, whereas this one is concerned with the input user interface. 

I would prefer them to be handled seperately.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -338,7 +345,32 @@ def render(element_html, data):
         )
 
         mcq_options = data["params"][answer_name]
-        mcq_options = filter_multiple_from_array(mcq_options, ["inner_html", "uuid"])
+        mcq_options = filter_multiple_from_array(mcq_options, ["inner_html", "uuid", "tag", "distractor-for"])
+
+        # visual pairing
+        correct_tags = set(block['tag'] for block in mcq_options + student_previous_submission if block['tag'] is not None)
```
eliotwrobson, line 351: You use `mcq_options + student_previous_submission` in multiple places, would be best to set  this as a variable and re-use.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -338,7 +345,32 @@ def render(element_html, data):
...
+        incorrect_tags = set(block['distractor-for'] for block in mcq_options + student_previous_submission if block.get('distractor-for', None)) 
+
+        if not incorrect_tags.issubset(correct_tags):
+            raise Exception("The following distractor-for tags do not have matching correct answer tags: " + str(incorrect_tags - correct_tags))
```
eliotwrobson, line 355:
```suggestion
            raise ValueError(f"The following distractor-for tags do not have matching correct answer tags: {incorrect_tags - correct_tags})
```

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -338,7 +345,32 @@ def render(element_html, data):
...
+
+        for block in student_previous_submission + mcq_options:
+            if block.get('distractor-for') is not None:
```
eliotwrobson, line 358: Will `"distractor-for"` always be in `block`? If so, you can just use `block["distractor-for"]`, the `.get()` function is only needed if you want to return a default value if the desired key is not in the given dictionary.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -338,7 +345,32 @@ def render(element_html, data):
...
+                continue
+
+            distractors = [
+                block2 
+                for block2 in student_previous_submission + mcq_options 
+                if (block['tag'] == block2.get('distractor-for', None)) and (block['tag'] is not None)
```
eliotwrobson, line 364: Nit: the `.get()` function uses `None` as a default when the key is not present.
```suggestion
                if (block['tag'] == block2.get('distractor-for')) and (block['tag'] is not None)
```

eliotwrobson (commented):
Minor items but generally seems fine.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: done

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: addressed

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
SethPoulsen: It looks like `student_previous_submission` is actually an empty list some of the time and so its not even needed. I'm going to refactor a little bit.

SethPoulsen:
After my tweaks, final UI looks like:
image

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.css
```diff
@@ -72,3 +75,19 @@ div.pl-order-blocks-container {
...
+  border-radius: 10px;
+  border: 0px !important;
+  list-style: none;
+  margin-left: -10px;
+  margin-right: -10px;
+  padding: 4px 10px 4px 10px;
+  margin-top: 5px;
+  margin-bottom: 5px;
+  overflow: hidden;
+}
```
nwalters512, line 89: I'll overlook the pixel values since the rest of the element already uses them, but we typically want to use `rem` values instead.

docs/elements.md
```diff
@@ -454,13 +454,14 @@ Element to arrange given blocks of code or text that are displayed initially in
...
+| `tag`            | string             | ""      | Optional attribute when `grading-method="dag"`. Used to identify the block when declaring which other blocks depend on it.                                                                                                                                                                                                                                                |
+| `distractor-for` | string             | ""      | Optional attribute on blocks where `correct=false`. Used to visually group a distractor block with a correct block that it is similar to.                                                                                                                                                                                                                                 |
```
nwalters512, line 464:
```suggestion
| `tag`            | string             | ""      | Optional attribute when `grading-method="dag"`. Used to identify the block when declaring which other blocks depend on it.                                                                                                                                                                                                                                                |
| `distractor-for` | string             | ""      | Optional attribute on blocks where `correct="false"`. Used to visually group a distractor block with a correct block that it is similar to.                                                                                                                                                                                                                                 |
```

These docs don't actually explain what value `distractor-for` should actually have. Looking at the example question, I can see that it should be another element's tag, but that's not obvious here. I'd recommend updating the `tag` description as well, now that it's not only used for `grading-method="dag"`.

exampleCourse/questions/element/orderBlocks/question.html
```diff
@@ -75,11 +76,12 @@
...
-  <pl-answer correct="true" ranking="3" indent="1">sum += first</pl-answer>
-  <pl-answer correct="true" ranking="3" indent="1">sum += second</pl-answer>
+  <pl-answer correct="true" ranking="3" indent="1" tag="addfirst">sum += first</pl-answer>
+  <pl-answer correct="true" ranking="3" indent="1" tag="addsecond">sum += second</pl-answer>
   <pl-answer correct="true" ranking="4" indent="1">return sum</pl-answer>
-  <pl-answer correct="false">sum + first</pl-answer>
-  <pl-answer correct="false">sum + second</pl-answer>
+  <pl-answer correct="false" distractor-for="addfirst">sum + first</pl-answer>
+  <pl-answer correct="false" distractor-for="addfirst">sum = first</pl-answer>
+  <pl-answer correct="false" distractor-for="addsecond">sum + second</pl-answer>
```
nwalters512, line 84: It seems like `correct="true" distractor-for="..."` would be an invalid configuration, but AFAICT we don't have validation for that. Did you consider validating that during `prepare()`?

eliotwrobson (commented):
Didn't review everything but Python changes seem fine.

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.css
SethPoulsen: noted

exampleCourse/questions/element/orderBlocks/question.html
SethPoulsen: Nice catch! I will add validation for that.

docs/elements.md
SethPoulsen: Sounds good, fixed.

nwalters512 (approved):

PR #7793, opened 2023-05-29
Trigger response log on stream close

jonatanschroeder:
If an HTTP client (browser, test, API, etc.) reads from a URL but closes the stream before it's completely sent, the log response is not being triggered. This is because the `finish` event only triggers when the stream is completely sent. This often happens for responses that cause a redirection (e.g., most POST responses), since in that case the client only needs the header to identify the next location. This PR changes the logging to happen on the `close` event, which is triggered even when the stream is closed.

I have no means to test how this affects production environment and sentry logs.

### DIFF
--- a/apps/prairielearn/src/middlewares/logResponse.js
+++ b/apps/prairielearn/src/middlewares/logResponse.js
@@ -25,26 +25,20 @@ module.exports = function (req, res, next) {
         params: req.params,
         body: req.body,
         finished: res.writableFinished,
-        authn_user_id: res.locals && res.locals.authn_user ? res.locals.authn_user.user_id : null,
-        authn_user_uid: res.locals && res.locals.authn_user ? res.locals.authn_user.uid : null,
-        user_id: res.locals && res.locals.user ? res.locals.user.user_id : null,
-        user_uid: res.locals && res.locals.user ? res.locals.user.uid : null,
-        course_id: res.locals && res.locals.course ? res.locals.course.id : null,
-        course_short_name: res.locals && res.locals.course ? res.locals.course.short_name : null,
-        course_instance_id:
-          res.locals && res.locals.course_instance ? res.locals.course_instance.id : null,
-        course_instance_short_name:
-          res.locals && res.locals.course_instance ? res.locals.course_instance.short_name : null,
-        assessment_id: res.locals && res.locals.assessment ? res.locals.assessment.id : null,
-        assessment_directory:
-          res.locals && res.locals.assessment ? res.locals.assessment.tid : null,
-        assessment_instance_id:
-          res.locals && res.locals.assessment_instance ? res.locals.assessment_instance.id : null,
-        question_id: res.locals && res.locals.question ? res.locals.question.id : null,
-        question_directory:
-          res.locals && res.locals.question ? res.locals.question.directory : null,
-        instance_question_id:
-          res.locals && res.locals.instance_question ? res.locals.instance_question.id : null,
+        authn_user_id: res.locals?.authn_user?.user_id ?? null,
+        authn_user_uid: res.locals?.authn_user?.uid ?? null,
+        user_id: res.locals?.user?.user_id ?? null,
+        user_uid: res.locals?.user?.uid ?? null,
+        course_id: res.locals?.course?.id ?? null,
+        course_short_name: res.locals?.course?.short_name ?? null,
+        course_instance_id: res.locals?.course_instance?.id ?? null,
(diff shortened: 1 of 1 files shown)


mwest1066:
This looks sensible to me but I'm a bit worried about the fact that the [first](https://www.moesif.com/blog/technical/logging/How-we-built-a-Nodejs-Middleware-to-Log-HTTP-API-Requests-and-Responses/#handling-response-errors-and-client-aborts) [two](https://brightinventions.pl/blog/http-request-logging-in-node/) blog posts I found on this do something more similar to our existing code -- log full details on `finish` and only a "client aborted" message on `close`.

An alternative strategy to this PR would be to change this line: https://github.com/PrairieLearn/PrairieLearn/blob/5e4c719c596b49d214e4cfcc1d37fd8f4c1dacaa/apps/prairielearn/src/middlewares/logResponse.js#L56 to be a `verbose()` and change the message to something like `'request aborted by client'`.

However, as far as I can tell we will still do our full request processing even for a `close` event, so we will have all the data to log just like we currently do in the `finish` handler. I can't see why we shouldn't log all this data on `close`. @nwalters512 thoughts?

apps/prairielearn/src/middlewares/logResponse.js
```diff
@@ -24,39 +24,23 @@ module.exports = function (req, res, next) {
...
+        course_instance_id: res.locals?.course_instance?.id ?? null,
+        course_instance_short_name: res.locals?.course_instance?.short_name ?? null,
+        assessment_id: res.locals?.assessment?.id ?? null,
+        assessment_directory: res.locals?.assessment?.tid ?? null,
+        assessment_instance_id: res.locals?.assessment_instance?.id ?? null,
+        question_id: res.locals?.question?.id ?? null,
+        question_directory: res.locals?.question?.directory ?? null,
+        instance_question_id: res.locals?.instance_question?.id ?? null,
       };
       logger.verbose('response', access);
```
mwest1066, line 43: I'm wondering if we should additionally add:
```
if (!res.writableFinished) {
  logger.verbose('request aborted by client', {response_id: res.locals.response_id, timestamp: new Date().toISOString()});
}
```
This isn't adding any new information, but in the past it's been helpful when scanning the logs to be able to see when there are a lot of the old `response was not logged` messages, which typically indicated that something was going wrong somewhere. With our improved observability these days this might no longer matter, however. @nwalters512 ?

nwalters512: I see that Jonatan is already logging this as a `finished` property on the log info above, but I do agree for the case where we're manually tailing logs, it's helpful to have a specific and recognizable prefix to look for (as opposed to looking for a specific key/value pair in the entire response context).

Re: observability, it's probably worth adding OTel metrics for requests/responses - think `http.requests` to track the total number of requests and `http.requests.aborted` to track the number of, well, aborted requests. That's outside the scope of this PR, of course.

jonatanschroeder: Done in 1272708f1.

apps/prairielearn/src/middlewares/logResponse.js
```diff
@@ -24,37 +24,29 @@ module.exports = function (req, res, next) {
...
-    // install a handler that will always be called, so we can
-    // check whether we correctly logged the response
-    res.on('close', () => {
-      if (!res.locals.response_logged) {
-        logger.error('response was not logged', {
+      // Print additional message in this case to simplify grepping logs for this scenario
+      if (!res.writableFinished) {
+        logger.verbose('request aborted by client', {
           response_id: res.locals.response_id,
+          timestamp: new Date().toISOString(),
```
nwalters512, line 49: I was about to comment on this before seeing that @mwest1066 specifically requested this in his comment (it was hidden behind a horizontal scroll for me). Matt, is there a specific reason to add this? Our production logs always include timestamps by default.

mwest1066: Excellent point that we are adding the timestamp by default with https://github.com/PrairieLearn/PrairieLearn/blob/765dc616efabef65340d4e55675c6d00b8970654/packages/logger/src/index.ts#L23
In this case we should definitely not bother to include it again here. We could also remove the existing unnecessary timestamp at https://github.com/PrairieLearn/PrairieLearn/blob/765dc616efabef65340d4e55675c6d00b8970654/apps/prairielearn/src/middlewares/logResponse.js#L20

nwalters512 (approved):

PR #7799, opened 2023-05-30
Fix Python grader docs

nwalters512:
This was just straight up incorrect. Thankfully, all our other examples use the correct class name.


nwalters512:
> Am assuming these are the only references to the incorrect names in this document.

Yep - the only reference anywhere in the codebase, in fact!

PR #7801, opened 2023-05-30
Deprecate feature_grants.type column

nwalters512:
This column will no longer be useful now that we're clarified how feature flags are meant to be used. This PR preps it for removal:

- `feature_grants.type` is allowed to be `NULL` so that we can insert new rows without it
- All indexes have been recreated to not include the `type` column


apps/prairielearn/src/pages/administratorFeatures/administratorFeatures.html.ts
```diff
@@ -1,25 +1,32 @@
...
+  course_title: z.string().nullable(),
+  course_short_name: z.string().nullable(),
+  course_instance_id: z.string().nullable(),
+  course_instance_short_name: z.string().nullable(),
+  course_instance_long_name: z.string().nullable(),
+  user_id: z.string().nullable(),
+  user_uid: z.string().nullable(),
+  user_name: z.string().nullable(),
+});
+type FeatureGrantRow = z.infer<typeof FeatureGrantRowSchema>;
```
mwest1066, line 21: Good to standardize on putting these `...Row` types in the `.html.ts` files 👍

mwest1066 (approved):

PR #7803, opened 2023-05-30
Upgrade to latest Redis library

nwalters512:
This PR makes a number of Redis-related changes:

- Upgrades to v4 of the `redis` library
- Replaces `socket.io-redis` with `@societ.io/redis-adapter`
- Drops `redis-lru` in favor of Redis's own ability to evict items when facing memory pressure (https://redis.io/docs/reference/eviction/)


apps/prairielearn/src/pages/authCallbackLti/authCallbackLti.js
```diff
@@ -71,7 +73,7 @@ router.post('/', function (req, res, next) {
       // https://oauth.net/core/1.0/#nonce
       var nonceReused = false;
       var nonceKey = 'authCallbackLti:' + parameters.oauth_timestamp + ':' + parameters.oauth_nonce;
-      cache.get(nonceKey, (err, val) => {
+      util.callbackify(cache.get)(nonceKey, (err, val) => {
```
nwalters512, line 76: Rather than putting callback garbage in the cache implementation itself, I'm pushing the callbackification out to this one place that it's actually used. I think we should do this in general as we convert from callbacks to promises.

apps/prairielearn/src/lib/cache.ts
```diff
@@ -1,138 +1,132 @@
...
+    }
+
+    case 'redis': {
+      let cursor = 0;
+      do {
+        const reply = await client.scan(cursor, { MATCH: '*', COUNT: 1000 });
+        cursor = reply.cursor;
+        console.log('deleting keys', reply.keys);
+        await client.del(reply.keys);
+      } while (cursor !== 0);
```
nwalters512, line 116: I stole this from Rails:

https://github.com/rails/rails/blob/e88857bbb9d4e1dd64555c34541301870de4a45b/activesupport/lib/active_support/cache/redis_cache_store.rb#L210-L229

apps/prairielearn/src/lib/cache.ts
```diff
@@ -1,138 +1,132 @@
...
+    case 'redis': {
+      const value = await client.get(key);
+      if (typeof value === 'string') {
+        return JSON.parse(value);
+      }
+      return null;
+    }
+
+    default: {
+      return null;
```
mwest1066, line 92: It's a super-minor thing, but I think the convention for built-in JS data types like `Map` is to [return `undefined`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/get#return_value) if the key is not present.

apps/prairielearn/src/lib/cache.ts
mwest1066: Cute!

apps/prairielearn/src/pages/authCallbackLti/authCallbackLti.js
mwest1066: I'm on board!

apps/prairielearn/src/server.js
```diff
@@ -2125,12 +2125,7 @@ if (require.main === module && config.startServer) {
       },
       async () => await codeCaller.init(),
       async () => await assets.init(),
-      (callback) => {
-        cache.init((err) => {
-          if (ERR(err, callback)) return;
-          callback(null);
-        });
-      },
+      async () => cache.init(),
```
mwest1066, line 2128: For stack traces?
```suggestion
      async () => await cache.init(),
```

mwest1066 (approved):

apps/prairielearn/src/lib/cache.ts
nwalters512: Nice! I'll go with that.

apps/prairielearn/src/server.js
nwalters512: Done in 3890eea.

PR #7807, opened 2023-06-01
Clarify docs for `advanceScorePerc`

mwest1066:
Cinda asked me about exactly how `advanceScorePerc` worked and I realized that the docs didn't quite clarify the unlocking behavior for future questions. I had to check the code to make sure I understood whether a question with `advanceScorePerc` only blocks the _next_ question or _all subsequent_ questions (it's the latter). This PR attempts to make the docs clearer in this respect.

Cinda also misinterpreted the behavior of `advanceScorePerc` at the `zone` level and whether it acts as a block between zones or as a default for questions within the zone (it's the latter). This PR also attempts to make this clearer.


docs/assessment.md
```diff
@@ -226,42 +226,60 @@ Students are able to see their groupmates' UIDs, which can become a point of con
...
 
-If a student uses all of their attempts on a question and cannot submit any more attempts, the next question will automatically unlock, no matter what score they earned on the previous question. This is to prevent students from getting permanently stuck on an assessment, unable to receive further credit.
+The relevant score for comparing to `advanceScorePerc` is the student's _highest submission score_ for the question, not their percentage score on the question overall. For example, suppose `q1` above has `"points": [10, 4, 2, 1]`. Then a student who makes a 50%-correct submission on their second attempt will unblock the question, even though they only score 2 points out of 10 on the question (50% of the 4-point second-chance value).
```
jonatanschroeder, line 253: For auto points or combined (auto+manual)?

docs/assessment.md
```diff
@@ -226,42 +226,60 @@ Students are able to see their groupmates' UIDs, which can become a point of con
...
+}
+```
+
+In the example above, `q2` and `q5` will have an `advanceScorePerc` of 100 because the zone-level attribute is used as a default.
+
+Note that an `advanceScorePerc` of 0 is equivalent to not having the attribute at all.
+
+If a student uses all of their attempts on a question and cannot submit any more attempts, that question will automatically unblock, no matter what score they earned on it. This is to prevent students from getting permanently stuck on an assessment, unable to receive further credit.
+
+### Warning about in-order questions and high-stakes exams
```
jonatanschroeder, line 280: What about exam's that randomize the order of questions?

docs/assessment.md
mwest1066: Good question. Here's the code: https://github.com/PrairieLearn/PrairieLearn/blob/5b65573e6974251dbbcef171f68013a33174d3ef/apps/prairielearn/src/sprocs/question_order.sql#L22
Does `highest_submission_score` include manual grading? Do we want it to for this use case?

docs/assessment.md
mwest1066: I added the following. Is this what you meant? https://github.com/PrairieLearn/PrairieLearn/blob/e207091508b9e347aea9aefd839db2a16a6a42d3/docs/assessment.md?plain=1#L278

jonatanschroeder: Yes, that clarifies what is implemented. That said, it is probably a good idea to warn instructors that if there is a specific order required for this setting (i.e., if they use this option to ensure they work in a particular order) they should follow https://prairielearn.readthedocs.io/en/latest/assessment/#changing-question-order-randomization.

docs/assessment.md
jonatanschroeder: I checked the source, and it seems it's not consistent. From what I could tell, it is the auto score before any manual grading is done, and total score after it. This should probably reviewed to make it more consistent.

docs/assessment.md
jonatanschroeder: Sorry, read wrong. It will be auto score only, and null if there are no auto points.

docs/assessment.md
jonatanschroeder: Sorry, I read it wrong. It's auto score, with 0 if the internal/external grader never ran.

eliotwrobson (approved):
Substantially clearer than what was here originally.

docs/assessment.md
jonatanschroeder: In essence what this means is that the advance score percentage applies to auto points (i.e., internal or external grader). So this option should not be used for questions that are manual-only. This is probably an important note to add here, or something to change in the implementation (in the sense of ignoring this option if the question has manual-only points).

nwalters512 (approved):

docs/assessment.md
mwest1066: I added a note here: https://github.com/PrairieLearn/PrairieLearn/blob/442ca8a6220377b4c38d6c48dd91e48142859398/docs/assessment.md?plain=1#L253

I agree that it would probably be good to enforce that purely-manually-graded questions don't have `advanceScorePerc` set, or issue a warning, or something.

docs/assessment.md
mwest1066: Added: https://github.com/PrairieLearn/PrairieLearn/blob/9c28fc98b808a98cc84da2079f71c4db7e0de9d4/docs/assessment.md?plain=1#L278

jonatanschroeder (approved):

PR #7897, opened 2023-06-01
Upgrade all JavaScript dependencies

nwalters512:
Excludes potentially-incompatible major version upgrades.


mwest1066 (approved):

packages/postgres-tools/src/describe.ts
```diff
@@ -57,11 +56,6 @@ interface DescribeOptions {
   ignoreEnums?: string[];
 }
 
-function parsePostgresArray(arr: string): string[] {
-  // @ts-expect-error -- Incorrect type definitions in `pg-types`.
-  return pgArray.create(arr, String).parse();
-}
-
```
nwalters512, line 64: This was removed from `pg-types` in https://github.com/brianc/node-pg-types/pull/105. Thankfully, it was already just a small wrapper around `postgres-array`, so we can just use that directly.

PR #7901, opened 2023-06-01
Disable pl-order-blocks when new answers are no longer allowed

Frozemint:
As mentioned in #5540, `pl-order-blocks` does not become disabled when question submission is no longer allowed. 

This PR fixes the said defect by adding a `nodrag` class to `li` dragging blocks whenever the param `data['editable']` is set to `false`, and modifying the `sortable` dragging function within `pl-order-blocks.js` to ignore (and hence disallow) dragging of `li` items with the `nodrag` class. This is done to disable the dragging functionality as desired.

The CSS file (`pl-order-blocks.css`) has also been updated to show a `cursor: not-allowed` upon hovering over disabled elements, and disabled elements have opacity set to `0.6` to show that interaction is not allowed (see screencap below). This is done to show a visual cue to the user that dragging of the elements is not possible.

image

### DIFF
--- a/apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.css
+++ b/apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.css
@@ -22,10 +22,18 @@ li.pl-order-blocks-code {
   margin-bottom: 0px !important;
 }
 
-ul.pl-order-blocks-connected-sortable li:hover{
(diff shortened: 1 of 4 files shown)


SethPoulsen (approved):
Hey @Frozemint good to see you around here again!

This looks great, thanks for the contribution!

PR #7902, opened 2023-06-02
Replace workspace killing script with native functionality

nwalters512:
In lieu of `kill_running_workspaces.sh`, the workspace host process will now kill any Docker containers that don't correspond to anything in the database for this host.

This change facilitates #7363 by allowing us to remove Docker from our image to reduce the container size.


apps/workspace-host/src/interface.js
```diff
@@ -104,8 +105,23 @@ let workspace_server_settings = {};
...
+        path.join(APP_ROOT_PATH, 'config.json'),
+      ];
+
+      // If a config file was specified on the command line, we'll use that
+      // instead of the default locations.
+      if ('config' in argv) {
+        configPaths = [argv['config']];
+      }
+
+      await loadConfig(configPaths);
```
nwalters512, line 124: This same code exists in PrairieLearn; it supports the (legacy, I suppose) pattern of having a single `config.json` in the root of the repo that is ready by both `prairielearn` and `grader-host`.

apps/workspace-host/src/interface.js
```diff
@@ -166,11 +182,16 @@ async
...
+        try {
+          await pruneStoppedContainers();
+          await pruneRunawayContainers();
+          await sqldb.queryAsync(sql.update_load_count, {
+            instance_id: workspace_server_settings.instance_id,
+          });
+        } catch (err) {
+          logger.error('Error pruning containers', err);
+          Sentry.captureException(err);
+        }
```
nwalters512, line 194: A small thing I noticed while working on this: previously, if any of these functions threw, the pruning process would essentially die because we'd never schedule another one via the `setTimeout` call below this.

apps/workspace-host/src/interface.js
```diff
@@ -210,16 +231,24 @@ async
...
             debug(`Couldn't find container: ${err}`);
           }
-        } else if (ws.state === 'running') {
-          if (!ws.launch_uuid) {
-            await workspaceUtils.updateWorkspaceState(ws.id, 'stopped', 'Shutting down');
-            await sqldb.queryAsync(sql.clear_workspace_on_shutdown, {
-              workspace_id: ws.id,
-              instance_id: workspace_server_settings.instance_id,
-            });
-          }
```
nwalters512, line 220: @mwest1066 and I discussed this. As far as we can tell, it's impossible for a workspace to be in the `running` state and also have a null `launch_uuid`. Given that, we decided to remove this.

apps/workspace-host/src/interface.js
```diff
@@ -210,16 +231,24 @@ async
...
       });
+
+      // Especially in dev mode, there may be containers that are running but
+      // don't correspond to any known workspace. Clean them up.
+      const allContainers = await docker.listContainers({
+        all: true,
+        filters: { label: ['prairielearn.workspace-id'] },
+      });
+      for (const container of allContainers) {
+        const containerWorkspaceId = container.Labels['prairielearn.workspace-id'];
```
mwest1066, line 244: For dev mode outside of a container we could be even more polite and leave any container without a `prairielearn.workspace-id` label running. These would presumably be other containers running on the dev box, unrelated to PL. I _think_ this would still do what we want in prod, because we presumably label all of our containers?

mwest1066 (approved):

apps/workspace-host/src/interface.js
jonatanschroeder: I think that's what line 241 is doing?

nwalters512: Yeah, that was the intention of that line. Per https://docs.docker.com/engine/reference/commandline/ps, `label=foo` (which is what that line should translate to) finds only images where the label `foo` is present.

PR #7904, opened 2023-06-02
Show answer panel when external grading results are completed

### LINKED ISSUE #1735: Page update after external grading does not display correct answer panel
Steps to reproduce:

1. `fibonacciEditor` problem in the example course
2. Submit an answer and wait for grading to complete
3. Page updates with grading results
4. The “correct answer” panel does not display, which is a bug
5. Reload the page, which causes the “correct answer” panel to display.

jonatanschroeder:
Resolves #1735. In essence, if the answer panel is supposed to be hidden before the grading is completed but shown after it (e.g., in a homework where the external grader results in 100%), then shows the answer panel after submission.

Piggybacking this PR to move the client-side code that updates the panels (as well as mathjax) to an asset script, and to remove reliance on jQuery.


apps/prairielearn/src/pages/instructorQuestionStatistics/instructorQuestionStatistics.ejs
```diff
@@ -2,7 +2,6 @@
 <html lang="en">
   <head>
     <%- include('../partials/head', {pageNote: question.qid}); %>
-    <%- include('../partials/mathjax'); %>
```
jonatanschroeder, line 5: I don't see anywhere in this page where MathJax is actually used.

nwalters512: Lines like these get incorrectly copied around a lot when making new pages, yeah. Thanks for cleaning it up!

apps/prairielearn/assets/scripts/question.ts
```diff
@@ -0,0 +1,162 @@
...
+      // We're done with the socket for this incarnation of the page
+      socket.close();
+      if (msg.answerPanel) {
+        const answerContainer = document.querySelector('.answer-body');
+        answerContainer.innerHTML = msg.answerPanel;
+        answerContainer.closest('.grading-block').classList.remove('d-none');
+      }
+      if (msg.submissionPanel) {
+        document.getElementById('submission-' + submissionId).outerHTML = msg.submissionPanel;
+        window.MathJax.typesetPromise();
```
nwalters512, line 117: I don't think this is safe now that the `async` has been added to the MathJax script tag? We probably want to use `MathJax.config.onReady(() => ...)` to queue up typesetting instead.

apps/prairielearn/assets/scripts/question.ts
```diff
@@ -0,0 +1,162 @@
...
+      authorized_edit: authorizedEdit,
+    },
+    function (msg) {
+      // We're done with the socket for this incarnation of the page
+      socket.close();
+      if (msg.answerPanel) {
+        const answerContainer = document.querySelector('.answer-body');
+        answerContainer.innerHTML = msg.answerPanel;
+        answerContainer.closest('.grading-block').classList.remove('d-none');
+      }
```
nwalters512, line 114: I'm assuming this is the only meaningful change to this script?

For future reference, normally I'd encourage splitting this kind of PR up into "move all existing code almost entirely as-is to asset" and "make changes to asset".

nwalters512 (approved):
I tested this out locally and it looks to be working!

apps/prairielearn/assets/scripts/question.ts
jonatanschroeder: Indeed this is the only change of functionality, other changes are replacing jQuery with vanilla JS and other lint/type related changes while keeping functionality.

PR #7905, opened 2023-06-02
Add support for flash messages

nwalters512:
This PR builds upon sessions introduced in #7143 and adds support for [flash](https://api.rubyonrails.org/classes/ActionDispatch/Flash.html) [messages](https://docs.djangoproject.com/en/4.2/ref/contrib/messages/).

Screenshot 2023-06-07 at 11 00 38

I also rewrote the instructor assessment groups page to use flash messages, which in turn allows us to do proper POST/redirect/GET on that page 🎉


apps/prairielearn/src/pages/partials/navbar.ejs
```diff
@@ -506,6 +506,20 @@
...
+  <div class="alert alert-<%= color %> border-left-0 border-right-0 rounded-0 mb-0" role="alert">
+    <div class="container">
+      <div class="row">
+        <div class="col">
+          <%= message %>
+        </div>
+      </div>
+    </div>
+  </div>
+<% }); %>
```
nwalters512, line 521: Open question: is this the right place? On the instructor assessment groups page (the only page using flash-like things at the moment), errors were shown below the nav tabs, but before I saw that, I added it above the tabs.

In the following screenshot, "Nathan testing ..." is the old location, and the other error is the new one.

Screenshot 2023-06-05 at 16 57 50

jonatanschroeder: I'm more in favour of the old location. It looks closer to the actual page functionality.

nwalters512: I think I agree. However, the old location has the downside that we don't know if a page is normal-width or full-width when we render the navbar, so we can't make the width of the flash messages match that of the page itself.

packages/flash/src/index.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      flash('error', 'goodbye world');
+
+      assert.sameDeepMembers(flash(), [
+        { type: 'notice', message: 'hello world' },
+        { type: 'error', message: 'goodbye world' },
+      ]);
+    });
+  });
+
+  it('clears flash after it has been read', () => {
```
nwalters512, line 77: I should probably document this (I need to add a README to this package), but this is subtly different from what other frameworks (Rails/Laravel/etc.) do. Other frameworks typically persist flash messages *only* until the very next HTTP request, at which point they're cleared, whether or not they've been read. On the other hand, this implementation persists the messages until they're actually read. This is probably fine in practice?

For context, the API I used here is modeled after https://www.npmjs.com/package/connect-flash. I didn't use that package directly because I wanted to use `AsyncLocalStorage` so we don't have to pass around a reference to `req.session` everywhere.

jonatanschroeder: > I should probably document this (I need to add a README to this package), but this is subtly different from what other frameworks (Rails/Laravel/etc.) do. Other frameworks typically persist flash messages _only_ until the very next HTTP request, at which point they're cleared, whether or not they've been read. On the other hand, this implementation persists the messages until they're actually read. This is probably fine in practice?

I actually prefer the option used here. Other HTTP requests may, due to weird timing, end up being an unrelated request (e.g., favicon, deferred/delayed loading of a script/image/style of the previous page, AJAX request from a different tab, etc.).

apps/prairielearn/src/pages/partials/navbar.ejs
mwest1066: Can't you just put this directly into the page, using the `container` or `container-fluid` that's already there? This will then automatically make it the right width.

packages/flash/src/index.test.ts
mwest1066: I agree that this seems like a better system.

packages/flash/src/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    throw new Error('@prairielearn/flash requires session support');
+  }
+
+  const session = req.session as any;
+
+  return {
+    add(type: string, message: string) {
+      session.flash ??= {};
+      session.flash[type] = message;
+    },
```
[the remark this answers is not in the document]
mwest1066: In practice feels like this will be fine, but I can't really see the advantage of storing the flash messages as a Record rather than a simple array like `{type: string; message: string}[]`.

packages/flash/src/index.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export function flashMiddleware() {
+  return (req: Request, _res: Response, next: NextFunction) => {
+    const flashStorage = makeFlashStorage(req);
+    als.run(flashStorage, () => next());
+  };
+}
+
+export function flash(): FlashMessage[];
+export function flash(type: string): string | null;
```
mwest1066, line 19: Can we get some stronger typing on `type`?

mwest1066 (approved):

packages/flash/src/index.ts
nwalters512: I was hesitant to constrain the keys here because there are desirable usages for flash data beyond just error/success/etc. messages. For instance, we could use this on the API token page to persist a newly-generated token for exactly one request.

This might of course be a case of YAGNI; I'm certainly not chomping at the bit to update our tokens page to use flash messages instead. 🤷

apps/prairielearn/src/pages/partials/navbar.ejs
nwalters512: I really want to make this part of our nav so we don't have to repeat this snippet on every page.

GitHub renders these as a full-width thing all the time *and* renders blow the nav tabs. I think this is a good option?

Screenshot 2023-06-06 at 18 23 34

jonatanschroeder: I'm happy with that option.

nwalters512:
Oops, looks like GitHub is delayed in processing commits; there was supposed to be a commit before re-requesting review. Sorry! Hopefully the commit shows up soon.

apps/prairielearn/src/pages/partials/navbar.ejs
mwest1066: Agreed.

packages/flash/src/index.ts
mwest1066: If we ever do get around to using this for tokens (which I agree would be good) we could just add a `token` enum option at that point and handle it explicitly? If you do want to allow arbitrary strings here (I still don't see the use-case) then I'd suggest making that a `flashNonstandardType()` or something so that people will only use it if they really need it.

nwalters512: OK, fair, YAGNI 😛 I'll make this change.

apps/prairielearn/src/pages/partials/navbar.ejs
nwalters512: Moved below the tabs in 13577ec. I enabled dismissal, which required a small amount of additional CSS/JS.

nwalters512: Actually, I was able to achieve the desired behavior _without_ any extra CSS/JS! GitHub isn't showing the commit yet; I'll link to it once it's there.

mwest1066 (approved):

jonatanschroeder (approved):

apps/prairielearn/package.json
```diff
@@ -172,6 +173,7 @@
     "@types/chai-as-promised": "^7.1.5",
     "@types/cheerio": "^0.22.31",
     "@types/dockerode": "^3.3.18",
+    "@types/express": "^4.17.17",
```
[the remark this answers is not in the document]
nwalters512: We use Express anytime we `import`/`require` it; this dependency ensures that we have types when we do so.

apps/prairielearn/src/pages/partials/navbar.ejs
```diff
@@ -524,4 +524,16 @@ if (typeof navPage != 'undefined' && navPagesTabs.hasOwnProperty(navPage)) { %>
         </ul>
     </nav>
 <% } %>
+
+<div class="mb-3">
+  <% const globalFlashColors = { 'notice': 'info', 'success': 'success', 'warning': 'warning', 'error': 'danger' }; %>
+  <% flash(Object.keys(globalFlashColors)).forEach(({ type, message }) => { %>
+    <% const color = type === 'warn' ? 'warning' : type === 'error' ? 'danger' : type === 'notice' ? 'info' : 'success'; %>
```
jonatanschroeder, line 531:
```suggestion
```

jonatanschroeder (approved):

PR #7906, opened 2023-06-02
Parse authenticated user with Zod schema

nwalters512:
This helps ensure that `user_id` and `authn_user_id` will consistently be strings throughout the codebase.


mwest1066:
> The alternative to this PR is to introduce a Zod schema to parse the user/institution/etc. when authenticating - maybe that's the better option?

That sounds much nicer to me!

nwalters512:
@mwest1066 done!

SethPoulsen:
I thought the issue was that one query was using JSONB to return these IDs, and that's how they ended up as numbers. I don't see any queries involving JSONB being edited here. Did I miss something?

nwalters512:
@SethPoulsen this is the magic of ["parse, don't validate"](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)! In particular, check out the `IdSchema` type in `apps/prairielearn/src/lib/db-types.ts` and the corresponding comment:

https://github.com/PrairieLearn/PrairieLearn/blob/43ee9a5cfd4e5a8adab294ffb72c7e940a1de2f4/apps/prairielearn/src/lib/db-types.ts#L3-L5

SethPoulsen:
Ok cool! I guess I don't understand yet how all the Zod stuff is connected so I'll have to look at in more carefully.

Just for my understanding, does this mean that IDs returned through JSONB in queries are getting converted from postgres BIGINT to JS number to JS string? So we could be losing some precision if the BIGINT doesn't fit into the number, regardless of the fact that it is eventually being put into a string.

SethPoulsen (approved):

nwalters512:
Yep, we'll start losing precision at [`Number.MAX_SAFE_INTEGER`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER), about 9 quadrillion. Matt and I talked about this pretty extensively and decided that if we get that many of *anything* in our systems, we'll have significantly larger problems to worry about 😅

nwalters512:
Re: Zod, let me know if you have any questions! We need to document our strategy, which is essentially "parse + validate all data coming back from the database". This allows us to have strong type-checking downstream of the DB. We've been using this strategy for the past year or so in PrairieTest and it's been working very nicely.

Issue #6256, opened 2022-09-02
Cannot copy group assignments from one assessment to another


DallasTrinkle:
In the groups tab for assessments, there is the option to copy group assignments from a previous assessment. After selecting the assessment with the group assignment, it fails and returns with the error: 
```
Error processing request: 
duplicate key value violates unique constraint “group_configs_assessment_id_key”
```
There isn't an existing group assignment for this assessment. However, group assignments can still be uploaded from a CSV file successfully, so I suspect the underlying assessment is fine and this is a PL issue.

nwalters512:
AFAICT, this is because the [`assessment_group_copy` sproc](https://github.com/PrairieLearn/PrairieLearn/blob/bb8cae5491a96d05136be245b7925b67a8464632/sprocs/assessment_groups_copy.sql) is just straight-up broken. It tries to soft-delete old group configs before inserting a new one, but we have a unique constraint on `assessment_id` for the `group_configs` table.

I think this was actually working correctly when groups were first implemented, as the original unique constraint included `WHERE deleted_at IS NULL`:

https://github.com/PrairieLearn/PrairieLearn/blob/bb8cae5491a96d05136be245b7925b67a8464632/migrations/20200817231422_group_change_assessment_inst.sql#L19-L20

However, https://github.com/PrairieLearn/PrairieLearn/pull/3859 changed that to not filter on `deleted_at`, but didn't update the sproc to not perform soft-deletes.

nwalters512:
Related to this, I think it's a mistake to copy `group_configs` when copying group, as `infoAssessment.json` should always be the source of truth for that. We should remove the code that does this. We should ensure that the group assignments being copied are compatible with the group settings of the assignment they're being copied to. For instance, if we're copying from assessment A to B, we should ensure that none of A's groups are smaller or larger than the max or min sizes that are configured for B.

bowmnath:
I think Nathan is right that this worked originally, but it has been broken for a while now (see #3866).  If it is a difficult fix, it might be good to remove the option from the dropdown for now to avoid confusion.

firasm:
Can confirm this is still broken 😅

<img width="1366" alt="Screenshot 2023-02-13 at 9 55 51 PM" src="https://user-images.githubusercontent.com/2507459/218652403-ce80ebed-b643-4d88-89a7-1e33cd9da15b.png">

workaround of downloading the csv from one assignment and then re-uploading into another one seems to work fine.

[...]

PR #7907, opened 2023-06-03
Refactor instructor assessment groups page

nwalters512:
Prepping for some improvements to this page, namely, actually doing POST-Redirect-GET instead of the funky respond-directly-to-POST thing this page does at the moment. This will rely on the flash messages being introduced in #7905 instead of rendering error messages directly in the response.


jonatanschroeder:
Given #6256, do you want to at least remove the copy button to avoid confusion, assuming you don't want to fix it in this PR?

jonatanschroeder (approved):
I'm approving as from my tests it seems to be working as it did before, but I'm assuming you'll handle the POST-redirect-to-GET in a separate PR.

apps/prairielearn/src/pages/instructorAssessmentGroups/instructorAssessmentGroups.js
```diff
@@ -25,144 +29,156 @@ function obtainInfo(req, res, next) {
...
+        res.locals.assessment.id,
+        res.locals.authn_user.user_id,
+      ]);
+      res.redirect(req.originalUrl);
+    } else if (req.body.__action === 'add_group') {
+      const assessment_id = res.locals.assessment.id;
+      const group_name = req.body.group_name;
+      if (!group_name || String(group_name).length < 1) {
+        res.locals.errormsg = 'Please enter a group name when adding a group';
+        await obtainInfo(req, res);
```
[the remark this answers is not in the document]
nwalters512: Yes, I'm going to do proper POST/redirect/GET in a separate PR. This one is cleaning things up so it's slightly easier to make that change. I'm going to remove the "copy" button in another PR as well.

apps/prairielearn/src/pages/instructorAssessmentGroups/instructorAssessmentGroups.js
```diff
@@ -25,144 +29,156 @@ function obtainInfo(req, res, next) {
...
-      });
-    });
-  });
+    },
+    z.object({
+      group_id: IdSchema,
+      name: z.string(),
+      size: z.number(),
+      // TODO: what actually is the type of this thing?
+      uid_list: z.array(z.string()),
```
mwest1066, line 69: That seems like the correct type to me.

mwest1066 (approved):

apps/prairielearn/src/pages/instructorAssessmentGroups/instructorAssessmentGroups.js
nwalters512: Thanks for the reminder to remove this - left over from a WIP!

apps/prairielearn/src/pages/instructorAssessmentGroups/instructorAssessmentGroups.js
```diff
@@ -25,144 +29,156 @@ function obtainInfo(req, res, next) {
...
-        });
-      });
-    });
-  });
+    },
+    z.object({
+      group_id: IdSchema,
+      name: z.string(),
+      size: z.number(),
+      // TODO: what actually is the type of this thing?
```
nwalters512, line 68:
```suggestion
```

Issue #1758, opened 2019-10-19
Randomization of Code Scenarios in Starter Files through Mustache Templating


coatless:
**Goal:**
 
Better support for randomization within code scenarios.

**Idea:**

Given a starter code file and a set of dynamic parameters, run the code file through `mustache` to modify the template with the appropriate `data` parameters. Serve the template to the user.

**Example:**
```
<pl-code source-file-name='my_script.R" template="true"></pl-code>
```

**Use case:**

Customizing random seed given to students.

----

Present sketch...

**server.py**
```python
def generate(data): 
       # Scenario specific changes
       file_params = {
        'value': 1
       }

       # Write scenario into starter template with mustache
        with open('some_code_file.R', 'r', encoding='utf-8') as f:
           written_file = chevron.render(f, file_params).strip()

      # Release into problem
      data['params']['code_template'] = written_file
```

**question.html**
```html
<pl-code language='r'>
{{{params.code_template}}}
</pl-code>
```

nwalters512:
A general question: are there any languages with a double-curly-brace syntax that could conflict with this? Does mustache allow escaping of curly braces if so?

coatless:
@nwalters512 not that I'm aware of. 

Looks like mustache also has support for templating escapes:

```
* {{ default_tags }}
{{=<% %>=}}
* <% erb_style_tags %>
<%={{ }}=%>
* {{ default_tags_again }}
```

c.f. https://github.com/janl/mustache.js#setting-in-templates

coatless:
@eliotwrobson this isn't a zip file question?

eliotwrobson:
@coatless in the example, the zip file contains a python file generated from a template, similar to the example code you have here.

Issue #336, opened 2016-04-18
Expand server-side file templating


mwest1066:
@lotrfan pointed out today that we could expand the server-side templating capabilities that @kevinwang recently introduced to significantly improve the question generation for CS 255 and similar courses.

The rough idea is to allow the templating to have access to the question `params` object, and to add a mechanism for zipping the templated files into a single downloadable archive.

zillesc:
I do the poor man's version of this in my question verilogCombi using the adm-zip library (which provides synchronous zipping).  The question dynamically generates q1.v and q1_tb.v and then zips it up and sends it out.

 ```
       var dirname = options.dirname;
        var q1_zip = new zip();
        q1_zip.addFile(dirname + '/q1.v', new Buffer(question), "verilog file to edit");
        q1_zip.addFile(dirname + '/q1_tb.v', new Buffer(tb), "verilog testbench");
        var buffer = q1_zip.toBuffer();
        var fileData1 = encodeURIComponent(buffer.toString('base64'));
```

eliotwrobson:
@mwest1066 @nwalters512 is this still needed? Seems like plenty of templating / question customization features have been released since this issue was made.

mwest1066:
I guess the question here is whether we need something easier to use than hand-rolling your own templating/zipping. For example, we could provide a `pl-templated-download` element that provides a file download where it automatically runs it through mustache first with `params` available. Such an element could also take a directory and automatically template all the files and zip the result, providing the result as a download.

I know @jonatanschroeder was also working on templated files for external grading. Maybe this interacts in some way?

eliotwrobson:
@mwest1066 take a look at the linked PR. The code for doing this is short enough that I don't think it's worth writing a specific element.

jonatanschroeder:
> I know @jonatanschroeder was also working on templated files for external grading. Maybe this interacts in some way?

I believe you mean @steven-wolfman's work in #5485.

Issue #5372, opened 2022-02-09
Client files question path not available in file function


bowmnath:
The option `data["options"]["client_files_question_path"]` does not appear to exist in the `file` function.  The files are still accessible if the name is stored elsewhere, e.g. a `data['params']['client_files_question_path']` entry that was created in `generate`.  Don't know whether this is a bug or intended behavior, but this limitation isn't mentioned in [https://prairielearn.readthedocs.io/en/latest/question/#accessing-files-on-disk](https://prairielearn.readthedocs.io/en/latest/question/#accessing-files-on-disk).

You might not generally need to access that directory if you are dynamically generating files with `file`, but I am using the `file` function to dynamically modify a skeleton file stored in that directory, so I need to be able to read it.

This has an easy workaround for users as I mentioned above -- storing the path elsewhere.  But, it might be worth at least updating the docs if this is the case.

nwalters512:
I suspect this is a bug, yeah. I'd caution you or anyone else against persisting paths via `params` as a workaround, as we make absolutely no guarantees that those paths will remain unchanged over time. I'll do my best to look into this today. In the meantime, if you can post or send a link to a variant where this didn't work as expected, that'd help me debug here!

PR #7909, opened 2023-06-03
Add random zip file example question

eliotwrobson:
Added an example question to resolve https://github.com/PrairieLearn/PrairieLearn/issues/336 and resolve https://github.com/PrairieLearn/PrairieLearn/issues/1758. Still have to fix an issue where the generated zip file isn't being given to the student (the downloaded file right now is blank for some reason), but the generation logic itself works.

@jonatanschroeder do you have any idea why this might be happening? As I recall, you've used the file generation before.

EDIT: For reasons I cannot fathom, this was happening because the options given to `file()` did not include paths on the server. This was very hard to debug because console output from the `file` function is suppressed.


jonatanschroeder:
> EDIT: For reasons I cannot fathom, this was happening because the options given to `file()` did not include paths on the server.

Note that the question directory is the current working directory (unless your code changes it for some reason), so you can use a relative path instead. No need to use the paths info in your example. But I agree that paths should probably be provided in file() to allow it to use info from, e.g., clientFilesCourse.

> This was very hard to debug because console output from the `file` function is suppressed.

About the console output: it is not suppressed, it is added to an issue. That said, issues created by file() are typically generated after the page has already been rendered (since it's called from a separate request from the browser), so they don't show up in the page itself right away. You need to either reload the page with the same variant (e.g., by saving a submission) to see the issue, or go to the issues page.

eliotwrobson:
> Note that the question directory is the current working directory (unless your code changes it for some reason), so you can use a relative path instead. No need to use the paths info in your example. But I agree that paths should probably be provided in file() to allow it to use info from, e.g., clientFilesCourse.

Ahh, I did not know this. This is especially confusing because there aren't many guarantees on where files are supposed to be located. So generally, all code I write for PL uses the path info taken from options. Other people have reported this issue before: https://github.com/PrairieLearn/PrairieLearn/issues/5372

> About the console output: it is not suppressed, it is added to an issue. That said...

Ahh, I missed this. I'll double-check the issue list.

exampleCourse/courseInstances/SectionA/assessments/homework/extraCredit/infoAssessment.json
```diff
@@ -38,7 +38,8 @@
                 {"id": "demo/calculation",          "points": 1, "maxPoints": 3},
                 {"id": "demo/randomCheckbox",       "points": 1, "maxPoints": 3},
                 {"id": "demo/randomPlot",           "points": 1, "maxPoints": 3},
-                {"id": "demo/randomFakeData",       "points": 1, "maxPoints": 3}
+                {"id": "demo/randomFakeData",       "points": 1, "maxPoints": 3},
+                {"id": "demo/randomZippedFile",       "points": 1, "maxPoints": 3}
```
jonatanschroeder, line 42: Nit: 
```suggestion
                {"id": "demo/randomFakeData",       "points": 1, "maxPoints": 3},
                {"id": "demo/randomZippedFile",     "points": 1, "maxPoints": 3}
```

exampleCourse/questions/demo/randomZippedFile/question.html
```diff
@@ -0,0 +1,22 @@
...
+  </p>
+  <p>
+    Specifically, we generate a zip file <code>generated_files.zip</code>, containing two files,
+    <code>{{params.text_file_name}}</code>, whose contents are
+    <code>{{params.words_for_text}}</code>, and <code>{{params.python_file_name}}</code>, which
+    contains a function named <code>{{params.function_name}}</code> that prints
+    <code>Hello world!</code>.
+  </p>
+  <p>
+    Download the zipped file here
```
jonatanschroeder, line 15: Nit:
```suggestion
    Download the zipped file here:
```

jonatanschroeder (approved):

coatless:
@eliotwrobson maybe change the name to:

```
demo/randomizedTemplateCodeFileZipped
```

from:

```
demo/randomZippedFile
```

to better convey demo contents.

PR #7914, opened 2023-06-04
Add paths to option object in call to file()

### LINKED ISSUE #5372: Client files question path not available in file function

jonatanschroeder:
Based on discussion with @eliotwrobson in #7909. Resolves #5372.

### DIFF
--- a/apps/prairielearn/src/question-servers/freeform.js
+++ b/apps/prairielearn/src/question-servers/freeform.js
@@ -1603,6 +1603,7 @@ module.exports = {
         options: _.get(variant, 'options', {}),
         filename: filename,
       };
+      _.extend(data.options, module.exports.getContextOptions(context));
 
       const { data: cachedData, cacheHit } = await module.exports.getCachedDataOrCompute(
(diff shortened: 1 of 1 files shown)


eliotwrobson (approved):
LGTM!

PR #7918, opened 2023-06-05
feat: More flexibility in BigO grading

eliotwrobson:
See title, slight changes to the backend of the BigO grader to give unconditional correct feedback in more cases. Small API additions to the sympy code, but will not break any existing code (did not have to modify any sympy test cases).


nwalters512 (approved):
Looks reasonable!

PR #7919, opened 2023-06-05
Fix group deletion

nwalters512:
When groups were deleted, either alone or in bulk, we didn't previously actually delete the corresponding `group_users` rows. This meant that if students in a deleted group were ever assigned to another group in the future, they'd be unable to actually view the group because a `query...OneRow` would return multiple rows in the future.

This didn't cause problems before #6109, which introduced `getGroupId` that internally uses `queryValidatedSingleColumnZeroOrOneRow`. Before that, we had a `get_group_info` query that used the plain `query` function and just manually picked the first row off the result with `result.rows[0]`. So, we were always returning multiple rows, we just didn't previously validate our assumption that there would only be one row.


apps/prairielearn/src/lib/groups.sql
```diff
@@ -261,3 +261,70 @@ VALUES
...
+    UPDATE groups AS g
+    SET
+      deleted_at = NOW()
+    WHERE
+      g.id IN (
+        SELECT
+          id
+        FROM
+          assessment_groups
+      )
```
jonatanschroeder, line 319:
```suggestion
    FROM
      assessment_groups ag
    WHERE
      g.id = ag.id
```

Not sure how much this would impact performance, but I'd assume it has better potential for improvement and is cleaner.

apps/prairielearn/src/lib/group-update.js
```diff
@@ -28,7 +28,7 @@ module.exports = {
     let params = { assessment_id };
     sqldb.queryOneRow(sql.select_assessment_info, params, function (err, result) {
       if (ERR(err, callback)) return;
-      const assessmentLabel = result.rows[0].assessmentLabel;
+      const assessmentLabel = result.rows[0].assessment_label;
```
jonatanschroeder, line 31: It seems to me that this column name is implicitly derived from the expression in:
https://github.com/PrairieLearn/PrairieLearn/blob/2193f2cc25a80282fabe0e54bb8566addc1ebb22/apps/prairielearn/src/lib/group-update.sql#L3

I think it would be clearer if that column has an explicit name.

apps/prairielearn/src/lib/groups.sql
nwalters512: Agreed, good suggestion!

apps/prairielearn/src/lib/group-update.js
nwalters512: Added in 9adaf39!

apps/prairielearn/src/lib/groups.sql
```diff
@@ -261,3 +261,60 @@ VALUES
...
+  deleted_groups AS (
+    UPDATE groups AS g
+    SET
+      deleted_at = NOW()
+    FROM
+      assessment_groups AS ag
+    WHERE
+      g.id = ag.id
+    RETURNING
+      id
```
jonatanschroeder, line 311:
```suggestion
      g.id
```

This gave me an error when testing locally.

jonatanschroeder (approved):
Other than the error listed and the lint issue, it looks fine. Not sure if it's worth creating a migration to update existing records?

jonatanschroeder:
> I like the idea of a migration. What do you think that would look like, deleting all `group_users` rows where the corresponding `groups` row has `deleted_at IS NOT NULL`?

That is exactly what I had in mind.

nwalters512:
I like the idea, though to be absolutely sure the migration does the right thing, I think we'd want to merge + deploy it after this PR has been deployed. That way, the migration is guaranteed to catch any affected rows.

PR #7921, opened 2023-06-06
Add allow-blank attribute support in pl-order-blocks

### LINKED ISSUE #5496: allow-blank in pl-order-blocks
Some elements have an `allow-blank` attribute so they will accept a blank answer as a submission, but pl-order-blocks does not. 

I would benefit from having `allow-blank` in order blocks questions, for two reasons:

1. because sometimes that's the correct answer: I use pl-order-blocks to ask questions of the form "Identify which of these events will occur, and put the events that *will* occur in order". Sometimes the correct answer is "None of these events will occur", so I'd want the answer to be blank. 

    As a workaround, I could add a "None of these" source block, but that's awkward and not very intuitive for students.

2. for more forgiving assessments: sometimes a pl-order-blocks question is part of a longer question. If the student leaves the pl-order-blocks part empty, the question is considered invalid and not gradeable, so the student won't get any credit for the part they complete. 

    For some other elements, I can clear format errors in `parse()` if I want the rest of the question to be graded (and the blank part to get zero). But this doesn't work for pl-order-blocks.

Frozemint:
As mentioned in #5496, `pl-order-blocks` does not currently support the `allow-blank` attribute offered by some other PL elements. Currently empty answer submissions (when no blocks has been dragged into the answer area) results in the submission being flagged as an invalid, ungradable submission.

This PR implements the `allow-blank` attribute feature for `pl-order-blocks` (see the screenshots below). The PL element will now look for an `allow-blank` attribute in the `<pl-order-blocks>` HTML tag. If `allow-blank` is set to `true`, then the checks for blank answer submissions are skipped in the `parse` and `grade` function, and student submissions that are blank will now be graded as desired even if they are blank and results in a grade of 0. I have also added a default for `allow-blank` (`PL_ALLOW_EMPTY_SUBMISSION_DEFAULT = False`). This closes #5496.

Finally, I have also taken the liberty to tweak the error message for blank answer submissions to be slightly more informative. 

image
image


apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -624,15 +628,16 @@ def grade(element_html, data):
...
     true_answer_list = data["correct_answers"][answer_name]
 
     final_score = 0
     feedback = ""
     first_wrong = -1
 
-    if len(student_answer) == 0:
-        data["format_errors"][answer_name] = "Your submitted answer was empty."
+    if not allow_blank_submission and len(student_answer) == 0:
+        data["format_errors"][answer_name] = "Your submitted answer was blank; you did not drag any answer blocks into the answer area."
```
SethPoulsen, line 640: Shouldn't we just delete this code? I don't think the grade function will ever be run if there is already a format_error saved by the parse function. We can double check this with the other elements.

SethPoulsen:
Please run the formatter over the code so the CI will pass?

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
Frozemint: What you said about the parse function makes sense (I'm not familiar with the exact behaviour if format_error is set tho 😓 ). I commented out the entire if check, and it doesn't seem to break anything... maybe someone else can chime in on this?

SethPoulsen: I don't know if anyone will now the answer to this off the top of their head, the best thing to do is probably to just look at 4 or 5 of the other elements that implement this feature and see what they do. A lot of the other elements have much shorter code than this one, so it shouldn't take too long to do.

Frozemint: I had a look at `pl-string-input`, `pl-number-input`, `pl-integer-input`, and it looks like you are correct; the other elements only set `format_error` in parse. I'll go ahead and remove the if block check then?

SethPoulsen: Sounds good to me, thanks!

apps/prairielearn/elements/pl-order-blocks/pl-order-blocks.py
```diff
@@ -13,6 +13,7 @@
 
 PL_ANSWER_CORRECT_DEFAULT = True
 PL_ANSWER_INDENT_DEFAULT = -1
+PL_ALLOW_EMPTY_SUBMISSION_DEFAULT = False
```
SethPoulsen, line 16: I'd prefer to have this be `ALLOW_BLANK_DEFAULT = False`
to match what other elements have

SethPoulsen (approved):
Looks good, thanks!

SethPoulsen:
@Frozemint I just realized, we forgot to add this new option to the documentation `docs/elements.md`. Feel free to open another PR to do that if you like.

Frozemint:
Good catch @SethPoulsen . I'll do that in a bit.

PR #7926, opened 2023-06-06
Review use of MathJax promises

jonatanschroeder:
`MathJax` has some of its components loaded asynchronously, so a safe way to request a typeset requires waiting for it to load. Traditionally this has been done with a custom function `MathJax.config.onReady`, but this is not fully safe, as `MathJax.config` is only available once the mathjax script is partially loaded. Using `MathJax.onReady` is also not safe. Similarly, `MathJax.startup.promise` is only available when the script is partially loaded.

This PR introduces a promise that resolves when the script is fully loaded and ready for typeset to be called. It also introduces helper functions that wait for this promise and call typical calls like `typesetPromise` and `tex2svgPromise`.

### DIFF
--- a/apps/prairielearn/assets/scripts/mathjax.js
+++ b/apps/prairielearn/assets/scripts/mathjax.js
@@ -3,49 +3,64 @@
 // Default to SVG, as lines were sometimes disappearing when using the CHTML renderer.
 const outputComponent = 'output/svg';
 
-window.MathJax = {
-  tex: {
-    inlineMath: [
-      ['$', '$'],
-      ['\\(', '\\)'],
-    ],
-  },
-  svg: {
-    // Using local instead of global prevents a MathJax bug if the user
-    // tries to switch the renderer with the pop-up menu. The bug will
-    // supposedly be fixed in MathJax 4, and then this could be changed
-    // back to global.
-    // Refer to issues on MathJax:
-    // https://github.com/mathjax/MathJax/issues/2924
-    // https://github.com/mathjax/MathJax/issues/2956
-    // This PR was merged but won't be released until MathJax v4:
-    // https://github.com/mathjax/MathJax-src/pull/859
-    fontCache: 'local',
-  },
-  loader: {
-    load: ['input/tex', 'ui/menu', outputComponent],
-  },
-  // Register a callback to be run when MathJax is loaded, use MathJax.config.onReady().
-  readyQueue: [],
-  onReady: (cb) => {
-    window.MathJax.config.readyQueue.push(cb);
-  },
-  startup: {
-    ready: () => {
-      window.MathJax.startup.defaultReady();
-      window.MathJax.Hub = {
-        Queue: function () {
-          console.warn(
-            'MathJax.Hub.Queue() has been deprecated in 3.0, please use MathJax.typeset() or MathJax.typesetPromise()'
-          );
-          window.MathJax.typesetPromise();
-        },
-      };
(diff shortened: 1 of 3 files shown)


apps/prairielearn/elements/pl-rich-text-editor/pl-rich-text-editor.js
```diff
@@ -67,7 +67,6 @@ class MathFormula extends Embed {
       const formatted = html.innerHTML;
       // Without trailing whitespace, cursor will not appear at end of text if LaTeX is at end
       node.innerHTML = formatted + '&#8201;';
-      await MathJax.typesetPromise();
```
jonatanschroeder, line 70: This is not necessary, as translation is done explicitly a few lines above.

apps/prairielearn/elements/pl-big-o-input/pl-big-o-input.mustache
```diff
@@ -61,7 +61,7 @@
                     </div>`,
             });
             $('#big-o-input-submission-{{uuid}} [data-toggle="popover"]').on('shown.bs.popover', function() {
-                MathJax.typesetPromise();
+                MathJax.startup.promise.then(() => MathJax.typesetPromise());
```
nwalters512, line 64: I grepped across all questions and unfortunately, there are course elements that make direct use of `MathJax.typesetPromise()`, which means we'll have to keep that as an API to avoid breaking them.

jonatanschroeder: > I grepped across all questions and unfortunately, there are course elements that make direct use of `MathJax.typesetPromise()`, which means we'll have to keep that as an API to avoid breaking them.

Keeping `MathJax.typesetPromise` is not a problem in most cases, it would only cause a problem if it is called before Mathjax finishes the initialization. That said, I could add a new fake `typesetPromise` function that does an appropriate action if Mathjax is not yet loaded, for additional safety. Thoughts? Something similar to what was done with `Mathjax.startup.promise`.

apps/prairielearn/assets/scripts/mathjax.js
```diff
@@ -1,51 +1,66 @@
...
+module.exports = {
+  mathjaxTypeset: async () => {
+    await mathjaxPromise;
+    return window.MathJax.typesetPromise();
+  },
+  mathjaxConvert: async (value) => {
+    await mathjaxPromise;
+    return (window.MathJax.tex2chtmlPromise || window.MathJax.tex2svgPromise)(value);
   },
 };
```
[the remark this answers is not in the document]
jonatanschroeder: I didn't realize it was possible in JS, I thought it was mostly a TS thing. Thanks for pointing it out. Done in ae4fa4cb9fa674f21f39aaf94343759c7b304455.

apps/prairielearn/assets/scripts/mathjax.js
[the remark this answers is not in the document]
jonatanschroeder: Done in ae4fa4cb9fa674f21f39aaf94343759c7b304455.

apps/prairielearn/elements/pl-big-o-input/pl-big-o-input.mustache
jonatanschroeder: Done in 13862461454694f3157a70428e4b62ccc6ca0b02. In this case, a call to `MathJax.typesetPromise()` directly will work either before or after MathJax is finished loading.

nwalters512 (approved):

PR #6141, opened 2022-08-18
Add help link

liiabutler:
This allows instructors to add a link to an assessment that displays to students as a button on each question of an assessment where they can get help from the course staff, if enabled for a course.

Changes include: 
* Add column to assessments table for the help link and update relevant database files
* Add column to pl_courses table for feature enablement and update relevant database files
* Add relevant database migrations 
* Add relevant admin queries
* Add help link partial page
* Modify question score panel partial page

PR #7929, opened 2023-06-07
feat: Add help link

eliotwrobson:
See title, remake of https://github.com/PrairieLearn/PrairieLearn/pull/6141 (since that PR is stale and was not rebased).


apps/prairielearn/src/migrations/20230607155228_pl_courses__help_link__add.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE pl_courses ADD COLUMN IF NOT EXISTS help_link_enabled boolean NOT NULL DEFAULT False;
```
jonatanschroeder, line 1: Feature flags have moved from columns to a separate process, check 
https://github.com/PrairieLearn/PrairieLearn/blob/bc63ab17a0490f2b2221d5fe25a7da76c6485809/apps/prairielearn/src/pages/instructorAssessmentManualGrading/instanceQuestion/instanceQuestion.js#L37-L40
https://github.com/PrairieLearn/PrairieLearn/blob/bc63ab17a0490f2b2221d5fe25a7da76c6485809/apps/prairielearn/src/lib/features/index.ts#L3

eliotwrobson: Out of curiosity, does something this minor really need to be behind a feature flag? Not sure what the convention is for this type of thing.

jonatanschroeder: Good point, I think it makes sense to enable it by default. We've done more breaking changes in the past without a feature flag.

eliotwrobson: Yeah, users won't even see a change unless they have the help links added anyway, and nobody has them yet because the feature is new 😅

apps/prairielearn/src/pages/partials/questionScorePanel.ejs
```diff
@@ -46,6 +46,16 @@
     </tbody>
   </table>
 
+  <% if (assessment.help_link) {%>
+    <div class="card-footer">
+      <div id="linkNav">
```
jonatanschroeder, line 51: Is this id necessary?

apps/prairielearn/src/schemas/schemas/infoAssessment.json
```diff
@@ -136,6 +136,10 @@
             "type": "boolean",
             "default": false
         },
+        "helpLink": {
+          "description": "Link to help source for the assessment.",
```
jonatanschroeder, line 140: A better explanation here would be better. In particular, referencing the fact that this is a link used by students to get help from staff.

jonatanschroeder (commented):
There's a bunch of stuff missing here before this is ready for merging. In particular, this property needs to be listed in the docs, and it might be a good idea to include a CI test for it, in particular ensuring that the value of the field is properly shown in the panel, and that the absence of this field does not cause any issue, and does not show the link. It might also be a good idea to show this in the instructor UI in some capacity (e.g., in assessment settings).

apps/prairielearn/src/pages/partials/questionScorePanel.ejs
eliotwrobson: It may not be, this is from the files I copied.

apps/prairielearn/src/schemas/schemas/infoAssessment.json
eliotwrobson: Done here https://github.com/PrairieLearn/PrairieLearn/pull/7929/commits/a7ab952dc04d980f681aa8bcf73f04727f74ac7e

apps/prairielearn/src/pages/partials/questionScorePanel.ejs
eliotwrobson: Short test has been added here https://github.com/PrairieLearn/PrairieLearn/pull/7929/commits/4ea0074909caf29f922328ae4a149b2bb825bb0b

eliotwrobson:
Closing due to lack of interest for this feature. If needed, we can always reopen later.

PR #3364, opened 2020-10-20
Uid on submissions

astrosticks:
Closes #3248 by displaying the UID of the user who made a submission on the submission panel of a question. Useful for letting group users know who submitted what answer.

# Screenshots
Viewing as user who made the submission
submissionUIDme

Viewing as group member who didn't make the submission
submissionUIDthem

---

- Per Matt's comment, wait until #3518 is merged before merging this.

PR #7930, opened 2023-06-07
Show user UID on submissions

### LINKED ISSUE #3248: Show UID of student who submitted answer (useful for groupwork)
Currently, submissions on groupwork can leave the author of a submission a mystery to other teammates:

image

This makes it hard to reach out to the person who made the submission to get advice, give feedback, etc.

I would suggest showing the UID of the user who made a submission in the submission's description field, e.g. the above example can be reworded as `Submitted by user@domain.net at [timestamp]`. If the submission belongs to the student currently viewing the page, the field can instead say `You submitted at [timestamp]`.

eliotwrobson:
Resolves #3248, remade and rebased version of #3364.


apps/prairielearn/src/pages/partials/submission.ejs
```diff
@@ -73,7 +73,14 @@
           <% } %>
         </span>
       </div>
-      <small>Submitted at <%- submission.formatted_date %></small>
+      <span class="small">
+        <% if (authz_data.user.uid == submission.uid) { %>
```
jonatanschroeder, line 77: I'm wondering if this should be `authn_user` instead of `user`. In essence: if I use effective user, should it show "You" for the authenticated user or the effective user?

apps/prairielearn/src/pages/partials/submission.ejs
```diff
@@ -73,7 +73,14 @@
...
+          <span class="pl-submission-attribution">You</span> submitted
+        <% } else { %>
+          Submitted by <%= authz_data.user.name %> (<span class="pl-submission-attribution"><%= submission.uid %></span>)
```
jonatanschroeder, line 80: `authz_data.user.name` is the effective user's name, not the user associated to the submission.

eliotwrobson: Ahh, is the user associated to the submission just `submission.user`?

apps/prairielearn/src/pages/partials/submission.ejs
eliotwrobson: Good point, this should probably be effective user.

apps/prairielearn/src/pages/partials/submission.ejs
jonatanschroeder: The user associated to the submission is what is being queried in your change to `question.sql`. I don't think this is queried yet, you'd need to add it to the query.

apps/prairielearn/src/lib/question-render.sql
```diff
@@ -96,6 +97,7 @@ FROM
     LIMIT
       1
   ) AS gj ON TRUE
+  LEFT JOIN users u ON (s.auth_user_id = u.user_id)
```
nwalters512, line 100: `submissions.auth_user_id` is unfortunately nullable, so I think technically we should be `LEFT JOIN`ing here. We should almost certainly add a `NOT NULL` constraint to this column in another PR.

apps/prairielearn/src/pages/partials/submission.ejs
```diff
@@ -75,7 +75,15 @@
           <% } %>
         </span>
       </div>
-      <small>Submitted at <%- submission.formatted_date %></small>
+      <span class="small">
+        <% if (!submission.user_uid) { %>
+          Submitted at <%= submission.formatted_date %> 
+        <% } else if (authz_data.authn_user.uid === submission.user_uid) { %>
```
nwalters512, line 81: TODO before merging: `authz_data` is not available when we're rendering this partial outside the context of a full page load (that is, when loading old submissions on the fly). @eliotwrobson how important do you think it is to be "clever" and show "You" instead of the user's UID? We could save some complexity if we're willing to give that up.

eliotwrobson: I think just showing the uid instead of You is probably the way to go, I don't think the extra cleverness is really doing much in this case.

nwalters512: Removed in 6ebc5c6.

apps/prairielearn/src/pages/partials/submission.ejs
nwalters512: We're no longer showing the name.

apps/prairielearn/src/pages/partials/submission.ejs
nwalters512: We're no longer showing "You"; we always just show the UID.

nwalters512:
@eliotwrobson I know this is your PR, but I just took it over so we can get it merged. Want to give it a once-over and make sure things look good?

eliotwrobson:
@nwalters512 everything looks reasonable, though I'm pretty sure you know way more about what's going on here than I do 😂

jonatanschroeder (approved):

PR #7933, opened 2023-06-07
Allow admins to override feature flags on a per-request basis

nwalters512:
I'm doing this so that I can easily test #7128 in production. Specifically, I want to be able to quickly flip this feature on and off in arbitrary courses so that I can compare performance and results.

I borrowed this idea from https://github.com/jnunemaker/flipper/issues/283#issuecomment-327898906. However, I didn't implement all the pattern-matching stuff that that example used. It's just simple on/off for specific flags.

### DIFF
--- a/apps/prairielearn/src/lib/features/manager.ts
+++ b/apps/prairielearn/src/lib/features/manager.ts
@@ -1,5 +1,6 @@
 import { loadSqlEquiv, queryAsync, queryValidatedSingleColumnOneRow } from '@prairielearn/postgres';
 import { z } from 'zod';
+import { AsyncLocalStorage } from 'node:async_hooks';
 
 import { config } from '../config';
 
@@ -12,6 +13,8 @@ const DEFAULT_CONTEXT = {
   course_instance_id: null,
   user_id: null,
 };
+
+export type FeatureOverrides = Record<string, boolean>;
 
 type EmptyContext = Record<string, never>;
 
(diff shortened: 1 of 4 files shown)


apps/prairielearn/src/server.js
```diff
@@ -470,6 +471,13 @@ module.exports.initExpress = function () {
   app.use(require('./middlewares/authn')); // authentication, set res.locals.authn_user
   app.use('/pl/api', require('./middlewares/authnToken')); // authn for the API, set res.locals.authn_user
 
+  // Must come after the authentication middleware, as we need to read the
+  // `authn_is_administrator` property from the response locals.
+  //
+  // This means that feature flag overrides will not be available for
+  // unauthenticated routes.
+  app.use(featuresMiddleware((req, res) => res.locals.authn_is_administrator));
```
nwalters512, line 479: I wasn't entirely sure whether to use this of `is_administrator`; the latter only applies admin access if the `pl_access_as_administrator` cookie says that we should. I was thinking maybe there might be a use case for overriding feature flags when accessing something as a non-admin... but maybe there isn't?

apps/prairielearn/src/lib/features/manager.ts
```diff
@@ -60,6 +64,7 @@ export class FeatureManager<FeatureName extends string> {
       }
     });
     this.features = new Set(features);
+    this.als = new AsyncLocalStorage<FeatureOverrides>();
```
nwalters512, line 67: `AsyncLocalStorage` is so good!!

apps/prairielearn/src/server.js
mwest1066: I don't see anything wrong with the code as it is -- it makes it very simple to understand who can override features.

mwest1066 (approved):

Issue #7385, opened 2023-03-31
Add accessible labels to all element inputs


nwalters512:
All form elements (read: inputs) must have accessible labels (https://dequeuniversity.com/rules/axe/4.6/label). Currently, many have placeholders, and they may also have `aria-describedby` to reference prefixes/suffixes, but these aren't sufficient. We should add reasonable defaults (e.g. `integer` for `pl-integer-input`) and also add support for a `label` attribute that can be used to provide custom, more specific labels.

eliotwrobson:
I've added this issue to the discussion here: https://github.com/PrairieLearn/PrairieLearn/discussions/7353

reteps:
Related to #11857 and #11837

Issue #5902, opened 2022-06-13
"Invalid" responses should be identified beside the question (rather than only in the submitted answer panel)


firasm:
Easy change to make, I think the "invalid-ungradeable" feedback should be shown directly beside where the students put in their answer.

This is also going to apply to:

- pl-symbolic-input
- pl-string-input
- pl-file-editor
- pl-file-upload
- pl-rich-text-editor (min/max size)
- 
<img width="859" alt="173266028-fcb2567c-b186-4985-8472-6a371a93fd2a copy" src="https://user-images.githubusercontent.com/2507459/173266611-a7a19acd-0581-47a1-84e9-471d82005b23.png">

I can take a crack at submitting a PR for this as well.

eliotwrobson:
This is a duplicate of https://github.com/PrairieLearn/PrairieLearn/issues/4246

michaelfromyeg:
Going to pick this one up for `pl-file-upload` (so long as it's still available) with #1447, since they're ish-related (`pl-file-upload` warnings).

eliotwrobson:
This has now been done for all of the most commonly used text input elements:

- pl-string-input
- pl-symbolic-input
- pl-big-o-input
- pl-integer-input
- pl-number-input
- pl-units-input

PR #7934, opened 2023-06-07
Consistency fixes, many enhancements, and small accessibility additions for pl-number-input

### LINKED ISSUE #710: Indicate insufficient precision with client-side JS
For numeric input elements, client-side JavaScript could detect when the input values have insufficient precision and indicate this, for example by turning the notification button on the input element red.

Frozemint:
Per #6968 (~~so close~~), some input textfield elements do not support custom HTML placeholder text. 

This PR implements the said custom placeholder text feature for the PL element `pl-number-input`. If a user specifies a custom `placeholder` HTML text attribute in the `pl-number-input` HTML tag, then the default shortinfo placeholder text (the text that reads "number (rtol=0.01, atol=1e-08)") becomes overridden by the specified placeholder text.

| ℹ️ Edit: The scope of this PR has been expanded per the elaboration from @\eliotwrobson, please see [their comment](https://github.com/PrairieLearn/PrairieLearn/pull/7934#issuecomment-1581632148) and [this additional comment](https://github.com/PrairieLearn/PrairieLearn/pull/7934#discussion_r1251573789) below. Following the scope expansion, the following features are also implemented in this PR: |
| --- |

- The feature to suppress the partial credit score label feature (aka `show-score` from #878). 
- If a `pl-number-input` textfield is ungradable (e.g.: syntax error), a new submission invalid icon is now shown directly next to the offending textfield (#5902).
- Minor changes to the accessibility labelling is added to the input textfield (#7385). A more complete implementation will follow pending the completion of the overhaul of the element `pl-integer-input`.
- Minor changes to the invalid responses feedback UI that is shown to students (the error message is given a more prominent `background-color`). A more complete UI overhaul will follow pending the finalization of #6980 first.
- `pl-number-input` now warns the student if their submitted answer is incorrect and has insufficient precision (precision is not less than or equal to the relative tolerance, closes #710 ).


eliotwrobson:
@Frozemint I think this is good so far, but we were planning on bundling this with some element consistency features (i.e., features that should be enabled on all elements). Take a look at the discussion https://github.com/PrairieLearn/PrairieLearn/discussions/7353. If you had the time, it would be awesome if you could include these changes on this element; they should all be fairly minor things like what you've already done.

As an example, the first element we're uplifting with all these changes at once is the integer input: https://github.com/PrairieLearn/PrairieLearn/pull/6980

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -115,6 +117,7 @@ def render(element_html, data):
             "editable": editable,
             "size": pl.get_integer_attrib(element, "size", SIZE_DEFAULT),
             "uuid": pl.get_uuid(),
+            "placeholder": placeholder,
```
eliotwrobson, line 120: I don't think this needs to be included here?

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -210,10 +213,17 @@ def render(element_html, data):
             info_params.pop("format", None)
             # Within mustache, the shortformat generates the shortinfo that is used as a placeholder inside of the numeric entry.
             # Here we opt to not generate the value, hence the placeholder is empty.
-            info_params["shortformat"] = pl.get_boolean_attrib(
-                element, "show-placeholder", SHOW_PLACEHOLDER_DEFAULT
-            )
-            shortinfo = chevron.render(f, info_params).strip()
+            # The placeholder text may be overriden by setting the 'placeholder' attribute in the pl-number-input HTML tag
```
eliotwrobson, line 216: I think this can be done more cleanly, see https://github.com/PrairieLearn/PrairieLearn/pull/6980/files#diff-30a7667069c2672e849167650819b4eb9fc32c03fea4775a2810f1530a5b96a9R107-R112

eliotwrobson (changes requested):
See other comment, I think it makes sense to attack the consistency issues all at once with this element (also enabling type hints).

Frozemint:
Gotcha. I'll turn this into a draft while I work on your comments and adding more consistency fixes.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
Frozemint: You're right, this was before I realised I could override the `shortinfo` variable

apps/prairielearn/elements/pl-number-input/pl-number-input.py
Frozemint: Yeah, the refactored approach from `pl-integer-input` is definitely way nicer. I tried my hand at doing something similar in the latest commit. Could you have a look when you have the time.

eliotwrobson: I agree it looks much cleaner now! I think we have to keep the `show-placeholder` attribute for backward compatibility, so I think this is about as clean as we can get it. Should be good to start adding more changes to this PR.

Frozemint:
Hi @eliotwrobson, I've added a few more consistency changes to the element. Could you have a look when you have the chance? I haven't had a go at the improving format error UI in the submission panel feature because I'm not quite sure what exactly should be done based on the discussions I could read in the issues. For the other items, I tried my best to put together what I _think_ should be done based on the discussions.

eliotwrobson:
@Frozemint Everything looks reasonable so far, great work! I'll take a more detailed look shortly. As for the UI on the submission panel (and a few other related items), that has been implemented in this PR: https://github.com/PrairieLearn/PrairieLearn/pull/6980

However, that PR is still awaiting review, so I wouldn't spend too much time making everything here look like the way things are in that PR quite yet. In the meantime, the next thing that needs to be done as part of this PR is adding type annotations to the controller for this element. Then, once the integer input PR is reviewed, we can focus on making this element consistent with that one as far as documentation and display.

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
```diff
@@ -12,11 +12,24 @@
 {{#inline}}<span class="form-inline d-inline-block ml-2">{{/inline}}
     <span id="pl-number-input-{{uuid}}" class="input-group pl-number-input">
         {{#label}}
+        <label for="pl-number-input-textfield-{{uuid}}" hidden>{{accessibility_description}}</label>
```
eliotwrobson, line 15: I think this should be an `arialabel` on the input tag? cc @nwalters512

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -102,10 +106,13 @@ def render(element_html, data):
     custom_format = pl.get_string_attrib(
         element, "custom-format", CUSTOM_FORMAT_DEFAULT
     )
+    show_score = pl.get_boolean_attrib(element, "show-score", SHOW_SCORE_DEFAULT)
+    accessibility_description = pl.get_string_attrib(element, "accessibility-description", name)
```
eliotwrobson, line 110: I don't think we want to use `name` as the default here.

eliotwrobson (commented):
Looking good so far! Some minor comments. Like I said in my other comment, the main thing right now is getting typechecking set up in this element while this other style stuff gets finalized.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
Frozemint: Would something like "textfield for inputting answer variable `name`" work better?

eliotwrobson: @nwalters512 would know more. I'm not familiar with all of the accessibility options.

eliotwrobson (changes requested):
Mostly looks fine for now, the main thing is adding type annotations to this element. I won't be able to give this a detailed review until the integr input overhaul is done (and some things with accessibility are getting worked out there still).

Frozemint:
Thanks @eliotwrobson, I had a crack at adding type annotation based on the PR you sent me and checking the PL documentation. Please have a look and see if I have missed anything, and let me know if I should remove the accessibility changes from this PR since we are waiting on the `pl-integer-input` overhaul to be done first.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -326,7 +343,7 @@ def render(element_html, data):
     return html
 
 
-def get_format_string(is_complex=False, allow_fractions=False, message=None):
+def get_format_string(is_complex=False, allow_fractions=False, message=None) -> str:
```
eliotwrobson, line 346: Missing type annotations on input arguments. The type goes before the default argument, send me a DM if you have trouble finding examples.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -363,7 +380,7 @@ def parse(element_html, data):
         data["submitted_answers"][name] = None
 
 
-def grade(element_html, data):
+def grade(element_html, data: pl.QuestionData) -> None:
```
eliotwrobson, line 383: Missing a type annotation for element_html (this happens in a couple places in this file).

eliotwrobson (changes requested):
@Frozemint first set of annotations look good! The main thing is that you have to add pl-number-input to the pyproject.toml file to enable type checking by the CI and locally. That should be in the PR I sent you, but let me know if you can't find it.

With respect to accessibility changes, don't worry about changing those yet until the integer input is finalized. There are a couple other things in this element that can be simplified, I think it's fine to focus on those for now.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
Frozemint: Good catch, thanks.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
Frozemint: Fixed. Note to self if I dig this up again: https://docs.python.org/3/library/typing.html#typing.Optional and https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#functions

Frozemint: Also, since the argument `message` is of type `str`, the type checker is throwing an error because the default `None` is not a `str`. Should I change the default to an empty string instead?

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -294,7 +311,7 @@ def render(element_html, data):
                 else:
                     html_params["incorrect"] = True
             except Exception:
-                raise ValueError("invalid score" + score)
```
eliotwrobson, line 297: For this, the following badge-setting logic can be replaced with the helper function `determine_score_params` (https://github.com/PrairieLearn/PrairieLearn/blob/master/apps/prairielearn/python/prairielearn.py#L138). Take a look at other elements that use this helper function for examples.

Frozemint: Fixed. I'm surprised only 3 other elements currently use this.

eliotwrobson: @Frozemint You don't need to put this in a try-except block anymore, and you should also use this function on line 141 (where the score is set as well).

Frozemint: Thanks for catching that. Is there anything else that is missing in this PR?

eliotwrobson: Outside of changes waiting for the integer input PR, I think this is ok for now. There's a change I might want to make to the feedback that gets displayed to students, but I need to think about it first.

eliotwrobson: @Frozemint looking at this again (and since your other PR got closed out), I think it would be good to address https://github.com/PrairieLearn/PrairieLearn/issues/710 as part of this PR, and change the way that correct answer feedback is displayed / processed to match that of the units input element (for example). That way, the additional feedback in the linked issue can be a short addon to the grader code. Feel free to DM me if you would like a more detailed description.

Also, great work on your PRs so far!

Frozemint: *shot

Frozemint: @eliotwrobson I took a look at `pl-unit-input` and the way they handle feedback (correct unit, incorrect magnitude), and added something similar in the commits I just pushed. Since neither `pl-number-input` or `pl-unit-input` uses JS, this implementation is not JS based. My implementation is quite crude and quickly put together in case this isn't quite what you are looking for. Please have a look and feel free to make any changes you see fit.

eliotwrobson: @Frozemint thanks! I think this is a really good starting point. Sorry the issue was a bit unclear, we hashed out some discussion in the comments section about how the precision should be handled. This was exactly how you did it though (on the backend in Python and not in JS). There are a couple of minor notes I have, will leave in a review.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -372,68 +410,78 @@ def grade(element_html, data):
...
+            a_tru_converted = np.float64(a_tru)
+
+        # Compare submitted answer with true answer
+        if comparison is ComparisonType.RELABS:
+            rtol = pl.get_float_attrib(element, "rtol", RTOL_DEFAULT)
+            atol = pl.get_float_attrib(element, "atol", ATOL_DEFAULT)
+
+            a_sub_precision = get_string_precision(str(a_sub))
+            if not math.isclose(a_sub_precision, rtol, rel_tol=rtol):
+                feedback = ANSWER_INSUFFICIENT_PRECISION_WARNING
```
eliotwrobson, line 460: We only want to leave feedback in the case the student answer is incorrect: https://github.com/PrairieLearn/PrairieLearn/issues/710#issuecomment-1607032756

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -372,68 +410,78 @@ def grade(element_html, data):
...
+
+            return (
+                pl.is_correct_scalar_ra(a_sub_converted, a_tru_converted, rtol, atol),
+                feedback,
+            )
+        elif comparison is ComparisonType.SIGFIG:
+            digits = pl.get_integer_attrib(element, "digits", DIGITS_DEFAULT)
+            return (
+                pl.is_correct_scalar_sf(a_sub_converted, a_tru_converted, digits),
+                None,
```
eliotwrobson, line 470: @Frozemint So right now in the case of a correct answer, the following gets displayed:
image
This is effectively feedback that I'd rather have returned here (instead of stored in a popover). In the mustache template, can you remove this part of the submission panel and have that be handled as part of the grading logic here?

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -65,47 +87,71 @@ def prepare(element_html, data):
...
-            raise ValueError(
-                'method of comparison "%s" is not valid (must be "relabs", "sigfig", or "decdig")'
-                % comparison
-            )
+            assert_never(comparison)
     return a_tru
 
 
-def render(element_html, data):
+def get_string_precision(number_string: str) -> float:
```
eliotwrobson, line 117: (comment so I don't forget) we should have some test cases for these. However, part of this PR might include changing the name of this file so it's easier to write tests. Don't worry quite yet.

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
```diff
@@ -108,8 +124,22 @@
   {{/a_tru}}
 </span>
 {{/correct}}
-{{#partial}}<span class="badge badge-warning"><i class="fa fa-circle-o" aria-hidden="true"></i> {{partial}}%</span>{{/partial}}
-{{#incorrect}}<span class="badge badge-danger"><i class="fa fa-times" aria-hidden="true"></i> 0%</span>{{/incorrect}}
+{{#partial}}
```
eliotwrobson, line 127: It looks like the feedback is only in the partial / incorrect case? The template should be such that there can always be feedback displayed (see other comments about correct answer feedback).

Frozemint: Will do.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
Frozemint: So something that looks like this? 
image
Right now the `the correct answer used for grading` text is not part of the grading logic, it is hardcoded text in the mustache, what I did was move the text from a popover into its own feedback span instead.

eliotwrobson: > So something that looks like this?

Exactly! Looks great. Moving the correct answer feedback into the grading logic is the way to go.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
Frozemint: Fixed.

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
Frozemint: Fixed in the latest commits.

eliotwrobson:
@Frozemint https://github.com/PrairieLearn/PrairieLearn/pull/6980 has been merged! Thanks for your hard work and patience so far. Now that has been merged, should be good to copy the way the mustache template has been organized (among other things).

The main difference I noticed when playing around with the current code is that the invalid display here differs from the changed version in https://github.com/PrairieLearn/PrairieLearn/pull/6980, so that will need to be updated. Also, there were some differences between the attributes given to the `<input>` tag in the mustache template.

Frozemint:
That's great news 🎉 . I will take a look at copying the mustache template and changing the invalid display to be consistent with the other element. I will keep you updated.

Frozemint:
Hi @eliotwrobson, I've just pushed changes to `pl-number-input`'s mustache such that its invalid display and <input> tag are (more) consistent with that of `pl-integer-input`'s. Sorry for the radio silence, life got into the way.

eliotwrobson:
@Frozemint no worries! I'm working on a few things right now so I won't be able to look until later in the afternoon. For now, my only comment is that, in https://github.com/PrairieLearn/PrairieLearn/pull/7934/commits/8e48bad8b5d23bef4459c4978151815209ec4cf9, you removed the type annotations and `bool()` casts to some of the helper functions in `prairielearn.py`. Those need to be added back, since they fixed some subtle bugs with the return types of those functions.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -65,47 +86,71 @@ def prepare(element_html, data):
...
+    if "." in number_string:
+        return 10 ** -len(number_string.partition(".")[2])
+    else:
+        return 10 ** (len(number_string) - len(number_string.rstrip("0")))
+
+
+def get_string_significant_digits(number_string):
```
eliotwrobson, line 123:
```suggestion
def get_string_significant_digits(number_string: str) -> int:
```

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -372,68 +409,83 @@ def grade(element_html, data):
...
-    #   Casting as np.float64 avoids this error. This is reasonable in any case,
-    #   because <pl-number-input> accepts double-precision floats, not ints.
-    #
-    if np.iscomplexobj(a_sub) or np.iscomplexobj(a_tru):
-        a_sub = np.complex128(a_sub)
-        a_tru = np.complex128(a_tru)
-    else:
-        a_sub = np.float64(a_sub)
-        a_tru = np.float64(a_tru)
+    # a_sub = data["submitted_answers"].get(name, None)
```
eliotwrobson, line 417: Remove commented out code.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -372,68 +409,83 @@ def grade(element_html, data):
...
+            digits = pl.get_integer_attrib(element, "digits", DIGITS_DEFAULT)
+            return (
+                pl.is_correct_scalar_dd(a_sub_converted, a_tru_converted, digits),
+                None,
+            )
+        else:
+            assert_never(comparison)
+
+    pl.grade_answer_parameterized(data, name, grade_function, weight=weight)
+    # if correct:
```
eliotwrobson, line 482: Remove commented out code.

docs/elements.md
```diff
@@ -324,20 +324,34 @@ def fib(n):
...
+| `tag`            | string             | ""      | Optional attribute. Used to identify the block when declaring which other blocks depend on it or are a distractor for it.                                                                                                                                                                                                                                                 |
+| `distractor-for` | string             | ""      | Optional attribute on blocks where `correct=false`. Used to visually group a distractor block with a correct block that it is similar to, should match the `tag` attribute of the block that it should be visually paired with.                                                                                                                                           |
```
eliotwrobson, line 354: These docs changes are from order blocks??

pyproject.toml
```diff
@@ -75,7 +75,7 @@ include = [
     "./apps/prairielearn/elements/pl-hide-in-panel",
     "./apps/prairielearn/elements/pl-integer-input",
     "./apps/prairielearn/elements/pl-manual-grading-only",
-    "./apps/prairielearn/elements/pl-matrix-latex",
```
eliotwrobson, line 78: We still need to typecheck `matrix-latex` as well.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -372,68 +409,83 @@ def grade(element_html, data):
...
+            a_sub_converted = np.float64(a_sub_parsed)
+            a_tru_converted = np.float64(a_tru)
+
+        # Compare submitted answer with true answer
+        if comparison is ComparisonType.RELABS:
+            rtol = pl.get_float_attrib(element, "rtol", RTOL_DEFAULT)
+            atol = pl.get_float_attrib(element, "atol", ATOL_DEFAULT)
+            return (
+                pl.is_correct_scalar_ra(a_sub_converted, a_tru_converted, rtol, atol),
+                None,
```
eliotwrobson, line 464: It looks like `None` is always being used for feedback now? I thought feedback messages were introduced in a previous commit in this PR.

eliotwrobson (changes requested):
The core logic looks ok, but some oddities that seem like they came from issues during a merge commit. The mustache template looks ok now, but there were some things removed from `prairielearn.py` and logic in the python controller that need to be added back in.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
Frozemint: Removed as suggested; IIRC and for future reference, those were a leftover from when we weren't sure whether we have to check for blank submissions in grading since we weren't sure whether blank submissions would be caught in `parse()`, but as it turns out, it is caught in `parse()`.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
Frozemint: Removed.

pyproject.toml
Frozemint: Readded. I think I might have broken the laws of physics when trying to merge latest changes into this branch after 3 weeks of inactivity.

docs/elements.md
Frozemint: Yeah, I have no idea how these got here 💀

Frozemint:
> but there were some things removed from prairielearn.py and logic in the python controller that need to be added back in.

Yes, @eliotwrobson; I think I might have broken something (possibly the universe's laws of physics as well) while merging changes from master into this branch, ill check at your typechecking commit and add the changes back in.

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
```diff
@@ -12,26 +12,53 @@
...
               <span class="d-flex align-items-center">
             {{/show_info}}
-
-            {{#correct}}&nbsp;<span class="badge badge-success"><i class="fa fa-check" aria-hidden="true"></i> 100%</span>{{/correct}}
-            {{#partial}}&nbsp;<span class="badge badge-warning"><i class="fa fa-circle-o" aria-hidden="true"></i> {{partial}}%</span>{{/partial}}
-            {{#incorrect}}&nbsp;<span class="badge badge-danger"><i class="fa fa-times" aria-hidden="true"></i> 0%</span>{{/incorrect}}
-
+            {{#parse_error}}
+                 <a role="button"
+                   class="btn btn-light border d-flex align-items-center text-danger rounded-right"
```
eliotwrobson, line 46: As I have now discovered, we can just get rid of all the rounded rights: https://github.com/PrairieLearn/PrairieLearn/pull/8270

apps/prairielearn/elements/pl-number-input/pl-number-input.py
eliotwrobson: I think the feedback given for getting the question correct disappeared as well (where it output the reference solution).

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
```diff
@@ -12,26 +12,53 @@
...
+                   title="Format Error"
+                   data-placement="auto"
+                   data-trigger="focus"
+                   tabindex="0"
+                   data-content="{{parse_error}}">
+                    <span class="mr-1">Invalid </span> <i class="fa fa-exclamation-triangle" aria-hidden="true"></i>
+                </a>
+            {{/parse_error}}
+            {{#show_score}}
+              {{#correct}}&nbsp;<span class="badge badge-success"><i class="fa fa-check" aria-hidden="true"></i> 100%</span>{{/correct}}
```
eliotwrobson, line 58: Using `&nbsp;` is a bit hacky, would be better to use the `mr-1` (or similar) bootstrap class: https://getbootstrap.com/docs/4.0/utilities/spacing/

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
Frozemint: Removed as suggested.

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
```diff
@@ -12,26 +12,53 @@
 {{#inline}}<span class="form-inline d-inline-block ml-2">{{/inline}}
     <span id="pl-number-input-{{uuid}}" class="input-group pl-number-input">
         {{#label}}
+        <label for="pl-number-input-textfield-{{uuid}}" hidden>{{accessibility_description}}</label>
         <span class="input-group-prepend">
-            <span class="input-group-text">{{{label}}}</span>
+            <span class="input-group-text" id="pl-number-input-{{uuid}}-label">{{{label}}}</span>
         </span>
         {{/label}}
-        <input name="{{name}}" autocomplete="off" type="text" class="form-control {{^display_append_span}}rounded-right{{/display_append_span}} pl-number-input-input" size="{{size}}" {{^editable}}disabled{{/editable}} {{#raw_submitted_answer}}value="{{raw_submitted_answer}}"{{/raw_submitted_answer}} aria-describedby="basic-addon1" placeholder="{{shortinfo}}" />
```
eliotwrobson, line 19: Now that the rounded-right logic is removed here, I think we can remove the `display_append_span` key from the dict passed into this mustache template (I don't think that key is used anywhere else).

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
eliotwrobson: It doesn't seem like this label is needed? I don't even think the `accessibility_description` is being set anywhere.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -65,47 +86,71 @@ def prepare(element_html, data):
...
-    display = pl.get_string_attrib(element, "display", DISPLAY_DEFAULT)
+    display = pl.get_enum_attrib(element, "display", DisplayType, DISPLAY_DEFAULT)
     allow_fractions = pl.get_boolean_attrib(
         element, "allow-fractions", ALLOW_FRACTIONS_DEFAULT
     )
     custom_format = pl.get_string_attrib(
         element, "custom-format", CUSTOM_FORMAT_DEFAULT
     )
+    show_score = pl.get_boolean_attrib(element, "show-score", SHOW_SCORE_DEFAULT)
+    accessibility_description = pl.get_string_attrib(
```
eliotwrobson, line 146: I think this can be removed, since the custom description is just being run through a hidden label (but we have the `aria-labeledby` set independently).

apps/prairielearn/elements/pl-number-input/pl-number-input.py
eliotwrobson: Unresolving this since this change got lost in the merge.

docs/elements.md
eliotwrobson: Yeah I think git freaked out on you 😢

eliotwrobson (changes requested):
Left a more detailed review since, after the issues with merging get resolved, this is almost ready to go. It would be awesome if we could land this before the semester starts, since this element gets used pretty regularly and the changes are a substantial quality-of-life improvement.

Also @Frozemint thanks for sticking with this and for the great work so far! I know that the expanded scope + other things have been a bit of a drag, but I think this is really close.

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
Frozemint: I checked, and it looks like I have removed the `display_append_span` keys per [9743a97](https://github.com/PrairieLearn/PrairieLearn/pull/7934/commits/9743a9740eb46720989ebb29be003e26577beec8), when the rounded-right logic was removed. I think I was wise to the fact that the keys have no use aside from serving the function of `rounded-righ`, and so the keys were removed then.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
Frozemint: As I remember, these were added per #7385, but I think we could remove these for now and I can open another PR to address the accessibility issues later, since you mentioned it would be a good idea to get what we have here merged first before the flood gates of September opens. 

@Frozemint note to self: open the accessibility PR

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
eliotwrobson: It looks like the key is still being set @Frozemint ? Check line 275: https://github.com/PrairieLearn/PrairieLearn/blob/d7f4c02c81343c02646b3e6e5e2a423a7b2b0e69/apps/prairielearn/elements/pl-number-input/pl-number-input.py#L275

apps/prairielearn/elements/pl-number-input/pl-number-input.py
Frozemint: Fixed again in [bfb518d](https://github.com/PrairieLearn/PrairieLearn/pull/7934/commits/bfb518dc636477ea6722f16869a75d7447386577).

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
Frozemint: Good catch 💀

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
```diff
@@ -96,20 +125,21 @@
         });
     });
 </script>
+{{#show_correct}}
```
eliotwrobson, line 128: Slight regression, this was changed at some point to do this through the grading logic function and not a separate key. See discussion: https://github.com/PrairieLearn/PrairieLearn/pull/7934#discussion_r1253733443

Frozemint: By this, you mean that if the student answer submission is correct, then the answer used for grading should always be shown?

I definitely remember we have talked about this earlier at some point, I dont know if it's the merging nightmare that introduced this regression while fixing the [regression from earlier](https://github.com/PrairieLearn/PrairieLearn/pull/7934#discussion_r1285124690)

eliotwrobson: > By this, you mean that if the student answer submission is correct, then the answer used for grading should always be shown?

Yes

> I definitely remember we have talked about this earlier at some point, I dont know if it's the merging nightmare that introduced this regression while fixing the [regression from earlier](https://github.com/PrairieLearn/PrairieLearn/pull/7934#discussion_r1285124690)

Yeah, I think this got lost somewhere. I think one of the earlier versions (pre-merge) had this as part of the logic in the grading function, so it shouldn't be _too_ hard to fish out that code.

Frozemint: I had a quick look at the mustache file from the earlier commits before the big merge, and it looks like the code snippet (the div showing the correct answer used for grading) is always nested between `show_correct` for some reason. But since the line of code is nested between `{{#correct}}`, I think we could just take out the `show_correct` altogether since everything nested between `correct` wont be shown unless it's a correct 100% full credit submission (I think).

eliotwrobson: I agree, I think that everything in `{{#show_correct}}` can be removed and handled more cleanly by returning it as feedback in the `grade()` function, where it will get displayed automatically.

docs/elements.md
```diff
@@ -814,6 +814,7 @@ def generate(data):
...
 | `show-help-text`      | boolean                         | true     | Show the question mark at the end of the input displaying required input parameters.                                                                                                                                                                                                                                                                                                                                                                                                    |
 | `show-placeholder`    | boolean                         | true     | Show the placeholder text that shows the default comparison.                                                                                                                                                                                                                                                                                                                                                                                                                            |
```
eliotwrobson, line 820: This should be removed in favor of the new `placeholder` attribute.

Frozemint: That reminds me, the placeholder attribute is not in the docs yet.

That said, should the `show-score` documentation be removed? I thought this was one of the new features this PR adds.

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
Frozemint: RIGHT, now memories are coming back. I think what we did was append the correct answer used for grading inside whatever feedback the student is also getting. Now I understand why the code snippet in question seems like an anachronism. 

So this is what people mean by when you code, you are both the perpetrator and the detective.

Frozemint: Note to self on a discussion we probably had: I think we also talked about not liking the "this is the correct answer used for grading" being hardcoded in mustache (since this is ugly), and instead everything should be done in the `grade()` function. You'll probably have to dig up earlier reviews to find the traces of evidence of this discussion we had.

docs/elements.md
eliotwrobson: Yes, the show-score is added by this PR, so it should stay in the docs.

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
```diff
@@ -125,11 +125,9 @@
         });
     });
 </script>
-{{#show_correct}}
 <div class="d-inline pl-number-input-feedback">
```
eliotwrobson, line 128: Should remove what's inside this tag as well.

Frozemint: This should be removed in [ee51086](https://github.com/PrairieLearn/PrairieLearn/pull/7934/commits/ee51086858ce7c37114e79320ac1f2e5e4ca5c07).

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
```diff
@@ -86,6 +108,13 @@
...
+{{#feedback}}
+<div class="d-inline-flex pl-number-input-feedback">
+  {{feedback}}
+</div>
+{{/feedback}}
+
+
 {{#correct}}
 <script>
     $(function(){
```
eliotwrobson, line 120: I don't think the stuff in this script tag is needed anymore?  The correct badge doesn't need to be a popover anymore. This should be just a badge, the way that the partial and incorrect tags below are set up.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -65,47 +85,68 @@ def prepare(element_html, data):
...
 
-def render(element_html, data):
+def get_string_precision(number_string: str) -> float:
+    if "." in number_string:
+        return 10 ** -len(number_string.partition(".")[2])
+    else:
+        return 10 ** (len(number_string) - len(number_string.rstrip("0")))
+
+
+def get_string_significant_digits(number_string: str) -> int:
```
eliotwrobson, line 122: It doesn't look like this function gets called anywhere?

docs/elements.md
```diff
@@ -814,9 +814,11 @@ def generate(data):
...
```
eliotwrobson, line 820: The use of `show-placeholder` is deprecated with the introduction of the custom `placeholder` attribute. It should be removed from the docs (but should stay in the code, the current implementation is correct).

eliotwrobson (changes requested):
Last round of nitpicks. I haven't gotten to play with this myself yet, but I think this is very close to being ready to go!!

docs/elements.md
Frozemint: so ill just remove the `show-placeholder` line then?

apps/prairielearn/elements/pl-number-input/pl-number-input.py
Frozemint: Not yet; we are only addressing the adding of precision warning for the comparison type `RELABS` in this PR (since this is the scope of #710), but while researching how to do the precision checks I also went ahead and and looked at how to do the significant digits checks (for the other two comparison types). I could remove the code for now and leave this for another PR.

docs/elements.md
eliotwrobson: > so ill just remove the `show-placeholder` line then?

Yep!

apps/prairielearn/elements/pl-number-input/pl-number-input.py
eliotwrobson: Yeah, I think we should remove for now and we can add back later. Shouldn't be too difficult to do in a followup, but would like to avoid any dead code.

docs/elements.md
Frozemint: Fixed in [616c262](https://github.com/PrairieLearn/PrairieLearn/pull/7934/commits/616c26287a9ad0f2d323ed38ffe3aa742b5a2bcd).

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
Frozemint: That makes sense. Fixed in [82622b3](https://github.com/PrairieLearn/PrairieLearn/pull/7934/commits/82622b3b92a4429aa9c83c500d60c47d77c9b474).

apps/prairielearn/elements/pl-number-input/pl-number-input.py
Frozemint: Sounds good. Removed for now in [faf4997](https://github.com/PrairieLearn/PrairieLearn/pull/7934/commits/faf4997f2b01ae5e2028d597d003ca1a234773d0).

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -284,22 +297,19 @@ def render(element_html, data):
...
-            except Exception:
-                raise ValueError("invalid score" + score)
+            score_type, score_value = pl.determine_score_params(score)
+            html_params[score_type] = score_value
 
         html_params["error"] = html_params["parse_error"] or html_params.get(
             "missing_input", False
         )
 
+        feedback = partial_score.get("feedback", None)
```
eliotwrobson, line 309:
```suggestion
        html_params["feedback"] = partial_score.get("feedback", None)
```
Nit: `None` is "falsy" with respect to chevron rendering, so there's no need for the if guard below

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -372,68 +386,83 @@ def grade(element_html, data):
...
+        # Compare submitted answer with true answer
+        if comparison is ComparisonType.RELABS:
+            rtol = pl.get_float_attrib(element, "rtol", RTOL_DEFAULT)
+            atol = pl.get_float_attrib(element, "atol", ATOL_DEFAULT)
 
+            a_sub_precision = get_string_precision(str(a_sub))
+            is_correct = pl.is_correct_scalar_ra(
+                a_sub_converted, a_tru_converted, rtol, atol
+            )
+            if not is_correct and (a_sub_precision > rtol):
```
eliotwrobson, line 442: Nit: I know we decided against adding the precision warnings for the other grading modes, but might be nice to put a TODO to add that in with a link to this PR in this grading function.

eliotwrobson (approved):
A couple of small nitpicks. I haven't had a chance to play with this locally, so I think we should hold off merging until at least one maintainer does that, but other than that this looks great!

apps/prairielearn/elements/pl-number-input/pl-number-input.py
Frozemint: Added in [d5c5926](https://github.com/PrairieLearn/PrairieLearn/pull/7934/commits/d5c59269ada670bda3ebf158c0d4557f3dc579e1).

Frozemint: I will also open an issue for this after this PR gets merged. @Frozemint

apps/prairielearn/elements/pl-number-input/pl-number-input.py
Frozemint: Changed as suggested in [175f82c](https://github.com/PrairieLearn/PrairieLearn/pull/7934/commits/175f82c20eab5ac9efc8e19696f5669740fc160c).

apps/prairielearn/python/prairielearn.py
```diff
@@ -1588,37 +1589,39 @@ def is_correct_ndarray2D_ra(a_sub, a_tru, rtol=1e-5, atol=1e-8):
...
+        return real_comp and imag_comp
 
     # Get bounds on submitted answer
     eps = 0.51 * (10**-digits)
     lower_bound = a_tru - eps
     upper_bound = a_tru + eps
 
     # Check if submitted answer is in bounds
-    return (a_sub > lower_bound) & (a_sub < upper_bound)
+    return bool((a_sub > lower_bound) & (a_sub < upper_bound))
```
nwalters512, line 1614: Is there a specific reason this is using a bitwise `&` operator instead of the usual `and` operator?

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -65,47 +85,61 @@ def prepare(element_html, data):
...
 
-def render(element_html, data):
+def get_string_precision(number_string: str) -> float:
+    if "." in number_string:
+        return 10 ** -len(number_string.partition(".")[2])
+    else:
+        return 10 ** (len(number_string) - len(number_string.rstrip("0")))
+
+
+# TODO: add precision calculation function for other grading methods
```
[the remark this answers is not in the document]
Frozemint: TODO for a future PR, as discussed with @\eliotwrobson somewhere further up in this PR; we decided that it'll be a good idea to merge this PR first, as it is already dragging on and ballooning in scope and changes.

apps/prairielearn/python/prairielearn.py
Frozemint: I'm not sure; I believe this change was originally pushed by @\eliotwrobson when they added typechecking to the [prairielearn.py](https://github.com/PrairieLearn/PrairieLearn/pull/7934/files/f9ae7ffa00fe0f94b517feaade01ad6e920a6bcd#diff-58735c0379465a7c9e00a320feb090342b1561025053fbe98c118fd44484d375) file. You'll have to ask them.

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
```diff
@@ -132,9 +149,6 @@ Your answer must be a real number between <code>-1e308</code> and <code>1e308</c
...
 {{/show_info}}
-{{#show_correct}}
-The correct answer used for grading is <samp>{{a_tru}}</samp>
-{{/show_correct}}
```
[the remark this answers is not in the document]
eliotwrobson: This is currently being used, but changed to display in the feedback box instead of being a popover (and requiring the extra parameters). See https://github.com/PrairieLearn/PrairieLearn/blob/master/apps/prairielearn/elements/pl-number-input/pl-number-input.py#L186

If you take a look at the new `grade()` function, text similar to this gets added there.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -65,47 +85,61 @@ def prepare(element_html, data):
...
-        elif comparison == "decdig":
+        elif comparison is ComparisonType.DECDIG:
             digits = pl.get_integer_attrib(element, "digits", DIGITS_DEFAULT)
             a_tru = "{:.{ndigits}f}".format(a_tru, ndigits=digits)
         else:
-            raise ValueError(
-                'method of comparison "%s" is not valid (must be "relabs", "sigfig", or "decdig")'
-                % comparison
-            )
+            assert_never(comparison)
```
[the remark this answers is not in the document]
eliotwrobson: Yep! `pl.get_enum_attrib` displays the possible values of the enum to the user: https://github.com/PrairieLearn/PrairieLearn/blob/master/apps/prairielearn/python/prairielearn.py#L185

apps/prairielearn/elements/pl-number-input/pl-number-input.py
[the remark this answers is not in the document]
eliotwrobson: For the future. Adding the precision calculation for the other methods requires new helper functions, and we wanted to not make this PR any longer than it already is.

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
```diff
@@ -63,11 +89,7 @@
 {{#parse_error}}
     {{#label}}<span>{{{label}}}</span>{{/label}}
     {{#raw_submitted_answer}}<code class="user-output-invalid">{{raw_submitted_answer}}</code>{{/raw_submitted_answer}}
-
-    <span class="badge text-danger badge-invalid"><i class="fa fa-exclamation-triangle" aria-hidden="true"></i> Invalid</span>
-
-    <a href="javascript:void(0);" role="button" class="btn btn-sm btn-secondary small border" data-placement="auto" data-trigger="focus" data-toggle="popover" data-html="true" style="margin-left: 5px" title="Format Error" tabindex="0" data-content="{{parse_error}}"> Why <i class="fa fa-question-circle" aria-hidden="true"></i>
-    </a>
+    <a href="javascript:void(0);" role="button" class="badge text-danger badge-invalid btn btn-sm btn-secondary small border" data-placement="auto" data-trigger="focus" data-toggle="popover" data-html="true" style="margin-left: 5px" title="Format Error" tabindex="0" data-content="{{parse_error}}"> Invalid <i class="fa fa-exclamation-triangle" aria-hidden="true"></i></a>
```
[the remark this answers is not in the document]
eliotwrobson: Copied from other elements. A bunch of them have this for whatever reason: https://github.com/PrairieLearn/PrairieLearn/blob/master/apps/prairielearn/elements/pl-string-input/pl-string-input.mustache#L66

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -225,22 +255,9 @@ def render(element_html, data):
             html_params["popup_title"] = "Correct Answer"
 
         # Enable or disable the popup
-        if (
-            pl.get_boolean_attrib(element, "show-help-text", SHOW_HELP_TEXT_DEFAULT)
-            or show_correct
```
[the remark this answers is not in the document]
eliotwrobson: Seems like a bug? I guess the idea is if someone didn't show the correct answer, they didn't want to show the help text either. This is another weird artifact that's really old and I don't really know what the thinking behind it was.

apps/prairielearn/python/prairielearn.py
eliotwrobson: I don't think so, just what was here before (for whatever odd reason).

apps/prairielearn/elements/pl-number-input/pl-number-input.py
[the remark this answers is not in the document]
nwalters512: I'd encourage one of you to open an issue for this and explain in more detail what needs to be done.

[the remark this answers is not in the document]
Frozemint: Just opened #8365 to explain this.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
```diff
@@ -210,13 +235,18 @@ def render(element_html, data):
...
             # Within mustache, the shortformat generates the shortinfo that is used as a placeholder inside of the numeric entry.
             # Here we opt to not generate the value, hence the placeholder is empty.
-            info_params["shortformat"] = pl.get_boolean_attrib(
-                element, "show-placeholder", SHOW_PLACEHOLDER_DEFAULT
-            )
-            shortinfo = chevron.render(f, info_params).strip()
+            # The placeholder text may be overriden by setting the 'placeholder' attribute in the pl-number-input HTML tag
+            if pl.has_attrib(element, "placeholder"):
+                # 'placeholder' attribute is set, override the placeholder text
+                html_params["shortinfo"] = pl.get_string_attrib(element, "placeholder")
```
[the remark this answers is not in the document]
Frozemint: This does look like a simple mustache variable rename, but will this change break anything?

apps/prairielearn/elements/pl-number-input/pl-number-input.mustache
```diff
@@ -13,25 +13,51 @@
...
+          id="pl-number-input-textfield-{{uuid}}" 
+          name="{{name}}"
+          type="text" 
+          inputmode="text"
+          autocomplete="off" 
+          class="form-control pl-number-input-input" 
+          size="{{size}}" 
+          {{^editable}}disabled{{/editable}} 
+          {{#raw_submitted_answer}}value="{{raw_submitted_answer}}"{{/raw_submitted_answer}} 
+          aria-describedby="pl-integer-input-{{uuid}}-label pl-number-input-{{uuid}}-suffix"
```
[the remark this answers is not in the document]
Frozemint: As tested on Mac OS 13.5.1 and Firefox 116.0.3 with VoiceOver, it looks like the description read out by the screen reader is the html placeholder text, even when `aria-describedby` references a nonexistent ID.

apps/prairielearn/elements/pl-number-input/pl-number-input.py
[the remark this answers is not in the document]
eliotwrobson: @Frozemint this change shouldn't break anything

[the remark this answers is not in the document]
eliotwrobson: Renamed here https://github.com/PrairieLearn/PrairieLearn/pull/7934/commits/a9b061c0af36281056dfb8fcde9ebccd7a2b3163

nwalters512 (approved):
@eliotwrobson I'm going to lean a lot on your own reviews here; my approval is largely a rubber stamp.

PR #7935, opened 2023-06-07
Fix correct answer not showing up in the question template/numberInput

Frozemint:
While working on `pl-number-input` and using the template question `template/numberInput` for testing, I found a small bug where the correct answer wasn't showing up or used for grading for the question. This is caused by a minor typo in the question code, and this PR fixes the typo in the question code.


eliotwrobson (approved):
LGTM!

PR #7938, opened 2023-06-08
Improved documentation of local installation on Windows

jonatanschroeder:
The documentation currently lists WSL2 as an alternative on Windows, however this provides better performance, better functionality for jobs (particularly for permissions) and better use of local file systems (see #7928).

This PR updates the documentation to list WSL2 as the main option on Windows. It also reorganizes the installation documentation for clarity, and includes a brief description of the steps to clone the course directory locally. Some minor updates in related documentation are also included.

### DIFF
--- a/docs/externalGrading.md
+++ b/docs/externalGrading.md
@@ -230,7 +230,7 @@ A reference Python implementation for this can be seen in `PrairieLearn/graders/
 
 ## Writing questions
 
-To enable students to submit files, you can use one of PrairieLearn's file elements. `<pl-file-editor>` gives students an in-browser editor that they can use to write code. `<pl-file-upload>` allows students to upload files from their own computer. For examples of both style of question, you can look at `PrairieLearn/exampleCourse/questions/fibonacciEditor` and `PrairieLearn/exampleCourse/questions/fibonacciUpload`.
+To enable students to submit files, you can use one of PrairieLearn's file elements. `<pl-file-editor>` gives students an in-browser editor that they can use to write code. `<pl-file-upload>` allows students to upload files from their own computer. For examples of both style of question, you can look at `PrairieLearn/exampleCourse/questions/fibonacciEditor` and `PrairieLearn/exampleCourse/questions/fibonacciUpload`. For questions using workspaces, [the `gradedFiles` option](https://prairielearn.readthedocs.io/en/latest/workspaces/#infojson-for-externally-graded-workspace) identifies workspace files that will be made available to the external grader.
 
 If you want to write your own submission mechanism (as a custom element, for instance), you can do that as well. We expect files to be present in a `_files` array on the `submitted_answers` dict. They should be represented as objects containing the `name` of the file and base-64 encoded `contents`. Here's an example of a well-formed `_files` array:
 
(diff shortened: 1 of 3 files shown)


docs/externalGrading.md
```diff
@@ -258,44 +258,51 @@ In production, PrairieLearn runs external grading jobs on a distributed system t
 
 PrairieLearn supports two ways of running on your own machine: natively, and inside a Docker container. We support running external autograders for each way. If the `HOST_JOBS_DIR` environment variable is set (more on that later), PrairieLearn will assume it's running in a container; otherwise, it assumes it's running natively.
 
-#### Running locally (on Docker)
+### Running locally (on Docker)
```
nwalters512, line 261: Not requesting a change yet, but: I've thought for a while that burying this information on the external grading page is not the best; perhaps this should be present on `installing.md` instead?

docs/externalGrading.md
```diff
@@ -258,44 +258,51 @@ In production, PrairieLearn runs external grading jobs on a distributed system t
...
+    -v "$HOME/pl-tam212:/course" `# Replace the path with your course directory` \
     -v "$HOME/pl_ag_jobs:/jobs" `# Map jobs directory into /jobs` \
     -e HOST_JOBS_DIR="$HOME/pl_ag_jobs" \
     -v /var/run/docker.sock:/var/run/docker.sock `# Mount docker into itself so container can spawn others` \
+    --add-host=host.docker.internal:172.17.0.1 \
     prairielearn/prairielearn
 ```
 
-On Windows PowerShell, `cd` to your course directory and copy the following command **except** with your own username in `HOST_JOBS_DIR`:
+#### Running on Windows without WSL 2 support
```
nwalters512, line 297: Is it even worth documenting this? It seems like running without WSL 2 is just a horrific experience.

docs/externalGrading.md
jonatanschroeder: That is actually a good idea. The workspace page refers to it as well anyway, might as well have both pointing to the same doc. Will change.

docs/externalGrading.md
jonatanschroeder: > Is it even worth documenting this? It seems like running without WSL 2 is just a horrific experience.

I thought about that, and tried to err on the side of keeping it just in case, but indeed for external graders and workspaces the experience can be quite deceiving. I'll remove it and indicate that we don't support the use of Windows without WSL 2 when using external graders and workspaces.

nwalters512: I think that's a good idea, but I still wonder if we should even document it for running "just" PrairieLearn. Per the linked discussion, it took almost four minutes to boot everything on a very beefy machine. I can't imagine anyone wanting that when WSL exists.

docs/externalGrading.md
jonatanschroeder: Done in 720b5e1f6.

docs/externalGrading.md
jonatanschroeder: Done in 4c320bba6.

nwalters512 (approved):

PR #7939, opened 2023-06-08
Validate that assessment question points are non-increasing

### LINKED ISSUE #773: Ensure points_list is non-increasing in syncFromDisk
Check this in https://github.com/PrairieLearn/PrairieLearn/blob/master/sync/fromDisk/assessments.js

Immediately throw an error if points_list is increasing.

nwalters512:
Closes #773. See note on that issue about the safety of this change (tl;dr it is very safe).

### DIFF
--- a/apps/prairielearn/src/sync/course-db.js
+++ b/apps/prairielearn/src/sync/course-db.js
@@ -1165,6 +1165,19 @@ async function validateAssessment(assessment, questions) {
               'Cannot specify "maxPoints" or "maxAutoPoints" for a question in an "Exam" assessment'
             );
           }
+
+          const hasSplitPoints =
+            alternative.autoPoints !== undefined ||
+            alternative.maxAutoPoints !== undefined ||
+            alternative.manualPoints !== undefined;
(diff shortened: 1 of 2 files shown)


apps/prairielearn/src/sync/course-db.js
```diff
@@ -1165,6 +1165,19 @@ async function validateAssessment(assessment, questions) {
               'Cannot specify "maxPoints" or "maxAutoPoints" for a question in an "Exam" assessment'
             );
           }
+
+          const hasSplitPoints =
+            alternative.autoPoints !== undefined ||
+            alternative.maxAutoPoints !== undefined ||
+            alternative.manualPoints !== undefined;
```
jonatanschroeder, line 1172: Nit, feel free to ignore: this expression is repeated in line 1154, do you want to define and use it there?

jonatanschroeder (approved):

apps/prairielearn/src/sync/course-db.js
nwalters512: I think I'll leave this as-is, since the error message specifically mentions checking if these values are present. It's clearer to me to check them explicitly than to rely on `hasSplitPoints` which isn't immediately obviously the same check.

PR #7941, opened 2023-06-08
Small cleanup to grader_host code

nwalters512:
- Uses `async`/`await` in more places
- Removes unused `useDatabase` config option (we always want to operate with a database, AFAIK)
- Report idle-in-transaction errors to Sentry


apps/grader-host/src/index.js
```diff
@@ -56,32 +52,38 @@ async.series(
...
+        globalLogger.error('idle client error', err);
+        Sentry.captureException(err, {
+          level: 'fatal',
+          tags: {
+            // This may have been set by `sql-db.js`. We include this in the
+            // Sentry tags to more easily debug idle client errors.
+            last_query: err?.data?.lastQuery ?? undefined,
+          },
+        });
+        Sentry.close().finally(() => process.exit(1));
```
nwalters512, line 74: We kill the process on an idle error in both PrairieLearn and the workspace host; do you think there was a valid reason we weren't doing that for the grader host too?

apps/grader-host/src/index.js
```diff
@@ -39,15 +39,11 @@ process.on('SIGTERM', () => {
 
 async.series(
```
mwest1066, line 40: At some point it would be great to start getting rid of all the `async.series()` calls and just having regular code calling `await ...` in series 😄

apps/grader-host/src/index.js
mwest1066: I can't see any reason to not do this in the grader host. In fact the grader host is the one that is the most resilient to errors because of the grading job queue so we should definitely be dying here if something is wrong (like an idle timeout).

mwest1066 (approved):

apps/grader-host/src/index.js
jonatanschroeder: Isn't the grader host called in the same process as the main PL one in dev mode?

nwalters512: For better or worse, PL reimplements most of the external grading execution process for local dev 😭

apps/grader-host/src/index.js
nwalters512: Agreed! Two things standing partially in the way of that:

- We want ESM for top-level await (thought we could just do `async.series` with a single function for error handling at the top level)
- Some of the functions here are actually still callback-based - hopefully not for long!

nwalters512:
I had to move job cancellation checking out of `receiveFromQueue`, since it relies on the database and the database isn't currently set up in PG tests.

mwest1066 (approved):

nwalters512:
Tested this on staging, it looks good!

PR #7942, opened 2023-06-08
Fix handling of empty groups

nwalters512:
This was a regression introduced in #7907. Our query ended up with `uid_list = '{NULL}'` for empty groups, which we didn't handle correctly.

### DIFF
--- a/apps/prairielearn/src/pages/instructorAssessmentGroups/instructorAssessmentGroups.js
+++ b/apps/prairielearn/src/pages/instructorAssessmentGroups/instructorAssessmentGroups.js
@@ -130,12 +130,19 @@ router.post(
       const assessment_id = res.locals.assessment.id;
       const group_name = req.body.group_name;
       if (!group_name || String(group_name).length < 1) {
-        flash('error', 'Group name cannot be empty');
+        flash('error', 'Group name cannot be empty.');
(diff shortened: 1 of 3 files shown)


apps/prairielearn/src/pages/instructorAssessmentGroups/instructorAssessmentGroups.js
```diff
@@ -165,7 +171,7 @@ router.post(
       const assessment_id = res.locals.assessment.id;
       const group_id = req.body.group_id;
       const uids = req.body.add_member_uids;
-      const uidlist = uids.split(/[ ,]+/);
+      const uidlist = uids.split(/[ ,]+/).filter((uid) => !!uid);
```
nwalters512, line 174: This wasn't strictly a bug, but it did mean that an empty string would show a weird error message like `Failed to add the following users: . Please check if the users exist.`. Now an empty `uids` string is just a no-op.

jonatanschroeder (approved):
It seems to be working, but it may be a good idea to show `(empty)` for empty groups in the page, in the list of UIDs in a group. Currently the cell is blank. The students tab shows `(empty)` for these groups.

jonatanschroeder (approved):

mwest1066 (approved):

PR #7943, opened 2023-06-08
Rewrite editors with async/await writing

nwalters512:
I'd recommend reviewing with whitespace changes hidden.

### DIFF
--- a/apps/prairielearn/src/lib/editors.js
+++ b/apps/prairielearn/src/lib/editors.js
@@ -674,37 +674,37 @@ class AssessmentCopyEditor extends Editor {
     const result = await sqldb.queryAsync(sql.select_assessments_with_course_instance, {
       course_instance_id: this.course_instance.id,
     });
-    this.oldNamesLong = _.map(result.rows, 'title');
+    const oldNamesLong = _.map(result.rows, 'title');
 
     debug('Get all existing short names');
-    this.oldNamesShort = await this.getExistingShortNames(assessmentsPath, 'infoAssessment.json');
+    const oldNamesShort = await this.getExistingShortNames(assessmentsPath, 'infoAssessment.json');
 
     debug(`Generate TID and Title`);
     let names = this.getNamesForCopy(
       this.assessment.tid,
-      this.oldNamesShort,
+      oldNamesShort,
       this.assessment.title,
-      this.oldNamesLong
+      oldNamesLong
     );
-    this.tid = names.shortName;
-    this.assessmentTitle = names.longName;
-    this.assessmentPath = path.join(assessmentsPath, this.tid);
-    this.pathsToAdd = [this.assessmentPath];
-    this.commitMessage = `${this.course_instance.short_name}: copy assessment ${this.assessment.tid} to ${this.tid}`;
+    const tid = names.shortName;
+    const assessmentTitle = names.longName;
+    const assessmentPath = path.join(assessmentsPath, tid);
+    this.pathsToAdd = [assessmentPath];
+    this.commitMessage = `${this.course_instance.short_name}: copy assessment ${this.assessment.tid} to ${tid}`;
 
     const fromPath = path.join(assessmentsPath, this.assessment.tid);
-    const toPath = this.assessmentPath;
+    const toPath = assessmentPath;
     debug(`Copy template\n from ${fromPath}\n to ${toPath}`);
     await fs.copy(fromPath, toPath, { overwrite: false, errorOnExist: true });
 
     debug(`Read infoAssessment.json`);
-    const infoJson = await fs.readJson(path.join(this.assessmentPath, 'infoAssessment.json'));
+    const infoJson = await fs.readJson(path.join(assessmentPath, 'infoAssessment.json'));
 
     debug(`Write infoAssessment.json with new title and uuid`);
-    infoJson.title = this.assessmentTitle;
+    infoJson.title = assessmentTitle;
     this.uuid = uuidv4(); // <-- store uuid so we can find the new assessment in the DB
     infoJson.uuid = this.uuid;
-    await fs.writeJson(path.join(this.assessmentPath, 'infoAssessment.json'), infoJson, {
+    await fs.writeJson(path.join(assessmentPath, 'infoAssessment.json'), infoJson, {
       spaces: 4,
     });
   }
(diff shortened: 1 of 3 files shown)


apps/prairielearn/src/lib/editors.js
```diff
@@ -190,11 +194,19 @@ class Editor {
...
             if (ERR(err, (e) => logger.error('Error in write()', e))) {
               job.fail(err);
             } else {
-              job.succeed();
+              if (this.description == null || this.description === '') {
+                job.fail(new Error('Missing required description'));
+              } else if (this.pathsToAdd == null || this.pathsToAdd.length === 0) {
+                job.fail(new Error('Missing required pathsToAdd'));
+              } else if (this.commitMessage == null || this.commitMessage === '') {
+                job.fail(new Error('Missing required commitMessage'));
```
nwalters512, line 206: This is just some additional safety checking.

apps/prairielearn/src/lib/editors.js
```diff
@@ -468,105 +480,77 @@ class Editor {
...
+    const walk = async (relativeDir) => {
+      let directories;
+      try {
+        directories = await fs.readdir(path.join(rootDirectory, relativeDir));
+      } catch (err) {
+        // If the directory doesn't exist, then we have nothing to load
+        if (err.code === 'ENOENT' || err.code === 'ENOTDIR') {
+          return;
+        }
+        throw err;
```
nwalters512, line 531: Aside: this is basically the only case where I don't like async/await. I've written a little function like this in the past to make this slightly less annoying:

```ts
async function wrapErrorfn) {
  try {
    const data = await fn;
    return [data, null];
  } catch (error) {
    return [null, error];
  }
}
```

So you can then do something like this:

```ts
const [directories, err] = await wrapError(fs.readdir());
```

This is roughly what golang does, and not that far off from something like [Rust's `Result` type](https://doc.rust-lang.org/std/result/). We can also add TS generic types so that this works generically, informs `tsc` that the first value is not null if the second is, etc. Do you completely hate this? 😅

jonatanschroeder: I was reading about Promise.allSettled the other day, which always resolves with an object that can be used in something similar to what you have in mind. Does that help? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

nwalters512: Interesting, I hadn't thought about applying that function that way. It's actually not bad:

```ts
const [res] = await Promise.allSettled([fs.readdir(path.join(rootDirectory, relativeDir))]);
if (res.status === "rejected") {
  // If the directory doesn't exist, then we have nothing to load
  if (res.reason.code === "ENOENT" || res.reason.code === "ENOTDIR") {
    return;
  }
  throw res.reason;
}
const directories = res.value;
```

Wrapping + unwrapping with arrays is a little annoying; a hypothetical `Promise.settled()` would be nice, and would probably end up looking a lot like what I proposed above.

Compared to the proposed `wrapError`-like thing:

```ts
const [directories, err] = await wrapError(fs.readdir(path.join(rootDirectory, relativeDir)));
if (err) {
  // If the directory doesn't exist, then we have nothing to load
  if (err.code === "ENOENT" || err.code === "ENOTDIR") {
    return;
  }
  throw err;
}
```

If we actually called this `settled` and made if behave more like `Promise.allSettled`, we could use it like this:

```ts
const directoriesResult = await settled(fs.readdir(path.join(rootDirectory, relativeDir)));
if (directoriesResult.status === "rejected") {
  // If the directory doesn't exist, then we have nothing to load
  if (directoriesResult.reason.code === "ENOENT" || directoriesResult.reason.code === "ENOTDIR") {
    return;
  }
  throw directoriesResult.reason;
}
const directories = directoriesResult.value;
```

jonatanschroeder: What about:

```
      const directories = await fs.readdir(path.join(rootDirectory, relativeDir)).catch((err) => {
        // If the directory doesn't exist, then we have nothing to load
        if (err.code === 'ENOENT' || err.code === 'ENOTDIR') {
          return null;
        }
        throw err;
      });

File: apps/prairielearn/src/lib/editors.js [hunk 3]
```diff
@@ -468,105 +480,77 @@ class Editor {
...
-              callback(null);
-            } else {
-              debug('No files, deleting', delPath);
-              fs.remove(delPath, (err) => {
-                if (ERR(err, callback)) return;
-                callback(null);
-              });
-            }
-          });
+    for (const [index] of reverseFolders.entries()) {
```
mwest1066, line 494: Neat idiom!

apps/prairielearn/src/lib/editors.js
```diff
@@ -677,88 +661,52 @@ class AssessmentCopyEditor extends Editor {
...
+    const oldNamesShort = await this.getExistingShortNames(assessmentsPath, 'infoAssessment.json');
+
+    debug(`Generate TID and Title`);
+    let names = this.getNamesForCopy(
+      this.assessment.tid,
+      oldNamesShort,
+      this.assessment.title,
+      oldNamesLong
     );
+    const tid = names.shortName;
```
mwest1066, line 689: At some point we should switch to calling this `aid` (assessment ID) rather than the PL v1 `tid` (test ID). But not in this PR.

apps/prairielearn/src/lib/editors.js
```diff
@@ -677,88 +661,52 @@ class AssessmentCopyEditor extends Editor {
...
+    const assessmentTitle = names.longName;
+    const assessmentPath = path.join(assessmentsPath, tid);
+    this.pathsToAdd = [assessmentPath];
+    this.commitMessage = `${this.course_instance.short_name}: copy assessment ${this.assessment.tid} to ${tid}`;
+
+    const fromPath = path.join(assessmentsPath, this.assessment.tid);
+    const toPath = assessmentPath;
```
mwest1066, line 696: I'm not tracking the details, but I assume it's ok that we lost the `this` here?

mwest1066 (approved):
So awesome to see all those hideous callbacks vanish! 🚀

apps/prairielearn/src/lib/editors.js
nwalters512: That might actually be the cleanest, I like it!

apps/prairielearn/src/lib/editors.js
nwalters512: Yep! Previously some of these values were being written to `this` so that they could be accessed in the various `async.series` callbacks. I rewrote them as plain old local variables, which should be much easier to reason about.

apps/prairielearn/src/lib/editors.js
nwalters512: Done in 8254a4e.

PR #7944, opened 2023-06-08
Fix select_submission_info query

nwalters512:
This was introduced by #7272. As far as I can tell, the broken version resulting in joining to every row of `pl_courses`, which is obviously wrong.

### DIFF
--- a/apps/prairielearn/src/lib/question.sql
+++ b/apps/prairielearn/src/lib/question.sql
@@ -232,7 +232,7 @@ FROM
   LEFT JOIN assessment_sets AS aset ON (aset.id = a.assessment_set_id)
   LEFT JOIN course_instances AS ci ON (ci.id = v.course_instance_id)
   JOIN pl_courses AS c ON (c.id = v.course_id)
-  JOIN pl_courses AS qc ON (c.id = q.course_id)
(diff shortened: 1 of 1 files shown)


SethPoulsen (approved):

apps/prairielearn/src/lib/question.sql
```diff
@@ -232,7 +232,7 @@ FROM
   LEFT JOIN assessment_sets AS aset ON (aset.id = a.assessment_set_id)
   LEFT JOIN course_instances AS ci ON (ci.id = v.course_instance_id)
   JOIN pl_courses AS c ON (c.id = v.course_id)
-  JOIN pl_courses AS qc ON (c.id = q.course_id)
+  JOIN pl_courses AS qc ON (qc.id = q.course_id)
```
mwest1066, line 235: In my review of #7272 I considered whether we should always use `vc` and `qc` when both types of course join are used together. I now wish I'd suggested this! We could make the change now, if you want, but I'm also fine to just fix the bug here.

SethPoulsen: Ya, that seems like a good idea. I thought I had done this, to be honest, but I must have missed this one.

PR #7948, opened 2023-06-09
Mark sync_warnings as nullable

nwalters512:
I considered marking the columns themselves as `NOT NULL`, but in this case I think it's actually reasonable for them to be nullable - that just indicates a lack of warnings!


mwest1066 (approved):

mwest1066:
> I considered marking the columns themselves as `NOT NULL`, but in this case I think it's actually reasonable for them to be nullable - that just indicates a lack of warnings!

Very reasonable!

PR #7949, opened 2023-06-09
Rewrite editor saving and syncing with new server jobs

nwalters512:
I didn't touch `apps/prairielearn/src/pages/instructorFileEditor/instructorFileEditor.js` yet, which uses much of the same logic. I have a suspicion that we can share the same save + sync code between these two pages; I'll try to make that happen. It would be a big win if that ends up being possible!


apps/prairielearn/src/lib/courseUtil.js
```diff
@@ -31,7 +31,7 @@ module.exports.getCommitHashAsync = promisify(getCommitHash);
  * will also add the `commit_hash` property to the given course object.
  *
  * @param {Object} course
- * @param {(err?: Error, commitHash?: string) => void} callback
+ * @param {(err?: Error, commitHash: string) => void} callback
```
nwalters512, line 34: `instructorFileEditor` is the last place where these callback-based functions are used; we should be able to remove them very soon!

apps/prairielearn/src/lib/editors.js
```diff
@@ -56,425 +92,106 @@ class Editor {
...
+              try {
+                await job.exec('git', ['push'], {
+                  cwd: this.course.path,
+                  env: gitEnv,
+                });
+              } catch (err) {
+                // In the original implementation, we would *only* reset the
+                // repo state. AFAICT, it should be safe to do both a clean
+                // and reset if the push fails?
+                // TODO: answer above question.
```
nwalters512, line 176: 👀

apps/prairielearn/src/lib/editors.js
```diff
@@ -56,425 +92,106 @@ class Editor {
...
-          on_error: config.fileEditorUseGit ? _cleanupAfterWrite : _cleanup,
-          no_job_sequence_update: true,
-        };
-        serverJobs.createJob(jobOptions, (err, job) => {
-          if (ERR(err, (e) => logger.error('Error in createJob()', e))) {
-            _finishWithFailure();
-            return;
-          }
+              if (config.fileEditorUseGit) {
+                await cleanAndResetRepository(this.course, gitEnv, job);
```
nwalters512, line 122: `instructorFileEditor.js` will do a `git fetch` before this - is it worth adding the same here?

nwalters512: On a second look, it only actually fetches if the user explicitly requests that after a failed push; grep for `pull_and_save_and_sync`. It would be great to have consistent behavior here, but I think that may be outside the scope of this PR.

apps/prairielearn/src/lib/courseUtil.js
mwest1066: Yay!

apps/prairielearn/src/lib/editors.js
mwest1066: It feels to me like we can always clean whenever we are resetting.

mwest1066 (approved):
It's amazing how much cleaner the new code is!

apps/prairielearn/src/lib/editors.js
nwalters512: Cool, I finalized that here: fbfcd93. I also used `finally` to reduce failure and success into a single case.

PR #7956, opened 2023-06-12
Add infrastructure for plans with specific features

nwalters512:
These changes will make it possible for PrairieLearn admins to grant access for specific features to specific institutions and course instances. This will allow us to track paid customers directly in PrairieLearn, and lays the groundwork for our upcoming student-billing functionality.

## TODO

- [x] Audit events for plan changes?


apps/prairielearn/src/ee/lib/billing/components/InstructorInstanceAdminBillingForm.html.ts
```diff
@@ -0,0 +1,257 @@
...
+  const computeDidChange =
+    input.initialRequiredPlans.includes('compute') !==
+    input.desiredRequiredPlans.includes('compute');
+  let computeAlert: AlertProps | null = null;
+  if (!studentBillingEnabled && (computeEnabledByInstitution || computeEnabledByCourseInstance)) {
+    computeCanChange = false;
+    computeAlert = {
+      message: computeEnabledByCourseInstance
+        ? 'This course instance already has access to compute features without additional payment.'
+        : 'Courses in your institution already have access to compute features without additional payment.',
```
nwalters512, line 74: Will folks care about this distinction in practice?

apps/prairielearn/src/ee/lib/billing/plans-types.ts
```diff
@@ -0,0 +1,46 @@
...
+  features: PlanFeatureName[];
+  initialEnrollmentLimit?: number;
+}
+
+export const PLANS = {
+  // Enabled when student-pays is enabled for a course instance.
+  basic: {
+    features: ['course-instance-access'],
+  },
+  // Add-on to basic plan that enables workspaces and external grading.
```
nwalters512, line 21: This can actually be used without the `basic` plan (for instance: institution is paying only for basic, course wants to enable compute for their students).

apps/prairielearn/src/ee/lib/billing/plans.sql
```diff
@@ -0,0 +1,140 @@
...
+  plan_name,
+  institution_id,
+  course_instance_id,
+  enrollment_id
+) DO
+UPDATE
+SET
+type = EXCLUDED.type;
+
+-- BLOCK update_plan_grants_for_course_instance
```
nwalters512, line 93: This is quite similar to `update_plan_grants_for_institution` above, but I couldn't find any nice way to reuse the same query for both cases. Suggestions welcome, but I'm also happy to just retain two separate copies.

apps/prairielearn/src/ee/pages/institutionAdminGeneral/institutionAdminGeneral.html.ts
```diff
@@ -118,3 +134,62 @@ function StatisticsCard({ title, value }: { title: string; value: HtmlValue }) {
...
+        // The basic plan is never available at the institution level; it's only
+        // used for student billing for enrollments.
+        return null;
+      }
+
+      const planFeatures = PLANS[planName].features;
+      const planGrant = planGrants.find((grant) => grant.plan_name === planName);
+      const hasPlanGrant = !!planGrant;
+      const planGrantType = planGrant?.type ?? 'trial';
+      console.log(planGrant);
```
nwalters512, line 151:
```suggestion
```

apps/prairielearn/src/ee/pages/institutionAdminGeneral/institutionAdminGeneral.ts
```diff
@@ -41,6 +49,24 @@ router.post(
...
+        const planGranted = !!req.body[`plan_${plan}`];
+        const planGrantType = req.body[`plan_${plan}_grant_type`];
+        if (planGranted) {
+          updates.push({
+            plan,
+            grantType: planGrantType,
+          });
+        }
+      }
+      console.log('plan grant updates', updates);
```
nwalters512, line 67:
```suggestion
```

apps/prairielearn/src/lib/db-types.ts
```diff
@@ -101,3 +101,19 @@ export const UserSchema = z.object({
   user_id: IdSchema,
 });
 export type User = z.infer<typeof UserSchema>;
+
+export const PlanGrantSchema = z.object({
+  course_instance_id: IdSchema.nullable(),
+  created_at: z.date(),
+  // TODO: handled `deleted_at` everywhere we need to.
```
nwalters512, line 108: 👀 - do we actually want soft deletes here?

Related: should we be recording audit logs plan grants, enrollment limits, etc?

apps/prairielearn/src/lib/db-types.ts
```diff
@@ -101,3 +101,19 @@ export const UserSchema = z.object({
...
+  deleted_at: z.date().nullable(),
+  enrollment_id: IdSchema.nullable(),
+  // TODO: remove this column.
```
nwalters512, line 111: 👀

apps/prairielearn/src/lib/features/manager.ts
```diff
@@ -177,7 +178,30 @@ export class FeatureManager<FeatureName extends string> {
...
+
+  /**
+   * Globally sets the given feature overrides and automatically cleans them up
+   * once the provided function has executed.
+   *
+   * Note that this should generally only be used in tests. If used while serving
+   * actual requests, this will have unintended behavior since the overrides will
+   * apply to all code that is running at the same time.
+   */
+  async runWithGlobalOverrides<T>(overrides: FeatureOverrides, fn: () => Promise<T>): Promise<T> {
```
nwalters512, line 198: I ideally wanted to use `runWithOverrides` here, but in tests, overrides don't make their way down into the code that's actually handling the request. Hence, this _other_ way of doing overrides that doesn't rely on ALS.

apps/prairielearn/src/tests/utils/enrollments.ts
```diff
@@ -0,0 +1,59 @@
+import { assert } from 'chai';
+import fetch from 'node-fetch';
+import { z } from 'zod';
+import { queryRow } from '@prairielearn/postgres';
+
+// Must be imported so that `config.serverPort` is set.
```
nwalters512, line 6: This is kinda gross, but 🤷 

I looked into using Mocha hooks to apply these changes to each worker process so that we wouldn't have to rely on a side effect of importing this module, but couldn't figure out something that worked.

apps/prairielearn/src/lib/db-types.ts
nwalters512: Soft deletes removed in 3d49ac2.

apps/prairielearn/src/ee/pages/institutionAdminGeneral/institutionAdminGeneral.html.ts
nwalters512: Removed in 59ace1c.

apps/prairielearn/assets/scripts/instructorInstanceAdminBillingClient.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (basicPlanEnabled) requiredPlans.push('basic');
+    if (computePlanEnabled) requiredPlans.push('compute');
+
+    morphdom(
+      billingForm,
+      InstructorInstanceAdminBillingForm({
+        ...initialProps,
+        requiredPlans,
+      }).toString()
+    );
```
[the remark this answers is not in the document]
nwalters512: An alternative to this: use HTMX to hit a server endpoint to compute the new markup?

apps/prairielearn/src/ee/lib/billing/plans.sql
nwalters512: I ultimately rewrote this to do the logic in JS with a `reconcilePlanGrants` function (name bikesheddable).

apps/prairielearn/src/ee/lib/billing/components/InstructorInstanceAdminBillingForm.html.ts
```diff
@@ -0,0 +1,247 @@
...
+  const {
+    studentBillingEnabled,
+    studentBillingCanChange,
+    studentBillingAlert,
+    computeEnabled,
+    computeCanChange,
+    computeAlert,
+  } = instructorInstanceAdminBillingState(props);
+
+  const enrollmentLimitText = studentBillingEnabled ? '∞' : enrollmentLimit;
```
mwest1066, line 117: I personally found the infinity sign to be a bit overly cute. I'd prefer to just write "XX enrollments" without the `/ ∞` in this case (and keep the help text that explains we have unlimited enrollments because student billing is enabled). I'd also prefer to not show the (empty) progress bar when we have unlimited enrollments.

apps/prairielearn/src/ee/lib/billing/plans.ts
```diff
@@ -0,0 +1,154 @@
...
+    z.enum(PLAN_NAMES),
+  );
+}
+
+export async function updateRequiredPlansForCourseInstance(
+  course_instance_id: string,
+  plans: PlanName[],
+) {
+  await runInTransactionAsync(async () => {
+    // TODO: locking?
```
mwest1066, line 66: I suppose so, although it doesn't seem very likely to be important in practice. We could always take a row lock on the CI.

mwest1066 (approved):

apps/prairielearn/src/ee/lib/billing/plans.sql
mwest1066: I much prefer having this kind of logic in JS in any case. I definitely want to move towards getting this kind of thing out of SQL when possible.

apps/prairielearn/assets/scripts/instructorInstanceAdminBillingClient.ts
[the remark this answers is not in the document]
mwest1066: I think this is quite nice as is. I do worry a bit that we are essentially writing our own client-side framework, but maybe things like `morphdom` and `selector-observer` are actually the way to _avoid_ client-side frameworks? I don't know enough to have a good opinion about whether it would be cleaner to use React or Svelte or whatever. I _do_ like the fact that `morphdom` just does one simple thing and it's easy to understand with no magic.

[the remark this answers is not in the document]
nwalters512: Yeah, I definitely would like to avoid our own bespoke client-side "framework" here. If we start seeing this pattern appear in more places, I'll take a second look at this.

The problem with React and friends is that they don't necessarily plug in nicely with server-rendered HTML that wasn't also generated with React/etc. The "right" way to use them is with a meta-framework that takes care of hydration on the client, shuttling data from the backend to the frontend, etc. We could always build our own little helpers to do this, including rendering on the backend and hydrating on the frontend?

Regardless, let's kick this can down the road until we actually have working student-pays 😂

apps/prairielearn/src/ee/lib/billing/components/InstructorInstanceAdminBillingForm.html.ts
nwalters512: This is handling the case where we add a plan grant to the course instance itself, not to the parent institution. Recall that a course instance _plan grant_ is different from a course instance _required plan_ - the former makes features available, the latter requires students to have access to the feature (possibly through student-pays) to access the course instance.

If this is confusing to you, it's almost definitely confusing to end users. Maybe the only message we need is "This course instance already has access to compute features without additional payment"?

apps/prairielearn/src/lib/audit-log.ts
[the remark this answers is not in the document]
nwalters512: I can certainly start now if you'd like! I already did that in `ee` as you pointed out in your comment below.

apps/prairielearn/src/pages/partials/navTabs/instructorInstanceAdmin.ejs
```diff
@@ -23,6 +23,13 @@
     urlSuffix:      '/instance_admin/lti',
     iconClasses:    'fas fa-graduation-cap',
     tabLabel:       'LTI'}) %>
+<% if (billing_enabled) { %>
+    <%- include('navTab', {
+        activeSubPage:  'billing',
+        urlSuffix:      '/instance_admin/billing',
+        iconClasses:    'fas fa-dollar-sign',
```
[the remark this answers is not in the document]
nwalters512: A credit card does seem better 🙂 I'll make the change to use this icon: https://fontawesome.com/icons/credit-card?f=classic&s=solid

apps/prairielearn/assets/scripts/instructorInstanceAdminBillingClient.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    const requiredPlans: PlanName[] = [];
+    if (basicPlanEnabled) requiredPlans.push('basic');
+    if (computePlanEnabled) requiredPlans.push('compute');
+
+    morphdom(
+      billingForm,
+      InstructorInstanceAdminBillingForm({
+        ...initialProps,
+        requiredPlans,
```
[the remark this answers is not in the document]
nwalters512: Nice, I caught this independently this morning while testing! The problem is that the splatted `initialProps` had type `any`, which effectively shut off typechecking (did someone say [`noImplicitAny`](https://www.typescriptlang.org/tsconfig#noImplicitAny)?). I added a type parameter to `decodeData` in this file and TypeScript now complains as it should.

apps/prairielearn/src/ee/lib/billing/components/InstructorInstanceAdminBillingForm.html.ts
nwalters512: Hah, sure, I can remove the infinity sign. I'm not sure what to do about the progress bar though - showing/hiding it would cause layout shift that would move the checkbox out from under the cursor, which is what I was trying to avoid by having it become empty.

apps/prairielearn/src/ee/lib/billing/plans.ts
nwalters512: Yeah, that's probably true. I think we can go without locking here.

apps/prairielearn/src/pages/partials/navTabs/instructorInstanceAdmin.ejs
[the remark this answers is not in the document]
nwalters512: Done in b29b265.

apps/prairielearn/assets/scripts/instructorInstanceAdminBillingClient.ts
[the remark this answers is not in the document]
nwalters512: Done in 6fbead7 and e0446d4.

apps/prairielearn/src/ee/models/plan-grants.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export async function insertPlanGrant(planGrant: NewPlanGrant): Promise<void> {
+  const newPlanGrant = await queryRow(
+    sql.insert_plan_grant,
+    {
+      institution_id: null,
+      course_instance_id: null,
+      enrollment_id: null,
+      user_id: null,
+      ...planGrant,
```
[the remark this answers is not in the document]
nwalters512: I had a draft of that here: https://github.com/PrairieLearn/PrairieLearn/pull/5610. Might be nice to revisit that sometime!

Yeah, I can switch to passing them explicitly here.

apps/prairielearn/src/ee/models/plan-grants.ts
```diff
@@ -0,0 +1,98 @@
...
+    },
+    PlanGrantSchema,
+  );
+  // TODO: this doesn't yet associate plan grants with enrollments.
```
nwalters512, line 38: Once #8204 lands, I can address this and similar comments.

apps/prairielearn/src/ee/models/plan-grants.ts
```diff
@@ -0,0 +1,98 @@
...
+  await insertAuditLog({
+    // TODO: pipe this down?
```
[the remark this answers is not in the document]
nwalters512: I made this change in 04792ad. Sometimes I wish we had this user in ALS, but I suppose it's nice to be explicit, especially for testing.

apps/prairielearn/src/ee/models/plan-grants.ts
```diff
@@ -0,0 +1,98 @@
...
+    authn_user_id: null,
+    table_name: 'plan_grants',
+    action: 'insert',
+    institution_id: newPlanGrant.institution_id,
+    course_instance_id: newPlanGrant.course_instance_id,
+    // TODO: is this the correct usage of `user_id` here? Does this column
+    // represent the user takin an action, or the user being affected by the
+    // action?
```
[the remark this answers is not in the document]
nwalters512: Cool, ok - we should of course pick better names once we rework how audit events work 😂

apps/prairielearn/src/lib/audit-log.ts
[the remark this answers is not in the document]
nwalters512: I moved to a `models/` directory in af37d31.

apps/prairielearn/assets/scripts/institutionAdminGeneralClient.ts
```diff
@@ -0,0 +1,11 @@
+import { onDocumentReady } from '@prairielearn/browser-utils';
+
+onDocumentReady(() => {
+  document.querySelectorAll('.js-plan').forEach((plan) => {
+    const enabledCheckbox = plan.querySelector<HTMLInputElement>('.js-plan-enabled');
+    const enabledType = plan.querySelector<HTMLSelectElement>('.js-plan-type');
+    enabledCheckbox.addEventListener('change', () => {
+      enabledType.disabled = !enabledCheckbox.checked;
```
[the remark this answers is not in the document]
nwalters512: I'm going to pass on this for now in the interest of shipping this, if that's ok with you? Let's revisit in the future!

apps/prairielearn/src/ee/pages/instructorInstanceAdminBilling/instructorInstanceAdminBilling.ts
```diff
@@ -0,0 +1,139 @@
...
+  asyncHandler(async (req, res) => {
+    // This page is behind a feature flag for now.
+    if (!res.locals.billing_enabled) {
+      throw error.make(404, 'Not Found');
+    }
+
+    // Only course owners can manage billing.
+    if (!res.locals.authz_data.has_course_permission_own) {
+      throw error.make(403, 'Access denied (must be course owner)');
+    }
```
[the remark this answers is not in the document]
nwalters512: ah sure, especially since we're showing it on the navbar. I'll make that change!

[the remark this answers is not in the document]
nwalters512: How does this look?

Screenshot 2023-07-20 at 11 05 34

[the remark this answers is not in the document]
nwalters512: These changes were made in 635a65d and ded257d.

apps/prairielearn/src/ee/models/plan-grants.ts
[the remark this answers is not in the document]
nwalters512: I got, uh, clever with types in 70e3487 and made this work 😅 let me know if this is too much!

apps/prairielearn/src/ee/models/course-instance-required-plans.ts
```diff
@@ -0,0 +1,55 @@
...
+    sql.insert_required_plan_for_course_instance,
+    { course_instance_id, plan_name: plan },
+    CourseInstanceRequiredPlanSchema,
+  );
+  await insertAuditLog({
+    authn_user_id,
+    table_name: 'course_instance_required_plans',
+    action: 'insert',
+    institution_id: null, // TODO: get value for this
+    course_id: null, // TODO: get value for this
```
nwalters512, line 23: Would be nice to have `insertAuditLog` handle this for us, like it does in PT. @mwest1066 do you have (or could you make) a non-sproc version of that in mind that we could use in PL?

apps/prairielearn/assets/scripts/instructorInstanceAdminBillingClient.ts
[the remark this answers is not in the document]
mwest1066: Totally agree on kicking this down the road!

apps/prairielearn/src/ee/lib/billing/components/InstructorInstanceAdminBillingForm.html.ts
mwest1066: Right, thanks for the reminder! And yes, I think we should focus on the important information for the instructor as you say.

apps/prairielearn/assets/scripts/institutionAdminGeneralClient.ts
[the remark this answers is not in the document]
mwest1066: Totally fine with me, just wanted to point it out.

apps/prairielearn/src/lib/audit-log.ts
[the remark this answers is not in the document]
mwest1066: Nice, thanks!

apps/prairielearn/src/ee/lib/billing/plans.ts
nwalters512: Comments removed in 8a94eff.

apps/prairielearn/src/ee/models/plan-grants.ts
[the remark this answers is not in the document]
mwest1066: Hahaha, that's a pretty deep level of cleverness 😄. I don't follow all the details of how `WithRequiredKeys` is implemented but the usage of it looks clear and natural. I definitely like being able to avoid the splat.

apps/prairielearn/src/ee/models/plan-grants.ts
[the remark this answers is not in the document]
mwest1066: Double-plus in favor of better names in the future!

apps/prairielearn/src/ee/models/plan-grants.ts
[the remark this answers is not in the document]
mwest1066: Ah cool, sounds good.

apps/prairielearn/assets/scripts/instructorInstanceAdminBillingClient.ts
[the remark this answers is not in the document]
mwest1066: Excellent!

apps/prairielearn/src/ee/lib/billing/components/InstructorInstanceAdminBillingForm.html.ts
mwest1066: I agree having layout jump around is annoying. Two ideas:
1. Use `visibility: hidden` to hide the progress bar but leave the space intact.
2. Shift the progress bar to be displayed on the same line as the count, so we'd have:
    ```
    12 / 20 enrollments [xxxxxxxx---------]
    ```
    This would require using a fixed width progress bar, which I think would look better anyway (the current progress bar is so long that I initially thought it was a thick horizontal line rather than a progress bar.

Without having tried them I think I like option (2).

apps/prairielearn/src/ee/models/course-instance-required-plans.ts
mwest1066: I haven't yet made a non-sproc version of this but I plan to do so soon.

nwalters512: I took the "do this the stupid way so we can ship something" approach in b4d0aa9.

apps/prairielearn/src/ee/models/plan-grants.ts
nwalters512: I added `enrollment_id` to the audit logs in 2d8f239.

apps/prairielearn/src/ee/lib/billing/components/InstructorInstanceAdminBillingForm.html.ts
nwalters512: I simplified this to a single message in 828cfa7.

apps/prairielearn/src/ee/models/course-instance-required-plans.ts
mwest1066: lgtm

apps/prairielearn/src/ee/lib/billing/components/InstructorInstanceAdminBillingForm.html.ts
nwalters512: I switched to a width-capped, inline progress bar in 277f732. I fudged the numbers a bit and enforced a minimum width of 2% so that it wouldn't disappear completely for low enrollment counts with a high enrollment limit.

Screenshot 2023-07-21 at 13 02 34

PR #7957, opened 2023-06-12
Use supported Pandas testing utilities

nwalters512:
This broke when we updated Pandas to v2 in #7813; this change was missed when reviewing the changelog.

Related: https://github.com/pandas-dev/pandas/pull/30745


eliotwrobson (approved):
Seems good! Another case where type annotations on the autograder would be nice.

